Issue #533, opened 2014-12-16
Generic property injection on the purpose of plugins


Horusiath:
Since C# doesn't offer multiple inheritance (like Scala via traits) sometimes there is a need to bypass that problem by using composition over inheritance. Example is stashing with `StashFactory.CreateStash`: we use framework to automatically inject stash into actors marked with `IActorStash` interface.

This feature may be however used more often (we already could use it for example in logging and cluster actor initialization), and also be utilized by plugins. I've already found two use cases in one of the possible designs of Akka.Persistence plugin.

Should we provide some kind of a generic actor creation interceptor mechanism or should we dump the responsibility on shoulders of Akka users?

HCanber:
I like it, however I have two things I'd like to discuss.

__1__
_Created_ and  _Terminated_ are well defined concepts for an actor. An actor is created and terminated once, and only once, but the methods `AfterActorCreated` and `BeforeActorTerminated` are called several times during an actor's lifecycle. They are called every time the _underlying_ actor instance (i.e. the class that the user writes) is created/terminated.

Unfortunately there's no term for the underlying instance. In our, as well as Akka JVM's code, the only time (as I know of) a user is exposed to the underlying instance is in `TestActorRef` which has a property called `UnderlyingActor`. So maybe we should use that term?

I.e. call the methods `AfterUnderlyingActorCreated` and `BeforeUnderlyingActorTerminated`.

__2__
Do we really need to check every actor instance? 
``` C#
public interface IActorProducerPlugin
{
  bool CanBeAppliedTo(ActorBase actor);
  ...
}
```

Or could we change this to:
``` C#
public interface IActorProducerPlugin
{
  bool CanBeAppliedTo(Type actorType);
  ...
}
```

This would allow us to check the actor type once, build a pipeline with only the plugins applicable to the actor type and then store the pipeline (per type). Whenever `AfterUnderlyingActorCreated` and `BeforeUnderlyingActorTerminated` is to be called, we get the pipeline for the actor (type) and call the method, i.e. no need to call `CanBeAppliedTo` every time.

Horusiath:
[...]

PR #539, opened 2015-01-01
Plugable actor creation pipeline + IDisposable action handling

Horusiath:
cc #533

I've created `ActorProducerPipeline`, which accessible from `ExtendedActorSystem` (to be able to configure by plugins) and allows to inject custom hooks satisfying following interface:
```csharp
interface IActorProducerPlugin {
    bool CanBeAppliedTo(ActorBase actor);
    void AfterActorCreated(ActorBase actor, IActorContext context);
    void BeforeActorTerminated(ActorBase actor, IActorContext context);
}
```

- **CanBeAppliedTo** determines if plugin can be applied to specific actor instance.
- **AfterActorCreated** is applied to actor after it has been instantiated by an `ActorCell` and before `InitializableActor.Init` method will (optionally) be invoked.
- **BeforeActorTerminated** is applied before actor terminates and before `IDisposable.Dispose` method will be invoked (for disposable actors) - **auto handling disposable actors is second feature of this commit**.

For common use it's better to create custom classes inheriting from `ActorProducerPluginBase` and `ActorProducerPluginBase<TActor>` classes.

Pipeline itself provides following interface:

```csharp
class ActorProducerPipeline : IEnumerable<IActorProducerPlugin> {
    int Count { get; } // current plugins count - 1 by default (ActorStashPlugin)
    bool Register(IActorProducerPlugin plugin)
    bool Unregister(IActorProducerPlugin plugin)
    bool IsRegistered(IActorProducerPlugin plugin)
    bool Insert(int index, IActorProducerPlugin plugin)
}
```

- **Register** - registers a plugin if no other plugin of the same type has been registered already (plugins with generic types are counted separately). Returns true if plugin has been registered.
- **Insert** - same as register, but plugin will be placed in specific place inside the pipeline - useful if any plugins precedence is required.
- **Unregister** - unregisters specified plugin if it has been found. Returns true if plugin was found and unregistered.
- **IsRegistered** - checks if plugin has been already registered.

By default pipeline is filled with one already used plugin - `ActorStashPlugin`, which replaces stash initialization/unstashing mechanism used up to this moment.

### DIFF
--- a/src/core/Akka/Actor/ActorCell.cs
+++ b/src/core/Akka/Actor/ActorCell.cs
@@ -263,9 +263,13 @@ namespace Akka.Actor
                     {
                         disposable.Dispose();
                     }
-                    catch (Exception)
+                    catch (Exception e)
                     {
-                        //TODO: what if user defined Dispose method will fail?
+                        if (_systemImpl.Log != null)
+                        {
+                            _systemImpl.Log.Error(e, "An error occurred while disposing {0} actor. Reason: {1}", 
+                                actor.GetType(), e.Message);
+                        }
                     }
                 }
 
--- a/src/core/Akka/Actor/ActorCell.cs
+++ b/src/core/Akka/Actor/ActorCell.cs
@@ -183,14 +183,9 @@ namespace Akka.Actor
         {
             var actor = _props.NewActor();
 
-            //Check if the actor needs a stash
-            var actorStash = actor as IActorStash;
-            if(actorStash != null && actorStash.Stash==null)
-            {
-                //Only assign a new stash if one hasn't been created by the actor already
-                actorStash.Stash = StashFactory.CreateStash(this, actorStash);
-            }
-
+            // Apply default of custom behaviors to actor.
+            _systemImpl.ActorProducerPipeline.AfterActorCreated(actor, this);
+            
             var initializableActor = actor as InitializableActor;
             if(initializableActor != null)
             {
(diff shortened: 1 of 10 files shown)


src/core/Akka/Actor/ActorCell.cs
```diff
@@ -261,6 +256,19 @@ protected void ClearActor(ActorBase actor)
...
+                var disposable = actor as IDisposable;
+                if (disposable != null)
+                {
+                    try
+                    {
+                        disposable.Dispose();
+                    }
+                    catch (Exception)
+                    {
+                        //TODO: what if user defined Dispose method will fail?
```
rogeralsing: That shoud probably send an error message to the system log.

src/core/Akka/Actor/ActorProducerPipeline.cs
```diff
@@ -0,0 +1,251 @@
...
+
+        /// <summary>
+        /// Plugin behavior applied to <paramref name="actor"/> instance when the new one is being created.
+        /// </summary>
+        public virtual void AfterActorCreated(TActor actor, IActorContext context) { }
+
+        /// <summary>
+        /// Plugin behavior applied to <paramref name="actor"/> instance before the actor is being recycled.
+        /// </summary>
+        public virtual void BeforeActorDestroyed(TActor actor, IActorContext context) { }
```
rogeralsing: I don't think "Destroy" is used anywhere else in Akka.NET afaik, is this name clear enough for users to understand that it applies to when the instance is about to be killed?
Not saying that it's a bad name, just so we are consistent and clear with naming.

rogeralsing:
:+1: looks great.

Mabe the docs needs to be a bit extended, providing a bit more context to the summaries, e.g. how the flow of the creation process works, how ordering of plugins can affect the outcome etc, we are lacking in the rest of the framework so I think we should step that area up for every new feature/fix.

src/core/Akka/Actor/ActorProducerPipeline.cs
Horusiath: Yes, I've started with *Destroy* suffix, but then realized that actually naming convention calls it *Terminated*, so I've adjusted methods names to it. But it seems that I've missed one place ;)

Horusiath:
Where can I extend docs part considering this commit?

PR #541, opened 2015-01-12
API changes: pipeline caching based on actor type + renaming plugins methods

Horusiath:
As discussed in #533.

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorProducerPipelineTests.cs
+++ b/src/core/Akka.Tests/Actor/ActorProducerPipelineTests.cs
@@ -38,7 +38,7 @@ namespace Akka.Tests.Actor
 
         internal class GenericPlugin<T> : ActorProducerPluginBase<T> where T : PlugActor
         {
-            public override void AfterActorCreated(T actor, IActorContext context)
+            public override void AfterIncarnated(T actor, IActorContext context)
             {
                 actor.PluginMessages.Add(typeof(T).ToString());
             }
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
@Horusiath AppVeyor couldn't compile this PR: -------------------------------------------------------------------- 
  1) Building C:\projects\akka-net-335\src\Akka.sln failed with exitcode 1. 
  2) MSB5023: C:\projects\akka-net-335\src\Akka.sln(549,0): Error parsing the nested project section in solution file. A project with the GUID "{68D89B2A-6770-483E-B7B0-C7EDAE0C2A6D}" is listed as being nested under project "{E814B432-DBC2-4372-92C6-70B9198613BD}", but does not exist in the solution.

Could you investigate?

Aaronontheweb:
Looks like this was an issue with the previous commit too; my fault for not fixing the broken unit tests on AppVeyor for so long (we've just started ignoring them.)

Aaronontheweb:
I was able to fix this locally by just removing FSharp.Deploy.Local from the Solution, recompiling it, adding it back, and then running MSBuild from the command line. Seems like the issue is a funky "works on my machine but not with MSBUIlD" sort of issue: http://stackoverflow.com/questions/13014057/msbuild-solution-file-error-msb5010-no-file-format-header-found

Aaronontheweb:
Err, wrong link: http://stackoverflow.com/questions/6599966/team-foundation-server-2010-msbuild-error-parsing-the-nested-project-section

Aaronontheweb:
Going to pull this in and fix it on a subsequent PR.

Aaronontheweb:
Fix the solution file business so AppVeyor can MSBuild it, I mean :p

PR #549, opened 2015-01-19
fixed visual studio solution file

Aaronontheweb:
This PR adds nothing to the Akka.NET codebase itself - just fixes a mangling issue inside the `Akka.sln` file itself. Akka.NET currently compiles without issue in Visual Studio, but the mangled `.sln` file causes the MSBuild command line compile to fail.


Aaronontheweb:
Since this change doesn't affect any underlying code, I'm going to pull it in so we have a clean solution for the next PR.

Issue #553, opened 2015-01-22
ActorPath is not serializable using standard Newtonsoft JSON serializer


Horusiath:
While trying to use `ActorRef.Tell` on an object having an `ActorPath` as a reference I've got following exception:

```
'Newtonsoft.Json.JsonConvert.SerializeObject(d)' threw an exception of type 'Newtonsoft.Json.JsonSerializationException'
    base: {"Self referencing loop detected for property 'Root' with type 'Akka.Actor.RootActorPath'. Path 'Parent.Parent'."}
```

I think this deserves to be described as a bug. ActorPath should probably have it's own custom serialization rule (maybe serialization to stringified path and parsed on deserialization?).

rogeralsing:
When do you want to serialize an actor path instead of the actual ActorRef?

Horusiath:
While porting Akka.Persistence I've found, that canonical Akka implementation heavily utilizes `ActorPath`s to **at least once** delivery confirmations. I'm not quite sure yet if it can be replaced by `ActorRef` or even `ICanTell` instances, but either way it think it should be a serializable value.

Aaronontheweb:
@rogeralsing @Horusiath that makes sense - in order to resurrect a downed Actor inside `Akka.ClusterSharding`, which depends on `Akka.Persistence` internally you're going to care more about the `ActorPath` than the reference to a specific `ActorRef` instance itself. 

This is similar in concept to Orleans' virtual actors - only one instance of the Actor is actually doing work in the cluster at any given time, but it has hot replicas standing by elsewhere in the cluster in the event that the main actor dies. It uses Akka.Persistence under the hood to make sure that the internal state from the main actor gets shared with the replicas in a read-only fashion, so they can take over immediately in the event of main actor termination.

So we should make `ActorPath` serializable - I think stringifying / unstringifying it is pretty straightforward.

rogeralsing:
See #557

Aaronontheweb:
@rogeralsing does #581 solve this?

rogeralsing:
@Horusiath has already fixed this

PR #556, opened 2015-01-24
Serializable actor path

Horusiath:
This one will allow to properly serialize/deserialize `ActorPath` instances. It uses standard serialization to actor path's format string with address and deserializes it using TryParse method.

cc #553

### DIFF
--- a/src/core/Akka/Serialization/ActorPathSerializer.cs
+++ b/src/core/Akka/Serialization/ActorPathSerializer.cs
@@ -1,0 +1,51 @@
+﻿using System;
+using System.Text;
+using Akka.Actor;
+
+namespace Akka.Serialization
+{
+    public class ActorPathSerializer : Serializer
+    {
+        public ActorPathSerializer(ExtendedActorSystem system) : base(system)
+        {
+        }
+
+        public override int Identifier
+        {
+            get { return -4;}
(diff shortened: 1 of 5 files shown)


src/core/Akka/Serialization/ActorPathSerializer.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+using System.Text;
+using Akka.Actor;
+
+namespace Akka.Serialization
+{
+    public class ActorPathSerializer : Serializer
```
rogeralsing: I don't think this is the best way to serialize an `ActorPath`.
For this to work, the `ActorPath` needs to be the root object, the message itself.
If the `ActorPath` is part of a larger message structure, this will fail since the serializer for that specific message type will kick in.

I think we should use the default serializer for this, this can be acheived by using substitution types in Json.NET.
This way, we can send both the ActorPath as the root message or as part of a bigger message..
See https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/ActorRef.cs#L172
That would probably also require less code in the end, just those operator overloads.

Aaronontheweb:
What's the status of this PR? We scrapping this approach in favor of something else?

Issue #559, opened 2015-01-28
Add C# hello world sample project. 


avoxm:
Having hello world sample project from Akka.NET Wiki under examples folder will make debugging for new developers easier.

Aaronontheweb:
Closed per #561

PR #561, opened 2015-01-28
Adding C# hello world example #559

avoxm:
Hello world example project from Akka.NET wiki.


Aaronontheweb:
I think it's a good idea to include a Hello World sample - your thoughts @rogeralsing ?

rogeralsing:
The other examples have aimed to show specific features. Hello world would show none.

But I guess that it could be useful just to get going. 
So +1 from me

avoxm:
@rogeralsing I guess you are right in regards of presenting features. Frankly, my intention was not presenting features, but to have the most basic sample just to make debugging through sources easier for new developers who want to join.

Aaronontheweb:
Having a reference sample for the most basic stuff is useful - gives newbies a place to start looking.

I'll deal with the merge conflict a little later today and pull this in.

Issue #563, opened 2015-01-28
Document how MultiNode Test Runner works


Aaronontheweb:
Pre-requisite for #562

Should include:

* a readme in the [MultiNode Test Runner Folder](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.MultiNodeTestRunner)
* an entry for the Wiki, and
* additional XML-DOC comments.

Assigning this one to myself for the time being.

Aaronontheweb:
Marking this one as finished for now - I'm going to be changing the behavior of the TestRunner based on some of the other issues I've created and will update the README accordingly and will publish the finished product into the Wiki.

Aaronontheweb:
Also - going to continually add more XML-DOC comments as I go.

PR #574, opened 2015-01-29
Check in on MultiNodeTestRunner docs for #563

Aaronontheweb:
Don't have everything done for #563 yet, but I wanted to check in what I've done thus far so I can start linking  the images I need to use in the Wiki / README docs


Aaronontheweb:
Docs only in this pull request. No code changes, unless you count XML-DOC comments :p

Issue #570, opened 2015-01-29
Import scala's `require` function to Akka.Util


Aaronontheweb:
In many places within the original Akka codebase, the development team enforces code contracts using a simple scala function - `require`:

```scala
private[testkit] lazy val myself: RoleName = {
    require(_roles.size > MultiNodeSpec.selfIndex, "not enough roles declared for this test")
    _roles(MultiNodeSpec.selfIndex)
  }
```

See https://github.com/akka/akka/blob/master/akka-multi-node-testkit/src/main/scala/akka/remote/testkit/MultiNodeSpec.scala#L96

`require` seems pretty similar to `Debug.Assert` in that it throws an exception when a boolean expression evaluates to false. But I'm sure there are some nuances to it.

We've been hacking around the lack of this call inside tons of constructors in various parts within Akka.NET - I think it's time we added this to the codebase for parity purposes.

Please add this class to `Akka.Util` inside the code Akka library.

rogeralsing:
This leads to another q, should we use code contracts?

Skickat från min iPhone

Aaronontheweb:
I think using the specific "code contracts" .NET library that recently got OSSed is overkill, but I think we should use the concept itself liberally.

alex-bogomaz:
There is one place in the codebase (DeadlineFailureDetector.cs) where explicit "throw new Ex.." is replaced with the function like ```require```:  
```csharp 
Helios.Util.Guard.True(bool boolean, string errorMessage)
```

```csharp
public DeadlineFailureDetector(TimeSpan acceptableHeartbeatPause, 
        {
            //.....
            Guard.True(_acceptableHeartbeatPause > TimeSpan.Zero, 
                       "acceptable-heartbeat-pause must be greater than zero");
        }
```
So we may say that we already have such function.
On the other hand scala's implementation of require is straightforward and matches the implementation of Guard.True (see https://github.com/scala/scala/blob/2.11.x/src/library/scala/Predef.scala#L205)
```scala
def require(requirement: Boolean, message: => Any) {
    if (!requirement)
      throw new IllegalArgumentException("requirement failed: "+ message)
  }
```

Aaronontheweb:
great catch, @alex-bogomaz !

So in order to introduce this capability to Akka Core, we'd have to add a dependency on Helios - which I'd like to avoid in order to keep things simple.

IMHO, I think we should just clone Helios` `Guard` function into Akka.NET and add that to Akka.Util.

[...]

PR #578, opened 2015-01-30
Class for expressing coding assumptions #570

avoxm:
Adding utility class for expressing coding assumptions. This class will
be extended to support more conditions as needed.


Aaronontheweb:
@avoxm I think you might need to merge in the latest changes from `dev` before this can automatically merge, given that @Horusiath's latest PR added some new projects to the .SLN file

avoxm:
Hmmm.... running sync shows that everything is up to date.

avoxm:
yes you were right @Aaronontheweb, thanks. Now it looks much better

Aaronontheweb:
@avoxm ty sir!

Issue #573, opened 2015-01-29
Need Serialization tests for all actor primitives


Aaronontheweb:
Per #569, need to add some tests to `Akka.Tests.Serialization` that explicitly test the serializability of all  of the following primitives:

1. Props
1. Deploy
1. RouterConfig
1. Address
1. ActorPath
1. Config (?)

Should we break this up into one issue for each?

rogeralsing:
I'm taking this one

Aaronontheweb:
Resolved as of #581

AlbertoMonteiro:
@Aaronontheweb @rogeralsing the meaning of Serializability is the same that those links explain

- https://msdn.microsoft.com/pt-br/library/ms710099(v=vs.85).aspx
- https://en.wikipedia.org/wiki/Serializability

Aaronontheweb:
We're referring to Serialization, not Serializability - being able to guarantee that a primitive Akka.NET type can be represented as a byte array

Issue #569, opened 2015-01-29
Props are not serialized correctly using default Netwonsoft serializer


Horusiath:
Netwonsoft JSON.NET is not able to correctly serialize/deserialize `Props` objects. Test code presenting problem:

```
class Creator: UntypedActor
{
    protected override void OnReceive(object message)
    {
        var props = message as Props;
        var aref = Context.ActorOf(props);
        Sender.Tell(aref);
    }
}

var creator = ActorOf<Creator>();
creator.Tell(Props.Create(() => new BlackHoleActor()));
```

In following example props message received inside Creator (message serialization is turned on) is incorrectly initialized - ActorType is set to `ActorBase` and producer to `DefaultProducer`.

Aaronontheweb:
IMHO, related to #547 

Need strong equality by value semantics for the entirety of the `Props` class so this stuff *fails correctly* during its serialization unit tests.

rogeralsing:
Im taking this one

rogeralsing:
This is going to be a beast, props drags along all of the other primitives like router config, config, supervisorstrategy etc. all of them need correct serialization.

Aaronontheweb:
Resolved as of #581 

Will be creating a new issue to deal with `Props` serialization verification.

Issue #547, opened 2015-01-19
Need to make all Props primitives IEquatable<T>


Aaronontheweb:
One thing that would make it a lot easier to write unit tests for Akka.NET is making it easy to compare all of the following `Props` primitives by value:

* `RouterConfig`
* `Scope`
* `Deploy`
* `Config`

Given that these types are meant to be immutable, there should be an easy way to test that two `Deploy` or `Config` instances contain the same content.

I've started work on this for `Scope` and `RouterConfig`.

Aaronontheweb:
Resolved as of #581

PR #581, opened 2015-02-01
Improving serialization support

rogeralsing:
This is for discussion and review and yakshaving for #547 and #573 and #569 

Added JsonSerializer support for private constructors and private setters
Added Equality members to Akka primitives
Added Serialization tests for Akka primitives
Added Serialization support for `Config`, seems to work, but need extensive testing,
Added Serialization support for `Props` , seems to work, but need extensive testing,

The currently largest issue is the implementation of `SupervisorStrategy`, we use a delegate to deal with exception evaluation, this does not serialize.
The only option I can think of is to replace this with a KVP -> exception type + directive and serialize that instead.

All tests pass, but we need more eyes and extensive testing,

cc @akkadotnet/developers.

### DIFF
--- a/src/core/Akka/Actor/Props.cs
+++ b/src/core/Akka/Actor/Props.cs
@@ -7,6 +7,7 @@ using Akka.Util.Internal;
 using Akka.Util.Reflection;
 using Akka.Routing;
 using Akka.Util;
+using Newtonsoft.Json;
 
 namespace Akka.Actor
 {
@@ -22,8 +23,72 @@ namespace Akka.Actor
     ///   private Props otherProps = props.WithDeploy(deployment info);
     ///  </code>
     /// </summary>
-    public class Props
+    public class Props : IEquatable<Props>
     {
+        public bool Equals(Props other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+            return CompareDeploy(other) && CompareSupervisorStrategy(other) && CompareArguments(other) && CompareInputType(other);
+        }
+
+        private bool CompareInputType(Props other)
+        {
+            return inputType == other.inputType;
+        }
+
+        private bool CompareDeploy(Props other)
+        {
+            return Deploy.Equals(other.Deploy);
+        }
+
+        private bool CompareSupervisorStrategy(Props other)
+        {
+            return true;
+            return Equals(SupervisorStrategy, other.SupervisorStrategy);
+        }
+
+        private bool CompareArguments(Props other)
+        {
+            if (other == null)
+                return false;
+
+            if (Arguments == null && other.Arguments == null)
+                return true;
+
+            if (Arguments == null)
+                return false;
+
+            if (Arguments.Length != other.Arguments.Length)
+                return false;
+
+            //TODO: since arguments can be serialized, we can not compare by ref
+            //arguments may also not impement equality opertators, so we can not structurally compare either
+            //we can not just call a serializer and compare outputs either, since different args may require diff serializer mechanics
+
+            return true;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj.GetType() != this.GetType()) return false;
+            return Equals((Props) obj);
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                int hashCode = (Deploy != null ? Deploy.GetHashCode() : 0);
+              //  hashCode = (hashCode*397) ^ (SupervisorStrategy != null ? SupervisorStrategy.GetHashCode() : 0);
+              //  hashCode = (hashCode*397) ^ (Arguments != null ? Arguments.GetHashCode() : 0);
+                hashCode = (hashCode*397) ^ (inputType != null ? inputType.GetHashCode() : 0);
+                return hashCode;
+            }
+        }
+
         /// <summary>
         ///     The default deploy
         /// </summary>
@@ -32,7 +97,7 @@ namespace Akka.Actor
         /// <summary>
         ///     No args
         /// </summary>
-        private static readonly Object[] noArgs = new Object[] { };
+        private static readonly Object[] noArgs = { };
 
         /// <summary>
         ///     A Props instance whose creator will create an actor that doesn't respond to any message
@@ -67,7 +132,7 @@ namespace Akka.Actor
         /// <summary>
         ///     Initializes a new instance of the <see cref="Props" /> class.
         /// </summary>
-        protected Props()
+        protected Props() 
             : this(defaultDeploy, null, noArgs)
         {
         }
@@ -152,6 +217,7 @@ namespace Akka.Actor
         ///     Gets the type.
         /// </summary>
         /// <value>The type.</value>
+        [JsonIgnore]
         public Type Type
         {
             get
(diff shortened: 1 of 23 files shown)


rogeralsing:
There is something funky going on in the `CanSerializeProps` test, if one pass a concrete value to any of the `.WithXXXX` methods, the testrunner blows up randomly after that specific test.

Calling `.WithXXXX` in the `CanSerializeDeploy` test works fine.
So there has to be something more to it in the props test.

Everyone is more than welcome to try to hunt this bugger down :)

src/core/Akka/Actor/Props.cs
```diff
@@ -22,8 +22,76 @@ namespace Akka.Actor
...
     ///  </code>
     /// </summary>
-    public class Props
+    public class Props : IEquatable<Props>
     {
+        public bool Equals(Props other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+            return CompareDeploy(other) && CompareSupervisorStrategy(other) && CompareArguments(other) && CompareInputType(other) && CompareProducer(other);
```
rogeralsing: Currently mocked, how should we do equality checks on Props? arguments have been serialized and can be anything, so we cant expect args to support euquality and thus, comparing the args array will be weird..

src/core/Akka/Actor/Props.cs
```diff
@@ -22,8 +22,76 @@ namespace Akka.Actor
...
+            if (other == null)
+                return false;
+
+            if (Arguments == null && other.Arguments == null)
+                return true;
+
+            if (Arguments == null)
+                return false;
+
+            return Arguments.SequenceEqual(other.Arguments);
```
rogeralsing: This will not work.

src/core/Akka/Actor/SupervisorStrategy.cs
```diff
@@ -245,8 +245,7 @@ public OneForOneStrategy(Func<Exception, Directive> decider) : this(-1, -1, deci
         /// <summary>
         /// Serialization-friendly constructor
         /// </summary>
-        [Obsolete("For serialization purposes only!")]
-        public OneForOneStrategy() : this(DefaultDecider) { }
+        protected OneForOneStrategy() : this(DefaultDecider) { }
```
rogeralsing: Removed public obsolete hack in favor for protected ctor now that the json serializer supports this.

src/core/Akka/Configuration/Config.cs
```diff
@@ -6,11 +6,6 @@ namespace Akka.Configuration
 {
     public class Config
     {
-        private Config _fallback;
-        private HoconValue _node;
-        private IEnumerable<HoconSubstitution> _substitutions;
```
rogeralsing: removed private fields in favor for autoprops with private setters to support serialization.

Aaronontheweb:
@rogeralsing :clap: this issues took a lot of work. Thanks for taking the time to do it. I'll do a detailed review tonight.

src/core/Akka/Routing/RouterConfig.cs
```diff
@@ -129,9 +129,37 @@ public override bool Equals(RouterConfig other)
     }
 
 
-    //TODO: ensure this can be serialized/deserialized fully
-    public abstract class Pool : RouterConfig
+    //TODO: ensure this can be serialized/deserialized fully    
+    public abstract class Pool : RouterConfig, IEquatable<Pool>
     {
+        //TODO: add supervisor strategy to the equality compare
+        public bool Equals(Pool other)
```
Aaronontheweb: Let's rename the "overrides" region to "equality" and move all of these members inside of it - just for organization purposes.

src/core/Akka/Actor/Props.cs
Aaronontheweb: Definitely not the most performant way of doing it, but you could compare the serialized output of both objects?

Aaronontheweb: Err, whoops - that doesn't make sense in any context other than serialization.

The other idea that might be worth considering is doing some reflection magic - traversing down the tree of publicly exposed properties (including nested user-defined objects within nested user-defined objects) and going all the way down until you get to comparable base types.

Or you could have a short-circuit that tests for equality only if the args are a primitive type or explicitly support IEquatable or IComparable. In the event that it's a POCO object without any equality comparer, just make sure that at least the types are identical and that both values are null or both values are not null.

src/core/Akka/Actor/Props.cs
```diff
@@ -152,6 +221,7 @@ public Props(Deploy deploy, Type type, params object[] args)
         ///     Gets the type.
         /// </summary>
         /// <value>The type.</value>
+        [JsonIgnore]
```
Aaronontheweb: Why are these and some of the other fields marked as `JsonIgnore` ?

Aaronontheweb:
Bah, my bad for mixing diff comments with commit comments.

But that's it for my review thus far. Looks good but I left you a few questions

src/core/Akka/Actor/Props.cs
rogeralsing: Json.NET includes readonly properties into the serialized json.
That is because Json.NET is a bit too smart for its own good, it can use those readoly properties as arguments to a parameterized constructor.

We now use the argless default ctor and so we only need data from properties that have a setter.

Aaronontheweb: Ah got it, that makes sense. Thank you.

src/core/Akka/Actor/SupervisorStrategy.cs
Aaronontheweb: I think we do this in a few other places in the codebase too. Can we remove this hack from those areas too?

src/core/Akka/Actor/Props.cs
rogeralsing: Do we really want to go through that hoop of weird magic just to be able to do .Equals on Props for tests?

I assume that equality for Props will be mostly used for tests, or even just only for this specific test?

src/core/Akka/Actor/SupervisorStrategy.cs
rogeralsing: Fixed

rogeralsing:
Squashed commits and cleaned up the comment flow here, removed comments for corrected issues.

src/core/Akka/Actor/Props.cs
rogeralsing: If we should supports props equality, I vote for serializing the args and comparing that.
It's an ugly hack, but the best option we have here IMO.

Aaronontheweb: In that case, maybe we're better off writing a `PropsComparer` inside the `TestKit` that does this, but only exposes that code within our tests.

Aaronontheweb: Err, `PropsEqualityComparer`

Aaronontheweb:
For the `SupervisorStrategy` bit, we should still allow the use of delegates *but not as the default option*. There are some capabilities that supervisors need to do such as publish child failures to another subscriber (this gets used into the Akka.Remote internals for listening for transport errors.) Can't get rid of the delegate option entirely.

We should warn the end user that any supervisor strategy that uses a delegate is not serializable. Kind of like how we do with `DynamicProps` (which may no longer be needed, now that we have `IndirectActorProducer` support.)

For 80% of cases, using a `Decider` DSL that just does pattern matching around Exception types will get the job done and we can guarantee that that is serializable. We should make that the safe default and allow advanced users access to the delegate method, with the warning that you can't serialize this stuff over the wire.

rogeralsing:
This is good to go.
We should add new issues for serializing `SupervisorStrategy` and how to verify that `Props` have been properly serialized. (they do get serialized correctly, we can just not do an Equals compare on them)

Other than that, serialization of primitives are done.

Issue #560, opened 2015-01-28
Add F# hello world sample project.


avoxm:
see #559

PR #584, opened 2015-02-02
Add F# hello world sample project. #560

michal-franc:


michal-franc:
I messed up with git remote ... will create new PR

PR #538, opened 2014-12-27
F# API - README + more changes

Horusiath:
## Breaking Changes

I've exposed a `Watch/Unwatch` methods into separate `ICanWatch` interface. It's now implemented by `IActorContext`, `Inboxable` and F# actors. Since `Inboxable.Watch` had previously void as return type, now it returns actor reference (also have an `Unwatch` method, which was not present before). I've also removed a `watch` function from F# API (it has been replaced by `monitor`, mentioned below).

Also some of the F# modules (Configuration, Strategy) now require qualified access - calling their functions without module prefix was unclear and could result in enigmatic code i.e. `load` vs `Configuration.load`.

## New F# functions

1. `Configuration.load()` - alias to `ConfigurationFactory.Load` method.
2. `monitor`/`demonitor` functions - similar to `Watch`/`Unwatch` (naming inspired by Erlang). Can operate directly on Inbox or Actor instances.
3. `pub`, `sub`, `unsub` functions - Publish/Subscribe functions. Reason for abbreviation is that Pub/Sub is also a popular naming shortcut for related pattern.
4. Scheduling functions - `schedule` and `scheduleOnce` work on functions, while `scheduleTell` and `scheduleTellOnce` work on message passed to actor refs. Moreover they take native F# functions as a parameters, so that explicit conversion to `System.Action` is no longer necessary.

## README.md for F# project

All F# API functions, grouped and explained with examples. I hope, that this could be a foundation to the F# API Documentation on Akka.NET home page.

PR #585, opened 2015-02-03
Akka.Persistence - F# guaranteed delivery, more tests, moved projects

Horusiath:
- `Akka.Persistence.FSharp` and `PersistenceExample.FsApi` has been moved to src/core and src/examples directories (they were directly in src/ before).
- Changes considering F# persistence plugin API:
    - `exec` method doesn't affect the aggregate state now - it can be modified from that method by emitting events with callback in `Persist`, `AsyncPersist` and `Defer` methods of the eventsourced parameter.
    - Methods mentioned above has switched params (callback is defined first, sequence of events second) - I think it better reflects a F#-way of dealing with sequential data
- F# persistence plugin now exposes a function to spawning guaranteed delivery actors.
- `PersistentViewSpec.PersistentView_should_run_size_limited_updates_on_user_request` test has been fixed.


rogeralsing:
@Horusiath can you fix this PR?
Looks good but can't be merged atm

Horusiath:
I've fixed some of the compilation problems with F# projects and applied some of the @dsyme suggestions from #538 review:

- Changed Akka.FSharp from module to namespace, rest of the exposed API has been splitted into separate *AutoOpen* modules.
- Strategy factory functions have been refactored to static methods with optional arguments.
- `pub`, `sub` and `unsub` changed to `publish`, `subscribe` and `unsubscribe`

All of the changes has been updated in docs and examples.

PR #605, opened 2015-02-08
Added Async Await support for TaskDispatcher.

rogeralsing:
This means that any actor running under the TaskDispatcher can now safely use async await inside actors.

The behavior used here is what Orleans call a Reentrant actor, that is, while the actor awaits an async task, it can still process other messages.
Once the async task completes, the TaskDispatcher will marshall execution back into the actors context and re-apply ambient properties such as Sender, CurrentMessage etc.

This does come with a cost, setting state that can flow with async operations is costly, so the pingpong benchmark suffers a 50% throughput loss when this is enabled.
However, this will have little to no effect on real world implementations where there are real work being performed in the actors.

### DIFF
--- a/src/core/Akka/Actor/ActorRef.cs
+++ b/src/core/Akka/Actor/ActorRef.cs
@@ -3,10 +3,13 @@ using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.Remoting.Messaging;
 using System.Threading.Tasks;
 using Akka.Actor.Internals;
 using Akka.Dispatch.SysMsg;
 using System.Threading;
+using Akka.Actor.Internal;
+using Akka.Dispatch;
 using Akka.Event;
 using Akka.Util;
 using Akka.Util.Internal;
@@ -51,12 +54,16 @@ namespace Akka.Actor
         private readonly TaskCompletionSource<object> _result;
         private readonly Action _unregister;
         private readonly ActorPath _path;
-        private ActorRef _sender;
+        private readonly ActorRef _actorAwaitingResult;
 
-        public FutureActorRef(TaskCompletionSource<object> result, ActorRef sender, Action unregister, ActorPath path)
+        public FutureActorRef(TaskCompletionSource<object> result, ActorRef actorAwaitingResult, Action unregister, ActorPath path)
         {
+            if (ActorCell.Current != null)
+            {
+                _actorAwaitingResultSender = ActorCell.Current.Sender;
+            }
             _result = result;
-            _sender = sender ?? NoSender;
+            _actorAwaitingResult = actorAwaitingResult ?? NoSender;
             _unregister = unregister;
             _path = path;
         }
@@ -75,6 +82,8 @@ namespace Akka.Actor
         private const int INITIATED = 0;
         private const int COMPLETED = 1;
         private int status = INITIATED;
+        private readonly ActorRef _actorAwaitingResultSender;
+
         protected override void TellInternal(object message, ActorRef sender)
         {
             if (message is SystemMessage) //we have special handling for system messages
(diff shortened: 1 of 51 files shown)


src/core/Akka/Actor/ActorRef.cs
```diff
@@ -85,14 +94,23 @@ protected override void TellInternal(object message, ActorRef sender)
             {
                 if (Interlocked.Exchange(ref status, COMPLETED) == INITIATED)
                 {
-                    _unregister();
```
rogeralsing: I might have borked something here. 
Github PR compare for finding changes FTW...

src/core/Akka/Dispatch/ConcurrentQueueMailbox.cs
```diff
@@ -69,7 +69,7 @@ private void Run()
                     Mailbox.DebugPrint(ActorCell.Self + " processing message " + envelope);
 
                     //run the receive handler
-                    ActorCell.Invoke(envelope);
+                    ActorCell.Dispatcher.Dispatch(ActorCell, envelope);
```
rogeralsing: This is for aligning more to JVM Akka.
The Dispatcher can execute extra logic before processing each message.

The Mailbox should probably not know about the owner cell but the dispatcher when port is complete.

src/core/Akka/Dispatch/Dispatchers.cs
```diff
@@ -209,6 +209,7 @@ public MessageDispatcher Lookup(string dispatcherName)
         public MessageDispatcher FromConfig(string path)
         {
             //TODO: this should not exist, it is only here because we dont serialize dispathcer when doing remote deploy..
+            //TODO: @Aaron, we do now, should this be removed?
```
Aaronontheweb: @rogeralsing yes, we should remove this and then re-run the `RemoteDeploy` tests in Akka.Remote. Everything should go through.

src/core/Akka/Dispatch/ConcurrentQueueMailbox.cs
Aaronontheweb: I like it

rogeralsing:
cc @akkadotnet/developers can anyone fetch this one and try it out locally?
I'm getting quite a bit of weird XUnit exceptions/freezes, not sure if it is my XUnit runner or if it is the updated code that makes things blow up.

smalldave:
@rogeralsing tried your fork. I'm using the resharper xunit runner and weirdly it is only picking up the ActorAsyncAwaitSpec tests. Not sure why that is. Also are you supposed to install xunit.runner.visualstudio in each test project? That seems wrong but maybe that is what xunit expects?

rogeralsing:
@smalldave I actually have no idea how one is supposed to use the nuget based runner, but installing it per test project made the tests appear in the test explorer atleast :)

smalldave:
@rogeralsing looks like that is what you are supposed to do. Seems odd to add a runner as a project dependency. Afraid I may not be the best person to check this just at the minute. I've got the latest resharper xunit extension and it has 'issues' with old xunit tests.

Aaronontheweb:
We're holding off on this because of the XUnit issue, right?

rogeralsing:
@Aaronontheweb yes, either someone else verifies if this PR works, or we upgrade to 2.0.
I can't get 1.9.2 to work correctly with the new runner, and the old runner seems to be replaced.

rogeralsing:
this now uses xUnit 2.0,  tests pass except two which is unrelated to the async await stuff.

* ReachabilityTableMustHaveCorrectStatusForAMixOfNodes , has an invalid cast in it. possibly due to changes in xUnit api. (cc @Aaronontheweb )

*  PersistentView_should_run_size_limited_updates_on_user_request , fails due to a continuation that runs on the threadpool. this is initalized from a TaskCompletionSource , where the continuation runs w/o a actor context.. don't see how this one could ever have worked. (cc @Horusiath )

Other than that, it works and all code is unaffected by the async await stuff, which is only enabled if an actor uses the `TaskDispatcher`

Horusiath:
@rogeralsing If you have to, skip this test, I'll take a look on it after merge

Aaronontheweb:
Make sure you try this out with the MultiNodeTestRunner. Has a strong dependency on XUnit

PR #606, opened 2015-02-09
ClusterDeatchWatchSpecs now pass

### LINKED ISSUE #602: Severe latency with ClusterRemoteWatcher Terminated notifications for nodes marked as Cluster.Down
Working through the `ClusterDeathWatchSpec` today and discovered that the test is failing due to severe latency between these two network events.

### Expected

| Node1 | Node2 | Node3 | Node4 | Node5 |
|:-----------:|:------------:|:------------:|:------------:|:------------:|
| UP | UP | UP | UP | UP |
| DeathWatch --> Node2, Node3 | .... | .... | .... | .... |
| .... | .... | Cluster.Down Node2 | .... | .... |
| ExpectMsg<Terminated> from Node2 | DOWN | .... | .... | .... |
| Cluster.Down Node3 | DOWN | .... | .... | .... |
| ExpectMsg<Terminated> from Node3 | DOWN | DOWN | .... | .... |

### Actual

| Node1 | Node2 | Node3 | Node4 | Node5 |
|:-----------:|:------------:|:------------:|:------------:|:------------:|
| UP | UP | UP | UP | UP |
| DeathWatch --> Node2, Node3 | .... | .... | .... | .... |
| .... | .... | Cluster.Down Node2 | .... | .... |
| ExpectMsg<Terminated> from Node2 | DOWN | .... | .... | .... |
| **TIMEOUT** | DOWN | .... | .... | .... |
| DOWN | DOWN | .... | .... | .... |

### Issue
Eventually the `Terminated` message does arrive to the subscribed actor on Node1, but in some instances it's taken over 15 seconds. Should be closer to ~200ms.

I suspect one of the following is the source of the issue:
* The `AddressTerminatedTopic` isn't firing correctly when the `Gossip` arrives to Node1, notifying the node that node3 is now `MemberStatus.Down`.
* The `RemoteWatcher`, which the `ClusterRemoteWatcher` depends on, isn't firing correctly when the connection to Node3 is terminated.
* There's a scheduling problem behind `Gossip`, where information isn't being propagated quickly enough (this won't take long to rule out.)
* Cluster.Down does not correctly fire `AddressTerminated` or does not notify the `ClusterActorRefProvider` to begin letting subscribers know that this node is shutting down.

Aaronontheweb:
close #602 

`ClusterRemoteWatcher` is now fully covered per the original `ClusterDeathWatchSpec`.

Unearthed some bugs as part of this unit test and fixed them accordingly:
* Found a race condition that resulted in unhandled errors within Helios during periods of rapid successive calls to `Disconnect`. Pushed a new version of Helios, v1.3.6, with fixes for this and referenced it throughout Akka.NET.
* `RemoteWatcher` was incorrectly deployed under the `/user/` actor hierarchy. It now lives happily under `/system/`
* `AddressTerminated` did not correctly fire within `RemoteWatcher`
* Fixed a couple sources of exceptions within `Conductor` inside the Akka.Remote.TestKit
* Added some equality members and implicit case operators to `ActorPath.Surrogate` so it can be used inside `ReceiveActor.Receive<>` predicates that expect an ActorPath.

This is one of the more important multi-node tests because cluster deployment is pretty useless without reliable deathwatch, so that this passes makes me feel pretty optimistic about the state of the plumbing inside Akka.Cluster. Many more `MultiNodeSpec`s yet to go though.

### DIFF
--- a/src/core/Akka.Cluster.Tests/MultiNode/ClusterDeathWatchSpec.cs
+++ b/src/core/Akka.Cluster.Tests/MultiNode/ClusterDeathWatchSpec.cs
@@ -1,7 +1,16 @@
-﻿using Akka.Actor;
+﻿using System;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Threading;
+using Akka.Actor;
 using Akka.Configuration;
+using Akka.Event;
+using Akka.Remote;
 using Akka.Remote.TestKit;
 using Akka.TestKit;
+using Akka.TestKit.TestActors;
+using Akka.Util.Internal;
+using Xunit;
 
 namespace Akka.Cluster.Tests.MultiNode
 {
@@ -25,7 +34,7 @@ namespace Akka.Cluster.Tests.MultiNode
             _third = Role("third");
             _fourth = Role("fourth");
             _fifth = Role("fifth");
-
+            DeployOn(_fourth, @"/hello.remote = ""@first@""");
             CommonConfig = ConfigurationFactory.ParseString(@"akka.cluster.publish-stats-interval = 25s")
                 .WithFallback(MultiNodeLoggingConfig.LoggingConfig)
                 .WithFallback(DebugConfig(true))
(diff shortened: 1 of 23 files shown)


src/core/Akka.Cluster.Tests/MultiNode/ClusterDeathWatchSpec.cs
```diff
@@ -68,55 +77,314 @@ private ClusterDeathWatchSpec(ClusterDeathWatchSpecConfig config)
...
+        }
+
+        /* 
+         * NOTE: it's not possible to watch a path that doesn't exist in Akka.NET
+         * REASON: in order to do this, you fist need an ActorRef. Can't get one for
+         * a path that doesn't exist at the time of creation. In Scala Akka they have to use
+         * System.ActorFor for this, which has been deprecated for a long time and has never been
+         * supported in Akka.NET.
+        */
+        //public void AnActorWatchingARemoteActorInTheClusterMustReceiveTerminatedWhenWatchedPathDoesNotExist()
```
Aaronontheweb: cc @akkadotnet/developers - need a second opinion on this.

In canonical Akka this portion of the tests requires an `ActorRef` for a non-existent path, creating using `ActorFor` - which has been deprecated within Akka for many versions now and has never been supported in Akka.NET. Therefore I think this test case is more for legacy Akka apps and doesn't apply to Akka.NET.

Can anyone confirm the following for me as a sanity check? (from my code comment)

  /* 
+         * NOTE: it's not possible to watch a path that doesn't exist in Akka.NET
+         * REASON: in order to do this, you fist need an ActorRef. Can't get one for
+         * a path that doesn't exist at the time of creation. In Scala Akka they have to use
+         * System.ActorFor for this, which has been deprecated for a long time and has never been
+         * supported in Akka.NET.
+        */

smalldave:
Had a good look through this. Looks great. Decent logging makes a huge difference.

Issue #452, opened 2014-10-05
Implement bounded message queues


HCanber:
Currently all the following are just copies of the Unbounded versions.
- `BoundedMessageQueue`
- `BoundedDequeMessageQueue`
- `BoundedDequeBasedMailbox`

At least `BoundedMessageQueue` needs to be updated. Check the others as well.

Aaronontheweb:
Marking this issue as up for grabs.

avoxm:
I'll take this. 
What should be the default capacity if nothing is provided ? Should it be without an upper bound if nothing is provided ?

rogeralsing:
Completed in #676.

PR #611, opened 2015-02-10
Implementing bounded message queue #452

avoxm:
Updating BoundedMessageQueue class to act as blocking and bounded
thread-safe collection.

### DIFF
--- a/src/core/Akka/Dispatch/MessageQueues/BoundedMessageQueue.cs
+++ b/src/core/Akka/Dispatch/MessageQueues/BoundedMessageQueue.cs
@@ -7,12 +7,21 @@ namespace Akka.Dispatch.MessageQueues
     /// <summary>An Bounded mailbox message queue.</summary>
     public class BoundedMessageQueue : MessageQueue, BoundedMessageQueueSemantics
     {
-        //TODO: Implement BoundedMessageQueue. Currently it's just a copy of UnboundedMessageQueue
-        private readonly ConcurrentQueue<Envelope> _queue = new ConcurrentQueue<Envelope>();
+        private readonly BlockingCollection<Envelope> _queue;
+
+        public BoundedMessageQueue()
+        {
+            _queue = new BlockingCollection<Envelope>();
+        }
+
+        public BoundedMessageQueue(int boundedCapacity)
+        {
+            _queue = new BlockingCollection<Envelope>(boundedCapacity);
+        }
 
         public void Enqueue(Envelope envelope)
         {
-            _queue.Enqueue(envelope);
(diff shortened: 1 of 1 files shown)


src/core/Akka/Dispatch/MessageQueues/BoundedMessageQueue.cs
```diff
@@ -7,12 +7,21 @@ namespace Akka.Dispatch.MessageQueues
...
+
+        public BoundedMessageQueue(int boundedCapacity)
+        {
+            _queue = new BlockingCollection<Envelope>(boundedCapacity);
+        }
 
         public void Enqueue(Envelope envelope)
         {
-            _queue.Enqueue(envelope);
+            _queue.Add(envelope);
```
rogeralsing: Use this overload instead:
https://msdn.microsoft.com/en-us/library/dd287149(v=vs.110).aspx

`TryAdd(Envelope,Timespan)`

And use the `PushTimeout` property.
If I recall correctly, if the timeout occurs, the message should be sent to deadletters, not sure, check Scala impl/docs.

rogeralsing:
This needs to be configurable using the Akka config.

```
    bounded-mailbox {
      mailbox-capacity = 1000
      mailbox-push-timeout-time = 10s
      mailbox-type = "....
```

rogeralsing:
Looks good, just a few minor details that needs fixing

src/core/Akka/Dispatch/MessageQueues/BoundedMessageQueue.cs
avoxm: What is the equivalent class in scala version ?

rogeralsing:
config: https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/dispatch/Mailbox.scala#L642

rogeralsing:
enqueue (post) https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/dispatch/Mailbox.scala#L512

rogeralsing:
@avoxm are you still working on this? otherwise I can fix the last bits myself

avoxm:
@rogeralsing I'll check the scala version and finalize this shortly. Thanks

Aaronontheweb:
@avoxm any reason why this is closed? think all we need is to implement @rogeralsing's feedback

avoxm:
@Aaronontheweb I'll be opening a new pull request to address the notes from @rogeralsing

avoxm:
I guess it would be easier for you to merge this way :)

Aaronontheweb:
Ah, got it :)

PR #613, opened 2015-02-10
Some documentation fixes

nvivo:
Convert examples to C#, removed some references of JVM and Scala.


nvivo:
While modifying the "Working with Actors", there are many places saying "Don't forget the second argument on Tell(....)!!!", but it seems Akka.NET by default passes the Self inside the actor. I wonder if this should be removed from most examples.

Maybe leave it only if explaining explicitly that something that requires it, like a child actor answering on behalf of the parent.

rogeralsing:
Looks great, thanks!

rogeralsing:
Yes, the sender in a call to `.Tell` is optional, it will be resolved from the current context.
So it can be omitted, and I think your suggestion is good, leave it in if there is an Ask/ReqResp type of communication going on.

nvivo:
@rogeralsing, I'll keep translating and fixing things first, then I'll go back and try to improve the examples.

rogeralsing:
:+1:

PR #614, opened 2015-02-11
Example application for stashing.

nvivo:
This is a proposal for a stashing sample app. I was going to create this in the documentation, but the code would be too long. Please check if this sample is good enough to add to repository.

I'm not using any supervisor, but it seems to do the trick. If there is anything you'd like me to change, I'm all ears.


skotzko:
This looks good — thanks for putting it together! :+1:

nvivo:
I'm closing this, as I'll probably put in an external repository.

PR #618, opened 2015-02-12
Port of the FaultTolerance sample from Java

nvivo:
This is a port of the code from http://doc.akka.io/docs/akka/snapshot/java/fault-tolerance-sample.html
I ported this as a single file, as the original, to be include as reference in the Akka.NET documentation.

I thought it was important to be able to execute this code and see the results, so I put it in an example project. If it shouldn't be here, let me know.


rogeralsing:
This is great!
We are somewhat restrictive with new examples in the core repo, but this is exactly the thing we should have there!
Thanks!

Issue #234, opened 2014-07-25
Throw InvalidActorNameException if actor is created with invalid name


rogeralsing:
If an actor is created with an invalid name, `InvalidActorNameException` should be thrown.

We need consistency for documentation

rogeralsing:
Impelemented

PR #623, opened 2015-02-13
Fixes issue #234 (Throw InvalidActorNameException if actor is created wi...

kstaruch:
...th invalid name) by porting Akkas non-regex solution for name validation.

### DIFF
--- a/src/core/Akka/Actor/ActorPath.cs
+++ b/src/core/Akka/Actor/ActorPath.cs
@@ -1,7 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text.RegularExpressions;
 using Akka.Util;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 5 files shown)


src/core/Akka/Actor/ActorPath.cs
```diff
@@ -92,13 +91,42 @@ public override int GetHashCode()
...
-        /// The regex that actor names must conform to RFC 2396, http://www.ietf.org/rfc/rfc2396.txt
+        /// Method that checks if actor name conforms to RFC 2396, http://www.ietf.org/rfc/rfc2396.txt
         /// Note that AKKA JVM does not allow parenthesis ( ) but, according to RFC 2396 those are allowed, and 
         /// since we use URL Encode to create valid actor names, we must allow them.
         /// </summary>
-        public static readonly Regex ElementRegex =
-            new Regex(@"^(?:[-a-zA-Z0-9:@&=+,.!~*'_;\(\)]|%[0-9a-fA-F]{2})(?:[-a-zA-Z0-9:@&=+,.!~*'\$_;\(\)]|%[0-9a-fA-F]{2})*$", RegexOptions.Compiled | RegexOptions.CultureInvariant);
+        public static bool IsValidPathElement(string s)
+        {
+            return s.Length > 0 && !s.StartsWith("$") && Validate(s.ToCharArray(), s.Length);
```
rogeralsing: We should probably have a null check also (?)

kstaruch:
BTW, there's a small inconsistency with Akka regarding null name - Akka throws while Akka.net generates random name instead. I suppose it's intentional, so left it as it was.

src/core/Akka/Actor/ActorPath.cs
kstaruch: Currently nulls are not going to happen as we generate the name when null is passed, but I agree since it's public. I'll add it.

rogeralsing:
Looks good!

rogeralsing:
And thank you for your contribution! :)

PR #628, opened 2015-02-14
Working async await ftw

rogeralsing:
Finally got the testrunners and fixed some broken Nuget references.

This PR gives us the ability to use async await for both Ask and other async operations when using the `TaskDispatcher`
It is completely separate from all other infrastructure so it does not affect other usecases.

How it works.
If an actor is configured with `.WithDispatcher("akka.actor.task-dispatcher"),`
It will dispatch all messages through the task dispatcher, which in turn places all relevant continuation state into the `LogicalCallContext` which makes it possible for state to flow async.
It also executes the mailbox using Tasks and applies an `ActorTaskScheduler` to manage all task execution within the current context.

### DIFF
--- a/src/core/Akka.Tests/Akka.Tests.csproj
+++ b/src/core/Akka.Tests/Akka.Tests.csproj
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\packages\xunit.runner.visualstudio.2.0.0-rc1-build1030\build\net20\xunit.runner.visualstudio.props" Condition="Exists('..\..\packages\xunit.runner.visualstudio.2.0.0-rc1-build1030\build\net20\xunit.runner.visualstudio.props')" />
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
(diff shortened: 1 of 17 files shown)


rogeralsing:
This is finally safe to merge.
All tests pass except for the already broken cluster config localiation parser

Aaronontheweb:
Whoops, meant to comment on this issue instead of the previous PR - can we make sure that the MultiNodeTestRunner plays nice with the new XUnit?

rogeralsing:
Multi node tests works fine.
(Still using 1.9.2 here, 2.0 was far too unstable)

rogeralsing:
Any objections to this one or are we good to go?

src/core/Akka.Tests/Akka.Tests.csproj
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\packages\xunit.runner.visualstudio.2.0.0-rc1-build1030\build\net20\xunit.runner.visualstudio.props" Condition="Exists('..\..\packages\xunit.runner.visualstudio.2.0.0-rc1-build1030\build\net20\xunit.runner.visualstudio.props')" />
```
HCanber: This line and a line below in the same file are mistakes and should not have been merged, right?

rogeralsing:
Those are refs to the new nuget deployed  xunit testrunner.
The xunit testrunner plugin has been marked obsolete.

HCanber:
So we have switched to use xunit 2? you said "(Still using 1.9.2 here, 2.0 was far too unstable)"

rogeralsing:
We are still using 1.9.2.
The vs xunit testrunner plugin is however no longer functional.
The vs gallery download points to instructions on how to use the new vs testrunner from nuget.

PR #634, opened 2015-02-16
Bringing Akka.Remote up to code with canonical Akka

Aaronontheweb:
In the course of trying to fix a bug that turned out to be an issue with Akka.Remote.TestKit, I ended up modernizing the Endpoints system to use what canonical Akka currently uses.

Here's the full list of changes:

1. Rewrote `EndpointWriter` from being an FSM to a normal actor
	1. Implemented `IPiorityMessage` handling, which means that heartbeats get delivered first once a connection opens.
	2. Removed all instances of `PatternMatch`
	3. Added adaptive backoff support
1. Implemented the `ThrottleTransportAdapter` and `FailureInjectorTransportAdapter`
	1. Added specs that cover both (`AkkaProtocolStressTest` for `FailureInjectorTransportAdapter`)
	2. Fixed fully qualified namespaces inside the `Remote.conf` to correctly load them
	3. Integration tested with `HeliosTcpTransport`
1. Fixed the `TransportAdapter` system
	1. Fixed a HOCON bug that stopped dictionaries from being parsed correctly
	2. Previously the `TransportAdapter` table was a no-op - wired it up and it now correctly layers and nests transports within adapters.
	3. Correctly prefixes Akka.NET addresses with applied adapters schemes - i.e. `akka.thrttl.tcp` or `akka.gremlin.tcp` are now baked into the address whenever you used an adapter on top of a transport.
1. Made small changes to parts of Akka.Cluster, mostly removed `PatternMatch` and replaced it with `if...`
2. Ported the `PromiseActorRef` and added it to `Futures.cs` not sure if we want it or not. I ended up not using it inside the `ThrottleTransportAdapter`. It's part of the new structure canonical Akka is using for Ask.
3. Made some modifications to `TestKit`, mostly to support different layers of post-test termination hooks. Used for shutting down additional `ActorSystem` instances inside Akka.Remote.Tests usually.

I'm going to resume work on the multinode testkit stuff, but please take a look through this and let me know if I should make any changes before this PR can be accepted.

### DIFF
--- a/src/contrib/testkits/Akka.TestKit.Xunit/TestKit.cs
+++ b/src/contrib/testkits/Akka.TestKit.Xunit/TestKit.cs
@@ -61,7 +61,7 @@ namespace Akka.TestKit.Xunit
         /// the system. Otherwise you'll leak memory.
         /// </remarks>
         /// </summary>
-        protected virtual void AfterTest()
+        protected virtual void AfterAll()
         {
             Shutdown();
         }
@@ -104,7 +104,7 @@ namespace Akka.TestKit.Xunit
                 {
                     if(disposing)
                     {
-                        AfterTest();                        
+                        AfterAll();                        
                     }
                 }
                 _isDisposed = true;
--- a/src/core/Akka.Cluster.Tests/MultiNode/InitialHeartbeatSpec.cs
+++ b/src/core/Akka.Cluster.Tests/MultiNode/InitialHeartbeatSpec.cs
@@ -72,6 +72,7 @@ namespace Akka.Cluster.Tests.MultiNode
                 : base(config)
             {
                 _config = config;
+                MuteMarkingAsUnreachable();
             }
 
             [MultiNodeFact]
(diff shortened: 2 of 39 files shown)


Aaronontheweb:
Oh yeah, also implemented the `RARP` actor system extension - used for forcing key internal classes to use the remoting dispatcher within Akka.Remote.

smalldave:
Wow. That is a lot of code! Clustering stuff looks fine. Getting rid of Match was on my (large) todo list.
Took a cursory look at Remoting stuff and bits I understand looked good. Lots of interesting stuff in there.

Aaronontheweb:
@smalldave I removed the `PatternMatch` stuff mostly to make it easier for me to stick some breakpoints in the individual message receive blocks :p, but the added performance benefits are a bonus.

I'm going to make a habit of gradually cleaning those up inside Akka.Remote and Akka.Cluster while I work on other parts of them

PR #643, opened 2015-02-18
Added exception handling to async await support

rogeralsing:
Added Receive API for tasks, that is, the receive actor can handle async receive methods correctly now.
Added more tests to verify that async behavior is correct

This PR makes the async await support handle exceptions correctly, so any exception thrown in a continuation/await block will be propagated to the supervisor.


Aaronontheweb:
That failing tests here aren't the result of issues with your code. Merging this in.

PR #644, opened 2015-02-18
Fixing problematic specs

Aaronontheweb:
removed skip attribute from RemotingSpecs

brought RoutingSpec up to code with canonical Akka

adjusting timeouts on regularly crashing specs

We have some specs that appear to be pretty racy - sensitive to GC pauses and other stuff. Did some work to bring those up to code and adjust timeouts here and there. Would like to have tests pass consistently.

### DIFF
--- a/src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
+++ b/src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
@@ -110,7 +110,7 @@ namespace Akka.Remote.Tests
             for (var i = 0; i < 5; i++)
             {
                 router.Tell("", probe.Ref);
-                var expected = probe.ExpectMsg<ActorRef>();
+                var expected = probe.ExpectMsg<ActorRef>(TimeSpan.FromSeconds(5));
                 replies.Add(expected.Path);
             }
 
@@ -130,7 +130,7 @@ namespace Akka.Remote.Tests
             for (var i = 0; i < 5; i++)
             {
                 router.Tell("", probe.Ref);
-                var expected = probe.ExpectMsg<ActorRef>();
+                var expected = probe.ExpectMsg<ActorRef>(TimeSpan.FromSeconds(5));
                 replies.Add(expected.Path);
             }
 
@@ -150,7 +150,7 @@ namespace Akka.Remote.Tests
             for (var i = 0; i < 5000; i++)
             {
                 router.Tell("", probe.Ref);
-                var expected = probe.ExpectMsg<ActorRef>();
+                var expected = probe.ExpectMsg<ActorRef>(TimeSpan.FromSeconds(5));
                 replies.Add(expected.Path);
             }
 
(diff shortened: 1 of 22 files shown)


src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
```diff
@@ -110,7 +110,7 @@ public void RemoteRouter_must_deploy_its_children_on_remote_host_driven_by_confi
             for (var i = 0; i < 5; i++)
             {
                 router.Tell("", probe.Ref);
-                var expected = probe.ExpectMsg<ActorRef>();
+                var expected = probe.ExpectMsg<ActorRef>(TimeSpan.FromSeconds(5));
```
rogeralsing: I think we should use this TestKit helper:

```csharp
        /// <summary>
        /// If <paramref name="timeout"/> is defined it is returned; otherwise
        /// the config value "akka.test.single-expect-default" is returned.
        /// </summary>
        public TimeSpan GetTimeoutOrDefault(TimeSpan? timeout)
```

`GetTimeoutOrDefault(null)` gives you the default expect timeout by config
Just to avoid magic constants IMO.

Aaronontheweb: That's a good idea - I'll modify this commit and do that.

Looks like the Inbox and Resizer specs need some love too.

Aaronontheweb:
Nice, adding a new commit to the PR kicked off a new build. Wanted to make sure that works :)

Aaronontheweb:
The Resizer and Inbox specs have always been a little weird. Going to do into those now.

As for the PersistentView specs, those are failing a lot - going to mark them as SKIP = FIXME for now cc @Horusiath

Aaronontheweb:
Aside from the persistence stuff (which is an early alpha, after all) I think there's a legitimate bug with routers evicting routees. Seeing that issue pop up in a couple of different specs now. Going to investigate.

Aaronontheweb:
Well this makes it a little easier:

   at Akka.Tests.Routing.ResizerSpec.DefaultResizer_must_backoff_within_10_seconds() in d:\BuildAgent\work\49b164d63843fb4\src\core\Akka.Tests\Routing\ResizerSpec.cs:line 213
------- Stdout: -------
[WARNING][2/19/2015 1:27:34 AM][Thread 0009][akka://RuntimeType-208/user] DeadLetter from [akka://RuntimeType-208/user] to [akka://RuntimeType-208/user]: <Received dead system message: <DeathWatchNotification>: [akka://RuntimeType-208/user/testActor222], ExistenceConfirmed=True, AddressTerminated=False>
[WARNING][2/19/2015 1:27:34 AM][Thread 0009][akka://RuntimeType-208/user/$b] DeadLetter from [akka://RuntimeType-208/user/$b] to [akka://RuntimeType-208/user/$b]: <Received dead letter from [akka://RuntimeType-208/user/$b]: Akka.Routing.Resize>
[WARNING][2/19/2015 1:27:34 AM][Thread 0009][akka://RuntimeType-208/user] DeadLetter from [akka://RuntimeType-208/user] to [akka://RuntimeType-208/user]: <Received dead system message: <DeathWatchNotification>: [akka://RuntimeType-208/user/$b], ExistenceConfirmed=True, AddressTerminated=False>

Looks like the `Resize` message that this test needs to passed gets delivered as a Deadletter before it has a chance to complete its run. I'll see what I can do to sort that out.

PR #650, opened 2015-02-19
added small pause for mailbox init overhead (ResizerSpec)

Aaronontheweb:
Figured out the fundamental issue with the ResizerSpecs - there's a tiny amount of mailbox pressure that happens at the very beginning of each actor being created: the mailboxes start with a status of busy. The Resizer detects this and increases capacity.

If you pause for a short amount of time (after all of the mailbox create / ActorCell wireup is done) the pressure values appear to return to normal.

Going to test this theory with a PR


Aaronontheweb:
Going to take a look at the ReceiveTimeoutSpec and InboxSpec, but otherwise I think this should help stabilize ResizerSpec issues.

PR #660, opened 2015-02-19
Proposal for an AsyncUntypedActor.

nvivo:
This actor allows to use tasks inside Receive but have the same behavior as a normal actor. It's implemented as a state machine. If any messages arrive during async processing, they will be stashed until the current processing is done. If exceptions are thrown during async operations, they will be handled by the supervisor as if it was a synchronous operation.

Example:

```cs
public class MyAsyncActor : AsyncUntypedActor
{
    protected override async Task OnReceiveAsync(object message)
    {
        // keep the sender, because it will lose context after "await" in 0.8
        var sender = Sender;

        var values = await DoSomeDatabaseWork();

        sender.Tell(values);
    }
}
``` 

It will also behave as a normal UntypedActor if the task is synchronous. That means that if you return a completed task, it won't cause any overhead in processing.

Let me know if there are any issues or suggestions.

### DIFF
--- a/src/core/Akka/Actor/AsyncUntypedActor.cs
+++ b/src/core/Akka/Actor/AsyncUntypedActor.cs
@@ -1,0 +1,56 @@
+﻿using System.Runtime.ExceptionServices;
+using System.Threading.Tasks;
+
+namespace Akka.Actor
+{
+    public abstract class AsyncUntypedActor : UntypedActor, WithUnboundedStash
+    {
+        public IStash Stash { get; set; }
+        bool _awaiting;
+        readonly object AwaitComplete = new object();
+
+        protected sealed override void OnReceive(object message)
+        {
+            if (_awaiting)
+            {
+                if (message == AwaitComplete)
+                {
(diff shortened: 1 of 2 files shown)


rogeralsing:
I will dig into this,
Its a good start, but needs to work with the ActorTaskScheduler too.
We should also discuss how to fit async into existing actors.
Eg. TypedActor would be pretty easy to alter,
But we have some extra for persistence too.
Ill open an issue for discussion

nvivo:
Ok. I'll keep using it and see how it goes, if there are problems or improvements I'll try to bring up.

nvivo:
@rogeralsing Is there an issue to discuss this yet? Didn't find anything.

Aaronontheweb:
Closing this - the approach we'd like to go with is something that can be plugged into all actors (although it'll be easier to use in some actors over others - i.e. `ReceiveActor` will have much better built-in support than `UntypedActor` will), rather than just a specific type of actor.

See the latest on #686

PR #661, opened 2015-02-19
Async Await stacktraces fix

rogeralsing:
Updated async await support to give correct exception stacktraces instead of throwing `AggregateException`


Aaronontheweb:
The `RemoteRouterSpec` crash should be resolved in `/dev` - fixed the gating timeout setting for those specs so they can reestablish a Helios connection before the timeout expires. As for the other failing specs those are known issues

PR #665, opened 2015-02-20
Serializable supervisor

rogeralsing:
Added decider type that replaces the decider lambda
Added Decider serialization spec
Added SupervisorStrategy Serialization spec
Added Decider DSL

PR for #665 

Serializable Deciders can now be created like this:
```csharp
   var decider = Decider.From(
                Directive.Restart,
                Directive.Stop.When<ArgumentException>(),
                Directive.Stop.When<NullReferenceException>());
```

Old lambda support is still valid but not serializable.

### DIFF
--- a/src/core/Akka/Actor/SupervisorStrategy.cs
+++ b/src/core/Akka/Actor/SupervisorStrategy.cs
@@ -154,7 +154,7 @@ namespace Akka.Actor
             }
         }
 
-        protected abstract bool LoggingEnabled { get;}
+        protected bool LoggingEnabled { get; set; }
 
         private void Publish(IActorContext context, LogEvent logEvent)
         {
@@ -174,10 +174,7 @@ namespace Akka.Actor
         ///     <see cref="DefaultDecider" />.
         /// </summary>
         /// <value>The default.</value>
-        public static SupervisorStrategy DefaultStrategy
-        {
-            get { return new OneForOneStrategy(DefaultDecider); }
-        }
+        public static readonly SupervisorStrategy DefaultStrategy = new OneForOneStrategy(DefaultDecider);    
 
         /// <summary>
         /// This method is called after the child has been removed from the set of children.
@@ -193,10 +190,9 @@ namespace Akka.Actor
     /// </summary>
     public class OneForOneStrategy : SupervisorStrategy
     {
-        private readonly int _maxNumberOfRetries;
-        private readonly int _withinTimeRangeMilliseconds;
-        private readonly Func<Exception, Directive> _decider;
-        private readonly bool _loggingEnabled;
+        public int MaxNumberOfRetries { get; private set; }
+        public int WithinTimeRangeMilliseconds { get; private set; }
+        public IDecider Decider { get; private set; }
 
         /// <summary>
         ///     Applies the fault handling `Directive` (Resume, Restart, Stop) specified in the `Decider`
@@ -208,11 +204,28 @@ namespace Akka.Actor
         ///     if the limit is exceeded the child actor is stopped.
         /// </param>
         /// <param name="withinTimeRange">duration of the time window for maxNrOfRetries, Duration.Inf means no window.</param>
-        /// <param name="decider">mapping from Exception to <see cref="Directive" /></param>
-        public OneForOneStrategy(int? maxNrOfRetries, TimeSpan? withinTimeRange, Func<Exception, Directive> decider)
-            : this(maxNrOfRetries.GetValueOrDefault(-1), withinTimeRange.GetValueOrDefault(Timeout.InfiniteTimeSpan).Milliseconds, decider)
+        /// <param name="localOnlyDecider">mapping from Exception to <see cref="Directive" /></param>
+        public OneForOneStrategy(int? maxNrOfRetries, TimeSpan? withinTimeRange, Func<Exception, Directive> localOnlyDecider)
+            : this(maxNrOfRetries.GetValueOrDefault(-1), withinTimeRange.GetValueOrDefault(Timeout.InfiniteTimeSpan).Milliseconds, localOnlyDecider)
         {
+            //Intentionally left blank
+        }
 
+        /// <summary>
+        ///     Applies the fault handling `Directive` (Resume, Restart, Stop) specified in the `Decider`
+        ///     to all children when one fails, as opposed to <see cref="OneForOneStrategy" /> that applies
+        ///     it only to the child actor that failed.
+        /// </summary>
+        /// <param name="maxNrOfRetries">
+        ///     the number of times a child actor is allowed to be restarted, negative value means no limit,
+        ///     if the limit is exceeded the child actor is stopped.
+        /// </param>
+        /// <param name="withinTimeMilliseconds">duration in milliseconds of the time window for <paramref name="maxNrOfRetries"/>, negative values means no window.</param>
+        /// <param name="localOnlyDecider">Mapping from an <see cref="Exception"/> to <see cref="Directive"/></param>
+        /// <param name="loggingEnabled">If <c>true</c> failures will be logged</param>
+        public OneForOneStrategy(int maxNrOfRetries, int withinTimeMilliseconds, Func<Exception, Directive> localOnlyDecider, bool loggingEnabled = true) : this(maxNrOfRetries,withinTimeMilliseconds,new LocalOnlyDecider(localOnlyDecider),loggingEnabled)
+        {
+            //Intentionally left blank
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Did you merge my latest changes into your branch first before you pulled? `ReceiveTimeoutSpec` should have been fixed in them.

rogeralsing:
cc @akkadotnet/developers 
Is the DSL for creating serializable OK?
Do we need to change it?

Horusiath:
Ok for me. F# API probably should be updated, but this commit probably won't break anything in this matter.

Aaronontheweb:
This looks good - ready to merge it when you are @rogeralsing

src/core/Akka/Actor/SupervisorStrategy.cs
```diff
@@ -193,10 +190,9 @@ public static SupervisorStrategy DefaultStrategy
     /// </summary>
     public class OneForOneStrategy : SupervisorStrategy
     {
-        private readonly int _maxNumberOfRetries;
-        private readonly int _withinTimeRangeMilliseconds;
-        private readonly Func<Exception, Directive> _decider;
-        private readonly bool _loggingEnabled;
+        public int MaxNumberOfRetries { get; private set; }
+        public int WithinTimeRangeMilliseconds { get; private set; }
+        public IDecider Decider { get; private set; }
```
HCanber: Please change this and `AllForOneStrategy` back to use `private readonly`. You've made this class mutable when it used to be immutable.

src/core/Akka/Actor/SupervisorStrategy.cs
```diff
@@ -443,4 +514,74 @@ public enum Directive
...
+        }
+
+        public Directive Decide(Exception cause)
+        {
+            return _decider(cause);
+        }
+    }
+
+    public class DeployableDecider : IDecider
+    {
```
HCanber: Make this immutable as well by using `private readonly` fields instead of `{ get; private set; }` properties.

src/core/Akka/Actor/SupervisorStrategy.cs
```diff
@@ -443,4 +514,74 @@ public enum Directive
...
+        public DeployableDecider(Directive defaultDirective, params KeyValuePair<Type, Directive>[] pairs)
+        {
+            DefaultDirective = defaultDirective;
+            Pairs = pairs;
+        }
+
```
HCanber: I think we should have this overload as well:
``` C#
public DeployableDecider(Directive defaultDirective, IEnumerable<KeyValuePair<Type, Directive>> pairs)
```

HCanber:
I like the DSL. See line comments as well.
One thought: Are there any cases when a user might want to specify an exception type and all deriving from it? For example `WhenDerive<ArgumentException>` which would match `ArgumentException` as well as `ArgumentNullException`, `ArgumentOutOfRangeException` and so on. (`WhenDerive` needs a better name in that case)

src/core/Akka/Actor/SupervisorStrategy.cs
rogeralsing: That won't serialize

src/core/Akka/Actor/SupervisorStrategy.cs
rogeralsing: All the auto prop changes are for serialization support.

src/core/Akka/Actor/SupervisorStrategy.cs
HCanber: Oh yes it will, if you expose them as Get-properties and have a constructor with arguments that has the same name as the properties. This works:
``` C#
public class Test
{
	private readonly int _nummer;
	private readonly string _sträng;

	public Test(int nummer, string sträng)
	{
		_nummer = nummer;
		_sträng = sträng;
	}
	public int Nummer { get { return _nummer; } }
	public string Sträng { get { return _sträng; } }
}

var test = new Test(4711, "Håkan");
var json = JsonConvert.SerializeObject(test);
var test2 = JsonConvert.DeserializeObject<Test>(json);
System.Diagnostics.Debug.Assert(test.Nummer == test2.Nummer);
System.Diagnostics.Debug.Assert(test.Sträng == test2.Sträng);
```

rogeralsing:
Good point, I do think we want derivate checks since that is how the `is` operator would do it (and `.Match<>`

code like `if (cause.GetType() == typeof(ArgumentException)` would be unlikely in the real world, right?
If you want to match on any subtypes one would just specify those earlier in the list of types.

rogeralsing:
Name suggestions for the DSL?

`When`
`WhenDerive`
`WhenCauseIs`
`WhenInstanceOf`

src/core/Akka/Actor/SupervisorStrategy.cs
rogeralsing: Yepp, but this requires that there is an ctor that matches the properties in the Json data exactly, if an argument is added to a ctor, or if a property is changed in some way, it breaks. super fragile, Properties map 1-1 to the Json data which is mych less fragile.
We have had problems with this in the past with the other Akka primitives.

src/core/Akka/Actor/SupervisorStrategy.cs
rogeralsing: I've added this to the DSL now, any reason why you would want it on the `DeployableDecider` also?

HCanber:
Are you thinking we should have 
- One that matches exactly and one that matches all derivates (meaaning the kvp has to be turned into a 3-tuple)? Then `When` (matches type and all derivates) and `WhenExactly` (matches only the type)

or
- Only one, that matches derivates?
`When` is good for that.

src/core/Akka/Actor/SupervisorStrategy.cs
HCanber: That's what tests are for. Write a test that proves it can be serialized and deserialized. Much better than to make it mutual.

src/core/Akka/Actor/SupervisorStrategy.cs
HCanber: The minimum datatype needed by `DeployableDecider` is `IEnumerable<KVP>` so I think it should support that.
`params KVP[]` is really only syntactic that's nice to have.

src/core/Akka/Actor/SupervisorStrategy.cs
rogeralsing: I this is the wrong way to go, this makes a super strong coupling to Json.NET since that is pretty much the only standard serializer that supports this kind of ctor deserialization, while most do support property based deserialization.
This would also mean that every primitive would have to expose a ctor that matches all of its properties. Maybe it supports private/protected ctors? otherwise we would expose an API that is for the serializer only and not intended for end users.
We would also have to verify how this plays with surrogates.

If we do this, it should be applied to all the primitives.
I think its a very high price for the sake of purism.

rogeralsing: this is not a joke, but I just got bitten by this exact issue.

I did add the IEnumerable overload to the DeployableDecider, so there is one for arrays and one for IEnumerable.

```
Result Message:	Newtonsoft.Json.JsonSerializationException : Unable to find a constructor to use for type Akka.Actor.DeployableDecider. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute. Path 'DefaultDirective', line 1, position 75.
Result StackTrace:	
```

So the constructor resolution is apparently unable to decide which ctor is the most specific one to use.

HCanber: I think the problem here is that we try to serialize the same classes that are used internally? Shouldn't we use a dto (that might be tightly coupled to the serializer used) instead?

rogeralsing: We can do that too, but then we need surrogate support for each and every primitive.
Because a user might decide to reference one of these primitives. e.g.

```csharp
public class MyMessage
{

        public Props MyProps {get;private set;} 
        ....
}

So the default serializer needs to be able to resolve those DTO's and turn them into the real thing even if they are embedded deep in a user message.

We already do this for `ActorRef` and `ActorPath` and maybe as you say, that is the best way to do it for all the primitives in the end, to keep the core types clean from serialization logic and extract all that into the surrogates.

So far so good, but then again, some primitives are used by other serializers that does not support surrogates.
e.g. deploy info passed around in the remoting layer carried by Protobuf.

So it's not a super easy problem to solve entirely.

HCanber: Using JsonConverters is another option, that will take care of embedded primitives and the problem with using the correct ctor.

HCanber:
Was this really finished to be merge? I thought we were in the middle of a discussion.

rogeralsing:
We agreed yesterday on the meeting to open up a new issue regarding serialization yesterday and treat that separately.

>Only one, that matches derivates? When is good for that.
That is how the current code works, so all good?

Or was there something more?

HCanber:
Oh, ok. My understanding was that we should treat serialization of the other in another issue and only talk about serializing the strategies in this one.
I really dislike changing it to be mutable. I think surrogates is the way to go. But let's discuss that in another issue then, and fix serialization of the Strategies later. Have we got an issue where we can discuss this?

rogeralsing:
I opened this #681 to continue the serialization discussion

PR #676, opened 2015-02-25
Implementing bounded message queue #452

avoxm:


avoxm:
Some notes : 
In Scala version there is no parameter-less constructor and I assume at some point we will need to get rid of it too. But since there are some dependencies  (see BoundedDequeMessageQueue and BoundedDequeBasedMailbox) from it I am leaving it here. New issue can be opened for this if needed.

Also enqueue method in Scala version under MessageQueue trait/interface takes accepts ActorRef parameter, which later is used by BoundedMessageQueue to send to dead-letters if enqueue failed.

I see mostly if not always code that uses these queues has reference to ActorRef, so I assume it would be pretty easy to update the code and pass it to the queues, but  MessageQueue interface must be updated and all dependent classes must be updated accordingly. I took the safest root with leaving as is at this point and separate issue can be opened if this needs to be addressed.

rogeralsing:
Looks good,
The refactoring of the message queues is a bigger task than this, so I'll open up a new for that.
I'm merging this to allow us to start working on the queues.

Issue #652, opened 2015-02-19
Prefix user-facing interfaces with "I" to match .NET Convention


SeanKilleen:
Per #633, this should likely be undertaken after #651.

HCanber:
If this is to be done, it should be done before v1? Right?

Aaronontheweb:
@HCanber yeah, we should just bit the bullet and do this now. 

Does anyone want this issue assigned to them? Should be straightforward.

SeanKilleen:
I'll try to take this one on. I haven't gotten to really contribute yet but with that in mind I'm happy to give it a shot. :)

HCanber:
@Aaronontheweb  Was this closed intentionally, or automatically by merging my PR? Because it isn't fixed.

Aaronontheweb:
@HCanber whoops, I thought you said in your PR that this was finished. Looks like I misread it. My bad!

HCanber:
These are the ones @nvivo listed in #633, and which have been fixed. Once this list is complete, we need to recheck and make sure there aren't any more.

- [X] `Akka.Actor.ActorRefProvider, Akka` (Public)
- [X] `Akka.Actor.ActorRefScope, Akka` (Public)
- [X] `Akka.Actor.AutoReceivedMessage, Akka` (Public)
- [X] `Akka.Actor.Cell, Akka` (Public)
- [X] `Akka.Actor.Inboxable, Akka` (Public)
- [X] `Akka.Actor.IndirectActorProducer, Akka` (Public)
- [X] `Akka.Actor.Internal.ChildrenContainer, Akka` (Public)
- [X] `Akka.Actor.Internal.ChildStats, Akka` (Public)
- [X] `Akka.Actor.Internal.InternalSupportsTestFSMRef`2, Akka` (Public)
- [X] `Akka.Actor.Internal.SuspendReason+WaitingForChildren, Akka`
- [X] `Akka.Actor.Internals.InitializableActor, Akka` (Public)
- [X] `Akka.Actor.LocalRef, Akka`
- [ ] `Akka.Actor.LoggingFSM, Akka` (Public)
- [ ] `Akka.Actor.NoSerializationVerificationNeeded, Akka` (Public)
- [ ] `Akka.Actor.PossiblyHarmful, Akka` (Public)
- [X] `Akka.Actor.RepointableRef, Akka` (Public)
- [ ] `Akka.Actor.WithBoundedStash, Akka` (Public)
[...]

PR #685, opened 2015-02-27
WIP: Prefix Interfaces with "I"

SeanKilleen:
In support of #652, the purpose of this branch will be to prefix all interfaces in all assemblies with an `I` to match the .NET convention.

Do not merge this PR until the "WIP" is removed and it is ready to be considered.

Thanks!


SeanKilleen:
FYI, my standard for success at this point will be whether the project continues to build in Visual Studio and whether the tests pass (once I get them up and running).

I'll be utilizing ReSharper for a lot of the safe renaming, and I'll do my best not to replace strings, etc. where they aren't associated.

PS. This seems like a large task, and will be a lot to review. Does it make more sense to split this into several PRs? Feel free to chime in if you've got an opinion; otherwise I'll keep chugging along on here.

SeanKilleen:
Ooops; I'll open another PR.

PR #689, opened 2015-02-28
Refactor childStats requestRestartPermission algorithm

rodrigovidal:


HCanber:
Why was this changed? For non trivial "refactorings" I'd like to see an explanation as we might need to be able to go back and see why things change.

In this case, the changes aren't just a refactoring, this PR has altered the algorithm. Unfortunately it introduced several bugs, and needs to be reverted, as support for negative values have been completely removed. 

The values for `maxNrOfRetries` and `withinTimeMilliseconds` are specified when creating `AllForOneStrategy` and `OneForOneStrategy`. From the docs for these constructors:
-- `maxNrOfRetries`: the number of times a child actor is allowed to be restarted, negative value means no limit, if the limit is exceeded the child actor is stopped.
-- `withinTimeMilliseconds`: duration in milliseconds of the time window for maxNrOfRetries, negative values means no window.

These calls now does the incorrect thing:
``` C#
RequestRestartPermission(-1,-1)
RequestRestartPermission(10,-1)
RequestRestartPermission(-1, 10)
```

rodrigovidal:
@HCanber Yes, I was fooled by the variables names. I considered that maxNrOfRetries and withinTimeMilliseconds could not be negative. I was wrong. Do we have tests for this? I would be glad to help adding it. 

@rogeralsing sorry about this mistake.

Aaronontheweb:
@rodrigovidal looks like we don't have a test that would have caught this issue - would you mind adding one?

rodrigovidal:
@Aaronontheweb Sure! I'll do it :+1:

Aaronontheweb:
awesome, thanks @rodrigovidal

HCanber:
@rodrigovidal :+1:

PR #692, opened 2015-02-28
Does this optimization really make difference?

rodrigovidal:


rogeralsing:
Agree, anyone else @akkadotnet/developers is there subtile things we miss?

Aaronontheweb:
I'd like @HCanber's opinion on this one

HCanber:
As I can see it, the optimized version
- Avoids a funcion call to `SplitNameAndUid()` 
- Avoids creating an instance of `NameAndUid` class
- Avoids an if statement and comparison

Not a large optimization so from that point of view this PR is acceptable.

But, since this is part of the core of Akka, I'd like to see it aligned with Akka JVM as much as possible, unless there's a really good reason for diverging.
As Akka JVM has this optimization, and there is no real reason for diverging, I vote for not merging this PR.

rodrigovidal:
@HCanber I understand what you are saying, but are we doing this kind of optimization? To me this seems like code/logic duplication because the if statement exists in two places.

Although I understand the desire to be aligned with Akka JVM, it doesn't mean they are always right.

I would like to see at least one benchmarking or test that justify this optimization.

HCanber:
@rodrigovidal As I said, I don't want to keep this because of speed but because of keeping it aligned with Akka JVM. When they are wrong we should divert, but in this case they aren't really wrong, are they?

Aaronontheweb:
I'm fine with closing this PR, @HCanber - it's an instance where maintainability and understandability with Akka JVM is the more important issue. We still appreciate your hard work and thoughtfulness though, @rodrigovidal

Issue #72, opened 2014-03-15
Add descriptive ToString for all built in messages


rogeralsing:
Add descriptive ToString() impl for both SystemMessage and AutoReceived messages.

PR #695, opened 2015-02-28
Added meaningful ToString() to all system messages

Horusiath:
cc #72

Hope this will close the case.


Aaronontheweb:
@Horusiath I doubt these changes were the reason why the ThrottlerTransportAdapterSpec failed, so I queued up a new version of this build to run just to verify. Thanks for taking care of this!

Aaronontheweb:
Wow, that's not at all what I expected... Somehow these changes did effect these three tests consistently, even on different build servers. What could have caused that?

rogeralsing:
Found the issue..
It throws Null ref exception in Identify.ToString

```
        {
            return "<Identify>: " + MessageId.ToString();
        }
```

Remove ToString and it is fine

PR #697, opened 2015-03-01
Revert parts of #691 "refactor suspend children"

HCanber:
This reverts parts of commit a43580322ec573e9e5259ef6e885e45fd601ad1e.

The previous version was optimized for the most common scenario, when `exceptFor == null`.
The LINQ expression introduced in #691 is creating a list every time and uses more enumerators. It also makes a call to `List<T>.Contains()` when we know it's not necessary.

Normally LINQ expression can make loops more readable, bu in this case, and it might come down to a matter of taste, I don't think the LINQ version was more readable.

### DIFF
--- a/src/core/Akka/Actor/ActorCell.Children.cs
+++ b/src/core/Akka/Actor/ActorCell.Children.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


rodrigovidal:
Again, I would like to see this kind of optimization justified. Suspend actors is not a thing you do 1000x per second, I think.

I can't believe that you don't think the linq version é more readable. :)

PR #698, opened 2015-03-01
Add info on Title and Description for a PR in CONTRIBUTING.md

HCanber:
A PR should always be justified. 
As a few PR lately have slipped through without any description at all, except a title saying "Refactor Xyz" I thought it could be useful to have a paragraph we could point contributors to.

Maybe a native English speaker could have a look at what I've written. :)


Aaronontheweb:
@HCanber I think this is great - we should also include the following:

DON'T DO: <insert bad example of PR title here>

DO: <insert good example>

HCanber:
Any suggestions to what to put under those headings?

Issue #686, opened 2015-02-27
Discussion: How to provide better support for async/await inside actors


nvivo:
This issue is to discuss how to better support async/await inside actors.

Currently, all actors are synchronous and support for Tasks is achieved using PipeTo/Tell. Most of .NET nowadays makes heavy use of tasks, so even simple things like writing files and accessing databases can be async. It seems to me that Akka.NET should support async/await receive methods without requiring users to implement this logic by breaking the processing into multiple messages with PipeTo.

The "async" support shouldn't break the synchronous model where one message is processed at a time. That means that any implementation of an "async Receive method" should somehow wait for the "Receive" method to finish its processing (possibly awaiting for the task) before processing a new message. This behavior allows any mix of implementations like using "async/await" but also allowing "PipeTo/Tell" where needed.

I have been thinking about how to support this, and here are some thoughts about the problem and some proposed solutions to discuss. I must say my knowledge of the project code is  nowhere good enough to provide a complete list, and as @rogeralsing said, it must support the ActorTaskScheduler, but I hope this at least helps with the research.

### Solution 1 - Implementing Async versions for each actor

Pros:
- Can be achieved with no changes to existing code, thus less risk of breaking things
- Allows a very simple syntax, usually requiring the user to implement some form of `Task<bool> ReceiveAsync` instead of `bool Receive`

Cons: 
- Requires stashing support to prevent blocking, leaving the user without option to choose which stash to use
- Requires implementing (and maintaining) repetitive pattern for each existing actor

There is a sample implementation of an async actor here: https://github.com/akkadotnet/akka.net/pull/660. Also, not all actors require a new class. Both FSM and ReceiveActor can be implemented by just adding methods to the existing actors (`ReceiveAsync/WhenAsync`).

Actors like EventStream or Logger shouldn't require async implementations in my opinion, thus making the total of only 3 implementations (4 if Handler/TypedActor should be included) to fully support the official Actors. This seems to be the easiest way to achieve better async support today.

### Solution 2 - Extensions methods for ActorBase

Pros: 
- Exposes a single implementation that could in theory work with any actor

Cons:
- Still requires stashing support
- Would require users to add method calls to their implementations, making it more complex to use
- Different actors have different ways to implement Receive, which seems to indicate multiple extension implementations would be required anyway

I tried this solution, but any implementation I could think of would be at least as complex to maintain, and more complex to use than solution 1.

### Solution 3 - Change ActorBase to support Task directly

Pros: 
[...]

PR #700, opened 2015-03-01
Async Await support w/o performance hit and TypedActor/UntypedActor support

rogeralsing:
This PR removes the need to use the special "task-dispatcher" for async await inside ReceiveActor.
If the receive handler is an async method, it will apply the correct scheduling mechanics for that handler only.

This PR also introduces the `RunTask` method inside `UntypedActor` (possibly it should be in `ActorBase`)
This method lets you run tasks within the ActorTaskScheduler.

This does however not address "reentrancy", messages can still be processed during await periods (but never in parallell)
But this could easily be handled by extending how RunActor behaves, e.g. an argument for if reentrancy is allowed or not, and if not, suspend the mailbox.

The idea behind `RunTask` is that it allows you to run any task, or multiple tasks, within the safety of the ActorTaskScheduler. this means that reading and writing state is safe both using async/await and continueWith.

cc. @nvivo

### DIFF
--- a/src/core/Akka/Actor/UntypedActor.cs
+++ b/src/core/Akka/Actor/UntypedActor.cs
@@ -1,4 +1,9 @@
 ﻿using System;
+using System.Runtime.ExceptionServices;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.Dispatch;
 
 namespace Akka.Actor
 {
@@ -11,6 +16,16 @@ namespace Akka.Actor
         {
             OnReceive(message);
             return true;
+        }
+
+        protected void RunTask(AsyncBehavior behavior, Action action)
+        {
+            ActorTaskScheduler.RunTask(behavior,action);
+        }
+
+        protected void RunTask(AsyncBehavior behavior, Func<Task> action)
+        {
+            ActorTaskScheduler.RunTask(behavior,action);
         }
 
         /// <summary>
--- a/src/core/Akka.MultiNodeTestRunner.Shared/Reporting/SpecRunCoordinator.cs
+++ b/src/core/Akka.MultiNodeTestRunner.Shared/Reporting/SpecRunCoordinator.cs
@@ -80,7 +80,7 @@ namespace Akka.MultiNodeTestRunner.Shared.Reporting
         /// and report their results
         /// </summary>
         /// <returns>An awaitable task, since this operation uses the <see cref="Futures.Ask"/> pattern</returns>
-        private Task HandleEndSpec(EndSpec endSpec)
+        private void HandleEndSpec(EndSpec endSpec)
         {
             var futures = new Task<NodeData>[Nodes.Count];
 
@@ -94,7 +94,7 @@ namespace Akka.MultiNodeTestRunner.Shared.Reporting
             var sender = Context.Sender;
 
             //wait for all Ask operations to complete and pipe the result back to ourselves, including the ref for the original sender
-            return Task.WhenAll(futures)
+            Task.WhenAll(futures)
                 .PipeTo(Self, sender);
         }
 
--- a/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
+++ b/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
@@ -20,7 +20,7 @@ namespace Akka.Tests.Dispatch
             {
                 Self.Tell("change");
                 await Task.Delay(TimeSpan.FromSeconds(1));
-                //we expect that state should have changed due to an incoming message
+                //we expect that state should not have changed due to an incoming message
                 Sender.Tell(state);
             });
         }
--- a/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
+++ b/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
@@ -1,11 +1,45 @@
 ﻿using System;
 using System.Threading.Tasks;
 using Akka.Actor;
+using Akka.Dispatch;
 using Akka.TestKit;
 using Xunit;
 
 namespace Akka.Tests.Dispatch
 {
+    public class SuspendActor : ReceiveActor
+    {
+        public SuspendActor()
+        {
+            var state = 0;
+            Receive<string>(s => s == "change", _ =>
+            {
+                state = 1;
+            });
+            Receive<string>(AsyncBehavior.Suspend, async _ =>
+            {
+                Self.Tell("change");
+                await Task.Delay(TimeSpan.FromSeconds(1));
+                //we expect that state should have changed due to an incoming message
+                Sender.Tell(state);
+            });
+        }
+    }
+    public class ReentrantActor : ReceiveActor
+    {
+        public ReentrantActor()
+        {
+            var state = 0;
+            Receive<string>(s => s == "change", _ => state = 1);
+            Receive<string>(AsyncBehavior.Reentrant, async _ =>
+            {
+                Self.Tell("change");
+                await Task.Delay(TimeSpan.FromSeconds(1));
+                //we expect that state should have changed due to an incoming message
+                Sender.Tell(state);
+            });
+        }
+    }
     public class AsyncAwaitActor : ReceiveActor
     {
         public AsyncAwaitActor()
@@ -16,10 +50,30 @@ namespace Akka.Tests.Dispatch
                 var self = Self;
                 await Task.Yield();
                 await Task.Delay(TimeSpan.FromSeconds(1));
-                Assert.Same(sender,Sender);
+                Assert.Same(sender, Sender);
                 Assert.Same(self, Self);
                 Sender.Tell("done");
             });
+        }
+    }
+
+    public class UntypedAsyncAwaitActor : UntypedActor
+    {
+        protected override void OnReceive(object message)
+        {
+            if (message is string)
+            {
+                RunTask(AsyncBehavior.Suspend,  async () =>
+                {
+                    var sender = Sender;
+                    var self = Self;
+                    await Task.Yield();
+                    await Task.Delay(TimeSpan.FromSeconds(1));
+                    Assert.Same(sender, Sender);
+                    Assert.Same(self, Self);
+                    Sender.Tell("done");
+                });
+            }
         }
     }
 
@@ -39,11 +93,37 @@ namespace Akka.Tests.Dispatch
         }
     }
 
+    public class UntypedAsker : UntypedActor
+    {
+        private readonly ActorRef _other;
+
+        public UntypedAsker(ActorRef other)
+        {
+            _other = other;
+        }
+
+        protected override void OnReceive(object message)
+        {
+            if (message is string)
+            {
+                RunTask(AsyncBehavior.Suspend, async () =>
+                {
+                    var sender = Sender;
+                    var self = Self;
+                    var res = await _other.Ask("start");
+                    Assert.Same(sender, Sender);
+                    Assert.Same(self, Self);
+                    Sender.Tell(res);
+                });
+            }
+        }
+    }
+
     public class BlockingAsker : ReceiveActor
     {
         public BlockingAsker(ActorRef other)
         {
-            Receive<string>( _ =>
+            Receive<string>(_ =>
             {
                 //not async, blocking wait
                 var res = other.Ask("start").Result;
@@ -75,12 +155,6 @@ namespace Akka.Tests.Dispatch
     {
         private readonly ActorRef _callback;
 
-        protected override void PostRestart(Exception reason)
-        {
-            _callback.Tell("done");
-            base.PostRestart(reason);
-        }
-
         public AsyncExceptionActor(ActorRef callback)
         {
             _callback = callback;
@@ -89,6 +163,12 @@ namespace Akka.Tests.Dispatch
                 await Task.Yield();
                 ThrowException();
             });
+        }
+
+        protected override void PostRestart(Exception reason)
+        {
+            _callback.Tell("done");
+            base.PostRestart(reason);
         }
 
         private static void ThrowException()
@@ -104,11 +184,11 @@ namespace Akka.Tests.Dispatch
             Receive<string>(m =>
             {
                 //this is also safe, all tasks complete in the actor context
-                Task.Delay(TimeSpan.FromSeconds(1))
-                    .ContinueWith(t =>
-                    {
-                        Sender.Tell("done");
-                    });
+                RunTask(AsyncBehavior.Suspend, () =>
+                {
+                    Task.Delay(TimeSpan.FromSeconds(1))
+                        .ContinueWith(t => { Sender.Tell("done"); });
+                });
             });
         }
     }
@@ -116,33 +196,45 @@ namespace Akka.Tests.Dispatch
     public class AsyncTplExceptionActor : ReceiveActor
     {
         private readonly ActorRef _callback;
-        protected override void PostRestart(Exception reason)
-        {
-            _callback.Tell("done");
-            base.PostRestart(reason);
-        }
 
         public AsyncTplExceptionActor(ActorRef callback)
         {
             _callback = callback;
             Receive<string>(m =>
             {
-                Task.Delay(TimeSpan.FromSeconds(1))
-                    .ContinueWith(t =>
-                    {
-                        throw new Exception("foo");
-                    });
+                RunTask(AsyncBehavior.Suspend, () =>
+                {
+                    Task.Delay(TimeSpan.FromSeconds(1))
+                   .ContinueWith(t => { throw new Exception("foo"); });
+                });               
             });
+        }
+
+        protected override void PostRestart(Exception reason)
+        {
+            _callback.Tell("done");
+            base.PostRestart(reason);
         }
     }
 
     public class ActorAsyncAwaitSpec : AkkaSpec
     {
         [Fact]
+        public async Task UntypedActors_should_be_able_to_async_await_ask_message_loop()
+        {
+            var actor = Sys.ActorOf(Props.Create<UntypedAsyncAwaitActor>(), "Worker");
+            var asker = Sys.ActorOf(Props.Create(() => new UntypedAsker(actor)), "Asker");
+            var task = asker.Ask("start", TimeSpan.FromSeconds(5));
+            actor.Tell(123, ActorRef.NoSender);
+            var res = await task;
+            Assert.Equal("done", res);
+        }
+
+        [Fact]
         public async Task Actors_should_be_able_to_async_await_in_message_loop()
         {
-            var actor = Sys.ActorOf(Props.Create<AsyncAwaitActor>().WithDispatcher("akka.actor.task-dispatcher"));
-            var task = actor.Ask<string>("start", TimeSpan.FromSeconds(55));
+            var actor = Sys.ActorOf(Props.Create<AsyncAwaitActor>());
+            var task = actor.Ask<string>("start", TimeSpan.FromSeconds(5));
             actor.Tell(123, ActorRef.NoSender);
             var res = await task;
             Assert.Equal("done", res);
@@ -151,11 +243,9 @@ namespace Akka.Tests.Dispatch
         [Fact]
         public async Task Actors_should_be_able_to_async_await_ask_message_loop()
         {
-            var actor = Sys.ActorOf(Props.Create<AsyncAwaitActor>().WithDispatcher("akka.actor.task-dispatcher"),
-                "Worker");
-            var asker = Sys.ActorOf(Props.Create(() => new Asker(actor)).WithDispatcher("akka.actor.task-dispatcher"),
-                "Asker");
-            var task = asker.Ask("start", TimeSpan.FromSeconds(55));
+            var actor = Sys.ActorOf(Props.Create<AsyncAwaitActor>(), "Worker");
+            var asker = Sys.ActorOf(Props.Create(() => new Asker(actor)), "Asker");
+            var task = asker.Ask("start", TimeSpan.FromSeconds(5));
             actor.Tell(123, ActorRef.NoSender);
             var res = await task;
             Assert.Equal("done", res);
@@ -164,11 +254,9 @@ namespace Akka.Tests.Dispatch
         [Fact]
         public async Task Actors_should_be_able_to_block_ask_message_loop()
         {
-            var actor = Sys.ActorOf(Props.Create<AsyncAwaitActor>().WithDispatcher("akka.actor.task-dispatcher"),
-                "Worker");
-            var asker = Sys.ActorOf(Props.Create(() => new BlockingAsker(actor)).WithDispatcher("akka.actor.task-dispatcher"),
-                "Asker");
-            var task = asker.Ask("start", TimeSpan.FromSeconds(55));
+            var actor = Sys.ActorOf(Props.Create<AsyncAwaitActor>().WithDispatcher("akka.actor.task-dispatcher"),"Worker");
+            var asker =Sys.ActorOf(Props.Create(() => new BlockingAsker(actor)).WithDispatcher("akka.actor.task-dispatcher"),"Asker");
+            var task = asker.Ask("start", TimeSpan.FromSeconds(5));
             actor.Tell(123, ActorRef.NoSender);
             var res = await task;
             Assert.Equal("done", res);
@@ -177,9 +265,8 @@ namespace Akka.Tests.Dispatch
         [Fact(Skip = "Maybe not possible to solve")]
         public async Task Actors_should_be_able_to_block_ask_self_message_loop()
         {
-            var asker = Sys.ActorOf(Props.Create(() => new BlockingAskSelf()).WithDispatcher("akka.actor.task-dispatcher"),
-                "Asker");
-            var task = asker.Ask("start", TimeSpan.FromSeconds(55));
+            var asker = Sys.ActorOf(Props.Create(() => new BlockingAskSelf()),"Asker");
+            var task = asker.Ask("start", TimeSpan.FromSeconds(5));
             var res = await task;
             Assert.Equal("done", res);
         }
@@ -187,7 +274,7 @@ namespace Akka.Tests.Dispatch
         [Fact]
         public void Actors_should_be_able_to_supervise_async_exceptions()
         {
-            var asker = Sys.ActorOf(Props.Create(() => new AsyncExceptionActor(TestActor)).WithDispatcher("akka.actor.task-dispatcher"));
+            var asker = Sys.ActorOf(Props.Create(() => new AsyncExceptionActor(TestActor)));
             asker.Tell("start");
             ExpectMsg("done", TimeSpan.FromSeconds(5));
         }
@@ -195,17 +282,32 @@ namespace Akka.Tests.Dispatch
         [Fact]
         public async Task Actors_should_be_able_to_use_ContinueWith()
         {
-            var asker = Sys.ActorOf(Props.Create<AsyncTplActor>().WithDispatcher("akka.actor.task-dispatcher"));
-            var res = await asker.Ask("start",TimeSpan.FromSeconds(5));
+            var asker = Sys.ActorOf(Props.Create<AsyncTplActor>());
+            var res = await asker.Ask("start", TimeSpan.FromSeconds(5));
             Assert.Equal("done", res);
         }
 
         [Fact]
         public void Actors_should_be_able_to_supervise_exception_ContinueWith()
         {
-            var asker = Sys.ActorOf(Props.Create(() => new AsyncTplExceptionActor(TestActor)).WithDispatcher("akka.actor.task-dispatcher"));
+            var asker = Sys.ActorOf(Props.Create(() => new AsyncTplExceptionActor(TestActor)));
             asker.Tell("start");
-            ExpectMsg( "done", TimeSpan.FromSeconds(5));
+            ExpectMsg("done", TimeSpan.FromSeconds(5));
+        }
+        [Fact]
+        public async Task Actors_should_be_able_to_reenter()
+        {
+            var asker = Sys.ActorOf(Props.Create(() => new ReentrantActor()));
+            var res = await asker.Ask<int>("start",TimeSpan.FromSeconds(5));
+            res.ShouldBe(1);
+        }
+
+        [Fact]
+        public async Task Actors_should_be_able_to_suspend_reentrancy()
+        {
+            var asker = Sys.ActorOf(Props.Create(() => new SuspendActor()));
+            var res = await asker.Ask<int>("start", TimeSpan.FromSeconds(555));
+            res.ShouldBe(0);
         }
     }
 }
--- a/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
+++ b/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
@@ -169,6 +169,8 @@ namespace Akka.Actor
                 else if (m is Recreate) FaultRecreate((m as Recreate).Cause);
                 else if (m is Suspend) FaultSuspend();
                 else if (m is Resume) FaultResume((m as Resume).CausedByFailure);
+                else if (m is SuspendReentrancy) HandleSuspendReentrancy();
+                else if (m is ResumeReentrancy) HandleResumeReentrancy();
                 else if (m is Terminate) Terminate();
                 else if (m is Supervise)
                 {
@@ -184,6 +186,16 @@ namespace Akka.Actor
             {
                 HandleInvokeFailure(cause);
             }
+        }
+
+        private void HandleSuspendReentrancy()
+        {
+            Mailbox.Suspend(MailboxSuspendStatus.AwaitingTask);
+        }
+
+        private void HandleResumeReentrancy()
+        {
+            Mailbox.Resume(MailboxSuspendStatus.AwaitingTask);
         }
 
         private void HandleCompleteTask(CompleteTask task)
@@ -338,6 +350,16 @@ namespace Akka.Actor
         public void Suspend()
         {
             SendSystemMessage(Dispatch.SysMsg.Suspend.Instance);
+        }
+
+        public void SuspendReentrancy()
+        {
+            SendSystemMessage(Dispatch.SysMsg.SuspendReentrancy.Instance);
+        }
+
+        public void ResumeReentrancy()
+        {
+            SendSystemMessage(Dispatch.SysMsg.ResumeReentrancy.Instance);
         }
 
         private void SendSystemMessage(SystemMessage systemMessage)
--- a/src/core/Akka/Actor/ReceiveActor.cs
+++ b/src/core/Akka/Actor/ReceiveActor.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using Akka.Actor.Internals;
 using Akka.Tools.MatchHandler;
 using System.Threading.Tasks;
+using Akka.Dispatch;
 
 namespace Akka.Actor
 {
@@ -73,23 +74,23 @@ namespace Akka.Actor
             base.Become(m => ExecutePartialMessageHandler(m, newHandler), discardOld);
         }
 
-
         protected void Receive<T>(Func<T,Task> handler)
         {
             EnsureMayConfigureMessageHandlers();
             _matchHandlerBuilders.Peek().Match<T>( m =>
             {
-                var task = handler(m);
+                Func<Task> wrap = () => handler(m);
+                ActorTaskScheduler.RunTask(AsyncBehavior.Suspend, wrap);
+            });
+        }
 
-                    task.ContinueWith(x =>
-                    {
-                        if (x.IsFaulted)
-                        {
-                            ExceptionDispatchInfo.Capture(task.Exception.InnerException).Throw();
-                        }
-
-                    }, TaskContinuationOptions.None);
-
+        protected void Receive<T>(AsyncBehavior behavior, Func<T, Task> handler)
+        {
+            EnsureMayConfigureMessageHandlers();
+            _matchHandlerBuilders.Peek().Match<T>(m =>
+            {
+                Func<Task> wrap = () => handler(m);
+                ActorTaskScheduler.RunTask(behavior, wrap);
             });
         }
 
--- a/src/core/Akka/Akka.csproj
+++ b/src/core/Akka/Akka.csproj
@@ -168,6 +168,7 @@
     <Compile Include="Configuration\Hocon\HoconTokenizer.cs" />
     <Compile Include="Configuration\Hocon\HoconValue.cs" />
     <Compile Include="Configuration\Hocon\IHoconElement.cs" />
+    <Compile Include="Dispatch\ActorTaskScheduler.cs" />
     <Compile Include="Dispatch\BoundedDequeBasedMailbox.cs" />
     <Compile Include="Dispatch\ConcurrentQueueMailbox.cs" />
     <Compile Include="Dispatch\DequeBasedMailbox.cs" />
--- a/src/core/Akka/Dispatch/ActorTaskScheduler.cs
+++ b/src/core/Akka/Dispatch/ActorTaskScheduler.cs
@@ -1,0 +1,146 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Runtime.Remoting.Messaging;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Dispatch.SysMsg;
+
+namespace Akka.Dispatch
+{
+    public enum AsyncBehavior
+    {
+        Reentrant,
+        Suspend
+    }
+
+    public class AmbientState
+    {
+        public ActorRef Self { get; set; }
+        public ActorRef Sender { get; set; }
+        public object Message { get; set; }
+    }
+
+    public class ActorTaskScheduler : TaskScheduler
+    {
+        public static readonly TaskScheduler Instance = new ActorTaskScheduler();
+        public static readonly TaskFactory TaskFactory = new TaskFactory(Instance);
+        public static readonly string StateKey = "akka.state";
+        private const string Faulted = "faulted";
+        private static readonly object Outer = new object();
+
+        public static void SetCurrentState(ActorRef self, ActorRef sender, object message)
+        {
+            CallContext.LogicalSetData(StateKey, new AmbientState
+            {
+                Sender = sender,
+                Self = self,
+                Message = message
+            });
+        }
+
+        protected override IEnumerable<Task> GetScheduledTasks()
+        {
+            return null;
+        }
+
+        protected override void QueueTask(Task task)
+        {
+            if (task.AsyncState == Outer)
+            {
+                TryExecuteTask(task);
+                return;
+            }
+
+            //we get here if the task needs to be marshalled back to the mailbox
+            //e.g. if previous task was an IO completion
+            var s = CallContext.LogicalGetData(StateKey) as AmbientState;
+
+            s.Self.Tell(new CompleteTask(s, () =>
+            {
+                TryExecuteTask(task);
+                if (task.IsFaulted)
+                    Rethrow(task, null);
+
+            }), ActorRef.NoSender);
+        }
+
+        protected override bool TryExecuteTaskInline(Task task, bool taskWasPreviouslyQueued)
+        {
+            if (taskWasPreviouslyQueued)
+                return false;
+
+            var s = CallContext.LogicalGetData(StateKey) as AmbientState;
+            var cell = ActorCell.Current;
+
+            //Is the current cell and the current state the same?
+            if (cell != null &&
+                s != null &&
+                cell.Self == s.Self &&
+                cell.Sender == s.Sender &&
+                cell.CurrentMessage == s.Message)
+            {
+                var res = TryExecuteTask(task);
+                return res;
+            }
+
+            return false;
+        }
+
+        public static void RunTask(AsyncBehavior behavior, Action action)
+        {
+            RunTask(behavior, () =>
+            {
+                action();
+                return Task.FromResult(0);
+            });
+        }
+
+        public static void RunTask(AsyncBehavior behavior, Func<Task> action)
+        {
+            var context = ActorCell.Current;
+
+            //if reentrancy is not allowed, suspend user message processing
+            if (behavior == AsyncBehavior.Suspend)
+            {
+                context.SuspendReentrancy();
+            }
+
+            SetCurrentState(context.Self, context.Sender, null);
+
+            //wrap our action inside a task, so that everything executing 
+            //directly or indirectly from the action is executed on our task scheduler
+
+            Task.Factory.StartNew(async _ =>
+            {
+
+                //start executing our action and potential promise style
+                //tasks
+                await action()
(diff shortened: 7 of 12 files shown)


Aaronontheweb:
@rogeralsing so these tasks don't run on the dispatcher, but rather the general purpose `ThreadPool` right?

rogeralsing:
The tasks run like normal tasks by TPL, the continuations, be it the code after await or normal continuewith, is piped back to execute inside the actor mailbox loop, just like pipeto

Aaronontheweb:
That's what I figured - thanks!

Asking because I'm thinking about the Fork-Join Dispatcher stuff. Despite the fact that .NET has much better built-in concurrency tools than the JVM, being able to manage the locality of tasks + threads is one area where the JVM does it better.

Doesn't really apply to user-defined code, but in some system actors (Remoting, in particular) there are areas where key Tasks (like setting a Transport to listen for incoming connections) are explicitly told to run on top of the dedicated dispatcher those actors use.

(Personally, that's not something I would try to re-create on the TPL in .NET _without_ data indicating that it's something we need doing.)

Aaronontheweb:
You broke my MultiNodeTestRuner bro!

image

Aaronontheweb:
LOL, looks like one of my Receive handlers already returns a task... even though it's using PipeTo... WTF

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.MultiNodeTestRunner.Shared/Reporting/SpecRunCoordinator.cs#L64

rogeralsing:
like-a-boss

nvivo:
Roger,  I'll try that once I get home,  but looks like a good start. 

One thing though is that you don't need that action.ContinueWith in the RunTask.  All that code is already part of the await syntax,  and exceptions are already unwraped if you use await. 

I also wonder if the task itself shouldn't be running in the actor context instead of the continuation. This would allow the task to set values inside the actor. 

What's the purpose of only the continuation running in the actor context?

nvivo:
BTW,  it's amazing the speed and energy you guys have. It's sunday for God's sake. Really like a boss!  =)

rogeralsing:
@nvivo am I missing something?
The entire block of code runs within the actor context. all of it is scheduled to the `ActorTaskScheduler`

```csharp
 Task.Factory.StartNew(async () => 
 {
        await action().ContinueWith(x =>
        {
            if (x.IsFaulted)
            {
                 ExceptionDispatchInfo.Capture(x.Exception.InnerException).Throw();
            }
        });
 }, CancellationToken.None, TaskCreationOptions.None, ActorTaskScheduler.Instance);
```

`await` binds to the current `SynchronizationContext` or `TaskScheduler` if present.

rogeralsing:
Next step would be to extend `RunTask` to accept an enum describing if reentrancy should be allowed or not. and add the appropriate code to stash incoming messages if reentrancy is not allowed.
much like your code @nvivo , but w/o using stashing, but rather something specific to this case.

nvivo:
@rogeralsing,

Sorry, I misunderstood your comment above that "tasks run like normal tasks by TPL, the continuations ... piped back to execute inside the actor mailbox loop". Anyway, I smelled something yesterday, and after sleeping over it I got what it is.

The proposed code misses the point of supporting the "async/await" syntax, which is what I was trying to implement.

**And I know I'm repeating myself A LOT, but something is not getting through.**

The problem is that this implementation forces tasks to be scheduled. And now it makes sense to get 50% drop in performance in your benchmark if you did them this way.

Code emulating async/await:

1. should NEVER start tasks
2. should NEVER use ContinueWith unless a background task is already running
3. should be IDENTICAL to not having tasks performance-wise to not using tasks if handling completed tasks

That means that:

```cs
void Receive(object message) {
    RunTask(async () => { });
}
```
and
```cs
void Receive(object message)
{}
```
MUST by any means execute in the same way (except for an some extra method call that do nothing).

This code achieves another thing, that is scheduling tasks in the actor context, and that may be a good idea, but it's just not what I was proposing. If that is just what you want, than it's fine.

But in order to support async/await, your untyped actor should implement something like:

```cs
protected void RunTask(Func<Task> action)
{
	var task = action();

	// if task is null, treat as synchronous execution
	if (task == null)
		return;

	// if task throws an exception, capture and rethrow
	if (task.IsFaulted)
		ExceptionDispatchInfo.Capture(task.Exception.InnerException).Throw();

	// if task is completed, return synchronously
	if (task.IsCompleted)
		return;
	
	SomeOtherInternalHelperToHandleTaskAwaiting(task);
}
``` 

And this is interesting, because I can't think of a way to implement `SomeOtherInternalHelperToHandleTaskAwaiting` that wouldn't make the entire thing pointless and would go back to just supporting async/await directly. Every implementation of this would be just another wrapper doing the same thing.

I'm really trying to come up with a solution that doesn't require Receive returning a Task, but if it doesn't you are forced to start tasks which will hurt the performance the way you see.

There may be other ways to implement this but I'm not seeing it. Any help anyone?

rogeralsing:
You must take a look at the ActorTaskScheduler and get what that is doing.
Discussing the other bits is pointless if that part is missing.

Im away for a week now, feel free to modify the implementation as long as all asyncawaitspecs pass.

src/core/Akka/Actor/UntypedActor.cs
```diff
@@ -13,6 +18,31 @@ protected sealed override bool Receive(object message)
...
 
+        protected void RunTask(Action action)
+        {
+            RunTask(() =>
+            {
+                action();
+                return Task.FromResult(0);
+            });
+        }
+        protected void RunTask(Func<Task> action)
```
nvivo: Thinking about it, there is a way to implement this that wouldn't be so bad:

```
protected void RunTask(Func<Task> action)
{
	var task = action();

	if (task == null)
		return;

	if (task.IsFaulted)
		ExceptionDispatchInfo.Capture(task.Exception.InnerException).Throw();

	if (task.IsCompleted)
		return;

	ActorTaskScheduler.SetCurrentState(Self, Context.Sender, null);
	Task.Factory.StartNew(async () => { await task; }, CancellationToken.None, TaskCreationOptions.None, ActorTaskScheduler.Instance);
}
```

This is just a little bit worse in performance than pure async/await, but not by much. And should retain the same principles.

src/core/Akka/Actor/UntypedActor.cs
```diff
@@ -13,6 +18,31 @@ protected sealed override bool Receive(object message)
...
+                await action().ContinueWith(x =>
+                {
+                    if (x.IsFaulted)
+                    {
+                        ExceptionDispatchInfo.Capture(x.Exception.InnerException).Throw();
+                    }
+                });
+            }, CancellationToken.None, TaskCreationOptions.None, ActorTaskScheduler.Instance);
+        }
+
```
nvivo: Thinking about it, there is a way to implement this that wouldn't be so bad:

```
protected void RunTask(Func<Task> action)
{
	var task = action();

	if (task == null)
		return;

	if (task.IsFaulted)
		ExceptionDispatchInfo.Capture(task.Exception.InnerException).Throw();

	if (task.IsCompleted)
		return;

	ActorTaskScheduler.SetCurrentState(Self, Context.Sender, null);
	Task.Factory.StartNew(async () => { await task; }, CancellationToken.None, TaskCreationOptions.None, ActorTaskScheduler.Instance);
}
```

This is just a little bit worse in performance than pure async/await, but not by much. And should retain the same principles.

nvivo:
Nevermind. This solution above won't work as expected. After a lot of debugging and adjustments to the code, I'm back to my previous statement that any attempt to solve this correctly will result in trying to supporting async/await directly.

The problem is that you can either start a background task as Roger did and force everything async, or you need to return a Task on RunTask to await in the correct order (which would mean the Receive method returning a task as well, which would mean proper task support). Attempting to await like I did will result in the final await being scheduled before the user's await, causing deadlocks in most cases. 

In this case, I'm not sure how much adding this to Akka would help, because the only behavior it adds is using a slightly nicer syntax but loses a lot in performance, which shouldn't happen with proper async/await support.

If there are alternative solutions, I'd be really happy to see it. For now, it seems stashing and flags will perform better.

rogeralsing:
You are still not taking reentrancy into consideration.
Your approach will only solve the non reentrant scenario.
In our current implementation we can spin up x tasks in paralell and still end up on the actor context to safely set actor state.
And the perfloss still refers to the overhead of logicalcallcontext, not the task scheduling.
You are also assuming the scheduler schedules to run on a new thread, we pipe messages back to the self mailbox, since we already are inside an ongoing mailbox run, that is, we append a message to the mailbox.
When the current receive ends, we will pick that message up and execute it directly. No new threads involved.
Only time we will end up on a new thread is when the task is actually async eg task.delay or other io completion port triggers.

Receive -> schedule task -> pipe pack to mailbox -> end receive -> pick up next message -> repeat

nvivo:
I won't try to convince you about this again because you are sure which solution works better.

I'm really trying my best to explain what is the difference between these solutions and how supporting this correctly would solve those problems better (both reentrant and non-reentrant). But every argument comes back as if I'm missing some very basic point and ignores everything I said.

It's your project, and I must say you have been doing an amazing job so far. But I can't argument anymore about this because either I cannot explain this well enough or you don't want to hear what I'm saying.

If someone brings this up in the future, maybe they'll have a better argument in favor of it.

rogeralsing:
The `ActorTaskScheduler` now runs tasks inline if possible.
Thus no overhead except for some GC pressure and a few jumps when working with tasks that are not IO bound.

Aaronontheweb:
So, what's the status on merging this? still for review only?

rogeralsing:
Somewhat little feedback on it, but it is stable, I have tried it plenty now.
Its good to go imo.

nvivo:
Just for the record,  I tested it and seems to work fine.  I'm not against it in any way,  it was just not the type of support I was hoping for. 

But it works and should be enough for most people.

rogeralsing:
@nvivo using the `ReciveActor` the experience is pretty frictionless, right?

```
Receive<Foo>(async f => {
       //any async code goes here.. no config, no runtask
});
``` 

We do need to support non-reentrant style also, but that will require some redesign on mailbox level.

nvivo:
Yes, it is. The receive actor has this benefit, for other actors the RunTask is kinda odd IMHO.

Quick question: since this implementation relies on messages for continuations, would stashing affect them?

rogeralsing:
Its a system message, so no, they are processed in a separate pipe.

FSM actors should also be possible to build the same way as receiveactor, TypedActor should be possible to solve using a new IHandleAsync<TMessage>,
That leaves UntypedActor, which could be solved using our built in patternmatch also.

nvivo:
Great.

I know that you have been working on this before but I feel my discussion about async/await somehow rushed this PR ahead of others. If that is the case, I don't have any urgency in having this merged, as there are workarounds available. 

I'd rather having reentrancy figured out, as i would be the primary reason to use tasks for me.

But if there are other reasons to merge it, go ahead.

rogeralsing:
We already support Suspend and Resume for user messages on mailbox level, this feature is however dedicated for supervision today.
My thought is to adapt this for more purposes.

e.g.

`mailbox.Suspend(SuspendMailbox.Supervision);`
and 
`mailbox.Suspend(SuspendMailbox.AsyncProcess);`

Pretty much replace the suspended bool in the mailbox with a bitmask.
That would solve the reentrancy stuff while having minimal impact on the framework and performance.

rogeralsing:
I've added some initial reentrant/suspend behavior

The behavior can be provided on the async receive method.
This makes explicit to the dev what behavior is used.

```csharp
public class MyAsyncActor : ReceiveActor
{
    public MyAsyncActor()
    {
          Receive<Foo>(AsyncBehavior.Suspend,async f => {
                 await Task.Delay(TimeSpan.FromSeconds(1));
                 //dostuff
          });
    }
}
```

Suspension is the default behavior. so any async receive will behave like the sync couterpart.
w/o blocking ofc.

The current naming of the behaviors are `AsyncBehavior.Reentrant` and `AsyncBehavior.Suspend` Im not sure what the best names would be, but I use those for now.

Suspension is provided by suspending the mailbox through the same mechanism as supervision does, there is now a enum bit flag to describe what has suspended the mailbox.

I've provded tests for both usecases.

cc @nvivo this does what you want, right?

nvivo:
The code *looks* like what I want to write, but works differently than what I'd expect. But it's enough to provide support for most cases.

We need more people providing feedback on this indeed.

rogeralsing:
Here is the thing, somewhere on the way to the receive there _has_ to be a `Task.Run` or `Task.Factory.StartNew` since the actor model is based on fire and forget messaging. that is, the sender of a message should never have to wait for a `Tell` to complete. 

So `Tell` is a void by definition.. so we need to go from a void Tell, to an async Receive.. and the only thing that exists between those two places is the mailbox. `Receive` _could_ be async in the UntypedActor, but then we add a lot of overhead in terms of object allocation and garbage collection, we are extremely optimized in this pipeline so any little change matters much, and letting Receive be async by default would be too pricy. 
e.g. changing Envelope from class to struct gave a perfgain about 50% just because we got predictable  GC for Receive... 

So just saying "Tasks add no overhead" does not cut it, it does matter under heavy load.   thus, we need to be optionally async, and the only place we can be that is _inside_ the receive method.

rogeralsing:
Also, "most cases", can you give a concrete example when it would not be enough?

nvivo:
I defend that most of these problems you are worried are related to the path you choose for the implementation, not about adding task support in a more standard way.

The way I see it, there is nothing unique to how Akka handles messages. ASP.NET is very similar, it's also single-threaded (per request), and once a request is handled, you may use async/await, it's never parallel, it may switch threads between calls and HttpContext is set back correctly every time in a transparent way, the user never has to call a custom method to make this work. ASP.NET supports both sync and async operations, and the performance hit is taken only if the user implementation requires that hit. I don't see why Akka couldn't use the same mechanism and semantics that people are already used to.

Previously I have wasted a lot of time trying to express this with examples, benchmarks, etc, and that didn't work out. I don't want to go the same path as it's very hard to discuss these deep technical things here.

I'd rather see other people reviewing this and see what they think of the API in general instead of just if the name of an argument makes sense.

rogeralsing:
You did not provide an example where the current implementation does not work.

You are also misunderstanding the actor model if you are comparing it with how asp.net handles things. the actor model have multiple producers pushing messages into a single consumer.

actor model
```
producer ->
producer ->
producer ->     mailbox -> actor
producer ->
producer ->
```

asp.net
```
request -> mvc/api action
```

The latter is also not fire and forget. that request will await the mvc/api action and then it is considered done.

That is in no way similar to the actor model where concurrent producers push messages into a shared queue for processing by the same cosumer.

That would be like having multiple requests entering the same instance of a controller and safely multiplexing so that the controller does not corrupt its state.
Is it this part that is confusing you?

>the user never has to call a custom method

What do you mean here?
There is no call to a custom method that the end user sees in the ReceiveActor either.
And ofc there is somewhere in the .net stack that the incoming network stream of bytes in the request is turned into spawning a task somewhere.
Be it in the asp.net stack and/or in the networking code, somewhere they are spinning up a task and applying the correct synchronization context (asp.net has one)

If we only stick to the ReceiveActor here.
What is it exactly that is bothering you and exactly when will it not work?

nvivo:
I'm not confused by anything, in fact I understand how these things work in more detail than I really want to. Its just that we are talking about completely unrelated things again.

Also, you are focusing too much on ReceiveActor while it is the least of my concerns. You said you wanted to implement this in a standard way for all actors, and ReceiveActor is more of the exception than the rule. We should focus on the semantics of ActorBase.Receive if this is to be implemented in a standard way for all actors.

I don't want to discuss this here anymore because it's exhausting. I'm open for a skype call some time if you really want to talk about these things in a more productive way. Otherwise this will just be another thread with 100 comments that nobody is willing to read.

HCanber:
I don't think any of this should have been merged into dev. For me, this is too experimental, and I don't like the api of it. I think we need to discuss this further before making it public.

All these async-PRs @rogeralsing  have made should go into a async-experimental branch and be reverted in dev. This will allow us to explore other alternatives, and other ways to achieve async-await and awaitable message handlers in other experimental-branches, for example as @nvivo says, by supporting it directly in `ActorBase.Receive`

For v1 we should just say, don't use async await unless you know exactly what you do.

HCanber:
@rogeralsing is sha 84c2909009b55 from Feb 13 the first commit on this subject, i.e. the one just before that is the one I need to go back to try out another async-await solution?

rogeralsing:
The async await support is only activated when you call either RunTask, or if you call `Receive<TMessage>(Func<object,Task> ` on the ReceiveActor.
Those are the only way that activate it.
So you can easily try something out side by side, or just remove those two methods if you still want to use the scheduler.

Some notes on the topic:
Supporting async await directly on Receive, receive needs to be altered to return Task (or an awaitable). that would be a major breaking change in itself.
It also comes with an extreme cost, just creating a completed task inside receive w/o doing anything with it comes with a 30% throughput penalty in the benchmark app, due to GC and allocation.
Actually handling it and returning it to the mailbox will increase that number.

Personally, I'm not willing to give up that much to support something that is a special case, not the default path.
Are you?

Also, the first commit was in Feb 13, why no feedback if you oppose the solution?`

What parts of the API is it that bothers you?
The RunTask alone? or the ReceiveActor API too?

HCanber:
>So you can easily try something out side by side, or just remove those two methods if you still want to use the scheduler.

Yeah, but there has been some changes to other things as well, like the mailboxes, dispatchers, uses of configure.await on Ask, and so on.

>That would be a major breaking change in itself.

If we were to modify ActorBase directly, then yes, but there might be other ways to introduce this.

>Personally, I'm not willing to give up that much to support something that is a special case, not the default path.
>Are you?

This isn't a case of "either you're with me or against me". I'm just looking for exploring other solutions as I think there exist one.

>Also, the first commit was in Feb 13, why no feedback if you oppose the solution?

The solution has been a moving target and the discussion has not always been so easy to follow, and quite frankly I was waiting for it to stop moving so much before reviewing it.
My original objection should have been written like "This was merged before any of us had a real chance at answering @rogeralsing s questions raised regarding the api, so it shouldn't have been merged. And now that I have reviewed it I think we should move it to an experimental branch".

>What parts of the API is it that bothers you?
>The RunTask alone? or the ReceiveActor API too?

Both. Not so sure if it's a good thing to be able to specify reentrancy on the Receive-method. It might be, but it certainly needs a bit more discussion.

nvivo:
@HCanber, @rogeralsing  I'm sorry about this discussion, I know this partly is my fault as I keep replying quickly every post.

When I posted the first discussion about this, I was starting to mess with the code to come up with a PR with some ideas on how these things could be solved. Unfortunately, I got bashed so hard just for proposing it that I just gave up and got carried away by the discussion.

So far, I was trying to focus on "what is needed" while the discussion moved quickly to "how it was already achieved in this implementation". I don't think the "what" was ever discussed so far, and this implementation tries to solve only part of the problem. Until now, I didn't say a word about "how" because it wouldn't make sense without the "what" and I feel nobody wants to hear anything from outside.

If you guys at Petabridge want to have full control over these discussions, that's fine for me. You are doing a great job so far and I was just trying to help the project.

rogeralsing:
@nvivo your PR only dealt with suspending reentrancy.
That part was great, but there are much much more to the issue at hand, which we have discussed.
The solution I ended up using does the same, but one level higher up. buffering user messages in the mailbox instead of buffering them inside the actor. so your solution lives on in a different way.

Shouldn't we instead try to get a clear picture of what everyone wants and what is actually possible.

e.g.
The main path is that message handlers are _not_ Task based. so we should optimize for that path IMO.
Task support is secondary to the main implementation, thus, Task support should be opt-in IMO.
Which it currently is. 
There are possibly other ways for opt-ing in, but that is my firm oppinion, that the default path should never have to pay the price of the "nice to have" path.

What are your oppinions on that?

HCanber:
I agree. If it comes down to which we need to optimize for, then we should optimize for handling synchronous messages.

nvivo:
@rogeralsing, that PR was just a hack to make it work in the current architecture, not a solution. I tried to discuss alternative solutions on #686 at a high level, but never went down to which objects to change.

About your question:

> The main path is that message handlers are not Task based. so we should optimize for that path IMO.

Agree.

> Task support is secondary to the main implementation, thus, Task support should be opt-in IMO.
Which it currently is. 

Agree. 

> There are possibly other ways for opt-ing in, but that is my firm oppinion, that the default path should never have to pay the price of the "nice to have" path. 

Agree.

> (<- and dont say its free, its not)

It's not free, but it doesn't need to be as expensive as you are saying if we change the implementation.

I'd like to focus more on "what the API should be" and later we can focus on GC, allocations, class x struct or even if any solution is possible.

If you want to talk about it here, or open another issue, or move to the chat, let me know.

rogeralsing:
Regarding the numbers.
You can try it out by adding `Task tmp = Task.FromResult(123);` inside receive of actorbase and run the benchmark app..
Throughput drops from 34 mil to 25 mil messages per second on my machine. 
And that has nothing to do with the current async await support,  that is just GC and allocation alone.

But yes, lets discuss the API;

IMO, we should be explicit on what sort of behavior the async flow would use.
As this is of much confusion for the Project Orleans users.

Also, we all agree that the support should be opt in.
Should it be opt in per actor or per message handler?

IMO, per message handler as you might want to use it for only a specific message and let the rest of the messages benefit from our optimized pipeline.

Thoughts?

nvivo:
First, can you share this benchmark code so we can have a baseline? 

What is the baseline handler doing? Because there is nothing expensive with Task.FromResult, it's just an [object allocation setting 2 fields](http://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/Task.cs,11a386e7d7cae64a). If you benchmark an empty method vs a method that allocates a simple string you'll get ~70% drop in throughput. I did a quick benchmark here, and while Task.FromResult is 3 times slower than doing nothing, it's 2x faster as allocating a StringBuilder. 

We need some representative code to measure the cost of this, even if it's an actor that is just works as in memory counter. Measuring a single object allocation 34 million times is not that helpful.

Makes sense?

nvivo:
About the API:

I'll focus on ActorBase support because it's the base class and once this works, everything else is just cosmetic on top of it. Also, I'm focusing only on "what" here, not "how".

I believe Akka should expose this support the same way most .NET frameworks do. 
In frameworks that support tasks (eg. winforms, asp.net, xunit) if you want an async handler, you just change your return from "void" to "Task" and it "just works". The behavior is that the framework awaits for that task to be completed (non-reentrant), and it's up to the user if that takes too long.

The reason I think this is important is because it's the same as other frameworks, so less learning curve for Akka, and it prevents the user to have access to a "RunTask" method which may create confusion and edge cases, like this:

```cs
void Receive(object message) {
    object result;

    RunTask(async() => {
          result = await DoSomething(message);
    });

    Sender.Tell(result);
}
```

This is an uncommon edge case, but is possible. No framework exposes methods like that just to allow async/await for this reason. Either you implement a sync or an async method. So, I think ActorBase should *somehow* support:

```cs
bool Receive(object message) { }
// or
Task<bool> Receive(object message) { }
```

It's important to note that reentrancy/synchronicity is something you decide by the code you write inside the handler, not by the handler you choose. That means adding a Task-enabled handler doesn't mean non-reentrant neither asynchronous. It just means you will handle a task and Akka will await for it. That may mean async+non-reentrant, but that's up to the runtime to decide.

If the developer screws his logic with async, it's his fault not Akka's. Akka just follows the same convention as everybody else.

In short, we don't need "task-running support", we need "task-awaiting support". And for that, all we should care is that the user handles us a task, we don't really care where it was created.

What do you think?

HCanber:
I don't think we should support reentrant actors at all. If the user wants that they can kick off operations in a temporary (inline) actor or start a task.
We should never abandon "An actor processes one message at a time".

I think we can achieve what's needed if we change `Receive` to:
``` C#
Task Receive(object message)
```

If `null` is returned, it's considered to be a synchronous handler (and no `Task.FromResult` is needed). If a `Task` is returned the mailbox waits for it to complete before continuing. 

We don't really need the `bool`. It's currently only used when logging (if `akka.actor.debug.receive==true` and the actor implements `ILogReceive`) to print out "received handled message xyz" or "received unhandled message xyz". I think we can live without that.

That's how an actor, from the framework perspective, should work. Then we can have abstract classes which slightly modifies the `Receive` syntax, like the `ActorBase` we have today, that has a `bool Receive(object message)` that does not support async-await at all.

(On the reentrant-actor note, to soften what I stated first, a reentrant actor could be created by inheriting from an actor base class in which the receive always returns null)

nvivo:
> I don't think we should support reentrant actors at all. If the user wants that they can kick off operations in a temporary (inline) actor or start a task.
We should never abandon "An actor processes one message at a time".

That was *exactly* my point, just in case my text was confusing. =)

rogeralsing:
> "An actor processes one message at a time".

Apparently that is not a requirement of the actor model, its how it works "on a conceptual level" according to Carl Hewitt himself (see channel9 interview)
Actors are allowed to optimize for parallelism if the next message behavior do not need to change.

(So our routers can still by definition be actors , which I thought they weren't)

Also, the reentrant support does not process real messages in parallel, it still does one at a time, but the await/complete flow of tasks is chopped up into one message each.

That beeing said, I would be totally OK with removing the reentrant support as it is sort of confusing.
It would probably even allow us to resolve tasks w/o messagepassing, as the actors Context will not mutate during the task execution, no new senders, no new messages.

And @nvivo I totally didn't get that from our previous discussions, I even recall that you wanted reentrant as default as it would be the most performant (?)

Either way, that would probably be a completely viable path forward.

nvivo:
> I even recall that you wanted reentrant as default as it would be the most performant

I know I write *a lot* and I may write confusing things sometimes, I have many flaws. But there was a feeling (at least I felt it very strongly) that "this new guy has no idea how Akka works and comes here trying to tell us how to do our jobs, let's educate him instead". That makes it really difficult to say anything.

This is a very complex topic, and I don't know how to discuss it in 140 chars =). I don't remember where I said that, but there must be more context involved. I'd prefer to avoid pointing to that heated discussion and try to talk about the solution now that we are almost in the same page if you're ok with that.

There are some differences between what @HCanber is proposing and how this was implemented. Specifically what `RunTask` implies vs what `await` does, and just removing the reentrancy parameter won't solve that, if that is what you're proposing.

> "An actor processes one message at a time".

The idea here is not that this is a strict rule of the actor model, but more that in Akka actors the "Receive" method starts and finishes, and only then a new message is processed. PipeTo doesn't actually break this behavior, but making a reentrant option with tasks would.

IMO, for the user there shouldn't be a different behavior between these two methods:

```cs
void Receive(object msg) {
     var value = GetValueFromDatabase(msg);
     // you don't expect another message to be processed here
     Sender.Tell(value);
}

async Task Receive(object msg) {
     var value = await GetValueFromDatabaseAsync(msg);
     // you shouldn't expect another message to be processed here
     Sender.Tell(value);
}
```

But there is a different behavior between these: 
```cs
void Receive(object msg) {
     var value = GetValueFromDatabase(msg);
     Sender.Tell(value);
}

void Receive(object msg) {
    RunTask(async () => {
         var value = await GetValueFromDatabaseAsync(msg);
         // if reentrancy is supported, another message can be processed here
         Sender.Tell(value);
    });
    // anything after here executes before the previous code starts
}
```

If the user wants reentrancy, PipeTo/Self.Tell is the best option. If during the runtime `GetValueFromDatabaseAsync` is synchronous, it will execute as fast as if no tasks were there with no changes or optimizations in Akka. This allow scenarios not possible today.

Does it make sense why I believe providing a task returning method on ActorBase would be better?

It seems you don't like or don't agree with this API for some reason. What problems do you see with it?

rogeralsing:
@nvivo I thnik with @HCanber approach, we can still have one method that supports both cases.
There is a natural "start to code here" entry point.

Your approach if I get you right, relies on duck typing, that is, if there is a `void Receive` then go sync, if there is a `Task Receive` then go async, right?

If you have to chose either or, then that actor can only support either or, you can not mix and use sync for some messages and Tasks for some messages, is that a correct understanding of your approach?

nvivo:
@rogeralsing  As far as I can tell, me and @HCanber are proposing exactly the same thing. I didn't specify any implementation, just the idea so far. Is this why I'm confusing everybody here?

I just wanted to discuss the idea first and why it would be better. I may propose an actual implementation path if you're ok with the idea.

For reference, the implementation starts with what @HCanber said, exposing a single method that returns a task, but optimizing by returning null on sync actors. There are some other changes that would require more discussion then.

rogeralsing:
My impression was that you suggested to have support like xUnit, you either specify a void or a Task method..
While I interpreted @HCanber's solution to be to just change bool to Task on ActorBase.

nvivo:
No! =) When I talked about xunit/asp.net, etc, I was just discussing the differences between what those frameworks expose and Akka having to call `RunTask` inside a void method.

Anyway... Talking about the actual implementation, this would require other changes to run correctly and without overhead. That could cause the code to start diverging from the JVM implementation, but should provide the same results.

This is the part where I started trying to change the code, and conceptually this should work, but requires some internal knowledge to adapt correctly and I may be wrong about some choices, so **please be patient with me while reading this**:

1. In order to allow "await" correctly inside actors, the ActorBase.Receive method must be running already inside the ActorTaskScheduler - this allows for the optimization mentioned before: if the returned task is null, treat it as if it was a completed task and go to the next message - *this alone should cause the performance to be the same as it is today with no extra allocations* if the actor is synchronous. 

2. It seems to me that the easiest path to achieve this is to schedule `Mailbox.Run` as a task inside the ActorTaskScheduler instead of the Dispatcher - this seems to require some changes to the implementations to make the ActorTaskScheduler a more active role in the message processing, instead of just a helper.

3. `CallContext.LogicalGetData` needs to be replaced by a custom `SynchronizationContext`, as .NET automatically captures and passes that back to continuations, so there is no need for the RunTask logic to keep the state anymore.

4. With continuations being automatically posted to the current SynchronizationContext instead of passed as a system messages, there are no changes to envelopes  - not sure if it's clear here, but there is no need to pause the mailbox, as far as it's concerned it will just wait for Receive to complete, possibly even using "await" on the task - this can be optimized to add no overhead to synchronous processing as well.

This is no simple change, but would be more "standard" in comparison to other frameworks.

As for the API changes, all current sync actors are kept the same but returns `null` to ActorBase. ReceiveActor and FSM would have a Task version of `Receive/When` (as you did already in ReceiveActor), but that just returns the task back to ActorBase.

UntypedActor requires some though. It could either implement 2 virtual methods instead of 1 abstract method, or it could keep the same API and have a new `AsyncUntypedActor` version to implement only the async handler (I like this idea better).

The `UntypedActor` solution is particularly important, because it gives an example of how to solve this problem for `PersistentActor` for example, that have the similar API.

@HCanber, are these changes somewhere along the lines of what you were thinking?

@rogeralsing, makes sense? Comments?

rogeralsing:
Regarding 3.
The `CallContext` is _also_ passed back and captured, that is why we use it.
The only reason we have the message based stuff is for reentrancy. 
If reentrancy is removed, CallContext also flow in async transitions, just like SyncContext.
We can not rely on synchronizationcontext, as the synchronizationcontext should be set by the thread owner. e.g. xUnit or Asp.NET, we should never overwrite those.
So if we want to execute stuff on the asp.net threadpool, thats a no go.

As Akka.NET can run on different dispatchers, we should never assume that we are allowed to overwrite the current SynchronizationContext.

Stephen Cleary has some post about that, I will see if I can dig it out.
But in short, if you own the thread, use `SynchronizationContext`, if you run on a leased thread, use `TaskScheduler`


4. 
If `Receive` is a `Task`, and we call `Receive` from the mailbox.
How do you envision the mailbox to wait for the task? 
Should it just exit the mailbox Run after setting up the task execution and then apply the correct mailbox status on task completion?

That _might_ work, not sure on what effects that would have in general. 
e.g. Should system messages be suspended during long running tasks also?
What if the actors parent decide to try to terminate children and the async actor is waiting for a Task.Delay(2 hours).. 
Imo, System message processing should be run in parallel to async tasks, or we might get actors that appear as zombies to the rest of the system.
This is ofc true for the sync stuff too, but I think that it might be that users beleive that they can have message processing that runs indefinetly due to async..  

That is at least an important aspect to think about.

And regarding the actor API, I'd very much like to keep the number of different actor types to a minimum, we already have too many :-)

rogeralsing:
If we only support non-reentrant tasks, it couldl possibly be viable to use this for scheduling instead of having our own scheduler at all:
https://msdn.microsoft.com/en-us/library/system.threading.tasks.concurrentexclusiveschedulerpair%28v=vs.110%29.aspx

the only thing we would need to do then, is to replace the `[ThreadStatic]` Context and store it in the `LogicalCallContext` instead, as that does flow through async actions.

That would get rid of all the magic stuff.

HCanber:
>Should it just exit the mailbox Run after setting up the task execution and then apply the correct mailbox status on task completion?

Pretty much.

> e.g. Should system messages be suspended during long running tasks also?

Yes, the semantics should not change if you wait or await a task inside the handler, so the semantics should be exactly the same as if it was a long-running-synchronous message handler.

rogeralsing:
If we stick with the same semantics, we need to be super clear to end users that long running handlers could cause weird supervision behaviors.

All of this would be super easy to solve. 
ThreadStatic -> CallContext.

It wont even affect speed as the call context only needs to be set once per mailboxrun, like the ThreadStatic today.
The perf penalty earlier was due to setting it for every message to support reentrancy.


Skickat från min iPhone

HCanber:
>If we stick with the same semantics, we need to be super clear to end users that long running handlers could cause weird supervision behaviors.

In what way?

rogeralsing:
Let's say a supervisor tries to restart all of it children, one child is not responding to sys messages. then that branch will not be in a restarted known to work state for however long the async actor is running.

And this is ofcourse exactly the same as if there is a long running sync actor, _but_ as our userbase knows that blocking is bad, most of them will know not to block actors.
But in the async case, many might think that it would be ok to spinn up a very long running process.

It might not be a problem at all in the end, but if I saw an async API, I would assume I could do a long running flow in there.

HCanber:
Isn't this expected in an asynchronous system? What you're describing is exactly what can happen if an actor is deployed on another server, and the supervisor restarts it. There are no guarantees that the children are restarted directly after the supervisor has said them to. 
And the alternative is worse, that the supervisor would interrupt and possibly restart an actor in the middle of it processing a user message.

nvivo:
I agree that the semantics of async handlers should be the same of the sync ones. In general we should assume async handlers will take as much CPU time as sync ones + some time for context switches.

Task support in Akka is intended to provide support for the async API in .NET, things like `DbCommand.ExecuteNonQueryAsync` or `Stream.FlushAsync`, not to create long running tasks. We can make it clear in the documentation.

>  if I saw an async API, I would assume I could do a long running flow in there.

IMO, I think this is more in your head. I don't have this impression when I see an async API. Or at least when I say "long-running", I usually mean of ~100ms (time to do the regular work + some context switches for continuations), maybe a second, *never* minutes or hours.

> How do you envision the mailbox to wait for the task? 
Should it just exit the mailbox Run after setting up the task execution and then apply the correct mailbox status on task completion?

If the `Mailbox.Run` method is scheduled in the task scheduler, it can return a task, so you could just turn it async and await only if needed. something like:

```cs
var task = dispatcher.Dispatch(ActorCell, envelope);

if (task != null)
    await task;
```

This would prevent avoid small overhead for sync methods, and cause the same semantics as if the task was not there.

rogeralsing:
ah, yes that would be the same. 
That sounds fair to me.

So, what about tasks that are not awaited, if someone by mistake or trying to be smart starts tasks that are not awaited, should the devs be on their own or should the framework try to prevent parallelism?

rogeralsing:
@nvivo that would assume that the mailbox run is also async.
So the `Run` would have to be scheduled also (I think you already said this, just want to make sure I get your vision on it)

nvivo:
> So, what about tasks that are not awaited

What do you mean?

> So the Run would have to be scheduled also

This is a part I probably need more research or change the implementation of the scheduler to be more active as I said before. but today you have:

```cs
protected override void Schedule()
{
    //only schedule if we idle
    if (Interlocked.Exchange(ref status, MailboxStatus.Busy) == MailboxStatus.Idle)
    {
        dispatcher.Schedule(Run);
    }
}
```

The scheduling line would be changed to something like:

```cs
actorTaskScheduler.Schedule(Run);
```

This remains fire and forget. Makes sense?

rogeralsing:
>What do you mean?

Lets say someone calls some code like this

```csharp
protected override async Task Receive(object message)
{
....
                 Task.Run(() => this.SomeState = 123);
                 //or
                 SomeAsyncMethod(); //user forgot to use await
....                   
              
}
```

Should that be safe, that is, the framework coordinates scheduling of that task to run within the actor concurrency constraint.
Or
Should we just let that run async w/o any safety net and possibly result in corrupt state.

rogeralsing:
> Makes sense?

Sort of, but we should not assume that the tasks nor the mailbox should run on the standard thread pool.
We need to keep the support to run anywhere. 
This can most likely be solved by letting each type of dispatcher have its own instance of the TaskScheduler and decide how/where  the scheduler schedules the task execution.

nvivo:
> Should that be safe, that is, the framework coordinates scheduling of that task to run within the actor concurrency constraint.

From what I understand, when you do just Task.Run, it uses the DefaultScheduler, not the current. That means that tasks created anywhere need to be considered unsafe and outside of Akka control - this is still the case of RunTask currently, the `action` in `RunTask(action)` runs inside akka, but akka cannot control where the user or .net creates tasks inside of that - that's why I always said RunTask doesn't actually provide all this safety.

So, IMO akka shouldn't care about that because there is no way to ensure every task always goes to your scheduler. Just like in a WPF app by default it uses the UI dispatcher, but you can force using something else and crash your app.

>  but we should not assume that the tasks nor the mailbox should run on the standard thread pool.

Yes, I agree. When I write these things, I'm talking more of the concept than the usage with the current code.

The idea is that ActorTaskScheduler and the Dispatcher would be somewhat interchangeable. If you schedule something in that actor task scheduler instance, you are guaranteed to be running that code within Akka in the right context. I know today this is not exactly what happens, as it depends on which actor is invoking the scheduler. In my proposal, the scheduler instance is tied to the dispatcher.

As you said, maybe each dispatcher have its own scheduler. I don't know what are the effects of this  though.

rogeralsing:
Some heads up.
I've done some testing using `CallContext` instead of `ThreadStatic`
This works, but_

1. Big perf hit. 35 mil to 20 mil msg/sec
2. xUnit goes completely bonkers and blows up... (example apps work, but something in xUnit does not like this)

implicit context :(

nvivo:
> It wont even affect speed as the call context only needs to be set once per mailboxrun
...
> Big perf hit. 35 mil to 20 mil msg/sec, I suspect reading callcontext is expensive

@rogeralsing, do you mean that calling it only once before running through the mailbox causes that?

Isn't this due to benchmarks producing messages at the same rate they are being consumed? If benchmarks run like the DedicatedThreadPool, you may actually calling CallContext almost at every message as the mailbox never has time to fill up and is always restarting the loop.

rogeralsing:
The Benchmark does an explicit pre fill up before starting for that very reason.

The issue here is that we use the implicit actor context all over the place, e.g. to find current Sender.
The implicit context is currently thread static, which it can't be if we want to remove the magic in the old ActorTaskScheduler, to remove that magi we have to replace the thread static with CallContext.

This work, but it is much much more expensive both reading and writing the CallContext state than normal ThreadSatic state.

We could ofcourse use some hybrid of the current scheduler (not the runtask, just the scheduler) and remove all the message passing stuff in it, but keep the parts that marshalls the thread static info into the call context.
That would solve that problem (and possibly the xUnit breaking issue)

But it would have been nice to just have CallContext as we might possibly be able to not use a scheduler at all.. if CallContext flows the actor context, and we only support awaited tasks, then the other stuff is not needed.

tstojecki:
@rogeralsing
What is the outcome here? Lots of back and forth in those two threads, which is a good discussion in and of itself, but no clear indication on what the outcome is?

Is pipeTo still the recommended way of working with async/await? Do you plan on integrating @nvivo  suggestions, if so do you have a roadmap you can share?

If you have made any updates already, can you reference them here? 

Thanks!

rogeralsing:
@tstojecki PipeTo is absolutely the recommended way of doing async operations from within an actor.
The async await support for ReceiveActor does work and no known bugs.
(the actual implementation is not optimal but is a result of the constraints of the JVM Akka API)

We will most likely not push the async await support any further but stick to PipeTo until Akka.Typed is ported, which solves all of the issues we faced in the current API.

nvivo:
@tstojecki For your knowledge (and Roger's despair :-D) I just released an updated version of my async actors that I'm using in my projects, If you're interested.

https://github.com/nvivo/akka-async-actors
https://www.nuget.org/packages/Akka.Contrib.AsyncActors/

tstojecki:
Thanks for the update @rogeralsing and thanks for pointing out the signalR 'edge case'.

@nvivo 
I will be sure to check it out. Props to you for this way of solving differences in opinion :)

colin-jack:
Great thread, lots of useful information to someone new to Akka.NET like myself, even the final point about calling SignalR was interesting.

Anyway although the thread is full of great information some of it is quite difficult to follow so I wondered if there's any plans to write any of this up?

Not even suggesting a summary, from my perspective this is a topic where more detail is better than less.

vchekan:
Hmm, is there any summary what are rules and limitations of current async implementation?
Please correct/add below observations.

1. Actor is blocked while executing async call in Receive
2. Actor reaction to Kill, while in async is ???
3. Task, invoked from within Receive, without explicit scheduler, is executed on default task scheduler and not on the Actor's dispatcher scheduler.
4. It (is|is not)??? safe to access Actor's members from within code executed by task from Receive.
5. Code after "await" is executed in usual Actor's thread and thus is safe to access Actor's members.
6. Because of (1), attempt to Self.Ask, from within a task will cause a deadlock.

nvivo:
@vchekan,

A lot have been said, but **the important thing is that the current setup works fine, its just not optimal**. It's easy to make it optimal though without any changes to the current architecture, even though I admit I failed hard in leaving that clear.

I plan to make a small PR to add some better support for async/await to all actors without any performance hit and without changing anything in the current architecture, and ditching the async actors I created.

About your points, from my understanding:

1. Yes, the behavior of a `Receive(async () => {...})` is exactly the same as `Receive(() => ...)`, that is, no messages are processed until the method finishes executing (and the task is completed). It blocks as it would from a synchronous method.

2. The the actor will be killed during await, but the task you started will still finish fine, because even though the actor stopped or restarted, the actor instance (the class itself, not the actor ref) will be there (because .net will keep the reference), and you will be able to access members just fine. If you do a `Self.Tell(msg)`, the new instance should get the message or it will go to deadletters if the actor is terminated.

3. (see 5)
4. (see 5)
5. The "safety" of akka is about the fact that there is only one thread changing the state of the actor at any point in time. There is no issue in starting tasks in the default scheduler, as almost any code awaited inside a `Receive<T>(async...)` will be running outside of akka scope anyway. But for any code you create that touch actor members, **as long as you start tasks that access the members one at a time, and await for it inside the actor, everything is safe**. (see the example below).

6. Yes, it will deadlock, but It doesn't make any sense to do `Self.Ask`, as you already have access to any state you need.

Example:

```cs
private int _state;

Receive<Foo>(async => {
     // safe
    _state++;

     await Task.Run(async () => {
          //safe
          _state++;

          await Task.Run(async() => {
              //safe
              _state++;
         });
     });

     // safe
    _state++;
});
```

But if you create code that starts concurrent tasks, it's on you:
```cs
Receive<Foo>(async => {
     // safe
    _state++;

     // unsafe - your code is awaiting, but internally 2 threads may change the state at the same time
     await Task.WhenAll(new [] {
         Task.Run(async() => _state++);
         Task.Run(async() => _state++);
     });

    // safe again
    _state++;
});
```

Now if your member or code was thread safe (ex, using ConcurrentBag or Interlocked.Increment) it would be safe to do whatever you want, but it wouldn't make sense to use Akka that way.

vchekan:
@nvivo yes, I have read both long threads, your initial proposal and this one. I have some thoughts, but want to understand current implementation.

In your example, I'm still not convinced that one task example is thread-safe.

Receive<Foo>(async => {
     // (#1) Thread - 1: 0+1=1
    _state++;

     await Task.Run(async () => {
          // (#2)  Thread - 2: 1+1 = 2
          _state++;
     });

     // (#3) Thread - 1: (2 OR 1)+1!!!?
    _state++;

My source of concerns is L1 cache (cache per-core). I have not tested, but is it possible that Thread-1 (the one handling actor's mailbox), would cache the result of first increment (0+1=1), and in (#3), instead of fetching the value 2 from memory, would find cached "1" in L1 cache?

nvivo:
@vchekan I don't think you have this kind of control in managed code. The runtime will put barriers everywhere to make it work as expected. Take a look at https://msdn.microsoft.com/magazine/jj863136.

I know there are edge cases where you may find bugs, but this is not something you should have to care. The threadpool it will take care of these things for you. That's why it's C#, not C.

reijerh:
Just wanted to say that as a new Akka user I'm very happy that I can use an async handler in Receive(). Tasks are unavoidable in modern C#.

I don't understand why @nvivo's simple solution he published here: https://github.com/nvivo/akka-async-actors isn't used though. It seems to me there aren't any real downsides to it. Consider this my +1 vote (I don't know of a better place to express my support).

JeffCyr:
@reijerh Do you have any concerns with the current implementation? It is conceptually similar to nvivo's solution.

His solution is a bit more lightweight, but the issue is that the MessageDispatcher of the actor is not respected when an async receive continue after an await, this is why it can't be used as a general solution.

Also, async receives should not be used when high message frequency is expected. The actor is not handling messaged when an async receive is in progress, so you can end up with OutOfMemoryException.

Since async receives should not be used with high frequency, the cost of using a TaskScheduler per actor is insignificant.

nvivo:
@JeffCyr a quick comment here:

> async receives should not be used when high message frequency is expected.
..
> Since async receives should not be used with high frequency, the cost of using a TaskScheduler per actor is insignificant.

Both statements are actually false. The reason why it's slower today is only because original akka implementation is synchronous (because JVM doesn't support the concept of tasks yet), and implementing an async api on top of a synchronous api requires a lot of hacks to make it work. If the core of akka took tasks into account, this would be virtually free.

This would of course require changes to the APIs and diverge from the JVM implementation in favor of a better and more scalable API, but it's perfectly possible.

I'm gave up on trying to change people's mind about this, but I hate these wrong ideas are perpetuated.

JeffCyr:
@nvivo I'm not sure I follow you on this. Are you talking about the fact that the async receives are not reentrant? The actor is suspended while an async receive is in progress in your implementation too.

nvivo:
@JeffCyr, Sorry, I'm not going into all of this again. =) I had enough discussions about this for a year.

I already proved that it's possible to support tasks with zero costs at the cost of API changes. The reason this is not implemented correctly is political, not technical.

Horusiath:
@nvivo @JeffCyr @reijerh fully integrated async support is something, that will be probably implemented as part of major change set in Akka.NET (currently it's proposed in 2.0). Until then PipeTo and async Receive are preferred ways to work with tasks.

nvivo:
@Horusiath I have faith that day will come. =) Good to know that's part of the plan.

Is this being tracked somewhere? any issues or wiki for the 2.0 plans?

Aaronontheweb:
https://docs.google.com/document/d/1r-ao3JDL92pha2aXJI6fqpMK18lPeWX1ynNdtIAjGBM/edit?ts=56a9f9a9

Until then, locking the thread - no more of this async await flamewar forever please.

PR #702, opened 2015-03-02
Added support to the F# API for actor objects

jorgef:
Added spawnObj and spawnObjOpt to support creating OO actors from F#.

Usage:

```fsharp
spawnObj mailbox.Context "tailActor" (<@ (fun () -> new TailActor(reporter, filePath)) @>)
spawnObjOpt mailbox.Context "tailActor" (<@ (fun () -> new TailActor(reporter, filePath)) @> ) [SpawnOption.SupervisorStrategy(SupervisorStrategy.DefaultStrategy)]
```

This is required to create actors that are able to handle methods like PreStart() and PostStop(). Without this there is no easy way to release disposable resources at the moment.


rogeralsing:
This has some overlap with this PR 
https://github.com/akkadotnet/akka.net/pull/682

Horusiath:
@jorgef I think, I found another solution for problem of releasing `IDisposable` resources with #682 (new Defer method for Actor<'t> interface). Your commit however will greatly simplify life for F#-ers, who for some reasons want to use standard Actor API for their needs.

I'll review it soon.

src/core/Akka.FSharp/FsApi.fs
```diff
@@ -424,6 +429,31 @@ module Spawn =
...
 
     /// <summary>
+    /// Spawns an actor using specified actor quotation, with custom spawn option settings.
+    /// The actor can only be used locally. 
+    /// </summary>
+    /// <param name="actorFactory">Either actor system or parent actor</param>
+    /// <param name="name">Name of spawned child actor</param>
+    /// <param name="f">Used to create a new instance of the actor</param>
+    /// <param name="options">List of options used to configure actor creation</param>
+    let spawnObjOpt<'Actor when 'Actor :> ActorBase> (actorFactory : ActorRefFactory) (name : string) (f : Quotations.Expr<(unit -> 'Actor)>)
```
Horusiath: In this case `<'Actor when 'Actor :> ActorBase>` is unnecessary. Equivalent function signature is:
`let spawnObjOpt (actorFactory : ActorRefFactory) (name : string) (f : Quotations.Expr<(unit -> #ActorBase)>)`

jorgef:
@Horusiath I will have a look to #682 and try to use it whenever I can.
Thanks for the review, you are right, I just improved both signatures.

Issue #633, opened 2015-02-16
Fix interface naming conventions before 1.0?


nvivo:
Are there any plans to standardize interface names before 1.0?

It feels strange to use interfaces without an `I` prefix in .NET. I think it would help the project to have the port using .NET conventions. Just as properties from java are converted from `getContext`  to `Context`, an `LoggingAdapter` feels more natural in .NET as `ILoggingAdapter`.

I saw this was quickly discussed before [here](https://github.com/akkadotnet/akka.net/issues/109#issuecomment-39038065) but was probably forgotten.

Scanning the project, I found the following interfaces that don't use the .NET naming conventions.

* `Akka.Actor.ActorRefProvider, Akka` (Public)
* `Akka.Actor.ActorRefScope, Akka` (Public)
* `Akka.Actor.AutoReceivedMessage, Akka` (Public)
* `Akka.Actor.Cell, Akka` (Public)
* `Akka.Actor.Inboxable, Akka` (Public)
* `Akka.Actor.IndirectActorProducer, Akka` (Public)
* `Akka.Actor.Internal.ChildrenContainer, Akka` (Public)
* `Akka.Actor.Internal.ChildStats, Akka` (Public)
* `Akka.Actor.Internal.InternalSupportsTestFSMRef`2, Akka` (Public)
* `Akka.Actor.Internal.SuspendReason+WaitingForChildren, Akka`
* `Akka.Actor.Internals.InitializableActor, Akka` (Public)
* `Akka.Actor.LocalRef, Akka`
* `Akka.Actor.LoggingFSM, Akka` (Public)
* `Akka.Actor.NoSerializationVerificationNeeded, Akka` (Public)
* `Akka.Actor.PossiblyHarmful, Akka` (Public)
* `Akka.Actor.RepointableRef, Akka` (Public)
* `Akka.Actor.WithBoundedStash, Akka` (Public)
* `Akka.Actor.WithUnboundedStash, Akka` (Public)
* `Akka.Dispatch.BlockingMessageQueueSemantics, Akka` (Public)
* `Akka.Dispatch.BoundedDequeBasedMessageQueueSemantics, Akka` (Public)
* `Akka.Dispatch.BoundedMessageQueueSemantics, Akka` (Public)
* `Akka.Dispatch.DequeBasedMailbox, Akka` (Public)
* `Akka.Dispatch.DequeBasedMessageQueueSemantics, Akka` (Public)
[...]

PR #707, opened 2015-03-03
ConsistentHashRouter rewrite with Virtual Nodes, Cluster support for pool routers, and ActorRefProvider Updates

### LINKED ISSUE #550: Need to develop way to iterate through Fallback Deploy, RouterConfig, etc during deployment
Right now `Deploy`, `RouterConfig`, and other pieces of configuration plumbing have a `WithFallback` function just like their partners in canonical Akka, but we aren't using the fallback values anywhere in `Deployer`s, `ActorRefProvider`s, et al.

We need to do the following:

1. Define some reliable __merge__ behavior for fallbacks, just like how we do with `Config`.
1. Invoke the merges at the time of deployment, serialization, actor ref creation, etc...

Scala Akka has code like this (taken from their `RemoteActorRefProvider`) that handles the merging through pattern matching and aggregation:

````
Iterator(props.deploy) ++ deployment.iterator reduce ((a, b) ⇒ b withFallback a) match {
        case d @ Deploy(_, _, _, RemoteScope(addr), _, _) ⇒
````

See: https://github.com/akka/akka/blob/cb05725c1ec8a09e9bfd57dd093911dd41c7b288/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala#L273

We should do likewise, as it will help make it easier to do remote and local deployments through configuration.

### LINKED ISSUE #673: Consistent hash routing and resizable pools, potential bug
I don't think we handle resizable pools and consistent hashrouting correctly.
here is my reasoning.

Lets say the pool consists of two workers.
We send a message, say a SomeCustomerCommand with CustomerId (the hash) = 1
This message ends up on say worker 2.
Worker 2 starts processing the message.

If the pool resizes  now at this point.
Another message with the same CustomerId (again, the hash in this case)
This message may now end up on another worker since the the number of workers have now changed and the affinity between hash and worker no longer holds.
So this time, the message might end up on worker 3.
So potentially, worker 2 and worker 3 are now processing a message each at the same time, that really should have been handled by the same worker.

Thoughts?

### PR DESCRIPTION
**Pull request by Aaronontheweb**
From the original commit message

---
close #550
helps resolve #673

Reimplemented Murmur3, ConsistentHashRouting to use virtual nodes
Added deploy and routerconfig fallback support
Rewrote ActorOf method to LocalActorRefProvider to match Akka
Rewrote ActorOf method to RemoteActorRefProvider to match Akka
Breaking change - renamed ConsistentHashable interface to IConsistentHashable (per #633)
Added MultiNodeTests for ClusterConsistentHashRouting
Implemented Pool routers for Akka.Cluster

Updates
---
This is a major update of three systems:

1. *Props, Deploy, RouterConfig, and ActorRefProvider* - this code now matches what's inside Akka. Deploy and RouterConfig fallbacks now work as expected. This was necessary in order to support `ClusterPoolRouters` in combination with `ConsistentHashingPool`, which relies on a combination of configured routing behavior and programmatically defined routing behavior (`ConsistentHashMapping`.)
2. *ConsistentHashRouter, MurmurHash, and ConsistentHashMapping* - the entire infrastructure for consistent hash routing has been updated to support virtual nodes and the latest Akka API. More on this in a moment.
3. *ClusterGroupRouter and ClusterPoolRouter* - pool routing is now supported in Akka.Cluster and is covered by multi-node tests, which pass.

### ConsistentHashRouting

The APIs are largely the same, but with two major exceptions:
#### `ConsistentHashMapping` Delegate
There are three ways to instruct a router to hash a message:
1. Wrap the message in a `ConsistentHashableEnvelope`;
2. Implement the `IConsistentHashable` interface on your message types; or
3. Or, write a `ConsistentHashMapper` delegate and pass it to a `ConsistentHashingGroup` or a `ConsistentHashingPool` programmatically at create time.

Here's an example, taken from the `ConsistentHashSpecs`:

```csharp
ConsistentHashMapping hashMapping = msg =>
{
    if (msg is Msg2)
    {
        var m2 = msg as Msg2;
        return m2.Key;
    }

    return null;
};
var router2 =
    Sys.ActorOf(new ConsistentHashingPool(1, null, null, null, hashMapping: hashMapping)
    .Props(Props.Create<Echo>()), "router2");
```

Alternatively, you don't have to pass the `ConsistentHashMapping` into the constructor - you can use the `WithHashMapping` fluent interface built on top of both `ConsistentHashingGroup` and `ConsistentHashingPool`:

```csharp
var router2 =
    Sys.ActorOf(new ConsistentHashingPool(1).WithHashMapping(hashMapping)
    .Props(Props.Create<Echo>()), "router2");
```

### `ActorOf` changes

Now you might be wondering, what the hell does `ConsistentHashRouting` have to do with `ActorRefProvider`? The answer is - in order to support `ConsistentHashMapper`, there are many situations where half of the router will be defined in configuration and the other half programmatically.

In order to support this, *especially* with ClusterPoolRouter wrapping a `ConsistentHashingPool`, we needed to follow the Akka conventions to have Deploy and RouterConfig properly fall back.

I had to shave a couple of large, hairy, and kind of angry yaks in order to do that but it's done and all tests pass. Both the `RemoteActorRefProvider` and `LocalActorRefProvider`'s ActorOf method matches what's in Akka now.

### Breaking Changes
Per #633, `ConsistentHashable` interface has been renamed to `IConsistentHashable` - this will break any existing user-defined message types which implement the `ConsistentHashable` interface.

### DIFF
--- a/src/core/Akka/Actor/ActorRefProvider.cs
+++ b/src/core/Akka/Actor/ActorRefProvider.cs
@@ -1,7 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using Akka.Actor.Internals;
+using Akka.Configuration;
 using Akka.Dispatch;
 using Akka.Dispatch.SysMsg;
 using Akka.Event;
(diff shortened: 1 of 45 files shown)


src/core/Akka/Actor/ActorRefProvider.cs
```diff
@@ -341,67 +343,96 @@ private ActorRef ResolveActorRef(InternalActorRef actorRef, IReadOnlyCollection<
...
+            else //routers!!!
             {
-                deploy = deploy.WithRouterConfig(props.RouterConfig);
-            }
-
-            var routerDispatcher = system.Dispatchers.FromConfig(props.RouterConfig.RouterDispatcher);
-            var routerMailbox = _system.Mailboxes.CreateMailbox(props, null);
-            //TODO: Should be val routerMailbox = system.mailboxes.getMailboxType(routerProps, routerDispatcher.configurator.config)
+                var lookup = (lookupDeploy ? Deployer.Lookup(path) : null) ?? Deploy.None;
+                var fromProps = new List<Deploy>() { props.Deploy.WithRouterConfig(props.Deploy.RouterConfig.WithFallback(props.Deploy.RouterConfig)), deploy, lookup };
```
HCanber: Maybe `new Deploy[] { ... }` instead?

HCanber:
This was a lot. :) I looked through the things I know of (which really isn't much) and except for that tiny remark in `ActorRefProvider` i didn't see anything that upset me (I'm doing emotional based reviewing here :) Wouldn't mind if someone else looked through it too.
Renaming the settings property `DebugRouterMisConfiguration` to `DebugRouterMisconfiguration` is a breaking change as well. Right?

Aaronontheweb:
@HCanber > (I'm doing emotional based reviewing here :) 

this actually made me laugh out loud; I wish all PR feedback started this way :p 

> Renaming the settings property DebugRouterMisConfiguration to DebugRouterMisconfiguration is a breaking change as well. Right?

For some reason this property was defined twice inside `Settings` before:

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/Settings.cs#L229

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/Settings.cs#L241

It's only *supposed* to be used internally by `LocalActorRefProvider* - we weren't using it before but we are now. So to answer your question, I have no idea if this counts as a breaking change :\ - you tell me and we'll just go with that haha

src/core/Akka/Actor/ActorRefProvider.cs
Aaronontheweb: Are you referring to this crazy-assed thing?

`props.Deploy.WithRouterConfig(props.Deploy.RouterConfig.WithFallback(props.Deploy.RouterConfig))`

Because if so, yeah, we should probably get rid of that - it's something that didn't make a lot of sense in the scala IMPL either.

Aaronontheweb:
@smalldave I like your feedback - I'll make those changes today.

smalldave:
@Aaronontheweb doh. got distracted and never finished this. i've focused on the clustering bits (as that's what I know). Looks great. Very exciting to see the stuff in those tests working!

Aaronontheweb:
@smalldave the ClusterGroup routers are not working as expected in the multi-node tests. Appear to work fine in the actual samples though. Not sure why yet. Would have dug deeper on it in this PR but I felt like it had already grown out of control as-is :p

src/core/Akka/Actor/ActorRefProvider.cs
HCanber: The remark was to create an array instead of a list, as the number of items are known and no more is added to it.
But really, as it's only three items in the collection, instead of an aggregation it should probably be rewritten using if statements.

HCanber:
@Aaronontheweb Ok, no breaking change then. :)

Aaronontheweb:
@smalldave on closer inspection, looks like there are some *tiny* dependencies on the Cluster Pool hashing tests - mostly just some changes in who's up / down in the cluster.

I'd like to leave this for the time being, but I implemented all of the other suggestions both you and @HCanber suggested.

Issue #636, opened 2015-02-17
Remoting does not work on Mono


loosechainsaw:
When starting up an actor system that uses remoting on mono it hangs.

Simple code sample.

using System;
using System.Threading;
using Akka;
using Akka.Actor;

namespace AkkaRemoting
{

    public class RequestMessage
    {

    }

    public class BackendActor : ReceiveActor
    {

        public static Random random = new Random(41);

        public BackendActor()
        {
            Receive<RequestMessage>(x =>
            {
                var sender = Sender;
               Console.WriteLine("The value is: " + random.Next());
            });

        }
    }

    public class FrontendActor : ReceiveActor
    {
        public FrontendActor()
        {

            Receive<RequestMessage>(x =>
            {
                var sender = Sender;

                var selection = Context.System.ActorSelection("akka.tcp://System1@localhost:8888/user/backend");
                selection.Tell(new RequestMessage());
            });
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            var system1configuration = FluentConfig.Begin().StartRemotingOn("localhost", 8888);
            var system2configuration = FluentConfig.Begin().StartRemotingOn("localhost", 8889);

            var system1 = ActorSystem.Create("System1", system1configuration.Build());
            var system2 = ActorSystem.Create("System2", system2configuration.Build());

            var backend = system1.ActorOf(Props.Create<BackendActor>(),"backend");
            var frontend = system2.ActorOf(Props.Create<FrontendActor>(),"frontend");

            Thread.Sleep(1000);

            var result = frontend.Ask<string>(new RequestMessage());
            result.ContinueWith(x => Console.WriteLine(x.Result));

            Console.WriteLine("Press any key to exit....");
            Console.ReadKey();
            system1.Shutdown();
            system2.Shutdown();
        }
    }
}

rogeralsing:
I'll dig into this.

rogeralsing:
The endpoint manager seems to fail somehow, the last message that is logged is "Starting prune timer for endpoint manager"

stefansedich:
Any news on this one? I was just today deploying to our mono test server and noticed this :(

stefansedich:
Interesting I tried this out running mono 3.12 on windows and it works fine, so I upgraded my Ubuntu box to 3.12 just then but it still does not work there, just hands in the same place.


Cheers

stefansedich:
[...]

Issue #687, opened 2015-02-27
ActorSelection fails on mono on OSX when paths start with /


bruinbrown:
During ActorPath parsing, a check is made to load in the actor path as an absolute Uri. On Mono (on OSX at least), when parsing an ActorPath which starts with /, this is a valid absolute file URI. This means that any ActorSelection of this form returns the dead letters address instead. It's currently possible to temporarily fix this by adding akka://<system-name> before all absolute paths.

rogeralsing:
oooh... thanks for finding the cause for this!

migueldeicaza:
The issue here is that the Uri API is broken by design.

Microsoft allowed Uris to take paths without the "file://" prefix.   This means that it is valid to call Uri like this:

     new Uri ("c:/autoexec.bat")

Usually the issue comes up with code like this:

    foreach (var f in Directory.GetFiles (...)) 
        xxx (new Uri (f));

So we had to make Uri allow paths, which in Unix includes "/"

The solution is to use the full Uri constructor and full specifiers.

Aaronontheweb:
Thanks @migueldeicaza - this is extremely helpful.

Aaronontheweb:
#708 should fix this. Going to keep the PR open to get some confirmations from our Mono users first.

PR #708, opened 2015-03-03
Change ordering within ActorSelection to fix issue where /foo paths were...

stefansedich:
... not working within mono on linux.


stefansedich:
I hope that this can fix #636 and #687 I was unable to run tests locally due to a stack overflow exception but they failed both before and after my change.

I think it makes sense to have the "no path" check at the top and then check for the /path/path before checking to see if is an absolute Uri.


Thoughts?

Aaronontheweb:
@stefansedich the results from the build server look good - kicked off a second build just to make sure.

PR #711, opened 2015-03-04
Removed a bunch of redundant type checks

sean-gilliam:
Most of these are in the form of the following:

if (message is ClusterUserAction.JoinTo)
{
    var jt = message as ClusterUserAction.JoinTo;
    ...
}

Here message is being checked twice for being of type
ClusterUserAction.JoinTo

So instead I replaced the 'as' cast to a direct cast.

if (message is ClusterUserAction.JoinTo)
{
    var jt = (ClusterUserAction.JoinTo)message;
    ...
}

I would have done the more familiar 'as' cast/check for
null convention but most of these are part of a long
if/else chain that doesn't make sense to break apart.

### DIFF
--- a/src/core/Akka.Cluster/ClusterDaemon.cs
+++ b/src/core/Akka.Cluster/ClusterDaemon.cs
@@ -742,17 +742,17 @@ namespace Akka.Cluster
             }
             else if (message is ClusterUserAction.JoinTo)
             {
-                var jt = message as ClusterUserAction.JoinTo;
+                var jt = (ClusterUserAction.JoinTo)message;
                 Join(jt.Address);
             }
             else if (message is InternalClusterAction.JoinSeedNodes)
             {
-                var js = message as InternalClusterAction.JoinSeedNodes;
+                var js = (InternalClusterAction.JoinSeedNodes)message;
                 JoinSeedNodes(js.SeedNodes);
             }
             else if (message is InternalClusterAction.ISubscriptionMessage)
             {
-                var isub = message as InternalClusterAction.ISubscriptionMessage;
+                var isub = (InternalClusterAction.ISubscriptionMessage)message;
                 _publisher.Forward(isub);
             }
             else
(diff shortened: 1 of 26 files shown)


sean-gilliam:
It's a fact that the is + direct cast is slower than the as + null check, but is + as is slower
than the is + direct cast. Not by much, but its still there.

Testing 30000000 objects. Results in milliseconds
Is and Cast: 470
Is and As: 475
As and null check: 433
Cast: 445
As: 422

Aaronontheweb:
`as` also avoids `NullReferenceException`s in a few areas, whereas unboxing does. It's generally a safer operation.

I appreciate your thoroughness, but I feel like this is an unnecessary micro-optimization.

sean-gilliam:
@Aaronontheweb i hadn't about NRE's it just looked weird with is/as combo. i'll close this out.

HCanber:
I agree with @sean-gilliam _is+as_ is a strange and a very unusual combo (at least in code I've read).
``` C#
//Strange combo
if (message is ClusterUserAction.JoinTo)
{
  var jt = message as ClusterUserAction.JoinTo;
  Join(jt.Address);
}
```

The usual patterns are _is+cast_:
``` C#
if (message is ClusterUserAction.JoinTo)
{
  var jt = (ClusterUserAction.JoinTo) message;
  Join(jt.Address);
}
```
or (for reference types) _as+null-check_
``` C#
var jt = message as ClusterUserAction.JoinTo
if (jt != null)
{
  Join(jt.Address);
}
```

@Aaronontheweb  _is+cast_ is just as safe as _is+as_ (as when we cast we know it will succeed). Right?

Aaronontheweb:
@HCanber TBH this is a pattern that I don't see very often anywhere outside of Akka.NET pattern matching, so I don't have a clear sense of what's "standard".

I fixed a bunch of bugs in my PR with things like router fallbacks where folks were originally using boxing when they should have been using as, which can happen when you're casting from an object.

However, in this case it's safe given the IS syntax above there.

I personally don't think is / as is any more or less strange than boxing.

Issue #639, opened 2015-02-17
Add SandCastle support to build script


Aaronontheweb:
To make it easier to implement #637

PR #719, opened 2015-03-11
added SandCastle support to build system #639

Aaronontheweb:
Should now run SandCastle on build server.

Right now the docs are about ~100mb in size per build, so we're going to run the docs as a separate build from the normal PR builds.

Also, there appears to be an XML-DOC issue inside `Akka.TestKit` that causes SandCastle to crash - I think it's an improperly closed tag somewhere. I couldn't find it myself even with verbose logging turned on, but I'll create a separate issue for that.


Aaronontheweb:
Referencing #639

Aaronontheweb:
BTW, this is the issue I run into when documenting TestKit:

http://shfb.codeplex.com/discussions/452473

I'll include the output of the SandCastle build log into a new issue

Aaronontheweb:
Going to merge this in so I can test on build server

Issue #482, opened 2014-10-20
DequeWrapperMessageQueue does not correctly EnqueueFirst


HCanber:
This is a tricky one. :)
@rogeralsing originally used a Stack as PrependBuffer, then came me and @Aaronontheweb along and said it was wrong, and that a Queue should be used in #424. That was wrong. It fixed the stashing problem, but introduced a new one.
``` C#
q.EnqueueFirst(A)
q.EnqueueFirst(B)

q.Dequeue() //returns A, should be B
q.Dequeue() //returns B, should be A
```

So a Stack _should_ be used internally.

However this introduced a problem that @Aaronontheweb wrote about in #424:
>items that are being prepended to the front of the mailbox can't jump ahead of the other messages that were previously scheduled to be prepended. The PrependBuffer needs to maintain the same FIFO order that the Stash does.

I think what #424 should have fixed was to UnstashAll messages in reverse order. So either we do that or we create an overload in `DequeWrapperMessageQueue` that takes an `IEnumerable` that respects fifo order.

Aaronontheweb:
@HCanber @rogeralsing did we already fix this?

rogeralsing:
Dont think we have

HCanber:
Can't find any commits that suggests this has been fixed.

Aaronontheweb:
@HCanber @rogeralsing ok, we should definitely fix this before V1 then. Any takers?

PR #721, opened 2015-03-11
Fixed the DequeWrapperMessageQueue so it correctly calls EnqueueFirst (#482)

Aaronontheweb:
Fixed the DequeWrapperMessageQueue so it correctly calls EnqueueFirst per issue #482

CC @HCanber @rogeralsing could you take a look at the changes I made to `AbstractStash` and `DequeWrapperMessageQueue` and verify that the behavior is correct?

I reverse the contents of the `Stack` when we call `UnstashAll` and as a result none of the Remoting or Clustering tests were affected. Ran the MultiNodeTest suite on my local machine and got the same results as  before, so I think we're set.


rogeralsing:
Looks good

HCanber:
+1

PR #726, opened 2015-03-12
Introduce IScheduler and use it everywhere instead of Scheduler.

### LINKED ISSUE #468: Change the Scheduler api, Take 2
Right now the Scheduler is a class where you can 
- schedule to send a message in the future, once or repeatably
- schedule an action to be executed in the future, once or repeatably

It looks like this

```c#
//Tell in the future
Task ScheduleOnce(TimeSpan initialDelay, ActorRef receiver, object message)
Task ScheduleOnce(TimeSpan initialDelay, ActorRef receiver, object message, CancellationToken cancellationToken)
Task Schedule(TimeSpan initialDelay, TimeSpan interval, ActorRef receiver, object message)
Task Schedule(TimeSpan initialDelay, TimeSpan interval, ActorRef receiver, object message, CancellationToken cancellationToken)

//Schedule action
Task ScheduleOnce(TimeSpan initialDelay, Action action)
Task ScheduleOnce(TimeSpan initialDelay, Action action, CancellationToken cancellationToken)
Task Schedule(TimeSpan initialDelay, TimeSpan interval, Action action)
Task Schedule(TimeSpan initialDelay, TimeSpan interval, Action action, CancellationToken cancellationToken)
```

__Remove schedule an action__
Scheduling an action within an actor is not a good thing to do and is discouraged. It may be a source for strange bugs. So me and @rogeralsing suggest that all those overloads are removed.
If you need to do it outside an actor, there are things in TPL and Task you can use.

__Remove Task__
When inside an actor it would be very strange to await a `ScheduleOnce` like this:
``` C#
await Scheduler.ScheduleOnce(TimeSpan.FromMinutes(1), Self, theMessage);
```
Returning a `Task` opens up for more bugs, and inside our codebase we never utilizes that a Task is returned.
So we suggest we should not return `Task`.

__Open up for other types of schedulers__
It would be really useful to have a scheduler in TestKit that let's you control the time. 
Instead of running tests in real time, you can run them much faster, and still be able to test that things happens on timeouts. For example, something like this:

``` C#
[Fact]
public Retrier_should_resend_if_no_response_is_received_within_a_minute()
{
  var retrier = ActorOf<Retrier>();
  retrier.Tell(new Start(TestActor)); //This will kick off everything, and all messages should be sent to TestActor
  ExpectMsg<Message>(); //Expect a message immediately
  Time.Advance(TimeSpan.FromSeconds(59);
  ExpectNoMsg();  //We haven't reached the timeout yet, so no message should have been sent.
  Time.Advance(TimeSpan.FromSeconds(2));
  ExpectMsg<Message>(); //Expect that a new message to be sent since a minute has passed
}
```

In order to do that we need to have full control over how timeouts and delayed sends are scheduled.
We also need full control over how cancellation works. For example, `CancellationTokenSource` has a `CancelAfter(TimeSpan)` which creates a `Timer` that we have no control over and it cannot be swapped out with anything else.

__IScheduler & ICancellable__
So the suggestion is that we introduce and return `ICancellable` and change scheduler to an interface like this:

``` C#

public interface IScheduler
{
  ICancellable ScheduleOnce(TimeSpan initialDelay, ICanTell receiver, object message)
  ICancellable Schedule(TimeSpan initialDelay, TimeSpan interval, ICanTell receiver, object message)
}

public interface ICancellable
{
    void Cancel();
    bool IsCancelled { get; }
}
```

We would also have extension methods to link two or more `ICancellable` together, just like `CancellationTokenSource`:
``` C#
var c1 = Scheduler.ScheduleOnce(interval1, Self, message1);
var c2 = Scheduler.ScheduleOnce(interval2, Self, message2);
var c2 = Scheduler.ScheduleOnce(interval3, Self, message3);

_cancellable= c1.LinkWith(c2, c3);
//Or another variant:
_cancellable= Cancellable.Create(c1, c2, c3);
```

__Delayed ActorRef.Tell__
To simplify the api new extension methods on `ActorRef` are created that would allow this:
``` C#
actor.Tell(msg, TimeSpan.FromMinutes(1));
actor.Tell(msg, TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1), );
```
Or with new names to make it clearer
``` C#
actor.TellOnce(msg, TimeSpan.FromMinutes(1));
actor.TellRepeatedly(msg, TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(2));
```

Thoughts?

### LINKED ISSUE #709: Scheduler should never allow a TimeSpan.Zero or equivalent value for its interval
Creates an infinite loop and spams an actor's mailbox faster than it can process. Need to use a `Guard.Assert` statement to prevent this value from being captured.

This issue was directly responsible for an Akka.Cluster bug that took @smalldave and I a long time to track down.

HCanber:
Fixes #468.
Fixes #709.

__What is it__
The reason for this PR is to prepare for a scheduler that can be
used in tests without having to run the tests in real time.

`ICancelable` & `Cancelable` is basically a wrapper around `CancellationTokenSource`. 

__Why?__
`CancellationTokenSource` is replaced to be able to customize what `CancelAfter` does.  `CancellationTokenSource.CancelAfter` always runs in real time, but as we want to be able to use a scheduler that runs in it's own time `CancellationTokenSource` cannot be used. 

However, if the scheduler used, actually runs in real time, like the normal scheduler does, then `CancellationTokenSource.CancelAfter` is used under the hood, so during normal execution the scheduler uses the  `CancellationTokenSource` just as before this PR.

__API changes__
These are the ways to schedule things introduced in this PR:

``` C#
Context.System.Scheduler
  .ScheduleTellOnce(TimeSpan delay, ICanTell receiver, object message, ActorRef sender);
  .ScheduleTellOnce(TimeSpan delay, ICanTell receiver, object message, ActorRef sender, ICancelable cancelable);
  .ScheduleTellRepeatedly(TimeSpan initialDelay, TimeSpan interval, ICanTell receiver, object message, ActorRef sender);
  .ScheduleTellRepeatedly(TimeSpan initialDelay, TimeSpan interval, ICanTell receiver, object message, ActorRef sender, ICancelable cancelable);

Context.System.Scheduler.Advanced
  .ScheduleOnce(TimeSpan delay, Action action);
  .ScheduleOnce(TimeSpan delay, Action action, ICancelable cancelable);
  .ScheduleRepeatedly(TimeSpan initialDelay, TimeSpan interval, Action action);
  .ScheduleRepeatedly(TimeSpan initialDelay, TimeSpan interval, Action action, ICancelable cancelable);
```
There are also oveloads/extension methods for specifying delays/intervals in milliseconds.
I've also created extension methods for all four variants (`ScheduleTellOnceCancelable`, `ScheduleTellRepeatedlyCancelable`, `ScheduleOnceCancelable`, `ScheduleRepeatedlyCancelable`) that creates a cancelable, schedules, and returns the cancelable. This was the most common scenario for how the scheduler was used in our code, so I thought I'd make helper methods to simplify the code. Example:
``` C#
ICancelable cancelable = Context.System.Scheduler.ScheduleTellRepeatedlyCancelable(PruneInterval, PruneInterval, Self, new Prune(), Self)
```

See issue #468 for more information and background. A scheduler that can be used during tests will be submitted in a separate PR later.

__Reviewing__
What to look out for when reviewing:
- F# api changes: @Horusiath could you have a look at that?
- Akka.Cluster changes, @Aaronontheweb could you have a look at the changes? A side-by-side/split comparison is probably better than the unified for these


src/core/Akka.FSharp/FsApi.fs
```diff
@@ -604,8 +610,20 @@ module Scheduler =
     /// <param name="message">Message to be sent to the receiver by the scheduler.</param>
     /// <param name="receiver">Message receiver.</param>
     /// <param name="scheduler"></param>
-    let scheduleTell (after: TimeSpan) (every: TimeSpan) (message: 'Message) (receiver: ActorRef) (scheduler: Scheduler): Async<unit> =
-        Async.AwaitTask (scheduler.Schedule(after, every, receiver, message).ContinueWith taskContinuation)
+    let scheduleTell (after: TimeSpan) (every: TimeSpan) (message: 'Message) (receiver: ActorRef) (scheduler: IScheduler): unit =
+         DeprecatedSchedulerExtensions.Schedule(scheduler, after, every, receiver, message)
```
rogeralsing: Just so I get this, what does this one do that requires it to use deprecated scheduler mechanics?

rogeralsing:
Beautiful :heart:

src/core/Akka.FSharp/FsApi.fs
HCanber: The new `ScheduleTellOnce` and -`Repeatedely` requires a sender (no implicit sender), so I would need to introduce a new argument in the F# code. I think, by looking at the other F# code, I know how to that, but I'd rather have someone, like @Horusiath fix this instead. :)

HCanber: I think the entire F# api for scheduling needs a rewrite...

Horusiath: I'll adjust F# API to those changes.

Aaronontheweb:
Running the build one more time just because I've never seen the EMWASpec fail before. I'll take a look at this and run the multinode test suite too, which is where any issues with Akka.Cluster would probably come up.

Aaronontheweb:
@HCanber does this also resolve #709?

Aaronontheweb:
MultiNodeTests pass and the changes you made to the clustering internals look good to me.

However, this test failure is legitimate:

```xml
System.InvalidOperationException : The Counter layout for the Category specified is invalid, a counter of the type:  AverageCount64, AverageTimer32, CounterMultiTimer, CounterMultiTimerInverse, CounterMultiTimer100Ns, CounterMultiTimer100NsInverse, RawFraction, or SampleFraction has to be immediately followed by any of the base counter types: AverageBase, CounterMultiBase, RawBase or SampleBase.
   at System.Diagnostics.CategorySample.GetCounterDefinitionSample(String counter)
   at System.Diagnostics.PerformanceCounter.NextSample()
   at System.Diagnostics.PerformanceCounter.NextValue()
   at Akka.Cluster.PerformanceCounterMetricsCollector.SystemLoadAverage() in d:\BuildAgent\work\49b164d63843fb4\src\core\Akka.Cluster\ClusterMetricsCollector.cs:line 780
   at Akka.Cluster.PerformanceCounterMetricsCollector.Sample() in d:\BuildAgent\work\49b164d63843fb4\src\core\Akka.Cluster\ClusterMetricsCollector.cs:line 762
   at Akka.Cluster.Tests.EWMASpec.Calculate_the_EWMA_for_multiple_variable_datastreams() in d:\BuildAgent\work\49b164d63843fb4\src\core\Akka.Cluster.Tests\EWMASpec.cs:line 106
```

I can't replicate the error on my machine, but in this PR it happens here https://github.com/HCanber/akka.net/blob/scheduler-468/src/core/Akka.Cluster/ClusterMetricsCollector.cs#L780

This file has been modified as part of this PR, but none of the performance counters were actually updated. So I don't actually know why this test is failing, but it only fails on this PR thus far :(

HCanber:
@Aaronontheweb Yes this also resolves #709. These tests verifies that an exception is thrown in these cases: 
https://github.com/akkadotnet/akka.net/pull/726/files#diff-fc5bbf50574e056648320f491e0303aeR158
https://github.com/akkadotnet/akka.net/pull/726/files#diff-e82e2dcc65ff0e69feae14e3e28c95e5R145

Regarding the failing test, I can't see how any of my changes could affect that.

Aaronontheweb:
@HCanber might be an issue with the build agent. In that case I'll merge in the PR and see about dealing with the ClusterMetrics stuff later - I need to refactor all of that into a new NuGet package anyway

Issue #722, opened 2015-03-11
F# API: Configuration.load is broken


Horusiath:
@jorgef drew my attention, that `Configuration.load()`, which is responsible for returning configuration from *.config files, fails silently returning empty HOCON node. 

Is this somewhat related to #679 ?

Aaronontheweb:
@Horusiath does the `ConfigurationManager` behave differently for FSharp? Should be the same.

Oh snap, you know what it might be - we need to reference `System.Configuration` in `Akka.FSharp` in order for this to work.

rodrigovidal:
Is it fixed?

Horusiath:
It looks so. I've created a test case in Akka.FSharp.Tests to check if loading is performed without problems.

Issue #680, opened 2015-02-26
F# API: use F# quotation expression for serializable strategy


Horusiath:
Since #599 has been done, F# provides another way of implementing serializable deciders. The idea is to create F#-only version of the `IDecider` interface using serializable quotation expressions and expose it as another `Strategy` member.

rogeralsing:
:+1:

Horusiath:
Closed with #732

PR #732, opened 2015-03-14
F# API strategy decider + schedulers

Horusiath:
List of changes:

- New type of `IDecider` using F# quotation expressions and reflected this change in `Strategy` methods inside F# API.
- Completely new version of supervision strategy example to include actor lifecycle presentation - since F# actors don't expose lifcycle hook methods explicitly, new example presents how equivalents of the `PreStart`/`PostStop` methods could be achieved.
- (**breaking change!**) Also due to changes made by @HCanber all scheduling functions has been removed from F# API and replaced with extension methods to `IScheduler` and `IAdvancedScheduler` interfaces, which could work smoothly with F# functions and optional parameters.
- Last change (**also breaking one**) is making ask operator `<?` generic. Now it leverages F# type inference - this has been reflected in examples, which now have reduced number of explicit casts and type declarations.

cc #680 and #722 (fix proven in test case)


Aaronontheweb:
@Horusiath is this good to go?

Horusiath:
Yes. It is.

src/core/Akka.FSharp/Schedulers.fs
```diff
@@ -0,0 +1,67 @@
+﻿[<AutoOpen>]
+module Schedulers
+
+open Akka.Actor
+open System
+
+type Akka.Actor.IAdvancedScheduler with
```
HCanber: Better to change this to `type Akka.Actor.IActionScheduler`

src/core/Akka.FSharp/Schedulers.fs
```diff
@@ -0,0 +1,67 @@
...
+    /// <param name="after">Delay before calling the function.</param>
+    /// <param name="fn">Function called by the scheduler.</param>
+    /// <param name="cancelable">Optional cancelation token</param>
+    member this.ScheduleOnce(after: TimeSpan, fn: unit->unit, ?cancelable: ICancelable) : unit =
+        let action = Action fn
+        match cancelable with
+        | Some c -> this.ScheduleOnce(after, action, c)
+        | None -> this.ScheduleOnce(after, action)
+        
+type Akka.Actor.IScheduler with
```
HCanber: Change to `type Akka.Actor.ITellScheduler with`

Issue #706, opened 2015-03-03
Issues with DI Support


stefansedich:
Looking over the DI support I spot what I see as a fairly big issue (for disposable components referenced within an actor), we do not handle releasing any components, for example in Autofac support we resolve on a root IContainer but we do not handle lifetime scopes or anything like that, if we had disposable services injected into actors these would not be disposed when the actor is done with.

Ideally the actors life cycle should be wrapped by an ILifetimeScope which is disposed of when the actor is, that would ensure we clean up things properly.

Thoughts?

Thanks
Stefan

Horusiath:
I found it hard to achieve by any external plugin/framework. Actors have their own life cycle, which does not cover with what you may usually expect - i.e. don't expect that actor will be GCed simply when you stop sending messages to it.

Right now we can deal with it in several ways:

1. Let your actor implement `IDisposable` interface and dispose it's resources there.
2. Override `PostStop` method and dispose your objects at the end.
3. Create your own pipeline plugin and use it's `BeforeIncarnated` method to release resources.

Option no. 3 gives some chances, that it can be used at DI plugin level, but I don't know internals of specific DIs enough to be sure for that.

stefansedich:
I think at the moment if we have some guidance around using the various plugins, i.e. if using autofac ensure you use ExternallyOwned or use an Owned<T> and handle disposing the dependencies on your own inside your actor etc...

I have not really looked too deeply into how IOC is implemented to see how it could actually be resolved properly to use lifetime scopes etc but I think some guidance around these things could be a good start, thoughts?


Thanks
Stefan

Aaronontheweb:
Senor @Horusiath did a better job than I could explaining how to tackle this issue. 

> I think at the moment if we have some guidance around using the various plugins, i.e. if using autofac ensure you use ExternallyOwned or use an Owned and handle disposing the dependencies on your own inside your actor etc...

I think you're right on @stefansedich - that does need to be explained better. I've never used Autofac, but would you mind sending a PR to amend this page in our docs explaining just that? 

https://github.com/akkadotnet/akka.net/blob/dev/documentation/wiki/Dependency%20injection.md

Maybe also include that in the XML-DOC comments for the module too. Does that sound like a good start?

Aaronontheweb:
So @jcwrequests wrote a really detailed proposal for how to solve this problem here: http://www.message-flow.com/2015/03/ioc-contrib-design-flaw.html

I have a couple of questions for him here:

[...]

PR #733, opened 2015-03-15
Patch to Issues with DI Support #706

jcwrequests:
Patch to Issues with DI Support #706


Aaronontheweb:
MSBuild failed: http://petabridge-ci.cloudapp.net/viewLog.html?buildId=264&tab=buildResultsDiv&buildTypeId=AkkaNet_AkkaNetPrBuild

nvivo:
I'm not sure if this is planned or not, but since the DI producer is registered as an extension inside the actor system, shouldn't the ActorOf methods  (both from ActorSystem and children Actors) automatically support it internally somehow?

Having to [extract the DIExt](https://github.com/akkadotnet/akka.net/pull/733/files#diff-5cfcb16334e195be265545548eb09aa1R62) in every actor to create children seems too low level for general use.

jcwrequests:
@nvivo I based my implementation off the jvm side so it should theoretically work the same. Have you tried using the contrib and testing if that would work? Either way send me some examples and I will take a look.

nvivo:
@jcwrequests, I don't know much about the JVM implementation, so I'm sorry if the question is out of place. I was looking at the examples you provided in this PR, and inside an actor you do:

```cs
var producer = Context.System.GetExtension<DIExt>();
Context.ActorOf(producer.Props("TypedWorker"));
```

This actor must know a DI framework is being used and code accordingly, which seems strange as the DI should be "transparent" to the application of code.

Since the extension is registered inside the ActorSystem, shouldn't be able to just call `System/Context.ActorOf<TypedWorker>()` and that would use the DI extension internally if registered?

I'm currently using the contrib from nuget with Autofac, but it seems this part didn't change in this PR.

jcwrequests:
@nvivo I will take a look as soon as I can to see what can be done. There maybe some changes that need to occur to Akka itself. My contrib was using the extensibility extensions with the framework. This maybe something that could be solved with an extension method. Either way I will take a look.

jcwrequests:
@nvivo The problem is with the extension method which is not taking into account the possibility of DI being used. That was not something I coded but I think it is worth looking at seeing if there is a way around the issue. @aaronontheweb @hcanber @rogeralsing what do you guys think?

Aaronontheweb:
> Since the extension is registered inside the ActorSystem, shouldn't be able to just call System/Context.ActorOf<TypedWorker>() and that would use the DI extension internally if registered?

@nvivo that's not how actor system extensions work. They're meant to be called directly.

The actor construction pipeline, a special type of extension does work this way however - we don't have detailed docs for it but we do have a description of its API here: https://github.com/akkadotnet/akka.net/releases/tag/v0.8

It's configurable and can perform special initialization and shutdown logic on actors. It's how stashes get created and destroyed currently. Th'at's something worth investigating cc @jcwrequests

nvivo:
@Aaronontheweb, I'm not questioning anything related to the how the extensions or Akka works. I'm sorry if that was the impression I gave, this was really not my intention.

@jcwrequests, please it was just a question, don't bother trying to change anything. I'm under the impression that for this to work, Akka would need to recognize DI directly, so it's really transparent to the developer. In theory, if DI is supported we should be able to do something like `Context.ActorOf<IServiceInterface>()`, but that's a long shot and I'm just brainstorming here.

I need to take a look at how the JVM works for this, but I'd be glad to help with anything.

jcwrequests:
@nvivo I think you have a valid point that may be solvable and help everyone using the extension. Don't worry I would spend to much time it's an interesting problem.

Aaronontheweb:
>  I'm not questioning anything related to the how the extensions or Akka works. I'm sorry if that was the impression I gave, this was really not my intention.

No worries man - was just trying to educate you on how they work. No need to apologize.

jcwrequests:
@Aaronontheweb @nvivo Just made the changes and added an extension method and adapter in the core to make created child actors with DI simple.

rogeralsing:
@jcwrequests could you fix the current conflicts and prefferably squash to a single commit?

jcwrequests:
@rogeralsing Not sure if squashing at this time will screw anything up. Sorry about the conflicts for some reason I could not get my fork's upstream to merge properly so I was not seeing the rename changes made to ConsistentHashable.

jcwrequests:
@rogeralsing I see there are some new conflicts I will look into it.

rogeralsing:
@akkadotnet/developers should we pull this in?
Should we squash the commits? anyone that is comfortable with git that can assist on that?

Aaronontheweb:
Replacing this PR with #775

Issue #745, opened 2015-03-19
Make Akka.Persistence.TestKit more intuitive


Horusiath:
Right now the minimal configuration needed to create custom Journal spec using persistence test kit is: 

```csharp
public class MyJournalSpec : JournalSpec
{
        private static readonly Config SpecConfig = ConfigurationFactory.ParseString(@"
            akka.persistence {
                publish-plugin-commands = on
                journal {
                    plugin = ""akka.persistence.journal.my""
                    my {
                        class = ""TestPersistencePlugin.MyJournal, TestPersistencePlugin""
                        plugin-dispatcher = ""akka.actor.default-dispatcher""
                    }
                }
            }
        ");
        public MyJournalSpec() : base(SpecConfig) { }
}
```

Most of this is a boilerplate. Desired result should be more like:

```csharp
public class MyJournalSpec : JournalSpec
{
        public MyJournalSpec() : base(typeof(MyJournal)) { }
}
```
Task is fairly simple - we need to extend Akka.Persistence.TestKit to generate necessary configuration based on journal type.

Same rules applies to `SnapshotStoreSpec`.

PR #749, opened 2015-03-23
Akka persistence bugfixes

Horusiath:
Change log:

1. Minor changes in F# Persistence API:
    - Added `Defer` method from core F# actors API
    - Renamed `Persist`, `AsyncPersist` and `Defer` to `PersistEvent`, `AsyncPersistEvent` and `DeferEvent` due to `Defer` method naming conflict with core F# API for `Defer` method.
2. Altered state machine for `PersistentView` - these changes mirror latest updates of canonical akka-persistence API: 
    - Removed *PrepareRestart* state (it's logic has been moved to *ReplayFailed* state). 
    - Moved update scheduler initialization from *ReplayStarted* state to `PreStart` method. Also it now uses new `ScheduledUpdate` message instead of `Update`. This message has a dedicated behavior in *Idle* state.
3. Better support for `LocalSnapshotStore`:
    - Asynchronous tasks are now activated using Akka dispatchers.
    - Local snapshot store has now custom settings class defined and configurable number (it was fixed before) of subsequent load attempts of the file-based snapshots - it's used in case of i.e. corrupted snapshots. 
4. Additional Journal/SnapshotStore specs constructors (see #745)
5. Local snapshot store folder for persistence spec is now generated for each test case separately.
6. Introduced new test kit method - `ExpectMsgAnyOf` (from canonical akka-testkit).
7. Fixed all failing tests in `PersistentActorSpec`.
8. Changed all actor path strings in `GuaranteedDeliverySpec` back to `ActorPath`, since they are correctly serialized/deserialized now.

### DIFF
--- a/src/core/Akka.TestKit/TestKitBase_Expect.cs
+++ b/src/core/Akka.TestKit/TestKitBase_Expect.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Threading;
 using Akka.Actor;
 using Akka.TestKit.Internal;
+using Akka.Util;
 
 namespace Akka.TestKit
 {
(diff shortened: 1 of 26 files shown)


src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -253,6 +254,20 @@ private void InternalExpectNoMsg(TimeSpan duration)
             }
         }
 
+        public T ExpectMsgAnyOf<T>(params T[] messages)
```
HCanber: Add comments to this method as it is public.

HCanber:
Squash "unnecessary" commits into the parents and write commit messages that tells a bit of what the commit changes (and why) so that others reading them get a clue what's going on.

Horusiath:
Thx @HCanber . Now everything is squashed and commented.

HCanber:
thx @Horusiath . Maybe you're already aware of this: PersistanceExample doesn't seem to work. Nothing except `--- SNAPSHOTED ACTOR EXAMPLE ---` is printed.

PR #757, opened 2015-03-28
Extension method for dispatcher returning task

Horusiath:
This is proposition of generic dispatcher `Schedule` variation returning TPL task used for tracking completion status. I've found it a pretty useful, because sometimes I want to delegate action through Akka dispatcher but still be able to wait/continue/pipe work from it. 

It works with any type of dispatcher, also one not using TPL internally. 

@akkadotnet/developers What do you think about this?

### DIFF
--- a/src/core/Akka/Dispatch/Dispatchers.cs
+++ b/src/core/Akka/Dispatch/Dispatchers.cs
@@ -14,6 +14,7 @@ namespace Akka.Dispatch
         PinnedDispatcher,
         SynchronizedDispatcher,
     }
+
     public static class DispatcherTypeMembers
     {
         public static string GetName(this DispatcherType self)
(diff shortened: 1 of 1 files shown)


rogeralsing:
Looks good to me.
But under what conditions do we want to schedule things using the dispatcher?
(Might sound unnecessary pessimistic here now, but just want to understand the use case)
 
The dispatcher is at least on the JVM very closely tied to the actors mailbox.
Also if there is a legit reason to schedule things on the mailbox dispatchers, shouldn't this be combined with the scheduler API?

Horusiath:
I've found that canonical implementation of akka persistence exposes it's API to return futures from most methods. In default snapshot store implementation this API is consumed by actor by delegating the work through akka dispatcher - so everything stays in the family. I've thought, it's probable that this pattern may be repeated in other plugins over the course of time.

HCanber:
My objection is the name `ScheduleTask` as it's not scheduling any tasks at all. `ScheduleAsync` maybe?

Aaronontheweb:
I think this is a really good idea - I'd love to be able to run specific scheduler tasks on top of a `ForkJoinDispatcher`. Will this allow me to do that?

rogeralsing:
@Aaronontheweb that is not what this does, this schedules an action on the dispatcher, but wraps that in a TaskCompletionSource so that you can be notified when the action has run to completion.
So as @HCanber says, the name is somewhat misleading.

It simply allows for notification when something is done executing

rogeralsing:
I'm with @HCanber `ScheduleAsync` would be the .NET idiomatic name for this method, it does schedule and return a task that you can await.

Horusiath:
Ok, it's changed to `ScheduleAsync`.

PR #759, opened 2015-03-29
DispatcherConfigurators, ForkJoinDispatcher, and more

### LINKED ISSUE #607: Add named dispatchers to Akka.Remote and Akka.Cluster
We have a bunch of `todo:`s inside Akka.Remote and Akka.Cluster for using specific, dedicated, named dispatchers.

## Akka.Remote Example

Here's an example inside Akka.Remote (inside `Remote.conf`)

````
### Default dispatcher for the remoting subsystem

    default-remote-dispatcher {
      type = Dispatcher
      executor = "fork-join-executor"
      fork-join-executor {
        # Min number of threads to cap factor-based parallelism number to
        parallelism-min = 2
        parallelism-max = 2
      }
    }
````

And a class that actually depends on it: https://github.com/akkadotnet/akka.net/blob/dev/src%2Fcore%2FAkka.Remote%2FRemoting.cs#L273

```csharp
internal class TransportSupervisor : UntypedActor
    {
        private readonly SupervisorStrategy _strategy = new OneForOneStrategy(3, TimeSpan.FromMinutes(1), exception => Directive.Restart);
        protected override SupervisorStrategy SupervisorStrategy()
        {
            return _strategy;
        }

        {
            PatternMatch.Match(message)
                .With<RegisterTransportActor>(r =>
                {
                    /*
                     * TODO: need to add support for RemoteDispatcher here.
                     * See https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/RemoteSettings.scala#L42 
                     * and https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Remoting.scala#L95
                     */
                    Sender.Tell(Context.ActorOf(r.Props.WithDeploy(Deploy.Local), r.Name));
                });
        }
    }
```
These need to be backed up by an actual dedicated threadpool dispatcher.

### LINKED ISSUE #629: Memory footprint
I've been profiling the Akka.NET footprint today.
It was not as good as I expected.
We seem to be at about 1300ish bytes per actor when they are fully created.
This is far too expensive.

I found a few problems.

1. Dispatchers are being recreated per actor, so each actor has its own dispatcher..
1. Watchers is eagerly initialized per actor as a HashSet. 
1. Same goes for WhatchedBy
1. Same goes for TerminateQueue

There should be a good amount of bytes to save by fixing the above issues.
e.g. replace hashset to some other structure, or make it lazy initialized or with a lower capacity

Aaronontheweb:
close #629
affects #629
close #607

Added ForkJoinDispatcher, Dispatchers, DispatcherConfigurators, and the ability
for plugins to write their config settings directly into Dispatchers.

### Using ForkJoinDispatcher
So ForkJoinDispatcher is special - it uses [`Helios.Concurrency.DedicatedThreadPool`](https://github.com/helios-io/DedicatedThreadPool) to create a dedicated set of threads for the exclusive use of the actors configured to use a particular `ForkJoinDispatcher` instance. All of the remoting actors depend on the `default-remote-dispatcher` for instance.

Here's how you can create your own ForkJoinDispatcher instances via Config:

```
myapp{
  my-forkjoin-dispatcher{
    type = ForkJoinDispatcher
    throughput = 100
    dedicated-thread-pool{ #settings for Helios.DedicatedThreadPool
      #deadlock-timeout = 3s #optional timeout for deadlock detection
      threadtype = background #values can be "background" or "foreground"
    }
  }
}
}
```

You can then use this specific `ForkJoinDispatcher` instance by configuring specific actors to use it, whether it's via config or the fluent interface on `Props`:

**Config**
```
akka.actor.deploy{
     /myActor1{
       dispatcher = myapp.my-forkjoin-dispatcher
     }
}
```

**Props**
```csharp
var actor = Sys.ActorOf(Props.Create<Foo>().WithDispatcher("myapp.my-forkjoin-dispatcher"));
```

### Dispatchers
We now have a proper implementation of the `Dispatchers` class, which caches configurations for specific dispatchers and ensures that the same `MessageDispatcher` instance is returned each time when an actor requests access to a specific one.

The only exception to this behavior is the `SynchronizedDispatcher` - that particular one returns a new instance every time its called because it's associated with a specific `ExecutionContext`.

But other than that, all dispatchers are now cached and able to be configured independently.

### Remoting and Persistence actors now run on their own ForkJoinDispatcher
All of the system actors needed to support `Akka.Remote` and `Akka.Persistence` now run on their own named `ForkJoinDispatcher` instances, per what's baked into the Akka configuration sections. I've run the test suite probably 30 times now and everything seems to be ok.


Aaronontheweb:
Apparently TeamCity doesn't like the old timestamp on this rebased commit :p - going to close and send a new PR real quick

PR #760, opened 2015-03-29
DispatcherConfigurators, ForkJoinDispatcher, and more

### LINKED ISSUE #607: Add named dispatchers to Akka.Remote and Akka.Cluster

### LINKED ISSUE #629: Memory footprint

Aaronontheweb:
close #629
affects #629
close #607

Added ForkJoinDispatcher, Dispatchers, DispatcherConfigurators, and the ability
for plugins to write their config settings directly into Dispatchers.

### Using ForkJoinDispatcher
So ForkJoinDispatcher is special - it uses [`Helios.Concurrency.DedicatedThreadPool`](https://github.com/helios-io/DedicatedThreadPool) to create a dedicated set of threads for the exclusive use of the actors configured to use a particular `ForkJoinDispatcher` instance. All of the remoting actors depend on the `default-remote-dispatcher` for instance.

Here's how you can create your own ForkJoinDispatcher instances via Config:

```
myapp{
  my-forkjoin-dispatcher{
    type = ForkJoinDispatcher
    throughput = 100
    dedicated-thread-pool{ #settings for Helios.DedicatedThreadPool
      #deadlock-timeout = 3s #optional timeout for deadlock detection
      threadtype = background #values can be "background" or "foreground"
    }
  }
}
}
```

You can then use this specific `ForkJoinDispatcher` instance by configuring specific actors to use it, whether it's via config or the fluent interface on `Props`:

**Config**
```
akka.actor.deploy{
     /myActor1{
       dispatcher = myapp.my-forkjoin-dispatcher
     }
}
```

**Props**
```csharp
var actor = Sys.ActorOf(Props.Create<Foo>().WithDispatcher("myapp.my-forkjoin-dispatcher"));
```

### Dispatchers
We now have a proper implementation of the `Dispatchers` class, which caches configurations for specific dispatchers and ensures that the same `MessageDispatcher` instance is returned each time when an actor requests access to a specific one.

The only exception to this behavior is the `SynchronizedDispatcher` - that particular one returns a new instance every time its called because it's associated with a specific `ExecutionContext`.

But other than that, all dispatchers are now cached and able to be configured independently.

### Remoting and Persistence actors now run on their own ForkJoinDispatcher
All of the system actors needed to support `Akka.Remote` and `Akka.Persistence` now run on their own named `ForkJoinDispatcher` instances, per what's baked into the Akka configuration sections. I've run the test suite probably 30 times now and everything seems to be ok.


src/core/Akka/Dispatch/CachingConfig.cs
```diff
@@ -0,0 +1,277 @@
...
+    /// <summary>
+    /// INTERNAL API
+    /// 
+    /// A <see cref="CachingConfig"/> is a <see cref="Config"/> that wraps another <see cref="Config"/> and is used to
+    /// cache path lookup and string retrieval, which we happen to do in some critical paths of the actor creation
+    /// and mailbox selection code.
+    /// 
+    /// All other <see cref="Config"/> operations are delegated to the wrapped <see cref="Config"/>.
+    /// </summary>
+    class CachingConfig : Config
```
rogeralsing: This should be used carefully as it could interfere with `InjectTopLevelFallback`. where the config might cache a value that is later overridden by a submodule, e.g. remoting/persistence.

Is this handled in the current impl? is this recreated when new fallbacks are applied?

src/core/Akka/Dispatch/Dispatchers.cs
```diff
@@ -186,84 +147,211 @@ public MessageDispatcher DefaultGlobalDispatcher
...
+
+        private MessageDispatcherConfigurator LookupConfigurator(string id)
+        {
+            MessageDispatcherConfigurator configurator;
+            if (!_dispatcherConfigurators.TryGetValue(id, out configurator))
             {
-                var disp = new ThreadPoolDispatcher
+                // It doesn't matter if we create a dispatcher configurator that isn't used due to concurrent lookup.
+                // That shouldn't happen often and in case it does the actual ExecutorService isn't
+                // created until used, i.e. cheap.
```
rogeralsing: Isn't this all solved by the `ConcurrentDictionary<.>.GetOrAdd(...)` ?

https://msdn.microsoft.com/en-us/library/ee378676%28v=vs.110%29.aspx

rogeralsing:
Looks great!

Regarding the name `ForkJoinDispatcher` should we stick with that in .NET too?
Pros:
It fills the same purpose as on JVM, so it is familiar.

Cons:
It is familiar but does not behave the same, the `ForkJoinDispatcher` is the fastest on JVM, it is not on .NET (atm).
It really doesn't use Fork-Join mechanics in our case.

This is all I have ;-)

src/core/Akka/Dispatch/CachingConfig.cs
rogeralsing: We are also passing bits of configurations across the wire, e.g. some deployment primitives holds a `Config` objects.
This _should_ be handled by the surrogate stuff, but I guess we need to make sure?

Aaronontheweb: @rogeralsing truth be told, I'm not clear on how canonical Akka handles this. So I came up with my own solution.

Here's what I did:
Have the `Settings` class publish changes to `Dispatchers`
https://github.com/Aaronontheweb/akka.net/blob/dispatcher-squash-timestamp/src/core/Akka/Actor/Settings.cs#L27

And then have `Dispatchers` rebuild its `CachingConfig`:
https://github.com/Aaronontheweb/akka.net/blob/dispatcher-squash-timestamp/src/core/Akka/Dispatch/Dispatchers.cs#L167

This issue came up when I was integrating @Horusiath's latest Akka.Persistence changes. In all other cases when `InjectTopLevelFallback` is called it's done through the `ActorRefProvider` (i.e. in Akka.Remote / Akka.Cluster) but not in Akka.Persistence. So this is the solution I came up with to allow a plugin to inject its own dispatchers into the registry when it gets called.

The only other way of doing this AFAIK would be to scan all of the assemblies loaded into an Akka.NET AppDomain for embedded HOCON files (we would have to standardize the name) and chain all of them at the same time, rather than do it when they're invoked (plugins get invoked after `ActorRefProvider` and after `Dispatchers` is setup, which is what created this problem.)

I think that's what the JVM does and why all of their resource files are found in the same filename and path in each .jar.

src/core/Akka/Dispatch/Dispatchers.cs
Aaronontheweb: I think so but all of the funky error handling and lookup stuff would have to be done inside a lambda. So I ran with this just because it was easier to debug.

Aaronontheweb:
Regarding the `ForkJoinDispatcher` name: 

¯\_(ツ)_/¯

Doesn't make much of a difference to me - it doesn't use fork-join mechanics, but it's meant to be familiar. I considered calling it `DedicatedThreadPoolDispatcher`.

PR #766, opened 2015-03-29
Small optimization in ActorCell: Create behavior stack with size 1

HCanber:
This PR creates the behavior stack with capacity 1, i.e. `new Stack<Receive>(1)` instead of the initial size 0.
It also makes the stack private and renames it.

### Explanation
The behavior stack will _always_ contain at least 1 item, as all actors, the first thing they do, is to call Become, which pushes an item to the stack.

On first call to `Become`, as the stack has 0 capacity, a new array with size 4 is created.
As most actors do not use `Become`-stacked to push more items on the stack, this means we're wasting bytes, and allocating an empty array that's not used for long.

If we create the behavior stack with initial size 1, we avoid the creation of a new array on the first `Become`-call and for most common case the underlying array isn't larger than it has to be.

### Note
This is really a micro optimization. We might need to rethink the behavior stack and optimize for that most actors don't use it
One question for the future: As most actors don't use Become-stacked, is the behavior stack something that should exist in `ActorCell` at all? 
Or should we push the functionality down to the actor class instead?

### DIFF
--- a/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
+++ b/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
@@ -115,7 +115,7 @@ namespace Akka.Actor
 
         internal void ReceiveMessage(object message)
         {
-            var wasHandled = _actor.AroundReceive(behaviorStack.Peek(), message);
+            var wasHandled = _actor.AroundReceive(_behaviorStack.Peek(), message);
 
             if (System.Settings.AddLoggingReceive && _actor is ILogReceive)
             {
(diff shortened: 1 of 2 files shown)


rogeralsing:
Agree, there are a some features that are not always used that we still pay for.

Would it be possible to extract watchers, watchees and behavior into a separate object, that can be optimized for the default usecase? 1 behavior, 1 watcher. And once the actor diverges from that, that default object is replaced.
Sort of like childcontainer works

rogeralsing:
There should be a questionmark in there somewhere too ;)

HCanber:
Yep, might be doable.

PR #768, opened 2015-03-29
Immutable SupervisorStrategy 

rogeralsing:
The final step in making primitives immutable using surrogates.
`SupervisorStrategy` is now back to its original form


HCanber:
Looks good.

PR #769, opened 2015-03-29
BecomeStacked instead of Become?

HCanber:
This is sort of a discussion and PR at the same time. :)

The api for using `Become` and the options to store or not store the latest handler on the stack is very strange.

It's very hard to understand what differs between these calls:
``` C#
Become(newHandler);
Become(newHandler, false);
Become(newHandler, true);
```

In fact two of the calls are equivalent. Can you see which two? 
Which stores the handler on the stack so you can restore it later?

The `IActorDsl` used for inline actors has always had `Become(handler)`, `BecomeStacked(handler)` and `UnbecomeStacked()`. It's much better.

I suggest we do the same everywhere, i.e. 
- Obsolete `Become(handler,bool)` and add `Become(handler)` and `BecomeStacked(handler)`
- Obsolete `Unbecome()` and add `UnbecomeStacked()`

And incidentally, it's exactly that this PR does. ;)

So what do you think?

### DIFF
--- a/src/core/Akka.Persistence.Tests/PersistentActorSpec.Actors.cs
+++ b/src/core/Akka.Persistence.Tests/PersistentActorSpec.Actors.cs
@@ -186,7 +186,7 @@ namespace Akka.Persistence.Tests
                     Persist(new Evt(cmd.Data + "-22"), evt =>
                     {
                         UpdateState(evt);
-                        Context.Unbecome();
+                        Context.UnbecomeStacked();
                     });
                     return true;
                 }
@@ -224,7 +224,7 @@ namespace Akka.Persistence.Tests
                     Persist(new Evt(cmd.Data + "-21"), evt =>
                     {
                         UpdateState(evt);
-                        Context.Unbecome();
+                        Context.UnbecomeStacked();
                     });
                     Persist(new Evt(cmd.Data + "-22"), UpdateStateHandler);
                     return true;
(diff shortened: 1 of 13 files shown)


rogeralsing:
I like it

Aaronontheweb:
Looks like @skotzko and I are going to have a lot of READMEs to update for Akka.NET bootcamp when V1 ships :shipit: 

Why remove `Unbecome` ? How do you become something in the previous stack?

rogeralsing:
It was replaced with `UnbecomeStacked` here.
To make the name consistent i guess.
Become
BecomeStacked
UnbecomeStacked 

As `Become` replaces, it dont have an unbecome.

Aaronontheweb:
Ah ok, this is where I was confused I thought we were obseleting `UnbecomeStacked` too. That must refer to a different implementation of the method then, right @HCanber ?

HCanber:
>That must refer to a different implementation of the method then, right @HCanber ?

I'm not sure I follow. I say nothing about obsoleteing `UnbecomeStacked`. I'm saying:

>Obsolete `Unbecome()` and add `UnbecomeStacked()`

Or is it something in the code, you mean?

rogeralsing:
@Aaronontheweb it's to make the Become/Unbecome API more readable.

```
//Old API
Become(newHandler, false);
Become(newHandler, true);
Unbecome();

//New API:
Become(newHandler);
BecomeStacked(newHandler);
UnbecomeStacked();
```

Same functionality, cleaner API.

Aaronontheweb:
Got it. Welp, ok then - in it goes.

PR #778, opened 2015-03-31
Close #327

### LINKED ISSUE #327: Implement TerminationHooks
Used as part of graceful shutdown for built-in system actors - in particular it's for actors like the `RemoteDaemon` and others who need to send final `DeathWatch` notifications during the shut-down process.

smalldave:
All I've done really is uncomment @HCanber's original code and added some tests.

I have added this

```csharp
// guardian MUST NOT lose its children during restart
{
}
```
to GuardianActor.

It isn't necessary for this PR but seemed sensible so I've added it.
Perhaps I should add this as a separate PR?


smalldave:
I'll try this again

Issue #62, opened 2014-03-03
Mono support


rogeralsing:
We need to ensure that Akka.NET works on mono.

rogeralsing:
related to #694

rogeralsing:
This works fine now. closing

PR #779, opened 2015-03-31
TerminationHooks, Close #327

### LINKED ISSUE #327: Implement TerminationHooks

smalldave:
All I've done really is uncomment @HCanber's original code and added some tests.

I have added this

```csharp
// guardian MUST NOT lose its children during restart
{
}
```
to GuardianActor.

It isn't necessary for this PR but seemed sensible so I've added it.
Perhaps I should add this as a separate PR?

### DIFF
--- a/src/core/Akka.Remote/RemoteDaemon.cs
+++ b/src/core/Akka.Remote/RemoteDaemon.cs
@@ -103,7 +103,7 @@ namespace Akka.Remote
             {
                 var addressTerminated = (AddressTerminated) message;
                 //stop any remote actors that belong to this address
-                ForeachActorRef(@ref =>
+                ForEachChild(@ref =>
                 {
                     if(@ref.Parent.Path.Address == addressTerminated.Address) _system.Stop(@ref);
                 });
--- a/src/core/Akka.Tests/Actor/SystemGuardianTests.cs
+++ b/src/core/Akka.Tests/Actor/SystemGuardianTests.cs
@@ -1,0 +1,59 @@
+﻿using System.CodeDom;
+using Akka.Actor;
+using Akka.Dispatch.SysMsg;
+using Akka.TestKit;
+using Akka.TestKit.TestActors;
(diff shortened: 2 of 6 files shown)


smalldave:
On thing I'm not sure about. I've terminated the TestActor in one of the tests.
It works fine but is it sensible?

HCanber:
Regarding `PreRestart` on `GuardianActor` (used for Root guardian and User guardian). When should any of these be restarted?

>It works fine but is it sensible?

It works, although a bit strange. Why not create a `Akka.TestKit.TestActors.BlackHoleActor`.

Also see my on  question https://github.com/akkadotnet/akka.net/issues/327#issuecomment-88187372

Would you mind amending and fixing the commit message and then force-push. A sentence mentioning TerminationHooks or something is enough. I like the commit history to say something about what/why.
From [Contributing.md](https://github.com/akkadotnet/akka.net/blob/dev/CONTRIBUTING.md):
>__Descriptive commit messages__ If a commit's message isn't descriptive, change it using [interactive rebase](https://help.github.com/articles/about-git-rebase). Refer to issues using `#issue`. Example of a bad message ~~"Small cleanup"~~. Example of good message: _"Removed Security.Claims header from FSM, which broke Mono build per #62"_. Don't be afraid to write long messages, if needed. Try to explain _why_ you've done the changes. The Erlang repo has some info on [writing good commit messages](https://github.com/erlang/otp/wiki/Writing-good-commit-messages).

/Commit Police :)

smalldave:
I'll make the changes and amend the comments. Focusing too much on automatically closing the issue after I found out that works :)
No idea about restarting Guardians. If it won't happen I'll remove the code.

smalldave:
Changes made

PR #780, opened 2015-03-31
Remoting terminator

smalldave:
This is for review really.

This depends on #779 so ignore that.

It appears to work but I don't have any tests. 
I can look through the logs and see the expected messages when running tests.
I'm not really sure how to go about testing this. I'd like to test the behaviour (i.e. Remoting shuts down correctly) but I'm not sure where to start on that.
Any ideas?

### DIFF
--- a/src/core/Akka.Remote/RemoteActorRefProvider.cs
+++ b/src/core/Akka.Remote/RemoteActorRefProvider.cs
@@ -43,7 +43,7 @@ namespace Akka.Remote
                 return _internals ??
                        (_internals =
                            new Internals(new Remoting(_system, this), _system.Serialization,
-                               new RemoteDaemon(_system, RootPath / "remote", SystemGuardian, _log)));
+                               new RemoteDaemon(_system, RootPath / "remote", SystemGuardian, _remotingTerminator, _log)));
             }
         }
 
@@ -86,8 +86,8 @@ namespace Akka.Remote
             _local.UnregisterTempActor(path);
         }
 
-        //TODO: Why volatile?
-        private IActorRef _remoteWatcher;
+        private volatile IActorRef _remoteWatcher;
+        private volatile IActorRef _remotingTerminator;
 
         public virtual void Init(ActorSystemImpl system)
         {
@@ -95,7 +95,12 @@ namespace Akka.Remote
 
             _local.Init(system);
 
-            //TODO: RemotingTerminator
+            _remotingTerminator =
+                system.SystemActorOf(
+                    RemoteSettings.ConfigureDispatcher(Props.Create(() => new RemotingTerminator(_local.SystemGuardian))),
+                    "remoting-terminator");
+
+            _remotingTerminator.Tell(RemoteInternals);
 
             Transport.Start();
             _remoteWatcher = CreateRemoteWatcher(system);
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
@smalldave what was the multi-node test that kept failing as a result of not having this? We can use that as a simple test for now, but ultimately we are going to need the Akka.Remote multi-node tests too IIRC.

Here's one that might be able to help: https://github.com/akka/akka/blob/master/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteNodeShutdownAndComesBackSpec.scala

smalldave:
That looks like the test I want.
Cluster one is RestartNodeSpec but it relies on Clustering shutting down as well so I'll start on remoting test.

Aaronontheweb:
@smalldave It'll be a bit of work to get the MultiNodeTestRunner to do both of these multi-node assemblies correctly (although I think the stuff I added to the FAKE script should make this easier) but the long-term benefits will be huge. I'd be happy to follow your lead here and help out.

smalldave:
@Aaronontheweb How far did you get with getting throttling to work. When I send a ThrottleMsg I am getting 

`Message missing required fields: failure`

from 
```csharp
 .With<ThrottleMsg>(
                        throttle =>
                            w.SetFailure(
                                InjectFailure.CreateBuilder()
                                    .SetAddress(Address2Proto(throttle.Target))
                                    .SetDirection(Direction2Proto(throttle.Direction))
                                    .SetRateMBit(throttle.RateMBit)))
```
in MsgEncoder.cs

Mean anything to you.
Looks like it is from code generated by protogen. 
Out of interest how is that generated?

Aaronontheweb:
@smalldave that's weird - not sure where that's coming from. The `ThrottleAdapter` for transports definitely works now though - tested that pretty thoroughly.

smalldave:
@Aaronontheweb. It's definitely comes from the protobuf library. Not sure where to go with this.
How did you create TestConductorProtocol in the first place?

Aaronontheweb:
@smalldave used the google C# protobuf generator tool - did they update the .proto file recently?

smalldave:
Closing. Will create a new PR with test/s included.

 Issue with Throttling was in MsgEncoder. Error said failure field missing. I assumed it was on the Wrapper class. Was actually on the InjectFailure object.
Long way of saying it's sorted now.

Aaronontheweb:
@smalldave do you have these changes pushed to your branch anywhere? we should pull these in.

smalldave:
@Aaronontheweb I'll try and push up everything that is relevant.
FYI my understanding of what the RemotingTerminator does is that it stops all remotely deployed actors (i.e. those created using DaemonMsgCreate).
I don't think it will (and as I've ported it it doesn't) fix the issue with remote systems failing to associate when you bring one back up (I think that has to do with the AkkaProtocolActor dying on the side that hasn't gone down, current theory anyway).

smalldave:
BTW I don't really understand what DaemonMsgCreate is used for. Reading what I've written it doesn't entirely make sense. Either way I'm pretty confident it isn't responsible for shutting down remoting infrastructure.

smalldave:
@Aaronontheweb check out RemotingTerminator branch on my fork.
Also RestartNodeSpec branch has cluster test for #918

PR #781, opened 2015-03-31
Rename interfaces

HCanber:
Renames a few interfaces specified in #652 and prefix them with `I`.

Should we keep the commits, or should they be squashed? I'm guessing merge conflicts might be more frequent with more commits, but handling the merge conflicts are easier if the commit doesn't change too many things. Let me know. :)


Aaronontheweb:
I'll pull this one in after the others that are currently staged, but looks good

Issue #765, opened 2015-03-29
Remove Guard.Assert and throw exceptions manually


HCanber:
I've looked at how `Guard.Assert` is being used and I don't like it.

Here are a few examples:

``` C#
1. Guard.Assert(!IsEmpty, string.Format("Can't get node for [{0}] from an empty node ring", key));
2. Guard.Assert(TryParseLogMessage(messageStr, out log), "could not parse log message: " + messageStr);
```

Whats wrong? Compare to what they would have looked like without `Guard.Assert`:
``` C#
1. if(!IsEmpty) throw new ArgumentException(string.Format("Can't get node for [{0}] from an empty node ring", key));
2. if(!TryParseLogMessage(messageStr, out log)) throw new ArgumentException("could not parse log message: " + messageStr));
```

In the `Guard.Assert` case a string concatenation and allocation (and in _1._ a `string.Format`) is done __every__ time, compared to only when needed in the if-throw case.

Worse, for _1:_ it shouldn't be an `ArgumentException` at all. It should be `InvalidOperationException`. Not to mention an extra function call is made. In some places of the core framework we don't want that.

I don't want to see this spread.

The scala variant `require` is used for static code analysis tools. That's one reason they're not using if-throw. In scala the expression that's used to create the error message get's compiled into a new class with an `string apply()` function that returns the message, so the expression isn't evaluated until needed.

Do we really need `Guard.Assert`? 
If we keep it where is it ok to use it? 
How do we guarantee that string.Format and string allocations aren't used?

rogeralsing:
I agree, it's better if we can be consistent through the entire code base.
Better to promote correct exception types and consistency over the small benefit of a DbC DSL that is just a few characters shorter than the native boilerplate code.

I did violate this in my old Type.GetType .. != null PR that we reverted for this reason.

Aaronontheweb:
> Do we really need Guard.Assert? 
If we keep it where is it ok to use it? 
How do we guarantee that string.Format and string allocations aren't used? 

`Guard.Assert` is really meant for two things:
1. Future-proofing against standardizing on exception types - i.e. if we want to wrap all of these argument / invalid blah blah exceptions into some type derived from `AkkaException`.
2. Convenience for the developer.

>Better to promote correct exception types and consistency over the small benefit of a DbC DSL that is just a few characters shorter than the native boilerplate code.

[...]

PR #784, opened 2015-04-01
Issue #765

KadekM:
In first commit I removed Guard which would never throw.

In second commit I removed guards and changed them to manual exception throwing.

Should I also remove Guard class since it's now no longer in use ?


KadekM:
Forgot  to read guidelines, sorry! Will re-post later.

PR #785, opened 2015-04-01
Issue #765

KadekM:
In first commit I removed Guard which would never throw.

In second commit I removed guards and changed them to manual exception throwing.

Should I also remove Guard class since it's now no longer in use ?

### DIFF
--- a/src/core/Akka.Remote/DeadlineFailureDetector.cs
+++ b/src/core/Akka.Remote/DeadlineFailureDetector.cs
@@ -28,7 +28,7 @@ namespace Akka.Remote
         {
             _acceptableHeartbeatPause = acceptableHeartbeatPause;
             _acceptableHeartbeatMillis = Convert.ToInt64(acceptableHeartbeatPause.TotalMilliseconds);
-            Guard.Assert(_acceptableHeartbeatPause > TimeSpan.Zero, "acceptable-heartbeat-pause must be greater than zero");
+            if (_acceptableHeartbeatPause <= TimeSpan.Zero) throw new ArgumentException("acceptable-heartbeat-pause must be greater than zero");
         }
 
         /// <summary>
--- a/src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
+++ b/src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
@@ -175,7 +175,6 @@ namespace Akka.Cluster.Routing
                 throw new ConfigurationException("Resizer can't be used together with cluster router.");
             _settings = settings;
             _local = local;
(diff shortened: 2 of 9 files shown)


Aaronontheweb:
looks good to me - how about you @HCanber ?

src/core/Akka.Remote/DeadlineFailureDetector.cs
```diff
@@ -28,7 +28,7 @@ public DeadlineFailureDetector(TimeSpan acceptableHeartbeatPause, Clock clock =
         {
             _acceptableHeartbeatPause = acceptableHeartbeatPause;
             _acceptableHeartbeatMillis = Convert.ToInt64(acceptableHeartbeatPause.TotalMilliseconds);
-            Guard.Assert(_acceptableHeartbeatPause > TimeSpan.Zero, "acceptable-heartbeat-pause must be greater than zero");
+            if (_acceptableHeartbeatPause <= TimeSpan.Zero) throw new ArgumentException("acceptable-heartbeat-pause must be greater than zero");
```
rogeralsing: If the value refers to a setting in a config, the type should be `Akka.Configuration.ConfigurationException`
The hyphen'ed name in the message hints that this is the case.

rogeralsing:
Looks good, IMO we should probably supply parameter names to the Argument*Exceptions, not just bake them into the exception message, as we do provide param names in the rest of the codebase.

src/core/Akka.Remote/DeadlineFailureDetector.cs
Aaronontheweb: Ah yes, that's correct - should be an `Akka.Configuration.ConfigurationException`. Could you please change that @KadekM ?

KadekM:
I will fix it ASAP (today / tommorow) . So far thank you for review. I'm kind of new to open source.

Aaronontheweb:
thanks @KadekM! No worries - most of the people working on this project were new to open source before they started sending PRs ;) you're in good company.

KadekM:
Great :+1:
So should I also delete Akka.Util.Guard since it's now no longer in use ?

KadekM:
Test is passing on my machine - also I don't suppose change in my commit would cause this fail, is the test somewhat unstable ?

KadekM:
Ok, I ammended my commit (added second parameters to Argument*Exception) and now two different tests are failing... what's going on ? :)

Aaronontheweb:
@KadekM racey unit tests - not your fault. I'll re-run the build.

KadekM:
Ok. Is there anything else need to be done?

Aaronontheweb:
@KadekM nope, good to go! thanks for your pr!

PR #786, opened 2015-04-02
Update Examples of use cases for Akka.md

codedemonuk:
Removed stray heading that appeared at the and of the page.


rogeralsing:
Thanks!

PR #795, opened 2015-04-04
Bugfix - unused supervisorStrategy in props create is fixed

KadekM:
I believe I found a tiny bug with unused strategy argument passed to Props.Create in Props.cs.

Tests demonstrate expected (at least by me) functionality - on upstream `Props_created_with_strategy_must_have_it_set` would not pass.

After merge both tests should pass.


rogeralsing:
@akkadotnet/developers Why do we even have `SupervisorStrategy` on props?
Java Akka doesnt.

rogeralsing:
Thanks!

PR #805, opened 2015-04-07
Removing docs _again_

rogeralsing:
Removing documentation from the main repo as this was generating too much noise (issues and PR's)
Documentation now lives in the GetAkka.NET repo together with all other site resources.


stefansedich:
@rogeralsing I don't seem to have permissions to kick off another build, you mind doing it?

Aaronontheweb:
@rogeralsing could you please undelete the logo and the sandcastle build file? need both of those for the API docs

Issue #802, opened 2015-04-07
Are there a few more interfaces left to change?


SeanKilleen:
A continuation of the work done in #652, #782, and #801.

Hi all, I did a quick regex search in VS for `interface [^Ii]` and the following still appear to be left.

Is there any guidance for how to determine which (if any) of these should be / should not be changed?

-  [x] `EventFilter` (`Akka.Event` namespace)
-  [x]  `EventFilterApplier` (`Akka.TestKit`)
-  [x]  `UnmutableFilter` (`Akka.TestKit`)
-  [x]  `NoImplicitSender` (`Akka.TestKit`)
-  [ ]  `TestKitAssertions` (`Akka.TestKit`)
-  [x]  ~~`T` (`Akka.Tests.Event`)~~
-  [x]  ~~`AT` (`Akka.Tests.Event`)~~
-  [x]  ~~`ATT` (`Akka.Tests.Event`)~~
-  [x]  ~~`BT` (`Akka.Tests.Event`)~~
-  [x]  ~~`BTT` (`Akka.Tests.Event`)~~
-  [ ]  `ActorRefFactory` (`Akka.Actor`)

Aaronontheweb:
The T / AT / ATT ones you can leave alone - looks like they're specific to a single unit test.

As for the others - we should change those. Any objections @HCanber @rogeralsing ?

SeanKilleen:
With the merge of #810, this is considered closed.

PR #810, opened 2015-04-07
Actually the last interface renames. :)

SeanKilleen:
In support of #802, the last of the interfaces.

- [x] `TestKitAssertions` (`Akka.TestKit`)
- [x] `ActorRefFactory` (`Akka.Actor`)


SeanKilleen:
@Aaronontheweb assuming that last commit checks out, you can merge this and close the related issue.

PR #815, opened 2015-04-07
added first draft of V1 release notes

Aaronontheweb:
## Needs review

I spent some time today organizing all of the major PRs that introduced breaking changes in V1.0 and tried to capture them all in `RELEASE_NOTES.md`.

I need some help from the following people:

@akkadotnet/fsharpteam and @Horusiath - what are the major F# changes that were introduced? Could you add a blurb in the comments here?

@HCanber did I miss anything related to interface / scheduler API changes?

@rogeralsing could you add a blurb to the comments here for `async` / `await` support for the release notes so I can capture it correctly?

Please add any other line notes to the file included in this PR and I'll update it accordingly.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -107,8 +107,8 @@ var actorSystem = ActorSystem.Create("MySystem"); //automatically loads App/Web.
 **Dispatchers**
 Akka.NET v1.0 introduces the `ForkJoinDispatcher` as well as general purpose dispatcher re-use.
 
-##### Using ForkJoinDispatcher
-So ForkJoinDispatcher is special - it uses [`Helios.Concurrency.DedicatedThreadPool`](https://github.com/helios-io/DedicatedThreadPool) to create a dedicated set of threads for the exclusive use of the actors configured to use a particular `ForkJoinDispatcher` instance. All of the remoting actors depend on the `default-remote-dispatcher` for instance.
+**Using ForkJoinDispatcher**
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Also for @HCanber - any TestKit changes that need to be captured?

rogeralsing:
Sure.

## Async Await Support

`ReceiveActor`s now support Async/Await out of the box.

```csharp
public class MyActor : ReceiveActor
{
       public MyActor()
       {
             Receive<SomeMessage>(async some => {
                    //we can now safely use await inside this receive handler
                    await SomeAsyncIO(some.Data);
                    Sender.Tell(new EverythingIsAllOK());                   
             });
       }
}
```

It is also possible to specify the behavior for the async handler, using `AsyncBehavior.Suspend` and  `AsyncBehavior.Reentrant` as the first argument.
When using `Suspend` the normal actor semantics will be preserved, the actor will not be able to process any new messages until the current async operation is completed.
While using `Reentrant` will allow the actor to multiplex messages during the `await` period.
This does not mean that messages are processed in parallel, we still stay true to "one message at a time", but each await continuation will be piped back to the actor as a message and continue under the actors concurrency constraint.

However, `PipeTo` pattern is still the preferred way to perform async operations inside an actor, as it is more explicit and clearly states what is going on.

---

Do we need more on this?

Aaronontheweb:
@rogeralsing looks good - thanks!

Aaronontheweb:
Sending a new PR on a rebased branch.

Issue #818, opened 2015-04-08
Docs: Move from CloudCopy --> AzCopy


SeanKilleen:
[AzCopy](http://azure.microsoft.com/en-us/documentation/articles/storage-use-azcopy/) is Microsoft's tool for copying files to Azure storage blog.

I think the Azure PowerShell tools unfortunately mostly support copying between blobs.

As of now, I recommend this instead of Azure PowerShell tools.

### Build Server Changes
- [x] Move [Azure Storage Tools](http://azure.microsoft.com/en-us/documentation/articles/storage-use-azcopy/) into the tools directory

### Code Changes
- [x] Understand current syntax of build process for copying items using AzCopy
- [x] Replace CloudCopy call with a command-line call to AzCopy

### Testing
Not sure I can really help out with this one since it has to be done on the build server. I'll do my best to launch the command locally and at least ensure that it fails in an understandable way.

Aaronontheweb:
@SeanKilleen if we can make this work with a local copy of the AzCopy binary, per #819, then that'll make life a lot easier all around :p

SeanKilleen:
@Aaronontheweb 100% agreed, already put that into #819 for that exact reason. Wasn't thinking it through during my initial write-up.

Aaronontheweb:
Still have some bugs I need to sort out with AzCopy - setting right content-headers and such.

SeanKilleen:
@Aaronontheweb sorry that I don't have more visibility into what you're seeing. If you want to post a little more about it here (error messages, general background, etc.) I can try to do some legwork on it.

Aaronontheweb:
no error messages - just not seeing the results I expect inside Azure blob storage.

Saw two problems basically:

1. Some content didn't look like it was being overwritten correctly (replacing an existing blob with a local file from source)
2. Uploaded content had the incorrect MIME type

Just closed a PR that fixes problem 2 - going to see if the issue with problem 1 was related to something else I was doing last night.

Aaronontheweb:
@SeanKilleen turns out that the content not being overwritten correctly is an Azure CDN issue... I need to find a way to set a better default content-cache expiration header.

Might just disable the CDN for now and serve everything directly from blob storage.

Aaronontheweb:
But otherwise, everything else is working.

Aaronontheweb:
[...]

Issue #811, opened 2015-04-07
Need to create local build of CloudCopy to fix build server issue


Aaronontheweb:
IT'S ALWAYS SOMETHING :(

[CloudCopy](http://cloudcopy.codeplex.com/), the tool we use to uploading Azure hasn't been updated since 2011. And apparently it's not very build-server friendly.

[20:24:06][Step 1/1] Unhandled Exception: System.InvalidOperationException: Cannot see if a key has been pressed when either application does not have a console or when console input has been redirected from a file. Try Console.In.Peek.
[20:24:06][Step 1/1]    at System.Console.get_KeyAvailable()
[20:24:06][Step 1/1]    at CloudCopy.Commands.UploadCommand.Execute()
[20:24:06][Step 1/1]    at CloudCopy.Program.RunCommands()
[20:24:06][Step 1/1]    at CloudCopy.Program.Main(String[] args)

Basically it's trying to do some standard I/O redirection stuff and failing: http://stackoverflow.com/questions/3961542/checking-standard-input-in-c-sharp

I think we can just wrap this in a try...catch block and resolve the issue that way. Either way we'll need to fork the source and hotfix it so our build server can properly upload its API docs to Azure.

SeanKilleen:
@Aaronontheweb would the group be open to documentation? I looked at the docs and since they're already markdown formatted, it might make sense to move them to Github Pages / Jekyll. 

I did this with [my blog](http://seankilleen.com/) [(repo)](https://github.com/SeanKilleen/seankilleen.github.io) -- it builds on the fly to static pages every checkin, all hosted on Github so I don't pay a dime. Also, it's Github's server, so as Zach Holman puts it, "when my site goes down, Github gets paged."

I know it doesn't solve this specific issue, but is there interest in me branching the site and exploring options around that? Seems like it could really reduce some overhead.

Aaronontheweb:
@SeanKilleen the Wiki, I.E. the documentation written by humans, is done on Github pages through drone.io - so we're covered there.

This is the other half of the documentation - the API docs, generated via reflection from all of the XML-DOC comments on the code. It's synthesized automatically by SandCastle.

We considered integrating this stuff using Github pages, but we'd quickly exceed the maximum size of a Github repository (4 gigs I think) just doing our weekly unstable doc builds.

So we decided to go with Azure Blob Storage + CDN for doing this. Simpler to do (easier to do lots of small uploads than one mega commit + push) and just as reliable.

[...]

PR #819, opened 2015-04-08
Move from CloudCopy --> AzCopy

SeanKilleen:
In support of #818, originating from conversation on #811.


SeanKilleen:
@Aaronontheweb give that a spin if you can -- it should at least get us close.

Aaronontheweb:
thanks! I'll give this a try.

PR #820, opened 2015-04-08
Script to add Copyright headers & modify files to add header

SeanKilleen:
In support of #718, this PR will contain two changes:

* A PowerShell script in the tools directory that adds copyright files to `.cs` and `.fs` files that don't have them
* A pass on every C# and F# file that adds in the new


rogeralsing:
Imo the C# line should be changed into a copyright line too, apache 2 license is clear on this matter, you own the modifications you make.

So something along the lines of 
```
Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
Copyright (C) 2013-2015 Akka.NET project <http://getakka.net/>
```

And personally, I'd rather see this github link instead of getakka.net.
As github is the actual source. people reading our source will most likely already know that we have a site.

SeanKilleen:
@rogeralsing for further reference, there appears to be [a guide here](http://www.apache.org/dev/apply-license.html). 

Most examples I've seen so far involve adding some more text -- while the script is mostly ready to go, I won't run the script across all files until we finish the discussion -- #718 is likely a better place. I'm thinking.

SeanKilleen:
@Aaronontheweb @rogeralsing ready for review.

SeanKilleen:
@rogeralsing @Aaronontheweb it occurs to me that this is a huge commit. While it builds and tests fine, would you prefer I break it up into smaller commits? I didn't see a problem at the time but let me know if you want me to revert this and commit several smaller chunks.

Aaronontheweb:
@SeanKilleen PR is fine as-is. I'm just waiting on getting the API docs stuff working perfectly before I pull this in, just because I'm going to have to make some on-the-fly adjustments to that script I suspect. I'll pull this in within a couple of hours unless @akkadotnet/contributors object :p

Aaronontheweb:
err mean @akkadotnet/developers

PR #825, opened 2015-04-09
F# DU serialization fix

rogeralsing:
F# DU serialization support.

The test case @Horusiath provided now passes.
We do need more tests for this to see if there are any special cases where this does not work.
But it solves the standard case where you pass a single DU as a message.

Can we live with such solution untill Json.NET comes out with a true fix for this?

### DIFF
--- a/src/core/Akka/Serialization/NewtonSoftJsonSerializer.cs
+++ b/src/core/Akka/Serialization/NewtonSoftJsonSerializer.cs
@@ -129,10 +129,12 @@ namespace Akka.Serialization
                     var caseTypeName = j["Case"].Value<string>();
                     var caseType = type.GetNestedType(caseTypeName);
                     var ctor = caseType.GetConstructors(BindingFlags.NonPublic | BindingFlags.Instance)[0];
-                    //var values = ????? j["Fields"]["$values"].Values<object>();
-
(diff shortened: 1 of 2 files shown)


src/core/Akka/Serialization/NewtonSoftJsonSerializer.cs
```diff
@@ -122,6 +123,17 @@ private static object TranslateSurrogate(object deserializedValue,ActorSystem sy
...
                     return GetValue(value);
                 }
+
+                if (j["Case"] != null)
+                {
+                    var caseTypeName = j["Case"].Value<string>();
+                    var caseType = type.GetNestedType(caseTypeName);
+                    var ctor = caseType.GetConstructors(BindingFlags.NonPublic | BindingFlags.Instance)[0];
+                    //var values = ????? j["Fields"]["$values"].Values<object>();
+
```
rogeralsing: If we can fetch the raw objects from the JObject here, we are good to go.

rogeralsing:
Added test for nested DU types

Aaronontheweb:
Running the multinode test suite locally and did all of the rebase / dance stuff in my local fork. Will submit a clean PR shortly

Aaronontheweb:
Refer to #827

Issue #831, opened 2015-04-10
v1.0 F# scheduling API not sending any scheduled messages


raymens:
Updating the Akka.net library from 0.8 to 1.0 seems to be breaking the F# scheduler. 
The actor doesn't receive any messages, which worked fine on 0.8.
Had to do a small change because of a breaking API change in 1.0 from:
```
    system.Scheduler
    |> scheduleTell (TimeSpan.Zero) (TimeSpan.FromSeconds(30.)) () fetchRef
    |> ignore
```
To:
```
system.Scheduler.ScheduleTellRepeatedly(TimeSpan.Zero, TimeSpan.FromSeconds(10.), fetchRef, ())
```

The following stripped down code can be used to possibly reproduce it:
```
open Akka.FSharp
open System

let system = System.create "NotificationWorker" (Configuration.defaultConfig())
let handleMsg msg = printfn "Hi!"
let fetchRef = spawn system "NotificationFetch" (actorOf handleMsg)
system.Scheduler.ScheduleTellRepeatedly(TimeSpan.FromSeconds(5.), TimeSpan.FromSeconds(10.), fetchRef, ())
```

Horusiath:
Internal construction of `ScheduleTellRepeatedly` (and possibly other F# method overrides too) doesn't call underlying C# method implementation having the same name, going into infinite recursive loop call instead.

Horusiath:
Fixed by #834

raymens:
> EDIT: I see that,  ITellScheduler  extensions inside F# API could be dropped completely.

Yeah since 1.0 the difference between the C# and F# version is minimal.

PR #834, opened 2015-04-10
removed ITellScheduler extensions from F# API

Horusiath:
This should fix an #831 issue. New implementations of `ScheduleTellOnce` and `ScheduleTellRepeatedly` are now ripped off from optional parameters - they provide a minimal set of parameters required by their previous versions, while more parametrized ones are still available from akka core `ITellScheduler` implementation.

### DIFF
--- a/src/core/Akka.FSharp/Schedulers.fs
+++ b/src/core/Akka.FSharp/Schedulers.fs
@@ -47,15 +47,8 @@ type Akka.Actor.ITellScheduler with
     /// <param name="every">Interval.</param>
     /// <param name="message">Message to be sent to the receiver by the scheduler.</param>
     /// <param name="receiver">Message receiver.</param>
-    /// <param name="sender">Optional actor reference set up as message sender</param>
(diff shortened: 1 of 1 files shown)


rogeralsing:
We should add tests for this imo

Aaronontheweb:
@rogeralsing @Horusiath does that mean that we need to add them to this PR? Or do you want to create a separate issue for that?

rogeralsing:
IMO we should at least have a test that proves the broken bits are now fixed for each bug fix PR.
also the entire F# API is weak on tests, we should have tests for each function of the F# API, but that could be a separate task to improve on.

Horusiath:
Ok, I'll update PR.

Horusiath:
Ok, I've realized it's quite cumbersome without F# support for akka testkit. Can we get this one up (I've tested in localy on some example) and cover tests after TestKit.FSharp plugin will be out?

Aaronontheweb:
@Horusiath who's doing the TestKit for FSharp?

Horusiath:
Me... eventually.

PR #835, opened 2015-04-11
Patch for Issues 832 & 833

jcwrequests:
Based @rogeralsing  suggestion I removed the using a string to produce the actor and am instead using the type. This should resolve the issues with 832 & 833. @nvivo @danthar could you please check this version to see if it resolves your issues.  Thanks

### DIFF
--- a/src/contrib/dependencyInjection/Akka.DI.AutoFac/AutoFacDependencyResolver.cs
+++ b/src/contrib/dependencyInjection/Akka.DI.AutoFac/AutoFacDependencyResolver.cs
@@ -73,11 +73,10 @@ namespace Akka.DI.AutoFac
         /// </summary>
         /// <param name="actorName">Name of the ActorType</param>
         /// <returns>factory delegate</returns>
-        public Func<ActorBase> CreateActorFactory(string actorName)
+        public Func<ActorBase> CreateActorFactory(Type actorType)
         {
             return () =>
             {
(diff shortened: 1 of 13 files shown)


Aaronontheweb:
@jcwrequests looks good to me. If @nvivo or @danthar have anything to add they can chime in here. Will pull this in on Monday.

nvivo:
Looks good by reading,  but I couldn't test it yet. Would be nice to have a new "rc" on nuget before releasing the next version,  so we can better check this and the other issues.

One note is that I think there are some comments or text files in a folder somewhere in the tree that explains how to create a resolver that might need an update after this.

Aaronontheweb:
@nvivo we had a preview release of this up on NuGet about a week before the official release http://www.nuget.org/packages/Akka/1.0.0-dev1504032244

Each PR, when successful, also generates NuGet packages you can download as artifacts off of TeamCity.

But you're right - for longer term releases we should probably put weekly pre-releases out there rather than depend on people compiling from source.

Aaronontheweb:
And thanks for doing your review too @nvivo - much appreciated!

jcwrequests:
@nvivo @aaronontheweb you are probably right about the docs I will double check. BTW @nvivo sorry about the earlier confusion about your suggestion. I was in the middle of doing work around the house and was trying to review everything from my phone.

PR #836, opened 2015-04-11
Fix for async await suspend-resume mechanics

rogeralsing:
Hotfix for async await suspend-resume mailbox mechanics.


rogeralsing:
cc @nvivo

nvivo:
Thanks!

Aaronontheweb:
> There is really no excuse for this one, we rushed the v1 release.There is really no excuse for this one, we rushed the v1 release.

We had several weeks to find this bug from the time the PR was accepted to when the NuGet package was pushed. The release was definitely not rushed. But bugs happen - not a big deal. Don't turn it into a fault-finding mission.

Issue #667, opened 2015-02-21
Plugin proposal: Akka.Persistence.SqlServer


Horusiath:
Add **Journal** and (optionally) **SnapshotStore** implementations using SqlServer database on the backend. Since SqlServer is considered an important part of a .NET dev culture, I think it should be placed in contrib lib with support from Akka.NET community.

- It also should provide some easy way to check and run necessary database schema update if required.
- I think that it should support SqlServer 2008 and above.

Minimal Journal configuration necessary to make plugin work with Akka.Persistence correctly:
```hocon
akka.persistence.journal {
  plugin = "akka.persistence.journal.sql-server"
  sql-server {
      class = "Akka.Persistence.SqlServer.SqlServerJournal, Akka.Persistence.SqlServer"
      plugin-dispatcher = "akka.actor.default-dispatcher"
      
      # connection string to SqlServer database
      connection-string= "Server=myServerAddress;Database=myDataBase;User Id=myUsername;"
  }
}
```

In case if plugin will also implement SnapshotStore based on SqlServer, this is minimal configuration necessary to make plugin work with Akka.Persistence correctly:
```hocon
akka.persistence.snapshot-store {
  plugin = "akka.persistence.snapshot-store.sql-server"
  sql-server {
    class = "Akka.Persistence.SqlServer.SqlServerSnapshotStore, Akka.Persistence.SqlServer"
    plugin-dispatcher = "akka.persistence.dispatchers.default-plugin-dispatcher"

     # connection string to SqlServer database
     connection-string= "Server=myServerAddress;Database=myDataBase;User Id=myUsername;"
  }
}
```
Integration tests can be done fairly simply using `Akka.Persistence.TestKit` library (see [Akka.Persistence.TestKit.Tests](https://github.com/akkadotnet/akka.net/tree/dev/src/core/Akka.Persistence.TestKit.Tests) for more details).

References:

- [MemoryJournal](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Persistence/Journal/MemoryJournal.cs) implementation based on `AsyncWriteProxy`. For SQL databases more adviced is to use `SyncWriteJournal` instead (see canonical Akka's [LevelDbJournal](https://github.com/akka/akka/tree/master/akka-persistence/src/main/scala/akka/persistence/journal/leveldb))

Horusiath:
Completed by #842 and #901

PR #842, opened 2015-04-12
Akka.Persistence.SqlServer plugin

Horusiath:
This PR introduces plugin with support for persistent Journal and Snapshot Store backed by SqlServer database.

- Setup and configuration details are available in README.md. Public API is commented.
-  ~~Test cases by default assume that test database (named: `AkkaPersistenceSqlServerSpecDb`) already exists on `.\SQL-EXPRESS` server~~ Tests uses `AkkaPersistenceSqlServerSpecDb.mdf` database file located in Akka.Persistence.SqlServer.Tests/Resources directory.
- Also `Akka.Persistence.TestKit` base specs in form of `JournalSpec` and `SnapshotStoreSpec` has been changed to require to run `Initialize()` method at the end of spec constructor in order to properly initialize the test cases **[BREAKING CHANGE]**.

cc #667


src/contrib/persistence/Akka.Persistence.SqlServer.Tests/SqlServerSnapshotStoreSpec.cs
```diff
@@ -0,0 +1,37 @@
...
+            publish-plugin-commands = on
+            snapshot-store {
+                plugin = ""akka.persistence.snapshot-store.sql-server""
+                sql-server {
+                    class = ""Akka.Persistence.SqlServer.Snapshot.SqlServerSnapshotStore, Akka.Persistence.SqlServer""
+                    plugin-dispatcher = ""akka.actor.default-dispatcher""
+                    table-name = SnapshotStore
+                    schema-name = dbo
+                    auto-initialize = on
+                    connection-string = ""Data Source=.\\SQLEXPRESS;Initial Catalog=AkkaPersistenceSqlServerSpecDb;Integrated Security=True""
```
rogeralsing: This needs to be configured somehow. we can't expect the build serer and users to have this specific setup.

src/contrib/persistence/Akka.Persistence.SqlServer/Journal/QueryBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        public DefaultJournalQueryBuilder(string tableName, string schemaName)
+        {
+            _tableName = tableName;
+            _schemaName = schemaName;
+
+            _insertMessagesSql =
+                string.Format("INSERT INTO [{0}].[{1}] (PersistenceID, SequenceNr, IsDeleted, PayloadType, Payload) VALUES (@PersistenceId, @SequenceNr, @IsDeleted, @PayloadType, @Payload)",
+                    _schemaName, _tableName);
+            _selectHighestSequenceNrSql = string.Format(@"SELECT MAX(SequenceNr) FROM [{0}].[{1}] WHERE CS_PID = CHECKSUM(@pid)", _schemaName, _tableName);
```
rogeralsing: Maybe a non-issue but this is open for sql injection incase someone manages to manipulate tableNAme or schemaName.
(if that is the case the user is already in trouble, but maybe we should throw some validation at this just to be sure)

rogeralsing:
IMO, just to  be on the safe side, we should escape schema and table names when building queries.
This might be the correct way `SqlCommandBuilder.QuoteIdentifier` : https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlcommandbuilder.quoteidentifier%28v=vs.110%29.aspx

rogeralsing:
As for configuration settings on connection strings, I guess we should complete the substitution support in HOCON.

Aaronontheweb:
.SLN file currently has an MSBUILD failure according to TeamCity - need to update PR accordingly

jcwrequests:
@Horusiath @rogeralsing Do you think it makes sense to also provide guidance on protecting the app config file? Whenever I deploy anything that has something sensitive like connection strings I always encrypted the config. I know other frameworks don't bother since it most shops that's an IT responsibility but that is not always the case. For the last 15 plus years I have been not only responsible for development but operations end and am alway concern about security. I am just sharing my thoughts. Once this has settled down I would love to see if that same thing could be done with Postgres and Cassandra.

Aaronontheweb:
Build Server can't run MSBuild at the moment, according to the error log. Would you mind taking a look at the .SLn @Horusiath ?

Horusiath:
@Aaronontheweb If fixed a sln file and build.cmd didn't notice any problems, so I guest it should be ok now.

rogeralsing:
Can this be merged or do we have to wait for sql support to be installed on the build server?

Aaronontheweb:
@rogeralsing I need some instructions from @Horusiath on how to update the build agents first

Horusiath:
I've updated FAKE script to check if Local DB is installed on host machine and to skip Akka.Persistence.SqlServer.Tests in case when it's not. cc #891

Horusiath:
As discussed with @Aaronontheweb I've created a separate build task (`RunSqlServerTests`) in FAKE for Akka.Persistence.SqlServer. It was possible to automatically recognize if Local DB is installed on host but it could potentially cause some problems on Linux. So right now it's a explicit choice of the user running a build script.

Aaronontheweb:
looking good buddy - looking forward to seeing this hit the airwaves in v1.0.1

Aaronontheweb:
@Horusiath one minor thing - I believe we need a .nuspec file for this :p

PR #843, opened 2015-04-12
Hotfix async await when awaiting IO completion port based tasks

rogeralsing:
Another hotfix for async await.
Turns out when using the `Suspend` behavior and awaiting tasks that complete on IO completion ports, e.g. `Task.Delay` or IO operations, the current context was lost.
This PR fixes this.


rogeralsing:
Ty @nvivo for reporting this stuff

rogeralsing:
#### Do not merge until @nvivo confirms that this solves the issues in his app.

nvivo:
The fix seems to work. Didn't see any crashes in my app anymore using ReceiveActor and the async Receive handler.

PR #845, opened 2015-04-12
Fix stopwatch ticks evalution

mj1856:
Stopwatch ticks are not the same as TimeSpan or DateTime ticks.

The `Stopwatch.ElapsedTicks` property retrieves raw timer ticks, which are of varying size depending on the capabilities of the system.  They are typically around 16 nanoseconds per tick on modern systems that have high frequency performance counters.  To use them directly, you have to consider the value of `Stopwatch.Frequency`.

It's easier just to use `Stopwatch.Elapsed`, which takes the frequency into account.  Then you can get the `Ticks` property from the resulting timespan, which will indeed have a fixed size of 100 nanos-per-tick.

So - one period changes everything  `ElapsedTicks` => `Elapsed.Ticks`.

:)

See also: https://msdn.microsoft.com/en-us/library/system.diagnostics.stopwatch.elapsedticks.aspx

### DIFF
--- a/src/core/Akka.Remote/SystemNanoTime.cs
+++ b/src/core/Akka.Remote/SystemNanoTime.cs
@@ -28,7 +28,7 @@ namespace Akka.Remote
 
         public static long GetNanos()
         {
-            return StopWatch.ElapsedTicks.ToNanos();
+            return StopWatch.Elapsed.Ticks.ToNanos();
         }
 
         internal const long NanosPerTick = 100;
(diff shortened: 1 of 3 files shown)


mj1856:
It's also worth noting that since it has been being used incorrectly, there may have been other side-effects throughout the system.  By fixing it, you may uncover timing-related errors that you didn't notice before.

mj1856:
Quite a lot of tests fail once this change is made - but it is the correct change to make.  Someone with deeper expertise of AKKA.NET should dig into this.

Aaronontheweb:
@mj1856 the failed tests may not have had anything to do with your code - looks like the XUnit test runner may have crashed while running one of the test assemblies. XUnit 1.9.2 does some weird stuff with IO interception locally and sometimes this happens. We're going to upgrade to 2.0 (just came out) which supposedly fixes this.

Going to kick off a rebuild and see what happens.

Aaronontheweb:
@mj1856 thanks for your contribution!

PR #848, opened 2015-04-13
DO NOT MERGE

rogeralsing:
Checking if the buildserver is broken due to the `CallContext` Stuff used in the Async Await support


rogeralsing:
cc @Aaronontheweb

Aaronontheweb:
So in some PRs (from latest dev snapshot) I'm seeing intermittent failures. Looks like async / await support is definitely causing some issues with XUnit 1.9.2

rogeralsing:
Closing this, I will try to find a way to solve this but I don't have too high hopes for it as the issue is deep inside xUnit

PR #854, opened 2015-04-14
Modified default Cluster.conf to use a ForkJoinDispatcher just like remo...

annymsMthd:
...te


Aaronontheweb:
I'll test this locally and see how the multi-node tests hold up. More work might be required to do some of the integrations here - not sure yet how well the clustering actors respect this value.

Issue #853, opened 2015-04-14
Fix all racy unit tests


Aaronontheweb:
We have a handful of evil unit tests inside Akka core that are either finding:

* legitimately bad behavior that we need to fix
* race conditions within specs
* race conditions within core actor code.

Time to hunt these bugs down and fix them so we can have full confidence in our test suite.

Here's the list thus far:

* [ ] DeathWatchSpec Bug209_any_user_messages_following_a_Terminate_message_should_be_forwarded_to_DeadLetterMailbox	
* [x]  ResizerSpec DefaultResizer_must_grow_as_needed_under_pressure	
* [x] ResizerSpec DefaultResizer_must_backoff	
* [x] InboxSpec Inbox_have_a_default_and_custom_timeouts	
* [x]  ConsistentHashingRouterSpec ConsistentHashingRouterMustAdjustNodeRingWhenRouteeDies	
* [ ] RoutingSpec Router_in_general_must_evict_terminated_routees	
* [x]   SupervisorHierarchySpec A_supervisor_hierarchy_must_handle_failure_in_creation_when_supervision_strategy_returns_Resume_and_Restart	
* [x] TailChoppingSpec Tail_chopping_router_must_throw_exception_if_no_result_will_arrive_within_the_given_time
* [ ] RemoteRouterSpec RemoteRouter_must_deploy_its_children_on_remote_host_driven_by_configuration

annymsMthd:
I'll try to tackle some of these now. I noticed them in my PR on TC

Aaronontheweb:
Here's a new one:

* []   SupervisorHierarchySpec A_supervisor_hierarchy_must_handle_failure_in_creation_when_supervision_strategy_returns_Resume_and_Restart

Aaronontheweb:
Two more.... http://petabridge-ci.cloudapp.net/viewLog.html?buildId=670&tab=buildResultsDiv&buildTypeId=AkkaNet_AkkaNetPrBuild?guest=1

TailChoppingSpec.Tail_chopping_router_must_throw_exception_if_no_result_will_arrive_within_the_given_time	

```
Failed: Expected a message of type Akka.Actor.Status+Failure, but received {ack} (type System.String) instead  from [akka://all-systems/]
[...]

PR #857, opened 2015-04-15
853 Fixing racy tests

annymsMthd:
Adding this as a PR already to start testing on Akka's CI.

### DIFF
--- a/src/core/Akka.Tests/Actor/InboxSpec.cs
+++ b/src/core/Akka.Tests/Actor/InboxSpec.cs
@@ -105,7 +105,7 @@ namespace Akka.Tests.Actor
                     o.ShouldBe(0);
                 }
 
-                //The inbox should be empty now, so receiving should result in a timeout             
+                //The inbox should be empty now, so receiving should result in a timeout
                 Intercept<TimeoutException>(() =>
                 {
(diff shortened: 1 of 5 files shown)


annymsMthd:
Relates to #853

src/core/Akka.Tests/Actor/InboxSpec.cs
```diff
@@ -122,11 +122,12 @@ public void Inbox_have_maximum_queue_size()
...
-            Within(TimeSpan.FromSeconds(5), TimeSpan.FromSeconds(6), () =>
+            Within(TimeSpan.FromSeconds(4), TimeSpan.FromSeconds(6), () =>
             {
                 Intercept<TimeoutException>(() => _inbox.Receive());
                 return true;
             });
+
             Within(TimeSpan.FromSeconds(1), () =>
             {
                 Intercept<TimeoutException>(() => _inbox.Receive(TimeSpan.FromMilliseconds(100)));
```
Aaronontheweb: Looks like this part of the test is still subject to race conditions. Let's take a look at the Inbox code and see if anything's up there.

Aaronontheweb:
@annymsMthd looks like that was a bug inside `Inbox` - so good job. I'll pull it in and see if this alleviates the issue. Can only tell with time.

Issue #789, opened 2015-04-02
Unreliable Save/Restore snapshot for GuaranteedDeliveryActor


aprooks:
hi, I was trying to implement a GuaranteedDelivery sample: https://github.com/aprooks/akka.net/commit/2f25bd9e3eed9deb483fef7be9fdc5bbf063470c

and found, that if you throw exception right after  saving snapshot, new instance is reloaded with an old journal, and `foo` message is lost:

```
            deliverer.Tell(new Message("foo"));
            System.Threading.Thread.Sleep(1000);  //Remove this and restores none 
            deliverer.Tell("boom");
```

Did I miss something? Is there's a way to have a 100% guarantee of not missing a message?

Horusiath:
@aprooks thx for the note. It looks like a bug from snapshot store point of view - for some reason when the actor restarts it sends a request to snapshot store for the latest version of it's state, but snapshot store is not responding correctly - it doesn't save a snapshot on local file system on time, neither it returns data from the currently saving snapshots log.

aprooks:
@Horusiath thak you for confirmation.

Horusiath:
I've recreated the same issue using JVM akka persistence code ([sample](https://gist.github.com/Horusiath/bd00d33a130e4add2491)). Tracker: akka/akka#17139

Aaronontheweb:
Nice @Horusiath !

Horusiath:
As discussed on Akka JVM thread, this is not a bug. Akka.Persistence mechanism uses snapshotting only as optimization technique. It shouldn't be used solely without event's persistence as it's not a safe mechanism in environment when infrastructure is not error-proof (basically any non-academic case).

PR #859, opened 2015-04-15
Guaranteed delivery sample

aprooks:
Guaranteed delivery with default journal. I didn't use snapshotting as there are some discussions in progress #789 and akka/akka#17139

### DIFF
--- a/src/examples/PersistenceExample/Program.cs
+++ b/src/examples/PersistenceExample/Program.cs
@@ -16,14 +16,44 @@ namespace PersistenceExample
             {
                 //BasicUsage(system);
 
-                // FailingActorExample(system);
+                //FailingActorExample(system);
 
-                SnapshotedActor(system);
+                //SnapshotedActor(system);
 
                 //ViewExample(system);
 
+                GuaranteedDelivery(system);
+
                 Console.ReadLine();
             }
+        }
+
+        private static void GuaranteedDelivery(ActorSystem system)
+        {
+            Console.WriteLine("\n--- GUARANTEED DELIVERY EXAMPLE ---\n");
+            var delivery = system.ActorOf(Props.Create(()=> new DeliveryActor()),"delivery");
+
+            var deliverer = system.ActorOf(Props.Create(() => new GuaranteedDeliveryExampleActor(delivery.Path.ToString())));
+            delivery.Tell("start");
+            deliverer.Tell(new Message("foo"));
+            
+
+            System.Threading.Thread.Sleep(1000); //making sure delivery stops before send other commands
+            delivery.Tell("stop");
+
+            deliverer.Tell(new Message("bar"));
+
+            Console.WriteLine("\nSYSTEM: Throwing exception in Deliverer\n");
+            deliverer.Tell("boom");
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Hmmm..... since the commit is from 14 days ago looks like our build server didn't pick it up. Weird :\

aprooks:
Strange, as I was amending my old commit. Didn't expect date to be saved

Aaronontheweb:
@aprooks git works in mysterious ways - looks good now. I'll pull this in tomorrow so @Horusiath has time for feedback (if necessary.)

src/examples/PersistenceExample/Program.cs
```diff
@@ -16,16 +16,46 @@ static void Main(string[] args)
...
                 Console.ReadLine();
             }
         }
 
+        private static void GuaranteedDelivery(ActorSystem system)
+        {
+            Console.WriteLine("\n--- GUARANTEED DELIVERY EXAMPLE ---\n");
+            var delivery = system.ActorOf(Props.Create(()=> new DeliveryActor()),"delivery");
+
+            var deliverer = system.ActorOf(Props.Create(() => new GuaranteedDeliveryExampleActor(delivery.Path.ToString())));
```
Horusiath: You don't need to stringify `ActorPath`, since it's parsed later in the code. Just push it through.

Horusiath:
Only one little suggestion. Rest is good IMO. Great job!

aprooks:
@Horusiath removed stringification

Also I've just noticed, that foo is sent twice: 
```
--- GUARANTEED DELIVERY EXAMPLE ---

sending: foo, with deliveryId: 1
Confirming delivery of message id: 1 and data: foo

SYSTEM: Throwing exception in Deliverer

sending: bar, with deliveryId: 2
Ignoring message id: 2 and data: bar
recovered foo
recovered confirmation 1
recovered bar
[ERROR][16.04.2015 12:47:20][Thread 0011][akka://example/user/$b] Controlled dev
astation
Cause: System.Exception: Controlled devastation
   exception log

SYSTEM: Enabling confirmations in 3 seconds

sending: foo, with deliveryId: 1
Ignoring message id: 1 and data: foo
sending: bar, with deliveryId: 2
sending: bar1, with deliveryId: 3
Ignoring message id: 2 and data: bar
Ignoring message id: 3 and data: bar1

SYSTEM: Enabled confirmations

Confirming delivery of message id: 2 and data: bar
Confirming delivery of message id: 3 and data: bar1
```

seems like it is sent, then confirmed by recovery, so only 1 extra message. Should there be any validation if we are in recovery mode or not, to wait for all confirmations being recovered?

Horusiath:
@aprooks You always can check if persistent actor is at recovery phase using it's `IsRecovering` property. But now I see, that you've applied basically the same logic for both on command and on recovery receives and I'm not yet sure if it's valid at this point. I'll check that later.

aprooks:
@Horusiath ok I'll check docs as well

aprooks:
@Horusiath I've found issue: I shouldn't have persisted during recovery. Only message handling. I will update this later

Issue #812, opened 2015-04-07
Add Mono framework target for Akka.Cluster nuget package


Aaronontheweb:
Unfortunately some of the code built into Akka.Cluster is very platform specific - namely the `MetricsCollector` responsible for reporting on CPU / Memory utilization for each individual node in the cluster.

There's a specific performance counter we have to read to get total available memory for Mono, and a totally different method for Windows. Neither is x-plat compatible with the other (AFAIK.)

So the solution is to use pre-compilation directives and build platform-specific distributions of Mono for modules like Akka.Cluster that have this problem.

One of the things we'll need to do is add a Mono framework target to the NuGet package for these distributions - kind of like we'd do if we wanted to distribute a .NET 4.0 version of Akka.NET.

Does anyone have any idea how to do this *for Mono specifically*? I couldn't find what I was looking for inside the nuget.org docs.

kekekeks:
A lot of people who use Mono are deploying binaries built with Visual Studio (since Mono C# compiler and xbuild aren't that great). So it's not a good idea to ship a separate binary for Mono, it's better to select the implementation at run-time based on `Type.GetType("Mono.Runtime") != null`

Aaronontheweb:
@kekekeks in this case the issue is that on Windows we depend on a specific library that isn't available in Mono (don't laugh, but it's the VB.NET runtime - [here's why we use it](http://stackoverflow.com/questions/105031/how-do-you-get-total-amount-of-ram-the-computer-has).)

If we can come up with a platform-independent way of getting this data (it might be as simple as using a performance counter supported on both platforms) then that will eliminate this as a requirement.

Aaronontheweb:
> it might be as simple as using a performance counter supported on both platforms

(I have looked into this, but haven't settled on anything conclusive yet.)

Aaronontheweb:
(I might also be missing something really obvious, which happens more often than I'd like to admit ;) )

kekekeks:
@Aaronontheweb
>in this case the issue is that on Windows we depend on a specific library that isn't available in Mono

It's actually not an issue. If class isn't used, CLR/Mono won't try to load libraries it depends on. So on Mono we can safely use binaries compiled for .NET.

If one wants to use MonoDevelop for development (very strange idea from my point of view, because you can't use ReSharper) or as debugger (way more useful for this purpose), just wrap code that uses VB.NET library in
```csharp
#if __MonoCS__ 
#else
//Your code
#endif
```
[...]

PR #862, opened 2015-04-16
Proposed fix for #812

kekekeks:
#812


Aaronontheweb:
Looks good - PR's run on the CI server has been affected by #853 but we have some fixes for that.

PR #863, opened 2015-04-16
Select ConcurrentQueue implementation at runtime (Mono)

kekekeks:
See #812


rogeralsing:
As stated in 812, I think we can use the System version on both mono and .NET now. I think it was some 9 months ago that the concurrent queue impl in Mono was fixed.

Aaronontheweb:
@rogeralsing so close this PR and just remove the separate Mono implementation for `ConcurrentQueue` altogether then?

rogeralsing:
@Aaronontheweb  yes.

PR #868, opened 2015-04-16
 #853 fixed race condition in RoutingSpec

annymsMthd:
test Router_in_general_must_evict_terminated_routees


Aaronontheweb:
Aha! There we go - finally got a failure, but on a `ResizerSpec` test instead

http://petabridge-ci.cloudapp.net/viewLog.html?buildId=726&tab=buildResultsDiv&buildTypeId=AkkaNet_AkkaNetPrBuild&Guest=1

The frequency at which they're coming up is much less common now though, so I think you're making good progress.

PR #869, opened 2015-04-17
Test runner improvements

annymsMthd:
Opening pull request to test on CI. This may not be ready for merge and should be reviewed.

PR #870, opened 2015-04-17
Test runner improments

annymsMthd:
updated fake and used xunit2helper in fake script
Had to update two tests to pass
Upgraded to XUnit 2.0 RTM

#869 previous PR


src/core/Akka.Remote.Tests/Serialization/DaemonMsgCreateSerializerSpec.cs
```diff
@@ -34,6 +34,8 @@ protected override void OnReceive(object message)
         public DaemonMsgCreateSerializerSpec()
             : base(@"akka.actor.provider = ""Akka.Remote.RemoteActorRefProvider, Akka.Remote""")
         {
+            Akka.Serialization.Serialization.CurrentTransportInformation = null;
```
annymsMthd: I'd like to call this fix out. Without this when you run a test runner on the assembly the objects get the system name of the other tests.

HCanber:
Haven't reviewed this as I think we need to discuss this first:
This is a breaking change, isn't it?
Should we upgrade TestKit to xUnit 2 or should we create a new package `Akka.TestKit.Xunit2` (and then switch to use that internally) and have support for both. Is there any reason why people would not want to use v2?

Aaronontheweb:
@HCanber if we're able to swap out the underlying XUnit implementation to XUnit 2 and still preserve the integrity of the testkit (all tests that previously passed still pass) then I'd say it's not a breaking change.

The one annoying edge-case would be for users who mixed and matched their unit test project - some parts of the project use regular old XUnit without the TestKit and other parts do. If that's an issue they'll just need to break it up into two separate projects.

So TL;DR;: as long as XUnit 2 doesn't do anything that forces us to do a catastrophic rewrite of the TestKit.XUnit I think we'll be ok.

More important question in my opinion: what sort of weird shit does XUnit 2 do under the hood? ;)

src/core/Akka.Tests/Actor/ActorRefSpec.cs
```diff
@@ -73,6 +74,7 @@ public void An_ActorRef_should_not_equal_another_ActorRef_when_path_differs()
         public void An_ActorRef_should_not_allow_actors_to_be_created_outside_an_ActorOf()
         {
             Shutdown();
+            InternalCurrentActorCellKeeper.Current = null;
```
HCanber: Hmm... if the test fails without this row, we need to fix this somewhere else. Did it start failing by switching to xUnit 2?

HCanber:
@Aaronontheweb xUnit 2 has breaking changes, see http://xunit.github.io/docs/test-migration.html#update-tests and http://xunit.github.io/docs/upgrade-extensions.html

src/core/Akka.Tests/Actor/ActorRefSpec.cs
annymsMthd: I believe so. Something weird is happening with the threads and i noticed a few ThreadStatic attributes in the codebase.

Aaronontheweb:
@HCanber ok, in that case do you want @annymsMthd to move all of the XUnit 2.0 stuff into a new nuget package and have our internal tests reference that instead?

annymsMthd:
I'm going to fix that test, rebase and squash.

Issue #846, opened 2015-04-13
DateTime Concerns


mj1856:
In many places throughout the code the current time is retrieved by calling `DateTime.Now`.

For example, here are just a few of the lines I found:

````csharp
return Receiver.Ask(new Get(DateTime.Now + timeout), Timeout.InfiniteTimeSpan);
```

```csharp
var cancelable = Context.System.Scheduler.ScheduleTellOnceCancelable(next.Deadline - DateTime.Now, Self, new Kick(), Self);
```

This is not a good practice, as it means that code could fail during [daylight saving time](http://stackoverflow.com/tags/dst/info) transitions.

`DateTime.Now` returns the local date and time based on the time zone setting of the computer that the code is running on, however it does *not* take that time zone into account when doing math.

There are few valid choices for remedies:

1. You could track all values in UTC, which does not have DST transitions.  The current time would be obtained by calling `DateTime.UtcNow`.

2. If the local time is relevant, then you need to be working with `DateTimeOffset` values instead of `DateTime` values.  The fields and properties would have to change, and the current time would come from `DateTimeOffset.Now`.

3. You could consider using the [Noda Time](http://nodatime.org) library, in which case the individual values could be stored in any of `Instant`, `OffsetDateTime`, or `ZonedDateTime`, depending on exactly what purpose they serve.

See also the following articles on my blog:
- http://codeofmatt.com/2013/04/25/the-case-against-datetime-now/
- http://codeofmatt.com/2015/03/06/common-daylight-saving-time-mistakes-for-net-developers/

Aaronontheweb:
I think we should just switch to UTC on those calls - since we're really only using the current time to measure elapsed time, right?

rogeralsing:
@Aaronontheweb yepp

mj1856:
That would probably be the simplest thing to do to avoid DST issues.

mj1856:
Though - you may also want to consider what to do with existing data.  If the value is persisted anywhere, then switching to UTC would mean possibly comparing the previous local data with the new UTC value.   If that's something you're concerned with, the `DateTimeOffset` approach would solve it.

Aaronontheweb:
@mj1856 I'll take a deeper look at the code base, but most of those calls appear to be for things that are ephemeral to the state of an Akka.NET application as it runs in-memory. Scheduled recurring messages and so forth.

I'll look closely to see if the local clock gets accidentally exposed to the end-user anywhere they might need to store something.

[...]

PR #876, opened 2015-04-18
Monotonic clock for #846

kekekeks:
#846


Aaronontheweb:
So I guess we'd need to integrate this inside `Deadline`, `IScheduler`, and so forth... I agree that this is the right way to provide consistent + high resolution clock behavior internally.

Any thoughts as to how difficult this will be to integrate into those other areas?

HCanber:
Do not pull this is in yet, see discussion in #846

Aaronontheweb:
@HCanber does this look like it's ready to pull in?

Aaronontheweb:
The MonotonicClock values still need to be integrated with the actual `Scheduler`, but for now I'm going to pull it in.

HCanber:
Sorry, missed this.
Can we somehow mark these as experimental? Although they fulfill what's needed, I don't like the api, and I'm not sure the user will understand when to use them.
Maybe mark them with `[Obsolete("Experimental. Might be removed and replaced with something else")]`.

PR #877, opened 2015-04-19
Update nuspecs to point to new icon location

bruinbrown:
The NuGet packages were missing icons, this changes the nuspec files to point to the new icon location


Aaronontheweb:
No idea why XUnit is crashing after all of the specs pass. Merging this in anyway.

PR #878, opened 2015-04-19
QueueUserWorkItem vs UnsafeQueueUserWorkItem issue in .NetFiddle

sergey-litvinov:
Hey guys,

As you asked i've made a pull request for the issue. The reason why default thread-pool dispatcher doesn't work in .NetFiddle is that you are using `ThreadPool.UnsafeQueueUserWorkItem(wc, null);` in `ThreadPoolDispatcher` instead of `ThreadPool.QueueUserWorkItem(wc, null);` 

The main difference between these two calls is that `UnsafeQueueUserWorkItem` call doesn't check for Code Access Security(CAS), so it works a bit faster without these checks, but it requires FullTrust permissions. In .NetFiddle we can't allow to use FullTrust permissions as code executes on server, so logic that requires FullTrust doesn't work here.

In this pull request i've added check that is based on `AppDomain.CurrentDomain.IsFullyTrusted`, so if current AppDomain is FullTrusted, then the code will use `UnsafeQueueUserWorkItem`, otherwise it will use safe version `QueueUserWorkItem`. I've tested it with local .NetFiddle and it works fine on original code sample

### DIFF
--- a/src/core/Akka/Dispatch/Dispatchers.cs
+++ b/src/core/Akka/Dispatch/Dispatchers.cs
@@ -50,7 +50,8 @@ namespace Akka.Dispatch
     /// </summary>
     public class ThreadPoolDispatcher : MessageDispatcher
     {
-	    private static readonly bool _isFullTrusted = AppDomain.CurrentDomain.IsFullyTrusted;
+
+        private static readonly bool _isFullTrusted = AppDomain.CurrentDomain.IsFullyTrusted;
 
         /// <summary>
         /// Takes a <see cref="MessageDispatcherConfigurator"/>
@@ -66,11 +67,12 @@ namespace Akka.Dispatch
         public override void Schedule(Action run)
         {
             var wc = new WaitCallback(_ => run());
-			if (_isFullTrusted)
-				ThreadPool.UnsafeQueueUserWorkItem(wc, null);
-			else
-				ThreadPool.QueueUserWorkItem(wc, null);
-		}
+            // we use unsafe version if current application domain is FullTrusted
+            if (_isFullTrusted)
+                ThreadPool.UnsafeQueueUserWorkItem(wc, null);
+            else
+                ThreadPool.QueueUserWorkItem(wc, null);
+        }
     }
 
     /// <summary>
--- a/src/core/Akka/Dispatch/Dispatchers.cs
+++ b/src/core/Akka/Dispatch/Dispatchers.cs
@@ -50,6 +50,8 @@ namespace Akka.Dispatch
     /// </summary>
     public class ThreadPoolDispatcher : MessageDispatcher
     {
+	    private static readonly bool _isFullTrusted = AppDomain.CurrentDomain.IsFullyTrusted;
+
         /// <summary>
         /// Takes a <see cref="MessageDispatcherConfigurator"/>
         /// </summary>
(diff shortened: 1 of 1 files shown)


rogeralsing:
Thanks!
cc @akkadotnet/developers should we allow the dispatcher to be this smart? 
I'm not sure but this can have an impact on how actors behaves, right? if a message is scheduled from a CAS restricted client, the actor will not be able to execute any unsafe operations, while in the fully trusted version it would.

sergey-litvinov:
There are two differences - permission to the call itself and also CAS validations during thread execution.

From security point of view original implementation is ignoring CAS. So for example if I have such code:


	string path = "c:\\test";
	// allow only Read permissions for the file
	new FileIOPermission(FileIOPermissionAccess.Read, path).PermitOnly();

	var callback = new WaitCallback((o) =>
	{
		try
		{
			new FileIOPermission(FileIOPermissionAccess.Write, path).Demand();
			Console.WriteLine("Allowed");
		}
		catch (Exception)
		{
			Console.WriteLine("Denied");
		}
	});

	ThreadPool.QueueUserWorkItem(callback, null);
	ThreadPool.UnsafeQueueUserWorkItem(callback, null);

In that sample i've allowed just readonly access to "c:\\test" file, and in the callback i'm checking for write access for that file. As i allowed just read access, then write check should fail. If i use `QueueUserWorkItem`, then my permissions are propagated with copied call stack to executing thread, and when callback code will check for Write access - it will fail due to denied write access.

But in case of `UnsafeQueueUserWorkItem` - call stack wont be copied to the executing thread, and it will successfully pass and display "Allowed" string. So it will ignore my call stack permissions modifications that isn't correct.

It's in case when both client and actor are FullTrusted, but call stack was manually updated like in my sample. I'm not sure that this scenario is used often in real world.



Also there are 3 cases. If partially trusted client tries to schedule schedule a message:
1. With `UnsafeQueueUserWorkItem` it will fail, as `UnsafeQueueUserWorkItem` requires FullTrust and partially trust client just can't execute it at all.
2. With `QueueUserWorkItem` and correct client CAS permissions, then it will be scheduled and execued fine.
3. With `QueueUserWorkItem` and restricted client CAS permissions, then it will be scheduled and started fine, but somewhere during callback processing - it will fail with `SecurityException`.

So currently in original implementation it wont work at all for partially trusted client as it will always throw 
`Attempt by security transparent method 'Akka.Dispatch.ThreadPoolDispatcher.Schedule(System.Action)' to access security critical method 'System.Threading.ThreadPool.UnsafeQueueUserWorkItem(System.Threading.WaitCallback, System.Object)' failed.`
But with this change for partial trusted client will schedule it fine and will validate CAS permissions.

About FullTrust its hard to say should `QueueUserWorkItem` be used here or not, because its a trade between CAS validation and perfomance for it. Thats why i implemented `QueueUserWorkItem` just for partial trusted clients, and full trust clients will use unsafe version.

Aaronontheweb:
@rogeralsing do a few runs of the pingpong benchmark with and without this PR. Let's quantify how big the perf hit is when we're checking for access, since this function is literally the most performance critical one :p

rogeralsing:
Tried it and there is no impact perf wise. it's just an `if` on a static field bool.
I'm OK with merging this

sergey-litvinov:
Thanks!

PR #880, opened 2015-04-21
Added generic interface method for Props

Danthar:
Since the latest DI api changes, we can now resolve dependencies based
on type. However a generic convenience method is missing.
Since generics are used throughout Akka already. I see no reason not to
add this method.


Danthar:
I absolutely missed this: Context.DI().ActorOf<T>()

PR #867, opened 2015-04-16
Fixed a number of compiler warnings

sean-gilliam:
The compiler warnings are getting out of control with at least 102 in
HEAD. Most of these can be ignored since they look to be Obsolete tags
throwing the warnings. This PR reduces the number of warnings to 60.

PR #883, opened 2015-04-21
fixed `IDisposable` calls on EMWASpec and MetricsCollectorSpec

Aaronontheweb:
#867 inadvertently included a commit that introduced some recursive disposal calls in these two specs.

This undoes those changes and leaves a comment explaining why they're there.

### DIFF
--- a/src/core/Akka.Cluster.Tests/EWMASpec.cs
+++ b/src/core/Akka.Cluster.Tests/EWMASpec.cs
@@ -126,13 +126,13 @@ namespace Akka.Cluster.Tests
 
         #region IDisposable members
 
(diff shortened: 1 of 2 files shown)


sean-gilliam:
D'oh sorry about that :(

Aaronontheweb:
@sean-gilliam it's cool man - my fault for not noticing it when we were reviewing your PR. We're so used to XUnit being unreliable that we might occasionally miss stuff like this. Doesn't help that TeamCity and the console runner don't make it easy to find the `StackOverflowException`. I had to attach the debugger to the test run to find that.

PR #894, opened 2015-04-23
Remove unneeded references from the projects

sean-gilliam:
Most of the projects were referencing the following assemblies which
aren't being used by said project

System.Data
System.Data.DataSetExtensions
System.Xml
System.Xml.Linq


sean-gilliam:
There's still room for improvement, but i didn't want to get too aggressive. 

Yay green check mark!!

Issue #887, opened 2015-04-22
DeathWatch AddressTerminated issue


annymsMthd:
Just found this one out doing multinode tests.
JVM: https://github.com/akka/akka/blob/8485cd2ebb46d2fba851c41c03e34436e498c005/akka-actor/src/main/scala/akka/actor/dungeon/DeathWatch.scala line 165

NET: https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/ActorCell.DeathWatch.cs line 255

annymsMthd:
Fixed in #888

Aaronontheweb:
@annymsMthd would you mind breaking this out into its own PR, so we can get this fix in separately from any of the discussions around XUnit 2? Or would that be a big pain in the ass?

annymsMthd:
Crap I didn't see this. I can do that first thing tomorrow

PR #899, opened 2015-04-24
Fixed to notification on DeathWatch watching.

annymsMthd:
Also some boyscout cleanup

### DIFF
--- a/src/core/Akka/Actor/ActorCell.DeathWatch.cs
+++ b/src/core/Akka/Actor/ActorCell.DeathWatch.cs
@@ -25,7 +25,7 @@ namespace Akka.Actor
                 MaintainAddressTerminatedSubscription(() =>
                 {
                     a.Tell(new Watch(a, Self)); // ➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS
-                    _state = _state.AddWatching(a);                   
+                    _state = _state.AddWatching(a);
                 }, a);
             }
             return a;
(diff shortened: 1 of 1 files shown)


annymsMthd:
#887

annymsMthd:
@Aaronontheweb here we go

src/core/Akka/Actor/ActorCell.DeathWatch.cs
```diff
@@ -202,27 +207,23 @@ protected void RemWatcher(IActorRef watchee, IActorRef watcher)
...
-            watchedBy = _state.GetWatchedBy();
-
             // send DeathWatchNotification to self for all matching subjects
             // that are not child with existenceConfirmed = false because we could have been watching a
             // non-local ActorRef that had never resolved before the other node went down
             // When a parent is watching a child and it terminates due to AddressTerminated
             // it is removed by sending DeathWatchNotification with existenceConfirmed = true to support
             // immediate creation of child with same name.
-            foreach (var a in watchedBy.Where(a => a.Path.Address == address))
+            foreach (var a in _state.GetWatching().Where(a => a.Path.Address == address))
```
annymsMthd: This is the fix

Aaronontheweb: So was this a bug that got added when we extracted the `ActorCell` state, or has this been a persistent issue for some time? Only asking because it may explain some of the cluster deathwatch behavior I've observed during multi-node test runs.

annymsMthd: Looks like this was introduced about 29 days ago. It affects Cluster DeathWatch directly because a node that has been removed from the cluster can't tell the DeathWatchers from another node it's dead:)

rogeralsing:
Ohhh my bad, thanks!

PR #903, opened 2015-04-25
async/await support via custom awaiter

kekekeks:
Usage
```csharp
protected override async void OnReceive(object message)
{
    await Task.Delay(1000).ActorAwait(AsyncBehavior.Suspend);
}
```

Advantages:
- no overhead of using `Task.Factory.StartNew`
- no any additional threads
- one can await from UI thread
- one can mix `AsyncBehavior.Reentrant` and `AsyncBehaviour.Suspended` inside one handler

Disadvantages:
- People might forget to call ActorAwait

Example usage:

```csharp
private class AsyncActor : UntypedActor
{
    protected override async void OnReceive(object message)
    {
        if (message.Equals("crash"))
        {
            try
            {
                await ThrowExceptionAsync().ActorAwait();
            }
            catch (Exception e)
            {
                Console.WriteLine("Catched: "+ e);
            }
        }
        else if (message is string)
        {
            Console.WriteLine("Before first delay, message: " + message);
            await Task.Delay(1000).ActorAwait(AsyncBehavior.Suspend);
            Console.WriteLine("After suspended delay, message: " + message);
            await Task.Delay(1000).ActorAwait(AsyncBehavior.Reentrant);
            Console.WriteLine("After reentrant delay, message: " + message);
        }
    }

    private async Task ThrowExceptionAsync()
    {
        await Task.Delay(100);
        throw new Exception("crash");
    }
}

public static void Main()
{
    var system = ActorSystem.Create("test");
    var actor = system.ActorOf(Props.Create<AsyncActor>());
    actor.Tell("1");
    actor.Tell("2");
    actor.Tell("crash");
    system.AwaitTermination();
}
```

### DIFF
--- a/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
+++ b/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
@@ -98,6 +98,7 @@ namespace Akka.Actor
             else if (m is PoisonPill) HandlePoisonPill();
             else if (m is ActorSelectionMessage) ReceiveSelection(m as ActorSelectionMessage);
             else if (m is Identify) HandleIdentity(m as Identify);
+            else if (m is CompleteReentrantAwaitedTask) HandleCompleteReentrantAwaitedTask(m as CompleteReentrantAwaitedTask);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


rogeralsing:
As the async await support have been a hot potato ever since it was introduced. 
We should wait with merging anything that adds to that, and instead first have a long talk about exactly what we should support on async await.

HCanber:
Yep, agree.

Aaronontheweb:
cc @rogeralsing is this PR still relevant?

nvivo:
Just to add some feedback, some comments against this by [Phil Haack](http://haacked.com/archive/2014/11/11/async-void-methods/) and [Stephen Cleary](http://stackoverflow.com/a/19415703/1902354).

As a library, it would be a terrible idea to provide official support for a bad pattern. We need to find a better way to solve this.

kekekeks:
It may be used later as sugar for PipeTo. But current implementation isn't suitable for that

Issue #886, opened 2015-04-22
ConfigureAwait(false) all async/await stuff


Horusiath:
It's a good practice anyway, so in order to prevent for mysterious deadlocks in synchronous code using async API (in example `IActorRef.Ask` method) we should add `ConfigureAwait(false)` to all of our task-based operations exposed as async methods.

rogeralsing:
Are you sure?
lets say I have an `ReceiveActor` with an `async` receive handler.
and lets assume this receive handler calls `.Ask` on another actor.
Will that returned task complete outside of the asking actors TaskScheduler?

rogeralsing:
Changing the Ask method to:

```
        public static Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeout = null)
        {
            var t = AskProxy<T>(self, message, timeout);
            t.ConfigureAwait(false);
            return t;
        }

        private static async Task<T> AskProxy<T>(ICanTell self, object message, TimeSpan? timeout)
        {
            IActorRefProvider provider = ResolveProvider(self);
            if (provider == null)
                throw new NotSupportedException("Unable to resolve the target Provider");

            ResolveReplyTo();
            var result = (T) await Ask(self, message, provider, timeout);
            return result;
        }
```

Still passes all the tests.
But TBH, I have no clue if this change makes anything better in any way.
Would the above solve the deadlock issue you saw @Horusiath ?

Aaronontheweb:
@rogeralsing `ConfigureAwait(false)` is generally good idea inside libraries and frameworks - we had to do this EVERYWHERE inside the MarkedUp SDK. This is one of the stupid / evil default behaviors `await` introduced to our otherwise glorious concurrent programming libraries.

rogeralsing:
The semantics of it is pretty unclear to me at least, exactly how does this affect async state and schedulers?

Skickat från min iPhone

kekekeks:
@rogeralsing
```
        public static Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeout = null)
        {
            var t = AskProxy<T>(self, message, timeout);
            t.ConfigureAwait(false);
            return t;
        }
```
`t.ConfigureAwait(false);` here does nothing. This method doesn't modify the Task, it returns [ConfiguredTaskAwaitable](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.configuredtaskawaitable%28v=vs.110%29.aspx) that can be awaited.

[...]

PR #905, opened 2015-04-25
Proposed fix for #886 asp.net deadlock

kekekeks:
See #886

### DIFF
--- a/src/core/Akka/Actor/Futures.cs
+++ b/src/core/Akka/Actor/Futures.cs
@@ -29,15 +29,14 @@ namespace Akka.Actor
             return self.Ask<object>(message, timeout);
         }
 
-        public static async Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeout = null)
+        public static Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeout = null)
         {
             IActorRefProvider provider = ResolveProvider(self);
             if (provider == null)
                 throw new NotSupportedException("Unable to resolve the target Provider");
 
             ResolveReplyTo();
-            var result = (T)await Ask(self, message, provider, timeout);
-            return result;
+            return Ask(self, message, provider, timeout).CastTask<object, T>();
         }
 
         internal static IActorRef ResolveReplyTo()
--- a/src/core/Akka/Akka.csproj
+++ b/src/core/Akka/Akka.csproj
@@ -245,6 +245,7 @@
     <Compile Include="Util\Internal\Collections\ImmutableAvlTreeMap.cs" />
     <Compile Include="Util\Internal\Collections\ImmutableTreeSet.cs" />
     <Compile Include="Util\Internal\StringBuilderExtensions.cs" />
+    <Compile Include="Util\Internal\TaskExtensions.cs" />
     <Compile Include="Util\ISurrogate.cs" />
     <Compile Include="Util\StandardOutWriter.cs" />
     <Compile Include="Event\Subscription.cs" />
--- a/src/core/Akka/Util/Internal/TaskExtensions.cs
+++ b/src/core/Akka/Util/Internal/TaskExtensions.cs
@@ -1,0 +1,35 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Akka.Util.Internal
+{
+    public static class TaskExtensions
+    {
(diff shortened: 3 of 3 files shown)


rogeralsing:
Looks good to me, anyone verified if this solves the asp.net deadlocks?

kekekeks:
```
    public class ActorController : Controller
    {
        //
        // GET: /Actor/


        public class MyActor : UntypedActor
        {
            {
                Sender.Tell(message + "-pong");
            }
        }

        public ActionResult Index()
        {
            var actor = MvcApplication.ActorSystem.ActorOf(Props.Create<MyActor>());
            var res = actor.Ask<string>("ping").Result;
            return new ContentResult() {Content = res};
        }

    }
```

At least this code works now

nvivo:
There are more things happenning here.

1. The Ask's are in the reverse order. Today, Ask<object> call Ask<T> that calls self.Ask<object> with await. This is all unnecessary. Just by moving all logic to Ask<object> and making Ask<T> just cast the result achieves the same result. The CastTask method is unecessary.

2. I suspect this doesn't actually fix the deadlock. Can you run this with actor selection to an actor running somewhere else?

kekekeks:
It seems, that it solves the problem only for ActorRef.

kekekeks:
```csharp
    {

        public ActionResult Index()
        {

            var actor = MvcApplication.ActorSystem.ActorSelection("/user/test");
            var res = actor.Ask<string>("ping").Result;
            return new ContentResult() {Content = res};
        }

    }
```

```csharp
    public class MvcApplication : System.Web.HttpApplication
    {

        public class MyActor : UntypedActor
        {
            public MyActor()
            {
                
            }
            {
                Sender.Tell(message + "-pong");
            }
        }

        public static ActorSystem ActorSystem = ActorSystem.Create("sandbox");
        protected void Application_Start()
        {
            ThreadPool.QueueUserWorkItem(_ => MvcApplication.ActorSystem.ActorOf(Props.Create<MyActor>(), "test"));
            AreaRegistration.RegisterAllAreas();

            WebApiConfig.Register(GlobalConfiguration.Configuration);
            FilterConfig.RegisterGlobalFilters(GlobalFilters.Filters);
            RouteConfig.RegisterRoutes(RouteTable.Routes);
            BundleConfig.RegisterBundles(BundleTable.Bundles);
        }
    }
```

Works in this scenario

kekekeks:
>Just by moving all logic to Ask and making Ask just cast the result achieves the same result. The CastTask method is unecessary.

You can't do that. You have to move the cast to `FutureActorRef` (which is public and is used in a lot of places) and change it from `FutureActorRef` to `FutureActorRef<T>`

nvivo:
Talking different things here. What I mean is:

```cs
public static Task<object> Ask(this ICanTell self, object message, TimeSpan? timeout = null)
{
    return self.Ask<object>(message, timeout);
}

public static async Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeout = null)
{
    IActorRefProvider provider = ResolveProvider(self);
    if (provider == null)
        throw new NotSupportedException("Unable to resolve the target Provider");

    ResolveReplyTo();
    var result = (T)await Ask(self, message, provider, timeout);
    return result;
}
``` 

First should read as:

```cs
public static Task<object> Ask(this ICanTell self, object message, TimeSpan? timeout = null)
{
    IActorRefProvider provider = ResolveProvider(self);
    if (provider == null)
        throw new NotSupportedException("Unable to resolve the target Provider");

    ResolveReplyTo();

    return Ask(self, message, provider, timeout);
}

public static async Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeout = null)
{
    return (T) await Ask(self, message, timeout);
}
```  

The internal Ask already returns a `Task<object>`, there is no need for all the indirection. This doesn't change the problem, just makes the code more readable and in the correct order.

Then you can optimize the await with some ContinueWith logic to execute in the same thread that just casts the value, but that is much more simple than that.

With code in the right order, you could just avoid the entire casting logic inside
```cs
var result = (string) actor.Ask("ping").Result;
```

That's what I mean.

Aaronontheweb:
@kekekeks could you please run the test suite with the debugger attached and see what it's failing on Akka.Tests? Normally when this issue comes up it's a `StackOverflowException` related to XUnit and the TestKit both racing to dispose a resource, but the error message will not show up in the build server.

kekekeks:
Should work now

rogeralsing:
Im pulling this in, tests are passing and the code is cleaner than we have atm.

Issue #904, opened 2015-04-25
Receive with AsyncBehavior.Suspend doesn't suspend actor's mailbox completely


kekekeks:
```csharp
private class AsyncActor : ReceiveActor
{
    public AsyncActor()
    {
        Receive <SuspendedMessage>(AsyncBehavior.Suspend, async m =>
        {
            Console.WriteLine("Suspended before await");
            await Task.Delay(1000);
            Console.WriteLine("Suspended after await");
        });
        Receive<ReentrantMessage>(AsyncBehavior.Reentrant, async m =>
        {
            Console.WriteLine("Reentrant before await");
            await Task.Delay(500);
            Console.WriteLine("Reentrant after await");
        });



    }
}

class SuspendedMessage
{
            
 
}

public class ReentrantMessage
{
             
}

        

public static void Main()
{
    var system = ActorSystem.Create("test");
    var actor = system.ActorOf(Props.Create<AsyncActor>());
    actor.Tell(new ReentrantMessage());
    actor.Tell(new SuspendedMessage());
    system.AwaitTermination();
}
```

Console output:
```
Reentrant before await
Suspended before await
Reentrant after await
Suspended after await
```

That's because continuations to reentrant tasks are also scheduled via `CompleteTask` which is a system message and bypasses suspend rules of the mailbox.

I think it should be fixed by adding auto-received message.

kekekeks:
Fixed by reentrancy removal.

PR #906, opened 2015-04-25
Proposed fix for #904

kekekeks:
#904

### DIFF
--- a/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
+++ b/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
@@ -98,6 +98,7 @@ namespace Akka.Actor
             else if (m is PoisonPill) HandlePoisonPill();
             else if (m is ActorSelectionMessage) ReceiveSelection(m as ActorSelectionMessage);
             else if (m is Identify) HandleIdentity(m as Identify);
+            else if (m is CompleteTaskReentrant) HandleCompleteTask(m as CompleteTaskReentrant);
         }
 
         /// <summary>
@@ -209,6 +210,14 @@ namespace Akka.Actor
             Sender = task.State.Sender;
             task.SetResult();
         }
+
+        private void HandleCompleteTask(CompleteTaskReentrant task)
+        {
+            CurrentMessage = task.State.Message;
+            Sender = task.State.Sender;
+            task.SetResult();
+        }
+
         public void SwapMailbox(DeadLetterMailbox mailbox)
         {
             Mailbox.DebugPrint("{0} Swapping mailbox to DeadLetterMailbox", Self);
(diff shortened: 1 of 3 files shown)


rogeralsing:
This looks good.

I'm waiting with merging this a bit, I need to think a bit on the directon where we should go with the async await stuff, just so we don't dig us even deeper into a hole by adding more code ontop of the current impl.

kekekeks:
I think it would be nice to create some general "system" messages to invoke stuff on mailbox's context. In this case there won't be `CompleteTask`/`CompleteTaskReentrant`, but `InvokeDelegateSystemMessage` and `InvokeDelegateMessage` with `Action` as the only parameter (and `AmbientState` would be embedded into closure inside `ActorTaskScheduler`). Then one can experiment with async/await outside Akka codebase to their heart's content.

rogeralsing:
@kekekeks yes the proposition of InvokeDelegateSystemMessage and InvokeDelegateMessage sounds good to me.

Another note that is worth considering.
When suspending the mailbox, only user messages are suspended, so even in that mode, system messages for restart and other management messages might slip through to the actor.
This was by design from my part originally, as we do recommend the PipeTo pattern, which acts the same way, each task is piped back as a message.
And as that is our default approach, it seemed like a good idea to do the same for the suspend behavior. so that the async await stuff is sort of like a DSL ontop of the actor.

I'm not sure I still think that is a good idea. the suspend behavior should probably mimic the semantics of normal actors as far as possible.
that is, no messages what so ever are allowed to be processed while the actor is running an async hander, untill the entire async flow is completed.

The main issue at hand in this async await mess, is that there is no right or wrong in how to interpret how it should be applied.
it's totally ok to say that it _should_ be a DSL ontop of actors that becomes some sort of syntactic sugar for the PipeTo pattern.
But it is also completely fine to say that it should act exactly as normal actors do.

And everyone and his mother have a different view on how it should be applied.
But I do think that keeping the exact same semantics as for normal actors when using suspend would not break the rule of least surprise, or?

kekekeks:
Well, there are two completely different applications of async/await in case of actors.
1) Original Akka's pattern with `PipeTo`, where one might want some syntactic sugar.
2) Complex asynchronous handler of a single message that asks other actors, does async I/O, etc.

Most people in .NET world would want/expect the second scenario, but the first one is also very useful.

So in some ideal world where Akka.Net was designed from scratch all actors should return `Task` and no messages are processed until it is complete/failed/whatever. And when people want some reentrancy, they use something like `ConfigureAwait` (called `ReentrantAwait`, `PipeToAwait`, whatever), that returns a custom awaiter that pipes the continuation back to actor's mailbox and resumes message processing.

rogeralsing:
If we remove reentrancy, this goes away to some extent.
If actors should or should not process system messages during an async suspend needs to be decided, as there are different opinions on this atm.

But I still agree that having the flexibility to send delegate system messages could be a nice addition.

Aaronontheweb:
cc @rogeralsing is this PR still relevant?

kekekeks:
Shouldn't be

Issue #908, opened 2015-04-26
IndexOutOfRangeException with RoundRobinRoutingLogic & SmallestMailboxRoutingLogic


jordansjones:
Assuming `Routee[] routees` has 1 or more elements in the array, when `Interlocked.Increment(ref _next)` rolls over to a negative number, an `IndexOutOfRangeException` will be thrown.

The following is a simplistic example that demonstrates what is happening:
```csharp
var items = new [] { 1, 2, 3, 4, 5 };
var count = int.MaxValue - 5;

for (var i = 0; i < 20; i++)
{
	items[Interlocked.Increment(ref count) % items.Length].Dump();
}
```

The following is a simplistic program that demonstrates the above while using Akka.
```csharp
public class Program
{

    public const long Total = ((long) int.MaxValue) + 15;

    public static Stopwatch Stopwatch;

    public static void DoIt()
    {
        var system = ActorSystem.Create("GreetSystem");
        var greeter = system.ActorOf(
            Props.Create(() => new OutputActor())
                    .WithRouter(new RoundRobinPool(10).WithResizer(new DefaultResizer(5, 150)))
            );

        Stopwatch = Stopwatch.StartNew();
        for (var i = 0L; i < Total; i++)
        {
            greeter.Tell(i);
        }
    }

    public static void Main(string[] args)
    {
        DoIt();
    }

    public class OutputActor : ReceiveActor
    {

        private static long _count = -1;

        private static float _percent = -1f;

        private static readonly object Guard = new object();

        public OutputActor()
        {
            Receive<long>(x => DoStuff(x));
        }

        protected void DoStuff(long message)
        {
            var count = Interlocked.Increment(ref _count);
            lock (Guard)
            {
                var percent = (float) Math.Round((double) count / Total, 4);
                if (Math.Abs(percent - _percent) < .00001F) { return; }

                _percent = percent;
                Console.SetCursorPosition(0, 0);
                Console.Write("                    ");
                Console.SetCursorPosition(0, 0);
                Console.Write("{0:P2}", _percent);
            }
        }

    }

}
```

PR #909, opened 2015-04-26

jordansjones:
When `_next` rolls over to a negative value, this will ensure that the
negative value is corrected to be positive.

Reference: http://stackoverflow.com/questions/4412179/best-way-to-make-javas-modulus-behave-like-it-should-with-negative-numbers/4412200#4412200

### DIFF
--- a/src/core/Akka/Routing/RoundRobin.cs
+++ b/src/core/Akka/Routing/RoundRobin.cs
@@ -35,7 +35,7 @@ namespace Akka.Routing
             {
                 return Routee.NoRoutee;
             }
-            return routees[Interlocked.Increment(ref _next)%routees.Length];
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
there's a better way of doing this - just mask the sign bit off. 

See https://github.com/helios-io/DedicatedThreadPool/blob/dev/src/core/Helios.DedicatedThreadPool/Helios.Concurrency.DedicatedThreadPool.cs#L325

Closing this PR

Aaronontheweb:
This is definitely a bug, so thanks for pointing it out - but I recommend doing it by just bitshifting

jordansjones:
No problem.. I will update and re-submit.

PR #910, opened 2015-04-27

jordansjones:
Per the suggestion from @Aaronontheweb (#909), bitmask Interlocked.Increment(ref _next)
to ensure that it will always be a positive number.

### DIFF
--- a/src/core/Akka/Routing/RoundRobin.cs
+++ b/src/core/Akka/Routing/RoundRobin.cs
@@ -35,7 +35,7 @@ namespace Akka.Routing
             {
                 return Routee.NoRoutee;
             }
-            return routees[Interlocked.Increment(ref _next)%routees.Length];
+            return routees[(Interlocked.Increment(ref _next) & int.MaxValue) % routees.Length];
         }
     }
 
--- a/src/core/Akka/Routing/SmallestMailbox.cs
+++ b/src/core/Akka/Routing/SmallestMailbox.cs
@@ -35,7 +35,7 @@ namespace Akka.Routing
             var winningScore = long.MaxValue;
 
             // round robin fallback
-            var winner = routees[Interlocked.Increment(ref next) % routees.Length];
+            var winner = routees[(Interlocked.Increment(ref next) & int.MaxValue) % routees.Length];
 
             for (int i = 0; i < routees.Length; i++)
             {


annymsMthd:
Did we have a test that triggered this bug? If not would it be possible to create on while it is still fresh in your mind?

jordansjones:
No test that triggered the bug. I happened to notice this while looking at the code.

Putting a test in for this would require setting an initial value for `_next` aside from `-1`. This may sound like me being lazy, but I hesitate in making moderate changes to this code base because I am still new to it and I don't want to step on any toes.. If yourself, or anyone else, would like to give some direction, I would be happy to make it happen.

src/core/Akka/Routing/SmallestMailbox.cs
```diff
@@ -35,7 +35,7 @@ private Routee SelectNext(Routee[] routees)
             var winningScore = long.MaxValue;
 
             // round robin fallback
-            var winner = routees[Interlocked.Increment(ref next) % routees.Length];
+            var winner = routees[(Interlocked.Increment(ref next) & int.MaxValue) % routees.Length];
```
Aaronontheweb: Needs to be wrapped in an `unchecked` block

src/core/Akka/Routing/RoundRobin.cs
```diff
@@ -35,7 +35,7 @@ public override Routee Select(object message, Routee[] routees)
             {
                 return Routee.NoRoutee;
             }
-            return routees[Interlocked.Increment(ref _next)%routees.Length];
+            return routees[(Interlocked.Increment(ref _next) & int.MaxValue) % routees.Length];
```
Aaronontheweb: Needs to be wrapped in an `unchecked` block

Aaronontheweb:
You should refactor the `RoutingLogic` instance for both of these to be testable and add specs.

Add two constructors to each - a default constructor that passes an initial counter value of zero to a second constructor which takes that counter value as an argument. Shouldn't create any serialization problems or anything like that.

src/core/Akka/Routing/SmallestMailbox.cs
jordansjones: Pardon my ignorance, but why would this need to be wrapped in `unchecked`?

Aaronontheweb: No sweat @jordansjones - glad that you asked!

When next == Int32.MaxValue the `Interlocked.Increment` operation will cause `next` to "overflow" from a positive to a negative value. In the CLR's normal behavior this will throw an exception because all arithmetic is "checked" for overflows and underflows (subtracting Int32.MinValue to make it positive) by default.

Wrapping this call in an `unchecked` block tells the CLR not to check any arithmetic operations for underflows / overflows and lets the application handle those issues themselves, which we're doing using bitshifting.

So this basically just makes sure that the operation always succeeds.

See the full docs for `unchecked` here https://msdn.microsoft.com/en-us/library/a569z7k8(v=vs.71).aspx

jordansjones: Hmm.. I got the distinct impression that the Interlocked.Increment handled the overflow scenario internally.

Aaronontheweb: Looks like you're right: https://msdn.microsoft.com/en-us/library/vstudio/dd78zt0c(v=vs.100).aspx

jordansjones:
Closed and created new pull request with unit tests (#916).

Issue #907, opened 2015-04-26
TestKit ImplicitSender - async await broken


rogeralsing:
Another async await issue incoming..

in the `TestKitBase`, we have this lines:

```csharp
 if(!(this is INoImplicitSender))
{
       InternalCurrentActorCellKeeper.Current = (ActorCell)((ActorRefWithCell)testActor).Underlying;
}
_testActor = testActor;
```

This sets the testActor as the current sender for all tests that are not decorated with `INoImplicitSender`

So far so good.

But as test methods can be async, this does not work, for the very same reasons we had to hack the world to get async await to work inside actors.

The implicit context is lost in the `await` and all code running in the test method be incorrect.
And worse:
if there are two parallel tests running at the same time, where one has `INoImplicitSender` and one hasnt'
Its completely possible that the context set above leaks over to the other test if that happens to inherit the same thread while awaiting.

So now good folks, here we have correct Task based methods, with broken async await support.
As these tests do return tasks, if I get what everyone else is saying here, it should be an easy fix?

(In my head it's not, I still think the implicit actor context needs to async flow in the tests also.
But I've been wrong before on this topic :-) )

cc @nvivo @kekekeks @Danthar @HCanber 

Someone reported this in a PR recently, that the context had leaked over to another test.
And had to set `.Current = null`

So, shoot, anyone have any ideas how to fix this?

rogeralsing:
actually, it will leak context even if two or more tests are running and are async, 

e.g.
test 1 sets context in thread 1,  test 2 sets context in thread 2
test 1 calls await , test 2 calls await
test 1 resumes on thread 2, test 2 resumes on thread 1

Now implicit `Sender` is swapped for both tests..

Danthar:
Even then its definitely not a silver bullet. This guy explains it perfectly: http://blog.stephencleary.com/2013/04/implicit-async-context-asynclocal.html

rogeralsing:
![flippin the table](http://i.kinja-img.com/gawker-media/image/upload/s--0IrKEf86--/i8izjqpwgxpozsktndnc.gif)

rogeralsing:
cc @StephenCleary we need you! <3
The async await stuff with implicit contexts are killing us

Aaronontheweb:
![abandon-thread](https://cloud.githubusercontent.com/assets/326939/7338818/f17c085a-ec0c-11e4-9737-4bceb057f9ec.gif)

nvivo:
[...]

PR #912, opened 2015-04-27
SynchronizationContext for #907

kekekeks:
This should restore current actor cell before calling continuation

### DIFF
--- a/src/core/Akka.TestKit/ActorCellKeepingSynchronizationContext.cs
+++ b/src/core/Akka.TestKit/ActorCellKeepingSynchronizationContext.cs
@@ -24,7 +24,10 @@ namespace Akka.TestKit
             ThreadPool.UnsafeQueueUserWorkItem(_ =>
             {
                 var oldCell = InternalCurrentActorCellKeeper.Current;
+	            var oldContext = Current;
+				SetSynchronizationContext(this);
                 InternalCurrentActorCellKeeper.Current = _cell;
+				
                 try
                 {
                     d(state);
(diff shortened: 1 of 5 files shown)


kekekeks:
#907

rogeralsing:
Do we have to account for the xUnit synchronizationcontext or is that applied ontop of this one once the async test starts?

kekekeks:
Xunit just [wraps](https://github.com/xunit/xunit/blob/43fa87d47a593aab97ff9f84dc5d2c2ac54b1a33/src/xunit.execution/Sdk/AsyncTestSyncContext.cs) anything it finds [before calling the method](https://github.com/xunit/xunit/blob/3f8b35c989af05e7b80beae37938b00036c670ce/src/xunit.execution/Sdk/Frameworks/Runners/TestInvoker.cs#L187), but after [calling constructor](https://github.com/xunit/xunit/blob/3f8b35c989af05e7b80beae37938b00036c670ce/src/xunit.execution/Sdk/Frameworks/Runners/TestInvoker.cs#L157).

Aaronontheweb:
Could we add a test to verify the existence and resolution of the bug in addition to the fix?

kekekeks:
Obsoleted by #913

kekekeks:
CallContext is a slowpoke, so it seems that we are going to use this

rogeralsing:
As #913 is too costly in terms of processing overhead, we still need this fix.
The test seems to fail, are we losing context even with this applied?

PR #913, opened 2015-04-27
Use CallContext for current ActorCell

kekekeks:
#912 #907


kekekeks:
Too slow for production use. Meh.

PR #915, opened 2015-04-27
Added .editorconfig

kekekeks:
http://editorconfig.org/

Needed for [TabSanity](https://visualstudiogallery.msdn.microsoft.com/ac4d4d6b-b017-4a42-8f72-55f0ffe850d7) to work properly.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+; This file is for unifying the coding style for different editors and IDEs.
+; More information at http://EditorConfig.org
+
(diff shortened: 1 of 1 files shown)


HCanber:
Hmm... we use 4 spaces for indentation and not tab so it should say:
``` 
indent_style = space
indent_size = 4
```

Right?

kekekeks:
ARGH

kekekeks:
#923

PR #916, opened 2015-04-28

jordansjones:
Per the suggestion from @Aaronontheweb (#909) bitmask Interlocked.Increment(ref _next)
to ensure that it will always be a positive number.
Also, added unit tests that verify this behavior (#910).


annymsMthd:
Nice.

PR #917, opened 2015-04-28
Removed Reentrancy behavior

rogeralsing:
Removing all traces of async reentrancy.

### DIFF
--- a/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
+++ b/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
@@ -8,8 +8,6 @@
 using System;
 using System.Threading.Tasks;
 using Akka.Actor;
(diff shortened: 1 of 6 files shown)


rogeralsing:
The build error, is that something else or what causes that?
"it works on my machine"...

annymsMthd:
Looks like a nuget issue

nvivo:
I'm all for it, but doesn't this change the API? Where do we draw the line?

rogeralsing:
Yes it does change the API, but hopefully reentrancy was not widely used, and there is just too much noise around it that steals focus from the project itself. 
So I'm all for breaking that part..

nvivo:
Agree.

HCanber:
Agree

Aaronontheweb:
yeah science

Pulling this in.

PR #919, opened 2015-04-28
Link to existing SharedAssemblyInfo.cs instead of own copy

HCanber:
Tiny fix. `SharedAssemblyInfo.cs` was a copy, instead of a link to the existing one.


Aaronontheweb:
thanks @HCanber!

PR #931, opened 2015-04-30
added clearer logging and validation to Remote / Cluster

Aaronontheweb:
Added a null argument check for supervisor strategy inside ClusterPoolRouters - saw this come up inside the WebCrawler demo.
Made it so heartbeat timeouts get logged to the error log.


annymsMthd:
Nice. This should help in debugging.

PR #932, opened 2015-04-30
All Pool routers from configuration now use DefaultSupervisorStrategy

Aaronontheweb:
Fixed a bug that caused clustered pool routers to have a null `SupervisoryStrategy` by default when loaded from configuration. Big problem if one of your routees under a clustered router dies :p


annymsMthd:
Looks good, short and to the point. Just waiting for the build.

Issue #851, opened 2015-04-14
F# API enchancements


Horusiath:
These are some thoughts provoked by akka-typed package.

## More `actor { ... }` computation expression results

Right now the only returned values of actor computation expression are Cont (changes the actor behavior or remains the same) or Return (stops actor). My proposition is to:

1. Rename Return->Stop to better describe it's true intent.
2. Make a Return/Stop bound to it's own custom computation expression called `stop`.
3, Introduce new result, Unhandled returned is result of calling next custom computation expression called 'unhandled`.

This could allow us to introduce actor combinators (if it is possible). Combinator is the operator which combines two other behaviors and returns an behavior as a result:

- Or eg. `behavior1 || behavior2` will generate result behavior invoking both of provided ones in way, that behavior2 will be executed if behavior1 result is either `Cont` or `Unhandled`.
- And eg. `behavior1 && behavior2` will generate result behavior invoking both of provided ones in way, that behavior2 will be executed upon message only when result of behavior1 is `Cont`.

This is how it could work:

```fsharp
let aref = spawn system "actor" 
    <| fun mailbox ->
        // first behavior
        let rec behavior1 () = actor {
            let msg = mailbox.Receive()
            match msg with
            | Msq1 -> unhandled!  // returns an unhandled result
            | StoppingMessage -> stop!     // stops an actor
            | _ -> return! behavior1()     // normal actor return
        }
        // second behavior
        let rec behavior2 () = actor {
            let msg = mailbox.Receive()
            match msg with
            | StoppingMessage -> stop!     // stops an actor
            | _ -> return! behavior2()     // normal actor return
        }
        // combine both behaviors together into one and call it
        (behavior1 || behavior2) ()
```

## Make `spawn` functions return typed version of IActorRef

Spawning methods may return a typed version of IActorRef, so that only messages of certains types + system control messages (which could be recognized by `IAutoReceiveMessage` interface) are allowed to push through that actor ref.

Issue #841, opened 2015-04-12
F# Persistence API: add switchable behavior to command receiver.


Horusiath:
The idea is to add FSM support to F# just like it's in the core F# API - passes `Aggregate` exec method should be switchable using mutual recursion.

PR #935, opened 2015-05-01
Typed actor ref in F# API + new Akka.Persistence.FSharp API

Horusiath:
This is a proposition of typed actor refs in F# API and new idea of Persistence F# plugin. My proposal is to push them (after review) to -pre version to let users play with it before official publish. Make it something like `1.1-RC1` version.

## Changes in Akka.FSharp

- Introduced strongly typed version of `IActorRef` (`ActorRef<'Message>`) and `TypedActorSelection<'Message>` - right now they'll allow only to tell/ask using messages of specific type (compile time checked). `IActorRef<>` is also serializable (test attached).
- All F# APi functions have been changed to use these new interfaces.
- Event stream subscribe/unsubscribe functions now doesn't need `Type` as first param - it's inffered from provided `IActorRef<'Mesage>`.
- One more fix - until now when message of incorrect type was sent to an actor, an `InvalidCastException` was thrown. Now it will be checked and in case of invalid type of message it will become `Unhandled`.

## Changes in Akka.Persistence.FSharp

- F# API changes are also reflected here.
- `PersistentView` works more similar to basic F# Actors. Main difference is passed actor param (which is `View<'Message, 'State>` instead of `Actor<'Message>`).
- PersistentActor and AtLeastOnceDelivery actors now have changed spawning functions signature - they work much more like original F# actors. This way we can leverage things such as behavior switching with mutually recursive functions. See PersistentExample.FsApi for samples.

cc #851 #841


rogeralsing:
Anyone who can review this?
This is way way out of my F# knowledge.
Anyone? @akkadotnet/developers

rodrigovidal:
I'll have a look. ASAP.

Aaronontheweb:
> This is a proposition of typed actor refs in F# API and new idea of Persistence F# plugin. My proposal is to push them (after review) to -pre version to let users play with it before official publish. Make it something like 1.1-RC1 version.

We can't really merge this into `dev` if that's what we'd like to do - this is the main branch. Either these changes go out as a planned update to all Akka.FSharp users or this gets tucked into a different feature branch that gets pulled into dev later.

Does this break the current F# API? Or is this an extension of it?

Horusiath:
@Aaronontheweb changes will break existing F# API. In terms of github repo I thnik that separate feature branch to be pulled later into `dev` sounds good. But how about NuGet packages? Could we publish this as a part of `-Pre` release channel to get first users impressions and feedback?

Aaronontheweb:
> @Aaronontheweb changes will break existing F# API.

So we committed to preserving the F# APIs in 1.0 - I know there were some issues with the F# 1.0 API with the scheduler, but we resolved those in 1.0.1.

We need to preserve the integrity of our commitment to the 1.0 API and not break it again. So I think this PR needs to be redone as an extension of the 1.0 APIs without breaking them. We can't treat F# like an experiment - it's just as important as the C# APIs.

I think the feature set et al is a good idea though, that being said. Can this be redone without breaking the F# APIs?

Horusiath:
@Aaronontheweb Concerning each of points I've mentioned on top of PR:

1. This will break the API - because we're changing from untyped to typed actor refs - there is no way around it ;) . *This could be easily fixed thou by explicitly specifying type of the messages in conflicting cases, where F# type inference is not enough (which is rare use case from my experience)* .
2. Since we leverage the information brought by typed actors, the untyped ones will be conflicting. This is potential problem when we interop with C# API and pass i.e. `IActorRef` as a message field. *Solution for this is using `typed` function on `IActorRef` which returns it's wrapped typed version.*
3. This is breaking change only because function takes 2 params instead of 3 (missing param is message type, which can now be infered directly from passed actor ref). I could split this function back into two versions (2- and 3-args) to maintain backward compatibility.
4. I don't think about this as a breaking changes, more like a bug fix. Casting exception is probably not desired behavior in case when message passed to function has invalid type.

sebfia:
This looks like a great feature to have! +1

src/core/Akka.FSharp/Spawning.fs
```diff
@@ -0,0 +1,140 @@
...
+
+module System = 
+    /// Creates an actor system with remote deployment serialization enabled.
+    let create (name : string) (config : Akka.Configuration.Config) : ActorSystem = 
+        let system = ActorSystem.Create(name, config)
+        Serialization.exprSerializationSupport system
+        system
+
+[<AutoOpen>]
+module Spawn =
```
neoeinstein: In general, I think the name and the actor continuation/expression should be the last two parameters to all of the spawning functions. That will help greatly with composability and partial evaluation. Here's how I foresee this looking:

    spawn system "spawned" <| actorOf2 (reduce dict)
    spawne system options "spawned" <@ ... @>
    spawnOpt system options "spawned" <| actorOf2 (reduce dict)

src/core/Akka.FSharp/Spawning.fs
```diff
@@ -0,0 +1,140 @@
...
+[<AutoOpen>]
+module Spawn = 
+    type SpawnOption = 
+        | Deploy of Deploy
+        | Router of Akka.Routing.RouterConfig
+        | SupervisorStrategy of SupervisorStrategy
+        | Dispatcher of string
+        | Mailbox of string
+    
+    let rec applySpawnOptions (props : Props) (opt : SpawnOption list) : Props =
```
neoeinstein: `Spawn` is `[<AutoOpen>]`, and this looks like a utility function only used by this module. I'd recommend making this function `internal` so that it doesn't pollute the namespace when opening the package.

src/core/Akka.FSharp/Spawning.fs
```diff
@@ -0,0 +1,140 @@
...
+    /// <param name="name">Name of spawned child actor</param>
+    /// <param name="f">Used to create a new instance of the actor</param>
+    let spawnObj (actorFactory : IActorRefFactory) (name : string) (f : Quotations.Expr<unit -> #ActorBase>) : ActorRef<'Message> = 
+        spawnObjOpt actorFactory name f []
+    
+    /// <summary>
+    /// Wraps provided function with actor behavior. 
+    /// It will be invoked each time, an actor will receive a message. 
+    /// </summary>
+    let actorOf (fn : 'Message -> unit) (mailbox : Actor<'Message>) : Cont<'Message, 'Returned> =
```
neoeinstein: Would it be worth changing the signature of this just a bit? It's effectively re-arranging deck chairs for this one, but the direct consumer of `actorOf` is not usually going to provide a `mailbox`. What about:

    let actorOf (fn : 'Message -> unit) : (Actor<'Message> -> Cont<'Message, 'Returned>) =
        fun mbx ->
            let rec loop() = actor {
                let! msg = mbx.Receive()
                fn msg
                return! loop()
            }
        loop()

Similar with `actorOf2` below. Again, due to associativity, this doesn't change much, but it makes it more clear that what `actorOf` is creating for you is a function which takes a mailbox and returns a continuation, which is what the above `spawn` variations actually want.

src/core/Akka.FSharp/FaultHandling.fs
```diff
@@ -0,0 +1,135 @@
...
+    
+    interface IDecider with
+        member this.Decide(e : exn) : Directive = this.Compiled.Value(e)
+    
+    interface ISurrogated with
+        member this.ToSurrogate _ = 
+            let fsp = Nessos.FsPickler.FsPickler.CreateBinary()
+            ExprDeciderSurrogate(Serialization.serializeToBinary fsp this.Expr) :> ISurrogate
+
+type Strategy =
```
neoeinstein: Convert this to a module with functions instead of a type with static methods?

src/core/Akka.FSharp/Actors.fs
```diff
@@ -0,0 +1,389 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="Actors.fs" company="Akka.NET Project">
+//     Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+//     Copyright (C) 2013-2015 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+namespace Akka.FSharp
```
neoeinstein: Reduce the indenting by changing this to `module Akka.FSharp.Actors` and adding the `[<AutoOpen>]` attribute here. If that causes you to need to move the `OptionHelper` out of the module so that it can be accessed, move it to a `Prelude.fs`.

neoeinstein:
Added several notes on style and format. Much of it is pretty minor stuff, but if we are going to make changes that include any breakage, then we should consolidate a couple of signature changes in with this too.

src/core/Akka.FSharp/FaultHandling.fs
```diff
@@ -0,0 +1,135 @@
...
+namespace Akka.FSharp
+
+open Akka.Actor
+open System
+open Microsoft.FSharp.Quotations
+open Microsoft.FSharp.Linq.QuotationEvaluation
+
+
+[<AutoOpen>]
+module Logging =
```
neoeinstein: Does this really go with the supervisor strategies below? If not, kick it out into its own file?

src/core/Akka.FSharp/Spawning.fs
neoeinstein: Other thought that I just had about this part of the API: the module name is `Spawn`, and then we have several variations on `spawn` as functions below. Could we instead rename some of them and then remove the `[<AutoOpen>]`? Instead have it be used as:

    Spawn.create system "spawned" <| Actor.ofFun (reduce dict)
    Spawn.create' system options "spawned" <| Actor.withMailbox (reduce dict)
    Spawn.ofExpr system options "spawned" <@ ... @>

To me that looks like a better way to refer to these functions, but your preference may vary.

Horusiath: Concerning first (moving options forward on function arg list) - IMO this is bad. 

- First you have a set of similar functions, each using same subset of function arguments, but depending on the function they are placed in different order (i.e. if I have `spawn <system> <name> <actor-logic>` setting, other spawn functions where this order is broken looks counter-intuitive to me).
- Second, I think it's a good practice (especially from partial application perspective) that function args should be in order from most general to most specialized. In that case list of spawn options is in fact more specific than actor behavior.

About `Spawn` module. The only reason for it is to keep code in order. Spawn naming conventions are borrowed from Erlang and used as well in Scala's akka-typed package. In your example accidental using simple `open Spawn` would end with making function names unreasonable. This could be fixed by using [RequireQualifiedAccess](https://msdn.microsoft.com/en-us/library/ee340423.aspx) attribute. But again - it's breaking existing API in heavier way, than proposed changes, while possible gains are questionable.

src/core/Akka.FSharp/FaultHandling.fs
Horusiath: This was module with functions at the beginning, but after discussion and some advises from Don Syme we decided to change it to current shape.

src/core/Akka.FSharp/Spawning.fs
neoeinstein: I agree, definitely, with the partial application perspective. I was thinking that options was more general, though. Similar to `Props` creation, this makes it more amenable to having a partially applied:

    let createWithDefaultPool = spawnOpt system [SpawnOption.Router (RoundRobinPool(10))]

I thought that it was more natural for such purposes, but this is of course easily done by consumers if they want.

On the naming, my recommendation would be for `[<RequireQualifiedAccess>]`, but I wholly understand keeping things more in line with the Erlang and Scala roots.

src/core/Akka.FSharp/FaultHandling.fs
neoeinstein: I took a look back at 85545c3cc1159c48acab9a35f615f070b82f2d60, which is where I believe the current shape came from. While I definitely agree that the previous shape was wrong, I'm not sure this shape is quite right either. This effectively becomes a thin wrapper around the C# API, and I think we can expose it in a more F#-friendly way. I'll push up a changeset shortly that incorporates what i was thinking.

neoeinstein: Here's a link to the changeset I was thinking of: neoeinstein@66f9d86.

Horusiath:
I've made some minor refactoring changes (mostly code organization - thx @neoeinstein for pointing that out). Also renamed `CanTell<>` &rArr; `ICanTell<>` and `ActorRef<>` &rArr; `IActorRef<>` to maintain naming compatibility with possible C# typed actor refs in the future.

Aaronontheweb:
after discussing this in the Akka.NET contributor's meeting today, senor @Horusiath will be releasing these changes as a separate package outside of the core Akka.NET project. Closing this PR.

PR #938, opened 2015-05-01
Added constructor for ConsistentHashingPool

Danthar:
Constructor for ConsistentHashingPool which also allows to the set the
Resizer without having to specify all defaults.


Aaronontheweb:
@Danthar for this you should be able to use the fluent interface:

```csharp
var pool = new ConsistentHashingPool(10).WithResizer(myResizer);
```

Will that work in your case?

Danthar:
Ah. Didn't realise. 
I switched from the `RoundRobinPool` to the `ConsistentHashingPool`. The former does has a constructor accepting the `nrOfInstances` and a `Resizer`. 
Thats why I thought the `ConsistentHashingPool` was missing an overload.

Aaronontheweb:
@Danthar that constructor probably shouldn't exist for `RoundRobinPool` either, but can't get rid of it now due to breaking API changes.

Issue #926, opened 2015-04-29
Akka.Persistence: dynamic journal/snapshot store resolution


Horusiath:
Right now Akka.Persistence plugin uses the same journal and snapshot store across all persistent actors. This has changed in [JVM Akka](https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/Persistence.scala) already.

The purpose of this task is to update Akka.Persistence to use dynamic journal/snapshot store resolution mechanism for persistent actor based on it's *journalPluginId* and *snapshotStoreId*. Just like the way, canonical Akka does it. This will allow us to use multiple journals and snapshot stores of different types in scope of the same actor system.

PR #945, opened 2015-05-05
Dynamic resolution of snapshot stores and journals

Horusiath:
cc #926

This PR for Akka.Persistence will allow to resolve different snapshot stores and journals depending on `SnapshotPluginId` and `JournalPluginId` values provided by persistent actors or views. This way we could handle multiple peristence providers inside single actor system.

This change is backward compatible.


woodpk:
Is there any update on when this issue will be resolved/closed?￼

PR #946, opened 2015-05-05
For review only: Async/await fixes and optimizations

nvivo:
This is a proposal **for review and discussion only** to fix the remaining async/await issues (as seen in #907 and others). I only ran the tests and they pass, didn't do any extensive testing.

The idea of this implementation is:

* Fix possible context flow issues during async/await by using CallContext for everything - the fix for xunit works, but other environments are still susceptible
* Pay the price of CallContext overhead only when needed
* Simplify the ActorTaskScheduler logic and prepare for future enhancements

This is working is by storing the entire `InternalCurrentActorCellKeeper` in the CallContext by default, unless the actor implements the `ISyncActor` marker interface. In that case, it uses only the old ThreadStatic field. The check happens only when the mailbox is about to run, so the price decreases when the throughput increases.

The reason I'm doing `ISyncOnly` instead of some form of `IEnableAsyncAwait` is because I believe most users will expect a language feature working by default. All system actors have the marker, so we avoid overhead there.

Even though there is a real drop in performance without the marker in the benchmark, most user code shouldn't notice. From my tests, the benchmark app performance should be similar to what it was with the interface.

Some notes:

* Since the ActorCell.Current is kept in CallContext, there is no logic required in the TaskScheduler to keep the state anymore
* Without async reentrancy, there is no need to pass messages to complete the task in the actor - it only receives messages when the task failed so it can throw the exception correctly
* XUnit still requires a custom SynchronizationContext because it runs in a different AppDomain - TestKitBase forces ThreadStatic so the SyncCtx can use that correctly

Let me know what you guys think and if this is a viable option.

### DIFF
--- a/src/benchmark/PingPong/ClientActorBase.cs
+++ b/src/benchmark/PingPong/ClientActorBase.cs
@@ -11,7 +11,7 @@ using Akka.Actor;
 
 namespace PingPong
 {
-    public class ClientActorBase : ActorBase
+    public class ClientActorBase : ActorBase, ISyncActor
     {
         private readonly IActorRef _actor;
         private readonly TaskCompletionSource<bool> _latch;
--- a/src/benchmark/PingPong/ClientReceiveActor.cs
+++ b/src/benchmark/PingPong/ClientReceiveActor.cs
@@ -11,7 +11,7 @@ using Akka.Actor;
 
 namespace PingPong
 {
-    public class ClientReceiveActor : ReceiveActor
+    public class ClientReceiveActor : ReceiveActor, ISyncActor
     {
         public ClientReceiveActor(IActorRef actor, long repeat, TaskCompletionSource<bool> latch)
         {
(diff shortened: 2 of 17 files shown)


rogeralsing:
To keep the noise at a minimum with yet another async await thread, some short notes:

>Fix possible context flow issues during async/await by using CallContext for everything - the fix for xunit works, but other environments are still susceptible

Do we have any known ones?

>Pay the price of CallContext overhead only when needed

We already do

>Simplify the ActorTaskScheduler logic and prepare for future enhancements

This does have some merit, but considering this introduces breaking, or rather forces every non async actor to implement a marker interface.

kekekeks:
>using CallContext for everything

We've tried it already, it significantly drops performance

kekekeks:
Oh, I see. Your approach with `UseThreadStatic` won't work.

1) Async actor's cell is saved in `CallContext`
2) Sync actor is processed on the thread, `UseThreadStatic` is set to false, `Current` is set to null
3) Task continued on the same thread (or thread from pool), `UseThreadStatic` is false, `Current` is null, `CallContext` isn't used, `ActorCell` isn't restored.

nvivo:
As discussed on gitter, let's leave 1.x as it is and wait for Akka.Typed to solve all problems.

PR #947, opened 2015-05-05
Cleanup in the Event namespace

stefansedich:
- Cleanup  some of the auto generated comments.
- Code cleanup in some places for readability and to be consistent with coding standards.

If anyone has any better ideas for some of the xmldoc comments let me know (documentation is not a strong point of mine).


stefansedich:
Just to add I killed the Obsolete Warn() method, I assume now that 1.0 is shipped we can do this?

src/core/Akka/Event/ILoggingAdapter.cs
```diff
@@ -46,10 +44,6 @@ public interface ILoggingAdapter
         /// <param name="args">The arguments.</param>
         void Warning(string format, params object[] args);
 
-        [Obsolete("Use Warning instead!")]
```
Aaronontheweb: Probably a good idea to delete `Warn` from the logging adapter interface now - any public interface changes will break the ability to do binding redirects on anything that depends on Akka core and uses this interface, so better off doing that now before stuff starts to pile up.

stefansedich: Do you mean "not a good idea"?

Aaronontheweb: No, I meant that this is a good idea. Either you end up leaving obsoleted stuff around permanently or you tear the band-aid off right away. I'm a fan of tearing the band-aid off.

Issue #941, opened 2015-05-02
Exception creating actor implementing IWithUnboundedStash


sadprofessor:
When an actor that implements IWithUnboundedStash that was resolved from DIExt (autofac in this case, although I doubt that matters) calls it's first Become, this exception is thrown:
```
[ERROR][02/05/2015 23:54:32][Thread 0012][ActorSystem(System)] An exception occurred while trying to apply plugin of type Akka.Actor.ActorStashPlugin to the newly created actor (Type = Machina.
MessagePipeline.MessageBufferingActor, Path = akka://System/user/$b/$b)
Cause: System.NotSupportedException: DequeBasedMailbox required, got: UnboundedMailbox
An (unbounded) deque-based mailbox can be configured as follows:
    my-custom-mailbox {
        mailbox-type = "Akka.Dispatch.UnboundedDequeBasedMailbox"
    }
   at Akka.Actor.Internal.AbstractStash..ctor(IActorContext context, Int32 capacity)
   at Akka.Actor.Internal.UnboundedStashImpl..ctor(IActorContext context)
   at Akka.Actor.StashFactory.CreateStash(IActorContext context, Type actorType)
   at Akka.Actor.ActorStashPlugin.AfterIncarnated(ActorBase actor, IActorContext context)
   at Akka.Actor.ActorProducerPipeline.AfterActorIncarnated(ActorBase actor, IActorContext context)
```

rogeralsing:
As the exception complains about the mailbox type, have you configured the actor to use some custom mailbox type?

Not all mailboxes support stashing, as the mailbox has to be able to prepend messages at the front when using stashing.

sadprofessor:
I haven't changed the configuration of the mailbox type. It's just a vanilla ReceiveActor with no custom configuration.

This sample actor has the same issue with resolving through DI

```
   class TestActor : ReceiveActor, IWithUnboundedStash
    {
        public IStash Stash { get; set; }

        protected override void PreStart()
        {
            base.PreStart();
            Become(FirstState);
        }

        private void FirstState()
        {
            
        }
    }
```

When resolving it through 
```
Context.ActorOf(Context.System.GetExtension<DIExt>().Props(typeof(TestActor)))
```

it errors, when creating it directly through
```
Context.ActorOf(Props.Create(() => new TestActor()));
```

it's fine

The registration in autofac is just:
```
builder.RegisterType<TestActor>().AsSelf();
```

Cheers

Danthar:
[...]

PR #950, opened 2015-05-06
Fix for DIActorProducer #941

Danthar:
Now returns the proper ActorType instead of the Type of the
depencyResolver.


Aaronontheweb:
thanks @Danthar!

Issue #940, opened 2015-05-02
akka.loglevel = OFF doesn't work


nvivo:
[Pigeon.conf states](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Configuration/Pigeon.conf#L27) `akka.loglevel = OFF` should work, but actually causes an exception:

```
Unknown LogLevel: "OFF". Valid values are: "DEBUG", "INFO", "WARNING", "ERROR"
```

rogeralsing:
Fix for this have been merged. closing

PR #951, opened 2015-05-06
"virtual" log level for OFF #940

kekekeks:
#940


kekekeks:
It may look like a hack, but shouldn't break anything, just no messages will be logged.

The reason why LogLevel.Off shouldn't be exposed is that it shouldn't ever be used by user. User shouldn't create loggers that won't log anything and shouldn't log messages with this level.
I'm not sure that LoggingBus should be aware of that and don't start any logging at all.

rogeralsing:
Scala does the same thing: 
```scala
private final val OffLevel = LogLevel(Int.MinValue)
```

If it works for them, it works for me :)

Issue #849, opened 2015-04-13
ClusterRouterGroup not removing downed nodes


annymsMthd:
We have traced this one to line 78 in https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Routing/RoutedActorCell.cs

The type of the routee is an ActorSelectionRoutee, not ActorRefRoutee,  and it is being ignored and not removed. 

Also line 173 in https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Routing/RouterConfig.cs looks like where the routees are created.

Aaronontheweb:
Related to #743

Aaronontheweb:
I'll look into this one

Aaronontheweb:
First glance - still having lots of issues with my multi-node tests aimed at this behavior even after making some changes to `RouterConfig`. Not totally sure what the issue is yet.

annymsMthd:
Working on this in a branch on my repo. This looks like it goes deep. I've fixed up the router stuff but it looks like it exposes some issues with ActorSelection. I have tests failing on Group Routers and i chased it down the rabbit hole.

akka:
https://github.com/akka/akka/blob/8485cd2ebb46d2fba851c41c03e34436e498c005/akka-actor/src/main/scala/akka/actor/ActorSelection.scala

us:
https://github.com/akkadotnet/akka.net/blob/8ac06436f53df7e1622e7a107abc6c2823d72799/src/core/Akka/Actor/ActorSelection.cs

I'm way too tired for recursion tonight but I will start on this fresh tomorrow.

annymsMthd:
#953

Aaronontheweb:
Is this considered to be "fixed" now?

annymsMthd:
The dlls with the nightly package are still marked 1.0.1 and I'm not seeing
the changed messages. I need to check this when I get home

Aaronontheweb:
Argh! Damn TeamCity. I'm getting weird stuff from the feed too.
In that case, go to the build server and download the artifacts from there.

annymsMthd:
I'm investigating

annymsMthd:
This is working properly now in our cluster with the nightly build.

PR #953, opened 2015-05-07
#849 Fixes to ClusterRouterGroup

annymsMthd:


src/core/Akka/Routing/RoutedActorCell.cs
```diff
@@ -126,20 +122,17 @@ public override void Start()
...
+                        AddRoutees(Vector.Fill<Routee>(nrOfRoutees)(() => pool.NewRoutee(_routeeProps, this)));
                 })
-                .With<Group>(r =>
+                .With<Group>(group =>
                 {
-                    var routees = _routerConfig.GetRoutees(this).ToArray();
-                    AddRoutees(routees);
+                    var paths = group.Paths;
+                    if (paths.NonEmpty())
+                        AddRoutees(paths.Select(p => group.RouteeFor(p, this)).ToList());
```
annymsMthd: Down the rabbit hole.

annymsMthd:
Going to squash these commits when i get home tonight. @Aaronontheweb mind giving this a look over when you can?

Aaronontheweb:
yeessir - can do! I will personally be thrilled to see this spec pass. Been one that's eluded me for a while!

annymsMthd:
I branched off this and merged in my app domain changes. Makes it really
easy with a debugger attached

Aaronontheweb:
once we get an overload with the `Scheduler` running on a `DedicatedThreadPool` things should be easier for the `AppDomain` implementation - I strongly suspect threadpool saturation was at play with the inconsistent behavior I got.

annymsMthd:
I'm going to be opening up a PR for it soon just to keep it around until we can get them to consistently pass. I also added the teamcity output.

src/core/Akka/Util/Internal/ArrayExtensions.cs
```diff
@@ -27,6 +27,16 @@ public static bool IsNullOrEmpty(this Array obj)
         }
 
         /// <summary>
+        /// Determines if an array is not null or empty.
+        /// </summary>
+        /// <param name="obj">The array to check.</param>
+        /// <returns>True if not null or empty, false otherwise.</returns>
+        public static bool NonEmpty(this Array obj)
```
Aaronontheweb: Wouldn't linq `.Any()` work for this? I just realized that right after I closed the PR :\

src/core/Akka/Util/Vector.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Akka.Util
+{
+    public static class Vector
```
Aaronontheweb: Could you add some XML-DOC comments for this on a subsequent PR @annymsMthd ?

annymsMthd: you got it

PR #959, opened 2015-05-08
Add appdomain to MultiNode tests [Not Ready for Merge]

annymsMthd:


annymsMthd:
This one may be more trouble than it's worth. I've been working on it for a bit and I'm still seeing weird behavior

PR #963, opened 2015-05-09
Changes in Akka.Persistence SQL backend

Horusiath:
- I've changed Akka.Persistence.SqlServer to use asynchronous/non-blocking version of the ADO.NET where it's possible. 
- This PR also introduces `Akka.Persistence.Sql.Common` - project which may be used for faster development of SQL-based persistence plugins.
- Minor bugfix for `AsyncWriteJournal` - issuse with null EventStream reference in task continuation on delete messages.

These changes should not break existing API.


rogeralsing:
:+1:

Issue #944, opened 2015-05-04
Plugin proposal: Akka.Persistence.PostgreSQL


davidflerlage:
I just finished the initial implementation of a PostgreSQL persistence plugin.  I am hoping to submit a pull request later today or tomorrow.  If you are curious, you can find the code here: https://github.com/davidflerlage/akka.net/tree/akka-persistence-postgresql.

This plugin is largely a port of the Sql Server plugin, with some inspiration from the JVM Akka PostgreSQL plugin, https://github.com/okumin/akka-persistence-sql-async.  There is some duplication between the Sql Server and PostgreSql plugins that we might want to address.  I imagine that any future sql plugins will also need some of the same code.

I have a question about the best way to handle database initialization on the build server during automated testing.  I see two main options:

1) Drop and create the database each time.
2) Drop and recreate just the tables each time.

In either case, I would provide a script.

The Sql Server plugin uses LocalDb, so it can easily recreate the database each time the tests run.  I do not think PostgreSQL has a LocalDb equivalent, so that's not really an option here.  I also don't think integrated security is an option, so user credentials need to be persisted somewhere.

I am leaning towards option 1, but I am eager to hear everyone's thoughts.

Horusiath:
@davidflerlage The Akka.Persistence.SqlServer plugin will be rewritten to using ADO.NET async API soon - we know that this could probably introduce some problems with testing. Also we decided that only limited set of the persistence plugins should be part of akka core repository and have official support. @akkadotnet/developers should we include PostrgreSQL to our contrib repo and give support to it?

annymsMthd:
+1

skotzko:
+1

taswar:
+1

davidflerlage:
I refactored my plugin to use the new Sql.Common project.  I am pretty much done with this plugin, but I am experiencing some random test failures.  I am trying to track the issue down.  The tests run fine when run individually and in small groups, but not when all of them are run together.

Horusiath:
Merged

PR #971, opened 2015-05-13
PostgreSQL persistence plugin for both event journal and snapshot store

davidflerlage:
See issue #944

 -Largely ported from Akka.Persistence.SqlServer plugin

 -Event journal and snapshot store specs passing

 -Added separate build task for PostgreSql tests

 -Added initial draft of readme and nuspec file

 -There is still one potentially unresolved issue pertaining to passing in a connection string to the test runner inside build.fsx.  Right now, I put the connection string (with credentials) in the app.config of the test project.  There is currently no way to pass in a custom connection string or custom credentials using the test runner inside build.fsx.  I was talking to @Aaronontheweb last week, and one of his suggestions was to create a "PersistenceTestRunner" that would modify the app.config before the tests are run.  I wanted to double check before I wrote a test runner.  Comments?

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -217,7 +217,8 @@ Target "RunTests" <| fun _ ->
     let xunitTestAssemblies = !! "src/**/bin/Release/*.Tests.dll" -- 
                                     "src/**/bin/Release/Akka.TestKit.VsTest.Tests.dll" -- 
                                     "src/**/bin/Release/Akka.TestKit.NUnit.Tests.dll" --
-                                    "src/**/bin/Release/Akka.Persistence.SqlServer.Tests.dll"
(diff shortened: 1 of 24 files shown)


Horusiath:
Great job! In my opinion, code looks good. As I understand, you've run the `Akka.Persistence.TestKit` tests and they were passing (I'm asking to make sure, plugin is Akka.Persistence-compatible)? If so I think, we could merge this PR and resolve problem of the test runner as a separate issue.

davidflerlage:
Yes, I ran the tests from Akka.Persistence.TestKit, and everything is passing.  For what it's worth, I ran them a bunch of times without any issues.

davidflerlage:
Wow, that was quick.  Thank you very much.  Now, onto the next issue!

PR #976, opened 2015-05-14
Fixes for the cluster and remoting dispatchers

annymsMthd:


annymsMthd:
Currently just for discussion. More work needs to be done here.

src/core/Akka/Configuration/Pigeon.conf
```diff
@@ -238,55 +238,35 @@ akka {
...
-      attempt-teamwork = on
- 
-      # If this dispatcher requires a specific type of mailbox, specify the
-      # fully-qualified class name here; the actually created mailbox will
-      # be a subtype of this type. The empty string signifies no requirement.
-      mailbox-requirement = ""
+      type = ForkJoinDispatcher
+      throughput = 1
+      dedicated-thread-pool{ #settings for Helios.DedicatedThreadPool
+          thread-count = 1 #number of threads
```
annymsMthd: Should probably bump this up. I was unsing 1 for debugging

src/core/Akka/Configuration/Pigeon.conf
```diff
@@ -238,55 +238,35 @@ akka {
...
-			thread-count = 3 #number of threads
-			#deadlock-timeout = 3s #optional timeout for deadlock detection
-			threadtype = background #values can be "background" or "foreground"
-		}
-	}
+    default-fork-join-dispatcher{
+        type = ForkJoinDispatcher
+        throughput = 100
+        dedicated-thread-pool{ #settings for Helios.DedicatedThreadPool
+            thread-count = 40 #number of threads
```
annymsMthd: should probably bring this one back down

annymsMthd:
Checking out failing tests

annymsMthd:
Looks like a couple of racy tests i need to hunt down. I'm going to implement the custom thread pool tomorrow that grows, deadlock detection, etc tomorrow

rogeralsing:
So what I dont get is this:

If we run the scheduler on its own thread, like JVM does.
And we run all Akka System actors in a dedicated threadpool (per module)
Why do we need to mess with the default dispatcher?
If _all_ of the sensitive stuff is moved away from the thread pool. it should be fine, no?

The current DedicatedThreadPool implementation is horrible, its a hack and can in no way compete with the real ThreadPool in terms of throughput.

Im pretty sure this will gimp the performance of user actors severly if we change the default dispatcher instead of isolating the critical parts of the system.

nvivo:
I agree with Roger. A single dedicated thread should solve the problem for schedulers. In fact, even some critical actors like the ones handling cluster heartbeats could run in this thread without problems.

Trying to develop a full blown threadpool is an overkill for this problem.

annymsMthd:
The threadpool is a place for short running tasks.

annymsMthd:
Take a node with 1000+ actors and start it up in a cluster while messages
are coming into it. You'll see why

rogeralsing:
@annymsMthd I still dont get it.
most messages are short running tasks.

Why is this a problem?

annymsMthd:
Even with short running tasks the ThreadPool queue is filling up with work. The threadpool also has to decide to add more threads while this is going on.

annymsMthd:
http://doc.akka.io/docs/akka/snapshot/scala/dispatchers.html I'm going to point out the fact that akka uses a fork-join as the default again.

src/examples/Cluster/Samples.Cluster.Simple/Program.cs
```diff
@@ -6,7 +6,10 @@
 //-----------------------------------------------------------------------
```
annymsMthd: I'll revert this near the end

rogeralsing:
There are two incorrect assumptions here:

1) Our fork-join is not a fork-join. its a hack
2) JVM uses the forkjoin because of the global queue used for the default JVM threadpool causes contention and this cause horrible throughput

annymsMthd:
I'll set the default back to threadpool and finish my work up here. Our cluster config, my company, is set so we are gtg. I'm sure this issue will come up in the future but for the time now I'll let it go and just clean up the work with the dedicated pool as an option.

annymsMthd:
I'm going to change this to just the fixes on the remoting/clustering dispatchers. #982 solves the issue with heartbeat scheduling.

src/core/Akka/Helios.Concurrency.DedicatedThreadPool.cs
```diff
@@ -400,9 +404,10 @@ public PoolWorker(WorkerQueue work, DedicatedThreadPool pool, bool errorRecovery
                 _pool = pool;
                 _workQueue = _work.WorkQueue;
                 _work.ReplacePoolWorker(this, errorRecovery);
-
+                
                 _thread = new Thread(() =>
                 {
+                    Thread.CurrentThread.Name = pool.Settings.Name + "_1";
```
annymsMthd: should give this the worker number instead

annymsMthd:
Gonna go ahead and close this PR for a new one.

PR #978, opened 2015-05-14
Add PipeTo for non-generic Tasks for exception handling

adamhathcock:
This feels like something I want to do.  Let me know if this is a bad idea and I won't be doing in my own code.

### DIFF
--- a/src/core/Akka/Actor/PipeToSupport.cs
+++ b/src/core/Akka/Actor/PipeToSupport.cs
@@ -24,10 +24,24 @@ namespace Akka.Actor
             sender = sender ?? ActorRefs.NoSender;
             return taskToPipe.ContinueWith(tresult =>
             {
-                if(tresult.IsCanceled  || tresult.IsFaulted)
+                if (tresult.IsCanceled || tresult.IsFaulted)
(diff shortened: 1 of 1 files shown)


nvivo:
The idea of PipeTo is to pipe the result of the task to the recipient. If the task doesn't have a result, what are you piping? Only the failure?

adamhathcock:
Correct.  I'd want the parent is try to handle the error and not just let it die in Task-land

rogeralsing:
Looks good, this is in line with what JVM Akka does.
´´´scala
  final class PipeableFuture[T](val future: Future[T])(implicit executionContext: ExecutionContext) {
    def pipeTo(recipient: ActorRef)(implicit sender: ActorRef = Actor.noSender): Future[T] = {
      future onComplete {
        case Success(r) ⇒ recipient ! r
        case Failure(f) ⇒ recipient ! Status.Failure(f)
      }
      future
    }
```

stefansedich:
I lile it! 

Do we need to have a test around it?

nvivo:
Agreed. Makes sense to me.

PR #982, opened 2015-05-15
Dedicated thread Scheduler

rogeralsing:
This is a PoC to see what happens when the scheduler runs on a dedicated thread.
When running with the TaskBasedScheduler (the default) and the thread pool gets flooded, Akka.Remote can no longer push Heartbeats through and endpoints gets disassociated.

When using this scheduler, the disassociation no longer occur.

I have tested this by letting the chat application spin up 600 blocking actors on a special command.

scheduler

That turned out to delay Heartbeats by almost 40 seconds.

When running the new scheduler, everything works fine in my test.

### DIFF
--- a/src/core/Akka/Actor/Internals/ActorSystemImpl.cs
+++ b/src/core/Akka/Actor/Internals/ActorSystemImpl.cs
@@ -51,11 +51,11 @@ namespace Akka.Actor.Internals
             if(config == null)
                 throw new ArgumentNullException("config");
 
-            _name = name;
-            ConfigureScheduler();
+            _name = name;            
             ConfigureSettings(config);
             ConfigureEventStream();
             ConfigureProvider();
+            ConfigureScheduler();
             ConfigureSerialization();
             ConfigureMailboxes();
             ConfigureDispatchers();
@@ -127,7 +127,7 @@ namespace Akka.Actor.Internals
 
         private void ConfigureScheduler()
         {
-            _scheduler = new TaskBasedScheduler();
+            _scheduler = new DedicatedThreadScheduler(this);;// new TaskBasedScheduler();
         }
 
         /// <summary>
--- a/src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
+++ b/src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
@@ -32,7 +32,7 @@ namespace Akka.Actor
                     while(_workQueue.TryDequeue(out work))
                     {
                         //has work already expired?
-                        if (work.UtcTickExpires < now)
+                        if (work.TickExpires < now)
                         {
                             work.Action();
                         }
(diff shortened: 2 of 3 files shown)


rogeralsing:
cc @Aaronontheweb @annymsMthd

src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
```diff
@@ -0,0 +1,158 @@
...
+            var expected = DateTime.UtcNow + initialDelay;
+            Action executeAction = () =>
+            {
+                if (token.IsCancellationRequested)
+                    return;
+
+                try
+                {
+                    var actual = DateTime.UtcNow;
+                    Console.WriteLine("InternalScheduleTellOnce diff {0}", (actual - expected).TotalMilliseconds);
```
rogeralsing: debugging info just to see how this performs..

annymsMthd:
This does the trick.

src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
```diff
@@ -0,0 +1,158 @@
...
+        protected override void InternalScheduleRepeatedly(TimeSpan initialDelay, TimeSpan interval, Action action, ICancelable cancelable)
+        {
+            var cancellationToken = cancelable == null ? CancellationToken.None : cancelable.Token;
+            InternalScheduleRepeatedly(initialDelay, interval, action, cancellationToken);
+        }
+
+
+        private void InternalScheduleOnce(TimeSpan initialDelay, Action action, CancellationToken token)
+        {
+            var expected = DateTime.UtcNow + initialDelay;
```
Aaronontheweb: can we modify this to start using the monotonic clock per #846?

rogeralsing:
I've updated the code to use monotonic clock, and removed the debug writelines.

We need some sort of mechanism to kill the thread when the system exits.
Do we have one that would be appropriate here?
`system.what?`

rogeralsing:
We should probably pass a Config to the pool to, and read the precision value from there

Aaronontheweb:
@rogeralsing you can use the termination hooks for that

rogeralsing:
`system.TerminationTask` or something else?

rogeralsing:
Even if this is largely untested, it's still better than what we have right now.
Do we pull this in or should someone port the JVM scheduler instead?

src/core/Akka/Actor/Internals/ActorSystemImpl.cs
```diff
@@ -51,11 +51,11 @@ public ActorSystemImpl(string name, Config config)
             if(config == null)
                 throw new ArgumentNullException("config");
 
-            _name = name;
-            ConfigureScheduler();
+            _name = name;            
             ConfigureSettings(config);
             ConfigureEventStream();
             ConfigureProvider();
+            ConfigureScheduler();
```
rogeralsing: provider and settings needs to be in place before we can config the scheduler

annymsMthd:
I'm all for this. Tests are good. Did any check the multinode tests?

rogeralsing:
Nope, my VS + xUnit is giving me a really hard time ATM. cannot compile any tests myself right now

Aaronontheweb:
I'm all for bringing this in. Death to unnecessary disassociations!

Aaronontheweb:
I'll run the MultiNode tests now

PR #985, opened 2015-05-17
small sharding progress

rogeralsing:
Checking in my latest sharding updates


rogeralsing:
merging this myself as the sharding branch is work in progress only

PR #990, opened 2015-05-19
Added generic extensions to EventStream subscribe/unsubscribe.

nvivo:
EventStream is one of the few pieces that don't have generic counterparts.

This patch allows for:
```cs
EventStream.Subscribe<SomeMessage>(Self);
EventStream.Unsubscribe<SomeMessage>(Self);
```
instead of
```cs
EventStream.Subscribe(Self, typeof(SomeMessage));
EventStream.Unsubscribe(Self, typeof(SomeMessage));
```

I was going to add these methods directly to EventStream, but it seems these generic versions are usually extensions methods. One caveat with the extension method is that most code don't require an `using Akka.Event`, so people may miss it entirely.

Please let me know if they should be moved.

### DIFF
--- a/src/core/Akka/Akka.csproj
+++ b/src/core/Akka/Akka.csproj
@@ -217,6 +217,7 @@
     <Compile Include="Event\Error.cs" />
     <Compile Include="Event\EventBus.cs" />
     <Compile Include="Event\EventStream.cs" />
+    <Compile Include="Event\EventStreamExtensions.cs" />
     <Compile Include="Event\ILogMessageFormatter.cs" />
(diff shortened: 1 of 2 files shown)


rogeralsing:
Looks good to me :+1:

Aaronontheweb:
> One caveat with the extension method is that most code don't require an using Akka.Event, so people may miss it entirely.

Is the `EventStream` in the Akka.Event namespace too?

rogeralsing:
Yes

Aaronontheweb:
Cool - I think that should be fine then.

PR #991, opened 2015-05-19
Fix the url for Building Akka docs

lnmunhoz:
The old url seems not exists anymore.


skotzko:
nice catch @lnmunhoz !

PR #993, opened 2015-05-20
Convert null to NoSender.

rogeralsing:
This PR makes it possible to explicitly pass null as `sender` to `Tell` as it is a fairly common mistake and we have no real reason to throw on it.

This PR also fixes some context leakage in the testkit

### DIFF
--- a/src/core/Akka.TestKit/TestKitBase.cs
+++ b/src/core/Akka.TestKit/TestKitBase.cs
@@ -110,6 +110,12 @@ namespace Akka.TestKit
             {
                 InternalCurrentActorCellKeeper.Current = (ActorCell)((ActorRefWithCell)testActor).Underlying;
             }
+            else if(!(this is TestProbe)) 
+                //HACK: we need to clear the current context when running a No Implicit Sender test as sender from an async test may leak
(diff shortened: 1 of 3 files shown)


src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -110,6 +110,12 @@ private TestKitBase(ITestKitAssertions assertions, ActorSystem system, Config co
             {
                 InternalCurrentActorCellKeeper.Current = (ActorCell)((ActorRefWithCell)testActor).Underlying;
             }
+            else if(!(this is TestProbe))
```
rogeralsing: This is obviously ugly in every possible way. but the testkit currently requires two different behaviors based on type.
(it could be extracted to a virt method, but then again, virt methods from ctors and all that.)

Aaronontheweb: Could you put a //TODO: on this line, letting someone in the future know why it's there?

rogeralsing: there is a //HACK: on the line below already :)

Issue #987, opened 2015-05-18
Plugin Proposal: Akka.Persistence.Cassandra


LukeTillman:
I'm almost finished with an Akka.Persistence journal and snapshot store implementation for Cassandra and I wanted to open this issue for any discussion.  I'm hoping to squash things down to a single commit and send a PR in the next day or so. You can find the code in a branch here:

https://github.com/LukeTillman/akka.net/tree/akka-persistence-cassandra

A couple things of note:

- The plugin is making use of `async` and `await` because writing things with continuations would have been *really* complicated in a few spots. I've tried to follow along with some of the issues here around `async`/`await` usage in Akka.NET and it seems like they are OK to use?
- The tests assume you have a local Cassandra cluster (which can be a single node) running at `127.0.0.1`, but I'm hoping to do some investigation into providing a way to spin up a cluster from scratch in any environment (probably using the [CCM tool](https://github.com/pcmanus/ccm)).

Any thoughts? Objections? Suggestions?

skotzko:
my first thought: :+1:

Horusiath:
@LukeTillman nice! Can't wait for your PR. Concerning your points:

- There shouldn't be any problems with async/await used from `AsyncWriteJournal` derivatives.
- Right now we only decided that each persistence tck plugin specs which needs non standard configuration should have it's own task in `build.fsx` file (sql server and postgres plugins already have their own). We should think, what better solution can we find for this in the future.

LukeTillman:
Closing since the PR has been merged. Thanks!

PR #995, opened 2015-05-20
Cassandra persistence plugin (issue #987)

LukeTillman:
- Based partially on the great Scala version by Martin Krasser (https://github.com/krasserm/akka-persistence-cassandra)

- Modified from Scala version:
    * Different schema for the Journal to try and avoid reading tombstones if using permanent deletions
    * Ensures batches of writes in the Journal will only go to one partition to ensure atomicity and isolation
    * Share ISession instance from DataStax driver across journal and session by default
    * Make ISession resolution pluggable (allows for more advanced creation options and possible DI integration)

- Implementations for both Journal and Snapshot Store passing specs

- Readme with instructions for setup

- Separate task in build script for running tests


rogeralsing:
Looks great!

rogeralsing:
btw. FYI, We will start moving persistence modules into their own repositories soon. so they can have their own release cycles and issue tracker.

PR #1002, opened 2015-05-21
Dedicated Thread Scheduler Tests

rogeralsing:
* Removed the old TaskBasedScheduler
* Updated TaskBasedScheduler tests for the new scheduler
* Removed Skip directive on two tests that was marked as unstable due to threadpool timings.
They work fine locally, lets see how they perform on the build server.
* Added logging to the scheduler so that failed actions are no longer silently ignored.


rogeralsing:
Tests seems to pass, including previously problematic ones.

stefansedich:
Look good @rogeralsing I am merging this in so that I can update my PR and not cause conflicts to this PR.

Issue #788, opened 2015-04-02
akka.net.io namespace 


bobanco:
Hi Guys, 

Firstly, i want to say thanks to all who did effort for porting this nice framework to .net community, i want to use the actor model to build fault tolerant telecom project which perfectly fits my requirements, since java/scala has the akka.io package in order to work with underlying TCP,UDP sockets i was not able to find the corresponding implementation in akka.net, so i want to ask is there any progress over the IO namespace or any plans for this to be ported in future? 

Thanks in advance,
Bob.

Aaronontheweb:
hey @bobanco - welcome to the project!

@Horusiath has done some exploratory work on Akka.IO and to be honest it's an essential ingredient for some of the important mobile scenarios we want to enable (running Akka.NET clients on Xamarin / Windows Phone that can communicate with an Akka.NET cluster.)

**We will port Akka.IO to Akka.NET**, but at the moment don't have any dedicated resources for it. The core team is pretty focused on Akka.NET core, Akka.Cluster, and Akka.Persistence at the moment - but our contributor base is expanding every week and this would be a project we'd be happy to see some  pull requests for.

willieferguson:
I have completed most of the Akka IO port (TCP driver only). I'm busy porting the remainder of the Akka IO specs, and ironing out some inconsistencies between .NET IO and JVM IO. But it should be done in the next couple of days.

https://github.com/willieferguson/akka.net/tree/AkkaIO

rogeralsing:
@willieferguson that is awesome!! :+1:

bobanco:
@willieferguson Thanks for your effort to port the IO part, i will test the  TCP driver and will let you know if there are any issues.

willieferguson:
@bobanco The current branch on my fork is broken. I've ported most of the test (not all pushed yet). Once all the tests are ported, I will experiment with various implementation of SocketChannel - the adapter for .NET sockets. I should have an working alpha version next week.

I've also started with the Akka Streams port and plan to have an alpha version in June.

Horusiath:
@willieferguson How much of the Akka.Streams did you ported already? There are multiple attempts for this, including mine.

Horusiath:
@willieferguson would you consider a PR for this one? If it's not finished yet but it's close-to-complete point, I think we could make a separate branch for it and make able to contribute from rest of the community. How do you think @akkadotnet/developers ?

willieferguson:
[...]

PR #1023, opened 2015-06-02
Initial Akka IO port with TCP driver

willieferguson:
This is a line-by-line port of the Akka IO TCP driver as discussed in #788.

### Known issues
This is an early alpha version with probably many issues. These are the ones I know about:
#### SocketChannel
The .NET BCL does not have a SocketChannel class. To keep the port close to JVM Akka, I've created an SocketChannel adapter. The current implementation use blocking calls, which throws an exceptions when the socket is set to non blocking. This is likely to cause some performance issues. I will create load tests and experiment with a implementation based on SocketAsyncEventArgs. 
#### SelectionHandler
Socket.Select is currently used, which requires the allocation of lists for each call, which will cause a huge amount of allocations. This is necessary for the current  SocketChannel implementation, and will likely be eliminated once SocketChannel is based on SocketAsyncEventArgs.
#### File IO
I have no requirements for File IO, but might be convinced to port it if required.
#### Unit Tests
Not all the specs have been ported yet. I plan port the remaining ones in the near future.

### DIFF
--- a/src/core/Akka/IO/ByteBuffer.cs
+++ b/src/core/Akka/IO/ByteBuffer.cs
@@ -1,0 +1,115 @@
+//-----------------------------------------------------------------------
+// <copyright file="ActorBase.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+//     Copyright (C) 2013-2015 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+
+namespace Akka.IO
+{
+    public class ByteBuffer
+    {
+        private byte[] _array;
+        private int _limit;
+        private int _position;
+        private ByteOrder _order;
+
+        public ByteBuffer(byte[] array)
+            : this(array, 0, array.Length)
+        {
+        }
+
+        public ByteBuffer(byte[] array, int offset, int length)
+        {
+            _array = array;
+            _position = offset;
+            _limit = offset + length;
+        }
+
+        public void Clear()
+        {
+            _position = 0;
+            _limit = _array.Length;
+        }
+
+        public void Limit(int maxBufferSize)
+        {
+            _array = new byte[maxBufferSize];
+            _limit = maxBufferSize;
+        }
+
+        public void Flip()
+        {
+            _limit = _position;
+            _position = 0;
+        }
+
+        public bool HasRemaining
+        {
+            get { return _position < _limit; }
+        }
+
+        public int Remaining
+        {
+            get { return _limit - _position; }
+        }
+
+        public byte[] Array()
+        {
+            return _array;
+        }
+
+        public void Put(byte[] src)
+        {
+            var len = Math.Min(src.Length, Remaining);
+            System.Array.Copy(src, 0, _array, _position, len);
+            _position += len;
+        }
+
+        public static ByteBuffer Wrap(byte[] array, int start, int len)
+        {
+            return new ByteBuffer(array, start, len);
+        }
+        public static ByteBuffer Wrap(byte[] array)
+        {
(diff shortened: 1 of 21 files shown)


src/core/Akka/IO/ByteBuffer.cs
```diff
@@ -0,0 +1,115 @@
...
+//     Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+//     Copyright (C) 2013-2015 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+
+namespace Akka.IO
+{
+    public class ByteBuffer
```
Horusiath: @Aaronontheweb how does it compare to buffer pools in Helios? Should we stay with that or maybe we could reuse Helios implementation?

src/core/Akka/IO/IO.cs
```diff
@@ -0,0 +1,16 @@
...
+//     Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+//     Copyright (C) 2013-2015 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using Akka.Actor;
+
+namespace Akka.IO
+{
+    public abstract class Extension : IExtension
```
Horusiath: My suggestion is to rename this to `IOExtension` since while this is clearly logical in scope of the Akka.IO, the name could be missleading for the people using intelli sense while typing `Extension` when using akka.

src/core/Akka/IO/SelectionHandler.cs
```diff
@@ -0,0 +1,377 @@
...
+                    return false;
+                };
+            }
+        }
+
+        /* 
+         * Special supervisor strategy for parents of TCP connection and listener actors.
+         * Stops the child on all errors and logs DeathPactExceptions only at debug level.
+         */
+        private class ConnectionSupervisorStrategyImp : OneForOneStrategy
```
Horusiath: typo ;) I think it should be `ConnectionSupervisorStrategyImpl`

src/core/Akka/Util/ByteIterator.cs
```diff
@@ -0,0 +1,471 @@
...
+            private int _from;
+
+            public ByteArrayIterator(byte[] array, int @from, int until)
+            {
+                _array = array;
+                _from = @from;
+                _until = until;
+            }
+
+            public override int Len
```
Horusiath: While this is actually direct compatibility from Akka, my vote is to get rid of abbreviations - we should keep close to C# naming conventions in C# code. cc/ @akkadotnet/developers

Horusiath:
First of all: this is great contribution! :D I've shared with few of my thoughts, but they are mostly concerning style and not the main implementation.

Another question to @akkadotnet/developers : are we gonna to merge this into dev branch directly, or create a separate one and finish the remaining pieces of the namespace (and also complete any possibly missing parts from the TCP implementation) before merging?

Aaronontheweb:
@Horusiath this should go into a separate branch before merging, since Akka.IO goes into the core Akka module.

src/core/Akka/IO/ByteBuffer.cs
Aaronontheweb: This is pretty similar, although Helios' `ByteBuffer` offers more methods for doing things like "just add an integer to my byte array" and takes care of the underlying bit conversion.

I think using this class is probably fine. Anyone who's going to be doing socket IO with Akka.NET is already going to be familiar with byte-formatting their messages.

willieferguson:
@Aaronontheweb please create a branch that I can use for the PR

rogeralsing:
I've created a branch for this: akka-io

PR #1029, opened 2015-06-02
Revert ILoggingAdapter Warn method removal

Aaronontheweb:
I made a mistake on #947 - should not have approved the removal of the `ILoggingAdapter.Warn` method. Needs to stay in there until next major version upgrade per our change management rules around breaking changes.


stefansedich:
Doh! merging this one now.

PR #1030, opened 2015-06-02
close #1004 - Akka.Persistence now added as dependency to all Akka.Persistence.* nuget packages

### LINKED ISSUE #1004: Akka.Persistence.SQLServer nuget package doesn't take dependency on Akka.Persistence package
Need to update the FAKE script to make sure that all Akka.Persistence implementations take an explicit NuGet dependency on Akka.Persistence.

Aaronontheweb:
Resolves #1004.


annymsMthd:
might need a rebase here

Aaronontheweb:
@annymsMthd sure thing, I'll take care of that

Issue #1036, opened 2015-06-03
NoSender actor serialization problems


ssoshchepkov:
`NoSender` actor has two serialization related bugs:

1. `Serialization.SerializedActorPath` for `NoSender` returns 'akka://all-systems' instead of 'akka://all-systems/NoSender' because `NoSender` actor's name is a part of `RootActorPath`.
2. When `RemoteActorRefProvider` is used as an actor provider, deserialization of `NoSender` causes `RemoteTransportException`. It happens because `public IActorRef ResolveActorRef(string path)` method of `RemoteActorRefProvider` expects an empty string for `NoSender` actor's path, not 'akka://all-systems' nor 'akka://all-systems/NoSender'.

rogeralsing:
Closed with #1041

PR #1037, opened 2015-06-03
Fixes for NoSender serialization bugs

ssoshchepkov:
Fixes for `NoSender` actor serialization bugs described in #1036.

### DIFF
--- a/src/core/Akka.Remote/RemoteActorRefProvider.cs
+++ b/src/core/Akka.Remote/RemoteActorRefProvider.cs
@@ -282,7 +282,7 @@ namespace Akka.Remote
 
         public IActorRef ResolveActorRef(string path)
         {
-            if (path == "")
+            if (path == "" || path == ActorRefs.NoSender.Path.ToStringWithAddress())
                 return ActorRefs.NoSender;
 
             ActorPath actorPath;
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
thanks, @ssoshchepkov - looks like we've overlooked this one.

rogeralsing:
Im not sure this fix is correct.
The scala version of the code uses the old "/nosender" path.
I think there could be more to this than we see here

Aaronontheweb:
@rogeralsing ok, I reverted it. I'll let you take a closer look and see what's really going on.

rogeralsing:
Turns out that its `NoBody` that uses that notation in scala.
`NoSender` in scala is just a wrapper for `null`

@ssoshchepkov if you want, could you adapt the PR to treat NoSender as null instead?
It's better if we stay as close as possible to the Scala code.

Aaronontheweb:
@rogeralsing won't that conflict with your PR that turns `null` into `NoSender`? :p

rogeralsing:
That code could just go away if we do this.
There is no reason to have weird comparisons all over the codebase if nosender really means null imo.

Or?

ssoshchepkov:
@rogeralsing I've made changes to treat `NoSender` as `null`. Could you reopen this PR? Or should I create another?

rogeralsing:
I think you have to send a fresh one. I dont think a PR can be re-opened once merged (as this was merged and then a revert patch applied)

PR #1040, opened 2015-06-03
pass exception to NLogLogger exception logging

ashic:
Currently, NLogLogger simply logs exception messages as strings. This commit passes the exception to NLog as well. This allows clients to use ${exception:...} layout renderers to choose if they want to log exception details (including stack traces) or not.

### DIFF
--- a/src/contrib/loggers/Akka.Logger.NLog/NLogLogger.cs
+++ b/src/contrib/loggers/Akka.Logger.NLog/NLogLogger.cs
@@ -34,7 +34,7 @@ namespace Akka.Logger.NLog
         /// </summary>
         public NLogLogger()
         {
-            Receive<Error>(m => Log(m, logger => logger.Error("{0}", m.Message)));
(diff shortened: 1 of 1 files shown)


ashic:
I'm thinking about this a bit more. It seems that the NLogLogger uses the LogEvent's Message property as the Message passed to NLog. However, LogEvent.ToString() provides a much richer message that includes the logger, correct thread id, etc. This is quite helpful (and in my apps, I'm using a custom logger for this reason specifically), and there are no real alternatives; NLog's layout renderers' ${threadid} won't have the correct thread id, as it'll be the id of the logging actor's thread, and not where the LogEvent originated. 

Would it be better if we logged LogEvent.ToString() as the message in NLogLogger?

Aaronontheweb:
> Would it be better if we logged LogEvent.ToString() as the message in NLogLogger?

@ashic yes,I think so

PR #1046, opened 2015-06-08
Initial Akka IO port with TCP driver

willieferguson:
This is a port of the Akka IO TCP driver. I've initially created PR #1023 on the dev branch. This is the same PR on the akka-io branch.

@Horusiath - Thanks for the feedback in that PR - I will include it in my next contribution to Akka IO


rogeralsing:
Thanks!

Issue #1042, opened 2015-06-04
Add timeout to F# Ask operator


Aaronontheweb:
Feedback from one of our trainees

> You were correct that the <? (Ask) operator does not specify any timeout. That makes it of pretty limited usefulness, since it seems like most times we want to use a timeout. Maybe this operator should have a default timeout? Otherwise, the documentation should warn users about how they could be setting themselves up for a deadlock when using this operator.

Sounds like we need to add the ability to pass a timeout to the F# Ask operator. What's the easiest way to do that?

jtmueller:
The problem is that custom operators can only have two arguments, and the `<?` operator already has two. The only way to add in a timeout argument would be to change one of the arguments to a compound value like a tuple. Something like...

`let! result = actor <? (SomeMessage, TimeSpan.FromSeconds(1.0))`

But that syntax means you're required to specify a timeout. Another option would be to keep the syntax as-is, but have the operator itself specify a default timeout. When you want a non-default timeout, or no timeout, you use the `Ask` method.

Speaking of the `Ask` method, I ended up writing this useful extension method that I expect other users of the F# API could benefit from...

    type ICanTell with
        member x.Ask(msg:obj, ?timeout:TimeSpan) =
            // As of F# 4.0, I'd use Option.toNullable instead of this match statement
            let timeout = match timeout with Some t -> Nullable(t) | None -> Nullable()
            x.Ask(msg, timeout) |> Async.AwaitTask

This extension method lets me write code like this:

    async {
        let! response = selection.Ask(Internal(GetActorRef), TimeSpan.FromSeconds(1.5))
        // do stuff
        return something
    }

Without that extension method, I'm forced to write all this boiler-plate every time I call `Ask`:

    async {
        let! response = selection.Ask(Internal(GetActorRef), Nullable(TimeSpan.FromSeconds(1.5))) |> Async.AwaitTask
        // do stuff
        return something
    }

Horusiath:
We could actually introduce overload over `ICanTell` interface to use F#-style optional timeout parameter. But changing ask operator would be harder - it will break existing API.

Scala has a notion of implicit parameters. Once variable with *implicit* keyword is introduced, it gets applied implicitly to all `?` operators in the same scope. We could mock similar behavior in Akka.NET - however this would be probably require some changes in akka core, rather than on F# API level.

[...]

PR #1053, opened 2015-06-11
F# API - ask wrapper + making messages of wrong types unhandled

Horusiath:
Two little changes:

1. Overload for `ICanTell.Ask` method. As discussed in #1042 up to this moment there was no elegant way to use Ask with timeout specified using F# API. To be compatible, ask had to look like this: `actorRef.Ask(message, Nullable(timeout)) |> Async.AwaitTask`. After this change it's equivalent would be: `actorRef.Ask(message, timeout)`.
2. Until now F# Actor tried direct casting message to input type expected by behavior function. It means, that if message of wrong type was sent, invalid cast exception would be thrown and actor would crash and restart. Now it will check if message is of expected type and if not, it will become unhandled, which is actually an expected and akka idiomatic behavior.

Review first ;)

### DIFF
--- a/src/core/Akka.FSharp/FsApi.fs
+++ b/src/core/Akka.FSharp/FsApi.fs
@@ -39,6 +39,14 @@ module Actors =
 
     /// Pipe operator which sends an output of asynchronous expression directly to the recipients mailbox
     let inline (<!|) (recipient : ICanTell) (computation : Async<'T>) = pipeTo computation recipient ActorRefs.NoSender
+    
+    type ICanTell with
+        
+        /// <summary>
+        /// Sends a message to an actor and asynchronously awaits for a response back until timeout occur.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
cc @jtmueller

jtmueller:
I strongly approve of messages of the wrong type being unhandled instead of throwing invalid cast exceptions.

I also like this version of Ask, but in order to deal with timeouts, the caller has to know that they need to wrap the call to Ask in a try/catch, and specifically handle TaskCanceledException for a timeout. In my own project, I have a version of Ask that returns a Success/Error/Timeout discriminated union so that I cannot accidentally forget to handle the TaskCanceledException on timeout.

Horusiath:
@jtmueller while your proposition is IMO very good approach, I think that for this library, we should keep close to default F# Async behavior. Writing wrapper over it to achieve the same behavior is easy, while it may potentially block some patterns used by other people in their projects.

PR #1064, opened 2015-06-17
RemoteDaemon bug, not removing children

rogeralsing:
Pinpoints the bug found in this conversation:

```
 @AndrewEgorov if the Context.Stop isn't working, you can try sending a PoisonPill.Instance to that remote actor and see if that removes it
but this sounds like a bug
I'll log it and see if we can fix it
Horusiath 21:02
@Aaronontheweb I tried @AndrewEgorov example with PoisonPill - same result. Looks like bug
rogeralsing 21:38
I found the problem
we dont have any code for handling deathwatchnotification in RemoteDaemon..
```

cc @AndrewEgorov

I will port the missing code. opening this PR so everyone knows we are working on it

### DIFF
--- a/src/core/Akka.Remote/RemoteSystemDaemon.cs
+++ b/src/core/Akka.Remote/RemoteSystemDaemon.cs
@@ -286,17 +286,17 @@ namespace Akka.Remote
             }
         }
 
-        private void RemoveChildParentNeedsUnwatch(IActorRef parent, IActorRef child)
+        private bool RemoveChildParentNeedsUnwatch(IActorRef parent, IActorRef child)
         {
             const bool weDontHaveTailRecursion = true;
             while (weDontHaveTailRecursion)
             {
                 IImmutableSet<IActorRef> children;
                 if (!_parent2Children.TryGetValue(parent, out children)) 
-                    return; //parent is missing, so child does not need to be removed
+                    return false; //parent is missing, so child does not need to be removed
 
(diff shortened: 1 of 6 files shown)


src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -183,6 +278,37 @@ public override IActorRef GetChild(IEnumerable<string> name)
             }
             return ActorRefs.Nobody;
         }
+
+        private void AddChildParentNeedsWatch(IActorRef parent, IActorRef child)
```
rogeralsing: We need some eyes on this. I think this part is threadsafe but it should be reviewed.

src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -183,6 +278,37 @@ public override IActorRef GetChild(IEnumerable<string> name)
...
+                IImmutableSet<IActorRef> children;
+                if (_parent2Children.TryGetValue(parent, out children))
+                {
+                    if (_parent2Children.TryUpdate(parent, children.Add(child), children))
+                        return; //child successfully added
+                }
+            }
+        }
+
+        private bool RemoveChildParentNeedsUnwatch(IActorRef parent, IActorRef child)
```
rogeralsing: This too

rogeralsing:
I think this is done.
calling system/context stop on a remote deployed actorref now terminates and removes the remote actor

Issue #1062, opened 2015-06-15
Creating a cluster router before the creating node has joined the cluster results in a broken router


rogeralsing:
If you try to create a cluster aware router before the node that is creating it have managed to join its seed node, the router will be permanently broken, even after the node manages to join the cluster.
This results in racy systems if all your nodes are started about the same time.

I'm marking this as a bug as I believe that the router is intended to adapt even under those conditions.
As it would be very weird if one would need to add some initialization code to wait until the node manages to join and then start creating your actors or routers.

Aaronontheweb:
This is not a bug. It's an explicit design choice.

> I'm marking this as a bug as I believe that the router is intended to adapt even under those conditions.

The router is meant to be a transparent message distribution point for other actors - it's inherently racy as a result of the high-throughput, mailbox-less design of routers themselves. If you create a clustered router and immediately send a message to it before it receives any gossip messages, it won't have any routees.

Clustered routers are fundamentally different from every other type of router because they pick up their routees over time, as a result of changes in the network. Unlike local or Akka.Remote routers, they can't create their routees at start-time because clustered routers have to wait for gossip information before they know who they can route to.

> As it would be very weird if one would need to add some initialization code to wait until the node manages to join and then start creating your actors or routers.

Why is that weird? If your node doing the routing is up before any of the routees are up, what should the router do? Block on startup (bad?) Add a mailbox and queue messages (run out of memory and still not deliver any messages if a routee node never comes online?)

Routers should stay dumb. Shouldn't over-engineer them to avoid some use cases where you have to be aware that you're using clustering, because ultimately the problem you're dealing with is a race condition that is totally in the hands of the end-user to control. 

I posted a solution on how to deal with this issue in a straightforward way in Gitter chat late last week, and we have a sample inside WebCrawler that we recently added.

I agree that this can surprise developers  - it certainly has caught me by surprise before, but this is intentional behavior.

rogeralsing:
> immediately send a message to it before it receives any gossip messages, it won't have any routees.

That is not the behavior Im describing here.
The router is dead completely, even after the node has joined the cluster.

So even if you send it messages after it joins the cluster, nothing goes through

Aaronontheweb:
Ah crap, I misread this

> If you try to create a cluster aware router before the node that is creating it have managed to join its seed node, the router will be permanently broken, even after the node manages to join the cluster.
This results in racy systems if all your nodes are started about the same time.

[...]

PR #1068, opened 2015-06-19
RemoteDaemon bug, not removing children

rogeralsing:
This fixes the issue that remote deployed actors was not removed from the remote daemon when `.Stop(remoteref)`. which resulted in that it was impossible to re-deploy remote actors with the same name,

Renamed RemoteDaemon to RemoteSystemDaemon (was changed in JVM)
Added Hooks 
Added spin-retry updates of parent2children
Added fake remote terminator argument (using deadletters instead of true termination hooks as those are still missing)
Added AddChildParentNeedsWatch


rogeralsing:
I've found a big issue in the last change to this PR.

Apparently, when the RemoteSystemDaemon tries to send messages back to the calling system, there is no `Serialization.CurrentTransport` 
This line `parent.Tell(new Watch(parent, this),this);` tries to communicate back to the creating system, _but_ the `ActorRef` for the remote daemon gets serailized w/o transport information, that is, no address, no protocol.
Which causes the calling side to fail when trying to deserialize that message.

This looks critical to me, if the remoting layer "forgets" to set the active transport before sending messages under some conditions.

cc @Aaronontheweb

rogeralsing:
I Updated the PR, fixed the problem with outbound actorrefs in RemoteSystemDaemon. (which was an error in Serialization.SerializeActorRef, not as I first suspected, an error with active transports)

However, there is another bug that surfaces if the `parent.Tell(new Watch(parent, this),this);` line is uncommented.
The remoting layer will try to re-watch the actorref over and over, so that needs fixing, but I'd say that is a separate issue.
I'm leaving the line commented for now, as this is new functionallity that we have not had before, and which relies on the remote watching re-watch problem to be resolved first.

rogeralsing:
The failing test might be racy or something, it works locally, rerun?

Aaronontheweb:
This is great work @rogeralsing - I'll give it a full review tomorrow.

rogeralsing:
Just to avoid any confusion, this is unrelated to #1062

PR #1073, opened 2015-06-21
Akka IO: Remote Transport, bug fixes and improvements

willieferguson:
This PR address a number of issues to get Akka IO ready to merge to dev, including:

#### Akka IO Transport
I've created a new contrib project with the AkkaIOTransport. Initial testing shows better performance that Helios by 20-40%. This was done using a simple ping-pong test and result might be substantially different for more realistic work loads.

#### Bug fixes
- SelectionHandler now creates and use a dedicated thread for Select. (A thread synchronization issue  caused the intermittent failures of some tests)
- Couple of small bugs

#### Other improvements
- SocketChannel: Use Socket.Poll to test for non blocking instead of using exceptions
- Unit tests seems to consistently pass. Let me know is anybody still get intermittent failures.


rogeralsing:
:+1: looking forward to try this out

PR #1075, opened 2015-06-22
 Akka IO: fixes #1070 - StackOverflowException

### LINKED ISSUE #1070: Akka.IO: StackOverflowException received when ByteString is send back to Sender
Consider that we have a TCP listener actor having following logic (basic ECHO service):

```csharp
Receive<Tcp.Received>(received => Sender.Tell(Tcp.Write.Create(received.Data)));
```

I've found that this would result in error inside [TcpConnection.DoWrite](https://github.com/akkadotnet/akka.net/blob/akka-io/src/core/Akka/IO/TcpConnection.cs#L656) method. Basically recursive call turns into infinite loop, since first condition `_buffer.HasRemaining` (*data has been sent*) is always false, while second `data.NonEmpty` (*ByteIterator.Len is 0 and ByteString.Count != 0 at the same time*) is always true.

This problem can be reproduced by using sending some data with telnet command on example I've attached in #1069. 

/cc @willieferguson

willieferguson:
ByteStrings should create a new ByteIterator every time.

### DIFF
--- a/src/core/Akka/Util/ByteString.cs
+++ b/src/core/Akka/Util/ByteString.cs
@@ -91,12 +91,10 @@ namespace Akka.IO
         internal class ByteString1C : CompactByteString
         {
             private readonly byte[] _bytes;
-            private readonly ByteIterator.ByteArrayIterator _iterator;
 
             public ByteString1C(byte[] bytes)
             {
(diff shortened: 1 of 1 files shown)


Horusiath:
cc #1070

Issue #1077, opened 2015-06-22
Create a remote test base class for two systems


rogeralsing:
There are some tests that do set up a remote system and tests the interaction between the two, but this involves some not completely obvious config string replacements to make system name and port number dynamic.

It would be nice to create a base class for this kind of tests so that we can test remote interaction w/o throwing too much effort on the test setup.
`RemoteRouterSpec` does this and I think this can be pretty much copy pasted into a base type.

I'd like to see better coverage of akka.remote in general, we have had quite a few issues with remote deployment and want to fix this.

Thoughts on this?
Should we abstract the double system setup to a base class in Akka.Remote.Tests?

rogeralsing:
Or is this exactly what the MultiNodeSpec does?

Aaronontheweb:
This is what `MultiNodeSpec` does - but as @Horusiath has ported out, it needs to be documented.

PR #1082, opened 2015-06-22
RemoteDeploySpec

rogeralsing:
Started to clean up the RemoteDeploySpec, which previously contained only two local tests copied from Akka.Tests RoutingSpec for some reason.
Added some basic tests to verify that remote deployment works. 

This also introduces a PoC of  #1077, thoughts on this?


Aaronontheweb:
Good idea, but we should just port the multi-node tests and add a test run for Akka.Remote.MultiNodeTests. I move to close this PR and do that instead. I'll help get that set up - got stuck in training today but I'm able to get back onto it now.

rogeralsing:
There are a few other tests that use the same strategy as this PR, they should be transformed into multinode testkit tests then I guess.
E.g. RemoteRouterSpec and others

Right?

Issue #1065, opened 2015-06-18
NUnit tests fail if the fixture contains more than one test


garethbudden:
I've hit a problem when trying to use the NUnit test suite, given the following example...

```c#
public class NUnitTestFixture
{
    [Test]
    public void SayHello()
    {
        ActorOf<Parrot>("test").Tell(new SayHello());
    
        ExpectMsg<Hello>();
    }

    [Test]
    public void SayGoodbye
    {
        ActorOf<Parrot>("test").Tell(new SayGoodbye());
    
        ExpectMsg<Goodbye>();
    }
}
```
When I run the tests the first succeeds but the second fails with ```Akka.Actor.InvalidActorNameException : Actor name "test" is not unique!```

If I instead run the same tests but with XUnit then they succeed, I believe this could be related to #972, which says that NUnit keeps the ActorSystem around for the full test fixture, but XUnit creates a new ActorSystem for each test.

rogeralsing:
This is definitely a bug in the NUnit TestKit, this package is pretty new and thus not very well tested in the field.
We use XUnit TestKit internally.

garethbudden:
One option would be to wipe the ActorSystem after each test run, as was mentioned in #972 this may not be performant, on the other hand it would be more consistent with how XUnit works.  Or is there a better way to handle this entirely?

rogeralsing:
Yes, imo, we should spin up a  new instance of the system and init all the defaults after each test.
This is after all what happens in the XUnit Testkit, each parallel test has its own system and infrastructure

Aaronontheweb:
Is this really a bug? The current design seems right to me - the actor system should be shared across all of the tests in a fixture. Our multi-node tests would all break if this wasn't the case.

Aaronontheweb:
Ah, I take that back about the multi-node tests - that's not the case there.

rogeralsing:
If the system was shared between tests, we would end up with a gazillion racy tests with completely undeterministic behavior

Aaronontheweb:
Where inside the XUnit testkit does the actor system get shut down after each test currently?

rogeralsing:
Yes, there is one instance per test in XUnit, all isolated

Aaronontheweb:
> Yes, there is one instance per test in XUnit, all isolated

Can you walk me through that inside the code of the XUnit fixture itself? Like what's the call chain that actually makes this happen per-test? Everything looks per-fixture to me.

[...]

PR #1092, opened 2015-06-24
Fix for issue #1065 NUnit to destroy ActorSystem after each test run inline with XUnit

garethbudden:
I've made changes as outlined in #1065 to allow NUnit's TestKit to work in the same way as xunit.

##### Providing an ActorSystem in the constructor will fail
Unlike xunit (and I think mstest), NUnit does not create a new instance of the test class for each test run.  If the user provides an ```ActorSystem``` in the constructor, all tests will attempt to use the same system, however after each test the system is destroyed.  One possible fix would be to take a ```Func<ActorSystem>``` instead, however that would introduce a breaking change to the API for all TestKit implementations and I wasn't sure how you'd feel about that.  Another option is simply to throw a ```NotSupportedException``` in the NUnit TestKit if the user provides an instance of ```ActorSystem```.

##### Redundant methods
The NUnit TestKit has ```AfterAll(), Dispose(), and Dispose(bool disposing)``` methods, however these are redundant as the class now uses a TearDown attribute to clean up.  Removing these methods makes sense but again introduces a breaking change.  Would this be a problem?

### DIFF
--- a/src/contrib/testkits/Akka.TestKit.NUnit.Tests/Akka.TestKit.NUnit.Tests.csproj
+++ b/src/contrib/testkits/Akka.TestKit.NUnit.Tests/Akka.TestKit.NUnit.Tests.csproj
@@ -46,6 +46,7 @@
   <ItemGroup>
     <Compile Include="AssertionsTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="TestKitTestFixtureTest.cs" />
     <Compile Include="TestKitTests.cs" />
   </ItemGroup>
   <ItemGroup>
--- a/src/contrib/testkits/Akka.TestKit.NUnit.Tests/TestKitTestFixtureTest.cs
+++ b/src/contrib/testkits/Akka.TestKit.NUnit.Tests/TestKitTestFixtureTest.cs
@@ -1,0 +1,30 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="TestKitTests.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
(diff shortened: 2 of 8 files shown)


Aaronontheweb:
>  Another option is simply to throw a NotSupportedException in the NUnit TestKit if the user provides an instance of ActorSystem.

Code-bombing people is a much better alternative than introducing breaking changes to the API, so I approve :)

Aaronontheweb:
> The NUnit TestKit has AfterAll(), Dispose(), and Dispose(bool disposing) methods, however these are redundant as the class now uses a TearDown attribute to clean up. Removing these methods makes sense but again introduces a breaking change. Would this be a problem?

The TestKit API is stable as of 1.0. There's no possibility of changing it. NUnit must work with it.

Aaronontheweb:
@garethbudden this looks great - just finished my review.

Could you have `AfterAll` / `Dispose` etc called inside the testfixture's `TearDown`-decorated methods? That should work.

garethbudden:
I've added the code :bomb:, and made the TearDown method call ```AfterAll()```. 

I haven't made the TearDown method call ```Dispose()``` however, because I realised this would mean that the test fixture _instance_ would be Disposed multiple times, one for each test!  Which felt wrong, and may cause adverse affects on people who rely on dispose being called once when the Fixture finishes.

Aaronontheweb:
Ty sir!

mchandschuh:
@Aaronontheweb -- nunit 3.13 adds `[FixtureLifeCycle(LifeCycle.InstancePerTestCase)]` which creates a new test class instance per test, but the not supported exception prevents it's use. What needs to be updated to allow this to work?
> https://docs.nunit.org/articles/nunit/writing-tests/attributes/fixturelifecycle.html

Aaronontheweb:
@mchandschuh we might need to patch Akka.NET first in order to fix this.

PR #1093, opened 2015-06-24
Fixed build scripts to work around known issues with Mono 4.0

Aaronontheweb:
Updated our build scripts to work around https://github.com/fsharp/fsharp/issues/427 and resolve https://github.com/fsharp/FAKE/issues/839


Aaronontheweb:
Build currently fails on Mono as a result of #1094, but this PR has all of the necessary script modifications to make builds on Mono possible to begin with.

PR #1095, opened 2015-06-24
[api-doc] Documentation of namespaces

sean-gilliam:
Documented namespaces.

Note: I put a TBD on Akka.Pattern since it looks like the only class in
that namespace could be put into another.

Note2: As mentioned in gitter chat, it looks like Akka.Actor.Internal
and Akka.Actor.Internals is an oversight and should be merged.


Aaronontheweb:
Yes, we should consider merging those two. I'll have to take a look at `ActorSystemImpl` in particular and see if that gets exposed in any third-party plugins (custom Akka.Persistence impls and Akka.Monitoring spring to mind)

PR #1098, opened 2015-06-25
Implemented BackoffSupervisor

Horusiath:
**Ported from Akka JVM**: This is implementation of the exponential backoff algorithm for actor supervisors. It's useful in cases, when for example persistent actor fails due to underlying persistence mechanism failures or unavailability. In that case it's quite probable that in this case immediate restart of an actor will fail again. Solution is to apply some delay before restart.

This implementation uses exponential back-off i.e. given minimal back-off 3 sec. and maximal back-off 30 sec. the following start attempts will be delayed with 3, 6, 12, 24, 30 seconds. In addition back-off uses random delay multiplier configured by `randomFactor` variable (example 0.2 &rArr; 20%). Reason for that is to avoid hitting the persistence backend by all restarting actors at the same time.

**Don't merge yet!**

### DIFF
--- a/src/core/Akka.Tests/Akka.Tests.csproj
+++ b/src/core/Akka.Tests/Akka.Tests.csproj
@@ -142,6 +142,7 @@
     <Compile Include="MatchHandler\MatchExpressionBuilder_CreateArgumentValuesArray_Tests.cs" />
     <Compile Include="MatchHandler\MatchHandlerBuilderTests.cs" />
     <Compile Include="MatchHandler\PartialActionBuilderTests.cs" />
+    <Compile Include="Pattern\BackoffSupervisorSpec.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Routing\BroadcastSpec.cs" />
     <Compile Include="Routing\ConsistentHashingRouterSpec.cs" />
--- a/src/core/Akka.Tests/Pattern/BackoffSupervisorSpec.cs
+++ b/src/core/Akka.Tests/Pattern/BackoffSupervisorSpec.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using Akka.Actor;
+using Akka.Pattern;
+using Akka.TestKit;
+using Xunit;
+
+namespace Akka.Tests.Pattern
+{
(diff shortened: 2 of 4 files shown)


rogeralsing:
Looks good to me :+1:

Aaronontheweb:
@rogeralsing @Horusiath IMHO this does not belong in Akka.Persistence.

I would create an Akka.Contrib.Pattern project and put this in there so we can start porting over some of the other re-usable patterns.

On the JVM is this something that's specific to Akka.Persistence? I haven't looked. But if it is, then this is fine as-is.

Horusiath:
@Aaronontheweb  It's part of persistence plugin on akka JVM, but for me it should be part of Akka.Patterns namespace. I've asked about this on akka/dev gitter and I'm waiting for the response.

Aaronontheweb:
@Horusiath great minds think alike :+1:

Horusiath:
`BackoffSupervisor` has been moved to *Akka.Pattern* and corresponding proposal has been made for JVM akka [here](https://github.com/akka/akka/issues/17842).

We could document this pattern somewhere in [Fault Tolerance](http://getakka.net/docs/Fault%20tolerance) section.

rogeralsing:
@Aaronontheweb can we do something about the multi node tests? its hard to reason about the status of a PR when the tests are failing

Aaronontheweb:
@rogeralsing opened up #1105 to deal with the readability. At this rate MultiNodeTests are here to stay - open PRs and fix them if they fail.

Aaronontheweb:
Technically, doesn't this need to live in its own project and package, just like it does on the JVM?

https://github.com/akka/akka/tree/master/akka-contrib/src/main/scala/akka/contrib

Horusiath:
`BackoffSupervisor` is used in the latest version of cluster sharding plugin - I'm not so sure if we want to have obligatory dependency on patterns project there. I'd treat it on the same rights as `CircuitBreaker` which is part of core akka package.

rogeralsing:
I agree with @Horusiath, if this is used by other soon to be core modules, then this should go directly into Akka.
All agree ?
@akkadotnet/developers

Issue #1090, opened 2015-06-23
Remote watching is repeated


rogeralsing:
If like 243 is uncommented in the RemoteSystemDaemon: 
```csharp
           //TODO: figure out why current transport is not set when this message is sent
          parent.Tell(new Watch(parent, this),this);
```

The system that trigger the watch will repeatedly send `Re-watching` to the remote system.
![remotewatch](https://cloud.githubusercontent.com/assets/647031/8315934/1b5c5250-19f4-11e5-9213-77a02f3c74d2.png)

I guess this is not expected(?)
What and why is the reason of a re-watch?

cc @Aaronontheweb

Aaronontheweb:
Yep, that's a mystery. Might be a bug in the `ReliableDeliveryActor` responsible for guaranteed delivery of system messages inside Akka.Remote

rogeralsing:
Or is this intended, I found this peice of code:

```csharp
        private void ReceiveHeartbeatRsp(int uid)
        {
            var from = Sender.Path.Address;

            if (_failureDetector.IsMonitoring(from))
            {
                _log.Debug("Received heartbeat rsp from [{0}]", from);
            }
            else
            {
                _log.Debug("Received first heartbeat rsp from [{0}]", from);
            }

            if (_watchingNodes.Contains(from) && !_unreachable.Contains(from))  //<--- checks if we are watching this node
            {
                if (!_addressUids.ContainsKey(from) || _addressUids.ContainsKey(from))
                    ReWatch(from);                                                                                         //<-- sends rewatch
                _addressUids[from] = uid;
                _failureDetector.Heartbeat(from);
            }
        }
```

See inline comments

Aaronontheweb:
Oh yeah, definitely looks intended. Which class is this from? I'd be interested to see if it's still done this way on the JVM.

rogeralsing:
RemoteWatcher.cs

rogeralsing:
In any case, the above code explans why the rewatch appears in sync with the heartbeats.
But it seems expensive to do this for each beat imo.

Aaronontheweb:
@rogeralsing looks like they still do this on the JVM too https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/RemoteWatcher.scala#L150

so is there something we need to uncomment in order to enable this?

[...]

PR #1099, opened 2015-06-25
fixing repeated remote rewatch

rogeralsing:
fixing #1090

there was a condition that always evaluated true that is now fixed to be in line with the scala code


rogeralsing:
cc @Aaronontheweb

Issue #1100, opened 2015-06-26
Update mono build targets to exclude MSTest projects


Aaronontheweb:
MSTest doesn't play nicely on Mono, so we're going to have to create a build configuration for Mono using the `AllTests` FAKE target that doesn't attempt to compile this binary or run any of its associated unit tests.

stefansedich:
This is done as part of the #1103 PR but we are now experiencing further issues, I am closing off this item and creating a new issue realted to our mono issues and builds.

PR #1103, opened 2015-06-26
Fixes for mono FAKE builds

stefansedich:
Fixes for #1100


Aaronontheweb:
@stefansedich could you squash these commits?

Aaronontheweb:
Pull in latest from dev - #1102 might have fixed this.

stefansedich:
Ah awesome! Wil give that a shot today sometime.

Aaronontheweb:
@stefansedich mind including a short update about where you're at with the Mono stuff? All of the fun stuff from our Skype chat ;)

stefansedich:
Sure sorry @Aaronontheweb forgot to write an update :)

So thus far this has been mostly a task of one step forwards 500 back, after various small fixes to actually get the build running on Mono (fix some output paths, ignore MSTest tests under mono) we have now run into an issue with xunit2 on mono: 

https://github.com/xunit/xunit/issues/158

According to this thread the master branch (4.3.x) actually works with xunit2 (this is something I am currently validating as being true for akka.net).

I will update as soon as I get the latest mono compiling and have a chance to run the build.

stefansedich:
Ok running the latest nightly mono build the xunit tests at least start, but at the first cluster test it goes boom and seems to hang.

Will investigate this further!

stefansedich:
@Aaronontheweb should I kill this PR for now and shelve my changes? I am not sure I see a fix in the immediate term, I prefer to not just change the tests to mask the issue we are seeing in Mono :(.

Aaronontheweb:
@stefansedich yes, that's probably for the best - but keep #1110 up to date.

stefansedich:
Ok no worries, closing this off!

PR #1114, opened 2015-07-02
NinjectDependencyResolver should use GetOrAdd not TryAdd

sblackler:
From reading the original code, it seems the intention was to use the cache in the case that `GetTypeValue()` assembly scanned. Unfortunately, this was always being evaluated which made the use of the cache pretty pointless.   This PR changes the implementation to `GetOrAdd`.


rogeralsing:
:+1: Thanks

Issue #1126, opened 2015-07-07
bug in F# discriminated union serialization


arcturius:
Hello, 

I have decided to give akka.net a try and unfortunately I faced a bug in 30 minutes after my start. The following my unit test crashes on my PC with the latest source code: 

``` F#
type testType1 = 
    string * string

type testType2 = 
    | V2 of testType1

[<Fact>]
let MyTest () =
    let x = V2("hello!","world")
    use sys = System.create "system" (Configuration.defaultConfig())
    let serializer = sys.Serialization.FindSerializerFor x
    let bytes = serializer.ToBinary x
    let des = serializer.FromBinary (bytes, typeof<testType2>) :?> testType2
    des
    |> equals x

```
Stack:

Test Outcome:	Failed
Test Duration:	0:00:00.471

Result Message:	System.NullReferenceException : Object reference not set to an instance of an object.

Trace:	
at Akka.Serialization.NewtonSoftJsonSerializer.TranslateSurrogate(Object deserializedValue, ActorSystem system, Type type) in Akka\Serialization\NewtonSoftJsonSerializer.cs:line 132
   at Akka.Serialization.NewtonSoftJsonSerializer.FromBinary(Byte[] bytes, Type type) in Akka\Serialization\NewtonSoftJsonSerializer.cs:line 114
   at Akka.FSharp.Tests.ApiTests.MyTest() in Akka\src\core\Akka.FSharp.Tests\ApiTests.fs:line 76

I see other unit tests implemented to test this feature. they work ok. But they obviously dont test it thoroughly. Not sure how to fix the bug myself. I really appreciate your quick help. 

Best regards,
Pavel.

rogeralsing:
Yes, this is a known issue, Json.NET does not play well with DU's in many cases.
We still have no good workaround for it as Json.NET does not emit type information for DU's.

Related to:
#999 #737

rogeralsing:
So, what does this actually compile into?
cc @Horusiath 

```fsharp
type testType1 = 
    string * string

type testType2 = 
    | V2 of testType1
```

Other DU's contains nested types for each entry in the DU, which allows us to do duType.GetNestedTypes() to get the possible types for the DU.
But in this case, there are no nested types, and the DU type itself does not have any ctor, neither public or private.

So what is `V2` behind the scenes?

rogeralsing:
NVM, I figured it out, I submitted a PR for this here #1127

arcturius:
[...]

Issue #999, opened 2015-05-21
F# API - problem with discriminated union serialization


Horusiath:
@neoeinstein found a problem with DU serialization/deserialization:

> I tried passing a TraceableMessage<ItemAvailabilityMessage>, which failed complaining about receiving the wrong token in the JSON. Then I tried unwrapping things and just sending over the ItemAvailabilityMessage, but that gave be the NullReferenceException on remote deserialize.

Gist with message types: https://gist.github.com/neoeinstein/30a38229c368313b721d

erdeszt:
Having the same issue. Sending discriminated unions to remotes fail with ```NullReferenceException``` (even with a simple Option type). Works fine with local actors.
I'm running mono 4.1 and f# 3.0 on linux mint 17(x64 the build target is x86). 
Example code: https://gist.github.com/erdeszt/aece58b48bc4ebfe7090

rogeralsing:
I suspect that the actual serialization works, but that when we put that into the remoting envelopes we might lack the type information to actually deserialize it on the other end.

I will look into it

Horusiath:
@erdeszt This is JSON.NET issue (to be exact it's < 7.0 version issue) - your example doesn't apply to Akka.NET, since it uses default settings (akka serializer uses custom configuration, which provokes the problem). To fix this you may upgrade Newtonsoft.Json to version >= 7 . Also I've created a custom fork of Akka.FSharp called [Akkling](https://www.nuget.org/packages/Akkling/0.1.0-beta), which solves that issue by upgrading to the newest JSON.NET by default.

erdeszt:
@Horusiath thanks for the info and sorry for my confusion, I didn't know about the custom configuration of Json.Net.
Is there a chance for a patch release with the upgraded Json.Net lib? I'm not that keen on using my own fork or experimental libs.
Not that I don't like your approach I was also playing around with the F# api to make it more typesafe, so I'd be happy to see your changes in the official Akka.FSharp repo. Any plans on merging it when you have a stable version?

Horusiath:
I don't suspect to merge it into Akka.FSharp any time soon, since it contains some breaking changes to the API, and we can't afford to make them since library is already officially 1.0. I've decided to make independent lib, which is not bound by core Akka tooling (we can use F# specific tools like i.e. Paket or Project Scaffold this way) and release cycle. It will be probably merged back at some point in time, but not anytime soon.

Aaronontheweb:
[...]

PR #1128, opened 2015-07-07
Json net 7

rogeralsing:
This PR introduces Json.NET and removes the old DU hacks we had in the serializer.

This closes: #1127 #1126 #999 #827

### DIFF
--- a/src/core/Akka/Serialization/NewtonSoftJsonSerializer.cs
+++ b/src/core/Akka/Serialization/NewtonSoftJsonSerializer.cs
@@ -117,16 +117,18 @@ namespace Akka.Serialization
             var j = deserializedValue as JObject;
             if (j != null)
             {
+                //The JObject represents a special akka.net wrapper for primitives (int,float,decimal) to preserve correct type when deserializing
                 if (j["$"] != null)
                 {
                     var value = j["$"].Value<string>();
                     return GetValue(value);
                 }
 
+                //The JObject represents a discriminated union
                 if (j["Case"] != null)
                 {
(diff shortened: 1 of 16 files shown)


src/core/Akka/Serialization/NewtonSoftJsonSerializer.cs
```diff
@@ -107,40 +110,32 @@ public override byte[] ToBinary(object obj)
...
-                    var paramTypes = ctor.GetParameters().Select(p => p.ParameterType).ToArray();
-                    var values =
-                        j["Fields"].ToObject<object[]>()
-                            .Select((o, i) => TranslateSurrogate(o, system, paramTypes[i]))
-                            .ToArray();
-                    var res = ctor.Invoke(values.ToArray());
-                    return res;
-                }
+                //The JObject is not of our concern, let Json.NET deserialize it.
+                return j.ToObject(type, parent._serializer);
```
rogeralsing: This does the magic, if we get a JObject, we can pass in the same settings as we use for other serialzation, and things turns out correct :)

Aaronontheweb:
I'm kind of the opinion that we should just pull this in anyway buuuuuuuuuut 

image

What sort of risk is there for this change to cause Akka.NET to be incompatible with other stuff that depends on JSON.NET, such as ASP.NET MVC?

rogeralsing:
@Aaronontheweb I have absolutely no idea :) maybe a simple assembly version redirect would work in those cases. ¯\_(ツ)_/¯

Aaronontheweb:
A BRAVE NEW WORLD

PR #1129, opened 2015-07-07
Cluster sharding spec (complete)

Horusiath:
This is a completed (but not yet passing) spec suite for basic akka cluster sharding behaviors.


rogeralsing:
Yay!

PR #1132, opened 2015-07-08
Either usability improvements

kekekeks:
**Construction**
Before:
```csharp
static Either<int, string> Foo(bool left)
{
    if (left)
        return new Left<int, string>(123);
    else
        return new Right<int, string>("321");
}
```
After (first variant is still supported):
```csharp
static Either<int, string> Bar(bool left)
{
    if (left)
        return Either.Left(123);
    else
        return Either.Right("123");
}
```

**Transformations**
```csharp
var either = new Left<int, string>(123);
return either.
    .MapLeft(x => x + 1)
    .MapRight(s => s.Substring(1))
    .Map(x => x + 2, s => s + "123")
    .Fold(x => x/2, s => s.Length);
```

### DIFF
--- a/src/core/Akka/Util/Either.cs
+++ b/src/core/Akka/Util/Either.cs
@@ -5,6 +5,8 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System;
+
 namespace Akka.Util
 {
     public abstract class Either<TA,TB>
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
@kekekeks what's the output of the transformation at the end here? :p

kekekeks:
63
 `MapRight` and the right part of `Map` do nothing, since its `Left`, so its `(123+1+2)/2`

Aaronontheweb:
Got it! awesome. Makes sense.

PR #1137, opened 2015-07-09
A virtual time TestScheduler

sadprofessor:
Needed for testing actors that use the scheduler. Included unit tests demonstrate usage.


Aaronontheweb:
Akka.NET TestKit may now go back in time. Safety Not Guaranteed. Must bring your own weapons.

In all seriousness, is this setup as the default scheduler in the TestKit going forward?

sadprofessor:
I didn't feel 100% comfortable making it the default scheduler, so there's a new TestScheduler.conf you can use for tests that require deterministic scheduling. Of course it could be made the default if you think it should. I tried it briefly but it broke a bunch of remoting tests for non-obvious reasons.

Aaronontheweb:
Good to know. Would you mind sending a page to https://github.com/akkadotnet/getakka.net explaining how to enable it and how it works (i.e. how does time advance?)

sadprofessor:
Yeah np

PR #1140, opened 2015-07-09
GracefulStop bug fix - didn't properly handle timeouts

Aaronontheweb:
GracefulStop wouldn't correctly unwatch an actor if the actor failed to respond to the shutdown request in time and the task was cancelled.

### DIFF
--- a/src/core/Akka/Actor/GracefulStopSupport.cs
+++ b/src/core/Akka/Actor/GracefulStopSupport.cs
@@ -63,19 +63,26 @@ namespace Akka.Actor
             return promise.Task.ContinueWith(t =>
             {
                 var returnResult = false;
-                PatternMatch.Match(t.Result)
-                    .With<Terminated>(terminated =>
-                    {
-                        returnResult = (terminated.ActorRef.Path.Equals(target.Path));
-                    })
-                    .Default(m =>
-                    {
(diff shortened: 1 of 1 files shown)


src/core/Akka/Actor/GracefulStopSupport.cs
```diff
@@ -63,19 +63,25 @@ public static Task<bool> GracefulStop(this IActorRef target, TimeSpan timeout, o
...
+                            returnResult = (terminated.ActorRef.Path.Equals(target.Path));
+                        })
+                        .Default(m =>
+                        {
+                            internalTarget.Tell(new Unwatch(target, fref));
+                            returnResult = false;
+                        });
+                }
+                else //need to Unwatch in the event of a timeout or failure
+                {
```
rogeralsing: Do we need to care about race conditions here?
what if `t` is not completed at line 66, but now that this line executes, it might be.

Aaronontheweb: No race conditions (because this occurs inside a continuation) but I just realized that `Task.IsCompleted` is the wrong flag to check.

Aaronontheweb:
So, it looks like the original `GracefulStop` implementation actually works per the specification. No need for this PR.

PR #1145, opened 2015-07-13
Completed Akka.Cluster.Tools.Tests

Horusiath:
This is completed ported code for specs concerning `CluterClient` and `DistributedPubSubMediator`. Now we have to adapt these specs to work with .NET version of Multi Node Test Runner and patch missing/broken parts.


Aaronontheweb:
I think there's something messed up with your MultiNodeSpec impls. Please double check.

src/contrib/cluster/Akka.Cluster.Tools.Tests/PubSub/DistributedPubSubMediatorSpec.cs
```diff
@@ -0,0 +1,738 @@
...
+            public readonly object Message;
+
+            public Talk(string path, object message)
+            {
+                Path = path;
+                Message = message;
+            }
+        }
+
+        [Serializable]
```
Aaronontheweb: Why are the `[Serializable]` attributes necessary?

src/contrib/cluster/Akka.Cluster.Tools.Tests/PubSub/DistributedPubSubMediatorSpec.cs
```diff
@@ -0,0 +1,738 @@
...
+            AwaitAssert(() =>
+            {
+                Mediator.Tell(Count.Instance);
+                Assert.Equal(expected, ExpectMsg<int>());
+            });
+        }
+
+        #endregion
+
+        [MultiNodeFact(Skip = "TODO")]
```
Aaronontheweb: From http://getakka.net/docs/testing/multinode-testkit:

If you have multiple phases that need to be executed as part of a test, you can write them like this:

```csharp
[MultiNodeFact]
public void ConvergenceSpecTests()
{
    AClusterOf3MembersMustReachInitialConvergence();
    AClusterOf3MembersMustNotReachConvergenceWhileAnyNodesAreUnreachable();
    AClusterOf3MembersMustNotMoveANewJoiningNodeToUpWhileThereIsNoConvergence();
}
```

This unfortunate design is a byproduct of Xunit and how it recreates the entire test class on each method.

Horusiath: Got it ;)

src/contrib/cluster/Akka.Cluster.Tools.Tests/PubSub/DistributedPubSubMediatorSpec.cs
Horusiath: I use it for convenience to find message types faster.

Aaronontheweb:
@Horusiath looks like this might be your problem

[15:59:02][Step 1/1] [RUNNER][3:59 PM]: <----------- BEGIN NODE 5 ----------->
[15:59:02][Step 1/1] [RUNNER][3:59 PM]:  --> Akka.MultiNodeTests.ClusterDeathWatchMultiNode5.ClusterDeathWatchSpecTests FAIL
[15:59:02][Step 1/1] [RUNNER][3:59 PM]:  --> Message: Could not load file or assembly 'System.Collections.Immutable, Version=1.0.34.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
[15:59:02][Step 1/1] [RUNNER][3:59 PM]:  --> StackTrace:    at Akka.Remote.TestKit.MultiNodeConfig..ctor()
[15:59:02][Step 1/1] [RUNNER][3:59 PM]:  --> Type: System.IO.FileLoadException
[15:59:02][Step 1/1] [RUNNER][3:59 PM]: <----------- END NODE 5 ----------->

Issue #1107, opened 2015-06-29
Hocon Persistence Configuration : connection-string


mmisztal1980:
Please consider replacing this key with connection-string-name or something similar. so that We can keep connection strings where ther're supposed to be kept (between 'connectionStrings' tags)

Horusiath:
I think we could give people choice, either they want to insert connection string directly in HOCON using `connection-string` or they want to specify name in the connection string section using `connection-string-name` key.

mmisztal1980:
Yeah, I agree - mutually exclusive choice is good.

Horusiath:
Closed by #1151 and #1306

PR #1151, opened 2015-07-14
New config setting akka.persistence.xyz.connection-string-name

Horusiath:
This is a new key for `JournalSettings` and `SnapshotStoreSettings` for persistence plugin based on SQL-backends to give users ability to specify their connection strings not only inside HOCON configs, but as references to <connectionStrings> sections of Web.config/App.config files as well. 

This will also need a corresponding update in Akka.Persistence.SqlServer and Akka.Persistence.PostgreSql in order to work - this PR only adds new common key for settings classes.

/cc #1107


Aaronontheweb:
cc @davidflerlage @mmisztal1980

davidflerlage:
Hi Aaron and Bartosz, I'm on vacation this week, but I am happy to take a look when I return. Thanks!

Issue #774, opened 2015-03-30
Implement RemotingTerminator


Aaronontheweb:
Special termination hook for cleanly shutting down transports within Akka.Remote. Depends on #327

smalldave:
@Aaronontheweb check out RemotingTerminator branch on my fork.
Also RestartNodeSpec branch has cluster test for #918

Horusiath:
Could this one be closed?

Aaronontheweb:
Hasn't been implemented yet. Caused a bunch of side-effects. I'll have to take a closer look at @smalldave's code and give it another try.

PR #1154, opened 2015-07-15
Implemented RemotingTerminator

Aaronontheweb:
Third try's a charm. See #774


src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -211,9 +223,16 @@ private void TerminationHookDoneWhenNoChildren()
...
         {
             OnReceive(message);
+            Sender = sender ?? ActorRefs.Nobody;
         }
 
         /// <summary>
+        /// For system messages that require a reply back to the <see cref="IActorRef"/> responsible
+        /// for telling a message to the <see cref="RemoteSystemDaemon"/>.
+        /// </summary>
+        protected IActorRef Sender { get; private set; }
```
rogeralsing: This is racy.
The `RemoteSystemDaemon` is not an actor, just an impl of `IActorRef` (VirtualPathContainer).
So there is no concurrency constraints applied here.

(I could be misreading something ofc)

rogeralsing: I checked the scala version and they do not do this, they handle all sender related stuff inside their huge `!` tell handler.

Aaronontheweb: I'm fine with option 2 - I need this data either way.

Aaronontheweb:
Working on this alongside some other internals fixes for Akka.Remote.

Aaronontheweb:
Made some more changes to the `RemotingTerminator` and also fixed a bug inside Helios (threw an exception upon shutdown.)

Looks like I'm locking up one of the Akka.Remote.TestKit.Tests with that Helios change now, so I'll start digging some more into that.

Aaronontheweb:
Waiting on the latest build to finish, but otherwise this PR is ready to review.

src/core/Akka.Remote/EndpointManager.cs
```diff
@@ -562,31 +581,27 @@ protected void Accepting(object message)
...
+                    shutdownStatus.ContinueWith(tr => Task.WhenAll(_transportMapping.Values.Select(x => x.Shutdown())).ContinueWith(
+                              result =>
+                              {
+                                  if (result.IsFaulted || result.IsCanceled)
+                                  {
+                                      if (result.Exception != null)
+                                          result.Exception.Handle(e => true);
+                                      return false;
+                                  }
+                                  return result.Result.All(x => x) && tr.Result;
```
rogeralsing: what is the `x => x` thing doing?
Seems odd too compare each element to itself (?)

src/core/Akka.Remote/Transport/Helios/HeliosHelpers.cs
```diff
@@ -247,7 +247,7 @@ public override Task<bool> OpenAsync()
                 {
                     var listener = s.Result;
                     RegisterListener(channel, listener, msg, remoteSocketAddress);
-                }, TaskContinuationOptions.AttachedToParent & TaskContinuationOptions.ExecuteSynchronously & TaskContinuationOptions.NotOnCanceled & TaskContinuationOptions.NotOnFaulted);
+                }, TaskContinuationOptions.ExecuteSynchronously & TaskContinuationOptions.NotOnCanceled & TaskContinuationOptions.NotOnFaulted);
```
rogeralsing: You are mixing up `&` and `|` here. that expression will eval to 0

src/core/Akka.Remote/Transport/Helios/HeliosTcpTransport.cs
```diff
@@ -110,7 +110,9 @@ protected override void OnMessage(NetworkData data, IConnection responseChannel)
         /// <param name="erroredChannel">The handle to the Helios channel that errored.</param>
         protected override void OnException(Exception ex, IConnection erroredChannel)
         {
-            ChannelLocalActor.Notify(erroredChannel, new UnderlyingTransportError(ex, "Non-fatal network error occurred inside underlying transport"));
+            // Want to notify only for encoding exceptions
+            if(!(ex is HeliosConnectionException))
+                ChannelLocalActor.Notify(erroredChannel, new UnderlyingTransportError(ex, "Non-fatal network error occurred inside underlying transport"));
```
rogeralsing: Just taking a stab in the dark here, but aren't we interested in any "fatal" error too? 
Are we getting those exceptions in some other observable way?

rogeralsing:
Questiong
The code base (not just here in the PR) is sprinkled with `ExecuteSynchronously` in a lot of our `PipeTo` scenarios, and _some_ of those also have `AttachedToParent`.
Are we doing this consistently? are there good reasons for when we use and not use `AttachedToParent` 
And are they really proven to have an effect together? it seems strange if `ExecuteSynchronously` wouldn't do this by itself as it must continue directly on the same thread.

src/core/Akka.Remote/EndpointManager.cs
Aaronontheweb: Yeah, basically I just need to make sure that N endpoints and N transports all shut down successfully and the aggregate `bool` value should be either "all succeeded" or "not all succeeded"

src/core/Akka.Remote/Transport/Helios/HeliosTcpTransport.cs
Aaronontheweb: Those exceptions get surfaced as disassociations, and the only other possible type of exception coming from Helios are the recoverable non-fatal ones, such as `CorruptedFrameException`.

Aaronontheweb:
https://msdn.microsoft.com/en-us/library/system.threading.tasks.taskcontinuationoptions.aspx

Using `ExecuteSynchronously` supersedes `AttachedToParent` as far as I can tell, so I've been removing `AttachedToParent` in a bunch of areas. `ExecuteSynchronously` is an optimization that's appropriate for most of the areas where we use it, where the continuation does something simple like formats a message or handles an exception. It's not appropriate for anything that has I/O or is otherwise CPU-intensive. It basically inlines the execution of the next `Task`.

Part of the reason why I haven't removed `AttachedToParent` EVERYWHERE is that despite what the documentation says, it's a flag that is applied somewhat non-deterministically it seems, so I'm not sure what the impact of removing it everywhere will be. Call me "paranoid" :p

src/core/Akka.Remote/Transport/Helios/HeliosHelpers.cs
Aaronontheweb: Argh! You're right! I've been doing that wrong.

Aaronontheweb:
Fixed up the bitmasking on the `TaskContinuationOptions`

PR #1166, opened 2015-07-18
Adds support for a default Ask timeout in HOCON conf.

### LINKED ISSUE #1163: Add default Ask timeout to HOCON configuration
In Scala, ask timeout can be set implicitly, while in C#/F# this is not possible, and therefore it may feel repetitive and not safe (when user decided not to set it at all, there is possibility to hang forever). The idea is to add HOCON config key for setting the default timeout on `IActorRef.Ask` method in case, when user didn't specified it explicitly.

Proposal is to create `akka.actor.ask-timeout` key with TimeSpan value (set to **null** by default to stay backward compatible). This setting should be applied to all [IActorRef.Ask](https://github.com/akkadotnet/akka.net/blob/928ff2b3b23457ea1e4e1d9f0819fca945506a1b/src/core/Akka/Actor/Futures.cs) invocations without explicit timeouts. This should also be reflected in updated documentation.

jordansjones:
This is an attempt to resolve #1163. Instead of defaulting to `null`, I wasn't sure on how that should be represented, it defaults to `infinite`. This should provided the same effective result.

### DIFF
--- a/src/core/Akka.Tests/Actor/AskTimeoutSpec.cs
+++ b/src/core/Akka.Tests/Actor/AskTimeoutSpec.cs
@@ -1,0 +1,54 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="AskSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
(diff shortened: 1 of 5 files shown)


rogeralsing:
Looks good to me :+1:

PR #1168, opened 2015-07-18
Removed debug code from earlier F# DU serialization fix

rogeralsing:
This removes a line of code used for debugging when we fixed the F# DU serialization problems earlier.
It sets the Json default settings, thus mutating shared state, resulting in racy code if anything else uses JsonConvert in our code, in our tests or **in external code**.


Aaronontheweb:
Nice catch!

Issue #1164, opened 2015-07-17
How much of Hocon should we support


rogeralsing:
I've been looking into the Hocon spec a bit more and some of the stuff in there are just crazy.

We do support substitutions already, but according to the Hocon spec, substitutions should be able to self reference

```
a : some value
a : {$a} and some more
a : {$a} and even more
```

In the above case the result should be "some value  and some more  and even more"
There is some conflicting info in the spec saying that substitutions should be done at the end, after all files have been included and parsed.
Which means that somehow, the parser result should remember all three of those lines.

The next part is includes.

```
a : { x : 123 }
a : { include 'somefile.bar' }
```

In the above case, the contents of somefile.bar should be merged with the existing value of `a`, which means that we probably have to load the other files during parsing, and prepend the result onto the string that is being parsed.

And to combine the two into a joy of brain pain.
somefile.bar can contain substitutions that ar relative to the document itself when loaded, but once inserted into the including config, the paths should be rewritten relative to the place where they are inserted into the inserting config.

I really don't want to change the parser to some multi pass parser that does the same'ish thing over and over untill the final config is produced.

So which of these features do we need? should we just let this be for the moment?

rogeralsing:
I've actually got some progress on this, we can resolve includes, and fixup substitutions in includes.
Not sure what to do with the self ref substitutions though.

The include resolver is a `Func<string, HoconRoot>` which we can optionally pass to the `ConfiguratonFactory.ParseString`, we can use this to resolve external configs as long as we return a hocon object back.

I also realized that since Hocon is a superset of json, we can serialize .NET objects to json, and then parse the json as hocon.
Which allows us to create typed objects that can be turned into config objects with a single method call.

So this will probably open up for some nice use cases

rogeralsing:
This has been merged.
Will open up a separate issue if we ever think we need self referencing substitutions.

PR #1169, opened 2015-07-18
Hocon include support

rogeralsing:
Related to #1164 

* Added Hocon Include support
* Added Hocon include resolver support
* Added Hocon Include substitution support
* Added more Hocon tests
* Added HoconObject.Merge support
* Added Config from Object support (.NET object to Hocon Config)

This allows us to use Hocon includes:

```javascript
a {
   include "some resource"
}
```

The 'some resource' can then be resolved by an optional resolver callback. so the resource could be a file on disk, an url, or whatever you want that could result in a new hocon sub config.

**see inline comments for explanations of the code to help the review**

### DIFF
--- a/src/core/Akka/Configuration/ConfigurationFactory.cs
+++ b/src/core/Akka/Configuration/ConfigurationFactory.cs
@@ -11,6 +11,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Reflection;
 using Akka.Configuration.Hocon;
+using Newtonsoft.Json;
 
 namespace Akka.Configuration
 {
@@ -34,11 +35,24 @@ namespace Akka.Configuration
         /// HOCON (Human-Optimized Config Object Notation) string.
         /// </summary>
         /// <param name="hocon">A string that contains configuration options to use.</param>
+        /// <param name="includeCallback">callback used to resolve includes</param>
+        /// <returns>The configuration defined in the supplied HOCON string.</returns>
+        public static Config ParseString(string hocon, Func<string,HoconRoot> includeCallback)
+        {
+            HoconRoot res = Parser.Parse(hocon, includeCallback);
+            return new Config(res);
+        }
+
+        /// <summary>
+        /// Generates a configuration defined in the supplied
+        /// HOCON (Human-Optimized Config Object Notation) string.
+        /// </summary>
+        /// <param name="hocon">A string that contains configuration options to use.</param>
         /// <returns>The configuration defined in the supplied HOCON string.</returns>
         public static Config ParseString(string hocon)
         {
-            HoconRoot res = Parser.Parse(hocon);
-            return new Config(res);
+            //TODO: add default include resolver
+            return ParseString(hocon, null);
         }
 
         /// <summary>
@@ -125,6 +139,18 @@ namespace Akka.Configuration
                     return ParseString(result);
                 }
             }
+        }
+
+        /// <summary>
+        /// Creates a configuration based on the supplied source object
+        /// </summary>
+        /// <param name="source">The source object</param>
+        /// <returns>The configuration created from the source object</returns>
+        public static Config FromObject(object source)
+        {
+            var json = JsonConvert.SerializeObject(source);
+            var config = ParseString(json);
+            return config;
         }
     }
 }
--- a/src/core/Akka/Configuration/Hocon/HoconObject.cs
+++ b/src/core/Akka/Configuration/Hocon/HoconObject.cs
@@ -177,6 +177,33 @@ namespace Akka.Configuration.Hocon
             }
             return text;
         }
+
+        public void Merge(HoconObject other)
+        {
+            var thisItems = Items;
+            var otherItems = other.Items;
+
+            foreach (var otherItem in otherItems)
+            {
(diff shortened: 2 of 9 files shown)


rogeralsing:
ohh, shiny "needs review" auto label :)

src/core/Akka/Configuration/ConfigurationFactory.cs
```diff
@@ -126,6 +140,18 @@ public static Config FromResource(string resourceName, Assembly assembly)
...
+        /// <summary>
+        /// Creates a configuration based on the supplied source object
+        /// </summary>
+        /// <param name="source">The source object</param>
+        /// <returns>The configuration created from the source object</returns>
+        public static Config FromObject(object source)
+        {
+            var json = JsonConvert.SerializeObject(source);
+            var config = ParseString(json);
+            return config;
```
rogeralsing: This lets us create Hocon Configs from typed objects.

a convenience method only, but opens up for using objects as includes, or creating typed configuration support.

src/core/Akka/Configuration/Hocon/HoconObject.cs
```diff
@@ -177,6 +177,33 @@ private string QuoteIfNeeded(string text)
...
+                        thisItem.GetObject().Merge(otherItem.Value.GetObject());
+                    }
+                }
+                else
+                {
+                    //other key was not present in this object, just copy it over
+                    Items.Add(otherItem.Key,otherItem.Value);
+                }
+            }            
+        }
```
rogeralsing: This makes it possible to merge Hocon Objects. which could eventually be used to rewrite the `WithFallback` support.
Right now it is used for the Hocon Include support, merging the sub config with the object that does the include.

src/core/Akka/Configuration/Hocon/HoconParser.cs
```diff
@@ -73,11 +76,25 @@ private void ParseObject(HoconValue owner, bool root)
...
+                        var substitutions = included.Substitutions;
+                        foreach (var substitution in substitutions)
+                        {
+                            //fixup the substitution, add the current path as a prefix to the substitution path
+                            substitution.Path = currentPath + "." + substitution.Path;
+                        }
+                        _substitutions.AddRange(substitutions);
+                        var otherObj = included.Value.GetObject();
+                        owner.GetObject().Merge(otherObj);
+
```
rogeralsing: if the token is an Include, then we call the `_includeCallback` to get the requested sub object.
the sub object is then merged with the currently parsed hocon object.
any substitutions in the sub object is copied into the current parsing pass, and each of those substitutions are rewritten to point to the sub objects new location in the object hierarchy.

src/core/Akka/Configuration/Hocon/HoconSubstitution.cs
```diff
@@ -104,20 +104,6 @@ public HoconObject GetObject()
...
-        /// Performs an implicit conversion from <see cref="HoconSubstitution" /> to <see cref="HoconObject" />.
-        /// </summary>
-        /// <param name="substitution">The HOCON object that contains the substitution.</param>
-        /// <returns>The HOCON object contained in the substitution.</returns>
-        public static implicit operator HoconObject(HoconSubstitution substitution)
-        {
-            return substitution.GetObject();
-        }
-
-        #endregion
```
rogeralsing: I have no idea why we had this strange implicit operator support.  casting from a hocon substitution to a hocon object implicitly.
It was not used, and it was harmful as the cast could fail as not all substitutions are objects.

src/core/Akka/Configuration/Hocon/HoconTokenizer.cs
```diff
@@ -19,6 +20,17 @@ public class Tokenizer
...
+
+        public void Push()
+        {
+            _indexStack.Push(_index);
+        }
+
+        public void Pop()
+        {
+            _index = _indexStack.Pop();
+        }
```
rogeralsing: Improved the Hocon parser to let the tokenizer to use transactions.
we can Push the current parser position and then consume some tokens, and then rollback if the consumed text didnt match what we were looking for.

( The Include support is ambigious with normal unquoted hocon keys. and thus we need to "look ahead" to verify if this is actually an include or not)

src/core/Akka/Configuration/Hocon/HoconTokenizer.cs
```diff
@@ -602,6 +636,34 @@ public bool IsSubstitutionStart()
...
+                        {
+                            return true;
+                        }
+                    }
+                }
+                return false;
+            }
+            finally
+            {
+                Pop();
```
rogeralsing: This is only looking/peeking at the current pos, not consuming, thus we _always_ rollback the parser position.

rogeralsing:
the failing test "ConfigurationSpec.CanCreateConfigFromSourceObject" is fixed by my other PR.

rogeralsing:
If the tests are re-run, does it run on top of newly merged PR's ?

Aaronontheweb:
@rogeralsing it runs exactly the code that you have in your PR - so anything that's in `dev` but not in your PR won't be included in the test run.

rogeralsing:
I updated my branch to include the racy Json config fix merged yesterday.  everything pass now.

PR #1173, opened 2015-07-21
Issue badges & more contribution guidelines

skotzko:
Added Waffle.IO badges & more info/guidelines about contributing. Also added Readme logo for TeamCity build support / OSS license.


skotzko:
Looks like multinode tests crapped out. Fortunately, this is a README only change.

Aaronontheweb:
LOL, man the build server had some issues tonight. Ran into trouble with `StructureMap` tests, Akka.IO, multi-node, you name it... I'll leave this here for a bit in case anyone else objects, but I think the changes look great. Appreciate you adding TeamCity as a sponsor since they gave us that generous OSS license for the build system.

Issue #1148, opened 2015-07-14
Replace SystemNanoTime calls with the MonotonicClock


Aaronontheweb:
Replace all instances of using https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote/SystemNanoTime.cs

With https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Util/MonotonicClock.cs

ProTip: expose a method on the `MonotonicClock` class that implements the behavior of `SystemNanoTime.GetNanos()` - don't convert all of the nano calls into using `TimeSpan` objects.

PR #1174, opened 2015-07-21
Replace SystemNanoTime with MonotonicClock. #1148

jordansjones:


Aaronontheweb:
Looks great, thanks @jordansjones!

PR #1176, opened 2015-07-22
TaskContinuationOptions bitwise fix

rogeralsing:
This PR cleans up some of the broken bitwise operations used in the `ContinueWith`.
The cases that have been fixed is ones that use  `&` instead of `|` to combine `TaskContinuationOptions`, which always result in `TaskContinuationOptions.None`


rogeralsing:
The TCP tests that fails does not fail locally, and I'd be surprised if any of the changes here (remoting/multinode testkit) would have anything to to with akka io

PR #1178, opened 2015-07-23
First autodetected bugs

rogeralsing:
# Do not merge

This PR demonstrates the initial work on the code analyzer.
It correctly detected the use of Self/Sender/Context inside `ContinueWith`

Currently, the analyzer only comments on these issues, I think that is probably safest so that we can reason about the change and fix it as we like.

### DIFF
--- a/src/analyzers/AkkaCodeAnalyzer/AkkaCodeAnalyzer/ClassReferenceCounter.cs
+++ b/src/analyzers/AkkaCodeAnalyzer/AkkaCodeAnalyzer/ClassReferenceCounter.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Editing;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.FindSymbols;
+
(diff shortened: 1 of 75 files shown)


Aaronontheweb:
This is super cool @rogeralsing! Should we be recommending that developers close over `Self` and `Parent`? I thought that since those properties were immutable it didn't matter - is there something I'm missing?

rogeralsing:
`Parent` is only available on `Context`, so to access it you have to close over `Context`

`Self` has this impl;

```csharp
protected IActorRef Self { get { return HasBeenCleared ? _clearedSelf : Context.Self; } }
```

Which means that as long as the actor is a normal actor, it goes through the thread static `Context`, so even if Self never changes, Context does (because of the thread static magic we have for scala implicit values)

So yes, both needs to be closed over.

rogeralsing:
I created another analyzer that checks the number of references to each class.
There are like 70+ classes that has 0 references via code atm.
that does not however mean that they are not used, they could be used via reflection, but we should at least verify their usage.
either remove them, or add tests that show how they should be used..

rogeralsing:
(Some might be tests that the analyzer doesnt yet see as tests.)

rogeralsing:
cc @akkadotnet/developers please take a peek at all the "not used" classes. we need better coverage than this

Danthar:
This is pretty damn cool stuff. Ill see if I can spend some time this weekend to verify some of the unused code.

rogeralsing:
I've updated the analyzer to ignore test classes, as those generally report 0 references as they are managed via reflection in the testrunners.
sending a clean PR with a more accurate list of files

PR #1179, opened 2015-07-23
Akka.NET framework code analyzer

rogeralsing:
This is the initial commit of the code analyzer I'm building for Akka.NET. 
The analyzer is based on Roslyn and we can add different analyzers to it that detect different types of issues.

Currently, it only detects non-closed over Context/Self/Sender inside task continuations.

It does not patch the code, it only applies comments, making it easy to see where the problems was found and then let a human deal with the issue.
I placed the code inside "src/analyzers" as this does not need to be part of the actual akka solution itself, maybe it should even live in its own repo.

Eventually, this could probably be turned into a tool for Akka.NET users too, not only for the framework.


Aaronontheweb:
This is amazing! Does it all run on .NET 4.5? I don't know much about Roslyn's built-in tools.

rogeralsing:
I'm currently coding it in 4.6 but I think there is no technical limit that makes 4.5 impossible.
as this is a separate tool, it doesnt have to live by the same standards as the actual framework

rogeralsing:
I've moved this to a separate repo as it is not really part of the framework code

PR #1195, opened 2015-07-30
fixed #1194 (ThrottleTransportAdapter overflow bug)

### LINKED ISSUE #1194: ThrottleTransportAdapter throws OverflowException if system up time is longer than a few days
This bug was hilarious - never was able to find it on the build server because the system up time for each build agent is small (they're created on-demand,) but it's come up every time I've tried to run the Akka.Remote test suite on `dev` on my local machine since #1174 was merged in.

`ThrottlerTransportAdapter_must_maintain_average_message_rate` would fail, repeatedly, every time.

Here's the issue:

```csharp
sealed class TokenBucket : ThrottleMode
{
    readonly int _capacity;
    readonly double _tokensPerSecond;
    readonly long _nanoTimeOfLastSend;
    readonly int _availableTokens;

    public TokenBucket(int capacity, double tokensPerSecond, long nanoTimeOfLastSend, int availableTokens)
    {
        _capacity = capacity;
        _tokensPerSecond = tokensPerSecond;
        _nanoTimeOfLastSend = nanoTimeOfLastSend;
        _availableTokens = availableTokens;
    }

    int TokensGenerated(long nanoTimeOfSend)
    {
        // OVERFLOW EXCEPTION!!!
        return Convert.ToInt32(((double)(nanoTimeOfSend - _nanoTimeOfLastSend) / TimeSpan.TicksPerMillisecond) * _tokensPerSecond / 1000);
    }
}
```

The reason we have an `OverflowException` occur here is because `_nanoTimeOfLastSend` is set to 0 on the first calculation, so the result of `nanoTimeOfSend - _nanoTimeOfLastSend` is equal to the **entire system up time** on the first calculation, which causes the `OverflowException` when we attempt to cast down to an `int`.

Way to fix this is to set the `_nanoTimeOfLastSend` equal to the current `MontonicClock.Ticks` value when the `TokenBucket` is created. That way the delta is going to be something smaller (like, seconds worth of ticks) that can be downcast.

Aaronontheweb:
Trivial bug to fix - had one bug inside the `TokenBucket` where I forgot to convert nanoseconds to ticks   (1 tick = 100 nanoseconds) and a second bug inside the `TokenBucketSpec` where I also, forgot to convert nanoseconds to ticks.

This has been fixed and resolved the `OverflowException` that was occurring on my system.


Aaronontheweb:
Whoops, forgot the link: #1194

PR #1200, opened 2015-08-02
Reorganize  samples

wmeints:
Moved samples into structure that represents the various features of
Akka.NET. This fixes issue #1200


wmeints:
Oops, forgot that there's an overall solution file. Will fix the issue and update the request :-)

rogeralsing:
I've cloned your repo to get a better overview and I really like the changes, much more clear on what features are being shown.
(And it also highlights that we have a lack of F# examples)

The solution file seems to not have been updated though, thus the build server fails.
That should be updated before merging.

rogeralsing:
I'm closing this as it is extremely outdated and merge conflicts, please resubmit if issues gets fixed

Issue #147, opened 2014-06-18
Add CircuitBreaker support


rogeralsing:
We should to add CircuitBreaker support.

docs: http://doc.akka.io/docs/akka/snapshot/common/circuitbreaker.html

impl: https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/pattern/CircuitBreaker.scala

rogeralsing:
This will not be included in Akka.net, @HCanber might publish a separate package for the circuitbreaker

HCanber:
Oh, I haven't decided yet. :) 
I realised when I started coding it that there is very little dependency on Akka so it would be very easy to create it as a separate package (which would be useful outside Akka as the C# implementations I've seen so far don't handle concurrency that well). 
However there is also a small part that needs to be created to integrate it with Akka, so until decided we should keep this issue opened.

MAOliver:
I translated Akka's circuitbreaker to C# a while back
https://github.com/MAOliver/AkkaCircuitBreaker_InCS (Jun 21, 2013) along with the unit tests and test cases. 

It needs Option and some revisiting of some of the threading pieces (I remember this being a huge pain in the ass -- I was used to futures / promises and async / await was frustrating).

Hopefully, it's useful for someone : ).

rogeralsing:
Anyone? should we bring in @MAOliver's code or do we have some other impl we should use?

MAOliver:
After I finish up some side-work (another week or two), I can devote ~5 hours a week to updating this to the latest version from akka.io and writing up the new tests. Assuming no one has grabbed this one by then.

MAOliver:
Initial integration complete and all tests passing.

Still to complete:

- Proper XMLDocs for all new classes
- Proper Test Descriptions
- Verify awaited tasks that exceed the timeout are handled (thread should die)
- Look at thread execution context for this implementation and verify that there isn't any surprising behavior compared to scala implementation
- Look at the akka.io current version / tests and pull in changes since 2.0.5

Horusiath:
@MAOliver I've seen that you removed atomic counters in your implementation. Any specific reason for this?

MAOliver:
My original implementation used:

```
public abstract class AtomicState : AtomicInteger, IState
{
// base implementation
}
```

I saw that akka.net got rid of AtomicInteger in favor of AtomicCounter and AtomicCounterLong, so I changed my base "AtomicState" to:

```
[...]

PR #1201, opened 2015-08-03
147 circuit breaker

MAOliver:
#147 

- Add CircuitBreaker to codebase
- Add AtomicState, IAtomicState, OpenState, HalfOpenState, ClosedState, CircuitBreakerSpec

Ultimately I wound up ignoring the IScheduler and ExecutionContext / SynchronizationContext. I have implemented the tests from the Akka.io CircuitBreakerSpec tests and we can always come back later and add in the scheduler and execution context in a later commit and add suitable default values to a constructor.

### DIFF
--- a/src/core/Akka/Pattern/IllegalStateException.cs
+++ b/src/core/Akka/Pattern/IllegalStateException.cs
@@ -10,6 +10,10 @@ using Akka.Actor;
 
 namespace Akka.Pattern
 {
+    /// <summary>
+    /// Signals that a method has been invoked at an illegal or
+    /// inappropriate time.
+    /// </summary>
     public class IllegalStateException : AkkaException
     {
 
--- a/src/core/Akka/Util/Internal/AtomicState.cs
+++ b/src/core/Akka/Util/Internal/AtomicState.cs
@@ -1,0 +1,198 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="AtomicState.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+//     Copyright (C) 2013-2015 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Concurrent;
+using System.Runtime.ExceptionServices;
+using System.Threading.Tasks;
+
+namespace Akka.Util.Internal
+{
+    /// <summary>
+    /// Internal state abstraction
+    /// </summary>
+    internal abstract class AtomicState : AtomicCounterLong, IAtomicState
+    {
+        private readonly ConcurrentQueue<Action> _listeners;
+        private readonly TimeSpan _callTimeout;
+
+        protected AtomicState( TimeSpan callTimeout, long startingCount )
+            : base( startingCount )
+        {
+            _listeners = new ConcurrentQueue<Action>( );
+            _callTimeout = callTimeout;
+        }
+
+        /// <summary>
+        /// Add a listener function which is invoked on state entry
+        /// </summary>
+        /// <param name="listener">listener implementation</param>
+        public void AddListener( Action listener )
+        {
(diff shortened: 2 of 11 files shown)


src/core/Akka/Pattern/IllegalStateException.cs
```diff
@@ -10,6 +10,12 @@
 
 namespace Akka.Pattern
 {
+    /// <summary>
+    /// Signals that a method has been invoked at an illegal or
+    /// inappropriate time.  In other words, the Java environment or
+    /// Java application is not in an appropriate state for the requested
+    /// operation.
+    /// </summary>
```
Horusiath: This summary is quite inappropriate here.

src/core/Akka/Util/Internal/AtomicState.cs
```diff
@@ -0,0 +1,186 @@
...
+
+        /// <summary>
+        /// Notifies the listeners of the transition event via a 
+        /// </summary>
+        protected void NotifyTransitionListeners( )
+        {
+            if ( !HasListeners ) return;
+            Task
+                .Factory
+                .StartNew
```
Horusiath: Starting a new Task, but neither passing it out of the method nor waiting. Is it safe?

src/core/Akka/Util/Internal/AtomicState.cs
```diff
@@ -0,0 +1,186 @@
...
+            try
+            {
+                return await task;
+            }
+            finally
+            {
+                switch ( task.Status )
+                {
+                    case TaskStatus.RanToCompletion:
+                        if ( DateTime.Now.CompareTo( deadline ) < 0 )
```
rogeralsing: See this thread on datetime.now https://github.com/akkadotnet/akka.net/issues/846

src/core/Akka/Util/Internal/AtomicState.cs
rogeralsing: This does indeed look strange.
Why notify subscribers in a separate task? they are still notified sequentially inside the new task, so if one blocks, the others have to wait.
Not saying this is wrong, but some description on why would be interestng

src/core/Akka/Pattern/IllegalStateException.cs
MAOliver: @Horusiath I missed that one -- will correct and resubmit the pull request

src/core/Akka/Util/Internal/AtomicState.cs
MAOliver: @Horusiath @rogeralsing No -- it could lead to undesirable and hard to debug side effects depending on what type of actions users pass into the "AddListener" method. The original akka.io code adds objects which extend "Runnable" to a queue. It then passes each runnable to an executor to execute without blocking.

Since I've decided to ignore ExecutionContext for now -- the best way for me to handle this is to propagate the Async through the internal / protected methods up to the "CallThrough method" and await the result there. I'll try to implement the change this way and run through the tests and verify that it doesn't block the main thread.

src/core/Akka/Util/Internal/AtomicState.cs
MAOliver: My mistake. Will change to UtcNow.

src/core/Akka/Util/Internal/AtomicState.cs
MAOliver: This is going to take a bit longer. I have two failing tests because it's swallowing the exception. Will resubmit once I resolve the failures.

MAOliver: Actually -- I think I'm going to retract my statement about this being dangerous. I found a couple blogs, SO questions, MSDN docs in which this appears to be "normal" in the case where you want to "fire and forget" a task. I opted to await the task and make "NotifyTransitionListeners" async. 

I put a comment above the "Enter" task in AtomicState specifying that not awaiting the thread was intentional.

MAOliver:
@rogeralsing @Horusiath let me know if this works. I squashed all the commits.

Issue #1192, opened 2015-07-29
Exception naming inconsistencies


sean-gilliam:
I was going through the code about to further document what methods throw what exceptions and noticed a couple of things.

* in \src\core\Akka.Remote\Endpoint.cs these classes don't follow the .NET Exception naming convention, ```ShutDownAssociation```,  ```InvalidAssociation```, and ```HopelessAssociation```

* in \src\core\Akka.Remote.TestKit.BarrierCoordinator.cs these classes don't follow the .NET Exception naming convention, ```BarrierTimeout```, ```FailedBarrier```, ```DuplicateNode```, ```WrongBarrier```, ```BarrierEmpty```, and ```ClientLost```

* there are two exception with virtually the same name with one being a more general form of the other. ```InvalidAssociationException``` being the general form of ```InvalidAssociation```. It looks like ```InvalidAssociation``` is more address related than the former. Maybe rename it to ```InvalidAddressAssociation```?
  * InvalidAssociationException is in \src\core\Akka.Remote\Transport\Transport.cs
  * InvalidAssociation is in \src\core\Akka.Remote\Endpoint.cs

sean-gilliam:
I also noted quite a few ```throw new Exception(...)``` instead of a more specific ```Exception``` being thrown.

Aaronontheweb:
@sean-gilliam yeah, you're right - we should clean this up.

> I also noted quite a few throw new Exception(...) instead of a more specific Exception being thrown.

Might not be a big deal - depends on where it is. Are these inside the core modules?

> there are two exception with virtually the same name with one being a more general form of the other. InvalidAssociationException being the general form of InvalidAssociation. It looks like InvalidAssociation is more address related than the former. Maybe rename it to InvalidAddressAssociation?


yep, I like this idea. Go ahead and rename those - these exception types aren't propagated externally so there's no risk of breaking the API.

rogeralsing:
Ill add a Scan for "new Exception" to the analyzers too. We really should clean those up imo

sean-gilliam:
There are quite a few bare exceptions being thrown in the core modules.

```
C:\src\git\Akka.Net\akka.net\src\core\Akka.TestKit\AutoPilots.cs(78):            throw new Exception("Must not call");
  C:\src\git\Akka.Net\akka.net\src\core\Akka.TestKit\TestKitBase.cs(352):                    throw new Exception(string.Format(msg, system.Name, durationValue, ""));
[...]

PR #1205, opened 2015-08-04
Replaced generic exceptions with more specific ones

sean-gilliam:
Reference #1192

This PR replaces the generic exceptions with more specific ones. This
should help make debugging easier for apps that work with Akka.NET.

### DIFF
--- a/src/core/Akka.Remote.TestKit/Conductor.cs
+++ b/src/core/Akka.Remote.TestKit/Conductor.cs
@@ -63,7 +63,7 @@ namespace Akka.Remote.TestKit
         /// <returns></returns>
         public async Task<INode> StartController(int participants, RoleName name, INode controllerPort)
         {
-            if(_controller != null) throw new Exception("TestConductorServer was already started");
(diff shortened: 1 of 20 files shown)


rogeralsing:
Looks great!
The failed tests looks a bit funky though, do we have code that expects the old exception types anywhere?

sean-gilliam:
Thanks :) 

Hmm. At first glance, it looks to be timeout related. Maybe my changes on src/core/Akka.TestKit/TestKitBase.cs from Exception to TimeoutException could be causing this?

I'll dig it to it.

sean-gilliam:
Well I just ran those tests again on my end and I get green check marks. Maybe something going on with the build server?

rogeralsing:
@Aaronontheweb can we do something about the buildserver rebuilds, add a bot that we ourselves can activate or something?
Its been frustrating for a long time now that nobody else can activate this

Aaronontheweb:
@rogeralsing yeah, that's something I want to be able to do with Hubot but we don't have it working yet. cc @skotzko 

In the meantime, I can give you guys accounts for restarting builds. Would that work?

Issue #1206, opened 2015-08-05
Outdated FsPickler version in Akka.FSharp can lead to runtime errors when other FsPickler-dependent packages are installed


rdavisau:
Hi!

The current version of FsPickler referenced by Akka.FSharp is a pre 1.0 release and the latest version is 1.2.21. Between these versions, there has been a rationalisation of the API surface, making the newer FsPickler incompatible with the version Akka.FSharp has been written against. If another package referencing FsPickler (e.g. [Vagabond](https://github.com/nessos/Vagabond)) is installed alongside Akka.FSharp, NuGet will update FsPickler to the newer, incompatible version. This results in a `MissingMethodException` at runtime when Akka.FSharp attempts to call one of the removed overloads of `Serialize`. 

I will follow this with a PR :rocket:

Horusiath:
Closed by #1207

PR #1207, opened 2015-08-05
Update FsPickler version used by Akka.FSharp

rdavisau:
Ref #1206

I am new to both Akka and F#, so please go easy on me!

Per the referenced issue, Akka.FSharp is using an old version of FsPickler, which I have updated in this PR. The key change in the FsPickler API was the removal of overloads for `Serialize`/`Deserialize` taking a type as a method parameter; instead it is a generic parameter on the method. I made the few minor changes to account for this, hopefully they are idiomatic/appropriate.

I ran the F# tests by hand and they passed. I was also able to use the updated Akka.FSharp.dll in conjunction with Vagabond in an FSI session, which I was unable to do with the existing FsPickler.0.9.11-based version.


Aaronontheweb:
Welcome to the Akka.NET project.

PR #1209, opened 2015-08-05
added Google Analytics tracking to docs

Aaronontheweb:
Want to start measuring how much people are actually using the docs, so I embedded the getakka.net Google Analytics script into the sandcastle output.


Aaronontheweb:
Darnit! looks like this won't work. I'll have to come up with something else to solve this.

PR #1213, opened 2015-08-07
Fix for null sender #1212

### LINKED ISSUE #1212: Fix Null Sender issue
Per what we discussed on Gitter:

```csharp
Receive<GetCurrentLifecycleStage>(stage =>
            {
                if(_currentSessionState != null)
                    Sender.Tell(_currentSessionState); //SENDER can be null!
            });
```

We need to make it so `Sender` can't be null. Solution is to do what JVM Akka does:

```scala
  /*
   * ACTOR CONTEXT IMPLEMENTATION
   */

  final def sender(): ActorRef = currentMessage match {
    case null                      ⇒ system.deadLetters
    case msg if msg.sender ne null ⇒ msg.sender
    case _                         ⇒ system.deadLetters
  }
```

Coalesce the null into Deadletters.

Aaronontheweb:
Resolves #1212 - any messages for a `null` Sender now to go `Deadletters`.

Added a spec to verify.

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorRefSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorRefSpec.cs
@@ -279,6 +279,17 @@ namespace Akka.Tests.Actor
             Assert.True(!(bool)t2.Result);
         }
 
+        [Fact]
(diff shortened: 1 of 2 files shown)


rogeralsing:
Perfect

PR #1220, opened 2015-08-08
1088 - Skip attribute on MultiNodeFact does not work

barambani:
It seems that the discovery process was ignoring the presence of the Skip in the MultiNodeFact attribute.
I added a filter that checks if it is null or not and prevents the skipped facts from being added to the tests to run.


barambani:
There's another pull request for this issue and it passes all the tests. I'm closing this one.

Issue #1088, opened 2015-06-23
`Skip` attribute on MultiNodeFact does not work


Aaronontheweb:
Still ran the tests I wanted to skip :\

![image](https://cloud.githubusercontent.com/assets/326939/8315198/afe4bc6a-19a3-11e5-8b76-513ce94e40e7.png)

schatekar:
Looking at the source for `MultiNodeFact` below.

This is returning `null` from skip if the test is run using MultiNodeTestRunner or NodeTestRunner. I am not sure if XUnit treats a null `Skip` value as test not to be skipped or not but if I assume that then clearly any test run using MultiNodeTestRunner would never be skipped, isn't it?

```csharp
    public class MultiNodeFactAttribute : FactAttribute
    {
        public static Lazy<bool> ExecutedByMultiNodeRunner =
            new Lazy<bool>(() =>
            {
                var args = Environment.GetCommandLineArgs();
                if (args.Length == 0) return false;
                var firstArg = args[0];
                return firstArg.Contains("Akka.MultiNodeTestRunner") 
                    || firstArg.Contains("Akka.NodeTestRunner");
            });

        public override string Skip
        {
            get
            {
                return ExecutedByMultiNodeRunner.Value
                    ? null
                    : "Must be executed by multi-node test runner";
            }
            set { base.Skip = value; }
        }
    }
```

cpx:
Just started looking at this as well :) @schatekar, a null value means test should be run, and non-null means it should be skipped.

schatekar:
I think I know how to fix this. If you are not fixing it I can take a crack
at it in few hours.

cpx:
It's all yours :)

Aaronontheweb:
You guys are awesome :)

schatekar:
This is more involved than I had imagined. I would give it another try over the weekend.

Aaronontheweb:
> This is more involved than I had imagined. I would give it another try over the weekend.

I know that feeling well, my friend.

barambani:
I gave a try to this as well, I guess because is less intimidating than the others :), it looks like that removing the null is not enough. I could enable the runner to skip the multinode facts checking the attribute in the

schatekar:
@barambani  I did not get a message that you have worked on the issue as well. I just went ahead and submitted my fix without knowing that you too have submitted a fix.

barambani:
@schatekar no problem, I will close mine.

schatekar:
[...]

PR #1221, opened 2015-08-08
#1088 MultiNode tests can now be skipped by specifying a SkipReason

schatekar:
#1088 MultiNode tests can now be skipped by specifying a SkipReason

I have learned that XUnit works in two phases - discovery of tests and execution of tests. 

The tests that are to be skipped are skipped during execution and not discovery. I believe this is because you would want to show a message that a test was not executed but skipped for a particular reason. Since MultiNodeTestRunner has its own test execution logic, a change was needed in that logic to skip the tests.


annymsMthd:
Skipped tests are bad

schatekar:
@annymsMthd I get the point but a mechanism is needed to stop the randomly failing tests from failing the build until we fix them.

rogeralsing:
:+1:

Aaronontheweb:
@annymsMthd currently the clustered router tests are skipped - there's something inherently racy with the tests themselves.

PR #1229, opened 2015-08-10
make Received and Connected inherit from Tcp.Event

johlrich:
Noticed these two messages do not inherit from Tcp.Event while playing with the new version


rogeralsing:
Nice catch!

rogeralsing:
cc @Aaronontheweb we need a re-run on this. two failing IO tests, probably not related but better make sure

PR #1230, opened 2015-08-10
Do not send if payload size > max allowed. Fixes #879

### LINKED ISSUE #879: Akka.Remote: implement "log frame size exceeding"
Is there a way to log *everything* that's happening in an actor and, if possible, the actor system in general?

I'm using TopShelf for service hosting and Serilog (min level: debug) for logging.

Including:

    akka {
        (...)
         actor {
              loglevel = DEBUG
              stdout-loglevel = DEBUG
              provider = "Akka.Cluster.ClusterActorRefProvider, Akka.Cluster"
              debug {
                 receive = on
                 autoreceive = on
                 fsm = on
                 lifecycle = on
                 unhandled = on
                 event-stream = on
                 router-misconfiguration = on
              }
        loggers = ["Akka.Logger.Serilog.SerilogLogger,Akka.Logger.Serilog"]
        remote {
              log-remote-lifecycle-events = DEBUG
              (...)
         }
    }

in the configuration doesn't change anything when you compare the log output without the `debug` section specified at all.

I'm asking because I'm trying to track a really bizarre behavior (one message, which has 4 collections in it, all of collection items have `Date` (DateTime) member, is going nowhere unless I set members' dates to `default(DateTime)` in at least one of the collections...?!) but it might be hard to reproduce. So I'd rather exhaust all my monitoring and debugging options first before spending half my day setting up a separate solution...

cpx:
- Added property MaximumPayloadBytes to Transport. For HeliosTransport
  this is the same as MaxFrameSize.
- If payload size > MaximumPayloadBytes, log OversizedPayloadException
  and do not send message.

Corresponding Akka Scala code:
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Endpoint.scala#L745
https://github.com/akka/akka/blob/master/akka-remote/src/test/scala/akka/remote/RemotingSpec.scala#L527
https://github.com/akka/akka/search?utf8=%E2%9C%93&q=maximumpayloadbytes

### DIFF
--- a/src/core/Akka.Remote/Endpoint.cs
+++ b/src/core/Akka.Remote/Endpoint.cs
@@ -1234,17 +1234,28 @@ namespace Akka.Remote
 
                 //todo: RemoteMetrics https://github.com/akka/akka/blob/dc0547dd73b54b5de9c3e0b45a21aa865c5db8e2/akka-remote/src/main/scala/akka/remote/Endpoint.scala#L742
 
-                //todo: max payload size validation
-
-                var ok = _handle.Write(pdu);
-
-                if (ok)
+                if (pdu.Length > Transport.MaximumPayloadBytes)
                 {
-                    _ackDeadline = NewAckDeadline();
-                    _lastAck = null;
+                    var reason = new OversizedPayloadException(
+                        string.Format("Discarding oversized payload sent to {0}: max allowed size {1} bytes, actual size of encoded {2} was {3} bytes.",
+                            send.Recipient,
+                            Transport.MaximumPayloadBytes,
+                            send.Message.GetType(),
+                            pdu.Length));
+                    _log.Error(reason, "Transient association error (association remains live)");
                     return true;
                 }
+                else
+                {
+                    var ok = _handle.Write(pdu);
 
+                    if (ok)
+                    {
+                        _ackDeadline = NewAckDeadline();
+                        _lastAck = null;
+                        return true;
+                    }
+                }
                 return false;
             }
             catch (SerializationException ex)
@@ -1519,6 +1530,7 @@ namespace Akka.Remote
             _provider = RARP.For(Context.System).Provider;
         }
 
+        private readonly ILoggingAdapter _log = Context.GetLogger();
         private readonly AkkaPduCodec _codec;
         private readonly IActorRef _reliableDeliverySupervisor;
         private readonly ConcurrentDictionary<EndpointManager.Link, EndpointManager.ResendState> _receiveBuffers;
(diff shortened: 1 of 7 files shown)


rogeralsing:
Looks solid

cpx:
All of those failed tests seem to pass on my machine. Build server instability, or did I miss something?

Aaronontheweb:
@cpx thanks for sending this in! Some of the tests are ones that are known to be racy, but your new test that you added has failed too. The smaller hardware we run the build server on is good at exposing race conditions that our powerful development laptops miss. Take a look at the error message and see if there's something going on with the code you added or the test you wrote, because if the test fails then it means it's sensitive to the latency from lower degrees of concurrency (crappy CPUs and less of them :p )

cpx:
@Aaronontheweb Well, it fails on the exception expectation; it received 2 OversizedPayloadException events instead of one. Seems strange to me since we're only sending one message and there's only one place where the error is logged. I"m a bit stumped; any tips on how that might happen?

src/core/Akka.Remote/Transport/TestTransport.cs
```diff
@@ -51,10 +54,11 @@ private static string GetConfigString(Config conf, string name)
             return value;
         }
 
-        public TestTransport(Address localAddress, AssociationRegistry registry, string schemeIdentifier = "test")
+        public TestTransport(Address localAddress, AssociationRegistry registry, long maximumPayloadBytes = 32000, string schemeIdentifier = "test")
```
Aaronontheweb: Does the Scala JVM transport support this constructor out of the box? Been a while since I've looked

src/core/Akka.Remote/Transport/Transport.cs
```diff
@@ -21,6 +21,7 @@ public abstract class Transport
         public ActorSystem System { get; protected set; }
 
         public virtual string SchemeIdentifier { get; protected set; }
+        public virtual long MaximumPayloadBytes { get; protected set; }
```
Aaronontheweb: I just double checked to make sure that this is also exposed on the JVM transport, and it is: https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/transport/Transport.scala#L92

So :+1:

src/core/Akka.Remote/Transport/TestTransport.cs
Aaronontheweb: Why... yes it does. https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/transport/TestTransport.scala#L36

:+1:

Aaronontheweb:
@cpx do we have an equivalent of this https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Endpoint.scala#L934 on the `EndpointReader` in this PR? Could you add it?

src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1234,17 +1234,27 @@ private bool WriteSend(EndpointManager.Send send)
...
-                    _ackDeadline = NewAckDeadline();
-                    _lastAck = null;
-                    return true;
+                    var reason = new OversizedPayloadException(
+                        string.Format("Discarding oversized payload sent to {0}: max allowed size {1} bytes, actual size of encoded {2} was {3} bytes.",
+                            send.Recipient,
+                            Transport.MaximumPayloadBytes,
+                            send.Message.GetType(),
+                            pdu.Length));
+                    _log.Error(reason, "Transient association error (association remains live)");
```
Aaronontheweb: To get the spec to pass, you need to add `return true;` on a new line below here - this will signal the `EndpointWriter` not to queue up the oversized message for backoff + retry.

This is also equivalent to what the Scala code looks like: https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Endpoint.scala#L748

cpx:
@Aaronontheweb Ah, what a silly mistake from my side. Thank you for the help!
About the case in the receive function, I will check that ASAP. Definitely looks like we're missing that part.

cpx:
There. Now we are dropping both sent and received messages if payload is too large.

One thing I noticed when debugging though which I found a bit strange, is that the calculated payload size in the EndpointWriter is much larger than the corresponding size in the EndpointReader. It seems like it includes the envelope when sending, but only includes the actual message when receiving. Is that by design or mistake?

Aaronontheweb:
@cpx looks like my pre-build macro I added in Akka.MultiNodeTestRunner in my previous PR caused some problems this time. Please remove it if you don't mind :p 

That's an interesting point.... I think it's intentional though but I'll have to double check.

Once you've fixed that build issue, could you squash all of these commits onto dev? `Contributing.md` outlines the steps for doing that if you need help.

Aaronontheweb:
Looks great - nicely done @cpx !

PR #1232, opened 2015-08-11
Migrant serializer

rogeralsing:
# Do not merge

This is a first PoC using the Migrant serializer instead of Json.NET

Migrant handles all of the problems that we have had with Json.NET, primitives are deserialized with their correct type, int will be int, long will be long etc. F# Discriminated Unions works fine, even when stored in an object property.

There are a lot more to investigate here, e.g. is migrant threadsafe ? does it alter its own state when serializing/deserializing? (e.g. caching)
But from my early tests, this is exactly what we have been looking for.

@Horusiath can you pull this one and experiment with F# and persistence modules to see if there are anything funky that I have missed so far?

### DIFF
--- a/src/core/Akka.Tests/Serialization/SerializationSpec.cs
+++ b/src/core/Akka.Tests/Serialization/SerializationSpec.cs
@@ -495,7 +495,7 @@ namespace Akka.Tests.Serialization
             Sys.Serialization.FindSerializerFor(null).GetType().ShouldBe(typeof(NullSerializer));
             Sys.Serialization.FindSerializerFor(new byte[]{1,2,3}).GetType().ShouldBe(typeof(ByteArraySerializer));
             Sys.Serialization.FindSerializerFor("dummy").GetType().ShouldBe(typeof(DummySerializer));
-            Sys.Serialization.FindSerializerFor(123).GetType().ShouldBe(typeof(NewtonSoftJsonSerializer));
+            Sys.Serialization.FindSerializerFor(123).GetType().ShouldBe(typeof(MigrantSerializer));
         }
 
 
--- a/src/core/Akka/Akka.csproj
+++ b/src/core/Akka/Akka.csproj
@@ -54,6 +54,10 @@
     <Optimize>false</Optimize>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Migrant, Version=0.10.3.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Migrant.0.10.3\lib\net40\Migrant.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="Newtonsoft.Json, Version=7.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\packages\Newtonsoft.Json.7.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
@@ -266,6 +270,7 @@
     <Compile Include="IO\TcpListener.cs" />
     <Compile Include="IO\TcpManager.cs" />
     <Compile Include="IO\TcpOutgoingConnection.cs" />
+    <Compile Include="Serialization\MigrantSerializer.cs" />
     <Compile Include="Util\ByteIterator.cs" />
     <Compile Include="Util\ByteString.cs" />
     <Compile Include="Util\ContinuousEnumerator.cs" />
@@ -379,7 +384,6 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
-  <ItemGroup />
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
--- a/src/core/Akka/Configuration/Pigeon.conf
+++ b/src/core/Akka/Configuration/Pigeon.conf
@@ -391,7 +391,7 @@ akka {
 
     serializers {
         json = "Akka.Serialization.NewtonSoftJsonSerializer"
-		migrant = "Akka.Serialization.MigrantSerializer"
+        migrant = "Akka.Serialization.MigrantSerializer"
         java = "Akka.Serialization.JavaSerializer"				# not used, reserves java serializer identifier
         bytes = "Akka.Serialization.ByteArraySerializer"
     }
--- a/src/core/Akka/Configuration/Pigeon.conf
+++ b/src/core/Akka/Configuration/Pigeon.conf
@@ -391,6 +391,7 @@ akka {
 
     serializers {
         json = "Akka.Serialization.NewtonSoftJsonSerializer"
+		migrant = "Akka.Serialization.MigrantSerializer"
         java = "Akka.Serialization.JavaSerializer"				# not used, reserves java serializer identifier
         bytes = "Akka.Serialization.ByteArraySerializer"
     }
@@ -404,7 +405,7 @@ akka {
     # "java.io.Serializable" = none
     serialization-bindings {
       "System.Byte[]" = bytes
-      "System.Object" = json
+      "System.Object" = migrant
     }
   }
  
--- a/src/core/Akka/Serialization/MigrantSerializer.cs
+++ b/src/core/Akka/Serialization/MigrantSerializer.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.IO;
-using System.Security.Policy;
 using Akka.Actor;
 using Akka.Util;
 
@@ -25,17 +24,10 @@ namespace Akka.Serialization
               public MigrantSerializer(ExtendedActorSystem system) : base(system)
         {
             _serializer = new Antmicro.Migrant.Serializer();
-
-            _serializer.ForObject<ISurrogated>().SetSurrogate(x => 
-                x.ToSurrogate(system));
-
-            _serializer.ForSurrogate<ISurrogate>().SetObject(x => 
-                x.FromSurrogate(system));
-
-            _serializer.ForObject<Type>().SetSurrogate(t => 
-                new TypeSurrogate(t.AssemblyQualifiedName));
-            _serializer.ForSurrogate<TypeSurrogate>().SetObject(t => 
-                t.Restore());
+            _serializer.ForObject<ISurrogated>().SetSurrogate(x => x.ToSurrogate(system));
+            _serializer.ForSurrogate<ISurrogate>().SetObject(x => x.FromSurrogate(system));
+            _serializer.ForObject<Type>().SetSurrogate(t => new TypeSurrogate(t.AssemblyQualifiedName));
+            _serializer.ForSurrogate<TypeSurrogate>().SetObject(t => t.Restore());
         }
         
 
--- a/src/core/Akka/Serialization/MigrantSerializer.cs
+++ b/src/core/Akka/Serialization/MigrantSerializer.cs
@@ -1,0 +1,72 @@
+﻿using System;
+using System.IO;
+using System.Security.Policy;
+using Akka.Actor;
+using Akka.Util;
+
+namespace Akka.Serialization
+{
+    public class TypeSurrogate
+    {
+        public TypeSurrogate(string typeName)
+        {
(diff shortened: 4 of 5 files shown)


rogeralsing:
Seeing the same persistence issues as I saw locally, the odd thing is that those eventually passed when re-running the tests a few times.
Probably related to the serializer but not sure is actually causing it.

Good news is that migrant also serializes Expressions, so we might even be able to remove the FsPickler dependency.

rdavisau:
Hello, as an interested party I thought I'd have a play - [this gist](https://gist.github.com/rdavisau/0b4077b56fe60a5fca06) testing DUs with Akka.Cluster works with migrant when run as a console application, but it fails when executed in FSI. Actually, after further fiddling, it appears to me that only primitives and strings are OK when using FSI + cluster, other complex types fail as below:

```
[ERROR][11/08/2015 10:01:31 PM][Thread 0020][akka://cluster/system/endpointManager/reliableEndpointWriter-akka.tcp%3a%2f%2fcluster%40localhost%3a54054-1/endpointWriter] Failed to write message to the transport
Cause: Akka.Remote.EndpointException: Failed to write message to the transport ---> System.ArgumentException: Multimoduled assemblies are not supported yet.
   at Antmicro.Migrant.VersionTolerance.AssemblyDescriptor..ctor(Assembly assembly)
   at Antmicro.Migrant.TypeDescriptor.Init(Type t)
   at Antmicro.Migrant.TypeDescriptor.CreateFromType(Type type)
   at Antmicro.Migrant.ObjectWriter.TouchAndWriteTypeId(Type type)
   at Antmicro.Migrant.ObjectWriter.WriteObjectInner(Object o)
   at Antmicro.Migrant.ObjectWriter.InvokeCallbacksAndWriteObject(Object o)
   at Antmicro.Migrant.ObjectWriter.WriteObject(Object o)
   at Antmicro.Migrant.Serializer.Serialize(Object obj, Stream stream)
   at Akka.Serialization.MigrantSerializer.ToBinary(Object obj) in C:\Users\rdavis\Source\Repos\akka.net-roger\src\core\Akka\Serialization\MigrantSerializer.cs:line 48
   at Akka.Remote.Serialization.MessageContainerSerializer.SerializeActorSelectionMessage(ActorSelectionMessage sel) in C:\Users\rdavis\Source\Repos\akka.net-roger\src\core\Akka.Remote\Serialization\MessageContainerSerializer.cs:line 69
```

You can see the test for multimodule [here](https://github.com/antmicro/Migrant/blob/e0dbbaf608a3424cd4e0ca84167b1f80d44d3a46/Migrant/VersionTolerance/AssemblyDescriptor.cs#L82-#L101); I have to admit I don't quite understand what they are or what the implication of dealing with one is.

rogeralsing:
@rdavisau so the problem only occurs in the REPL for F#? 
The error message is from within Migrant, and a fair guess would be that FSI compiles some types as you go along and then place them in different dynamic assemblies. not sure but that would be my guess.

rdavisau:
@rogeralsing yes, just in the REPL. First result on google for multi-module assemblies yields [this old post](http://blogs.msdn.com/b/junfeng/archive/2004/07/15/183813.aspx) which hints at reasons why they probably aren't common or preferred. I see you opened an issue at Migrant; I wonder how much work would be involved in relaxing the single-module assumption..

Horusiath:
@rogeralsing @rdavisau If that's the case, I think we should not hesitate to write an issue on Migrant repo for it. If we won't signal the demand for that feature, we cannot hope, it will be fixed any time soon.

rogeralsing:
Done and done :) they are looking into it

rogeralsing:
https://github.com/antmicro/Migrant/issues/88

rdavisau:
As I mentioned in Gitter, I've been playing with the Migrant fix and it's all working well. I have noticed that when I did something incorrect that I saw an error that might suggest there are a few more changes needed in the serializer replacement.

I wrote: 
```fsharp
{list of actors}
        |> List.iter (fun a -> a <? Marco |> Async.RunSynchronously)
```

Here I was changing from having tested a `Tell`, to testing an `Ask`. Of course, I forgot to change to `List.map` and to cast the result, but the code above produced this:

```
System.AggregateException: One or more errors occurred. ---> System.InvalidCastException: Unable to cast object of type 'Akka.Serialization.MigrantSerializer' to type 'Akka.Serialization.NewtonSoftJsonSerializer'.
   at Akka.FSharp.Actors.tryCast[Message](Task`1 t)
   at Akka.FSharp.Actors.op_LessQmark@87.Invoke(Task`1 arg)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
>    at Microsoft.FSharp.Control.AsyncBuilderImpl.commit[a](Result`1 res)
   at Microsoft.FSharp.Control.CancellationTokenOps.RunSynchronously[a](CancellationToken token, FSharpAsync`1 computation, FSharpOption`1 timeout)
   at Microsoft.FSharp.Control.FSharpAsync.RunSynchronously[T](FSharpAsync`1 computation, FSharpOption`1 timeout, FSharpOption`1 cancellationToken)
   at Microsoft.FSharp.Primitives.Basics.List.iter[T](FSharpFunc`2 f, FSharpList`1 x)
   at Microsoft.FSharp.Collections.ListModule.Iterate[T](FSharpFunc`2 action, FSharpList`1 list)
   at <StartupCode$FSI_0010>.$FSI_0010.main@() in C:\Users\rdavis\Documents\Visual Studio 2015\Projects\LostAlpakka\LostAlpakka.Control\Main.fs:line 130
---> (Inner Exception #0) System.InvalidCastException: Unable to cast object of type 'Akka.Serialization.MigrantSerializer' to type 'Akka.Serialization.NewtonSoftJsonSerializer'.
   at Akka.FSharp.Actors.tryCast[Message](Task`1 t)
   at Akka.FSharp.Actors.op_LessQmark@87.Invoke(Task`1 arg)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()<---
```

I'm not suggesting what I wrote should be supported, rather that the invalid cast might indicate part of the serialization code that did not anticipate the Migrant serializer being used. 

As far as I can tell however, everything is working great - cluster + discriminated unions + FSI :+1:

Horusiath:
@rdavisau My guess is that's because of my JSON.NET hack I've introduced before. It should be removed, when we settle on using migrant.

rdavisau:
Ah - good to know!

I forgot - I have not tested on mono yet. That might be covered by your tests already, but I will give it a shot over the next day or so.

rdavisau:
Have not looked at mono yet, but in increasing the complexity of experiments I found that F# option types do not seem to be supported - https://github.com/antmicro/Migrant/issues/91. The Migrant team has been pretty responsive so I expect they'll comment on viability fairly quickly.

rogeralsing:
I did some perf tests today to compare Migrant to Json.NET in terms of perf.
That didnt look tóo good TBH. Json.NET is 24 times faster than migrant here. 

The test is very naive but it should give some pointers to the general performance.
https://gist.github.com/rogeralsing/7b57167935ce048859e7

100 000 objects takes 4.4 sec on my machine using Migrant, and 0.18 sec for Json.NET.

rogeralsing:
I've made some progress addresing the performance fixes for Migrant.
**My local fork of Migrant is now 3 times slower than Json.NET, instead of 27 times slower.**
If we can get it down to 2 times slower, I think it would be fine to use it, Json.NET is insanely fast now a days (it wasnt earlier) so twice as slow is actually really fast.

The reasons Migrant had these problems is not really their fault, the lib was simply not designed to be used the way we use it, serializing individual objects one at a time, but rather batches of objects, where Migrant is 5 times faster than json.net.

rogeralsing:
I'm closing this.
Even after patches, Migrant just cant deliver the performance we need here.
I'm going to invest my time in Wire instead.

wirevsmigrant

konrad-kruczynski:
Hi Roger,
in our tests Migrant is close to Protobuf.NET in performance, sometimes being faster, sometimes slower, so your results are quite surprising for me - and we were definitely much faster than binary formatter. Even if you do not plan to use Migrant, could you share the performance testing code? I wonder what is the reason.

rogeralsing:
@konrad-kruczynski are the fixes in the Master branch or are they elsewhere?
I downloaded the migrant repo and built the assembly from there.
Are there any specific settings I need to check for the serializer, I did not see any new settings so maybe I'm using the old code

rogeralsing:
```csharp
        public class Poco
        {
           public string Name { get; set; }
           public int Age { get; set; }
        }

        private static readonly Poco poco = new Poco
        {
            Age = 123,
            Name = "Hello",
        };
        private static void SerializePocoProtoMigrant()
        {
            var ser = new Antmicro.Migrant.Serializer(new Settings());
            var s = new MemoryStream();
            ser.Serialize(poco,s);
            RunTest("Migrant", () =>
            {
                var stream = new MemoryStream();
                ser.Serialize(poco,stream);
            }, () =>
            {
                s.Position = 0;
                ser.Deserialize<Poco>(s);
            });
        }

        private static void RunTest(string testName, Action serialize, Action deserialize)
        {
            var tmp = Console.ForegroundColor;
            Console.ForegroundColor = ConsoleColor.Yellow;
            Console.WriteLine($"{testName}");
            Console.ForegroundColor = tmp;
            var sw = Stopwatch.StartNew();
            for (var i = 0; i < 1000000; i++)
            {
                serialize();
            }
            sw.Stop();
            Console.WriteLine($"   {"Serialize".PadRight(30,' ')} {sw.ElapsedMilliseconds} ms");
            var sw2 = Stopwatch.StartNew();
            for (var i = 0; i < 1000000; i++)
            {
                deserialize();
            }
            sw2.Stop();
            Console.WriteLine($"   {"Deseralize".PadRight(30, ' ')} {sw2.ElapsedMilliseconds} ms");
            Console.WriteLine($"   {"Total".PadRight(30, ' ')} {sw.ElapsedMilliseconds + sw2.ElapsedMilliseconds} ms");
        }
```

konrad-kruczynski:
Some patches are not yet there, however these results surprise even without them - especially the deserialization part.

rogeralsing:
@konrad-kruczynski the current Migrant Master branch yields _a lot_ better result than before,

migrantperf

Actually faster than the current serializer settings we have in Akka.NET. so I will continue to evaluate and see which serializer best fits our needs.

mateusz-holenko:
We are still working on few tweaks to make `Migrant` even better. We hope to commit them until the end of the week.

As for now, did you use new setting (called  `disableTypeStamping`) while running presented test? We introduced it in order to improve performance by limiting number of type structure scanning operations and stream size.

rogeralsing:
@mateusz-holenko yes, I enabled that one

konrad-kruczynski:
Hi Roger,
you can try the newest master. You may experience some slight
additional improvements. However, the whole performance subject is
broader.

First of all maybe you could use `OpenStreamSerializer`. It is
currently tied to the stream it is constructed with. This is because
we serialize the metadata only once for each OSS. If you could have
one stream and OpenStreamSerializer for each receiver, then it would
be beneficial - the times in your simple would be 4 - 5 times better
than with regular serialization.

If you need to send your own data through the same stream, we might
change the API to allow it - so you could send one package of
Migrant-serialized data, one package of your own data, one package
from Migrant, and so on.

The other thing is whether this test actually represents your real
life scenario. The serialized object is very simple (just a POCO, not
a complicated tree of objects referencing each other), in which case
even small bookkeeping stuff, like writing a header, takes relatively
a lot of time. If that's really the case, then I would say that
Migrant is indeed not too well suited here. But I guess that you're
actually interested in a more complicated scenario in which Migrant
could really excel. We can also investigate further performance
improvements but for that we would need a more complex test resembling
real life scenario case.

konrad-kruczynski:
Hi Roger,
have you come to any conclusions regarding this case?

rogeralsing:
@konrad-kruczynski we are still holding every option open.
What we cannot do though, is use the OpenStreamSerializer due to two main reasons.
1) Our API and meachanics are bound to mimic the JVM version as close as possible, the serialization support is pluggable and must rely on per message independent serialization.
2) Akka-Persistence uses the same serailizer support to write messages to a persistent store, deserializing event logs would in such case need to read the entire history in order to get all relevant data.

But Migrant is still a viable alternative to what we have now (the highly adapted Json.NET setup).
We still need to evaluate which serializer gives us the best over all experience, speed/size/ability to serialize anything etc.

I'll keep you updated once we have any progress on this

konrad-kruczynski:
@rogeralsing I see, thanks for the info.

konrad-kruczynski:
BTW, if one guaranteed that the assemblies are used on both sides, we would be able to use metadata tokens as IDs, instead of assembly qualified names which could speed up serialization. Maybe this is a useful scenario.

PR #1234, opened 2015-08-11
akka-io: Ported a few tests in ByteStringSpec with bug fixes

willieferguson:
I found a few issues with ByteString while using akka-io. This PR contains a few tests from the ByteStringSpec, as well as fixes to bugs identified with these tests.

ByteStringSpec uses FsCheck, a port of ScalaCheck.


rogeralsing:
:+1:

PR #1240, opened 2015-08-12
akka-io: Stop select loop on idle

### LINKED ISSUE #1225: High CPU load using the Akka.IO TCP server
The following simple program:
namespace Proto1.Client
{
    using Akka.Actor;
    using Akka.IO;
    using System;

    class Program
    {
        {
            var system = ActorSystem.Create("proto1");
            var manager = system.Tcp(); //this causes 100% CPU
            Console.ReadLine();
        }
    }
}

On my machine it uses around 25% CPU. Starting the console app 4 times gives me 100% CPU utilization:

2015-08-08 23_45_53-performance monitor

I am using i7 cpu:
2015-08-08 23_50_42-task manager

Using nighly build Akka 1.0.4.64-beta

willieferguson:
The PR fixes #1225. The select loop is now only started once sockets are registered for selection, and stopped when no sockets are interested anymore.

### DIFF
--- a/src/core/Akka/IO/SelectionHandler.cs
+++ b/src/core/Akka/IO/SelectionHandler.cs
@@ -196,7 +196,6 @@ namespace Akka.IO
             {
                 _log = log;
                 _executionContext = new SingleThreadExecutionContext();
-                Execute(Select);
             }
 
             private void Execute(Action action)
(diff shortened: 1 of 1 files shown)


src/core/Akka/IO/SelectionHandler.cs
```diff
@@ -262,14 +260,24 @@ private void EnableInterest(SocketChannel channel, SocketAsyncOperation op)
...
-                            break;
-                        case SocketAsyncOperation.Connect:
-                        case SocketAsyncOperation.Send:
-                            Execute(() => _write.Add(channel.Socket, channel));
-                            break;
+                    case SocketAsyncOperation.Accept:
+                    case SocketAsyncOperation.Receive:
+                        Execute(() =>
+                        {
+                            _read.Add(channel.Socket, channel);
```
rogeralsing: this looks racy to me, _read is a dictionary and is now executed on another thread.
Or?

willieferguson: All access to _read and _write are currently on a single thread - the one used for the select loop.

rogeralsing: :+1:

PR #1242, opened 2015-08-12
Upgrade FAKE to version 4.1.0

### LINKED ISSUE #1241: Upgrade Fake version to cope with VS 2015 only machines.
The version of Fake used means that machines with only 2015 can't run any VS tests

GraemeBradbury:
Altered build command to install 4.1.0
Added .fake/ directory to .gitignore. Directory contains build artifacts.
Removed Fake 3.17.0 from packages.config as build.cmd dicates version.

Fixes #1241


rogeralsing:
:+1:

Issue #1223, opened 2015-08-08
Port Event Adapters to Akka.Persistence


Horusiath:
[Link](http://doc.akka.io/docs/akka/2.4-M2/scala/persistence.html#Event_Adapters) to documentation. **tl;dr** Event adapters are a layer between journals and persistent actors, that allows to define custom mappings and interpretation of stored events, before they reach an actor. It's usefull:

- in cases then events need to be versioned
- when events may be mapped to data types understood by persistence backend providers

Horusiath:
Closed by #1262

PR #1262, opened 2015-08-15
implemented event adapters

Horusiath:
This is implementation of [Event Adapters](http://doc.akka.io/docs/akka/snapshot/scala/persistence.html#Event_Adapters) feature from canonical akka. They major role is ability to separate model stored by journal from domain model send by persistent actors through configurable mappers implementing `IEventAdapter` interface. Examples of use cases may be journal event versioning or addjusing domain model to specific schema expected by persistent backend.

Test suite is attached.
This feature should be documented.

cc #1223


Aaronontheweb:
> This feature should be documented.

I agree - can you add a page to the docs as part of this pull request? Drop a link to your PR against http://getakka.net/ as part of this PR por favor.

rogeralsing:
:+1:

PR #1265, opened 2015-08-19
Added ability to filter multinode test assemblies

GraemeBradbury:
By specifying 'spec-assembly' as part of the build it'll filter
out multinode tests assemblies that don't have the supplied word
in their name.

eg: build multinodetests spec-assembly=remote
Will run all multinode tests from all assemblies with 'remote' in
their name.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -248,8 +248,12 @@ Target "RunTestsMono" <| fun _ ->
         xunitTestAssemblies
 
 Target "MultiNodeTests" <| fun _ ->
+    let testSearchPath =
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
This is awesome @GraemeBradbury - just started re-running the tests.

Could you do me a favor and add some documentation to this page in our repo, including this?

https://github.com/akkadotnet/getakka.net/blob/master/src/docs/Building%20and%20Distributing%20Akka.NET.md

We explain how to generally use our FAKE command line, but not how to launch the multi-node tests specifically. That would be really helpful!

rogeralsing:
:+1:

Issue #1247, opened 2015-08-14
Port Akka.Remote MultiNodeSpec: RemoteDeliverySpec


Aaronontheweb:
https://github.com/akka/akka/blob/master/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteDeliverySpec.scala

GraemeBradbury:
I'll take this.

Aaronontheweb:
Resolved via #1266

PR #1266, opened 2015-08-19
Remote Delivery Spec

GraemeBradbury:
Addresses #1247

### DIFF
--- a/src/core/Akka.Remote.Tests.MultiNode/RemoteDeliverySpec.cs
+++ b/src/core/Akka.Remote.Tests.MultiNode/RemoteDeliverySpec.cs
@@ -1,0 +1,131 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using Akka.Actor;
+using Akka.Remote.TestKit;
+
+namespace Akka.Remote.Tests.MultiNode
+{
+    public class RemoteDeliveryMultiNetSpec : MultiNodeConfig
+    {
+        public RemoteDeliveryMultiNetSpec()
+        {
+            First = Role("first");
+            Second = Role("second");
+            Third = Role("third");
+
+            CommonConfig = DebugConfig(false);
+        }
+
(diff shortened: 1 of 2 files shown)


src/core/Akka.Remote.Tests.MultiNode/RemoteDeliverySpec.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        public class Postman : UntypedActor
+        {
+            protected override void OnReceive(object message)
+            {
+                var letter = message as Letter;
+                if (letter != null)
+                {
+                    letter.Route[0].Tell(new Letter(letter.N, letter.Route.Skip(1).ToList()));
```
Aaronontheweb: technically `letter.Route.Skip(1)` will only work for a list of size 2 - you should call `Skip(N-1)`

Aaronontheweb: Oh gosh, nevermind! I misread the original Scala source. Your version is just fine. This is the REST of the array MINUS the head. Not the last element in the array. My bad :p

src/core/Akka.Remote.Tests.MultiNode/RemoteDeliverySpec.cs
```diff
@@ -0,0 +1,132 @@
...
+        protected override int InitialParticipantsValueFactory
+        {
+            get
+            {
+                return Roles.Count;
+            }
+        }
+
+        [MultiNodeFact]
+        public void RemoteDeliverSpecs()
```
Aaronontheweb: Rename this to `Remoting_with_TCP_must_not_drop_messages_under_normal_circumstances`

Aaronontheweb:
@GraemeBradbury ok, finished my review - one very small change (renaming the spec) but that's it.

PR #1271, opened 2015-08-20
Added BuildRelease and CleanTests targets as dependency for MultiNodeTests target

schatekar:
`MultiNodeTests` target should have dependency on the following two targets in order to build the solution before running the tests

- BuildRelease
- CleanTests


schatekar:
@Aaronontheweb This is the PR for the `MultiNodeTests` target. There is no issue number for this.

Aaronontheweb:
Thanks @schatekar - I appreciate it!

Issue #1235, opened 2015-08-11
NullReferenceException in Akka.Persistence.Serialization.MessageSerializer.PersistentPayloadToProto


LukeTillman:
I'm trying to run tests after Akka.Persistence.Cassandra was upgraded to use Akka 1.0.4 and I'm consistently getting a NRE like this:

```
System.NullReferenceException was unhandled by user code
  HResult=-2147467261
  Message=Object reference not set to an instance of an object.
  Source=Akka.Persistence
  StackTrace:
       at Akka.Persistence.Serialization.MessageSerializer.PersistentPayloadToProto(Object payload)
       at Akka.Persistence.Serialization.MessageSerializer.PersistentToProto(IPersistentRepresentation p)
       at Akka.Persistence.Serialization.MessageSerializer.ToBinary(Object obj)
       at Akka.Persistence.Cassandra.Journal.CassandraJournal.Serialize(IPersistentRepresentation message) in d:\DataStax\Akka.Persistence.Cassandra\src\Akka.Persistence.Cassandra\Journal\CassandraJournal.cs:line 361
       at Akka.Persistence.Cassandra.Journal.CassandraJournal.WriteConfigurationValue[T](String key, T value) in d:\DataStax\Akka.Persistence.Cassandra\src\Akka.Persistence.Cassandra\Journal\CassandraJournal.cs:line 349
       at Akka.Persistence.Cassandra.Journal.CassandraJournal.PreStart() in d:\DataStax\Akka.Persistence.Cassandra\src\Akka.Persistence.Cassandra\Journal\CassandraJournal.cs:line 85
       at Akka.Actor.ActorBase.AroundPreStart()
       at Akka.Actor.ActorCell.<>c__DisplayClass4c.<Create>b__4b()
       at Akka.Actor.ActorCell.UseThreadContext(Action action)
       at Akka.Actor.ActorCell.Create(Exception failure)
  InnerException: 
```

Looks like it's coming from this line here:

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Persistence/Serialization/MessageSerializer.cs#L140

When `TransportInformation` is null. I'm guessing https://github.com/akkadotnet/akka.net/commit/fb03e9d687fabd2bd61ce3a341ce9c57460d1d38 may have introduced the issue. Not sure if fixing this is as simple as just adding a null check on `TransportInformation` before calling the `Address` property or if that property shouldn't be null in this case indicating some other bug.

[...]

PR #1273, opened 2015-08-22
Fixes the NullReferenceException in #1235 

trbngr:
...and appears to adhere to the practice of including an address with the serialized binary.


rogeralsing:
Thanks!

PR #1276, opened 2015-08-24
serialization fixes #1275 

### LINKED ISSUE #1275: Serializing SupervisorStrategy containing LocalOnlyDecider silently ignores the decider
When serializing SupervisorStrategy's with a LocalOnly decider, Json.NET simply ignores the decider and continues with its business.
That might seem fair, but it will result in a supervisorstrategy that uses the default strategy when deserializing.
So if the sender did not use the default strategy, the receiving end will have a strategy with an unintended behavor.

The simple solution is to make the SupervisorStrategy throw an exception if the decider is a LocalOnly when translating to the serialization surrogate.
(only strategies with a deployable decider should be possible to serialize).

There are a couple of serializer tests that does this, they create different router pools w/o specifying a supervisor strategy, and thus gets the default strategy that uses a local only decider..

rogeralsing:
Fixes #1275 

This PR replaces the default deciders in SupervisorStrategy and Pool with deployable deciders with the same semantics.
There is also a fix that prevents supervisorstrategies with local only deciders to be serialized, current code just ignores the local decider and sends an incomplete message

This could have some impact on user code, that is, if the user have used local only deciders in a remote deploy scenario, there will now be an exception thrown.
One could argue that this is a breaking change, but it is really a bug in user code if a local only decider is used in remote cases.

### DIFF
--- a/src/core/Akka/Actor/SupervisorStrategy.cs
+++ b/src/core/Akka/Actor/SupervisorStrategy.cs
@@ -80,17 +80,10 @@ namespace Akka.Actor
         /// </summary>
         /// <param name="exception">The exception.</param>
         /// <returns>Directive.</returns>
-        public static Directive DefaultDecider(Exception exception)
-        {
-            if (exception is ActorInitializationException)
-                return Directive.Stop;
-            if (exception is ActorKilledException)
-                return Directive.Stop;
-            if (exception is DeathPactException)
-                return Directive.Stop;
-
-            return Directive.Restart;
-        }
+        public static IDecider DefaultDecider = Decider.From(Directive.Restart,
+            Directive.Stop.When<ActorInitializationException>(),
+            Directive.Stop.When<ActorKilledException>(),
+            Directive.Stop.When<DeathPactException>());
 
         /// <summary>
         ///     Restarts the child.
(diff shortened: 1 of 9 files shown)


src/core/Akka/Actor/SupervisorStrategy.cs
```diff
@@ -364,6 +357,8 @@ public ISurrogated FromSurrogate(ActorSystem system)
 
         public override ISurrogate ToSurrogate(ActorSystem system)
         {
+            if (Decider is LocalOnlyDecider)
+                throw new NotSupportedException("Can not serialize LocalOnlyDecider");
```
rogeralsing: This line prevents objects referring to a local only decider from being serialized.
Json.NET silently ignored this and passed the message without the decider, potentially leading to incorrect error handling on the remote system.

src/core/Akka/Actor/SupervisorStrategy.cs
```diff
@@ -80,17 +80,10 @@ public bool HandleFailure(ActorCell actorCell, Exception cause, ChildRestartStat
...
-                return Directive.Stop;
-            if (exception is DeathPactException)
-                return Directive.Stop;
-
-            return Directive.Restart;
-        }
+        public static IDecider DefaultDecider = Decider.From(Directive.Restart,
+            Directive.Stop.When<ActorInitializationException>(),
+            Directive.Stop.When<ActorKilledException>(),
+            Directive.Stop.When<DeathPactException>());
```
rogeralsing: The old default decider have been replaced with a deployable decider with the exact same behavior.
I consider this a better default value as supervisorstrategies and pool routers are deployable when created using their small constructors.

Horusiath:
Looks good for me :+1:

PR #1278, opened 2015-08-29
F# API - added support to override actor lifetime

tjaskula:
Hi,

I thought it would be useful to have a possibility to override PreStart, PostStop, PreRestart, PostRestart lifetime methods when using F# API and actor computation expression.
I didn't change any existing api call.

### DIFF
--- a/src/core/Akka.FSharp.Tests/ApiTests.fs
+++ b/src/core/Akka.FSharp.Tests/ApiTests.fs
@@ -72,3 +72,81 @@ let ``can serialize and deserialize discriminated unions over remote nodes`` () 
     response
     |> equals msg
 
+[<Fact>]
+let ``can override PreStart method when starting actor with computation expression`` () =
+    
+    let preStartCalled = ref false
+    let preStart = Some(fun (baseFn : unit -> unit) -> preStartCalled := true)
+    
+    use system = System.create "testSystem" (Configuration.load())
+    let actor = 
+        spawnOvrd system "actor" 
+        <| actorOf2 (fun mailbox msg ->
+                mailbox.Sender() <! msg)
+        <| {defOvrd with PreStart = preStart}
+    let response = actor <? "msg" |> Async.RunSynchronously
+    (!preStartCalled, response) |> equals (true, "msg")
+
+[<Fact>]
+let ``can override PostStop methods when starting actor with computation expression`` () =
+    
+    let postStopCalled = ref false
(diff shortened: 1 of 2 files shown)


Horusiath:
I've created an example of how to solve problem with actor lifecycle events [here](https://gist.github.com/Horusiath/ba2e5b33dc4377a0f018) using existing tools. I'm not sure if we need to introduce more complexity to the API in case, where we can handle it this way.

rogeralsing:
I vote for closing as there is an existing idiomatic aporoach already.

tjaskula:
There is no an idiomatic approach for PreRestart and PostRestart method. Also I don't think API is more complex as existing functions are not changed, but more complete. If I need to override all the lifecycle I have to go away from computation expressions and this is frustrating. But it's up to you. I'm closing the issue

Horusiath:
@tjaskula there was few more proposed solutions for this:

1. Make `mailbox.Defer` and function accepting `exn option` as input argument. If arg was `None` it means a PostStop event, otherwise it's `PreRestart` event with exception as an argument.
2. Create set of dedicated message types like `PreStart`, `PreRestart`, `PostRestart`, `PostStop` and invoke actor behavior function with them, when a corresponding actor lifecycle event occurs. This idea is borrowed from cannonical akka-typed module.

However, before we introduce any change, we need justification that current design makes real use cases impossible to complete.

rogeralsing:
Oh, the event idea is neat. and so obvious once you read it :)

tjaskula:
@Horusiath This makes sense. I understand that you have to justify the wider usage. That's why I though that it would better play out for me if I had just an extension to the existing API. So I made it and it works well for what I need : https://github.com/tjaskula/akka.net-fsharp.extensions/blob/master/src/ComposeIt.Akka.FSharp.Extensions/FsApi.Extensions.fs

And it's specific to my need. I can spawn actor with overridden functions and you don't need to bend the API for this specific need. It's a win win scenario:)

tjaskula:
Anyway. I have a question. Do you have any roadmap or discussion thread of what you plan to put in the API or something. It may help to avoid to bother you with no needed PRs. Just in a case

Horusiath:
There are milestone tasks and roadmap for Akka.NET core, but no such thing on for it's F# API - I have a bunch of improvement tasks in [Akkling](https://github.com/Horusiath/Akkling/issues) - my fork of Akka.FSharp - where I'm testing new ideas and improvements. Most of the discussions is made on our [gitter channel](https://gitter.im/akkadotnet/akka.net).

Issue #1279, opened 2015-08-30
Remote deployed actors fail with super unintuitive exception when remote system does not reference actor types


rogeralsing:
Found this from an issue in https://github.com/IdleLands/IdleLandsRedux/tree/akka-refactor today.

The scenario was that the remote system did not reference the assembly holding the actor types.
Which made the `DaemonMsgCreateSerializer` create a `Props` object with no type, which in turn resulted in `DefaultProducer` being used, which in turn resulted in a exception, far far away from the actual problem.
There should be two fixes here IMO.
1. Make `Props` throw if the type passed is `null`
2. Make  `DaemonMsgCreateSerializer` throw if `var clazz = Type.GetType(proto.Props.Clazz);` results in `null`

In the latter case, if the proto.Props.Class holds a value, and `clazz` is null, then clearly there is a missing reference in the remote system

rdavisau:
Hi @rogeralsing, I can look at this one.

I think we can't perform the null check in the [main constructor](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/Props.cs#L238:L244), because the [parameterless constructor](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/Props.cs#L165:L168) passes `null` in the type parameter; I guess it's there to support deserialization and maybe some subclasses?
```csharp
protected Props() 
    : this(defaultDeploy, null, noArgs)
{
}
```
In that case, I would be adding the check to each of the other ctors and static `create`-style methods. Does this sound reasonable?

rogeralsing:
Yes that sounds like the correct way to go about it.

PR #1286, opened 2015-08-31
Better exception message when a remote actor deploy fails because of missing reference #1279

rdavisau:
This addresses #1279. 

As suggested, I added checks to the `Props` constructors to handle cases where the `type` constructor parameter was `null`. I added a single test for the six methods changed (if preferred, I can split them out). Finally, I added the check to `DaemonMsgCreateSerializer` - when `GetType` returns `null` for the actor type name, an exception is thrown informing the user that the type should be referenced.

### DIFF
--- a/src/core/Akka.Remote/Serialization/DaemonMsgCreateSerializer.cs
+++ b/src/core/Akka.Remote/Serialization/DaemonMsgCreateSerializer.cs
@@ -113,7 +113,23 @@ namespace Akka.Remote.Serialization
         public override object FromBinary(byte[] bytes, Type type)
         {
             var proto = DaemonMsgCreateData.ParseFrom(bytes);
-            var clazz = Type.GetType(proto.Props.Clazz);
+            Type clazz; 
+
+            try
+            {
+                clazz = Type.GetType(proto.Props.Clazz, true);
+            }
+            catch (TypeLoadException ex)
+            {
+                var msg = string.Format(
+                       "Could not find type '{0}' on the remote system. " +
+                       "Ensure that the remote system has an assembly that contains the type {0} in its assembly search path", 
+                       proto.Props.Clazz);
+
+
+                throw new TypeLoadException(msg, ex);
+            }
+
             var args = GetArgs(proto);
             var props = new Props(GetDeploy(proto.Props.Deploy), clazz, args);
             return new DaemonMsgCreate(
--- a/src/contrib/testkits/Akka.TestKit.Xunit2/Akka.TestKit.Xunit2.csproj
+++ b/src/contrib/testkits/Akka.TestKit.Xunit2/Akka.TestKit.Xunit2.csproj
@@ -53,6 +53,7 @@
     <Compile Include="Internals\AkkaAssertEqualityComparer.cs" />
     <Compile Include="Internals\AkkaAssertEqualityComparerAdapter.cs" />
     <Compile Include="Internals\AkkaEqualException.cs" />
+    <Compile Include="Internals\Loggers.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="TestKit.cs" />
     <Compile Include="XunitAssertions.cs" />
--- a/src/contrib/testkits/Akka.TestKit.Xunit2/Internals/Loggers.cs
+++ b/src/contrib/testkits/Akka.TestKit.Xunit2/Internals/Loggers.cs
@@ -1,0 +1,21 @@
+﻿using Akka.Actor;
+using Akka.Event;
+using Xunit.Abstractions;
+
+namespace Akka.TestKit.Xunit2.Internals
+{
+    public class TestOutputLogger : ReceiveActor
+    {
+        public TestOutputLogger(ITestOutputHelper output)
+        {
+            Receive<Debug>(e => output.WriteLine(e.ToString()));
+            Receive<Info>(e => output.WriteLine(e.ToString()));
+            Receive<Warning>(e => output.WriteLine(e.ToString()));
+            Receive<Error>(e => output.WriteLine(e.ToString()));
+            Receive<InitializeLogger>(e =>
+            {
+                e.LoggingBus.Subscribe(Self, typeof (LogEvent));
+            });
+        }
+    }
+}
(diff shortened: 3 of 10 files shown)


Aaronontheweb:
@rdavisau could you please squash these commits per `contributing.md`?

rdavisau:
@Aaronontheweb that response time! 
I've squashed them back up. :+1:

src/core/Akka.Remote/Serialization/DaemonMsgCreateSerializer.cs
```diff
@@ -114,6 +114,17 @@ public override object FromBinary(byte[] bytes, Type type)
...
             var clazz = Type.GetType(proto.Props.Clazz);
+
+            if (clazz == null)
+            {
+                var msg =
+                    string.Format(
+                        "Could not find '{0}' in loaded assemblies. If this is a remote/cluster deploy scenario, " +
+                        "ensure the remote system has referenced the assembly containing this type.", proto.Props.Clazz);
+                    
+                throw new ActorInitializationException(msg);
```
rogeralsing: Two minor issues here:
1. `ActorInitializationException` feels like the wrong exception, at this stage we are just materializing the stuff needed from the wire.

2. "If this is a remote/cluster deploy" that is a given.  as the DaemonMsgCreate is only used for this very scenario. 
And "has referenced the assembly containing this type" is somewhat too strict, It would work if the assembly is available in the GAC or if the assembly simply resides in one of the directories where the host process loads assemblies.
So a more generic description of the problem would be good IMO.

rogeralsing: There is an overload for `Type.GetType(typename, throwOnError)` that throws a `TypeLoadException` should we try catch this and use that exception as an inner exception for whatever we throw?
Or, would that exception on itself be enough to figure out the problem?

cc @akkadotnet/developers

rdavisau: Good point - the message from `TypeLoadException` is as follows:
```
TypeLoadException: Could not load type 'TypeName' from assembly 
'AssemblyName, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
```
For someone familiar with the exception, it will be clear that the remote system does not have access either to the containing assembly or one of its dependencies. For someone unfamilar, I still think it would be enough to get started with. So I am in favour and can make the change if you're happy with that. 

I guess the case for wrapping it would be that it could make it immediately obvious to the user that a known specific issue with a known specific solution (review the referenced/available assemblies) had occurred. But going down that track can get you into trouble if later on another set of circumstances can trigger the same exception but with a different cause/resolution - now the error message is too specific to the old problem and inaccurate for this new one.

rogeralsing: I made an SO post about it http://stackoverflow.com/questions/32315574/throw-what-on-type-gettype-null

We do have the same issue on other places in the framework where we manually check for null and then throw.
And possibly some places where we have missed null checks too.

It would be nice if we can come up with an approach that we all agree on :)

sean-gilliam: I think ```TypeLoadException``` is perfectly fine exception as is. IMO, most .NET developers would know exactly what was going on if they ran across that exception.

But if you wanted it to be more user friendly, maybe create a custom exception like RemoteTypeLoadException and then do something like

```csharp
Type clazz;
try
{
    clazz = Type.GetType(proto.Props.Clazz, true);
}
catch(TypeLoadException tle)
{
    var msg = string.Format(
        "Could not find type '{0}' on the remote system. Ensure that the remote system has an assembly that contains {0} in its assembly search path", proto.Props.Clazz);

    throw new RemoteTypeLoadException(msg, tle);
}
```

or have our custom RemoteTypeLoadException already have msg's verbiage and just plug in the {0} value as a property on that exception.

```csharp
throw new RemoteTypeLoadException(tle) { Type = proto.Props.Clazz };
```

rogeralsing: Yes, you're right, we shouldnt really clutter things up by introducing more and more exception types.
The TypeLoadException is the thing that one should look for in this scenario I guess, so lets just go with `Type.GetType(...,true)` then.

Aaronontheweb: I like @sean-gilliam's idea

rdavisau: I'm good with either approach, so will wait for consensus on this :palm_tree:

Horusiath: :+1: for @sean-gilliam idea... but instead of try-catching on `Type.GetType(..., true)` it's faster just to do `Type.GetType(...)` and throw on null result ;)

rogeralsing: If we just check for null, we will not get the real inner type load exception.
I figure it would be good to actually get that specific exception, even if it is just an inner exception for the user, as that is the "real" cause of the problem.

Aaronontheweb: > If we just check for null, we will not get the real inner type load exception.
I figure it would be good to actually get that specific exception, even if it is just an inner exception for the user, as that is the "real" cause of the problem.

I vote to do this:

```
Type.GetType(..., true)
```

And throw a `TypeLoadException`, rather than add a custom exception. Reduce clutter, get the specific details, and we can customize the error message to indicate that this was an issue with remote deployment. Agreed?

rogeralsing: :+1:

sean-gilliam: :+1:

rdavisau:
I've made the changes discussed in the comment thread on `DaemonMsgCreateSerializer`; but it looks like another akka commit has snuck into the PR too. I did attempt to pull in changes from upstream following the commands on `CONTRIBUTING`, any hints on how to get rid of the extra commit?

Aaronontheweb:
@rdavisau that other commit is already in the history for this repo, so you wouldn't be able to squash it I suppose. TBH, not sure - that's beyond my Git skill :p 

But the PR looks good. I'll merge it in.

PR #1287, opened 2015-08-31
Xunit2 output logger for parallel TestKit runs

Horusiath:
This is an enhancement for TestKit using Xunit2, that allows you to see your logging data in output of xunit test runner. Even when you're running tests in parallel.

Xunit2 defines a dedicated interface `ITestOutputHelper`, that will be automatically injected to each test class constructor by the test runner itself. To allow this you must only expose it:

```csharp
class MyActorSpec : Akka.TestKit.Xunit2.TestKit 
{
    public MyActorSpec(ITestOutputHelper output) : base(output: output) { }
}
```

`AkkaSpec` used for internal testing also has this *output* param exposed in it's constructor. I think we could expose it in all of internal test projects, but this should be done in separate PR. This will help with getting log info while executing specs.

### DIFF
--- a/src/core/Akka.Tests.Shared.Internals/AkkaSpec.cs
+++ b/src/core/Akka.Tests.Shared.Internals/AkkaSpec.cs
@@ -12,6 +12,7 @@ using System.Text.RegularExpressions;
 using System.Threading;
 using Akka.Configuration;
 using Xunit;
+using Xunit.Abstractions;
 using Xunit.Sdk;
 
 // ReSharper disable once CheckNamespace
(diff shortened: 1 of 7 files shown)


src/core/Akka.Tests.Shared.Internals/AkkaSpec.cs
```diff
@@ -39,13 +40,13 @@ public abstract class AkkaSpec : Xunit2.TestKit    //AkkaSpec is not part of Tes
 
         private static int _systemNumber = 0;
 
-        public AkkaSpec(string config)
-            : this(ConfigurationFactory.ParseString(config).WithFallback(_akkaSpecConfig))
+        public AkkaSpec(string config, ITestOutputHelper output = null)
```
Aaronontheweb: Do we really need to do this? Can't we just load this XUnit logger from HOCON config by default inside the XUnit2 testkit?

Horusiath: `ITestOutputHelper` can be added only by the test runner and only as a constructor param for test class. It's out of our control.

Aaronontheweb: Ok, that makes sense - and it looks like this is only for XUnit2 TestKit and AkkaSpec. I originally misread what you wrote and thought that this was for the TestKit base class itself, which would be a much larger change obviously.

I'm good with this change. cc @rogeralsing ?

rogeralsing: Me too, having log outputs for the tests will help a lot.

PR #1295, opened 2015-09-06
Semantc Serilog

rogeralsing:
This PR contains an alternative Serilog implementation.

The standard serilog implementation does not deal well with log templates of the `string.Format` kind.
That is. if there is a logging statement like so: `log.Debug("Bar {0}, Foo {1}",bar, foo)` , this will be logged as property `0` and property `1`.
Which gives very little value for log analyzers.

The logger in this PR, deals with both the specialized Serilog format, _and_ the old `string.Format` style log statements.
This is extremely useful since Akka.NET internally uses the `string.Format` approach, so if you want to analyze what is going on inside Akka.NET itself, this logger lets you do that.

The SemanticSerilogger here solves this by checking if a log statement use the `string.Format` style template, if so, it will extract any textual prefix before any property placeholder, like so:
`"Connecting to {0} using protocol {1}"` -> `"Connecting to {ConnectingTo} using protocol {UsingProtocol}"`

This is of-course a compromise, using semantic logging internally in Akka.NET would have been a better way to solve this.
But as a non-intrusive addition to the contrib packages, this works surprisingly well.

For example, the dashboard below is generated only from the built in logging statements inside Akka.NET(Remote+Cluster) using this logger
akkadashboard

The reason why I did not make this an addition to the default Serilogger package is that this also takes dependency on Sprache parsing library.
There are also some behaviors in the other Serilog package that I don't want to touch due to backwards compat issues.

### DIFF
--- a/src/Akka.sln
+++ b/src/Akka.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+VisualStudioVersion = 12.0.40629.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Examples", "Examples", "{69279534-1DBA-4115-BF8B-03F77FC8125E}"
 EndProject
--- a/src/contrib/loggers/Akka.Logger.Serilog/Akka.Logger.Serilog.csproj
+++ b/src/contrib/loggers/Akka.Logger.Serilog/Akka.Logger.Serilog.csproj
@@ -49,6 +49,8 @@
     </Compile>
     <Compile Include="Event\Serilog\SerilogLogMessageFormatter.cs" />
     <Compile Include="MessageTemplateCache.cs" />
+    <Compile Include="SemanticAdapter\FormatPart.cs" />
+    <Compile Include="SemanticAdapter\TemplateTransform.cs" />
     <Compile Include="SerilogLogMessageFormatter.cs" />
     <Compile Include="SerilogLogger.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
--- a/src/contrib/loggers/Akka.Logger.Serilog/SemanticAdapter/FormatPart.cs
+++ b/src/contrib/loggers/Akka.Logger.Serilog/SemanticAdapter/FormatPart.cs
@@ -1,0 +1,16 @@
+﻿
+namespace Akka.Logger.Serilog.SemanticAdapter
+{
+    internal class FormatPart
+    {
+        public FormatPart(string text, int index)
+        {
+            Text = text;
+            Index = index;
+        }
+
+        public string Text { get;private set; }
+
+        public int Index { get;private set; }
+    }
+}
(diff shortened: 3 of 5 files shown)


src/contrib/loggers/Akka.Logger.SemanticSerilog/StringFormat.cs
```diff
@@ -0,0 +1,11 @@
+namespace Akka.Logger.SemanticSerilog
+{
+    internal class StringFormat
```
Horusiath: Maybe struct is a better choice?

src/contrib/loggers/Akka.Logger.SemanticSerilog/LogParser.cs
```diff
@@ -0,0 +1,24 @@
+using System.Collections.Generic;
+using Sprache;
+
+namespace Akka.Logger.SemanticSerilog
+{
+    internal static class LogParser
+    {
+        public static Parser<StringFormat> StringFormat =
```
Horusiath: Does the Sprache is really a justified dependency in this case? Doesn't this feel like taking a canon to kill a mosquito?

src/contrib/loggers/Akka.Logger.SemanticSerilog/StringFormat.cs
rogeralsing: Maybe, the templates are however cached, so it is a one time translation, thus the parsing perf/resource usage is largely irrelevant

src/contrib/loggers/Akka.Logger.SemanticSerilog/LogParser.cs
rogeralsing: Removed sprache dependency, so maybe this could go in the standard serilog package as just an alternative logger class

Mpdreamz:
I quite like the guestimate of the semantic names! I know this feature would be very be useful in the https://github.com/serilog/serilog-sinks-elasticsearch as well.

Since this no longer relies on sprache, could be something very useful to port back to serilog as an option? cc @nblumhardt @mivano

nblumhardt:
Definitely watching with interest!

I'd be hesitant to bake this into Serilog's core because the edge cases and assumptions would need to be baked in to, but given it's generally going to be applied at integration points where older code is being adapted it may be able to be packaged up as a library function and used in adapters?

There's also a question of event quality even in the Akka.NET case - laborious as it would be, sifting through the code and replacing `{0}`s with `{Name}`s will yield better results. Having an adapter from Serilog-style to indexed format strings is also much more straightforward and doesn't need any heuristics (the holes are always numbers `0..n`).

rogeralsing:
I have merged this implementaton with the original `SerilogLogger`, logging anything that contains serilog property names continue to work as previously, no change there.
But if an old string.format template is passed to serilog, it will apply the guesstimate for property names, as this is far better than just logging `{0}` for every property.

adamchester:
Would it be that difficult to switch to {name} format internally inside akka? Since it's possible to [reuse the Serlig code](https://github.com/adamchester/MessageTemplates)...

rogeralsing:
@adamchester it wouldnt be techically difficult, its more a question of what effects it has to existing users, performance etc.

Adding a bonus featrue to one of the loggers is a much less intrusive action than changng the internals of Akka.NET.
(I personally think we should change. but it needs to be decided first)

adamchester:
On the question of performance, I can help with that... Here's one sample

perf

Mpdreamz:
@adamchester mind opensourcing your benchmark code? I feel functionally liblog+serilog is a winning combo for library authors would be cool to see if those numbers can come down a bit.

adamchester:
This is the version that produced the above chart

https://github.com/adamchester/MessageTemplates/blob/8bea72188a987dacbb6697c4aae11b996febc698/tests/MessageTemplates.PerfTests/PerfTests.fsx

Feedback welcome...

Regards,
Adam Chester

Horusiath:
Is this ready to merge?

Issue #1283, opened 2015-08-31
Replace string concatenation log statements with format templates


rogeralsing:
Currently, there are a few cases where we do things like:

`_log.Debug("Foo" + bar + "Baz" + yo);`
This is however incompatible with loggers that use special formatting. (e.g. Serilog)

This should be replaced with:
`_log.Debug("Foo {0} Baz {1}",bar,yo);`

That way, loggers can apply whatever logic they need to the formatting template.

alex-kondrashov:
Can I help on this one? seems to be easy to start with.

Dim0N22:
I've found only [one line](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/Inbox.Actor.cs#L52) with string concatenation in entire solution.
Everything else looks fine to me.

rogeralsing:
Ill see what sources i find, i see some log posts coming in with no message templates in elastic search

rogeralsing:
@Dim0N22 did you also check the other log levels or only Debug?

Dim0N22:
@rogeralsing I've checked all log levels. 
Double checked it with regexp, nothing found. Maybe I missed something?

alex-kondrashov:
I've found couple of places [like this](https://github.com/akkadotnet/akka.net/blob/928ff2b3b23457ea1e4e1d9f0819fca945506a1b/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs#L606)
but I'm not sure if it's relevant.

rogeralsing:
@alex-kondrashov that is absolutely relevant. all dynamic content should be injected through the template. otherwise we can not extract the value using semantic logging

rogeralsing:
Thanks for all the help guys!

PR #1297, opened 2015-09-07
Replaced string concatenation with format templates in log statements

alex-kondrashov:
These fixes relate to [this issue #1283](https://github.com/akkadotnet/akka.net/issues/1283).

Basically I corrected all the places that I've found. The main idea behind fixes is
to change log statements from this 
```c#
_log.Debug("Foo" + bar + "Baz" + yo)
```
to this
```c#
_log.Debug("Foo {0} Baz {1}",bar,yo);
```
I squashed all the commits into one.

### DIFF
--- a/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
+++ b/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
@@ -603,7 +603,7 @@ namespace Akka.Remote.TestKit
                         // controller node is finished/exited before r.addr is run
                         // on the other nodes
                         var unresolved = "akka://unresolved-replacement-" + r.Role.Name;
-                        Log.Warning(unresolved + " due to: " + e.ToString());
+		         Log.Warning(unresolved + " due to: {0}", e.ToString());
                         replaceWith = unresolved;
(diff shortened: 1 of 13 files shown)


src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -603,7 +603,7 @@ protected void InjectDeployments(ActorSystem system, RoleName role)
                         // controller node is finished/exited before r.addr is run
                         // on the other nodes
                         var unresolved = "akka://unresolved-replacement-" + r.Role.Name;
-                        Log.Warning(unresolved + " due to: " + e.ToString());
+		         Log.Warning(unresolved + " due to: {0}", e.ToString());
```
rogeralsing: Could you remove the .ToString() part on the `e`? normal loggers will do the tostringing by themselves and serilog needs the entire object to write the exception details

Aaronontheweb:
Could you rebase to the latest changes on `dev`? Should fix the problem with this failing unit test.

rogeralsing:
Merging this

PR #1274, opened 2015-08-22
Add remote random spec

KadekM:
#1257

PR #1299, opened 2015-09-08
do-not-merge

KadekM:
... trying to find problem

### DIFF
--- a/src/core/Akka.Remote.Tests.MultiNode/LookupRemoteActorMultiNetSpec.cs
+++ b/src/core/Akka.Remote.Tests.MultiNode/LookupRemoteActorMultiNetSpec.cs
@@ -62,7 +62,7 @@
             }
         }
 
-        [MultiNodeFact]
+        //[MultiNodeFact]
         public void LookupRemoteActorSpecs()
         {
             RunOn(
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
think it could be the `Resizer` that's causing the problem?

KadekM:
Yes - i have done some poking around it. SomeActor receives 'hit' resize is initiated, at first is resized by one (all's good), but then another and another resizing starts to happen...

KadekM:
(another resizimgs happen before 'hits' are sent... I will try play with for some more later in the day...

Aaronontheweb:
We've had issues with the Resizer before - it's somewhat imprecise. I wouldn't be surprised if that was the culprit.

rogeralsing:
Should this also be closed as the Resizer problem was found and fixed?

Aaronontheweb:
@rogeralsing senor @KadekM should re-integrate that change into this PR and uncomment out the failing test and see if that resolves the issue.

KadekM:
@Aaronontheweb: the test is on dev brach - it was never commented out. Issue was patched before PR which was commenting it out was merged (#1302), thus those test pass (whole *RoundRobinSpec) and this issue should by resolved.

KadekM:
Do not close yet please... there still seems to be something wrong (#1274) spec just failed.

Aaronontheweb:
No worries - we'll keep it open

PR #1302, opened 2015-09-09
Ignore one test as per #1301

KadekM:
#1301

PR #1304, opened 2015-09-09
Fix for Issue 1301

### LINKED ISSUE #1301: Imprecise Resizer
There seems to be issue with Resizer, not resizing properly (resizing too much?).

It can be observed with on this test.
https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote.Tests.MultiNode/RemoteRoundRobinSpec.cs#L169

which is failing our builds, hence I will comment it out until we resolve this issue.

KadekM:
Should fix #1301 

If build server agrees then #1302 does should not be merged.


rogeralsing:
:+1:

Issue #1196, opened 2015-07-31
Multi node test runner visualizer


Horusiath:
This was already mentioned during the contributors meeting. We noticed that it's quite hard to track/debug the behavior of cluster spec based only on console/file output. For this reason, we wanted to create an API for multi node test runner, that could be used to visualize events occurring on each node in form of a time line (one per each node) with some convenient player-like functionality.

It's hard to describe this in words, so let's use ASCII graphic to the rescue (`#` marks logs/events):

    Node 1 |--#-----#----------------------#--------->| (stop)
    Node 2 |------#------------#---#------------#---->| (stop)
    Node 3 |---#---------#-------#----->X (crash)

Aaronontheweb:
To echo the idea I put down in Gitter for this, the MultiNodeTestRunner.Shared library includes a number of sinks for structuring the per-node log output into a time-ordered set of per-test / per-node structures. The entirety of this output gets saved to a .JSON file today.

We could write a HTML5 / JS static web app that could render this JSON file into a timeline for each test. Would that work?

Horusiath:
I think so. The major concern is that it should be easy to use and IMO it shouldn't need any additional installation or configuration steps. Maybe a light HTTP server that can be embedded in multinode test runner and open with command line arg?

rogeralsing:
The data we want to display, is it a stream or are we just aiming to show the end result when everything is done?

Aaronontheweb:
@rogeralsing it'd be a lot more work to make it streaming. I was thinking just something that would load the final output file (was gets updated incrementally with each test run.) That would be a lot easier to build. Doing a streaming version of this would require websocket support and an actual server.

Aaronontheweb:
@Horusiath if we go with the static HTML approach, I can write the test output directly into the HTML and just pop open a web browser window immediately afterwards.

Aaronontheweb:
A separate UX issue with the MultiNodeTestRunner... making it easy to run JUST ONE test versus having to run the entire suite. That'd save a lot of time... Not sure how to solve that issue without writing a visual studio plugin though :p

rogeralsing:
I vote for AngularJS, would be easy to customize how the stuff is rendered and could give some nice interactivity too.

GraemeBradbury:
This looks good for our needs. http://visjs.org/examples/timeline/groups/groups.html 
Should be a case of adapting `TestRunTree` and spitting out a html file.

Aaronontheweb:
[...]

PR #1305, opened 2015-09-09
Multinode Test Visualizer

GraemeBradbury:
VisualizerPersistenTestRunStore creates a html file with a timeline
visual showing the node events.

This file sits alongside the json file and has the same name
(different file extension)

#1196


sean-gilliam:
Why include all the vis.js files in the PR? Maybe include it as CDN links in the html or reference it through a tool like bower for the dev to pull in on their own.

Just a suggestion. Otherwise looks great :+1:

GraemeBradbury:
IMO it's a strength of the Akka.Net project that I can pull, type`build build` and it all just works. So putting a dependency on nodejs just to build the test report feels wrong to me.

Thinking about using a CDN raises a couple of questions
* What to do with our css? CDN as well?
* What about offline/intermittent connections? is it a problem?

GraemeBradbury:
capture

Screenshot of visualizer in Chrome

Aaronontheweb:
Figured out why TC didn't pick up this build. The timestamp of this commit is from 9 days ago, so TC won't pick it up.

Could do the following and update the timestamp to present time? http://stackoverflow.com/questions/454734/how-can-one-change-the-timestamp-of-an-old-commit-in-git

GraemeBradbury:
@Aaronontheweb This runs consistently on local.
On build server it'll run the Cluster multinode tests but will consistently fail on the Remote multinode tests.

The build log doesn't contain any reason for failure beyond exit code 42. Which I've not been able to find in MSDN docs.

Have found mention of code 42 in reference to TeamCity not finding a file but the build logs show that both the runner and the dll are being created successfully in the locations expected.

A remote possibility could be that the exit code is 0x42 (BAD_DEV_TYPE - incorrect network resource type) which sorta makes sense considering the tests. It just seems unlikely considering Cluster builds and runs successfully.

GraemeBradbury:
The test sinks seem to be working properly

````
[08:23:42][Step 1/1] Writing test state to: D:\BuildAgent\work\49b164d63843fb4\src\core\Akka.Remote.Tests.MultiNode\bin\Release\Akka.Remote.Tests.MultiNode-635775565262939944.json
[08:23:42][Step 1/1] Writing test state to: D:\BuildAgent\work\49b164d63843fb4\src\core\Akka.Remote.Tests.MultiNode\bin\Release\Akka.Remote.Tests.MultiNode-635775565262939944.html
```

Looks like its the RemoteRoundRobinSpec being consistently racy
```
[08:24:45][Step 1/1] [RUNNER][8:24 AM]: Results for Akka.Remote.Tests.MultiNode.RemoteRoundRobinMultiNode.RemoteRoundRobinSpecs
[08:24:45][Step 1/1] [RUNNER][8:24 AM]: Start time: 9/11/2015 8:22:06 AM
[08:24:45][Step 1/1] [RUNNER][8:24 AM]:  --> Node 1: PASS [00:00:27.6249069 elapsed]
[08:24:45][Step 1/1] [RUNNER][8:24 AM]:  --> Node 2: FAIL [00:01:02.6693675 elapsed]
[08:24:45][Step 1/1] [RUNNER][8:24 AM]:  --> Node 3: FAIL [00:00:42.6693895 elapsed]
[08:24:45][Step 1/1] [RUNNER][8:24 AM]:  --> Node 4: FAIL [00:00:27.5746810 elapsed]
```

Aaronontheweb:
@GraemeBradbury I'm going to see about adding the multi-node test visualizer as an artifact of each build, so it can be viewed as part of the TC builds

GraemeBradbury:
@Aaronontheweb Made some tweaks to make the visualizer work better from TC.
Basically took @sean-gilliam 's advice and used the CDN for VisJS with no other references. So each file is standalone.

rogeralsing:
Do we need to have the full namespace written out on the events? 
`Akka.Cluster.Tests.MultiNode.LeaderLeavingSpecConfg.` wouldnt it be possible to somehow reduce the number of times we see the same prefix?
Add legends to charts or add headers or whatever.

Love it all but the event names are a mouthfull

Issue #1291, opened 2015-09-03
Multiple active result sets exception


sadprofessor:
we occasionally see an exception coming from SQL akka persistence moaning about multiple active result sets not being supported. Might it need enabling?

	`System.Data.dll!System.Data.SqlClient.SqlInternalConnectionTds.ValidateConnectionForExecute(System.Data.SqlClient.SqlCommand command)	Unknown
 	System.Data.dll!System.Data.SqlClient.SqlConnection.ValidateConnectionForExecute(string method, System.Data.SqlClient.SqlCommand command)	Unknown
 	System.Data.dll!System.Data.SqlClient.SqlCommand.ValidateCommand(string method, bool async)	Unknown
 	System.Data.dll!System.Data.SqlClient.SqlCommand.RunExecuteReader(System.Data.CommandBehavior cmdBehavior, System.Data.SqlClient.RunBehavior runBehavior, bool returnStream, string method, System.Threading.Tasks.TaskCompletionSource<object> completion, int timeout, out System.Threading.Tasks.Task task, bool asyncWrite)	Unknown
 	System.Data.dll!System.Data.SqlClient.SqlCommand.BeginExecuteReaderInternal(System.Data.CommandBehavior behavior, System.AsyncCallback callback, object stateObject, int timeout, bool asyncWrite)	Unknown
 	System.Data.dll!System.Data.SqlClient.SqlCommand.BeginExecuteReaderAsync(System.Data.CommandBehavior behavior, System.AsyncCallback callback, object stateObject)	Unknown
 	mscorlib.dll!System.Threading.Tasks.TaskFactory<System.Data.SqlClient.SqlDataReader>.FromAsyncImpl<System.Data.CommandBehavior>(System.Func<System.Data.CommandBehavior,System.AsyncCallback,object,System.IAsyncResult> beginMethod, System.Func<System.IAsyncResult,System.Data.SqlClient.SqlDataReader> endFunction, System.Action<System.IAsyncResult> endAction, System.Data.CommandBehavior arg1, object state, System.Threading.Tasks.TaskCreationOptions creationOptions)	Unknown
[...]

PR #1306, opened 2015-09-10
Query mechanism for Akka.Persistence SQL-based journals

Horusiath:
**WARNING**: this PR contains breaking changes for SQL-based plugins. After merging, changes should be applied in dependent sql plugins. Querying mechanism should also be documented.

### Changes regarding Akka.Persistence

- All `Akka.Persistence` messages from now implement structural equality and meaningfull `ToString` messages.
- Some of the message constructors have been hardened with requirement for not-null arguments.
- All Persistence specs use xUnit2 `ITestOutputHelper` for spec output logging.
- More meaningfull error messages, when journal/snapshot plugin couldn't be loaded because dependent module it's not linked by user project.

### Changes regarding Akka.Persistence.Sql.Common

- Querying mechanism - now user can query the journal using `Query` message. Message may contain set o `IHint` instances used to filter the data (currently only Manifest field and collection of PersistenceIds are supported, more to be written in the future). Response is collection of `QueryResponse` messages containing events filtered for query followed by `QuerySuccess` message which finishes response transmission. If an error occurred final message send is `QueryFailure` message
- **(breaking change)** Database column `PayloadType` has been renamed to `Manifest`. Now it's more general and can contain user defined manifests if necessary.
- SQL-based journals can now use common abstraction in form of `SqlJournal` class. This simplifies further plugin development and handles query messages.
- Attached new project `Akka.Persistence.Sql.Common.TestKit` that contains `SqlJournalQuerySpec` to be inherited by dependent plugins specs in order to verify if querying is working correctly.
- Extracting connection string from either HOCON or App.config is now handled at this library level.

### Changes regarding Akka.Persistence.SqlServer

- `Akka.Persistence.SqlServer` is again joined into core repo - reason for that is that `Akka.Persistence.Sql.Common` contained dead code, that was not used/tested anywhere in core solution. Me and @rogeralsing agreed that at least one sub-plugin should be presented here.
- Plugin has been updated to use features from the current `Akka.Persistence.Sql.Common` lib.
- Extension will now log at `Warning` level info, if used connection string don't specify *MultipleActiveResultSets=True*, as it may be reason of future problems. /cc #1291
- Journal will also log at `Error` level if after insertion command will return result other than *1* for any reason (which is not a desired behavior).


Aaronontheweb:
Weird... why didn't the build server pick this up?

Aaronontheweb:
Ah, figured it out - the timestamp of this commit is from August 18th, so TC won't pick it up.

Could do the following and update the timestamp to present time? http://stackoverflow.com/questions/454734/how-can-one-change-the-timestamp-of-an-old-commit-in-git

Horusiath:
@Aaronontheweb all tests except those, which require SQL Server to be initialized, have passed.

Horusiath:
I've introduced few more changes:

- No more `DbConnection` caching - now it's created from scratch for each operation.
- `SnapshotStore` delete methods are now async (to reflect incomming changes on akka 2.4).
- Whole sql persistence API has been remade with asyn/await in mind.

rogeralsing:
error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections.

Horusiath:
@rogeralsing @Aaronontheweb Builds are passing. Could you review the changes?

Issue #1292, opened 2015-09-04
Unable to use await while message processing


alexkool:
Scheme:
`System -> MasterActor -> ChildrenActors.`

We pass `MsgReqBalance` message `MaserActor` in foreach loop we trying to `Ask<MsgRespBalance>(new MsgReqBalance())` all children and get blanace from all children. If we try to use `awat` we will fall while getting `Context` in `MasterActor`

so. This code raise an error `There is no active ActorContext, this is most likely due to use of async operations from within this actor.`:
```
    public async void Handle(MsgReqBalance message)
        {
            List<BalanceEntry> entries = new List<BalanceEntry>();
            foreach (var clientActor in Context.GetChildren())
            {
                var resp = await clientActor.Ask<MsgRespBalance>(new MsgReqBalance(), ActSys.OperationTimeout);
                entries.Add(resp.BalanceEntry);
            }
            Context.Sender.Tell(new MsgRespComplexBalance() // << here we fall
            {
                BalanceEntries = entries
            });
        }
```

but this code works fine:
```
    ConcurrentBag<BalanceEntry> entries = new ConcurrentBag<BalanceEntry>();
    var actions = Context.GetChildren().Select(actor => new Action(() =>
    {
        var task = actor.Ask<MsgRespBalance>(new MsgReqBalance(), ActSys.OperationTimeout);
        task.Wait();
        entries.Add(task.Result.BalanceEntry);
    }));
    Parallel.Invoke(actions.ToArray());
    Context.Sender.Tell(new MsgRespComplexBalance()
    {
        BalanceEntries = entries.ToList()
    });
```
The goal is using `Parallel.Invoke`. It blocks the execution but code works!

Please explain why this is so.

alexkool:
I found the another way :)
before using `await` I save the sender:
`var sender = Context.Sender;`

and then:
```
sender.Tell(new MsgRespComplexBalance()
            {
                BalanceEntries = entries
            });
```
So. We can use `await` in this way! all tests are passes)

Horusiath:
Using `async/await` and `Ask` between the actors is generaly a strong anti-pattern: it results in suboptimal performance and introduces races (just like in your case). Not mentioning that `Parallel.Invoke` will actually actively block your actor untill all of the messages will arrive.

alexkool:
thanks for a comment! 
*suboptimal and are obsoleted* sorry, I dont know that
If u can, please write a sample.
Regards!

Caldas:
[...]

PR #1308, opened 2015-09-11
Support of "await" continuation

vchekan:
I know, I know, there were several (heated) discussions about async support for actor model (#946), also there were some blogs and documentation, encouraging usage of PipeTo.

Why am I trying to do it again than? The answer is, because PipeTo, UntypedActor.RunTask, etc are not support of async, but a synchronous invocation of asynchronous operation. 

I do understand the reasoning behind it. State of the actor can change after "await" and it might introduce subtle bugs and make code less reliable.

But let's think about some use cases one more time. Lets say I have a dynamic configuration module, which does long http polls. (https://www.consul.io/) When change in config happen, the module will receive a response from http, update its internal configuration structures, and issue another poll. 
  At the same time this module is responsible for handling "get me config X" requests. Obviously, if long poll was issued with PipeTo, all requests to get configuration property will be queued and not handled. So this is not the way to go.

In general, c# programmers are well trained by now, that Task.Run means no lock, so the behavior of PipeTo comes as a surprise.

Once I learn limitations of PipeTo, I started thinking, how to work around it. At first, I thought about having another child actor, which would do only polls, thus freeing configuration component to respond to configuration property requests. But it would complicate otherwise simple code. And I thought about all other cases, when I need to perform async action in actor. 

Next iteration was to execute polling in a dedicated thread, separate from actor, and send messages to "self" with updates. This would solve async problem, but what about supervision? The main point of akka is to supervise the code which is prone to failures. And http request is exactly the code, I would like to keep an eye on. But now it is executed in a separate thread, and I have to resort to old school try-catch. Still not the experience I want from akka.

So I started to think that what "await" does is so similar to akka's mailbox. In fact, Await split the function into "before await" and "after await" functions. When 1st function is complete, the second one is invoked with result from the previous one. What if I could intercept this moment and instead of invoking the invisible function after "await", send it to the actor's mailbox? This way, every span of code which is split by "await" becomes a message and is routed through actor's mailbox.

Luckily I do have experience with doing exactly this. The fundamental fifference between "await" and Task.ContinueWith is that await will check for presence of SynchronizationContext, and if present, will delegate invocation of continuation to the context. With custom context set, we can schedule execution of continuation as a message to ourselves. And this is exactly what this patch does. I introduce new type of message, AsyncContinuation, set async context before handling the message to capture asyncs, and when task is done, it delegates continuation to my sync context, which delegates it to the actor's mailbox. I even manage to preserve Sender, so it is set correctly even after "await" (which solves #1292). Pay attention to the way I do async. Instead of spawning a task, I mark OnReceive as async. This means that the code before 1st await has System Self and Sender set up properly. After await is scheduled to be executed through mailbox, so context is set up properly again. It seems to me, my solution solves context problem completely.

But I can see a problem. For example, I have no idea how FSM would work.
Another problem is that actor state can change between awaits. If you have someting like

    if(_myList.Length == 0) {
        await _myList.Add(LoadList());
bad things may happen, because by the time load is complete, list might be not empty anymore. This problem cant be solved with akka, and developer must be aware of await side effects.

I am not too confident in unit test at the moment, not sure how to test it properly. Because Dispatch can schedule different runs onto different threads, I can't validate that multiple messages are not executed simultenously. So test needs improvement. But debugging indicates that code does right things.

Let me know what do you think.

### DIFF
--- a/src/core/Akka.Tests/Actor/AkkaSyncContextSpec.cs
+++ b/src/core/Akka.Tests/Actor/AkkaSyncContextSpec.cs
@@ -43,13 +43,12 @@ namespace Akka.Tests.Actor
                 switch(message as string)
                 {
                     case "start":
-                        var sender = Sender;
                         System.Diagnostics.Debug.WriteLine("{0:O} Before async thread: {1}", DateTime.Now, Thread.CurrentThread.ManagedThreadId);
                         await Task.Delay(100);
                         System.Diagnostics.Debug.WriteLine("{0:O} After async thread: {1}", DateTime.Now, Thread.CurrentThread.ManagedThreadId);
                         Thread.Sleep(300);
                         System.Diagnostics.Debug.WriteLine("{0:O} After sleep thread: {1}", DateTime.Now, Thread.CurrentThread.ManagedThreadId);
-                        sender.Tell("finish");
+                        Sender.Tell("finish");
                         break;
                     case "start2":
                         System.Diagnostics.Debug.WriteLine("{0:O} Got start2 in thread {1}", DateTime.Now, Thread.CurrentThread.ManagedThreadId);
--- a/src/core/Akka.Tests/Actor/AkkaSyncContextSpec.cs
+++ b/src/core/Akka.Tests/Actor/AkkaSyncContextSpec.cs
@@ -1,0 +1,65 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Event;
+using Akka.TestKit;
+using Xunit;
+
+namespace Akka.Tests.Actor
+{
+    public class AkkaSyncContextSpec : AkkaSpec
+    {
+        private readonly IActorRef _actor;
+
+        [Fact]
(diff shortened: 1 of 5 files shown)


Aaronontheweb:
> But I can see a problem. For example, I have no idea how FSM would work.
> Another problem is that actor state can change between awaits. If you have someting like

> if(_myList.Length == 0) {
>    await _myList.Add(LoadList());
> bad things may happen, because by the time load is complete, list might be not empty anymore. This > problem cant be solved with akka, and developer must be aware of await side effects.

This problem is currently solved by Akka, which is why our long-standing position about `await` and `PipeTo` is what it is. It's not because we hate async, but rather because `async` / `await`'s design isn't compatible with the conceptual integrity of Akka.NET actors. Actors internal state is inherently thread-safe because they process messages serially and mutate their state synchronously - that's one of about the sacrosanct concepts behind the project.

If you need to do long-polling, write an actor who polls. Better yet: use sockets and eliminate the need for polling to begin with.

rogeralsing:
1. Synchronizationcontext should only be used by threads you own, not someone röses threads. 

2. Setting/restoring sync context in the receive pipeline will severely degrade performance of normal messages.

Async await is a "nice to have" feature, it should not come with a cost for the standard pipeline

rogeralsing:
Autocorrection :-/ "someone Elses"

Issue #1309, opened 2015-09-12
Incorrect(?) Async CircuitBreaker + tests


KadekM:
I believe our CircuitBreaker async impl is not correct - this is our current interface
```
public async Task<T> WithCircuitBreaker<T>( Task<T> body )
``` 
while scala has
```
def withCircuitBreaker[T](body: ⇒ Future[T]): Future[T]
```

They are using call by name - as it makes sense, since Future starts instantly when instantiated... and passing already-running-future into circuit breaker would not make sense, since what can we circuit break then?

So I believe our current imlementation does not really make sense:
1. If you pass in running task (as we do in tests), then what are you circuit breaking ? Sure, OPEN event will fire, but you are already running task (and thus potentially contacting external service, and kicking it while it's down on ground, potentially killing it)
2. If you pass in not running task, then noone in circuit breaker starts it (of course, why should they).

In this sense I think it's now working.

So I believe, idiomatic port should be

```
public async Task<T> WithCircuitBreaker<T>( Func<Task<T>> body )
``` 

Or otherwise we would have to require clients to pass in not-yet-started task, and start it ourselves in circuit breaker (if its closed)... but then that would be pretty stupid, usually when you do `MyExternalService.DoThisRiskyOperation` it returns you already started task...

I will submit a PR, with changes to interface as described here.

rogeralsing:
Merged.

PR #1310, opened 2015-09-12
fix circuit breaker as per #1309

KadekM:
#1309


rogeralsing:
I'm merging this. the circuitbreaker is not related to the failing multinode test runner.

Issue #1285, opened 2015-08-31
1.0.4 breaks F# actors that send unit to actors that allow _ messages


raymens:
Actors that accept `_` in F# will not receive any messages send as a `()` *(unit)*. If the message is something else like a string, it does work. Or if the actor specifically accepts a `()` *(unit)* it also works.

Our codebase contains several pieces of code where this issue happens and as such we can't upgrade to 1.0.4 without altering our codebase.

Code sample that shows the issue:
```fsharp
open Akka.FSharp

let getWhateverHandler (mailbox : Actor<'a>) _ =
    printfn "Is this working?"

let system = System.create "my-system" (Configuration.load())
let parentActor = spawn system "Whatever" (actorOf2 getWhateverHandler)

// This doesn't work
parentActor <! ()

// This works
parentActor <! "This works"
```

rogeralsing:
cc @Horusiath is this a result of the JObject hack?

Aaronontheweb:
Time to add a unit test for this.

rogeralsing:
Is this really a serialization bug? the example above does not use remoting and no serialization

raymens:
Are these valid tests to use?

```
[<Fact>]
// FAILS currently
let ``actor that accepts _ will receive unit message`` () =    
    let timeoutConfig =
        """
        akka { 
            actor {
                ask-timeout = 5s
            }
        }
        """
        |> Configuration.parse 

    let getWhateverHandler (mailbox : Actor<_>) _ = 
        mailbox.Sender() <! "SomethingToReturn"

    let system = System.create "my-system" timeoutConfig
    let aref = spawn system "UnitActor" (actorOf2 getWhateverHandler)

    let response = aref <? () |> Async.RunSynchronously
    response
    |> equals "SomethingToReturn"

[<Fact>]
// SUCCEEDS
let ``actor that accepts unit will receive unit message`` () =    
    let timeoutConfig =
        """
        akka { 
            actor {
                ask-timeout = 5s
            }
        }
        """
        |> Configuration.parse 

    let getWhateverHandler (mailbox : Actor<unit>) () = 
        mailbox.Sender() <! "SomethingToReturn"

    let system = System.create "my-system" timeoutConfig
    let aref = spawn system "UnitActor" (actorOf2 getWhateverHandler)

    let response = aref <? () |> Async.RunSynchronously
    response
    |> equals "SomethingToReturn"
```

rogeralsing:
[...]

PR #1317, opened 2015-09-17
Added tests for F# null bug

rogeralsing:
This PR contains the F# tests that @raymens created in #1285

The failing test is commented out for now.
@Horusiath the actual problem is in here:

```fsharp
   let private tryCast (t:Task<obj>) : 'Message =
        match t.Result with
        | :? 'Message as m -> m
        | o ->                         //<------------- here, if the user Tells () to an actorref, the () will not be matched by the above line as it is not of that type. 
                                         //instead the unit will be processed here, as a JObject, which it is not..
                                         //my F#-fu is not strong enough to know how to deal with the F# type system :-/

            let context = Akka.Actor.Internal.InternalCurrentActorCellKeeper.Current
            if context = null 
            then failwith "Cannot cast JObject outside the actor system context "
            else
                let serializer = context.System.Serialization.FindSerializerForType typeof<'Message> :?> Akka.Serialization.NewtonSoftJsonSerializer
                match Serialization.tryDeserializeJObject serializer.Serializer o with
                | Some m -> m
                | None -> raise (InvalidCastException("Tried to cast JObject to " + typeof<'Message>.ToString()))
```

So we either need to fix this, or just decide that we should not be able to pass () to F# actors

### DIFF
--- a/src/core/Akka.FSharp.Tests/ApiTests.fs
+++ b/src/core/Akka.FSharp.Tests/ApiTests.fs
@@ -72,3 +72,71 @@ let ``can serialize and deserialize discriminated unions over remote nodes`` () 
     response
     |> equals msg
 
+//[<Fact>]
+// FAILS
+let ``actor that accepts _ will receive unit message`` () =    
+    let timeoutConfig =
+        """
+        akka { 
+            actor {
+                ask-timeout = 5s
+            }
+        }
+        """
+        |> Configuration.parse 
+
+    let getWhateverHandler (mailbox : Actor<_>) _ = 
+        mailbox.Sender() <! "SomethingToReturn"
+
+    let system = System.create "my-system" timeoutConfig
+    let aref = spawn system "UnitActor" (actorOf2 getWhateverHandler)
+
+    let response = aref <? () |> Async.RunSynchronously
+    response
+    |> equals "SomethingToReturn"
+
+[<Fact>]
+// SUCCEEDS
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Thanks @raymens, for coming up with this - we appreciate you!

Aaronontheweb:
@Horusiath @rogeralsing do we want to pull this in and create a separate issue for fixing the failing test?

rodrigovidal:
@rogeralsing @Aaronontheweb I've investigated this a little bit. In the case of the test that fails it doesn't seem to even reach the "tryCast" function. I removed the timeout configuration, and it just hang forever in the .Result.

@rogeralsing Have you experienced a different behavior?

rogeralsing:
@Aaronontheweb I think we should pull it in, the json.net f# stuff might take a while...

raymens:
@rodrigovidal that was indeed the case and that's why I added the timeout

@Aaronontheweb no problem, thank you for akka.net ;)

rodrigovidal:
@raymens Oh, I see. What I was trying to say is the problem doesn't seem to be where @rogeralsing pointed. I'll investigate further.

Aaronontheweb:
@raymens Merged. could you please create a Github issue to get the test included in this pull request to pass?

raymens:
@rodrigovidal oh sorry, I misunderstood. Thanks for investigating.

@Aaronontheweb I will, thanks for the merge.

PR #1319, opened 2015-09-17
Wire Serializer

rogeralsing:
This PR introduces the Wire serializer as an alternative serializer for Akka.NET.

The serializer is not set as the default serializer here, it is only included so users can activate it manually by changing what objects should use it in the config.
All tests pass when the serializer is enabled instead of the json.net serializer.

I would like users to evaluate this for some time and try it out in real world scenarios.

To enable the serializer:

```javascript
    serialization-bindings {
      "System.Byte[]" = bytes
      "System.Object" = wire //instead of json
    }
```

The main features this addresses

* Fully supports all F# types
* 27 times faster than our current Json.NET setup

### DIFF
--- a/src/core/Akka/Akka.csproj
+++ b/src/core/Akka/Akka.csproj
@@ -63,6 +63,10 @@
     <Reference Include="System.Core" />
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Xml" />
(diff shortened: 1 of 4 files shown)


Horusiath:
Merged

Aaronontheweb:
I'll try this out in my local builds here

rogeralsing:
:+1:

PR #1320, opened 2015-09-17
Use ActorTaskScheduler.AwaitTask instead of RunTask for ReceiveActor

nvivo:
This is a proposal to make `ActorTaskScheduler.AwaitTask` the standard API for async message handlers. 

It's a simple optimization to avoid calling RunTask and pausing and resuming when the task is synchronous. This is more inline with how "await" is implemented.

The main benefit is to make code that return completed tasks perform exactly the same as if an action was used, and there is no change to the API.

@rogeralsing, this should be straightforward. See if that works for you.

### DIFF
--- a/src/core/Akka/Actor/ReceiveActor.cs
+++ b/src/core/Akka/Actor/ReceiveActor.cs
@@ -116,7 +116,7 @@ namespace Akka.Actor
             _matchHandlerBuilders.Peek().Match<T>( m =>
             {
                 Func<Task> wrap = () => handler(m);
-                ActorTaskScheduler.RunTask(wrap);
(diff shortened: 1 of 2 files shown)


rogeralsing:
Looks great @nvivo !

Aaronontheweb:
@nvivo restarting the build - looks our poor little build agent had a rough morning

nvivo:
Unfortunately it was too good to be true. =) After digging into why some async tests stopped working, I realized that without changing the logic of the ActorTaskScheduler to do what it was supposed to do with a solution based on #946, it's not possible to optimize this at library level. Since there is no interest in adding these kind of optimization currently, this remains to be something to optimize at application level. I'm closing this.

Issue #1246, opened 2015-08-14
Port Akka.Remote MultiNodeSpec: PiercingShouldKeepQuarantineSpec


Aaronontheweb:
https://github.com/akka/akka/blob/master/akka-remote-tests/src/multi-jvm/scala/akka/remote/PiercingShouldKeepQuarantineSpec.scala

KadekM:
Are there known issues with this?
I have written the test but it seems, that even though I speak to actor from system which I quarantineed, I am getting answers.

KadekM:
I'd like to do this

KadekM:
I wrote the test - can someone take a look at this please ?
https://github.com/KadekM/akka.net/blob/issue-1246/src/core/Akka.Remote.Tests.MultiNode/PiercingShouldKeepQuarantineSpec.cs#L94-L106

Are there some known issues in this regards or am I doing something stupid ? (expecting none ->failing that we received some  |  expecting some -> failing that none was received).

According to log messages communicating between each other...
```
[NODE2][8/22/2015 1:00:04 PM][DEBUG][[akka://PiercingShouldKeepQuarantineSpec/system/endpointManager/endpointWriter-akka.tcp%3a%2f%2fPiercingShouldKeepQuarantineSpec%40localhost%3a57955-3]]: received local message [RemoteMessage: ActorSelectionMessage - Message: getuid - WildCartFanOut: False - Elements: user/subject to [akka://PiercingShouldKeepQuarantineSpec/]<+akka://PiercingShouldKeepQuarantineSpec/ from [akka.tcp://PiercingShouldKeepQuarantineSpec@localhost:57955/user/testActor1]]
[NODE1][8/22/2015 1:00:04 PM][DEBUG][[akka://PiercingShouldKeepQuarantineSpec/system/endpointManager/reliableEndpointWriter-akka.tcp%3a%2f%2fPiercingShouldKeepQuarantineSpec%40localhost%3a57956-4/endpointWriter]]: Drained buffer with maxWriteCount: 50, fullBackoffCount: 1,smallBackoffCount: 0, noBackoffCount: 0,adaptiveBackoff: 10000
[...]

PR #1322, opened 2015-09-19
Issue #1246 - PiercingShouldKeepQuarantineSpec

KadekM:
Ported spec from Akka, but with larger akka.remote.retry-gate-closed-for-timer.

How can akka spec actually pass ? If I set akka.remote.retry-gate-closed-for = 0.5s (as they have in their spec), then in a for part, then spec would not pass for some i > 0... as what I believe happens is, that after 0.5s quarantine is over and next assert ExpectNoMessage fails (since it waits for 2s)...

With increased retry-gate-closed-for it should pass...What am I not seeing ?


Aaronontheweb:
Need to update the commit date to latest in order to get this to run on the build server - current commit timestamp is 28 days old.

PR #1338, opened 2015-09-29
Cluster singleton proxy finished

Horusiath:
I've managed to finish missing parts in ClusterSingletonProxy implementation an make it working with cluster singleton example projects.


Aaronontheweb:
@Horusiath what's the issue with using Linq 4.0 and NuGet 2.8.1 not working?

PR #1340, opened 2015-10-01
Added an overload using non-generic System.Type

cdmdotnet:
As discussed with Andrew Skotzko, when you dynamically want to resolve an actor, without an overload using System.Type, and only having generic types, you are forced to hard code large switch statements and also have the class know the entire universe.

This doesn't change the implementation, it just adds an overload.

### DIFF
--- a/src/contrib/dependencyInjection/Akka.DI.AutoFac/AutoFacDependencyResolver.cs
+++ b/src/contrib/dependencyInjection/Akka.DI.AutoFac/AutoFacDependencyResolver.cs
@@ -89,7 +89,17 @@ namespace Akka.DI.AutoFac
         /// <returns>The configuration object for the given actor type</returns>
         public Props Create<TActor>() where TActor : ActorBase
         {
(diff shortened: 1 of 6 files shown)


cdmdotnet:
I've checked the build server and I could be reading things wrong, but it appears something didn't pass a test in an area I didn't touch... but TC didn't report a broken test. Did i fork from the wrong branch?

cdmdotnet:
How do I re-trigger a build if if was a transient issue.

Aaronontheweb:
@cdmdotnet we have some known race conditions inside the multi-node testkit (part of the Akka.Cluster and Akka.Remote testing infrastructure; still in beta), which is an absolute bear to debug. 

You can ping @akkadotnet/contributors - a bunch of us can restart builds. Eventually we want a bot that can do this.

cdmdotnet:
Cool. I feel your pain.

rogeralsing:
:+1:

Issue #1332, opened 2015-09-25
Akka.Cluster.Routing.ClusterRouterGroupSettings Mono Linux issue


evertmulder:
When creating a cluster router in code the following code works perfectly on windows but fails when running under linux mono
```csharp
using Akka.Cluster.Routing;
using System;
using System.Collections.Immutable;

namespace MonoTestClusterRouterGroupSettings
{
    class Program
    {
        static void Main(string[] args)
        {
            var test = new ClusterRouterGroupSettings(10, false, "sink", ImmutableHashSet.Create("/user/blocklist"));
            Console.WriteLine("OK {0}", test.RouteesPaths.Count);
            Console.ReadLine();
        }
    }
}
```
Running under mono in windows I get the expected OK 1, running under Linux I get 
```
Unhandled Exception:
System.ArgumentException: routeesPaths [/user/blocklist] is not a valid relative actor path.
Parameter name: routeesPaths
  at Akka.Cluster.Routing.ClusterRouterGroupSettings..ctor (Int32 totalInstances, Boolean allowLocalRoutees, System.String useRole, System.Collections.Immutable.ImmutableHashSet`1 routeesPaths) [0x00000] in <filename unknown>:0
  at MonoTestClusterRouterGroupSettings.Program.Main (System.String[] args) [0x00000] in <filename unknown>:0
[ERROR] FATAL UNHANDLED EXCEPTION: System.ArgumentException: routeesPaths [/user/blocklist] is not a valid relative actor path.
Parameter name: routeesPaths
  at Akka.Cluster.Routing.ClusterRouterGroupSettings..ctor (Int32 totalInstances, Boolean allowLocalRoutees, System.String useRole, System.Collections.Immutable.ImmutableHashSet`1 routeesPaths) [0x00000] in <filename unknown>:0
  at MonoTestClusterRouterGroupSettings.Program.Main (System.String[] args) [0x00000] in <filename unknown>:0
```
Tested on Rasbian Linux and Ubuntu 12.04.
Used Akka.Cluster.1.0.5.86-beta

Aaronontheweb:
leading slashes for the win here

Aaronontheweb:
In more detail - "/user/blocklist" is a valid absolute path on Linux. It's a valid relative path on Windows.

Hmmm.... We need to make this workable with Linux... Let me take a look at ClusterRouterGroupSettings.

evertmulder:
[...]

PR #1342, opened 2015-10-02
Fix for  #1332 Akka.Cluster.Routing.ClusterRouterGroupSettings Linux issue

evertmulder:
New PR with 1 commit. See issue for more details


Aaronontheweb:
@akkadotnet/developers pulling this in, but would love some additional verification of the fix on Mono boxes (since Mono CI is fried due to Mono 4.0 bugs)

PR #1349, opened 2015-10-02
rewrote ActorPath.TryParseAddrss to now test Uris by use Uri.TryCreate instead of try / catch

Aaronontheweb:
I get a flurry of handled `UriFormatException`s whenever I run Akka.NET specs, and that's been by design. Rewrote that code to use `Uri.TryCreate` instead so it no longer throws period.


rogeralsing:
:+1:

Issue #1325, opened 2015-09-22
Implement comparability in value-based data types


Horusiath:
Right now we have a problem with using some of the Akka.NET data types, like `IActorRef`, as keys in our AVL-based maps - including all `IImmutableMap<,>`, `IImmutableSet<>` and F# `Map` implementations. All of them requires, that key type is implementing `IComparable<TKey>` interface in order to place key-value properly in tree structure.

For this reason, we should implement `IComparable` interface for as many types as reasonable. I think, that basically any type that contains custom equality and hashCode generation method, should be able to implement interface.

Moreover, in F# we are able to create comparability, equality and hash code generation in pretty automatic way for most of the data types. Do you think we could try to automatize this process using Roslyn compiler and just generate IComparable impl on the fly?

Aaronontheweb:
> Do you think we could try to automatize this process using Roslyn compiler and just generate IComparable impl on the fly?

I think this is a good idea.

For `IActorRef` at least, the underlying `ActorPath` implements `IComparable<ActorPath>` - we could just have the `IComparable<IActorRef>` implementation call down to that.

rogeralsing:
Agree, with IActorRef, just compare the underlying path.

As for Roslyn, do you mean we should make a plugin to roslyn that runs when we compile`?
OR that we make a tool that just walks through all interfaces and implement that parts for us. and we just run it once to do this?

maxim-s:
I can try to implement it

rogeralsing:
IActorRef (+impls) and Props are good candidates, and maybe also `IDecider`, `Deploy`, `RouterConfig` + subclasses of those should have this support.
@Aaronontheweb @Horusiath any other types that also needs this?

maxim-s:
It seems like
interface ```IActorRef``` (impls) already has ```IComparable<TKey>``` implementation
and it's possible to use it as a key for AVL-based maps

rogeralsing:
This is what we get when trying to create an F# Set of IActorRef:

![actorrefset](https://cloud.githubusercontent.com/assets/647031/10220882/cfe453a4-684a-11e5-9425-aab103969a77.png)

rogeralsing:
It's strange, because as @maxim-s  says, it does implement the interface
Does it have to implement a non generic version of the interface too for F#?
Or [CustomEquality] / [CustomComparison] attributes?

```csharp
[...]

PR #1351, opened 2015-10-04
implemented IComparable  genericless variant

maxim-s:
fix https://github.com/akkadotnet/akka.net/issues/1325


rogeralsing:
Looking goo @maxim-s !
Welcome on board

Issue #1321, opened 2015-09-18
Deployer doesn't work with config fallbacks


kantora:
I am using configs stack (system has modules and every module has it's own default config, that are stacked in application).

I've found out that if one part of akka.actor.deployment is defined in one config (for one actor path) and second is defined in another (for other actor path) - only the one on the top will be used (the second one is ignored).

rogeralsing:
I'll take a look at this

rogeralsing:
I've added a test for this scenario, are your configurations chained with `WithFallback` ?
Can you provide a minimal example that reproduce the issue?

see #1353 for more info

rogeralsing:
I'm unable to replicate this one, anyone? @akkadotnet/developers 
Close?

kantora:
Sorry, needed to switch to other project for some time.
Issue is still reproduced on Akka 1.0.6.16

I have the following test:
some actor has initialization of subworkers like
`this.workers = Context.ActorOf(Props.Create(typeof(Worker)).WithRouter(FromConfig.Instance), "workers");`

In case I define 
`
akka.actor.deployment {
    /MyActor/workers {
            router = round-robin-pool
            nr-of-instances = 5
        }  
} 
`
on top-level config - everything is working fine.

But if this is defined after series of Fallback for final config generation, I am getting Akka.Configuration.ConfigurationException on actor start

kantora:
I'll try to provide fully functional test for this issue.

kantora:
[DeployerFallbackTest.zip](https://github.com/akkadotnet/akka.net/files/156635/DeployerFallbackTest.zip)
This test demonstrates the problem.

Danthar:
Fixed via PR

PR #1353, opened 2015-10-04
Deployer Spec WIP

rogeralsing:
WIP for #1321.

I am unable to reproduce the problem for the reported issue.


kantora:
I've attached working test to reproduce the problem
[https://github.com/akkadotnet/akka.net/issues/1321#issuecomment-191712404](comment)

Issue #1345, opened 2015-10-02
MultiNodeSpecs: add millisecond time resolution to log messages


Aaronontheweb:
Per-second messages are too coarse, i.e.

```
[NODE2][9/30/2015 12:56:48 AM][DEBUG][EventStream(TestConductorSpecNode2)]: Logger log1-TestEventListener [TestEventListener] started
[NODE2][9/30/2015 12:56:48 AM][DEBUG][EventStream(TestConductorSpecNode2)]: StandardOutLogger being removed
[NODE2][9/30/2015 12:56:48 AM][DEBUG][EventStream(TestConductorSpecNode2)]: Default Loggers started
[NODE1][9/30/2015 12:56:48 AM][DEBUG][remoting]: Starting remoting
[NODE2][9/30/2015 12:56:48 AM][DEBUG][remoting]: Starting remoting
[NODE1][9/30/2015 12:56:48 AM][DEBUG][remoting]: Starting prune timer for endpoint manager...
[NODE2][9/30/2015 12:56:48 AM][DEBUG][remoting]: Starting prune timer for endpoint manager...
[NODE1][9/30/2015 12:56:49 AM][DEBUG][Akka.Remote.Transport.FailureInjectorTransportAdapter]: FailureInjectorTransport is active on this system. Gremlins might munch your packets.
[NODE2][9/30/2015 12:56:49 AM][DEBUG][Akka.Remote.Transport.FailureInjectorTransportAdapter]: FailureInjectorTransport is active on this system. Gremlins might munch your packets.
[NODE1][9/30/2015 12:56:49 AM][DEBUG][remoting]: Remoting started
[NODE1][9/30/2015 12:56:49 AM][DEBUG][remoting]: Remoting now listens on addresses: [akka.trttl.gremlin.tcp://TestConductorSpecNode1@localhost:19124]
[NODE2][9/30/2015 12:56:49 AM][DEBUG][remoting]: Remoting started
[NODE2][9/30/2015 12:56:49 AM][DEBUG][remoting]: Remoting now listens on addresses: [akka.trttl.gremlin.tcp://TestConductorSpecNode2@localhost:19125]
[NODE2][9/30/2015 12:56:50 AM][DEBUG][PlayerHandler]: connected to 127.0.0.1:4711
[NODE1][9/30/2015 12:56:50 AM][DEBUG][Akka.Remote.TestKit.ConductorHandler]: connection from 127.0.0.1:19126
[...]

PR #1358, opened 2015-10-05
Update LogEvent.cs

AntoineGa:
Modify ToString method to add milliseconds

### DIFF
--- a/src/core/Akka/Event/LogEvent.cs
+++ b/src/core/Akka/Event/LogEvent.cs
@@ -67,7 +67,7 @@ namespace Akka.Event
         /// <returns>A <see cref="System.String" /> that represents this LogEvent.</returns>
         public override string ToString()
         {
-            return string.Format("[{0}][{1}][Thread {2}][{3}] {4}", LogLevel().ToString().Replace("Level", "").ToUpperInvariant(), Timestamp, Thread.ManagedThreadId.ToString().PadLeft(4, '0'), LogSource, Message);
+            return string.Format("[{0}][{1}][Thread {2}][{3}] {4}", LogLevel().ToString().Replace("Level", "").ToUpperInvariant(), Timestamp.ToString("yyyy-MM-dd HH:mm:ss.fff"), Thread.ManagedThreadId.ToString().PadLeft(4, '0'), LogSource, Message);
         }
     }
 }


AntoineGa:
related to https://github.com/akkadotnet/akka.net/issues/1345

rogeralsing:
This could potentionally kill monitoring systems, monitoring tools that rely on the current format could stop working.

Im not sure that changing the log events to fix a problem in the multi node testkit is the right way to go about this. Making a logger for the testkit that does the correct extraction feels more solid imo. @Horusiath @Aaronontheweb  thoughts?

Aaronontheweb:
Having the millisecond times is actually a bigger issue than just the multi-node testkit; in soft real-time scenarios, which many users of Akka.NET have, it's important to have this data to review the order of concurrent events precisely, which we don't support right now. This should be part of the logging system in general.

sean-gilliam:
How about making it a configuration option which defaults to using MS if not specified?

rogeralsing:
>which we don't support right now. This should be part of the logging system in general.

What?
Ofc we do, you have access to all the stuff inside the log events in your logger, just dont use .ToString() to create the logging output.
Serilog does this just fine https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/loggers/Akka.Logger.Serilog/SerilogLogger.cs#L31
This is about the integrity of the project, if we start scotch taping stuff that arent broken just to keep moving, things will eventually fall apart.

If milliseconds are needed to create a good visualizer, make a custom logger that outputs the timestamp as ticks.

We should solve this the same way as JVM Akka does.

AntoineGa:
Sorry, i though the use of .ToString() was intended.
Should we switch to Serilog / SerilogLogMessageFormatter and remove all the ToString usage ?

rogeralsing:
@AntoineGa in my opinion, we should handle this with a special logger for the multinode testkit, all the relevant data we need is in the log message, we could extract things like exception details better than we do today.
The special logger could use similar logic as Serilog does for extracting the data from the log messages.

That would solve both the issue with exception details and timing precision at the same time

Aaronontheweb:
> @AntoineGa in my opinion, we should handle this with a special logger for the multinode testkit, all the relevant data we need is in the log message, we could extract things like exception details better than we do today.

I disagree with this - it's an issue bigger than the multi-node testkit. Time resolution issues will come up in any HA environment. A custom logger for this is obnoxious.

> This is about the integrity of the project, if we start scotch taping stuff that arent broken just to keep moving, things will eventually fall apart.

Adding milliseconds to the default time resolution of a log message doesn't seem significant enough to merit this, but your point is well-made.

What I'd propose we do instead is add a format string to HOCON that sets the format string used in printing dates inside log messages. Leave the default as-is and allow folks who need higher time resolution to just update the configuration. That should work, no?

rogeralsing:
btw. there are sections for this in jvm Akka: http://doc.akka.io/docs/akka/snapshot/java/logging.html
http://doc.akka.io/docs/akka/snapshot/java/logging.html#More_accurate_timestamps_for_log_output_in_MDC

We should look into how Akka solves this too

rogeralsing:
Ok, so me and @Aaronontheweb have taked about this one.

The cleanest way forward would be to introduce an overload of the LogEvent ToString that takes a format string.
see https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Event/StandardOutLogger.cs#L123

We could replace that call with something like `StandardOutWriter.WriteLine(logEvent.ToString(_format), color);`
The format could be read from the config so that users can overrde it.
The absolutely easiest way forward would be to just use .NET's String.Format style. e.g. `[{0}][{1}] : Message {2}` that way we could still use standard formatting args. e.g. how dates and times are presented.
The problem in that case would be that nobody would know what the heck 0,1 and 2 means w/o reading the docs.
so an alternative might be to just replace the placeholders

 `[{ThreadId}][{Timestamp:yyyy-MM-dd HH:mm:ss.fff}] : Message {Message}`

That way, we could have a readable template, while still having support for value formatting.

Thoughts on this?

sean-gilliam:
:+1: on the named fields template

AntoineGa:
{Timestamp:yyyy-MM-dd HH:mm:ss.fff} is the Serilog way to manage custom format, is it rightly managed by the other logger ? Or do you refer to C#6 String Interpolation ( https://roslyn.codeplex.com/discussions/540869 ) ?

rogeralsing:
@AntoineGa see #1367. it enables us to write named arguments but those are just placeholders for argument indexes in a normal string format.
so `{Timestamp:yyyy-MM-dd HH:mm:ss.fff}` in this case, would be translated to: `{0:yyyy-MM-dd HH:mm:ss.fff}`

This would allow us to have named logging template arguments while still being able to format each argument with standard .NET string format rules.

rogeralsing:
Closing this in favor for #1367
The log format in this PR can be used in the new log template

PR #1360, opened 2015-10-05
Ported AttemptSysMsgRedeliverySpec from Akka; Fixed MNTK bugs

Aaronontheweb:
Should resolve `AttemptSysMsgRedeliverySpec`, which now passes.
Fixed issue with PlayerFSM that broke Throttle in MNTK
Moved Helios to single thread in order to preserve message ordering, which has the added benefit of allowing the `AkkaProtocolStressSpec` to finally work.

Going to look into perf issues with Helios and add a benchmark for it so I can gauge the impact, but having the correct message order over the network is more important than speed.


Aaronontheweb:
Bah, C6 issues on the build server

Aaronontheweb:
@GraemeBradbury looking at the logs, it appears as though `AttemptSysMsgRedeliverySpec` really did expose an issue with the re-delivery of system messages!

[attempt-sys-message-redelivery-spec.txt](https://github.com/akkadotnet/akka.net/files/8118/attempt-sys-message-redelivery-spec.txt)

Aaronontheweb:
@GraemeBradbury although, looking at it more closely, could be an issue with turning off black-holing

Aaronontheweb:
Even more complete logs: 
[multinode-output.txt](https://github.com/akkadotnet/akka.net/files/8165/multinode-output.txt)

Aaronontheweb:
Looks to me like the throttling system isn't ever able to turn itself off - either that or there's an issue where the `Watch` message that is being redelivered never gets ACKed so the `Terminated` message right behind it never gets delivered either.

PR #1362, opened 2015-10-06
Switching to Paket - fixed edition

Horusiath:
This is fixed version of @forki Paket PR, with updated dependecies for all existing projects.


Aaronontheweb:
Still failing at the MSBuild stage :\

forki:
yes @Horusiath told me to look into this after work.

forki:
granate

Issue #1296, opened 2015-09-07
Port EventBusUnsubscribers


rogeralsing:
In JVM Akka, there is a `EventBusUnsubscribers` (https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/event/EventBusUnsubscribers.scala) 
Which deals with unsubscription from the eventstream for terminated actors.
We don't have that, so if an actor subscribes and then terminates, it will still be registered as a subscriber on the eventstream, thus propagating messages to deadletters.

I am marking this as a bug, because that is what the users are seeing. even if it in reality is more of a "not yet implemented" kind of problem.

maxim-s:
I can try to do it

rogeralsing:
Merged, closing

PR #1363, opened 2015-10-07
#1296 Port EventBusUnsubscribers

maxim-s:


src/core/Akka/Event/EventBusUnsubscriber.cs
```diff
@@ -0,0 +1,113 @@
...
+           
+        }
+
+        protected override bool Receive(object message)
+        {
+            return message.Match().With<Register>(register =>
+            {
+                if (_debug)
+                    _eventStream.Publish(new Debug(this.GetType().Name, GetType(),
+                        $"watching {register.Actor} in order to unsubscribe from EventStream when it terminates"));
```
Horusiath: We still have VS2013 users and compatibility with C# 5 / .NET 4.5. I don't think this is valid notation.

src/core/Akka/Event/EventBusUnsubscriber.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        protected override void PreStart()
+        {
+            if (_debug)
+                _eventStream.Publish(new Debug(this.GetType().Name, GetType(),
+                    $"registering unsubscriber with {_eventStream}"));
+            _eventStream.InitUnsubscriber(Self, _system);
+        }
+
+        internal class Register
```
Horusiath: It's better to keep all messages immutable by convention. Please, don't expose mutable fields in messages.

src/core/Akka/Event/EventStream.cs
```diff
@@ -89,9 +97,83 @@ public override bool Unsubscribe(IActorRef subscriber)
...
+                }
+            }).WasHandled;
+        }
+
+        private void RegisterWithUnsubscriber(IActorRef subscriber)
+        {
+            _initiallySubscribedOrUnsubscriber.Match().With<Left<ISet<IActorRef>>>(v =>
+            {
+                if (_initiallySubscribedOrUnsubscriber.CompareAndSet(v,
+                    Either.Left<ISet<IActorRef>>(new HashSet<IActorRef>(v.Value.Union(new[] {subscriber})))))
```
Horusiath: Immutable set is better choice here, we probably don't want to rebuild a whole `HashSet` on each change. Same concerns line 167.

rogeralsing:
Looks solid to me.
Could you also squash the commits into one? we try to minimize the commit count in PR's

(just in case: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

rogeralsing:
Great job @maxim-s !

PR #1364, opened 2015-10-07
[api-docs] Fixed various x-references

sean-gilliam:
This PR fixes various x-reference issues in the xmldocs.

Most of these are in the form of replacing ```<param name="method parameter"/>``` with ```<paramref name="method parameter"/>``` when included in the ```<summary/>```. 

Sandcastle misinterprets the former when in a summary and puts a placeholder **[!parameter name]** instead of the correct reference *parameter name*. See this for an [example](http://api.getakka.net/docs/stable/html/F9E07D50.htm) of what it does when referencing the *address* parameter.


rogeralsing:
Thanks!

PR #1366, opened 2015-10-08
Renamed internal exceptions to follow .NET standard

### LINKED ISSUE #1192: Exception naming inconsistencies

sean-gilliam:
Closes #1192.

This PR renames various internal exceptions to follow the .NET standard
of suffixing exceptions with "Exception". Since these are internal and
not public it shouldn't break the public API.


rogeralsing:
:+1:

PR #1367, opened 2015-10-08
Named log format

rogeralsing:
This is a PoC for #1358, introducing `LogEvent` format templates.
This is just to get a feel for how a template _could_ look.
I went for the standard .NET string format approach, but with named arguments. 
It is somewhat on the hackish side of things, but you would still have to make an effort to actually break the templates.

Are we OK with this format?

There are still room for improvement in terms of perf and GC, we could easily cache the parsed templates in a concurrent dictionary if translating names turns out to be on the heavy side.

### DIFF
--- a/src/core/Akka/Event/LogEvent.cs
+++ b/src/core/Akka/Event/LogEvent.cs
@@ -6,6 +6,8 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Threading;
 using Akka.Actor;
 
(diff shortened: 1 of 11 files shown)


src/core/Akka/Event/LogEvent.cs
```diff
@@ -67,7 +67,23 @@ protected LogEvent()
...
+        public string ToString(string format)
+        {
+            format = format
+                .Replace("{LogLevel", "{0")
+                .Replace("{Timestamp", "{1")
+                .Replace("{ThreadId", "{2")
+                .Replace("{LogSource", "{3")
+                .Replace("{LogClass", "{4")
+                .Replace("{Message", "{5")
+                .Replace("{CrLf", "{6");
```
rogeralsing: Ok, CrLf might be silly, it could just as well be escaped in the template string

Aaronontheweb:
This is cool - could you add the HOCON config and an example?

rogeralsing:
Are we all in consensus on using named args for the templates then?
if so, I will expand this PR with the template config and specs

Aaronontheweb:
Yep, I'm a fan

rogeralsing:
Here is the output and config to manage this:

logwithms

logtemplate

rogeralsing:
I'll need to add some specs for this before pulling it

Aaronontheweb:
@rogeralsing What do you need to be able to merge in this spec, and will there be any impact on #1295?

rogeralsing:
@Aaronontheweb no, not related to #1295 

The missing bits of this PR is tests, and I'm not quite sure what would be the best way to ensure the behavior of this would be.
Possibly just make a few tests that act upon the `LogEvent` types directly, and some tests that verify that the templates can be read from the hocon file.

I'll try to throw some love at this this weekend.

rogeralsing:
I've updated this PR with some tests for the log events.

annymsMthd:
+1

rogeralsing:
The code have been updated to support a general log format and one for the stdout log.
Both live on the `system.Settings` object.

For other loggers that want to support templating, they should either access the `system.Settings.LogTemplate` or override that in their own settings.

How other loggers handle configuration is omitted here as that is a separate issue not related to how templating is defined or how the `LogEvent` types render themselves as strings.

rogeralsing:
@Aaronontheweb whats up with the build system here?

rogeralsing:
@Aaronontheweb CI is broken, what do I do with this one?

maxim-s:
@rogeralsing http://stackoverflow.com/a/31540373/377476 - use this to update the commit timestamp to now and TeamCity will pick it up and run the usual CI steps. (c) @Aaronontheweb

rogeralsing:
Thanks @maxim-s that seems to have done the trick

rogeralsing:
Should I update this one or should we just close?

Danthar:
Im a fan. So I vote to update.

PR #1373, opened 2015-10-11
Remote PingPong benchmark

rogeralsing:
PingPong benchmark adapted to run two actor systems in the same process using Akka.Remote.
Shows throughput on loopback.

remotepingponglocalhost

### DIFF
--- a/src/Akka.sln
+++ b/src/Akka.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+VisualStudioVersion = 12.0.40629.0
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
This benchmark isn't as useful in Remote due to lack of parallelism on EndpointWriter (network message order preservation comes at a price) but I still think this is very useful to have in the short run

rogeralsing:
A better use for this benchmark might be to alter it to ignore the dispatcher throughput and instead pass different sized messages.
That would give a hint on how message sizes and different serializers affect remote throughput.

Aaronontheweb:
@rogeralsing I like that idea

rogeralsing:
I'm closing this for now, I'll work on it to see if I can get more meaningful data out of it, but until then, closed

PR #1374, opened 2015-10-12
Akka IO Transport: framing support

### LINKED ISSUE #1357: AkkaIOTransport broken
The AkkaIOTransport is broken is several ways.

1. Since the introduction of max frame size exceptions, the IO Transport does not read those settings and thus fails to send any messages
2. Connection back to the calling system seems to fail

I have tried this on the Chat examples in the main repo.

cc @willieferguson

willieferguson:
Fixes #1357: AkkaIOTransport broken


rogeralsing:
Rerunning this a million times as the muntinode stuff bugs out..

PR #1377, opened 2015-10-13
fix #1376 EventStreamUnsubscriber is created under user space

### LINKED ISSUE #1376: EventStreamUnsubscriber and RemoteDeploymentWatcher are created under `/user`
I've noticed that the two new actors EventStreamUnsubscriber and RemoteDeploymentWatcher are created under user space, they should be created in system space.

It should not be more than a change from `ActorOf` to `SystemActorOf`
I noticed that they both interfere with deployment configs as they are matched by deploy wildcards.

cc @maxim-s

maxim-s:
…ated under `/user`


rogeralsing:
:+1:

PR #1379, opened 2015-10-14
Tweaking the visualizer

### LINKED ISSUE #1337: MultiNode Test Run Visualizer: improve time resolution
Make it easier to view the full timeline for the specs themselves by removing / changing the display of the built-in runner messages, which far in the past.

GraemeBradbury:
Replaced passed messages with tick images.
Defaulted display to just the node facts rather than full test.
    (full test is still in timeline but needs to be scrolled to,
    or zoomed out to become visible)
Simple instructions at head of page

resolves #1337


GraemeBradbury:
capture

Example of what changes look like.

rogeralsing:
Love it!!

PR #1380, opened 2015-10-14
Removed F# hack over JSON.NET serializer

Horusiath:
This PR removes an JSON.NET hack hardcoded inside F# API. Also existing end2end discriminated union serialization test between remote nodes has been updated by replacing JSON.NET with Wire.

### DIFF
--- a/src/core/Akka.FSharp.Tests/ApiTests.fs
+++ b/src/core/Akka.FSharp.Tests/ApiTests.fs
@@ -40,13 +40,16 @@ type TestUnion2 =
     | D of int
 
 [<Fact>]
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
So does this mean that Akka.FSharp is using Wire and not JSON.NET now as its active transport, or am I misunderstanding?

Horusiath:
No, it just mean, that the test suite against serializing Discriminated Unions is now run using wire instead of default serializer (which we know, that would fail that test).

GraemeBradbury:
Looks fine to me :shipit:

PR #1381, opened 2015-10-15
Add optional PipeTo completion handlers

emilingerslev:
Added an optional success and failure handler to PipeTo, to simplify use of Tasks. PipeTo falls back to original behavior either handler isn't supplied.

Some examples:

```csharp
Context.ActorSelection("akka.tcp://remote@remote:1234/user/PingActor")
    .ResolveOne(TimeSpan.FromSeconds(1))
    .PipeTo(Self,
        success: x => new PleasePing(x),
        failure: e => new TryAgainLater(e.ToString()));
        
actorRef.Ask<Pong>(new Ping())
    .PipeTo(Self,
        failure: e => new TryAgainLater(e.ToString()));
```

PR #1383, opened 2015-10-20
Add optional PipeTo completion handlers

emilingerslev:
Added an optional success and failure handler to PipeTo, to simplify
use of Tasks. PipeTo falls back to original behavior when not supplied.

Added test for both default handlers and the new handlers.

### DIFF
--- a/src/core/Akka.Tests/Actor/PipeToSupportSpec.cs
+++ b/src/core/Akka.Tests/Actor/PipeToSupportSpec.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Event;
+using Akka.TestKit;
+using Xunit;
+
+namespace Akka.Tests.Actor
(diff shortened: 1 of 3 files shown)


emilingerslev:
Resubmit with tests of PR #1381

emilingerslev:
Test seems to fail due to unrelated MultiNodeTestRunner test

emilingerslev:
The tests still seems to fail due to some recurring failure.

scottt732:
Quick question/possible bug.  In this commit, the `PipeTo` overload shows `TaskContinuationOptions.ExecuteSynchronously | TaskContinuationOptions.AttachedToParent`.  Elsewhere in the docs (https://github.com/petabridge/akkadotnet-code-samples/tree/master/PipeTo), the `&` operator is used.  I'm not sure if this would be an issue w/the docs or with the code.

Aaronontheweb:
@scottt732 see #1347

PR #1388, opened 2015-10-22
Multinode test runner alterations

GraemeBradbury:
Refactored NodeTestRunner to allow MultinodeTestRunner to call it in
two different manners.

`build multinodetests` works in the same manner as in the past. Starting
a seperate process for each node that needs to be created.

Using `build multinodetests spec=<name of spec>` will limit the runner to
starting just the named test. In addition, the nodes will be spun up
inside isolated AppDomains rather than as seperate processes. The final
difference between single test and multi test mode, is that the runner
will call `Debugger.Launch()` to attach VS to the test.

By automatically launching the debugger, I've been able to remove
the `Thread.Sleep()`s that were needed previously, reducing the test run
time by 50%-60%.

**app.config issues**
I did have issues with app.config at various times (Akka.NodeTestRunner uses the Akk.MultinodeTestRunner app config when run in the app domains, and doesn't use
a config file at all then run as a seperate process.)
So i'm not sure how to go forward keeping the binding redirects up to date.


Horusiath:
:+1:

Aaronontheweb:
@GraemeBradbury I'll take a look at the binding redirects issue and a code review of this PR on Friday (traveling at the moment) but I'm excited to see that you've done this! Should make life less frustrating for multi-node test users.

Aaronontheweb:
Taking a look at this now

src/core/Akka.Remote.TestKit/TestSettings.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Runtime.Remoting.Messaging;
+
+namespace Akka.Remote.TestKit
+{
+    /// <summary>
+    ///     Command line argument parser for individual node tests during a <see cref="MultiNodeSpec" />.
+    ///     Parses arguments from <see cref="Environment.GetCommandLineArgs" /> using the same conventions as canonical Akka.
+    ///     For example (from the Akka.NodeTestRunner source):
+    ///     <code>
+    ///     var nodeIndex = CommandLine.GetInt32("multinode.index");
```
Aaronontheweb: Need to update this XML-DOC comment

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -90,60 +106,22 @@ static void Main(string[] args)
...
+                        {
+                            PublishRunnerMessage(string.Format("Skipping test {0}.{1}. Reason - Doesn't match multinode.spec-name.", test.Value.First().TestName, test.Value.First().MethodName));
+                            continue;
+                        }
 
-                        StartNewSpec(test.Value);
+                        PublishRunnerMessage(string.Format("Starting test {0}", test.Value.First().MethodName));
 
-                        foreach (var nodeTest in test.Value)
+                        if (!restrictToTestName)
```
Aaronontheweb: Do you want to pass in a global flag here to specify a strategy for running all tests? I.E. run ALL in AppDomains OR processes? Or is there a technical reason why this might be a bad idea - I.E. unpredictable results across multiple specs as a result of thread contention.

Aaronontheweb:
@GraemeBradbury this PR was very easy to review - left 1 comment and 1 question for you.

PR #1389, opened 2015-10-23
Add log4net logging adapter support

onatm:
Added support for log4net logging. :dart:


rogeralsing:
Looks good.
cc @akkadotnet/developers should we include more contrib loggers in the core repo or should we opt for keeping this externally?

annymsMthd:
I think log4net is a pretty widely used logger and we should probably add it to core.
@onatm can you rebase this PR?

onatm:
@annymsMthd done :+1:

onatm:
I thought a quick rebase was enough but after b293dae377f9f90979a7c5a6819e7891b88395d2 there are so many build errors. I'm trying to fix the problems.

PR #1396, opened 2015-11-03
Implement enumerator for Composite ByteString.

tintoy:
Hi. I ran into this while working with Akka I/O (specifically, concatenated `ByteString`s can't be enumerated). I was going to add a couple of tests for this, but I see that the existing spec for `ByteString` doesn't currently cover IEnumerable at all. I'm wondering if this is because there's a separate "iterator" implemented on `ByteString` (a Scala artefact, perhaps?)...

Anyway, let me know if some tests would be helpful, I don't mind writing them.


tintoy:
Ah, just read the contributor's guide - I should have done this on a feature branch. That was a bit lazy of me. Sorry.

Issue #1403, opened 2015-11-04
Persistent actor receives duplicated events after restart during recovery


Horusiath:
We have some sign, that actor can receive duplicated events, if it restarts during recovery. What's wrong here, is that we have a request for replay events send to journal from actor incarnation A1. Once journal starts to replay events, actor may crash and get restarted for some reason. New incarnation of the actor (let's call it A2) will again send replay request. In result it may receive some of the replayed messages twice (first time those sent for incarnation A1 before restart, and later the same for A2).

patriknw:
That could be a bug (also in Akka). For the writes we have this instanceId to detect such things, see for example here: https://github.com/akka/akka/blob/release-2.3/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala#L119

but that is not done for the replay.

Have you migrated to Akka 2.4? There we have less of this trouble in 2.4 because we unconditionally stop the actor when there are any journal failures, but restarts can still happen due to application exceptions.

ktoso:
(a bit more details on how 2.4 deals with this here: https://github.com/akka/akka/issues/18850#issuecomment-153721647 )

patriknw:
right, it's handled in 2.4: https://github.com/akka/akka/blob/master/akka-persistence/src/main/scala/akka/persistence/Eventsourced.scala#L487

Danthar:
Closing this issue: A PR was merged

thedevd:
I am even facing the same problem with akka 2.6.14, should I raise new ticket

patriknw:
good to create new issue, thanks

Issue #1397, opened 2015-11-03
PersistenceExtension - CreatePlugin in race condition


Zetanova:
PersistenceExtension.SnapshotStoreFor() is creating the plugin with the help of Lazy<PluginHolder>.
But is not using the LazyThreadSafetyMode.ExecutionAndPublication Flag to lock the initializion Thread.

The Plugin SnapshotStore.sql-server will fail with Cause:
Akka.Actor.InvalidActorNameException: Actor name "akka.persistence.snapshot-store.sql-server" is not unique!

All Lazy<> Helper that initialize complex components should force the initializion to only once
with the help of LazyThreadSafetyMode.ExecutionAndPublication

Danthar:
Closing this issue: A PR was merged

PR #1404, opened 2015-11-04
Persistent actor stops on recovery failures

Horusiath:
This PR contains two changes:

1. It attaches `LazyThreadSafetyMode.ExecutionAndPublication` in order to make lazy journal/snapshot store creation more thread safe (see issue #1397).
2. It changes behavior of persistent actor in case of failure during recovery - now it causes actor to stop. This is desired behavior in sync with canonical akka-persistence from version 2.4 (see issue #1403). Also it changes singature of method `OnReplayFailure(Exception reason)` to `OnReplayFailure(Exception reason, object message = null)` - now it will not only pass an exception that caused an error but also responsible event. Another change is that default implementation will log this cause on Error level - therefore all persistent actors contain logging adapters by default.


rogeralsing:
Looks good to me

Issue #1398, opened 2015-11-03
Akka.Persistence.Sql.Common.Journal - JournalDbEngine Not Thread Safe


Zetanova:
LinkedList.Remove() is not Thread safe and will Fail with a NullReferenceException under load.

Horusiath:
The problem lies in the PendingOperations list of journal engine, which is mutated from async tasks themselves, causing race conditions. This part should be redesigned - journal engine state mutations should not be shared or occur inside tasks.

Horusiath:
The problem is existence of `PendingOperations` list itself - it's not necessary for us. Actually what we might want to do is to cleanup all pending database tasks once we are stoping/restarting the journal - for this we need only one shared cancellation token source that will be used for all database calls and canceled at `PostStop` journal method.

Danthar:
Closing this issue: A PR was merged

PR #1407, opened 2015-11-05
Fixed: data races inside sql journal engine

Horusiath:
This PR fixes racing conditions caused by concurrent access to `JournalDbEngine.PendingOperations` list (see #1398). 

Instead of using list, we now have a single CancellationTokenSource per engine instance, that is shared by all tasks created by it. Once journal gets closed (by default when sql journal calls post stop), all pending tasks are cancelled from there.


rogeralsing:
Are these changes guaranteed to be enough to prevent races? cancellation tokens never insta-kills any tasks, it is more of a best effort behavior, sometimes tasks have reached a point of no return.
Just so that we are not camouflaging the problems and they still exists but happens more rarely, which would be even harder to debug later on.

Horusiath:
In general most of akka parts don't care about Futures at all (same history for Akka.NET and Tasks) - once actor dies, all dependent tasks simply gets orphaned. Here it's more like best effort mechanism to try to cancel tasks if it's possible.

Issue #1399, opened 2015-11-03
FAKE Build - unable to push NuGet packages to nightly or production NuGet feeds


Aaronontheweb:
Since #1362 we've not been able to push any NuGet packages to our nightly MyGet feed, because the FAKE script for publishing packages wasn't updated to handle the switch from NuGet to Paket.

See relevant build errors in the log here:

http://petabridge-ci.cloudapp.net/viewLog.html?buildId=2551&buildTypeId=AkkaNet_AkkaNetUnstableNightlyBuilds&tab=buildLog#_focus=10680&guest=1

Horusiath:
@Zetanova reported, that there are some problems with getting .pdb files for nightly builds. Could this be related?

Horusiath:
Reason behind that is, that we are still using nuget package manager for publishing packages. This should be replaced with FAKE [Paket.push](http://fsharp.github.io/FAKE/apidocs/fake-paket.html) command. F# project scaffold does it by default - example [here](https://github.com/Horusiath/Akkling/blob/master/build.fsx#L142).

PR #1422, opened 2015-11-14
Moved from NuGet to Paket for package release process

Horusiath:
This PR changes FAKE build script to Paket for the process of releasing nuget packages. Hopefully this will solve problems described in #1399. Few notes:

- `*.nuspec` files have been replaced with their `*.paket.template` equivalents - most notable difference is notation - paket.template is more YAML style, while nuspec is XML.
- In case when some nuget field is not provided, Paket will try to infere it from project's assembly info.
- Package dependencies are taken directly from `*.csproj`/`*.fsproj` projects themselves.
- Paket doesn't require providing list of files to generate nuget from, instead it discovers all projects with associated paket.template files inside solution and generates nuget packages for them automatically.


Aaronontheweb:
Haven't tested this yet, but I'm not a fan... I don't think Paket has a long lifespan in the context of .NET tooling. It will be killed by DNX. I'm fine using it for restores, but I'd much rather stick with NuGet and the tools we already have. Just re-download the NuGet.exe binary and that's it - why go through all of this trouble?

Aaronontheweb:
Just ran an integration test locally - stills runs into the same issue as #1399.

Horusiath:
Strange. AFAIK Paket doesn't need to use NuGet for publishing packages.

PR #1424, opened 2015-11-16
A logger that logs using the System.Diagnostics.Debugger

Jared314:
During the development of my applications, I've found it very useful for log entries to output to the standard debugger output and break on errors. This has made my development errors easier to find and fix, compared to the other loggers I've tried.

### DIFF
--- a/src/core/Akka/Event/DebugLogger.cs
+++ b/src/core/Akka/Event/DebugLogger.cs
@@ -15,9 +15,11 @@ namespace Akka.Event
     /// </summary>
     public class DebugLogger : ReceiveActor
     {
+        private const string CONFIGURATION_KEY = "akka.logger.debuglogger.break-on-error";
         private readonly ILoggingAdapter _log = Context.GetLogger();
+        private bool breakOnErrorMessage = false;
 
-        private static void Log(LogEvent logEvent)
+        private static void Log(LogEvent logEvent, bool breakOnError)
         {
             System.Diagnostics.Debug.WriteLine(
                 logEvent.ToString(),
@@ -33,7 +35,7 @@ namespace Akka.Event
                         logEvent.ToString());
                 }
 
-                if (logEvent.LogLevel() == LogLevel.ErrorLevel)
+                if (breakOnError && logEvent.LogLevel() == LogLevel.ErrorLevel)
                 {
                     Debugger.Break();
                 }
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
Holy crap this is an amazing idea - cc @akkadotnet/contributors any different opinions on this? Looks very useful to me.

sean-gilliam:
:+1:

danbarua:
Yeah, I would make the break on exceptions behaviour optional, otherwise :+1:

--  
Dan Barua

Aaronontheweb:
@danbarua yep, I agree with that.

@Jared314 would you mind updating this PR to include a HOCON configuration option that toggles whether or not this debugger breaks on exceptions?

Jared314:
@Aaronontheweb I can make it configurable. Can you point me to an example I can use as a base for those modifications?

I only used it in situations where I would want that breaking behavior, like local development and dev environments with remote debugging. And, I replace it in all other environments with the a better logger (e.g. Serilog).

rogeralsing:
@Aaronontheweb no clue on the recursion part, looks OK to me

Jared314:
@Aaronontheweb I've updated this PR to include an option to turn off the breaking behavior.

src/core/Akka/Event/DebugLogger.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        public DebugLogger()
+        {
+            Receive<Error>(m => Log(m, breakOnErrorMessage));
+            Receive<Warning>(m => Log(m, breakOnErrorMessage));
+            Receive<Info>(m => Log(m, breakOnErrorMessage));
+            Receive<Akka.Event.Debug>(m => Log(m, breakOnErrorMessage));
+            Receive<InitializeLogger>(m =>
+            {
+                breakOnErrorMessage = Context.System.Settings.Config.GetBoolean(CONFIGURATION_KEY, true);
```
Aaronontheweb: Looks good to me, @Jared314 

Sorry for being annoying, but could you also write a spec (Akka.Tests) that verifies the:

* Default settings work as expected AND
* Can be successfully overridden.

Purpose of that is more for regression testing, to guard against future changes to this code or the code around it.

rogeralsing:
@Aaronontheweb @Horusiath @akkadotnet/contributors 

How should we handle PR's that are going stale?
Most work is done but lacks final touches.

Should we pull into separate branch and fix the missing bits our selves ? (depending on how important the PR is)

Danthar:
Let start with a definition of a 'stale' PR.
I think any PR on which the Author hasn't responded for more then 1 month can be marked as stale.
On that moment we can reach out to the author via a mention. Then give him another week (one?, two? more?) to respond. 
If he does not we can decide what to do with it.

> Should we pull into separate branch and fix the missing bits our selves ? (depending on how important the PR is)

Short of waiting even longer for the original author, I think if its a PR that merits the effort. Sure. Otherwise, reject the PR.
Might sound a bit harsh, but short of doing everything yourself, its the only way to maintain some level of quality on the PR's. You don't want to end up fostering a culture where its ok not to follow through on a PR because you know someone else will pick up your slack anyways.

My 2 ct

sean-gilliam:
@Danthar I totally agree.

Jared314:
@Danthar I also agree.

Closing due to lack of focused time to update this PR with the required parts and match the current dev branch. (Work got in the way)

Issue #1426, opened 2015-11-16
Race condition with Akka.Persistence FSM


maxim-s:
Looks like we have a race condition with Akka.Persistence FSMs:

http://petabridge-ci.cloudapp.net/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=2672#_focus=13023&guest=1

Aaronontheweb:
Thanks @maxim-s !

maxim-s:
looks like it's actor naming problem in the PersistentFSMSpec

PR #1429, opened 2015-11-17
#1426 fix actor naming in the PersistentFSMSpec 

maxim-s:


Aaronontheweb:
Failing due to other insanity with the build server. At this rate I no longer care - run this build 10 times and it fixes an issue that's failing others.

PR #1440, opened 2015-11-18
DONT MERGE

GraemeBradbury:
Stuff to see what's what with the build hangs.

Please don't merge.

### DIFF
--- a/src/core/Akka.MultiNodeTestRunner.Shared/Persistence/VisualizerPersistentTestRunStore.cs
+++ b/src/core/Akka.MultiNodeTestRunner.Shared/Persistence/VisualizerPersistentTestRunStore.cs
@@ -4,6 +4,7 @@
 //  </copyright>
 // -----------------------------------------------------------------------
 
+using System;
 using System.IO;
 
 using Akka.MultiNodeTestRunner.Shared.Reporting;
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
So I just noticed that the files actually do get written to disk successfully (RDP into the build server to verify) but this call never appears to return. It's like the process for the multinode test runner never exits.

GraemeBradbury:
Did the json/html contain more than 1 message/test?
The entire file gets re-written each message the sink receives.
So I wonder if the something is messing up the output.

Aaronontheweb:
## JSON
[Akka.Cluster.Tests.MultiNode-635834854169362547-JSON.txt](https://github.com/akkadotnet/akka.net/files/38459/Akka.Cluster.Tests.MultiNode-635834854169362547-JSON.txt)

## HTML
[Akka.Cluster.Tests.MultiNode-635834854169362547-HTML.txt](https://github.com/akkadotnet/akka.net/files/38460/Akka.Cluster.Tests.MultiNode-635834854169362547-HTML.txt)

Aaronontheweb:
Nope, only 1 test

GraemeBradbury:
Think this is a race condition in the coordinator.

PR #1451, opened 2015-11-20
Paket targets file fixes

forki:
* Updated bootstrapper
* Set ContinueOnError to true during VS restore process
* Fixed csproj files so that VS builds sln when opening the first time


Aaronontheweb:
@forki could you squash these commits?

forki:
Sure on Monday.

forki:
done

PR #1434, opened 2015-11-18
close #1399 - modified build script to download NuGet for package publishing

Aaronontheweb:
Modifies just the `build.cmd` and `build.sh` files to download Nuget.exe to where it's expected for package publishing.

PR #1452, opened 2015-11-20
Remove NuGet -self restore

forki:
Not sure why this was still in place. AFAIK we use NuGet.CommandLine package to pack new nupkgs.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -2,23 +2,6 @@
 
 pushd %~dp0
 
-SETLOCAL
-SET CACHED_NUGET=%LocalAppData%\NuGet\NuGet.exe
-
-IF EXIST %CACHED_NUGET% goto copynuget
-echo Downloading latest version of NuGet.exe...
-IF NOT EXIST %LocalAppData%\NuGet md %LocalAppData%\NuGet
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
I just added this in because NuGet **publishing** was broken after Paket integration.

Aaronontheweb:
See #1434

forki:
But then we don't need the nuget.commandline package in the dependencies
file.

Aaronontheweb:
@forki I'm not familiar with that - could you show me what you're referring to?

forki:
IIRC we have NuGet.Commandline in the paket.dependencies file. So after
paket restore there will be a nuget.exe in
/packages/NuGet.Commandline/tools

In other words we now have two versions.

Aaronontheweb:
Ah, ok then - well in that case just update this PR to include a modification to the FAKE file here so I can find NuGet.exe on the correct path https://github.com/akkadotnet/akka.net/blob/dev/build.fsx#L59

forki:
Yep will do

forki:
done.

Aaronontheweb:
Make sure you rebase to the latest off of dev - need to have the changes from #1446 included in your PR in order for the multi-node testrunner to pass.

Aaronontheweb:
going to give this a try locally first before merging

PR #1462, opened 2015-11-25
[WIP] Paket revert

Aaronontheweb:
Quick integration test to see if the build server can run the changes.

Already tested NuGet publishing and it appears to work.

Will send a new rebased PR on a separate branch if this goes through.


Aaronontheweb:
looking good looking good...

PR #1463, opened 2015-11-25
removed Paket

Aaronontheweb:
This PR removes Paket and (hopefully) all of the massive .csproj and app.config modifications it made.


Aaronontheweb:
discussed this with @rogeralsing - if Paket doesn't actually make our lives easier, which after a month of usage it does not, then it's getting removed.

forki:
Hallo  b

forki:
Hallo  b

forki:
Hallo  b

rogeralsing:
@forki we gave it a try and we had massive problems, not sure if it was us holding it wrong or what caused it.
We ended up with hundreds of changes in the csproj files, case statements for a bazillion different versions of things. things ended up not compiling, switching from debug to release and vice versa took minutes (I assume that was related to the case statements in the project files(?))

forki:
But did you try to put "framework: net451" on top of dependencies file? It
would have installed only for .NET 4.5.1?

Issue #1410, opened 2015-11-08
ActorTaskScheduler QueueTask for LongRunning


Zetanova:
Some external Sources are lazy initializing background threads with the LongRunning Task hint. 
Like with RX Subject on the first OnNext()
With the current ActorTaskScheduler behavior the actor will lock at Line64 TryExecuteTask(task);
The deault behavir of the TaskPoolScheduler would be to start a detached explicit Thread for this Task.

Zetanova:
The Cancled Check is missing at line65
if (task.IsFaulted || task.IsCanceled)
      Rethrow(task, null);

Danthar:
Can you confirm this bug with a Test/Spec ?
After which, you are most welcome to do a PR for this :)

Zetanova:
I created a Spec, but of couse it is failing.

There are multiple possible options for the ActorTaskScheduler 
1) It creates only a detetched Thread on TaskCreationOptions.LongRunning
2) It uses the CompleteTask-Message only for a TaskCreationOptions.AttachedToParent

Aaronontheweb:
@Zetanova so your spec proves definitively that this is an issue. 

What's the appropriate fix?

Danthar:
@Zetanova your right about the check on line 65. That should include the IsCanceled check.

However i'm not sure what you mean with the possible options for the ActorTaskScheduler. As in, how would that solve the problem?

Zetanova:
Currently the ActorTaskScheduler is executing all task-types with the help of the CompleteTask-Message.
Some Libraries can have hidden side-effects that they are requesting a task internaly
for a long running process.
Like System.Reactive on the first call of OnNext() method.

if currently something is requesting a LongRunning Task it will probably lock the actor with it.

If something requires a Long Running Task for lets say an infinit message loop
they normaly requesting the Task with the TaskCreationOptions.LongRunning flag
and correctly assume that they can run this task for a long period.
A separate and new created thread is used for this task by the TaskPoolScheduler.

I didn't thought it through but maybe the ActorTaskScheduler can behave like the normal TaskPoolScheduler for all Task-types beside  TaskCreationOptions.AttachedToParent
and only using the CompleteTask-Message on tasks that are attached as childs
with the AttachedToParent flag set.

The ActorTaskScheduler schould allways create a new thread for the a LongRunning -Task 
and yes, it is running then outside of the akka context.

Aaronontheweb:
[...]

PR #1470, opened 2015-11-26
ActorTaskScheduler QueueTask for LongRunning #1410

Aaronontheweb:
For @Zetanova

### DIFF
--- a/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
+++ b/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
@@ -243,8 +243,99 @@ namespace Akka.Tests.Dispatch
         }
     }
 
+    public class AsyncLongRunningActor : ReceiveActor
+    {
+        Task longRunning;
+        System.Threading.CancellationTokenSource cts;
+
+        public AsyncLongRunningActor()
+        {
+            Receive<string>(m =>
+            { 
+                if(m.Equals("start"))
+                {
+                    RunTask(() =>
+                    {
+                        cts = new System.Threading.CancellationTokenSource();
+                        //long running task as side effect
+                        longRunning = Task.Factory.StartNew(() =>
+                        {
+                            cts.Token.WaitHandle.WaitOne();
+                            return "done";
+                        }, TaskCreationOptions.LongRunning)
+                        .PipeTo(Sender);
+                    });
+                }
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
@JeffCyr mind taking a look at this issue while you're at it?

JeffCyr:
@Aaronontheweb Sure,

I think the TaskCreationOptions got mixed up here.

TaskCreationOptions.LongRunning is a hint that means:
> Hey scheduler, I'm going to send you a task that may take quite some time to complete, you may want to execute my task on a dedicated thread instead of the ThreadPool to prevent a starvation of your precious threads.

The changes proposed by @Zetanova breaks the ActorTaskScheduler contract (which is to execute tasks in the actor context) by executing the `LongRunning` task with `TaskScheduler.Default` (and potentially starving the ThreadPool) outside the actor context (so the task may execute concurrently with other tasks or with other Receives, this is bad).

Concerning the AttachedToParent option in @Zetanova's commit, I think the flag is mixed up as well. I tried to explain this one in issue #1346. This flag is poorly documented and this is probably due to the fact that it is generally harmful or useless. The proposed change would allow concurrent execution of tasks in the actor context when this flag is specified, so it is bad.

I implemented the LongRunning flag in the refactoring of ActorTaskScheduler included in PR #1484, can you validate that it address your issue appropriately?

rogeralsing:
Completely agree with @JeffCyr, this PR breaks actor concurrency due to `TaskScheduler.Default`

JeffCyr:
I just realized that my implementation of the LongRunning task is useless in PR #1484. I'm starting a thread just to send a msg in the actor mailbox, the task will still be executed on the actor's IScheduler.

I think there is no fix to this issue, the ActorTaskScheduler cannot satisfy the LongRunning hint by design.

Do you think we should throw an exception and refuse to execute LongRunning tasks on an ActorTaskScheduler? I'm not sure what is the best trade-off here, it could work to just ignore the flag too.

rogeralsing:
I think ignoring the flag would be the most correct thing to do, as that is inline with how other schedulers work as you describe.

The fact that actors shouldn't do long running tasks is something that developers/users of Akka.NET need to learn anyway, as the same apply to the synchronous message pipeline.

rogeralsing:
Closing as this breaks actor concurrency..

PR #1471, opened 2015-11-26
made Akka.Persistence.SqlServer LinkedList ThreadSafe

Aaronontheweb:
work done by @zetanova


Horusiath:
Actually in this context, list of pending operations isn't even necessary, as the only thing we use pending operations list for is to cancel all async operations, that work in the background when actor is stopped. This can be done easy using single shared `CancellationTokenSource` as it's in [JournalDbEngine](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/JournalDbEngine.cs#L61) right now.

rogeralsing:
@Horusiath so you are saying that this PR fixes a non issue?

Horusiath:
I've submitted PR #1507 , where I've (hopefully) solved this problem in the simpler manner.

Aaronontheweb:
@Horusiath can this be closed then?

Horusiath:
Fixed in another PR (#1507)

Issue #1481, opened 2015-11-29
Multiple edge-case issues in ActorTaskScheduler


JeffCyr:
I reviewed the ActorTaskScheduler and found multiple subtle issues:

##### The failing message is not set in ReceiveActor.PreRestart
see Actor_PreRestart_should_give_the_failing_message.

##### The actor can restart even if the exception is catched in an async receive
see Actor_should_not_restart_if_exception_is_catched.

##### ActorTaskScheduler can execute task concurrently
see ActorTaskScheduler_reentrancy_should_not_be_possible.
This one is bad, it is possible to execute a task in the ActorTaskScheduler concurrently when the actor is busy.

It's because of a wrong usage of CallContext to flow the state in the ActorTaskScheduler. CallContext is only restored in the Task's body, so it is possible that the CallContext has not yet been set in TaskScheduler.QueueTask.

I created a LinqPad gist to expose the issue: https://gist.github.com/JeffCyr/9c0864c0d3d1b31c7823

rogeralsing:
:+1: well spotted

PR #1482, opened 2015-11-29
Actor task scheduler refactoring

JeffCyr:
This pull request is a refactoring to fix the issues discussed in issue #1481. The major change with the old design is that each actor now have its own (lazy loaded) ActorTaskScheduler. CallContext is no longer required to flow the state so it should achieve better performance.

### DIFF
--- a/src/core/Akka/Actor/ActorCell.cs
+++ b/src/core/Akka/Actor/ActorCell.cs
@@ -30,6 +30,7 @@ namespace Akka.Actor
         private bool _actorHasBeenCleared;
         private Mailbox _mailbox;
         private readonly ActorSystemImpl _systemImpl;
+        private ActorTaskScheduler m_taskScheduler;
 
 
         public ActorCell(ActorSystemImpl system, IInternalActorRef self, Props props, MessageDispatcher dispatcher, IInternalActorRef parent)
@@ -65,6 +66,20 @@ namespace Akka.Actor
         public int NumberOfMessages { get { return Mailbox.NumberOfMessages; } }
         internal bool ActorHasBeenCleared { get { return _actorHasBeenCleared; } }
         internal static Props TerminatedProps { get { return terminatedProps; } }
+
+        public ActorTaskScheduler TaskScheduler
+        {
+            get
+            {
+                var taskScheduler = Volatile.Read(ref m_taskScheduler);
+
+                if (taskScheduler != null)
+                    return taskScheduler;
+
+                taskScheduler = new ActorTaskScheduler(this);
+                return Interlocked.CompareExchange(ref m_taskScheduler, taskScheduler, null) ?? taskScheduler;
+            }
+        }
 
         public void Init(bool sendSupervise, Func<Mailbox> createMailbox /*, MailboxType mailboxType*/) //TODO: switch from  Func<Mailbox> createMailbox to MailboxType mailboxType
         {
--- a/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
+++ b/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.TestKit;
@@ -25,7 +26,7 @@ namespace Akka.Tests.Dispatch
             Receive<string>(async s =>
             {
                 _replyTo = Sender;
-                
+
                 await Task.Delay(TimeSpan.FromMilliseconds(100));
                 SetReceiveTimeout(TimeSpan.FromMilliseconds(100));
             });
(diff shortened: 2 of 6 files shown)


rogeralsing:
Looks good! :+1:

rogeralsing:
Can you squash the commits into one?

JeffCyr:
@rogeralsing Done!

rogeralsing:
Thanks! :+1:

Aaronontheweb:
I'm sorry, but this should be performance tested with a benchmark before it's merged.

> CallContext is no longer required to flow the state so it should achieve better performance.

Probably, but I'll never take a person's word for it. 

Reverting this until we have benchmark data to back this up.

JeffCyr:
Sure I'll create a benchmark

Aaronontheweb:
Thanks @JeffCyr !

rogeralsing:
I do like the design used here.
There is a lot of headache that disappears by removing the `CallContext`.
As far as I can tell, @JeffCyr is correct, executing tasks and flowing async state w/o call context is faster.
That being said, there is a tradeoff by using more memory instead, as each async actor also holds a ref to its own `TaskScheduler`

This will allocate 8 bytes extra for _all_ actors due to the task scheduler reference.
And an unknown amount of memory for actors that use the task scheduler.

I'd be more interested in seeing the memory footprint of async actors than performance in this case as there is nothing affecting the default message pipeline.

But all in all, I'm in favor for this design, it would probably solve the current Mono async problem we have also where the actor Context is null after async operations.

JeffCyr:
I updated my branch to revert the revert and include a benchmark, can we reopen this PR or do I need to create another one?

I added an optional AsyncActor test in the PingPong benchmark `PingPong --async`.

Results shows that the refactored ActorTaskScheduler is 2-3x faster than the original.

Before:
benchmark-without-refactoring

After:
benchmark-with-refactoring

JeffCyr:
@rogeralsing Regarding the memory footprint, there's not much we can do for the 8 extra bytes, but since `ActorCell` already have a lot of fields, it should not make a big difference proportionally.

For actors that are initializing their TaskScheduler, if you check http://referencesource.microsoft.com/#mscorlib/system/threading/Tasks/TaskScheduler.cs,b76a4a6f77962f28

You can see that TaskScheduler is very lightweight (it only have an int field), it's just sad that they add a weakref of every TaskScheduler in a static dictionary (custom hash table) which is only for debugging purpose. They should have used a debugging flag like `Task.s_asyncDebuggingEnabled` does. I will probably create an issue in coreclr about this but we won't see the fix anytime soon.

We could benchmark the memory pressure when millions of actors initialize their TaskScheduler, but I guess it would be that same as keeping a hash table referencing all actors, which would probably be acceptable.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -30,6 +30,7 @@ public partial class ActorCell : IUntypedActorContext, ICell
         private bool _actorHasBeenCleared;
         private Mailbox _mailbox;
         private readonly ActorSystemImpl _systemImpl;
+        private ActorTaskScheduler m_taskScheduler;
```
stefansedich: Any reason for the m_ ?

JeffCyr: Old habits :) I'll fix that

JeffCyr:
I created the [issue 2189](https://github.com/dotnet/coreclr/issues/2189) about TaskScheduler on coreclr.

Here's the benchmark result for the creation of 10 millions TaskScheduler:
> Real TaskScheduler test:
> Memory usage: 420.48 MB
> Total time: 00:00:13.9793632
> 
> Fake TaskScheduler test:
> Memory usage: 148.89 MB
> Total time: 00:00:01.0347952

The memory usage is not that bad and the cpu usage is bad but would probably be amortized by the application without notice.

There is a workaround by creating the ActorTaskScheduler with `FormatterServices.GetUninitializedObject` to bypass the constructor, but that would be a very nasty hack!

Aaronontheweb:
@JeffCyr you'll need to open a new PR

Horusiath:
@JeffCyr concerning your screens from the benchmark, did you noticed serious perf drop in standard actors after the changes?

JeffCyr:
@Horusiath The code path for standard actor is not affected because the ActorTaskScheduler is only created if an actor uses it. My screenshots may have small variations because the payload on my machine was not the same when I did both runs, I ran the benchmark again and got similar results with and without the changes.

rogeralsing:
The reason I brought up the 8 extra bytes earlier is that once the actor reaches a certain size, there will be effects on the CPU mem cache pipeline. 
JVM Akka have worked hard to bring the size down to prevent this. 

I think that the .NET actors are still too big to benefit from any low level optimizations, but there could be potential side effects when altering the size of a raw actor cell.
(I dont think there is in this case, but in theory)

JeffCyr:
If this becomes an issue, we could group some infrequently accessed fields in another class and access them with a level of indirection.

PR #1484, opened 2015-12-01
ActorTaskScheduler refactoring with benchmark

JeffCyr:
This pull request is a refactoring to fix the issues discussed in issue #1481. The major change with the old design is that each actor now have its own (lazy loaded) ActorTaskScheduler. CallContext is no longer required to flow the state so it should achieve better performance.

For the full comment history see Pull Request #1482

### DIFF
--- a/src/benchmark/PingPong/ClientAsyncActor.cs
+++ b/src/benchmark/PingPong/ClientAsyncActor.cs
@@ -1,0 +1,48 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ClientReceiveActor.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+//     Copyright (C) 2013-2015 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Threading.Tasks;
+using Akka.Actor;
+
+#pragma warning disable 1998 //async method lacks an await
+
+namespace PingPong
+{
+    public class ClientAsyncActor : ReceiveActor
+    {
+        public ClientAsyncActor(IActorRef actor, long repeat, TaskCompletionSource<bool> latch)
+        {
+            var received = 0L;
+            var sent = 0L;
+            Receive<Messages.Msg>(async m =>
+            {
+                received++;
+                if (sent < repeat)
+                {
+                    actor.Tell(m);
+                    sent++;
+                }
+                else if (received >= repeat)
+                {
+                    latch.SetResult(true);
+                }
+            });
+            Receive<Messages.Run>(r =>
+            {
+                var msg = new Messages.Msg();
+                for (int i = 0; i < Math.Min(1000, repeat); i++)
+                {
+                    actor.Tell(msg);
+                    sent++;
+                }
+            });
+            Receive<Messages.Started>(s => Sender.Tell(s));
+        }
+    }
+}
+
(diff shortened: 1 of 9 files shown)


Aaronontheweb:
Re-running the build

JeffCyr:
I updated this PR according to the discussion about the LongRunning flag in PR #1470.

Zetanova:
Looks good.

There is no point to have two different behaivios in the execution for AsyncTask().Wait() and RunTask(AsyncTask). If someone want to enforce single-threading then there should be a possibility with an explicit method like AsyncTaskSchudeler.RunTaskSync(AsyncTask)

The problem with a single-threaded implementation is that there are many deathlock problems like
RunTask(() => { AsyncTask().Wait(); }); would allready deathlock. Sometimes it is inside a 3th libary.

The LongRunning behaivior should be equal to Task.Factory.StartNew() with and without 
running it inside of AsyncTaskSchudeler.RunTask()
Sometimes a 3th libary lazy loads an internal message-thread with the LongRunning Flag
it has nothing to do with Akka-Actor. It just wants a non-workpool-thread for long-running.

Zetanova:
Maybe to queue all task of ONE RunTask in the ActorTaskSchudeler with the help of a LinkList and execute them in sequence over a worker-thread that holds the ActorContext. Then there is the possibility to implement TryDequeue() and GetScheduledTasks()

If the first-task resumes  the mailbox there would be a fine position to generate an error or warning if  GetScheduledTasks() is not empty. It would mean that there are still free-floating work-pool-task around.

JeffCyr:
@Zetanova 
This is a **Actor**TaskScheduler not a general purpose scheduler like TaskScheduler.Default.

`ActorTaskScheduler.RunTask(Func<Task> asyncAction)` can only be called within the context of an actor, message processing for that actor will be suspended until `asyncAction` completes and meanwhile all tasks scheduled in that TaskScheduler will be pushed on the captured actor and executed single-threaded.

So yeah you can deadlock if you block inside the TaskScheduler, but that's by design, the same goes for any single-threaded TaskScheduler like the UI thread with TaskScheduler.FromCurrentSynchronizationContext().

I will create an AsyncActorSample to clarify the usage, but the basic scenario is to support async receives in an actor like this:

```c#
public class MyActor : ReceiveActor
{
    public MyActor()
    {
        Receive<SomeMessage>(async m =>
        {
            //Do something
            ...
            await SomethingAsync(); // This release the current thread but message processing is still suspended

            //Do something else (here we are back in the single-threaded context of the actor)
            ...
        }
    }
}
```

The purpose of an async Receive is to perform IO operation without blocking a ThreadPool thread. If we performed blocking IO in a normal Receive, we would potentially starve the ThreadPool and have bad performance.

It's true that some third party library may use Task.Factory.StartNew and Task.ContinueWith without specifying TaskScheduler.Default, so they end up executing arbitrary operation (possibly LongRunning) on TaskScheduler.Current which is bad if its an ActorTaskScheduler. This would be a bug in the third-party library and should be reported and fixed.

If you think that an async call might not follow good practice and hijack the ActorTaskScheduler, you can still `await Task.Run(SomethingAsync)` and you will be safe.

rogeralsing:
This is perf tested and solid I'm pulling this.

stefansedich:
As awesome!

I will test this to see if it fixes the mono issues I was having with
Context and async tomorrow!

JeffCyr:
Great!

@rogeralsing See my last comment in PR #1347, most usage of Task.Factory.StartNew and Task.ContinueWith in Akka should explicitly use TaskScheduler.Default, otherwise some continuations might get pushed back in the calling actor context. The same is true for async methods that await without Task.ConfigureAwait(false).

PR #1487, opened 2015-12-02
Removed wire serializer

rogeralsing:
This PR removes the Wire serializer from core Akka.Actor lib.

I will publish a standalone replacement either as a contrib package or in separate repo, toughts on what would be best?


Horusiath:
Make it a contrib package.

Aaronontheweb:
Yep, push a contrib package :+1:

Aaronontheweb:
Yep, push a contrib package :+1:

PR #1491, opened 2015-12-02
Fixed the sqlite.conf as well as the readme

### LINKED ISSUE #1488: Fix Akka.Persistence.Sqlite default config entry
This is easy, but I want to keep track on it. [Default HOCON config for sqlite](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/persistence/Akka.Persistence.Sqlite/sqlite.conf#L7) contains wrong type name for the journal. Fix it and make sure that all of the type names for journals and snaphot stores are correct.

Silv3rcircl3:
closes #1488


Aaronontheweb:
@Silv3rcircl3 thanks for this! could you squash these two commits together?

Issue #1499, opened 2015-12-04
Supervisorstrategy does not preserve stacktrace


Danthar:
As reported by @stevemesser.
When an unhandled exception gets sent to the Supervisor strategy the stacktrace is not preserved.

This has to do with the way the ActorCell handles faults. It simply rethrows errors.
This can be resolved using the ExceptionDispatchInfo system. 

IMO its very annoying missing out on the stacktrace. Coincidently i've been bit by this yesterday. Ill pick this up myself and make a PR today.

Aaronontheweb:
Resolved by #1500

PR #1500, opened 2015-12-04
Fix 1499

Danthar:
Instead of simply rethrowing an exception we use
ExceptionDispatchInfo.Capture to preserve the stacktrace. This is an  >=
Net 4.5 feature only.

### DIFF
--- a/src/core/Akka/Actor/ActorCell.FaultHandling.cs
+++ b/src/core/Akka/Actor/ActorCell.FaultHandling.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.ExceptionServices;
 using Akka.Actor.Internal;
 using Akka.Dispatch.SysMsg;
 using Akka.Event;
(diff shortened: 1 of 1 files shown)


Danthar:
I know there has been some work to support lower versions of .Net. Since this PR for #1499 contains an 4.5 feature. We might want to look into alternatives ?

I saw some work is done in regards to [backporting](http://stackoverflow.com/questions/20171877/is-there-an-analog-of-exceptiondispatchinfo-in-microsoft-bcl-async) I wonder if mono has an implementation for this class, which we might use.

Danthar:
It seems mono used to have its own implementation which was .net 4 compatible. https://github.com/mono/mono/tree/5.7/mcs/class/corlib/System.Runtime.

Danthar:
To be clear. I could use some feedback on this issue regarding which direction we should take.

Aaronontheweb:
@Danthar For the time being, I think we use this - so it should still work on Mono 4.+ then?

Danthar:
I checked the mono repo. Its been in there for a while 3.* already has it. Only its not until mono 4.0 that the `ifdef .net 4.5` got removed. So in short yes it should work on mono 4.

Aaronontheweb:
Sounds good - I think in terms of doing whatever we need to do support Unity3D.... not a priority right now. There's plenty of other stuff, like the TPL, they don't have access to that's already heavily used in Akka.NET.

PR #1502, opened 2015-12-04
[WIP] Akka.Cluster.Tools & Akka.Cluster.Sharding with tests and examples

Horusiath:
This PR introduces two cluster contrib modules:

- `Akka.Cluster.Tools` gives cluster singleton, distributed topic-based publish/subscribe mechanism and cluster client to allow to communicate with akka cluster from actor systems not joined to the cluster itself. Example can be found under /Examples/Cluster/ClusterTools path. Known issue is that JSON.NET is not able to correctly deserialize `Dictionary<Address, ...>` type, which is used in one of the messages by distributed pub/sub and cluster client.
- `Akka.Cluster.Sharding` gives ability to create sharded on-demand actor creation with automatic rebalancing and handoff on eventual cluster node failures. Example can be found under /Examples/Cluster/ClusterSharding path and uses Akka.Persistence.Sqlite as (required) persistence backend. It used *bin/Debug/store.db* database file for automatic cluster resolution - **it must be removed between separate usages, as cluster nodes doesn't get removed automatically when someone will hard kill the process**.

Additional changes:

- `SerializerWithManifest` class ported from JVM akka-core. 
- Serializer Identifier can be configured through HOCON config. Can be resolved using `SerializerIdentifierHelper` class.

Known issues, that still need to be resolved:

- [ ] Default serializer support for Akka.Cluster.Tools.
- [ ] Making all multinode test runner specs working.
- [ ] There are some known problems with cluster shards rebalancing when nodes containing shards are crashing.


rogeralsing:
Code bomb, impossible to review, but the code is isolated from all other bits so I'm pulling this and I'll manually run tests after merge..

PR #1503, opened 2015-12-05
Add wire back as contrib package + Serialization TestKit

rogeralsing:
Adding Wire back as a contrib package.

I've also added a Akka Serialization Testkit which can be inherited much like the Persistence spec.

The tests for the Wire serializer is implemented using the SerializationTestkit


Horusiath:
:+1:

PR #1504, opened 2015-12-05
[WIP] async/await fixes with MNTR

Aaronontheweb:


src/core/Akka.MultiNodeTestRunner.Shared/Sinks/SinkCoordinator.cs
```diff
@@ -142,13 +142,14 @@ private void InitializeReceives()
...
 
+                var self = Self;
                 foreach (var sink in Sinks)
                 {
                     sink.RequestExitCode(Self);
                     sink.Close(Context.System)
                         .ContinueWith(r => new SinkClosed(),
                         TaskContinuationOptions.ExecuteSynchronously)
-                        .PipeTo(Self);
+                        .PipeTo(self);
```
rogeralsing: Self and Sender can never change in this context, only if used within a closure.
PipeTo is fed with the correct self here.

Aaronontheweb: Good to know, but I was just changing it to closures everywhere rather than leave it up to the next contributor's TPL knowledge.

PR #1505, opened 2015-12-05
[WIP] [Test] Reverting ActorTaskScheduler

Aaronontheweb:
This reverts commit e724d600283862fe78293c46aa62533fcc3a46a6, reversing
changes made to eb2067d2743ff8a9739d170373617c168027923d.

This is just a test - need to see if the consistent failure / deadlock following 1.0.5 with the MNTR was caused by this change.


Aaronontheweb:
This doesn't look like the issue. Still fails just fine.

PR #1511, opened 2015-12-07
prevent nuget from self-upgrading

Aaronontheweb:


Aaronontheweb:
This is a temporary measure until https://github.com/NuGet/Home/issues/1813 is resolved

PR #1515, opened 2015-12-08
decoupled the MNTR from MSBuild

Aaronontheweb:


Aaronontheweb:
Using this to test build changes

PR #1520, opened 2015-12-08
added initial performance specs using NBench

Aaronontheweb:
Uses [NBench](https://github.com/petabridge/nbench) for performing some simple, but automated benchmarks on each build.

Adding this and actually uses it requires me to change the build process on the server, so I wanted to ship a minimal performance testing footprint immediately and expand it later - that way this commit can begin making it into every other contributor's dev branch going forward.

I'll link to some of the results this PR produces....

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,9 @@ bld/
 [Tt]est[Rr]esult*/
 [Bb]uild[Ll]og.*
 
+# NBench results
+[Pp]erf[Rr]esult*/
+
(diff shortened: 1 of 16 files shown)


Aaronontheweb:
Benchmark data: http://petabridge-ci.cloudapp.net/viewLog.html?buildId=3520&buildTypeId=AkkaNet_AkkaNetWindowsPerformanceTests&tab=artifacts&guest=1

Issue #60, opened 2014-03-03
ActorSystemSpec


rogeralsing:
Port the ActorSystemSpec.scala to C#
(Download the Scala source code from https://github.com/akka/akka)

alex-kondrashov:
I*d like to help out here. Is it ok if I work on this one?

rogeralsing:
Absolutely, go for it! :)

alex-kondrashov:
there is ActorSystemSpec.cs already in repository.  However ActorSystemSpec.scala is slightly different. Should ActorSystemSpec.cs be merged/overwritten according to scala*s version or left as it is?

Danthar:
Is this still an active issue ?

Aaronontheweb:
Our version should be updated to include all of the specs found in the Scala version. @alex-kondrashov or @Silv3rcircl3 or @Danthar, would any of you like to:

1. Verify that we've ported all of the specs and close out this issue or
2. Finish porting those specs and then close out the issue?

Silv3rcircl3:
I will give it a try.

Aaronontheweb:
Ty sir

Silv3rcircl3:
I think I need some help here :)
This is what I found

Scala implementation | .Net implementation
------------ | -------------
[use scala.concurrent.Future's InternalCallbackEC](https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/scala/akka/actor/ActorSystemSpec.scala#L158) | I'm not sure if [this](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Tests/Actor/ActorSystemSpec.cs#L93) is the equivalent
[reject invalid names](https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/scala/akka/actor/ActorSystemSpec.scala#L162) | Is implemented [here](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Tests/Actor/ActorSystemSpec.cs#L29)
[allow valid names](https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/scala/akka/actor/ActorSystemSpec.scala#L178) | Is implemented [here](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Tests/Actor/ActorSystemSpec.cs#L46)
[...]

PR #1523, opened 2015-12-09
Implemented the RegisterOnTermination feature.

### LINKED ISSUE #1519: Implement RegisterOnTermination
In the [scala](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/ActorSystem.scala#L376) version it's possible to register callbacks which will be executed after all actors have been terminated. 
This is also referenced in the [Shutdown](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/ActorSystem.cs#L144) comment but not implemented. 

See also ActorSystemSpec #60

Silv3rcircl3:
closes #1519 
It also implements some of the tests for #60.

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
@@ -28,7 +28,7 @@ namespace Akka.Tests.Actor
         [Fact]
         public void AnActorSystemMustRejectInvalidNames()
         {
-            (new List<string> { 
+            new List<string> { 
                   "hallo_welt",
                   "-hallowelt",
                   "hallo*welt",
@@ -36,7 +36,7 @@ namespace Akka.Tests.Actor
                   "hallo#welt",
                   "hallo$welt",
                   "hallo%welt",
-                  "hallo/welt"}).ForEach(n =>
+                  "hallo/welt"}.ForEach(n =>
                   {
                       XAssert.Throws<ArgumentException>(() => ActorSystem.Create(n));
                   });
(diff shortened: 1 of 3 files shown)


src/core/Akka/Util/AtomicReference.cs
```diff
@@ -81,6 +81,18 @@ public bool CompareAndSet(T expected, T newValue)
             return false;
         }
 
+        /// <summary>
+        /// Returns the current value and overrides it with the new one.
+        /// </summary>
+        /// <param name="newValue">The new value</param>
+        /// <returns>The current value</returns>
+        public T GetAndSet(T newValue)
```
Horusiath: This implementation doesn't seems to be atomic.

Silv3rcircl3: Yeah I struggled with this one, thanks for the hint, will update the PR.

Silv3rcircl3: ```Interlocked.Exchange``` unfortunately only allows reference types  in the generic version.
Would this also work?
```
Monitor.Enter(_lock);
var current = Value;
Value = newValue;
Monitor.Exit(_lock);
return current;
```

Aaronontheweb:
Looks great! How does this integrate with the termination hooks for Akka.Remote? Or does this live as a separate system?

https://github.com/akkadotnet/akka.net/blob/efdacf8bf9d179956d5af16269b837fb2dd70d57/src/core/Akka.Remote/RemoteActorRefProvider.cs#L440

If the `RemotingTerminator` needs to be modified, we can do that as a separate issue.

Silv3rcircl3:
Sorry, but through my low understanding about Akka.Remote I can't give an appropriate answer.

Aaronontheweb:
@Silv3rcircl3 ok, no worries Marc - I'll add that as something for me to investigate

Issue #1524, opened 2015-12-10
GuardianSupervisorStrategyConfigurator from Akka is not implemented


akoshelev:
At this time, Akka.NET does not support configuring guardian supervisor strategy: option ```akka.actor.guardian-supervisor-strategy``` is ignored and TODO sections left in the code.

https://github.com/akkadotnet/akka.net/blob/cf7cd6f274cd56af6403afc640a038564dc4d03b/src/core/Akka/Actor/ActorRefProvider.cs#L158

Aaronontheweb:
Closed by #1746

PR #1525, opened 2015-12-10
Support akka.actor.guardian-supervisor-strategy.

akoshelev:
Fixed issue #1524. LocalActorRefProvider now supports option 'akka.actor.guardian-supervisor-strategy'

### DIFF
--- a/src/core/Akka/Actor/ActorRefProvider.cs
+++ b/src/core/Akka/Actor/ActorRefProvider.cs
@@ -135,6 +135,7 @@ namespace Akka.Actor
         private LocalActorRef _userGuardian;    //This is called guardian in Akka
         private Func<Mailbox> _defaultMailbox;  //TODO: switch to MailboxType
         private LocalActorRef _systemGuardian;
+        private readonly ISupervisorStrategyConfigurator _guardianSupervisorStrategyConfigurator;
 
         public LocalActorRefProvider(string systemName, Settings settings, EventStream eventStream)
             : this(systemName, settings, eventStream, null, null)
@@ -155,7 +156,9 @@ namespace Akka.Actor
             _tempNumber = new AtomicCounterLong(1);
             _tempNode = _rootPath / "temp";
 
-            //TODO: _guardianSupervisorStrategyConfigurator = dynamicAccess.createInstanceFor[SupervisorStrategyConfigurator](settings.SupervisorStrategyClass, EmptyImmutableSeq).get
+            var supervisorStrategyType = Type.GetType(settings.SupervisorStrategyClass, true);
+
+            _guardianSupervisorStrategyConfigurator = (ISupervisorStrategyConfigurator)Activator.CreateInstance(supervisorStrategyType);
             _systemGuardianStrategy = SupervisorStrategy.DefaultStrategy;
 
         }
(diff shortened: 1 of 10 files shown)


akoshelev:
I'm a little bit worried about ```Pigeon.conf``` modification as it breaks backward compatibility. May be we should support lowercase syntax for types as well?

rogeralsing:
>I'm a little bit worried about Pigeon.conf modification as it breaks backward compatibility. May be we should support lowercase syntax for types as well?

I dont think this is a real problem as the old value was never used.
Unlikely that anyone have a dependency on a non used value IMO.

src/core/Akka/Actor/ActorRefProvider.cs
```diff
@@ -155,7 +156,14 @@ public LocalActorRefProvider(string systemName, Settings settings, EventStream e
             _tempNumber = new AtomicCounterLong(1);
             _tempNode = _rootPath / "temp";
 
-            //TODO: _guardianSupervisorStrategyConfigurator = dynamicAccess.createInstanceFor[SupervisorStrategyConfigurator](settings.SupervisorStrategyClass, EmptyImmutableSeq).get
+            var supervisorStrategyType = Type.GetType(settings.SupervisorStrategyClass);
+            if (supervisorStrategyType == null)
+            {
+                // TODO: is it ok to throw exception here?
+                throw new ArgumentException(string.Format("Type \"{0}\" is not found.", settings.SupervisorStrategyClass));
```
rogeralsing: We throw `Akka.ConfigurationException` if the source of the value is the HOCON config.
`ArgumentException` is more of a user provided value.

There is also a third option here and use `Type.GetType(settings.SupervisorStrategyClass,true)`
Which will throw one of the following exceptions : https://msdn.microsoft.com/en-us/library/a87che9d(v=vs.110).aspx 

I am leaning towards the latter as it would be the most idiomatic .NET way to deal with missing types

akoshelev:
Thanks @rogeralsing for the remark. I've changed the code, you may review the updated version now.
It seems though, we may support lowercase types in HOCON configuration out of the box, by setting the third parameter of ```GetType``` method to ```true```

```csharp
Type.GetType(settings.SupervisorStrategyClass, true, true)
```
But it's up to you.

Silv3rcircl3:
@akoshelev there are also two missing tests in the ActorSystemSpec #60 about the supervisor strategy, maybe it would be make sense to include them in this PR?

You can find them in the scala version [here](https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/scala/akka/actor/ActorSystemSpec.scala#L322) and [here](https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/scala/akka/actor/ActorSystemSpec.scala#L342).

akoshelev:
@Silv3rcircl3 Yep, I like  the idea. Will do

akoshelev:
@Silv3rcircl3  @rogeralsing @Aaronontheweb I've got one problem with these tests. As I'm creating an actor system inside the test method, I'm not able to use ```EventFilter``` to filter ```Error``` messages - this filter works with ```Sys``` ActorSystem only. So formally tests are incomplete. Could you guys help me find a way to make this work?

Aaronontheweb:
@akoshelev so we can't apply the `EventFilter` to a second `ActorSystem` created in the spec? Am I understanding you correctly?

akoshelev:
@Aaronontheweb Yes, you're right. ```EventFilter``` property of ```TestKit``` instance works with ```Sys``` actor only. This behaviour is defined in ```InternalEventFilterApplier``` that is used by ```EventFilterFactory```. This class uses ```TestKit.Sys``` property to filter events.

```csharp

        public T ExpectOne<T>(Func<T> func)
        {
            return Intercept(func, _testkit.Sys, null, 1);
        }

```
https://github.com/akkadotnet/akka.net/blob/cf7cd6f274cd56af6403afc640a038564dc4d03b/src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs#L60

Aaronontheweb:
@akoshelev in that case, I guess we'll need to add an overload to the `EventFilter` to take an external `ActorSystem`. I had to do the same thing with either the `TestProbe` or something else a while ago. May need to do that on a separate PR.

akoshelev:
@Aaronontheweb Ok, I'll try to add this overload and will update this PR accordingly.

Aaronontheweb:
Implemented by #1746, but we still appreciate #1753!

PR #1513, opened 2015-12-07
Revert "[WIP] Akka.Cluster.Tools & Akka.Cluster.Sharding with tests and examples"

Aaronontheweb:
Reverts akkadotnet/akka.net#1502 - this PR is still a work in progress and was not ready to go into dev just yet.

PR #1530, opened 2015-12-10
Akka.Cluster.Tools & Akka.Cluster.Sharding with tests and examples

Horusiath:
Work in progress. Re-added again after #1502 and #1513 . MNTK cases are there, but still they need some work. I'll create tasks for them in separate issues.


Aaronontheweb:
TAKE THIS, `DEV` BRANCH

Issue #1365, opened 2015-10-08
Write specs to verify Akka.Remote untrusted mode and trusted actorselection paths


Aaronontheweb:
# Enable untrusted mode for full security of server managed actors, prevents
    # system messages to be send by clients, e.g. messages like 'Create',
    # 'Suspend', 'Resume', 'Terminate', 'Supervise', 'Link' etc.
    untrusted-mode = off
    
    # When 'untrusted-mode=on' inbound actor selections are by default discarded.
    # Actors with paths defined in this white list are granted permission to receive actor
    # selections messages. 
    # E.g. trusted-selection-paths = ["/user/receptionist", "/user/namingService"]   
    trusted-selection-paths = []

maxim-s:
I can try

Aaronontheweb:
Thank you sir - check the JVM source to see if there are specs already. I've looked before but can't remember :(

Aaronontheweb:
Do you expect to have a PR in for this soon @maxim-s ? Giving you some alternative issues per #1475 ;)

maxim-s:
@Aaronontheweb yes, it will be done today

maxim-s:
PR #1535

Aaronontheweb:
#1546 had to revert due to race failure in the spec - could you please review @maxim-s ?

Aaronontheweb:
Reposting my comment from there, with links to the build log:

http://petabridge-ci.cloudapp.net/viewLog.html?buildId=3846&tab=buildResultsDiv&buildTypeId=AkkaNet_AkkaNetPrBuild&guest=1

maxim-s:
PR #1552

PR #1535, opened 2015-12-12
#1365 added untrustedspec

maxim-s:
fixed dropping inbound actor selection in the untrusted mode 
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Endpoint.scala#L79

fixed dropping inbound PossiblyHarmful message
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Endpoint.scala#L86

added path to remote daemon
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala#L167

added actor selection by given target

replaced type matching to `if else` , for doing complicated conditions


rogeralsing:
Solid as usual!

Aaronontheweb:
This failed on the dev pr. Looks like the untrusted spec has a race condition within it.

Issue #1522, opened 2015-12-09
Multiple Persistence actors of same type, different names throw Actor name \"akka.persistence.journal.inmem\" is not unique!


MrTortoise:
My scenario involves an AtLeastOnce delivery actor that i am using as a client for 2 different services. 

So i have 2 instances of the same actor type but with 2 different names - they differ as the props passes in the address to use as a path.

In the actor prestart i set up a scheduler to snapshot every second - this causes a cant stash messages more than once error however on actor restart i then get this:
```
GetCandidatePromotion
2015-12-09 12:32:11 [Error] "Can't stash the same message PromotionPlanning.Akka.GuranteedDelivererClientProxy+SnapshotMessages more than once"
Akka.Actor.IllegalActorStateException: Can't stash the same message PromotionPlanning.Akka.GuranteedDelivererClientProxy+SnapshotMessages more than once
   at Akka.Actor.Internal.AbstractStash.Stash()
   at Akka.Persistence.Eventsourced.AroundReceive(Receive receive, Object message)
   at Akka.Actor.ActorCell.ReceiveMessage(Object message)
   at Akka.Actor.ActorCell.Invoke(Envelope envelope)
2015-12-09 12:32:11 [Error] "Can't stash the same message PromotionPlanning.Akka.GuranteedDelivererClientProxy+SnapshotMessages more than once"
Akka.Actor.IllegalActorStateException: Can't stash the same message PromotionPlanning.Akka.GuranteedDelivererClientProxy+SnapshotMessages more than once
   at Akka.Actor.Internal.AbstractStash.Stash()
   at Akka.Persistence.Eventsourced.AroundReceive(Receive receive, Object message)
   at Akka.Actor.ActorCell.ReceiveMessage(Object message)
   at Akka.Actor.ActorCell.Invoke(Envelope envelope)
2015-12-09 12:32:11 [Debug] "Restarting"
2015-12-09 12:32:11 [Debug] "Restarting"
2015-12-09 12:32:11 [Debug] "now supervising akka://clientActorSystem/system/akka.persistence.journal.inmem"
2015-12-09 12:32:11 [Error] "Actor name \"akka.persistence.journal.inmem\" is not unique!"
Akka.Actor.PreRestartException: Exception of type 'Akka.Actor.PreRestartException' was thrown.
2015-12-09 12:32:11 [Error] "Object reference not set to an instance of an object."
[...]

PR #1543, opened 2015-12-13
Fix for race conditions in presistence plugins

Horusiath:
This PR should fix race problems with lazy-loaded persistent plugins (#1522). It turns out that lazy loading isn't actually needed. Also plugins cache has been refactored to work in safer manner.

[EDIT:30.12.2015] **More description**

Actual problem seemed to lay in the way the plugin handlers were registered in `Persistence` extension. Because of concurrent access, all parties (i.e. persistent actors) were trying to lazily initialize plugin handler and `AddOrUpdate` it to concurrent dictionary - this part was handled correctly. However when there are no entry in dictionary, new plugin was initialized and overriding concurrently updated entry, that may appeared in the meantime, as the default dictionary conflict resolution was to take latest plugin created. However one of the plugin handler actions was new journal actor creation - that means, the latest plugin handler would trigger actor creation failure as that actor was already created by earlier plugin handlers already registered.

I think there were two solutions for this problem:

- Use atomic writes + recursive plugin handler retrieving - that is the way, JVM akka followed. It uses basically no locks and that one I've decided to use.
- Use `ConcurrentDictionary.GetOrAdd` method - this also seems to solve an issue, but lack of good example to recreate the problem on the first place seems a more riskier to me. Also turning back from concurrent dictionary into combination of atomic + immutable collection seems to be better option.

### DIFF
--- a/src/core/Akka.Persistence/Persistence.cs
+++ b/src/core/Akka.Persistence/Persistence.cs
@@ -13,6 +13,8 @@ using Akka.Actor;
 using Akka.Configuration;
 using Akka.Dispatch;
 using Akka.Persistence.Journal;
+using Akka.Util;
+using Akka.Util.Internal.Collections;
 
 namespace Akka.Persistence
 {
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
@Horusiath test failures

`MemoryEndToEndAdapterSpec.EventAdapters_in_end_to_end_scenarios_should_allow_using_an_adapter_when_write_was_performed_without_an_adapter`

`MemoryEventAdapterSpec.EventAdapter_should_store_events_after_applying_adapter`

Aaronontheweb:
@Horusiath saw that the tests were re-run - is there a known issue with the `MemoryEndToEndAdapterSpec` and `MemoryEventAdapterSpec`?

Horusiath:
I'd like to know it too. I've re-runed them to check if this are racy tests or repeatable issue associated with changes I've applied.

Horusiath:
As the previous attempt was not 100% thread safe, I've changed it to implementation which is more in line with JVM version - now plugin holders are synced through atomic references on immutable dictionaries.

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -80,13 +82,14 @@ public IActorRef SnapshotStoreFor(string snapshotPluginId)
...
             var configPath = string.IsNullOrEmpty(snapshotPluginId) ? _defaultSnapshotPluginId.Value : snapshotPluginId;
             Lazy<PluginHolder> pluginContainer;
-            if (!_snapshotPluginExtensionIds.TryGetValue(configPath, out pluginContainer))
+            var extensionIdMap = _snapshotPluginExtensionIds.Value;
+            if (!extensionIdMap.TryGet(configPath, out pluginContainer))
             {
-                var plugin = new Lazy<PluginHolder>(() => CreatePlugin(configPath, _ => DefaultPluginDispatcherId), LazyThreadSafetyMode.ExecutionAndPublication);
-                pluginContainer = _snapshotPluginExtensionIds.AddOrUpdate(configPath, plugin, (key, old) => plugin);
+                pluginContainer = new Lazy<PluginHolder>(() => CreatePlugin(configPath, _ => DefaultPluginDispatcherId), LazyThreadSafetyMode.ExecutionAndPublication);
+                _snapshotPluginExtensionIds.CompareAndSet(extensionIdMap, extensionIdMap.AddOrUpdate(configPath, pluginContainer));
```
Aaronontheweb: Looks good to me - immutable copy + CAS

Aaronontheweb:
I think this approach should work. Going to merge it in and give it a try.

Issue #1547, opened 2015-12-14
Race conditions within Akka.Remote.Transport.TestTransport


Aaronontheweb:
```
[20:35:37][Step 1/1] [ERROR][12/14/2015 8:35:38 PM][Thread 0029][[akka://SystemB4/system/endpointManager/reliableEndpointWriter-akka.test%3a%2f%2fSystemA3%40127.0.0.1%3a10-1/endpointWriter]] AssociationError [akka.test://SystemB4@127.0.0.1:10] <- akka.test://SystemA3@127.0.0.1:10: Error [Shut down address: akka.test://SystemA3@127.0.0.1:10] [   at Akka.Remote.EndpointReader.HandleDisassociated(DisassociateInfo info) in D:\BuildAgent\work\49b164d63843fb4\src\core\Akka.Remote\Endpoint.cs:line 1694
[20:35:37][Step 1/1]    at Akka.Actor.UntypedActor.Receive(Object message) in D:\BuildAgent\work\49b164d63843fb4\src\core\Akka\Actor\UntypedActor.cs:line 22
[20:35:37][Step 1/1]    at Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message) in D:\BuildAgent\work\49b164d63843fb4\src\core\Akka\Actor\ActorBase.cs:line 155
[20:35:37][Step 1/1]    at Akka.Actor.ActorCell.ReceiveMessage(Object message) in D:\BuildAgent\work\49b164d63843fb4\src\core\Akka\Actor\ActorCell.DefaultMessages.cs:line 144
[20:35:37][Step 1/1]    at Akka.Actor.ActorCell.Invoke(Envelope envelope) in D:\BuildAgent\work\49b164d63843fb4\src\core\Akka\Actor\ActorCell.DefaultMessages.cs:line 62]
[20:35:37][Step 1/1] Cause: Unknown
[20:35:37][Step 1/1] [ERROR][12/14/2015 8:35:38 PM][Thread 0027][akka://SystemB4/system/endpointManager/reliableEndpointWriter-akka.test%3a%2f%2fSystemA3%40127.0.0.1%3a10-1] Shut down address: akka.test://SystemA3@127.0.0.1:10
[...]

PR #1550, opened 2015-12-15
added dispatcher throughput specs for ThreadPoolDispatcher and ForkJoinDispatcher

Aaronontheweb:


Aaronontheweb:
Warmups don't appear to make much of a difference - I think what I'd need to do to recreate some of the behavior we see in Akka.Remote is not fill up the work queue all at once, but rather intermittently add items to it once every few ms or so... That way we can measure the overhead of the `ConsumingEnumerable` becoming empty / going to sleep / waiting on a `ResetEvent` or whatever it does internally

rogeralsing:
I just noticed that the `backoff-remote-dispatcher` does not cause the per problem, it can run on the dedicated thread pool w/o any (human) observable perf loss

rogeralsing:
blockingqueue

For some reason the blocking queue `MoveNext` is causing a lot of spin waits when used from Akka Remote.

I'm not really seeing the same behavior when using the DedicatedThreadPool manually.

Another interesting thing, if the test is changed from full ping > pong > wait, to a ping * 300000 > pong * 30000 > wait. 
Then the problem goes away, the ForkJoinDispatcher performs as well as the normal Dispatcher under those conditions.

So it seems to be workload related somehow, even if the warm ups or cold starts doesnt point in that direction

rogeralsing:
We should have two different approaches for testing remoting throughput.

1. Low Workload, Full Ping Pong , this is where we are seeing the DedicatedThreadPool issues.
2. High Workload Streaming, here are where we are seeing ActorRef TryParse, and other optimizations

The issues occur under different conditions

Aaronontheweb:
@rogeralsing that's a good idea - let me add a spec that behaves more like this:

> ping * 300000 > pong * 30000 > wait

Aaronontheweb:
Looks like the build was affected by #1547

Aaronontheweb:
Other PR, #1560, includes a fix for this issue

Aaronontheweb:
Rebased this on dev

Aaronontheweb:
cc @JeffCyr what changes should we make to these benchmarks in order to test the `DedicatedThreadPool` changes properly? Any suggestions?

JeffCyr:
@Aaronontheweb The worst case I found with the original Helios ThreadPool was this one:

```c#
static void RunLatencyTest()
{
    const int LatencyIteration = 1000000;
    ManualResetEventSlim evt;
    Stopwatch sw;

    evt = new ManualResetEventSlim(false);
    WaitCallback callback = state =>
    {
        evt.Set();
    };

    sw = Stopwatch.StartNew();

    for (int i = 0; i < LatencyIteration; i++)
    {
        ThreadPool.UnsafeQueueUserWorkItem(callback, null);
        evt.Wait();
        evt.Reset();
    }

    Console.WriteLine("Global ThreadPool : " + sw.Elapsed);

    Action action = () => evt.Set();

    DedicatedThreadPool heliosThreadPool = new DedicatedThreadPool(new DedicatedThreadPoolSettings(Environment.ProcessorCount));

    sw = Stopwatch.StartNew();

    for (int i = 0; i < LatencyIteration; i++)
    {
        heliosThreadPool.QueueUserWorkItem(action);
        evt.Wait();
        evt.Reset();
    }

    Console.WriteLine("HeliosThreadPool  : " + sw.Elapsed);
}
```

Aaronontheweb:
I'll add a benchmark for that on the `DedicatedThreadPool` repository: https://github.com/helios-io/DedicatedThreadPool

JeffCyr:
Ok, you want me to create a pull request for the refactoring on that repo now or should I wait until it has been reviewed here?

Aaronontheweb:
Go ahead and have it reviewed here first - more people paying attention to this repo than that one

rogeralsing:
Looks good, is this one ready for merge? (not sure if the conversation implies there are other things going on that needs to be done first)

Aaronontheweb:
@rogeralsing this is ready for merge.

PR #1552, opened 2015-12-15
#1365 added untrustedspec

maxim-s:
fixed dropping inbound actor selection in the untrusted mode 
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Endpoint.scala#L79

fixed dropping inbound PossiblyHarmful message
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Endpoint.scala#L86

added path to remote daemon
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala#L167

added actor selection by given target

replaced type matching to if else , for doing complicated conditions

### DIFF
--- a/src/core/Akka.Remote/RemoteActorRefProvider.cs
+++ b/src/core/Akka.Remote/RemoteActorRefProvider.cs
@@ -45,13 +45,16 @@ namespace Akka.Remote
 
         private Internals RemoteInternals
         {
-            get
-            {
-                return _internals ??
-                       (_internals =
-                           new Internals(new Remoting(_system, this), _system.Serialization,
-                               new RemoteSystemDaemon(_system, RootPath / "remote", SystemGuardian, _remotingTerminator, _log)));
-            }
+            get { return _internals ?? (_internals = CreateInternals()); }
+        }
+
+        private Internals CreateInternals()
+        {
+            var internals =
+                new Internals(new Remoting(_system, this), _system.Serialization,
+                    new RemoteSystemDaemon(_system, RootPath/"remote", SystemGuardian, _remotingTerminator, _log));
+            _local.RegisterExtraName("remote", internals.RemoteDaemon);
+            return internals;
         }
 
         public IInternalActorRef RemoteDaemon { get { return RemoteInternals.RemoteDaemon; } }
--- a/src/core/Akka/Actor/ActorRefFactoryExtensions.cs
+++ b/src/core/Akka/Actor/ActorRefFactoryExtensions.cs
@@ -14,6 +14,17 @@ namespace Akka.Actor
             return factory.ActorOf(Props.Create<TActor>(), name: name);
         }
 
+        /// <summary>
+        ///     Construct an <see cref="Akka.Actor.ActorSelection"/> from the given string representing a path
+        ///     relative to the given target. This operation has to create all the
+        ///     matching magic, so it is preferable to cache its result if the
+        ///     intention is to send messages frequently.
+        /// </summary>
+        public static ActorSelection ActorSelection(this IActorRefFactory factory, IActorRef anchorRef, string actorPath)
+        {
+            return ActorRefFactoryShared.ActorSelection(anchorRef, actorPath);
+        }
+
     }
 }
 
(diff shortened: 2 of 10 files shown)


Aaronontheweb:
Perf issues will be fixed by #1560 once that's merged

Aaronontheweb:
@maxim-s could you rebase this on dev? should resolve the issue with the failing spec

maxim-s:
@Aaronontheweb done

rogeralsing:
Looks good but unclear (to me) why the F# API changes are for?

Aaronontheweb:
@maxim-s @Horusiath what's up with the F# API changes in this pr, as @rogeralsing asked?

maxim-s:
@rogeralsing @Horusiath @Aaronontheweb I've changed F# API because, one new method   ```ActorSelection ActorSelection(IActorRef anchorRef, string actorPath);``` has been added into interface ```IActorRefFactory.cs``` and ```ActorSystem``` scala akka has this method
https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/ActorSelection.scala#L150

Aaronontheweb:
@rogeralsing @Horusiath that interface change won't create any issues will it? I think it's safe to add new members but would love to get a second opinion on that.

Otherwise, I'm good to go with this PR. Ready for merge as soon as that question is answered.

rogeralsing:
@Aaronontheweb the interface change will break any user defined class that implements the interface, which is the reason MS is moving away from interfaces in vNext in favor for abstract classes, as you can add new members with a default impl.

That being said, I would be extremely surprised if anyone have implemented their own custom actorrreffactory. 
So while it is not safe from a binary perspective, I do think it is safe from a "users wouldnt do that anyway" perspective.

maxim-s:
I can implement it wothout breaking changes

maxim-s:
@Aaronontheweb @rogeralsing Updated without changing interface

Aaronontheweb:
@maxim-s build server didn't pick it up due to the older commit date - could you please rebase 1 more time and run this command? This will refresh the timestamp and cause the build server to detect it:

http://stackoverflow.com/a/31540373/377476

Aaronontheweb:
Nevermind Maxim, looks like this was an issue with Azure.

Apparently our Azure cloud service we use for auto-scaling build agents was permanently faulted by something that occurred at midnight 12/31. Switched to a new cloud service and everything is running correctly again.

maxim-s:
@Aaronontheweb updated

src/core/Akka/Actor/ActorRefFactoryExtensions.cs
```diff
@@ -14,6 +14,17 @@ public static IActorRef ActorOf<TActor>(this IActorRefFactory factory, string na
...
 
+        /// <summary>
+        ///     Construct an <see cref="Akka.Actor.ActorSelection"/> from the given string representing a path
+        ///     relative to the given target. This operation has to create all the
+        ///     matching magic, so it is preferable to cache its result if the
+        ///     intention is to send messages frequently.
+        /// </summary>
+        public static ActorSelection ActorSelection(this IActorRefFactory factory, IActorRef anchorRef, string actorPath)
+        {
+            return ActorRefFactoryShared.ActorSelection(anchorRef, actorPath);
```
Aaronontheweb: Brand new extension method - not a breaking change. Should be fine.

src/core/Akka.Remote/RemoteActorRefProvider.cs
```diff
@@ -45,13 +45,16 @@ public RemoteActorRefProvider(string systemName, Settings settings, EventStream
...
-            }
+            get { return _internals ?? (_internals = CreateInternals()); }
+        }
+
+        private Internals CreateInternals()
+        {
+            var internals =
+                new Internals(new Remoting(_system, this), _system.Serialization,
+                    new RemoteSystemDaemon(_system, RootPath/"remote", SystemGuardian, _remotingTerminator, _log));
+            _local.RegisterExtraName("remote", internals.RemoteDaemon);
```
Aaronontheweb: @maxim-s what's this line for?

Aaronontheweb:
@maxim-s looks good - could you rebase this on `dev` and follow the procedure for making modifications to the public API?

http://getakka.net/docs/akka-developers/public-api-changes

Also left a couple of comments for you

src/core/Akka.Remote/RemoteActorRefProvider.cs
maxim-s: @Aaronontheweb system name for provider
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala#L167

maxim-s:
@Aaronontheweb updated

PR #1558, opened 2015-12-15
[api-docs] Added more routing documentation

sean-gilliam:
This PR adds more Akka.Routing documentation in the form of listeners,
routers and router messages.


Aaronontheweb:
Thanks @sean-gilliam - great work as always.

PR #1559, opened 2015-12-16
fixed bug with AddressSpec that caused it not to actually measure Address.Parse operations

Aaronontheweb:
Whoops - forgot to actually call the counter we use to instrument this spec.

### DIFF
--- a/src/core/Akka.Tests.Performance/Actor/AddressSpec.cs
+++ b/src/core/Akka.Tests.Performance/Actor/AddressSpec.cs
@@ -25,6 +25,7 @@ namespace Akka.Tests.Performance.Actor
         public void ParseLocalThroughput(BenchmarkContext context)
         {
             Address.Parse("akka.tcp://sys@localhost:9091");
+            _parseThroughput.Increment();
         }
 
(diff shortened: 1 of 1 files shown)


src/core/Akka.Tests.Performance/Actor/AddressSpec.cs
```diff
@@ -25,6 +25,7 @@ public void Setup(BenchmarkContext context)
         public void ParseLocalThroughput(BenchmarkContext context)
         {
             Address.Parse("akka.tcp://sys@localhost:9091");
+            _parseThroughput.Increment();
```
Aaronontheweb: Calls the NBench `Counter` object that's used to measure the throughput of the `Address.Parse` function, which I omitted by accident before.

src/core/Akka.Tests.Performance/Actor/AddressSpec.cs
```diff
@@ -33,6 +34,7 @@ public void ParseLocalThroughput(BenchmarkContext context)
         public void ParseRemoteThroughput(BenchmarkContext context)
         {
             Address.Parse("akka://sys");
+            _parseThroughput.Increment();
```
Aaronontheweb: Calls the NBench `Counter` object that's used to measure the throughput of the `Address.Parse` function, which I omitted by accident before.

PR #1540, opened 2015-12-13
Akka.Remote Performance - String.Format logging perf fix

rogeralsing:
Fixes #1539 

The log message is being formatted in the default pipeline, even if logging is not enabled. 
This PR moves the message formatting into the conditionals that logs.

PR #1561, opened 2015-12-16
added an independent performance spec for InboundMessageDispatcher

Aaronontheweb:
Makes it possible to test performance improvements made to the read side of Akka.Remote independently from the rest of the message processing pipeline. Great for changes such as #1540.

### DIFF
--- a/src/core/Akka.Remote.Tests.Performance/Akka.Remote.Tests.Performance.csproj
+++ b/src/core/Akka.Remote.Tests.Performance/Akka.Remote.Tests.Performance.csproj
@@ -30,6 +30,14 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Google.ProtocolBuffers, Version=2.4.1.521, Culture=neutral, PublicKeyToken=55f7125234beb589, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Google.ProtocolBuffers.2.4.1.521\lib\net40\Google.ProtocolBuffers.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Google.ProtocolBuffers.Serialization, Version=2.4.1.521, Culture=neutral, PublicKeyToken=55f7125234beb589, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Google.ProtocolBuffers.2.4.1.521\lib\net40\Google.ProtocolBuffers.Serialization.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="NBench, Version=0.1.5.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\..\packages\NBench.0.1.5\lib\net45\NBench.dll</HintPath>
       <Private>True</Private>
@@ -47,6 +55,7 @@
     <Compile Include="Transports\TestTransportAssociationStressSpec.cs" />
     <Compile Include="Transports\TestTransportRemoteMessagingThroughputSpec.cs" />
     <Compile Include="ThreadPoolDispatcherRemoteMessagingThroughputSpec.cs" />
+    <Compile Include="InboundMessageDispatcherSpec.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Akka.Remote\Akka.Remote.csproj">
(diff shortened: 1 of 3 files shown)


src/core/Akka.Remote.Tests.Performance/Akka.Remote.Tests.Performance.csproj
```diff
@@ -30,6 +30,14 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Google.ProtocolBuffers, Version=2.4.1.521, Culture=neutral, PublicKeyToken=55f7125234beb589, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Google.ProtocolBuffers.2.4.1.521\lib\net40\Google.ProtocolBuffers.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="Google.ProtocolBuffers.Serialization, Version=2.4.1.521, Culture=neutral, PublicKeyToken=55f7125234beb589, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Google.ProtocolBuffers.2.4.1.521\lib\net40\Google.ProtocolBuffers.Serialization.dll</HintPath>
```
Aaronontheweb: Had to add a reference to Protobuffs here since we're going to be manually constructing ones to test `IInboundDispatcher`

src/core/Akka.Remote.Tests.Performance/InboundMessageDispatcherSpec.cs
```diff
@@ -0,0 +1,97 @@
...
+        private ActorSystem _actorSystem;
+        private Address _systemAddress;
+        private IInboundMessageDispatcher _dispatcher;
+
+        private IInternalActorRef _targetActorRef;
+
+        /// <summary>
+        /// Not thread-safe, but called by a single thread in the benchmark
+        /// </summary>
+        private class BenchmarkActorRef : MinimalActorRef
```
Aaronontheweb: Using a `MinimalActorRef` implementation instead of an actor class so we can test this code in a single-threaded manner and not use any dispatchers.

src/core/Akka.Remote.Tests.Performance/InboundMessageDispatcherSpec.cs
```diff
@@ -0,0 +1,97 @@
...
+              }
+        ");
+
+        [PerfSetup]
+        public void Setup(BenchmarkContext context)
+        {
+            _actorSystem = ActorSystem.Create("MessageDispatcher" + Counter.GetAndIncrement(), RemoteHocon);
+            _systemAddress = RARP.For(_actorSystem).Provider.DefaultAddress;
+            _inboundMessageDispatcherCounter = context.GetCounter(MessageDispatcherThroughputCounterName);
+            _message = SerializedMessage.CreateBuilder().SetSerializerId(0).SetMessage(ByteString.CopyFromUtf8("foo")).Build();
```
Aaronontheweb: Manually construct a `ByteString` using protobuffs here. Going to use the same message every time on the test.

src/core/Akka.Remote.Tests.Performance/InboundMessageDispatcherSpec.cs
```diff
@@ -0,0 +1,97 @@
...
+            _dispatcher = new DefaultMessageDispatcher(_actorSystem, RARP.For(_actorSystem).Provider, _actorSystem.Log);
+            _targetActorRef = new BenchmarkActorRef(_inboundMessageDispatcherCounter);
+        }
+
+        [PerfBenchmark(Description = "Tests the performance of the Default", RunMode = RunMode.Throughput, NumberOfIterations = 13, TestMode = TestMode.Measurement)]
+        [CounterMeasurement(MessageDispatcherThroughputCounterName)]
+        public void DispatchThroughput(BenchmarkContext context)
+        {
+            _dispatcher.Dispatch(_targetActorRef, _systemAddress, _message);
```
Aaronontheweb: This is a throughput spec, so this one line will be executed continuously until roughly a second passes - might be shorter or longer than that depending on how long it took the warmups to run.

src/core/Akka.Remote.Tests.Performance/InboundMessageDispatcherSpec.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        [PerfCleanup]
+        public void Cleanup()
+        {
+            _actorSystem.Shutdown();
```
Aaronontheweb: Needed an `ActorSystem` in order to satisfy the constructor requirements of the `DefaultMessageDispatcher`, and it also had to have Akka.Remote enabled with a transport since the `DefaultMessageDispatcher` depends on being able to get a reference to a `RemoteActorRefProvider` internally.

rogeralsing:
Looks good, no remarks

PR #1562, opened 2015-12-16
Player closure fix

rogeralsing:
This PR fixes a non closed over Self reference in the Remote.TestKit Player.
`Self` is not safe to use in continuations as the `Self` property is pointing to the currently active context.

From ActorBase.cs
```csharp
protected IActorRef Self { get { return HasBeenCleared ? _clearedSelf : Context.Self; } }

protected static IActorContext Context
{
      get
      {
         var context = InternalCurrentActorCellKeeper.Current;
         if (context == null)
            throw new NotSupportedException(
                        "There is no active ActorContext, this is most likely due to use of async operations from within this actor.");

         return context.ActorHasBeenCleared ? null : context;
     }
}
```


Aaronontheweb:
Failure caused by the `RemoteRoundRobinSpec`, which is the usual suspect for failures

PR #1565, opened 2015-12-16
Remove string interpolation and replace with string format

### LINKED ISSUE #1563: Solution does no longer build from Dev
In `TestTransportRemoteMessagingThroughputSpec`

There is a line:

```csharp
var remoteAddress = $"test://{actorSystemName}@{ipOrHostname}:{port}";
```

Which uses C#6 string interpolation, while the project itself is set to "Default" on language level.

rogeralsing:
Fixes #1563

Removes C#6 code from Akka.Remote.Tests.Performance


Aaronontheweb:
@rogeralsing is this still necessary given what we discussed at the contributor's meeting yesterday?

PR #1568, opened 2015-12-17
[WIP] MultiNodeTestRunner UDP logging system

### LINKED ISSUE #1475: Create a Tcp Server/Clientfor the MultiNode Testkit
As discussed on the last Core dev meeting, we should replace the current StdOut logging for MultiNode Testkit to instead log using a custom logger that sends all events over Tcp using Akka.IO.

So we need a Tcp server that aggregates all the events, and for each node we need a custom logger that can connect and talk  to the server.

@Aaronontheweb and @Horusiath has more details about this

Aaronontheweb:
Resolves #1475 - should preserve full stack traces.


src/core/Akka.Remote.TestKit/TcpLogWriter.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        /// <summary>
+        /// Initializes the <see cref="TcpLogWriter"/> using arguments passed into
+        /// the multi-node test runner via <see cref="CommandLine"/>
+        /// </summary>
+        protected override void PreStart()
+        {
+            try
+            {
+                _udpLogWriter = Context.ActorOf(Props.Create(() => new UdpLogger()), "udpLogWriter");
```
Silv3rcircl3: shouldn't this be ```new TcpLogWriter()``` ?

Aaronontheweb:
I don't think this approach will work >_<

PR #1570, opened 2015-12-17
[api-docs]Enabled missing xmldoc output option

sean-gilliam:
Some of the newer assemblies forgot to enable this option when they were
merged in. This PR enables that option so that Sandcastle can generate
their respective documentation pages.

Also some of the newer namespaces were missing namespace documentation,
so this PR adds those docs as well.


documentation/akkadoc.shfbproj
```diff
@@ -186,6 +215,30 @@
         The Akka.Persistence.TestKit.Snapshot namespace contains classes that provide base testing functionality used to test
         snapshotting in new persistence extensions.
       </NamespaceSummaryItem>
+      <NamespaceSummaryItem name="Akka.Persistence.Sql.Common" isDocumented="True">
+        The Akka.Persistence.Sql.Common namespace contains common classes used by the various SQL/NoSQL persistence extensions.
```
rogeralsing: > SQL/NoSQL

cc @Horusiath , the Sql.Common is only for the Sql based DB's, right?

Horusiath: @rogeralsing right. It's used only for SQL stores, as they share many properties.

sean-gilliam: I thought NoSQL used it as well. My bad. I'll change it.

PR #1583, opened 2015-12-18
Enabling some tests where the problem have been solved.

rogeralsing:
There are a few issues that have been fixed that makes it possible to re-enable these tests.
Not sure about the status of ManageLogLevels, it runs good on local machine. going to give it a few runs on the build server.

### DIFF
--- a/src/core/Akka.Tests/Actor/FSMTimingSpec.cs
+++ b/src/core/Akka.Tests/Actor/FSMTimingSpec.cs
@@ -194,7 +194,7 @@ namespace Akka.Tests.Actor
         /// <summary>
         /// receiveWhile is currently broken
         /// </summary>
-        [Fact(Skip = "receiveWhile is currently broken")]
+        [Fact]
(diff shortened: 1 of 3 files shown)


src/core/Akka.Remote.Tests/Transport/AkkaProtocolStressTest.cs
```diff
@@ -178,7 +178,7 @@ public AkkaProtocolStressTest() : base(AkkaProtocolStressTestConfig)
 
         #region Tests
 
-        [Fact(Skip = "fails due to out-of-order processing as a result of Helios eventing")]
+        [Fact]
         public void AkkaProtocolTransport_must_guarantee_at_most_once_delivery_and_message_ordering_despite_packet_loss()
         {
```
rogeralsing: This still fails, wasn't message order for Helios solved some time ago? (fixed by limiting worker threads IIRC?)
Will revert this change if this is still an issue

cc @Aaronontheweb

Horusiath:
:+1:

PR #1584, opened 2015-12-18
Enabling bounded dequeue stash and receive timeout test

rogeralsing:
Enabling some tests that previously have been set to Skip due to broken or not implemented code.

### DIFF
--- a/src/core/Akka.TestKit.Tests/TestActorRefTests/TestActorRefSpec.cs
+++ b/src/core/Akka.TestKit.Tests/TestActorRefTests/TestActorRefSpec.cs
@@ -143,11 +143,11 @@ namespace Akka.TestKit.Tests.TestActorRefTests
             Assert.Equal("Hejsan!", actor.ReceivedString);
         }
 
-        [Fact(Skip = "Context.ReceiveTimeout is not implemented. Test cannot run")]
+        [Fact]
         public void TestActorRef_must_set_ReceiveTimeout_to_None()
         {
             var a = new TestActorRef<WorkerActor>(Sys, Props.Create<WorkerActor>());
-            //TODO: When Context.ReceiveTimeout is implemented: Assert.Equal(((IInternalActor)a.UnderlyingActor).ActorContext.ReceiveTimeout, not sure what value to put here: null or Timeout.InfiniteTimeSpan);
+            ((IInternalActor)a.UnderlyingActor).ActorContext.ReceiveTimeout.ShouldBe(null);
         }
 
         [Fact]
--- a/src/core/Akka.Tests/Actor/StashMailboxSpec.cs
+++ b/src/core/Akka.Tests/Actor/StashMailboxSpec.cs
@@ -32,13 +32,12 @@ namespace Akka.Tests.Actor
             intRef.Cell.Mailbox.GetType().ShouldBe(typeof(UnboundedDequeBasedMailbox));
         }
 
-        [Fact(Skip = "We do not have a BoundedDequeBasedMailbox yet")] //TODO: Remove Skip when we have a BoundedDequeBasedMailbox
+        [Fact]
         public void When_creating_actor_marked_with_WithBoundedStash_a_mailbox_which_supports_unbounded_stash_is_created()
         {
             var actorRef = ActorOf<BoundedStashActor>();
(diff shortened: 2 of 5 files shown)


src/core/Akka.Tests/Actor/StashMailboxSpec.cs
```diff
@@ -32,13 +32,12 @@ public void When_creating_actor_marked_with_WithUnboundedStash_a_mailbox_which_s
...
 
-        [Fact(Skip = "We do not have a BoundedDequeBasedMailbox yet")] //TODO: Remove Skip when we have a BoundedDequeBasedMailbox
+        [Fact]
         public void When_creating_actor_marked_with_WithBoundedStash_a_mailbox_which_supports_unbounded_stash_is_created()
         {
             var actorRef = ActorOf<BoundedStashActor>();
             var intRef = (LocalActorRef)actorRef;
-            //intRef.Cell.Mailbox.GetType().ShouldBe(typeof(BoundedDequeBasedMailbox));
-            throw new Exception("Incomplete. Remove the comment on the line above this, and remove this line, when we have BoundedDequeBasedMailbox");
+            intRef.Cell.Mailbox.GetType().ShouldBe(typeof(BoundedDequeBasedMailbox));
```
rogeralsing: The BoundedDequeBasedMailbox have been ported for quite some time ago.
Test now passes.

src/core/Akka/Actor/ActorCell.ReceiveTimeout.cs
```diff
@@ -19,12 +19,20 @@ public void SetReceiveTimeout(TimeSpan? timeout=null)
...
         }
 
+        public TimeSpan? ReceiveTimeout
+        {
+            get
+            {
+                return _receiveTimeoutDuration;
+            }
+        }
+
```
rogeralsing: ReceiveTimeout getter is available in the Scala code base (together with SetReceiveTimeout) so we are inline with them.
This property is needed in the `TestActorRef_must_set_ReceiveTimeout_to_None` test (also ported from Scala)

src/core/Akka.TestKit.Tests/TestActorRefTests/TestActorRefSpec.cs
```diff
@@ -143,11 +143,11 @@ public void TestActorRef_must_allow_access_to_internals()
             Assert.Equal("Hejsan!", actor.ReceivedString);
         }
 
-        [Fact(Skip = "Context.ReceiveTimeout is not implemented. Test cannot run")]
+        [Fact]
         public void TestActorRef_must_set_ReceiveTimeout_to_None()
         {
             var a = new TestActorRef<WorkerActor>(Sys, Props.Create<WorkerActor>());
-            //TODO: When Context.ReceiveTimeout is implemented: Assert.Equal(((IInternalActor)a.UnderlyingActor).ActorContext.ReceiveTimeout, not sure what value to put here: null or Timeout.InfiniteTimeSpan);
+            ((IInternalActor)a.UnderlyingActor).ActorContext.ReceiveTimeout.ShouldBe(null);
```
rogeralsing: This test is from the Scala tests.

src/core/Akka/Actor/IActorContext.cs
```diff
@@ -135,6 +135,11 @@ public interface IActorContext : IActorRefFactory, ICanWatch
         /// <param name="timeout">The timeout. Pass in <c>null</c> to switch off this feature.</param>
         void SetReceiveTimeout(TimeSpan? timeout);
 
+        /// <summary>
+        /// Gets the inactivity deadline timeout set using <see cref="SetReceiveTimeout"/>.
+        /// </summary>
+        TimeSpan? ReceiveTimeout { get; }
```
rogeralsing: This does not version well, Adding members to interfaces.
But..
It is a member in Scala and thus belongs here. and there is currently no way of checking if receive timeout has been set to a value.
It is also very unlikely that there are custom implementations of the interface out there that does not inherit `ActorCell`.

Thoughts on how to handle this? there are probably more places where we haven't yet fully ported the Scala code and need to add more members to support new features.

Horusiath:
Looks good for me.

PR #1587, opened 2015-12-19
Enabling AtLeastOnceDelivery_should_not_send_when_actor_crashes

rogeralsing:
cc @Horusiath 
This test now passes. 
The description of the Skip message was only "Fix me"... was this disabled due to the test failing or due to racy-ness?


rogeralsing:
Apparently it doesnt workon the build server..

Horusiath:
It was probably fixed by one of the changes made in the past. If it works now, it's great :)

PR #1588, opened 2015-12-19
Fixes PersistentView_should_run_updates_again_on_failure_during_an_update_cycle test

rogeralsing:
This PR fixes the previously skipped test PersistentView_should_run_updates_again_on_failure_during_an_update_cycle.

There was a message that wasnt handled in the view which caused the test to fail.
Adding the message handler from scala fixed the test.

cc @Horusiath


Horusiath:
:+1:

PR #1589, opened 2015-12-19
Fixes AtLeastOnceDelivery_must_redeliver_many_lost_messages test

rogeralsing:
This PR fixes the AtLeastOnceDelivery_must_redeliver_many_lost_messages which was previously Skipped.

The old code reads as follows:

```csharp
var resA = new SortedSet<string>(probeA.ReceiveN(n, deliverWithin).Cast<Action>().Select(x => x.Payload));
var resB = new SortedSet<string>(probeB.ReceiveN(n, deliverWithin).Cast<Action>().Select(x => x.Payload));
var resC = new SortedSet<string>(probeC.ReceiveN(n, deliverWithin).Cast<Action>().Select(x => x.Payload));
  
resA.ShouldOnlyContainInOrder(a);
resB.ShouldOnlyContainInOrder(b);
resC.ShouldOnlyContainInOrder(c);
```

It is explicitly trying to compare the result **in order**.

The Scala code does not do this, it only checks that the sequences contains the same data:

```scala
probeA.receiveN(N, deliverWithin).map { case a: Action ⇒ a.payload }.toSet should be((1 to N).map(n ⇒ "a-" + n).toSet)
probeB.receiveN(N, deliverWithin).map { case a: Action ⇒ a.payload }.toSet should be((1 to N).map(n ⇒ "b-" + n).toSet)
probeC.receiveN(N, deliverWithin).map { case a: Action ⇒ a.payload }.toSet should be((1 to N).map(n ⇒ "c-" + n).toSet)
```

`toSet` in scala returns a `Set`, Scala sets are not ordered, they have `SortedSet` for that.

The C# code have been updated to reflect the Scala non ordered comparison.

cc @Horusiath

### DIFF
--- a/src/core/Akka.Persistence.Tests/AtLeastOnceDeliverySpec.cs
+++ b/src/core/Akka.Persistence.Tests/AtLeastOnceDeliverySpec.cs
@@ -477,7 +477,7 @@ namespace Akka.Persistence.Tests
             Sys.Stop(sender);
         }
 
-        [Fact(Skip = "FIXME")]
+        [Fact]
         public void AtLeastOnceDelivery_must_redeliver_many_lost_messages()
         {
             var probeA = CreateTestProbe();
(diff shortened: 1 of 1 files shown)


Horusiath:
Great! :+1:

rogeralsing:
The reason why the old ordered test failed was simply because the C# SortedSet sorted things in alpha numeric order.
`a-38, a-39, a-4, a-40`
Unrelated to Akka.Persistence

PR #1591, opened 2015-12-19
Preparations before Akka.Cluster.Tools and Akka.Cluster.Sharding packaging

Horusiath:
### Motivation

First PR of `Akka.Cluster.Tools` and `Akka.Cluster.Sharding` introduced two fresh modules. However they were missing some parts, like lack of API encapsulation. Purpose of this PR is to prepare them for nightly releases.

### Modifications

- Added nuspec files for `Akka.Cluster.Tools` and `Akka.Cluster.Sharding` and included those assemblies in build.fsx.
- Implemented proto-buf based serializer for Akka.Cluster.Sharding messages.
- Moved cluster sharding and cluster tools control messages out of class nesting. Now they reside directly in correlated namespaces.
- Added async API for retrieving `ShardRegion` and `ShardRegionProxy` from `ClusterSharding` extension.
- Marked message definitions used by plugins internally as internal instead of public.
- Reorganized Cluster sharding example to be more readable.
- Renamed "entry" to "entity" in xml-docs (as they should be called by current convention).

### Result

Enclosed some of the API to reduce noise generated by types and prevent accidental missuse. Changed control message to follow C# namespacing convention. Improved contol messages serialization.


src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
```diff
@@ -391,6 +523,36 @@ public IActorRef StartProxy(string typeName, string role, IdExtractor idExtracto
...
+        /// Function to determine the shard id for an incoming message, only messages that passed the `extractEntityId` will be used
+        /// </param>
+        /// <returns>The actor ref of the <see cref="Sharding.ShardRegion"/> that is to be responsible for the shard.</returns>
+        public async Task<IActorRef> StartProxyAsync(string typeName, string role, IdExtractor idExtractor, ShardResolver shardResolver)
+        {
+            var timeout = _system.Settings.CreationTimeout;
+            var settings = ClusterShardingSettings.Create(_system).WithRole(role);
+            var startMsg = new ClusterShardingGuardian.StartProxy(typeName, settings, idExtractor, shardResolver);
+            var started = await _guardian.Value.Ask<ClusterShardingGuardian.Started>(startMsg, timeout);
+            _regions.TryAdd(typeName, started.ShardRegion);
```
rogeralsing: What happens if two calls to `StartProxyAsync` is called?
The current `TryAdd` will only add the first and drop the 2nd on the floor.
Is that the intended behavior?

src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
```diff
@@ -595,7 +784,7 @@ private void Done(bool ok)
     /// Check if we've received a shard start request.
     /// </summary>
     [Serializable]
-    public sealed class ResendShardHost
+    internal sealed class ResendShardHost
```
rogeralsing: IMO, it is somewhat arcane to have a serializable type where the type is not visible to the serializer.
(Most serializers will just ignore this and do magic anyway, but still, somewhat blurred intentions)

src/examples/Cluster/ClusterTools/ClusterToolsExample.Node/App.config
```diff
@@ -49,4 +49,19 @@
...
   </akka>
-<runtime><assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1" /></runtime></configuration>
\ No newline at end of file
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1" />
+  </runtime>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1" >
+      <dependentAssembly>
+        <assemblyIdentity name="Google.ProtocolBuffers.Serialization" publicKeyToken="55f7125234beb589" culture="neutral" />
```
rogeralsing: Do we need redirects or are we just mixing versions again?

src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
Horusiath: This doesn't necessarily matters, as this asynchronous async call will return the same actor ref (see [ClusterShardingGuardian handler](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Sharding/ClusterShardingGuardian.cs#L92)).

src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
Horusiath: I'm not sure I get it. AFAIK protocol buffers are able to serialize hidden types.

src/examples/Cluster/ClusterTools/ClusterToolsExample.Node/App.config
Horusiath: Samples are using the latest version of Google.ProtocolBuffers (which doesn't paste unnecessary folders and files). Beside that, yes, it's probably mixing versions - result of preferring NuGet over Paket ;)

Aaronontheweb: > result of preferring NuGet over Paket ;)

LOL

Should we upgrade to the latest version of protobuffs then? Could you create a Github issue for that @Horusiath ?

Horusiath: @Aaronontheweb  Done #1605

Issue #453, opened 2014-10-05
Fix bounded stashing


HCanber:
Once #452 **Implement bounded message queues** has been implemented bounded stashing can be fixed.
- Check if `BoundedStashImpl` needs updating
- Remove the `Obsolete` attribute on `WithBoundedStash` that warns the user that bounded stashing is not supported.
- Add tests to `ActorWithStashSpec` that verifies blocking and `PushTimeOut`

Aaronontheweb:
@HCanber does this still need to be done?

HCanber:
Yes, it does. It should be fairly simple now that #452 is implemented. The main thing is writing tests. It should probably go into a new test class `ActorWithBoundedStashSpec`, see https://github.com/akka/akka/blob/8485cd2ebb46d2fba851c41c03e34436e498c005/akka-actor-tests/src/test/scala/akka/actor/ActorWithBoundedStashSpec.scala
for inspiration

Aaronontheweb:
@HCanber do you want to take this one or should I mark it as up for grabs?

irperez:
Guys, I've been looking into this one, but I'm kind of stuck on one particular spot while unit testing.  I'm new to akka, so please be gentle :)

In particular, I'm calling context.Props.Mailbox inside the AbstractStash class.  And this is always = empty string.  And this is causing me a lot of headache as I'm not able to find how to get this to be any other value than an empty string.

Now just so its clear, here is the code I used so far in the constructor:

```C#
protected AbstractStash(IActorContext context, int capacity = 100)
        {
            var actorCell = (ActorCell)context;
            var mailbox = actorCell.Mailbox as IDequeBasedMailbox;
            if(mailbox == null)
            {
                string message = @"DequeBasedMailbox required, got: " + actorCell.Mailbox.GetType().Name + @"
An (unbounded) deque-based mailbox can be configured as follows:
    my-custom-mailbox {
        mailbox-type = ""Akka.Dispatch.UnboundedDequeBasedMailbox""
    }";
                throw new NotSupportedException(message);
            }
            _theStash = new LinkedList<Envelope>();
            _actorCell = actorCell;

            // TODO: capacity needs to come from dispatcher or mailbox config
            // https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/Stash.scala#L126            

            var dispatcher = context.System.Settings.Config.GetConfig(context.Props.Dispatcher);
[...]

PR #1598, opened 2015-12-21
[WIP] Activating bounded stashing

rogeralsing:
Enabling Bounded stashing as we now have bounded message queues.

This will of course need some tests before pulling it in.
I could use some help remembering how this interface galore setup works.
It's not clear how the `DequeWrapperMessageQueue` can interact with the push timeout on the `IBoundedDequeBasedMessageQueueSemantics`

Related to #453

### DIFF
--- a/src/core/Akka/Dispatch/MessageQueues/UnboundedDequeMessageQueue.cs
+++ b/src/core/Akka/Dispatch/MessageQueues/UnboundedDequeMessageQueue.cs
@@ -5,6 +5,8 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System;
+
 namespace Akka.Dispatch.MessageQueues
 {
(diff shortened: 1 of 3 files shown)


src/core/Akka/Dispatch/MessageQueues/UnboundedDequeMessageQueue.cs
```diff
@@ -19,6 +21,8 @@ public BoundedDequeMessageQueue()
             : base(new BoundedMessageQueue())
         {
         }
+
+        public TimeSpan PushTimeOut { get; set; }
```
rogeralsing: Maybe this needs a new base class that understands bounded semantics and thus can use the PushTimeout.
(?)

rogeralsing: This class really depends on two base classes, yay scala traits.
I need to merge the code from both classes in here

rogeralsing:
Closing, no requests for this feature for a very long time

Issue #1251, opened 2015-08-14
Port Akka.Remote MultiNodeSpec: RemoteNodeRestartDeathWatchSpec


Aaronontheweb:
https://github.com/akka/akka/blob/master/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteNodeRestartDeathWatchSpec.scala

maxim-s:
PR #1600

maxim-s:
PR #1600 Updated

maxim-s:
Closed by #1600

PR #1600, opened 2015-12-24
#1251 Port Akka.Remote MultiNodeSpec: RemoteNodeRestartDeathWatchSpec

maxim-s:


maxim-s:
still WIP

maxim-s:
Balckhole RateMBit should be zero https://github.com/akka/akka/blob/master/akka-multi-node-testkit/src/main/scala/akka/remote/testconductor/Player.scala#L229

Helios has both trttl and gremlin  
```
gremlin = "Akka.Remote.Transport.FailureInjectorProvider,Akka.Remote"
      trttl = "Akka.Remote.Transport.ThrottlerProvider,Akka.Remote")
```
MsgEncoder should handle ThrottleMsg with ThrottleMsg SetFailure(TCP.FailType.Throttle) 
https://github.com/akka/akka/blob/master/akka-multi-node-testkit/src/main/scala/akka/remote/testconductor/DataTypes.scala#L93

src/core/Akka.Remote.TestKit/MsgEncoder.cs
```diff
@@ -76,6 +76,7 @@ public void Encode(IConnection connection, object message, out List<IByteBuf> en
                         throttle =>
                             w.SetFailure(
                                 InjectFailure.CreateBuilder()
+                                    .SetFailure(TCP.FailType.Throttle)
```
Aaronontheweb: nice catch

Issue #1300, opened 2015-09-08
Throw exception when Type.GetType returns null


rogeralsing:
Related to #1286 

There are plenty of places in our codebase where we use `Type.GetType(typename)` and don't handle null results.

Here is a scan for `Type.GetType(`, some do handle null results correctly, e.g. either throw configuration exceptions or log warnings.
But in some cases, there is no handling for this. e.g. Akka.IO.DNS `Activator.CreateInstance(Type.GetType(Settings.ProviderObjectName));`

Imo, we should handle unresolved types the same way as in #1286 

```
Searching 1679 files for "Type.GetType(" (whole word)

C:\Projects\Git\akka.net\src\core\Akka.Cluster.Tests.MultiNode\MultiNodeClusterSpec.cs:
  445          public bool IsFailureDetectorPuppet()
  446          {
  447:             return Type.GetType(Cluster.Settings.FailureDetectorImplementationClass) == typeof (FailureDetectorPuppet);
  448          }
  449  

C:\Projects\Git\akka.net\src\core\Akka.Cluster.Tests\ClusterConfigSpec.cs:
   29              Assert.Equal(TimeSpan.FromMilliseconds(100), settings.FailureDetectorConfig.GetTimeSpan("min-std-deviation"));
   30              Assert.Equal(TimeSpan.FromSeconds(3), settings.FailureDetectorConfig.GetTimeSpan("acceptable-heartbeat-pause"));
   31:             Assert.Equal(typeof(PhiAccrualFailureDetector), Type.GetType(settings.FailureDetectorImplementationClass));
   32              Assert.Equal(ImmutableList.Create<Address>(), settings.SeedNodes);
   33              Assert.Equal(TimeSpan.FromSeconds(5), settings.SeedNodeTimeout);

C:\Projects\Git\akka.net\src\core\Akka.Cluster\ClusterMetricsCollector.cs:
  858              if (fqcn == typeof (PerformanceCounterMetricsCollector).AssemblyQualifiedName) return new PerformanceCounterMetricsCollector(system);
  859              
  860:             var metricsCollectorClass = Type.GetType(fqcn);
  861              if (metricsCollectorClass == null)
  862              {

C:\Projects\Git\akka.net\src\core\Akka.Persistence\Journal\EventAdapters.cs:
  167              var bindings = Sort(adapterBindings.Select(kv =>
  168              {
  169:                 var type = Type.GetType(kv.Key);
  170                  var adapter = kv.Value.Length == 1
[...]

PR #1602, opened 2015-12-26
adding TypeLoadException handling

alex-kondrashov:
These fixes relate to this issue #1300.

Basically I corrected all the places that I've found. The main idea behind those fixes is
 wrapping Type.GetType statement with try-catch and handling of TypeLoadException

There is only one commit with all the changes included.

### DIFF
--- a/src/core/Akka.Cluster/ClusterMetricsCollector.cs
+++ b/src/core/Akka.Cluster/ClusterMetricsCollector.cs
@@ -856,17 +856,15 @@ namespace Akka.Cluster
         {
             var fqcn = settings.MetricsCollectorClass;
             if (fqcn == typeof (PerformanceCounterMetricsCollector).AssemblyQualifiedName) return new PerformanceCounterMetricsCollector(system);
-            
-            var metricsCollectorClass = Type.GetType(fqcn);
-            if (metricsCollectorClass == null)
-            {
-                throw new ConfigurationException(string.Format("Could not create custom metrics collector {0}", fqcn));
-            }
-
             try
             {
-                var metricsCollector = (IMetricsCollector) Activator.CreateInstance(metricsCollectorClass, system);
+                var metricsCollectorClass = Type.GetType(fqcn, true);
+                var metricsCollector = (IMetricsCollector)Activator.CreateInstance(metricsCollectorClass, system);
                 return metricsCollector;
+            }
+            catch (TypeLoadException ex)
+            {
+                throw new ConfigurationException(string.Format("Could not create custom metrics collector {0}", fqcn), ex);
             }
             catch (Exception ex)
             {
--- a/src/core/Akka.Persistence/Journal/EventAdapters.cs
+++ b/src/core/Akka.Persistence/Journal/EventAdapters.cs
@@ -164,13 +164,22 @@ namespace Akka.Persistence.Journal
 
             // bindings is a enumerable of key-val representing the mapping from Type to handler.
             // It is primarily ordered by the most specific classes first, and secondly in the configured order.
+
             var bindings = Sort(adapterBindings.Select(kv =>
             {
-                var type = Type.GetType(kv.Key);
-                var adapter = kv.Value.Length == 1
-                    ? handlers[kv.Value[0]]
-                    : new CombinedReadEventAdapter(kv.Value.Select(h => handlers[h]));
-                return new KeyValuePair<Type, IEventAdapter>(type, adapter);
+                try
+                {
+                    var type = Type.GetType(kv.Key, true);
+                    var adapter = kv.Value.Length == 1
+                            ? handlers[kv.Value[0]]
+                            : new CombinedReadEventAdapter(kv.Value.Select(h => handlers[h]));
+                    return new KeyValuePair<Type, IEventAdapter>(type, adapter);
+                }
+                catch (Exception ex)
+                {
+                    var msgException = string.Format("Could not find type '{0}'", kv.Key);
+                    throw new TypeLoadException(msgException, ex);
+                }
             }).ToList());
 
             var backing = new ConcurrentDictionary<Type, IEventAdapter>();
@@ -235,18 +244,26 @@ namespace Akka.Persistence.Journal
 
         private static T Instantiate<T>(string qualifiedName, ExtendedActorSystem system)
         {
-            var instanceType = Type.GetType(qualifiedName);
-            if (!typeof(T).IsAssignableFrom(instanceType))
-                throw new ArgumentException(string.Format("Couldn't create instance of [{0}] from provided qualified type name [{1}], because it's not assignable from it",
-                    typeof(T), qualifiedName));
-
             try
             {
-                return (T)Activator.CreateInstance(instanceType, system);
+                var instanceType = Type.GetType(qualifiedName, true);
+                try
+                {
+                    if (!typeof(T).IsAssignableFrom(instanceType))
+                        throw new ArgumentException(string.Format("Couldn't create instance of [{0}] from provided qualified type name [{1}], because it's not assignable from it",
+                            typeof(T), qualifiedName));
+
+                    return (T)Activator.CreateInstance(instanceType, system);
+                }
+                catch (MissingMethodException)
+                {
+                    return (T)Activator.CreateInstance(instanceType);
+                }
             }
-            catch (MissingMethodException)
+            catch (TypeLoadException ex)
             {
-                return (T)Activator.CreateInstance(instanceType);
+                var msgException = string.Format("Could not find type '{0}'", qualifiedName);
+                throw new TypeLoadException(msgException, ex);
             }
         }
 
--- a/src/core/Akka.Persistence/Persistence.cs
+++ b/src/core/Akka.Persistence/Persistence.cs
@@ -156,17 +156,25 @@ namespace Akka.Persistence
 
             var pluginConfig = _system.Settings.Config.GetConfig(configPath);
             var pluginTypeName = pluginConfig.GetString("class");
-            var pluginType = Type.GetType(pluginTypeName, true);
 
-            var shouldInjectConfig = pluginConfig.HasPath("inject-config") && pluginConfig.GetBoolean("inject-config");
-            var pluginDispatcherId = pluginConfig.HasPath("plugin-dispatcher")
-                ? pluginConfig.GetString("plugin-dispatcher")
-                : dispatcherSelector(pluginType);
-            var pluginActorArgs = shouldInjectConfig ? new object[] { pluginConfig } : null;
-            var pluginActorProps = new Props(pluginType, pluginActorArgs).WithDispatcher(pluginDispatcherId);
+            try
+            {
+                var pluginType = Type.GetType(pluginTypeName, true);
+                var shouldInjectConfig = pluginConfig.HasPath("inject-config") && pluginConfig.GetBoolean("inject-config");
+                var pluginDispatcherId = pluginConfig.HasPath("plugin-dispatcher")
+                    ? pluginConfig.GetString("plugin-dispatcher")
+                    : dispatcherSelector(pluginType);
+                var pluginActorArgs = shouldInjectConfig ? new object[] { pluginConfig } : null;
+                var pluginActorProps = new Props(pluginType, pluginActorArgs).WithDispatcher(pluginDispatcherId);
 
-            var pluginRef = _system.SystemActorOf(pluginActorProps, configPath);
-            return new PluginHolder(pluginRef, CreateAdapters(configPath));
+                var pluginRef = _system.SystemActorOf(pluginActorProps, configPath);
+                return new PluginHolder(pluginRef, CreateAdapters(configPath));
+            }
+            catch (TypeLoadException ex)
+            {
+                var msgException = string.Format("Could not find type '{0}'", pluginTypeName);
+                throw new TypeLoadException(msgException, ex);
+            }   
         }
 
         private EventAdapters CreateAdapters(string configPath)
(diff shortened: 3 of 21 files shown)


rogeralsing:
Looks good over all, I'm not sure if we should go with a generic error message such as "Could not find type ...." or if we should try to be more explicit per use case.

In the reference PR we do have a more descriptive message :
https://github.com/akkadotnet/akka.net/pull/1286/files#diff-e0a72019db08bc7dfe1a953689e00917R124

cc @akkadotnet/contributors

Aaronontheweb:
I agree with @rogeralsing 

Could you make the error messages more explicit, @alex-kondrashov ? Like if the exception happens because we were unable to load a `Serializer` type could you change the error message to mention that we failed to load a Serializer? Does that make sense?

Aaronontheweb:
@alex-kondrashov would you mind updating this?

rogeralsing:
@Aaronontheweb can you provide some additional info on some of the cases here, it might be hard for an oudsider to know the exact details of each usecase, I'll see what I can fill in myself

Aaronontheweb:
@rogeralsing yep, I'll help start putting together a list

src/core/Akka/Serialization/Serializer.cs
```diff
@@ -121,7 +121,17 @@ public static int GetSerializerIdentifierFromConfig(Type type, ExtendedActorSyst
         {
```
rogeralsing: cc @Horusiath what does this method actually do?
Each serializer does have an identifier on the serializer, why is this needed?
Can we not just use ` Sys.Serialization.FindSerializerFor(obj).Identifier`
Does the identifier need to be configurable?

Horusiath: At some point Typesafe guys added ability to specify identifier for each serializer directly through config i.e. for case, when you have conflicting serializers.

src/core/Akka/IO/Dns.cs
```diff
@@ -124,11 +124,19 @@ public DnsSettings(Config config)
...
-            Settings = new DnsSettings(system.Settings.Config.GetConfig("akka.io.dns"));
-            //TODO: system.dynamicAccess.getClassFor[DnsProvider](Settings.ProviderObjectName).get.newInstance()
-            Provider = (IDnsProvider) Activator.CreateInstance(Type.GetType(Settings.ProviderObjectName));
-            Cache = Provider.Cache;
+            try
+            {
+                _system = system;
+                Settings = new DnsSettings(system.Settings.Config.GetConfig("akka.io.dns"));
+                //TODO: system.dynamicAccess.getClassFor[DnsProvider](Settings.ProviderObjectName).get.newInstance()
+                Provider = (IDnsProvider)Activator.CreateInstance(Type.GetType(Settings.ProviderObjectName, true));
```
rogeralsing: IMO, `Type.GetType(Settings.ProviderObjectName, true)`should go on a separate line for clarity.
There should also be a check to ensure that the created object from `Activator.CreateInstance` actually is assignable to `IDnsProvder`

The message for the TypeLoadException should indicate that the type is to be used as a Dns Provider.
If the type is not of the correct type, there needs to be a `ConfigurationException` thrown explaining that the type is not a IDnsProvider.

src/core/Akka/Dispatch/Mailboxes.cs
```diff
@@ -111,8 +120,16 @@ public Type FromConfig(string path)
...
-            var mailboxType = Type.GetType(type);
-            return mailboxType;
+            try
+            {
+                var mailboxType = Type.GetType(type, true);
+                return mailboxType;
+            }
+            catch (TypeLoadException ex)
+            {
+                var msgException = string.Format("Could not find type '{0}'", type);
```
rogeralsing: Here we have the full `path`, `path + ",mailbox-type"` so we can provide that in the type load exception.
"Could not find the mailbox type ${type} provided in the configuration path ${path + ".mailbox-path"}`"
We should also have a check that ensures that the type is assignable to `Akka.Dispatch.Mailbox`, if not, then we need to throw a configuration exception saying that the provided type is not a mailbox type

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -268,10 +276,18 @@ private void ConfigureMailboxes()
...
+            try
+            {
+                Type providerType = Type.GetType(_settings.ProviderClass, true);
+                global::System.Diagnostics.Debug.Assert(providerType != null, "providerType != null");
+                var provider = (IActorRefProvider)Activator.CreateInstance(providerType, _name, _settings, _eventStream);
+                _provider = provider;
+            }
+            catch (TypeLoadException ex)
+            {
+                var msgException = string.Format("Could not find type '{0}'", _settings.ProviderClass);
```
rogeralsing: We should indicate that the type should be used as an `ActorRefProvider`..
e.g. "Could not find the actorref provider type '{0}'"
We also need to check that the type is assignable to `Akka.Actor.IActorRefProvider`, if not, throw a configuration exception indicating that the type is not a valid actor ref provider.

src/core/Akka/Actor/Deployer.cs
```diff
@@ -100,11 +100,20 @@ private RouterConfig CreateRouterConfig(string routerTypeAlias, Config deploymen
...
             var routerTypeName = _settings.Config.GetString(path);
-            var routerType = Type.GetType(routerTypeName);
-            Debug.Assert(routerType != null, "routerType != null");
-            var routerConfig = (RouterConfig)Activator.CreateInstance(routerType, deployment);
 
-            return routerConfig;
+            try
+            {
+                var routerType = Type.GetType(routerTypeName, true);
+                Debug.Assert(routerType != null, "routerType != null");
```
rogeralsing: The debug.assert calls can be removed as they will never be hit as we are throwing before the line is executed.

src/core/Akka/Actor/Deployer.cs
```diff
@@ -100,11 +100,20 @@ private RouterConfig CreateRouterConfig(string routerTypeAlias, Config deploymen
...
+                var routerConfig = (RouterConfig)Activator.CreateInstance(routerType, deployment);
+
+                return routerConfig;
+            }
+            catch (TypeLoadException ex)
+            {
+                var msgException = string.Format("Could not find type '{0}'", routerTypeName);
```
rogeralsing: We should indicate that the type is intended for a Rotuer Config.
e.g. "Could not find the router config type ${type}"

we should also throw an `ConfigurationException` if the type is not assignable to `Akka.Routing.RouterConfig`

src/core/Akka/Serialization/Serializer.cs
rogeralsing: Has this replaced the use of `serializer.Identifier` through out the Scala code? or are they assigning the value to the Identifier property using this?

(It seems to be a strange way to go about things as you might end up with one type of serializer on one node claiming to be identifier 1, and a completely different serializer with the same identifier on another node.)

Horusiath: It isn't touching `Identifer` property, it can be only used to assign some value to it.

rogeralsing: So ise the serializer.identifier obsolete and should be replaced with this new way of handling things?
It seems conflicting to have both? or maybe I'm just not understanding the difference :-)

Horusiath: No. `Serializer.Identifier` is still valid. What only have changed, is that beside of having statically assigned number, now it can be a configurable value instead.

alex-kondrashov:
I've fixed the messages in the cases mentioned in the list above. Please have a review on that.

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -132,8 +132,16 @@ public override ActorSelection ActorSelection(string actorPath)
 
         private void ConfigureScheduler()
         {
-            var schedulerType = Type.GetType(_settings.SchedulerClass, true);
-            _scheduler = (IScheduler) Activator.CreateInstance(schedulerType, this);
+            try
+            {
+                var schedulerType = Type.GetType(_settings.SchedulerClass, true);
+                _scheduler = (IScheduler)Activator.CreateInstance(schedulerType, this);
```
rogeralsing: We need to check that the type is an IScheduler or throw configuration exception

src/core/Akka.Cluster/ClusterMetricsCollector.cs
```diff
@@ -856,18 +856,16 @@ public static IMetricsCollector Get(ExtendedActorSystem system, ClusterSettings
...
-            if (metricsCollectorClass == null)
-            {
-                throw new ConfigurationException(string.Format("Could not create custom metrics collector {0}", fqcn));
-            }
-
             try
             {
-                var metricsCollector = (IMetricsCollector) Activator.CreateInstance(metricsCollectorClass, system);
+                var metricsCollectorClass = Type.GetType(fqcn, true);
+                var metricsCollector = (IMetricsCollector)Activator.CreateInstance(metricsCollectorClass, system);
```
rogeralsing: We should check that the type is a IMetricsCollector or throw configuration exception

src/core/Akka.Persistence/Journal/EventAdapters.cs
```diff
@@ -164,13 +164,22 @@ private static EventAdapters Create(ExtendedActorSystem system, IDictionary<stri
...
+                    var type = Type.GetType(kv.Key, true);
+                    var adapter = kv.Value.Length == 1
+                            ? handlers[kv.Value[0]]
+                            : new CombinedReadEventAdapter(kv.Value.Select(h => handlers[h]));
+                    return new KeyValuePair<Type, IEventAdapter>(type, adapter);
+                }
+                catch (Exception ex)
+                {
+                    var msgException = string.Format("Could not find type '{0}'", kv.Key);
+                    throw new TypeLoadException(msgException, ex);
```
rogeralsing: @Horusiath what would be the correct messages here?

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -156,17 +156,25 @@ private PluginHolder CreatePlugin(string configPath, Func<Type, string> dispatch
...
-            return new PluginHolder(pluginRef, CreateAdapters(configPath));
+            try
+            {
+                var pluginType = Type.GetType(pluginTypeName, true);
+                var shouldInjectConfig = pluginConfig.HasPath("inject-config") && pluginConfig.GetBoolean("inject-config");
+                var pluginDispatcherId = pluginConfig.HasPath("plugin-dispatcher")
+                    ? pluginConfig.GetString("plugin-dispatcher")
+                    : dispatcherSelector(pluginType);
+                var pluginActorArgs = shouldInjectConfig ? new object[] { pluginConfig } : null;
+                var pluginActorProps = new Props(pluginType, pluginActorArgs).WithDispatcher(pluginDispatcherId);
```
rogeralsing: cc @Aaronontheweb @Horusiath should pluginType be checked for ActorBase compat here?

src/core/Akka.Remote/FailureDetectorRegistry.cs
```diff
@@ -65,13 +65,16 @@ public static class FailureDetectorLoader
         /// <returns>A configured instance of the given <see cref="FailureDetector"/> implementation.</returns>
         public static FailureDetector Load(string fqcn, Config config, ActorSystem system)
         {
-            var failureDetectorClass = Type.GetType(fqcn);
-            if (failureDetectorClass == null)
+            try
+            {
+                var failureDetectorClass = Type.GetType(fqcn, true);
+                var failureDetector = (FailureDetector)Activator.CreateInstance(failureDetectorClass, config, system.EventStream);
```
rogeralsing: Should check for FailureDetector compat or throw configuration exception

src/core/Akka.Remote/Transport/TransportAdapters.cs
```diff
@@ -66,11 +66,16 @@ public TransportAdapters(ExtendedActorSystem system)
             {
                 try
                 {
-                    var adapterTypeName = Type.GetType(adapter.Value);
+                    var adapterTypeName = Type.GetType(adapter.Value, true);
                     // ReSharper disable once AssignNullToNotNullAttribute
                     var newAdapter = (ITransportAdapterProvider)Activator.CreateInstance(adapterTypeName);
```
rogeralsing: should check for ITransportAdapterProvider compat or throw configuration exception

src/core/Akka/Serialization/Serializer.cs
```diff
@@ -121,7 +121,17 @@ public static int GetSerializerIdentifierFromConfig(Type type, ExtendedActorSyst
...
-                .ToDictionary(pair => Type.GetType(pair.Key, true), pair => pair.Value.GetInt());
+                .ToDictionary(pair => {
+                    try
+                    {
+                        return Type.GetType(pair.Key, true);
+                    }
+                    catch (TypeLoadException ex)
+                    {
+                        var msgException = string.Format("Could not find serialization identifier type '{0}'", pair.Key);
+                        throw new TypeLoadException(msgException, ex);
```
rogeralsing: @Horusiath is this the correct message?

src/core/Akka.Persistence/Persistence.cs
Horusiath: Yes, it should. But we could think of it as a separate issue.

src/core/Akka/Serialization/Serializer.cs
Horusiath: It's fine.

src/core/Akka.Persistence/Journal/EventAdapters.cs
Horusiath: Something like "Could not find event adapter of type '{0}'". It's more informative, but this one is still good.

PR #1603, opened 2015-12-29
Added basic TCP-Logger support for multi node tests

Silv3rcircl3:
This PR is a another attempt to implement #1475, currently it's only a starting point for further steps and not ready.

### DIFF
--- a/src/core/Akka.MultiNodeTestRunner/Program.cs
+++ b/src/core/Akka.MultiNodeTestRunner/Program.cs
@@ -11,8 +11,11 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Net;
+using System.Reflection;
 using System.Threading;
 using Akka.Actor;
+using Akka.IO;
 using Akka.MultiNodeTestRunner.Shared;
 using Akka.MultiNodeTestRunner.Shared.Persistence;
 using Akka.MultiNodeTestRunner.Shared.Sinks;
@@ -70,6 +73,9 @@ namespace Akka.MultiNodeTestRunner
         {
             TestRunSystem = ActorSystem.Create("TestRunnerLogging");
             SinkCoordinator = TestRunSystem.ActorOf(Props.Create<SinkCoordinator>(), "sinkCoordinator");
+
+            var tcpLogger = TestRunSystem.ActorOf<TcpLoggingServer>();
+            TestRunSystem.Tcp().Tell(new Tcp.Bind(tcpLogger, new IPEndPoint(IPAddress.Any, 5678)));
 
             var assemblyName = Path.GetFullPath(args[0]);
             EnableAllSinks(assemblyName);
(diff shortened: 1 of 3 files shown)


src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -71,6 +74,9 @@ static void Main(string[] args)
             TestRunSystem = ActorSystem.Create("TestRunnerLogging");
             SinkCoordinator = TestRunSystem.ActorOf(Props.Create<SinkCoordinator>(), "sinkCoordinator");
 
+            var tcpLogger = TestRunSystem.ActorOf<TcpLoggingServer>();
+            TestRunSystem.Tcp().Tell(new Tcp.Bind(tcpLogger, new IPEndPoint(IPAddress.Any, 5678)));
```
Horusiath: Does it need to be a hardcoded port? Maybe a config value or program parameter would be a better idea?

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -225,5 +231,29 @@ static void PublishToAllSinks(string message)
...
+    class TcpLoggingServer : ReceiveActor
+    {
+        public TcpLoggingServer()
+        {
+            var currentDirectory = Path.GetDirectoryName(Assembly.GetEntryAssembly().Location);
+            var logFilePath = Path.Combine(currentDirectory, "TcpLog.txt");
+
+            Receive<Tcp.Connected>(connected =>
+            {
+                File.AppendAllText(logFilePath, $"Node connected on {Sender}{Environment.NewLine}");
```
Horusiath: We're still bound to C# 5 syntax. Interpolated strings are not allowed.

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -58,6 +69,9 @@ static void Main(string[] args)
                                 specFail.FailureStackTraces.Add(innerEx.StackTrace);
                             }
                             Console.WriteLine(specFail);
+
+                            //make sure message is send over the wire
+                            Task.Delay(TimeSpan.FromSeconds(10)).Wait();
```
Horusiath: This is equivalent of `Thread.Sleep(10000)` ;)

src/core/Akka.MultiNodeTestRunner/Program.cs
Aaronontheweb: I wrote a version that does this - let me see if I can dig it up

src/core/Akka.MultiNodeTestRunner/Program.cs
Aaronontheweb: @Horusiath we decided to change that at the last meeting, did we not?

src/core/Akka.NodeTestRunner/Program.cs
Aaronontheweb: I second this change :+1:

Aaronontheweb:
Closing this in favor of #1604

PR #1604, opened 2015-12-31
Resolves 1475 - reliable error capture and logging system for MultiNodeTests

Aaronontheweb:
Builds on the work by @Silv3rcircl3 and uses a combination of STD OUT redirection and TCP logging to capture a complete picture of what actually happened inside a multi-node test.

### DIFF
--- a/src/core/Akka.MultiNodeTestRunner.Shared/Akka.MultiNodeTestRunner.Shared.csproj
+++ b/src/core/Akka.MultiNodeTestRunner.Shared/Akka.MultiNodeTestRunner.Shared.csproj
@@ -60,6 +60,7 @@
     </Compile>
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Sinks\ConsoleMessageSinkActor.cs" />
+    <Compile Include="Sinks\FileSystemAppenderActor.cs" />
     <Compile Include="Sinks\FileSystemMessageSinkActor.cs" />
     <Compile Include="Sinks\IMessageSink.cs" />
     <Compile Include="Reporting\MultiNodeMessage.cs" />
--- a/src/core/Akka.MultiNodeTestRunner.Shared/Sinks/FileSystemAppenderActor.cs
+++ b/src/core/Akka.MultiNodeTestRunner.Shared/Sinks/FileSystemAppenderActor.cs
@@ -1,0 +1,25 @@
+using System;
+using System.IO;
+using Akka.Actor;
+
+namespace Akka.MultiNodeTestRunner.Shared.Sinks
+{
+    /// <summary>
+    /// Actor that just writes dumb messages to the file system - used for capturing
+    /// raw logs from individual nodes.
+    /// </summary>
+    public class FileSystemAppenderActor : ReceiveActor
+    {
+        private readonly string _fullFilePath;
+
+        public FileSystemAppenderActor(string fullFilePath)
+        {
+            _fullFilePath = fullFilePath;
+
+            ReceiveAny(o =>
+            {
+                File.AppendAllText(_fullFilePath, o.ToString() + Environment.NewLine);
+            });
+        }
+    }
+}
--- a/src/core/Akka.MultiNodeTestRunner/Program.cs
+++ b/src/core/Akka.MultiNodeTestRunner/Program.cs
@@ -15,6 +15,7 @@ using System.Net;
 using System.Reflection;
 using System.Threading;
 using Akka.Actor;
+using Akka.Event;
 using Akka.IO;
 using Akka.MultiNodeTestRunner.Shared;
 using Akka.MultiNodeTestRunner.Shared.Persistence;
@@ -33,7 +34,10 @@ namespace Akka.MultiNodeTestRunner
 
         protected static IActorRef SinkCoordinator;
 
-
+        /// <summary>
+        /// file output directory
+        /// </summary>
+        protected static string OutputDirectory;
 
         /// <summary>
         /// MultiNodeTestRunner takes the following <see cref="args"/>:
@@ -67,15 +71,39 @@ namespace Akka.MultiNodeTestRunner
         ///                  as the test binary.
         ///     </description>
         /// </item>
+        /// <item>
+        ///     <term>-Dmultinode.listen-address={ip}</term>
+        ///     <description>
+        ///             Determines the address that this multi-node test runner will use to listen for log messages from
+        ///             individual NodeTestRunner.exe processes.
+        /// 
+        ///             Defaults to 127.0.0.1
+        ///     </description>
+        /// </item>
+        /// <item>
+        ///     <term>-Dmultinode.listen-port={port}</term>
+        ///     <description>
+        ///             Determines the port number that this multi-node test runner will use to listen for log messages from
+        ///             individual NodeTestRunner.exe processes.
+        /// 
+        ///             Defaults to 6577
+        ///     </description>
+        /// </item>
         /// </list>
         /// </summary>
         static void Main(string[] args)
         {
+            OutputDirectory = CommandLine.GetProperty("multinode.output-directory") ?? string.Empty;
             TestRunSystem = ActorSystem.Create("TestRunnerLogging");
             SinkCoordinator = TestRunSystem.ActorOf(Props.Create<SinkCoordinator>(), "sinkCoordinator");
 
-            var tcpLogger = TestRunSystem.ActorOf<TcpLoggingServer>();
-            TestRunSystem.Tcp().Tell(new Tcp.Bind(tcpLogger, new IPEndPoint(IPAddress.Any, 5678)));
+
+            var listenAddress = IPAddress.Parse(CommandLine.GetPropertyOrDefault("multinode.listen-address", "127.0.0.1"));
+            var listenPort = CommandLine.GetInt32OrDefault("multinode.listen-port", 6577);
+            var listenEndpoint = new IPEndPoint(listenAddress, listenPort);
+
+            var tcpLogger = TestRunSystem.ActorOf(Props.Create(() => new TcpLoggingServer(SinkCoordinator)), "TcpLogger");
+            TestRunSystem.Tcp().Tell(new Tcp.Bind(tcpLogger, listenEndpoint));
 
             var assemblyName = Path.GetFullPath(args[0]);
             EnableAllSinks(assemblyName);
@@ -101,7 +129,6 @@ namespace Akka.MultiNodeTestRunner
                         var processes = new List<Process>();
 
                         StartNewSpec(test.Value);
-
                         foreach (var nodeTest in test.Value)
                         {
                             //Loop through each test, work out number of nodes to run on and kick off process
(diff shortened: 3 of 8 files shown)


src/core/Akka.MultiNodeTestRunner.Shared/Akka.MultiNodeTestRunner.Shared.csproj
```diff
@@ -60,6 +60,7 @@
     </Compile>
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Sinks\ConsoleMessageSinkActor.cs" />
+    <Compile Include="Sinks\FileSystemAppenderActor.cs" />
```
Aaronontheweb: Added a new type of actor that just does dumb file-appending to a logfile.

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/FileSystemAppenderActor.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.IO;
+using Akka.Actor;
+
+namespace Akka.MultiNodeTestRunner.Shared.Sinks
+{
+    /// <summary>
+    /// Actor that just writes dumb messages to the file system - used for capturing
+    /// raw logs from individual nodes.
+    /// </summary>
+    public class FileSystemAppenderActor : ReceiveActor
```
Aaronontheweb: Aforementioned file appender actor.

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -64,13 +71,40 @@ class Program
         ///                  as the test binary.
         ///     </description>
         /// </item>
+        /// <item>
+        ///     <term>-Dmultinode.listen-address={ip}</term>
```
Aaronontheweb: New commandline parameters used to specify which IP and port the TCP logger should listen on. Optional - has a set of safe defaults implemented below.

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -64,13 +71,40 @@ class Program
...
         /// </list>
         /// </summary>
         static void Main(string[] args)
         {
+            OutputDirectory = CommandLine.GetProperty("multinode.output-directory");
             TestRunSystem = ActorSystem.Create("TestRunnerLogging");
             SinkCoordinator = TestRunSystem.ActorOf(Props.Create<SinkCoordinator>(), "sinkCoordinator");
 
+
+            var listenAddress = IPAddress.Parse(CommandLine.GetPropertyOrDefault("multinode.listen-address", "127.0.0.1"));
```
Aaronontheweb: Grabs the specified commandline parameter for IP, or defaults to `127.0.0.1`

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -64,13 +71,40 @@ class Program
...
+            var listenPort = CommandLine.GetInt32OrDefault("multinode.listen-port", 6577);
```
Aaronontheweb: Grabs the specified port, or defaults to `6577`

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -64,13 +71,40 @@ class Program
...
+            var listenEndpoint = new IPEndPoint(listenAddress, listenPort);
+
+            var tcpLogger = TestRunSystem.ActorOf(Props.Create(() => new TcpLoggingServer(SinkCoordinator)), "TcpLogger");
```
Aaronontheweb: Initializes the logger actor, which takes a reference to the `SinkCoordinator` actor who's responsible for determining the state of the multi-node test in aggregate across all nodes.

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -104,22 +137,23 @@ static void Main(string[] args)
                             process.StartInfo.UseShellExecute = false;
                             process.StartInfo.RedirectStandardOutput = true;
                             process.StartInfo.FileName = "Akka.NodeTestRunner.exe";
-                            process.StartInfo.Arguments = String.Format(@"-Dmultinode.test-assembly=""{0}"" -Dmultinode.test-class=""{1}"" -Dmultinode.test-method=""{2}"" -Dmultinode.max-nodes={3} -Dmultinode.server-host=""{4}"" -Dmultinode.host=""{5}"" -Dmultinode.index={6}",
-                                assemblyName, nodeTest.TypeName, nodeTest.MethodName, test.Value.Count, "localhost", "localhost", nodeTest.Node - 1);
+                            process.StartInfo.Arguments =
```
Aaronontheweb: We pass on the bound IP and Port onto the child `NodeTestRunner.exe` processes we spawned. Refactored this into string interpolation.

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -104,22 +137,23 @@ static void Main(string[] args)
...
-                                    string message = line.Data;
-                                    if (!message.StartsWith("[NODE", true, CultureInfo.InvariantCulture))
-                                    {
-                                        message = "[NODE" + nodeIndex + "]" + message;
-                                    }
-                                    PublishToAllSinks(message);
-                                };
-
+                            //TODO: might need to do some validation here to avoid the 260 character max path error on Windows
+                            var folder = Directory.CreateDirectory(Path.Combine(OutputDirectory, nodeTest.MethodName));
```
Aaronontheweb: Sets up STD OUT logging to capture the full trace for all nodes into their own files, all grouped together into a folder that shares the name of the test. Had some issues where I hit the 260 character path limit earlier, so we might need to possibly hash or shorten the name of the folder.

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -104,22 +137,23 @@ static void Main(string[] args)
...
+                            var logFilePath = Path.Combine(folder.FullName, "node" + nodeIndex + ".txt");
+                            var fileActor =
+                                TestRunSystem.ActorOf(Props.Create(() => new FileSystemAppenderActor(logFilePath)));
```
Aaronontheweb: Creates a `FileSystemAppenderActor` for each individual log file and has the STD OUT `process.OutputDataReceived` event pipe all messages to that actor.

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -225,5 +259,31 @@ static void PublishToAllSinks(string message)
             SinkCoordinator.Tell(message, ActorRefs.NoSender);
         }
     }
+
+    class TcpLoggingServer : ReceiveActor
```
Aaronontheweb: Akka.IO logging actor implementation

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -23,6 +30,14 @@ static void Main(string[] args)
             var testName = CommandLine.GetProperty("multinode.test-method");
             var displayName = testName;
 
+            var listenAddress = IPAddress.Parse(CommandLine.GetProperty("multinode.listen-address"));
```
Aaronontheweb: Grab the specified IP address and port from the commandline - if these arguments are not specified to `NodeTestRunner.exe` then the process will fail, unlike the `MultiNodeTestRunner.exe` binary which gets called directly by the end-user.

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -36,7 +51,7 @@ static void Main(string[] args)
                 Console.WriteLine("Running specs for {0} [{1}]", assemblyName, assemblyFileName);
                 using (var discovery = new Discovery(assemblyName, typeName))
                 {
-                    using (var sink = new Sink(nodeIndex))
+                    using (var sink = new Sink(nodeIndex, tcpClient))
```
Aaronontheweb: the `Sink` now takes a directly reference to the `RunnerTcpClient` actor reference and writes to that.

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -58,6 +73,9 @@ static void Main(string[] args)
                                 specFail.FailureStackTraces.Add(innerEx.StackTrace);
                             }
                             Console.WriteLine(specFail);
+
+                            //make sure message is send over the wire
+                            FlushLogMessages();
```
Aaronontheweb: Helper method used to prevent the process from exiting until all messages have been flushed onto the network.

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -67,14 +85,60 @@ static void Main(string[] args)
...
+
                         Environment.Exit(sink.Passed ? 0 : 1);
                     }
                 }
             }
         }
+
+        private static void FlushLogMessages()
+        {
+            _logger.GracefulStop(TimeSpan.FromSeconds(2)).Wait();
```
Aaronontheweb: I implemented this blocking flush mechanism using `GracefulStop`, which uses a `PoisonPill` by default - this doesn't necessarily guarantee that the actor has pushed all of its messages onto the other end of the wire but it's a pretty good approximation.

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -67,14 +85,60 @@ static void Main(string[] args)
...
+        }
     }
+
+     class RunnerTcpClient : ReceiveActor, IWithUnboundedStash
```
Aaronontheweb: Akka.IO Tcp client logger implementation

src/core/Akka.NodeTestRunner/Sink.cs
```diff
@@ -20,10 +21,13 @@ class Sink : MarshalByRefObject, IMessageSink, IDisposable
...
         readonly int _nodeIndex;
 
-        public Sink(int nodeIndex)
+        private IActorRef _logger;
+
+        public Sink(int nodeIndex, IActorRef logger)
         {
             _nodeIndex = nodeIndex;
             Finished = new ManualResetEvent(false);
+            _logger = logger;
```
Aaronontheweb: We now write to this `IActorRef` instead of the `Console`

src/core/Akka.Remote.TestKit/CommandLine.cs
```diff
@@ -43,10 +43,20 @@ public static string GetProperty(string key)
             return Values.Value[key];
         }
 
+        public static string GetPropertyOrDefault(string key, string defaultStr)
```
Aaronontheweb: Added some overloads to `CommandLine` to make it easier to specify defaults for optional parameters.

Aaronontheweb:
Should probably reference #1475 here

Aaronontheweb:
Got the specs to fail after enough runs so I could see the output - looks like our stacktrace still did not show up correctly. Will take a closer look at that.

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -64,13 +71,40 @@ class Program
...
+        ///             individual NodeTestRunner.exe processes.
+        /// 
+        ///             Defaults to 6577
+        ///     </description>
+        /// </item>
         /// </list>
         /// </summary>
         static void Main(string[] args)
         {
+            OutputDirectory = CommandLine.GetProperty("multinode.output-directory");
```
Silv3rcircl3: This would return null if this property doesn't exist and therefore Path.Combine would throw an ArgumentNullException later on. Should probably return an empty string instead of null.

Aaronontheweb: Good call

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -165,7 +199,7 @@ static void EnableAllSinks(string assemblyName)
 
             // if multinode.output-directory wasn't specified, the results files will be written
             // to the same directory as the test assembly.
-            var outputDirectory = CommandLine.GetProperty("multinode.output-directory");
+            var outputDirectory = OutputDirectory;
```
Aaronontheweb: Could get rid of this local variable and just use `OutputDirectory` directly

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -57,7 +72,11 @@ static void Main(string[] args)
                                 specFail.FailureMessages.Add(innerEx.Message);
                                 specFail.FailureStackTraces.Add(innerEx.StackTrace);
                             }
+                            _logger.Tell(specFail.ToString());
```
Aaronontheweb: Realized we weren't capturing test failure output here - logging it now.

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -66,15 +85,62 @@ static void Main(string[] args)
                             specFail.FailureExceptionTypes.Add(ex.GetType().ToString());
                             specFail.FailureMessages.Add(ex.Message);
                             specFail.FailureStackTraces.Add(ex.StackTrace);
+                            _logger.Tell(specFail.ToString());
```
Aaronontheweb: Realized we weren't capturing test failure output here - logging it now.

src/core/Akka.NodeTestRunner/Sink.cs
```diff
@@ -31,6 +35,7 @@ public bool OnMessage(IMessageSinkMessage message)
             var resultMessage = message as ITestResultMessage;
             if (resultMessage != null)
             {
+                _logger.Tell(resultMessage.Output);
```
Aaronontheweb: In all areas within the `Sink`, we log both to the `Console` and to the Akka.IO Logger - that way in the event that anything goes wrong with the Akka,IO transport we still have a local build log for the node that experienced a failure.

Aaronontheweb:
I implemented @Silv3rcircl3 suggestion and I think this PR is ready for merge, pending build results.

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -66,15 +91,80 @@ static void Main(string[] args)
...
                             Console.WriteLine(specFail);
+
+                            //make sure message is send over the wire
+                            FlushLogMessages();
                             Environment.Exit(1); //signal failure
+                            return 1;
```
Aaronontheweb: Added this for Mono / Unix compatibility in the future. We got burned by this issue on NBench so I thought it best to preemptively add it here. Effectively a no-op on Windows.

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -66,15 +91,80 @@ static void Main(string[] args)
...
                         }
-                        sink.Finished.WaitOne();
-                        Environment.Exit(sink.Passed ? 0 : 1);
+
+                        var timedOut = false;
+                        if (!sink.Finished.WaitOne(MaxProcessWaitTimeout)) //timed out
```
Aaronontheweb: **IMPORTANT**

I suspect we have some issues with deadlocks inside the `TestConductor` inside `Akka.Remote.TestKit` - in order to work around that and not have the MNTR hang the build agent for 60 minutes (that's the timeout specified by FAKE), I set up a timeout here within each local `NodeTestRunner.exe` process.

If any one of these processes takes longer than `MaxProcessWaitTimeout` (1.5 minutes, currently) the process will exit early and report a failure. 

We will likely need to tune this code later, but for the time being I think it's a reasonable safeguard to prevent hangs.

src/core/Akka.NodeTestRunner/Program.cs
```diff
@@ -66,15 +91,80 @@ static void Main(string[] args)
...
                     }
                 }
             }
         }
+
+        private static void FlushLogMessages()
+        {
+            try
+            {
+                _logger.GracefulStop(TimeSpan.FromSeconds(2)).Wait();
```
Aaronontheweb: We use `GracefulStop` on the actor that writes out to Akka.IO as our means of flushing the transport - any new log messages added after the `FlushLogMessages` method is called will not be pushed onto the transport.

Aaronontheweb:
Heh, probably should have mentioned #1475 in the comments so we'd get a crosslink :p

PR #1613, opened 2016-01-05
Clear current message - fixes #1609

### LINKED ISSUE #1609: Akka seems to keep a reference to the last handled message
When I send a very large message (object containing an XDocument for a 800-mb xml file) to a ReceiveActor and handle it, then garbage collect, it seems that the 1.7 gb for the document is not being cleaned up. If I follow up the large message with a small message, the large object DOES get cleaned up. Does Akka keep a reference to the last handled message?

rogeralsing:
Fixes #1609

cc @JeffCyr please review that the current message on async `ActorTaskSchedulerMessage` messages looks OK.

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorCellSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorCellSpec.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.TestKit;
+using Xunit;
+
+namespace Akka.Tests.Actor
+{
+
+    public class ActorCellSpec : AkkaSpec
+    {
+        public class DummyActor : ReceiveActor
+        {
+            public DummyActor(AutoResetEvent autoResetEvent)
+            {
+                ReceiveAny(m => autoResetEvent.Set());
+            }
+        }
+
+        public class DummyAsyncActor : ReceiveActor
+        {
+            public DummyAsyncActor(AutoResetEvent autoResetEvent)
+            {
+                Receive<string>(async m =>
+                {
+                    await Task.Delay(500);
+                    autoResetEvent.Set();
+                });
+            }
+        }
+
+        [Fact]
+        public void Cell_should_clear_current_message_after_receive()
+        {
+            AutoResetEvent autoResetEvent = new AutoResetEvent(false);
+            var actor = Sys.ActorOf(Props.Create(() => new DummyActor(autoResetEvent)));
+            actor.Tell("hello");
+            //ensure the message was received
+            autoResetEvent.WaitOne();
+            var refCell = actor as ActorRefWithCell;
+            var cell = refCell.Underlying as ActorCell;
+            //wait while current message is not null (that is, receive is not yet completed/exited)
+            SpinWait.SpinUntil(() => cell.CurrentMessage == null, TimeSpan.FromSeconds(2));
+
+            cell.CurrentMessage.ShouldBe(null);
+        }
+
+        [Fact]
+        public void Cell_should_clear_current_message_after_async_receive()
+        {
+            AutoResetEvent autoResetEvent = new AutoResetEvent(false);
+            var actor = Sys.ActorOf(Props.Create(() => new DummyAsyncActor(autoResetEvent)));
+            actor.Tell("hello");
+            //ensure the message was received
+            autoResetEvent.WaitOne();
+            var refCell = actor as ActorRefWithCell;
+            var cell = refCell.Underlying as ActorCell;
+            //wait while current message is not null (that is, receive is not yet completed/exited)
+            SpinWait.SpinUntil(() => cell.CurrentMessage == null, TimeSpan.FromSeconds(2));
+
+            cell.CurrentMessage.ShouldBe(null);
+        }
+    }
+}
--- a/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
+++ b/src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
@@ -57,10 +57,14 @@ namespace Akka.Tests.Dispatch
             {
                 state = 1;
             });
-            Receive<string>(async _ =>
+            Receive<string>(async m_ =>
             {
                 Self.Tell("change");
                 await Task.Delay(TimeSpan.FromSeconds(1));
+                var cell = (ActorCell) Context;
+                var current = cell.CurrentMessage;
+                //ensure we have the correct current message here
+                Assert.Same(m_, current);
                 //we expect that state should not have changed due to an incoming message
                 Sender.Tell(state);
             });
--- a/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
+++ b/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
@@ -60,6 +60,7 @@ namespace Akka.Actor
                     AutoReceiveMessage(envelope);
                 else
                     ReceiveMessage(message);
+                CurrentMessage = null;
             }
             catch (Exception cause)
             {
@@ -212,6 +213,9 @@ namespace Akka.Actor
 
         private void HandleActorTaskSchedulerMessage(ActorTaskSchedulerMessage m)
         {
+            //set the current message captured in the async operation
+            //current message was cleared earlier when the async receive handler completed
+            CurrentMessage = m.Message;
             if (m.Exception != null)
             {
                 HandleInvokeFailure(m.Exception);
@@ -219,6 +223,7 @@ namespace Akka.Actor
             }
 
             m.ExecuteTask();
+            CurrentMessage = null;
         }
 
         public void SwapMailbox(DeadLetterMailbox mailbox)
--- a/src/core/Akka/Dispatch/ActorTaskScheduler.cs
+++ b/src/core/Akka/Dispatch/ActorTaskScheduler.cs
@@ -17,6 +17,7 @@ namespace Akka.Dispatch
     public class ActorTaskScheduler : TaskScheduler
     {
         private readonly ActorCell _actorCell;
+        public object CurrentMessage { get; private set; }
 
         internal ActorTaskScheduler(ActorCell actorCell)
         {
@@ -57,8 +58,9 @@ namespace Akka.Dispatch
         }
 
         private void ScheduleTask(Task task)
-        {
-            _actorCell.Self.Tell(new ActorTaskSchedulerMessage(this, task), ActorRefs.NoSender);
+        {            
+            //we are in a max concurrency level 1 scheduler. reading CurrentMessage should be OK
+            _actorCell.Self.Tell(new ActorTaskSchedulerMessage(this, task, CurrentMessage), ActorRefs.NoSender);
         }
 
         internal void ExecuteTask(Task task)
@@ -95,6 +97,7 @@ namespace Akka.Dispatch
             mailbox.Suspend(MailboxSuspendStatus.AwaitingTask);
 
             ActorTaskScheduler actorScheduler = context.TaskScheduler;
+            actorScheduler.CurrentMessage = context.CurrentMessage;
 
             Task<Task>.Factory.StartNew(() => asyncAction(), CancellationToken.None, TaskCreationOptions.None, actorScheduler)
                               .Unwrap()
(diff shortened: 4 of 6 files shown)


src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -60,6 +60,7 @@ public void Invoke(Envelope envelope)
                     AutoReceiveMessage(envelope);
                 else
                     ReceiveMessage(message);
+                CurrentMessage = null;
             }
```
rogeralsing: From the Scala code

```scala
final def invoke(messageHandle: Envelope): Unit = try {
    currentMessage = messageHandle
    cancelReceiveTimeout() // FIXME: leave this here???
    messageHandle.message match {
      case msg: AutoReceivedMessage ⇒ autoReceiveMessage(messageHandle)
      case msg                      ⇒ receiveMessage(msg)
    }
    ///########## current message is cleared here
    currentMessage = null // reset current message after successful invocation
  } catch handleNonFatalOrInterruptedException { e ⇒
    handleInvokeFailure(Nil, e)
  } finally {
    checkReceiveTimeout // Reschedule receive timeout
  }
```

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -214,6 +215,9 @@ private void HandleActorTaskSchedulerMessage(ActorTaskSchedulerMessage m)
         {
             if (m.Exception != null)
             {
+                //set the current message captured in the async operation
+                //current message was cleared earlier when the async receive handler completed
+                CurrentMessage = m.Message;
                 HandleInvokeFailure(m.Exception);
```
rogeralsing: If we are handling a `ActorTaskSchedulerMessage`, we need to re-set the current message captured in the async operation. 
The actorcells current message was cleared when the async receive method completed.

src/core/Akka/Dispatch/ActorTaskScheduler.cs
```diff
@@ -93,6 +93,7 @@ public static void RunTask(Func<Task> asyncAction)
 
             //suspend the mailbox
             mailbox.Suspend(MailboxSuspendStatus.AwaitingTask);
+            var currentMessage = context.CurrentMessage;
```
rogeralsing: Capture the current message from the actor cell

src/core/Akka/Dispatch/ActorTaskScheduler.cs
```diff
@@ -109,7 +110,7 @@ public static void RunTask(Func<Task> asyncAction)
                                   }
                                   else
                                   {
-                                      context.Self.Tell(new ActorTaskSchedulerMessage(exception), ActorRefs.NoSender);
+                                      context.Self.Tell(new ActorTaskSchedulerMessage(exception,currentMessage), ActorRefs.NoSender);
```
rogeralsing: If the async operation failed, pass back the exception together with the current user message (captured above)

src/core/Akka/Dispatch/SysMsg/ISystemMessage.cs
```diff
@@ -261,12 +261,14 @@ public ActorTaskSchedulerMessage(ActorTaskScheduler scheduler, Task task)
         ///     Initializes a new instance of the <see cref="ActorTaskSchedulerMessage" /> class.
         /// </summary>
         /// <param name="exception">The exception.</param>
-        public ActorTaskSchedulerMessage(Exception exception)
+        public ActorTaskSchedulerMessage(Exception exception,object message)
         {
             Exception = exception;
+            Message = message;
         }
```
rogeralsing: `ActorTaskSchedulerMessage`  now also carries the user message that caused the exception.

rogeralsing:
Some notes.
The rationale for the implementation here is that we should stick to how Scala does this, and all the magic of async operations is dealt with externally.
We set and clear the current message on entry and exit of the `ActorCell.Invoke`.
When doing this, we also need to make sure that async operations have access to the current message in their callbacks, this is solved by capturing the current user message when an async operation starts, and then pass that message back to the actor cell using the `ActorTaskSchedulerMessage` message.

JeffCyr:
ActorCell.CurrentMessage won't be set after the first await in the async receive. It will be restored properly before calling HandleInvokeFailure in case of an exception, but it won't be set for other usage:

AbstractStash.Stash
DefaultResizer.Pressure
PersistentView.ReplayStarted

Stash won't work after the first await, not sure how it impacts the other usages.

This could be fixed by creating an ActorTaskScheduler instance per async receive, the current message would be kept in the instance so it can be restored on each ActorTaskSchedulerMessage handling.

An instance per message is not that bad for async receive, the throughput of async receive should not be high anyway. However this [issue](https://github.com/dotnet/coreclr/issues/2189) may impact the performance of this approach.

An alternative would be to use a custom ActorSynchronizationContext instead of an ActorTaskScheduler.

rogeralsing:
>ActorCell.CurrentMessage won't be set after the first await in the async receive. 

I have updated the code to restore the current message on all ActorTaskSchedulerMessage messages.

JeffCyr:
Looks good!

src/core/Akka/Dispatch/ActorTaskScheduler.cs
```diff
@@ -57,8 +57,8 @@ protected override void QueueTask(Task task)
         }
 
         private void ScheduleTask(Task task)
-        {
-            _actorCell.Self.Tell(new ActorTaskSchedulerMessage(this, task), ActorRefs.NoSender);
+        {            
+            _actorCell.Self.Tell(new ActorTaskSchedulerMessage(this, task, _actorCell.CurrentMessage), ActorRefs.NoSender);
```
JeffCyr: Hmm wait, I think this approach doesn't work if the task is scheduled from outside of the actor's context.

JeffCyr:
@rogeralsing Just saw your edit, that's right ActorTaskScheduler could capture CurrentMessage in a field as long as it is cleared when the async receive completes to not recreate the original issue.

src/core/Akka/Dispatch/ActorTaskScheduler.cs
rogeralsing: AFAIK, no one could ever use this outside of the actor context as each scheduler is bound to an individual actor and only used from within a context.
Are there any cases I am missing?

JeffCyr:
```c#
Receive(async m =>
{
    // ...
    await Task.Delay(1000);

});
```

src/core/Akka/Dispatch/ActorTaskScheduler.cs
```diff
@@ -109,9 +112,10 @@ public static void RunTask(Func<Task> asyncAction)
                                   }
                                   else
                                   {
-                                      context.Self.Tell(new ActorTaskSchedulerMessage(exception), ActorRefs.NoSender);
+                                      context.Self.Tell(new ActorTaskSchedulerMessage(exception, actorScheduler.CurrentMessage), ActorRefs.NoSender);
                                   }
-
+                                  //clear the current message field of the scheduler
+                                  actorScheduler.CurrentMessage = null;
```
rogeralsing: Clear the current message of the scheduler

src/core/Akka/Dispatch/SysMsg/ISystemMessage.cs
```diff
@@ -251,22 +251,26 @@ internal sealed class ActorTaskSchedulerMessage : ISystemMessage
...
-        public ActorTaskSchedulerMessage(Exception exception)
+        /// <param name="message">The message causing the exception</param>
+        public ActorTaskSchedulerMessage(Exception exception,object message)
         {
             Exception = exception;
+            Message = message;
         }
 
         public Exception Exception { get; private set; }
+        public object Message { get;private set; }
```
rogeralsing: Placeholder for the current user message inside the scheduler

src/core/Akka.Tests/Dispatch/AsyncAwaitSpec.cs
```diff
@@ -57,10 +57,14 @@ public SuspendActor()
...
-            Receive<string>(async _ =>
+            Receive<string>(async m_ =>
             {
                 Self.Tell("change");
                 await Task.Delay(TimeSpan.FromSeconds(1));
+                var cell = (ActorCell) Context;
+                var current = cell.CurrentMessage;
+                //ensure we have the correct current message here
+                Assert.Same(m_, current);
                 //we expect that state should not have changed due to an incoming message
```
rogeralsing: Assert that we in fact do have the correct current message after awaiting a Task.Delay

rogeralsing:
cc @JeffCyr , I've updated the scheduler to keep a field for the current message.
I've also updated the tests to ensure that we do have the correct current message after awaiting a Task.Delay.

JeffCyr:
Looks good! :thumbsup:

Aaronontheweb:
Should we add a spec to ensure that the current message is cleared?

src/core/Akka.Tests/Actor/ActorCellSpec.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        [Fact]
+        public void Cell_should_clear_current_message_after_receive()
+        {
+            AutoResetEvent autoResetEvent = new AutoResetEvent(false);
+            var actor = Sys.ActorOf(Props.Create(() => new DummyActor(autoResetEvent)));
+            actor.Tell("hello");
+            //ensure the message was received
+            autoResetEvent.WaitOne();
```
rogeralsing: wait until the actor have received our message

src/core/Akka.Tests/Actor/ActorCellSpec.cs
```diff
@@ -0,0 +1,40 @@
...
+            var refCell = actor as ActorRefWithCell;
+            var cell = refCell.Underlying as ActorCell;
```
rogeralsing: get the actorcell from the actorref

src/core/Akka.Tests/Actor/ActorCellSpec.cs
```diff
@@ -0,0 +1,40 @@
...
+            //wait while current message is not null (that is, receive is not yet completed/exited)
+            SpinWait.SpinUntil(() => cell.CurrentMessage != null, TimeSpan.FromSeconds(2));
+
+            cell.CurrentMessage.ShouldBe(null);
```
rogeralsing: wait for the current message to become null, or timeout
then assert that current message is null

rogeralsing:
@Aaronontheweb I've added a spec to ensure that the current message is cleared.

src/core/Akka.Tests/Actor/ActorCellSpec.cs
```diff
@@ -0,0 +1,68 @@
...
+            var actor = Sys.ActorOf(Props.Create(() => new DummyActor(autoResetEvent)));
+            actor.Tell("hello");
+            //ensure the message was received
+            autoResetEvent.WaitOne();
+            var refCell = actor as ActorRefWithCell;
+            var cell = refCell.Underlying as ActorCell;
+            //wait while current message is not null (that is, receive is not yet completed/exited)
+            SpinWait.SpinUntil(() => cell.CurrentMessage == null, TimeSpan.FromSeconds(2));
+
+            cell.CurrentMessage.ShouldBe(null);
```
rogeralsing: wait for current message to become null, or timeout
Then assert that current message is null

src/core/Akka.Tests/Actor/ActorCellSpec.cs
```diff
@@ -0,0 +1,68 @@
...
+            var actor = Sys.ActorOf(Props.Create(() => new DummyAsyncActor(autoResetEvent)));
+            actor.Tell("hello");
+            //ensure the message was received
+            autoResetEvent.WaitOne();
+            var refCell = actor as ActorRefWithCell;
+            var cell = refCell.Underlying as ActorCell;
+            //wait while current message is not null (that is, receive is not yet completed/exited)
+            SpinWait.SpinUntil(() => cell.CurrentMessage == null, TimeSpan.FromSeconds(2));
+
+            cell.CurrentMessage.ShouldBe(null);
```
rogeralsing: Same test as above but for async await actor

Aaronontheweb:
Looks like #1608 dirtied the working directory and failed this spec, which is bad. This is an issue with how I've configured TeamCity, not anything to do with our code. I'll get to work on fixing this.

Aaronontheweb:
All of the tests pass - will review this when I get back from a meeting

src/core/Akka.Tests/Actor/ActorCellSpec.cs
Aaronontheweb: You can use `AwaitAssert` for this in the future, but this change is fine

Aaronontheweb:
Looks fine to me.

PR #1615, opened 2016-01-05
Rewrite of the AtomicReference

Silv3rcircl3:
Based on the gitter discussion I have rewritten the AtomicReference based on the DotNetty version.

- it is now restricted to reference types
- it uses Volatile Read/Write instead of Interlocked.MemoryBarrier for the Value property
- it has a new method to atomically get the current value and replace it with a new one

### DIFF
--- a/src/core/Akka.Cluster.Tests.MultiNode/FailureDetectorPuppet.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/FailureDetectorPuppet.cs
@@ -5,10 +5,10 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System.Threading;
 using Akka.Configuration;
 using Akka.Event;
 using Akka.Remote;
-using Akka.Util;
 
 namespace Akka.Cluster.Tests.MultiNode
 {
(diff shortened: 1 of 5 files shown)


src/core/Akka.Cluster.Tests.MultiNode/FailureDetectorPuppet.cs
```diff
@@ -69,6 +69,97 @@ public override void HeartBeat()
         {
             _status.CompareAndSet(Status.Unknown, Status.Up);
         }
+
```
Silv3rcircl3: This test has used the old version with a value type, therefore I have copied the old version into this class.

src/core/Akka.Tests/IO/SimpleDnsCacheSpec.cs
```diff
@@ -79,5 +80,96 @@ public void Cache_should_be_updated_with_the_latest_resolved()
             cache.Put(cacheEntryTwo, ttl);
             cache.Cached("test.local").ShouldBe(cacheEntryTwo);
         }
+
```
Silv3rcircl3: This test has used the old version with a value type, therefore I have copied the old version into this class.

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -383,7 +383,7 @@ public TerminationCallbacks(Task upStreamTerminated)
 
             upStreamTerminated.ContinueWith(_ =>
             {
-                _terminationTask = Interlocked.Exchange(ref _atomicRef, new AtomicReference<Task>(null)).Value;
+                _terminationTask = _atomicRef.GetAndSet(null);
```
Silv3rcircl3: This was the original reason for the discussion, in the old version it wasn't possible to implement GetAndSet because Interlocked.Exchange is restricted to reference types.

rogeralsing:
Looks good to me.
cc Mr. Threading @JeffCyr

JeffCyr:
Looks good for GetAndSet, but CompareAndSet doesn't look thread safe, it should use Interlocked.CompareExchange.

JeffCyr:
I reviewed the other atomic utils and noted this:
- AtomicBoolean could be updated to also use Volatile.Read/Write
- AtomicCounter.Current should use Volatile.Read

Silv3rcircl3:
@JeffCyr I thought CompareExchange was replaced because of this null checks, but according to the history it was never implemented via CompareExchange, thx. 

Should I add the other classes to this PR as well?

JeffCyr:
@Silv3rcircl3 [Interlocked.CompareExchange](https://msdn.microsoft.com/en-us/library/bb297966(v=vs.110).aspx) returns the original value

```c#
var previous = Interlocked.CompareExchange(ref atomicValue, newValue, expected);
return Object.ReferenceEquals(previous, expected);
```

Silv3rcircl3:
@JeffCyr alright, thx for the info!

Aaronontheweb:
We'll merge this in once @JeffCyr's comments are addressed

Silv3rcircl3:
@JeffCyr @Aaronontheweb updated.

rogeralsing:
>CoreAPISpec.ApproveCore	

cc @Aaronontheweb how do we fix the above?

Silv3rcircl3:
It's working locally, don't know why it doesn't work with the CI server... 
Maybe it's this https://github.com/akkadotnet/akka.net/pull/1615/files#diff-0b623d3c0315003b5801b3c3ce2d5a4dL3770 line, but locally it says I need to change it this way. 
Could this be some sort of local language settings issue since I have a german windows + keyboard layout?

rogeralsing:
@Silv3rcircl3 could be, what happens if you revert that line and push it here again?

Silv3rcircl3:
that's it... will also update my other PR's

Horusiath:
:+1:

Aaronontheweb:
Going to merge this one in

PR #1617, opened 2016-01-06
Implemented StarTime and Uptime

Silv3rcircl3:
The JVM version has this two properties which I need in the ActorSystemSpec.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -315,8 +315,10 @@ namespace Akka.Actor
         public abstract Akka.Actor.IScheduler Scheduler { get; }
         public abstract Akka.Serialization.Serialization Serialization { get; }
         public abstract Akka.Actor.Settings Settings { get; }
+        public System.TimeSpan StartTime { get; }
         [System.ObsoleteAttribute("Use WhenTerminated instead. This property will be removed in future versions")]
         public abstract System.Threading.Tasks.Task TerminationTask { get; }
+        public System.TimeSpan Uptime { get; }
         public abstract System.Threading.Tasks.Task WhenTerminated { get; }
         public abstract Akka.Actor.IActorRef ActorOf(Akka.Actor.Props props, string name = null);
         public abstract Akka.Actor.ActorSelection ActorSelection(Akka.Actor.ActorPath actorPath);
--- a/src/core/Akka/Actor/ActorSystem.cs
+++ b/src/core/Akka/Actor/ActorSystem.cs
@@ -12,6 +12,7 @@ using Akka.Actor.Internal;
 using Akka.Configuration;
 using Akka.Dispatch;
 using Akka.Event;
+using Akka.Util;
 
 namespace Akka.Actor
 {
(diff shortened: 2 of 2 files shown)


src/core/Akka/Actor/ActorSystem.cs
```diff
@@ -74,6 +74,24 @@ public abstract class ActorSystem : IActorRefFactory, IDisposable
...
+
+        /// <summary>
+        /// Up-time of this actor system in seconds.
+        /// </summary>
+        public long Uptime => (MillisecondsSinceEpoch() - StartTime) / 1000;
+
+        /// <summary>
+        /// Returns the time in milliseconds since the epoch
+        /// </summary>
+        private static long MillisecondsSinceEpoch()
```
Silv3rcircl3: Should this be maybe moved into Utils?

src/core/Akka/Actor/ActorSystem.cs
```diff
@@ -74,6 +74,24 @@ public abstract class ActorSystem : IActorRefFactory, IDisposable
...
+        /// Start-up time in milliseconds since the epoch.
+        /// </summary>
+        public long StartTime => _startTime;
+
+        private readonly long _startTime = MillisecondsSinceEpoch();
+
+        /// <summary>
+        /// Up-time of this actor system in seconds.
+        /// </summary>
+        public long Uptime => (MillisecondsSinceEpoch() - StartTime) / 1000;
```
rogeralsing: Is seconds really the format one would expect from this?
Why not Timespan?

Silv3rcircl3: I have used the same signature as the scala version but would also prefer Timespan.

src/core/Akka/Actor/ActorSystem.cs
Aaronontheweb: You should use the methods we have in the `MonotonicClock` under utils and see if that suits your needs here. If it doesn't, then yes you should move this into a utility class in that namespace.

src/core/Akka/Actor/ActorSystem.cs
Aaronontheweb: Let's use a `TimeSpan` here - it's a value type so there shouldn't be any allocations.

Silv3rcircl3:
@Aaronontheweb @rogeralsing updated

rogeralsing:
LGTM, I'm restarting the multi node tests before merge

Aaronontheweb:
I'm going to fix that racy one soon, now that I have better logs

Aaronontheweb:
@Silv3rcircl3 this will need to be rebased on dev due to #1639 being merged in.

Bear in mind you'll need to run the unit tests locally first to generate a new .approved.txt file per the new process we're using to control how the public API gets modified. I outlined this here: http://getakka.net/docs/akka-developers/public-api-changes

Could you give that a try?

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -3765,7 +3767,7 @@ namespace Akka.Routing
     }
     public class DefaultResizer : Akka.Routing.Resizer, System.IEquatable<Akka.Routing.DefaultResizer>
     {
-        public DefaultResizer(int lower, int upper, int pressureThreshold = 1, double rampupRate = 0.2, double backoffThreshold = 0.3, double backoffRate = 0.1, int messagesPerResize = 10) { }
+        public DefaultResizer(int lower, int upper, int pressureThreshold = 1, double rampupRate = 0,2, double backoffThreshold = 0,3, double backoffRate = 0,1, int messagesPerResize = 10) { }
```
Silv3rcircl3: I don't know were this is coming from since I didn't have changed this class.

Silv3rcircl3:
@Aaronontheweb updated

Silv3rcircl3:
It's working on my machine :wink: 
image

Silv3rcircl3:
Fixed Api approve test.

PR #1620, opened 2016-01-06
Akka Android (Do Not Merge)

annymsMthd:
This PR is to discuss the proper way to add android support for akka.net


src/core/Akka/Configuration/ConfigurationFactory.cs
```diff
@@ -55,6 +55,7 @@ public static Config ParseString(string hocon)
             return ParseString(hocon, null);
         }
 
+#if !ANDROID
```
Aaronontheweb: Is it only the configuration stuff that needed to be  `IFDEF`-ed out in order for this to work?

annymsMthd: Pretty much. I'll be picking this up again within the next few days, it's
in our sprint

annymsMthd:
@Aaronontheweb the demo app works on this now. I'm creating a nuspec for this project now.

annymsMthd:
@Aaronontheweb We need to figure out a nuspec file that will bring this in with the main akka core package but targeting monoandroid

Horusiath:
Maybe we should create a separate branch for this one?

PR #1624, opened 2016-01-07
F#: Async support in actor workflow / remove Nullable object constraint in the actor workflow 

eloraiby:
This adds supports for async operations in the actor workflow for F#, allowing let! to work on async.
Also it removes the nullable constraint on disposable objects from F# workflow.


Horusiath:
The way how `let!` interops with async is questionable (it's blocking), but I think we could change it eventually. Beside that looks good.

eloraiby:
The Async.RunSynchronously has the problem of blocking the executing thread waiting for an answer. I am trying another idea by changing FunActor. Not sure this is the best way either. Help is appreciated.

Aaronontheweb:
@eloraiby @Horusiath what's the status of this going forward?

eloraiby:
@Aaronontheweb will have time in few days to continue working on it.

eloraiby:
Added  `AsyncResult (exp: Async<'Out>)  : Cont<obj, 'Out>` which is an unblocking wait for async result. same for `Rpc` which sends a message an waits for an answer from the remote agent.

Waiting for feedback/suggestions to proceed further if needed.

eloraiby:
Any update ?

src/core/Akka.FSharp/Rpc.fs
```diff
@@ -0,0 +1,105 @@
...
+            async {
+                let! result = exp
+                orig.Self.Tell (Result result, orig.Self)
+            } |> Async.Start
+    
+            let rec loop() = actor {
+                let! msg = orig.Receive()
+                let sender = orig.Sender()
+                match msg with
+                | :? AsyncResult<'Out> as m when sender.Path = orig.Self.Path ->
```
rogeralsing: Is there a reason why you dont just compare the actorrefs instead of the parths here?

src/core/Akka.FSharp/Rpc.fs
```diff
@@ -0,0 +1,105 @@
...
+                    let (Result resp) = m
+                    orig.UnstashAll()
```
rogeralsing: IMO, the F# API should use the same semantics as the C# API when it comes to async support.
Here you use the stashing support to acheive mailbox suspension, but that comes with the cost that the F# API will not be able to use stashing.

Instead, try to mimic this code but for F# async.
https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Dispatch/ActorTaskScheduler.cs#L92

You can suspend and then resume the context on start/completion of the async operation

rogeralsing: Btw. also pay attention to this one #2241, there is currently some issue in the async Task suspension.

PR #1626, opened 2016-01-08
Excludes DeadLetter from serialization verification

akoshelev:
Serialization verification in combination with Wire serializer causes ```StackOverflowException``` to be thrown as ```DeadLetter``` deserialization forces ActorSystem to send another ```DeadLetter```.


Aaronontheweb:
Hmmm.... doesn't look like the JVM uses this: https://github.com/akka/akka/blob/9972914da07a397df87ab81b457250dc40d5f53e/akka-actor/src/main/scala/akka/actor/ActorRef.scala#L480

Definitely a bug though, but it sounds like an issue with the serializer implementation itself.

akoshelev:
oops, sorry I missed this. I'll create an issue related to the serializer then.

PR #1631, opened 2016-01-10
PersistentActorSpec for Persit on ActorTaskScheduler

Zetanova:
If a Task ran before the Persist call, the journal WriteMessageSuccess Message will be queued somewhere and on actor stop pushed to deadletters.

### DIFF
--- a/src/core/Akka.Persistence.Tests/PersistentActorSpec.Actors.cs
+++ b/src/core/Akka.Persistence.Tests/PersistentActorSpec.Actors.cs
@@ -9,9 +9,11 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
@Zetanova looks like the spec you added is failing, according to the build log

Zetanova:
@Aaronontheweb its a spec for a bug

Zetanova:
It is working with only the Persist call and without the dummy-task OR with the dummy-task and without the Persist call. But with both the Persist Method will be called but the persist-callback not.
The actor will process new messages and on actor-stop the journal WriteMessageSuccess Message will be pushed to deadletters.

JeffCyr:
This bug may be caused by issue #1346 

All usages of AttachedToParent should be removed from the code base.

PR #1632, opened 2016-01-11
ActorAsk UserCancellation Extension

Zetanova:
The option to cancel an ICanTell.Ask was missing.
With this extension both timeout and user cancellation is working.

### DIFF
--- a/src/core/Akka/Actor/Futures.cs
+++ b/src/core/Akka/Actor/Futures.cs
@@ -93,13 +93,15 @@ namespace Akka.Actor
                 var cancellationSource = new CancellationTokenSource();
                 cancellationSource.Token.Register(() => result.TrySetCanceled());
                 cancellationSource.CancelAfter(t);
-                result.Task.ContinueWith(_ => cancellationSource.Dispose());
+                result.Task.ContinueWith(_ => cancellationSource.Dispose(), 
+                    CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
             }
 
             if (cancellationToken.CanBeCanceled)
             {
                 var ctr = cancellationToken.Register(() => result.TrySetCanceled());
-                result.Task.ContinueWith(_ => ctr.Dispose());
+                result.Task.ContinueWith(_ => ctr.Dispose(), 
+                    CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
             }                
         
             //create a new tempcontainer path
--- a/src/core/Akka/Actor/Futures.cs
+++ b/src/core/Akka/Actor/Futures.cs
@@ -26,17 +26,37 @@ namespace Akka.Actor
         //when asking from outside of an actor, we need to pass a system, so the FutureActor can register itself there and be resolvable for local and remote calls
         public static Task<object> Ask(this ICanTell self, object message, TimeSpan? timeout = null)
         {
-            return self.Ask<object>(message, timeout);
+            return self.Ask<object>(message, timeout, CancellationToken.None);
+        }
+
+        public static Task<object> Ask(this ICanTell self, object message, CancellationToken cancellationToken)
+        {
+            return self.Ask<object>(message, null, cancellationToken);
+        }
+
+        public static Task<object> Ask(this ICanTell self, object message, TimeSpan? timeout, CancellationToken cancellationToken)
+        {
+            return self.Ask<object>(message, timeout, cancellationToken);
         }
 
         public static Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeout = null)
+        {
+            return self.Ask<T>(message, timeout, CancellationToken.None);
+        }
+
+        public static Task<T> Ask<T>(this ICanTell self, object message, CancellationToken cancellationToken)
+        {
+            return self.Ask<T>(message, null, cancellationToken);
+        }
+
+        public static Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeout, CancellationToken cancellationToken)
         {
             IActorRefProvider provider = ResolveProvider(self);
             if (provider == null)
                 throw new NotSupportedException("Unable to resolve the target Provider");
 
             ResolveReplyTo();
-            return Ask(self, message, provider, timeout).CastTask<object, T>();
+            return Ask(self, message, provider, timeout, cancellationToken).CastTask<object, T>();
         }
 
         internal static IActorRef ResolveReplyTo()
@@ -62,19 +82,22 @@ namespace Akka.Actor
         }
 
         private static Task<object> Ask(ICanTell self, object message, IActorRefProvider provider,
-            TimeSpan? timeout)
+            TimeSpan? timeout, CancellationToken cancellationToken)
         {
             var result = new TaskCompletionSource<object>(TaskContinuationOptions.AttachedToParent);
 
-            timeout = timeout ?? provider.Settings.AskTimeout;
+            var t = timeout.GetValueOrDefault(provider.Settings.AskTimeout);
 
-            if (timeout != Timeout.InfiniteTimeSpan && timeout.Value > default(TimeSpan))
+            if (t != Timeout.InfiniteTimeSpan && t > default(TimeSpan))
             {
                 var cancellationSource = new CancellationTokenSource();
                 cancellationSource.Token.Register(() => result.TrySetCanceled());
-                cancellationSource.CancelAfter(timeout.Value);
+                cancellationSource.CancelAfter(t);
             }
 
+            if (cancellationToken.CanBeCanceled)
+                cancellationToken.Register(() => result.TrySetCanceled());
+        
             //create a new tempcontainer path
             ActorPath path = provider.TempPath();
             //callback to unregister from tempcontainer
--- a/src/core/Akka/Actor/Futures.cs
+++ b/src/core/Akka/Actor/Futures.cs
@@ -93,10 +93,14 @@ namespace Akka.Actor
                 var cancellationSource = new CancellationTokenSource();
                 cancellationSource.Token.Register(() => result.TrySetCanceled());
                 cancellationSource.CancelAfter(t);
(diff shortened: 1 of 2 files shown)


Zetanova:
Did the perf test got faster?

Aaronontheweb:
I don't think we can merge this - adds an overload to the `Ask` extension method, which will break binary compatibility with libraries built on previous versions of Akka.NET. cc @akkadotnet/contributors for second opinions.

Zetanova:
The signature of the original public extension method didnt change. The overhead of CancellationToken.None is minimal.  

What makes the better performance
to dispose the CancelationTokenSource-Timer or to let it run,
i dont know.

src/core/Akka/Actor/Futures.cs
```diff
@@ -26,17 +26,37 @@ public static class Futures
...
         {
+            return self.Ask<T>(message, timeout, CancellationToken.None);
+        }
+
+        public static Task<T> Ask<T>(this ICanTell self, object message, CancellationToken cancellationToken)
+        {
+            return self.Ask<T>(message, null, cancellationToken);
+        }
+
+        public static Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeout, CancellationToken cancellationToken)
```
JeffCyr: timeout + cancellationToken is redundant.

If the caller really needs to combine a timeout and a CancellationToken, he can use [CancellationTokenSource.CreateLinkedTokenSource](https://msdn.microsoft.com/en-us/library/dd642265(v=vs.110).aspx)

Zetanova: The caller dont know the default timeout and without explicit setting it, the default timeout should be allways used.

src/core/Akka/Actor/Futures.cs
```diff
@@ -62,19 +82,26 @@ internal static IActorRefProvider ResolveProvider(ICanTell self)
...
+            var t = timeout.GetValueOrDefault(provider.Settings.AskTimeout);
 
-            if (timeout != Timeout.InfiniteTimeSpan && timeout.Value > default(TimeSpan))
+            if (t != Timeout.InfiniteTimeSpan && t > default(TimeSpan))
             {
                 var cancellationSource = new CancellationTokenSource();
                 cancellationSource.Token.Register(() => result.TrySetCanceled());
-                cancellationSource.CancelAfter(timeout.Value);
+                cancellationSource.CancelAfter(t);
+                result.Task.ContinueWith(_ => cancellationSource.Dispose());
```
JeffCyr: Should use ExecuteSynchronously and force TaskScheduler.Default.

src/core/Akka/Actor/Futures.cs
```diff
@@ -62,19 +82,26 @@ internal static IActorRefProvider ResolveProvider(ICanTell self)
...
             }
 
+            if (cancellationToken.CanBeCanceled)
+            {
+                var ctr = cancellationToken.Register(() => result.TrySetCanceled());
+                result.Task.ContinueWith(_ => ctr.Dispose());
```
JeffCyr: Should use ExecuteSynchronously and force TaskScheduler.Default.

rogeralsing:
@Aaronontheweb 
> adds an overload to the Ask extension method, which will break binary compatibility with libraries built on previous versions of Akka.NET

How does that break binary compat?
There is a new separate extension method that does not conflict with the old one.
There is no signature change to any existing method nor any ambiguity due to default values.
Is there something I'm not seeing here?

JeffCyr:
Food for thoughts

If this extension method was provided:

```c#
public static Task<T> WithCancellation<T>(this Task<T> task, CancellationToken ct)
{
    return task.ContinueWith(t => t.Result, ct, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
}
```

You could achieve the same behavior without adding another overload:

```c#
actor.Ask<string>(msg).Wait(cancellationToken);
or
actor.Ask<string>(msg).WithCancellation(cancellationToken).Result
```

Zetanova:
The user cancelationToken is required not only to free the waiting thread,
but to stop the temp-actor and unregister the timer.

JeffCyr:
Good point about the timer unregister.

However looking at the code, it looks like the temp actor is never unregistered if there is a timeout or cancellation.

Zetanova:
Yes i saw it, have not enough inside, its the same behavior as it would timeout.
Just to call provider.UnregisterTempActor(path)? Can it be called twice?

JeffCyr:
I think the best approach would be to make only one ContinueWith with all the cleanup work and remove the unregister callback from FutureActorRef.

```c#
result.Task.ContinueWith(_ => {
    if (timeoutCancellation != null)
        timeoutCancellation.Dispose();

    if (customCancellation != null)
        customCancellation.Dispose();

    provider.UnregisterTempActor(path);

}, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
```

While you're at it, `ResolveReplyTo()` can be deleted, it's no longer used.

Aaronontheweb:
@rogeralsing 

> How does that break binary compat?

Here's a PR that made it into 1.0.5 that broke binary compatibility: https://github.com/akkadotnet/akka.net/pull/1383

Issue in this case was adding additional optional parameters to PipeTo, which utterly broke all Akka.Persistence plugins. 

So if all we're doing is adding *new extension method overloads* and not *modifying any existing extension methods* then we should be ok. Does that make sense?

trbngr:
@Aaronontheweb I believe that was a different case. #1383 broke compat because of the default parameter that was added to the existing method. If it was a simple overload, it would've been fine.

Aaronontheweb:
@trbngr that's what I meant - adding optional parameters to the existing method (effectively modifying the signature) is what broke it

Aaronontheweb:
Revisiting this again - what all needs to be cleaned up in order to accept this PR? cc @Zetanova @JeffCyr

JeffCyr:
@Aaronontheweb The concerns expressed in my last comment can be resolved in a separate issue, otherwise I think the changes are ok.

Aaronontheweb:
Still might need to run this PR through this process first:

http://getakka.net/docs/akka-developers/public-api-changes cc @Zetanova

ZigMeowNyan:
I just started looking for this functionality (proactively cancelling an Ask() call without waiting for a timeout to be reached) today and found this PR from nine months ago.

Am I correct in assuming that what's blocking the merging of this PR is these three steps?

1. Run the unit tests locally to generate files containing public API information.
2. Merge the changes from the generated `CoreAPISpec.ApproveCore.received.txt` file into the `src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt` file and include those changes in the PR. According to the approval process docs, it seems this is expected to be added into the PR by the submitter (in this case @Zetanova).
3. _Other contributors_ then confirm the expected impact of the API changes present in the PR, and ACK it if it's minimal or urgent, or NAK it if it's likely to cause problems.

I think that this expectation of submitter participation got lost in the churn. The docs aren't explicit about who does what, so confusion is entirely possible depending on English comfort level. What's more, they might not even be on Github anymore, since there's little public interaction from @Zetanova since January. If we're really just waiting for these last few steps to get the automated builds passing, then we should tag them again and give them a proper chance to finish up the merging requirements themselves. If they don't respond in a week or two, we can look into making the PR changes ourselves if the code's still admissible.

Danthar:
@ZigMeowNyan I dont think Zetanova is going to respond to this PR anymore. Its been months. You are more then welcome to pick up his PR clean it up and resubmit it.

Zetanova:
Sry, im still on vs2013, dont think that i can run the tests now.

ZigMeowNyan:
@Zetanova Would you mind if I redid this and ran the tests for you? I attempted to rebase it on dev to directly pull your commits, but there's been nearly 1000 commits since this PR was initially submitted, and that includes conflicting lines. Rebasing or cherry-picking is problematic.

If that's alright with you, I'll collapse the changes and try to get your extensions into a PR with a minimum of churn. On the other hand, if you want to do that portion of it yourself, go ahead, publish your branch and tag me and I'll assist by running the tests.

Zetanova:
@ZigMeowNyan Pls cherry-pick this small code fragment to include the user cancellation token. I will close this PR after yours is open. The actor model overwhelmed myself a bit and i got far behind. Until i dont update to vs20115, i cant run or compile the akka.net poject

ZigMeowNyan:
@Zetanova - [Here's the changes directly applied to dev](https://github.com/akkadotnet/akka.net/compare/dev...ZigMeowNyan:AskUserCancellationRebase). In this branch I tried maintaining your commits throughout the merge. There's [an alternate branch](https://github.com/akkadotnet/akka.net/compare/dev...ZigMeowNyan:ActorAsk-UserCancellation2) where I squashed all the changes into a single commit to minimize churn. I started from here to use as a reference in the merging process of the rebase.

Here's the changes:

1. The change to `CoreAPISpec.ApproveCore.approved.txt` is the result of the API policy above. This was that last little bit they were waiting for.
2. The naming convention in `AskSpec.cs` has changed from PascalCase since the original submission, so I used the same casing + underscore approach.
3. `Futures.cs` is mostly the same, but in the merge I left out this change to the timeout variable on line 76 from
`timeout = timeout ?? provider.Settings.AskTimeout;` to `var t = timeout.GetValueOrDefault(provider.Settings.AskTimeout);` and the subsequent changed lines from the use of the new variable. They should be functionally identical, so I avoided the extra churn.

Do you feel the changes in this first branch accurately represent your intent?

Zetanova:
@ZigMeowNyan Looks good, should i close this PR now?

To 3) When a property value is used more then 3 times, i am copying it into an temp variable. Its only a micro optimisation and it avoids maybe costly hidden calls.

ZigMeowNyan:
@Zetanova I put in the new PR (#2349). Let's wait and see how it's handled before we prematurely close things. Sorry for the delay - things got hectic.

ZigMeowNyan:
Code for this got merged in #2349, so it should be safe to close this now. Thanks again, @Zetanova Zetanova.

PR #1639, opened 2016-01-12
API Diff with fixed Approval file

Aaronontheweb:


Aaronontheweb:
Found a tiny, tiny issue in the `Approved.txt` file that I believe was causing the issue - had a comma instead of a period approved. cc @danielmarbach

Aaronontheweb:
Ok, got this passing now - going to add some docs to getakka.net that explain what to do in the event of a change to a public API since it adds some new steps that developers have to be aware of in order for their builds to pass.

PR #1643, opened 2016-01-13
[docs] Added blurb about updated PRs and build server

sean-gilliam:
This PR adds a section to the Contributing.md document instructing users
how to update git timestamps in order for the build server to pick up
updated PRs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,6 +109,9 @@ If we ask you to change already published commits using interactive rebase (like
 git push -f origin my-new-branch-123
 ```
 
+### The build server isn't picking up a Pull request that I've modified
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
@sean-gilliam can you run the tests locally and let me know if there are any API changes that need to be approved? Going to be a bit of a transition period here while the in-flight PRs that haven't rebased on `dev` all hit this failure http://getakka.net/docs/akka-developers/public-api-changes

sean-gilliam:
@Aaronontheweb I ran the tests as outlined in that document, but it didn't generate a received.txt file to compare against. Do the tests generate that file if no code changes are made?

Aaronontheweb:
yes, that file will always be generated but I think it gets deleted if there are no changes

Aaronontheweb:
There could also be some funkiness going on with line endings and other stuff

sean-gilliam:
gotcha

PR #1645, opened 2016-01-14
Issue1202 boekabart

boekabart:
This fixes #1202 which was caused by Remove sometimes returning an unbalanced tree. This happened in after the 'core' remove, node X became unbalanced, and either X.L.R was higher than X.L.L  or X.R.L was higher than X.R.R .

This has been reproduced in 2 unit tests.

Mostly this would not lead to a problem: a next remove or add would fix it, but in some cases (haven't been able to reproduce in a Unit Test, though) doing adds close to this tree (not to the tree itself!) would lead to a NullReferenceException when Rotating the now-unbalanced tree.

The fix is trivial: DoubleRotate should be done when the tree-after-remove is left (resp right) heavy - not only when it's left/right imbalanced. The helper functions were incorrectly named.

PR #1648, opened 2016-01-15
Fix #1522 Ensure extensions and persistence plugins are only registered/created once

### LINKED ISSUE #1522: Multiple Persistence actors of same type, different names throw Actor name \"akka.persistence.journal.inmem\" is not unique!

cconstantin:
Use ConcurrentDictionary.GetOrAdd instead of TryGet + AddOrUpdate

### DIFF
--- a/src/core/Akka.Persistence/Persistence.cs
+++ b/src/core/Akka.Persistence/Persistence.cs
@@ -13,8 +13,6 @@ using Akka.Actor;
 using Akka.Configuration;
 using Akka.Dispatch;
 using Akka.Persistence.Journal;
-using Akka.Util;
-using Akka.Util.Internal.Collections;
 
 namespace Akka.Persistence
 {
@@ -41,8 +39,8 @@ namespace Akka.Persistence
         private readonly Lazy<string> _defaultJournalPluginId;
         private readonly Lazy<string> _defaultSnapshotPluginId;
 
-        private readonly AtomicReference<IImmutableMap<string, Lazy<PluginHolder>>> _journalPluginExtensionIds = new AtomicReference<IImmutableMap<string, Lazy<PluginHolder>>>(ImmutableTreeMap<string, Lazy<PluginHolder>>.Empty);
-        private readonly AtomicReference<IImmutableMap<string, Lazy<PluginHolder>>> _snapshotPluginExtensionIds = new AtomicReference<IImmutableMap<string, Lazy<PluginHolder>>>(ImmutableTreeMap<string, Lazy<PluginHolder>>.Empty);
+        private readonly ConcurrentDictionary<string, Lazy<PluginHolder>> _journalPluginExtensionIds = new ConcurrentDictionary<string, Lazy<PluginHolder>>();
+        private readonly ConcurrentDictionary<string, Lazy<PluginHolder>> _snapshotPluginExtensionIds = new ConcurrentDictionary<string, Lazy<PluginHolder>>();
 
         public PersistenceExtension(ExtendedActorSystem system)
         {
@@ -82,14 +80,13 @@ namespace Akka.Persistence
         {
             var configPath = string.IsNullOrEmpty(snapshotPluginId) ? _defaultSnapshotPluginId.Value : snapshotPluginId;
             Lazy<PluginHolder> pluginContainer;
-            var extensionIdMap = _snapshotPluginExtensionIds.Value;
-            if (!extensionIdMap.TryGet(configPath, out pluginContainer))
+            if (!_snapshotPluginExtensionIds.TryGetValue(configPath, out pluginContainer))
             {
-                pluginContainer = new Lazy<PluginHolder>(() => CreatePlugin(configPath, _ => DefaultPluginDispatcherId), LazyThreadSafetyMode.ExecutionAndPublication);
-                _snapshotPluginExtensionIds.CompareAndSet(extensionIdMap, extensionIdMap.AddOrUpdate(configPath, pluginContainer));
-                return SnapshotStoreFor(snapshotPluginId);
+                var plugin = new Lazy<PluginHolder>(() => CreatePlugin(configPath, _ => DefaultPluginDispatcherId), LazyThreadSafetyMode.ExecutionAndPublication);
+                pluginContainer = _snapshotPluginExtensionIds.AddOrUpdate(configPath, plugin, (key, old) => plugin);
             }
-            else return pluginContainer.Value.Ref;
+
+            return pluginContainer.Value.Ref;
         }
 
         /// <summary>
@@ -100,18 +97,17 @@ namespace Akka.Persistence
         {
             var configPath = string.IsNullOrEmpty(journalPluginId) ? _defaultJournalPluginId.Value : journalPluginId;
             Lazy<PluginHolder> pluginContainer;
-            var extensionIdMap = _journalPluginExtensionIds.Value;
-            if (!extensionIdMap.TryGet(configPath, out pluginContainer))
+            if (!_journalPluginExtensionIds.TryGetValue(configPath, out pluginContainer))
             {
-                pluginContainer = new Lazy<PluginHolder>(() => CreatePlugin(configPath, type =>
+                var plugin = new Lazy<PluginHolder>(() => CreatePlugin(configPath, type =>
                     typeof (AsyncWriteJournal).IsAssignableFrom(type)
                         ? Dispatchers.DefaultDispatcherId
                         : DefaultPluginDispatcherId), 
                         LazyThreadSafetyMode.ExecutionAndPublication);
-                 _journalPluginExtensionIds.CompareAndSet(extensionIdMap, extensionIdMap.AddOrUpdate(configPath, pluginContainer));
-                return JournalFor(journalPluginId);
+                pluginContainer = _journalPluginExtensionIds.AddOrUpdate(configPath, plugin, (key, old) => plugin);
             }
-            else return pluginContainer.Value.Ref;
+
+            return pluginContainer.Value.Ref;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
@Horusiath @zetanova care to comment on this, Akka.Persistence gurus?

Zetanova:
The combination of AtomicReference TrySet+CompareAndSet(of IImmutableMap)+Get  should be thread safe. If there is a race condition then it is likly to be in likly ImmutableTreeMap or unlikly in AtomicReferenece 

Maybe fix for ImmutableTreeMap:
https://github.com/akkadotnet/akka.net/pull/1645

Maybe Fix for AttomicReference:
https://github.com/akkadotnet/akka.net/pull/1615

cconstantin:
I have a test that's calling SnapshotFor in two threads concurrently. Fails on `dev`, passes with the fix. I'll find an appropriate location for it and push it up.





On Fri, Jan 15, 2016 at 7:58 PM -0800, "Zetanova" <notifications@github.com> wrote:












The combination of AtomicReference TrySet+CompareAndSet(of IImmutableMap)+Get  should be thread safe. If there is a race condition then it is likly to be in likly ImmutableTreeMap or unlikly in AtomicReferenece 



Maybe fix for ImmutableTreeMap:

#1645



Maybe Fix for AttomicReference:

#1615

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -177,10 +177,8 @@ private void ConfigureExtensions(IEnumerable<IExtensionId> extensionIdProviders)
         public override object RegisterExtension(IExtensionId extension)
         {
             if(extension == null) return null;
-            if(!_extensions.ContainsKey(extension.ExtensionType))
-            {
-                _extensions.TryAdd(extension.ExtensionType, new Lazy<object>(() => extension.CreateExtension(this)));
-            }
```
rogeralsing: This part was the cause for race conditions in the old code.
there is a check if something does not exist, and if so, it adds an extension.. classic race condition semantics.

The new code fixes that.

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -81,15 +79,13 @@ public string PersistenceId(IActorRef actor)
...
         {
             var configPath = string.IsNullOrEmpty(snapshotPluginId) ? _defaultSnapshotPluginId.Value : snapshotPluginId;
-            Lazy<PluginHolder> pluginContainer;
-            var extensionIdMap = _snapshotPluginExtensionIds.Value;
-            if (!extensionIdMap.TryGet(configPath, out pluginContainer))
-            {
-                pluginContainer = new Lazy<PluginHolder>(() => CreatePlugin(configPath, _ => DefaultPluginDispatcherId), LazyThreadSafetyMode.ExecutionAndPublication);
-                _snapshotPluginExtensionIds.CompareAndSet(extensionIdMap, extensionIdMap.AddOrUpdate(configPath, pluginContainer));
-                return SnapshotStoreFor(snapshotPluginId);
-            }
```
rogeralsing: This part was the cause for race conditions in the old code.
there is a check if something does not exist, and if so, it adds an extension.. classic race condition semantics.

The new code fixes that.

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -99,19 +95,17 @@ public IActorRef SnapshotStoreFor(string snapshotPluginId)
         public IActorRef JournalFor(string journalPluginId)
         {
             var configPath = string.IsNullOrEmpty(journalPluginId) ? _defaultJournalPluginId.Value : journalPluginId;
-            Lazy<PluginHolder> pluginContainer;
-            var extensionIdMap = _journalPluginExtensionIds.Value;
-            if (!extensionIdMap.TryGet(configPath, out pluginContainer))
-            {
-                pluginContainer = new Lazy<PluginHolder>(() => CreatePlugin(configPath, type =>
-                    typeof (AsyncWriteJournal).IsAssignableFrom(type)
```
rogeralsing: This part was the cause for race conditions in the old code.
there is a check if something does not exist, and if so, it adds an extension.. classic race condition semantics.

The new code fixes that.

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -124,18 +118,17 @@ public IActorRef JournalFor(string journalPluginId)
         public EventAdapters AdaptersFor(string journalPluginId)
         {
             var configPath = string.IsNullOrEmpty(journalPluginId) ? _defaultJournalPluginId.Value : journalPluginId;
-            Lazy<PluginHolder> pluginContainer;
-            var extensionIdMap = _journalPluginExtensionIds.Value;
-            if (!extensionIdMap.TryGet(configPath, out pluginContainer))
-            {
-                pluginContainer = new Lazy<PluginHolder>(() =>
-                    CreatePlugin(configPath, type => typeof (AsyncWriteJournal).IsAssignableFrom(type)
```
rogeralsing: This part was the cause for race conditions in the old code.
there is a check if something does not exist, and if so, it adds an extension.. classic race condition semantics.

The new code fixes that.

rogeralsing:
The problems in the old code does not relate to atomic references or immutable lists, it comes down to standard race conditions... 

```
IF something THEN  //this 
    DOSTUFF              //is racy...
```

src/core/Akka.Persistence/Persistence.cs
Zetanova: It should be thread safe. CompareAndSet Sets only the new value if extensionIdMap is still the same. SnapshotStoreFor should return allways the valid value

src/core/Akka.Persistence/Persistence.cs
Zetanova: It should be thread safe. CompareAndSet Sets only the new value if extensionIdMap is still the same. JournalFo rshould return allways the valid value

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
Zetanova: this can be the couse. Should that "return extension.Get(this);" not actualy using the result of _extensions.GetOrAdd ?

rogeralsing: yes, it could just do a return on that call.

Aaronontheweb:
I reviewed this also and agreed with @rogeralsing - better to use a `ConcurrentDictionary` here.

Aaronontheweb:
Just for reference, in @Horusiath's original PR he talked about some of the potential trade-offs of switching to `ConcurrentDictionary`. Putting this here just in case merging this code comes back to bite us in the ass :p 

https://github.com/akkadotnet/akka.net/pull/1543

Issue #1646, opened 2016-01-14
Actor does not receive "Terminated" message if remoting is used and it is not monitored actor's parent


akoshelev:
Hi,

I am trying to implement [Reaper of Souls](http://letitcrash.com/post/30165507578/shutdown-patterns-in-akka-2) pattern in Akka.NET and when remoting is enabled my Reaper actor does not receive ```Terminated``` message from its watchees, if, at the time it receives ```WatchMe``` message, watchee is already dead, but direct parent of the watchees is receiving these messages regardless of watchee's state. Here is the code to reproduce this problem (you'll need three separate c# projects)
```csharp

/* system1.csproj */
namespace System1
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var @as = ActorSystem.Create("system1", ConfigurationFactory.ParseString(@"
                akka {  
                    log-config-on-start = on
                    stdout-loglevel = INFO
                    loglevel = INFO
                    actor {
                        provider = ""Akka.Remote.RemoteActorRefProvider, Akka.Remote""
        
                        debug {  
                          receive = on 
                          autoreceive = on
                          lifecycle = on
                          event-stream = on
                          unhandled = on
                        }

                        deployment {
                            /coordinator {
                               remote = ""akka.tcp://system2@localhost:8080""
                            }
                        }
                    }
                    remote {
                       helios.tcp {
		                   port = 8090
		                   hostname = localhost
                       }
                    }
                }
                ")))
            {
                var coordinator = @as.ActorOf(Props.Create(() => new Coordinator()), "coordinator");
                coordinator.Tell("Start");
                Console.ReadLine();
            }

        }
    }
}

/* system2.csproj */
namespace System2
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var @as = ActorSystem.Create("system2", ConfigurationFactory.ParseString(@"
                               akka {  
                    log-config-on-start = on
                    stdout-loglevel = DEBUG
                    loglevel = INFO
                    actor {
                        provider = ""Akka.Remote.RemoteActorRefProvider, Akka.Remote""
        
                        debug {  
                          receive = on 
                          autoreceive = on
                          lifecycle = on
                          event-stream = on
                          unhandled = on
                        }
                    }
                    remote {
                        helios.tcp {
		                    port = 8080
		                    hostname = localhost
                        }
                    }
                }
                ")))
            {
                Console.ReadLine();
            }

        }
    }
}

/* ClassLibrary.csproj */
[...]

PR #1649, opened 2016-01-17
Remote ActorRef Provider Fix for #1646

rogeralsing:
Fix for #1646

In the `RemoteDeadletterActorRef` there is some missing code to propagate deadletters to the base implementation:

```
.With<DeadLetter>(
    deadLetter =>
    {
        // else ignore: it is a reliably delivered message that might be retried later, and it has not yet deserved
        // the dead letter status
        var deadSend = deadLetter.Message as EndpointManager.Send;
        if (deadSend != null && deadSend.Seq == null)
        {
            base.TellInternal(deadSend.Message, deadSend.SenderOption ?? ActorRefs.NoSender);
        }
       
        //##############
       // the issue here was that the With<DeadLetter> consumed _all_ dead letters, 
       // when this clause was just intended to consume those containing EndpointManager.Send
    })
```

### DIFF
--- a/src/core/Akka.Remote/RemoteActorRefProvider.cs
+++ b/src/core/Akka.Remote/RemoteActorRefProvider.cs
@@ -532,30 +532,27 @@ namespace Akka.Remote
 
             protected override void TellInternal(object message, IActorRef sender)
             {
-                message
-                    .Match()
-                    .With<EndpointManager.Send>(
-                        send =>
-                        {
-                            // else ignore: it is a reliably delivered message that might be retried later, and it has not yet deserved
-                            // the dead letter status
-                            if (send.Seq == null)
-                            {
-                                base.TellInternal(send.Message, send.SenderOption ?? ActorRefs.NoSender);
-                            }
-                        })
-                    .With<DeadLetter>(
-                        deadLetter =>
-                        {
-                            // else ignore: it is a reliably delivered message that might be retried later, and it has not yet deserved
-                            // the dead letter status
-                            var deadSend = deadLetter.Message as EndpointManager.Send;
-                            if (deadSend != null && deadSend.Seq == null)
-                            {
-                                base.TellInternal(deadSend.Message, deadSend.SenderOption ?? ActorRefs.NoSender);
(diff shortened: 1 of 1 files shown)


src/core/Akka.Remote/RemoteActorRefProvider.cs
```diff
@@ -532,30 +532,27 @@ public RemoteDeadLetterActorRef(IActorRefProvider provider, ActorPath actorPath,
...
+                var send = message as EndpointManager.Send;
+                var deadLetter = message as DeadLetter;
+                if (send != null)
+                {
+                    if (send.Seq == null)
+                    {
+                        base.TellInternal(send.Message, send.SenderOption ?? ActorRefs.NoSender);
+                    }
+                }
+                else if (deadLetter?.Message is EndpointManager.Send)
```
rogeralsing: I've replaced the pattern matching code with normal if blocks, this specific check only consumes deadletters containing EndpointManager.Send messages now.
(unlike before where it incorrectly consumed all deadletters)

rogeralsing:
See Scala code for ref implementation

```scala
override def !(message: Any)(implicit sender: ActorRef): Unit = message match {
      case Send(m, senderOption, _, seqOpt) ⇒
        // else ignore: it is a reliably delivered message that might be retried later, and it has not yet deserved
        // the dead letter status
        if (seqOpt.isEmpty) super.!(m)(senderOption.orNull)
      case DeadLetter(Send(m, senderOption, recipient, seqOpt), _, _) ⇒
        // else ignore: it is a reliably delivered message that might be retried later, and it has not yet deserved
        // the dead letter status
        if (seqOpt.isEmpty) super.!(m)(senderOption.orNull)
      case _ ⇒ super.!(message)(sender)
    }
```

Aaronontheweb:
I've reviewed this and yep, the bug is pretty clearly a greedy match statement and @rogeralsing's fix matches the JVM source. Merging it in prior to 1.0.6.

Aaronontheweb:
Crap, hit the wrong button the first time

PR #1652, opened 2016-01-18
Fixed static Uptime context.

Silv3rcircl3:


src/core/Akka/Actor/ActorSystem.cs
```diff
@@ -82,7 +82,9 @@ public abstract class ActorSystem : IActorRefFactory, IDisposable
         /// <summary>
         /// Up-time of this actor system.
         /// </summary>
-        public TimeSpan Uptime => MonotonicClock.ElapsedHighRes;
+        public TimeSpan Uptime => MonotonicClock.ElapsedHighRes - _creationTime;
+
+        private readonly TimeSpan _creationTime = MonotonicClock.ElapsedHighRes;
```
Aaronontheweb: yep, looks right

PR #1657, opened 2016-01-20
tuned up AkkaProtocolStressTest and enabled it

Aaronontheweb:


Aaronontheweb:
Looks like there are still problems - been able to get it to pass locally. I'll investigate and do some performance tuning.

PR #1664, opened 2016-01-23
Fixes #1663

### LINKED ISSUE #1663: PersistentShard throws NRE on recovery.
Looks like we don't respect the `snapshot-after` setting. 

This causes a brand new shard to save a snapshot with null entries. Next time the system boots up, the shard with crash with NRE.

trbngr:
Added a bit of safety in recovery.
Only save snapshots after the configured amount of messages have been persisted.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
@@ -47,12 +47,12 @@ namespace Akka.Cluster.Sharding
             if (message is EntityStarted)
             {
                 var started = (EntityStarted)message;
-                State = new ShardState(State.Entries.Add(started.EntityId));
+                State = new ShardState(State.Entries?.Add(started.EntityId));
             }
             else if (message is EntityStopped)
             {
                 var stopped = (EntityStopped)message;
-                State = new ShardState(State.Entries.Remove(stopped.EntityId));
+                State = new ShardState(State.Entries?.Remove(stopped.EntityId));
             }
             else if ((offer = message as SnapshotOffer) != null && offer.Snapshot is ShardState)
             {
@@ -79,10 +79,11 @@ namespace Akka.Cluster.Sharding
         protected void SaveSnapshotIfNeeded()
         {
             PersistCount++;
-            if ((PersistCount & Settings.TunningParameters.SnapshotAfter) == 0)
+            if (PersistCount >= Settings.TunningParameters.SnapshotAfter)
             {
                 Log.Debug("Saving snapshot, sequence number [{0}]", SnapshotSequenceNr);
                 SaveSnapshot(State);
+                PersistCount = 0;
             }
         }
 


src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
```diff
@@ -79,10 +79,11 @@ protected override void ProcessChange<T>(T evt, Action<T> handler)
         protected void SaveSnapshotIfNeeded()
         {
             PersistCount++;
-            if ((PersistCount & Settings.TunningParameters.SnapshotAfter) == 0)
+            if (PersistCount >= Settings.TunningParameters.SnapshotAfter)
```
Horusiath: This will cause snapshots to be made on every even above configured threshold. The original intent was to make a snapshot every X persists, therefore it should be `%` operator instead of `>=`.

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
```diff
@@ -47,12 +47,12 @@ protected override bool ReceiveRecover(object message)
             if (message is EntityStarted)
             {
                 var started = (EntityStarted)message;
-                State = new ShardState(State.Entries.Add(started.EntityId));
+                State = new ShardState(State.Entries?.Add(started.EntityId));
```
Horusiath: I think it will be easier to make the change in the `ShardState` constructor instead of null checking everytime:

```csharp
public ShardState(IImmutableSet<EntityId> entries) : this()
{
    Entries = entries ?? ImmutableHashSet<string>.Empty;
}
```

trbngr: Tried that. The problem is with the snapshot offer having a null entries

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
trbngr: Ahh. Makes sense

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
Horusiath: This is sign of another issue, which lies [here](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs#L57). I suspect, that the reason behind that may be that `ShardState` is struct. By default when no snapshot has been found, it's `SnapshotOffer.Snapshot` field should contain null value. But could it be possible that in case of structs, that field could be `default(ShardState)` instead? In that case all null checks would be omitted while still passing an invalid state (with uninitialized entries) through.

Aaronontheweb:
will review again once @trbngr addresses @Horusiath's remarks

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
trbngr: You're absolutely right. Changing `ShardState` to a class resolved all my issues here.

I can't foresee any problems with this change. Can you?

Aaronontheweb:
looks like this needs to be rebased

trbngr:
You can just ditch this one. This PR was dealing with two separate issues. At the time I didn't realize it. I issued PRs for the two issues and they were already merged.

Horusiath:
Closing this one.

Issue #1277, opened 2015-08-24
C# 6 + Akka.???.DSL


rogeralsing:
This is related to #1263 

If we move to DNX and C# 6.
We would gain the ability to use the new "using static classes".(http://davefancher.com/2014/08/07/c-6-0-using-static/)

I propose that each namespace is extended with a static class called DSL, containing static methods relevant for that namespace.

For example.

``` csharp
namespace Akka.Actor
{  
     public static class DSL
     {
             public static Props Props<TActor>();
             public static Props Props(Type actorType ....);
             public static PoisonPill PoisonPill => Akka.Actor.PoisonPill.Instance;
     }
```

This will allow users to write code like:

``` csharp
using Akka.Actor.DSL;
...
//using just "Props<..>" instead of "Props.Create<...>"
var someActor = system.ActorOf(Props<SomeActor>());
// "PoisonPill" instead of "PoisonPill.Instance"
someActor.Tell(PoisonPill);
```

This takes us a lot closer to Scala Akka API-Wise.

The same idea can be applied to every aspect of akka. creating routers, DI etc.

KadekM:
Does it have any downsides ? I don't see one - At first I was thinking whether it's idiomatic NET but then there were no static imports before.

rogeralsing:
I have done some PoC'ing with this:
```csharp
public static class DSL
    {
        public static Akka.Actor.ActorSystem ActorSystem(string name)
        {
            return Akka.Actor.ActorSystem.Create(name);
        }

        public static Props Props<TActor>() where TActor : ActorBase
        {
            return Akka.Actor.Props.Create<TActor>();
        }

        public static object Identify(object messageId) => new Identify(messageId);
        public static object PoisonPill() => Akka.Actor.PoisonPill.Instance;
    }
```

That will let you do things like:

```csharp
using (var system = ActorSystem("mysystem"))
{
    var actor = system.ActorOf(Props<MyActor>());
    actor.Tell(Identify(null));
    actor.Tell(PoisonPill());
}
```

Which I think looks pretty clean.
Too bad you need to put () after poison pill though, otherwise it get ambiguous with the existing type PoisonPill.
But other than that, it looks like it would work w/o breaking anything else.

sean-gilliam:
This issue hasn't had any activity in 3 years and the POC from 2 years ago was closed. Closing this issue.

PR #1668, opened 2016-01-24
WIP Actor DSL

rogeralsing:
PoC of #1277

This adds some initial DSL support by using the new C#6 `using static` feature.

We can get our API closer to Scala and it is completely opt-in.. if you don't use `using static Akka.Actor.ActorDsl` then everything behaves exactly as before.

When the DSL is enabled, we get convenience features such as:

`ActorSystem(name,config)` instead of `ActorSystem.Create`
`Props<T>()` instead of `Props.Create<T>()`
`FromConfig()` instead of `FromConfig.Instance`
`Success(status)` instead of `new Status.Success(status)`
`Failure(cause)` instead of `new Status.Failure(cause)`

This can ofcourse be improved upon.
I just wanted to submit a PoC to show how it could look and start a discussion around it.

### DIFF
--- a/src/examples/RemoteDeploy/System1/Program.cs
+++ b/src/examples/RemoteDeploy/System1/Program.cs
@@ -10,6 +10,7 @@ using Akka.Actor;
 using Akka.Configuration;
 using Akka.Routing;
 using Shared;
+using static Akka.Actor.ActorDsl;
 
 namespace System1
 {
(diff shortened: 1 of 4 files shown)


src/examples/RemoteDeploy/System1/Program.cs
```diff
@@ -61,14 +62,14 @@ private static void Main(string[] args)
...
+            using (var system = ActorSystem("system1", config))
             {
                 var reply = system.ActorOf<ReplyActor>("reply");
                 //create a local group router (see config)
-                var local = system.ActorOf(Props.Create(() => new SomeActor("hello", 123)).WithRouter(FromConfig.Instance), "localactor");
+                var local = system.ActorOf(Props(() => new SomeActor("hello", 123)).WithRouter(FromConfig()), "localactor");
 
                 //create a remote deployed actor
-                var remote = system.ActorOf(Props.Create(() => new SomeActor(null, 123)).WithRouter(FromConfig.Instance), "remoteactor");
+                var remote = system.ActorOf(Props(() => new SomeActor(null, 123)).WithRouter(FromConfig()), "remoteactor");
```
rogeralsing: Added some example usage of the DSL support.

Horusiath:
:+1: Good job :)

Danthar:
What are we going to do with this ? I still think its a nice idea. But it could use an article on the getakka.net. And it could probably be expanded some more (?)

rogeralsing:
Is this worth updating?

PR #1682, opened 2016-02-01
Akka IO: ByteIterator and ByteStringbuilder bug fixes

willieferguson:
Minor bug fixes for Akka IO


Horusiath:
@willieferguson could you describe this issue a little bit more? For the sake of future generations :P

willieferguson:
Sure:

1. This was a copy/paste bug: The code cast an integer to a short, which throws when the number is larger than a short.
2. Convert.ToByte throws an Overflow exception if the argument is larger than Byte.MaxValue, which is not what we want in this case.

Horusiath:
:+1:

Issue #1694, opened 2016-02-07
Inbox.Receive always gives an ArgumentOutOfRangeException


Oipo:
The following code always gives an exception on akka 1.0.6.

```
using (var system = ActorSystem.Create("System"))
{
  var inbox = Inbox.Create(system);
  while(true)
  {
    var message = inbox.Receive();
  }
}
```

```
[ERROR][2/7/2016 10:34:39 PM][Thread 0009][akka://DispatchSystem/system/inbox-1] Delay must be >=0. It was -00:00:00.0010000
Parameter name: delay
Cause: System.ArgumentOutOfRangeException: Delay must be >=0. It was -00:00:00.0010000
Parameter name: delay
  at Akka.Actor.SchedulerBase.ValidateDelay (TimeSpan delay, System.String parameterName) <0x412a4130 + 0x000bb> in <filename unknown>:0 
  at Akka.Actor.SchedulerBase.Akka.Actor.ITellScheduler.ScheduleTellOnce (TimeSpan delay, ICanTell receiver, System.Object message, IActorRef sender, ICancelable cancelable) <0x412cc660 + 0x00037> in <filename unknown>:0 
  at Akka.Actor.SchedulerExtensions.ScheduleTellOnceCancelable (IScheduler scheduler, TimeSpan delay, ICanTell receiver, System.Object message, IActorRef sender) <0x412cc5b0 + 0x0007c> in <filename unknown>:0 
  at Akka.Actor.InboxActor.Receive (System.Object message) <0x412ca770 + 0x0064b> in <filename unknown>:0 
  at Akka.Actor.ActorBase.AroundReceive (Akka.Actor.Receive receive, System.Object message) <0x4126a310 + 0x00024> in <filename unknown>:0 
  at Akka.Actor.ActorCell.ReceiveMessage (System.Object message) <0x41269fa0 + 0x0005b> in <filename unknown>:0 
  at Akka.Actor.ActorCell.Invoke (Envelope envelope) <0x412698b0 + 0x001fb> in <filename unknown>:0 
```

Horusiath:
[This](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/Inbox.Actor.cs#L172) is actual place where error occurs. It looks like measuring time left based on scheduler's monotonic clock isn't free of issues - my guess (unconfirmed) is that time resolution unit for monotonic clock is millisecond.

[...]

PR #1702, opened 2016-02-12
#1694 Added safe check in InboxActor when receive timeout is already expired

JeffCyr:
The InboxActor could crash when scheduling a query timeout if that timeout was already expired.

I added a safety check to send a Kick to self immediately if the timeout is already expired.


annymsMthd:
Did this close the issue?

JeffCyr:
Yes, I closed the issue

PR #1707, opened 2016-02-16
Bugfix: DeadLetter filter with Type parameter should call IsInstanceOfType with the correct argument

tvsonar:


Aaronontheweb:
you'll need to rebase this on `dev` in order to get the performance tests to pass, due to #1708

tvsonar:
Thanks for the help, I rebased it.

rogeralsing:
Thanks

PR #1717, opened 2016-02-19
Akka.Persistence - update code base to akka JVM v2.4

### LINKED ISSUE #1402: Akka.Persistence - update code base to akka JVM v2.4
Since our akka persistence module has been ported, some changes occurred on the JVM side of it. One of them is introduction of `AtomicWrite` and `ReplayFilter`. Associated with them are abilities to replay corrupted events emitted by overlapping writers and to reject events before they are stored. 

**[EDIT: 2016.02.11]**

List of changes, that probably needs to be included:

- `Persist` on sequence of events should be renamed to `PersistAll` and works as `AtomicWrite` (new kind of message) - either all or none events should be written. If journal doesn't support that it should inform about it.
- Rejecting writes - jorunal can reject saving an event, informing persistent actor about it.
- Obsoleting isPermanent parameter in delete operations - it should be supported for backward compatibility, but for future versions all deletes would be considered a hard ones.
- `Deliver` support for `ActorSelection`.
- Removing `SyncJournal`.
- Renaming all messages with *Replay* prefix to corresponding *Recovery*-prefixed equivalents.
- Introduce `Recovery` config object.
- Sender reference on replayed events should be always DeadLetters.
- Remove `Saved` method from `SnapshotStore`.

Obsoleting persistent view is not possible at the moment - because we still don't have support for Akka.Streams.

cconstantin:
Resolves  #1402.

Some of the changes:

-     Persist on sequence of events should be renamed to PersistAll and works as AtomicWrite (new kind of message) - either all or none events should be written. If journal doesn't support that it should inform about it.
-     Rejecting writes - journal can reject saving an event, informing persistent actor about it.
-     Obsoleting isPermanent parameter in delete operations - it should be supported for backward compatibility, but for future versions all deletes would be considered a hard ones.
-     Deliver support for ActorSelection.
-     Removing SyncJournal.
-     Renaming all messages with Replay prefix to corresponding Recovery-prefixed equivalents.
-     Introduce Recovery config object.
-     Sender reference on replayed events should be always DeadLetters.
-     Remove Saved method from SnapshotStore.

Full list at http://doc.akka.io/docs/akka/current/project/migration-guide-2.3.x-2.4.x.html#Akka_Persistence

Only one thing left to do is to add StateChangeEvent persistence to PersistentFSM, but I believe it can be done is a subsequent release.


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/JournalDbEngine.cs
```diff
@@ -211,39 +212,58 @@ void IDisposable.Dispose()
         /// Specific table used for message persistence may be defined through configuration within 
         /// 'akka.persistence.journal.sql-server' scope with 'schema-name' and 'table-name' keys.
         /// </summary>
-        public async Task WriteMessagesAsync(IEnumerable<IPersistentRepresentation> messages)
+        public async Task<IImmutableList<Exception>> WriteMessagesAsync(IEnumerable<AtomicWrite> messages)
```
Horusiath: I'm pretty sure `IEnumerable` here is sufficient. It's immutable enough in .NET space.

src/core/Akka.Persistence.Tests/Journal/ChaosJournal.cs
```diff
@@ -53,18 +55,18 @@ public ChaosJournal()
...
 
-        public override Task ReplayMessagesAsync(string persistenceId, long fromSequenceNr, long toSequenceNr, long max, Action<IPersistentRepresentation> replayCallback)
+        public override Task ReplayMessagesAsync(IActorContext context, string persistenceId, long fromSequenceNr, long toSequenceNr, long max, Action<IPersistentRepresentation> replayCallback)
         {
+            var promise = new TaskCompletionSource<object>();
             if (ChaosSupportExtensions.ShouldFail(_replayFailureRate))
             {
                 var replays = Read(persistenceId, fromSequenceNr, toSequenceNr, max).ToArray();
-                var top = replays.Take(_random.Next(replays.Length + 1));
+                var top = replays.Take(_random.Next(replays.Length + 1)).ToArray();
```
Horusiath: Do we really need to allocate array here?

src/core/Akka.Persistence/Persistent.cs
```diff
@@ -55,28 +68,75 @@ public NonPersistentMessage(object payload, IActorRef sender)
...
-        /// </summary>
         public object Payload { get; private set; }
+        public IActorRef Sender { get; private set; }
+        public int Size { get; private set; }
+    }
 
-        /// <summary>
-        /// Sender of this message.
-        /// </summary>
+    public sealed class AtomicWrite : IPersistentEnvelope, IMessage
```
Horusiath: This will be created quite often. Any arguments against making it a `struct`?

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/JournalDbEngine.cs
cconstantin: Agreed. Will change.

cconstantin: Actually, on second thought, IEnumerable is immutable, but the underlying implementation is not guaranteed to be, so it's possible for the creator of the IEnumerable to make changes after passing the reference back. Should we ask for a second opinion @Horusiath? :)

src/core/Akka.Persistence.Tests/Journal/ChaosJournal.cs
cconstantin: If we don't, we're going to enumerate multiple times since `top` is a param of `new ReplayFailedException(top)`. This code is only used in testing, so not really critical.

src/core/Akka.Persistence/Persistent.cs
cconstantin: I'd say let's create a task to profile.

cconstantin:
Please note that the 3 failing unit tests are as a result of additional checks in TestKit. I'd recommend addressing them after merging this PR.

Horusiath:
I think it's good. Eventual changes can be implemented in separate steps. However when it comes to failing tests, they must be either fixed or skipped - without that merging this to dev will doom our CI for all other PRs waiting to be merged.

src/core/Akka.Persistence.TestKit.Tests/MemoryJournalSpec.cs
```diff
@@ -13,9 +14,11 @@ namespace Akka.Persistence.TestKit.Tests
     public class MemoryJournalSpec : JournalSpec
     {
         public MemoryJournalSpec(ITestOutputHelper output)
-            : base(actorSystemName: "MemoryJournalSpec", output: output)
+            : base(typeof(MemoryJournal), "MemoryJournalSpec")
```
Horusiath: Missing output propagation to base constructor - this will cause xUnit parallel tests to not print any results.

cconstantin:
@Horusiath should I open new issues for the two tests (one for journals, one for snapshot stores) that test behavior not supported yet?

cconstantin:
@Aaronontheweb I believe this is ready to be merged. Some work will be needed on updating the documentation, and to update the journals and event stores. There's also a bug in JVM akka-persistence that I'm following and will submit a PR as soon as they agree on a fix (it seems quite serious: https://github.com/akka/akka/issues/19828).

Horusiath:
@cconstantin please open the issues for these tests - we don't want to forget them.

cconstantin:
@Aaronontheweb @Horusiath should I squash all commits into one before merging?

Aaronontheweb:
@cconstantin yes please!

cconstantin:
@Aaronontheweb done

PR #1722, opened 2016-02-23
added remote metrics

maxim-s:


Aaronontheweb:
@maxim-s are there are any unit tests for this we should add?

maxim-s:
@Aaronontheweb done!

Aaronontheweb:
Great! I'll review this soon

src/core/Akka.Remote/RemoteMetricsExtension.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// <summary>
+    ///     INTERNAL API
+    ///     Extension that keeps track of remote metrics, such
+    ///     as max size of different message types.
+    /// </summary>
+    internal class RemoteMetricsExtension : ExtensionIdProvider<IRemoteMetrics>
+    {
+        public override IRemoteMetrics CreateExtension(ExtendedActorSystem system)
+        {
+            if (system.Settings.Config.GetString("akka.remote.log-frame-size-exceeding").ToLowerInvariant() == "off")
```
Aaronontheweb: yep, this is the same as the JVM - this feature is only really used for Akka.Remote / Cluster users who want to keep some historical track as to whether or not they're close to hitting the ceiling on the maximum frame size.

src/core/Akka.Remote/RemoteMetricsExtension.cs
```diff
@@ -0,0 +1,132 @@
...
+                {
+                    type = msg.GetType();
+                }
+
+                // 10% threshold until next log
+                var newMax = Convert.ToInt64(payloadBytes*1.1);
+                Check(type, payloadBytes, newMax);
+            }
+        }
+        private void Check(Type type, long payloadBytes, long newMax)
```
Aaronontheweb: C# probably doesn't do tail recursion as well as Scala does, so it might be better to inline this as a loop eventually. But I'll merge it as-is.

Aaronontheweb:
Awesome job @maxim-s - this is great work.

PR #1729, opened 2016-02-29
Akka.IO bind failed must notify bindCommander of failure

tstojecki:
This is to align with scala's AkkaIO:
https://github.com/akka/akka/blob/ddcfb5e4d07b07ab5221175264795235cbcd2492/akka-actor/src/main/scala/akka/io/TcpListener.scala

Origin of the problem:
When running in IIS, the app pool recycle event will likely cause another Bind command that might result in:
"Only one usage of each socket address (protocol/network address/port) is normally permitted". 

In that instance, as the Bind command runs in the ctor, the actor crashes without any notification to commander of the failure, so no easy way to detect the error.


Silv3rcircl3:
Looks good to me :+1:

Silv3rcircl3:
@tstojecki could you create an issue or another PR for that, so that we keep track of it?

PR #1730, opened 2016-02-29
Removed VsTestKit

Silv3rcircl3:


Aaronontheweb:
@Silv3rcircl3 FAKE needs to be updated here - it's still looking for VS Tests. Make sure you copy that FAKE routine over to the new VSTest repo.

Silv3rcircl3:
@Aaronontheweb I removed it from the fake script and from what I can see in the windows-build log the VsTest project wasn't build at all (like it should since I removed it :smile: ) so maybe the VsTest.dll comes from a previous build?

PR #1693, opened 2016-02-06
#1676 nuget packages

MrTortoise:
This is just the nuget portion of this issue. Got a nod from Aaron regarding this pull request.

I am working on the rest as well. Bit busy next couple of days and its larger than i expected but itl come.

My first pull request from a feature branch in git like this so if i fubar please explain and ill sort it asap ;)
I think i followed the instructions though!

Issue #1698, opened 2016-02-09
Downgrade System.Collections.Immutable to 1.1.36


Aaronontheweb:
See previous discussion here: #1676

maxwellb:
The API changes look good. I ran Performance through NBench.. some tests have standard deviations thinner or wider, but averages/maxes don't seem to jump up in any alarming increments.

https://gist.github.com/maxwellb/c1eb44422427a52fa015

Aaronontheweb:
Closed by #1733 #1693

Issue #1676, opened 2016-01-27
Unify immutable collections


Horusiath:
At the present moment we have our own implementation of immutable collections - reason behind that was that Akka.NET predates `System.Collections.Immutable` package. However right now a lot of core plugins uses the latter or even mix of both.

This is an idea to move direct dependency on `System.Collections.Immutable` into Akka.dll, refactor internals to use this library and obsolete `Akka.Util.Internal.Collections`. NBench tests should give us some insights into how this change would affect an overall performance.

@akkadotnet/developers I hope we're on the same page here.

/cc #1175

rogeralsing:
Agree, we should also try to get rid of the custom AVL tree. someone pointed out that the Dictionary in Immutable collections are also AVL trees, so it would be safer to use that IMO.

mattnischan:
This is correct. ImmutableDictionary in System.Collections.Immutable (formerly Microsoft.Bcl.Immutable) is an immutable AVL tree.

Microsoft did release these in late 2013, though, so I think this was just an oversight rather than Akka.Net predating it. There a some good blog posts [here](http://blogs.msdn.com/b/dotnet/archive/2013/09/25/immutable-collections-ready-for-prime-time.aspx) and [here](https://blogs.msdn.microsoft.com/bclteam/2013/03/19/inner-workings-of-immutable-collections-on-channel-9/).

praneetloke:
Hey guys, on that note, System.Collections.Immutable's latest version 1.1.37 seems to add a dependency on .net 5. This is preventing me from being able to add packages like akka.cluster.tools on top of akka.cluster in VS2013. Akka.cluster seems to require system.Collections.immutable >= 1.1.37. It complains that it needs nuget client 3.0 to install akka.cluster.tools, which is not available for 2013 and below. So one would have to upgrade to vs2015 to use those additions.

I can submit another issue for this but this thread seemed related. Please let me know otherwise and I'll open a new issue for a separate discussion about this.

mattnischan:
Yeah, don't use 1.1.37. I don't know what versioning they're smoking, but it sure isn't SemVer. We found the same issue on our production code where we are using System.Collections.Immutable, so we reverted to 1.1.36 as a dependency. That doesn't pull in the .Net nuget deps.

Aaronontheweb:
@praneetloke @mattnischan 

[...]

PR #1733, opened 2016-02-29
Clean rebase of 1693

Aaronontheweb:
Original commit message:

>Have depended upon current implementations enumerable over values collections being in ascending order following a load of tests on the collections i performed elsewhere. This is not explicitly documented - but does follow the standard 'walk tree' behavior so is perhaps dependable.

I did not write any of this code - all credit is due to @MrTortoise. 

All I did was use my `git` super powers to squash the commits on #1693


Aaronontheweb:
Referencing the original issue too: #1676

maxwellb:
Bah, I commented in #1698 -- there's only 5 of them ;-)

The API changes look good. I ran Performance through NBench.. some tests have standard deviations thinner or wider, but averages/maxes don't seem to jump up in any alarming increments.

https://gist.github.com/maxwellb/c1eb44422427a52fa015

As noted in the gist, there is one failing test when I ran, but it was failing on dev, too, so ymmv?

PR #1735, opened 2016-03-01
renamed test methods

### LINKED ISSUE #1510: Test method naming format
Currently there is not single convention about the naming format, some are written in [pascal case](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Tests/Actor/ActorSystemSpec.cs#L29) and others with [underscores](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Tests/Actor/ActorSystemSpec.cs#L65).

As discussed with @Horusiath and @rogeralsing, the underscore notation should be used.

Silv3rcircl3:
closes #1510


sean-gilliam:
LGTM :+1:  Looks like the CI build needs to be re-run though.

PR #1737, opened 2016-03-01
Added a new argument to the MultiNodeTestRunner to filter specs

Silv3rcircl3:
While I was testing the RemoteGatePiercingSpec I found it annoying that I need to run all specs in the assembly and therefore I have created another argument ```-Dmultinode.spec``` which takes a string and only executes specs that contain the given name. 
It checks the method name that is marked with the ```MultiNodeFact``` attribute.

For example: 
>...Runner.exe "Akka.Remote.Tests.MultiNode.dll" -Dmultinode.spec=RemoteGatePiercing


Aaronontheweb:
I know @GraemeBradbury would like this :)

PR #1742, opened 2016-03-02
Added RemoteNodeShutdownAndComesBackSpec

### LINKED ISSUE #1253: Port Akka.Remote MultiNodeSpec: RemoteNodeShutdownAndComesBackSpec
https://github.com/akka/akka/blob/master/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteNodeShutdownAndComesBackSpec.scala

Silv3rcircl3:
closes #1253

### DIFF
--- a/src/core/Akka.Remote.Tests.MultiNode/RemoteNodeShutdownAndComesBackSpec.cs
+++ b/src/core/Akka.Remote.Tests.MultiNode/RemoteNodeShutdownAndComesBackSpec.cs
@@ -1,0 +1,166 @@
+﻿using System;
+using System.Text;
+using System.Threading;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Remote.TestKit;
+using Akka.Remote.Transport;
+using Akka.TestKit;
+
+namespace Akka.Remote.Tests.MultiNode
+{
+    public class RemoteNodeShutdownAndComesBackMultiNetSpec : MultiNodeConfig
+    {
+        public RemoteNodeShutdownAndComesBackMultiNetSpec()
+        {
+            First = Role("first");
+            Second = Role("second");
+
+            CommonConfig = DebugConfig(false).WithFallback(ConfigurationFactory.ParseString(@"
+                  akka.loglevel = INFO
+                  akka.remote.log-remote-lifecycle-events = INFO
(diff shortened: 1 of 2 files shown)


src/core/Akka.Remote.Tests.MultiNode/RemoteNodeShutdownAndComesBackSpec.cs
```diff
@@ -0,0 +1,166 @@
...
+            Second = Role("second");
+
+            CommonConfig = DebugConfig(false).WithFallback(ConfigurationFactory.ParseString(@"
+                  akka.loglevel = INFO
+                  akka.remote.log-remote-lifecycle-events = INFO
+                  ## Keep it tight, otherwise reestablishing a connection takes too much time
+                  akka.remote.transport-failure-detector.heartbeat-interval = 1 s
+                  akka.remote.transport-failure-detector.acceptable-heartbeat-pause = 3 s
+                  ## the acceptable pause is too long and therefore the test will fail, it pass when we use a lower value like the default one
+                  ## akka.remote.watch-failure-detector.acceptable-heartbeat-pause = 60 s
```
Silv3rcircl3: I'm wondering how the test is passing on the jvm with this setting

Aaronontheweb: I'll take a look

Aaronontheweb: I took a look and I have to agree with you - I don't know why this value doesn't work with Akka.NET. I wonder if maybe it's an issue with the way the `watch-failure-detector` is implemented. I'll go and check that in a second.

Aaronontheweb:
Aside from the above comments about `akka.remote.watch-failure-detector.acceptable-heartbeat-pause = 60 s`, this matches what is on the JVM and is excellent work @Silv3rcircl3.

I'm going to take a look at the underlying C# classes for the remote watch failure detector and see if there's something in there that can explain the differences in configuration.

Aaronontheweb:
#1763 may explain some of the discrepancies, but I'm going to pull in this spec anyway. It'll reveal other issues that might occur as we work on fixing the cluster rejoin issues.

PR #1746, opened 2016-03-03
Implemented `guardian-supervisor-strategy` hocon setting

JeffCyr:
This PullRequest makes it possible to change the User Guardian supervision strategy with the guardian-supervisor-strategy` hocon setting.

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
@@ -7,6 +7,7 @@
 
 using System.Threading.Tasks;
 using Akka.Actor;
+using Akka.Actor.Dsl;
 using Akka.TestKit;
 using Xunit;
 using System;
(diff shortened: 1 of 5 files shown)


Silv3rcircl3:
I think this is a duplicate of #1525 :)

JeffCyr:
Doh!

Aaronontheweb:
@Silv3rcircl3 @JeffCyr yeah, but #1525 looks like it's still a work in progress - need that TestKit.EventFilter to be able to apply to multiple actor systems in order to test this, right?

Silv3rcircl3:
yep, right and theoretically this PR would also need this. I assume that's the reason why @JeffCyr has changed the implementation of these tests compared to the scala version.

JeffCyr:
@Silv3rcircl3 @Aaronontheweb I didn't understand why the scala version used the EventFilter, don't you think it is a bit zealous? The test is more about asserting that the actor has stopped than observing that an exception was thrown from the actor.

src/core/Akka.Tests/Actor/ActorSystemSpec.cs
```diff
@@ -238,6 +239,50 @@ public void Support_using_a_customer_scheduler()
...
+            {
+                actor.Receive<string>((msg, context) => { throw new Exception("Boom"); });
+            });
+
+            var probe = CreateTestProbe(system);
+            probe.Watch(a);
+
+            a.Tell("die");
+
+            var t = probe.ExpectTerminated(a);
```
Aaronontheweb: Yeah, this looks fine to me - don't think we need #1753 to complete this. The `ExpectTerminated` does the job.

src/core/Akka.Tests/Actor/ActorSystemSpec.cs
```diff
@@ -238,6 +239,50 @@ public void Support_using_a_customer_scheduler()
...
+            var system = ActorSystem.Create(Guid.NewGuid().ToString(), config);
+
+            var a = system.ActorOf(actor =>
+            {
+                actor.Receive<string>((msg, context) => { throw new Exception("Boom"); });
+            });
+
+            a.Tell("die");
+
+            Assert.True(system.WhenTerminated.Wait(1000));
```
Aaronontheweb: Yep, also looks fine without needing the `EventFilter`

src/core/Akka/Actor/ActorRefProvider.cs
```diff
@@ -155,9 +156,8 @@ public LocalActorRefProvider(string systemName, Settings settings, EventStream e
             _tempNumber = new AtomicCounterLong(1);
             _tempNode = _rootPath / "temp";
 
-            //TODO: _guardianSupervisorStrategyConfigurator = dynamicAccess.createInstanceFor[SupervisorStrategyConfigurator](settings.SupervisorStrategyClass, EmptyImmutableSeq).get
             _systemGuardianStrategy = SupervisorStrategy.DefaultStrategy;
-
+            _userGuardianStrategyConfigurator = SupervisorStrategyConfigurator.CreateConfigurator(Settings.SupervisorStrategyClass);
```
Aaronontheweb: Looks correct

Aaronontheweb:
Compared to the JVM source and it seemed complete and correct to me.

PR #1747, opened 2016-03-03
Replaced Receive(Func<T, Task> handler) by ReceiveAsync(...)

JeffCyr:
Marked Receive(Func<T, Task> handler) as obsolete and replaced it by ReceiveAsync.

This prevents from calling the asynchronous Receive by mistake (e.g. EndpointManager)

### DIFF
--- a/src/benchmark/PingPong/ClientAsyncActor.cs
+++ b/src/benchmark/PingPong/ClientAsyncActor.cs
@@ -19,7 +19,7 @@ namespace PingPong
         {
             var received = 0L;
             var sent = 0L;
-            Receive<Messages.Msg>(async m =>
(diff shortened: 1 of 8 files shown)


Aaronontheweb:
@JeffCyr need to run the API approval change process: http://getakka.net/docs/akka-developers/public-api-changes

JeffCyr:
@Aaronontheweb Done

rogeralsing:
I'm all in favor for this, its the correct design for the API.
There are a few overloads of the sync `Receive`. we should probably add those as well for the async versions.
(there are overloads with predicates etc)

JeffCyr:
I implemented the missing overloads except `protected void Receive<T>(System.Func<T, bool> handler)` which is hard to implement with current design.

Issue #1630, opened 2016-01-09
TestKit EventFilter should support specifying external ActorSystem to filter events


akoshelev:
please see https://github.com/akkadotnet/akka.net/pull/1525 for details

akoshelev:
@Aaronontheweb As ```TestKitBase.EventFilter``` is a property, not a method, I'm unable to create an overload. Would you like me to create a separate method, like ```TestKitBase.CreateEventFilter(ActorSystem)``` instead? Another way would be to create overloads for ```EventFilterFactory.Exception```, ```EventFilterFactory.DeadLetter``` etc

Aaronontheweb:
@akoshelev 

> As TestKitBase.EventFilter is a property, not a method, I'm unable to create an overload. Would you like me to create a separate method, like TestKitBase.CreateEventFilter(ActorSystem) instead?

Yes, that's a good idea - would you like to move forward with it?

Silv3rcircl3:
@akoshelev any progress on this ?

akoshelev:
@Silv3rcircl3 As I've received response from the team, I'll continue working on it

Silv3rcircl3:
@akoshelev could you maybe explain again what response you need? The only one I can see here is about the separated method which @aaronontheweb already answered.

akoshelev:
@Aaronontheweb @sadprofessor It's almost done. I've got one more question. After such refactoring, the following code is no longer valid
```csharp
        protected T Intercept<T>(Func<T> func, ActorSystem system, TimeSpan? timeout, int? expectedOccurrences, MatchedEventHandler matchedEventHandler = null)
        {
            var timeoutValue = timeout.HasValue ? _testkit.Dilated(timeout.Value) : TestKitExtension.For(system).TestEventFilterLeeway;

```
The problem lies with TestKitExtension. If I create an actor system, TestKit extensions will not be available for it, so the code above fails with NRE. I may only suggest to always use ```TestKit.Sys``` actor system here, but I'm not sure this is a good way to do it. Any ideas?

Silv3rcircl3:
@akoshelev you could check if the given system has the TestKitExtension and if so use the leeway from it or otherwise use it from the _testkit. 

```
var leeway = system.HasExtension<TestKitSettings>()
       ? TestKitExtension.For(system).TestEventFilterLeeway
       : _testkit.TestKitSettings.TestEventFilterLeeway;

[...]

PR #1753, opened 2016-03-03
External ActorSystem for Testkit event filters.

akoshelev:
This commit resolves issue described in #1630. EventFilters now can be
created for user-defined actor systems, not only TestKit.Sys.

### DIFF
--- a/src/core/Akka.TestKit/TestKitBase.cs
+++ b/src/core/Akka.TestKit/TestKitBase.cs
@@ -173,6 +173,15 @@ namespace Akka.TestKit
         /// </summary>
         public EventFilterFactory EventFilter { get { return _testState.EventFilterFactory; } }
 
+        /// <summary>
+        /// Creates a new event filter for the specified actor system.
+        /// </summary>
+        /// <param name="system">Actor system.</param>
+        /// <returns>A new instance of <see cref="EventFilterFactory"/>.</returns>
+        public EventFilterFactory CreateEventFilter(ActorSystem system)
+        {
+            return new EventFilterFactory(this, system);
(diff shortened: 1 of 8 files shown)


src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -173,6 +173,15 @@ protected void InitializeTest(ActorSystem system, Config config, string actorSys
         /// </summary>
         public EventFilterFactory EventFilter { get { return _testState.EventFilterFactory; } }
 
+        /// <summary>
+        /// Creates a new event filter for the specified actor system.
+        /// </summary>
+        /// <param name="system">Actor system.</param>
+        /// <returns>A new instance of <see cref="EventFilterFactory"/>.</returns>
+        public EventFilterFactory CreateEventFilter(ActorSystem system)
```
Aaronontheweb: Looks good - so this means that the default `EventFilter` will still work exactly as expected then. That was the only major thing I wanted to check first.

Aaronontheweb:
The code changes look fine - but could you please add a spec that verifies that the testkit works for a secondary `ActorSystem`? Want to have a spec for behavior verification against regressions in the future.

Aaronontheweb:
ping @akoshelev - did you see my comment above?

akoshelev:
@Aaronontheweb Hi, 
sorry I did not have a chance to respond. 
It's all done, I've changed existing unit tests to cover this case as well

Aaronontheweb:
thanks @akoshelev - I've reviewed it and I think it looks good. Thanks for your contribution.

Issue #1700, opened 2016-02-10
Clustered routers fail to add new routee after leader becomes unreachable, new routee node joins, leader becomes reachable, node is up


Aaronontheweb:
Been able to replicate this bug inside WebCrawler by killing off lighthouse, launching a new crawler instance, bringing lighthouse back, and watching the new crawler node join the cluster. Cluster appears to not gossip about the new node to other existing nodes.

Aaronontheweb:
@annymsMthd might want to take a look at #1189 as well

annymsMthd:
https://github.com/akkadotnet/akka.net/compare/dev...syncromatics:feature/reverted-endpoint-managment-files?diff=split&expand=1&name=feature%2Freverted-endpoint-managment-files

I reverted these files back to around 1.0.3 and the issue is fixed. Something in these changes broke the rejoin of a node to a cluster.

annymsMthd:
Looks like the changes to EndpointManager and EndpointRegistry here:
https://github.com/akkadotnet/akka.net/commit/b642984c7fdb26f11107b17aa9481d76454c72ab#diff-1ac936535300cfff2e1ff58c1ad7b75d

Still narrowing it down

annymsMthd:
Narrowed it down farther to the refuseUid stuff in the registry. When i revert the registry to the last commit before this one it works properly with changes in the manager. Going to look at how refuseUid is being used and why it would cause this issue.

maxcherednik:
I've got similar issue, but not exactly. I wonder if it is related.
Here is my scenario:

1. I have a cluster of 1 'Master' node and 3 'Slave' nodes
2. Each Slave node hosts one Actor called 'widget'
3. Master node creates a group router with following config:

`deployment {
                              /widgetmanager/widgetsRouter {
                                router = broadcast-group
                                routees.paths = ["/user/widget"]
									                  cluster {
										                  enabled = on
                                      allow-local-routees = off
										                  use-role = riskenginewidget
									                  }
								              }
                          }`

4. I am trying to send the message through the Router.
5. It's received only by one Actor
6. I kill it and messages are received by the next Actor and so on

I tried different routers - broadcast, round robin. The whole system behaves the same.

annymsMthd:
https://github.com/akkadotnet/akka.net/compare/dev...syncromatics:feature/reverted-endpoint-managment-files?expand=1

@Aaronontheweb these files changes fix the problem. Not sure exactly which part and Im wondering what features this change added.

[...]

PR #1755, opened 2016-03-04
#1700 Feature/reverted endpoint managment files to fix rejoin issues

annymsMthd:
Reverting these changes fixes cluster rejoin behaviour.  See https://github.com/akkadotnet/akka.net/issues/1700


Aaronontheweb:
Reverting isn't the right thing to do here, because this did fix other issues with Akka.Remote - need to find the bugs and correct the behavior

Aaronontheweb:
But this is helpful

annymsMthd:
Cool. Let me know when this is fixed so we can move off of 1.0.3

PR #1756, opened 2016-03-05
Ensure internal stash is unstashed on writes after recovery

cconstantin:
See https://github.com/akka/akka/issues/19899

Fix in JVM is not yet merged in, but I thought I'd get this in before next week's release. I'll submit a subsequent if JVM team makes updates to the code.


Horusiath:
It's great to be so much up-to-date ;)

PR #1757, opened 2016-03-05
Wrap user stash to avoid confusion between PersistentActor.UnstashAll…

cconstantin:
… and PersistentActor.Stash.UnstashAll

I do find it fascinating (for lack of a better word) that `Unstash()` and `UnstashAll(predicate)` do not get the same treatment as `UnstashAll()`, i.e. prepending the user stash messages to the `_internalStash`. It may come up in the future.

### DIFF
--- a/src/core/Akka.Persistence/Eventsourced.cs
+++ b/src/core/Akka.Persistence/Eventsourced.cs
@@ -76,6 +76,7 @@ namespace Akka.Persistence
 
         protected readonly PersistenceExtension Extension;
         private readonly ILoggingAdapter _log;
+        private IStash _stash;
 
         protected Eventsourced()
         {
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
All of your recent PRs are failing this test: #1762

src/core/Akka.Persistence/Eventsourced.cs
```diff
@@ -501,6 +497,52 @@ private void StashInternally(object currentMessage)
...
+            public void UnstashAll()
+            {
+                // Internally, all messages are processed by unstashing them from
+                // the internal stash one-by-one. Hence, an unstashAll() from the
+                // user stash must be prepended to the internal stash.
+
+                _internalStash.Prepend(ClearStash());
+            }
+
+            public void UnstashAll(Func<Envelope, bool> predicate)
```
Horusiath: Isn't this a little confusing? I mean, this stash consists of two internally, and while parameterless `UnstashAll` method is working on internal stash, the one with predicate works only with user stash. Maybe we should name them in a way, that is more informative about their purpose.

cconstantin: @Horusiath this matches the akka jvm, where `unstashAll` overrides method from `super`.

Horusiath: Ok. Lets settle on that.

Issue #1750, opened 2016-03-03
Racy Spec: SqliteJournalQuerySpec.Journal_queried_on_Manifest_returns_events_with_particular_manifest	


Aaronontheweb:
Fails pretty often.

https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/persistence/Akka.Persistence.Sqlite.Tests/SqliteJournalQuerySpec.cs#L13

```
Failed: Timeout 00:00:03 while waiting for a message of type Akka.Persistence.WriteMessagesSuccessful 
Expected: True
Actual:   False
   at Akka.TestKit.Xunit2.XunitAssertions.Fail(String format, Object[] args) in D:\BuildAgent\work\d26c9d7f36545acd\src\contrib\testkits\Akka.TestKit.Xunit2\XunitAssertions.cs:line 22
   at Akka.TestKit.TestKitBase.InternalExpectMsgEnvelope[T](Nullable`1 timeout, Action`2 assert, String hint, Boolean shouldLog) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.TestKit\TestKitBase_Expect.cs:line 196
   at Akka.TestKit.TestKitBase.InternalExpectMsgEnvelope[T](Nullable`1 timeout, Action`1 msgAssert, Action`1 senderAssert, String hint) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.TestKit\TestKitBase_Expect.cs:line 179
   at Akka.TestKit.TestKitBase.InternalExpectMsg[T](Nullable`1 timeout, Action`1 msgAssert, String hint) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.TestKit\TestKitBase_Expect.cs:line 154
   at Akka.TestKit.TestKitBase.ExpectMsg[T](Nullable`1 duration, String hint) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.TestKit\TestKitBase_Expect.cs:line 29
   at Akka.Persistence.Sql.Common.TestKit.SqlJournalQuerySpec.WriteEvents() in D:\BuildAgent\work\d26c9d7f36545acd\src\contrib\persistence\Akka.Persistence.Sql.Common.TestKit\SqlJournalQuerySpec.cs:line 124
[...]

PR #1760, opened 2016-03-07
Fixed global ActorContext in SqlJournal

alexvaluyskiy:
Fixed global ActorContext in SqlJournal

#1750


Horusiath:
Once this racy tests pass, it's good to go.

PR #1765, opened 2016-03-09
close #1762 changed ActorCell.FaultHandling to call correct shutdown method

### LINKED ISSUE #1762: Racy Spec: ActorSystemSpec.Reliable_deny_creation_of_actors_while_shutting_down
Starting failing consistently now - this is a recent failure.

```
Assert.Empty() Failure
   at Akka.Tests.Actor.ActorSystemSpec.Reliable_deny_creation_of_actors_while_shutting_down() in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.Tests\Actor\ActorSystemSpec.cs:line 195
« Hide stacktrace
```

Appears to have started after the Akka.Persistence 2.4 changes were merged in CC @cconstantin

Aaronontheweb:
Related to #1762 

Looks like the use of System.Collections.Immutable significantly increased the overhead of the `ChildContainer` operations, and as a result this spec (which is inherently a little bit racy) started failing on occasion.

This PR fixes a bug where we weren't calling the correct shutdown method on child actors during shutdown before and as a result, starting a bunch of actors during shutdown when we should not have been.

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
@@ -160,11 +160,15 @@ namespace Akka.Tests.Actor
             Assert.Equal(typeof(LocalActorRef), ExpectMsg<Type>());
         }
 
-        [Fact]
+        // TODO: https://github.com/akkadotnet/akka.net/issues/1979
+        [Fact(Skip = "Failing due to https://github.com/akkadotnet/akka.net/issues/1979 not being implemented")]
         public void Reliable_deny_creation_of_actors_while_shutting_down()
         {
-            var sys = ActorSystem.Create("DenyCreationWhileShuttingDone");
-            Task.Delay(500).ContinueWith(_ => sys.Terminate());
+            var sys = ActorSystem.Create("DenyCreationWhileShuttingDown");
+            sys.Scheduler.Advanced.ScheduleOnce(TimeSpan.FromMilliseconds(100), () =>
+            {
+                sys.Terminate();
+            });
             var failing = false;
             var created = new HashSet<IActorRef>();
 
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
I ended up ripping open a ton of other stuff while trying to fix this today and discarded most of those changes. Kept it pared down to the absolute simplest stuff.

Aaronontheweb:
so this is bad news - the NBench tests now take roughly 20 minutes to execute as a result of this change, which is 5 minutes more than the maximum 15 minute timeout allows.

The behavior in this spec is now correct - but the amount of overhead that goes into shutting down an `ActorSystem` has gone up significantly.

JeffCyr:
This PR caused a performance issue that is fixed by PR #1772.

PR #1772 should be merged before this one.

Danthar:
Test are still failing.

Aaronontheweb:
@Danthar haven't updated this PR yet - working on a benchmark for actor system shutdowns

JeffCyr:
Might be a good idea to put the same timeouts as the scala test
https://github.com/akka/akka/blob/b6cbc7f13a52ef20505e931419e8fb83f1706cea/akka-actor-tests/src/test/scala/akka/actor/ActorSystemSpec.scala#L290-L290

Aaronontheweb:
@JeffCyr Updated - referenced the Git issue #1979 as something that'll need to be addressed before the Reliable Deny Create on Shutdown spec will pass.

Aaronontheweb:
Had an issue with cluster deathwatch MNTR first time. Known issue: https://github.com/akkadotnet/akka.net/issues/1976

Aaronontheweb:
Interesting! this change has revealed a real problem with cluster deathwatch. Time for me to play with that some :)

Aaronontheweb:
Merging this in with another related PR....

PR #1768, opened 2016-03-09
Port StreamTestDefaultMailbox

cconstantin:
Also included are ports of LoggerMailbox and LoggerMessageQueueSemantics,
plus updates to Mailboxes.GetMailboxType to support 'mailbox-requirement'
and 'mailbox-type' in dispatcher configuration.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -1046,7 +1046,7 @@ namespace Akka.Actor
         public abstract Akka.Actor.IActorRef GetChild(System.Collections.Generic.IEnumerable<string> name);
         public abstract void Restart(System.Exception cause);
         public abstract void Resume(System.Exception causedByFailure = null);
-        public void SendSystemMessage(Akka.Dispatch.SysMsg.ISystemMessage message, Akka.Actor.IActorRef sender) { }
+        public virtual void SendSystemMessage(Akka.Dispatch.SysMsg.ISystemMessage message, Akka.Actor.IActorRef sender) { }
         public abstract void Start();
         public abstract void Stop();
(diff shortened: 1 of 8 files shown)


Horusiath:
It's good for me, but maybe @JeffCyr @Aaronontheweb or @rogeralsing have something to add?

cconstantin:
Updated Copyright notice in LoggerMailbox.cs, rest is unchanged.

cconstantin:
@Horusiath @Aaronontheweb most of the changes in this PR are not related to streams, so maybe it's better to send a PR to `dev` with only those changes, and rebase streams? Would avoid merge pain when streams is complete.

src/core/Akka/Dispatch/Mailboxes.cs
```diff
@@ -76,23 +93,53 @@ public Type GetMailboxType(Props props, Config dispatcherConfig)
...
+            var hasMailboxType = dispatcherConfig.HasPath("mailbox-type") &&
+                                 dispatcherConfig.GetString("mailbox-type") != Deploy.NoMailboxGiven;
 
-            if (interfaces.Count > 0)
+            Func<Type, Type> verifyRequirements = mailboxType =>
             {
-                var config = _requirementsMapping[interfaces.First()];
-                var mailbox = config.GetString("mailbox-type");
-                var mailboxType = Type.GetType(mailbox);
+                // TODO Akka code after introducing IProducesMessageQueue
```
JeffCyr: Do we want to enforce the requirements anytime soon? If so another issue should be created to complete this.

cconstantin: Good point. Opened #1770.

Aaronontheweb:
> most of the changes in this PR are not related to streams, so maybe it's better to send a PR to dev with only those changes

Yes :+1: - good idea

PR #1769, opened 2016-03-09
Fixed the ScatterGatherFirstCompleted router logic

### LINKED ISSUE #1718: Router with ScatterGatherFirstCompletedRoutingLogic replies with Task<object>
Hi

I think, there is a bug in ScatterGatherFirstCompletedRoutingLogic. Router with it replies with task (with message as result) instead of a message. 
Routers with TailChoppingRoutingLogic reply with normal message.

Test project:
[Program.cs.txt](https://github.com/akkadotnet/akka.net/files/137694/Program.cs.txt)

Output:

> Using ScatterGatherFirstCompletedRoutingLogic = False
> Result is Task'1
> Result is MessageB
> Using ScatterGatherFirstCompletedRoutingLogic = True
> Result is Task'1
> Result is Task'1
> Result is MessageB

If it is not a desired behaviour, maybe additional unwrapping will help at akka.net/src/core/Akka/Routing/ScatterGatherFirstCompleted.cs:87
Smth like: `var any =  Task.Run(async () => await await Task.WhenAny(tasks));`

Thanks

annymsMthd:
Fixes #1718  The spec was using broadcast which never goes through the specific router routee logic

### DIFF
--- a/src/core/Akka.Tests/Routing/ScatterGatherFirstCompletedSpec.cs
+++ b/src/core/Akka.Tests/Routing/ScatterGatherFirstCompletedSpec.cs
@@ -7,16 +7,94 @@
 
 using System;
 using System.Threading;
+using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Routing;
 using Akka.TestKit;
 using Akka.Util.Internal;
+using FluentAssertions;
 using Xunit;
 
 namespace Akka.Tests.Routing
 {
     public class ScatterGatherFirstCompletedSpec : AkkaSpec
     {
+        [Fact]
+        public void Scatter_gather_router_must_be_started_when_constructed()
+        {
+            var routedActor = Sys.ActorOf(Props.Create<TestActor>().WithRouter(new ScatterGatherFirstCompletedPool(1)));
+            ((IInternalActorRef)routedActor).IsTerminated.ShouldBe(false);
+        }
+
+        [Fact]
+        public void Scatter_gather_router_must_deliver_a_broadcast_message_using_tell()
+        {
+            var doneLatch = new TestLatch(2);
+            var counter1 = new AtomicCounter(0);
+            var counter2 = new AtomicCounter(0);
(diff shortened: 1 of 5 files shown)


src/core/Akka/Routing/ScatterGatherFirstCompleted.cs
```diff
@@ -78,14 +79,18 @@ public ScatterGatherFirstCompletedRoutees(Routee[] routees, TimeSpan within)
         /// <param name="sender">The actor sending the message.</param>
         public override void Send(object message, IActorRef sender)
         {
```
annymsMthd: I'm a bit iffy on the error handling here. What happens if one of these tasks throws?

Aaronontheweb:
@annymsMthd looks like there are public API changes in here - you'll need to follow this: http://getakka.net/docs/akka-developers/public-api-changes

annymsMthd:
I have a few upgrades im doing for this

src/core/Akka/Routing/ScatterGatherFirstCompleted.cs
annymsMthd: Fixed this by using task completion sources and Status.Failure messages

src/core/Akka.Tests/Routing/ScatterGatherFirstCompletedSpec.cs
```diff
@@ -112,27 +159,41 @@ protected override void OnReceive(object message)
...
-            var routedActor = Sys.ActorOf(new ScatterGatherFirstCompletedGroup(paths, TimeSpan.FromSeconds(3)).Props());
+            private int _id;
+            private readonly TimeSpan _waitTime;
 
-            routedActor.Tell(new Broadcast(new Stop(1)));
-            shutdownLatch.Open();
-            var res = routedActor.Ask<int>(new Broadcast(0), TimeSpan.FromSeconds(10));
-            res.Wait();
-            res.Result.ShouldBe(14);
+            public DelayStopActor(int id, TimeSpan waitTime)
```
annymsMthd: Prob don't need this. I saw that the other actor delays based on id so i can just use that functionality

Issue #1766, opened 2016-03-09
System.Collections.Immutable performance impact on Actor create / terminate


Aaronontheweb:
So the PRs implemented to resolve #1676 appear to have had a negative performance impact, as measured by NBench.

# Before
# Akka.Tests.Performance.Actor.ActorMemoryFootprintSpec+ReceiveActorMemoryFootprint
__Measures the amount of memory used by 10,000 ReceiveActors__
_2/18/2016 7:31:02 PM_
### System Info
```ini
NBench=NBench, Version=0.1.5.0, Culture=neutral, PublicKeyToken=null
OS=Microsoft Windows NT 6.2.9200.0
ProcessorCount=4
CLR=4.0.30319.42000,IsMono=False,MaxGcGeneration=2
WorkerThreads=32767, IOThreads=4
```

### NBench Settings
```ini
RunMode=Iterations, TestMode=Measurement
NumberOfIterations=13, MaximumRunTime=00:00:01
```

## Data
-------------------

### Totals
          Metric |           Units |             Max |         Average |             Min |          StdDev |
---------------- |---------------- |---------------- |---------------- |---------------- |---------------- |
TotalBytesAllocated |           bytes |   28,008,832.00 |   26,902,120.00 |   25,250,416.00 |      843,641.85 |
[Counter] ActorCreateThroughput |      operations |       10,000.00 |       10,000.00 |       10,000.00 |            0.00 |

### Per-second Totals
          Metric |       Units / s |         Max / s |     Average / s |         Min / s |      StdDev / s |
---------------- |---------------- |---------------- |---------------- |---------------- |---------------- |
TotalBytesAllocated |           bytes |   69,230,977.62 |   54,114,108.66 |   43,021,475.73 |    9,931,200.88 |
[Counter] ActorCreateThroughput |      operations |       24,717.55 |       20,062.96 |       16,390.97 |        3,260.75 |

#### TotalBytesAllocated
           Run # |           bytes |       bytes / s |      ns / bytes |
---------------- |---------------- |---------------- |---------------- |
               1 |   27,785,784.00 |   65,412,083.32 |           15.29 |
               2 |   26,527,208.00 |   43,496,367.60 |           22.99 |
               3 |   26,171,392.00 |   43,021,475.73 |           23.24 |
               4 |   27,783,040.00 |   65,285,911.84 |           15.32 |
               5 |   26,294,304.00 |   43,098,902.10 |           23.20 |
[...]

PR #1772, opened 2016-03-10
Issue #1766 - Lazy evaluation of ChildrenContainer.Children and ChildrenContainer.Stats

JeffCyr:
ChildrenContainer.Children was returning a copy of all its child. This PR change the Children and Stats property to return the lazy evaluated LINQ expression instead of a copy.

- This is safe because the underlying collection is immutable
- This is more efficient because:
  - No memory is allocated
  - Enumerating the children from the ImmutableDictionary is not expensive
  - The returned IReadOnlyCollection is not iterated in most case

The performance problem of issue #1766 was caused by this line:
https://github.com/akkadotnet/akka.net/blob/d00159ee8be45babecc7d20b935559df1415f49f/src/core/Akka/Actor/ActorCell.FaultHandling.cs#L422

(Note that the implementation of HandleChildTerminated is not touching the children parameter)

With this PR, GetChildren is lazy evaluated, so calling this method is not expensive anymore.

### DIFF
--- a/src/core/Akka/Actor/ChildrenContainer/Internal/ChildrenContainerBase.cs
+++ b/src/core/Akka/Actor/ChildrenContainer/Internal/ChildrenContainerBase.cs
@@ -5,6 +5,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
Thanks Jeff! I'll review this before I merge the other one

src/core/Akka/Actor/ChildrenContainer/Internal/ChildrenContainerBase.cs
```diff
@@ -30,25 +54,29 @@ protected ChildrenContainerBase(IImmutableDictionary<string, IChildStats> childr
...
-                return (from stat in InternalChildren.Values
-                        let childRestartStats = stat as ChildRestartStats
-                        where childRestartStats != null
-                        select childRestartStats.Child).ToList();
+                var children = from stat in InternalChildren.Values
+                               let childRestartStats = stat as ChildRestartStats
+                               where childRestartStats != null
+                               select childRestartStats.Child;
+
+                return new WrappedReadonlyCollection<IInternalActorRef>(children, InternalChildren.Count);
```
JeffCyr: Problem: InternalChildren.Count will not be exact here, it does not take the where filter into account. (Same goes for Stats property).

JeffCyr:
> Problem: InternalChildren.Count will not be exact here, it does not take the where filter into account. (Same goes for Stats property).

I changed the Count to be lazy evaluated too, worst case the children will be enumerated twice if the Count property is accessed.

PR #1773, opened 2016-03-10
[Dispatch] Support for 'mailbox-requirement' and 'mailbox-type' in dispatcher config

cconstantin:


cconstantin:
@Aaronontheweb I've added this without the tests for now, since none of the mailbox specs are ported yet. I've created an issue for porting the specs: https://github.com/akkadotnet/akka.net/issues/1793.

cconstantin:
@Aaronontheweb we reached a point in akka-streams where we'd need this. Would it be possible to "fast track" this PR? :)

Horusiath:
@cconstantin I've restarted tests. For sake of the `dev` branch, all steps have to pass.

PR #1774, opened 2016-03-10
Port  LoggerMailbox and ILoggerMessageQueueSemantics

cconstantin:


rogeralsing:
Looks good to me, no remarks on the code.
IMO we should add a test that verifies that the queue is actually drained

cconstantin:
Good point, I'll add a test early next week.

cconstantin:
@rogeralsing @Aaronontheweb test added

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2659,7 +2660,7 @@ namespace Akka.Event
         public Debug(string logSource, System.Type logClass, object message) { }
         public override Akka.Event.LogLevel LogLevel() { }
     }
-    public class DefaultLogger : Akka.Actor.ActorBase
+    public class DefaultLogger : Akka.Actor.ActorBase, Akka.Dispatch.IRequiresMessageQueue<Akka.Event.ILoggerMessageQueueSemantics>
```
Aaronontheweb: cc @Danthar @rogeralsing hmmm... think we'll need to update the logging plugins if we add this? Or should this just work automatically since it doesn't really affect the public API?

Aaronontheweb:
aha, nevermind - I see that @cconstantin has already created issues for answering my question on the diff. So we will need to update those. I'm fine with that, especially considering that we already have open issues for it.

Aaronontheweb:
Ah nuts, just thought of one more thing (right after I merged)... Don't we need to add a validation check to logger Actor classes that requires them to add these semantics?

Danthar:
Good point.

Besides that. Ill pick up the issues in logging repos coming week. Plan to get the remaining logging plugins out of the main repo as well.

-----Original Message-----

Aaronontheweb:
@cconstantin what are your thoughts on:

> Ah nuts, just thought of one more thing (right after I merged)... Don't we need to add a validation check to logger Actor classes that requires them to add these semantics?

cconstantin:
You mean a test that loggers implement the interface?

PR #1778, opened 2016-03-11
[BENCHMARK] Lazy ChildContainer with full termination sequence

Aaronontheweb:
Benchmarking #1772 in combination with #1765

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
@@ -163,8 +163,11 @@ namespace Akka.Tests.Actor
         [Fact]
         public void Reliable_deny_creation_of_actors_while_shutting_down()
         {
-            var sys = ActorSystem.Create("DenyCreationWhileShuttingDone");
-            Task.Delay(500).ContinueWith(_ => sys.Terminate());
+            var sys = ActorSystem.Create("DenyCreationWhileShuttingDown");
+            sys.Scheduler.Advanced.ScheduleOnce(TimeSpan.FromMilliseconds(100), () =>
+            {
+                sys.Terminate();
+            });
             var failing = false;
             var created = new HashSet<IActorRef>();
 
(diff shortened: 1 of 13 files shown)


src/core/Akka/Actor/ChildrenContainer/Internal/ChildrenContainerBase.cs
```diff
@@ -30,25 +54,29 @@ protected ChildrenContainerBase(IImmutableDictionary<string, IChildStats> childr
...
-        public IReadOnlyList<IInternalActorRef> Children
+        public IReadOnlyCollection<IInternalActorRef> Children
         {
             get
             {
-                return (from stat in InternalChildren.Values
-                        let childRestartStats = stat as ChildRestartStats
-                        where childRestartStats != null
-                        select childRestartStats.Child).ToList();
+                var children = from stat in InternalChildren.Values
```
annymsMthd: it's kinda a personal thing but I find this linq syntax a little meh.
can we just do something like
`
return InternalChildren.Values
                    .OfType<ChildRestartStats>()
                    .Select(stats => stats.Child)
`

src/core/Akka/Actor/ChildrenContainer/Internal/ChildrenContainerBase.cs
```diff
@@ -30,25 +54,29 @@ protected ChildrenContainerBase(IImmutableDictionary<string, IChildStats> childr
...
             }
         }
 
-        public IReadOnlyList<ChildRestartStats> Stats
+        public IReadOnlyCollection<ChildRestartStats> Stats
         {
             get
             {
-                return (from stat in InternalChildren.Values
+                var children = from stat in InternalChildren.Values
```
annymsMthd: same here. I find:
`return InternalChildren.Values.OfType<ChildRestartStats>();`
a little easier to read

src/core/Akka.Tests/Actor/ActorSystemSpec.cs
```diff
@@ -185,14 +188,15 @@ public void Reliable_deny_creation_of_actors_while_shutting_down()
                 }
 
                 
-                if (!failing && sys.Uptime.TotalSeconds >= 5)
+                if (!failing && sys.Uptime.TotalSeconds >= 10)
                     throw new AssertionFailedException(created.Last() + Environment.NewLine +
                                                        "System didn't terminate within 5 seconds");
```
Silv3rcircl3: The failure message still says 5 seconds ;)

JeffCyr:
I updated PR #1772 to fix the Children.Count that might not have been accurate.

I also changed the LINQ queries to make @annymsMthd happy :)

src/core/Akka.Tests/Actor/ActorSystemSpec.cs
Aaronontheweb: Yeah, it's that way on the JVM too. Figured they were just being generous :p

PR #1780, opened 2016-03-11
Implemented PushPullGraphLogic

Horusiath:
Changes:

- `Attributes` and `ActorAttributes` have human-friendly ToString() methods. Improved readability and debugging experience.
- Implemented empty method bodies in `PushPullGraphLogic`.
- Modified `PushPullGraphLogic` to work on `AbstractStage<TIn, TOut, ISyncDirective, ISyncDirective, IContext<TOut>, ILifecycleContext>` instead of `AbstractStage<TIn, TOut, IDirective, IDirective, IContext<TOut>, ILifecycleContext>` - due to C# generic constraint rules previous cast inside `CreateLogicAndMaterializedValue` method was throwing an error. This may need further investigation.

/cc @Silv3rcircl3 @cconstantin

### DIFF
--- a/src/core/Akka.Streams/Attributes.cs
+++ b/src/core/Akka.Streams/Attributes.cs
@@ -24,6 +24,8 @@ namespace Akka.Streams
             {
                 Value = value;
             }
+
+            public override string ToString() => $"Name({Value})";
         }
 
         public sealed class InputBuffer : IAttribute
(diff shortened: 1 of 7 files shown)


cconstantin:
I hit that cast issue as well. We probably need to declare an IStage interface with covariance on those type arguments. Also, I think `CreateLogicAndMaterializedValue` should return a `Tuple` that includes the materialized value, rather than relying on a property. Using the property might fail in async usages.

Aaronontheweb:
why is nothing compiling correctly in the akka-streams branch?

cconstantin:
I had the same issue and I fixed it by removing and re-adding references to System.Reactive.Streams project. It resulted in no actual changes in the source files, not sure what is going on ...

PR #1782, opened 2016-03-11
Akka.Cluster.TestKit (internal use only)

Aaronontheweb:
Broke out the shared Akka.Cluster multi-node testing utilities into its own Akka.Cluster.TestKit so Akka.Cluster.Sharding and Akka.Cluster.Tools can use it for their multi-node tests.

**Should cut the multi-node test runner run time in half on TeamCity**, since all specs were being run 2x before.


Aaronontheweb:
image

Shows that this PR cuts the MNTR build time by 50% since we're no longer running the assemblies twice.

annymsMthd:
So much win

PR #1786, opened 2016-03-14
Actor creation optimisation

JeffCyr:
- Cached the default mailbox type
- Replaced the ImmutableDictionary in ChildrenContainer by ConcurrentDictionary


Horusiath:
Looks like this caused some failures in dependency injection plugins.

JeffCyr:
@Horusiath Could that be a timing issue exposed by this optimization?

I'll split this PR in two, the default mailbox type optimization is pretty straightforward, the replacement with ConcurrentDictionary requires more testing/benchmark

JeffCyr:
@Horusiath It was a real bug caused by the concurrentdictionary change.

I wasn't overwriting the ChildNameReserved value in ChildrenContainer.Add. It's a bit scary that only DI tests caught the issue. I'm gonna add more tests in another PR.

PR #1791, opened 2016-03-15
Akka.Streams: fixed sln file

Horusiath:
This PR should fix build step for akka-streams branch.


Danthar:
Is it expected for the tests to fail on this? Since its the akka-streams branch. If so, i'd say merge it. Because it does fix akka-streams not building.

Aaronontheweb:
@Danthar some of these test failures can probably be fixed with a rebase on `dev` - but the others are all failures related to Streams. Those specs should be passing.

Silv3rcircl3:
@aaronontheweb most of the test aren't passing, at least the ones I ported for example mostly all IO specs in the other PR , which is fine for the moment, we will fix them laiter

Aaronontheweb:
ok, I'll take the status check requirement off

PR #1792, opened 2016-03-15
Bugfix: Assert.All takes an 'Action' and not a 'Func', so it won't ch…

tvsonar:
…eck the return value of a 'Func<?, bool>'

`Assert.All` simply executes the action on all elements of the collection inside a try catch block. So unless `Exception`s are being checked, it typically contains an `Assert` statement.


Danthar:
Needs rebase.

Aaronontheweb:
@tvsonar please rebase this PR.

tvsonar:
@Aaronontheweb sorry for the delay. It's rebased now.

Issue #1784, opened 2016-03-13
ClusterSingleton does not start after oldest node shutdown


kantora:
I have cluster of three nodes (one seed and two workers).
Worker nodes are identical by code, have the same worker role and have a configured ClusterSingleton actor for this role.
At first everything is ok, actor starts on the first worker node and accepts messages (through the proxy nodes).
Then I terminate the first worker node.
I expect that singleton actor should go up on the second worker node. But this just not happen.
The last log message I get on the worker-2 is:
`[Information] Previous oldest removed ["akka.tcp://Cluster@worker1:43314"]`
and that is all.

If I'll bring the first worker node up again - it will start the singleton. And that is strange and unexpected.

Horusiath:
Could you show us the code used to instantiate cluster singleton on each node?

kantora:
They use the same code.

```
 context.ActorOf(
                ClusterSingletonManager.Props(
                    context.System.DI().Props(type),
                    new ClusterSingletonManagerSettings(
                        singletonName,
                        role,
                        actorConfig.GetTimeSpan("removal-margin", TimeSpan.FromSeconds(1), false),
                        actorConfig.GetTimeSpan("handover-retry-interval", TimeSpan.FromSeconds(5), false))),
                pathName);
```

Where type is sngleton actor end-type. 
SingletonManager is instantiated as a child of some other actor, but they are both created under same path.

kantora:
`singletonName`, `role` and `pathName` are all the same for both nodes.

Horusiath:
This may be a bug indeed, however it's also possible that:

1. Parent actor (creating a singleton actor) is not present on each node under the same path.
2. Maybe your node is not being marked as down. In that case other nodes may not be aware, that singleton has died, therefore they won't try to respawn it on another node.

kantora:
As I've said, both nodes have identical code, so parent actor and this code is executed on both nodes.
Also seed node received ClusterEvent.MemberRemoved event about worker1.
And I suppose that log message `[Information] Previous oldest removed ["akka.tcp://Cluster@worker1:43314"]` on worker2 also indicates that it is aware of worker1 state.

zbynek001:
I've run into this as well, and I've tracked it to these issues:
1. Member.UpNumber is not being propagated across cluster
2. MemberAgeComparer is sorting in a reverse order then it should
I have this fixes, so I can create PR for this if you are interested

Horusiath:
[...]

PR #1799, opened 2016-03-17
Member.UpNumber fix

zbynek001:
Fixes issue #1784.

Member.UpNumber is not being propagated across cluster
MemberAgeComparer is sorting in a reverse order then it should

### DIFF
--- a/src/core/Akka.Cluster/Member.cs
+++ b/src/core/Akka.Cluster/Member.cs
@@ -52,9 +52,9 @@ namespace Akka.Cluster
         readonly ImmutableHashSet<string> _roles;
         public ImmutableHashSet<string> Roles { get { return _roles; } }
 
-        public static Member Create(UniqueAddress uniqueAddress, MemberStatus status, ImmutableHashSet<string> roles)
+        public static Member Create(UniqueAddress uniqueAddress, int upNumber, MemberStatus status, ImmutableHashSet<string> roles)
         {
-            return new Member(uniqueAddress, 0, status, roles);
+            return new Member(uniqueAddress, upNumber, status, roles);
         }
 
         Member(UniqueAddress uniqueAddress, int upNumber, MemberStatus status, ImmutableHashSet<string> roles)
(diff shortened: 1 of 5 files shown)


src/core/Akka.Cluster/Member.cs
```diff
@@ -52,9 +52,9 @@ public static Member Removed(UniqueAddress node)
         readonly ImmutableHashSet<string> _roles;
         public ImmutableHashSet<string> Roles { get { return _roles; } }
 
-        public static Member Create(UniqueAddress uniqueAddress, MemberStatus status, ImmutableHashSet<string> roles)
+        public static Member Create(UniqueAddress uniqueAddress, int upNumber, MemberStatus status, ImmutableHashSet<string> roles)
         {
-            return new Member(uniqueAddress, 0, status, roles);
+            return new Member(uniqueAddress, upNumber, status, roles);
```
Aaronontheweb: Oh wow, **nice** catch. I could see this being responsible for some fun issues with Cluster rejoin scenarios too.

Aaronontheweb:
Looks like the timestamp on this commit is 29 days old, so it didn't show up on TeamCity - would you mind changing the timestamp on your commit? http://stackoverflow.com/a/31540373/377476

This will allow our CI system to scoop it up. Great job @zbynek001 :+1:

Aaronontheweb:
This is ready for merge once CI passes - it fixes an obvious bug in Akka.Cluster but the test suite might reveal others, and possibly faulty assumptions with the test suite.

src/core/Akka/Actor/ActorPath.cs
```diff
@@ -254,7 +254,8 @@ public bool Equals(ActorPath other)
             var a = path;
             foreach (string element in name)
             {
-                a = a / element;
+                if(!string.IsNullOrEmpty(element))
+                    a = a / element;
```
zbynek001: I've also added fix for ActorPath operator / used in singleton proxy. This way it's also handled in akka

PR #1809, opened 2016-03-20
Lazy Remote Sender WIP

rogeralsing:
This is an experiment to see if we can lazily resolve Remote senders.
Instead of resolving the remote sender directly, we instead pack the remote path inside a new `LazyActorRef` which will resolve the remote actor once any members have been touched.

This could potentially have a positive performance impact on Akka.Remote as most actor code never touch the Sender property. 

cc @JeffCyr

### DIFF
--- a/src/core/Akka.Remote/RemoteActorRef.cs
+++ b/src/core/Akka.Remote/RemoteActorRef.cs
@@ -9,6 +9,7 @@ using System;
 using System.Collections.Generic;
 using Akka.Actor;
 using Akka.Dispatch.SysMsg;
+using Akka.Util;
 
 namespace Akka.Remote
 {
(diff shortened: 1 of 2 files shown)


src/core/Akka.Remote/RemoteActorRef.cs
```diff
@@ -177,5 +178,42 @@ public override void Start()
...
         }
     }
+
+    public class LazyActorRef : IActorRef
+    {
+        private readonly Lazy<IActorRef> _lazyActorRef;
+
+        public LazyActorRef(RemoteActorRefProvider provider,Address address, string path)
+        {
+            _lazyActorRef = new Lazy<IActorRef>(() => provider.ResolveActorRefWithLocalAddress(path,address));
```
rogeralsing: This could have been agnostic of Akka Remote and just take the Lazy<IActorRef> instead.
But never mind that right now. I just want to see the perf impact on the experimental Remote Transport.

Aaronontheweb:
Hah! That's a clever idea - I like it

Horusiath:
Multinode tests are failing constantly on this one.

JeffCyr:
This PR is obsoleted by https://github.com/akkadotnet/akka.net/pull/1973

Danthar:
Then im closing it.

Issue #1695, opened 2016-02-08
Things to obsolete before v1.5 


Horusiath:
For some time we were talking about things that we should probably drop when 1.5 will came out. Until that time, there was an idea to inform about incoming changes in form of marking necessary classes as `Obsolete` when possible, and where the changes are not explicitly visible in the code - in form of warning log. This could look similar to the [JVM akka note about JavaSerializer usage](https://github.com/akka/akka/blob/master/akka-actor/src/main/resources/reference.conf#L576).
#### List of things to change

Note: not all of these changes can be made at the present moment, as some parts are still WIP.
- **Obsoleting `TypedActor`** - they are considered very bad for the performance reasons and heavily limit some of the Akka features (like Finite State Machines). 
- **Obsoleting `Inbox`** - it has a very limited usage and can be easily replaced with other mechanisms. Effectivelly just like in case of `TypedActor` it's a legacy feature.
- **Switching default serializer from Newtonsoft.Json &rArr; Wire** - JSON.NET was source of many issues. Ultimately it's not able to fit Akka.NET requirements. It's performance in this specific scenarios is also very questionable. Therefore we already encourage people to change it to dedicated serializers and are going to stop using JSON.NET due to number of issues it caused already.
- **Replacing our internal immutable collections library with System.Collections.Immutable** (see #1676) - this also becomed a leagacy code, as we're already using System.Collecions.Immutable in most of the core plugins.
- **Replacing Helios transport layer with DotNetty** (see #1465) - ultimately Helios didn't create such community as Akka.NET, making it expensive to maintain and develop. In the meantime Azure team has already open sourced it's [DotNetty](https://github.com/Azure/DotNetty) library, which perfectly serves our purpose.
- **Upgrading Akka.Persistence** (see #1402) - current version of this plugin is compatible with it's JVM counterpart from v2.3. From that time things have changed. There is a possibility for the breaking changes on that field as well.
- ~~**Changing Dependency Injection API** (see #1675) - as already discussed, current DI API lacks some features. Modifying it will be also a breaking change.~~
#### Why on 1.5?

To make it clear - Akka.NET is not using SemVer at the moment. The reason behind choosing version 1.5 to introduce changes was, that not all akkadotnet team members were convinced, that they are major API breaking ones - most of them is concerning underlying stack, not API itself - to require a major number version change. While we have some plans for v2.0 API already, we decided to follow the convention introduced by .NET team to introduce similar kind of changes in X.5 version.

[...]

PR #1811, opened 2016-03-20
Warning about JSON.NET obsolete in v1.5

Horusiath:
This PR includes information about obsoleting JSON.NET as the default serializer in Akka.NET version 1.5. (see #1695). Warning contains link to manual informing, how to change default serializer to Wire and info about how to suppress this warning message by setting `suppress-json-serializer-warning` config key to on.

### DIFF
--- a/src/core/Akka/Actor/Internal/ActorSystemImpl.cs
+++ b/src/core/Akka/Actor/Internal/ActorSystemImpl.cs
@@ -15,6 +15,7 @@ using Akka.Configuration;
 using Akka.Dispatch;
 using Akka.Dispatch.SysMsg;
 using Akka.Event;
+using Akka.Serialization;
 using Akka.Util;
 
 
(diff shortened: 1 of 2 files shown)


alexvaluyskiy:
Are you sure that Wire is ready for production use? 

For example I have this class and I try to serialize and deserialize it

```C#
    public class TestSerialization
    {
        public TestSerialization(byte[] document)
        {
            Document = document;
            DocumentReadOnly = new ReadOnlyCollection<byte>(document);
            DocumentImmutable = ImmutableArray.Create(document);
        }

        public byte[] Document { get; }

        public ImmutableArray<byte> DocumentImmutable { get; }

        public IReadOnlyList<byte> DocumentReadOnly { get; }
    }

    var expected = new TestSerialization(new byte[] { 2, 6, 8, 9, 3 });
```
I get this result
```C#
Document = [5, 0, 0, 0, 0, 0, 0, 0]
DocumentReadOnly = [2, 0, 6, 0, 8]
DocumentImmutable = null
```

But if I add a string property to TestSerialization class and try again I will get an exception {"Object of type 'System.Byte' cannot be converted to type 'System.String'."}
``` C#
public string Title { get; } = "Test string";
```

There is related bug https://github.com/rogeralsing/Wire/issues/18

rogeralsing:
The issue have been fixed in Wire.
Also a suggestion, if you have test code as the above, please submit that as a PR to Wire instead.

PR #1812, opened 2016-03-21
Completed port of ActorGraphInterpreterSpec

cconstantin:
Included are:
* Update to Graph Dsl to use `combineMaterializers` func
* Bug fix in `GraphAssembly` contructor
* Bug fix `StreamLayout.MaterializeModule`
* Bug fix in `GraphStageLogic.PassAlong`
* Bug fix in `FanOutShape` constructor

### DIFF
--- a/src/core/Akka.Streams/CodeGen/Dsl/GraphApply.cs
+++ b/src/core/Akka.Streams/CodeGen/Dsl/GraphApply.cs
@@ -1,8 +1,5 @@
 ﻿// --- auto generated: 05.02.2016 09:30:42 --- //
 using System;
-using System.Linq;
-using System.Reactive.Streams;
-using Akka.Streams.Implementation;
 using Akka.Streams.Dsl.Internal;
 
 namespace Akka.Streams.Dsl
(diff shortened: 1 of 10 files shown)


src/core/Akka.Streams/CodeGen/Dsl/GraphApply.cs
```diff
@@ -68,18 +65,18 @@ public partial class GraphDsl
...
 			where TShape2: Shape
 			
 		{
-			var builder = new GraphDsl.Builder<TMatOut>();
+			var builder = new Builder<TMatOut>();
 			
-			var shape0 = builder.Add(g0);
-			var shape1 = builder.Add(g1);
-			var shape2 = builder.Add(g2);
+			var shape0 = builder.Add<TShape0, TMat0, Func<TMat1, TMat2, TMatOut>>(g0, m0 => ((m1, m2) => combineMaterializers(m0, m1, m2)));
```
Horusiath: This generic type explicitness looks bad from usability point of view. Any ideas how we could describe it in such way, that compiler will be able to infer types?

cconstantin: Not just yet, I plan on tackling it once all tests pass.

PR #1816, opened 2016-03-23
Port GraphInterpreterPortsSpec, GraphInterpreterFailureModesSpec and InterpreterSupervisionSpec

cconstantin:


src/core/Akka.Streams.TestKit.Tests/Utils.cs
```diff
@@ -18,6 +18,24 @@ public static class Utils
         public class TE : SystemException
```
Horusiath: We should avoid abbreviated/cryptic names. Maybe just `TestException`?

Silv3rcircl3: The name comes from the jvm but I would also prefer a more meaningful name.

Horusiath: I'll merge it now. We can change this before releasing this module.

PR #1817, opened 2016-03-23
ActorPath UID fix

rogeralsing:
This PR fixes the issue with missing UID tags when using remote actorrefs..
There was a missing method in `ActorPath` `ToSerializationFormatWithAddress`, our code instead used the `ToStringWithAddress` which omits the UID.

This corrects the most immediate issues in Akka.Remote.
However there are a lot of uses of `ToStringWithAddress` inside Akka.Cluster and other modules, that probably should use the new method. (imo out of scope for this PR)

### DIFF
--- a/src/core/Akka.Remote.Tests/RemoteMessageLocalDeliverySpec.cs
+++ b/src/core/Akka.Remote.Tests/RemoteMessageLocalDeliverySpec.cs
@@ -86,7 +86,6 @@ namespace Akka.Remote.Tests
                     var ai =
                         Sys.ActorSelection(actorPath).Ask<ActorIdentity>(new Identify(null), TimeSpan.FromSeconds(3)).Result;
 
-                   
                     remoteActorRef.Tell(PoisonPill.Instance); // WATCH should be applied first
                     ExpectTerminated(remoteActorRef);
                 });
--- a/src/core/Akka/Actor/ActorPath.cs
+++ b/src/core/Akka/Actor/ActorPath.cs
@@ -10,6 +10,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Akka.Util;
 using Newtonsoft.Json;
+using static System.String;
 
 namespace Akka.Actor
 {
@@ -37,7 +38,7 @@ namespace Akka.Actor
                 Path = path;
             }
 
-            public string Path { get; private set; }
+            public string Path { get; }
 
             public ISurrogated FromSurrogate(ActorSystem system)
             {
@@ -49,22 +50,6 @@ namespace Akka.Actor
 
                 return null;
             }
-
-            #region Implicit conversion operators
-
-            public static implicit operator ActorPath(Surrogate surrogate)
-            {
-                ActorPath parse;
-                TryParse(surrogate.Path, out parse);
-                return parse;
-            }
-
-            public static implicit operator Surrogate(ActorPath path)
-            {
-                return (Surrogate)path.ToSurrogate(null);
-            }
-
-            #endregion
 
             #region Equality
 
@@ -78,7 +63,7 @@ namespace Akka.Actor
             public bool Equals(ActorPath other)
             {
                 if (other == null) return false;
-                return Equals(other.ToSurrogate(null));
+                return Equals(other.ToSurrogate(null)); //TODO: not so sure if this is OK
             }
 
             public override bool Equals(object obj)
@@ -86,8 +71,8 @@ namespace Akka.Actor
                 if (ReferenceEquals(null, obj)) return false;
                 if (ReferenceEquals(this, obj)) return true;
                 var actorPath = obj as ActorPath;
-
-                return Equals(actorPath);
+                if (actorPath != null) return Equals(actorPath);
+                return Equals(obj as Surrogate);
             }
 
             public override int GetHashCode()
@@ -98,7 +83,7 @@ namespace Akka.Actor
             #endregion
         }
 
-         /** INTERNAL API */
+        /** INTERNAL API */
         internal static char[] ValidSymbols = @"""-_.*$+:@&=,!~';""()".ToCharArray();
 
         /// <summary> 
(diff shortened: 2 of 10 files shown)


Horusiath:
I think - to make everything clear - we should ask here, if there are any backward compatibility issues? I.e. in case of incremental cluster upgrade, are we able to communicate between nodes having this fix and the ones without it?

Aaronontheweb:
@horusiath yes, it would cause a breaking change - this should go into the 1.5 branch then

rogeralsing:
The old code _is_ able to process paths with and without UID, e.g. actor resolution handles paths both with and without UID.
The issue was that the UID was not included at serialization level.

JeffCyr:
I think the fix is backward compatible. But it made a resolved RemoteActorRef behave like an ActorSelection, so the resolved RemoteActorRef was still working when it was not supposed to (if the remote actor was stopped and started with the same name)

So there is a behavior change that could create a bug, but this is an edge case.

src/core/Akka/Actor/ActorPath.cs
```diff
@@ -86,8 +86,8 @@ public override bool Equals(object obj)
                 if (ReferenceEquals(null, obj)) return false;
                 if (ReferenceEquals(this, obj)) return true;
                 var actorPath = obj as ActorPath;
-
-                return Equals(actorPath);
+                if (actorPath != null) return Equals(actorPath);
+                return Equals((Surrogate)obj);
```
JeffCyr: Should be `obj as Surrogate`

src/core/Akka/Actor/ActorPath.cs
```diff
@@ -298,10 +297,14 @@ private static bool TryParseAddress(string path, out Address address, out Uri ur
             //This code corresponds to AddressFromURIString.unapply
             uri = null;
             address = null;
-
-            if (!Uri.TryCreate(path, UriKind.Absolute, out uri))
+            try
```
JeffCyr: Why not TryCreate?

annymsMthd: Doesn't matter. Cluster code past 1.0.3 is broken anyways

annymsMthd: Anything that helps fix clustering should just go out.

rogeralsing: Looks like I have some rebase version issue, that older code somehow got in here

Aaronontheweb: @rogeralsing were you able to sort out the rebase issue?

Aaronontheweb:
Need to port https://github.com/akka/akka/blob/ddcfb5e4d07b07ab5221175264795235cbcd2492/akka-actor-tests/src/test/scala/akka/actor/UidClashTest.scala as part of this

rogeralsing:
@Aaronontheweb that linked test is for testing actor uniqueness (incarnation) upon restart, it is not related to actorpath UID, serialization or remoting.
The name of the test is a mystery as they don't touch path uid nor system uid.

rogeralsing:
I have added a test for the serialization layer that ensures that UID is passed along with actorpaths when using serialization.

UID for actorpaths seems to be a really last minute bolted on feature as there are no tests or verifications for this in Akka.Actor nor Akka.Remote on the JVM side..

Horusiath:
@rogeralsing what is the status of that. If it's ready, you need to rebase it.

rogeralsing:
Rebased,
All tests pass locally

rogeralsing:
¯\\\_(ツ)\_/¯

src/core/Akka.Remote.Tests/RemoteMessageLocalDeliverySpec.cs
```diff
@@ -86,7 +86,7 @@ public void RemoteActorRefProvider_should_correctly_resolve_valid_LocalActorRef_
                     var ai =
                         Sys.ActorSelection(actorPath).Ask<ActorIdentity>(new Identify(null), TimeSpan.FromSeconds(3)).Result;
 
-                   
+                    Task.Delay(TimeSpan.FromSeconds(3)).Wait(); // if we delay the initial send, this spec will fail
```
rogeralsing: @Aaronontheweb I need some help here.
Adding a delay before sending the poison pill makes the tests pass.
If the delay is not there, the remote actor will never send the ´DeathWatchNotification`
This makes me believe there is more going on here in terms of message delivery/order.

src/core/Akka.Remote.Tests/RemoteMessageLocalDeliverySpec.cs
```diff
@@ -86,7 +86,7 @@ public void RemoteActorRefProvider_should_correctly_resolve_valid_LocalActorRef_
                     var ai =
                         Sys.ActorSelection(actorPath).Ask<ActorIdentity>(new Identify(null), TimeSpan.FromSeconds(3)).Result;
 
-                   
+                    Task.Delay(TimeSpan.FromSeconds(3)).Wait();
```
rogeralsing: If this line is not here, the deathwatchnotification will for some reason not be sent from the remote actor

src/core/Akka/Actor/ActorPath.cs
```diff
@@ -254,8 +234,7 @@ public bool Equals(ActorPath other)
             var a = path;
             foreach (string element in name)
             {
-                if(!string.IsNullOrEmpty(element))
-                    a = a / element;
+                a = a / element;
```
Aaronontheweb: Removing the `if(!string.IsNullOrEmpty(element))` line here is what caused the `ClusterSingletonManagerStartupSpec` to fail - the `ActorSelection`'s used by the `ClusterSingletonProxy` can start with an empty first element due to the way their `split` function is implemented without this. Going to run a regression test on the build server real quick, but this looks like it.

Aaronontheweb:
Fixed the MNTR issues on #2073

PR #1822, opened 2016-03-24
A few more bug fixes and a lot more tests passing

cconstantin:
* StreamTestKitSpec
* ActorPublisherSpec (all but one passing)

### DIFF
--- a/src/core/Akka.Streams/Implementation/StreamLayout.cs
+++ b/src/core/Akka.Streams/Implementation/StreamLayout.cs
@@ -4,11 +4,9 @@ using System.Collections.Immutable;
 using System.Linq;
 using System.Reactive.Streams;
 using System.Runtime.Serialization;
-using System.Threading.Tasks;
 using Akka.Pattern;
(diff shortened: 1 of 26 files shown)


src/core/Akka.Streams/Implementation/StreamLayout.cs
```diff
@@ -1004,20 +1009,21 @@ private object ResolveMaterialized(StreamLayout.IMaterializedValueNode node, IDi
...
                     ResolveMaterialized(combine.Right, values, indent + "  "));
             }
             else if (node is StreamLayout.Transform)
             {
-                var transform = node as StreamLayout.Transform;
+                var transform = (StreamLayout.Transform) node;
                 result = transform.Transformator(ResolveMaterialized(transform.Node, values, indent + "  "));
             }
-            else result = null;
+            else result = Unit.Instance;
```
Horusiath: Are you sure about this one? Setting up Unit is not the same as not setting any value at all.

PR #1823, opened 2016-03-24
Updated StreamOfStreams

Silv3rcircl3:


src/core/Akka.Streams/Implementation/Fusing/StreamOfStreams.cs
```diff
@@ -1,576 +1,809 @@
...
-                    SetHandler(stage.Out,
-                        onPull: stage.PullParent,
-                        onDownstreamFinish: stage.CancelParent);
+                    _logic = logic;
                 }
 
-                public override void PreStart()
+                public bool HasInitialElement => FirstElem != null;
+
+                public T FirstElem { private get; set; }
```
Horusiath: Please, no abbreviations ;) (I know, scala has them, but we should follow C# style rules)

src/core/Akka.Streams/Implementation/Fusing/StreamOfStreams.cs
```diff
@@ -1,576 +1,809 @@
...
-                else Pull(_stage.In);
+                _subSink = subSink;
+
+                var ec = subSink._externalCallback;
+
+                SetHandler(subSink._in, onPush: () => ec(new OnNext(Grab(subSink._in))), onUpstreamFinish: () => ec(OnComplete.Instance), onUpstreamFailure: ex => ec(new OnError(ex)));
             }
 
-            private void OnUpstreamFinish()
+            private void SetCb(Action<SubSink.Command> cb)
```
Horusiath: Again, no abbreviations. I think it should be `SetCallback`, right?

src/core/Akka.Streams/Implementation/Fusing/StreamOfStreams.cs
```diff
@@ -1,576 +1,809 @@
...
         }
 
         #endregion
 
-        public readonly int Count;
-        public readonly Inlet<T> In = new Inlet<T>("PrefixAndTail.in");
-        public readonly Outlet<Tuple<IEnumerable<T>, Source<T, Unit>>> Out = new Outlet<Tuple<IEnumerable<T>, Source<T, Unit>>>("PrefixAndTail.out");
+        private readonly string _name;
+        private readonly Outlet<T> _out = new Outlet<T>("SubSource.out");
+        private readonly AtomicReference<object> _status = new AtomicReference<object>();
```
Horusiath: It looks like this container always have values of type `Action<IActorSubscriberMessage>`. Shouldn't we specify generic param to that type?

Silv3rcircl3: Status can also contain a IActorSubscriberMessage for example in this line 
```csharp
else if (!_status.CompareAndSet(null, OnComplete.Instance))
```

PR #1465, opened 2015-11-25
[WIP] DotNettyTransport

Aaronontheweb:

PR #1826, opened 2016-03-24
v1.5 - NetworkStreamTransport implementation

JeffCyr:
This pull request includes a new Remoting Transport based on a NetworkStream.

This Transport aims to get the lowest overhead possible on top of .Net Sockets.

A sub-goal is to be able to support easily any Transport based on System.IO.Stream. The PR include two other transports (SslStreamTransport and NamedPipeTransport), but it is still a work in progress.

Initial results are promising, the transport is able to achieve 120K msg/sec on a quad core workstation, compared to ~6K msg/sec with Helios.

The happy path is working, but the code remain largely untested. Other PRs will includes spec until there is 100% code & feature coverage.

(This PR does not include the ActorPath cache to achieve 120K msg/sec, without the patch it is doing ~70K msg/sec)

### DIFF
--- a/src/core/Akka.Remote/Configuration/Remote.conf
+++ b/src/core/Akka.Remote/Configuration/Remote.conf
@@ -587,6 +587,14 @@ akka {
 
       # Enables TCP Keepalive, subject to the O/S kernel's configuration
       tcp-keepalive = on
+
+      # Messages larger than this threshold will be read by chunk
+      # if the data is not already available in the buffer.
+      chunked-read-threshold = 4096b
+
+      # Messages larger than this limit will make the connection drop.
+      # Default 64MB.
+      frame-size-hard-limit = 67108864b
     }
 
     ### Default configuration for the failure injector transport adapter
--- a/src/core/Akka.Remote/Configuration/Remote.conf
+++ b/src/core/Akka.Remote/Configuration/Remote.conf
@@ -488,6 +488,107 @@ akka {
       }
     }
 
+    networkstream {
+      # The class given here must implement the akka.remote.transport.Transport
+      # interface and offer a public constructor which takes two arguments:
+      #  1) akka.actor.ExtendedActorSystem
+      #  2) com.typesafe.config.Config
+      transport-class = "Akka.Remote.Transport.Streaming.NetworkStreamTransport, Akka.Remote"
+
+      # Transport drivers can be augmented with adapters by adding their
+      # name to the applied-adapters list. The last adapter in the
+      # list is the adapter immediately above the driver, while
+      # the first one is the top of the stack below the standard
+      # Akka protocol
+      applied-adapters = []
+
+      # The default remote server port clients should connect to.
+      # Default is 2552 (AKKA), use 0 if you want a random available port
+      # This port needs to be unique for each actor system on the same machine.
+      port = 2552
+
+      # The hostname or ip clients should connect to.
+      hostname = "localhost"
+
+      # Use this setting to bind a network interface to a different port
+      # than remoting protocol expects messages at. This may be used
+      # when running akka nodes in a separated networks (under NATs or docker containers).
+      # Use 0 if you want a random available port. Examples:
+      #
+      # akka.remote.networkstream.port = 2552
+      # akka.remote.networkstream.bind-port = 2553
+      # Network interface will be bound to the 2553 port, but remoting protocol will
+      # expect messages sent to port 2552.
+      #
+      # akka.remote.networkstream.port = 0
+      # akka.remote.networkstream.bind-port = 0
+      # Network interface will be bound to a random port, and remoting protocol will
+      # expect messages sent to the bound port.
+      #
+      # akka.remote.networkstream.port = 2552
+      # akka.remote.networkstream.bind-port = 0
+      # Network interface will be bound to a random port, but remoting protocol will
+      # expect messages sent to port 2552.
+      #
+      # akka.remote.networkstream.port = 0
+      # akka.remote.networkstream.bind-port = 2553
+      # Network interface will be bound to the 2553 port, and remoting protocol will
+      # expect messages sent to the bound port.
+      #
+      # akka.remote.networkstream.port = 2552
+      # akka.remote.networkstream.bind-port = ""
+      # Network interface will be bound to the 2552 port, and remoting protocol will
+      # expect messages sent to the bound port.
+      #
+      # akka.remote.networkstream.port if empty
+      bind-port = ""
+
+      # Use this setting to bind to a specific ip
+      # than remoting protocol expects messages at.
+      # Use "0.0.0.0" to bind to all interfaces.
+      # akka.remote.networkstream.hostname if empty
+      bind-hostname = ""
+
+      # Sets the connectTimeoutMillis of all outbound connections,
+      # i.e. how long a connect may take until it is timed out
+      connection-timeout = 15s
+
+      # Sets the send buffer size of the Sockets.
+      send-buffer-size = 256000b
+
+      # Sets the receive buffer size of the Sockets.
+      receive-buffer-size = 256000b
+
+      # Sets the write buffer size of the Stream.
+      # This is the buffer that is given to Socket.BeginSend.
+      # Setting a higher value might cause send delays, only modify if you know what you are doing
+      stream-write-buffer-size = 4096b
+
+      # Sets the read buffer size of the Stream.
+      # This is the buffer that is given to Socket.BeginReceive.
+      stream-read-buffer-size = 65536b
+
+      # The time to wait for pending writes before closing the connection
+      flush-wait-on-shutdown = 2s
+
+      # Maximum message size the transport will accept, but at least
+      # 32000 bytes.
+      # Please note that UDP does not support arbitrary large datagrams,
+      # so this setting has to be chosen carefully when using UDP.
+      # Both send-buffer-size and receive-buffer-size settings has to
+      # be adjusted to be able to buffer messages of maximum size.
+      maximum-frame-size = 128000b
+
+      # Sets the size of the connection backlog
+      backlog = 4096
+
+      # Enables the TCP_NODELAY flag, i.e. disables Nagle's algorithm
+      tcp-nodelay = on
+
+      # Enables TCP Keepalive, subject to the O/S kernel's configuration
+      tcp-keepalive = on
+    }
+
     ### Default configuration for the failure injector transport adapter
 
     gremlin {
--- a/src/core/Akka.Remote/Transport/Streaming/AsyncQueue.cs
+++ b/src/core/Akka.Remote/Transport/Streaming/AsyncQueue.cs
@@ -1,74 +1,138 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Akka.Remote.Transport.Streaming
 {
-    //TODO This queue can be optimized by making it single reader
-    // and using a custom awaiter instead of TaskCompletionSource
     internal sealed class AsyncQueue<T> : IDisposable
     {
-        private class Waiter
+        internal interface IFutureItem : INotifyCompletion
         {
-            private readonly TaskCompletionSource<T> _completion;
+            T Value { get; }
+            bool IsCanceled { get; }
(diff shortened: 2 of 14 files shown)


src/core/Akka.Remote/Configuration/Remote.conf
```diff
@@ -488,6 +488,107 @@ akka {
...
+      # expect messages sent to the bound port.
+      #
+      # akka.remote.networkstream.port if empty
+      bind-port = ""
+
+      # Use this setting to bind to a specific ip
+      # than remoting protocol expects messages at.
+      # Use "0.0.0.0" to bind to all interfaces.
+      # akka.remote.networkstream.hostname if empty
+      bind-hostname = ""
```
Aaronontheweb: nitpick: we use "public-hostname" to achieve this same thing on the existing Helios transport. I don't have a strong opinion on whether or not those should be consistent.

src/core/Akka.Remote/Transport/Streaming/AsyncQueue.cs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        public IFutureItem Dequeue()
+        {
+            FutureItem futureItem = new FutureItem();
+
+            lock (_queue)
+            {
+                if (_pendingDequeue != null)
+                    throw new InvalidOperationException("Dequeue operation is already in progress. This class is single reader.");
```
Aaronontheweb: Nitpick: I would explain the intention of the design behind this queue in an XML-DOC comment, for the benefit of future contributors who might need to read and modify this code.

src/core/Akka.Remote/Transport/Streaming/NamedPipeTransport.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.IO.Pipes;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Configuration;
+
+namespace Akka.Remote.Transport.Streaming
+{
+    //TODO Work in progress
+    public class NamedPipeTranspotSettings : StreamTransportSettings
```
Aaronontheweb: Should be `NamedPipeTransportSettings`

src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
```diff
@@ -0,0 +1,289 @@
...
+                // will work for any address (ipv4 and ipv6)
+                if (IPAddress.Any.Equals(ip) && Socket.OSSupportsIPv6)
+                    ip = IPAddress.IPv6Any;
+
+                BindIp = ip;
+            }
+            else
+            {
+                // If hostname is not localhost or an ip, bind to IPAny
+                BindIp = Socket.OSSupportsIPv6 ? IPAddress.IPv6Any : IPAddress.Any;
```
Aaronontheweb: need to try DNS hostname resolution before binding to `0.0.0.0`

src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
```diff
@@ -0,0 +1,289 @@
...
+            Hostname = config.GetString("hostname");
+            Port = config.GetInt("port");
+
+            string bindHostname = config.GetString("bind-hostname");
+
+            if (string.IsNullOrEmpty(bindHostname))
+                bindHostname = Hostname;
+
+            IPAddress ip;
+            if (Hostname.Equals("localhost", StringComparison.OrdinalIgnoreCase))
```
Aaronontheweb: I'd change this to (pseudo-code):

```
if(!IPAddress.TryParse(Hostname, out ip)){
   //dns
   ip = Dns.GetHostEntry(HostName).AddressList[0]; // might want to cache somewhere in settings
} else{
   /* set the BIND IP  */
}
```

src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
Aaronontheweb: Also: if the user didn't specify a valid hostname OR ipaddress, throw an exception instead. Let them know their configuration is screwed up.

Aaronontheweb: Actually - nevermind. 0.0.0.0 is a fine default - consistent with what JVM Akka and Helios currently do.

src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
```diff
@@ -0,0 +1,289 @@
...
+            : base(config)
+        {
+            Hostname = config.GetString("hostname");
+
+            if (string.IsNullOrEmpty(Hostname))
+                Hostname = "localhost";
+
+            Port = config.GetInt("port");
+
+            Hostname = config.GetString("hostname");
```
Aaronontheweb: This will override the previous `Hostname` set above - delete.

src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
```diff
@@ -0,0 +1,289 @@
...
+            Port = config.GetInt("port");
```
Aaronontheweb: Redundant `Port` set here also.

src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
```diff
@@ -0,0 +1,289 @@
...
+
+        public IPEndPoint GetBindEndPoint()
+        {
+            return new IPEndPoint(BindIp, BindPort);
+        }
+    }
+
+    public class NetworkStreamTransport : StreamTransport
+    {
+        public const string ProtocolName = "tcp";
```
Aaronontheweb: Protocol name needs to be configurable via HOCON.

Aaronontheweb: Whoops, nevermind I stand corrected - Helios / Netty only do this because they support UDP and TCP. This is fine otherwise.

src/core/Akka.Remote/Transport/Streaming/SslStreamTransport.cs
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    public class SslStreamTransport : NetworkStreamTransport
+    {
+        protected new SslStreamTransportSettings Settings { get; }
+
+        public override string SchemeIdentifier
+        {
+            get { return "ssl"; }
```
Aaronontheweb: Should be `ssl.tcp` or whatever the underlying transport is - see equivalent JVM code: https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/transport/netty/NettyTransport.scala#L272

src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
```diff
@@ -0,0 +1,289 @@
...
+            return Task.FromResult((AssociationHandle)association);
+        }
+
+        protected async Task<AssociationHandle> CreateOutboundAssociation(Address remoteAddress)
+        {
+            Socket socket = CreateDualModeTcpSocket();
+            Settings.ConfigureSocket(socket);
+
+            //TODO Does it even work if the port is not set?
+            int port = remoteAddress.Port ?? 2552;
```
Aaronontheweb: No - blow it up if the remote port is not set.

src/core/Akka.Remote/Transport/Streaming/StreamAssociationHandle.cs
```diff
@@ -0,0 +1,238 @@
...
+                var payload = new InboundPayload(ByteString.Unsafe.FromBytes(payloadBuffer));
+                _eventListener.Notify(payload);
+            }
+        }
+
+        public override bool Write(ByteString payload)
+        {
+            _writeQueue.Enqueue(payload);
+
+            // Always return true, we don't want to use the backoff in EndpointWriter
```
Aaronontheweb: Using backoff might actually be the right thing to do in some cases, like if the writeQueue is full.

src/core/Akka.Remote/Transport/Streaming/StreamAssociationHandle.cs
```diff
@@ -0,0 +1,238 @@
...
+            _readBuffer = new byte[settings.StreamReadBufferSize];
+            _writeBuffer = new byte[settings.StreamWriteBufferSize];
+            _flushWaitOnShutdown = settings.FlushWaitTimeout;
+
+            _initialized = new TaskCompletionSource<bool>();
+            _stopped = new TaskCompletionSource<bool>();
+            _writeLoop = Task.Run(() => WriteLoop());
+        }
+
+        internal void Initialize(IHandleEventListener eventListener)
```
Aaronontheweb: So correct me if I'm wrong - but does each instance of this transport use two separate threads perpetually? One for writing and another for reading respectively?

Or do the `WriteAsync` / `ReadAsync` calls use an I/O completion port and release those threads back until data is available / flushed? I'm assuming the latter but just want to be certain.

Aaronontheweb:
@JeffCyr general comment on the stream transports: this is something that I screwed up with Helios because I didn't know any better at the time, but we should pick a default Endianness in all of the buffers flushed onto the network and not let the underlying hardware decide. That's going to be especially true when .NET Core ships and Akka.NET can start running in a lot more places.

I recommend going with BigEndian just because that's the expected host network order for most protocols: http://stackoverflow.com/a/13514942/377476

Aaronontheweb:
Ok, finished with this review - some high level things:

1. Add some XML-DOCs to all of these classes so poor @sean-gilliam doesn't have to do it later :p 
2. Add some unit tests to verify that all of the settings classes correctly parse their HOCON values from a `Config` object.
3. I thought it was really clever how you used the `IAssociationEventListener` and `IHandleEventListener` to cut out a ton of boilerplate that it's in the Helios and DotNetty transports. I found your implementation much easier to follow than two of mine.

Another thing to consider is the threading part of this - how well does this transport implementation perform across X simultaneous connections? DotNetty and Helios try to push all of the dispatching work that happens from the transport onto a fixed size number of threads so we don't saturate the global `ThreadPool` as the number of connections grows. How is this transport going to deal with that and why might it be or not be an issue?

src/core/Akka.Remote/Transport/Streaming/StreamAssociationHandle.cs
JeffCyr: These are async methods, so you can view them as logical thread, they do not block a real thread. WriteAsync/ReadAsync use the IO threadpool.

The thread is released when await is called on an uncompleted task. When the task completes, the loop continue on the thread that completed the task. Might be an IO completion thread if coming back from ReadAsync/WriteAsync or might be a threadpool thread if coming back from the AsyncQueue.

Aaronontheweb: Ok, cool - that's what I thought.

alexvaluyskiy:
@JeffCyr, have you compared performance with DotNetty transport from this PR #1465?

@Aaronontheweb, what the plans on Helios and DotNetty transports in the next release? Will DotNetty transport released?

JeffCyr:
@alexvaluyskiy I tried but there is currently a bug in DotNettyTransport or DotNetty itself that causes the throughput test to fail.

JeffCyr:
@Aaronontheweb I addressed your remarks in this [commit](https://github.com/akkadotnet/akka.net/pull/1826/commits/258c98dc97b1c3dc52a9f4bf4a2ae98a3bb03f15)

Here's my thought on some remarks:

##### `hostname` and `bind-hostname`
I followed the akka jvm settings here instead of helios. I also chose to bind to IPAny automatically unless localhost or a specific ip is specified in `hostname`. `bind-hostname` should rarely be used, `hostname` will work in most case, but it can be used to bind to a specific ip instead of IPAny.

##### Threading
The StreamTransports are designed to never block a thread. They will not cause a ThreadPool starvation, however some callbacks are called on the ThreadPool (e.g. when AsyncQueue.DequeueAsync is awaited). So the transport will be affected by a ThreadPool starvation if one occur in the current process. I will change the code to never hit the ThreadPool.

##### TODOs
- I will document the code thoroughly in another PR
- I will add 100% test coverage in a separate PR

PR #1827, opened 2016-03-25
All Implementation tests are now passing

cconstantin:


src/core/Akka.Streams.Tests/Implementation/Fusing/InterpreterStressSpec.cs
```diff
@@ -37,7 +33,8 @@ public InterpreterStressSpec(ITestOutputHelper helper = null) : base(helper)
         [Fact]
         public void Interpreter_must_work_with_a_massive_chain_of_maps()
         {
-            var ops = Enumerable.Range(1, ChainLength).Select(_=>_map).ToArray();
+            var ops = Enumerable.Range(1, ChainLength).Select(_ => _map)
+                .Cast<IStage<int, int>>().ToArray();
```
Silv3rcircl3: Is the extra cast necessary ?

Horusiath: I think, we can refactor this one during code cleanup if it's necessary. Right now I want to merge this one.

Issue #1828, opened 2016-03-25
ActorRefs.Nobody is not preserved by remoting


morganski:
If a remote actor returns an IActorRef of ActorRefs.Nobody, by the time it is received by the client end it's changed to a remote actor reference. Thus it becomes tricky to check if the remote end returned ActorRefs.Nobody as the object returned has been munged by remoting. I'd suggest that ActorRefs.Nobody is special cased and always turns up at the receiver as what was sent.

morganski:
I asked a question on SO about this recently... http://stackoverflow.com/questions/36021603/actorrefs-nobody-being-translated-by-remoting

Aaronontheweb:
Resolved by #1829

PR #1829, opened 2016-03-25
Issue #1828 Implemented NobodySurrogate

JeffCyr:
ActorRefs.Nobody is now deserialized back to its singleton instance.

This PR fix issue #1828.


Aaronontheweb:
Looks good to me - going to check the MNTR build logs and then restart.

PR #1846, opened 2016-04-04
[WIP] Akka.Persistence.Query

Horusiath:
This PR introduces `Akka.Persistence.Query` - plugin allowing building reactive streams on top of akka persistence backends. Currently it has implemented most (but not all) of the features needed by SQL-based backends, with SQLite specs as reference.


alexvaluyskiy:
These messages should be marked as `DeadLetterSuppression`
```
PersistenceIdAdded
EventAppended
CurrentPersistenceIds
TaggedEventAppended
ReplayedTaggedMessage
```
I've implemented this interface here (I don't know, why it fails on this random test)
https://github.com/akkadotnet/akka.net/pull/1847

Aaronontheweb:
@alex-kondrashov I'll get that restarted - still been working on that test since earlier in March.

Horusiath:
@alexvaluyskiy  I think, we can delay the implementation of that as a part of separate step.

cconstantin:
Looking good!

alexvaluyskiy:
@Horusiath you've implemented Delivery buffer in all three publishers. We can't use trait, like it was implemented in Akka JVM, maybe we could use composition and create helper class DeliveryBuffer<T>? something like this
```csharp
public class DeliverBuffer<T>
    {
        private readonly ActorPublisher<T> _actorPublisher;
        private ImmutableArray<T> _buffer = ImmutableArray<T>.Empty;

        public DeliverBuffer(ActorPublisher<T> actorPublisher)
        {
            _actorPublisher = actorPublisher;
        }

        public bool IsEmpty => _buffer.IsEmpty;

        public int Size => _buffer.Length;

        public void Add(T persistenceId)
        {
            _buffer = _buffer.Add(persistenceId);
        }

        public void AddRange(IEnumerable<T> persistenceId)
        {
            _buffer = _buffer.AddRange(persistenceId);
        }

        public void Deliver()
        {
            if (!_buffer.IsEmpty && _actorPublisher.TotalDemand > 0)
            {
                var totalDemand = (int)_actorPublisher.TotalDemand;
                if (_buffer.Length == 1)
                {
                    _actorPublisher.OnNext(_buffer[0]);
                    _buffer = ImmutableArray<T>.Empty; ;
                }
                else if (totalDemand <= int.MaxValue)
                {
                    for (var i = 0; i < totalDemand; i++)
                        _actorPublisher.OnNext(_buffer[i]);

                    _buffer = _buffer.RemoveRange(0, totalDemand);
                }
                else
                {
                    foreach (var element in _buffer)
                    {
                        _actorPublisher.OnNext(element);
                    }

                    _buffer = ImmutableArray<T>.Empty;
                }
            }
        }
    }
```

Horusiath:
Closing this one. I'll split changes in this PR into two separate.

PR #1851, opened 2016-04-06
Fixed FileSink. FileSinkSpecs is now passing.

Horusiath:


src/core/Akka/Util/ByteString.cs
```diff
@@ -166,6 +171,12 @@ public ByteString1(byte[] bytes)
                 get { return _bytes[checkRangeConvert(idx)]; }
             }
 
+            public override ByteBuffer AsByteBuffer()
+            {
+                var buffer = ByteBuffer.Wrap(_bytes, _startIndex, _length);
```
cconstantin: This variable can be inlined

cconstantin:
Looks good!

PR #1857, opened 2016-04-06
AtomicModule, optimized StreamLayout.Compose and minor IO spec fixes

Horusiath:
- Implemented AtomicModule and introduced it in inheriting classes.
- ByteString equality method (for spec asserts)
- Optimization step for compose working on two materialized values - now Keep.Left / Keep.Right check is made to minimize unnecessary steps.


cconstantin:
Looks good!

Issue #1861, opened 2016-04-07
Default journal field has been erased from default persistence config


Horusiath:
After one of the PRs we've missed that `akka.persistence.journal.plugin = "akka.persistence.journal.inmem"` config value has been erased, making a breaking change in the API. Before, when journal plugin used was not specified explicitly, in-memory journal was used by default. Now it will throw an exception instead.

Aaronontheweb:
@Horusiath ok, please go ahead and put this back in today and we'll do another patch immediately.

Aaronontheweb:
Btw, this is why you need to have "default configuration" specs like how Akka.Remote, Akka.Cluster, and so forth do.

Danthar:
PR was merged

PR #1864, opened 2016-04-07
Recovered default journal & snapshot store to default config

Horusiath:
See #1861

- Default config journal plugin id is restored to point at in-memory journal (was none, causing an error).
- Default config snapshot store plugin id is restored to point at local file system one (was none with warning log message, which didn't cause any explicit error, but was breaking change in API behavior).

EDIT:
- All persistence specs are now running on default config settings, making sure that default journal and snapshot store have been initialized.


Aaronontheweb:
Can we get spec coverage for this as well?

Horusiath:
Tests for default jorunal and snapshot store included.

Aaronontheweb:
Thank you sir

PR #1879, opened 2016-04-13
QueueSink implementation, Option as struct, and remove out param for materialized

cconstantin:


src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -18,57 +18,61 @@
 
 namespace Akka.Streams.Stage
 {
-    public interface IGraphStageWithMaterializedValue
+    public interface ILogicAndMaterializedValue<out TMaterialized>
```
Horusiath: Does it have to be an interface? Where there any downsides of using struct directly? Also, if we implement it as a struct and return and interface we have cost of additional boxing.

src/core/Akka.Streams.Tests/Dsl/FlowSpec.cs
```diff
@@ -675,7 +675,7 @@ protected override bool AroundReceive(Receive receive, object message)
             Func<Flow<TOut, TOut2, Unit>> createGraph = () =>
             {
                 var stage = new FaultyFlowStage<TOut, TOut2>();
-                var assembly = new GraphAssembly(new IGraphStageWithMaterializedValue[] { stage }, new[] { Attributes.None },
+                var assembly = new GraphAssembly(new IGraphStageWithMaterializedValue<Shape, object>[] { stage }, new[] { Attributes.None },
```
Horusiath: From what I can see in the usage of `IGraphStageWithMaterializedValue<Shape, object>`, the second generic parameter is always object. Feeling like something is wrong here. Either we should remove it and make object explicitly, or track this deep down to the point when object is cast to destination type.

src/core/Akka.Streams/Stage/GraphStage.cs
cconstantin: Variance can only be defined on interfaces. Should we make it a class to avoid boxing?

src/core/Akka.Streams.Tests/Dsl/FlowSpec.cs
cconstantin: Both type parameters are covariant and and always used as `<Shape, object>`. They are there purely to type-check return types of the methods in `IGraphStagedWithMaterializedValue`.

PR #1885, opened 2016-04-15
SplitWhen/After and GroupBy implementation

cconstantin:
GroupBy has some issues, will investigate tomorrow.


src/core/Akka.Streams/Dsl/Source.cs
```diff
@@ -134,6 +134,14 @@ public IRunnableGraph<TMat> To<TMat2>(IGraph<SinkShape<TOut>, TMat2> sink)
             return new Source<TOut, TMat2>(Module.TransformMaterializedValue(mapFunc));
         }
 
+        internal Source<TOut2, TMat> DeprecatedAndThen<TOut2>(StageModule op)
```
Horusiath: Do we really need to port deprecated API?

Horusiath:
Beside one question attached, the rest is good to go.

src/core/Akka.Streams/Dsl/Source.cs
cconstantin: It's still there in the latest version, and I'm not sure which direction they're taking.

cconstantin:
@Horusiath let's hold off on merging this, while investigating the remaining GroupBy issues I found that there may be an easier way to implement GroupBy. I'll post a comment when ready.

cconstantin:
GroupBy fixed and all tests pass, ready to merge.

Issue #1852, opened 2016-04-06
ByteString.IndexOf throws IndexOutOfRangeException after concating two ByteStrings


Silv3rcircl3:
After concating two ByteString's IndexOf throws the exception if the element didn't exist.

Example: 
```csharp           
var b = ByteString.Create(new byte[] {1}) + ByteString.Create(new byte[] {2});
            
b.IndexOf(3);
```

wmeints:
I tried the scenario you suggested and indeed it is out of range.
But I expected that, because you are asking for an index that is larger than the size of the array.

The following scenario works correctly as expected:

```
var b = ByteString.Create(new byte[] { 1 }) + ByteString.Create(new byte[] { 2 });
b.IndexOf(1); // The bytestring is 2 bytes, so this should be valid.
```

wmeints:
Changed things around a bit. I found a better way to fix the problem.
After looking around a bit more, I think that HasNext should return false when there are no more sub iterators left in the list.

@Silv3rcircl3  Can you check if this works better for you?

Silv3rcircl3:
Closed via #1887

PR #1887, opened 2016-04-15
Fix IndexOf() method for concatenated byte strings

wmeints:
Fixes issue #1852 where the method IndexOf(x) for a concatenated set
of byte strings would throw an index out of range exception when the
element doesn't exist in the either one of the strings.

Included two tests to confirm that existing and non-existing element situations work.

### DIFF
--- a/src/core/Akka/Util/ByteIterator.cs
+++ b/src/core/Akka/Util/ByteIterator.cs
@@ -197,7 +197,11 @@ namespace Akka.Util
 
             public override bool HasNext
             {
-                get { return Current.HasNext; }
+                get
+                {
+                    if (!_iterators.IsEmpty) return Current.HasNext;
+                    return false;
+                }
             }
 
             public override byte Head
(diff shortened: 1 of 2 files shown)


src/core/Akka/Util/ByteIterator.cs
```diff
@@ -496,7 +500,11 @@ public bool IsEmpty
 
         public T Head
         {
-            get { return _array[_offset]; }
+            get
+            {
+                if (!IsEmpty) return _array[_offset];
+                return default(T);
```
Silv3rcircl3: I think head should throw if the list is empty, I would expect the same behavior as with First. Also suggests the name offset that it could throw indexoutofrange even if the list isn't empty, offset >= count

wmeints:
Thanks for reviewing. Will revert that part and see if there's a better way to solve the problem.

Silv3rcircl3:
Looks good to me :+1:

PR #1888, opened 2016-04-15
Io issues

Silv3rcircl3:


src/core/Akka.Streams/Implementation/IO/OutputStreamSourceStage.cs
```diff
@@ -231,7 +283,14 @@ private void SendData(ByteString data)
         {
             Send(() =>
             {
-                _dataQueue.Add(data);
+                try
+                {
+                    _dataQueue.Add(data);
+                }
+                catch (Exception ex)
```
Horusiath: > Yo dawg, I've heard that you like IOExceptions, so we'll catch your IOException and wrap it with IOException.

I belive, this should be `catch(Exception ex) when !(ex is IOException)` ;)

src/core/Akka.Streams/Implementation/IO/OutputStreamSourceStage.cs
```diff
@@ -245,24 +304,35 @@ private void SendMessage(IAdapterToStageMessage msg, bool handleCancelled = true
...
-                    throw PublisherClosedException;
+                    _stageWithCallback.WakeUp(msg).Wait(_writeTimeout);
+                    if (_downstreamStatus.Value is Canceled && handleCancelled)
+                    {
+                        //Publisher considered to be terminated at earliest convenience to minimize messages sending back and forth
+                        _isPublisherAlive = false;
+                        throw PublisherClosedException;
+                    }
+                }
+                catch (IOException)
```
Horusiath: Same as above.

src/core/Akka.Streams/Implementation/IO/OutputStreamSourceStage.cs
Silv3rcircl3: I wanted to stay as close as possible to the jvm  ;-) 
https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/impl/io/OutputStreamSourceStage.scala#L147

Should I report this back?

Horusiath: I think, the canonical impl just didn't care about that. Personally I would vote for avoiding IOException inception.

Silv3rcircl3:
@Horusiath addressed your remarks

PR #1892, opened 2016-04-18
Akka.Persistence.Query core plugin

Horusiath:
This PR contains core lib for Akka.Persistence.Query - persistent backend implementation must use it in order to support akka streams of their side.

Example usage:

```csharp
var journal = system.ReadJournalFor<SomeCoolReadJournal>(queryPluginConfigPath);
var events = journal.Query(new EventsByTag("mytag", 0L));
```


cconstantin:
Looks good to go

Issue #1793, opened 2016-03-15
Add ActorMailboxSpec


cconstantin:
We should port the tests for mailboxes, such as https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/scala/akka/actor/ActorMailboxSpec.scala

wmeints:
If nobody else is working on this I'd like to take this one up.

Silv3rcircl3:
Go ahead :+1:

wmeints:
I notice there's a file MailboxesSpec in Akka.Test/Dispatcher. This test implements a few test cases for the mailboxes. Would you like to have the new tests in the same spot?

Silv3rcircl3:
@wmeints I think that are different specs, the MailboxesSpec tests a actual implementation weather the ActorMailboxSpec only tests that the actor is using the correct mailbox. 

This seems to be the counterpart of the MailboxesSpec 
https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/scala/akka/dispatch/PriorityDispatcherSpec.scala

wmeints:
:+1: Will write a the tests separate of the mailboxesspec.

wmeints:
Does Akka.NET support bounded mailboxes right now? I'm running into an issue where an actor cannot be created because I used `IBoundedMessageQueueSemantics`.

```
public class BoundedQueueReportingActor : MailboxReportingActor, IRequiresMessageQueue<IBoundedMessageQueueSemantics>
{

}
```

wmeints:
Since there are only a few mailboxes that are supported I'm going to add tests for those and leave the rest for another time. Hope that's okay :-)

Horusiath:
@wmeints no problem ;)

Silv3rcircl3:
@wmeints could you create issues for the missing mailboxes or create a list here? 
We should keep track of them.

wmeints:
Sure, I can do that. Will take a look at it tonight and create issues for the individual mailbox types.

wmeints:
Just noticed that there is a BoundedQueBasedMailbox class. The only thing missing for the bounded mailboxes is BoundedMailbox class. I think this one should, just like the UnboundedMailbox, be an alias for a sane implementation.

alexvaluyskiy:
@wmeints could we close it?

PR #1897, opened 2016-04-19
Implement spec for standard mailbox combinations in Akka.NET

wmeints:
Added some basic specs for the mailbox/actor combinations in Akka.NET.
Bounded dispatcher and mailbox configurations haven't been added yet
as these are not supported right now.

Takes care of issue #1793

### DIFF
--- a/src/core/Akka.Tests/Dispatch/ActorMailboxSpec.cs
+++ b/src/core/Akka.Tests/Dispatch/ActorMailboxSpec.cs
@@ -1,0 +1,177 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="MailboxesSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


src/core/Akka.Tests/Dispatch/ActorMailboxSpec.cs
```diff
@@ -0,0 +1,177 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="MailboxesSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Typesafe Inc. <http://www.typesafe.com>
```
Silv3rcircl3: Could you change the name from Typesafe to Lightbend?
I already created another PR which does that for all other files.

src/core/Akka.Tests/Dispatch/ActorMailboxSpec.cs
```diff
@@ -0,0 +1,177 @@
...
+
+        private static string GetConfig()
+        {
+            return @"
+
+unbounded-dispatcher {
+    mailbox-type = """ + typeof(UnboundedDequeBasedMailbox).AssemblyQualifiedName + @"""
+}
+
+task-dispatcher {
```
Silv3rcircl3: for what is the task dispatcher used?

Silv3rcircl3:
Rest looks good to me :+1:

wmeints:
Fixed the copyright lines in the file. Will update the original issue with the missing mailboxes.

PR #1900, opened 2016-04-20
Removed dependency on Newtonsoft.Json in F# serialization and persistence

object:
I made the changes a bit in the blind because there don't seem to be any tests that trigger the affected code. All F# API tests pass, and I verified that the sample program PersistenceExample.FsApi work just as it worked before. Now it works with both JSON.NET and Wire serializer, before F# persistence wrapper expected only JSON.NET. But even using PersistenceExample.FsApi I was not able to test all code branches.

If you find something that doesn't work with the proposed changes, please let me know how I can reproduce the problem and I will try to fix it.


src/core/Akka.Persistence.FSharp/FsApi.fs
```diff
@@ -149,18 +149,22 @@ type FunPersistentActor<'Command, 'Event, 'State>(aggregate: Aggregate<'Command,
...
+            | _ -> x.Unhandled msg
     override x.OnRecover (msg: obj) = 
         match msg with
         | :? 'Event as e -> state <- aggregate.apply mailbox state e
         | _ -> 
-            let serializer = UntypedActor.Context.System.Serialization.FindSerializerForType typeof<obj> :?> Akka.Serialization.NewtonSoftJsonSerializer
-            match Serialization.tryDeserializeJObject serializer.Serializer msg with
-            | Some(e) -> state <- aggregate.apply mailbox state e
-            | None -> x.Unhandled msg
+            let serializer = UntypedActor.Context.System.Serialization.FindSerializerForType typeof<obj>
```
Horusiath: Shouldn't that be `typeof<'Event>`?

src/core/Akka.Persistence.FSharp/FsApi.fs
```diff
@@ -275,12 +279,13 @@ type FunPersistentView<'Event, 'State>(perspective: Perspective<'Event, 'State>,
             state <- perspective.apply mailbox state e
             true            
         | _ -> 
-            let serializer = UntypedActor.Context.System.Serialization.FindSerializerForType typeof<obj> :?> Akka.Serialization.NewtonSoftJsonSerializer
-            match Serialization.tryDeserializeJObject serializer.Serializer msg with
-            | Some(e) -> 
+            let serializer = UntypedActor.Context.System.Serialization.FindSerializerForType typeof<obj>
```
Horusiath: Same here: Shouldn't that be `typeof<'Event>`?

src/core/Akka.Persistence.FSharp/FsApi.fs
```diff
@@ -350,19 +355,23 @@ type Deliverer<'Command, 'Event, 'State>(aggregate: DeliveryAggregate<'Command,
         match msg with
         | :? 'Command as cmd -> aggregate.exec mailbox state cmd
         | _ -> 
-            let serializer = UntypedActor.Context.System.Serialization.FindSerializerForType typeof<obj> :?> Akka.Serialization.NewtonSoftJsonSerializer
-            match Serialization.tryDeserializeJObject serializer.Serializer msg with
-            | Some(cmd) -> aggregate.exec mailbox state cmd
-            | None -> x.Unhandled msg
+            match msg with
+            | :? (byte[]) as bytes -> 
+                let serializer = UntypedActor.Context.System.Serialization.FindSerializerForType typeof<obj>
```
Horusiath: And again: Shouldn't that be `typeof<'Command>`?

object:
I thought so too, that is should use concrete types instead of obj. But since I am new to the code base I wanted to be very conservative about changes and left the original code untouched as much as I could. But now that you pointed this out - yes I agree. It should be the actual types - 'Event, 'Command etc.

object:
Corrected type cast.

Horusiath:
It looks like we have some FAKE problems. /cc @Aaronontheweb

Silv3rcircl3:
@object you need to rebase with latest dev, that should fix the build problems

object:
Will do, thanks.

object:
@Silv3rcircl3 @Horusiath I rebased my changes, should be OK now.

PR #1901, opened 2016-04-20
fixed racy specs

Silv3rcircl3:


src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -1438,16 +1438,15 @@ public Logic(FlowShape<T, T> shape, Attributes inheritedAttributes, Delay<T> sta
                 });
             }
 
-            private long NextElementWaitTime => _stage._delay.Ticks - (DateTime.UtcNow.Ticks - _buffer.Peek().Item1);
+            private long NextElementWaitTime
+                => _stage._delay.Milliseconds - (DateTime.UtcNow.Ticks - _buffer.Peek().Item1)*1000*1000;
```
cconstantin: Shouldn't it be `TotalMiliseconds`? Also, from ticks to miliseconds you probably need to divide by `TimeSpan.TicksPerSecond/1000`

Horusiath:
:+1:

PR #1909, opened 2016-04-25
Added heartbeat verbose mode

alexvaluyskiy:
```hocon
debug {
  # log heartbeat events (very verbose, useful mostly when debugging heartbeating issues)
  verbose-heartbeat-logging = off
}
```


Aaronontheweb:
Looks good but one small change: could you please add a check to [ClusterConfigSpec](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Cluster.Tests/ClusterConfigSpec.cs#L23) to ensure that this setting is off by default?

alexvaluyskiy:
Done

PR #1921, opened 2016-04-29
Added default config fallback to cluster tools

alexvaluyskiy:
Cluster Tools and Sharding doesn't have any default configuration. So, I've added it and have written unit tests for default configuration

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientSettings.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientSettings.cs
@@ -19,7 +19,12 @@ namespace Akka.Cluster.Tools.Client
     {
         public static ClusterClientSettings Create(ActorSystem system)
         {
+            system.Settings.InjectTopLevelFallback(ClusterClientReceptionist.DefaultConfig());
+
             var config = system.Settings.Config.GetConfig("akka.cluster.client");
+            if (config == null)
+                throw new ArgumentException(string.Format("Actor system [{0}] doesn't have `akka.cluster.client` config set up", system.Name));
+
             var initialContacts = config.GetStringList("initial-contacts").Select(ActorPath.Parse).ToImmutableSortedSet();
 
             return new ClusterClientSettings(initialContacts,
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj
@@ -60,6 +60,7 @@
     </Reference>
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="ClusterShardingConfigSpec.cs" />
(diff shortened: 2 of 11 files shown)


Aaronontheweb:
excellent work @alexvaluyskiy 

@Horusiath could you double check the settings to make sure they're correct? I don't know enough offhand.

alexvaluyskiy:
I've compared default configs with JVM versions. There are two differences.
Akka JVM has
```hocon
akka.cluster.sharding.coordinator-singleton = ${akka.cluster.singleton}
```
Akka.Net has
```hocon
akka.cluster.sharding.coordinator-singleton = akka.cluster.singleton
```
Also I've found, that Akka Cluster Client 2.4.2 supports reconnect-timeout
https://github.com/akka/akka/commit/19c7017482933172ab54a8fad754952dcecf614d

Should we create an issue for that?

And, Distributed Pub/Sub, Cluster Sharding, and Cluster Client use singletons. Should we test that these components inject the default configuration for singleton too?

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientSettings.cs
```diff
@@ -19,7 +19,12 @@ public sealed class ClusterClientSettings
     {
         public static ClusterClientSettings Create(ActorSystem system)
         {
+            system.Settings.InjectTopLevelFallback(ClusterClientReceptionist.DefaultConfig());
```
Horusiath: This could cause setting the same fallback config multiple times, since there are no restrictions for using `ClusterClientSettings.Create`. It would be better to nest it inside cluster client extension itself (it's always instantiated once).

Horusiath:
@alexvaluyskiy Yes. Please create an issue for that. I've addressed some issues, as config fallback injection should not be running inside settings creator method, as those methods are not supposed to cause side effects in actor system itself.

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientSettings.cs
alexvaluyskiy: Yes, I agree, but ClusterClient, ClusterSingletonManager and ClusterSingletonProxy don't have GetExtension method

alexvaluyskiy:
@Horusiath I moved initialization to extensions. But I don't know where I should initialize config fallacks in ClusterClient, ClusterSingletonManager and ClusterSingletonProxy

Horusiath:
I didn't know either, that's why it was up to user to provide those configs. Ideally we should be able to distinguish configs using structural comparison (so we don't need to add them twice), but this is not related to this PR. I think we could settle on that design.

alexvaluyskiy:
@Horusiath I've squashed my commits, but right now many MNTK tests fail

alexvaluyskiy:
@Aaronontheweb @Horusiath could you merge it?

Aaronontheweb:
@alexvaluyskiy will do, but could you rebase? MNTR will pass this time using latest in dev.

alexvaluyskiy:
@Aaronontheweb done

PR #1924, opened 2016-05-01
Merge branch 'dev' into v1.5

JeffCyr:
Synced the v1.5 branch with the latest commit from dev branch.


Aaronontheweb:
MNTR is acting up, but I suspect that might be a result of the new VM size running on fewer cores. Going to swap in the new Helios implementation to run the control channel for that soon and I think that will help.

PR #1929, opened 2016-05-04
Implemented sample to ensure, that feedback loop won't deadlock

Horusiath:
/cc @Silv3rcircl3 @cconstantin 

This example shows the potential problem of deadlocks inside graphs. Similar example recreated in JVM Akka (using both [scaladsl](http://doc.akka.io/docs/akka/2.4.4/scala/stream/stream-graphs.html) and [javadsl](http://doc.akka.io/docs/akka/2.4.4/java/stream/stream-graphs.html) - look at the bottom of the page) is working correctly.

I'd appreciate to get some help here.

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/GraphBackedFlowSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/GraphBackedFlowSpec.cs
@@ -376,5 +376,29 @@ namespace Akka.Streams.Tests.Dsl
 
             ValidateProbe(probe, 4, new[] {0, 1, 2, 3});
         }
+        
+        [Fact]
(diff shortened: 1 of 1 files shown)


Silv3rcircl3:
@Horusiath what do we do with this one? It's still targeting the old stream branch. 
I also gave it a quick shot but wasn't able to resolve the issue. 
There are also todo's left inside the `StreamLayout` that indicate that something is missing for cycles

https://github.com/akkadotnet/akka.net/blob/65342aac2cf44198e26c15f1b98e78b6379309f3/src/core/Akka.Streams/Implementation/StreamLayout.cs#L27

Horusiath:
I can send it in another PR, but this sounds like a serious issue for me, and I think we should resolve it.

PR #1930, opened 2016-05-05
Fix logging in Eventsourced.OnPersistFailure

cconstantin:


Horusiath:
We have constant MNTK failures here. /cc @Aaronontheweb

cconstantin:
@Horusiath MNTK fail because of the changes in this PR, or other reasons?

PR #1932, opened 2016-05-06
StreamTransport TestKit + misc fixes

JeffCyr:
This PR contains the TestKit for StreamTransport.

### DIFF
--- a/src/benchmark/RemoteBenchmark/Program.cs
+++ b/src/benchmark/RemoteBenchmark/Program.cs
@@ -12,7 +12,9 @@ namespace RemoteBenchmark
     {
         static void Main(string[] args)
         {
-            var actorSystemProvider = new NetworkStreamActorSystemProvider();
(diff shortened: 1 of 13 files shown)


JeffCyr:
The tests in this PR covers:
- The basic functionalities of NetworkStreamTransport
- Stress test with random message size to cover the buffering logic in StreamAssociationHandle
- Stress test with many concurrent connections testing that all messages written before a Disassociate are received by the other end

The issues fixed by this PR:
- The outgoing stream in StreamAssociationHandle was not flushed when Disassociate was called, this could cause some messages to be lost before a shutdown.
- The tcp connection did not gracefully shutdown, this was because the NetworkStream closed the socket with Shutdown(SocketShutdown.Both). Now the Socket is Shutdown with SocketShutdown.Send, then the association wait for a receive 0 to close the socket and terminate. This could also cause some messages to be lost before a shutdown.

Issue #1914, opened 2016-04-27
Akka cluster node is shutting down


Pisha91:
Cluster node very often fails to start with error "Failed to startup Cluster. You can try to increase 'akka.actor.creation-timeout'.". Our solution hosted on Azure cloud service with one web role(client role in Akka cluster)  and two worker roles (seed node and working node).

Aaronontheweb:
Looking into this now - testing a fix for it.

Aaronontheweb:
This is now resolved.

PR #1945, opened 2016-05-11
Helios 2.0 Integration - Multi-node Testkit Only

Aaronontheweb:
Integrated Helios 2.0 into the MNTR - testing that first.


Aaronontheweb:
May have found the root cause of https://github.com/akkadotnet/akka.net/issues/1914 - a race condition that occurs inside the constructor of `Cluster`. Added some lazy-loading to work around this.

Horusiath:
@Aaronontheweb could you squash your commits before merging?

Aaronontheweb:
@Horusiath yessir - been blasting the build server with lots of repeat-attempts at the MNTR just to make sure I've resolved any issues with rejoins. I might be able to strip the 10 second startup time off of each spec too.

Aaronontheweb:
@Horusiath @JeffCyr this is ready for review / merge in my opinion

src/core/Akka.Cluster.Tests.MultiNode/ClusterDeathWatchSpec.cs
```diff
@@ -117,7 +117,7 @@ public void ClusterDeathWatchSpecTests()
 
         public void An_actor_watching_a_remote_actor_in_the_cluster_must_receive_terminated_when_watched_node_becomes_down_removed()
         {
-            Within(TimeSpan.FromSeconds(20), () =>
+            Within(TimeSpan.FromSeconds(30), () =>
```
Aaronontheweb: I could change this back to 20 - changed it when I was testing some issues that turned out to be the fault of the MNTR.

src/core/Akka.Cluster.TestKit/MultiNodeLoggingConfig.cs
```diff
@@ -19,7 +19,7 @@ public static class MultiNodeLoggingConfig
 // ReSharper disable once InconsistentNaming
         private static readonly Config _loggingConfig =
             ConfigurationFactory.ParseString(@"
-                akka.loggers = []");
+                akka.loggers = [Akka.Event.DefaultLogger]");
```
Aaronontheweb: Logging is now on by default in the Akka.Cluster.TestKit

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -80,10 +80,10 @@ public Cluster(ActorSystemImpl system)
             //create supervisor for daemons under path "/system/cluster"
             _clusterDaemons = system.SystemActorOf(Props.Create(() => new ClusterDaemon(_settings)).WithDeploy(Deploy.Local), "cluster");
 
-            //TODO: Pretty sure this is bad and will at least throw aggregateexception possibly worse. 
-            _clusterCore = GetClusterCoreRef().Result;
-
             _readView = new ClusterReadView(this);
+
+            // force the underlying system to warmup
+            GetClusterCoreRef();
```
Aaronontheweb: Blocking on `GetClusterCoreRef` in the constructor caused a deadlock reported by #1914, so we do two things now:

1. Call `GetClusterCoreRef` in the constructor to warm up the clustering system
2. Lazily acquire a reference to it as needed later, without blocking in the constructor.

src/core/Akka.Remote.TestKit.Tests/ControllerSpec.cs
```diff
@@ -33,7 +34,7 @@ public ControllerSpec()
         [Fact]
         public void Controller_must_publish_its_nodes()
         {
-            var c = Sys.ActorOf(Props.Create(() => new Controller(1, Node.Loopback(0))));
+            var c = Sys.ActorOf(Props.Create(() => new Controller(1, new IPEndPoint(IPAddress.Loopback, 0))));
```
Aaronontheweb: No more `INode` in Helios. Just `EndPoint`s.

src/core/Akka.Remote.TestKit/BarrierCoordinator.cs
```diff
@@ -440,7 +440,7 @@ protected void InitFSM()
                         else if (clients.Select(x => x.FSM).SequenceEqual(new List<IActorRef>() {Sender}))
                             nextState =
                                 Stay().Replying(new ToClient<BarrierResult>(new BarrierResult(barrier.Name, true)));
-                        else if (clients.All(x => x.FSM != Sender))
+                        else if (clients.All(x => !Equals(x.FSM, Sender)))
```
Aaronontheweb: Fixed referential equality in a few places here.

src/core/Akka.Remote.TestKit/Internals/Reference.conf
```diff
@@ -34,7 +34,7 @@ akka {
       # (I&O) Used to configure the number of I/O worker threads on server sockets
       server-socket-worker-pool {
         # Min number of threads to cap factor-based number to
-        pool-size-min = 1
+        pool-size-min = 2
```
Aaronontheweb: Back to running on at least 2 threads.

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -168,7 +168,7 @@ internal ImmutableList<RoleName> Roles
     /// `AskTimeoutException: sending to terminated ref breaks promises`. Using lazy
     /// val is fine.
     /// </summary>
-    public abstract class MultiNodeSpec : TestKitBase, IMultiNodeSpecCallbacks
+    public abstract class MultiNodeSpec : TestKitBase, IMultiNodeSpecCallbacks, IDisposable
```
Aaronontheweb: Important change: making the `MultiNodeSpec` `IDisposable` will allow XUnit to call the `Dispose` method after the test is finished, and we need specific termination behavior in order to eliminate races in the MNTR.

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -431,9 +429,9 @@ public void MultiNodeSpecAfterAll()
             if (SelfIndex == 0)
             {
                 TestConductor.RemoveNode(_myself);
```
Aaronontheweb: This method is now called on shutdown via `Dispose`, and orders the shutdown of the nodes such that the controller is always last to exit.

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -642,6 +640,38 @@ protected ActorSystem StartNewSystem()
...
+            // If disposing equals false, the method has been called by the
+            // runtime from inside the finalizer and you should not reference
+            // other objects. Only unmanaged resources can be disposed.
+
+            //Make sure Dispose does not get called more than once, by checking the disposed field
+            if (!_isDisposed)
+            {
+                if (disposing)
+                {
+                    Console.WriteLine("---------------DISPOSING--------------------");
```
Aaronontheweb: Should probably change this to use the logger.

src/core/Akka.Remote.TestKit/Player.cs
```diff
@@ -544,53 +546,98 @@ class PlayerHandler : IHeliosConnectionHandler
...
-            _fsm.Tell(new ClientFSM.ConnectionFailure(ex.ToString()));
+            _fsm.Tell(new ClientFSM.ConnectionFailure(FormatConnectionFailure(context, exception)));
         }
 
         private void Reconnect()
         {
+            _log.Debug("Connecting...");
             _nextAttempt = Deadline.Now + _backoff;
-            RemoteConnection.CreateConnection(Role.Client, _server, _poolSize, this);
+            RemoteConnection.CreateConnection(Role.Client, _server, _poolSize, this).ContinueWith(tr =>
```
Aaronontheweb: Recursive async retry routine for when initial attempts to connect to the `TestConductor` fail, usually as a result of a non-controller node starting faster than the controller itself.

src/core/Akka.Tests.Performance/Util/StandardOutWriterMemoryBenchmark.cs
```diff
@@ -31,9 +30,7 @@ public void SetUp(BenchmarkContext context)
         [CounterMeasurement(ConsoleWriteThroughputCounterName)]
         [MemoryAssertion(MemoryMetric.TotalBytesAllocated, MustBe.LessThan, ByteConstants.SixtyFourKb)]
         [GcMeasurement(GcMetric.TotalCollections, GcGeneration.AllGc)]
-        [PerformanceCounterMeasurement(".NET CLR Memory", "# of Pinned Objects", InstanceName = NBenchPerformanceCounterConstants.CurrentProcessName, UnitName = "objects")]
```
Aaronontheweb: Had some issues with NBench.PerformanceCounters - disabled for now.

src/core/Akka.Remote.TestKit/Internals/Reference.conf
JeffCyr: This hocon is for Akka.Remote which doesn't use Helios 2.0 yet isn't? Increasing the pool size might not cause reordering issue?

src/core/Akka.Remote/packages.config
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Google.ProtocolBuffers" version="2.4.1.555" targetFramework="net45" />
-  <package id="Helios" version="1.4.1" targetFramework="net45" />
+  <package id="Helios" version="2.0.1" targetFramework="net45" />
```
JeffCyr: Updating the Helios version in Akka.Remote, this PR isn't only for the MNTR?

Aaronontheweb: That's correct - bringing upgrades to Akka.Remote itself soon.

All of the old Helios APIs are still present in 2.0 and that's what the rest of the system is still currently using.

PR #1950, opened 2016-05-12
Make UnhandledMessage constructor public to enable serialization

cconstantin:


Silv3rcircl3:
You need to update the api approved file.

cconstantin:
Ugh, doing too many things at the same time. Updated. Thanks @Silv3rcircl3.

Issue #1696, opened 2016-02-09
"akka.actor.serialize-messages = on" breaks clustering.


BrainSlugs83:
I put together a really vanilla console app to demonstrate this issue.  If you remove the line that says "serialize-messages = on" everything works fine.  But if you leave that line in, the ActorSystem.Create call **_never_** returns. -- Basically having that line causes the Actor System Creation to fail.

This is making it really hard to debug intermittent serialization issues.

This is my entire console app using Akka 1.0.6:
```
    public static class Program
    {
    const string strCfg = @"
        
        ActorSystemName : ""System""

        akka : 
        {
            actor : 
            {
                provider : ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""
                serialize-messages = on
            }
            remote : 
            {
                helios : 
                {
                    tcp : 
                    {
                        transport-class : ""Akka.Remote.Transport.Helios.HeliosTcpTransport, Akka.Remote""
                        transport-protocol : tcp

                        hostname : 0.0.0.0
                        public-hostname: 127.0.0.1
                        port : 1309
                    }
                }
            }
            cluster : 
            {    
                roles : [""Backend""]
                seed-nodes : [""akka.tcp://System@127.0.0.1:1309""]
            }
        }";

        public static void Main(string[] args)
        {
            var cfg = ConfigurationFactory.ParseString(strCfg);
            var sysName = cfg.GetString("ActorSystemName");

            Console.WriteLine($"CREATING ACTOR SYSTEM: {sysName}");
            using(var sys = ActorSystem.Create(sysName, cfg))
            {
                Console.WriteLine("READY");
                Console.ReadLine();
            }
        }
    }
```

BrainSlugs83:
It literally gets stuck forever, like hours go by and it never comes back.

It looks like it's getting stuck in Akka.Remote.Remoting::Start on one of the first two lines in the Try Block (from Reflector):
```
TaskCompletionSource<IList<ProtocolTransportAddressPair>> addressesPromise = new TaskCompletionSource<IList<ProtocolTransportAddressPair>>();
ActorRefImplicitSenderExtensions.Tell(this._endpointManager, new EndpointManager.Listen(addressesPromise));

```

donmikel:
#1554

BrainSlugs83:
Yeah, it's the same issue.

Aaronontheweb:
Serialization bugs - yay!

[...]

PR #1953, opened 2016-05-13
Added missing INoSerializationVerification markers in Akka.Remote

### LINKED ISSUE #1554: Using "serialize-message" option causes hang starting node
For example, you can change the configuration in WebCrawler sample, by adding:
```
akka {
  actor {
    serialize-messages = on
  }
}
```

And when we trying to start node, we hang on this point
```
[INFO][15.12.2015 9:02:45][Thread 0001][remoting] Starting remoting
```

cconstantin:
Intended to fix #1554 and #1696


Aaronontheweb:
ty sir!

PR #1955, opened 2016-05-14
enabled concurrency mode for NBench; activated Helios 1.4 benchmarks

Aaronontheweb:
This PR does a couple of things:

1. Enables concurrency mode for NBench, which allows all cores to be utilized. You will notice a significant increase in performance for any multi-threaded NBench spec and an increase in standard deviation size for non-multi-threaded specs possibly.
2. Activates the benchmark for the current Helios Akka.Remote transport. This will be used to compare it to the Helios 2.0 implementation.
3. Disables and then restores Console.Out for the duration our `StandardOutWriter` specs, so we don't fill up the build log with garbage.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -333,6 +333,7 @@ Target "NBench" <| fun _ ->
         let args = new StringBuilder()
                 |> append assembly
                 |> append (sprintf "output-directory=\"%s\"" perfOutput)
+                |> append (sprintf "concurrent=\"%b\"" true)
                 |> toText
 
         let result = ExecProcess(fun info -> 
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
All previous "feature-level" benchmarks for Akka.NET will have different results after this - we're effectively changing the baseline. But right now it's actually not possible to run some types of specs (like ones using the actual Akka.Remote transports) without this.

build.fsx
```diff
@@ -333,6 +333,7 @@ Target "NBench" <| fun _ ->
         let args = new StringBuilder()
                 |> append assembly
                 |> append (sprintf "output-directory=\"%s\"" perfOutput)
+                |> append (sprintf "concurrent=\"%b\"" true)
```
Aaronontheweb: Turns concurrency mode on

PR #1962, opened 2016-05-17
Helios 2.0 integration into Akka.Remote

Aaronontheweb:
The implementation as it stands has some known issues - I think I know what they are but I need some data from the build server to confirm.

Will squash these commits and clean things up before submitting for review.


Aaronontheweb:
Looking good! Going to being uncommenting some of the MNTR tests we turned off due to race conditions from before and see if they pass as a result of these fixes plus what we introduced in #1945

src/core/Akka.Remote.Tests.Performance/Transports/HeliosRemoteMessagingThroughputSpec.cs
```diff
@@ -28,7 +27,7 @@ public override Config CreateActorSystemConfig(string actorSystemName, string ip
                     # Used to configure the number of I/O worker threads on server sockets
       server-socket-worker-pool {
         # Min number of threads to cap factor-based number to
-        pool-size-min = 1
+        pool-size-min = 2
```
Aaronontheweb: Migrated back to Akka Netty settings values

src/core/Akka.Remote.Tests.Performance/Transports/RemoteMessagingThroughputSpecBase.cs
```diff
@@ -21,7 +21,7 @@ namespace Akka.Remote.Tests.Performance.Transports
     public abstract class RemoteMessagingThroughputSpecBase
     {
         private const string RemoteMessageCounterName = "RemoteMessageReceived";
-        private const long RemoteMessageCount = 10000;
+        private const long RemoteMessageCount = 50000;
```
Aaronontheweb: Opted to run each individual iteration of our messaging benchmarks for a greater number of messages, but run fewer iterations (3 instead of 13.)

src/core/Akka.Remote.Tests/RemoteConfigSpec.cs
```diff
@@ -94,8 +94,9 @@ public void Remoting_should_contain_correct_heliosTCP_values_in_ReferenceConf()
             Assert.True(s.TcpKeepAlive);
             Assert.True(s.TcpReuseAddr);
             Assert.True(string.IsNullOrEmpty(c.GetString("hostname")));
-            Assert.Equal(1, s.ServerSocketWorkerPoolSize);
-            Assert.Equal(1, s.ClientSocketWorkerPoolSize);
+            Assert.Equal(2, s.ServerSocketWorkerPoolSize);
```
Aaronontheweb: Validates the changes made to the socket worker pool settings in config

src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
```diff
@@ -46,7 +46,7 @@ public RemoteRouterSpec()
             akka.remote.retry-gate-closed-for = 1 s #in the event of a Sys <--> System2 whoosh (both tried to connect to each other), retry quickly
             akka.actor.provider = ""Akka.Remote.RemoteActorRefProvider, Akka.Remote""
             akka.remote.helios.tcp {
-                hostname = localhost
+                hostname = 127.0.0.1
```
Aaronontheweb: Had a bug where DNS resolution wasn't performed properly on the client; changed this spec to IPs to confirm DNS was the issue. Now we have a blend of specs that use `localhost` and `127.0.0.1`. Which I think is fine - but we should have a more explicit spec that tests for DNS resolution on the default transport. I'll add an issue for that.

Aaronontheweb: Added https://github.com/akkadotnet/akka.net/issues/1971

src/core/Akka.Remote.Tests/Transport/AkkaProtocolStressTest.cs
```diff
@@ -178,10 +178,12 @@ public AkkaProtocolStressTest() : base(AkkaProtocolStressTestConfig)
 
         #region Tests
 
-        [Fact(Skip = "Still have intermittent timeouts")]
+        [Fact()]
         public void AkkaProtocolTransport_must_guarantee_at_most_once_delivery_and_message_ordering_despite_packet_loss()
```
Aaronontheweb: This spec may be racy as reported by @JeffCyr, but I'd like to keep it on so we can capture some record of that in the logs.

src/core/Akka.Remote/Configuration/Remote.conf
```diff
@@ -345,6 +345,9 @@ akka {
       # Enables SSL support on this transport
       enable-ssl = false
 
+	  # Enables backwards compatibility with Akka.Remote clients running Helios 1.*
+	  enable-backwards-compatibility = false
```
Aaronontheweb: This setting was added as Helios 2.0 needed a special adapter to talk to Helios 1.4* - may no longer be necessary as of Helios 2.1, which this package uses. I'll have to test first though.

src/core/Akka.Remote/Transport/Helios/HeliosHelpers.cs
```diff
@@ -21,224 +22,50 @@
 
 namespace Akka.Remote.Transport.Helios
 {
-    /// <summary>
-    /// INTERNAL API
-    /// 
-    /// Abstract base class for adapting Helios <see cref="IConnection"/> objects to work with Akka.Remote.Transports
-    /// </summary>
-    internal abstract class HeliosHelpers : IConnection
+    internal abstract class CommonHandlers : ChannelHandlerAdapter
```
Aaronontheweb: Apologies for the unreadable diff here - Helios 2.0 requires much less code for its transport handlers due to the pipeline model. Basically we hand the yet-to-be-serialized message over to protobuf and decrease the reference count on the buffer.

src/core/Akka.Remote/Transport/Helios/HeliosTcpTransport.cs
```diff
@@ -7,155 +7,106 @@
...
         {
-            _associationListenerTask.ContinueWith(r =>
+            InitInbound(context.Channel, (IPEndPoint)context.Channel.RemoteAddress, null);
+            base.ChannelActive(context);
+        }
+
+        void InitInbound(IChannel channel, IPEndPoint socketAddress, object msg)
+        {
+            // disable automatic reads
+            channel.Configuration.AutoRead = false;
```
Aaronontheweb: `AutoRead` is false by default - otherwise one of the handshake messages could make it through without an actor listening. When `AutoRead` is set back to true the channel will begin automatically reading messages from the socket as they arrive using `SocketAsyncEventArgs` internally.

src/core/Akka.Remote/Transport/Helios/HeliosTcpTransport.cs
```diff
@@ -207,18 +155,17 @@ public TcpAssociationHandle(Address localAddress, Address remoteAddress, HeliosT
 
         public override bool Write(ByteString payload)
         {
-            if (_channel.IsOpen())
+            if (_channel.IsOpen && _channel.IsWritable)
```
Aaronontheweb: If the channel is not writable (too many pending writes queued up already) then this will trigger Akka.Remote's adaptive backoff mechanism to run - and that's a good thing.

src/core/Akka.Remote/Transport/Helios/HeliosTcpTransport.cs
```diff
@@ -239,12 +186,30 @@ public HeliosTcpTransport(ActorSystem system, Config config)
 
         protected override Task<AssociationHandle> AssociateInternal(Address remoteAddress)
         {
-            var client = NewClient(remoteAddress);
+            var clientBootstrap = ClientFactory(remoteAddress);
+            var socketAddress = AddressToSocketAddress(remoteAddress);
+            var associate = clientBootstrap.ConnectAsync(socketAddress).ContinueWith(tr =>
+            {
+                var channel = tr.Result;
+                var handler = (TcpClientHandler) channel.Pipeline.Last();
```
Aaronontheweb: the `IChannel.Pipeline` is safe to access in this context since it's never modified after startup.

src/core/Akka.Remote/Transport/Helios/HeliosTransport.cs
```diff
@@ -87,6 +89,9 @@ private void Init()
             ServerSocketWorkerPoolSize = ComputeWps(Config.GetConfig("server-socket-worker-pool"));
             ClientSocketWorkerPoolSize = ComputeWps(Config.GetConfig("client-socket-worker-pool"));
             Port = Config.GetInt("port");
+
+            // used to provide backwards compatibility with Helios 1.* clients
+            BackwardsCompatibilityModeEnabled = Config.GetBoolean("enable-backwards-compatibility", false);
```
Aaronontheweb: Again, may not need this with Helios 2.1 but that needs to be tested.

src/core/Akka.Remote/Transport/Helios/HeliosTransport.cs
```diff
@@ -154,13 +161,18 @@ private int ComputeWps(Config config)
     /// </summary>
     abstract class HeliosTransport : Transport
     {
+
+        private readonly IEventLoopGroup _serverEventLoopGroup;
```
Aaronontheweb: The private threadpools managed by Helios, which are cleaned up when the transport is terminated.

src/core/Akka.Remote/Transport/Helios/HeliosTransport.cs
```diff
@@ -208,150 +217,255 @@ private TransportType InternalTransport
...
+                await server;
+
+                return all.IsCompleted && server.IsCompleted;
+            }
+            finally
+            {
+                // free all of the connection objects we were holding onto
+                ConnectionGroup.Clear();
+#pragma warning disable 4014 // shutting down the worker groups can take up to 10 seconds each. Let that happen asnychronously.
+                _clientEventLoopGroup.ShutdownGracefullyAsync();
```
Aaronontheweb: We shut down the event loops separately and only after the `IChannel`s themselves are shutdown - that way we can avoid a race condition where the event loops abort the threads the channels are using to perform cleanup.

src/examples/Cluster/ClusterSharding/ClusterSharding.Node/Customers.cs
```diff
@@ -44,6 +44,7 @@ public ItemPurchased(string itemName)
 
         public Customer()
         {
+            SetReceiveTimeout(TimeSpan.FromSeconds(60));
```
Aaronontheweb: Added passivation to this sample.

src/examples/Cluster/ClusterSharding/ClusterSharding.Node/Program.cs
```diff
@@ -60,7 +60,7 @@ private static void RunExample(ActorSystem system)
 
         private static void ProduceMessages(ActorSystem system, IActorRef shardRegion)
         {
-            var customers = new[] { "Yoda", "Obi-Wan", "Darth Vader", "Princess Leia", "Luke Skywalker", "Jar Jar Binks", "R2D2", "Han Solo", "Chewbacca", "Jabba" };
+            var customers = new[] { "Yoda", "Obi-Wan", "Darth Vader", "Princess Leia", "Luke Skywalker", "R2D2", "Han Solo", "Chewbacca", "Jabba" };
```
Aaronontheweb: Removed Jar Jar Binks, a blight upon the Star Wars franchise

Aaronontheweb:
This PR is now final and ready for review. I've taken the liberty of calling out the biggest changes in this code, other than the gutting of some of the Helios 1.4.1 code.

Aaronontheweb:
Hmm, the perf benchmark found an issue. Nice - I'll look into that now.

Aaronontheweb:
Ran the benchmark many many times and haven't been able to reproduce the issue. Tracing is in-place now so when it does come up again I'll be able to capture enough data to know where to look. Just rebased everything and otherwise this PR is ready to go.

PR #1964, opened 2016-05-18
Added NonInfluenceReceiveTimeout

alexvaluyskiy:
Messages marked with NotInfluenceReceiveTimeout will not reset the timer. This can be useful when ReceiveTimeout should be fired by external inactivity but not influenced by internal activity, e.g. scheduled tick messages.


alexvaluyskiy:
@Aaronontheweb I've never seen this test was falling
```
RemotingSpec.Bug_884_Remoting_must_support_reply_to_Routee\
```

PR #1969, opened 2016-05-20
NetworkStreamTransport - Fixed error handling for outbound connections

JeffCyr:
Fixed error handling for outbound connections + proper socket cleanup in case of Socket.Accept errors.

### DIFF
--- a/src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
+++ b/src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
@@ -111,7 +111,7 @@ namespace Akka.Remote.Transport.Streaming
                 s.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, 1);
         }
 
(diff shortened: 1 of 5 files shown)


src/core/Akka.Remote/Transport/Streaming/NetworkStreamTransport.cs
```diff
@@ -293,12 +292,8 @@ protected virtual Task<AssociationHandle> CreateInboundAssociation(Stream stream
             Socket socket = CreateSocket(endpoint);
             Settings.ConfigureSocket(socket);
 
-            var connectTask = Task.Factory.FromAsync(socket.BeginConnect, socket.EndConnect, remoteAddress.Host, port, null);
-
-            var completedTask = await Task.WhenAny(connectTask, Task.Delay(Settings.ConnectionTimeout));
```
JeffCyr: awaiting Task.WhenAny did not throw an exception when connectTask failed. So we were creating the association even in case of failure and it was being disassociated shortly after.

PR #1970, opened 2016-05-23
Added Iterator.Continually support

Danthar:
I have added a custom Iterator to support the scala style Iterator.Continually concept.
This was inspired by following scala snippet

```
    val lastFlow = Flow[Double]
      .expand(Iterator.continually(_))
```

and using this PR could be translated to
```
 double current = 0;
Func<double> next = () => current++;

var lastFlow = Flow.Create<double>()
     .Expand(_ => Iterator.Continually(next));
```

We could add more overloads to simplify/reduce the sample even more. However this would mean adding explicit overloads for each type you would want to support. 

/cc @akkadotnet/core what are your thoughts on this?

### DIFF
--- a/src/core/Akka.Streams/Util/Iterator.cs
+++ b/src/core/Akka.Streams/Util/Iterator.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Akka.Streams.Util {
+    public abstract class Iterator
+    {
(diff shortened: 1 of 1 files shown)


Danthar:
An underlying subject is probably also: How far are we going to go with porting scala akka-streams samples, when running into platform differences.

Danthar:
Decided that for this one example. Its to much. Im just going to change the sample.

PR #1975, opened 2016-05-24
added cluster termination hooks and added RegisterOnMemberRemoved

### LINKED ISSUE #1946: Add Akka.Cluster termination hooks
Should cause an `ActorSystem` to `Leave` the cluster upon `Terminate` being called.

Aaronontheweb:
close #1946

Note: does not call leave on ActorSystem shutdown - merely terminates the underlying clustering system. The end user should still shut down the cluster using the following sequence of commands:

`Cluster.RegisterOnMemberRemoved(myActorSystem.Terminate())` --> `Cluster.Leave`

### DIFF
--- a/src/core/Akka.Cluster.Tests/ClusterSpec.cs
+++ b/src/core/Akka.Cluster.Tests/ClusterSpec.cs
@@ -5,9 +5,11 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System;
 using System.Collections.Immutable;
 using System.Linq;
 using Akka.Actor;
+using Akka.Configuration;
 using Akka.TestKit;
 using Akka.Util.Internal;
 using Xunit;
@@ -105,6 +107,13 @@ namespace Akka.Cluster.Tests
         {
             _cluster.Join(_selfAddress);
             LeaderActions(); // Joining -> Up
+
+            var callbackProbe = CreateTestProbe();
+            _cluster.RegisterOnMemberRemoved(() =>
+            {
+                callbackProbe.Tell("OnMemberRemoved");
+            });
+
             _cluster.Subscribe(TestActor, new []{typeof(ClusterEvent.MemberRemoved)});
             // first, is in response to the subscription
             ExpectMsg<ClusterEvent.CurrentClusterState>();
@@ -112,6 +121,44 @@ namespace Akka.Cluster.Tests
             _cluster.Shutdown();
             var memberRemoved = ExpectMsg<ClusterEvent.MemberRemoved>();
             Assert.Equal(_selfAddress, memberRemoved.Member.Address);
+
+            callbackProbe.ExpectMsg("OnMemberRemoved");
+        }
+
+        // TODO: https://github.com/akkadotnet/akka.net/issues/1983
+        [Fact(Skip = "fails for now - will need to implement https://github.com/akkadotnet/akka.net/issues/1983")]
+        public void A_cluster_must_be_allowed_to_join_and_leave_with_local_address()
+        {
+            var sys2 = ActorSystem.Create("ClusterSpec2", ConfigurationFactory.ParseString(@"akka.actor.provider = ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""
+        akka.remote.helios.tcp.port = 0"));
+
+            try
+            {
+                var ref2 = sys2.ActorOf(Props.Empty);
+                Cluster.Get(sys2).Join(ref2.Path.Address); // address doesn't contain full address information
+                Within(TimeSpan.FromSeconds(5), () =>
+                {
+                    AwaitAssert(() =>
+                    {
+                        Cluster.Get(sys2).State.Members.Count.ShouldBe(1);
+                        Cluster.Get(sys2).State.Members.First().Status.ShouldBe(MemberStatus.Up);
+                    });
+                });
+
+                Cluster.Get(sys2).Leave(ref2.Path.Address);
+
+                Within(TimeSpan.FromSeconds(5), () =>
+                {
+                    AwaitAssert(() =>
+                    {
+                        Cluster.Get(sys2).IsTerminated.ShouldBe(true);
+                    });
+                });
+            }
+            finally
+            {
+                Shutdown(sys2);
+            }
         }
     }
 }
--- a/src/core/Akka.Cluster/Cluster.cs
+++ b/src/core/Akka.Cluster/Cluster.cs
@@ -35,7 +35,7 @@ namespace Akka.Cluster
     /// the cluster address of this actor system is [[#selfAddress]]. A member also has a status;
     /// initially <see cref="Akka.Cluster.MemberStatus.Joining"/> followed by <see cref="Akka.Cluster.MemberStatus.Up"/>.
     /// </summary>
-    public class Cluster :IExtension
+    public class Cluster : IExtension
     {
         //TODO: Issue with missing overrides for Get and Lookup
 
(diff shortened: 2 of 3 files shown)


src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -569,26 +577,26 @@ public ClusterDaemon(ClusterSettings settings)
 
             Context.ActorOf(Props.Create<ClusterHeartbeatReceiver>().WithDispatcher(settings.UseDispatcher), "heartbeatReceiver");
 
+            Receive<InternalClusterAction.GetClusterCoreRef>(msg => _coreSupervisor.Forward(msg));
```
Aaronontheweb: Removed `PatternMatch` and changed this to a `ReceiveActor`. Going to be doing that periodically as I touch the built-in cluster actors.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -569,26 +577,26 @@ public ClusterDaemon(ClusterSettings settings)
...
+            Receive<InternalClusterAction.AddOnMemberUpListener>(msg =>
+                       Context.ActorOf(Props.Create(() => new OnMemberStatusChangedListener(msg.Callback, MemberStatus.Up)).WithDispatcher(_settings.UseDispatcher).WithDeploy(Deploy.Local)));
+            Receive <InternalClusterAction.AddOnMemberRemovedListener>(msg =>
```
Aaronontheweb: Handler for a new message type (`AddOnMemberRemovedListener`) - used to help perform cleanup events when the current node is removed from the `Cluster`.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1996,27 +2004,29 @@ public GossipStats IncrementOlderCount()
     /// <summary>
     /// INTERNAL API
     /// 
-    /// The supplied callback will be run once when the current cluster member is <see cref="MemberStatus.Up"/>
+    /// The supplied callback will be run once when the current cluster member has the same status.
     /// </summary>
-    class OnMemberUpListener : ReceiveActor
+    internal class OnMemberStatusChangedListener : ReceiveActor
```
Aaronontheweb: Refactored this actor to handle any arbitrary `MemberStatus` now - this actor gets re-used for both `RegisterOnMemberUp` and `RegisterOnMemberRemoved` now

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -271,7 +286,16 @@ private static IScheduler CreateScheduler(ActorSystem system)
...
+        /// <summary>
+        /// INTERNAL API.
+        /// 
+        /// Shuts down all connections to other members, the cluster daemon and the periodic gossip and cleanup tasks.
+        /// Should not called by the user. 
+        /// 
+        /// The user can issue a <see cref="Leave"/> command which will tell the node
+        /// to go through graceful handoff process `LEAVE -> EXITING ->  REMOVED -> SHUTDOWN`.
+        /// </summary>
+        internal void Shutdown()
```
Aaronontheweb: Changed the access on this method to `internal` since it's not meant to be directly callable. Technically a breaking API change, but I don't think people were calling this before.

Aaronontheweb:
Added relevant comments where necessary - this is ready for review.

Aaronontheweb:
Merge this in before #1977 (assuming either passes muster) - there will be conflicts but it'll be easier for me to solve if this one goes in first.

Horusiath:
@Aaronontheweb  ups. Didn't read your note. Can you rebase please?

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -221,13 +225,24 @@ public void Down(Address address)
...
 
         /// <summary>
+        /// The supplied callback will be run once when the current cluster member is <see cref="MemberStatus.Removed"/>.
+        /// 
+        /// Typically used in combination with <see cref="Leave"/> and <see cref="ActorSystem.Terminate"/>.
+        /// </summary>
+        /// <param name="callback">The callback that will be run whenever the current member achieves a status of <see cref="MemberStatus.Down"/></param>
+        public void RegisterOnMemberRemoved(Action callback)
+        {
+            _clusterDaemons.Tell(new InternalClusterAction.AddOnMemberRemovedListener(callback));
```
alexvaluyskiy: Why your implementation is differ from Akka Jvm code? Jvm has the code like this
```C#
public void RegisterOnMemberRemoved(Action callback)
{
  if (IsTerminated)
    callback()
  else
    _clusterDaemons.Tell(new InternalClusterAction.AddOnMemberRemovedListener(callback));
}
```

alexvaluyskiy:
**A_cluster_must_publish_member_removed_when_shutdown** test should use cluster.registerOnMemberRemoved method

Also ClusterSpec miss some tests from Akka Jvm

alexvaluyskiy:
And Akka.Cluster.Sharding.Tests.IllustrateGracefulShutdown also should use registerOnMemberRemoved

Aaronontheweb:
@alexvaluyskiy ah, thanks for catching that! I went through `ClusterSpec` and thought we had caught that already but it looks like I need to update that spec. I'll include that change when I update this PR.

src/core/Akka.Cluster/Cluster.cs
Aaronontheweb: Ah, nice catch - otherwise the hook would never fire if the cluster was already terminated.

src/core/Akka.Cluster.Tests/ClusterSpec.cs
```diff
@@ -105,13 +107,54 @@ public void A_cluster_must_publish_member_removed_when_shutdown()
...
 
             _cluster.Shutdown();
             var memberRemoved = ExpectMsg<ClusterEvent.MemberRemoved>();
             Assert.Equal(_selfAddress, memberRemoved.Member.Address);
+
+            callbackProbe.ExpectMsg("OnMemberRemoved");
+        }
+
+        [Fact]
+        public void A_cluster_must_be_allowed_to_join_and_leave_with_local_address()
```
Aaronontheweb: Added this spec also while I was at it.

src/core/Akka.Cluster.Tests/ClusterSpec.cs
```diff
@@ -105,13 +107,54 @@ public void A_cluster_must_publish_member_removed_when_shutdown()
         {
             _cluster.Join(_selfAddress);
             LeaderActions(); // Joining -> Up
+
+            var callbackProbe = CreateTestProbe();
+            _cluster.RegisterOnMemberRemoved(() => callbackProbe.Tell("OnMemberRemoved"));
```
Aaronontheweb: We now test `Cluster.RegisterOnMemberRemoved` here as well.

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -220,13 +220,27 @@ public void Down(Address address)
...
 
         /// <summary>
+        /// The supplied callback will be run once when the current cluster member is <see cref="MemberStatus.Removed"/>.
+        /// 
+        /// Typically used in combination with <see cref="Leave"/> and <see cref="ActorSystem.Terminate"/>.
+        /// </summary>
+        /// <param name="callback">The callback that will be run whenever the current member achieves a status of <see cref="MemberStatus.Down"/></param>
+        public void RegisterOnMemberRemoved(Action callback)
+        {
+            if (IsTerminated)
```
Aaronontheweb: Invoke the callback immediately if the cluster is already terminated

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -2015,34 +2018,42 @@ public GossipStats IncrementOlderCount()
...
+            {
+                if (IsTriggered(removed.Member))
                     Done();
             });
         }
 
         protected override void PreStart()
         {
-            _cluster.Subscribe(Self, new[] { typeof(ClusterEvent.MemberUp) });
+            _cluster.Subscribe(Self, new[] { typeof(ClusterEvent.MemberUp), typeof(ClusterEvent.MemberRemoved) });
```
Aaronontheweb: Needed to subscribe to and handle `ClusterEvent.MemberRemoved` as well.

Aaronontheweb:
Wow, every MNTR spec failed.... WTF happened here

Aaronontheweb:
Ah shit, it was something the mergetool automatically resolved and screwed up from my previous PR

Aaronontheweb:
Ok, should be fixed now. Re-running.

Aaronontheweb:
@alexvaluyskiy spec revealed one more area where I needed to add proper termination support - in the `PostStop` lifecycle handler of the `MemberStatusChangedListener`. That's been implemented and the `ClusterSpec` passes again locally.

PR #1977, opened 2016-05-25
adds proper lazy initialization to avoid cluster startup deadlock

### LINKED ISSUE #1959: [Racy] ClusterRouterSupervisorSpec.Cluster_aware_routers_must_use_default_supervisor_strategy_when_none_specified
Looks like the changes I made to Cluster startup have caused some issues with this spec occasionally failing using the current Helios 1.4.1 transport. Going to look into it.

Aaronontheweb:
closes #1959

### DIFF
--- a/src/core/Akka.Cluster.Tests/StartupWithOneThreadSpec.cs
+++ b/src/core/Akka.Cluster.Tests/StartupWithOneThreadSpec.cs
@@ -1,0 +1,73 @@
+﻿using System;
+using System.Threading;
+using Akka.Actor;
+using Akka.Actor.Dsl;
+using Akka.Configuration;
+using Akka.Event;
+using Akka.TestKit;
+using Akka.Util;
+using Xunit;
+
+namespace Akka.Cluster.Tests
+{
+    public class StartupWithOneThreadSpec : AkkaSpec
+    {
+        private static readonly Config Configuration = ConfigurationFactory.ParseString(@"
+            akka.actor.creation-timeout = 10s
+            akka.actor.default-dispatcher.Type = ForkJoinDispatcher
+            akka.actor.default-dispatcher.dedicated-thread-pool.thread-count = 1
+            akka.actor.provider = ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""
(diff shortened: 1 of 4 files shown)


src/core/Akka.Cluster.Tests/StartupWithOneThreadSpec.cs
```diff
@@ -0,0 +1,73 @@
...
+using Akka.Actor.Dsl;
+using Akka.Configuration;
+using Akka.Event;
+using Akka.TestKit;
+using Akka.Util;
+using Xunit;
+
+namespace Akka.Cluster.Tests
+{
+    public class StartupWithOneThreadSpec : AkkaSpec
```
Aaronontheweb: Spec designed to reproduce deadlock failures using a very small dedicated thread pool.

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -82,8 +82,8 @@ public Cluster(ActorSystemImpl system)
 
             _readView = new ClusterReadView(this);
 
-            // force the underlying system to warmup
-            GetClusterCoreRef();
+            // force the underlying system to start
+            _clusterCore = GetClusterCoreRef().Result;
```
Aaronontheweb: Going back to blocking in the constructor here, which is the correct behavior.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -550,32 +551,31 @@ public ClusterEvent.IClusterDomainEvent Event
...
             _settings = settings;
         }
 
         protected override void OnReceive(object message)
         {
             message.Match()
-                .With<InternalClusterAction.GetClusterCoreRef>(msg => _coreSupervisor.Forward(msg))
+                .With<InternalClusterAction.GetClusterCoreRef>(msg =>
+                {
+                    if(_coreSupervisor == null)
```
Aaronontheweb: Moved the lazy initialization to here instead - avoid calling the `Cluster` object in the constructors of any actors who are needed for the cluster to startup.

PR #1986, opened 2016-05-27
upgraded to Helios 2.1.1

Aaronontheweb:


Aaronontheweb:
Fixes the binary compatibility issue reported by @JeffCyr

Issue #1575, opened 2015-12-18
[Akka.Cluster.Tools] ClusterSingletonManagerChaosSpec MNTK spec


Horusiath:
This is up for grabs task for making [ClusterSingletonManagerChaosSpec](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonManagerChaosSpec.cs) work on MultiNode TestKit Runner.

PR #2000, opened 2016-05-30
fixed ClusterSingletonManagerChaosSpec

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/issues/1575

I could not run these two specs inside one MultiNodeSpec class. Remote TestKit has some problems with the state of cluster between different tests


Aaronontheweb:
> Remote TestKit has some problems with the state of cluster between different tests

This is an XUnit design. Each `Fact` method has to be run in isolation. Can't build up a spec incrementally across multiple facts like how the JVM can. A could achieve that by switching to NUnit, or glob all "facts" together into one `MultiNodeFact` method.

You're always going to have a clean slate between each `Fact`

Issue #2001, opened 2016-05-31
Port Akka.Cluster MultiNodeSpec: MinMembersBeforeUpSpec


alexvaluyskiy:
https://github.com/akka/akka/blob/master/akka-cluster/src/multi-jvm/scala/akka/cluster/MinMembersBeforeUpSpec.scala

PR #2002, opened 2016-05-31
Implemented MinMembersBeforeUpSpec

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/issues/2001


alexvaluyskiy:
I saw this failed spec yesterday in another PR. Should we create an issue for a racy spec?

Aaronontheweb:
`AkkaProtocolStressTest.AkkaProtocolTransport_must_guarantee_at_most_once_delivery_and_message_ordering_despite_packet_loss` - @JeffCyr and I believe the issue is with how the `ThrottleTransportAdapter` and the `FailureInjectorTransportAdapter` are implemented. 

As for the remote router specs - there's definitely something iffy in our router implementations. The round robin one, currently disabled, is particularly bad. I'm going to be taking on routing issues once I'm finished with deathwatch, which I expect to have done this morning.

Aaronontheweb:
crap - looks like this needs to be rebased

PR #2025, opened 2016-05-31
cleaned up SendSystemMessage usages in Deathwatch

Aaronontheweb:
fixed up PromiseActorRef and am now using it in `GracefulStop`.

WIP because I'm not sure if  `AttemptSysMsgDeliverySpec` passes just yet. Will be working on that next.

PR #2030, opened 2016-06-02
added spec to verify that AddressUidExtension is threadsafe

Aaronontheweb:


Aaronontheweb:
Simple spec in this case.

Aaronontheweb:
Cluster Deathwatch MNTR spec failed. This is addressed and fixed in #2025

Issue #1992, opened 2016-05-30
Port Akka.Cluster MultiNodeSpec: ClientDowningNodeThatIsUnreachableSpec


alexvaluyskiy:
https://github.com/akka/akka/blob/master/akka-cluster/src/multi-jvm/scala/akka/cluster/ClientDowningNodeThatIsUnreachableSpec.scala

alexpantyukhin:
I don't know why but my PR didn't attached automatically. Please find it there https://github.com/akkadotnet/akka.net/pull/2033

alexpantyukhin:
please find my PR there https://github.com/akkadotnet/akka.net/pull/2034

maxim-s:
Close via #2037

PR #2034, opened 2016-06-03
Add ClientDowningNodeThatIsUpSpec #1992

alexpantyukhin:


Aaronontheweb:
great work @alexpantyukhin - compared it to JVM and this is spot on.

PR #2048, opened 2016-06-06
Specs for checking register same journal/snapshot store under different config paths

Horusiath:
This PR introduces 2 persistence tests which checks if the same `Journal` / `SnapshotStore` class can be registered many times using different hocon paths, and if the config passed to each instance matches registered path.

### DIFF
--- a/src/core/Akka.Persistence.Tests/PersistenceConfigSpec.cs
+++ b/src/core/Akka.Persistence.Tests/PersistenceConfigSpec.cs
@@ -4,13 +4,102 @@
 //     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
 // </copyright>
 //-----------------------------------------------------------------------
+
+using Akka.Actor;
+using Akka.Configuration;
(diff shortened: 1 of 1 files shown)


src/core/Akka.Persistence.Tests/PersistenceConfigSpec.cs
```diff
@@ -4,13 +4,102 @@
...
+                    test-value = ""D""
+                }
+            }";
+
+        public PersistenceConfigSpec(ITestOutputHelper output = null) : base(SpecConfig, output)
+        {
+        }
+
         [Fact]
         public void Persistence_should_use_inmem_journal_by_default()
```
Aaronontheweb: Shouldn't this spec be renamed "Persistence_should_use_**filesys_snapshotstore**_by_default" ?

Aaronontheweb:
One minor remark, but otherwise these tests look correct.

Aaronontheweb:
Nevermind - my remark was caused by not fully expanding the diff under the spec.

PR #2051, opened 2016-06-06
Added support for akka.actor.deployment with actorPaths/wildcards w/ dots

### LINKED ISSUE #2050: Allow dots in configuration akka.actor.deployment-wildcard
Currently, it's not possible to use configuration to change the deployment parameters for an actor if you need the path-with-wildcards for that actor to include a dot in any of the actor names in it. According to spec, it should be allowed as long as the key (path, in this case) is between quotes. 
If one tries this, the config is parsed correctly but creating an actorsystem with it throws an exception.

`akka.actor.deployment {
  "/master/a.child/*" { dispatcher = some-dispatcher }
}`

I believe this is a problem in the Config implementation (not understanding quoted path sections for queries, but only for configuration files), combined with a bug in Deployer that uses Keys as parameter to a function that expects a Path, which is not valid if the key contains a dot.

boekabart:
Fixes #2050


Aaronontheweb:
Need to http://getakka.net/docs/akka-developers/public-api-changes

boekabart:
Added the new extension methods to the api spec file

boekabart:
According to @Silv3rcircl3 , the failed perf test is unrelated to this PR. How can we 'rerun' ?

PR #2054, opened 2016-06-07
fixed ClusterAccrualFailureDetector MNTR spec

Aaronontheweb:
Needed to combine all of the three spec methods into one `MultiNodeFact`

### DIFF
--- a/src/core/Akka.Cluster.Tests.MultiNode/ClusterAccrualFailureDetectorSpec.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/ClusterAccrualFailureDetectorSpec.cs
@@ -30,6 +30,8 @@ namespace Akka.Cluster.Tests.MultiNode
             CommonConfig= DebugConfig(false)
                             .WithFallback(ConfigurationFactory.ParseString("akka.cluster.failure-detector.threshold = 4"))
                             .WithFallback(MultiNodeClusterSpec.ClusterConfig());
+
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
`Akka.Cluster.Tests.MultiNode.ClusterAccrualFailureDetectorMultiNode1.ClusterAccrualFailureDetectorSpecs` failed - it has never passed. Passed because of the wonky way it was implemented and totaled before. Going to need to look at it in more detail....

src/core/Akka.Cluster.Tests.MultiNode/ClusterAccrualFailureDetectorSpec.cs
```diff
@@ -30,6 +30,8 @@ public ClusterAccrualFailureDetectorMultiSpec()
             CommonConfig= DebugConfig(false)
                             .WithFallback(ConfigurationFactory.ParseString("akka.cluster.failure-detector.threshold = 4"))
                             .WithFallback(MultiNodeClusterSpec.ClusterConfig());
+
+            TestTransport = true;
```
Aaronontheweb: Just needed to turn the `TestTransport` on in order to get the specs to pass; otherwise Throttling / Blackholing doesn't work

Aaronontheweb:
@rogeralsing they're meant to be executed in sequence. XUnit recreates the test class every time a `Fact`, so the test isn't valid as three separate facts.

Issue #1576, opened 2015-12-18
[Akka.Cluster.Tools] DistributedPubSubMediatorSpec MNTK spec


Horusiath:
This is up for grabs task for making [DistributedPubSubMediatorSpec](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs) work on MultiNode TestKit Runner.

Silv3rcircl3:
Closed via #2057

Issue #2056, opened 2016-06-07
DistributedPubSub: put messages without route into the dead letters


alexvaluyskiy:
Implement this PR from Akka 2.4.5
https://github.com/akka/akka/issues/19009

Issue #2055, opened 2016-06-07
DistributedPubSub: Race-condition issue when handling Topic/Group actors.


alexvaluyskiy:
Implement this PR from Akka 2.4.2
https://github.com/akka/akka/issues/19017

PR #2032, opened 2016-06-02
Updated ClusterTools.Client to 2.4.8

alexvaluyskiy:
- Code and specs was aligned with Cluster Client 2.4.8 (Jvm)
- Implemented https://github.com/akkadotnet/akka.net/issues/1980 and https://github.com/akkadotnet/akka.net/issues/1981

PR #2057, opened 2016-06-07
Enabled all DistributedPubSub specs

alexvaluyskiy:
Ported DistributedPubSub code from here https://github.com/akkadotnet/akka.net/pull/2032
This version was aligned with Akka 2.4.1
Also, I've created two stories related with DistributedPubSub
- DistributedPubSub: race condition regarding children/topic actor liveness (https://github.com/akkadotnet/akka.net/issues/2055)
- DistributedPubSub: put messages without route into the dead letters (https://github.com/akkadotnet/akka.net/issues/2056)

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
@@ -19,6 +19,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
         /// </summary>
         public static DistributedPubSubSettings Create(ActorSystem system)
         {
+            system.Settings.InjectTopLevelFallback(DistributedPubSub.DefaultConfig());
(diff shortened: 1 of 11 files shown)


src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
```diff
@@ -19,6 +19,7 @@ public class DistributedPubSubSettings
         /// </summary>
         public static DistributedPubSubSettings Create(ActorSystem system)
         {
+            system.Settings.InjectTopLevelFallback(DistributedPubSub.DefaultConfig());
```
alexvaluyskiy: I'm not sure 100%, but this method doesn't update serializers and serializer's maps from reference.conf

Horusiath:
@alexvaluyskiy I've merged #1918 . Please decide if we still need that PR and rebase it if so.

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/issues/1576

Issue #2053, opened 2016-06-07
Debug on stdout not working


qwoz:
When I use the sample config at http://getakka.net/docs/Logging I can receive `_log.Info()` calls on stdout but `_log.Debug()` calls don't output. Is there something else needed to enable debug level logging?
The config on start output:
```
  akka : {
    log-config-on-start : on
    stdout-loglevel : DEBUG
    loglevel : DEBUG
    actor : {
      debug : {
        receive : on
        autoreceive : on
        lifecycle : on
        event-stream : on
        unhandled : on
      }
    }
  }
```
The first few output lines show:
```
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] subscribing [akka://all-systems/] to channel Akka.Event.Info
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] subscribing [akka://all-systems/] to channel Akka.Event.Warning
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] subscribing [akka://all-systems/] to channel Akka.Event.Error
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] StandardOutLogger started
```
Note there's no Akka.Event.Debugthere. Later it shows:
```
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] subscribing [akka://MySystem/system/log1-DefaultLogger] to channel Akka.Event.Debug
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] subscribing [akka://MySystem/system/log1-DefaultLogger] to channel Akka.Event.Info
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] subscribing [akka://MySystem/system/log1-DefaultLogger] to channel Akka.Event.Warning
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] subscribing [akka://MySystem/system/log1-DefaultLogger] to channel Akka.Event.Info
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] subscribing [akka://MySystem/system/log1-DefaultLogger] to channel Akka.Event.Warning
[DEBUG][2016-06-07 3:44:21 AM][Thread 0016][EventStream] subscribing [akka://MySystem/system/log1-DefaultLogger] to channel Akka.Event.Error
```

Danthar:
Problem is in the EventStream subscribe method

[...]

PR #2066, opened 2016-06-07
Fix for #2053

Danthar:
Moved the subscribe debug-log message in such a way that the message is
published after the subscribe completes.


rogeralsing:
Looks correct to me, merging

PR #2073, opened 2016-06-07
Actorpath uid fix

Aaronontheweb:


Aaronontheweb:
Replaces https://github.com/akkadotnet/akka.net/pull/1817

Aaronontheweb:
cc @rogeralsing preserved your original commit in this PR. Just applied my patch on top of it.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
```diff
@@ -36,7 +36,7 @@ namespace Akka.Cluster.Tools.Singleton
     /// Note that this is a best effort implementation: messages can always be lost due to the distributed nature of the actors involved.
     /// </remarks>
     /// </summary>
-    public sealed class ClusterSingletonProxy : UntypedActor
+    public sealed class ClusterSingletonProxy : ReceiveActor
```
Aaronontheweb: Converted to `ReceiveActor` for performance reasons, but mostly to make it easier for me to debug the failing MNTR spec (`PatternMatch` obfuscated the stack trace.)

alexvaluyskiy: Why you change ClusterSingletonProxy inside "Actorpath uid fix"? Is it related changes?

alexvaluyskiy: Do you have performance comparison between ReceiveActor and Match.With? if you have, I think you should write " Contribution Guideline" and say, that we should use ReceiveActor, instead untyped actor and Match.With.

Also I'm not sure about Actor DSL, it looks like very experimental, but I saw it in some specs.

alexvaluyskiy: @Aaronontheweb you've already answered :)
https://github.com/akkadotnet/akka.net/pull/311#issuecomment-52966650

Aaronontheweb:
@alexvaluyskiy I don't have a PR comparing `PatternMatch` to `Receive`, but we do have benchmarks for `UntypedActor` (without `PatternMatch`) vs. `ReceiveActor` and the numbers are about equal.

`PatternMatch` allocates objects each time it's used, whereas `ReceiveActor` does not. That's the performance incentive for one over the other.

alexvaluyskiy:
@Aaronontheweb so, maybe it would be better, to change it when we have performance tests for ClusterSingleton?

Aaronontheweb:
@alexvaluyskiy in the grand scheme of things I don't think it'd make much of a difference either way performance wise for a clustered actor - bottleneck will usually be Akka.Remote for that. But I don't need a benchmark to tell you that this implementation allocates fewer objects per message than the previous - that's just the nature of the `PatternMatch`.

Issue #2028, opened 2016-06-01
Bug: need to properly dispose of all dispatchers on shutdown


Aaronontheweb:
Especially `PinnedDispatcher` and `ForkJoinDispatcher` instances. Right now these are left running even after the `ActorSystem` terminates.

This should be a simple fix - just need to add a termination hook for `Dispatchers`.

alexvaluyskiy:
@Aaronontheweb could we close it?

Aaronontheweb:
Yep

PR #2086, opened 2016-06-12
Dispatcher, MailboxType, Mailbox, and ActorCell updates: guaranteed actor termination, dispatcher cleanup, etc

### LINKED ISSUE #2024: ActorCell updates, rewrites to support proper child termination / recreation
I spent the past day or so debugging issues with Cluster Deathwatch. Turns out this was the culprit.

The entire way we handle child create / terminate is out of date or incorrectly implemented. See: https://github.com/akka/akka/blob/65a9e0a0f9372d2e2ecb945c8a9dae129a438f77/akka-actor/src/main/scala/akka/actor/ActorCell.scala#L421

TL;DR; - we need to be able to guarantee the following:

1. That all children are shutdown before a parent calls `FinishTerminate`.
2. That no children can be allocated while the parent is shutting down.
3. That system messages that don't have anything to do with the shutting down of children during termination are stashed.

The JVM has all of this implemented, but this will be a multi-day project for someone. The good news? The `ActorSystemSpec.Reliable_deny_create_on_shutdown` spec and scores of other unit tests will EXPLODE if this is implemented incorrectly. The bad news? This will require some yak-shaving.

Aaronontheweb:
close #2024 - guaranteed that all children are terminated before parent terminates
added mailboxtype; adjusted dispatcher system to match JVM
modified SystemMessages to support SystemMessageList
implemented IRequiresMessageQueue and IProducesMessageQueue


Aaronontheweb:
Also resolves https://github.com/akkadotnet/akka.net/issues/2028

src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
```diff
@@ -336,11 +336,11 @@ public void CanSerializeActorPathContainer()
         [Fact]
         public void CanSerializeSingletonMessages()
         {
-            var message = Terminate.Instance;
+            var message = PoisonPill.Instance;
```
Aaronontheweb: `Terminate` is no longer a singleton message type, therefore not valid for this test.

src/contrib/dependencyInjection/Akka.DI.TestKit/DiResolverSpec.cs
```diff
@@ -335,7 +336,7 @@ public void DependencyResolver_should_inject_into_normal_mailbox_Actor()
 
             var internalRef = (LocalActorRef)stashActor;
 
-            Assert.IsType<UnboundedMailbox>(internalRef.Cell.Mailbox);
+            Assert.IsType<UnboundedMessageQueue>(internalRef.Cell.Mailbox.MessageQueue);
```
Aaronontheweb: There is one and only one `Mailbox` implementation, but it can use different message queues. Test has been adjusted to reflect that.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -87,7 +87,7 @@ namespace Akka.Actor
         public static Akka.Actor.IActorRef GetCurrentSenderOrNoSender() { }
         [System.ObsoleteAttribute("Use TryGetSingleChild")]
         public Akka.Actor.IInternalActorRef GetSingleChild(string name) { }
-        public void Init(bool sendSupervise, System.Func<Akka.Dispatch.Mailbox> createMailbox) { }
+        public void Init(bool sendSupervise, Akka.Dispatch.MailboxType mailboxType) { }
```
Aaronontheweb: Using proper `MailboxType` on `ActorCell` now.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -107,11 +107,10 @@ namespace Akka.Actor
         protected void SetTerminated() { }
         public static Akka.Actor.NameAndUid SplitNameAndUid(string name) { }
         public virtual void Start() { }
+        protected void Stash(Akka.Dispatch.SysMsg.SystemMessage msg) { }
```
Aaronontheweb: Internal stash for `SystemMessage`s that can't be processed during different types of suspensions

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -107,11 +107,10 @@ namespace Akka.Actor
...
         public void Stop() { }
         public void Stop(Akka.Actor.IActorRef child) { }
         public void Suspend() { }
-        public void SwapMailbox(Akka.Actor.DeadLetterMailbox mailbox) { }
```
Aaronontheweb: Removed this method and made it `internal`.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -485,15 +484,9 @@ namespace Akka.Actor
         protected override bool SpecialHandle(object message, Akka.Actor.IActorRef sender) { }
         protected override void TellInternal(object message, Akka.Actor.IActorRef sender) { }
     }
-    public class DeadLetterMailbox : Akka.Dispatch.Mailbox
+    public sealed class DeadLetterMailbox : Akka.Dispatch.Mailbox
```
Aaronontheweb: marking class as `sealed`. Modifying this class could break DeathWatch and other capabilities.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1124,14 +1117,13 @@ namespace Akka.Actor
     }
     public class LocalActorRef : Akka.Actor.ActorRefWithCell, Akka.Actor.IActorRefScope, Akka.Actor.ILocalRef
     {
-        public LocalActorRef(Akka.Actor.Internal.ActorSystemImpl system, Akka.Actor.Props props, Akka.Dispatch.MessageDispatcher dispatcher, System.Func<Akka.Dispatch.Mailbox> createMailbox, Akka.Actor.IInternalActorRef supervisor, Akka.Actor.ActorPath path) { }
-        protected internal LocalActorRef(Akka.Actor.ActorSystem system, Akka.Actor.Props props, Akka.Dispatch.MessageDispatcher dispatcher, System.Func<Akka.Dispatch.Mailbox> createMailbox, Akka.Actor.IInternalActorRef supervisor, Akka.Actor.ActorPath path, System.Func<Akka.Actor.LocalActorRef, Akka.Actor.ActorCell> createActorCell) { }
+        public LocalActorRef(Akka.Actor.Internal.ActorSystemImpl system, Akka.Actor.Props props, Akka.Dispatch.MessageDispatcher dispatcher, Akka.Dispatch.MailboxType mailboxType, Akka.Actor.IInternalActorRef supervisor, Akka.Actor.ActorPath path) { }
```
Aaronontheweb: There is one and only one constructor for `LocalActorRef`. The `protected internal` one has been removed.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1124,14 +1117,13 @@ namespace Akka.Actor
...
         public Akka.Actor.ActorCell Cell { get; }
         public override System.Collections.Generic.IEnumerable<Akka.Actor.IActorRef> Children { get; }
-        protected System.Func<Akka.Dispatch.Mailbox> CreateMailbox { get; }
         protected Akka.Dispatch.MessageDispatcher Dispatcher { get; }
         public override bool IsLocal { get; }
         public override bool IsTerminated { get; }
+        protected Akka.Dispatch.MailboxType MailboxType { get; }
```
Aaronontheweb: Using the new `MailboxType` internally

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1141,6 +1133,7 @@ namespace Akka.Actor
         public override Akka.Actor.ICell Underlying { get; }
         public override Akka.Actor.IActorRef GetChild(System.Collections.Generic.IEnumerable<string> name) { }
         public override Akka.Actor.IInternalActorRef GetSingleChild(string name) { }
+        protected virtual Akka.Actor.ActorCell NewActorCell(Akka.Actor.Internal.ActorSystemImpl system, Akka.Actor.IInternalActorRef self, Akka.Actor.Props props, Akka.Dispatch.MessageDispatcher dispatcher, Akka.Actor.IInternalActorRef supervisor) { }
```
Aaronontheweb: overrideable method for producing an `ActorCell` - this replaces the delegates we were using in the now-removed `protected internal` LocalActorRef constructor.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1381,7 +1374,7 @@ namespace Akka.Actor
     }
     public class RepointableActorRef : Akka.Actor.ActorRefWithCell, Akka.Actor.IActorRefScope, Akka.Actor.IRepointableRef
     {
-        public RepointableActorRef(Akka.Actor.Internal.ActorSystemImpl system, Akka.Actor.Props props, Akka.Dispatch.MessageDispatcher dispatcher, System.Func<Akka.Dispatch.Mailbox> createMailbox, Akka.Actor.IInternalActorRef supervisor, Akka.Actor.ActorPath path) { }
+        public RepointableActorRef(Akka.Actor.Internal.ActorSystemImpl system, Akka.Actor.Props props, Akka.Dispatch.MessageDispatcher dispatcher, Akka.Dispatch.MailboxType mailboxType, Akka.Actor.IInternalActorRef supervisor, Akka.Actor.ActorPath path) { }
```
Aaronontheweb: Same change: no more mailbox producer functions in any `IActorRef` constructors. This is all done via `MailboxType` now.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2252,6 +2245,16 @@ namespace Akka.Configuration.Hocon
 }
 namespace Akka.Dispatch
 {
+    public sealed class ActionRunnable : Akka.Dispatch.IRunnable
```
Aaronontheweb: The `IRunnable` classes are used by the `MessageDispatcher` to run tasks - they may not stick around depending on how performance review goes. The idea is to avoid constant delegate allocations by using classes that can encapsulate all of the state needed to execute an asynchronous action internally, and then use a static delegate to execute those. This is similar to how the task queues work in DotNetty and other parts of the TPL.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2262,29 +2265,26 @@ namespace Akka.Dispatch
         public static void RunTask(System.Func<System.Threading.Tasks.Task> asyncAction) { }
         protected override bool TryExecuteTaskInline(System.Threading.Tasks.Task task, bool taskWasPreviouslyQueued) { }
     }
-    public class BoundedDequeBasedMailbox : Akka.Dispatch.Mailbox<Akka.Dispatch.MessageQueues.BoundedMessageQueue, Akka.Dispatch.MessageQueues.BoundedDequeMessageQueue>, Akka.Dispatch.IDequeBasedMailbox
+    public sealed class BoundedDequeBasedMailbox : Akka.Dispatch.MailboxType, Akka.Dispatch.IProducesMessageQueue<Akka.Dispatch.MessageQueues.BoundedDequeMessageQueue>
```
Aaronontheweb: All of the previous built-in `Mailbox` classes have been preserved and simply had their implemented types switched to `MailboxType` - which gives us backwards compatibility inside the reference configuration.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2262,29 +2265,26 @@ namespace Akka.Dispatch
...
-        protected override void Schedule() { }
-        protected bool TryDequeue(out Akka.Actor.Envelope envelope) { }
+        public BoundedMailbox(Akka.Actor.Settings settings, Akka.Configuration.Config config) { }
+        public int Capacity { get; }
+        public System.TimeSpan PushTimeout { get; }
+        public override Akka.Dispatch.MessageQueues.IMessageQueue Create(Akka.Actor.IActorRef owner, Akka.Actor.ActorSystem system) { }
     }
     public class CurrentSynchronizationContextDispatcher : Akka.Dispatch.MessageDispatcher
     {
+        protected static readonly System.Action<object> Executor;
```
Aaronontheweb: Static delegate used for executing `IRunnable`s on this dispatcher

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2351,86 +2343,72 @@ namespace Akka.Dispatch
         Akka.Actor.Settings Settings { get; }
     }
     public interface IMultipleConsumerSemantics : Akka.Dispatch.ISemantics { }
+    public interface IProducesMessageQueue<TQueue>
```
Aaronontheweb: New `IProducesMessageQueue<TQueue>` interface - acts as the compliment for the existing `IRequiresMessageQueue<TQueue>` type implemented by @cconstantin. This type checking is now actually implemented inside the `Mailboxes` class.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2351,86 +2343,72 @@ namespace Akka.Dispatch
...
-        protected Akka.Dispatch.MessageDispatcher dispatcher;
-        protected System.Boolean modreq(System.Runtime.CompilerServices.IsVolatile) hasUnscheduledMessages;
-        protected int status;
-        protected Mailbox() { }
-        protected Akka.Actor.ActorCell ActorCell { get; }
-        public abstract bool IsClosed { get; }
-        public bool IsSuspended { get; }
-        public abstract void BecomeClosed();
-        public abstract void CleanUp();
+    public class Mailbox
```
Aaronontheweb: The new `Mailbox` type - **breaking changes here**.
 
This type is now firmly aligned with the JVM's design, and in implementing this I remove methods that were not relevant to that design. Many properties that were on the previous implementation are now callable methods instead, so we could take advantage of `AggressiveInling` for key state checking operations.

All mailbox instances are more or less the same, the major difference being the `IMessageQueue` instance produced by the `MailboxType` - which will be injected into this class at the time of construction.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2351,86 +2343,72 @@ namespace Akka.Dispatch
...
     {
         public static readonly string DefaultMailboxId;
         public static readonly string NoMailboxRequirement;
         public Mailboxes(Akka.Actor.ActorSystem system) { }
         public Akka.Actor.DeadLetterMailbox DeadLetterMailbox { get; }
-        public Akka.Dispatch.Mailbox CreateMailbox(Akka.Actor.Props props, Akka.Configuration.Config dispatcherConfig) { }
         public System.Type FromConfig(string path) { }
-        public System.Type GetMailboxType(Akka.Actor.Props props, Akka.Configuration.Config dispatcherConfig) { }
-        public System.Type LookupByQueueType(System.Type queueType) { }
+        public Akka.Dispatch.MailboxType GetMailboxType(Akka.Actor.Props props, Akka.Configuration.Config dispatcherConfig) { }
```
Aaronontheweb: Type checking methods for:
* Loading the correct `MailboxType` consistent with what's been requested in a HOCON declaration
* Loading the correct `MailboxType` consistent with what's been requested in a `IRequiresMessageQueue<TQueue>` declaration
* Verifying that the `MailboxType` produces the correct type of `IMessageQueue` needed to satisfy a `IRequiresMessageQueue<TQueue>` declaration

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2351,86 +2343,72 @@ namespace Akka.Dispatch
...
+        public System.Type GetRequiredType(System.Type actorType) { }
+        public bool HasRequiredType(System.Type actorType) { }
+        public Akka.Dispatch.MailboxType Lookup(string id) { }
+        public Akka.Dispatch.MailboxType LookupByQueueType(System.Type queueType) { }
+        public bool ProducesMessageQueue(System.Type mailboxType) { }
     }
-    [System.FlagsAttribute()]
-    public enum MailboxSuspendStatus
+    public abstract class MailboxType
```
Aaronontheweb: The new `MailboxType` class - used to instrument the creation of mailboxes and ensure they're instantiated with the correct `IMessageQueue` implementation.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2351,86 +2343,72 @@ namespace Akka.Dispatch
...
     public abstract class MessageDispatcher
     {
         public const int DefaultThroughput = 100;
         protected MessageDispatcher(Akka.Dispatch.MessageDispatcherConfigurator configurator) { }
         public Akka.Dispatch.MessageDispatcherConfigurator Configurator { get; }
-        public string Id { get; set; }
+        public static bool DebugDispatcher { get; }
+        public Akka.Event.EventStream EventStream { get; }
+        public string Id { get; }
+        protected long Inhabitants { get; }
```
Aaronontheweb: API modifications to `MessageDispatcher`. 

`Inhabitants` is essentially the reference count for each `MessageDispatcher`.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2540,45 +2523,47 @@ namespace Akka.Dispatch.MessageQueues
...
+        public UnboundedPriorityMessageQueue(System.Func<object, int> priorityGenerator, int initialCapacity) { }
         protected override int LockedCount { get; }
         protected override void LockedEnqueue(Akka.Actor.Envelope envelope) { }
         protected override bool LockedTryDequeue(out Akka.Actor.Envelope envelope) { }
     }
 }
 namespace Akka.Dispatch.SysMsg
 {
-    public sealed class ActorTask : Akka.Actor.INoSerializationVerificationNeeded, Akka.Dispatch.SysMsg.ISystemMessage
+    public sealed class ActorTask : Akka.Dispatch.SysMsg.SystemMessage
```
Aaronontheweb: `ISystemMessage` still exists, but now all sysmsg instances subclass from `SystemMessage` - an implementation class which includes all of the capabilities we need for linking system messages into lists.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2540,45 +2523,47 @@ namespace Akka.Dispatch.MessageQueues
...
     }
-    public sealed class Escalate : Akka.Actor.INoSerializationVerificationNeeded, Akka.Dispatch.SysMsg.ISystemMessage
+    public sealed class Escalate : Akka.Dispatch.SysMsg.SystemMessage
     {
         public Escalate(System.Exception reason) { }
         public System.Exception Reason { get; }
         public override string ToString() { }
     }
-    public sealed class Failed : Akka.Actor.INoSerializationVerificationNeeded, Akka.Dispatch.SysMsg.ISystemMessage
+    public sealed class Failed : Akka.Dispatch.SysMsg.SystemMessage, Akka.Dispatch.SysMsg.IStashWhenFailed
```
Aaronontheweb: System messages can now be marked with certain interfaces, such as this one `IStashWhenFailed`, which signals to the `ActorCell` that we should not process this system message until after the actor is no longer in a failed state.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2624,21 +2605,27 @@ namespace Akka.Dispatch.SysMsg
...
+    public abstract class SystemMessage : Akka.Actor.INoSerializationVerificationNeeded, Akka.Dispatch.SysMsg.ISystemMessage
+    {
+        protected SystemMessage() { }
+        public bool Unlinked { get; }
+        public void Unlink() { }
+    }
+    public sealed class Terminate : Akka.Dispatch.SysMsg.SystemMessage, Akka.Actor.IPossiblyHarmful
     {
-        public static Akka.Dispatch.SysMsg.Terminate Instance { get; }
+        public Terminate() { }
```
Aaronontheweb: `Terminate` is no longer a singleton. No system message instance can ever be shared between two actors, due to the linked lists built into them now.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -4709,6 +4695,22 @@ namespace Akka.Util
         bool IsEmpty { get; }
         Akka.Util.ILinearSeq<T> Tail();
     }
+    public class Index<TKey, TValue>
```
Aaronontheweb: Index - used for debugging inside the dispatchers. Not used in release builds.

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
```diff
@@ -463,7 +463,8 @@ public void SetDeliverySnapshot(AtLeastOnceDeliverySnapshot snapshot)
 
         public void Cancel()
         {
-            _redeliverScheduleCancelable.Cancel();
+            // need a null check here, in case actor is terminated before StartRedeliverTask() is called
+            _redeliverScheduleCancelable?.Cancel();
```
Aaronontheweb: As explained in the comment, this eliminated some `NullReferenceException`s I saw thrown in the `AtLeastOnceDeliveryActor` specs.

src/core/Akka.Remote.Tests/RemotingTerminatorSpecs.cs
```diff
@@ -68,5 +70,37 @@ public void RemotingTerminator_should_shutdown_promptly_with_some_associations()
                 Assert.True(terminationTask.Wait(RemainingOrDefault), "Expected to terminate within 10 seconds, but didn't.");
             });
         }
+
+        [Fact]
+        public void RemotingTerminator_should_shutdown_properly_with_remotely_deployed_actor()
```
Aaronontheweb: This spec helped me reproduce a bug with the `RemoteDeploymentWatcher` where we weren't properly sending clearing out dead remotely deployed actors. Passes now.

src/core/Akka.Remote/RemoteDeploymentWatcher.cs
```diff
@@ -19,8 +19,8 @@ namespace Akka.Remote
     internal class RemoteDeploymentWatcher : ReceiveActor, IRequiresMessageQueue<IUnboundedMessageQueueSemantics>
     {
 
-        private readonly IImmutableDictionary<IActorRef, IInternalActorRef> _supervisors =
-            ImmutableDictionary<IActorRef, IInternalActorRef>.Empty;
+        private readonly IDictionary<IActorRef, IInternalActorRef> _supervisors =
```
Aaronontheweb: ImmutableDictionary is what caused the bug (because of how we were using it) with not removing `IActorRef`s for terminated remotely deployed actors. Changed to a good, old-fashioned mutable dictionary instead.

src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -185,46 +204,25 @@ protected void OnReceive(object message, IActorRef sender)
...
-
-        /// <summary>
-        ///     Tells the internal.
-        /// </summary>
-        /// <param name="message">The message.</param>
-        /// <param name="sender">The sender.</param>
-        protected override void TellInternal(object message, IActorRef sender)
-        {
-            OnReceive(message, sender);
+                base.SendSystemMessage(message);
```
Aaronontheweb: Fixed the `RemoteSystemDaemon` to use the `SendSystemMessage` capabilities of the `MinimalActorRef` correctly.

src/core/Akka.Tests.Performance/Dispatch/MailboxMemoryFootprintSpec.cs
```diff
@@ -17,24 +20,43 @@ public class MailboxMemoryFootprintSpec
 
         private const string CreateThroughputCounter = "MailboxCreateThroughput";
         private const int MailboxCreateNumber = 10000;
+        private ActorSystem _actorSystem = ActorSystem.Create("Testing");
```
Aaronontheweb: Had to adjust the mailbox memory footprint specs to support the new `MailboxType` class.

src/core/Akka.Tests/Actor/ActorSystemSpec.cs
```diff
@@ -160,7 +160,7 @@ public void Find_actors_that_just_have_been_created()
             Assert.Equal(typeof(LocalActorRef), ExpectMsg<Type>());
         }
 
-        [Fact(Skip = "Context.AttachChild still needs to be implemented")]
+        [Fact()]
```
Aaronontheweb: Booyeah, fixed #2024 so we can run this spec without issue now.

src/core/Akka.Tests/Actor/FSMTimingSpec.cs
```diff
@@ -365,7 +365,7 @@ public StateMachine(IActorRef tester)
                             var contextLocal = Context;
                             var numberOfMessages = ((ActorCell) contextLocal).Mailbox.NumberOfMessages;
                             SetTimer("hallo", new Tock(), TimeSpan.FromMilliseconds(1));
-                            StaticAwaitCond(() => contextLocal.AsInstanceOf<ActorCell>().Mailbox.HasUnscheduledMessages, TimeSpan.FromSeconds(1), TimeSpan.FromMilliseconds(50));
+                            StaticAwaitCond(() => contextLocal.AsInstanceOf<ActorCell>().Mailbox.HasMessages, TimeSpan.FromSeconds(1), TimeSpan.FromMilliseconds(50));
```
Aaronontheweb: Name of the property is now `HasMessages`. Scheduling is solely determined by the `MessageDispatcher`.

src/core/Akka.Tests/Serialization/SerializationSpec.cs
```diff
@@ -453,7 +453,7 @@ public void Can_serialize_ActorPathContainer()
         [Fact]
         public void Can_serialize_singleton_messages()
         {
-            var message = Terminate.Instance;
+            var message = new Terminate();
```
Aaronontheweb: **NEED TO FIX**: should change this to `PoisonPill`; otherwise the test is invalid.

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -115,7 +115,13 @@ public void Stop(IActorRef child)
         /// <returns>The third value of the tuple that <paramref name="updater"/> returned.</returns>
         private TReturn UpdateChildrenRefs<TReturn>(Func<IChildrenContainer, Tuple<bool, IChildrenContainer, TReturn>> updater)
         {
-            return InterlockedSpin.ConditionallySwap(ref _childrenContainerDoNotCallMeDirectly, updater);
+            while (true)
```
Aaronontheweb: Inlined the code we were using before.

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -233,7 +239,7 @@ private void ResumeChildren(Exception causedByFailure, IActorRef perpetrator)
             foreach (var stats in ChildrenContainer.Stats)
             {
                 var child = stats.Child;
-                var cause = child.Equals(perpetrator) ? causedByFailure : null;
+                var cause = (perpetrator != null && child.Equals(perpetrator)) ? causedByFailure : null;
```
Aaronontheweb: found an edge case where the `perpetrator` can be null which can cause a `NullReferenceException` to be thrown inside the `Equals` method. Figured it would be better to add the null check here rather than modify the `Equals` method.

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -372,7 +378,7 @@ private IInternalActorRef MakeChild(Props props, string name, bool async, bool s
 
             // In case we are currently terminating, fail external attachChild requests
             // (internal calls cannot happen anyway because we are suspended)
-            if (ChildrenContainer.IsTerminating)
+            if (IsFastTerminating() || ChildrenContainer.IsTerminating)
```
Aaronontheweb: `IsFastTerminating()` checks a state flag I added to indicate that this `ActorCell` is terminating. Added it to solve #2024, where long story short: the immutable collection operations happening inside the constructor of the `ChildContainer` take too long and result in the `ChildContainer.IsTerminated` check on `ActorCell.MakeChild` always winning the race.

Adding this flag turned that check into a constant time operation and eliminated speediness of the `IChildContainer` as a potential race condition.

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -177,53 +178,116 @@ private void ReceiveSelection(ActorSelectionMessage m)
...
         }
 
-        /// <summary>
-        ///   Used to invoke system messages.
-        /// </summary>
-        /// <param name="envelope">The envelope.</param>
-        public void SystemInvoke(Envelope envelope)
+        /* SystemMessage handling */
+
+        private int CalculateState()
```
Aaronontheweb: Used to determine if we will stash an `ISystemMessage` or not.

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -353,8 +417,8 @@ private void Create(Exception failure)
         /// </summary>
         public virtual void Start()
         {
-            PreStart();
-            Mailbox.Start();
+            // This call is expected to start off the actor by scheduling its mailbox.
+            Dispatcher.Attach(this);
```
Aaronontheweb: `Dispatcher.Attach` starts the mailbox and registers the actor's mailbox for execution

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -376,25 +441,27 @@ public void Resume(Exception causedByFailure)
...
         }
 
+        /// <remarks>➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅</remarks>
         public void SendSystemMessage(ISystemMessage systemMessage)
         {
             try
             {
-                // TODO: use a dispatcher or the mailbox to post a message directly
-                _mailbox.Post(Self, new Envelope() { Message = systemMessage, Sender = ActorRefs.NoSender});
+                Dispatcher.SystemDispatch(this, (SystemMessage)systemMessage);
```
Aaronontheweb: `SystemDispatch` signals to the mailbox that it's OK to push a new system message into the actor, but not to touch any user messages.

src/core/Akka/Actor/ActorCell.FaultHandling.cs
```diff
@@ -20,12 +22,12 @@ public partial class ActorCell
...
         private void SuspendNonRecursive()
         {
-            Mailbox.Suspend();
+            Dispatcher.Suspend(this);
         }
 
         private void ResumeNonRecursive()
         {
-            UseThreadContext(() => Mailbox.Resume());
+            Dispatcher.Resume(this);
```
Aaronontheweb: The dispatcher will activate the mailbox and schedule it for execution here.

src/core/Akka/Actor/ActorCell.FaultHandling.cs
```diff
@@ -20,12 +22,12 @@ public partial class ActorCell
     {
         private void SuspendNonRecursive()
         {
-            Mailbox.Suspend();
+            Dispatcher.Suspend(this);
```
Aaronontheweb: Dispatcher will suspend the mailbox.

src/core/Akka/Actor/ActorCell.FaultHandling.cs
```diff
@@ -180,16 +182,39 @@ private void FinishCreate()
...
         }
 
+        /*
+         * Used to resolve the racy shutdown issue reported here: https://github.com/akkadotnet/akka.net/issues/2024
+         * Fundamentally the issue was that the ImmutableDictionary operations executed in the IChildContainer's constructor
+         * are so slow that several hundred `ChildContainer.IsTerminated` operations could pass even in the middle of signaling
+         * termination. So we signal shutdown using a byte flag instead so Termination can be signaled independently of the speed
+         * of the child container.
+         */
+        private volatile byte _fastTerminatingDoNotCallMeDirectly = 0;
```
Aaronontheweb: Read comment above for explanation of what this code does.

src/core/Akka/Actor/ActorCell.FaultHandling.cs
```diff
@@ -180,16 +182,39 @@ private void FinishCreate()
...
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private bool IsFastTerminating() { return _fastTerminatingDoNotCallMeDirectly == 1; }
+        
+
         /// <summary>Terminates this instance.</summary>
         private void Terminate()
         {
+            SetFastTerminate();
```
Aaronontheweb: Signal termination immediately.

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -172,6 +174,38 @@ public override MessageDispatcher Dispatcher()
...
+        /* dispatcher debugging helpers */
+        public static bool DebugDispatcher { get; } = false; // IMPORTANT: make this a compile-time constant so compiler will elide debug code in production
+        internal static readonly Lazy<Index<MessageDispatcher, IInternalActorRef>> Actors = new Lazy<Index<MessageDispatcher, IInternalActorRef>>(() => new Index<MessageDispatcher, IInternalActorRef>(), LazyThreadSafetyMode.PublicationOnly);
+
+        /// <summary>
+        /// INTERNAL API - Debugging purposes only! Should be elided by compiler in release builds.
+        /// </summary>
+        internal static void PrintActors()
+        {
+            if (DebugDispatcher)
```
Aaronontheweb: In release mode, elides the use of the code below

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -209,25 +257,177 @@ protected MessageDispatcher(MessageDispatcherConfigurator configurator)
...
+        /// or the system default specified.
+        /// </summary>
+        public TimeSpan ShutdownTimeout { get; internal set; }
+
+        /// <summary>
+        /// The number of actors attached to this <see cref="MessageDispatcher"/>
+        /// </summary>
+        protected long Inhabitants => Volatile.Read(ref _inhabitantsDoNotCallMeDirectly);
+
+        private long AddInhabitants(long add)
```
Aaronontheweb: Updates the reference count for this dispatcher

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -241,7 +441,62 @@ public virtual void SystemDispatch(ActorCell cell, Envelope envelope)
         /// <param name="cell">The ActorCell belonging to the actor who's attaching to this dispatcher.</param>
         public virtual void Attach(ActorCell cell)
         {
-            
+            Register(cell);
+            RegisterForExecution(cell.Mailbox, false, true);
```
Aaronontheweb: Turns on the mailbox for the very first time when an actorcell attaches to this dispatcher.

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -241,7 +441,62 @@ public virtual void SystemDispatch(ActorCell cell, Envelope envelope)
...
+        /// 
+        /// Suggest to register the provided mailbox for execution
+        /// </summary>
+        /// <param name="mbox">The mailbox</param>
+        /// <param name="hasMessageHint">Do we have any messages?</param>
+        /// <param name="hasSystemMessageHint">Do we have any system messages?</param>
+        /// <returns><c>true</c> if the <see cref="Mailbox"/> was scheduled for execution, otherwise <c>false</c>.</returns>
+        internal bool RegisterForExecution(Mailbox mbox, bool hasMessageHint, bool hasSystemMessageHint)
+        {
+            if (mbox.CanBeScheduledForExecution(hasMessageHint, hasSystemMessageHint)) //This needs to be here to ensure thread safety and no races
```
Aaronontheweb: Guarantees that only one mailbox run can occur at a time.

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -241,7 +441,62 @@ public virtual void SystemDispatch(ActorCell cell, Envelope envelope)
...
+            {
+                if (mbox.SetAsScheduled())
```
Aaronontheweb: Second check here - only start the run if we can successfully set the mailbox to `Scheduled`. Several reasons why this may not happen:
1. Mailbox has been closed
2. Mailbox is already scheduled

src/core/Akka/Dispatch/IRequiresMessageQueue.cs
```diff
@@ -5,8 +5,16 @@
...
 
+using Akka.Actor;
+using Akka.Dispatch.MessageQueues;
+
 namespace Akka.Dispatch
 {
+    /// <summary>
+    /// Used to help give hints to the <see cref="ActorSystem"/> as to what types of <see cref="IMessageQueue"/> this
+    /// actor requires. Used mostly for system actors.
+    /// </summary>
```
Aaronontheweb: Decided to take mercy on @sean-gilliam and add XML-DOC comments wherever I was spending a lot of time :p

Aaronontheweb:
NOOOOOOOOOOOOOOOOOOOOOOOOOOOOO

image

Aaronontheweb:
Well, at least it's down from several hundred to just 1 :p

Aaronontheweb:
This thing is ready for review - I'll be taking a look at the perf numbers and making some adjustments there.

Aaronontheweb:
cc @rogeralsing @Horusiath @JeffCyr per what we were discussing in Gitter, I think we're eventually going to have to use something like a lock to fully solve the race issue with the top-level actors. As the first run of the spec showed, even with the CAS stuff I have in place it may not be enough.

JeffCyr:
@aaronontheweb You need to use AttachChild to create top level actors in ActorSystem.ActorOf

src/core/Akka.Tests/Actor/ActorSystemSpec.cs
```diff
@@ -463,12 +463,17 @@ public DispatcherImpl(MessageDispatcherConfigurator configurator) : base(configu
             {
             }
 
-            public override void Schedule(Action run)
+            public override void Schedule(IRunnable run)
             {
-                Task.Run(run);
+                run.Run();
```
rogeralsing: da du run run

rogeralsing:
I've gotten through half of the PR so far.
Looks solid so far. will continue later today

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -54,7 +54,7 @@ private bool TryGetChild(string name, out IActorRef child)
 
         public virtual IActorRef AttachChild(Props props, bool isSystemService, string name = null)
```
alexvaluyskiy: why public, not internal?

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -54,7 +54,7 @@ private bool TryGetChild(string name, out IActorRef child)
...
         {
-            return ActorOf(props, name, true, isSystemService);
+            return MakeChild(props, name ?? GetRandomActorName(), true, isSystemService);
```
alexvaluyskiy: Should call CheckName if name is not empty.

src/core/Akka/Actor/ActorCell.Children.cs
alexvaluyskiy: why public, not internal?

src/core/Akka/Actor/RepointableActorRef.cs
```diff
@@ -24,16 +24,16 @@ public class RepointableActorRef : ActorRefWithCell, IRepointableRef
         private readonly ActorSystemImpl _system;
         private readonly Props _props;
         private readonly MessageDispatcher _dispatcher;
-        private readonly Func<Mailbox> _createMailbox;
+        private readonly MailboxType _mailboxType;
```
alexvaluyskiy: I need all that fields in RoutedActorRef, maybe we should make them as protected?

alexvaluyskiy:
@Aaronontheweb I've found a bug with GetRandomActorName
```C#
        [Fact]
        public void Actor_names_must_starts_from_a()
        {
            Sys.ActorOf(Props.Create<EchoActor>(TestActor)).Path.ToStringWithoutAddress().Should().Be("/user/$a");
        }
```
This spec fails, it returns "/user/$b"

alexvaluyskiy:
@Aaronontheweb why you have not changed Post to SendMessage? We've already have breaking changes here, and will have them in Routing PR.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Aaronontheweb: cc @rogeralsing one note; when we make an `async` / `await` call inside actors right now, SystemMessages are still allowed to be processed. I personally think that's fine - the only real thing that could cause an `ActorContext` to significantly change in the midst of that would be a child terminating, which we couldn't block anyway (since it happens inside the context of another actor.)

That being said, if we find a case where we need to suspend certain system messages during async await we can extend this `IStashWhen__` convention to support that.

Aaronontheweb:
@JeffCyr 

> You need to use AttachChild to create top level actors in ActorSystem.ActorOf

So I took a look at this when I was going through it:

This is `ActorSystemImpl.ActorOf`
```csharp
public override IActorRef ActorOf(Props props, string name = null)
        {
            return _provider.Guardian.Cell.ActorOf(props, name: name);
        }
```

Cell.ActorOf calls

```csharp
private IActorRef ActorOf(Props props, string name, bool isAsync, bool isSystemService)
        {
            if (name == null)
                name = GetRandomActorName();
            else
                CheckName(name);

            return MakeChild(props, name, isAsync, isSystemService);
        }
```

`ActorCell.AttachChild` does the following:

```csharp
public virtual IActorRef AttachChild(Props props, bool isSystemService, string name = null)
        {
            return MakeChild(props, name ?? GetRandomActorName(), true, isSystemService);
        }
```

Basically they both end up doing the same things. I can consolidate these, but the behavior is the same. It's the same with Scala https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/dungeon/Children.scala#L43

```scala
private[akka] def attachChild(props: Props, systemService: Boolean): ActorRef =
    makeChild(this, props, randomName(), async = true, systemService = systemService)
  private[akka] def attachChild(props: Props, name: String, systemService: Boolean): ActorRef =
    makeChild(this, props, checkName(name), async = true, systemService = systemService)
```

Aaronontheweb:
Only real difference is the extra `CheckName` call in `ActorOf`

Aaronontheweb:
@alexvaluyskiy 

> why you have not changed Post to SendMessage? We've already have breaking changes here, and will have them in Routing PR.

Eh, fair enough. I'll change that for the sake of consistency with the JVM.

src/core/Akka/Actor/ActorCell.Children.cs
Aaronontheweb: That happens in the `ActorOf` method.

src/core/Akka/Actor/ActorCell.Children.cs
Aaronontheweb: Good point - I'll change that.

src/core/Akka/Actor/RepointableActorRef.cs
Aaronontheweb: yeah I noticed that earlier too that we had duplicates of these fields in `RouterActorRef`. I'll make this change and consolidate them.

Aaronontheweb:
@alexvaluyskiy 

> I've found a bug with GetRandomActorName

Hah, I wonder how that happened? Maybe we're starting a counter at 1 instead of at 0 or something like that...

src/core/Akka.Tests/Actor/ActorSystemSpec.cs
Aaronontheweb: *Barbara annnnnnnnnnnnnnnnnnnnnn*

Aaronontheweb:
Some performance items: in particular, I'm interested in seeing how the dispatcher and mailbox changes might have affected throughput.

Let's do this top-down:

## Before
# Akka.Remote.Tests.Performance.Transports.HeliosRemoteMessagingThroughputSpec+OneWay
__Measures the throughput of Akka.Remote over a particular transport using one-way messaging__
_6/11/2016 8:45:51 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
NumberOfIterations=3, MaximumRunTime=00:00:01
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalCollections [Gen0] |     collections |          191.00 |          184.00 |          174.00 |            8.89 |
TotalCollections [Gen1] |     collections |           11.00 |           10.00 |            9.00 |            1.00 |
TotalCollections [Gen2] |     collections |            1.00 |            1.00 |            1.00 |            0.00 |
[Counter] RemoteMessageReceived |      operations |       50,000.00 |       50,000.00 |       50,000.00 |            0.00 |

### Per-second Totals
TotalCollections [Gen0] |     collections |           74.76 |           73.08 |           70.94 |            1.95 |
TotalCollections [Gen1] |     collections |            4.33 |            3.98 |            3.52 |            0.41 |
TotalCollections [Gen2] |     collections |            0.41 |            0.40 |            0.39 |            0.01 |
[Counter] RemoteMessageReceived |      operations |       20,386.25 |       19,873.65 |       19,570.59 |          446.38 |


## After 
# Akka.Remote.Tests.Performance.Transports.HeliosRemoteMessagingThroughputSpec+OneWay
__Measures the throughput of Akka.Remote over a particular transport using one-way messaging__
_6/12/2016 10:59:30 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
NumberOfIterations=3, MaximumRunTime=00:00:01
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalCollections [Gen0] |     collections |          190.00 |          179.67 |          169.00 |           10.50 |
TotalCollections [Gen1] |     collections |           15.00 |           12.00 |            8.00 |            3.61 |
TotalCollections [Gen2] |     collections |            3.00 |            2.67 |            2.00 |            0.58 |
[Counter] RemoteMessageReceived |      operations |       50,000.00 |       50,000.00 |       50,000.00 |            0.00 |

### Per-second Totals
TotalCollections [Gen0] |     collections |          102.19 |           98.06 |           92.02 |            5.35 |
TotalCollections [Gen1] |     collections |            7.89 |            6.54 |            4.36 |            1.91 |
TotalCollections [Gen2] |     collections |            1.70 |            1.46 |            1.09 |            0.32 |
[Counter] RemoteMessageReceived |      operations |       28,386.28 |       27,305.28 |       26,305.83 |        1,042.62 |

The most recent perf runs on the `dev` branch have all placed the Helios transport's throughput during one-way messaging in the 20k-22k msg / s range. This spec has it clocking in between 26.3-28.3k msg / s. Looks promising!

Aaronontheweb:
Now looking at *just* the `ForkJoinDispatcher` running the normal remoting pipeline:

## Before
# Akka.Remote.Tests.Performance.ForkJoinDispatcherRemoteMessagingThroughputSpec+OneWay
__Measures the throughput of Akka.Remote over a particular transport using one-way messaging__
_6/11/2016 8:44:31 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
NumberOfIterations=3, MaximumRunTime=00:00:01
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalCollections [Gen0] |     collections |          118.00 |          117.33 |          117.00 |            0.58 |
TotalCollections [Gen1] |     collections |            7.00 |            7.00 |            7.00 |            0.00 |
TotalCollections [Gen2] |     collections |            2.00 |            1.33 |            1.00 |            0.58 |
[Counter] RemoteMessageReceived |      operations |       50,000.00 |       50,000.00 |       50,000.00 |            0.00 |

### Per-second Totals
TotalCollections [Gen0] |     collections |           99.17 |           97.34 |           95.45 |            1.86 |
TotalCollections [Gen1] |     collections |            5.88 |            5.81 |            5.71 |            0.09 |
TotalCollections [Gen2] |     collections |            1.68 |            1.11 |            0.82 |            0.49 |
[Counter] RemoteMessageReceived |      operations |       42,020.92 |       41,477.23 |       40,788.96 |          628.58 |

### Raw Data
#### TotalCollections [Gen0]
           Run # |     collections | collections / s |ns / collections |
               1 |          118.00 |           99.17 |   10,083,759.32 |
               2 |          117.00 |           97.40 |   10,267,462.39 |
               3 |          117.00 |           95.45 |   10,477,110.26 |

#### TotalCollections [Gen1]
           Run # |     collections | collections / s |ns / collections |
               1 |            7.00 |            5.88 |  169,983,371.43 |
               2 |            7.00 |            5.83 |  171,613,300.00 |
               3 |            7.00 |            5.71 |  175,117,414.29 |

#### TotalCollections [Gen2]
           Run # |     collections | collections / s |ns / collections |
               1 |            2.00 |            1.68 |  594,941,800.00 |
               2 |            1.00 |            0.83 |1,201,293,100.00 |
               3 |            1.00 |            0.82 |1,225,821,900.00 |

#### [Counter] RemoteMessageReceived
               1 |       50,000.00 |       42,020.92 |       23,797.67 |
               2 |       50,000.00 |       41,621.82 |       24,025.86 |
               3 |       50,000.00 |       40,788.96 |       24,516.44

## After
# Akka.Remote.Tests.Performance.ForkJoinDispatcherRemoteMessagingThroughputSpec+OneWay
__Measures the throughput of Akka.Remote over a particular transport using one-way messaging__
_6/12/2016 10:58:22 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
NumberOfIterations=3, MaximumRunTime=00:00:01
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalCollections [Gen0] |     collections |          118.00 |          117.33 |          117.00 |            0.58 |
TotalCollections [Gen1] |     collections |            7.00 |            7.00 |            7.00 |            0.00 |
TotalCollections [Gen2] |     collections |            2.00 |            1.33 |            1.00 |            0.58 |
[Counter] RemoteMessageReceived |      operations |       50,000.00 |       50,000.00 |       50,000.00 |            0.00 |

### Per-second Totals
TotalCollections [Gen0] |     collections |          113.05 |          110.44 |          107.68 |            2.69 |
TotalCollections [Gen1] |     collections |            6.71 |            6.59 |            6.44 |            0.13 |
TotalCollections [Gen2] |     collections |            1.92 |            1.26 |            0.92 |            0.57 |
[Counter] RemoteMessageReceived |      operations |       47,903.78 |       47,061.55 |       46,016.56 |          959.81 |

### Raw Data
#### TotalCollections [Gen0]
           Run # |     collections | collections / s |ns / collections |
               1 |          117.00 |          110.60 |    9,041,712.82 |
               2 |          117.00 |          107.68 |    9,286,882.91 |
               3 |          118.00 |          113.05 |    8,845,415.25 |

#### TotalCollections [Gen1]
           Run # |     collections | collections / s |ns / collections |
               1 |            7.00 |            6.62 |  151,125,771.43 |
               2 |            7.00 |            6.44 |  155,223,614.29 |
               3 |            7.00 |            6.71 |  149,108,428.57 |

#### TotalCollections [Gen2]
           Run # |     collections | collections / s |ns / collections |
               1 |            1.00 |            0.95 |1,057,880,400.00 |
               2 |            1.00 |            0.92 |1,086,565,300.00 |
               3 |            2.00 |            1.92 |  521,879,500.00 |

#### [Counter] RemoteMessageReceived
               1 |       50,000.00 |       47,264.32 |       21,157.61 |
               2 |       50,000.00 |       46,016.56 |       21,731.31 |
               3 |       50,000.00 |       47,903.78 |       20,875.18 |

Previously clocked in between 40.8k - 42k msg / s. Doing about 46-48k msg / s now. Seems consistent with the performance change we saw when using Helios.

Aaronontheweb:
Now let's look at the `ReceiveActor` throughput numbers:

## Before
# Akka.Tests.Performance.Actor.ReceiveActorThroughputSpec+Benchmark
__Measures the throughput of an actor implementation__
_6/11/2016 8:57:44 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalCollections [Gen0] |     collections |           55.00 |           53.46 |           52.00 |            1.05 |
TotalCollections [Gen1] |     collections |           24.00 |           20.46 |           18.00 |            1.56 |
TotalCollections [Gen2] |     collections |            7.00 |            5.38 |            4.00 |            0.96 |
[Counter] MessageReceived |      operations |   10,000,000.00 |   10,000,000.00 |   10,000,000.00 |            0.00 |

### Per-second Totals
TotalCollections [Gen0] |     collections |           22.59 |           20.20 |           17.74 |            1.28 |
TotalCollections [Gen1] |     collections |            9.86 |            7.74 |            6.48 |            0.82 |
TotalCollections [Gen2] |     collections |            2.55 |            2.03 |            1.36 |            0.37 |
[Counter] MessageReceived |      operations |    4,107,453.28 |    3,778,396.57 |    3,412,137.21 |      226,468.61 |

## After
# Akka.Tests.Performance.Actor.ReceiveActorThroughputSpec+Benchmark
__Measures the throughput of an actor implementation__
_6/12/2016 11:11:28 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalCollections [Gen0] |     collections |           46.00 |           45.15 |           44.00 |            0.69 |
TotalCollections [Gen1] |     collections |           19.00 |           18.38 |           17.00 |            0.65 |
TotalCollections [Gen2] |     collections |            6.00 |            5.38 |            4.00 |            0.65 |
[Counter] MessageReceived |      operations |   10,000,000.00 |   10,000,000.00 |   10,000,000.00 |            0.00 |

### Per-second Totals
TotalCollections [Gen0] |     collections |           18.83 |           17.11 |           15.61 |            0.82 |
TotalCollections [Gen1] |     collections |            7.78 |            6.97 |            6.38 |            0.41 |
TotalCollections [Gen2] |     collections |            2.46 |            2.04 |            1.52 |            0.28 |
[Counter] MessageReceived |      operations |    4,092,658.78 |    3,788,223.95 |    3,541,850.70 |      148,096.66 |

The numbers are virtually identical. No real changes there. Both of these tests use the `ThreadPoolDispatcher`.

JeffCyr:
@Aaronontheweb With AttachChild, isAsync is true, this will create a RepointableActorRef instead of LocalActorRef and this will fix the remaining race condition you were observing

Aaronontheweb:
@JeffCyr oh man, I didn't even notice that! Awesome - I'll make that change.

JeffCyr:
This will break a lot of spec that were hardcoded to expect a LocalActorRef, good luck :-)

Aaronontheweb:
Last but not least, and the area where I expect the biggest improvement: memory consumption.

First up, `ReceiveActor`

## Before
# Akka.Tests.Performance.Actor.ActorMemoryFootprintSpec+ReceiveActor_memory_footprint
__Measures the amount of memory used by 10,000 ReceiveActors__
_6/11/2016 8:50:49 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalBytesAllocated |           bytes |   27,889,440.00 |   27,439,748.92 |   26,021,592.00 |      480,147.59 |
[Counter] ActorCreateThroughput |      operations |       10,000.00 |       10,000.00 |       10,000.00 |            0.00 |

## After
# Akka.Tests.Performance.Actor.ActorMemoryFootprintSpec+ReceiveActor_memory_footprint
__Measures the amount of memory used by 10,000 ReceiveActors__
_6/12/2016 11:03:14 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalBytesAllocated |           bytes |   20,782,904.00 |   18,024,289.23 |   17,336,352.00 |      863,440.75 |
[Counter] ActorCreateThroughput |      operations |       10,000.00 |       10,000.00 |       10,000.00 |            0.00 |

Aaronontheweb:
10,000 `ReceiveActors` used to consume 27.4mb of memory, using the previous `Mailbox` implementation -  2.74kb per actor.

That number is now **18mb** for 10,000 receive actors: 1.8kb per actor.

**Net memory footprint has been reduced by 34%**. I call that a success :)

Aaronontheweb:
Now comparing the memory footprints for *just* the mailboxes, using the default `UnboundedMessageQueue` mailbox implementation.

## Before
# Akka.Tests.Performance.Dispatch.MailboxMemoryFootprintSpec+ConcurrentQueueMailbox
__Measures the amount of memory used by 10,000 ConcurrentQueueMailboxes__
_6/11/2016 8:50:26 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalBytesAllocated |           bytes |   14,801,752.00 |   14,796,950.77 |   14,794,000.00 |        1,930.39 |
[Counter] MailboxCreateThroughput |      operations |       10,000.00 |       10,000.00 |       10,000.00 |            0.00 |

## After
# Akka.Tests.Performance.Dispatch.MailboxMemoryFootprintSpec+UnboundedMailbox
__Measures the amount of memory used by 10,000 ConcurrentQueueMailboxes__
_6/12/2016 11:02:44 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalBytesAllocated |           bytes |    7,759,504.00 |    7,757,964.92 |    7,755,696.00 |        1,214.36 |
[Counter] MailboxCreateThroughput |      operations |       10,000.00 |       10,000.00 |       10,000.00 |            0.00 |

Aaronontheweb:
Memory footprint of the default mailbox is roughly half of what it was before. No surprises there - deleting the second `ConcurrentQueue` for system messages and replacing it with a linked list would do that.

alexvaluyskiy:
@Aaronontheweb you have created a lot of benchmarks. Could you create the one for the routers? :)

Aaronontheweb:
@alexvaluyskiy sure thing - would be happy to do that on another PR today.

rogeralsing:
Dev branch:

dev


This branch:
aaron

There is a 20%ish perfloss with the new mailboxes on the fast path.

rogeralsing:
That being said, I personally don't mind, the PingPong benchmark is super artificial and holds little value.
But at least we should know what changes affect the system

Aaronontheweb:
@rogeralsing I agree with you. There's actually a lot we can do to improve the performance around the `MessageDispatcher` - I opted not to do that because I wanted to get a sanity check from the Akka.NET team on these changes overall first.

What would you prefer: we create an issue to tackle performance stuff inside the `MessageDispatcher` later or we try to do it now as part of this PR?

I'd opt for doing it in a separate step just so we don't block @alexvaluyskiy and his router changes he's working on right now. I also think it'd be easier for other folks to try making their own modifications / improvements that way.

alexvaluyskiy:
@Aaronontheweb 
> Hah, I wonder how that happened? Maybe we're starting a counter at 1 instead of at 0 or something like that...
Because they use GetAndIncrement, but we are using something like IncrementAndGet

src/core/Akka/Actor/ActorCell.Children.cs
alexvaluyskiy: But AttachChild is not related with ActorOf. Jvm version uses checkName in both of them

Aaronontheweb:
Added some new commits that address all of the issues and feedback thus far.

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -18,7 +18,7 @@ namespace Akka.Actor
     public partial class ActorCell
     {
         private volatile IChildrenContainer _childrenContainerDoNotCallMeDirectly = EmptyChildrenContainer.Instance;
-        private long _nextRandomNameDoNotCallMeDirectly;
+        private long _nextRandomNameDoNotCallMeDirectly = -1; // Interlocked.Increment automatically adds 1 to this value. Allows us to start from 0.
```
alexvaluyskiy: Cluster router uses AtomicCounter for this purposes. Could we use it here. Or it has some drawbacks?

src/core/Akka/Actor/RepointableActorRef.cs
```diff
@@ -342,7 +342,7 @@ public bool TryGetChildStatsByName(string name, out IChildStats child)
             return false;
         }
 
-        public void Post(IActorRef sender, object message)
+        public void SendMessage(IActorRef sender, object message)
```
alexvaluyskiy: Still without Envelope parameter :)

Aaronontheweb: Going to skip that method for now - requires some yak-shaving in how we implement message sending and I need to spend time working on other things. Not essential at the moment.

src/core/Akka/Actor/ActorCell.Children.cs
Aaronontheweb: AtomicCounter is just a wrapper around this.

Aaronontheweb:
Implemented what I consider to be the "final" set of changes to this PR. Should be ready for another pass now.

Aaronontheweb:
Since the `AttachChild` change these specs are failing racily:

 Test collection for Akka.Streams.Tests.Dsl.FlowIntersperseSpec (1): Akka.Streams.Tests.Dsl (1)
  FlowIntersperseSpec.A_Intersperse_must_demonstrate_how_to_prepend_only	
 Test collection for Akka.Streams.Tests.Dsl.FlowLimitSpec (1): Akka.Streams.Tests.Dsl (2)
  FlowLimitSpec.A_Limit_must_produce_n_messages_before_throwing_a_StreamLimitReachedException_when_n_lower_than_input_size	
  FlowLimitSpec.A_Limit_must_throw_a_StreamLimitReachedException_when_n_lower_than_0	
 
First failure:  	2086	#1228	No changes	13 Jun 16 22:41
System.AggregateException : One or more errors occurred.
---- Akka.Streams.StreamLimitReachedException : Limit of -1 reached
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Akka.Streams.Dsl.Sink.<>c__3`1.<First>b__3_0(Task`1 e) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.Streams\Dsl\Sink.cs:line 149
   at Akka.Streams.Implementation.Module.<>c__DisplayClass26_0`2.<TransformMaterializedValue>b__0(Object x) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.Streams\Implementation\StreamLayout.cs:line 408
   at Akka.Streams.Implementation.MaterializerSession.ResolveMaterialized(IMaterializedValueNode node, IDictionary`2 values, String indent) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.Streams\Implementation\StreamLayout.cs:line 1160
   at Akka.Streams.Implementation.MaterializerSession.ResolveMaterialized(IMaterializedValueNode node, IDictionary`2 values, String indent) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.Streams\Implementation\StreamLayout.cs:line 1154
   at Akka.Streams.Implementation.MaterializerSession.MaterializeModule(IModule module, Attributes effectiveAttributes) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.Streams\Implementation\StreamLayout.cs:line 1134
   at Akka.Streams.Implementation.MaterializerSession.Materialize() in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.Streams\Implementation\StreamLayout.cs:line 1080
   at Akka.Streams.Implementation.ActorMaterializerImpl.Materialize[TMat](IGraph`2 runnable, Func`2 subFlowFuser) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.Streams\Implementation\ActorMaterializerImpl.cs:line 242
   at Akka.Streams.Tests.Dsl.FlowLimitSpec.A_Limit_must_throw_a_StreamLimitReachedException_when_n_lower_than_0() in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.Streams.Tests\Dsl\FlowLimitSpec.cs:line 88
----- Inner Stack Trace -----
« Hide stacktrace

Aaronontheweb:
Going to take a look - I assume they're guilty of the same type of assumption that the `TestScheduler` specs were making.

Aaronontheweb:
Merged in @cconstantin's changes and that seemed to help with the Akka.Streams specs. 

We know there's some issues with routers - and @alexvaluyskiy is working on that. But this PR needs to come in first before we can address those, as his changes depend on this. Are there any further changes that need to be made before this can be merged?

src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -100,12 +100,27 @@ public RemoteSystemDaemon(ActorSystemImpl system, ActorPath path, IInternalActor
...
+            _terminating.WhileOn(() =>
+            {
+                if (!HasChildren)
+                {
+                    _terminator.Tell(TerminationHookDone.Instance, this);
+                }
+            });
+        }
+
+  //      def terminationHookDoneWhenNoChildren(): Unit = terminating.whileOn {
```
cconstantin: this Scala code can be removed.

cconstantin:
@Aaronontheweb looks good to me.

src/core/Akka.Remote/RemoteSystemDaemon.cs
Aaronontheweb: Ok, I'll delete that

Aaronontheweb:
@cconstantin alrighty, all set

PR #2087, opened 2016-06-13
A persistent view replays events during the recovery despite the fact…

artem-karnaukh:
A persistent view replays events during the recovery despite the fact, that Recovery property is overriden and returns Recovery.None

### DIFF
--- a/src/core/Akka.Persistence.Tests/PersistentViewSpec.Actors.cs
+++ b/src/core/Akka.Persistence.Tests/PersistentViewSpec.Actors.cs
@@ -293,6 +293,36 @@ namespace Akka.Persistence.Tests
         }
 
         #endregion
+
+        internal class RecoveryTestPersistentView : PersistentView
+        {
+            private readonly string _name;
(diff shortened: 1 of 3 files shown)


cconstantin:
@artem-karnaukh thanks for submitting fix. Would you mind including a test to illustrate the issue?

Aaronontheweb:
@artem-karnaukh looks good to me - please go ahead and squash these commits and we'll merge them in.

cconstantin:
@artem-karnaukh @Aaronontheweb: I put this on hold. Can you please update the test to demonstrate that events are replayed when `Recovery.None`? It does not do that in the current form. Sorry for insisting on this, but it appears to me that the existing code works correctly.

Danthar:
Whats the status on this @cconstantin ? Does this PR addresses a valid problem?

cconstantin:
@Danthar there was no reply after my request for a proper test. I don't believe this is a valid issue.

Danthar:
Stale, and regarding the above comments. Im closing this.

PR #2090, opened 2016-06-14
ported QuickRestartSpec; fixed EndpointManager errors

Aaronontheweb:
added ActorsLeakSpec
resolved two EndpointManager bugs that would have prevented reconnects

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -49,6 +49,7 @@ namespace Akka.Actor
         public const int UndefinedUid = 0;
         public ActorCell(Akka.Actor.Internal.ActorSystemImpl system, Akka.Actor.IInternalActorRef self, Akka.Actor.Props props, Akka.Dispatch.MessageDispatcher dispatcher, Akka.Actor.IInternalActorRef parent) { }
         protected Akka.Actor.ActorBase Actor { get; }
+        public Akka.Actor.Internal.IChildrenContainer ChildrenContainer { get; }
         public int CurrentEnvelopeId { get; }
         public object CurrentMessage { get; }
         public Akka.Dispatch.MessageDispatcher Dispatcher { get; }
@@ -158,6 +159,7 @@ namespace Akka.Actor
         protected ActorPath(Akka.Actor.ActorPath parentPath, string name, long uid) { }
         public Akka.Actor.Address Address { get; }
         public System.Collections.Generic.IReadOnlyList<string> Elements { get; }
+        public System.Collections.Generic.IReadOnlyList<string> ElementsWithUid { get; }
         public string Name { get; }
         public abstract Akka.Actor.ActorPath Parent { get; }
         public abstract Akka.Actor.ActorPath Root { get; }
@@ -644,6 +646,7 @@ namespace Akka.Actor
         protected ExtendedActorSystem() { }
         public abstract Akka.Actor.ActorProducerPipelineResolver ActorPipelineResolver { get; }
         public abstract Akka.Actor.IInternalActorRef Guardian { get; }
+        public abstract Akka.Actor.IInternalActorRef LookupRoot { get; }
         public abstract Akka.Actor.IActorRefProvider Provider { get; }
         public abstract Akka.Actor.IInternalActorRef SystemGuardian { get; }
         public abstract Akka.Actor.IActorRef SystemActorOf(Akka.Actor.Props props, string name = null);
@@ -921,6 +924,7 @@ namespace Akka.Actor
     }
     public interface ICell
     {
+        Akka.Actor.Internal.IChildrenContainer ChildrenContainer { get; }
         bool HasMessages { get; }
         bool IsLocal { get; }
         bool IsTerminated { get; }
@@ -931,6 +935,7 @@ namespace Akka.Actor
         Akka.Actor.ActorSystem System { get; }
         Akka.Actor.Internal.ActorSystemImpl SystemImpl { get; }
         Akka.Actor.IInternalActorRef GetChildByName(string name);
+        [System.ObsoleteAttribute("Used ChildrenRefs instead")]
         System.Collections.Generic.IEnumerable<Akka.Actor.IInternalActorRef> GetChildren();
         Akka.Actor.IInternalActorRef GetSingleChild(string name);
         void Restart(System.Exception cause);
@@ -1608,6 +1613,7 @@ namespace Akka.Actor
     public class UnstartedCell : Akka.Actor.ICell
     {
         public UnstartedCell(Akka.Actor.Internal.ActorSystemImpl system, Akka.Actor.RepointableActorRef self, Akka.Actor.Props props, Akka.Actor.IInternalActorRef supervisor) { }
+        public Akka.Actor.Internal.IChildrenContainer ChildrenContainer { get; }
         public bool HasMessages { get; }
         public bool IsLocal { get; }
         public bool IsTerminated { get; }
@@ -1748,6 +1754,7 @@ namespace Akka.Actor.Internal
         public override Akka.Event.EventStream EventStream { get; }
         public override Akka.Actor.IInternalActorRef Guardian { get; }
         public override Akka.Event.ILoggingAdapter Log { get; }
+        public override Akka.Actor.IInternalActorRef LookupRoot { get; }
         public override Akka.Dispatch.Mailboxes Mailboxes { get; }
         public override string Name { get; }
         public override Akka.Actor.IActorRefProvider Provider { get; }
--- a/src/core/Akka.Cluster.Tests.MultiNode/QuickRestartSpec.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/QuickRestartSpec.cs
@@ -1,0 +1,141 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading;
+using Akka.Actor;
+using Akka.Cluster.TestKit;
+using Akka.Configuration;
+using Akka.Remote.TestKit;
+using Akka.TestKit;
+using Akka.Util;
+
+namespace Akka.Cluster.Tests.MultiNode
+{
+    public class QuickRestartSpecConfig : MultiNodeConfig
+    {
+        public RoleName First { get; }
+        public RoleName Second { get; }
+        public RoleName Third { get; }
+
+        public QuickRestartSpecConfig()
+        {
+            First = Role("first");
+            Second = Role("second");
+            Third = Role("third");
+
+            CommonConfig = DebugConfig(false)
+                .WithFallback(ConfigurationFactory.ParseString(@"
+                    akka.cluster.auto-down-unreachable-after = off
+                "))
+                .WithFallback(MultiNodeClusterSpec.ClusterConfig());
+        }
+    }
+
+    public class QuickRestartMultiNode1 : QuickRestartSpec { }
+    public class QuickRestartMultiNode2 : QuickRestartSpec { }
+    public class QuickRestartMultiNode3 : QuickRestartSpec { }
+
+    public abstract class QuickRestartSpec : MultiNodeClusterSpec
+    {
+        private readonly QuickRestartSpecConfig _config;
+        private readonly Lazy<ImmutableList<Address>> _seedNodes;
+        private int _rounds = 3;
+
+        protected override TimeSpan ShutdownTimeout => TimeSpan.FromSeconds(45 * _rounds);
+
+        protected QuickRestartSpec() : this(new QuickRestartSpecConfig())
+        {
+
+        }
+
+        protected QuickRestartSpec(QuickRestartSpecConfig config) : base(config)
+        {
+            _config = config;
+            _seedNodes = new Lazy<ImmutableList<Address>>(() => ImmutableList<Address>.Empty
+            .Add(GetAddress(_config.First))
+            .Add(GetAddress(_config.Second))
+            .Add(GetAddress(_config.Third)));
+        }
+
+        [MultiNodeFact]
+        public void QuicklyRestartingNodeMust()
+        {
+            SetupStableSeedNodes();
+            JoinAndRestart();
+        }
+
+        private void SetupStableSeedNodes()
+        {
+            Within(TimeSpan.FromSeconds(15), () =>
+            {
+                Cluster.JoinSeedNodes(_seedNodes.Value);
+                AwaitMembersUp(Roles.Count);
+                EnterBarrier("stable");
+            });
+        }
+
+        private void JoinAndRestart()
+        {
+            var totalNumberOfNodes = Roles.Count + 1;
+            ActorSystem restartingSystem = null; // only used on second
+            for (var i = 1; i <= _rounds; i++)
+            {
+                var round = i; //non-loop variable closure
+                Log.Info("Round-{0}", i);
+                RunOn(() =>
+                {
+                    restartingSystem = restartingSystem == null
+                        ? ActorSystem.Create(Sys.Name, ConfigurationFactory.ParseString($"akka.cluster.roles=[round-{round}]")
+                            .WithFallback(Sys.Settings.Config))
+                        : ActorSystem.Create(Sys.Name, ConfigurationFactory.ParseString($"akka.cluster.roles=[round-{round}]")
+                            .WithFallback(
+                                $"akka.remote.helios.tcp.port={Cluster.Get(restartingSystem).SelfAddress.Port}")
+                            .WithFallback(Sys.Settings.Config));
+                    Log.Info("Restarting node has address {0}", Cluster.Get(restartingSystem).SelfUniqueAddress);
+                    Cluster.Get(restartingSystem).JoinSeedNodes(_seedNodes.Value);
+                    Within(TimeSpan.FromSeconds(20), () =>
+                    {
+                        AwaitAssert(() =>
+                        {
+                            Cluster.Get(restartingSystem).State.Members.Count.ShouldBe(totalNumberOfNodes);
+                            Cluster.Get(restartingSystem).State.Members.All(x => x.Status == MemberStatus.Up).ShouldBeTrue();
+                        });
+                    });
+                }, _config.Second);
+
+                EnterBarrier("joined-"+i);
+                Within(TimeSpan.FromSeconds(20), () =>
+                {
+                    AwaitAssert(() =>
+                    {
+                        Cluster.Get(Sys).State.Members.Count.ShouldBe(totalNumberOfNodes);
+                        Cluster.Get(Sys).State.Members.All(x => x.Status == MemberStatus.Up).ShouldBeTrue();
+                        // use the role to test that it is the new incarnation that joined, sneaky
+                        Cluster.Get(Sys).State.Members.SelectMany(x => x.Roles).ShouldBe(new [] { $"round-{round}" });
+                    });
+                });
+                EnterBarrier("members-up-"+i);
+
+                // gatting occurred after a while
+                if (i > 1)
+                    Thread.Sleep(ThreadLocalRandom.Current.Next(15) * 1000);
+
+                Cluster.Get(Sys).State.Members.Count.ShouldBe(totalNumberOfNodes);
+                Cluster.Get(Sys).State.Members.All(x => x.Status == MemberStatus.Up).ShouldBeTrue();
+                Cluster.Get(Sys).State.Unreachable.Count.ShouldBe(0);
+
+                EnterBarrier("before-terminate-"+i);
+                RunOn(() =>
+                {
+                    restartingSystem.Terminate().Wait(RemainingOrDefault).ShouldBeTrue("Timed out when terminating actorsystem");
+                }, _config.Second);
+
+                // don't wait for it to be removed, new incarnation will join in next round
+                EnterBarrier("terminated-"+i);
+                Log.Info("End of round-"+i);
+            }
+        }
+
+    }
+}
--- a/src/core/Akka.MultiNodeTestRunner.Shared/Sinks/MessageSink.cs
+++ b/src/core/Akka.MultiNodeTestRunner.Shared/Sinks/MessageSink.cs
@@ -102,10 +102,10 @@ namespace Akka.MultiNodeTestRunner.Shared.Sinks
         /*
          * Regular expressions - go big or go home. [Aaronontheweb]
          */
-        private const string NodeLogMessageRegexString = @"\[(\w){4}(?<node>[0-9]{1,2})\]\[(?<level>(\w)*)\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\](?<thread>\[(\w|\s)*\])\[(?<logsource>(\[|\w|:|/|\(|\)|\]|\.|-|\$|%|\+|\^|@)*)\]\s(?<message>(\w|\s|:|<|\.|\+|>|,|\[|/|-|]|%|\$|\+|\^|@|\(|\))*)";
+        private const string NodeLogMessageRegexString = @"\[(\w){4}(?<node>[0-9]{1,2})\]\[(?<level>(\w)*)\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\](?<thread>\[(\w|\s)*\])\[(?<logsource>(\[|\w|:|/|\(|\)|\]|\.|-|\$|%|\+|#|\^|@)*)\]\s(?<message>(\w|\s|:|<|\.|\+|>|,|\[|/|-|]|%|\$|\+|\^|@|\(|\))*)";
         protected static readonly Regex NodeLogMessageRegex = new Regex(NodeLogMessageRegexString);
 
-        private const string RunnerLogMessageRegexString = @"\[(?<level>(\w)*)\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\](?<thread>\[(\w|\s)*\])\[(?<logsource>(\[|\w|:|/|\(|\)|\]|\.|-|\$|%|\+|\^|@)*)\]\s(?<message>(\w|\s|:|<|\.|\+|>|,|\[|/|-|]|%|\$|\+|\^|@)*)";
+        private const string RunnerLogMessageRegexString = @"\[(?<level>(\w)*)\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\](?<thread>\[(\w|\s)*\])\[(?<logsource>(\[|\w|:|/|\(|\)|\]|\.|-|\$|%|\+|#|\^|@)*)\]\s(?<message>(\w|\s|:|<|\.|\+|>|,|\[|/|-|]|%|\$|\+|\^|@)*)";
         protected static readonly Regex RunnerLogMessageRegex = new Regex(RunnerLogMessageRegexString);
 
         private const string NodeLogFragmentRegexString = @"\[(\w){4}(?<node>[0-9]{1,2})\](?<message>(.)*)";
--- a/src/core/Akka.Remote.Tests.MultiNode/RemoteNodeRestartDeathWatchSpec.cs
+++ b/src/core/Akka.Remote.Tests.MultiNode/RemoteNodeRestartDeathWatchSpec.cs
@@ -65,8 +65,8 @@ namespace Akka.Remote.Tests.MultiNode
                     .GetAwaiter()
                     .GetResult();
                 TestConductor.Shutdown(_specConfig.Second).GetAwaiter().GetResult();
-                ExpectTerminated(subject, TimeSpan.FromSeconds(20));
-                Within(TimeSpan.FromSeconds(10), () =>
+                ExpectTerminated(subject, TimeSpan.FromSeconds(15));
+                Within(TimeSpan.FromSeconds(5), () =>
                 {
                     // retry because the Subject actor might not be started yet
                     AwaitAssert(() =>
--- a/src/core/Akka.Remote.Tests/ActorsLeakSpec.cs
+++ b/src/core/Akka.Remote.Tests/ActorsLeakSpec.cs
@@ -1,0 +1,223 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Akka.Actor;
+using Akka.Actor.Internal;
+using Akka.Configuration;
+using Akka.Remote.Transport;
+using Akka.TestKit;
+using Akka.TestKit.TestActors;
+using Akka.Util.Internal;
+using Xunit;
+
+namespace Akka.Remote.Tests
+{
+    public class ActorsLeakSpec : AkkaSpec
+    {
+        public static readonly Config Confg = ConfigurationFactory.ParseString(@"
+            akka.actor.provider = ""Akka.Remote.RemoteActorRefProvider, Akka.Remote""
+            akka.loglevel = DEBUG
+            akka.remote.helios.tcp.applied-adapters = [trttl]
+            akka.remote.helios.tcp.hostname = 127.0.0.1
+            akka.remote.log-lifecycle-events = on
+            akka.remote.transport-failure-detector.heartbeat-interval = 1 s
+            akka.remote.transport-failure-detector.acceptable-heartbeat-pause = 3 s
+            akka.remote.quarantine-after-silence = 3 s
+            akka.test.filter-leeway = 12 s
+        ");
+
+        public ActorsLeakSpec() : base(Confg)
+        {
+        }
+
+        private static ImmutableList<IActorRef> Recurse(IActorRef @ref)
+        {
+            var empty = new List<IActorRef>();
+            var list = empty;
+            if (@ref is ActorRefWithCell)
+            {
+                var cell = @ref.AsInstanceOf<ActorRefWithCell>().Underlying;
+                if (cell.ChildrenContainer is EmptyChildrenContainer ||
+                    cell.ChildrenContainer is TerminatedChildrenContainer ||
+                    cell.ChildrenContainer is TerminatingChildrenContainer) list = empty;
+                else list = cell.ChildrenContainer.Children.Cast<IActorRef>().ToList();
+            }
+
+            return ImmutableList<IActorRef>.Empty.Add(@ref).AddRange(list.SelectMany(Recurse));
+        }
+
+
+        private static ImmutableList<IActorRef> CollectLiveActors(IActorRef root)
+        {
+            return Recurse(root);
+        }
+
+        class StoppableActor : ReceiveActor
+        {
+            public StoppableActor()
+            {
+                Receive<string>(str => str.Equals("stop"), s =>
+                {
+                    Context.Stop(Self);
+                });
+            }
+        }
+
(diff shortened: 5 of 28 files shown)


Aaronontheweb:
This might still be a work in progress - waiting on CI to tell me. Strong chance this spec will depend on the actor shutdown mechanics in #2086 in order to run reliably.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -49,6 +49,7 @@ namespace Akka.Actor
         public const int UndefinedUid = 0;
         public ActorCell(Akka.Actor.Internal.ActorSystemImpl system, Akka.Actor.IInternalActorRef self, Akka.Actor.Props props, Akka.Dispatch.MessageDispatcher dispatcher, Akka.Actor.IInternalActorRef parent) { }
         protected Akka.Actor.ActorBase Actor { get; }
+        public Akka.Actor.Internal.IChildrenContainer ChildrenContainer { get; }
```
Aaronontheweb: Needed to make `ChildrenContainer` visible as part of the `ICell` interface; it's like this on the JVM, but also necessary for the `ActorsLeakSpec`

src/core/Akka.Cluster.Tests.MultiNode/QuickRestartSpec.cs
```diff
@@ -0,0 +1,140 @@
...
+using System.Threading;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Remote.TestKit;
+using Akka.TestKit;
+using Akka.Util;
+
+namespace Akka.Cluster.Tests.MultiNode
+{
+    public class QuickRestartSpecConfig : MultiNodeConfig
```
Aaronontheweb: Used to verify that a cluster node who very quickly became unreachable, restarted, and then joined again can successfully rejoin the cluster

src/core/Akka.Remote.Tests/ActorsLeakSpec.cs
```diff
@@ -0,0 +1,220 @@
...
+using Akka.Configuration;
+using Akka.Remote.Transport;
+using Akka.TestKit;
+using Akka.TestKit.TestActors;
+using Akka.Util.Internal;
+using Xunit;
+
+namespace Akka.Remote.Tests
+{
+    public class ActorsLeakSpec : AkkaSpec
```
Aaronontheweb: This spec is pretty genius - it maps the system remoting actor heirarchies and verifies that after all of the scenarios below that we haven't accidentally "leaked" any remoting actors; that they're always cleaned up properly after disassociations under all of these scenarios.

This spec revealed two bugs with the `ReliableDeliverySupervisor` and the `EndpointManager`

src/core/Akka.Remote/Endpoint.cs
```diff
@@ -500,9 +500,12 @@ protected void Gated(bool writerTerminated, bool earlyUngateRequested)
                 {
                     if (earlyUngateRequested)
                         Self.Tell(new Ungate());
+                    else
```
Aaronontheweb: `ReliableDeliverySupervisor` bugfix here - when we switch from a `Receiving` behavior to a `Gated` behavior (common, after an unplanned termination) we will wait for verification that the `EndpointWriter` (child of the RDS) is `Terminated` and then either immediately `Ungate` ourselves or schedule an `Ungate` to occur in the future - the default setting is 5 seconds.

In previous implementations of this code I mis-wrote this `if` statement so the `Ungate` operation was never scheduled, causing the node to be permanently gated. That would definitely cause rejoin issues. This fixes that.

src/core/Akka.Remote/EndpointManager.cs
```diff
@@ -404,7 +404,7 @@ protected override SupervisorStrategy SupervisorStrategy()
                     {
                         if (hopeless.Uid.HasValue)
                         {
-                            _log.Error("Association to [{0}] with UID [{1}] is irrecoverably failed. Quarantining address.",
+                            _log.Error(hopeless.InnerException, "Association to [{0}] with UID [{1}] is irrecoverably failed. Quarantining address.",
```
Aaronontheweb: We weren't properly propagating the internal exceptions thrown during `HopelessAssociation` events earlier; this fixes that.

src/core/Akka.Remote/EndpointManager.cs
```diff
@@ -564,42 +564,67 @@ protected void Accepting()
                 }
                 else if (policy.Item1 is Pass && policy.Item2 != null)
                 {
-                    var pass = policy.Item1 as Pass;
-                    if (pass.Uid == quarantine.Uid)
+                    var pass = (Pass) policy.Item1;
```
Aaronontheweb: This changes mirrors a bug report and subsequent fix Patrik Nordwall told us about: https://github.com/akka/akka/issues/20523

basically during `Quarantine` the `EndpointManager` would not properly account for the state of the quarantined UID and as a result, would leak `ReliableDeliverySupervisor` instances from previous incarnations of that association.

So now we explicitly update what the quarantined UID is inside the `EndpointRegistry` and check on `Quarantine` events to make sure that the policies are always being applied to the correct, current UID.

Additionally this fix cleans up the `GotUid` handler and will not correctly apply `Quarantined` policies if the node that joins has a UID that corresponds with the current `RefuseUid` value of the `Pass` policy.

Aaronontheweb:
A number of MNTR failures here as a result of nodes restarting or downing. Pretty sure that's because there's more Cluster updates that need to be made, such as https://github.com/akkadotnet/akka.net/pull/2091 - will keep working on this.

Aaronontheweb:
looks like all of the specs that failed are ones that involve restarting a cluster node - that helps me narrow it down a lot. I'm starting to look into that now.

Aaronontheweb:
#2091 won't have much impact on these failing tests - you can see that line still working (because the node gracefully terminated in a `MemberStatus.Up` state)

```
[INFO][6/14/2016 11:53:20 PM][Thread 0028][[akka://MultiNodeClusterSpec/system/cluster/core/daemon]] New incarnation of existing member [UniqueAddress: (akka.tcp://MultiNodeClusterSpec@localhost:49501, 1718532469)] is trying to join. Existing will be removed from the cluster and then new member will be allowed to join.
[INFO][6/14/2016 11:53:20 PM][Thread 0028][[akka://MultiNodeClusterSpec/system/cluster/core/daemon]] Marking node [akka.tcp://MultiNodeClusterSpec@localhost:49501] as Down
```

Aaronontheweb:
Looks like the issue is that the node is able to successfully connect, but can't complete the joining process on restart. Saw some issues with the Akka.Remote multi node tests too. So there's something still off in the UID system here. Going to drill into it.

alexvaluyskiy:
@Aaronontheweb look at this spec too, https://github.com/akkadotnet/akka.net/pull/2092
It has problems with UID system

Aaronontheweb:
Need to rebase this if it has any new cluster specs

src/core/Akka/Actor/Cell.cs
```diff
@@ -73,9 +73,11 @@ public interface ICell
         void Post(IActorRef sender, object message);
 
 
-
+        [Obsolete("Used ChildrenRefs instead")]
```
alexvaluyskiy: But you have removed ChildrenRefs...

Aaronontheweb: `ChildrenRefs` was `private` and was, itself, marked as `Obsolete`. I decided just to use the `ChildContainer` instead.

Aaronontheweb:
Getting closer - last two specs appear to be failing as a result of an Akka.Remote DeathWatch issue where UID is not being respected. Going to port https://github.com/akka/akka/blob/master/akka-remote/src/test/scala/akka/remote/RemoteDeathWatchSpec.scala since it has a case that tests explicitly for this.

alexvaluyskiy:
@Aaronontheweb you could do another PR with QuickRestartSpec only. And merge it after all my cluster PRs

Aaronontheweb:
@alexvaluyskiy I want to make sure that the changes I needed to introduce in order to get `QuickRestartSpec` to pass didn't break something else. Right now it looks like they might have exposed a bug with death watch.

Aaronontheweb:
Looks like the issue here is that the `ActorPath.Elements` needs to include the UID as the final element in the sequence: https://github.com/akka/akka/pull/1244, which we don't right now.

Since that gets consumed in so many different places where we're reconstructing actor references for remote actors inside the remoting system. Going to resolve that now.

Aaronontheweb:
Found a number of different UID issues littered throughout Akka core and Akka.Remote. Going to re-run the test suite now to see if I caught all of them. The specs that were failing in the midst of this change are now passing locally.

Aaronontheweb:
Will provide additional comments once CI passes

Aaronontheweb:
Looks like there are some changes that still need to be made with Pool routers and respecting UIDs... I'll work on that now

Aaronontheweb:
Figured it out - we weren't consistent in reading and writing the UID on the remote side of the remote deployment. Got that sorted and now the remote deployment specs appear to pass.

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/MessageSink.cs
```diff
@@ -102,10 +102,10 @@ public enum MultiNodeTestRunnerMessageType
         /*
          * Regular expressions - go big or go home. [Aaronontheweb]
          */
-        private const string NodeLogMessageRegexString = @"\[(\w){4}(?<node>[0-9]{1,2})\]\[(?<level>(\w)*)\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\](?<thread>\[(\w|\s)*\])\[(?<logsource>(\[|\w|:|/|\(|\)|\]|\.|-|\$|%|\+|\^|@)*)\]\s(?<message>(\w|\s|:|<|\.|\+|>|,|\[|/|-|]|%|\$|\+|\^|@|\(|\))*)";
+        private const string NodeLogMessageRegexString = @"\[(\w){4}(?<node>[0-9]{1,2})\]\[(?<level>(\w)*)\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\](?<thread>\[(\w|\s)*\])\[(?<logsource>(\[|\w|:|/|\(|\)|\]|\.|-|\$|%|\+|#|\^|@)*)\]\s(?<message>(\w|\s|:|<|\.|\+|>|,|\[|/|-|]|%|\$|\+|\^|@|\(|\))*)";
```
Aaronontheweb: Even though we don't use the Regex parsing system inside the MNTR much anymore, I fixed the `Regex` to account for the fact that the UID fragment always gets appended to the end of the `IActorRef` whenever its `ToString()` method is called now.

src/core/Akka.Remote.Tests.MultiNode/RemoteNodeRestartDeathWatchSpec.cs
```diff
@@ -65,8 +65,8 @@ public void Must_receive_terminated_when_remote_actor_system_is_restarted()
                     .GetAwaiter()
                     .GetResult();
                 TestConductor.Shutdown(_specConfig.Second).GetAwaiter().GetResult();
-                ExpectTerminated(subject, TimeSpan.FromSeconds(20));
-                Within(TimeSpan.FromSeconds(10), () =>
+                ExpectTerminated(subject, TimeSpan.FromSeconds(15));
+                Within(TimeSpan.FromSeconds(5), () =>
```
Aaronontheweb: Hardening

src/core/Akka.Remote.Tests/RemoteDeathWatchSpec.cs
```diff
@@ -0,0 +1,146 @@
...
+using Akka.Configuration;
+using Akka.Event;
+using Akka.TestKit;
+using Akka.TestKit.TestEvent;
+using Akka.Util;
+using Xunit;
+
+namespace Akka.Remote.Tests
+{
+    public class RemoteDeathWatchSpec : AkkaSpec
```
Aaronontheweb: Ported a missing spec to verify some remote `DeathWatch` scenarios that weren't previously covered by tests.

src/core/Akka.Remote.Tests/RemotingSpec.cs
```diff
@@ -161,12 +158,77 @@ public void Remoting_must_support_remote_lookups()
...
             //the remoting layer uses some "initialdelay task.delay" for 4 seconds.
             //so the token is cancelled before the delay completed.. 
-            var msg = await here.Ask<Tuple<string,IActorRef>>("ping", TimeSpan.FromSeconds(1.5));
+            var msg = await here.Ask<Tuple<string, IActorRef>>("ping", TimeSpan.FromSeconds(1.5));
             Assert.Equal("pong", msg.Item1);
             Assert.IsType<FutureActorRef>(msg.Item2);
         }
 
         [Fact]
+        public void Remoting_must_not_send_remote_recreated_actor_with_same_name()
```
Aaronontheweb: Guarantees that both remote `ActorSystem` instances respect the UIDs for actors discovered via remote messaging, and that old `IActorRef`s pointing to the dead actor now get their messages delivered to `DeadLetters`.

src/core/Akka.Remote.Tests/RemotingSpec.cs
```diff
@@ -161,12 +158,77 @@ public void Remoting_must_support_remote_lookups()
...
+
+            remoteSystem.ActorSelection("/user/otherEcho1").Tell(75);
+            ExpectMsg(75);
+
+            Sys.ActorSelection("akka.test://remote-sys@localhost:12346/user/otherEcho1").Tell(76);
+            ExpectMsg(76);
+        }
+
+        [Fact]
+        public void Remoting_must_lookup_actors_across_node_boundaries()
```
Aaronontheweb: Does a bunch of crazy cross-boundary remote deployments and verifies that our ability to resolve and select those actors still works back and forth over the network.

src/core/Akka.Remote/EndpointManager.cs
```diff
@@ -658,8 +679,29 @@ protected void Accepting()
             Receive<EndpointWriter.TookOver>(tookover => RemovePendingReader(tookover.Writer, tookover.ProtocolHandle));
             Receive<ReliableDeliverySupervisor.GotUid>(gotuid =>
             {
-                _endpoints.RegisterWritableEndpointUid(gotuid.RemoteAddress, gotuid.Uid);
-                HandleStashedInbound(Sender, writerIsIdle: false);
+                
+                var policy = _endpoints.WritableEndpointWithPolicyFor(gotuid.RemoteAddress);
```
Aaronontheweb: During a `GotUid` message we check to see if the endpoint we've just received a UID for has a quarantined policy or not. If it does, we avoid overwriting the quarantine and stay in a quarantined state. Otherwise we register a writeable endpoint as usual.

src/core/Akka.Remote/RemoteActorRefProvider.cs
```diff
@@ -294,10 +294,10 @@ internal IInternalActorRef ResolveActorRefWithLocalAddress(string path, Address
                     // HACK: needed to make ActorSelections work
                     if (actorPath.ToStringWithoutAddress().Equals("/"))
                         return RootGuardian;
-                    return _local.ResolveActorRef(RootGuardian, actorPath.Elements);
+                    return _local.ResolveActorRef(RootGuardian, actorPath.ElementsWithUid);
                 }
                     
-                return new RemoteActorRef(Transport, localAddress, new RootActorPath(actorPath.Address) / actorPath.Elements, ActorRefs.Nobody, Props.None, Deploy.None);
+                return new RemoteActorRef(Transport, localAddress, new RootActorPath(actorPath.Address) / actorPath.ElementsWithUid, ActorRefs.Nobody, Props.None, Deploy.None);
```
Aaronontheweb: All three of these `ActorPath.ElementsWithUid` calls preserve the UID of the actor whose reference we are trying to resolve on the very last element in the `IEnumerable<string>`. Needed to guarantee that UIDs are respected for actor references.

src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -217,15 +217,18 @@ private void HandleDaemonMsgCreate(DaemonMsgCreate message)
             ActorPath childPath;
             if(ActorPath.TryParse(message.Path, out childPath))
             {
-                IEnumerable<string> subPath = childPath.Elements.Drop(1); //drop the /remote
-                ActorPath path = Path/subPath;
+                IEnumerable<string> subPath = childPath.ElementsWithUid.Drop(1); //drop the /remote
+                ActorPath p = Path/subPath;
+                var s = subPath.Join("/");
+                var i = s.IndexOf("#", StringComparison.Ordinal);
+                var childName = i < 0 ? s : s.Substring(0, i); // extract the name without the UID
```
Aaronontheweb: Chops the remotely deployed actor's UID off from the name this actor will use inside the remote daemon's `ChildContainer`, since the UID isn't actually part of the name.

src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -291,28 +294,18 @@ public override IActorRef GetChild(IEnumerable<string> name)
 
         private IInternalActorRef GetChild(string name)
         {
+            var nameAndUid = ActorCell.SplitNameAndUid(name);
             IInternalActorRef child;
-            if (TryGetChild(name, out child))
+            if (TryGetChild(nameAndUid.Name, out child))
```
Aaronontheweb: Split the name and UID before we attempt to lookup the remotely deployed actors.

src/core/Akka.Remote/RemoteWatcher.cs
```diff
@@ -498,7 +498,7 @@ private void SendHeartbeat()
 
         private void TriggerFirstHeartbeat(Address address)
         {
-            if (WatchingNodes.Contains(address) && !_failureDetector.IsMonitoring(address))
+            if (WatcheeByNodes.ContainsKey(address) && !_failureDetector.IsMonitoring(address))
```
Aaronontheweb: `WatchingNodes` is an alias for the keys collection for `WatcheeByNodes`, so this code effectively does the same thing that it did before. Difference is that it reads exactly like the JVM's now.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -134,7 +134,7 @@ public T ExpectMsg<T>(T expected, Func<T, T, bool> comparer, TimeSpan? timeout =
         /// </summary>       
         public Terminated ExpectTerminated(IActorRef target, TimeSpan? timeout = null, string hint = null)
         {
-            var msg = string.Format("Terminated {0}. {1}", target.Path, hint ?? "");
+            var msg = string.Format("Terminated {0}. {1}", target, hint ?? "");
```
Aaronontheweb: Adjusted the failure message printed out when an `ExpectTerminated` call fails, so I can see the UID number.

src/core/Akka/Actor/ActorCell.DeathWatch.cs
```diff
@@ -87,8 +87,7 @@ public void TerminatedQueuedFor(IActorRef subject)
 
         private bool WatchingContains(IActorRef subject)
         {
-            return _state.ContainsWatching(subject) ||
-                   (subject.Path.Uid != UndefinedUid && _state.ContainsWatching(new UndefinedUidActorRef(subject)));
+            return _state.ContainsWatching(subject);
```
Aaronontheweb: `UndefinedUidActorRef` does not need to exist in Akka.NET, because we never had `ActorFor`. All of our actors always have UIDs.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -241,8 +240,12 @@ void IUntypedActorContext.Become(UntypedReceive receive, bool discardOld)
 
         private long NewUid()
         {
-            var auid = Interlocked.Increment(ref _uid);
-            return auid;
+            // Note that this uid is also used as hashCode in ActorRef, so be careful
+            // to not break hashing if you change the way uid is generated
+            var uid = ThreadLocalRandom.Current.Next();
```
Aaronontheweb: Replaced the incrementing counter with `ThreadLocalRandom` due to an easily reproducible issue where if a remote actor system restarts and creates its actors in the same order each time, the UIDs will always be the same. Observed this happening in some of the MNTR specifications I was debugging. Instead we'll use a random number identical to what the JVM does:https://github.com/akka/akka/blob/65a9e0a0f9372d2e2ecb945c8a9dae129a438f77/akka-actor/src/main/scala/akka/actor/ActorCell.scala#L350

Minus tail recursion.

src/core/Akka/Actor/ActorPath.cs
```diff
@@ -175,6 +175,31 @@ public IReadOnlyList<string> Elements
...
 
         /// <summary>
+        /// INTERNAL API.
+        /// 
+        /// Used in Akka.Remote - when resolving deserialized local actor references
+        /// we need to be able to include the UID at the tail end of the elements.
+        /// 
+        /// It's implemented in this class because we don't have an ActorPathExtractor equivalent.
+        /// </summary>
+        public IReadOnlyList<string> ElementsWithUid
```
Aaronontheweb: Since we don't have the concept of an `ActorPathExtractor`, I bolted this onto `ActorPath` directly. Works the same as `ActorPath.Elements` except that it bolts the UID on as a Uri fragment to the very last element. Needed for making all of the Akka.Remote UID guarantees workable.

src/core/Akka/Actor/ActorPath.cs
```diff
@@ -419,7 +445,7 @@ public string ToStringWithAddress()
 
         public string ToSerializationFormat()
         {
-            return ToStringWithAddress();
+            return AppendUidFragment(ToStringWithAddress());
```
Aaronontheweb: SerializationFormat must *always* include the UID fragment.

src/core/Akka/Actor/ActorRef.cs
```diff
@@ -186,7 +186,8 @@ public void Tell(object message, IActorRef sender)
 
         public override string ToString()
         {
-            return string.Format("[{0}]", Path);
+            if(Path.Uid == ActorCell.UndefinedUid) return $"[{Path}]";
+            return $"[{Path}#{Path.Uid}]";
```
Aaronontheweb: UIDs are supposed to be included in the `IActorRef.ToString()` output - so you can determine from the logs if an actor was recreated or not. https://github.com/akka/akka/blob/65a9e0a0f9372d2e2ecb945c8a9dae129a438f77/akka-actor/src/main/scala/akka/actor/ActorRef.scala#L159

src/core/Akka/Actor/ExtendedActorSystem.cs
```diff
@@ -27,6 +27,10 @@ public abstract class ExtendedActorSystem : ActorSystem
         /// </summary>
         public abstract IInternalActorRef Guardian { get; }
 
+        /// <summary>
+        /// The <see cref="RootGuardianActorRef"/>, used as the lookup for <see cref="IActorRef"/> resolutions.
+        /// </summary>
+        public abstract IInternalActorRef LookupRoot { get; }
```
Aaronontheweb: Needed to expose the `RootGuardianActorRef` on the `ActorSystemImpl` so it could be accessed from `ActorLookupSpecs`.

Aaronontheweb:
Alright, CI passes and I'm all finished with this. Left extensive comments - please review this; I'm pretty certain the fixes in this PR are responsible for some of the other MNTR test failures related to quarantines and possibly some of the router issues (improved detection of dead remotely deployed routees.)

src/core/Akka.Cluster.Tests.MultiNode/QuickRestartSpec.cs
```diff
@@ -0,0 +1,141 @@
...
+                    Thread.Sleep(ThreadLocalRandom.Current.Next(15) * 1000);
+
+                Cluster.Get(Sys).State.Members.Count.ShouldBe(totalNumberOfNodes);
+                Cluster.Get(Sys).State.Members.All(x => x.Status == MemberStatus.Up).ShouldBeTrue();
+                Cluster.Get(Sys).State.Unreachable.Count.ShouldBe(0);
+
+                EnterBarrier("before-terminate-"+i);
+                RunOn(() =>
+                {
+                    restartingSystem.Terminate().Wait(RemainingOrDefault).ShouldBeTrue("Timed out when terminating actorsystem");
```
alexvaluyskiy: JVM version doesn't have ShouldBeTrue check

src/core/Akka.Remote.Tests.MultiNode/RemoteNodeRestartDeathWatchSpec.cs
```diff
@@ -137,7 +137,7 @@ public RemoteNodeRestartDeathWatchSpecConfig()
             First = Role("first");
             Second = Role("second");
 
-            CommonConfig = DebugConfig(false).WithFallback(ConfigurationFactory.ParseString(
+            CommonConfig = DebugConfig(true).WithFallback(ConfigurationFactory.ParseString(
```
alexvaluyskiy: Why DebugConfig(true)?

src/core/Akka.Cluster.Tests.MultiNode/QuickRestartSpec.cs
maxim-s: looks like scala version has missed assert ```Cluster(system).state.members.map(_.status == MemberStatus.Up)``` https://github.com/akka/akka/blob/master/akka-cluster/src/multi-jvm/scala/akka/cluster/QuickRestartSpec.scala#L105

src/core/Akka.Remote.Tests.MultiNode/RemoteNodeRestartDeathWatchSpec.cs
Aaronontheweb: I can disable it now - this spec was failing earlier and I needed the additional logs to diagnose why. That's how I discovered the UID issue

src/core/Akka.Cluster.Tests.MultiNode/QuickRestartSpec.cs
Aaronontheweb: The `ShouldBeTrue` check is innocuous - if the actor system doesn't shut down in time then the test fails anyway. This just helped make it clearer *why* the test failed.

I'll add that missing assertion @maxim-s

Aaronontheweb: Hah, you're right @maxim-s - they forgot to put an assertion on that :p; somehow I thought you meant *we* forgot to add that.

src/core/Akka.Tests/Actor/ActorLookupSpec.cs
```diff
@@ -0,0 +1,179 @@
...
+        public static readonly Props P = Props.Create<Node>();
+
+        public class Node : ReceiveActor
+        {
+            private IActorRefProvider Provider { get { return Context.System.AsInstanceOf<ActorSystemImpl>().Provider; } }
+
+            public Node()
+            {
+                Receive<Create>(create => Sender.Tell(Context.ActorOf(P, create.Child)));
+                Receive<LookupElems>(elems => Sender.Tell(Provider.ResolveActorRef(new RootActorPath(Provider.DefaultAddress) / elems.Path)));
```
alexvaluyskiy: Maybe we should implement their outdated method - actorFor?

Aaronontheweb: Definitely not needed - they're trying to deprecate and discourage its use, plus it interferes with the UID system. Don't want to port anything that has no future if we can help it.

src/core/Akka.Tests/Actor/ActorLookupSpec.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        [Fact]
+        public void ActorSystem_must_find_actors_by_looking_up_their_path()
+        {
+            Provider.ResolveActorRef(c1.Path).Should().Be(c1);
+            Provider.ResolveActorRef(c2.Path).Should().Be(c2);
+            Provider.ResolveActorRef(c21.Path).Should().Be(c21);
+            Provider.ResolveActorRef(user.Path / "c1").Should().Be(c1);
+            Provider.ResolveActorRef(user.Path / "c2").Should().Be(c2);
+            Provider.ResolveActorRef(user.Path / "c2" / "c21").Should().Be(c21);
```
alexvaluyskiy: You missed these asserts
```
      system.actorFor(system child "c2" child "c21") should ===(c21) // test Java API
      system.actorFor(system / Seq("c2", "c21")) should ===(c21)

      import scala.collection.JavaConverters._
      system.actorFor(system descendant Seq("c2", "c21").asJava) // test Java API
```

src/core/Akka.Tests/Actor/ActorLookupSpec.cs
```diff
@@ -0,0 +1,179 @@
...
+            ExpectNoMsg(1.Seconds());
+
+            // not equal, because it's terminated
+            Provider.ResolveActorRef(a1.Path.ToString()).Should().NotBe(a1);
+
+            var a2 = Sys.ActorOf(P, name);
+            a2.Path.Should().Be(a1.Path);
+            a2.Path.ToString().Should().Be(a1.Path.ToString());
+            a2.Should().NotBe(a1);
+            a2.ToString().Should().NotBe(a1.ToString());
```
alexvaluyskiy: Missed
```
      watch(a2)
      a2 ! PoisonPill
      expectTerminated(a2)
```

src/core/Akka.Tests/Actor/ActorLookupSpec.cs
Aaronontheweb: I skipped those because `ResolveActorRef` has different semantics than `ActorFor` - `ActorSelection` handles this type of functionality now. `ResolveActorRef` only works for absolute lookups.

src/core/Akka.Tests/Actor/ActorLookupSpec.cs
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        [Fact]
+        public void ActorSystem_must_lookup_temporary_actors()
```
alexvaluyskiy: The name should be "ActorSystem_must_find_temporary_actor"

src/core/Akka.Remote.Tests/RemotingSpec.cs
```diff
@@ -20,10 +21,10 @@
 
 namespace Akka.Remote.Tests
 {
-    
+
     public class RemotingSpec : AkkaSpec
     {
-        public RemotingSpec():base(GetConfig())
+        public RemotingSpec() : base(GetConfig())
```
alexvaluyskiy: Can't review it properly. There is lack of specs. I think if we haven't ported or won't ported any spec (or part of spec) to Akka.NET we should write a comment about it. Right not I don't know - this spec is outdated or not

src/core/Akka.Remote.Tests/RemoteDeathWatchSpec.cs
```diff
@@ -0,0 +1,146 @@
...
+            akka {
+                actor {
+                    provider = ""Akka.Remote.RemoteActorRefProvider, Akka.Remote""
+                    deployment {
+                        /watchers.remote = ""akka.tcp://other@localhost:2666""
+                    }
+                }
+                remote.retry-gate-closed-for = 1 s
+                remote.initial-system-message-delivery-timeout = 3 s
+                remote.netty.tcp {
```
alexvaluyskiy: netty?

src/core/Akka.Remote.Tests/RemoteDeathWatchSpec.cs
```diff
@@ -0,0 +1,146 @@
...
+                }
+            }
+        ");
+
+        private readonly ActorSystem _other;
+
+        public RemoteDeathWatchSpec() : base(_config)
+        {
+            _other = ActorSystem.Create("other",
+                ConfigurationFactory.ParseString(@"akka.remote.netty.tcp.port=2666").WithFallback(_config));
```
alexvaluyskiy: netty again?

src/core/Akka.Remote.Tests/RemoteDeathWatchSpec.cs
```diff
@@ -0,0 +1,146 @@
...
+            var probe = CreateTestProbe();
+            Sys.EventStream.Subscribe(probe.Ref, typeof(QuarantinedEvent));
+            var rarp = RARP.For(Sys).Provider;
+            // pick an unused port (not going to use a socket address generator here; just a port not used by either actor system)
+            int port = rarp.DefaultAddress.Port.Value;
+            while (port == rarp.DefaultAddress.Port.Value || port == 2666)
+                port = ThreadLocalRandom.Current.Next(1, 65535);
+            // simulate de-serialized ActorRef
+            var @ref = rarp.ResolveActorRef($"akka.tcp://OtherSystem@localhost:{port}/user/foo/bar#1752527294");
+            Action<IActorDsl> act = dsl =>
```
alexvaluyskiy: If we use ActorDsl for tests, we should have tests for DSL. Right now we don't

Aaronontheweb: We do have DSL specs: https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Tests/Actor/ActorDslSpec.cs

alexvaluyskiy: Some of them are skipped

src/core/Akka.Remote.Tests.MultiNode/RemoteNodeShutdownAndComesBackSpec.cs
```diff
@@ -23,7 +23,7 @@ public RemoteNodeShutdownAndComesBackMultiNetSpec()
             First = Role("first");
             Second = Role("second");
 
-            CommonConfig = DebugConfig(false).WithFallback(ConfigurationFactory.ParseString(@"
+            CommonConfig = DebugConfig(true).WithFallback(ConfigurationFactory.ParseString(@"
```
alexvaluyskiy: DebugConfig(true)?

src/core/Akka.Cluster.Tests.MultiNode/QuickRestartSpec.cs
```diff
@@ -0,0 +1,141 @@
...
+                });
+                EnterBarrier("members-up-"+i);
+
+                // gatting occurred after a while
+                if (i > 1)
+                    Thread.Sleep(ThreadLocalRandom.Current.Next(15) * 1000);
+
+                Cluster.Get(Sys).State.Members.Count.ShouldBe(totalNumberOfNodes);
+                Cluster.Get(Sys).State.Members.All(x => x.Status == MemberStatus.Up).ShouldBeTrue();
+                Cluster.Get(Sys).State.Unreachable.Count.ShouldBe(0);
```
alexvaluyskiy: Use FluentAssertions everywhere or nowhere :)

src/core/Akka.Tests/Actor/ActorLookupSpec.cs
Aaronontheweb: Added

src/core/Akka.Cluster.Tests.MultiNode/QuickRestartSpec.cs
Aaronontheweb: Noted. Not going to do that on this PR though.

src/core/Akka.Remote.Tests/RemotingSpec.cs
alexvaluyskiy: Your comments about "why you have not implemented the specific spec" could help to future generations

alexvaluyskiy:
I created this ugly code in my previous commit
```
return (new RootActorPath(path.Root.Address
    .WithProtocol(SelfAddress.Protocol)
    .WithSystem(SelfAddress.System)
    .WithHost(SelfAddress.Host)
    .WithPort(SelfAddress.Port)) / string.Join("/", path.Elements)).WithUid(path.Uid);
```
Could you changes help me to improve it?

alexvaluyskiy:
Reviews this PR, looks good

Aaronontheweb:
@alexvaluyskiy I'll rebase on dev again so I can go through and update that as well. Which area is that in?

PR #2091, opened 2016-06-14
relaxed constraints for automatically downing previous incarnations of Cluster members

Aaronontheweb:
When a new incarnation of a node joins a Cluster, i.e. restarting after becoming `Unreachable`, the `Cluster` will automatically `Down` the previous incarnation of that node was `MemberStatus.Up`. 

This change relaxes that constraint so it doesn't matter what the previous `MemberStatus` was - so if a node ends up becoming unreachable and restarting during the middle of leaving the cluster, a new incarnation can still successfully join.

I've noticed this issue from time to time in our MNTR logs, but also saw that the JVM had an issue for it earlier this year: https://github.com/akka/akka/issues/19859


Danthar:
lgtm. Also part of a larger effort of fixing bugs.

Issue #1254, opened 2015-08-14
Port Akka.Remote MultiNodeSpec: RemoteQuarantinePiercingSpec


Aaronontheweb:
https://github.com/akka/akka/blob/master/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteQuarantinePiercingSpec.scala

PR #2092, opened 2016-06-15
Implemented RemoteQuarantinePiercingSpec

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/issues/1254

### DIFF
--- a/src/core/Akka.Remote.Tests.MultiNode/RemoteQuarantinePiercingSpec.cs
+++ b/src/core/Akka.Remote.Tests.MultiNode/RemoteQuarantinePiercingSpec.cs
@@ -1,0 +1,136 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="RemoteQuarantinePiercingSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+
+using System;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Remote.TestKit;
+using FluentAssertions;
+
+namespace Akka.Remote.Tests.MultiNode
+{
+    public class RemoteQuarantinePiercingSpecConfig : MultiNodeConfig
+    {
+        public RoleName First { get; }
+        public RoleName Second { get; }
+
+        public RemoteQuarantinePiercingSpecConfig()
+        {
+            First = Role("first");
+            Second = Role("second");
+
+            CommonConfig = DebugConfig(false)
+                .WithFallback(ConfigurationFactory.ParseString(@"
+                  akka.loglevel = INFO
+                  akka.remote.log-remote-lifecycle-events = INFO
+                "));
+        }
+
+        public sealed class Subject : ReceiveActor
+        {
+            public Subject()
+            {
+                Receive<string>(str => str == "shutdown", c => Context.System.Terminate());
+                Receive<string>(str => str == "identify", c =>
+                {
+                    Sender.Tell(Tuple.Create(AddressUidExtension.Uid(Context.System), Self));
+                });
+            }
+        }
+    }
+
+    public class RemoteQuarantinePiercingMultiNode1 : RemoteQuarantinePiercingSpec { }
+    public class RemoteQuarantinePiercingMultiNode2 : RemoteQuarantinePiercingSpec { }
+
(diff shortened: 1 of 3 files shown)


src/core/Akka.Remote.Tests.MultiNode/RemoteQuarantinePiercingSpec.cs
```diff
@@ -0,0 +1,133 @@
...
+using Akka.Remote.TestKit;
+using Akka.TestKit;
+
+namespace Akka.Remote.Tests.MultiNode
+{
+    public class RemoteQuarantinePiercingConfig : MultiNodeConfig
+    {
+        public RoleName First { get; }
+        public RoleName Second { get; }
+        public RoleName Third { get; }
```
maxim-s: unused role

src/core/Akka.Remote.Tests.MultiNode/RemoteQuarantinePiercingSpec.cs
```diff
@@ -0,0 +1,133 @@
+﻿using System;
```
maxim-s: missed copyrights

src/core/Akka.Remote.Tests.MultiNode/RemoteQuarantinePiercingSpec.cs
alexvaluyskiy: And no copyrights

src/core/Akka.Remote.Tests.MultiNode/RemoteQuarantinePiercingSpec.cs
```diff
@@ -0,0 +1,139 @@
...
+        public RemoteQuarantinePiercingConfig()
+        {
+            First = Role("first");
+            Second = Role("second");
+
+            CommonConfig = DebugConfig(false)
+                .WithFallback(ConfigurationFactory.ParseString(@"
+                  akka.loglevel = INFO
+                  akka.remote.log-remote-lifecycle-events = INFO
+                "));
```
Danthar: This does not do what you think it does. 
First the correct values for log-remote-lifecycle-events is either 'on' or 'off'. 
See: https://github.com/akkadotnet/akka.net/blob/master/src/core/Akka.Remote/Configuration/Remote.conf

Second. The WithFallback does exactly that. Fallback. It does _not_ override any existing values. Which means that if the DebugConfig(false) loads from base, it will include all the base values from pigeon.conf. 
This fallback definition is effectively doing nothing. 

And only works because the defaults are exactly the same. Also it does not crash on akka.remote.log-remote-lifecycle-events because the fallback value is never evaluated, since the default is loaded.

src/core/Akka.Remote.Tests.MultiNode/RemoteQuarantinePiercingSpec.cs
```diff
@@ -0,0 +1,139 @@
...
+                EnterBarrier("actor-identified");
+                Sys.WhenTerminated.Wait(TimeSpan.FromSeconds(30));
+
+                Log.Error($"akka.remote.helios.tcp.hostname = {addr.Host}");
+                Log.Error($"akka.remote.helios.tcp.port = {addr.Port.Value}");
+
+                var freshSystem = ActorSystem.Create(Sys.Name, ConfigurationFactory.ParseString($@"
+                    akka.remote.helios.tcp.hostname = {addr.Host}
+                    akka.remote.helios.tcp.port = {addr.Port.Value}
+                ").WithFallback(Sys.Settings.Config));
```
Danthar: And this is the correct usage of the config system. (see my previous remark)

src/core/Akka.Remote.Tests.MultiNode/RemoteQuarantinePiercingSpec.cs
alexvaluyskiy: We have a lot of examples here in the code, when we are suing INFO/on/off with this setting

alexvaluyskiy:
Strange, this spec worked before I rebased it to the newest dev

Aaronontheweb:
@alexvaluyskiy other than the changes that were just pulled in about an hour ago, there's nothing in the `dev` branch I can think of that would have caused a big change here. But based on your comment in #2090 I think we've found some EndpointManager issues.

alexvaluyskiy:
Still fails, on the newest sources. Waiting for some UID fix

Aaronontheweb:
@alexvaluyskiy I got blocked on some of the dispatcher stuff I was working on (TL;DR; more yak-shaving is required to do what i wanted to do; skipping for now.) I'm back to work on some of the UID issues.

Aaronontheweb:
Woohoo!

PR #2103, opened 2016-06-21
Rebased Router Update

Aaronontheweb:


Aaronontheweb:
I need to work on other changes in parallel locally - decided to hand this over to the build server and let it run while I work on that.

Aaronontheweb:
Persistent failure with 

```
[17:14:43][Step 2/2] [RUNNER][5:14 PM]: <----------- BEGIN NODE 1 ----------->
[17:14:43][Step 2/2] [RUNNER][5:14 PM]:  --> Akka.Cluster.Tools.Tests.MultiNode.Singleton.ClusterSingletonManagerStartupNode1.ClusterSingletonManagerStartupSpecs
[17:14:43][Step 2/2]  FAIL
[17:14:43][Step 2/2] [RUNNER][5:14 PM]: <----------- END NODE 1 ----------->
[17:14:43][Step 2/2] [RUNNER][5:14 PM]: <----------- BEGIN NODE 2 ----------->
[17:14:43][Step 2/2] [RUNNER][5:14 PM]:  --> Akka.Cluster.Tools.Tests.MultiNode.Singleton.ClusterSingletonManagerStartupNode2.ClusterSingletonManagerStartupSpecs
[17:14:43][Step 2/2]  FAIL
[17:14:43][Step 2/2] [RUNNER][5:14 PM]: <----------- END NODE 2 ----------->
[17:14:43][Step 2/2] [RUNNER][5:14 PM]: <----------- BEGIN NODE 3 ----------->
[17:14:43][Step 2/2] [RUNNER][5:14 PM]:  --> Akka.Cluster.Tools.Tests.MultiNode.Singleton.ClusterSingletonManagerStartupNode3.ClusterSingletonManagerStartupSpecs
[17:14:43][Step 2/2]  FAIL
[17:14:43][Step 2/2] [RUNNER][5:14 PM]: <----------- END NODE 3 ----------->
```

src/core/Akka.Cluster.Tests.MultiNode/Routing/ClusterConsistentHashingRouterSpec.cs
```diff
@@ -129,17 +116,19 @@ protected void AssertHashMapping(IActorRef router)
...
         {
             A_cluster_router_with_consistent_hashing_pool_must_start_cluster_with2_nodes();
-            A_cluster_router_with_consistent_hashing_pool_must_create_routees_from_configuration();
-            A_cluster_router_with_consistent_hashing_pool_must_select_destination_based_on_hash_key();
-            A_cluster_router_with_consistent_hashing_pool_must_deploy_routees_to_new_member_nodes_in_the_cluster();
-            A_cluster_router_with_consistent_hashing_pool_must_deploy_programatically_defined_routees_to_the_member_nodes_in_the_cluster();
-            A_cluster_router_with_consistent_hashing_pool_must_handle_combination_of_configured_router_and_programatically_defined_hash_mapping();
-            A_cluster_router_with_consistent_hashing_pool_must_handle_combination_of_configured_router_and_programatically_defined_hash_mapping_and_cluster_config();
-            A_cluster_router_with_consistent_hashing_pool_must_remove_routees_from_downed_node();
+            // A_cluster_router_with_consistent_hashing_pool_must_create_routees_from_configuration();
```
Aaronontheweb: Why are all of these commented out?

alexvaluyskiy: Because they dont work

Aaronontheweb:
Going to work on the router MNTR specs

Aaronontheweb:
All set - ready for review

src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
```diff
@@ -145,7 +145,7 @@ public void CanSerializeProps()
         [Fact]
         public void CanSerializeDeploy()
         {
-            var message = new Deploy(RouterConfig.NoRouter).WithMailbox("abc");
+            var message = new Deploy(NoRouter.Instance).WithMailbox("abc");
```
alexvaluyskiy: We should write about this breaking change in Release Notes

src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
```diff
@@ -230,7 +230,7 @@ public void CanSerializeTailChoppingPool()
 
             var supervisor = new OneForOneStrategy(decider);
 
-            var message = new TailChoppingPool(10, null, supervisor, "abc", TimeSpan.FromSeconds(10), TimeSpan.FromSeconds(2));
+            var message = new TailChoppingPool(10, null, TimeSpan.FromSeconds(10), TimeSpan.FromSeconds(2), supervisor, "abc");
```
alexvaluyskiy: I've changes parameters' ordering to align with Scala version. Maybe I was wrong

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
```diff
@@ -120,75 +143,133 @@ protected static int GetMaxTotalNrOfInstances(Config config)
             }
             else
             {
-                return number;
+                return number; ;
```
alexvaluyskiy: extra symbol

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
```diff
@@ -120,75 +143,133 @@ protected static int GetMaxTotalNrOfInstances(Config config)
...
         }
 
-        public override RouterConfig WithFallback(RouterConfig routerConfig)
+        public override ISurrogate ToSurrogate(ActorSystem system)
         {
-            var localFallback = routerConfig as ClusterRouterGroup;
-            if (localFallback != null && (localFallback.Local is ClusterRouterGroup)) throw new ConfigurationException("ClusterRouterGroup is not allowed to wrap a ClusterRouterGroup");
-            if (localFallback != null) return Copy(Local.WithFallback(localFallback.Local).AsInstanceOf<Group>());
-            return Copy(Local.WithFallback(routerConfig).AsInstanceOf<Group>());
+            throw new NotImplementedException();
```
alexvaluyskiy: We have unimplemented ToSurrogate method. Do we really need surrogates for routers?

src/core/Akka/Actor/SupervisorStrategy.cs
```diff
@@ -197,7 +197,7 @@ private void Publish(IActorContext context, LogEvent logEvent)
     /// <summary>
     ///     Class OneForOneStrategy. This class cannot be inherited.
     /// </summary>
-    public class OneForOneStrategy : SupervisorStrategy
+    public class OneForOneStrategy : SupervisorStrategy, IEquatable<OneForOneStrategy>
```
alexvaluyskiy: Do we really need for IEquatable here? Maybe just Equals would be good?

src/core/Akka/Routing/Broadcast.cs
```diff
@@ -35,69 +37,34 @@ public override Routee Select(object message, Routee[] routees)
     /// <summary>
     /// This class represents a <see cref="Pool"/> router that sends messages it receives to all of its <see cref="Routee">routees</see>.
     /// </summary>
-    public class BroadcastPool : Pool
+    public sealed class BroadcastPool : Pool
```
alexvaluyskiy: Made all routers classes - sealed, as scala version did

src/core/Akka/Routing/Broadcast.cs
```diff
@@ -185,60 +145,116 @@ public override RouterConfig WithFallback(RouterConfig routerConfig)
...
-    /// This class represents a <see cref="Group"/> router that sends messages it receives to all of its routees.
-    /// </summary>
-    public class BroadcastGroup : Group
-    {
-        /// <summary>
-        /// This class represents a surrogate of a <see cref="BroadcastGroup"/> router.
-        /// Its main use is to help during the serialization process.
-        /// </summary>
-        public class BroadcastGroupSurrogate : ISurrogate
+        private RouterConfig OverrideUnsetConfig(RouterConfig other)
```
alexvaluyskiy: We have this method in each pool router. I can't move it to the base class, as it was before. Because in that case we will have api inconsistency with Scala version

alexvaluyskiy:
@Aaronontheweb Can't review all PR, because it is my own changes :)

src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
Aaronontheweb: Is this breaking change necessary? If so, why?

src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
Aaronontheweb: We should probably change it back - better off having backwards compat for current users

src/core/Akka.Cluster.Tests/Akka.Cluster.Tests.csproj
```diff
@@ -42,6 +42,10 @@
       <HintPath>..\..\packages\FluentAssertions.3.3.0\lib\net45\FluentAssertions.Core.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="FSharp.Core, Version=4.3.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
```
Aaronontheweb: I need to remove this - was going to use FsCheck but then decided against it

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -604,7 +604,7 @@ public ClusterDaemon(ClusterSettings settings)
             Receive<InternalClusterAction.PublisherCreated>(msg =>
             {
                 if (_settings.MetricsEnabled)
-                    Context.ActorOf(Props.Create<ClusterMetricsCollector>().WithDispatcher(_settings.UseDispatcher), "metrics");
+                    Context.ActorOf(Props.Create(() => new ClusterMetricsCollector(msg.Publisher)).WithDispatcher(_settings.UseDispatcher), "metrics");
```
Aaronontheweb: This was incorrect before and would fail to start to ClusterMetricsCollector

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
Aaronontheweb: If we didn't before then probably not

src/core/Akka/Actor/SupervisorStrategy.cs
Aaronontheweb: Equals would be fine

src/core/Akka/Routing/Broadcast.cs
Aaronontheweb: Agree

src/core/Akka/Routing/Broadcast.cs
Aaronontheweb: I think you resolved this issue by making sure the underlying fluent methods on RouterConfig all return the correct implementation class now, whereas this did some of that before. Plus we're using null instead of NoRouter - so I think this is good.

Aaronontheweb: Err, actually not sure about bull - commenting on my mobile :p

src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
alexvaluyskiy: To align with scala version

src/core/Akka.Cluster/ClusterDaemon.cs
alexvaluyskiy: Yeap. I saw an exceptions in mntk specs logs

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
alexvaluyskiy: But we have this interface in RouterConfig class. Why we need ISurrogated interface in routers?

src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
Aaronontheweb: I think we can have our cake and eat it too here. Let me try something out.

src/core/Akka.Remote/Routing/RemoteRouterConfig.cs
```diff
@@ -20,114 +20,87 @@ namespace Akka.Remote.Routing
     /// routees on defined target nodes. Delegates other duties to the local <see cref="Pool"/>,
     /// which makes it possible to mix this with built-in routers such as <see cref="RoundRobinGroup"/> or custom routers.
     /// </summary>
-    public sealed class RemoteRouterConfig : Pool 
+    public sealed class RemoteRouterConfig : Pool, IEquatable<RouterConfig>
     {
-        public class RemoteRouterConfigSurrogate : ISurrogate
```
Aaronontheweb: We may need to add this back in order to maintain serialization support for people who are remotely deploying routers.

Aaronontheweb: Derp, my bad - this code is still here. Just moved further down in the file.

src/core/Akka/Routing/RouterConfig.cs
```diff
@@ -20,227 +20,155 @@ namespace Akka.Routing
...
-        /// <param name="message">The message to inspect.</param>
-        /// <returns><c>true</c> if this message is handled by the router; otherwise <c>false</c>.</returns>
-        public virtual bool IsManagementMessage(object message)
+        public virtual Props RoutingLogicController(RoutingLogic routingLogic)
         {
-            return
-                message is IAutoReceivedMessage ||
-                // in akka.net this message is a subclass of AutoReceivedMessage - so removed condition that "message is Terminated ||"
-                message is RouterManagementMessage;
+            return null;
```
Aaronontheweb: Looks good - and I'm looking forward to seeing this feature used!

src/core/Akka/Actor/SupervisorStrategy.cs
Aaronontheweb: Decided to leave the `IEquatable` in there for now.

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
Aaronontheweb: So an update on this: `ISurrogate` is used for serialization of built-in Akka.NET primitives. Not needed for clustered routers because they're not meant to be remotely deployed onto other nodes.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -4235,39 +4250,52 @@ namespace Akka.Routing
...
+    public sealed class TailChopping : Akka.Routing.RoutingLogic
+    {
+        public TailChopping(Akka.Actor.IScheduler scheduler, System.TimeSpan within, System.TimeSpan interval) { }
+        public override Akka.Routing.Routee Select(object message, Akka.Routing.Routee[] routees) { }
+    }
     public sealed class TailChoppingGroup : Akka.Routing.Group
     {
         public TailChoppingGroup(Akka.Configuration.Config config) { }
-        public TailChoppingGroup(string[] routeePaths, System.TimeSpan within, System.TimeSpan interval) { }
+        public TailChoppingGroup(System.Collections.Generic.IEnumerable<string> routeePaths, System.TimeSpan within, System.TimeSpan interval) { }
```
Aaronontheweb: Even though this is a public signature change, it's backwards compatible with what we had before. I'm fine with this.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -4235,39 +4250,52 @@ namespace Akka.Routing
...
             public System.TimeSpan Within { get; set; }
             public Akka.Util.ISurrogated FromSurrogate(Akka.Actor.ActorSystem system) { }
         }
     }
     public sealed class TailChoppingPool : Akka.Routing.Pool
     {
-        public TailChoppingPool(int nrOfInstances, Akka.Routing.Resizer resizer, Akka.Actor.SupervisorStrategy supervisorStrategy, string routerDispatcher, System.TimeSpan within, System.TimeSpan interval, bool usePoolDispatcher = False) { }
         public TailChoppingPool(Akka.Configuration.Config config) { }
         public TailChoppingPool(int nrOfInstances, System.TimeSpan within, System.TimeSpan interval) { }
+        public TailChoppingPool(int nrOfInstances, Akka.Routing.Resizer resizer, System.TimeSpan within, System.TimeSpan interval, Akka.Actor.SupervisorStrategy supervisorStrategy, string routerDispatcher, bool usePoolDispatcher = False) { }
```
Aaronontheweb: Reverting this back to its original signature. Backwards compat trumps 100% alignment with JVM.

src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
Aaronontheweb: Came up with a compromise - added `RouterConfig.NoRouter` back, but it's implemented like this:

```csharp
public static RouterConfig NoRouter => NoRouter.Instance;
```

I also marked that API as obsolete and told users to start using `NoRouter.Instance` instead.

Aaronontheweb:
MNTR failure due to https://github.com/akkadotnet/akka.net/pull/2118

alexvaluyskiy:
@Aaronontheweb reviewed your changes. Looks good.

@akkadotnet/developers could anyone made quick-review of these changes and merge?

mrrd:
Hi,

Is this pull request included in any kind of release nuget package yet, even a pre-release version?

As this is currently preventing us from being able to make use of clustering currently due to a problem that is happening the same as this https://github.com/petabridge/akkadotnet-code-samples/issues/31 which I believe this is supposed to fix.

alexvaluyskiy:
@mrrd yes, it was released in 1.1.0

mrrd:
@alexvaluyskiy really? as it still doesn't seem to work even on 1.1.1 in my case.

mrrd:
@alexvaluyskiy in my scenario I have two nodes, both joining the cluster with both supporting the same role, both have an actor at the address the broadcast-group is looking for but when a separate actor on role2 sends a message to the broadcast-group address only the actor on role1 (the first role started) is ever getting the message, and as a part of the message handling on both I am using GetRoutees to get the members in the group just as the example but only one is being returned.

This is I gather the same sort of behaviour as was observed from the sample code by the original reporter.

Hence my request if this fix which is apparently supposed to resolve this is actually available yet.

Aaronontheweb:
@mrrd do you have allow-local-routees turned on? https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Cluster/Configuration/Cluster.conf#L204

mrrd:
@Aaronontheweb I do yes the deployment section is as follows:

              /state/broadcaster {
                router = broadcast-group
                routees.paths = ["/user/state"]
                cluster {
                  enabled = on
                  max-nr-of-instances-per-node = 1
                  allow-local-routees = on
                  use-role = exportstate
                }
              }

Aaronontheweb:
@mrrd got it - I can confirm that the issue with WebCrawler was fixed by this PR. Verified that before we shipped 1.1. We also verify this with a multi-node spec that runs a full cluster on each pull request. Your issue is likely something else; just from reading your comment above it's not clear to me how many roles you have and how many nodes of each role have joined in the cluster.

mrrd:
Hi @Aaronontheweb, I am performing this with two nodes until I am convinced things are working as they should be, the nodes both join the cluster ok as far as I can discern from viewing the debug logging in place and both also contain practically identical code in the actor that is in each with the same identical address in each of "/user/state", they also contain the same configuration section in deployment as posted above, I am really scratching my head as to what may be at fault as the code itself is very similar to the examples, I would really appreciate it if you could point me to where I should be looking next.

When a message is send by a third entirely separate actor to the address of the router using ActorSelection only the first actor (i.e. the role that started first) is ever actually receiving it.

mrrd:
@Aaronontheweb is there any config flags I can enable to allow more insight into what the clustered routers are doing as I currently have the following sections in the hcon of both nodes:

`
            debug {
              receive = on      # log any received message
              autoreceive = on  # log automatically received messages, e.g. PoisonPill
              lifecycle = on    # log actor lifecycle changes
              event-stream = on # log subscription changes for Akka.NET event stream
              unhandled = on    # log unhandled messages sent to actors
            }
            `

and the following as well (obviously with different port numbers for each node as they are both currently running locally to test it)

`remote {
            log-remote-lifecycle-events = DEBUG
            log-received-messages = on
            
            helios.tcp {
              hostname = localhost
              port = 8090
            }
          }`

I ask this because the only entry I have in the log is the following on the first node to indicate the router has sent the message to the actor on the first node sent by the third actor running on the second node which gives no real indication as to what has gone on in between as preceding and following this message are just heartbeat gossip messages.

`2016-08-18 15:21:41.8158|DEBUG|Akka.Remote.EndpointWriter|received local message [RemoteMessage: ActorSelectionMessage - Message: DataExport.Shared.Messages.GetExportState - WildCartFanOut: False - Elements: user/state to [akka://DataExport/]<+akka://DataExport/ from [akka.tcp://DataExport@localhost:8091/user/portal#1249539263]]`

The second node is definitely joining and being marked as up as far as I can see by the following messages:

`2016-08-18 15:21:37.4631|INFO|Akka.Cluster.ClusterCoreDaemon|Node [akka.tcp://DataExport@localhost:8091] is JOINING, roles [exportstate]`

`2016-08-18 15:21:38.5653|INFO|Akka.Cluster.ClusterCoreDaemon|Leader is moving node [akka.tcp://DataExport@localhost:8091] to [Up]`

mrrd:
@Aaronontheweb where are the broadcast-group cluster tests you mention located as browsing the repository I cannot seem to find one, only **ClusterRoundRobinSpec**, **ClusterConsistentHashingGroupSpec** and **ClusterConsistentHashingRouterSpec** ?

I am trying to write my own following the documentation for testing a clustered broadcast group and unfortunately it is failing and the command line runner doesn't really provide enough insight as to why that maybe.

I have started with something similar to ClusterRoundRobinSpec but it is failing at the first hurdle of the following code block with nothing else other than the configuration which is also very similar.

private void A_cluster_router_with_a_BroadcastGroup_router_must_start_cluster_with_2_nodes() {
            AwaitClusterUp(_config.First, _config.Second);
            EnterBarrier("after-1");
        }

mrrd:
@Aaronontheweb I have just worked out why by test was not passing, when downloading the Akka source I noticed it is .NET 4.5, I am using .NET 4.6.1 in my project, it seems if the tests are inside a 4.6.1 class library they immediately fail, and if that class library is changed to .NET 4.5 it passes, is Akka restricted to working on .NET 4.5 only? Could that be the whole reason for the other problems I am experiencing?

Issue #1255, opened 2015-08-14
Port Akka.Remote MultiNodeSpec: RemoteReDeploymentSpec


Aaronontheweb:
https://github.com/akka/akka/blob/master/akka-remote-tests/src/multi-jvm/scala/akka/remote/RemoteReDeploymentSpec.scala

alexvaluyskiy:
@maxim-s could you make PR?

maxim-s:
@alexvaluyskiy sure, but node2 still failed, I don't want to do failed PR

alexvaluyskiy:
Closed https://github.com/akkadotnet/akka.net/pull/2121

PR #2110, opened 2016-06-22
added RemoteReDeploymentSpec #1255

maxim-s:


Aaronontheweb:
@maxim-s might need to follow these instructions to update the timestamp so CI can pick this up http://stackoverflow.com/a/31540373/377476

Aaronontheweb:
Think I've found the issue here.... `System.Abort` needs to be implemented for this spec to work properly; this is the first test that's used that code path at all. Looking into it now.

Aaronontheweb:
Replaced by https://github.com/akkadotnet/akka.net/pull/2121

PR #2116, opened 2016-06-25
Gating should not overrwite quarantine

Aaronontheweb:
Implements https://github.com/akka/akka/pull/20594


Aaronontheweb:
This is ready to be reviewed. Name says it all - prevents a `Gated` policy from overwriting a `Quarantine` policy.

alexvaluyskiy:
@Aaronontheweb Looks good.
Have you tested any RemotingSpec with these changes?

Aaronontheweb:
@alexvaluyskiy yep - had to modify some tests inside the `EndpointRegistrySpec` but other than that it's pretty kosher.

This is a bit of a rare edge case but as you can see on the original Akka thread, not impossible.

PR #1666, opened 2016-01-23
#1252 added RemoteNodeRestartGateSpec

maxim-s:

PR #2120, opened 2016-06-27
1252 remote node restart gate spec

Aaronontheweb:
replaces https://github.com/akkadotnet/akka.net/pull/1666 - fixed some minor issues with the spec which allow it to pass.

### DIFF
--- a/src/core/Akka.Remote.Tests.MultiNode/RemoteNodeRestartGateSpec.cs
+++ b/src/core/Akka.Remote.Tests.MultiNode/RemoteNodeRestartGateSpec.cs
@@ -1,0 +1,177 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="RemoteNodeRestartGateSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Text;
+using System.Text.RegularExpressions;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Remote.TestKit;
+using Akka.Remote.Transport;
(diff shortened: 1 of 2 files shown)


src/core/Akka.Remote.Tests.MultiNode/RemoteNodeRestartGateSpec.cs
```diff
@@ -0,0 +1,178 @@
...
+                    {
+                        freshSystem.ActorSelection(new RootActorPath(firstAddress)/"user"/"subject")
+                            .Tell(new Identify("subject"), probe);
+                        probe.ExpectMsg<ActorIdentity>();
+                    });
+                });
+
+                // Now the other system will be able to pass, too
+                freshSystem.ActorOf(Props.Create(() => new Subject()), "subject");
+                freshSystem.WhenTerminated.Wait(TimeSpan.FromSeconds(30));
```
Aaronontheweb: this line needed to be changed from using `Sys` to using `freshSystem` - otherwise node 2 would immediately bail out.

src/core/Akka.Remote.Tests.MultiNode/RemoteNodeRestartGateSpec.cs
```diff
@@ -0,0 +1,178 @@
...
+                EnterBarrier("gated");
+                TestConductor.Shutdown(_specConfig.Second).Wait();
+                Within(TimeSpan.FromSeconds(10), () =>
+                    {
+                        AwaitAssert(
+                            () =>
+                                {
+                                    Sys.ActorSelection(new RootActorPath(secondAddress) / "user" / "subject")
+                                        .Tell(new Identify("subject"));
+                                    ExpectMsg<ActorIdentity>().Subject.ShouldNotBe(null);
```
Aaronontheweb: This line needed to be changed to perform an assertion on `ActorIdentity.Subject` to make sure that it was non-null, otherwise node 1 would bail out early.

Aaronontheweb:
@maxim-s preserved your original commit, BTW

maxim-s:
:+1:

PR #2121, opened 2016-06-28
#1255 Remote redeployment spec

Aaronontheweb:
Fixed some bugs with the Akka.Remote.TestKit that caused this spec not to pass. Also implemented `ActorSystemImpl.Abort`.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -651,6 +651,7 @@ namespace Akka.Actor
         public abstract Akka.Actor.IInternalActorRef LookupRoot { get; }
         public abstract Akka.Actor.IActorRefProvider Provider { get; }
         public abstract Akka.Actor.IInternalActorRef SystemGuardian { get; }
+        public abstract void Abort();
         public abstract Akka.Actor.IActorRef SystemActorOf(Akka.Actor.Props props, string name = null);
         public abstract Akka.Actor.IActorRef SystemActorOf<TActor>(string name = null)
             where TActor : Akka.Actor.ActorBase, new ();
@@ -1767,6 +1768,7 @@ namespace Akka.Actor.Internal
         [System.ObsoleteAttribute("Use WhenTerminated instead. This property will be removed in future versions")]
         public override System.Threading.Tasks.Task TerminationTask { get; }
         public override System.Threading.Tasks.Task WhenTerminated { get; }
+        public override void Abort() { }
         public override Akka.Actor.IActorRef ActorOf(Akka.Actor.Props props, string name = null) { }
         public override Akka.Actor.ActorSelection ActorSelection(Akka.Actor.ActorPath actorPath) { }
         public override Akka.Actor.ActorSelection ActorSelection(string actorPath) { }
--- a/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
+++ b/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
@@ -588,7 +588,7 @@ namespace Akka.Remote.TestKit
 
         protected void InjectDeployments(ActorSystem system, RoleName role)
         {
-            var deployer = Sys.AsInstanceOf<ExtendedActorSystem>().Provider.Deployer;
+            var deployer = system.AsInstanceOf<ExtendedActorSystem>().Provider.Deployer;
             foreach (var str in _deployments(role))
             {
                 var deployString = _replacements.Values.Aggregate(str, (@base, r) =>
@@ -631,7 +631,7 @@ namespace Akka.Remote.TestKit
         protected ActorSystem StartNewSystem()
         {
             var sb =
-                new StringBuilder("helios.tcp{").AppendLine()
+                new StringBuilder("akka.remote.helios.tcp{").AppendLine()
                     .AppendFormat("port={0}", _myAddress.Port)
                     .AppendLine()
                     .AppendFormat(@"hostname=""{0}""", _myAddress.Host)
--- a/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
+++ b/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
@@ -12,6 +12,8 @@ using System.Diagnostics;
 using System.Linq;
 using System.Net;
 using System.Net.Sockets;
+using System.Text;
+
 using Akka.Actor;
 using Akka.Configuration;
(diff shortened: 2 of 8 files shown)


Aaronontheweb:
Preserved your original commit @maxim-s

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -651,6 +651,7 @@ namespace Akka.Actor
         public abstract Akka.Actor.IInternalActorRef LookupRoot { get; }
         public abstract Akka.Actor.IActorRefProvider Provider { get; }
         public abstract Akka.Actor.IInternalActorRef SystemGuardian { get; }
+        public abstract void Abort();
```
Aaronontheweb: Made `Abort` a public member, although a user will have to know to cast an `ActorSystem` to `ActorSystemImpl` to get access to this method.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1767,6 +1768,7 @@ namespace Akka.Actor.Internal
         [System.ObsoleteAttribute("Use WhenTerminated instead. This property will be removed in future versions")]
         public override System.Threading.Tasks.Task TerminationTask { get; }
         public override System.Threading.Tasks.Task WhenTerminated { get; }
+        public override void Abort() { }
```
Aaronontheweb: Concrete implementation of `ExtendedActorSystem.Abort` lives on the `ActorSystemImpl` class, which is why this shows up twice in the approval file.

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -586,7 +588,7 @@ public Replacement(string tag, RoleName role, MultiNodeSpec spec)
 
         protected void InjectDeployments(ActorSystem system, RoleName role)
         {
-            var deployer = Sys.AsInstanceOf<ExtendedActorSystem>().Provider.Deployer;
+            var deployer = system.AsInstanceOf<ExtendedActorSystem>().Provider.Deployer;
```
Aaronontheweb: Nasty bug - needed to use the instance variable rather than the property `Sys` to work, otherwise we could never successfully call `MultiNodeSpec.NewActorSystem()` and get a correct result if we used any programmatic deployments.

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -628,11 +630,15 @@ protected void InjectDeployments(ActorSystem system, RoleName role)
 
         protected ActorSystem StartNewSystem()
         {
+            var sb =
+                new StringBuilder("akka.remote.helios.tcp{").AppendLine()
```
Aaronontheweb: Needed to include the full `akka.remote` here

src/core/Akka.Remote.TestKit/Player.cs
```diff
@@ -474,9 +476,7 @@ public void InitFSM()
                         }
                         if (terminateMsg.ShutdownOrExit.IsLeft && terminateMsg.ShutdownOrExit.ToLeft().Value == true)
                         {
-                            //TODO: terminate more aggressively with Abort
-                            //Context.System.AsInstanceOf<ActorSystemImpl>().Abort();
-                            Context.System.Terminate();
+                            Context.System.AsInstanceOf<ActorSystemImpl>().Abort();
```
Aaronontheweb: Implemented `Abort` - using it here inside the `PlayerFsm`. Same as the JVM.

src/core/Akka.Remote.Tests.MultiNode/RemoteReDeploymentSpec.cs
```diff
@@ -0,0 +1,307 @@
...
+                    .Wait();
+                TestConductor.Shutdown(_config.Second, true).Wait();
+                if (ExpectQuarantine)
+                {
+                    
+                    Within(SleepAfterKill, () => 
+                    {
+                        ExpectMsg("PostStop");
+                        //need to pad the timing here, since `ExpectNoMsg` will wait until exactly SleepAfterKill and fail the spec
+                        ExpectNoMsg(Remaining - TimeSpan.FromSeconds(0.2));
```
Aaronontheweb: Found a fun bug with the normal `Akka.TestKit` here which I will create an issue for. This was my workaround.

src/core/Akka/Actor/ActorCell.FaultHandling.cs
```diff
@@ -215,14 +215,15 @@ private void Terminate()
...
 
-            //TODO: Implement when we have ActorSystem.Abort
-            //    if (systemImpl.aborting) {
-            //      // separate iteration because this is a very rare case that should not penalize normal operation
-            //      children foreach {
-            //        case ref: ActorRefScope if !ref.isLocal ⇒ self.sendSystemMessage(DeathWatchNotification(ref, true, false))
-            //        case _                                  ⇒
-            //      }
-            //    }
+            if (SystemImpl.Aborting)
```
Aaronontheweb: Aborting implementation - basically doesn't actually wait to terminate any remotely deployed actors. Assumes they're dead and shuts down anyway.

Aaronontheweb: This is "safe" to do because the `RemoteWatcher` will kill the remotely deployed children eventually.

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -98,6 +98,14 @@ public override IActorRef SystemActorOf<TActor>(string name = null)
             return _provider.SystemGuardian.Cell.AttachChild(Props.Create<TActor>(), true, name);
         }
 
+        internal volatile bool Aborting = false;
```
Aaronontheweb: Field gets read by the `ActorCell` of all terminating actors.

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
alexvaluyskiy: Good catch

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -98,6 +98,14 @@ public override IActorRef SystemActorOf<TActor>(string name = null)
...
+
+        public override void Abort()
```
alexvaluyskiy: Good. We should add XML docs here later

PR #2122, opened 2016-06-29
Added nuspec for Akka.Persistence.Query.Sql

Horusiath:
This will add `Akka.Persistence.Query.Sql` as a new NuGet package. 

Question: does `Akka.Persistence.Query.Sql.Tests` should be a nuget package too? (It will be necessary for 3rd party SQL plugins)


alexvaluyskiy:
Yes, I think it should be, as Akka.Persistence.Testkit does

cconstantin:
Maybe it should also be renamed to .TestKit

PR #2124, opened 2016-06-30
Added API approval tests for Remote/Cluster/Persistence/Cluster.Tools

alexvaluyskiy:
These packages is going to be stable in the near future, so we need to check api changes in PRs


Aaronontheweb:
I like it - we should add one of these for testkit too.

PR #2129, opened 2016-07-02
fixed bug with TransitionSpec

Aaronontheweb:


src/core/Akka.Cluster.Tests.MultiNode/TransitionSpec.cs
```diff
@@ -55,12 +55,12 @@ protected TransitionSpec(TransitionSpecConfig config) : base(config)
 
         private RoleName Leader(params RoleName[] roles)
         {
-            return Roles.First();
+            return roles.First();
```
Aaronontheweb: Needed to use the parameter, not the property

Aaronontheweb:
Still have a failure - going to need to root around in there some more.

Aaronontheweb:
I can't get this spec to fail locally at all. But it will reliably fail on the build server - I'm convinced it's a race with the node boot order affecting how the code responsible for selecting the leader not working. In other words, the node that gets picked as the "leader" isn't actually the leader when this spec fails. I have a patch I'll try once I can a chance to review some logs with the error conditions.

Aaronontheweb:
So this is weird - I explicitly enabled logging and the test didn't produce any of those messages. Is TeamCity caching a previous build or something? Going to force-kill all of the agents and retry.

Aaronontheweb:
Opted to disable this spec for now - been able to narrow down to an issue with this code:

```csharp
var leader12 = Leader(_config.First, _config.Second);
var tmp = Roles.Where(x => x != leader12).ToList();
var other1 = tmp.First();
var other2 = tmp.Skip(1).First();
```

This gets run on all 3 participating nodes. Looks like the issue is that the roleNames selected by this code can be racily different on some of the nodes, resulting in a configuration that can never pass. Basically we'll need to mess around with the equality operators for `RoleName` and their orderings to get that right.

I'm opting to disable the spec for now so we can just move on and release 1.1 without any further delays, since this issue is with the test and not what's being tested.

PR #2132, opened 2016-07-02
Update version: Newtonsoft.Json and System.Collections.Immutable

marcinrybaa:
When I used Akka.Net in my ASP MVC .NET Core 1.0 project I had problem with versions of Newtonsoft.Json and System.Collections.Immutable. Exception was thrown in runtime when I started ActorSystem. Problem was with versions:
- Newtonsoft.Json - Akka.Net - 7.0.0, Microsoft.AspNetCore.Mvc - 9.0.1
- System.Collections.Immutable - Akka.Net - 1.1.36, .NET Core 1.0 - 1.2.0

I updated this libraries in all csproj and Nuget files.


Aaronontheweb:
We will not be upgrading System.Collections.Immutable. It was intentionally rolled back to this version. Not interested in upgrading JSON.NET prior to this release. Closing.

PR #2133, opened 2016-07-03
Removed metrics from Cluster

alexvaluyskiy:
As part of this issue: https://github.com/akkadotnet/akka.net/issues/2068
The metrics will be added later in a separate package


Aaronontheweb:
Looks good - and yep, we can add them later

PR #2135, opened 2016-07-05
disabled Helios STDOUT logging; eliminated ERROR logs for shutdown associations

Aaronontheweb:
Just some logging cleanup before we push 1.1

### DIFF
--- a/src/core/Akka.Remote/Endpoint.cs
+++ b/src/core/Akka.Remote/Endpoint.cs
@@ -1030,7 +1030,9 @@ namespace Akka.Remote
 
         private void PublishAndThrow(Exception reason, LogLevel level)
         {
-            reason.Match().With<EndpointDisassociatedException>(endpoint => PublishDisassociated())
+            reason.Match()
+                .With<EndpointDisassociatedException>(endpoint => PublishDisassociated())
+                .With<ShutDownAssociation>(shutdown => {}) // don't log an error for planned shutdowns
                 .Default(msg => PublishError(reason, level));
 
             throw reason;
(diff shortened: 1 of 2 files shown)


src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1030,7 +1030,9 @@ private Deadline NewAckDeadline()
 
         private void PublishAndThrow(Exception reason, LogLevel level)
         {
-            reason.Match().With<EndpointDisassociatedException>(endpoint => PublishDisassociated())
+            reason.Match()
+                .With<EndpointDisassociatedException>(endpoint => PublishDisassociated())
+                .With<ShutDownAssociation>(shutdown => {}) // don't log an error for planned shutdowns
```
Aaronontheweb: This is not present in the JVM, but I think it may be a good idea. Basically we won't log disassociations that occur as part of a shutdown (planned termination) as errors. Disabling logging for this since we get a steady drumbeat of questions from developers who are worried they're doing something wrong. Since the answer is "no, you're not doing anything wrong" I figured we may as well not log it as an exception.

The shutdown association gets logged as a `DEBUG` message by the `EndpointManager` after this, so it will still appear in the log files.

src/core/Akka.Remote/Transport/Helios/HeliosTransport.cs
```diff
@@ -64,6 +65,11 @@ public HeliosTransportSettings(Config config)
 
         private void Init()
         {
+            // Disable STDOUT logging for Helios in release mode
+#if !DEBUG
+            LoggingFactory.DefaultFactory = new NoOpLoggerFactory();
```
Aaronontheweb: disables Helios' built-in logging

PR #2143, opened 2016-07-06
added missing copyright headers; added spec to verify joining when a seed node is offline

Aaronontheweb:


RELEASE_NOTES.md
```diff
@@ -1,4 +1,4 @@
-#### 1.0.9 April 26 2016 ####
```
Aaronontheweb: Ignore. Going to replace in a subsequent commit.

src/core/Akka.Cluster.Tests.MultiNode/JoinWithOfflineSeedNodeSpec.cs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    public class JoinWithOfflineSeedNodeSpecNode1 : JoinWithOfflineSeedNodeSpec { }
+    public class JoinWithOfflineSeedNodeSpecNode2 : JoinWithOfflineSeedNodeSpec { }
+
+    /// <summary>
+    /// Tests to ensure that if we have 2 seed nodes defined and 1 is offline during a <see cref="Cluster.JoinSeedNodes"/>
+    /// command, the joining node will still be marked as <see cref="MemberStatus.Up"/> as long as there are no unreachable nodes
+    /// within the cluster.
+    /// </summary>
```
Aaronontheweb: Comment explains the purpose of this spec - tests a common scenario where only 1 seed node needs to be Up in order for a cluster to be joined. Other previous seed nodes can be Downed / offline and the join will still go through correctly.

src/core/Akka.Cluster/ClusterEvent.cs
```diff
@@ -202,6 +202,11 @@ public override int GetHashCode()
                     return hash;
                 }
             }
+
+            public override string ToString()
+            {
+                return $"{GetType()}(Member={Member})";
```
Aaronontheweb: Added user-friendly stringifications of all `IClusterDomainEvent`s

src/core/Akka.Cluster/Gossip.cs
```diff
@@ -95,15 +94,6 @@ public Gossip(ImmutableSortedSet<Member> members, GossipOverview overview, Vecto
                 return Overview.Reachability.RemoveObservers(downed.Select(m => m.UniqueAddress).ToImmutableHashSet());
             });
 
-            _reachability = new Lazy<Reachability>(() =>
```
Aaronontheweb: Removed. Wasn't used.

src/SharedAssemblyInfo.cs
```diff
@@ -1,8 +1,16 @@
-﻿// <auto-generated/>
+﻿//-----------------------------------------------------------------------
+// <copyright file="SharedAssemblyInfo.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2015 Typesafe Inc. <http://www.typesafe.com>
+//     Copyright (C) 2013-2015 Akka.NET project <https://github.com/akkadotnet/akka.net>
```
Silv3rcircl3: This looks like one of the old headers, everywhere else we already set 2016 as year and Lightbend as company name

Silv3rcircl3:
URL is still http://www.typesafe.com, will create another PR to fix it.

PR #2152, opened 2016-07-07
close #2151 - fixes NRE on ActorPath.ElementsWithUid

### LINKED ISSUE #2151: NullReferenceException thrown during remote message routing if address conflict
Definitely a bug; investigating the reproduction steps now.

NRE gets thrown here:
```
   at Akka.Actor.ActorPath.get_ElementsWithUid()
   at Akka.Remote.RemoteActorRefProvider.ResolveActorRefWithLocalAddress(String path, Address localAddress)
   at Akka.Remote.Transport.AkkaPduProtobuffCodec.DecodeMessage(ByteString raw, RemoteActorRefProvider provider, Address localAddress)
   at Akka.Remote.EndpointReader.TryDecodeMessageAndAck(ByteString pdu)
```

Appears to occur during an address mismatch; i.e. one node bound to `127.0.0.1`, other is trying to contact `localhost`. Resolves to same physical address, but different logical addresses.

Aaronontheweb:
Pretty nasty bug - not sure how it occurs in the wild exactly but I observed it happening inside Lighthouse. Have more tests to cover it and a small fix.

### DIFF
--- a/src/core/Akka/Actor/ActorPath.cs
+++ b/src/core/Akka/Actor/ActorPath.cs
@@ -189,7 +189,7 @@ namespace Akka.Actor
                 var current = this;
                 var elements = new List<string>() { AppendUidFragment(current.Name) };
                 current = current.Parent;
-                while (!(current is RootActorPath))
+                while (!(current is RootActorPath || current == null))
                 {
                     elements.Add(current.Name);
                     current = current.Parent;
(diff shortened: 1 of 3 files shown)


src/core/Akka/Actor/ActorPath.cs
```diff
@@ -189,7 +189,7 @@ public IReadOnlyList<string> ElementsWithUid
                 var current = this;
                 var elements = new List<string>() { AppendUidFragment(current.Name) };
                 current = current.Parent;
-                while (!(current is RootActorPath))
+                while (!(current is RootActorPath || current == null))
```
JordanJones: Shouldn't that be `current != null`?

PR #2154, opened 2016-07-07
[api-docs] Updated Sandcastle build file

sean-gilliam:
This PR updates the Sandcastle build file to add the api docs of the newly added
projects (Akka.Persistence.Query, Akka.Streams and Akka.Streams.TestKit).

I also rearranged the entries in alphabetical order so that it would be
easier to find missing entries in the future.


Aaronontheweb:
@sean-gilliam nice, we'll get this in as part of 1.1.1.

PR #2155, opened 2016-07-08
Added shapshot rewritting to Sqlite

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/issues/2156
Also I've removed three redundant files from Sql.Common and have added config tests


alexvaluyskiy:
@Horusiath could you review it?

Issue #2157, opened 2016-07-08
GracefulStop causes an unnecessary DeadLetter.


veblush:
The following code makes an unnecessary `DeadLetter` while executing `GracefulStop`.

```csharp
var greeter = system.ActorOf<GreetingActor>("greeter");
greeter.Tell(new Greet("World"));
greeter.GracefulStop(TimeSpan.FromSeconds(1)).Wait();
```

DeadLetter log is

```
[akka://MySystem/deadLetters]
Message DeadLetter from akka://S/user/greeter to akka://S/deadLetters was not delivered.
1 dead letters encountered.
```

This is caused because there is a difference between Akka and Akka.NET.

`GracefulStop` of [Akka](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/pattern/GracefulStopSupport.scala#L54)
sends `Unwatch` message only when the target actor is not terminated because it is not necessary.

But `GracefulStop` of [Akka.NET](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/GracefulStopSupport.cs#L63) always
sends `Unwatch` message even when the actor is terminated, which makes `DeadLetter` message.

alexvaluyskiy:
@veblush could you make a PR with the fix and test for it?

veblush:
@alexvaluyskiy what kinds of test do you mean? there seem to be three tests for `GracefulStop` in PatternSpec.cs.

alexvaluyskiy:
yes, but none of them was failed before

alexvaluyskiy:
Test - subscribe to deadletters and send a terminate request

veblush:
@alexvaluyskiy ok, I got the point. it's a good idea.

PR #2158, opened 2016-07-08
Remove an unnecessary DeadLetter at GracefulStop. #2157

veblush:


alexvaluyskiy:
@veblush Looks good, but, It is a bad idea to change standard Scala specs if we want to verify a new bug. The better idea is - create a new test with name like "GracefulStopSupport_must_not_send_unnecessary_Deadletter_bug_2157"
And squash this PR into one commit, please

Aaronontheweb:
@veblush I agree with @alexvaluyskiy. Could you revert the original spec to as it was and then add a new spec to that file, as @alexvaluyskiy suggested?

veblush:
Oh good point. I didn't know that. I'll take it.

veblush:
Oh i failed to update this PR to have squash one commit. I create new PR and close this one.

PR #1847, opened 2016-04-04
Implemented DeadLetterSupression

alexvaluyskiy:
Implemented DeadLetterSupression
http://doc.akka.io/docs/akka/2.4.3/scala/event-bus.html#Dead_Letters

PR #2159, opened 2016-07-08
Fix a bug that handled DeadLetters are published.

veblush:
DeadLetterActorRef.TellInternal should process DeadLetter like [ActorRef.scala](https://github.com/akka/akka/blob/60fb163331aa2f9c22e27a9fc424f5975bc45d13/akka-actor/src/main/scala/akka/actor/ActorRef.scala#L582). Without this fix, `Watch` sent to already stopped actor will publish unncessary `DeadLetter` event.

### DIFF
--- a/src/core/Akka/Actor/BuiltInActors.cs
+++ b/src/core/Akka/Actor/BuiltInActors.cs
@@ -174,7 +174,7 @@ namespace Akka.Actor
 
         protected override void TellInternal(object message, IActorRef sender)
         {
-            if(message == null) throw new InvalidMessageException("Message is null");
+            if (message == null) throw new InvalidMessageException("Message is null");
             var i = message as Identify;
             if (i != null)
             {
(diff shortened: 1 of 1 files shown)


alexvaluyskiy:
I have the same changes here
https://github.com/akkadotnet/akka.net/pull/1847
But I still have one failed test with DeadLetter and GracefullStop. Could you help me to debug it?

alexvaluyskiy:
We are sending here a PoisonPill with NoSender
https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/GracefulStopSupport.cs#L49

PoisonPill marked as IDeadLetterSupression.
And this behavior should be handles here
https://github.com/alexvaluyskiy/akka.net/blob/99d440f3f493de266d11d279f38130b5dcfa8915/src/core/Akka/Actor/EmptyLocalActorRef.cs#L89

But the message never going through this method.
And I get an error here
https://github.com/alexvaluyskiy/akka.net/blob/99d440f3f493de266d11d279f38130b5dcfa8915/src/core/Akka/Event/DeadLetter.cs#L74

I know, that we are sending NoSender, which equals null, but how is it works in Scala? And why I got this error in DeadLetter constructor instead of SupressedDeadLetter?

Aaronontheweb:
@alexvaluyskiy do you have a stacktrace for that error?

src/core/Akka/Actor/BuiltInActors.cs
```diff
@@ -174,26 +174,28 @@ public DeadLetterActorRef(IActorRefProvider provider, ActorPath path, EventStrea
...
             if (i != null)
             {
                 sender.Tell(new ActorIdentity(i.MessageId, ActorRefs.Nobody));
                 return;
             }
             var d = message as DeadLetter;
-            if(d != null && !SpecialHandle(d.Message, d.Sender)) { _eventStream.Publish(d);
+            if (d != null)
+            {
+                if (!SpecialHandle(d.Message, d.Sender)) { _eventStream.Publish(d); }
```
Aaronontheweb: How is this any different than what we had before?

Aaronontheweb: Nevermind - I got it now.

PR #2165, opened 2016-07-09
Akka dependencies fixes

alexvaluyskiy:
- removed System.Web assembly
- replaced fastJson to NewtonsoftJson
- removed redundant SqliteQueryBuilder.cs file


alexvaluyskiy:
We can't remove Akka.Remote dependency from Akka.Persistence.Tests, because we test persistent message with RemoteActorRef. Do we really need these tests here?

PR #2168, opened 2016-07-10
DistributedPubSub: Use TypeQualifiedNameForManifest with Manifest required Serializers

jordansjones:
Instead of `.FullName`.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Serialization/DistributedPubSubMessageSerializer.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Serialization/DistributedPubSubMessageSerializer.cs
@@ -275,7 +275,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Serialization
             else
             {
                 if (serializer.IncludeManifest)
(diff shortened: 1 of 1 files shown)


Horusiath:
Any rationale for that? Problem with assembly qualified names is that they contains things like version or token number - which are not used elsewhere in serialization pipeline and can cause unexpected behavior when incremental upgrades are going to happen while pub/sub is working.

alexvaluyskiy:
I think we should have an extension like this and use it everywhere, instead of AssemblyQualifiedName
```csharp
    public static class TypeExtensions
    {
        public static string TypeQualifiedNameForManifest(this Type type)
        {
            return type == null ? string.Empty : $"{type.FullName}, {type.Assembly.GetName().Name}";
        }
    }
```

cconstantin:
@alexvaluyskiy iirc correctly I added such an extension when updating persistence a while back, but it may be in persistence.

JordanJones:
@Horusiath the rational is that without the assembly name, `Type.GetType(...)` returns `null`. But I'm all for having an extension method.

I'll create one if there isn't one already in the core `Akka` project and update my PR either tonight or tomorrow sometime.

jordansjones:
Turns out that I missed the `TypeQualifiedNameForManifest` method on the base serializer. I've opted to use that instead of the `AssemblyQualifiedName` value.

PR #2173, opened 2016-07-10
close #2172 - fixed current synchronization context dispatchers

### LINKED ISSUE #2172: Dispatcher bug: "CurrentSynchronizationContext may not be used as task scheduler"
Found the following bug when booting up a client application sample to use Akka.NET 1.1:

```
The current SynchronizationContext may not be used as a TaskScheduler.
Cause: System.InvalidOperationException: The current SynchronizationContext may not be used as a TaskScheduler.
   at System.Threading.Tasks.SynchronizationContextTaskScheduler..ctor()
   at Akka.Dispatch.CurrentSynchronizationContextExecutorServiceFactory.Produce(String id)
   at Akka.Dispatch.Dispatcher.LazyExecutorServiceDelegate.<>c__DisplayClass3_0.<.ctor>b__0()
   at Akka.Util.FastLazy`1.get_Value()
   at Akka.Dispatch.Dispatcher.ExecuteTask(IRunnable run)
   at Akka.Dispatch.MessageDispatcher.Attach(ActorCell cell)
   at Akka.Actor.ActorCell.Start()
   at Akka.Actor.RepointableActorRef.Point()
   at Akka.Actor.ActorCell.HandleSupervise(IActorRef child, Boolean async)
   at Akka.Actor.ActorCell.Supervise(IActorRef child, Boolean async)
   at Akka.Actor.ActorCell.SysMsgInvokeAll(EarliestFirstSystemMessageList messages, Int32 currentState)
```

Aaronontheweb:


src/core/Akka.Tests/Actor/Dispatch/CurrentSynchronizationContextDispatcherSpecs.cs
```diff
@@ -0,0 +1,36 @@
...
+                dispatcher = akka.actor.synchronized-dispatcher
+               }
+        ");
+
+        public CurrentSynchronizationContextDispatcherSpecs() : base(_config) { }
+
+        [Fact]
+        public void CurrentSynchronizationContextDispatcher_should_start_without_error_Fix2172()
+        {
+            var uiActor = Sys.ActorOf(EchoActor.Props(this), "some-ui-actor");
```
Aaronontheweb: Prior to fixing this bug, this spec would fail. Actor would fail while it was being created (first time the dispatcher gets used.)

src/core/Akka/Dispatch/CurrentSynchronizationContextDispatcher.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    /// <summary>
+    /// Used to create instances of the <see cref="PinnedDispatcher"/>. 
+    /// 
+    /// Each actor created using the pinned dispatcher gets its own unique thread.
+    /// <remarks>
+    /// Always returns a new instance.
+    /// </remarks>
+    /// </summary>
+    internal sealed class CurrentSynchronizationContextDispatcherConfigurator : MessageDispatcherConfigurator
```
Aaronontheweb: Decided to wrap the `CurrentSynchronizationContextDispatcher` into its own class - it needs similar semantics to the `PinnedDispatcher` (this dispatcher belongs to exactly 1 actor instance) and it also needs to be used exclusively with a single executor type.

src/core/Akka/Dispatch/CurrentSynchronizationContextDispatcher.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// <summary>
+    /// Behaves like a <see cref="PinnedDispatcher"/> and always executes using <see cref="CurrentSynchronizationContextExecutorServiceFactory"/>
+    /// </summary>
+    public sealed class CurrentSynchronizationContextDispatcher : Dispatcher
+    {
+        public CurrentSynchronizationContextDispatcher(MessageDispatcherConfigurator configurator, string id, int throughput, long? throughputDeadlineTime, ExecutorServiceFactory executorServiceFactory, TimeSpan shutdownTimeout) 
+            : base(configurator, id, throughput, throughputDeadlineTime, executorServiceFactory, shutdownTimeout)
+        {
+            /*
+             * Critical: in order for the CurrentSynchronizationContextExecutor to function properly, it can't be lazily
```
Aaronontheweb: This is where we really fixed the bug; normally the `ExecutorService` is lazily initiated. This won't work - we need to activate the `ExecutorService` immediately so the correct `SynchronizationContext` is captured. So we do that by just running a no-op task at the tail end of the constructor. This is safe because all of the code needed for this method to run is constructed in the base class constructor, so there's no possibility of a race condition there.

src/core/Akka/Dispatch/PinnedDispatcher.cs
```diff
@@ -15,14 +15,14 @@
 namespace Akka.Dispatch
 {
     /// <summary>
-    /// Used to create instances of the <see cref="SingleThreadDispatcher"/>. 
+    /// Used to create instances of the <see cref="PinnedDispatcher"/>.
```
Aaronontheweb: Fixed a documentation bug.

PR #2177, opened 2016-07-12
Added epsilon value to the Within assertion.

### LINKED ISSUE #2130: Bug: Akka.TestKit Within and ExpectNoMsg can't be used together
Found this bug while working on https://github.com/akkadotnet/akka.net/pull/2121#discussion_r68680668

A TestKit code block written like this:

```csharp
Within(TimeSpan.FromSeconds(10), () => {
 // other stuff
ExpectNoMsg();
});
```

`ExpectNoMsg()` will wait until the defined timespan inside `Within` is hit, which *causes* the `Within` block to report that it executed longer than expected.

Therefore we should use an epsilon value to allow the `Within` block to check for a slightly longer time than reported by its configured timespan - so `Within`'s check should check for 10.01 seconds in this case instead of *exactly* 10:00, which will fail because `ExpectNoMsg` waits until *exactly* 10:00.

Silv3rcircl3:
closes #2130

### DIFF
--- a/src/core/Akka.TestKit.Tests/TestKitBaseTests/WithinTests.cs
+++ b/src/core/Akka.TestKit.Tests/TestKitBaseTests/WithinTests.cs
@@ -1,0 +1,21 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="WithinTests.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using Xunit;
+
+namespace Akka.TestKit.Tests.TestKitBaseTests
+{
+    public class WithinTests : AkkaSpec
+    {
+        [Fact]
+        public void Within_should_increase_max_timeout_by_the_provided_epsilon_value()
+        {
+            Within(TimeSpan.FromSeconds(1), () => ExpectNoMsg(), TimeSpan.FromMilliseconds(50));
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/core/Akka.TestKit/TestKitBase_Within.cs
```diff
@@ -93,7 +93,8 @@ public T Within<T>(TimeSpan min, TimeSpan max, Func<T> function, string hint = n
             }
             if (!_testState.LastWasNoMsg)
             {
-                var tookTooLong = elapsed > maxDiff;
+                epsilonValue = epsilonValue ?? TimeSpan.Zero;
```
Silv3rcircl3: Keep the current behavior intact

src/core/Akka.TestKit.Tests/TestKitBaseTests/WithinTests.cs
```diff
@@ -0,0 +1,21 @@
...
+using Xunit;
+
+namespace Akka.TestKit.Tests.TestKitBaseTests
+{
+    public class WithinTests : AkkaSpec
+    {
+        [Fact]
+        public void Within_should_increase_max_timeout_by_the_provided_epsilon_value()
+        {
+            Within(TimeSpan.FromSeconds(1), () => ExpectNoMsg(), TimeSpan.FromMilliseconds(50));
```
Aaronontheweb: Good!

src/core/Akka.TestKit/TestKitBase_Within.cs
```diff
@@ -58,7 +58,7 @@ public T Within<T>(TimeSpan max, Func<T> function)
         /// the innermost enclosing `within` block.</para>
         /// <remarks>Note that the max duration is scaled using <see cref="Dilated(TimeSpan)"/> which uses the config value "akka.test.timefactor"</remarks>
         /// </summary>
-        public T Within<T>(TimeSpan min, TimeSpan max, Func<T> function, string hint = null)
+        public T Within<T>(TimeSpan min, TimeSpan max, Func<T> function, string hint = null, TimeSpan? epsilonValue = null)
```
Aaronontheweb: Technically a breaking change to the TestKit (adding a new optional parameter.) if we release updated NUnit and VsTest packages right away this should be ok. Would you or @Danthar be up for this?

Silv3rcircl3: Yup, I can do it.

PR #2178, opened 2016-07-13
close #2176 - child can be created during async/await

### LINKED ISSUE #2176: If actor is created inside RunTask action, it never receives messages from its child actors
I've attached a small example to demonstrate this issue. The biggest problem with this issue is in case of deep nested structure of the actors it is really hard to understand what the problem is. It seems to me that it is better to throw an exception if an operation of creating a child actor inside ```RunTask``` method is invalid.

```c#
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading;
using System.Threading.Tasks;
using Akka.Actor;

namespace ConsoleApplication1
{
    class Actor1 : ReceiveActor
    {
        public Actor1()
        {
            Receive<string>(m =>
            {
                Console.WriteLine("Received message: " + m);
                Context.Stop(Self);
            });

            // if RunTask is ommitted, everything works fine
            // otherwise actor2 never receives messages from its child actor
            RunTask(() =>
            {
                Context.ActorOf(Props.Create<Actor2>(), "actor-2");
            });
        }
    }

    class Actor2 : ReceiveActor
    {
        public Actor2()
        {
            Receive<string>(m =>
            {
                Context.Parent.Tell(m + ",world");
                Context.Stop(Self);
            });
        }

        protected override void PreStart()
        {
            Context.ActorOf(Props.Create<Actor3>(), "actor-3");
        }
    }

    class Actor3 : ReceiveActor
    {
        public Actor3()
        {
            Self.Tell("start");

            Receive<string>(s =>
            {
                Context.Parent.Tell("hello");
                Context.Stop(Self);
            });
        }
    }

    class Program
    {
        {
            using (var @as = ActorSystem.Create("actor-system"))
            {
                @as.ActorOf(Props.Create<Actor1>(), "actor-1");
                @as.WhenTerminated.Wait();
            }
        }
    }
}

```

Aaronontheweb:


Aaronontheweb:
closes #2176

src/core/Akka.Tests/Actor/BugFix2176Spec.cs
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            protected override void PreStart()
+            {
+                Self.Tell("started");
+            }
+        }
+
+        [Fact]
+        public void Fix2176_Constructor_Should_create_valid_child_actor()
```
Aaronontheweb: Failed before the fix, passes now.

src/core/Akka.Tests/Actor/BugFix2176Spec.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            var actor = Sys.ActorOf(Props.Create(() => new Actor1NonAsync(TestActor)), "actor1");
+            ExpectMsg("started");
+        }
+
+        [Fact]
+        public void Fix2176_RunTask_Should_create_valid_child_actor()
```
Aaronontheweb: Always passed - verified that the issue was with `RunTask` and the parent starting its children in a suspended state if it created children while processing an async / await state machine.

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -374,7 +374,7 @@ private IInternalActorRef MakeChild(Props props, string name, bool async, bool s
                     throw;
                 }
 
-                if (Mailbox != null)
+                if (Mailbox != null && IsFailed)
```
Aaronontheweb: The fix - we check to see if the parent is faulted before applying recursive suspend semantics to our children. Normally this code would normally only be hit if the parent was faulted and expected to restart (that was the only condition under which an actor could be suspended on the JVM.) But because of how we implement async / await, we should only trigger a suspension for a child if the suspension for the parent is because it's faulted - not because it's processing a `RunTask`.

PR #2182, opened 2016-07-14
Multiple enumeration exception in UnstartedCell.ReplaceWith

kantora:
After update to latest Akka night build I've broke several tests in my project.
Logs revealed strange exceptions like this:
```
2016-07-14 11:00:21.508 +03:00 [Error] "Guardian failed. Shutting down system"
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at Akka.Actor.UnstartedCell.ReplaceWith(ICell cell) in C:\Users\Kantora\AppData\Local\JetBrains\Shared\v05\DecompilerCache\decompiler\FE738B23-F14D-4A04-9754-5DA100BDB627\7a\99f2e440\UnstartedCell.cs:line 156
   at Akka.Actor.RepointableActorRef.Point() in C:\Users\Kantora\AppData\Local\JetBrains\Shared\v05\DecompilerCache\decompiler\FE738B23-F14D-4A04-9754-5DA100BDB627\4d\b12fdbd7\RepointableActorRef.cs:line 160
   at Akka.Actor.ActorCell.HandleSupervise(IActorRef child, Boolean async) in C:\Users\Kantora\AppData\Local\JetBrains\Shared\v05\DecompilerCache\decompiler\FE738B23-F14D-4A04-9754-5DA100BDB627\d1\cf6b63d7\ActorCell.cs:line 1418
   at Akka.Actor.ActorCell.Supervise(IActorRef child, Boolean async) in C:\Users\Kantora\AppData\Local\JetBrains\Shared\v05\DecompilerCache\decompiler\FE738B23-F14D-4A04-9754-5DA100BDB627\d1\cf6b63d7\ActorCell.cs:line 1405
   at Akka.Actor.ActorCell.SysMsgInvokeAll(EarliestFirstSystemMessageList messages, Int32 currentState) in C:\Users\Kantora\AppData\Local\JetBrains\Shared\v05\DecompilerCache\decompiler\FE738B23-F14D-4A04-9754-5DA100BDB627\d1\cf6b63d7\ActorCell.cs:line 1339
```
(I've used dotPeek to recreate symbol files)

It was strange as `_messageQueue` was protected with lock everywhere... Further examination revealed that with the use of calling-thread dispatcher - it executes message process in the thread of SendMessage (in the current thread). And it can send messages to `this.Self` which is still `UnstartedCell` that is also processed in the current thread, so `lock` has no use...

This is the proposed solution.

### DIFF
--- a/src/core/Akka/Actor/RepointableActorRef.cs
+++ b/src/core/Akka/Actor/RepointableActorRef.cs
@@ -276,12 +276,19 @@ namespace Akka.Actor
                 {
                     DrainSysMsgQueue(cell);
 
-                    foreach (var envelope in _messageQueue)
+                    // In some scenarios `cell.SendMessage` can send additional messages to Self in current thread, that would change _messageQueue
+                    while (this._messageQueue.Any())
                     {
-                        cell.SendMessage(envelope.Sender, envelope.Message);
+                        var queue = new List<Envelope>(this._messageQueue);
+                        this._messageQueue.Clear();
 
-                        // drain sysmsgQueue in case a msg enqueues a sys msg
-                        DrainSysMsgQueue(cell);
+                        foreach (var envelope in queue)
+                        {
+                            cell.SendMessage(envelope.Sender, envelope.Message);
+
+                            // drain sysmsgQueue in case a msg enqueues a sys msg
+                            DrainSysMsgQueue(cell);
+                        }
                     }
                 }
                 finally


Horusiath:
@Aaronontheweb Is it a hot path? If so, I think, we'll need to observe perf changes for this one.
Also @kantora could you attach example to recreate the exception you had?

kantora:
@Horusiath it's is not so easy... But I'll try...
Also, on second thought... Suppose that _messageQueue replacement from `List` to `Queue` should fix the problem without performance issues...

kantora:
@Horusiath How it would be better to provide test?
I can commit it in this pull request - but it is working with my changes.

Nevertheless, here is the code:
```
namespace Akka.Tests.Actor
{
    using Akka.Actor;
    using Akka.Configuration;
    using Akka.TestKit.Xunit2;

    using Xunit;
    using Xunit.Abstractions;

    public class UnstartedCellSpec : TestKit
    {
        public UnstartedCellSpec(ITestOutputHelper output = null)
            : base(CreateTestActorSystem(), output)
        {
        }

        [Fact]
        public void InitializationTest()
        {
            var actor = this.ActorOf(Props.Create<ActorForTest>().WithDispatcher("akka.test.calling-thread-dispatcher"), "test-actor");
            this.ExpectNoMsg();
            actor.Tell(new StateRequest(), this.TestActor);
            var state = this.ExpectMsg<ActorForTest.EnState>();
            Assert.Equal(ActorForTest.EnState.Stage2, state);
        }

        /// <summary>
        /// Creating actor system for test
        /// </summary>
        /// <returns>
        /// actor system for test
        /// </returns>
        private static ActorSystem CreateTestActorSystem()
        {
            var config = ConfigurationFactory.ParseString(@"
{
  akka {

    scheduler {
      implementation = ""Akka.TestKit.TestScheduler, Akka.TestKit"" 
    }

    test {
      # factor by which to scale timeouts during tests, e.g. to account for shared
      # build system load
      timefactor =  1.0

      # duration of EventFilter.intercept waits after the block is finished until
      # all required messages are received
      filter-leeway = 3s

      # duration to wait in expectMsg and friends outside of within() block
      # by default
      single-expect-default = 500ms

      # The timeout that is added as an implicit by DefaultTimeout trait 
      # This is used for Ask-pattern
      default-timeout = 500ms

      calling-thread-dispatcher {
        type = ""Akka.TestKit.CallingThreadDispatcherConfigurator, Akka.TestKit""        
        throughput = 2147483647
      }

      test-actor.dispatcher {
        type = ""Akka.TestKit.CallingThreadDispatcherConfigurator, Akka.TestKit""
        throughput = 2147483647
      }
    }
    actor : {
      debug {
        receive = on
        autoreceive = on
        lifecycle = on
        event-stream = on
        unhandled = on
      }

      deployment {
        ""/*"" {
          dispatcher = akka.test.calling-thread-dispatcher
        }

        ""/*/*"" {
          dispatcher = akka.test.calling-thread-dispatcher
        }

        ""/*/*/*"" {
          dispatcher = akka.test.calling-thread-dispatcher
        }
      }

      #default-dispatcher : {
      #  type : ""Akka.TestKit.CallingThreadDispatcherConfigurator, Akka.TestKit""
      #  throughput = 2147483647
      #}
    }
  }
}
");
            return ActorSystem.Create("test", config);
        }

        public class ActorForTest : ReceiveActor, IWithUnboundedStash
        {
            public enum EnState
            {
                Created,
                Stage1,
                Stage2
            }

            public ActorForTest()
            {
                this.Receive<InitState1>(s => this.Initialize());
                this.Receive<object>(m => this.Stash.Stash());
                this.Self.Tell(new InitState1());
            }

            public EnState State { get; set; } = EnState.Created;

            private void Initialize()
            {
                this.State = EnState.Stage1;
                this.Self.Tell(new InitState2());
                this.Become(
                    () =>
                        {
                            this.Receive<InitState2>(m => this.Init2());
                            this.Receive<StateRequest>(m => this.Sender.Tell(this.State));
                        });
                this.Stash.UnstashAll();
            }

            private void Init2()
            {
                this.State = EnState.Stage2;
            }

            public IStash Stash { get; set; }
        }

        public class InitState1
        {
            
        }

        public class InitState2
        {

        }

        public class StateRequest
        {
            
        }
    }
}
```

Horusiath:
I guess that `this.Self.Tell(new InitState1());` called from the constructor is causing an error. I don't know what is the statement of Lightbend for this, but in general I've never seen using `Self` tells (maybe it's an antipattern?) - if something like that is necessary it usually occurs as part of overridden `PreStart` method.

kantora:
Generealy, `this.Self.Tell(new InitState2());` causing exception... Moving it to `PreStart` doesn't fix the problem as it in the same thread..
Also whether this is antipattern or not - this is the bug in either way. Sending messages shouldn't break the system :)  IMHO

kantora:
@Horusiath Maybe I will swap `_messageQueue` to queue? So there will be no need in list cloning... But I don't think that this can be a performance problem.

Horusiath:
I was saying that, because I'm curious, if this problem can be reproduced on the JVM implementation. I'll try to do so after work.

kantora:
:) Waiting for the news.

Aaronontheweb:
@Horusiath this code will occur at startup for all top-level actors, after they transition from their temporary placeholder "repointable" actor reference to their concrete implementation. This would likely only have a significant performance issue if the top level actor received an enormous amount of messages immediately after being created and before being started. But the throughput benchmarks might tell you the whole story.

But yeah, this definitely seems like a bug.

Horusiath:
On the JVM side this works correctly.

Aaronontheweb:
I think I understand why we have the bug.....

```scala
 while (!queue.isEmpty) {
        cell.sendMessage(queue.poll())
        // drain sysmsgQueue in case a msg enqueues a sys msg
        drainSysmsgQueue()
      }
```
No enumeration of the queue in Scala. Just polling the front of it.... I like that solution a lot better than this one. Going to add a spec to check against regressions also.

Aaronontheweb:
Ok, have a fix for this that I like a bit better as it avoids extra copying / allocations. Also was able to use the reproduction scenario that @kantora included to recreate the bug and verify the fix.

Aaronontheweb:
Replaced with #2184

PR #2184, opened 2016-07-15
changed data structure in UnstartedCell to avoid multiple enumeration

Aaronontheweb:
Props to @kantora for finding, reporting, and providing an initial fix for this issue.

### DIFF
--- a/src/core/Akka.Tests/Actor/RepointableActorRefSpecs.cs
+++ b/src/core/Akka.Tests/Actor/RepointableActorRefSpecs.cs
@@ -1,0 +1,57 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="RepointableActorRefSpecs.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.TestKit;
+using Xunit;
+
+namespace Akka.Tests.Actor
+{
+    public class RepointableActorRefSpecs : AkkaSpec
+    {
+        public class Bug2182Actor : ReceiveActor, IWithUnboundedStash
+        {
+            public Bug2182Actor()
+            {
+                Receive<string>(str => str.Equals("init"), s => Become(Initialize));
+                ReceiveAny(o => Stash.Stash());
+                Self.Tell("init");
+            }
+
+            private void Initialize()
+            {
+                Self.Tell("init2");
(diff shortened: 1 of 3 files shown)


src/core/Akka.Tests/Actor/RepointableActorRefSpecs.cs
```diff
@@ -0,0 +1,57 @@
...
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.TestKit;
+using Xunit;
+
+namespace Akka.Tests.Actor
+{
+    public class RepointableActorRefSpecs : AkkaSpec
+    {
+        public class Bug2182Actor : ReceiveActor, IWithUnboundedStash
```
Aaronontheweb: Simplified @kantora's reproduction scenario and was able to reproduce the error using this spec.

src/core/Akka/Actor/RepointableActorRef.cs
```diff
@@ -238,7 +238,7 @@ public class UnstartedCell : ICell
         private readonly object _lock = new object();
 
        /* Both queues must be accessed via lock */
-        private readonly List<Envelope> _messageQueue = new List<Envelope>();
+        private readonly LinkedList<Envelope> _messageQueue = new LinkedList<Envelope>();
```
Aaronontheweb: Switched to using a `LinkedList` to conserve memory once the cell starts and the list is drained / empty.

src/core/Akka/Actor/RepointableActorRef.cs
```diff
@@ -276,9 +276,12 @@ public void ReplaceWith(ICell cell)
                 {
                     DrainSysMsgQueue(cell);
 
-                    foreach (var envelope in _messageQueue)
+                    while (_messageQueue.Count > 0)
```
Aaronontheweb: careful to avoid enumerating the collection itself....

src/core/Akka/Actor/RepointableActorRef.cs
```diff
@@ -276,9 +276,12 @@ public void ReplaceWith(ICell cell)
...
                     {
-                        cell.SendMessage(envelope.Sender, envelope.Message);
+                        // roughly equal to what "poll" does
```
Aaronontheweb: Leaving an implementation comment to note that what I provided below is functionally equivalent to http://www.tutorialspoint.com/java/util/linkedlist_poll.htm

src/core/Akka/Actor/RepointableActorRef.cs
```diff
@@ -366,7 +369,7 @@ private void SendMessage(object message, IActorRef sender)
                     }
                     else
                     {
-                        _messageQueue.Add(new Envelope { Message = message, Sender = sender });
+                        _messageQueue.AddLast(new Envelope { Message = message, Sender = sender });
```
Aaronontheweb: Append to the end of the linked list, consistent with what you'd expect from a queue structure.

Issue #1544, opened 2015-12-14
ActorSelection fails for ActorPath from remotely deployed actors


Horusiath:
Right now `ActorSelection` doesn't work when combined with path taken from remotely deployed actor:

``` csharp
// remoteRef is remotely deployed actor
system.ActorSelection(remoteRef.Path).Ask<ActorIdentity>(new Identify(null)).Result;
```

This operation will timeout - actor won't be find under provided path. Also any messages that will be send under actor selection specified that way will be send to dead letters on the remote node.

I don't know what is expected behavior on the JVM side for that.

rogeralsing:
That would be me:

https://github.com/rogeralsing/akka.net/blob/master/src/core/Akka.Remote/RemoteDaemon.cs#L165

Horusiath:
Scaka version for comparison:

```scala
override def getChild(names: Iterator[String]): InternalActorRef = {

    @tailrec
    def rec(s: String, n: Int): (InternalActorRef, Int) = {
      import akka.actor.ActorCell._
      val (childName, uid) = splitNameAndUid(s)
      getChild(childName) match {
        case null ⇒
          val last = s.lastIndexOf('/')
          if (last == -1) (Nobody, n)
          else rec(s.substring(0, last), n + 1)
        case ref if uid != undefinedUid && uid != ref.path.uid ⇒ (Nobody, n)
        case ref ⇒ (ref, n)
      }
    }

    val full = Vector() ++ names
    rec(full.mkString("/"), 0) match {
      case (Nobody, _) ⇒ Nobody
      case (ref, 0)    ⇒ ref
      case (ref, n)    ⇒ ref.getChild(full.takeRight(n).iterator)
    }
  }
```

alexpantyukhin:
I will take it.

Horusiath:
@alexpantyukhin would you also submit a test case to ensure that this won't reappear in the future?

alexpantyukhin:
okay. I will

rogeralsing:
Also, before starting this.
Try the `system.ActorSelection(remoteRef.Path).Ask<ActorIdentity>(new Identify(null)).Result;` and put a breakpoint in RemoteSystemDaemon.GetChild and see if you even get there.

If not, then the issue resides elsewhere.

maxim-s:
Also would be great to port missing specs from https://github.com/akka/akka/blob/master/akka-remote/src/test/scala/akka/remote/RemotingSpec.scala

[...]

PR #2196, opened 2016-07-18
Update ActorSelection and specs

### LINKED ISSUE #1748: ActorSelectionSpec does not cover JVM test criteria
Noticed today that our `ActorSelectionSpec` doesn't cover most of the stuff covered in the JVM's version of the spec:

https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/scala/akka/actor/ActorSelectionSpec.scala

For instance, https://github.com/akka/akka/blob/master/akka-actor-tests/src/test/scala/akka/actor/ActorSelectionSpec.scala#L344 - do we have a test that verifies that a message intended for a non-existent actor goes to DeadLetters? There's lots of others that need ot be covered here.

Silv3rcircl3:
closes #1748

### DIFF
--- a/src/core/Akka.Remote.Tests/RemotingSpec.cs
+++ b/src/core/Akka.Remote.Tests/RemotingSpec.cs
@@ -18,13 +18,14 @@ using Akka.Util;
 using Akka.Util.Internal;
 using Google.ProtocolBuffers;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Akka.Remote.Tests
 {
 
     public class RemotingSpec : AkkaSpec
     {
-        public RemotingSpec() : base(GetConfig())
+        public RemotingSpec(ITestOutputHelper helper) : base(GetConfig(), helper)
         {
             var c1 = ConfigurationFactory.ParseString(GetConfig());
             var c2 = ConfigurationFactory.ParseString(GetOtherRemoteSysConfig());
@@ -145,14 +146,14 @@ namespace Akka.Remote.Tests
         #region Tests
 
 
-        [Fact()]
+        [Fact]
         public void Remoting_must_support_remote_lookups()
         {
             here.Tell("ping", TestActor);
             ExpectMsg(Tuple.Create("pong", TestActor), TimeSpan.FromSeconds(1.5));
         }
 
-        [Fact()]
+        [Fact]
         public async Task Remoting_must_support_Ask()
         {
             //TODO: using smaller numbers for the cancellation here causes a bug.
@@ -190,7 +191,7 @@ namespace Akka.Remote.Tests
             ExpectMsg(76);
         }
 
-        [Fact(Skip = "races with TestTransport")]
+        [Fact]
         public void Remoting_must_lookup_actors_across_node_boundaries()
         {
             Action<IActorDsl> act = dsl =>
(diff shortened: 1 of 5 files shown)


src/core/Akka.Remote.Tests/RemotingSpec.cs
```diff
@@ -190,7 +191,7 @@ public void Remoting_must_not_send_remote_recreated_actor_with_same_name()
             ExpectMsg(76);
         }
 
-        [Fact(Skip = "races with TestTransport")]
```
Silv3rcircl3: Issue with the test transport was fixed so we can enable this spec again

src/core/Akka.Remote.Tests/RemotingSpec.cs
```diff
@@ -202,34 +203,151 @@ public void Remoting_must_lookup_actors_across_node_boundaries()
...
+            ExpectMsg<ActorIdentity>().Subject.ShouldBeSame(l);
+
+            grandchild.Tell(Tuple.Create(Props.Create<Echo1>(), "grandgrandchild"));
+            var grandgrandchild = ExpectMsg<IActorRef>();
+
+            Sys.ActorSelection("/user/looker/child").Tell(new Identify("idReq1"));
+            ExpectMsg<ActorIdentity>(i => i.MessageId.Equals("idReq1")).Subject.ShouldBe(child);
+            //TODO see #1544
+            //Sys.ActorSelection(child.Path).Tell(new Identify("idReq2"));
+            //ExpectMsg<ActorIdentity>(i => i.MessageId.Equals("idReq2")).Subject.ShouldBe(child);
```
Silv3rcircl3: ActorSelection with a deployed path is still not working

src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -258,37 +258,27 @@ private void HandleDaemonMsgCreate(DaemonMsgCreate message)
...
-            var joined = string.Join("/", parts);
-
-            var child = GetChild(joined);
-            if (child != null)
-            {
-                return child;
-            }
-            var last = joined.LastIndexOf("/", StringComparison.Ordinal);
-            var i = 1;
-            while (last != -1 || parts.Length <= i)
```
Silv3rcircl3: With a deployed actor path `last` was `-1 ` but `i` was < `parts.Length` so the `Substring` in the next line throws.

src/core/Akka/Actor/ActorSelection.cs
```diff
@@ -22,31 +23,32 @@ namespace Akka.Actor
...
         /// <value>The anchor.</value>
         public IActorRef Anchor { get; private set; }
 
         /// <summary>
-        ///     Gets or sets the elements.
+        /// Gets the elements.
         /// </summary>
         /// <value>The elements.</value>
-        public SelectionPathElement[] Path { get; set; }
+        public SelectionPathElement[] Path { get; private set; }
```
Silv3rcircl3: Need to keep `Anchor` and `Path` with private setter otherwise a remote selection wouldn't work anymore, I assume it's the serialization issue with Newtonsoft < 8 and getter only properties

src/core/Akka/Actor/ActorSelection.cs
```diff
@@ -22,31 +23,32 @@ namespace Akka.Actor
...
 
         /// <summary>
-        /// <see cref="string"/> representation of all of the elements in the <see cref="ActorSelection"/> path.
+        /// <see cref="string"/> representation of all of the elements in the <see cref="ActorSelection"/> path,
+        /// starting with "/" and separated with "/".
         /// </summary>
-        public string PathString { get { return string.Join("/", Path.Select(x => x.ToString())); } }
+        public string PathString => "/" + string.Join("/", Path.Select(x => x.ToString()));
```
Silv3rcircl3: According to the jvm the `PathString` should start with a /

src/core/Akka/Actor/ActorSelection.cs
```diff
@@ -57,24 +59,26 @@ public ActorSelection(IActorRef anchor, SelectionPathElement[] path)
...
+        /// Initializes a new instance of the <see cref="ActorSelection" /> class.
         /// </summary>
         /// <param name="anchor">The anchor.</param>
         /// <param name="elements">The elements.</param>
         public ActorSelection(IActorRef anchor, IEnumerable<string> elements)
         {
             Anchor = anchor;
+            
             Path = elements
+                .Where(s=>!string.IsNullOrWhiteSpace(s))
```
Silv3rcircl3: This handles empty elements, for example a trailing /, otherwise it would result in a additional `SelectChildName`

src/core/Akka/Actor/ActorSelection.cs
```diff
@@ -213,26 +226,35 @@ public override bool Equals(object obj)
 
         protected bool Equals(ActorSelection other)
         {
-            return Equals(Anchor, other.Anchor) && Equals(Path, other.Path);
+            return Equals(Anchor, other.Anchor) && Equals(PathString, other.PathString);
```
Silv3rcircl3: `Path` was checked for reference equality, so two selections were only equal if they used the same path array reference... Used `PathString` for simplicity instead

src/core/Akka.Tests/Actor/ActorSelectionSpec.cs
```diff
@@ -6,112 +6,371 @@
...
-            ExpectMsg("hello3");
+            Identify("a/b/c").ShouldBe(null);
+            Identify("a/b/c").ShouldBe(null);
+            Identify("akka://all-systems/Nobody").ShouldBe(null);
+            Identify("akka://all-systems/user").ShouldBe(null);
+            Identify("user/hallo").ShouldBe(null);
+            Identify("foo://user").ShouldBe(Nobody.Instance);
+            Identify("/deadLetters").ShouldBe(Nobody.Instance);
+            Identify("deadLetters").ShouldBe(Nobody.Instance);
+            Identify("deadLetters/").ShouldBe(Nobody.Instance);
```
Silv3rcircl3: On the jvm it should always return `null` but I think `Nobody` is fine too.

src/core/Akka.Remote.Tests/RemotingSpec.cs
Silv3rcircl3: I haven't searched for the actual fix, only saw maxims https://github.com/akkadotnet/akka.net/issues/1544#issuecomment-166070834
while I implemented 
> Remoting_must_select_actors_across_node_boundaries

So I enabled the spec again and it was passing consistently, therefore I assuemend it was fixed.

Aaronontheweb: It still may not be working. We should re-run the specs a few times and see if we can get it to fail.

PR #2198, opened 2016-07-18
Replaced Thread.Yield

Silv3rcircl3:


src/core/Akka.Streams/Implementation/Fusing/ActorGraphInterpreter.cs
```diff
@@ -267,43 +289,57 @@ public void TryAbort(Exception reason)
...
         {
             try
             {
-                var effectiveLimit = !_settings.IsFuzzingMode
-                    ? _eventLimit
-                    : (ThreadLocalRandom.Current.Next(2) == 0
-                        ? (Thread.Yield() ? 1 : 0)
-                        : ThreadLocalRandom.Current.Next(2));  // 1 or 0 events to be processed
+                var usingShellLimit = _shellEventLimit < actorEventLimit;
+                var remainingQuta = _interpreter.Execute(Math.Min(actorEventLimit, _shellEventLimit));
```
alexvaluyskiy: remainingQuota

src/core/Akka.Streams/Implementation/Fusing/ActorGraphInterpreter.cs
```diff
@@ -872,26 +908,41 @@ public static Props Props(GraphInterpreterShell shell)
...
+            _eventLimit = _initial.Materializer.Settings.SyncProcessingLimit;
+            _currentLimit = _eventLimit;
         }
 
         public ILoggingAdapter Log => _log ?? (_log = Context.GetLogger());
 
+        private void EnqueueToShortCircuit(object input)
+        {
+            if(_shortCircuitBuffer == null)
+                _shortCircuitBuffer = new Queue<object>();
```
alexvaluyskiy: Why Queue instead of LinkedList?

src/core/Akka.Streams/Implementation/Fusing/ActorGraphInterpreter.cs
```diff
@@ -910,7 +961,7 @@ private bool TryInit(GraphInterpreterShell shell)
         public IActorRef RegisterShell(GraphInterpreterShell shell)
         {
             _newShells.Enqueue(shell);
-            Self.Tell(ShellRegistered.Instance);
+            EnqueueToShortCircuit(ShellRegistered.Instance);
```
alexvaluyskiy: Is ShellRegistered the same as Resume in JVM?

alexvaluyskiy:
LGTM

src/core/Akka.Streams/Implementation/Fusing/ActorGraphInterpreter.cs
Silv3rcircl3: I see no reason for a LinkedList and Queue has the simpler API for that use case. 

```var element = queue.Dequeue()``` 

vs
```
var element = list.First;
list.RemoveFirst();
```

src/core/Akka.Streams/Implementation/Fusing/ActorGraphInterpreter.cs
Silv3rcircl3: Yeah I struggled with this one too, we needed two classes for the `Resume` version from the JVM. 
It's a `case class` which takes a `GraphInterpreterShell` as parameter and in some places it is used as instance with a shell but in this particular case only the "singelton" is used, for this cases we have `ShellRegistered`

Silv3rcircl3:
@alexvaluyskiy addressed your remarks

PR #2199, opened 2016-07-18
More serializer tests

rogeralsing:
This PR adds some new serializer tests to allow us to pin down potential future issues.
They were used to fix the latest Wire compatibility problems


alexvaluyskiy:
LGTM

PR #2200, opened 2016-07-18
Much better UnboundedMailboxQueue

Scooletz:
This PR provides a much better concurrent collection for `UnboundedMailboxQueue` than `ConcurrentQueue{T}`. The new structure called `ConcurrentEnvelopeQueue` has been written to store `Envelope` structs only. I checked the following performance tests repeatedly, still obtaining a meaningful positive difference.

1. `Akka.Tests.Performance.Dispatch.MailboxMemoryFootprintSpec+UnboundedMailbox` (Average result of total bytes allocated)
1. `Akka.Tests.Performance.Dispatch.MailboxBenchmarks+MailboxBatchRunPerf` (Average # of operations / s)

|Branch | Average msg/s | TotalBytesAllocated |
| --- | --- | --- |
| this PR | 6,501,868.72 | 7,274,546.46 bytes |
| dev| 6,145,230.66 | 7,756,292.31 bytes |

### DIFF
--- a/src/core/Akka.Tests/Akka.Tests.csproj
+++ b/src/core/Akka.Tests/Akka.Tests.csproj
@@ -204,6 +204,7 @@
     <Compile Include="TestUtils\Supervisor.cs" />
     <Compile Include="Util\ByteStringSpec.cs" />
     <Compile Include="Util\CollectionExtensionsSpec.cs" />
+    <Compile Include="Util\ConcurrentEnvelopeQueueTests.cs" />
     <Compile Include="Util\ContinuousEnumeratorSpec.cs" />
     <Compile Include="Util\IndexSpec.cs" />
     <Compile Include="Util\Internal\Collections\IteratorTests.cs" />
--- a/src/core/Akka.Tests/Util/ConcurrentEnvelopeQueueTests.cs
+++ b/src/core/Akka.Tests/Util/ConcurrentEnvelopeQueueTests.cs
@@ -1,0 +1,114 @@
+﻿using System;
+using System.Linq;
+using System.Threading;
+using Akka.Actor;
+using Akka.Dispatch.SysMsg;
+using Akka.IO;
+using Akka.TestKit;
+using Akka.Util;
+using Xunit;
+
+namespace Akka.Tests.Util
+{
+    public class ConcurrentEnvelopeQueueTests
+    {
+        [Fact]
+        public void When_empty_Should_not_dequeue()
+        {
+            var queue = new ConcurrentEnvelopeQueue();
+
+            Envelope result;
+            queue.TryDequeue(out result).ShouldBeFalse();
+            result.Sender.ShouldBe(null);
+            result.Message.ShouldBe(null);
+        }
+
+        [Fact]
+        public void When_empty_Should_look_like_empty()
+        {
+            var queue = new ConcurrentEnvelopeQueue();
+
+            queue.Count.ShouldBe(0);
+            queue.IsEmpty.ShouldBeTrue();
+        }
+
+        [Fact]
+        public void When_enqueued_Should_be_able_to_dequeue()
+        {
+            var queue = new ConcurrentEnvelopeQueue();
+            var message = new object();
+
+            var envelope = new Envelope
+            {
+                Sender = ActorRefs.NoSender,
+                Message = message
+            };
+
+            queue.Enqueue(ref envelope);
+
+            Envelope result;
+            queue.TryDequeue(out result).ShouldBeTrue();
+
+            result.Message.ShouldBe(message);
+
+            ReferenceEquals(result.Sender, ActorRefs.NoSender).ShouldBeTrue();
+        }
+
+        [Fact]
+        public void When_enqueued_from_multiple_producers_should_be_dequeued_in_fifo_way()
+        {
+            const int messagesToSendPerThread = 1 << 16;
+            const int producerCount = 2;
+            const int expectedMessages = messagesToSendPerThread*producerCount;
+
+            var startEvent = new ManualResetEventSlim(false);
+            var producers = new Thread[producerCount];
+            var endEvent = new CountdownEvent(producers.Length);
+            var lastSeenMessagePerProducer = new int[producerCount];
+
+            var queue = new ConcurrentEnvelopeQueue();
+
+            var messages = new[] {new object(), new object(), new object(), new object(),};
+            var envelopes = messages.Select(m => new Envelope {Message = m}).ToArray();
+
+            for (var i = 0; i < producers.Length; i++)
+            {
+                producers[i] = new Thread(msgs =>
+                {
+                    var toSend = (Envelope[]) msgs;
+                    startEvent.Wait();
+                    for (var j = 0; j < messagesToSendPerThread; j++)
+                    {
+                        queue.Enqueue(ref toSend[(j + 1)%2]);
+                    }
+                    endEvent.Signal();
+                })
+                {
+                    IsBackground = true,
+                    Name = "Producer #" + i
+                };
+                producers[i].Start(envelopes.Skip(i*2).Take(2).ToArray());
+            }
+
+            startEvent.Set();
+            var count = 0;
+            while (count < expectedMessages)
+            {
+                Envelope result;
+                if (queue.TryDequeue(out result))
+                {
+                    var index = Array.IndexOf(messages, result.Message);
+                    var producerIndex = index >> 1;
+                    var messageIndex = index & 1;
+
+                    lastSeenMessagePerProducer[producerIndex].ShouldNotBe(messageIndex);
+                    lastSeenMessagePerProducer[producerIndex] = messageIndex;
+
+                    count += 1;
+                }
+            }
+
+            endEvent.Wait(TimeSpan.FromSeconds(1)).ShouldBeTrue();
+        }
+    }
+}
--- a/src/core/Akka/Akka.csproj
+++ b/src/core/Akka/Akka.csproj
@@ -269,6 +269,7 @@
     <Compile Include="NotUsed.cs" />
     <Compile Include="Util\ByteIterator.cs" />
     <Compile Include="Util\ByteString.cs" />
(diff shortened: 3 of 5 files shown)


rogeralsing:
Impressive!
Is the queue implementation based on the .NET concurrent queue?
If not, then it will be a pain to verify that there are no hidden raceconditions like the Mono implementation suffered from for a long time.

Scooletz:
Cannot say how nice is to hear these words @rogeralsing :)
Yes, it's mostly based, with pruned API and the `status` that's totally unneeded taking into consideration the `Envelope` struct and how you can model a queue for it (as I did).
If I had to pull this in, I'd be on the fence as well. I think I cannot provide you better coverage than in a test producing envelopes with two threads and consuming with one within reasonable time boundaries.
 I think that, if we had a longer conversation about possible relaxing `Mailbox.ProcessMailbox` we could make it even faster (take a look at [my EventStore PR](https://github.com/EventStore/EventStore/pull/982)).

Aaronontheweb:
@rogeralsing @Scooletz the ActorModelSpec tests for those concurrency issues using the default dispatcher. Best way to test the safety for the queue itself though is with a property-based test that concurrently interleaves queue and dequeue operations and tests for a valid ordering. I did this with the Single Thread Event Executor in DotNetty / Helios by using math operations with commutative property (addition / subtraction) against an unsafe counter. The final value of the counter should always equal the sum of the operations. If the counter ever does not equal that, it means you found a non-linearizable interleaving, which means a race condition (unsafe write or read) somewhere. Run the test each time with 10s of thousands of runs, which will generate a larger and larger number of operations the longer the test goes.

JeffCyr:
Interesting, have you tried combining the message  and sender array in a single Envelope array? It would save an allocation by segment.

Scooletz:
@Aaronontheweb If you use a commutative operation it will not guarantee FIFO ordering for producers. Could you take a look at the test I've written in this PR [here](https://github.com/Scooletz/akka.net/blob/cc529db9d3002fa4f7e36cea1c34d51545baac76/src/core/Akka.Tests/Util/ConcurrentEnvelopeQueueTests.cs#L58-L113). It provides a different approach in my opinion. It uses a small (for now length is 2 but can be bigger) array of messages per producer and ensures that for every producer the order of messages is: 1, 2, 1, 2, 1, 2. This ensures FIFO ordering for every producer testing the consumer in the same time. What do you think?

Scooletz:
@JeffCyr `Envelope` is mutable but it uses properties to store values so yes a separate wrapper would be needed.

> It would save an allocation by segment.

I gave it a try and here are results:

|Branch | Average msg/s | TotalBytesAllocated |
| --- | --- | --- |
| this PR | 6,501,868.72 | 7,274,546.46 bytes |
| one struct |  6,133,574.53 | 6,956,831.38 bytes |
| dev|  6,145,230.66 | 7,756,292.31 bytes |

It improved `TotalBytesAllocated` but reduced the throughput to the `dev` lvl. I think I'm going to leave it as it is for now and it reduces the allocations and speeds up the queue itself.

JeffCyr:
I ran a benchmark which tests the raw speed of ConcurrentQueue instead of going through the Mailbox:

### x64
```
ConcurrentQueue:                        9 009 921 msg/sec
ConcurrentEnvelopeQueue:                19 159 676 msg/sec
ConcurrentEnvelopeQueue (One struct):   19 026 782 msg/sec

ConcurrentQueue:                        9 231 541 msg/sec
ConcurrentEnvelopeQueue:                17 077 986 msg/sec
ConcurrentEnvelopeQueue (One struct):   18 986 072 msg/sec

ConcurrentQueue:                        9 211 244 msg/sec
ConcurrentEnvelopeQueue:                15 678 632 msg/sec
ConcurrentEnvelopeQueue (One struct):   17 522 275 msg/sec
```

### x86
```
ConcurrentQueue:                        29 386 969 msg/sec
ConcurrentEnvelopeQueue:                8 926 675 msg/sec
ConcurrentEnvelopeQueue (One struct):   7 988 184 msg/sec

ConcurrentQueue:                        28 597 527 msg/sec
ConcurrentEnvelopeQueue:                9 528 867 msg/sec
ConcurrentEnvelopeQueue (One struct):   9 452 588 msg/sec

ConcurrentQueue:                        28 471 815 msg/sec
ConcurrentEnvelopeQueue:                9 538 180 msg/sec
ConcurrentEnvelopeQueue (One struct):   9 507 228 msg/sec
```

I can't explain the huge difference between x86 and x64, my guess is a memory alignment issue.

Here is my benchmark: https://gist.github.com/JeffCyr/174ae2724a47d21c195aeb8bf9d12046

Aaronontheweb:
@Scooletz 

>  If you use a commutative operation it will not guarantee FIFO ordering for producers. 

My test was for safety. Order preservation is a different story (linearizable is a superset of FIFO

@JeffCyr yeah, this data structure needs to be cache-aligned for folks who intend to use Akka in embedded scenarios. Can you take a look at it and see what it would take to bring up the X86 performance?

Aaronontheweb:
Is this test failure an issue introduced by this change?

<details>
<summary>Akka.FSharp.Tests.ApiTests.ApiTests.actor that accepts unit will receive unit message	</summary>
 Test collection for Akka.FSharp.Tests.ApiTests (1): Akka.FSharp.Tests (1)
  ApiTests.actor that accepts unit will receive unit message	
 
First failure:  	2200	#1587	szymon kulec <… (1)	18 Jul 16 17:52
System.AggregateException : One or more errors occurred.
---- System.AggregateException : One or more errors occurred.
-------- System.Threading.Tasks.TaskCanceledException : A task was canceled.
   at Microsoft.FSharp.Control.AsyncBuilderImpl.commit[a](Result`1 res)
   at Microsoft.FSharp.Control.CancellationTokenOps.RunSynchronously[a](CancellationToken token, FSharpAsync`1 computation, FSharpOption`1 timeout)
   at Microsoft.FSharp.Control.FSharpAsync.RunSynchronously[T](FSharpAsync`1 computation, FSharpOption`1 timeout, FSharpOption`1 cancellationToken)
   at Akka.FSharp.Tests.ApiTests.actor that accepts unit will receive unit message() in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.FSharp.Tests\ApiTests.fs:line 141
----- Inner Stack Trace -----
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Akka.FSharp.Actors.tryCast[Message](Task`1 t) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka.FSharp\FsApi.fs:line 48
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
----- Inner Stack Trace -----
</details>

Scooletz:
@JeffCyr Would you consider starting the producer and a consumer at the same time using a wait handle? Maybe the results are not based on the structure but running consumer slightly later. I see no reason why it can be slower on x86. The array of structs consisting of two references and the size of 32. 
I've reviewed the akka test. It's single threaded hence no cache trashing etc. Using one structure seems more reasonable.

Scooletz:
@Aaronontheweb Yes, that was public `ConcurrentEnvelopeQueue` . I made it internal.

Scooletz:
Could anyone take a look at the last failing test: `actor that accepts

Scooletz:
Considering two tests:

1. [actor that accepts unit will receive unit message](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.FSharp.Tests/ApiTests.fs#L126-L145) that passes a null message in the `Envelope`.
1. [MailboxBatchRunPerf](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Tests.Performance/Dispatch/MailboxBenchmarks.cs#L54-L70) that passes `ActorRefs.NoSender` which is null as sender

none of the fields from the `Envelope` can be used to mark the envelope as written. This was the reason for failing tests. Having this said, we can't apply this approach. Thoughts?

JeffCyr:
It could work by swapping the null nosender by a singleton and restore it to null on dequeue, but considering the performance trade-off between x86 and x64 I would say to keep using the .Net ConcurrentQueue. But the IsEmpty check vs Count can still be commited.

FYI I still don't understand the difference between x64 and x86, but I isolated the diff to TryAppend, for some unknown reason the volatile 32bit ref write + non-volatile 32bit ref write is way more expansive than 8bit volatile write + 32bit*2 non-volatile write.

Scooletz:
Yeah, I've tried this `NullObject` approach. I agree with you @JeffCyr . I could do the condition `if (IntPtr.Size == 8)` and return the new queue, but then again, I'm not sure if adding another queue only for `x64` would be a good thing.

@JeffCyr I'm in the same place :)

@Aaronontheweb I'm closing this and will issue a separate PR for `IsEmpty` property.

Scooletz:
I've debugged it and got some answers described on my blog in [.NET volatile write performance degradation in x86](https://blog.scooletz.com/2016/07/25/net-volatile-write-performance-degradation-in-x86/). Summing up, for x86 `Volatile.Write` on object generates a lot of checks whether or not the pusher reference is a stack reference and more. 
Would you consider providing conditionally a better structure for `x64`?

JeffCyr:
Good investigation! But now I want to know why this check is required in x86 and not in x64 :)

Anyhow, I'm not against using a separate implementation of the queue in x64, I would just like to see how it affects the full stack in the [PingPong benchmark](https://github.com/akkadotnet/akka.net/tree/dev/src/benchmark/PingPong). (Remove the "Prefer 32bit" flag from the .csproj).

Aaronontheweb:
@Scooletz @JeffCyr it should be easy to detect which architecture type we're using at runtime (disclaimer: no idea if that's still true on .NET Core.) I.E. http://stackoverflow.com/a/3782556/377476

I'd be fine doing that - adds a barely noticeable amount of overhead at the time the mailbox is created in exchange for better performance gains on X64 workloads (our HA users in particular) while maintaining the same performance on X86.

PR #2206, opened 2016-07-21
Fixed instantiation of default internal stash overflow strategy

cconstantin:


Danthar:
perf-tests failed due to akka-streams assertion failing, Restarted.

PR #2207, opened 2016-07-22
Using supported IsEmtpy for non-mono UnboundedMailboxQueue.

Scooletz:
For queues that are longer than one segment (32 messages) this may reduce the check.


Danthar:
LGTM

PR #2210, opened 2016-07-24
Check for empty message in Envelope class

alexvaluyskiy:
I've also removed all getters in the properties

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -101,6 +101,7 @@ namespace Akka.Actor
         public void ReserveChild(string name) { }
         public void Restart(System.Exception cause) { }
         public void Resume(System.Exception causedByFailure) { }
+        public virtual void SendMessage(Akka.Actor.Envelope message) { }
         public virtual void SendMessage(Akka.Actor.IActorRef sender, object message) { }
         public void SendSystemMessage(Akka.Dispatch.SysMsg.ISystemMessage systemMessage) { }
         protected void SetActorFields(Akka.Actor.ActorBase actor) { }
@@ -646,8 +647,10 @@ namespace Akka.Actor
     }
     public struct Envelope
     {
-        public object Message { get; set; }
-        public Akka.Actor.IActorRef Sender { get; set; }
+        public Envelope(object message, Akka.Actor.IActorRef sender, Akka.Actor.ActorSystem system) { }
+        public Envelope(object message, Akka.Actor.IActorRef sender) { }
+        public object Message { get; }
+        public Akka.Actor.IActorRef Sender { get; }
         public override string ToString() { }
     }
     public class EventStreamActor : Akka.Actor.ActorBase
@@ -4915,6 +4918,7 @@ namespace Akka.Util
     {
         public static bool Implements<T>(this System.Type type) { }
         public static bool Implements(this System.Type type, System.Type moreGeneralType) { }
+        public static string TypeQualifiedName(this System.Type type) { }
     }
     public class static Vector
     {
--- a/src/core/Akka.FSharp.Tests/ApiTests.fs
+++ b/src/core/Akka.FSharp.Tests/ApiTests.fs
@@ -121,8 +121,8 @@ let ``actor that accepts _ will receive string message`` () =
     response
     |> equals "SomethingToReturn"
 
-[<Fact>]
-// SUCCEEDS
+//[<Fact>]
+// FAILS
 let ``actor that accepts unit will receive unit message`` () =    
     let timeoutConfig =
         """
--- a/src/core/Akka/Actor/ActorCell.cs
+++ b/src/core/Akka/Actor/ActorCell.cs
@@ -296,7 +296,7 @@ namespace Akka.Actor
             }
         }
 
-        public virtual void SendMessage(IActorRef sender, object message)
+        public virtual void SendMessage(Envelope message)
         {
             if (Mailbox == null)
             {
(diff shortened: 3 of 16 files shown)


src/core/Akka/Actor/ActorCell.cs
```diff
@@ -308,33 +308,31 @@ public virtual void SendMessage(IActorRef sender, object message)
...
-
+                    Serializer serializer = _systemImpl.Serialization.FindSerializerFor(message.Message);
+                    byte[] serialized = serializer.ToBinary(message.Message);
 
                     var manifestSerializer = serializer as SerializerWithStringManifest;
                     if (manifestSerializer != null)
                     {
                         var manifest = manifestSerializer.Manifest(serialized);
-                        message = _systemImpl.Serialization.Deserialize(serialized, manifestSerializer.Identifier, manifest);
+                        message = new Envelope(_systemImpl.Serialization.Deserialize(serialized, manifestSerializer.Identifier, manifest), message.Sender);
```
alexvaluyskiy: Scala version has a bit different implementation
https://github.com/akka/akka/blob/4acc1cca6a27be0ff80f801de3640f91343dce94/akka-actor/src/main/scala/akka/actor/dungeon/Dispatch.scala#L138

I should check for null, and then throw an exception

src/core/Akka.FSharp.Tests/ApiTests.fs
```diff
@@ -121,8 +121,8 @@ let ``actor that accepts _ will receive string message`` () =
     response
     |> equals "SomethingToReturn"
 
-[<Fact>]
-// SUCCEEDS
+//[<Fact>]
+// FAILS
```
alexvaluyskiy: I've discussed about this test with @Horusiath . He told me, that I should remove this test if it is not actual at the moment

src/core/Akka/Actor/Message.cs
```diff
@@ -5,26 +5,43 @@
...
+            if (message == null)
+            {
+                throw new ArgumentNullException(nameof(message));
+            }
+
+            Message = message;
+            Sender = sender != ActorRefs.NoSender ? sender : system.DeadLetters;
+        }
+
+        public Envelope(object message, IActorRef sender)
```
alexvaluyskiy: I should have made this constructor as internal, but I did not manage to do it. I've got some conflicts in Persistence

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -101,6 +101,7 @@ namespace Akka.Actor
         public void ReserveChild(string name) { }
         public void Restart(System.Exception cause) { }
         public void Resume(System.Exception causedByFailure) { }
+        public virtual void SendMessage(Akka.Actor.Envelope message) { }
```
Aaronontheweb: Do we need this? Why not have the other `SendMessage` method construct the envelope?

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -646,8 +647,10 @@ namespace Akka.Actor
     }
     public struct Envelope
     {
-        public object Message { get; set; }
-        public Akka.Actor.IActorRef Sender { get; set; }
+        public Envelope(object message, Akka.Actor.IActorRef sender, Akka.Actor.ActorSystem system) { }
```
Aaronontheweb: I like this change a lot. Making the setters private and using a constructor instead.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Aaronontheweb: Ah, I see now down below why that is....

alexvaluyskiy: I did the same what Akka JVM has

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -296,7 +296,7 @@ public void UseThreadContext(Action action)
             }
         }
 
-        public virtual void SendMessage(IActorRef sender, object message)
+        public virtual void SendMessage(Envelope message)
         {
             if (Mailbox == null)
```
alexvaluyskiy: What is that? Akka JVM doesn't have this check

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -305,36 +305,41 @@ public virtual void SendMessage(IActorRef sender, object message)
...
 
-            var m = new Envelope
+                Dispatcher.Dispatch(this, message);
+            }
+            catch (Exception e)
             {
-                Sender = sender,
-                Message = message,
-            };
+                _systemImpl.EventStream.Publish(new Error(e, _self.Parent.ToString(), ActorType, "Swallowing exception during message send"));
```
alexvaluyskiy: Added missed exception handler

src/core/Akka/Util/TypeExtensions.cs
```diff
@@ -37,6 +38,16 @@ public static bool Implements(this Type type, Type moreGeneralType)
         {
             return moreGeneralType.IsAssignableFrom(type);
         }
+
+        /// <summary>
+        /// Utility to be used by implementors to create a manifest from the type.
+        /// The manifest is used to look up the type on deserialization.
+        /// </summary>
+        /// <returns>Returns the type qualified name including namespace and assembly, but not assembly version.</returns>
+        public static string TypeQualifiedName(this Type type)
```
alexvaluyskiy: We have the same method in Persistence. But we need it everywhere, not only in serializers

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -305,36 +305,43 @@ public virtual void SendMessage(IActorRef sender, object message)
...
-                        message = _systemImpl.Serialization.Deserialize(serialized, serializer.Identifier, message.GetType().AssemblyQualifiedName);
+                        Serializer serializer = _systemImpl.Serialization.FindSerializerFor(unwrapped);
+                        byte[] serialized = serializer.ToBinary(unwrapped);
+
+                        var manifestSerializer = serializer as SerializerWithStringManifest;
+                        if (manifestSerializer != null)
+                        {
+                            var manifest = manifestSerializer.Manifest(unwrapped);
+                            var deserialized = _systemImpl.Serialization.Deserialize(serialized, serializer.Identifier, manifest);
+                            message = new Envelope(deserialized, message.Sender);
```
alexvaluyskiy: Some tests don't work without this. But scala code doesn't have this assignment

PR #2213, opened 2016-07-26
Replaces fields and methods signatures to allow overriding the Snapshot/QueryExecutor class

ismaelhamed:
While working on a persistent driver for Oracle, I found the need to override some of the methods in the Snapshot/QueryExecutor class. This change is consistent with what the Journal/QueryExecutor already does.


Aaronontheweb:
Re-ran the perf specs. Nothing in this PR appears to affect them given that they're in a separate library.

Issue #2153, opened 2016-07-07
NetCore meta-issue / Status


alexvaluyskiy:
**What has been done**
- Converted all projects to the new MSBuild 15 (Visual Studio 2017 required)
- Added support for `netstandard1.6` for all core modules
- Switched all old `protobuf2` serialization for remoting to the new `protobuf3`
- All test projects target `net452` and `netcoreapp1.1`
- Rewrote all build scripts

**Critical**
- [x] Akka.IO - currently is not present in `netstandard1.6` build, Tests are commented on all platforms. Waiting for (https://github.com/akkadotnet/akka.net/pull/2405)
- [x] NetCore nightly builds on MyGet (https://github.com/akkadotnet/akka.net/issues/2755)

**In progress**
- [ ] Fix FsCheck tests on for Akka.Remote project (https://github.com/akkadotnet/akka.net/issues/2827)
- [ ] NetCore in FSharp projects (https://github.com/akkadotnet/akka.net/issues/2826)
- [x] Remove protobuf2 serialization from Persistence (https://github.com/akkadotnet/akka.net/pull/2748)
- [x] Akka.DI.Core and Akka.DI.TestKit should support netcore (https://github.com/akkadotnet/akka.net/pull/2732)
- [x] Akka.Persistence.Sqlite should support netcore (https://github.com/akkadotnet/akka.net/issues/2682)
- [x] Switch to protobuf3 in `Akka.Remote.Testkit` (https://github.com/akkadotnet/akka.net/issues/2664)
- [x] Fix FsCheck tests on for Akka project
- [x] Fix OutOfMemoryException for Windows Unit Tests CI task (https://github.com/akkadotnet/akka.net/issues/2658)
- [x] Fix unit tests on Linux (both Mono and NetCoreApp1.1) (https://github.com/akkadotnet/akka.net/issues/2659)
- [x] Run NBench tests on Windows for .NET (https://github.com/akkadotnet/akka.net/issues/2660)
- [x] Fix tests in Akka.Streams.Tests  (https://github.com/akkadotnet/akka.net/issues/2663)

**Tests**
- [ ] Run MultiNode tests for `netstandard1.6` target on Windows and Linux
- [ ] Run NBench tests for `netstandard1.6` target on Windows and Linux (https://github.com/akkadotnet/akka.net/pull/2768)
- [ ] Fix FsCheck tests on for Akka.Remote project (https://github.com/akkadotnet/akka.net/issues/2827)
[...]

PR #2215, opened 2016-07-26
Moving towards NetStandard

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/issues/2153


Aaronontheweb:
To be clear, the build process for this system WILL NOT RUN right now. Need to add dotnetcore support to the build agents to do that.

adamhathcock:
I haven't read everything as it's late in my day but just wanted to ask: do you think you should make the libraries target as low as possible (like standard 1.3) instead of 1.6?

It practically may not matter because you have net45 there and 1.3 is net451 but I think best practice is going as low as possible.

The PCL compatibility package is new to me so I should use that.  Thanks :)

src/core/Akka.Tests/Actor/Dispatch/ActorModelSpec.cs
```diff
@@ -218,7 +218,16 @@ public DispatcherActor()
...
                 Receive<Interrupt>(interrupt => { Ack(); Sender.Tell(new Status.Failure(new ActorInterruptedException(cause: new Exception(Ping)))); _busy.SwitchOff(); throw new Exception(Ping); }, interrupt => true);
-                Receive<InterruptNicely>(interrupt => { Ack(); Sender.Tell(interrupt.Expect); _busy.SwitchOff(); Thread.CurrentThread.Interrupt(); });
+                Receive<InterruptNicely>(interrupt =>
+                {
+                    Ack();
+                    Sender.Tell(interrupt.Expect);
+                    _busy.SwitchOff();
+#if !CORECLR
+                    // TODO: CORECLR FIX IT
+                    Thread.CurrentThread.Interrupt();
```
alexvaluyskiy: Netcore doesn't have Thread.Interupt

src/core/Akka/Actor/Props.cs
```diff
@@ -763,7 +764,12 @@ private DynamicProps(Props copy, Func<TActor> invoker)
         protected override Props Copy()
         {
             Props initialCopy = base.Copy();
+#if CLONEABLE
             var invokerCopy = (Func<TActor>)invoker.Clone();
+#else
+            // TODO: CORECLR FIX IT
+            var invokerCopy = (Func<TActor>)invoker;
```
alexvaluyskiy: Should fix it

annymsMthd:
We should probably have these commits squashed

Aaronontheweb:
@alexvaluyskiy go ahead and squash these and we can merge them in.

PR #2217, opened 2016-07-27
Align SupervisorStrategy with scala version

alexvaluyskiy:
- added an abstract IDecider property to an abstract SupervisorStrategy class
- added "child" parameter to ProcessFailure method

### DIFF
--- a/src/core/Akka/Actor/ActorCell.FaultHandling.cs
+++ b/src/core/Akka/Actor/ActorCell.FaultHandling.cs
@@ -385,7 +385,7 @@ namespace Akka.Actor
                 var statsUid = childStats.Child.Path.Uid;
                 if (statsUid == f.Uid)
                 {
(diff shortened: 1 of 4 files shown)


src/core/Akka/Actor/ActorCell.FaultHandling.cs
```diff
@@ -385,7 +385,7 @@ private void FinishRecreate(Exception cause, ActorBase failedActor)
                 var statsUid = childStats.Child.Path.Uid;
                 if (statsUid == f.Uid)
                 {
-                    var handled = _actor.SupervisorStrategyInternal.HandleFailure(this, f.Cause, childStats, ChildrenContainer.Stats);
+                    var handled = _actor.SupervisorStrategyInternal.HandleFailure(this, failedChild, f.Cause, childStats, ChildrenContainer.Stats);
```
alexvaluyskiy: I don't like this part. Why not just `_actor.SupervisorStrategy`?

alexvaluyskiy:
@akkadotnet/core ready for review

Danthar:
LGTM

PR #2219, opened 2016-07-27
[api-docs] Documenting Akka exceptions

sean-gilliam:
This PR documents the exceptions thrown in the various areas of the main
Akka project (Akka.csproj)


Aaronontheweb:
@sean-gilliam looks like an assertion failure here, but a simple one:

```
Assert.Throws() Failure
Expected: typeof(System.Exception)
Actual:   typeof(System.InvalidOperationException): ActorSystem already terminated.
   at Akka.Actor.Internal.TerminationCallbacks.Add(Action code) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka\Actor\Internal\ActorSystemImpl.cs:line 461
```

PR #2222, opened 2016-07-28
Fixed #2221 (DistributedPubSubMediator - an error in HandlePrune)

kantora:
Fixed the error in DistributedPubSubMediator.HandlePrune (https://github.com/akkadotnet/akka.net/issues/2221)

Issue #2221, opened 2016-07-28
DistributedPubSubMediator - an error in HandlePrune


kantora:
I've just tried to implement the cluster pub/sub and failed :(
Getting an error like:
```
worker_1            | 2016-07-28 07:11:50 [Error] "An item with the same key has already been added."
worker_1            | System.ArgumentException: An item with the same key has already been added.
worker_1            |   at System.ThrowHelper.ThrowArgumentException (ExceptionResource resource) <0x421d9c00 + 0x00027> in <filename unknown>:0
worker_1            |   at System.Collections.Generic.Dictionary`2[TKey,TValue].Insert (System.Collections.Generic.TKey key, System.Collections.Generic.TValue value, Boolean add) <0x415b2e20 + 0x0017b> in <filename unknown>:0
worker_1            |   at System.Collections.Generic.Dictionary`2[TKey,TValue].Add (System.Collections.Generic.TKey key, System.Collections.Generic.TValue value) <0x415b79a0 + 0x0001b> in <filename unknown>:0
worker_1            |   at Akka.Cluster.Tools.PublishSubscribe.DistributedPubSubMediator.HandlePrune () <0x421d7240 + 0x00364> in <filename unknown>:0
worker_1            |   at Akka.Cluster.Tools.PublishSubscribe.DistributedPubSubMediator.<.ctor>b__14_14 (Akka.Cluster.Tools.PublishSubscribe.Internal.Prune _) <0x421d7200 + 0x0000f> in <filename unknown>:0
[...]

PR #2225, opened 2016-07-29
DistributedPubSubMediator - an error in HandlePrune, Fix of fix in #2222 (removed multiple enumeration)

kantora:
Sorry, I've replaced one bug with another. This should help.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
@@ -521,6 +521,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
 
         private void HandlePrune()
         {
+            var modifications = new Dictionary<Address, Bucket>();
             foreach (var entry in _registry)
             {
                 var owner = entry.Key;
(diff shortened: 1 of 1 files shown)


kantora:
Hmmm... There is no more exception here, but it still works wrong... Now it silently drops active subscriptions.

kantora:
The original pull request: https://github.com/akkadotnet/akka.net/pull/2222
The original issue: https://github.com/akkadotnet/akka.net/issues/2221

alexvaluyskiy:
@kantora I can't merge this PR, because we don't have any test for "HandlePrune" neither in MNTK tests nor in unit tests. Could you create the one. If not, just share some examples of code, which can help me to reproduce this bug

kantora:
@alexvaluyskiy As for me, it reproduces just each time cluster pub/sub is used, if you give it at least 2 or 3 minutes to run. And, to be honest, this pr just hides the problem (as it prevents the exception to appear) and don't fixes it. I've described the main issue in https://github.com/akkadotnet/akka.net/issues/2226.

As for this particular PR - the fix is obvious from code as the previous version modified dictionary while was iterating it and fixes the recently merged PR https://github.com/akkadotnet/akka.net/pull/2222.

alexvaluyskiy:
@kantora I've tested your code, it really fixed the problem. But we shouldn't test our code manually. So we should create a test for it. I've created an issue for this case
https://github.com/akkadotnet/akka.net/issues/2232

Issue #2201, opened 2016-07-18
Akka.Serialization.Wire broken for Wire 0.7.1 due to namespace change.


Inverness:
The _converters_ namespace in Wire 0.7.1 was renamed, this breaks Akka.Serialization.Wire due to the constructor reference no longer being valid:
```
System.Reflection.TargetInvocationException was unhandled by user code
  HResult=-2146232828
  Message=Exception has been thrown by the target of an invocation.
  Source=mscorlib
  StackTrace:
       at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
       at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
       at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes, StackCrawlMark& stackMark)
       at System.Activator.CreateInstance(Type type, BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes)
       at System.Activator.CreateInstance(Type type, Object[] args)
       at Akka.Serialization.Serialization..ctor(ExtendedActorSystem system)
       at Akka.Actor.Internal.ActorSystemImpl..ctor(String name, Config config)
       at Akka.Actor.ActorSystem.Create(String name, Config config)
       at Petrolink.Analytix.Akka.AnalytixApp.InitializeConfiguration(RabbitConfiguration queueConfig, Config akkaConfig, RestConfiguration chronosConfiguration) in C:\Repos\PetroVaultHD_AnalytixCore\src\dotnet\Petrolink.Analytix.Akka\AnalytixApp.cs:line 70
       at Petrolink.Analytix.Akka.Program.OnAnalytixAppStarted(AnalytixApp service, HostControl hostControl) in C:\Repos\PetroVaultHD_AnalytixCore\src\dotnet\Petrolink.Analytix.Akka\Program.cs:line 97
       at Topshelf.Builders.DelegateServiceBuilder`1.DelegateServiceHandle.Start(HostControl hostControl)
       at Topshelf.Hosts.ConsoleRunHost.Run()
  InnerException: 
       HResult=-2146233069
[...]

PR #2233, opened 2016-08-02
Updated Wire to 0.7.1

alexvaluyskiy:
fixed https://github.com/akkadotnet/akka.net/issues/2201
@rogeralsing are we ready for this PR?


Danthar:
lgtm. If a new version comes up that needs to be integrated, we will just do another pr.

PR #2236, opened 2016-08-03
Fixed ClusterShardingMessageSerializer

alexvaluyskiy:
Fixed ClusterShardingMessageSerializer and added ClusterShardingMessageSerializerSpec

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using Akka.Actor;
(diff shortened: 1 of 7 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
```diff
@@ -133,6 +134,53 @@ public State Updated(IDomainEvent e)
 
                 return new State(shards ?? Shards, regions ?? Regions, regionProxies ?? RegionProxies, unallocatedShards ?? UnallocatedShards);
             }
+
+            #region Equals
+
+            public override bool Equals(object obj)
```
alexvaluyskiy: Not the best my equality comparer. We should compare objects in a different way

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
```diff
@@ -133,6 +134,53 @@ public State Updated(IDomainEvent e)
...
+                if (ReferenceEquals(other, null)) return false;
+                if (ReferenceEquals(other, this)) return true;
+
+                return Shards.SequenceEqual(other.Shards)
+                    && Regions.Keys.SequenceEqual(other.Regions.Keys)
+                    && RegionProxies.SequenceEqual(other.RegionProxies)
+                    && UnallocatedShards.SequenceEqual(other.UnallocatedShards);
+            }
+
+            public override int GetHashCode()
```
alexvaluyskiy: I don't like this GetHashCode

Danthar: The fixed integer values are a smell, and a problem if the same gethashcode is copy pasted everywhere. But other then that its not like we need encryptionlevel hashcodes here.

Danthar:
needs rebase.

Issue #2226, opened 2016-07-29
Cluster.Tools Pub/Sub removes active subscriptions


kantora:
The history of the problem:
https://github.com/akkadotnet/akka.net/issues/2221
https://github.com/akkadotnet/akka.net/pull/2222
https://github.com/akkadotnet/akka.net/pull/2225

In my case, I have multiple subscribers and rare publishes (I use only `Publish`/`Subscribe`). But almost always (not every time) in 120s after node start (this is the default prune timeout) `DistributedPubSubMediator.HandlePrune` is called and removes everything from `_registry` (I've fixed the obvious bugs, so now it does not throw exceptions and just removes what it was intended to remove).

kantora:
@alexvaluyskiy emmm... the problem was not solved, I've just removed the exception, but subscriptions are still removed in 120s after node start...

alexvaluyskiy:
@kantora I've found something. We are using Tick in NextVersion()
https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs#L590

But we are comparing with Milliseconds in HandlePrune();
https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs#L531

Scala version uses Milliseconds in the both places:
https://github.com/akka/akka/blob/master/akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala#L509
https://github.com/akka/akka/blob/master/akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala#L816

alexvaluyskiy:
@kantora could you check it?

kantora:
@alexvaluyskiy I'll try to do it in the beginning of the next week.

alexvaluyskiy:
@kantora The fix has been merged. I tried it with my example, and it worked fine. You should try on your example too.

kantora:
@alexvaluyskiy as I see from the commit, you left the problem with the midnight (after the midnight all new versions will be obsolete).
I suppose that `var current = DateTime.UtcNow.Ticks / 10000;` is much better.

alexvaluyskiy:
[...]

Issue #2232, opened 2016-08-02
Create a spec for Pruning in DistributedPubSub


alexvaluyskiy:
We had some errors in HandlePrune method, but this method is the one, which is not covered by our unit and integration tests. So we should create it. Something like that

``` csharp
        [Fact]
        public void DistributedPubSubmediator_must_handle_prune_correctly()
        {
            IActorRef registryActor = Sys.ActorOf(Props.Create<EchoActor>());

            TimeSpan pruneInterval = 10.Seconds();
            var settings = new DistributedPubSubSettings("", new RandomLogic(), 1.Seconds(), pruneInterval, 3000);
            var mediator = Sys.ActorOf(DistributedPubSubMediator.Props(settings));
            mediator.Tell(new Put(registryActor));

            mediator.Tell(Count.Instance);
            ExpectMsg<int>().Should().Be(1);

            mediator.Tell(Prune.Instance);

            mediator.Tell(Count.Instance);
            ExpectMsg<int>().Should().Be(0);
        }
```

kantora:
As I see, some additional checks should be added to the `DistributedPubSubMediatorSpecs`.

alexvaluyskiy:
Yes

PR #2242, opened 2016-08-07
Fix pruning in Distributed PubSub

alexvaluyskiy:
Fixed this bug: https://github.com/akkadotnet/akka.net/issues/2226


Danthar:
lgtm. Should eventually be better covered by https://github.com/akkadotnet/akka.net/issues/2232

PR #2244, opened 2016-08-08
Marked some methods as virtual in Journal/QueryExecutor

ismaelhamed:
These changes are related to the Pull Request #2213

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
@@ -428,7 +428,7 @@ namespace Akka.Persistence.Sql.Common.Journal
             command.Parameters.Add(parameter);
         }
 
-        public async Task SelectEventsAsync(DbConnection connection, CancellationToken cancellationToken, IEnumerable<IHint> hints, Action<IPersistentRepresentation> callback)
+        public virtual async Task SelectEventsAsync(DbConnection connection, CancellationToken cancellationToken, IEnumerable<IHint> hints, Action<IPersistentRepresentation> callback)
         {
             using (var command = GetCommand(connection, QueryEventsSql))
             {
@@ -445,7 +445,7 @@ namespace Akka.Persistence.Sql.Common.Journal
             }
         }
 
-        private string HintToSql(IHint hint, DbCommand command)
+        protected virtual string HintToSql(IHint hint, DbCommand command)
         {
             if (hint is TimestampRange)
             {


ismaelhamed:
@alexvaluyskiy while implementing a persistence plugin for Oracle, I needed to override a bunch of methods in both QueryExecutors (journal and snapshot).

alexvaluyskiy:
But you can see three implementation of SQL plugins: mysql, postgres and sqlserver. And they don't need to override these methods. What is really do you want to do in these methods?

ismaelhamed:
Well mostly, Oracle uses a colon (:name) instead of the "at" symbol (@name) for named parameters. Just for this one you have to override most of the SQL statements in the SqlCommon dll. 

But there were some more oddities like:

- Lack of a proper Boolean type, which requires you to use a char or number, and make a cast later on the code (IPersistentRepresentation).
- The need to force an order in SelectEventsAsync: otherwise some unit tests fail due to the way Oracle returns rows when no ORDER BY is specified (QueryEventsSql)
- ExecuteScalarAsync returns a decimal, not a long, which again force you to override a couple methods to do the right cast.

The driver for Oracle is completed (including support for Persistence Query), working, and passing the tests, but right now it requires a precompiled version of Akka.Persistence.Sql.Common.dll with all these changes. It’s in my repositories, you can take a look.

Horusiath:
- Regarding ordering returned rows: I think we should apply *ORDER BY persistence_id, sequence_nr* by default. This is also a problem @alexvaluyskiy seen in postgres plugin.
- "ExecuteScalarAsync returns a decimal" - isn't that dependent on the underlying provider? I know this might be a problem in Orlace plugin, but shouldn't be a problem in Devart i.e.

ismaelhamed:
All changes I’ve made to Akka.Persistence.Sql.Common in both PRs just involve marking certain methods as virtual, so that all these issues can be dealt with at the plugin level.

to11mtm:
@Horusiath The Decimal return is somewhat dependent on the underlying provider. My for Oracle plugin to get best results I had to define Number mappings for the devart provider, and instead of just doing select MAX(sequencenr) doing CAST(MAX(sequencenr) as NUMBER(18,0)). I believe that you can do similar with ODP.NET (Definitely can when dealing with EF).

It would have been nice for all the methods to be virtual, I wound up having to explicitly re-implement the interface and relevant methods to get everything working. The way you have to do things in oracle tends to be quite different, i.e. @ismaelhamed does his upserts very different than I do, and yet neither look a whole lot like how you would do so in a different DB, and yet they're probably the two most accepted ways to do it in Oracle.

Aaronontheweb:
@Horusiath given @to11mtm's comments, anything fundamentally "wrong" with accepting this pull request once the test suite passes?

Danthar:
Given that its reviewed positively. Im guessing it needs a rebase before it can be merged ?

ismaelhamed:
Github's "Update branch" button was so neat I neglected to rebase instead. Should I amend this PR? Whats the best way of getting rid of all those merges?

Aaronontheweb:
@ismaelhamed here's what I do:

`git fetch akkadotnet`
`git checkout dev`
`git merge --ff-only akkadotnet/dev`
`git checkout {your feature branch}`
`git rebase -i dev`
`git push origin --force {your feature branch}`

ismaelhamed:
@Aaronontheweb I wasn't sure about the `git rebase -i dev` so I did a `git reset --hard` instead (it was easier for me to reason about) and then rebased onto dev (which is what I should've done in the first place).

ismaelhamed:
Since this PR goes along with the #2213 that is already set to be release in the 1.1.2 milestone, I think this should be too.

PR #2248, opened 2016-08-10
Exposed API for for materialized graph builder and at-least-once-delivery semantic messages

Horusiath:
- New method: `Graph.CreateMaterialized` which allows to use GraphDsl.Builder with custom generic type parameter. This is necessary, because currently there is no way to use Graph API returning graphs with custom materializer type. 
- Changed signatures of internal `AtLeastOnceDeliverySemantic` messages to public. This is also necessary, since using this object as separate component (instead of internal field of `AtLeastOnceDeliveryActor`) causes problems when internal messages, such as `RedeliveryTick`, arrive to user actors.


Aaronontheweb:
@Horusiath needs a rebase my friend

Silv3rcircl3:
@Horusiath needs another rebase

PR #2251, opened 2016-08-11
End-to-end Mono compilation [WIP]

Aaronontheweb:


alexvaluyskiy:
You have to change these paths too
https://github.com/akkadotnet/akka.net/blob/dev/build.fsx#L226

Aaronontheweb:
@alexvaluyskiy yeah,  there's a bunch of stuff in the FAKE file that needs to be updated

Aaronontheweb:
In `build.sh`:

```
mono $SCRIPT_PATH/.nuget/nuget.exe install xunit.runners -OutputDirectory $SCRIPT_PATH/packages/FAKE -ExcludeVersion -Version 2.0.0
```

Needs to be updated to `xunit.runner.console`, which is what we have in `build.cmd`

alexvaluyskiy:
```
mono $SCRIPT_PATH/src/.nuget/NuGet.exe install NUnit.Console -OutputDirectory $SCRIPT_PATH/src/packages/FAKE -ExcludeVersion -Version 3.2.1
mono $SCRIPT_PATH/src/.nuget/NuGet.exe install xunit.runner.console -OutputDirectory $SCRIPT_PATH/src/packages/FAKE -ExcludeVersion -Version 2.0.0
```

alexvaluyskiy:
And I think Sqlite nuget package doesn't support Mono. So we should exclude it
And... we should change the paths of test results, because it puts all files in the root folder, instead of "TestResults" folder

Aaronontheweb:
@alexvaluyskiy I added a method inside FAKE (on NBench) for being able to skip platform-specific unit tests and benchmarks: 

https://github.com/petabridge/NBench/blob/dev/build.fsx#L37

and

https://github.com/petabridge/NBench/blob/dev/build.fsx#L114

I'll use that same approach here.

Aaronontheweb:
As for the pathing stuff, in some of those cases we just need to be using the `Path` class and some of the built-in FileHelpers in FAKE - they do a decent job of handling that in a platform-independent way. I'll go through and start testing more of that. First things first though - gotta squash some Helios bugs on Mono.

Aaronontheweb:
Waiting on #2294 before moving onto the next phases of this. Resolves a resource leak issue that blows up the Mono test runner.

Danthar:
Since #2294 has been merged. Any news on this ?

Aaronontheweb:
@Danthar we can close this

PR #2253, opened 2016-08-12
[WIP] Cluster Tools technical debt

alexvaluyskiy:
Did not change anything, just cleaned the API. Also I've updated *.DotSettings file, and added default *.ruleset file

### DIFF
--- a/src/Akka.ruleset
+++ b/src/Akka.ruleset
@@ -1,0 +1,60 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Rules for Akka.Net" Description="Code analysis rules for Akka.Net solution." ToolsVersion="14.0">
+  <Rules AnalyzerId="SonarAnalyzer.CSharp" RuleNamespace="SonarAnalyzer.CSharp">
+    <Rule Id="S100" Action="None" />
+    <Rule Id="S2357" Action="Warning" />
+    <Rule Id="S2360" Action="None" />
+    <Rule Id="S2551" Action="Warning" />
+  </Rules>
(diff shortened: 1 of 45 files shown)


alexvaluyskiy:
I have plans to change namespaces
Akka.Cluster.Tools.Singleton => Akka.Cluster.Singleton
Akka.Cluster.Tools.Client => Akka.Cluster.Client
Akka.Cluster.Tools.PublishSubscribe => Akka.Cluster.PubSub or (Akka.Cluster.PublishSubscribe)

and move all public PubSub messages under DistributedPubSubMediator class, like JVM did
@Aaronontheweb @Horusiath what do you think?

Horusiath:
Regarding changing namespaces - great! :) 
Regarding moving messages under mediator - it will be irritating to prefix every message with `DistributedPubSubMediator` (think about other .NET languages here as well).

alexvaluyskiy:
@Horusiath so Akka.Cluster.PubSub or Akka.Cluster.PublishSubscribe? Since we have a class DistributedPubSubMediator, we should use PubSub

> Regarding moving messages under mediator - it will be irritating to prefix every message with DistributedPubSubMediator
But we have all public message under ClusterClient. We should have consistency here.

Aaronontheweb:
I have no idea what happened with the MNTR on this run... Looks like something faulted early on in the test run and broke 100% of the specs that came after it... Process that failed to terminate or something else like it could do it.

Aaronontheweb:
@alexvaluyskiy mind updating this PR with the latest?

alexvaluyskiy:
@Aaronontheweb only when we decide to make Cluster.Tools as a stable

Issue #2194, opened 2016-07-16
Akka.Cluster / Akka.Remote uses ipv6 instead of ipv4 in ipv4 environment


kantora:
At first. I've reported it in https://github.com/akkadotnet/akka.net/issues/2161, but suppose it is some separate bug.

Have the Docker / Ubuntu / Mono environment.
Updated to the latest Akka.Net night build (1.1.1.248-beta) and Helios 2.1.2.
There are one seed node and several worker nodes. 

Remote is configured with fixed port, `0.0.0.0` `hostname` and `public-hostname` according to the container name. All names are resolved ok into the correct ipv4 address.

Unfortunately, a cluster is not formed. I see nothing in worker node logs and sometimes errors in the seed node logs:
```
[36mseed_1 | [0m2016-07-14 16:02:48 [Error] Error caught channel ["::ffff:172.18.0.5:3090"->"::ffff:172.18.0.2:50152"] (Id="ChannelId(-1914377680)")
[36mseed_1 | [0mSystem.Net.Sockets.SocketException: Operation aborted
[36mseed_1 | [0m  at Helios.Channels.Sockets.SocketChannelAsyncOperation.Validate () <0x42187b20 + 0x00057> in <filename unknown>:0
[36mseed_1 | [0m  at Helios.Channels.Sockets.AbstractSocketByteChannel+SocketByteChannelUnsafe.FinishRead (Helios.Channels.Sockets.SocketChannelAsyncOperation operation) <0x421926a0 + 0x0013b> in <filename unknown>:0
[36mseed_1 | [0m2016-07-14 16:02:48 [Information] "Message Disassociated from NoSender to akka://ClusterKit/system/transports/akkaprotocolmanager.tcp.0/akkaProtocol-tcp%3A%2F%2FClusterKit%40%5B%3A%3Affff%3A172.18.0.2%5D%3A50152-3 was notdelivered. 3 dead letters encountered."
```

As for windows environment `::ffff:172.18.0.2` and `172.18.0.2` are the same addresses, But in linux (as for mine configuration) are not.

```
root@52e368960bf4:/opt/clusterkit# ping 172.18.0.2
PING 172.18.0.2 (172.18.0.2) 56(84) bytes of data.
64 bytes from 172.18.0.2: icmp_seq=1 ttl=64 time=0.060 ms
64 bytes from 172.18.0.2: icmp_seq=2 ttl=64 time=0.086 ms
[...]

PR #2258, opened 2016-08-21
related to #2194

andreyleskov:
Hi all. I've faced issue #2194 running Akka inside Azure WebApp. 
It is related to Ipv6 dns lookup not available in some cases.
To solve it we need to enforce ipv4 usage inside sockets used by Helious.
I've introduced new configuration variable for it, "enforce-ip-family".
According to #2194 it will be always need in Mono, so in Mono environment 
default value is true. 
Most of Akka.Net should not be affected by #2194 and will not care about new setting.

### DIFF
--- a/src/core/Akka.Remote.Tests/RemoteConfigSpec.cs
+++ b/src/core/Akka.Remote.Tests/RemoteConfigSpec.cs
@@ -101,6 +101,27 @@ namespace Akka.Remote.Tests
         }
 
         [Fact]
+        public void When_remoting_works_in_Mono_ip_enforcement_should_be_defaulted_to_true()
+        {
+            HeliosTransportSettings.IsMono = true;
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
HI @andreyleskov - this change looks good but since the commit is 9 days old, TeamCity isn't picking it up for CI. This happens all the time when folks rebase.

Would you mind running the following command to update the timestamp to today? This will allow CI to be run: http://stackoverflow.com/a/31540373

Aaronontheweb:
Otherwise, this change looks good - have something similar to this in my Mono branch but I like using the setting here to solve issues with restricted environments like Azure WebApps :p

andreyleskov:
@Aaronontheweb I've updated timestamp. Thank you for review.  Azure WebApp behavior was a real surprise for me.

Aaronontheweb:
I have no idea why it required me to use "administrator privilege" to merge in this pull request... Might be because I had merged in another PR earlier and have branch protection turned on?

vladkosarev:
Is it possible to get this in to a release soon? This issue is blocking us from using akka in azure.

PR #2260, opened 2016-08-22
Bugfix: NRE in Router constructor

tvsonar:


src/core/Akka/Routing/Router.cs
```diff
@@ -187,14 +187,15 @@ public class Router
         //We need one unique constructor to handle this call: new Router(logic). The other constructor will handle that.
         //So in order to not confuse the compiler we demand at least one ActorRef. /@hcanber
         public Router(RoutingLogic logic, IActorRef routee, params IActorRef[] routees)
-        {
-            var routeesLength = routees.Length;
```
tvsonar: If an explicit {{null}} is being passed as {{routees}}, the {{.Length}} will throw a NRE.

tvsonar:
Note: This issue was found by [SonarLint for Visual Studio](https://visualstudiogallery.msdn.microsoft.com/47d1049d-bb27-454e-aab8-24566c85e548).

src/core/Akka/Routing/Router.cs
```diff
@@ -188,13 +188,14 @@ public class Router
         //So in order to not confuse the compiler we demand at least one ActorRef. /@hcanber
         public Router(RoutingLogic logic, IActorRef routee, params IActorRef[] routees)
         {
-            var routeesLength = routees.Length;
```
tvsonar: If an explicit `null` is being passed as `routees`, the `.Length` will throw a NRE.

Aaronontheweb:
Nice catch.

PR #2261, opened 2016-08-23
Akka.Cluster.DistributedData

Horusiath:

This PR includes @bruinbrown work on ddata module with code updated by me. Basically the core module is ported - not sure if I didn't miss something, I was sitting late on that yesterday - missing part I'm going to work on is test suite (especially multinode).

NOTE: originally ddata messages use protobuf for serialization, however I've found out that we use some generics there and originally we used reflection to make serialization possible. I've erased generic types where I could, however I think, Wire would be a better choice here. @akkadotnet/core aggree?

Status:
- Core project:
  - [x] Flag
  - [x] GCounter
  - [x] GSet
  - [x] LWWDictionary
  - [x] LWWRegister
  - [x] ORDictionary
  - [x] ORMultiDictionary
  - [x] ORSet
  - [x] PNCounter
  - [x] PNCounterDictionary
  - [x] PruningState
  - [x] Replicator
  - [x] VersionVector
- Tests:
  - [x] ReplicatedDataSerializerSpec (used Wire instead of custom protobuf-based)
  - [x] ReplicatorMessageSerializerSpec (used Wire instead of custom protobuf-based)
  - [x] FlagSpecsSpec (passing)
  - [x] GCounterSpec (passing)
  - [x] GSetSpec (passing)
  - [x] LWWDictionarySpec (passing)
  - [x] LWWRegisterSpec (passing)
  - [x] ORDictionarySpec (passing)
  - [x] ORMultiDictionarySpec (passing)
  - [x] ORSetSpec (passing)
  - [x] PNCounterSpec (passing)
  - [x] PNCounterDictionarySpec (passing)
  - [x] PruningStateSpec (passing)
  - [x] VersionVectorSpec (passing)
  - [x] LocalConcurrencySpec (passing)
  - [x] WriteAggregatorSpec (passing)
  - [ ] LotsOfDataBot - to be implemented as part of performance spec
- MultiNode tests - can be implemented in a separate PR:
  - [x] JepsenInspiredInsertSpec (ported, not verified)
  - [ ] PerformanceSpec
  - [x] ReplicatorSpec (ported, not verified)
  - [x] ReplicatorChaosSpec (ported, not verified)
  - [x] ReplicatorPruningSpec (ported, not verified)


Horusiath:
All tests pass locally. Once our CI will prove that, it should be safe to merge. Remaining work, that still needs to be done:

- Port and fix MNTK spec.
- Port LotsOfDataBot into some sort of NBench performance spec.
- Performance optimizations.
- Serialization tests - currently tests are only asserting local replication. We need to check if replicated collections are serialized/deserialized correctly.
- Replicated collections API - right now it's almost 1-1 copy of JVM model, with some changes mostly in naming conventions. I think, that replicated collections should follow design closer to what we have in `System.Collections.Immutable` (i.e. extracting ORSet and GSet methods into some kind of `IReplicatedSet` interface). Also pretty much everything is exposed public right now. We need to check what needs to be internal or private.

Aaronontheweb:
Need to approve the API changes made to Akka.Cluster.

Horusiath:
I don't know why it broke on the MNTK specs. @Aaronontheweb could you verify?

alexvaluyskiy:
Will you plan to create tests for protobuf serializer? Like this: https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Sharding.Tests/ClusterShardingMessageSerializerSpec.cs

alexvaluyskiy:
It would be better to make all protobuff serializer classes as internal

Horusiath:
I need to make sure, that JVM protobuf definitions can be used here. JVM version of replicated collection assumes, that all dictionaries use string keys. I've assumed, that they may be generic. Also bruinbrown introduced collection keys, which are generic types - I don't know how protobuf will behave in this situation. 

Those were the reasons, why I opted to use Wire instead.

Horusiath:
UPDATE: I've implemented serialization specs - they pass under Wire akka serializer. My proposition is to stay with that, and if necessary (and it may be necessary, as custom serializer provided compression and optimizations), make custom serializer based on Wire anyway.

Horusiath:
UPDATE: I've fixed failing tests and added `Akka.DistributedData.Local` versions for each CRDT - they are surrogated wrappers that always operate in the context of the current cluster. This way we may avoid passing node address each time we want to update current CRDT.

Aaronontheweb:
Nice work @Horusiath !

Issue #2274, opened 2016-08-28
Config.GetStringList : Collection was not of fixed size


mmisztal1980:
# Description
When reading a configuration containing an empty list of seed nodes using ```Config.GetStringList``` I was trying to inject an item into the list and got a 

>Collection was not of fixed size

# Expected
```Config.GetStringList``` 's signature and name imply that an ```IList<string>``` is returned, which is true, however in this case an empty array was returned, preventing me from adding any additional items to the list. I don't believe it should be that way - this prevents users from injecting items to the configuration in a way which is done in Lighthouse (for example).

I'll be making a PR regarding this today.

Danthar:
PR was merged

PR #2275, opened 2016-08-28
Issue #2274

mmisztal1980:
This is related to issue #2274 

```Config.GetStringList``` returns an empty array instead of an empty list


Danthar:
Good point (see https://github.com/akkadotnet/akka.net/issues/2274).
PR LGTM.

Issue #2271, opened 2016-08-26
Cluster PubSub with Wire serialization drops cluster nodes down


kantora:
I have a cluster with several nodes, each of them contains actors that subscribe to the same topics.
After start, I receive these exceptions in the log:

```
	Akka.Remote.EndpointException: Failed to write message to the transport ---> Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: `Wire.SerializerFactories.ArraySerializerFactory.WriteValues<Akka.Cluster.Tools.PublishSubscribe.Internal.Bucket>(System.Collections.Generic.IReadOnlyList<Akka.Cluster.Tools.PublishSubscribe.Internal.Bucket>, System.IO.Stream, System.Type, Wire.ValueSerializers.ValueSerializer, Wire.SerializerSession)' is inaccessible due to its protection level
  at (wrapper dynamic-method) System.Object:CallSite.Target (System.Runtime.CompilerServices.Closure,System.Runtime.CompilerServices.CallSite,Wire.SerializerFactories.ArraySerializerFactory,object,System.IO.Stream,System.Type,Wire.ValueSerializers.ValueSerializer,Wire.SerializerSession)
  at System.Dynamic.UpdateDelegates.UpdateAndExecuteVoid6[T0,T1,T2,T3,T4,T5] (System.Runtime.CompilerServices.CallSite site, System.Dynamic.T0 arg0, System.Dynamic.T1 arg1, System.Dynamic.T2 arg2, System.Dynamic.T3 arg3, System.Dynamic.T4 arg4, System.Dynamic.T5 arg5) <0x40a54550 + 0x00655> in <filename unknown>:0 
  at (wrapper delegate-invoke) System.Action`7[System.Runtime.CompilerServices.CallSite,Wire.SerializerFactories.ArraySerializerFactory,System.Object,System.IO.Stream,System.Type,Wire.ValueSerializers.ValueSerializer,Wire.SerializerSession]:invoke_void_T1_T2_T3_T4_T5_T6_T7 (System.Runtime.CompilerServices.CallSite,Wire.SerializerFactories.ArraySerializerFactory,object,System.IO.Stream,System.Type,Wire.ValueSerializers.ValueSerializer,Wire.SerializerSession)
[...]

PR #2277, opened 2016-08-29
#2271 Akka.Cluster.Tools.PublishSubscribe.Internal made public

kantora:
Fix for the https://github.com/akkadotnet/akka.net/issues/2271

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
@@ -29,8 +29,9 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Internal
         private Count() { }
     }
 
+    // Made public for proper Wire serialization (Issue #2271)
     [Serializable]
-    internal class Bucket : IEquatable<Bucket>
+    public class Bucket : IEquatable<Bucket>
     {
         public readonly Address Owner;
         public readonly long Version;
@@ -74,8 +75,9 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Internal
         }
     }
 
+    // Made public for proper Wire serialization (Issue #2271)
     [Serializable]
-    internal sealed class ValueHolder : IEquatable<ValueHolder>
+    public sealed class ValueHolder : IEquatable<ValueHolder>
     {
         public readonly long Version;
         public readonly IActorRef Ref;


alexvaluyskiy:
Maybe Wire should fix it?

kantora:
Did I break the tests or they was already broken?

kantora:
@alexvaluyskiy this is supposed to be the temporary patch until Wire is fixed (or the correct solution designed). Just to bring functionality up. As it was discussed with @Danthar in the original issue.

P.S. I know, no one likes kludges.

Danthar:
A few remarks: 
A: I dont think this PR broke the unit tests. I think its either a racy tests thingy, or something else.
B: Like i said in the issue topic, I vote to make a PR fixing this in the Akka codebase. And @kantora followed up on that, rightly so.

The reason why I think this should be fixed in akka is because on one hand I don't want to block fixing this issue on developments in Wire (its really a minor change from the akka codebase perspective). And on the other hand I dont believe this will be fixed in Wire.

I mean, it was an internal type. Internal types are generally not meant to be serialized. So really, its no surprise Wire barfed on this.

alexvaluyskiy:
@Danthar we are changing our public API only for temporary quick-fix?

alexvaluyskiy:
@kantora if you change public fields to readonly properties in these classes, will it fix the problem?

kantora:
@alexvaluyskiy Just to be clear. Despite the subject, only two classes were made public (`Akka.Cluster.Tools.PublishSubscribe.Internal.Bucket` and `Akka.Cluster.Tools.PublishSubscribe.Internal.ValueHolder`) in order to be serializable.

> @kantora if you change public fields to readonly properties in these classes, will it fix the problem?

As I understood Wire can't serialize internal classes.
The exception:
```
Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: `Wire.SerializerFactories.ArraySerializerFactory.WriteValues<Akka.Cluster.Tools.PublishSubscribe.Internal.Bucket>(System.Collections.Generic.IReadOnlyList<Akka.Cluster.Tools.PublishSubscribe.Internal.Bucket>, System.IO.Stream, System.Type, Wire.ValueSerializers.ValueSerializer, Wire.SerializerSession)' is inaccessible due to its protection level
```
But readonly fields worked well (so readonly properties should work well to I suppose).

@alexvaluyskiy so correct solution will be moving `Bucket` and `ValueHolder` from `Internal` namespace, replacing all public fields with properties. Am I right?

alexvaluyskiy:
> As I understood Wire can't serialize internal classes.
So, we should fix Wire

rogeralsing:
I'll provide a fix inside Wire instead.
I found out what the issue is.
The current ArraySerializer in wire do some magic tricks when calling a generic method, this fails for internal types.
Internal types works everywhere else in Wire, thus it has not been spotted before.

I vote for closing this, and I'll have a fix out ASAP.

Silv3rcircl3:
Closing, will be fixed in Wire.

Issue #1982, opened 2016-05-25
Cluster: API for pluggable cluster downing


alexvaluyskiy:
API for pluggable cluster downing

https://github.com/akka/akka/pull/20320
It was implemented in Akka 2.4.5

Aaronontheweb:
I think this was added to support some of the proprietary auto-downing components that Typesafe / Lightbend developed for the Reactive Platform, but it's a good idea. I could see someone writing a plugin that uses the ARM / AWS APIs to get data about whether or not a node has been removed or not and automatically downing it.

Aaronontheweb:
Implemented in #2284

PR #2284, opened 2016-08-31
Implemented downing providers

Horusiath:
This PR includes downing providers feature described in #1982 . This change is backward compatible with existing API (only part is obsoleting one of the cluster settings).

### DIFF
--- a/src/core/Akka.Cluster.Tests/DowningProviderSpec.cs
+++ b/src/core/Akka.Cluster.Tests/DowningProviderSpec.cs
@@ -1,0 +1,111 @@
+﻿using System;
+using System.Threading;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.TestKit;
+using Akka.Util;
+using FluentAssertions;
+using Xunit;
+
+namespace Akka.Cluster.Tests
+{
+    class FailingDowningProvider : IDowningProvider
+    {
+        public FailingDowningProvider(ActorSystem system)
+        {
+        }
+
+        public TimeSpan DownRemovalMargin { get; } = TimeSpan.FromSeconds(20);
+
+        public Props DowningActorProps
+        {
+            get
+            {
+                throw new ConfigurationException("this provider never works");
+            }
+        }
+    }
+
+    class DummyDowningProvider : IDowningProvider
+    {
+        public readonly AtomicBoolean ActorPropsAccessed = new AtomicBoolean(false);
+        public DummyDowningProvider(ActorSystem system)
+        {
+        }
+
+        public TimeSpan DownRemovalMargin { get; } = TimeSpan.FromSeconds(20);
+
+        public Props DowningActorProps
+        {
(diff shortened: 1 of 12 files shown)


src/core/Akka.Cluster.Tests/DowningProviderSpec.cs
```diff
@@ -0,0 +1,111 @@
...
+        {
+            get
+            {
+                ActorPropsAccessed.Value = true;
+                return null;
+            }
+        }
+    }
+
+    public class DowningProviderSpec : AkkaSpec
```
Aaronontheweb: The specs look good to me

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -78,6 +79,9 @@ public Cluster(ActorSystemImpl system)
 
             _scheduler = CreateScheduler(system);
 
+            // it has to be lazy - otherwise if downing provider will init a cluster itself, it will deadlock
+            _downingProvider = new Lazy<IDowningProvider>(() => Akka.Cluster.DowningProvider.Load(Settings.DowningProviderType, system), LazyThreadSafetyMode.ExecutionAndPublication);
```
Aaronontheweb: Does this also need to be thread-safe for execution? Or does it only need to be thread-safe upon creation?

src/core/Akka.Cluster/ClusterSettings.cs
```diff
@@ -16,207 +16,116 @@ namespace Akka.Cluster
...
 
-            _verboseHeartbeatLogging = cc.GetBoolean("debug.verbose-heartbeat-logging");
-        }
+            VerboseHeartbeatLogging = cc.GetBoolean("debug.verbose-heartbeat-logging");
 
-        public bool LogInfo
-        {
-            get { return _logInfo; }
+            var downingProviderClassName = cc.GetString("downing-provider-class");
+            if (!string.IsNullOrEmpty(downingProviderClassName))
```
Aaronontheweb: Looks correct

src/core/Akka.Cluster/DowningProvider.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using Akka.Actor;
+using Akka.Configuration;
+
+namespace Akka.Cluster
+{
+    /// <summary>
```
Aaronontheweb: Thanks for all of the detailed intellisense comments @Horusiath

src/core/Akka.Cluster/Cluster.cs
Horusiath: As we don't have control over how `IDowningProvider` is actually implemented, I wouldn't set `PublicationOnly` - from the programmer point of view it would be safe to assume, that his/her provider won't be created random number of times, especially as it might be backed by potentially heavy initialization logic.

PR #2287, opened 2016-09-02
add support for non abbreviated timespan strings

mrrd:
This pull request adds support for non abbreviated timespan values in hocon


Aaronontheweb:
Lots of test failures....

Danthar:
@mrrd it looks like you broke the parsing of normal values ?

A quick survey reveales that values like: 0.05 s  are causing problems now.

Check the remote conf for more samples: https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote/Configuration/Remote.conf
and the remoteSettings object for a few examples on which are expected to be TimeSpans. https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote/RemoteSettings.cs

I think if you fix those scenario's it should probably resolve most, if not all, failing unit tests.

mrrd:
@Danthar thanks I can see exactly what I have missed a + in the regex

Danthar:
@mrrd can you please squash your commits

PR #2294, opened 2016-09-06
HashedWheelTimerScheduler and IScheduler disposal guarantees

### LINKED ISSUE #2288: Guarantee DedicatedThreadScheduler clean shutdowns
Need to guarantee that the scheduler is always properly disposed of whenever the `ActorSystem` terminates. This is currently blocking our ability to run CI on Mono as we run into these sorts of exceptions when running larger test suites:

http://petabridge-ci.cloudapp.net/viewLog.html?buildId=15538&tab=buildResultsDiv&buildTypeId=AkkaNet_LinuxMonoBuild&guest=1
```
System.SystemException : Thread creation failed.
  at System.Threading.Thread.StartInternal (IPrincipal principal, System.Threading.StackCrawlMark& stackMark) <0x411f1470 + 0x000e7> in <filename unknown>:0 
  at System.Threading.Thread.Start (System.Threading.StackCrawlMark& stackMark) <0x411f1320 + 0x000c4> in <filename unknown>:0 
  at System.Threading.Thread.Start () <0x412bb920 + 0x0002b> in <filename unknown>:0 
  at Akka.Actor.DedicatedThreadScheduler..ctor (Akka.Actor.ActorSystem system) <0x412bb410 + 0x002e3> in <filename unknown>:0 
  at (wrapper managed-to-native) System.Reflection.MonoCMethod:InternalInvoke (System.Reflection.MonoCMethod,object,object[],System.Exception&)
  at System.Reflection.MonoCMethod.InternalInvoke (System.Object obj, System.Object[] parameters) <0x411618d0 + 0x00045> in <filename unknown>:0 
```

Aaronontheweb:
Closes #2288


src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -517,12 +517,13 @@ namespace Akka.Actor
         public static Akka.Actor.DeployableDecider From(Akka.Actor.Directive defaultDirective, System.Collections.Generic.IEnumerable<System.Collections.Generic.KeyValuePair<System.Type, Akka.Actor.Directive>> pairs) { }
         public static Akka.Actor.LocalOnlyDecider From(System.Func<System.Exception, Akka.Actor.Directive> localOnlyDecider) { }
     }
-    public class DedicatedThreadScheduler : Akka.Actor.SchedulerBase, Akka.Actor.IDateTimeOffsetNowTimeProvider, Akka.Actor.ITimeProvider
+    public class DedicatedThreadScheduler : Akka.Actor.SchedulerBase, Akka.Actor.IDateTimeOffsetNowTimeProvider, Akka.Actor.ITimeProvider, System.IDisposable
```
Aaronontheweb: Public API change: `DedicatedThreadScheduler` now implements `IDisposable`

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -517,12 +517,13 @@ namespace Akka.Actor
...
     {
-        public DedicatedThreadScheduler(Akka.Actor.ActorSystem system) { }
+        public DedicatedThreadScheduler(Akka.Configuration.Config config, Akka.Event.ILoggingAdapter log) { }
```
Aaronontheweb: **Breaking** public API change: all `IScheduler` implementations take a `Config` and `ILoggingAdapter` object in their constructors instead of an `ActorSystem` reference. This is because the `IScheduler` is instantiated *inside* the `ActorSystem` CTOR and this creates a bit of a weird code cycle that was unresolvable in an earlier implementation of this. This is a safer implementation and will only affect users who are using their own custom `IScheduler`s, but all they have to do is update the CTOR.

src/core/Akka.Tests.Performance/Actor/Scheduler/DefaultSchedulerPerformanceTests.cs
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Akka.Tests.Performance.Actor.Scheduler
+{
+    /// <summary>
+    /// Designed to measure how many operations per-second the scheduler can perform in a fixed window of time.
+    /// 
+    /// Meant to measure the efficiency of the default <see cref="IScheduler"/> and to see if there's much contention
+    /// around the 
+    /// </summary>
+    public class DefaultSchedulerPerformanceTests
```
Aaronontheweb: Benchmark for the default `IScheduler` implementation. Not sure if this is the best design or not - wanted to test how much pressure could be put onto the work queue by multiple simultaneous queuers. There's a lot more that could be done here.

src/core/Akka.Tests/Actor/Scheduler/SchedulerShutdownSpec.cs
```diff
@@ -0,0 +1,137 @@
...
+using Akka.Util.Internal;
+using FluentAssertions;
+using Xunit;
+
+namespace Akka.Tests.Actor.Scheduler
+{
+    /// <summary>
+    /// Need to guarantee that the <see cref="IScheduler"/>, if it implements <see cref="IDisposable"/>, is shutdown.
+    /// </summary>
+    public class SchedulerShutdownSpec
```
Aaronontheweb: Specs for verifying that:

1. It's possible to have an `IScheduler` automatically shut down via `IDisposable`.
2. That the default `IScheduler` does this.
3. That the default `IScheduler` throws a `SchedulerException` in the event of attempting to schedule after the scheduler has been shutdown.

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -119,20 +120,39 @@ public override void Abort()
         /// <summary>Starts this system</summary>
         public void Start()
         {
-            _provider.Init(this);
-            ConfigureLoggers();
-            LoadExtensions();
+            try
+            {
+                RegisterOnTermination(StopScheduler);
```
Aaronontheweb: Sets a `TerminationHook` for ensuring that the `IScheduler` is terminated before `ActorSystem.WhenTerminated` completes, which guarantees that we won't leak any of the `IScheduler` threads across multiple specs in the test suite.

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -170,7 +190,13 @@ public override ActorSelection ActorSelection(string actorPath)
         private void ConfigureScheduler()
         {
             var schedulerType = Type.GetType(_settings.SchedulerClass, true);
-            _scheduler = (IScheduler) Activator.CreateInstance(schedulerType, this);
+            _scheduler = (IScheduler) Activator.CreateInstance(schedulerType, _settings.Config, Log);
```
Aaronontheweb: Updated CTOR args for creating an `IScheduler`.

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -170,7 +190,13 @@ public override ActorSelection ActorSelection(string actorPath)
...
+        }
+
+        private void StopScheduler()
+        {
+            var sched = Scheduler as IDisposable;
+            sched?.Dispose();
```
Aaronontheweb: `IScheduler.Dispose` is a synchronous call on all implementations.

src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
```diff
@@ -9,37 +9,48 @@
...
         {
-            _log = Logging.GetLogger(system, this);
-            var precision = system.Settings.Config.GetTimeSpan("akka.scheduler.tick-duration");
+            var precision = SchedulerConfig.GetTimeSpan("akka.scheduler.tick-duration");
+            _shutdownTimeout = SchedulerConfig.GetTimeSpan("akka.scheduler.shutdown-timeout");
             var thread = new Thread(_ =>
             {
                 var allWork = new List<ScheduledWork>();
-                while (true)
+                while (_stopped.Value == null)
```
Aaronontheweb: Check for termination signal on each loop instance.

src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
```diff
@@ -70,12 +81,16 @@ public DedicatedThreadScheduler(ActorSystem system)
                         }
                     }
                 }
+
+                // shutdown has been signaled
+               FireStopSignal();
```
Aaronontheweb: Signal to the caller that termination has completed, which allows the rest of the shutdown process to complete.

src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
```diff
@@ -141,22 +161,61 @@ private void InternalScheduleRepeatedly(TimeSpan initialDelay, TimeSpan interval
...
+            }
+            catch (Exception)
+            {
+                
+            }
+        }
+
+        public void Dispose()
+        {
+            if (!Stop().Wait(_shutdownTimeout))
```
Aaronontheweb: We block termination here on purpose - the scheduler must be stopped before we can consider `ActorSystem` termination to be complete.

src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
```diff
@@ -141,22 +161,61 @@ private void InternalScheduleRepeatedly(TimeSpan initialDelay, TimeSpan interval
...
+                Log.Warning("Failed to shutdown DedicatedThreadScheduler within {0}", _shutdownTimeout);   
+            }
+        }
+
+        private static readonly Task Completed = Task.FromResult(true);
+
+        private Task Stop()
+        {
+            var p = new TaskCompletionSource<bool>();
+            if (_stopped.CompareAndSet(null, p))
```
Aaronontheweb: Once this signal is set, any currently scheduled work will be allowed to complete 1 iteration only.

Aaronontheweb:
TODO: need to guarantee that repeating scheduled tasks can't deadlock the termination process.
TODO: need to update dispatchers to use the updated SchedulerException.
TODO: need to guarantee that an exception while performing scheduled work doesn't disable the `IScheduler`

Aaronontheweb:
# Benchmark Results
# Akka.Tests.Performance.Actor.Scheduler.DefaultSchedulerPerformanceTests+SchedulerThroughputStressTest
__Tests to see how many concurrent jobs we can schedule and what the effects are on throughput__
_9/7/2016 12:43:51 AM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
NumberOfIterations=1, MaximumRunTime=00:00:01
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalBytesAllocated |           bytes |  277,380,432.00 |  277,380,432.00 |  277,380,432.00 |            0.00 |
TotalCollections [Gen0] |     collections |          432.00 |          432.00 |          432.00 |            0.00 |
TotalCollections [Gen1] |     collections |          194.00 |          194.00 |          194.00 |            0.00 |
TotalCollections [Gen2] |     collections |           56.00 |           56.00 |           56.00 |            0.00 |
[Counter] ScheduledJobs |      operations |      889,600.00 |      889,600.00 |      889,600.00 |            0.00 |
[Counter] ScheduleInvokes |      operations |   41,414,684.00 |   41,414,684.00 |   41,414,684.00 |            0.00 |

### Per-second Totals
TotalBytesAllocated |           bytes |   13,207,660.29 |   13,207,660.29 |   13,207,660.29 |            0.00 |
TotalCollections [Gen0] |     collections |           20.57 |           20.57 |           20.57 |            0.00 |
TotalCollections [Gen1] |     collections |            9.24 |            9.24 |            9.24 |            0.00 |
TotalCollections [Gen2] |     collections |            2.67 |            2.67 |            2.67 |            0.00 |
[Counter] ScheduledJobs |      operations |       42,358.92 |       42,358.92 |       42,358.92 |            0.00 |
[Counter] ScheduleInvokes |      operations |    1,971,988.70 |    1,971,988.70 |    1,971,988.70 |            0.00 |

### Raw Data
#### TotalBytesAllocated
           Run # |           bytes |       bytes / s |      ns / bytes |
               1 |  277,380,432.00 |   13,207,660.29 |           75.71 |

#### TotalCollections [Gen0]
           Run # |     collections | collections / s |ns / collections |
               1 |          432.00 |           20.57 |   48,614,540.28 |

#### TotalCollections [Gen1]
           Run # |     collections | collections / s |ns / collections |
               1 |          194.00 |            9.24 |  108,255,058.76 |

#### TotalCollections [Gen2]
           Run # |     collections | collections / s |ns / collections |
               1 |           56.00 |            2.67 |  375,026,453.57 |

#### [Counter] ScheduledJobs
               1 |      889,600.00 |       42,358.92 |       23,607.78 |

#### [Counter] ScheduleInvokes
               1 |   41,414,684.00 |    1,971,988.70 |          507.10 |

Aaronontheweb:
Biggest surprise to me on this benchmark isn't the speed (roughly 2 million msg /s) but the garbage collection. Feels like this code is blowing through a lot more gen 2 GC than it should.... And an incredible amount of memory allocation (277mb.)

The number of schedule attempts per second is pretty much fixed by the way I wrote the benchmark; not terribly important here to measure maximum speed IMHO. Rather to see if the changes we make result in an improvement or not.

Aaronontheweb:
Implemented first cut of HashedWheelTimerScheduler - waiting for benchmark data. A little worried that it's sleep detector isn't sensitive enough, but it's tough to tell when debugging. Probably going to need to go over it with a profiler. Passed all of the scheduler specs save one though (since this is a behavior I have not yet implemented)

Aaronontheweb:
## HashedWheelTimerScheduler Benchmark 1

# Akka.Tests.Performance.Actor.Scheduler.DefaultSchedulerPerformanceTests+SchedulerThroughputStressTest
__Tests to see how many concurrent jobs we can schedule and what the effects are on throughput__
_9/7/2016 9:58:44 PM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
NumberOfIterations=1, MaximumRunTime=00:00:01
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalBytesAllocated |           bytes |  100,706,168.00 |  100,706,168.00 |  100,706,168.00 |            0.00 |
TotalCollections [Gen0] |     collections |          961.00 |          961.00 |          961.00 |            0.00 |
TotalCollections [Gen1] |     collections |          202.00 |          202.00 |          202.00 |            0.00 |
TotalCollections [Gen2] |     collections |            7.00 |            7.00 |            7.00 |            0.00 |
[Counter] ScheduledJobs |      operations |      856,400.00 |      856,400.00 |      856,400.00 |            0.00 |
[Counter] ScheduleInvokes |      operations |   38,603,899.00 |   38,603,899.00 |   38,603,899.00 |            0.00 |

### Per-second Totals
TotalBytesAllocated |           bytes |    5,035,062.97 |    5,035,062.97 |    5,035,062.97 |            0.00 |
TotalCollections [Gen0] |     collections |           48.05 |           48.05 |           48.05 |            0.00 |
TotalCollections [Gen1] |     collections |           10.10 |           10.10 |           10.10 |            0.00 |
TotalCollections [Gen2] |     collections |            0.35 |            0.35 |            0.35 |            0.00 |
[Counter] ScheduledJobs |      operations |       42,817.91 |       42,817.91 |       42,817.91 |            0.00 |
[Counter] ScheduleInvokes |      operations |    1,930,100.87 |    1,930,100.87 |    1,930,100.87 |            0.00 |

### Raw Data
#### TotalBytesAllocated
           Run # |           bytes |       bytes / s |      ns / bytes |
               1 |  100,706,168.00 |    5,035,062.97 |          198.61 |

#### TotalCollections [Gen0]
           Run # |     collections | collections / s |ns / collections |
               1 |          961.00 |           48.05 |   20,812,668.99 |

#### TotalCollections [Gen1]
           Run # |     collections | collections / s |ns / collections |
               1 |          202.00 |           10.10 |   99,014,727.23 |

#### TotalCollections [Gen2]
           Run # |     collections | collections / s |ns / collections |
               1 |            7.00 |            0.35 |2,857,282,128.57 |

#### [Counter] ScheduledJobs
               1 |      856,400.00 |       42,817.91 |       23,354.71 |

#### [Counter] ScheduleInvokes
               1 |   38,603,899.00 |    1,930,100.87 |          518.11 |

Aaronontheweb:
Big drop in allocations (277mb --> 100 mb) and GEN2 GC.
 (56 --> 7).

Slight increase in scheduling throughput, which I think is the result of decreased contention around the `ConcurrentQueue` this design uses. 42,358.92 schedules per second --> 42,817.91.

Bit of a drop in actual scheduled calls per second, which could be the result of the fact that we're now scheduling tasks to be executed on the `ThreadPool` instead of doing it directly on the scheduler thread (adds a delay to executing the scheduled operation.) Let me play around with that. 1,971,988.70 --> 1,930,100.87.

Aaronontheweb:
Looking at the MNTR logs, seems like there's still some issues with my implementation:

```
[21:45:13]	[Step 2/2] 
[21:45:13]	[Step 2/2] Unhandled Exception: System.InvalidOperationException: SchedulerRegistration.Deadline [200113356] > Timer.Deadline [146742809]
[21:45:13]	[Step 2/2]    at Akka.Actor.HashedWheelTimerScheduler.Bucket.Execute(Int64 deadline) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka\Actor\Scheduler\HashedWheelTimerScheduler.cs:line 574
[21:45:13]	[Step 2/2]    at Akka.Actor.HashedWheelTimerScheduler.Run() in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka\Actor\Scheduler\HashedWheelTimerScheduler.cs:line 165
[21:45:13]	[Step 2/2]    at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
[21:45:13]	[Step 2/2]    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[21:45:13]	[Step 2/2]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[21:45:13]	[Step 2/2]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
[21:45:13]	[Step 2/2]    at System.Threading.ThreadHelper.ThreadStart()
[21:45:13]	[Step 2/2] 
[21:45:13]	[Step 2/2] Unhandled Exception: System.InvalidOperationException: SchedulerRegistration.Deadline [200112555] > Timer.Deadline [146803128]
[21:45:13]	[Step 2/2]    at Akka.Actor.HashedWheelTimerScheduler.Bucket.Execute(Int64 deadline) in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka\Actor\Scheduler\HashedWheelTimerScheduler.cs:line 574
[21:45:13]	[Step 2/2]    at Akka.Actor.HashedWheelTimerScheduler.Run() in D:\BuildAgent\work\d26c9d7f36545acd\src\core\Akka\Actor\Scheduler\HashedWheelTimerScheduler.cs:line 165
[21:45:13]	[Step 2/2]    at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
[21:45:13]	[Step 2/2]    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[21:45:13]	[Step 2/2]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[21:45:13]	[Step 2/2]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
[21:45:13]	[Step 2/2]    at System.Threading.ThreadHelper.ThreadStart()
```

Aaronontheweb:
## HashedWheelTimerScheduler Benchmark 2

# Akka.Tests.Performance.Actor.Scheduler.DefaultSchedulerPerformanceTests+SchedulerThroughputStressTest
__Tests to see how many concurrent jobs we can schedule and what the effects are on throughput__
_9/8/2016 12:25:32 AM_
### System Info
```ini
NBench=NBench, Version=0.3.0.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
WorkerThreads=32767, IOThreads=2
```

### NBench Settings
```ini
NumberOfIterations=1, MaximumRunTime=00:00:01
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
TotalBytesAllocated |           bytes |  111,339,936.00 |  111,339,936.00 |  111,339,936.00 |            0.00 |
TotalCollections [Gen0] |     collections |           20.00 |           20.00 |           20.00 |            0.00 |
TotalCollections [Gen1] |     collections |            7.00 |            7.00 |            7.00 |            0.00 |
TotalCollections [Gen2] |     collections |            1.00 |            1.00 |            1.00 |            0.00 |
[Counter] ScheduledJobs |      operations |      964,800.00 |      964,800.00 |      964,800.00 |            0.00 |
[Counter] ScheduleInvokes |      operations |   79,906,737.00 |   79,906,737.00 |   79,906,737.00 |            0.00 |

### Per-second Totals
TotalBytesAllocated |           bytes |    5,566,713.06 |    5,566,713.06 |    5,566,713.06 |            0.00 |
TotalCollections [Gen0] |     collections |            1.00 |            1.00 |            1.00 |            0.00 |
TotalCollections [Gen1] |     collections |            0.35 |            0.35 |            0.35 |            0.00 |
TotalCollections [Gen2] |     collections |            0.05 |            0.05 |            0.05 |            0.00 |
[Counter] ScheduledJobs |      operations |       48,237.54 |       48,237.54 |       48,237.54 |            0.00 |
[Counter] ScheduleInvokes |      operations |    3,995,133.22 |    3,995,133.22 |    3,995,133.22 |            0.00 |

### Raw Data
#### TotalBytesAllocated
           Run # |           bytes |       bytes / s |      ns / bytes |
               1 |  111,339,936.00 |    5,566,713.06 |          179.64 |

#### TotalCollections [Gen0]
           Run # |     collections | collections / s |ns / collections |
               1 |           20.00 |            1.00 |1,000,050,970.00 |

#### TotalCollections [Gen1]
           Run # |     collections | collections / s |ns / collections |
               1 |            7.00 |            0.35 |2,857,288,485.71 |

#### TotalCollections [Gen2]
           Run # |     collections | collections / s |ns / collections |
               1 |            1.00 |            0.05 |20,001,019,400.00 |

#### [Counter] ScheduledJobs
               1 |      964,800.00 |       48,237.54 |       20,730.74 |

#### [Counter] ScheduleInvokes
               1 |   79,906,737.00 |    3,995,133.22 |          250.30 |

Aaronontheweb:
I've run the benchmark 3-4 times now and this is the lowest recorded result; wanted to sanity check the figures. Biggest change I made? All scheduled work happens on the scheduler thread instead of being executed as a new `Task` on the `ThreadPool`.

As long as the scheduled operations don't run for very long (most don't) this design achieves significantly higher throughput with a fraction of the allocations and garbage collections. Memory consumption is about the same: it's determined largely by the number of scheduled jobs, and many more jobs are queued in this design.

## Throughput
This spec always runs for 20 seconds. This implementation has fundamentally the same contention overhead as the previous HashedWheelTimerScheduler, but this implementation is able to improve upon 	42,817.91 schedules added per second all the way up to 48,237.54 per second. At 12-13% increase. Moving away from the `ThreadPool` wouldn't affect that directly, but decreasing the amount of garbage collection attempts would.

In in this benchmark we also improve upon total scheduled actions executed throughput from the previous HashedWheelTimerScheduler implementation from 1,930,100.87 to 3,995,133.22 per second. More than double the performance.

I think this is because:

1. We're executing the work synchronously rather than scheduling it onto the ThreadPool, which makes the timing of *when* the scheduled work is performed much more accurate.
2. The previous increase in the rate of scheduling is going to have a compound effect on scheduled operations executed, given that all of these events are *repeating* events at 20ms intervals. So there's just much more work being scheduled.
3. The sleep algorithm is dynamic in the HashedWheelTimer - if we had a lot of work to execute in one bucket (position on the wheel) we may not sleep the worker thread at all and just proceed directly to the next bucket immediately. Sleep only occurs if there's at least 1 millisecond of time between one bucket finishing and another starting. This means that as the scheduler's load increases, the amount of time it spends sleeping decreases or becomes more spread out over longer intervals of activity.

I'm going to remove the `Task` code that I commented out in this benchmark as I think this is just more efficient and accurate over delegating execution of scheduled work to the `ThreadPool`.

## Memory
Memory allocation didn't noticeably change at all - consumption increased slightly due to the increase in work scheduled. I'd have to run a profiler on both versions to understand why, but my back-of-the-envelope math works out to the fact that the `Task`s we were allocating earlier were using a static delegate with an argument, therefore aside from the `Task` object itself we weren't allocating much memory in terms of total bytes although we were allocating significantly more objects.

## GC
The biggest difference between this design and the other two (the first one being the original `DedicatedThreadScheduler`) is that this one has virtually no garbage collection overhead compared to the others.

Each item here is for all three implementations in order:

* **GEN2**: 56 --> 7 --> 1 garbage collections over 20 seconds
* **GEN1**: 194 -->202 --> 7 garbage collections over 20 seconds
* **GEN0**:  432 --> 961 --> 20 garbage collections over 20 seconds

The original `DedicatedThreadScheduler` is pretty heavy on GC across all generations; the first draft of the `HashedWheelTimerScheduler` is much more efficient on GEN2, but extremely heavy on the earlier GC generations. Eliminating the use of the `Task` in the second draft of the `HashedWheelTimerScheduler` resulted in drastic drops in GC pressure everywhere. In addition to making the timer more precise it also allocates far fewer objects - a recurring task takes exactly three allocations for the duration of its lifetime (`SchedulerRegistration`, `IRunnable`, `ICancelable`) and those objects will be re-used indefinitely as long as the `ICancelable` is not invoked. There will be some GC pressure for "one shot" tasks, but we make that pretty easy for the garbage collector by stripping all of the references to it out of our linked lists. 

## Conclusion
The `HashedWheelTimer` looks like the way to go. I already made it the default and the entire test suite is running on it, including the MNTR. It might merit some additional profiling to ensure that it's not being wasteful with its sleep / wake routine but other than that I think it's a demonstrable improvement. Going to go try it out in conjunction with the Mono specs.

Aaronontheweb:
@akkadotnet/developers this is now ready to be reviewed once CI has a chance to run.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
alexvaluyskiy: Should we change the version to 1.2.0 if we have API breaking changes?

src/core/Akka.Tests.Performance/Actor/Scheduler/DefaultSchedulerPerformanceTests.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        private ActorSystem _actorSystem;
+        public const string ScheduledOps = "ScheduleInvokes";
```
alexvaluyskiy: Do we really need these public consts? Could we make them as private?

src/core/Akka.Tests/Actor/Scheduler/TaskBasedScheduler_TellScheduler_Schedule_Tests.cs
```diff
@@ -9,54 +9,66 @@
...
 using Xunit.Extensions;
 
 namespace Akka.Tests.Actor.Scheduler
 {
     // ReSharper disable once InconsistentNaming
-    public class DedicatedThreadScheduler_TellScheduler_Schedule_Tests : AkkaSpec
+    public class DefaultScheduler_TellScheduler_Schedule_Tests : AkkaSpec
     {
-      //  [Theory(Skip = "Tests that messages are sent with the specified interval, however due to inaccuracy of Task.Delay this often fails. Run this manually if you've made changes to DedicatedThreadScheduler")]
+        [Theory(Skip = "Tests that messages are sent with the specified interval, however due to inaccuracy of Task.Delay this often fails. Run this manually if you've made changes to DedicatedThreadScheduler")]
```
alexvaluyskiy: Why you skip it?

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -170,7 +190,13 @@ public override ActorSelection ActorSelection(string actorPath)
         private void ConfigureScheduler()
         {
             var schedulerType = Type.GetType(_settings.SchedulerClass, true);
-            _scheduler = (IScheduler) Activator.CreateInstance(schedulerType, this);
+            _scheduler = (IScheduler) Activator.CreateInstance(schedulerType, _settings.Config, Log);
+        }
+
+        private void StopScheduler()
+        {
+            var sched = Scheduler as IDisposable;
```
alexvaluyskiy: why not `var scheduler = ...`?

src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
```diff
@@ -9,37 +9,48 @@
...
+        }
+
+        public override TimeSpan HighResMonotonicClock
+        {
+            get { return Util.MonotonicClock.ElapsedHighRes; }
+        }
+
+        private TimeSpan _shutdownTimeout;
 
         //TODO: use some more efficient approach to handle future work
```
alexvaluyskiy: Is TODO still actual?

src/core/Akka.Tests/Actor/Scheduler/TaskBasedScheduler_TellScheduler_Schedule_Tests.cs
Aaronontheweb: It was already skipped - I just uncommented it so I could run it in ReSharper.

src/core/Akka.Tests.Performance/Actor/Scheduler/DefaultSchedulerPerformanceTests.cs
Aaronontheweb: It's just a benchmark. Doesn't really matter.

src/core/Akka/Actor/Scheduler/DedicatedThreadScheduler.cs
Aaronontheweb: Hah, you're right - I'll get rid of that.

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
Aaronontheweb: `_scheduler` is the backing field for the `Scheduler` property

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Aaronontheweb: Nah, this is an internal component which *can* be extended but we don't know of any implementations out in the wild. Plus the old API is frankly kind of dangerous (as I explained in my comment below.) 

What I've decided to do was change the `SchedulerBase` constructor that supports the new `Activator.CreateInstance` arguments , but I've left the old `DedicatedThreadScheduler` constructor intact in case anyone was instantiating one of those for some reason. I figured that was a reasonable compromise.

alexvaluyskiy:
LGTM, may I merge it?

Aaronontheweb:
@alexvaluyskiy I'm going to implement some of your remarks and do a bit of cleanup first.

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
```diff
@@ -0,0 +1,670 @@
+﻿using System;
```
Silv3rcircl3: Copyright header  ;)

Aaronontheweb: Doh! yep, will add that

Aaronontheweb:
Addressed all remarks, approved the API, etc... this should be ready to go now.

Issue #2296, opened 2016-09-06
MNTR: Unannounced node disconnects


Aaronontheweb:
Been a big increase in MNTR failures lately as we've expanded the test suite, looking into it there appears to be a possible fault with the MNTR Runner itself:

<summary>
Cause: Akka.Remote.TestKit.BarrierCoordinator+ClientLostException: unannounced disconnect of RoleName(fourth) - Akka.Cluster.Tests.ConvergenceSpec
<details>
[DEBUG][9/6/2016 9:52:50 PM][Thread 0035][Akka.Remote.TestKit.MsgDecoder] Decoded Akka.Remote.TestKit.EnterBarrier
[DEBUG][9/6/2016 9:52:50 PM][Thread 0018][akka://MultiNodeClusterSpec/user/controller] received handled message Akka.Remote.TestKit.Controller+ClientDisconnected: RoleName(fourth)
[DEBUG][9/6/2016 9:52:50 PM][Thread 0018][[akka://MultiNodeClusterSpec/user/controller/barriers#1509742045]] processing Event: <Akka.Remote.TestKit.Controller+ClientDisconnected: RoleName(fourth)>, StateData: <Akka.Remote.TestKit.BarrierCoordinator+Data> from [akka://MultiNodeClusterSpec/user/controller/%5B%3A%3A1%5D%3A52005%3A52005-server3#1539168185]
[DEBUG][9/6/2016 9:52:50 PM][Thread 0035][Akka.Remote.TestKit.ConductorHandler] message from [::1]:52008: Akka.Remote.TestKit.EnterBarrier
[DEBUG][9/6/2016 9:52:50 PM][Thread 0018][[akka://MultiNodeClusterSpec/user/controller/%5B%3A%3A1%5D%3A52008%3A52008-server4#8676766]] processing Event: <Akka.Remote.TestKit.EnterBarrier>, StateData: <> from [akka://MultiNodeClusterSpec/deadLetters]
[DEBUG][9/6/2016 9:52:50 PM][Thread 0018][[akka://MultiNodeClusterSpec/user/controller/%5B%3A%3A1%5D%3A52008%3A52008-server4#8676766]] transition Ready,  -> Ready, 
[...]

PR #2297, opened 2016-09-06
Fixed folder output, barrier spec failure handling issues in MNTR

Aaronontheweb:
Working on resolviong #2296

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/Singleton/ClusterSingletonManagerStartupSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/Singleton/ClusterSingletonManagerStartupSpec.cs
@@ -31,6 +31,7 @@ namespace Akka.Cluster.Tools.Tests.MultiNode.Singleton
             CommonConfig = ConfigurationFactory.ParseString(@"
                 akka.loglevel = INFO
                 akka.actor.provider = ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""
+                akka.remote.retry-gate-closed-for = 1s #fast restart
                 akka.remote.log-remote-lifecycle-events = off
                 akka.cluster.auto-down-unreachable-after = 0s")
             .WithFallback(ClusterSingletonManager.DefaultConfig())
--- a/src/core/Akka.MultiNodeTestRunner/Program.cs
+++ b/src/core/Akka.MultiNodeTestRunner/Program.cs
@@ -156,7 +156,7 @@ namespace Akka.MultiNodeTestRunner
                                     } -Dmultinode.listen-address={listenAddress} -Dmultinode.listen-port={listenPort}";
                             var nodeIndex = nodeTest.Node;
                             //TODO: might need to do some validation here to avoid the 260 character max path error on Windows
-                            var folder = Directory.CreateDirectory(Path.Combine(OutputDirectory, nodeTest.MethodName));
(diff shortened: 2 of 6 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/Singleton/ClusterSingletonManagerStartupSpec.cs
```diff
@@ -31,6 +31,7 @@ public ClusterSingletonManagerStartupConfig()
             CommonConfig = ConfigurationFactory.ParseString(@"
                 akka.loglevel = INFO
                 akka.actor.provider = ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""
+                akka.remote.retry-gate-closed-for = 1s #fast restart
```
Aaronontheweb: Hardening the ClusterSingletonManagerStartupSpec - in the event two nodes try to connect to eachother at the same time (happens sometimes during Akka.Cluster formation) this lowers the retry gate from 5s to 1s so the spec has an opportunity to retry the association before hitting the deadline.

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -156,7 +156,7 @@ static void Main(string[] args)
                                     } -Dmultinode.listen-address={listenAddress} -Dmultinode.listen-port={listenPort}";
                             var nodeIndex = nodeTest.Node;
                             //TODO: might need to do some validation here to avoid the 260 character max path error on Windows
-                            var folder = Directory.CreateDirectory(Path.Combine(OutputDirectory, nodeTest.MethodName));
+                            var folder = Directory.CreateDirectory(Path.Combine(OutputDirectory, nodeTest.TestName));
```
Aaronontheweb: Uses the correct test name for the folders used for each of the individual node logs -this prevents two implementations of the same spec from overriding eachother's output.

src/core/Akka.Remote.TestKit/BarrierCoordinator.cs
```diff
@@ -244,7 +244,7 @@ public override int GetHashCode()
         public sealed class WrongBarrierException : Exception
         {
             public WrongBarrierException(string barrier, IActorRef client, Data barrierData)
-                : base($"tried to enter '{barrier}' while we were waiting for '{barrierData.Barrier}'")
+                : base($"[{client}] tried to enter '{barrier}' while we were waiting for '{barrierData.Barrier}'")
```
Aaronontheweb: Clearer barrier exception error message.

src/core/Akka.Remote.TestKit/Player.cs
```diff
@@ -136,7 +136,7 @@ public void Enter(TimeSpan timeout, ImmutableList<string> names)
                     var askTimeout = barrierTimeout + Settings.QueryTimeout;
                     // Need to force barrier to wait here, so we can pass along a "fail barrier" message in the event
                     // of a failed operation
-                    _client.Ask(new ToServer<EnterBarrier>(new EnterBarrier(name, barrierTimeout)), askTimeout).Wait();
+                    var result = _client.Ask(new ToServer<EnterBarrier>(new EnterBarrier(name, barrierTimeout)), askTimeout).Result;
```
Aaronontheweb: We've had false positives in the MNTR previously where specs were still passing even though barriers were failing. This change ensures that an exception is thrown instead, failing the spec for this node if a barrier operation fails.

PR #2307, opened 2016-09-13
Akka IO redesign using async IO - SocketAsyncEventArgs

willieferguson:
This PR replaces Akka IO's current Socket.Select design with IOCP using SocketAsyncEventsArgs.

The Socket.Select design has the following disadvantages:
* A maximum of 65k sockets are allowed
* There is no way to cancel Socket.Select, and therefore the select loop has to run continuously, causing moderate CPU utilization even when idle.

The following issues are still outstanding:
* Performance testing - initial testing showed between 0 and 50% performance improvement, depending on load.
* More integration tests are required.
* When an actor is terminated any 'outstanding' events are send to deadletters. The best solution I could come up with is for the connection manager to subscribe to deadletters and return leaked SAEA's to the pool.
* Confirm SuspendReading works as expected.


Aaronontheweb:
cc @Silv3rcircl3 any ideas on the Akka.Streams error?

Silv3rcircl3:
@Aaronontheweb yup, will look into it laiter today, should be easy to fix

Silv3rcircl3:
@willieferguson can you implement the changes from this commit https://github.com/Silv3rcircl3/akka.net/commit/f559369fc2164c0b36c6a48f78e488b9daa5d934
?
The skipped tests need some further investigation, don't know whats happening there, can take a look after the update to 2.4.10.

willieferguson:
Do I need to squash the commits for the feature branch, or will it be squashed before merging into dev?

Silv3rcircl3:
We can squash the commits when we merge it, no need to do it manually.

Horusiath:
Is it ready to merge then?

Silv3rcircl3:
Still needs a review, hadn't had the time to do it.

alexvaluyskiy:
We should squash the commits only when we merge to dev, not to the custom branch

Aaronontheweb:
@willieferguson could you please resend this PR to the `dev` branch and squash the commits, as @alexvaluyskiy suggested?

alexvaluyskiy:
@Aaronontheweb I think we should merge it to Akka-IO branch, then write additional integration and performance tests, then fix Stream Tcp specs and only after that we should merge it to the dev branch

Danthar:
I agree with @alexvaluyskiy on this.

src/core/Akka.Streams.Tests/IO/TcpSpec.cs
```diff
@@ -457,14 +458,14 @@ public void Outgoing_TCP_stream_must_handle_when_connection_actor_terminates_une
                 .RunAggregate(0, (i, s) => i + s.Count, mat2);
 
             // Getting rid of existing connection actors by using a blunt instrument
-            system2.ActorSelection(system2.Tcp().Path/"selectors"/"$b"/"*").Tell(Kill.Instance);
+            system2.ActorSelection(system2.Tcp().Path/"$a"/"*").Tell(Kill.Instance);
```
Danthar: This line made me do a triple take :P Never new the ActorPath class implements the / operator

Danthar:
Do these changes take into account the fixes made in https://github.com/akkadotnet/akka.net/pull/2258 regarding the DNS system. It does not look like it.

willieferguson:
@Danthar The default DNS resolver in Akka IO only return IP4 addresses. There were no changes to Akka IO's DNS resolution in this PR.

alexvaluyskiy:
I would like to merge it to akka-io branch, and create tickets for missing features
@willieferguson what tickets should I create?

willieferguson:
@alexvaluyskiy I think we need one ticket to extend the akka io test suite, which should initially include a checklist for:
- Tests for IOCP's pool leaks
- Tests Suspend/Resume behavior 
- Tests closed events (this is indirectly tested with TCP stream tests)
- Tests large send/receive (pending write changes)

Additionally we might also create a ticket to evaluate the SocketAsyncEventArgsPool implementation.

PR #2310, opened 2016-09-16
Clustered routers now load correct default number of routees

### LINKED ISSUE #2266: Clustered Broadcast Group Routers do not seem to work correctly
I have been wrestling with clustering the broadcast group router.

I have written a testing spec to verify my findings that it just does not seem to work currently, and better visualise that.

The following repository contains my broadcast group clustering test spec:

https://github.com/mrrd/AkkaBroadcastGroupTests

This seems to show that the clustered broadcast-group is only adding a single node routee to it, it should really be two as far as I can see.

Aaronontheweb:
closes #2266

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -318,7 +318,7 @@ Target "MultiNodeTests" <| fun _ ->
                 |> append assembly
                 |> append "-Dmultinode.enable-filesink=on"
                 |> append (sprintf "-Dmultinode.output-directory=\"%s\"" testOutput)
-                |> appendIfNotNullOrEmpty spec "-Dmultinode.test-spec="
+                |> appendIfNotNullOrEmpty spec "-Dmultinode.spec="
                 |> toText
 
         let result = ExecProcess(fun info -> 
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -591,6 +591,7 @@ namespace Akka.Actor
     }
     public class Deployer
     {
+        protected readonly Akka.Configuration.Config Default;
         public Deployer(Akka.Actor.Settings settings) { }
         public Akka.Actor.Deploy Lookup(Akka.Actor.ActorPath path) { }
         public Akka.Actor.Deploy Lookup(System.Collections.Generic.IEnumerable<string> path) { }
--- a/src/core/Akka.Cluster.Tests.MultiNode/Routing/ClusterBroadcastRouter2266BugfixSpec.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/Routing/ClusterBroadcastRouter2266BugfixSpec.cs
@@ -1,0 +1,200 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ClusterBroadcastRouter2266BugfixSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Akka.Actor;
+using Akka.Cluster.TestKit;
+using Akka.Configuration;
+using Akka.Remote.TestKit;
+using Akka.Routing;
+using FluentAssertions;
+
+namespace Akka.Cluster.Tests.MultiNode.Routing
+{
+    public class ClusterBroadcastGroupSpecConfig : MultiNodeConfig
+    {
+        internal interface IRouteeType { }
+        internal class PoolRoutee : IRouteeType { }
+        internal class GroupRoutee : IRouteeType { }
+
+        internal class Reply
+        {
+            public Reply(IRouteeType routeeType, IActorRef actorRef)
+            {
+                RouteeType = routeeType;
+                ActorRef = actorRef;
+            }
+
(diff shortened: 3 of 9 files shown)


build.fsx
```diff
@@ -323,7 +323,7 @@ Target "MultiNodeTests" <| fun _ ->
                 |> append assembly
                 |> append "-Dmultinode.enable-filesink=on"
                 |> append (sprintf "-Dmultinode.output-directory=\"%s\"" testOutput)
-                |> appendIfNotNullOrEmpty spec "-Dmultinode.test-spec="
+                |> appendIfNotNullOrEmpty spec "-Dmultinode.spec="
```
Aaronontheweb: Allows the ability for the MNTR to correctly run a single spec from the FAKE commandline, i.e.

```
 ./build.cmd multinodetests spec=ClusterBroadcastGroup
```

src/core/Akka.Cluster.Tests.MultiNode/Routing/ClusterBroadcastRouter2266BugfixSpec.cs
```diff
@@ -0,0 +1,200 @@
...
+using Akka.Actor;
+using Akka.Cluster.TestKit;
+using Akka.Configuration;
+using Akka.Remote.TestKit;
+using Akka.Routing;
+using FluentAssertions;
+
+namespace Akka.Cluster.Tests.MultiNode.Routing
+{
+    public class ClusterBroadcastGroupSpecConfig : MultiNodeConfig
```
Aaronontheweb: Multinode spec from @mrrd, based on the round-robin group clustered router MNTR spec. 

Verifies that the router has the correct number of routees and that ALL of them receive each message.

src/core/Akka.Cluster.Tests/ClusterDeployerSpec.cs
```diff
@@ -85,6 +94,26 @@ public void RemoteDeployer_must_be_able_to_parse_akka_actor_deployment_with_spec
             deployment.Dispatcher.ShouldBe("mydispatcher");
         }
 
+        [Fact]
+        public void BugFix2266RemoteDeployer_must_be_able_to_parse_broadcast_group_cluster_router_with_default_nr_of_routees_routees()
```
Aaronontheweb: Reproduction spec which verifies that we weren't parsing the correct number of routees originally: https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Cluster/Configuration/Cluster.conf#L209

src/core/Akka.Cluster/ClusterActorRefProvider.cs
```diff
@@ -121,7 +121,7 @@ public override Deploy ParseConfig(string key, Config config)
                 && !config.HasPath("nr-of-instances"))
             {
                 var maxTotalNrOfInstances = config
-                    .WithFallback(ClusterConfigFactory.Default())
+                    .WithFallback(Default)
```
Aaronontheweb: Fixes the bug reported in #2266 - correctly pulls in the `akka.actor.deplyoment.default` section and has us search for the default `cluster.max-total-nr-of-instances` inside of there, which fixes the issue.

src/core/Akka/Actor/Deployer.cs
```diff
@@ -16,17 +16,20 @@
 
 namespace Akka.Actor
 {
+    /// <summary>
+    /// Used to configure and deploy actors.
+    /// </summary>
     public class Deployer
     {
-        private readonly Config _default;
+        protected readonly Config Default;
```
Aaronontheweb: Needed to make the `Default` section visible to child classes.

Aaronontheweb:
Have a consistent cluster client spec failure... Not sure if it's just an issue with that spec or if I did something to affect it in this PR. Going to take a closer look now.

Aaronontheweb:
Looks like it's just an issue with the spec itself that's unrelated to these changes; was able to get it to pass locally and I reviewed the code as well and there's nothing in the `ClusterClient` that would be affected by these changes.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -591,6 +591,7 @@ namespace Akka.Actor
     }
     public class Deployer
     {
+        protected readonly Akka.Configuration.Config Default;
```
Horusiath: Does it need to be here as an instance field? It looks repetitive - maybe place it somewhere else or use as a static?

Aaronontheweb: It was a private field before. Doesn't need to be static since there's only ever 1 instance per `ActorSystem`, but multiple actor systems could have different values for this field depending on which plugins they loaded (although it looks like only Akka core and Akka.Cluster really use it at the moment.)

Aaronontheweb:
@Danthar doh - looks like the new "Update Branch" button Github added on PRs creates a merge commit. I'll rebase that...

Aaronontheweb:
Rebased

alexvaluyskiy:
@Aaronontheweb rebase it again, please

Aaronontheweb:
@alexvaluyskiy done

Danthar:
mono CI failing is expected?
And we have a failing unit test:  RoutingSpec.Routers_in_general_must_evict_terminated_routees

Aaronontheweb:
Mono specs failed because of https://github.com/xunit/xunit/issues/979 - might be another issue where the Mono runtime can randomly crap out inside XUnit, similar to this bug that I reported: https://bugzilla.xamarin.com/show_bug.cgi?id=44546

As for the routee evicted spec - looks like it just timed out. Can happen sometimes when it gets busy. None of the code here affects local routers since they were already using the configuration data that now also gets exposed to clustered routers.

Issue #967, opened 2015-05-11
RoundRobinPool constructor API inconsistency


Danthar:
I noticed the `RoundRobinPool`, when compared with Akka, has an extra constructor. This has been noticed before, and remarked by @Aaronontheweb that it probably should be removed.

akkadotnet reference:
https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Routing/RoundRobin.cs

scala reference:
https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/routing/RoundRobin.scala

So im creating this issue so that its not forgotten with the next big release. This issue should probably be marked towards some future release.

jordansjones:
The extra constructor was added to for unit testing purposes: 7ad8c505ae579efce07e15f405a37557a3d72224

Danthar:
Sorry i should have been more specific. 
@jordansjones You are referring to the constructor which accepts the nrOfInstances. That one is not the issue. That constructor is also in Akka.

Im referring to this one:

          /// <summary>
        /// Simple form of RoundRobin constructor
        /// </summary>
        /// <param name="nrOfInstances">The nr of instances.</param>
        /// <param name="resizer">A <see cref="Resizer"/> for specifying how to grow the pool of underlying routees based on pressure</param>
        public RoundRobinPool(int nrOfInstances, Resizer resizer) : base(nrOfInstances, resizer, DefaultStrategy, null) { }

Danthar:
Closed as "Wont Fix". See corresponding PR for more info

PR #2329, opened 2016-09-28
Removed inconsistent constructor from RoundRobinPool per #967

guessmyname:
Corrected associated tests.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -4139,7 +4139,6 @@ namespace Akka.Routing
     {
         public RoundRobinPool(Akka.Configuration.Config config) { }
         public RoundRobinPool(int nrOfInstances) { }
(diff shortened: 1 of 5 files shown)


Danthar:
This change modifies the API. Therefore needs approval: http://getakka.net/docs/akka-developers/public-api-changes

Danthar:
@Aaronontheweb or @Horusiath Could one of you chime in on this. I would like to have some more consensus on merging this PR (provided the mono build completes). Since it has impact on the public API, albeit minor.

alexvaluyskiy:
@guessmyname why this constructor is inconsistent?

Danthar:
@alexvaluyskiy  This PR is about: https://github.com/akkadotnet/akka.net/issues/967

Aaronontheweb:
I don't think this PR adds any value really, even if the constructor is inconsistent. Just mark it as deprecated.

alexvaluyskiy:
Scala has much more powerful default parameters and named parameters features. That's why Scala has only one constructor, but we should have a couple of overloaded constructors

Danthar:
Closing this as a "wont fix" @guessmyname thx for the effort. A valuable discussion was had. And this issue can now be removed from our list.

PR #2330, opened 2016-09-30
Added WeaklyUp cluster member status

Horusiath:
This is still experimental feature on Akka JVM, and for compatibility reasons needs activation using `akka.cluster.allow-weakly-up-members = on` setting.

In theory to add new cluster nodes, cluster must reach convergence. That means, in situation where there are some unreachable nodes, new node can join the cluster, but it won't be promoted to **Up** status (as unreachable nodes are not able to see new joining nodes, so the cluster view state won't be consistent).

When enabled, this feature allows to mark incoming nodes as **WeaklyUp** even thou cluster itself may have not reach the convergence yet. Once unreachable members become reachable again, all weakly up nodes are promoted to Up. This way they may be used without being visible for all nodes in the cluster. This is especially useful for big clusters, where time necessary to propagate new cluster state is quite long and eventual churn of joining/leaving nodes is big.

Weakly up nodes should not be used in some specific situations i.e. quorum decisions.


alexvaluyskiy:
Cluster protobuf serialization should be changes in order to support WeaklyUp

alexvaluyskiy:
WeaklyUp should be added to DistributedPubSub https://github.com/akka/akka/blob/f0f755bf08936f1ba79e577837356ffbaf635aa6/akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala#L675

and into the spec: MinMembersBeforeUpSpec.cs

alexvaluyskiy:
@Horusiath  Akka.Cluster.Tests.MultiNode.MemberWeaklyUpSpec did not pass

alexvaluyskiy:
@Horusiath Could you rebase it?

Issue #2346, opened 2016-10-21
Cluster member cannot Leave when status is Joining


ayoung:
I'm finding that a member cannot `Leave()` the cluster if the status is `MemberStatus.Joining`. It is only allowed to leave if it is `MemberStatus.Up`. 

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Cluster/ClusterDaemon.cs#L1195
if you call Leave() when status is Joining, the member is considered unreachable by the cluster indefinitely.

PR #2347, opened 2016-10-21
Down if MemberStatus.Joining

ayoung:
This pull is intended to fix issue #2346.

I noticed that if a cluster member is in `Joining` state when you call `cluster.Leave()`, the member will appear unreachable to other nodes indefinitely. This adds a check to see if the state is `Joining`, then call `Downing()` instead.

The only caveat to this is that if you have a callback waiting on `cluster.RegisterOnMemberRemoved(someAction)` then this won't be called. Is there some other event that should be hooked into for this?


Aaronontheweb:
Wondering how we can test this to ensure that it works....

Danthar:
Status on this one? The change itself looks good. @Aaronontheweb

Aaronontheweb:
I think this one still needs a spec to verify that its behavior is inline with the cluster specifications

Aaronontheweb:
resolved via https://github.com/akkadotnet/akka.net/pull/3551

PR #2349, opened 2016-10-25
Updated Ask user cancellation rebase

ZigMeowNyan:
This is taking Zetanova's commits from [PR #1632](https://github.com/akkadotnet/akka.net/pull/1632) in January, rebasing and doing all the merging to keep them current against the dev branch, and adding in the [changes](https://github.com/akkadotnet/akka.net/pull/1632#issuecomment-193860389) necessary to make the public API tests pass. [Additionally](https://github.com/akkadotnet/akka.net/pull/1632#issuecomment-253962723), one untested optimization was left out in the merge, and a test was renamed to follow adjustments to naming conventions.

The purpose is to allow the `Ask` extensions on `ICanTell` to be cancelled externally beyond just waiting for timeouts.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -837,7 +837,11 @@ namespace Akka.Actor
     public class static Futures
     {
         public static System.Threading.Tasks.Task<object> Ask(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout = null) { }
+        public static System.Threading.Tasks.Task<object> Ask(this Akka.Actor.ICanTell self, object message, System.Threading.CancellationToken cancellationToken) { }
+        public static System.Threading.Tasks.Task<object> Ask(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout, System.Threading.CancellationToken cancellationToken) { }
         public static System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout = null) { }
+        public static System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Threading.CancellationToken cancellationToken) { }
+        public static System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout, System.Threading.CancellationToken cancellationToken) { }
     }
     public class static GracefulStopSupport
     {
--- a/src/core/Akka.Tests/Actor/AskSpec.cs
+++ b/src/core/Akka.Tests/Actor/AskSpec.cs
@@ -86,6 +86,15 @@ namespace Akka.Tests.Actor
             Assert.Throws<AggregateException>(() => { actor.Ask<string>("timeout", TimeSpan.FromSeconds(3)).Wait(); });
         }
 
+        [Fact]
+        public void CanCancelWhenAskingActor()
+        {            
+            var actor = Sys.ActorOf<SomeActor>();
+            var cts = new CancellationTokenSource(TimeSpan.FromSeconds(3));
+            Assert.Throws<AggregateException>(() => { actor.Ask<string>("timeout", Timeout.InfiniteTimeSpan, cts.Token).Wait(); });
+            Assert.True(cts.IsCancellationRequested);
+        }
+
         /// <summary>
         /// Tests to ensure that if we wait on the result of an Ask inside an actor's receive loop
         /// that we don't deadlock
--- a/src/core/Akka.Tests/Actor/AskSpec.cs
+++ b/src/core/Akka.Tests/Actor/AskSpec.cs
@@ -87,7 +87,7 @@ namespace Akka.Tests.Actor
         }
 
         [Fact]
-        public void CanCancelWhenAskingActor()
+        public void Can_cancel_when_asking_actor()
         {            
             var actor = Sys.ActorOf<SomeActor>();
             var cts = new CancellationTokenSource(TimeSpan.FromSeconds(3));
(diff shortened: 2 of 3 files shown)


Danthar:
LGTM

Aaronontheweb:
Needs to be updated: http://stackoverflow.com/questions/454734/how-can-one-change-the-timestamp-of-an-old-commit-in-git

Need to compress the timestamp so it gets picked up by CI.

Danthar:
@ZigMeowNyan Could you take a look at updating this PR ? We would like to get it in eventually :)

eisendle:
Pls keep in mind to consider PR #2381 when merge this PR, because it still contains a memory leak.

ZigMeowNyan:
Got involuntarily sucked into a huge timesink, but I'm looking at this now. The test details on the CI server aren't loading for me for some reason. I've spent 5 minutes watching the "Loading test details. Please wait.." spinner to no avail. So I'm running the tests locally to see if I can figure out what's wrong in those tests. Hopefully I can finish this off before the end of the month.

@eisendle It looks like PR #2381 got reverted 11 days ago in PR #2399 due to unreasonable log sizes. I just noticed this as I was rebasing. Not sure when that'll be resolved, but I'll take the reduced concerns for the moment.

ZigMeowNyan:
So, from what I can see, the failed tests were caused by the cleanup continuations. It seems either Zetanova put his own in, and then someone else also added them later, or I screwed up the merge. The result was that `timeoutCancellation` could be Disposed twice. I'm disinclined to go on a discovery voyage over 80+ commits, so I'll just assume it was me.

Beyond that, rather than having three possible cleanup continuations, I want to do them in one, but there can be 0-2 `CancellationTokenRegistration` objects lying around depending on timeouts and cancellation tokens. It's a struct, too, so null checks are out and I'm not sure about Default(T) comparison results. I went with a small List<T>, but I'm not sure of the performance impact of that on the code.

I'd appreciate some feedback on the final change to Futures.cs to see if there's a particular style preferred to dealing with this.

Side note: it took me a while to go through the tests, as it crashes my test runner locally about 1/3-1/2 through the test collection. I also get several "`System.AppDomainUnloadedException: Attempted to access an unloaded AppDomain. This can happen if the test(s) started a thread but did not stop it. Make sure that all the threads started by the test(s) are stopped before completion.`" errors when running the tests, which seems a bit odd, but maybe my VS install is "special". Has anyone else seen this behavior?

Aaronontheweb:
Looks good to me. Rebased it. Will merge once tests have had a chance to finish running.

Aaronontheweb:
Nice work!

Aaronontheweb:
Just to be clear, #2381 is covered by this also, right @ZigMeowNyan ?

ZigMeowNyan:
> Just to be clear, #2381 is covered by this also[?]

@Aaronontheweb, That PR was both merged and then reverted before I got around to dealing with the test failures. Not sure if it got resubmitted, but I imagine it would have complained more about the rebase in that situation. However, if I understand it right, the crux of it was the call to `provider.UnregisterTempActor(path);` which is now solely in the `unregister` action at [L116 in Futures.cs](https://github.com/akkadotnet/akka.net/pull/2349/commits/c525f3d575e01ec603b57719eb00f45bf71a24be#diff-087cfe232c174bfb770e0de44390a722R116).

The way Zetanova structured the code additions was by routing everything through the same code path via overloads (whether using a `CancellationToken` or a timeout. My contribution was just the merging and putting all of the `CancellationTokenRegistration` cleanup in the `unregister` action. I grabbed it because it was old, mergeable code that touched relatively little and it was something I wanted, but couldn't do as easily outside of the project base due to code visibility modifiers.

I'm not sure what logic calls the `unregister` action (it's not present in the current slice of code in any of the commits I looked through, so it's probably inside `FutureActorRef`, which is tucked inside another file), but if it's always called _except when cancelled_, perhaps a new continuation (with `TaskContinuationOptions.OnlyOnCanceled`) that just calls unregister() would give us the needed functionality.  On the other hand, it may be that we can adjust the original call to the `unregister` action to make sure that it also runs on cancellation. I like this idea more - and it has the benefit of making `FutureActorRef`'s behavior more predictable. Doing extra cleanup for FutureActorRef outside of it seems like it will add to the maintenance burden. But either way, we would still have one place where we clean up the tokens and unregister the temp actor instead of sprinkling different unregistration strategies around the codebase.

Aaronontheweb:
@ZigMeowNyan should I revert then?

Aaronontheweb:
oh snap, you meant the other one. Nevermind.

ZigMeowNyan:
> oh snap, you meant the other one.

Correct. I didn't try to fix #2381 since it had been reverted with cause, as it might just get this commit reverted.

> > if it's always called except when cancelled, perhaps a new continuation (with TaskContinuationOptions.OnlyOnCanceled) that just calls unregister() would give us the needed functionality.

@Aaronontheweb, It looks like we could do this in the [`FutureActorRef` constructor](https://github.com/akkadotnet/akka.net/blob/d86eb6eb4fe512aac603cbc20bddc420acaec74a/src/core/Akka/Actor/ActorRef.cs#L60) by just adding a continuation (default options) that calls `_unregister()` instead of [manually calling it](https://github.com/akkadotnet/akka.net/blob/d86eb6eb4fe512aac603cbc20bddc420acaec74a/src/core/Akka/Actor/ActorRef.cs#L99). This way, unregister shouldn't get missed - whether it's after `_result.TrySetResult(message);` or if the task gets cancelled/faulted. It makes more sense to me for `FutureActorRef` itself to ensure that it calls the `unregister` action if it's aware of in the first place.

How does this approach sound for dealing with #2381?

Aaronontheweb:
@ZigMeowNyan sounds good to me - should be easy to check the NBench logs to see if there's a problem next time around.

ZigMeowNyan:
> sounds good to me

Ok. you want to tackle that, or should I?

Also, this current PR reintroduces the code from #1632, so we can thank Zetanova and close that.

Aaronontheweb:
@ZigMeowNyan you should do it :p

ZigMeowNyan:
> you should do it :p

It was worth a shot.  ¯\\_(ツ)_/¯
Anyway, there's a new PR for you (#2422) that should tackle that (consistent unregistration from inside `FutureActorRef`).

PR #2351, opened 2016-10-26
Use of dynamic method resolution

Zetanova:
To use GetMethod + Invoke on each message looks like a major performance hit (unproofen)


Danthar:
It seems this change is causing issues in the MNTK.

Aaronontheweb:
Need to approve API changes

rogeralsing:
I think dynamic cannot deal with internal or nested classes. making things break if you pass such a message.
Not 100% on the details around this but there is such an issue in Wire where dynamic is used.
https://github.com/rogeralsing/Wire/issues/117

I vote for close.
The correct thing to do here is to code generate this if TypedActor is to continue to be used.

Danthar:
@rogeralsing isn't that only related to the fact that in Wire you need to serialize/deserialize the type ?
That is not the case here.

I manually compared the performance results of 3 historical builds for 20 akka.streams perf tests, with the results of this PR. And im either seeing the same performance, or small improvements. 
A better comparison might be warranted, but im not doing that by hand :P maybe @Aaronontheweb knows some NBench-fu to take perflogs and generate an comparison table ?

PR #2356, opened 2016-10-31
Fix Serialization KeyNotFoundException when looking up serializers

ayoung:
A `KeyNotFoundException` can occur if the serializer dll is not present in the search paths for binaries. This can occur when there is a proj A that depends on B and B depends on Akka. There is no direct reference to any of the serializers in either A or B besides being referenced in HOCON. MSBuild likes to optimize dlls out of the output if they are not directly referenced in code.

### DIFF
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -80,7 +80,10 @@ namespace Akka.Serialization
                     continue;
                 }
 
-                var serializer = namedSerializers[serializerName];
+                Serializer serializer;
+
+                namedSerializers.TryGetValue(serializerName, out serializer);
+
                 if (serializer == null)
                 {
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
@ayoung I like it. Seen this come up when working with Akka.Persistence and using Wire only.

ayoung:
@Aaronontheweb Yes. That was precisely my problem. Spent about an hour trying to figure out what was happening. What is the common solution for this problem? Currently I've been having to do dirty things like this in my code.

```c#
public static readonly Type[] References =
{
    typeof (IActorRef),
    typeof (Akka.Cluster.Cluster),
    typeof (Akka.Serialization.WireSerializer)
};
```

ayoung:
Also, its not a NullRefEx. Its a `KeyNotFoundException`. I'll update soon.

PR #2366, opened 2016-11-05
fixeded cluster sharding demo

Aaronontheweb:
Cluster.Sharding demo would power-vomit upon startup due to incorrect SQLite configuration and missing database. This PR fixes that.


Aaronontheweb:
Rebased, CI passed, then merged

PR #2369, opened 2016-11-05
added Cluster.LeaveAsync for graceful exit

### LINKED ISSUE #2280: Cluster.LeaveAsync()
Currently, to leave cluster before termination one needs to subscribe to `RegisterOnMemberRemoved`, do `Leave(cluster.SelfAddress)` and wait for member removal event.

This pattern is very easy to implement wrong. There are also edge cases when event doesn't guarantee arrive (member is in Joining state etc.), thus very easy to hang the process.

Situation becomes more complicated with support of CancellationToken etc.

I'd want to propose to add a convenience method `Task LeaveAsync()` into the Cluster.

The example code could look like this:

``` c#
// with or without cancellation token
await Cluster.Get(_actorSystem).LeaveAsync(cancellationToken);
await _actorSystem.Terminate();
```

Naive implementation as extension method, just for reference:

``` c#
public static Task LeaveAsync(this Cluster cluster)
{
    var tcs = new TaskCompletionSource<bool>();

    cluster.RegisterOnMemberRemoved(() => { tcs.TrySetResult(true); });
    cluster.Leave(cluster.SelfAddress);

    return tcs.Task;
}
```

Aaronontheweb:
close #2280 

Also tried to implement the `leave-on-termination` feature in HOCON, but unfortunately the `TerminationHooks` implemented in `ActorSystem` don't support this - they always run *after* the actors have shut down. We'd need something that runs *before* the ActorSystem has shut down the actors to implement this correctly.

Therefore, I think the smartest thing to do is when developing an Akka.Cluster application, always call `Cluster.LeaveAsync` -> `ActorSystem.Terminate` in that order when you're ready to shut down.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCluster.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCluster.approved.txt
@@ -28,6 +28,7 @@ namespace Akka.Cluster
         public void Join(Akka.Actor.Address address) { }
         public void JoinSeedNodes(System.Collections.Generic.IEnumerable<Akka.Actor.Address> seedNodes) { }
         public void Leave(Akka.Actor.Address address) { }
+        public System.Threading.Tasks.Task LeaveAsync() { }
         public void RegisterOnMemberRemoved(System.Action callback) { }
         public void RegisterOnMemberUp(System.Action callback) { }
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Once this is merged in, I want to take a look at #2347 and make some changes there to ensure that `LeaveAsync` can exit even in the event that a node failed to join.

Aaronontheweb:
Squashed and merged all commits after being reviewed positively by @Horusiath and rebased.

PR #2374, opened 2016-11-10
Get events from sql journal db in right sequence

andreyleskov:
After Akka.Persistence upgrade to 1.1.2 sql journal started to fetch events with undetermined order, 
not caring about SequenceNr column. 

After we changed PK from (PersistenceId, SequenceNr) to (Ordering) raws in table no longer ordered by SequenceNr, so ordering should be declared explicitly


Aaronontheweb:
Thanks for your contribution @andreyleskov

PR #2376, opened 2016-11-12
Fix CircuitBreaker swallowing exceptions cased it to open

andreyleskov:
CircuitBreaker swallowing stack trace of exception cased it to become Open. 
It makes debug harder

### DIFF
--- a/src/core/Akka/Pattern/CircuitBreaker.cs
+++ b/src/core/Akka/Pattern/CircuitBreaker.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


Horusiath:
@Aaronontheweb the only breaking thing is a racy test for cluster. Can we merge that anyway?

Aaronontheweb:
@Horusiath really weird that didn't work on the build server... I wonder why. Have to assume it's an issue with that VM size and Docker container being more sensitive to degrees of parallelism on that VM.

Aaronontheweb:
thanks for your help @andreyleskov !

Issue #2348, opened 2016-10-21
Router cannot be created from fallback configuration


eisendle:
I have two HOCON configuration files A and B. B is configured as fallback of A. B contains the deployment definition for a router. However if I try to create the router with `Props.Empty.WithRouter(FromConfig.Instance)` I get a configuration error that the router configuration is missing. If I put the router definition in configuration file A, everything works like expected, the router is created without an error.

I already checked the code. Looks like Deployer.cs loads the section **akka.actor.deployment** in its constructor, which does not consider the fallback configuration.

Danthar:
Can you share how you created your config and passed it into your ActorSystem instance ?

eisendle:
Sure. I´ve created a gist which shows the issue [here](https://gist.github.com/eisendle/2f8f92e2f9bf734093782fda7a3aa51b). In the example creation of **workerB** fails with an ConfigurationException. However, **workerBConfig** contains the valid configuration for the router.

kantora:
There is an old bug. I have even provided test to reproduce it.
https://github.com/akkadotnet/akka.net/issues/1321

kerch-ai:
Trying to learn the inner workings of akka.net. Not 100%, but it looks like Deployer only looks at the root config. Should it also take into account the fallaback config? Will create pull request with my solution

Danthar:
Fixed via PR

Issue #1087, opened 2015-06-23
Add concise, clear logging around disassociation messages


Aaronontheweb:
Lack of good information around disassocations is making it very difficult to debug Akka.Remote and Akka.Cluster. We need to add more details around why a disassociation occurred as well as how endpoint policies are being applied.

Aaronontheweb:
Just spent some time in a new branch working on this. Cleaned up the `ProtocolStateActor` - it's not really responsible for any of the `disassociate` log messages so I thought I'd post that here and save others the trouble :p

Telavian:
I am running into a bunch of disassociation messages which are very sporadic at first and then constant after a few days of running. How would I go about troubleshooting this? I understand failed heartbeats should not be happening however when they do what does that really mean when both sides are running? A simple restart makes everything magically work again.

http://stackoverflow.com/questions/31753052/akka-net-remote-disconnection

Aaronontheweb:
After talking @rkuhn, shutdown messages should be logged as `DEBUG` - not `ERROR`

maxcherednik:
Since it's about clean logging, I can try to look at. Also cleaned some known exception logging in #2438

Danthar:
@maxcherednik sounds good! 👍

Aaronontheweb:
Done as part of 1.1.3

PR #2379, opened 2016-11-16
Messages are logged recursively with custom logger.

boriskreminskimoldev:
When "loglevel = DEBUG" and "receive = on" are set in HOCON and custom
logger is in use, messages are logged recursively. In addition, some of
the messages are not implementing IAutoReceivedMessage, although seem to
be sent to Self, causing the log to be noisy.


src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1385,7 +1385,7 @@ public sealed class FlushAndStop
             public static FlushAndStop Instance { get { return _instance; } }
         }
 
-        public sealed class AckIdleCheckTimer
+        public sealed class AckIdleCheckTimer: IAutoReceivedMessage
```
Danthar: Im not a fan of this. Its an internal message, and its seems very random to only mark this message as `IAutoReceivedMessage`. (Which would cause it to also have `IDeadLetterSuppression`).

src/core/Akka.Remote/EndpointManager.cs
```diff
@@ -209,7 +209,7 @@ public ManagementCommandAck(bool status)
 
         #region Messages internal to EndpointManager
 
-        public sealed class Prune : INoSerializationVerificationNeeded { }
+        public sealed class Prune : INoSerializationVerificationNeeded, IAutoReceivedMessage { }
```
Danthar: `IAutoReceivedMessage ` already also implements: `INoSerializationVerificationNeeded`.
And again it seems rather random, to mark this message, but leave the others untouched.

Danthar:
Ok, if I understand it correctly this PR does 2 things. First it attempts to fix the recursion problem when `receive = on`. And second it attempts to have some messages be filtered out from logging, by marking them as `IAutoReceivedMessage`.

In my mind these are 2 separate issues. Lets focus on the recursion problem first and reduce this PR to the changes in the `ActorCell.DefaultMessages.cs` file. 
Do like the extended log msg though. 

I have run a local build with only the changes in 'ActorCell.DefaultMessages'. And that seemed to resolve the test errors.

src/core/Akka.Remote/Endpoint.cs
Aaronontheweb: Broke several specs inside Akka.Remote as well

boriskreminskimoldev:
@Danthar  and @Aaronontheweb: thanks for reviewing. I agree, these are two separate issues. I will roll back all the changes except those in ActorCell.DefaultMessages. Question: how do I make sure I'm not breaking the specs before commit? Run Build AllTests?

Danthar:
After doing BuildRelease. yes. But fair warning. It will take some time ^^

boriskreminskimoldev:
Looks like its's passing locally, but I'm not sure whats going on with the checks. Are these errors related to my submission?

Danthar:
We have quite a few akka.streams performance tests failing. Some are just outside their margin, or with a wide margin. 

I think that's related to your new message.format statement. Because that bit of code is in a hot-path its important its as fast as possible. I restarted the perf tests to verify the issues. 

The perf problems are two-fold. First perf tests currently has a hard margin. So if it 1ms to long, it fails. Which can be annoying. This problem is known with the team, and already discussed. 
Second: Some tests are actually way outside their margin, indicating that your change really does has an impact. 

All i can say is: Try and improve the perf characteristics of your code. In this case since the string format arguments are fixed. A simple string concatenation will probably be optimized by the compiler, making it faster. But that is an assumption that needs testing.

The MNTK failure is (seemingly) unrelated, and I already reported it internally.

boriskreminskimoldev:
Tried to get rid of string.Format in favor of string concatenation. Not sure if this caused performance improvement, though. windows-perf-tests seem to be passing.

Danthar:
Ok now we got some real MNTK tests failing:

AttemptSysMsgRedeliveryMultiNetSpec.AttemptSysMsgRedelivery

And with the  TcpListenerSpec.A_TCP_Listner_must_continue_to_accept_connections_after_a_previous_accept.
Its a timeout on 3 secs. It might be fixed if we relaxed that timeout a little bit but you'd have to test it.

boriskreminskimoldev:
Reversed the condition, as it appears this is not a bug in ActorCell, but in my code (sorry about that). However, still want to leave the change for extended debug output, as discussed earlier today on gitter (#2408).

Danthar:
And suddently the failing unit tests started working again ;)

boriskreminskimoldev:
Yes, this was my first timer's journey :). Thanks for reviewing and accepting it! I would also spend some time to  update the documentation page, such as this one http://getakka.net/docs/Logging, just to mention that in order to see the received messages from particular actor, this actor has to implement the marker ILogReceive interface (as mentioned here: http://gigi.nullneuron.net/gigilabs/actor-logging-with-akka-net/.
Now, if possible, I would like to keep contributing and pick one of the open issues, something lightweight to begin with. How about #1087? @Danthar , what do you think?

Danthar:
You are welcome to pick up whatever you want. #2348 is also an contended, although abit more work i bet.

Issue #2380, opened 2016-11-17
FutureActorRef leak when ask operation times out


eisendle:
After analysing a memory leak in my akka application, i´ve found that when an ask operation hits a time out, the created temp actor is never removed from the temp container.

Aaronontheweb:
@eisendle sounds like a possible bug; could you add a leak spec for it?

eisendle:
Added  a spec to the PR.

Aaronontheweb:
Need to address:

https://github.com/akkadotnet/akka.net/issues/1448
https://github.com/akkadotnet/akka.net/pull/2399

As part of this. Failed our performance test suite, but moreover: creates insane amounts of logs. Hundreds of MBs each.

Danthar:
@Aaronontheweb  Whats the status on this ?

Aaronontheweb:
@Danthar resolved by #2422

PR #2381, opened 2016-11-17
Unregister temp actor if ask operation got canceled.

eisendle:
Fixes issue #2380. Added call to UnregisterTempActor in case of cancellation.


eisendle:
Added a spec to verify the resource leak

Aaronontheweb:
Why in the hell do our perf tests keep failing? Let's relax the time constraints for those if this memory leak cleanup hurts performance. Would rather trade "correctness" over speed here.

Issue #2368, opened 2016-11-05
Akka.Cluster: change default system logger to always include the Cluster.SelfAddress in the path of the actor who logged message


Aaronontheweb:
This will make life infinitely better for people using the multi-node testkit as well as anyone who has to compare logs from actors across multiple nodes.

Aaronontheweb:
closed by #2791

PR #2384, opened 2016-11-20
 Allow to pass cluster info into ToString method of actor

alexpantyukhin:
#2368 
I'm not sure about this approach. Just an idea how information about cluster could be reflected in the `ToString` method. @Aaronontheweb  could you review it please and tell what do you think about it?


Aaronontheweb:
Have some thoughts on this  - don't think we should be rewriting this data inside the `IActorRef` necessarily... Might be better to do this by switching out the default logger (if possible) instead to append this data... Going to noodle on it.

alexpantyukhin:
Yeah.. agree that it would be better to do not touch `IActorRef` if it is 100% not needed. But I didn't find the way how to pass info about cluster into the logger when log actor explicitly.

Aaronontheweb:
thanks for your contribution @alexpantyukhin but we ended up implementing this in #2791

PR #2389, opened 2016-11-29
Batching SQL event journal

Horusiath:
This is a new approach to persisting events.

## Motivation

Currently all async journals handle incoming request 1-by-1, as they come from persistent actors. This means, that in case of SQL we're creating a separate `DbConnection` every time a persistent actor sends a request. Given i.e. 10 000 actors trying to persist events at the same time we hit a problem - we can saturate the entire connection pool in a blink of an eye.

## New approach

With new approach I'm using a horizontal batching - executing requests from multiple different actors using one db connection. We are defining 3 things:

1. A maximum number of concurrent operations. This can be equal to connection pool size - 1, but we can also make it lower i.e. if we want to partition connection pool between akka journal and user defined code. It's 64 by default.
2. A request buffer - as we have operations limit we store incoming requests inside a buffer/queue. Once an operation is completed and new connection can be assigned, we empty the buffer. Difference is that we don't assign a connection per request, but for a many of them. It's 500 000 by default.
3. Maximum chunk size - this is a maximal number of requests that can be executed using a single db connection. It's 100 by default.

## Improvements

New implementation passes persistence testkit already, including persistence query. I've also modified persistence benchmark to be able to run stress tests. Performance results so far (SQL Server on default settings):

- Given 1000 actors trying to persist 100 events:
    - old implementation: ~571 writes/sec
    - new implementation: ~4219 writes/sec
- Given 10 000 actors trying to persist 100 events:
    - old implementation broke here
    - new implementation: ~9500-10500 writes/sec

Also with the new implementation 10 000 actors recovering from 101 events each (so 1010 000 events read in total) took around ~10-14 sec. (depending if replay filter feature is set or not).

I didn't noticed connection pools over-saturation any longer (this is by design), but I needed to remove part of code associated with cancellation token - it was triggering spontaneously without reason (internal implementation bug?).


Danthar:
How does this change the assumptions a user can make regarding the Persistence API ?
By that I specifically mean the `Persist` method where one can assume the event has been saved at the time the handler is being invoked.

Does that assumption still holds with the current implementation ?

Horusiath:
@Danthar without it `Persist` would be unreliable ;) New journal doesn't change anything in this regard.

src/contrib/persistence/Akka.Persistence.Sqlite/Journal/BatchingSqliteJournal.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+        }
+
+        public BatchingSqliteJournal(BatchingSqliteJournalSetup setup) : base(setup)
+        {
+            var conventions = Setup.NamingConventions;
+            Initializers = ImmutableDictionary.CreateRange(new[]
+            {
+                new KeyValuePair<string, string>("CreateJournalSql", $@"
+                CREATE TABLE IF NOT EXISTS {conventions.FullJournalTableName} (
```
Aaronontheweb: Is this backwards compatible with previous SQL journals? (the schema, that is)

src/contrib/persistence/Akka.Persistence.Sqlite/packages.config
```diff
@@ -1,4 +1,13 @@
...
+  <package id="System.Collections" version="4.0.0" targetFramework="net45" />
+  <package id="System.Collections.Immutable" version="1.1.36" targetFramework="net45" />
+  <package id="System.Data.SQLite.Core" version="1.0.98.0" targetFramework="net45" />
+  <package id="System.Diagnostics.Debug" version="4.0.0" targetFramework="net45" />
+  <package id="System.Globalization" version="4.0.0" targetFramework="net45" />
+  <package id="System.Linq" version="4.0.0" targetFramework="net45" />
+  <package id="System.Resources.ResourceManager" version="4.0.0" targetFramework="net45" />
+  <package id="System.Runtime" version="4.0.0" targetFramework="net45" />
+  <package id="System.Runtime.Extensions" version="4.0.0" targetFramework="net45" />
+  <package id="System.Threading" version="4.0.0" targetFramework="net45" />
```
Aaronontheweb: Why are all of these System packages needed? .NET Core package format?

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/MultiValueDictionaryExtensions.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System.Collections.Generic;
+
+namespace Akka.Persistence.Sql.Common.Journal
+{
+    internal static class MultiValueDictionaryExtensions
```
Aaronontheweb: XML-DOC comments please @sean-gilliam

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
```diff
@@ -141,11 +141,11 @@ protected override bool ReceivePluginInternal(object message)
                     var batch = new WriteJournalBatch(eventToTags);
                     await QueryExecutor.InsertBatchAsync(connection, _pendingRequestsCancellation.Token, batch);
                 }
-            });
+            }).ToArray();
```
Aaronontheweb: Does this affect any sort of lazy evaluation at all?

Aaronontheweb: Looks like it doesn't given the code below.

src/contrib/persistence/Akka.Persistence.Sqlite/Journal/BatchingSqliteJournal.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.Collections.Immutable;
+using System.Configuration;
+using System.Data.Common;
+using System.Data.SQLite;
+using Akka.Configuration;
+using Akka.Persistence.Sql.Common.Journal;
+
+namespace Akka.Persistence.Sqlite.Journal
+{
+    public sealed class BatchingSqliteJournalSetup : BatchingSqlJournalSetup
```
Aaronontheweb: Will this automatically get loaded for existing *.SQL plugins that get updated? I.E. do they have to update their HOCON at all to take care of this?

Aaronontheweb:
@Horusiath other question: any reason this can't be generalized to work with any Akka.Persistence implementation and not just the SQL ones?

src/contrib/persistence/Akka.Persistence.Sqlite/Journal/BatchingSqliteJournal.cs
Horusiath: I should create constructor working with HOCON config. So far, it's up to implementing class to provide config-value mappings. By default all SQL implementations I've created, are working with default sql persistence config (so it's compatible with existing SQL settings). New config values have defaults, in case if user didn't specify them.

src/contrib/persistence/Akka.Persistence.Sqlite/Journal/BatchingSqliteJournal.cs
Horusiath: Yes, schema creation is backward compatible.

Horusiath:
I must check, maybe it's possible to generalize this journal even further.

Horusiath:
I think, that trying to generalize batching journal to cover non-SQL scenarios doesn't have much sense. Some of the internal components are quite specific, namely:

- wrapping batches in DB connection/transaction/command. 
- compatibility with persistent queries for SQL, which use messages not defined in akka.persistence itself (as it's still separate from persistence query module), but still are coupled with a logic used to handle journal operations.

Horusiath:
Things that may come in the future, but are not part of this PR:

- Further generalization of BatchingJournal, so that any provider could use it.
- Hooks for both batch processing and single event processing: potentially useful for bulk inserts, but will need more hacking over SQL parameters naming.
- Manifest based serialization support.

Danthar:
Considering our known issues with streams-perf. I am inclined to merge this. Any objections ? @akkadotnet/core

Aaronontheweb:
@Danthar squashed and merged it - previous PR that was merged was just a documentation update. No need for a rebase.

PR #2391, opened 2016-12-01
[WIP] fix static analized high priority issues

maxim-s:
all high priority issues here https://gist.github.com/maxim-s/41ef6139bebe0e728594f5ded28f2b58

### DIFF
--- a/src/core/Akka.Remote.TestKit/DataTypes.cs
+++ b/src/core/Akka.Remote.TestKit/DataTypes.cs
@@ -434,7 +434,7 @@ namespace Akka.Remote.TestKit
             {
                 int hashCode = (_node != null ? _node.GetHashCode() : 0);
                 hashCode = (hashCode * 397) ^ (_target != null ? _target.GetHashCode() : 0);
-                hashCode = (hashCode * 397) ^ (_direction != null ? _direction.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ _direction.GetHashCode();
                 hashCode = (hashCode * 397) ^ _rateMBit.GetHashCode();
                 return hashCode;
             }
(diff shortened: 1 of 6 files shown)


src/core/Akka.Remote.TestKit/DataTypes.cs
```diff
@@ -434,7 +434,7 @@ public override int GetHashCode()
             {
                 int hashCode = (_node != null ? _node.GetHashCode() : 0);
                 hashCode = (hashCode * 397) ^ (_target != null ? _target.GetHashCode() : 0);
-                hashCode = (hashCode * 397) ^ (_direction != null ? _direction.GetHashCode() : 0);
```
maxim-s: enum can't be null

src/core/Akka.Remote/EndpointManager.cs
```diff
@@ -730,7 +730,6 @@ protected void Accepting()
                         _endpoints.MarkAsQuarantined(gotuid.RemoteAddress, gotuid.Uid,
                             Deadline.Now + _settings.QuarantineDuration);
                         _eventPublisher.NotifyListeners(new QuarantinedEvent(gotuid.RemoteAddress, gotuid.Uid));
-                        Context.Stop(pass.Endpoint);
```
maxim-s: if policy is WasGated Context.Stop(pass.Endpoint) will cause null reference exception
https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/Remoting.scala#L704

src/core/Akka/Actor/ActorCell.FaultHandling.cs
```diff
@@ -59,31 +59,29 @@ private void FaultRecreate(Exception cause)
                 if (System.Settings.DebugLifecycle)
                     Publish(new Debug(_self.Path.ToString(), failedActor.GetType(), "Restarting"));
 
-                if (failedActor != null)
```
maxim-s: failedActor is _actor, _actor already checked on null

src/core/Akka/Actor/Futures.cs
```diff
@@ -40,18 +40,9 @@ public static Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeo
             if (provider == null)
                 throw new ArgumentException("Unable to resolve the target Provider", nameof(self));
 
-            ResolveReplyTo();
```
maxim-s: the same as do nothing

Danthar:
Neat. Nice work @maxim-s

src/core/Akka.Remote/EndpointManager.cs
Aaronontheweb: NICE catch.

PR #2395, opened 2016-12-03
Removed Xunit1 dependencies from Akka.TestKit.Tests and Akka.FSharp.Tests projects

alexvaluyskiy:
Need this for https://github.com/akkadotnet/akka.net/issues/2153

I've moved all files from XUnit2 folder and have replaced the same files from Xunit1
I have kept Akka.TestKit.Xunit. We could remove it later


Aaronontheweb:
thanks @alexvaluyskiy - merged

PR #2396, opened 2016-12-03
Replace Stream.Close to Stream.Dispose

alexvaluyskiy:
Need this for https://github.com/akkadotnet/akka.net/issues/2153


Aaronontheweb:
Didn't bother rebasing a second time, since the only other change merged was XML-DOC.

PR #2398, opened 2016-12-04
disabled Test mode for all Akka.Streams.Tests.Performance benchmarks

### LINKED ISSUE #2397: Relax Akka.Streams Perf Test Assertions
Failing too often and by relatively small margins. Can we relax those? cc @Silv3rcircl3

Aaronontheweb:
closes #2397

This is a temporary solution to a bigger problem. The relativistic / historical benchmarking tools I've planned to add for NBench are really the right solution to this problem, but for the time being: we're having a **TON** of false negatives on our Akka.Streams performance assertions due to small variances that are totally within normal ranges (i.e. being off by a few MS.) Benchmarking concurrent code is a dark art, and doing it with hard limits is predictably erroneous.

For the time being, I've disabled assertions on the benchmarks themselves via the use of constants on the classes. This can easily be toggled on and by changing the constant value `CurrentTestMode` on each individual benchmark class.

CC @Silv3rcircl3 @Horusiath @Danthar any issues with this?

### DIFF
--- a/src/core/Akka.Streams.Tests.Performance/FlowSelectBenchmark.cs
+++ b/src/core/Akka.Streams.Tests.Performance/FlowSelectBenchmark.cs
@@ -45,6 +45,8 @@ akka {
 }
 ");
 
+        private const TestMode CurrentTestMode = TestMode.Measurement;
+
         private ActorSystem _actorSystem;
         private ActorMaterializer _materializer8;
         private ActorMaterializer _materializer32;
@@ -73,7 +75,7 @@ akka {
 
 
 
-        [PerfBenchmark(RunMode = RunMode.Iterations, TestMode = TestMode.Test, NumberOfIterations = 3,
+        [PerfBenchmark(RunMode = RunMode.Iterations, TestMode = CurrentTestMode, NumberOfIterations = 3,
             Description ="Test the performance of a Select flow with InputBufferSize of 8, 1 Select and without GraphStages.Identity")]
         [TimingMeasurement]
         [ElapsedTimeAssertion(MaxTimeMilliseconds = 300)]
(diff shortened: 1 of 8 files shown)


Silv3rcircl3:
If I remember correctly the perf specs passed quite reliable in the passt, so maybe instead of disable them complete we could only increase the times so that we still detect real bad perf drops

Danthar:
@Silv3rcircl3 that is also an option. A lot more work though.
The false positives makes it really hard to evaluate PR's. Making the akka.streams perf test practically useless....

@Aaronontheweb in what timeframe do you think the relativistic / historical benchmarking  for NBench will be done?
I'd be fine with disabling the perf tests for now. But only if we know when we can expect to turn them back on again.

Aaronontheweb:
@Danthar haven't started on it; won't be able to until I get some full time resources onboarded early next year. Difficult to say how long it'd take as it's not a trivial effort. Just have a design right now.

Aaronontheweb:
@Silv3rcircl3 @Danthar I see both POVs here; the part that's a bitch is not knowing which specs have the issue (it's a bunch of different ones each time I've looked at it.)

One thing we could do that would solve this issue is to add TeamCity message formatting to NBench. That would allow us to pick up the worst offenders on the Flaky Tests Report. Does doing that step first sound more reasonable?

Aaronontheweb:
Honestly though, I don't think we can wait that long: look at how often it's failing, including on documentation PRs

image

Danthar:
I am for the TC formatting. It would at least give us a good idea of how many tests we are talking about. We can always opt to disable them after that anyways.

Aaronontheweb:
@Danthar ok, let me take a look today at what's involved in doing that and I'll get back to you guys with an update

Danthar:
Ok so now have updated NBench. The next step would be looking at the failing tests? If there are any anymore? Im not seeing alot of false positives anymore, so it seems i missed something ?

Aaronontheweb:
@Danthar weirdly enough, the intermittent failures stopped occurring with such frequency after upgrading. Might have been due to some of the other fixes included in NBench v0.3.3.

Going to be upgrading to NBench v0.3.4 shortly, which should help eliminate even more of these issues. Turns out that https://github.com/petabridge/NBench/issues/153 might have been the issue here.

PR #2401, opened 2016-12-05
[api-docs] Updated Sandcastle build/projects

sean-gilliam:
This PR updates the Sandcastle build file to include newer namespaces.
Also updated the various project files to output xml declaration file.


Aaronontheweb:
Crap, just realized SandCastle isn't installed on the new build server image.

PR #2405, opened 2016-12-08
[WIP] Akka.IO and Streams update

Horusiath:

This is work I took over from @willieferguson on the akka-io branch. This PR contains new design of akka.io - without previous abstractions in form of socket channels. It's working directly on `SocketAsyncEventArgs` not and should be compatible with .NET Standard 1.3+ (no BeginXXX and EndXXX methods anymore).

Once I'll fix them, I'll remove WIP marker. 
/cc #2153


Horusiath:
Looks like Mono build doesn't specify MONO compile flag - therefore compilation chain fails on `IPAddress.MapToIPv4` as it's not implemented on Mono.

alexvaluyskiy:
Just move this extension method to Akka.Core package
https://github.com/akkadotnet/akka.net/blob/103d5d0a43f2a53aaba606716c2d45d225c4e75e/src/core/Akka.Remote/IPExtensions.cs#L42

Horusiath:
@alexvaluyskiy I've moved it and since it's using reflection, I've added conditional flag to use it only for mono (#if MONO), and use normal .NET API on remaining platforms. But it looks like the MONO flag is not set when building on linux-mono agent.

alexvaluyskiy:
@Horusiath change this line https://github.com/akkadotnet/akka.net/blob/dev/build.fsx#L26 to this
`let configuration = if isMono then "Release" else "Release Mono"`

Horusiath:
Thanks (the condition should be reversed IMO ;) ).

alexvaluyskiy:
@Horusiath It is not necessary to use #MONO flag here. You can stay with extension method (with reflection inside). The extension method never execute if there is the original method with the same signature.

Horusiath:
@alexvaluyskiy 

> The extension method never execute if there is the original method with the same signature.

The current impl was result of the fact, that mono build was blowing up when I left this as an extension method before.

Aaronontheweb:
Looks like the Mono build is still having trouble on this PR.

```

[09:13:43]	./build.sh: line 13: wget: command not found
[09:13:43]	Cannot open assembly '/checkout/src/.nuget/nuget.exe': No such file or directory.
[09:14:31]	Running build failed.
[09:14:31]	Error:
[09:14:31]	Building /checkout/src/Akka.sln failed with exitcode 1.
[09:14:31]	Status:           Failure
[09:14:31]	  1) Building /checkout/src/Akka.sln failed with exitcode 1.
[09:14:31]	  2) CS1061: IO/IPExtensions.cs(28,65): Type `System.Net.IPAddress' does not contain a definition for `MapToIPv4' and no extension method `MapToIPv4' of type `System.Net.IPAddress' could be found. Are you missing an assembly reference?
[09:14:31]	  3) CS1061: IO/IPExtensions.cs(31,65): Type `System.Net.IPAddress' does not contain a definition for `MapToIPv6' and no extension method `MapToIPv6' of type `System.Net.IPAddress' could be found. Are you missing an assembly reference?
[09:14:31]	
[09:14:35]	Process exited with code 42
```

Horusiath:
I've created the initial [performance tests for Akka.IO](https://github.com/akkadotnet/akka.net/pull/2405/files#diff-6b21e5ea986d6de59b8803be1d3cc973) based on Helios horizontal scale stress test. However numbers from CI agent running them looks strange. I'll probably need some help here.

Another thing is problem on Mono - I need further investigation, but it looks like a problem with `IPAddress.MapToIPv4` and `IPAddress.MapToIPv6` which we brought as extension methods, because they are not supported on Mono.

Horusiath:
Missing parts, I'm working on:

- Failing linux/mono tests - I've fixed most of them locally (by default creating Socket on Mono assigns it an `AddressFamily.InterNetworkV6`... which is not supported by Mono itself).
- There are some other tests failing on linux/mono for reason I don't know.
- I've created a stress test similar in mind to Helios horizontal scale test... but it can boil down my comp. Not sure why yet.

alexvaluyskiy:
@Horusiath should this https://github.com/Horusiath/akka.net/blob/akka-io/src/core/Akka/IO/SimpleDnsManager.cs#L56 returns true?

Horusiath:
It looks like the current issues on linux/mono environment will block this PR, and we cannot do much about: https://bugzilla.xamarin.com/show_bug.cgi?id=53229

alexvaluyskiy:
Could we temporary skip that tests on Mono. And create a know issue.

Aaronontheweb:
I support @alexvaluyskiy's idea, although here's my question: how is what we're doing with Akka.IO different than say Helios or DotNetty, both of which work fine on Mono? (Disclaimer: have not reviewed code yet)

Aaronontheweb:
Re-running this because I want to see if some of the specs are racy on Mono. The `PeerClosed` vs. `ErrorClosed` issue definitely seems to be a Mono implementation issue, as you mentioned in Gitter @Horusiath. But the issues with not receiving messages on some of the specs (including a UDP spec) make me think there could be something else going on too.

src/core/Akka/Akka.csproj
```diff
@@ -253,19 +252,18 @@
...
     <Compile Include="Pattern\FutureTimeoutSupport.cs" />
     <Compile Include="Pattern\OpenCircuitException.cs" />
     <Compile Include="Routing\ConsistentHashRouter.cs" />
     <Compile Include="Serialization\JavaSerializer.cs" />
     <Compile Include="IO\ByteBuffer.cs" />
-    <Compile Include="IO\DirectByteBufferPool.cs" />
+    <Compile Include="IO\SocketEventArgsPool.cs" />
     <Compile Include="IO\Inet.cs" />
     <Compile Include="IO\IO.cs" />
-    <Compile Include="IO\SelectionHandler.cs" />
```
alexvaluyskiy: Why you have remove it?

Horusiath: Not me. Selection handlers were first thing, Willie removed while working on this branch. They were ported from the JDK NIO to provide an abstraction, that will make straightforward port of akka.io possible. But even on JVM people hate selection handlers design.

In .NET we don't have NIO, we use SocketAsyncEventArgs instead. This is one of the major changes on this branch.

alexvaluyskiy: You still have this file in the source code. What about `SocketChannel`?

Horusiath: Same thing. Probably those files reappeared at some point after rebasing from dev.

Horusiath:
I've refactored `ByteString` and its companion types. Basically what I did, was:

- replacement of existing `ByteBuffer` with `ArraySegment<byte>` (for closer interop with .NET)
- removed `ByteIterator` (I've used yield generator to create `IEnumerable<byte>` over byte string)
- removed `ILinearSeq<>` with its implementation (replaced with `IEnumerable<>` where it was used)
- totally redesigned `ByteString` - I've removed it's inheritors (`CompactByteString`, `ByteString1` and `ByteString1C`). 

Byte string still behaves as rope data structure - it essentially contains two fields: `Count` defining its total length and `ArraySegment<byte>[]`, so an array of byte buffers used by it. It also has 2 types of constructor methods instead of old `Create`:

- `CopyFrom` which copies provided buffers/byte arrays. It's equivalent of old Create.
- `FromBytes` which is an unsafe, non-allocating version - therefore is a lot faster and less GC-intensive, but once you pass an array to it, you're no longer allowed to mutate it.

After this change there are some tests failing on the Akka.Streams side, mostly around area of `Framing` class. I need to check what's wrong there and fix it.

Beside that I've discovered, that there were still unfinished parts from early days of this branch, that I need to put on TODO list:

- We used `SocketAsyncEventArgs` by pooling them. However if pool was empty, we were going into infinite loop in hope that other actor will release it at some point. This behavior was not visible in normal tests, but appeared in horizontal specs, when we allocated a lot of clients.
- The way we used buffer pools was quite strange. We had a single continuous byte array, that was sliced into equal segments (1 per each `SocketAsyncEventArgs` object from the pool). It was not configurable, each socket always took 128 bytes, and it couldn't be resized. I've implemented `BufferPool` which has more sane design (it can rent/release one or more `ArraySegment<byte>` so it plays nicely with new `ByteString` impl) and can be increased on demand in relatively performant manner.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2979,102 +2979,78 @@ namespace Akka.Event
...
+}
+namespace Akka.IO
+{
     public enum ByteOrder
     {
         BigEndian = 0,
         LittleEndian = 1,
     }
-    public abstract class ByteString : System.Collections.Generic.IEnumerable<byte>, System.Collections.Generic.IReadOnlyCollection<byte>, System.Collections.Generic.IReadOnlyList<byte>, System.Collections.IEnumerable, System.IEquatable<Akka.IO.ByteString>
+    public sealed class ByteString : System.Collections.Generic.IEnumerable<byte>, System.Collections.IEnumerable, System.IEquatable<Akka.IO.ByteString>
```
alexvaluyskiy: Should we keep this API as a public? I think it should be internal

src/core/Akka/Util/ByteHelpers.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Akka.IO;
+
+namespace Akka.Util
+{
+    public static class ByteHelpers
```
alexvaluyskiy: Why public?

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Horusiath: `ByteString` is core concept of akka.io. When we pass bytes to the user API, we pass them in chunks of byte strings.

alexvaluyskiy: Ok.
What about the rest of the new APIs? Like BufferPoolAllocationException, DefaultBufferPool, IBufferPool, ByteHelpers

Horusiath: `BufferPoolAllocationException` should stay public, the rest can be probably hidden.

Horusiath:
I've implemented `IsMatch(ByteString substring, int fromIndex)` method to optimize case when we check if substring starting from given index matches given one. This was pretty intensive case in `Framing` class, and it was using something like `Equals(x.Slice(possibleMatch, y.Count), y)` which was allocating way more objects than necessary. This version doesn't allocate any heap objects, it's just plain while loops.

Aaronontheweb:
cc @alexvaluyskiy doesn't this need to target the v1.3 branch going forward now?

alexvaluyskiy:
@Aaronontheweb yes, it should target v1.3

PR #2406, opened 2016-12-09
singleton handover message matching fix

zbynek001:
Singleton handover does not work because of matching messages against singleton instances instead of type


Aaronontheweb:
@zbynek001 nice catch. Changes look good to me. Re-running CI now.

Issue #2404, opened 2016-12-08
Optimize Config.GetNode


Horusiath:
As @rogeralsing noticed, there is a non-optimal code on actor creation path - it was probably hard to detect, as it's not directly related to actors. During creation every actor must first check if there is a coresponding entry in HOCON deployment config. Problem is that parsing it every time is pretty expensive. 

If I understand correctly problem lies [here](https://github.com/akkadotnet/akka.net/blob/d86eb6eb4fe512aac603cbc20bddc420acaec74a/src/core/Akka/Configuration/Config.cs#L99). When we want to get a hocon path based on a string, we are are parsing it. It's not a problem itself, but when we won't find anything, we are proceeding with a recursive call to fallback configs - there may be a lot of them, as basically every akka plugin adds its own fallback config to actor system. Thus we are again parsing path string on every call. 

What we could do instead, is to parse config path once at top level call, and use overloaded methods to use parsed path on nested recursive calls to config fallback.

rogeralsing:
I know we saw that the JVM impl do hit the config for each call, but doing any form of string manipulation per actor creation is horribly ineffective.
Creating actors should be cheap, this is both CPU intensive and causes a lot work for the GC.

Currently, Akka.NET is ~10 times slower than Akka in this area, According to the Skynet concurrency benchmark.

I think a lot could be improved here if there was a better story for the HOCON configs.
The design of the HOCON lib stems way way back, before there was any intention of building Akka.NET itself.
Not knowing all the oddities it would face later.
The parser itself is decent, but how the actual result is generated is messy.
It would have been nicer to actually be able to merge configs instead.

Rattenkrieg:
This issue seems perfect to begin with. Will be happy to grab it.

Aaronontheweb:
@Rattenkrieg go for it!

Rattenkrieg:
If I correctly understand @Horusiath point about "the parsing", the toughest part in the call to SplitDottedPathHonouringQuotes is the string splitting and linq on top of it which pressures cpu and gc.
So [I did obvious optimisation ](https://github.com/Rattenkrieg/akka.net/blob/dd8f6d1974582ad6292781b11b571c7d48bd7b2f/src/core/Akka/Configuration/Config.cs#L120) with storing results of "parsing" once in collection, and then calling overloaded method in fallback logic. Exactly as suggested.
[...]

PR #2410, opened 2016-12-13
GetNode do not lookups fallback config; config produced by WithFallbck contains all values from provided fallback

Rattenkrieg:
Fix for [2404](https://github.com/akkadotnet/akka.net/issues/2404)
Note that configs logging on start producing more verbose output as a side effect of changes in implementation;
Also note that we may no more needing `public Config Fallback { get; private set; }`
I didn't removed it since its public, however its only used in Config.cs;
There is also room for optimisation [see here](https://github.com/akkadotnet/akka.net/issues/2404#issuecomment-266321385)

### DIFF
--- a/src/core/Akka.Tests/Configuration/ConfigurationSpec.cs
+++ b/src/core/Akka.Tests/Configuration/ConfigurationSpec.cs
@@ -34,7 +34,7 @@ namespace Akka.Tests.Configuration
 
             // settings.ConfigVersion.ShouldBe(ActorSystem.Version);
             settings.Loggers.Count.ShouldBe(1);
-            settings.Loggers[0].ShouldBe(typeof(DefaultLogger).FullName);
+            settings.Loggers[0].ShouldBe(typeof (DefaultLogger).FullName);
             // settings.LoggingFilter.ShouldBe(typeof(DefaultLoggingFilter));
             settings.LoggersDispatcher.ShouldBe(Dispatchers.DefaultDispatcherId);
             settings.LoggerStartTimeout.Seconds.ShouldBe(5);
@@ -44,8 +44,8 @@ namespace Akka.Tests.Configuration
             settings.LogDeadLetters.ShouldBe(10);
             settings.LogDeadLettersDuringShutdown.ShouldBeTrue();
 
-            settings.ProviderClass.ShouldBe(typeof(LocalActorRefProvider).FullName);
-            settings.SupervisorStrategyClass.ShouldBe(typeof(DefaultSupervisorStrategy).FullName);
+            settings.ProviderClass.ShouldBe(typeof (LocalActorRefProvider).FullName);
+            settings.SupervisorStrategyClass.ShouldBe(typeof (DefaultSupervisorStrategy).FullName);
             settings.CreationTimeout.Seconds.ShouldBe(20);
             settings.AskTimeout.ShouldBe(Timeout.InfiniteTimeSpan);
             settings.SerializeAllMessages.ShouldBeFalse();
@@ -62,13 +62,13 @@ namespace Akka.Tests.Configuration
             settings.DebugUnhandledMessage.ShouldBeFalse();
             settings.DebugRouterMisconfiguration.ShouldBeFalse();
 
-            settings.SchedulerClass.ShouldBe(typeof(HashedWheelTimerScheduler).FullName);
+            settings.SchedulerClass.ShouldBe(typeof (HashedWheelTimerScheduler).FullName);
         }
 
         [Fact]
         public void Deserializes_hocon_configuration_from_net_config_file()
         {
-            var section = (AkkaConfigurationSection)ConfigurationManager.GetSection("akka");
+            var section = (AkkaConfigurationSection) ConfigurationManager.GetSection("akka");
             Assert.NotNull(section);
             Assert.False(string.IsNullOrEmpty(section.Hocon.Content));
             var akkaConfig = section.AkkaConfig;
@@ -82,7 +82,7 @@ namespace Akka.Tests.Configuration
             {
                 StringProperty = "aaa",
                 BoolProperty = true,
-                IntergerArray = new[]{1,2,3,4 }
+                IntergerArray = new[] {1, 2, 3, 4}
             };
 
             var config = ConfigurationFactory.FromObject(source);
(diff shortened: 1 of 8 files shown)


Aaronontheweb:
Need to go through our API approval process for this:  http://getakka.net/docs/akka-developers/public-api-changes

Rattenkrieg:
Decided to change signature of _MergeImmutable_ from public to internal. At first time have made it public just by analogy with _Merge_. See no point of being it public.
Failure of MultiNodeTestRunner looks related to test agent.

Danthar:
Looks like ClusterSingletonManagerLeaveSpecs is failing. Although the gigantic wall of text is hampering my understand as to **why**

Rattenkrieg:
Wow how that appeared successfull on second run? BTW tests on my machine are also ok.

Aaronontheweb:
cc @akkadotnet/core need to do a code review on this. Making a note.

src/core/Akka.Tests/Configuration/ConfigurationSpec.cs
```diff
@@ -136,6 +136,35 @@ public void Can_merge_objects()
 
         }
 
+        [Fact(DisplayName = @"Scalar value should not be overriden by an object during merge")]
+        public void Scalar_value_should_not_be_overriden_by_an_object_during_merge()
```
Aaronontheweb: Good

src/core/Akka.Tests/Configuration/HoconTests.cs
```diff
@@ -546,6 +546,55 @@ public void Can_use_fluent_multi_level_fallback()
             config.GetInt("foo.bar.borkbork").ShouldBe(-1);
         }
 
+        [Fact(DisplayName = @"Scalar value should not be overriden by an object in fallback and vice versa")]
+        public void Scalar_value_should_not_be_overriden_by_an_object_in_fallback_and_vice_versa() {
```
Aaronontheweb: Yep, all specs look good here

src/core/Akka/Configuration/Hocon/HoconObject.cs
```diff
@@ -161,6 +161,7 @@ public string ToString(int indent)
             var sb = new StringBuilder();
             foreach (var kvp in Items)
             {
+                if (kvp.Value.AdoptedFromFallback) continue;
```
Aaronontheweb: Got it, your intellisense comment was very helpful. So we use this flag to skip objects that were being merged in from a fallback?

src/core/Akka/Configuration/Hocon/HoconValue.cs
```diff
@@ -29,6 +29,17 @@ public HoconValue()
...
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="HoconValue"/> class.
+        /// </summary>
+        /// <param name="values">The list of elements inside this HOCON value.</param>
+        /// <param name="adoptedFromFallback">Indicates whether this instance was constructed during association with fallback <see cref="Config"/>.</param>
+        public HoconValue(List<IHoconElement> values, bool adoptedFromFallback = true)
+        {
+            Values = values;
+            AdoptedFromFallback = adoptedFromFallback;
```
Aaronontheweb: Is this the only place where `AdoptedFromFallback` can be set?

src/core/Akka/Configuration/Config.cs
```diff
@@ -456,16 +450,20 @@ public virtual Config WithFallback(Config fallback)
             if (fallback == this)
                 throw new ArgumentException("Config can not have itself as fallback", nameof(fallback));
 
-            Config clone = Copy();
-
-            Config current = clone;
+            var mergedRoot = fallback == null
```
Aaronontheweb: @Rattenkrieg would you mind explaining the algorithm here a bit?

src/core/Akka/Configuration/Hocon/HoconObject.cs
Rattenkrieg: Exactly, either way logs get bloated, even some tests become failing; First I tried to use subclass of HoconValue as a marker, but this somehow breaks tests e.g LoadPluginSpec. Couple of hours of debugging yield no result for me, I suspect the system can't instantiate correct dispatcher for specs. But how inheritance affect this remains mystery for me.

src/core/Akka/Configuration/Hocon/HoconValue.cs
Rattenkrieg: Thats right and this ctor is only called from _MergeImmutable_

src/core/Akka/Configuration/Config.cs
Rattenkrieg: This code mainly leverages _MergeImmutable_ which follows _Merge_ in its structure with nuance of producing new _HoconValue_ every time it found same key in rhs _HoconObject_. Also there is a little boilerplate with injecting result in final _Config_ instance. I'm keep preserving chain of fallbacks since it is property with public getter. Feel free to ask for more precise explanation if something left unclear.
Also I've missed obvious optimisation as to shortcut entire routine if _fallback_ argument is null. Will fix it now.

Rattenkrieg:
Thanks for review @Aaronontheweb 
Please note my very last statement in response to your comments for Config.cs. Will it be convenient to provide fix in separate commit or to squash alltogether?
Also I'm about to include tests from #2418 (props to @kerch-ai)

kerch-ai:
credit goes to @eisendle for the unit test. Are there pre-existing tests for the deployer?

Rattenkrieg:
Included test for #2418 
Did minor refactoring as stated in answer to @Aaronontheweb comment to Config.cs, squashed everything.

@kerch-ai you can find a couple in [Akka.Tests.Actor.DeployerSpec](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Tests/Actor/DeployerSpec.cs) or just type "Deployer" in solution explorer.

Aaronontheweb:
Need to re-approve API changes

Rattenkrieg:
Omg how thah happen, that this test not running in "build RunAllTests" target?
And is changing signature of protected method is really an API change?

Aaronontheweb:
@Rattenkrieg on the former question, not sure :p 

on the latter question, if the class is public then yes. Affects how user-defined sub classes are implemented.

PR #2417, opened 2016-12-16
upgraded to NBench v0.3.4 to help prevent false negative performance assertions

Aaronontheweb:


Aaronontheweb:
Replaces #2398

Aaronontheweb:
Going to run the benchmarks multiple times to verify

PR #2418, opened 2016-12-18
Potential fix for Router cannot be created from fallback configuration

### LINKED ISSUE #2348: Router cannot be created from fallback configuration

kerch-ai:
I believe the issue is in the Deployer, it does not seem to look at the fallback configuration, if there is a value in the main configuration for deployment. Added config and fallback before calling SetDeploy

fix #2348

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorSystemTests.cs
+++ b/src/core/Akka.Tests/Actor/ActorSystemTests.cs
@@ -8,6 +8,8 @@
 using System.Linq;
 using Akka.Actor;
 using Akka.Actor.Internal;
+using Akka.Configuration;
+using Akka.Routing;
 using Xunit;
 
 namespace Akka.Tests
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
Does this conflict with #2410 at all?

kerch-ai:
I do not believe so, but i will double check

kerch-ai:
You are right with the fix for  #2410 this pull request is redundant

Issue #1448, opened 2015-11-19
Warning: TempContainer trying to remove non-child actor...


Aaronontheweb:
I see this continuously inside the logging center for the multinode testrunner and I think it's related to `FutureActorRef` and `Ask`, but either way it's disconcerting and I want to understand the circumstances under which this warning is raised. Might not be a big deal, but we should investigate.

```
[WARNING][11/19/2015 11:27:19 PM][Thread 0005][LocalActorRefProvider(akka://TestRunnerLogging)] akka://TestRunnerLogging/temp trying to remove non-child pb
```

rogeralsing:
There are two scenarios where this could potentially happen.

1. Some code tries to unregister a non existing temp actorref. e.g. passing in the incorrect name
2. Some code for some reasons tries to remove the same child twice (or more than one time)

My guess would be that there is some code in the `Ask` mechanics that might do the latter, e.g. unregister temp in some combination of complete and failure of tasks.

rogeralsing:
The `VirtualPathContainer` which logs this is also used for hosting the remote deployed actors in Akka.Remote. 
`class RemoteSystemDaemon : VirtualPathContainer`

Aaronontheweb:
> Some code for some reasons tries to remove the same child twice (or more than one time)

I suspect that's what is happening

Aaronontheweb:
Closed in #2422

PR #2399, opened 2016-12-04
Revert "Unregister temp actor if ask operation got canceled."

Aaronontheweb:
Reverts akkadotnet/akka.net#2381

Have to revert this. Causes enormous, like hundreds of MBs worth of "unregister temp actor spam" in all of our logs. Can't ship it like this. Need to address that logging issue first.

PR #2422, opened 2016-12-21
Make FutureActorRef call _unregister() in a task continuation.

ZigMeowNyan:
This takes eisendle's idea of dealing with the temp actor unregistration issue on cancelled ask requests in #2380 and cherry-picks his test spec commit from PR #2381 , but handles it by calling the `unregister` action in a continuation inside of `FutureActorRef`. This should make the unregistration more consistent with `FutureActorRef` - ensuring that we clean up our resources once regardless of how the task turns out. I'm assuming there might be some back and forth in case it hits the same issue that caused the revert in #2399. Following up with this PR as a lot of people asked about #2381 when rebasing code in #2349.

It also adds two specs to make sure we clean up if we're basing the ask requests around cancellation requests or both timeouts and cancellation requests instead of just timeouts. It makes me nervous to just assume that it'll work the same in all three instances forever, so there's some extra tests.

### DIFF
--- a/src/core/Akka.Tests/Actor/AskSpec.cs
+++ b/src/core/Akka.Tests/Actor/AskSpec.cs
@@ -94,6 +94,33 @@ namespace Akka.Tests.Actor
             Assert.Throws<AggregateException>(() => { actor.Ask<string>("timeout", Timeout.InfiniteTimeSpan, cts.Token).Wait(); });
             Assert.True(cts.IsCancellationRequested);
         }
+        [Fact]
+        public void Cancelled_ask_with_null_timeout_should_remove_temp_actor()
+        {
+            var actor = Sys.ActorOf<SomeActor>();
+            var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(100));
+            Assert.Throws<AggregateException>(() => { actor.Ask<string>("cancel", cts.Token).Wait(); });
+            Assert.True(cts.IsCancellationRequested);
+            Are_Temp_Actors_Removed(actor);
+        }
+        [Fact]
+        public void Cancelled_ask_with_timeout_should_remove_temp_actor()
+        {
+            var actor = Sys.ActorOf<SomeActor>();
+            var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(100));
+            Assert.Throws<AggregateException>(() => { actor.Ask<string>("cancel", TimeSpan.FromSeconds(30), cts.Token).Wait(); });
+            Assert.True(cts.IsCancellationRequested);
+            Are_Temp_Actors_Removed(actor);
+        }
+        private void Are_Temp_Actors_Removed(IActorRef actor)
+        {
+            var actorCell = actor as ActorRefWithCell;
+            Assert.True(actorCell != null, "Test method only valid with ActorRefWithCell actors.");
+            var container = actorCell.Provider.TempContainer as VirtualPathContainer;
+            int childCounter = 0;
+            container.ForEachChild(x => childCounter++);
+            Assert.True(childCounter == 0, "Temp actors not all removed.");
+        }
 
         /// <summary>
         /// Tests to ensure that if we wait on the result of an Ask inside an actor's receive loop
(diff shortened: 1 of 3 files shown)


ZigMeowNyan:
> 	**FutreActorRef**: Call _unregister ...

I am apparently an speller.

Danthar:
nice work @ZigMeowNyan

ZigMeowNyan:
Thanks. Still waiting on the full CI test results, but it failed one of the integration tests in `Akka.Streams.Tests.Dsl.UnfoldResourceAsyncSourceSpec` : `A_UnfoldResourceAsyncSource_must_use_dedicated_blocking_io_dispatcher_by_default`. Dropping into the build log, it has the following details:

```
[Akka.Streams.Tests.Dsl.UnfoldResourceAsyncSourceSpec.A_UnfoldResourceAsyncSource_must_use_dedicated_blocking_io_dispatcher_by_default] System.IO.IOException : The process cannot access the file 'D:\temp\buildTmp\blocking-source-spec-4.tmp' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileInfo.Delete()
   at Akka.TestKit.Xunit2.TestKit.Dispose(Boolean disposing) in D:\work\d26c9d7f36545acd\src\contrib\testkits\Akka.TestKit.Xunit2\TestKit.cs:line 129
   at Akka.TestKit.Xunit2.TestKit.Dispose() in D:\work\d26c9d7f36545acd\src\contrib\testkits\Akka.TestKit.Xunit2\TestKit.cs:line 95
   at ReflectionAbstractionExtensions.DisposeTestClass(ITest test, Object testClass, IMessageBus messageBus, ExecutionTimer timer, CancellationTokenSource cancellationTokenSource) in C:\BuildAgent\work\cb37e9acf085d108\src\xunit.execution\Extensions\ReflectionAbstractionExtensions.cs:line 76
```

The weird bit for me is it had passed that test locally before I even made the PR. I ran it again, and it failed 3 times and succeeded the 4th. Is anyone else getting inconsistent behavior from this particular test? If that exception is actually supposed to be how the test fails, I can go back and make the temp actor unregistration continuation synchronous, but I don't plan on that unless it's required and relevant.

Danthar:
@ZigMeowNyan yes its an known racy/buggy test. I think i saw @Aaronontheweb create an issue for it, but cant find it atm

Aaronontheweb:
what @Danthar says is correct - we have a race condition when we clean up the file used for that spec. Documented the issue somewhere :p

ZigMeowNyan:
Understood. I'll leave this PR for review, then.

src/core/Akka.Tests/Actor/AskSpec.cs
```diff
@@ -94,6 +94,33 @@ public void Can_cancel_when_asking_actor()
             Assert.Throws<AggregateException>(() => { actor.Ask<string>("timeout", Timeout.InfiniteTimeSpan, cts.Token).Wait(); });
             Assert.True(cts.IsCancellationRequested);
         }
+        [Fact]
```
Aaronontheweb: The specs look great. Covers exactly what I wanted to know.

src/core/Akka.Tests/Actor/AskTimeoutSpec.cs
```diff
@@ -50,5 +50,25 @@ public AskTimeoutSpec()
             }
         }
 
+        [Fact]
+        public async Task TimedOut_ask_should_remove_temp_actor()
```
Aaronontheweb: Also looks good

src/core/Akka/Actor/ActorRef.cs
```diff
@@ -66,6 +66,7 @@ public FutureActorRef(TaskCompletionSource<object> result, Action unregister, Ac
             _result = result;
             _unregister = unregister;
             _path = path;
+            _result.Task.ContinueWith(_ => _unregister());
```
Aaronontheweb: These changes also appear to have fixed https://github.com/akkadotnet/akka.net/issues/1448 as well. Could have also been the previous PR, but I don't see that message appear in the logs for NBench, unit tests, or multi-node tests (spot checked a bunch of them.)

PR #2424, opened 2016-12-22
spec for verifying that DistributedPubSub mediator can be spawned from within child actor in testkit

Aaronontheweb:
http://stackoverflow.com/questions/41249465/how-to-test-distributedpubsub-with-the-testkit-in-akka-net - just wanted to verify if this bug exists or not. My guess is that the end user needed to specify the `ClusterActorRefProvider` in their HOCON configuration


Danthar:
lgtm

PR #2430, opened 2016-12-27
Fixed RandomLogic and RandomPool to work correctly.

gordey4doronin:
I was playing around with `RandomPool` and it turned out it works as thought it is `round robin`.
I checked the code and found the root of the problem :)
Looks like a simple copy-paste typo.

Here is a list of fixes and additional improvements:

 - RandomLogic `Select Routee` upper bound fixed.
 - RandomPool `CreateRouter` fixed to use proper random logic.
 - RandomSpec improved to have additional check for stochasticity (some of all random values definitely should be different from `iterationCount`).
 - RandomSpec additional tests added for `RandomGroup` and `RandomLogic`.
 - RoundRobinSpec fixed to instantiate with proper round robin pools.

### DIFF
--- a/src/core/Akka.Tests/Routing/RandomSpec.cs
+++ b/src/core/Akka.Tests/Routing/RandomSpec.cs
@@ -53,12 +53,12 @@ namespace Akka.Tests.Routing
             }
         }
 
-        private class BroadcastActor : ReceiveActor
+        private class RandomBroadcastActor : ReceiveActor
         {
             private readonly TestLatch _helloLatch;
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Looks like CI didn't pick this up due to the old age of the commit (19 days).

Please try this: http://stackoverflow.com/a/31540373/377476

gordey4doronin:
Ah, I see..
I had many commits then squashed them into one, and git picked up the first date in a row.
I've fixed and force pushed, let's see what CI will say.

Issue #2426, opened 2016-12-22
Actor with Resume strategy does not clear child actor's CurrentMessage


ali-ince:
I have discovered a strange behavior when using an actor with a kind of pool router configured with a OneForOneStrategy returning Directive.Resume.

When one of the child actors of the router fails due to an exception on one of its receive handlers, router's strategy returns resume directive which seems to work as expected. However, when I inspect that resumed child actor I see that its Cell.CurrentMessage property is not cleared up - which is not the case when using standalone actors.

Here is a small console application code to demonstrate the issue. Child actors' CurrentMessage properties are dumped in the console.

```csharp
    class Program
    {

        public class TestActor : ReceiveActor
        {

            public TestActor()
            {
                Receive<string>(m => HandleError(), m => m.Equals("error"));
                Receive<string>(m => HandleNormal(), m => m.Equals("normal"));
            }

            private void HandleError()
            {
                throw new Exception("Test Exception");
            }
            
            private void HandleNormal()
            {
                Thread.Sleep(500);
            } 

        }


        static void Main(string[] args)
        {
            ActorSystem sys = ActorSystem.Create("TestSystem");

            OneForOneStrategy commonOneForOneStrategy = new OneForOneStrategy(exc =>
            {
                return Directive.Resume;
            });

            RouterConfig defaultPoolConfig =
                new RoundRobinPool(5)
                    .WithSupervisorStrategy(commonOneForOneStrategy);

            Props tstActorProps = Props.Create<TestActor>(() => new TestActor());

            IActorRef tstActor = sys.ActorOf(tstActorProps.WithRouter(defaultPoolConfig), "tstActor");

            tstActor.Tell("error");
            tstActor.Tell("normal");
            tstActor.Tell("normal");
            tstActor.Tell("normal");
            tstActor.Tell("normal");

            IEnumerable<IActorRef> children = ((dynamic) tstActor).Children;
            foreach (IActorRef actorRef in children)
            {
                LocalActorRef localActorRef = actorRef as LocalActorRef;

                if (localActorRef != null)
                {
                    Console.WriteLine("{0} CurrentMessage: {1}", localActorRef.Path.Uid, localActorRef.Cell.CurrentMessage ?? "null");
                }
            }
        }
    }
```

When we have huge messages, those messages are pinned to memory and GC does not collect them since they're referenced by the CurrentMessage property - increasing process memory. Although this happens until a new message hits that specific child actor, that may take time depending on message throughput.

[...]

PR #2433, opened 2017-01-02
 Actor with Resume strategy does not clear child actor's CurrentMessage #2426 

ali-ince:
When a child actor fails and the parent takes the resume strategy, it seems that the failed child actor's CurrentMessage property is not cleared out.

After discussing with @Rattenkrieg, I have written some test cases using different actor hierarchies that showcases the issue.


Aaronontheweb:
@ali-ince interesting - the group router specs fail as well. Aren't those actors just regular child actors?

src/core/Akka.Tests/Actor/FailingActorSpec.cs
```diff
@@ -0,0 +1,490 @@
...
+                Receive<string>(m => child.Tell(m), m => true);
+
+                resetEvt.Set();
+            }
+
+            private void CreateGroupedChildRandom()
+            {
+                IActorRef realChild = Context.ActorOf(Props.Create<FailingActor>(), "failing");
+
+                child = Context.ActorOf(Props.Empty.WithRouter(new BroadcastGroup(new string[] { realChild.Path.ToString() })).WithSupervisorStrategy(strategy));
```
Aaronontheweb: The `SupervisionStrategy` isn't being used on any of these group router tests at the moment, because group routers aren't parents. They're just actors that use actor selections to send messages to other actors created by someone else. You'll need to put a parent actor with the specified supervision strategy above the child. Looks like this is true for all of the specs.

ali-ince: Sorry if I got it wrong, but I think that ParentActor is already doing what you do point in its SupervisorStrategy() at this [method](https://github.com/akkadotnet/akka.net/pull/2433/files/abdc679a0f6e4b41f75363a85bff9a5ad305fab7#diff-1c1211befa61f9dbf239d8c8b525ba74R217). I see that the  WithSupervisorStrategy method does not have any effect on group routers, but ParentActor handles it in this case. Am I missing something?

I am making some changes in order to check whether Decider is hit or not for all of the test cases, will commit it along with your requested changes - based on your response to this comment.

ali-ince:
Upon @Aaronontheweb's requested changes;

1. ResumeDecider class is introduced to count how many times its Decide method is called.
2. Updated all test specs to include a check whether ResumeDecider is called or not.
3. All group router initialization code is cleaned from unneeded WithSupervisorStrategy calls.

Danthar:
Besides needing a rebase. @Aaronontheweb can you take at @ali-ince latest changes ?

Aaronontheweb:
@Danthar I'll try to take a look this evening

PR #2421, opened 2016-12-21
swapped out Wire references for Hyperion package

Aaronontheweb:
This PR swaps out the underlying Wire dependencies for Hyperion, in order to stop our users from getting GPL'd by upgrading to a new version of Wire.

I propose doing this across two releases

1. Release the current Akka.Serailization.Wire, complete with Hyperion dependencies, to move net new users onto the latest Hyperion version.
2. Deprecate the Akka.Serialization.Wire package and replace it with a Akka.Serialization.Hyperion package and update the necessary documentation in a subsequent release.

Any thoughts on the above?

PR #2441, opened 2017-01-08
Continued work over Distributed Data

Horusiath:
This PR brings few bug fixes, more API methods for distributed data collections and optimized, Hyperion-based serializers for ddata module (original Akka uses Protobuf).

### DIFF
--- a/src/core/Akka/Util/Internal/StringBuilderExtensions.cs
+++ b/src/core/Akka/Util/Internal/StringBuilderExtensions.cs
@@ -63,7 +63,8 @@ namespace Akka.Util.Internal
                     index++;
                     sb.Append(separator);
                     // ReSharper disable CompareNonConstrainedGenericWithNull
-                    if (enumerator.Current != null)
+                    current = enumerator.Current;
+                    if (current != null)
                         // ReSharper restore CompareNonConstrainedGenericWithNull
                     {
                         valueAppender(sb, current, index);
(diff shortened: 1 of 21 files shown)


Aaronontheweb:
Does this need to be dove-tailed with https://github.com/akkadotnet/akka.net/pull/2421 at all?

Horusiath:
No, those are custom serializers, but they use Hyperion internally. On the other side I'm thinking, if we should fall back to proto-buf here: current version assumes, that all elements for serialized collections will use Hyperion as well, which doesn't have to be true. We might potentially run into situation where we want to use custom serializers for internal elements representation.

Horusiath:
Ultimatelly using protobuf here is not a good option - the major point is the fact, that we have heavily generic collections here, causing a lot of pain when working with protobuf. Therefore for the time being I'm proposing to use existing serializer (a slightly optimized one Hyperion-based).

This PR is ready for review and to merge.

src/core/Akka/Util/Internal/StringBuilderExtensions.cs
```diff
@@ -63,7 +63,8 @@ public static StringBuilder AppendJoin<T>(this StringBuilder sb, string separato
                     index++;
                     sb.Append(separator);
                     // ReSharper disable CompareNonConstrainedGenericWithNull
-                    if (enumerator.Current != null)
+                    current = enumerator.Current;
+                    if (current != null)
```
Aaronontheweb: Nice catch

Aaronontheweb:
@Horusiath looks good to me

PR #2444, opened 2017-01-11
Initial replacement of Helios with DotNetty

Horusiath:
This PR contains changes necessary to replace existing Helios transport with DotNetty.


src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/ClusterClient/ClusterClientSpec.cs
```diff
@@ -611,7 +611,7 @@ public void ClusterClient_must_reestablish_connection_to_receptionist_after_serv
                     // start new system on same port
                     var sys2 = ActorSystem.Create(
                         Sys.Name,
-                        ConfigurationFactory.ParseString("akka.remote.helios.tcp.port=" + Cluster.Get(Sys).SelfAddress.Port).WithFallback(Sys.Settings.Config));
+                        ConfigurationFactory.ParseString("akka.remote.dot-netty.tcp.port=" + Cluster.Get(Sys).SelfAddress.Port).WithFallback(Sys.Settings.Config));
```
alexvaluyskiy: Maybe we should remove the dash and use `dotnetty` ?

alexvaluyskiy:
Will you plan to support SSL and/or UDP in the initial version?

src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/ClusterClient/ClusterClientSpec.cs
Horusiath: Original name is DotNetty (which is PascalCase) therefore I've proposed dot-netty, as name convention for HOCON usually conforms to DashCase.

Horusiath:
SSL will be supported. UDP probably won't - UDP is very useful when designing higher level protocol on top of it as it allows us to go faster than TCP (see  [QUIC](https://en.wikipedia.org/wiki/QUIC) or [Aeron](https://github.com/real-logic/Aeron/wiki/Protocol-Specification)), but unless we build one by ourselves, UDP itself probably won't be useful enough for Akka.

Aaronontheweb:
@Horusiath we should add UDP support anyway. Let the user make that decision. I get requests for it on occasion the time from IOT users.

Aaronontheweb:
Although, doesn't need to be in THIS pull request. Can add it later.

Horusiath:
I've added test certificate for SSL tests to .gitignore, however it seems that git ignores it anyway.

Horusiath:
At this point there are still few remaining issues:

- SSL support is working, but in order to be able to test it I have to add a test certificate to certificate store. I'm adding/removing cert before/after the test itself. This allows us to correctly work with those tests in R# test runner. However this also requires to run VS with admin permissions. Same will apply to running build script on CI agent.
- DNS resolution specs inherited from Helios are failing for unknown reason. They are FsCheck-based and can fail i.e. after 78 valid assertions. However when I'm trying to use the input data that caused an error to run this as a straightforward Xunit fact, everything works.
- The last truly missing piece is compatibility with Helios. Right now when DotNetty transport won't detect any `akka.remote.dot-netty.tcp` path, it can fall back to `akka.remote.helios.tcp`. This gives us backward compatibility for configuration. However actor systems using DotNetty and Helios at this moment still are unable to talk with each other - I'm guessing that the reason for that is the endianness of both (DotNetty &rarr; big endian, Helios &rarr; little endian), that needs to be adjusted. However the problem will be with further verification as none of the transport layers gives any useful info about exceptions and I haven't found a good hook point to insert breakpoint before that error happens.

Horusiath:
At this point all tests are passing locally! Now we need to figure out how to enable secure socket layer tests working in CI agent in secure manner. I guess running pull request tests on agent with admin permissions is not the smartest move possible :P

alexvaluyskiy:
Have you compared the performance between Helios and DotNetty?

alexvaluyskiy:
> DNS resolution specs inherited from Helios are failing for unknown reason.
There is a difference between Helios and DotNetty as I remember. DotNetty can't resolve 'localhost' @Aaronontheweb Am I Right?

alexvaluyskiy:
> The last truly missing piece is compatibility with Helios. 

Do we need this compatibility? The best idea - to release a Helios transport as a separate package

Horusiath:
@alexvaluyskiy I haven't made perf test yet. First I want CI build passing. 

Resolving DNS is not a problem, as this is already implemented, in similar but not exactly the same way it works in Helios. And compatibility is desired, when people are upgrading their nuget packages, they expect to end up with working system. Helios will be released as a separate package (legacy transport layer). 

I've managed to make DotNetty working with Helios already - major change was to switch default byte order in DotNetty to little endian (I've made it configurable).

Aaronontheweb:
The backwards compatibility test still fails - is this more DNS stuff?

Horusiath:
Nope, I've accidentally removed helios dependency from akka.remote.tests when fixing solution file.

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
```diff
@@ -0,0 +1,479 @@
...
+
+            return server;
+        }
+
+        private async Task<IPEndPoint> ResolveNameAsync(DnsEndPoint address)
+        {
+            var resolved = await Dns.GetHostEntryAsync(address.Host);
+            //NOTE: for some reason while Helios takes first element from resolved address list
+            // on the DotNetty side we need to take the last one in order to be compatible
+            return new IPEndPoint(resolved.AddressList[resolved.AddressList.Length - 1], address.Port);
```
Aaronontheweb: I don't understand this, but moreover we should probably take the first address *that matches the preferred DNS family settings*. It's probably a bug that we don't do this today for Helios in the first place.

Aaronontheweb: I see that the overload below does this - does that get called by default?

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
```diff
@@ -0,0 +1,479 @@
...
+                // DNS resolution will be performed by the transport
+                listenAddress = new DnsEndPoint(address.Host, (int)address.Port);
+            }
+            return listenAddress;
+        }
+
+        #endregion
+    }
+
+    internal class HeliosBackwardsCompatabilityLengthFramePrepender : LengthFieldPrepender
```
Aaronontheweb: Is this ported over from the Helios 2.1 bits?

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransportSettings.cs
```diff
@@ -0,0 +1,185 @@
...
+                serverSocketWorkerPoolSize: ComputeWorkerPoolSize(config.GetConfig("server-socket-worker-pool")),
+                clientSocketWorkerPoolSize: ComputeWorkerPoolSize(config.GetConfig("client-socket-worker-pool")),
+                maxFrameSize: ToNullableInt(config.GetByteSize("maximum-frame-size")) ?? 128000,
+                ssl: config.HasPath("ssl") ? SslSettings.Create(config.GetConfig("ssl")) : SslSettings.Empty,
+                dnsUseIpv6: config.GetBoolean("dns-use-ipv6", false),
+                tcpReuseAddr: config.GetBoolean("tcp-reuse-addr", true),
+                tcpKeepAlive: config.GetBoolean("tcp-keepalive", true),
+                tcpNoDelay: config.GetBoolean("tcp-nodelay", true),
+                backlog: config.GetInt("backlog", 4096),
+                enforceIpFamily: RuntimeDetector.IsMono || config.GetBoolean("enforce-ip-family", false),
```
Aaronontheweb: I haven't checked yet, but do we know if the newer (4.6+) versions of Mono function correctly with IPV6 now?

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransportSettings.cs
```diff
@@ -0,0 +1,185 @@
...
+        public readonly bool TcpKeepAlive;
+        public readonly bool TcpNoDelay;
+        public readonly bool EnforceIpFamily;
+        public readonly int Backlog;
+        public readonly int? ReceiveBufferSize;
+        public readonly int? SendBufferSize;
+        public readonly int? WriteBufferHighWaterMark;
+        public readonly int? WriteBufferLowWaterMark;
+        public readonly bool BackwardsCompatibilityModeEnabled;
+        public readonly bool LogTransport;
```
Aaronontheweb: This is for activating that Logging `ChannelHandlerAdapter` from earlier, right?

src/core/Akka.Remote/Configuration/Remote.conf
```diff
@@ -310,14 +310,17 @@ akka {
...
     }
 
-    ### Default configuration for the Helios based transport drivers
+	# necessary to keep backwards compatibility
+	helios.tcp.transport-class = "Akka.Remote.Transport.Helios.HeliosTcpTransport, Akka.Remote.Transport.Helios"
 
-    helios.tcp {
+    ### Default configuration for the DotNetty based transport drivers
+
+    dot-netty.tcp {
```
Aaronontheweb: Is there a new setting in here for turning on transport logging?

Horusiath: Yes. I've added `log-transport` value in order to add `LoggingHandler` to DotNetty pipeline. It's used to log user events on DotNetty side.

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
Horusiath: It depends on `enforce-ip-family` setting. It IP address family is on, second option will be used. I also think, we should pick first correct address, this however was causing a failure between Helios and DotNetty backward compatibility spec - I don't know yet why.

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
Horusiath: Yes. I've taken it from there + added ByteOrder.LittleEndian (because of DotNetty :) )

Aaronontheweb:
@Horusiath looks like we should be able to disable cert validation on the TLS handler, which will allow the TLS specs to pass without needing to install a certificate on the build machine, which is good because I have no idea how trust stores work in environments like Docker.

Here's an example of the DotNetty settings where they allow cert validation to be turned off in the Azure IoT hub:
https://github.com/Azure/azure-iot-sdk-csharp/blob/master/device/Microsoft.Azure.Devices.Client/Transport/Mqtt/MqttTransportHandler.cs#L771

Another example of how to disable this from the DotNetty integration test suite: https://github.com/Azure/DotNetty/blob/dev/test/DotNetty.Tests.End2End/End2EndTests.cs#L74

I recommend that we add a new HOCON configuration setting for the .ssl section that disables cert validation, just like the Azure IoT hub is doing. We can run the unit tests with that setting enabled. Does that sound reasonable to you?

Horusiath:
Looks like everything is fine, except that SSL spec is failing on mono.

src/core/Akka.Remote.Tests/Transport/DotNettySslSupportSpec.cs
```diff
@@ -80,6 +81,9 @@ public DotNettySslSupportSpec(ITestOutputHelper output) : base(TestConfig(ValidC
         [Fact]
         public void Secure_transport_should_be_possible_between_systems_sharing_the_same_certificate()
         {
+            // skip this test due to linux/mono certificate issues
+            if (IsMono) return;
```
Aaronontheweb: Looks good

Horusiath:
@Aaronontheweb not sure - but maybe an issue is in endianness? By default DotNetty uses big endian. To make it little-endian compatible (like Helios), I'm using `WithOrder(ByteOrder)` everytime a `IByteBuffer` is being read/written - in most cases this means, that DotNetty will wrap byte buffer with `SwappedByteBuffer` object. However I wouldn't expect those object to being promoted to Gen1.

Aaronontheweb:
Apologies for the table formatting. Logged an issue in NBench for that.

## Helios 2.1 Implementation
# Akka.Remote.Tests.Performance.Transports.HeliosRemoteMessagingThroughputSpec+OneWay
__Measures the throughput of Akka.Remote over a particular transport using one-way messaging__
_2/2/2017 3:49:12 AM_

### Totals
TotalCollections [Gen0] |     collections |          213.00 |          208.67 |          205.00 |            4.04 |
TotalCollections [Gen1] |     collections |            6.00 |            5.67 |            5.00 |            0.58 |
TotalCollections [Gen2] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
[Counter] RemoteMessageReceived |      operations |       50,000.00 |       50,000.00 |       50,000.00 |            0.00 |

### Per-second Totals
TotalCollections [Gen0] |     collections |           80.09 |           77.56 |           74.87 |            2.61 |
TotalCollections [Gen1] |     collections |            2.34 |            2.11 |            1.87 |            0.24 |
TotalCollections [Gen2] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
[Counter] RemoteMessageReceived |      operations |       19,534.40 |       18,597.15 |       17,575.10 |          982.40 |

## DotNetty Implementation

# Akka.Remote.Tests.Performance.Transports.HeliosRemoteMessagingThroughputSpec+OneWay
__Measures the throughput of Akka.Remote over a particular transport using one-way messaging__

### Totals

TotalCollections [Gen0] |     collections |          210.00 |          203.67 |          198.00 |            6.03 |
TotalCollections [Gen1] |     collections |           25.00 |           21.67 |           19.00 |            3.06 |
TotalCollections [Gen2] |     collections |            1.00 |            0.67 |            0.00 |            0.58 |
[Counter] RemoteMessageReceived |      operations |       50,000.00 |       50,000.00 |       50,000.00 |            0.00 |

### Per-second Totals
TotalCollections [Gen0] |     collections |           63.64 |           62.58 |           61.37 |            1.15 |
TotalCollections [Gen1] |     collections |            7.47 |            6.65 |            5.74 |            0.87 |
TotalCollections [Gen2] |     collections |            0.32 |            0.21 |            0.00 |            0.18 |
[Counter] RemoteMessageReceived |      operations |       16,071.62 |       15,374.37 |       14,936.59 |          610.38 |

Aaronontheweb:
Looking at a measured performance difference of about ~3k msg / s but I've seen that vary quite a bit on some of the benchmarks for this spec; been a difference sometimes as large as 6k.

The culprit to me looks like GEN 1 GC, which is roughly 10x is high as it is in the Helios transport. My guess is that @Horusiath is right as to the reason why: we're allocating a number of `SwappedByteBuffer` objects in the process of doing this.

Other possibility is that something like the `log-transport` option defaulting to `true`, although after looking through the code it shouldn't be the case.

maxcherednik:
@Aaronontheweb @Horusiath guys, I haven't seen the exact changes here. Last time Aaron mentioned that dotnetty is nearly the same as Helios. I hope we keep the logging clean for the known case which we recently fixed on Helios.

Issue #2440, opened 2017-01-06
NuGet binary wrongly in the repository


SeanFarrow:
Having spoken with @aaronontheweb, it appears that the NuGet 3.4 binary has been erroneously commited in to the .NuGet folder of the repository.

I am proposing we remove this file along with the NuGet.targets and pull them from the users cache, ensuring that they are appropriately added to the .gitignore file.

The rationale behind this change would be that most developers have NuGet on there system already, and if they don't we can just refresh the file from the web. I'm happy to create a PR for this in the next few days.

Danthar:
Deleting the NuGet.targets breaks the build atm. But if you can resolve that. Go for it.
The simpler solution would be to only delete the NuGet.exe and add it to the .gitignore file (if its not in there already).
A PR is always welcome.

SeanFarrow:
The fact that removing the NuGet.targets file breaks the build means one or more projects is relying on the file. I'm looking at this currently, so should have something for you later today. Also, I'm sure this was in the .gitignore, so it shouldn't have been added in the first place.

PR #2445, opened 2017-01-11
Fix issue #2440

SeanFarrow:
Fixes issue #2440, but does not migrate away form using MSBuild style package restore, I'll do this as a separate issue.


Aaronontheweb:
@SeanFarrow  on Mono:

```
[15:39:14]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:14]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:14]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:14]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:14]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:14]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:14]	Cannot open assembly '/checkout/src/packages/FAKE/tools/FAKE.exe': No such file or directory.
[15:39:17]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:17]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:17]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:17]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:17]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:17]	Cannot open assembly '/checkout/src/.nuget/NuGet.exe': No such file or directory.
[15:39:17]	Cannot open assembly '/checkout/src/packages/FAKE/tools/FAKE.exe': No such file or directory.
```

Aaronontheweb:
@SeanFarrow Mono compilation step is still failing, same issue as before.

SeanFarrow:
@aaronontheweb,Where is this specifically failing?
I can’t replicate this locally!

Danthar:
@SeanFarrow The mono build uses the build.sh file to run the build. It seems the errors originate from that file. As far as i can see, it should download nuget.exe if its not there. But apparently there is a bug in that script.

SeanFarrow:
Ok, thanks, do you have any idea what the bug is?

Danthar:
It seems that the check that verifies if the nuget.exe is there does not work. Either that, or downloading the nuget.exe fails (or puts it in the wrong directory or something)

Aaronontheweb:
@SeanFarrow the process is still trying to access the `/src/.nuget/NuGet.exe'

SeanFarrow:
Which project specifically?
There could be a bug in the build.sh script—I don’t have a mono setup to test unfortunately!

Danthar:
@SeanFarrow me either, but you could check the code for obvious problems, and then abuse our CI integation to let you know if that fixed it ;). Long feedback cycle though. 

I can take a look this afternoon and see if i can spot any pathing problems. Ill let you know what i find.

Danthar:
Ah yes. The problem is in the build.sh file. nuget is downloaded as nuget.exe. While further down the process its being called as NuGet.exe (note the difference in capitals). Because its run on a unix machine it fails.

The reason current builds do work. Is because the NuGet.exe was already part of the artifacts. So even though the if condition that checks if the exe is there returns false, and it downloads a new version of nuget. There would effectively be 2 versions available side by side. A nuget.exe and a NuGet.exe. (again note the difference in capitals). 
Because you removed the NuGet.exe that was in git, this fails now.

If you want i can do a PR to fix this. Or you can make it part of this PR. Whatever you want.

SeanFarrow:
Hi,

I’ll fix this tomorrow, thanks!

Danthar:
ow man. You stirred a hornet nest here @SeanFarrow .
Next problem: "./build.sh: line 13: wget: command not found"

SeanFarrow:
Hi,
Any chance of you fixing?
Cheers
Sean.

Aaronontheweb:
The issue with wget is my fault. We switched Docker containers and wget is not included by default on the new container so we'll have to install it.

PR #2447, opened 2017-01-12
fixed race condition with creating priority queue

### LINKED ISSUE #2308: UnboundedPriorityMailbox's PriorityGenerator only called sometimes
I was talking to Aaron Stannard in the gitter chat about this. He asked I file a bug.

I created a priority mailbox and configured it, but sometimes the priority generator never gets invoked.

Here's a project that demonstrates the problem:

https://github.com/ronnieoverby/Akka.Mailbox.Bug

Aaronontheweb:
close #2308 - realized that this was the result of a weird design choice on my part back when we were working on releasing 1.1. Should never change the priority generator function at runtime. Now it's set via a constructor and is immutable, which should eliminate the issue.

CC @ronnieoverby

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -2614,6 +2614,8 @@ namespace Akka.Dispatch.MessageQueues
     }
     public class UnboundedPriorityMessageQueue : Akka.Dispatch.MessageQueues.BlockingMessageQueue
     {
+        [System.ObsoleteAttribute("Use UnboundedPriorityMessageQueue(Func<object, int> priorityGenerator, int initia" +
+            "lCapacity) instead.")]
         public UnboundedPriorityMessageQueue(int initialCapacity) { }
         public UnboundedPriorityMessageQueue(System.Func<object, int> priorityGenerator, int initialCapacity) { }
         protected override int LockedCount { get; }
@@ -4853,14 +4855,17 @@ namespace Akka.Util
         public override bool IsRight { get; }
         public TA Value { get; }
     }
-    public class ListPriorityQueue
+    public sealed class ListPriorityQueue
     {
+        [System.ObsoleteAttribute("Use ListPriorityQueue(initialCapacity, priorityCalculator) instead")]
         public ListPriorityQueue(int initialCapacity) { }
+        public ListPriorityQueue(int initialCapacity, System.Func<object, int> priorityCalculator) { }
         public int Count() { }
         public Akka.Actor.Envelope Dequeue() { }
         public void Enqueue(Akka.Actor.Envelope item) { }
         public bool IsConsistent() { }
         public Akka.Actor.Envelope Peek() { }
+        [System.ObsoleteAttribute("Use the constructor to set the priority calculator instead.")]
         public void SetPriorityCalculator(System.Func<object, int> priorityCalculator) { }
         public override string ToString() { }
     }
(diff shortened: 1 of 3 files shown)


src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2614,6 +2614,8 @@ namespace Akka.Dispatch.MessageQueues
     }
     public class UnboundedPriorityMessageQueue : Akka.Dispatch.MessageQueues.BlockingMessageQueue
     {
+        [System.ObsoleteAttribute("Use UnboundedPriorityMessageQueue(Func<object, int> priorityGenerator, int initia" +
```
Aaronontheweb: Marked all of the methods that don't have you explicitly set the `priorityGenerator` at construction time as obsolete.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -4853,14 +4855,17 @@ namespace Akka.Util
         public override bool IsRight { get; }
         public TA Value { get; }
     }
-    public class ListPriorityQueue
+    public sealed class ListPriorityQueue
```
Aaronontheweb: There's no really means for extending this queue given that all of the important parts of it are private fields.

src/core/Akka/Dispatch/MessageQueues/UnboundedPriorityMailboxQueue.cs
```diff
@@ -12,63 +12,63 @@
 namespace Akka.Dispatch.MessageQueues
 {
     /// <summary> 
-    /// Base class message queue that uses a priority generator for messages 
+    /// Base class for a message queue that uses a priority generator for messages
```
Aaronontheweb: Tried to fill in all of the TBDs that @sean-gilliam left as placeholders.

src/core/Akka/Util/ListPriorityQueue.cs
```diff
@@ -17,33 +17,53 @@ namespace Akka.Util
...
+        /// Creates a new priority queue.
         /// </summary>
-        /// <param name="initialCapacity">TBD</param>
-        public ListPriorityQueue(int initialCapacity)
+        /// <param name="initialCapacity">The initial capacity of the queue.</param>
+        /// <param name="priorityCalculator">The calculator function for assinging message priorities.</param>
+        public ListPriorityQueue(int initialCapacity, Func<object, int> priorityCalculator)
         {
             _data = new List<Envelope>(initialCapacity);
+            _priorityCalculator = priorityCalculator;
```
Aaronontheweb: Set all of the guts of the priority queue in the same place...

src/core/Akka/Dispatch/MessageQueues/UnboundedPriorityMailboxQueue.cs
```diff
@@ -12,63 +12,63 @@
...
         /// </summary>
-        /// <param name="priorityGenerator">TBD</param>
-        /// <returns>TBD</returns>
-        internal void SetPriorityGenerator(Func<object, int> priorityGenerator)
+        /// <param name="priorityGenerator">The calculator function for determining the priority of inbound messages.</param>
+        /// <param name="initialCapacity">The initial capacity of the queue.</param>
+        public UnboundedPriorityMessageQueue(Func<object, int> priorityGenerator, int initialCapacity)
         {
-            _prioQueue.SetPriorityCalculator(priorityGenerator);
+            _prioQueue = new ListPriorityQueue(initialCapacity, priorityGenerator);
```
Aaronontheweb: This is the key fix: create the queue WITH its priority generator at the same time and never change it again. That way we don't end up in a situation where the mailbox is having messages queued into it while the underlying priority generator is being changed concurrently on another thread.

PR #2451, opened 2017-01-13
verify that FSMs with constructor agruments can be instantiated

### LINKED ISSUE #1709: [TestKit] Modify ActorOfAsTestFSMRef to allow constructor arguments to underlying FSM<T,S>
I'd be an improvement, I think to the `ActorOfAsTestFSMRef` method if we could pass in constructor arguments that will be used to construct the underlying FSM we're testing. Otherwise, having to rely on a parameterless FSM in order to test it this way eliminates a lot of valid design choices.

Aaronontheweb:
close #1709 

Minor change, but it looks like #1709 is a non-issue. Will circle back with the user who reported it and see if I can get more information / verify that this spec actually addresses their issue.


Danthar:
Merged. some perf tests are failing. But only by a very small margin. And would attribute those to build server busyness.

PR #2457, opened 2017-01-19
Fix for #2456

### LINKED ISSUE #2456: Akka.Cluster.Tools client settings - setting buffer-size to 0 causes exception
I upgraded cluster tools from 1.1.1 to 1.1.2.30-beta and it appears that disabling the cluster client buffer now throws an exception:
**System.ArgumentException: BufferSize must be >= 0 and <= 10000**
```
                # If connection to the receptionist is not established the client will buffer
                # this number of messages and deliver them the connection is established.
                # When the buffer is full old messages will be dropped when new messages are sent
                # via the client. Use 0 to disable buffering, i.e. messages will be dropped
                # immediately if the location of the singleton is unknown.
                # Maximum allowed buffer size is 10000.
                buffer-size = 0
```

Danthar:
close #2456

- Changed range check in ClusterClientSettings
- Updated specs
- Verified that 0 is an actual valid value used in the ClusterClient


src/contrib/cluster/Akka.Cluster.Tools.Tests/ClusterClient/ClusterClientConfigSpec.cs
```diff
@@ -42,7 +42,7 @@ public void ClusterClientSettings_must_have_default_config()
         }
 
         [Theory]
-        [InlineData(0)]
+        [InlineData(-1)]
         [InlineData(10001)]
         public void ClusterClientSettings_must_throw_exception_on_wrong_buffer(int bufferSize)
```
Aaronontheweb: Need to add a separate spec that verifies that a 0 setting works and `ClusterClient` can start correctly, in addition to this one.

Danthar:
Disabled spec because it still has some issues. This way this fix can be merged with 1.1.3.
A new PR will be made to fix the spec.

Issue #2443, opened 2017-01-11
Akka.Remote generated types for Protobuf have no namespace and collide with framework types


julienadam:
In [src/core/Akka.Remote/Proto/Wireformats.cs](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote/Proto/Wireformats.cs), types are generated without a namespace.

Among them is CommandType which has the same name as System.Data.CommandType. This causes build issues like this :

``` csharp
using System.Data;
class Foo
{
    CommandType t = CommandType.StoredProcedure; // <-- Doesn't compile
}
```

Generated types should be in a namespace to avoid these name collisions. It shouldn't affect serialization, unless full type names are also serialized and / or used for serialization.

Danthar:
Its probably an oversight on our end that these are not in a proper namespace. We can enable this in our proto compile step. But im not sure if this would break backward compat. @akkadotnet/core Can someone comment on that ?

Aaronontheweb:
@Danthar these types are marked as `public`  but they're not really exposed in any meaningful way outside of the Akka.Remote built-in message type serializer. We could re-run the protobuf compiler with a namespace setting to correct this. Might have to stick a shim into the Akka.Remote serializers to account for the namespace difference but I don't think it'd be a catastrophic change.

Aaronontheweb:
Resolved by #2459

PR #2459, opened 2017-01-19
Fix for Issue #2443

Danthar:
I noticed that the resulting layout of the files are very different. Since the proto files are unchanged. It should not really matter.


Danthar:
This pr only addresses: #2443

However i noticed the other proto generated classes also, aren't in their respective namespaces.

Danthar:
Needs some more work to fix build issues.

Aaronontheweb:
@Danthar need to approve the updated API for Akka.Remote

src/core/Akka.API.Tests/CoreAPISpec.ApproveRemote.approved.txt
```diff
@@ -8,959 +8,6 @@
 [assembly: System.Runtime.InteropServices.ComVisibleAttribute(false)]
 [assembly: System.Runtime.InteropServices.GuidAttribute("78986bdb-73f7-4532-8e03-1c9ccbe8148e")]
 [assembly: System.Runtime.Versioning.TargetFrameworkAttribute(".NETFramework,Version=v4.5", FrameworkDisplayName=".NET Framework 4.5")]
-public sealed class AckAndEnvelopeContainer : Google.ProtocolBuffers.GeneratedMessage<AckAndEnvelopeContainer, AckAndEnvelopeContainer.Builder>
```
Aaronontheweb: While I think this change is good, we need to check that this doesn't break the wire format with versions of Akka.NET 1.1.3 and earlier

Danthar: uhm, your comparing the wrong side here?

src/core/Akka.API.Tests/CoreAPISpec.ApproveRemote.approved.txt
```diff
@@ -1309,6 +356,962 @@ namespace Akka.Remote
         public override string ToString() { }
     }
 }
+namespace Akka.Remote.Proto
+{
+    public sealed class AckAndEnvelopeContainer : Google.ProtocolBuffers.GeneratedMessage<Akka.Remote.Proto.AckAndEnvelopeContainer, Akka.Remote.Proto.AckAndEnvelopeContainer.Builder>
+    {
```
Danthar: This part is current code. 
As you can see, its still the same. Just wrapped in an namespace.

However because if the difference in layout. The entire file became hard to compare. Which is also why the api approval changes are huge.

Danthar:
I do understand your concerns though @Aaronontheweb . However short of manually lining up the ApiApproval before and after files. It seems like something hard to verify.

The main problem is probably that the newer version of protobuf used generated a different layout. Which makes comparing the changes hard.
What i could do, is regenerate the files without the namespace change. And then compare those against the files, with the namespace change. 
Then the layouts should be the same. So any differences should be visible much more clearly.

Im also open to other suggestions.

Aaronontheweb:
@Danthar run one actorsystem using the latest from NuGet and have it message a second instance created from your local branch with the protobuf changes over Akka.Remote. Only sure way to know.

Danthar:
Yeah I realized that. However I dismissed that option, because I feel like its not an exhaustive enough test ?

Either way, ill rebuild the ChatServer example for both versions of Akka and test with that.

Danthar:
Ok. I manually verified that 1.1.3 clients can connect and communicate with a client running the latest changes. And vice versa. 
Since i've used the ChatServer example for this. I have tested this with a server running the latest, and a server running the 1.1.3. And also with a mix of clients for both server versions, where i had a few clients running the 1.1.3 binaries, and a few running off of my branch.

It all worked like a charm. 👍

Aaronontheweb:
Cool, I'm satisfied then!

Aaronontheweb:
Tested this locally too. Was able to join a lighthouse cluster running Akka 1.1.3 using this branch, including the dotnetty transports!!!

Issue #2458, opened 2017-01-19
Add configuration option for Hyperion serializer


Horusiath:
I think, we should add some ability to configure hyperion serializer, as settings can have serious impact on its speed and size of the payload. For example, current setting of `versionTolerance=true` seriously impacts the size of the payload (for small messages manifest may be actually few times bigger than the payload itself due to requirement of encoding field info data), while some users may decide to resign from version tolerance for better throughput.

Therefore my proposition is to add few additional HOCON settings:

```hocon
hyperion {
    # When true, it will enable Hyperion sessions to keep object references 
    # used within serialized payload, instead of serializing objects each time. 
    # This is also necessary to handle circular references.
    preserve-references = true

    # When true, Hyperion will encode not only type name but also fields metadata 
    # inside payload manifest, giving limited support for modifications of message 
    # definitions.
    version-tolerance = true

    # Type name of the class implementing ITypeProvider interface used to supply
    # Hyperion with a list of types known to all communicating parties up front,
    # before staring the actor system. That class must implement a default constructor.
    known-types-provider = "Akka.Serialization.Hyperion.NoTypesProvided, Akka.Serialization.Hyperion"
}
```

While `ITypeProvider` can be customized to inform Hyperion about known types, so we don't need to encapsulate that knowledge in serialized manifest, using only identifiers instead.

```csharp
interface ITypeProvider
{
    IEnumerable<Type> GetKnownTypes();
}
```

This way user could i.e. provide all types withing given assembly, or identified by some attribute or interface.

Aaronontheweb:
Added in #2461

PR #2461, opened 2017-01-19
Akka.Serialization.Hyperion package

Aaronontheweb:


Horusiath:
Good to go once tests will pass - #2458 features can go in one of the future releases.

Danthar:
merged, failing perf test was a known stream-perf-test that almost always fails ;P

Issue #2197, opened 2016-07-18
Make Persist return a Task instead of taking callback


Horusiath:
Currently we have Persistance API directly matching the JVM version. Therefore, if we want to persist set of events at the present moment, we need something like that:

```csharp
Command<Cmd>(cmd => {
    var data = DoSomething(cmd);
    Persist(new Event(data), e => 
        this.Update(e));
});
```

The exploration area here would be to change that into:

```csharp
Command<Cmd>(cmd => {
    var data = DoSomething(cmd);
    Persist(new Event(data)).ContinueWith(e => 
        this.Update(e));
});
```

which can be simplified to:

```csharp
CommandAsync<Cmd>(async cmd => {
    var data = DoSomething(cmd);
    await Persist(new Event(data));
    this.Update(e);
});
```

Summarizing changing a Persist signature from `void Persist<T>(T evt, Action<T> callback)` to `Task Persist<T>(T evt)`.

Any thoughts on that?

alexpantyukhin:
Hi @JeffCyr @Horusiath  . I played a little bit with the current task. I faced a problem when substitute the call of `Action` with the running `Task` synchronosly the next task (which was added into the execution chain with the `ContinueWith` will run asynchronosly. Actually it's clear why it happens and how to fix it when add `Task` with `ContinueWith` (just add `TaskContinuationOptions.ExecuteSynchronously`) but it makes the api dirty. Is there way how to configure the `Task` when all `Task`'s which are added with `ContinueWith` will be runned synchronously without direct setting of `TaskContinuationOptions.ExecuteSynchronously` flag? Could you advise something?

JeffCyr:
I'm afraid not, however with async await it doesn't have this problem.

Issue #1956, opened 2016-05-14
Akka.Persistence enhancements 


Horusiath:
This is proposal of few possible changes that we could/should introduce in Akka.Persistence plugin to make it work in more reliable, user-friendly way.

##### Allow to use the same journal/snapshot store under different settings

This was mentioned many times already. It should be possible for one i.e. to use SqlServerJournal in two different scenarios (like custom domain events and cluster sharding), each of those working under different settings, writing events to different databases/tables.

##### Give ability to set persistence plugins programatically

Since we have no real support for HOCON, and it's whole schema is dynamic and not validated, there is a lot of config-related issues when working with persistence. To solve them, we could introduce gradual movement towards some more code-as-configuration way of working with settings. 

Example:

```csharp
using(var system = ActorSystem.Create("sys") {
    // set default journal, using default configuration path: akka.persistence.journal.sql-server
    system.UseJournal<SqlServer>(isDefault: true);
    // refer to some custom path  
    system.UseJournal<SqlServer>("akka.persistence.journal.cluster-sharding");
}
```

Any other ideas?

kantora:
Is it possible to set journal/snapshot configuration with use of `akka.actor.deployment` configuration section? So it can be actor specific.

Horusiath:
@kantora what would you need it for?

kantora:
Oops, sorry. Just missed it :(.

Danthar:
@Horusiath I like this idea. +1 from me.

Horusiath:
Ok some further ideas. Those could be used to ease the users experience through better async/await interop. Things like event persistence, snapshot related methods and recovery procedure all works in request/response model, where every request type has at least one success/failure acknowledgement. Example of an alternate persistent actor base class:

```csharp
abstract class PersistentActor<TState, TCommand, TEvent>
{
    protected PersistentActor(string persistenceId) { }

    /// Persistent state of an actor.
    public TState State { get; set; }

    /// Unique identifier of a persistent actor.
    public string PersistenceId { get; }

    /// Persistent actor's equivalent of Receive
    protected abstract Task OnCommand(TCommand command);

    /// Method called once persistent actor starts recovery procedure
    protected abstract void OnRecover(TEvent e);

    /// Method called when persistent actor's snapshot has been found.
    /// By default it just assigns it to <see cref="State"/>
    protected virtual void OnSnapshot(TState snapshot);

    /// Persists an event and completes once persist was successfully confirmed.
[...]

PR #2467, opened 2017-01-20
Added CommandAsync method to ReceivePersistentActor …

petercsala:
(based on ReceiveActor's ReceiveAsync method)

In order to be able to use ReceivePersistentActor in async fashion (just like you do it with ReceiveActor).

### DIFF
--- a/src/core/Akka.Persistence.Tests/ReceivePersistentActorTests.cs
+++ b/src/core/Akka.Persistence.Tests/ReceivePersistentActorTests.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Event;
 using Akka.TestKit;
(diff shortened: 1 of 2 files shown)


petercsala:
Hi Horusiath!

Thanks for your quick review. :)

1) Your are right. I'm gonna change them to Func<T, Task>.

2) In the ReceiveActor's ReceiveAsync methods I can't see how do they enforce "1 msg at the time" rule. May I ask you to please highlight where can I find  this rule enforcer?

3) Also could you give me a hint in which file (under Akka.Persistence.Tests project) should I put the tests?

Horusiath:
Regarding 2: I might be mistaken and your example may work indeed, but this needs to be tested.

Regarding 3:  Probably the best place will be ReceivePersistentActorTests.cs. What you need to check for sure, is that:

- No data races occur (so i.e. send 2 commands handled by async command handler, if first message will trigger `Persist` event and update of actor state, we must assert that second message will be handled after persist callback will be called and state will be updated).
- Mailbox is resumed correctly (so if we send a command with async handler, persist event callback for sure will be called).
- `Command` and `CommandAsync` can be mixed - if we send 1 command handled asynchronously, and then 2nd handled synchronously, we still won't have data races and mailbox will be resumed correctly (so we can process further messages and event callbacks).

petercsala:
In order to prevent data race it seems like we've to adjust Eventsourced's Persistent method as well, if i understand correctly. We have to know when does the persist end and when does the callback start. As I can see the Persisent method does not use the ActorTaskScheduler to call its callback async.

So I can see two ways:
- Decouple Persist method from its callback by making it await-able. 
- Use ActorTaskScheduler and Continuation (ContinueWith(handler))

Do you see any other way?

petercsala:
1) Command(handler) //1st message
 - Persist
2) Persist Callback
3) Command(handler) //2nd message

The persist callback is "injected" between commands (basically it's a continuation of the first command). It's done by the persist method (pendingInvocations list)

1) CommandAsync(async handler) //1st message
 - Persist
2) CommandAsync(async handler) //2nd message
3) Persist Callback

The async handlers are scheduled by ActorTaskScheduler. This scheduler doesn't know anything about the callback.

1) CommandAsync(async handler) //1st message
1/B) - await Persist
2) CommandAsync(async handler) //2nd message

Persist should be known by the ActorTaskScheduler as a continuation of the first command async handler. Unfortunately it seems way bigger project as it seemed previously. Fortunately there are also other initiatives that are trying to target the same.
(- https://github.com/akkadotnet/akka.net/issues/1956#issuecomment-267831403
- https://github.com/akkadotnet/akka.net/issues/2197)

Because of this I cancel this pull request.

Danthar:
This issue is not actively being worked on at the moment. So your effort is/was appreciated. But as you noticed. Its an significant amount of work :)

PR #2484, opened 2017-01-25
Possible fix for racy sql testkit specs

Danthar:
As i have discovered before. XUnit in combination with the akka TestKit causes some wierd stuff sometimes.
One thing that happens is that state you expect to be reinitialized for each test is sometimes reused between tests.

Reference to where i encountered a similar case: http://stackoverflow.com/questions/40767924/testing-behavior-not-consistent-when-watching-actor-for-termination/40929659#40929659

What this PR does, is move the SqlReadJournal reference to the local scope. Thereby eliminating any chance there is concurrent access to that resource where its not expected.

The EventsByTagSpec might need more refactoring. But worked consistently locally on the SqlLite tests implementation whereas before are very racy.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.TestKit/AllPersistenceIdsSpec.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.TestKit/AllPersistenceIdsSpec.cs
@@ -34,14 +34,14 @@ namespace Akka.Persistence.Sql.TestKit
         [Fact]
         public void Sql_query_AllPersistenceIds_should_implement_standard_AllPersistenceIdsQuery()
         {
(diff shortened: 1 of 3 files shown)


Danthar:
This might also help solve issues with: https://github.com/akkadotnet/Akka.Persistence.SqlServer/pull/56

Horusiath:
If it works I'm all for it :) My only remark is to keep up with conventions - if `_queries` are local now, there's no need to have underscore prefix.

Aaronontheweb:
@Danthar would you make those style changes that @Horusiath suggested?

Danthar:
Sure thing. First thing tomorrow.

PR #2487, opened 2017-01-27
[api-docs] Various fixes

sean-gilliam:
This PR contains various fixes including filling in TBD placeholders,
fixed numerous copyright headers and converted a repetitive class to a
T4 template.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -4639,7 +4639,7 @@ namespace Akka.Tools.MatchHandler
     public sealed class PartialHandlerArgumentsCapture<T, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15> : Akka.Tools.MatchHandler.IPartialHandlerArgumentsCapture<T>
     {
         public PartialHandlerArgumentsCapture() { }
-        public bool Handle(T message) { }
+        public bool Handle(T value) { }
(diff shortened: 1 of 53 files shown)


Danthar:
public api has changed. Dont think that was intentional :P

Aaronontheweb:
@Danthar @sean-gilliam marking a method with an `OSBSOLETE` attribute can trigger a public API change, since attribute metadata does get factored into that.

btw, what are the T4 templates for?

sean-gilliam:
It was just for one class. There was a lot of classes with the same boilerplate and signatures like Class<T, T1>, Class <T, T1, T2>,Class<T, T1, T2,  ...>. Just a lot of repetition. Changed it to make it easier to document.

The API change came in with one of that classes' methods. At the very end, it changes one of its parameters names. I fixed it to match the other method signatures. It's just a Utility class that user's prb wouldn't care about.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -4639,7 +4639,7 @@ namespace Akka.Tools.MatchHandler
     public sealed class PartialHandlerArgumentsCapture<T, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15> : Akka.Tools.MatchHandler.IPartialHandlerArgumentsCapture<T>
     {
         public PartialHandlerArgumentsCapture() { }
-        public bool Handle(T message) { }
+        public bool Handle(T value) { }
```
Aaronontheweb: Api change is just a name of the parameter, which shouldn't matter really.

PR #2489, opened 2017-01-30
[WIP] Support subscribe to distributed pubsub from Cluster Client

amichel:
Support subscribe to distributed pubsub from Cluster Client:

- Added support for subscribe message in client and subscribing actor to specific mesasge types
- Added support for subscribe message in receptionist, subscribing response tunnel on behalf of client
- Routed heartbeat response messages via tunnel to keep it alive as long as client is alive, otherwise subscription is lost
- Added example client and node applications


Danthar:
@amichel seems your PR causes a problem with package restore. Causing the build to fail. ..

amichel:
Not sure how to fix it. I can't reproduce locally.

amichel:
I tried clean rebuild in both debug and release modes and restore packages. All worked fine.

Danthar:
I can. You PR fails consistently for me locally. So im guessing you have some kind of project extension that needs some local support?

The error message is not very helpful either. So all i can say, is that for some reason the Fake helper cannot parse the solution file. 
All i can think of right now is some kind of VS extension that registers a project type GUID in one (or all) of your new projects that we dont have. Which causes the build to fail. 
But thats just a assumption on my part.

amichel:
I only have resharper and git extensions

amichel:
The problem was in SLN file, probably due to one of merges. Fixed it manually.

amichel:
Currently the tests that fail don't seem related to any of my commits.

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClient.cs
```diff
@@ -86,6 +88,35 @@ public Publish(string topic, object message)
...
         }
 
+        /// <summary> 
+        /// </summary>
+        [Serializable]
+        public sealed class Subscribe
+        {
+            public string Topic { get; }
+            public string Group { get; }
+            public Type MessageType { get; }
```
Horusiath: What is this for? Distributed pub/sub doesn't operate on type of the message.

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClient.cs
```diff
@@ -414,7 +471,11 @@ private void SendBuffered(IActorRef receptionist)
             while (_buffer.Count != 0)
             {
                 var t = _buffer.Dequeue();
-                receptionist.Tell(t.Item1, t.Item2);
+                var item1 = t.Item1 as Subscribe;
+                if (item1 != null)
+                    ProcessSubscribeMessage(item1, receptionist);
+                else
```
Horusiath: Why handling pub/sub as a special case?

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClient.cs
```diff
@@ -440,6 +501,74 @@ private void PublishContactPoints()
...
+                receptionist.Tell(new PublishSubscribe.Subscribe(subscribe.Topic, Self, subscribe.Group));
+        }
+
+        private void ProcessUnsubscribeMessage(Unsubscribe unsubscribe, IActorRef receptionist)
+        {
+            if (UnregisterExternalSubscriber(unsubscribe))
+                receptionist.Tell(new PublishSubscribe.Unsubscribe(unsubscribe.Subscribe.Topic, Self, unsubscribe.Subscribe.Group));
+        }
+
+        private bool RegisterExternalSubscriber(Subscribe subscribe)
```
Horusiath: It looks like you're trying to add every possible subscriber to the client instead of using mediators:

1. This means, that if we have 10 subscribers on node A and 10 on node B, we need to send 20 messages remotelly from the client, while we could just send 2 (to mediators on each of the nodes) and keep the rest of messages send local only.
2. You're not watching over actor refs, which means, that if an actor dies, it will never get removed from the subscribers set. Ultimatelly you may end up with garbage actor refs, that are never removed, pilling up in memory + a lot of dead letter messages.

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClient.cs
```diff
@@ -440,6 +501,74 @@ private void PublishContactPoints()
...
+            return subscribersCount == 0;
+        }
+
+        private bool TryHandleExternalSubcriptions(object message)
+        {
+            var type = message.GetType();
+            ImmutableHashSet<IActorRef> actors;
+            if (_externalSubscribers.TryGetValue(type, out actors))
+            {
+                actors.ForEach(a => a.Tell(message));
```
Horusiath: As mentioned before, we should contact with mediators instead of subscribers directly.

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Akka.Actor;
+
+namespace Akka.Cluster.Tools.Client
+{
+    public static class ClusterClientExtensions
+    {
+        public static Action Subscribe<T>(this IActorRef client, ClusterClient.Subscribe subscribe)
```
Horusiath: I understand reason behind this, but without having typed `IActorRef` this is risky API (this method can be called on any `IActorRef` not only on cluster client). We usually leave this kind of extension methods to be implemented by users for their convinience and keep message passing part explicit - not hidding it behind methods.

amichel:
I see you point. Probably too many years with MassTransit where pub/sub is strongly typed.
The limitation I see in current Akka implementation is that messages are delivered to subscribers as is without any envelope or header which contains topic and group, so I have no idea whether I need to route it to specific subscriber or not. 
Given this limitation, what you basically suggest is to simplify the solution, have a BroadcastGroup inside client and route all unresolved messages to it, then manage subscribers by adding/removing them to/from BroadcastGroup.

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClient.cs
amichel: Accepted. Answered in long comment below.

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClient.cs
amichel: To call Registersubscriber method and avoid duplicate code.

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClient.cs
amichel: 1. I only meant to have local subscribers - every node needs to subscribe on its own. So no case of sending 20 messages instead of 2 that I've seen, but it is not explicit, I agree with you. 
I could use DistributedPubSub mediator and publish with preference to local actors, but I can't do that due to limitationI explained in previous comment - that I don't have original topic/group that  message was published to when I receive it.
2. Right. Need to address it somehow. What is the best way to do it? Use router or have some Watcher implementation?

amichel:
Again. Failing tests don't seem related to my changes. Not sure what this test means: 
[http://petabridge-ci.cloudapp.net/viewLog.html?buildId=21101&buildTypeId=AkkaNet_AkkaNetWindowsUnitTests&guest=1#testNameId-3625424486371443843]

Rattenkrieg:
@amichel thats because you introduced changed somwhere in public methods. Run approval tests fron visual studio and you will be prompted with diff.

amichel:
Now I would appreciate another hint why tests are failing.

Aaronontheweb:
@amichel you need to rebase your changes. We fixed the perf issues that are failing. As for why the Mono tests are failing, no idea - probably one of the known racy specs was hit, but sometimes the Mono test runner process crashes before TeamCity gets a chance to capture the output.

In general though, why is this feature even necessary? Why does this capability have to be built directly into the `ClusterClient`? And what unit tests do you intend to add in order to ensure that the feature works and that there are no regressions?

amichel:
Hi @Aaronontheweb ,
1) This feature is necessary (at least for me) because I want to be able to connect and subscribe to topics from outside of cluster - I have different systems and services in different subnets and DMZ. I cannot open any/any between all subnets (regulations, security standards, and also it doesn't make sense) just to allow subscribers to get updates on their topics, just like with any other pubsub solution that I use  (RabbitMq, ZeroMq)where endpoint addresses and FW rules are known and can be easily defined and audited. 

2) I understand that Akka was not intended for client/server style communication, but ClusterClient is already providing out-of-cluster messaging Api and prevents p2p connections to the client itself. Since publishing is already implemented there and two-way communication is supported using temp address and Ask pattern, why not make it a subscriber as well? I considered doing it as extension and separate package, but it would be a much bigger effort and mostly duplicate code. Alternative solution was to tunnel messages to ZeroMq socket and use use ZMQ to extend pub/sub beyond cluster.

3) Regarding unit tests - I just started to work on it. Will need to extend the Cluster Client specs and then cover the changes in the Receptionist. 

If you think it is a wrong design decision or it doesn't comply with your roadmap I will switch to one of alternative solutions I mentioned.

Aaronontheweb:
@amichel I think this is a fine idea but we already have a number of built-in pub sub tools (DistributedPubSub and EventStream) and I don't see a need to add a third. When we take on a contribution, we inherit it and have to maintain it for the lifespan of the project. Therefore, that should be done carefully. This seems like something you could implement on top of `ClusterClient` using your own built-in messages and not something that the framework should support.

amichel:
@Aaronontheweb, I understand your concern. I don't see an easy way to do it over clusterclient/receptionist as most classes are sealed and I don't have any extensibility api to tap into. Practically it means I need to duplicate most of the code or maintain it in my fork and tell dev teams use custom clustertools nuget package from local feed.

Aaronontheweb:
@amichel this is something you should implement as shared message contracts between the actors your client communicates with vs. the actors who are listening in on the cluster in the server. Doesn't need to be built directly into `ClusterClient`

amichel:
I don't see how it's possible without keepalive of the tunnel I did and
setting the correct sender address. Messages will end up in dead letters
otherwise.

Horusiath:
@amichel a whole cluster client is implemented using the same API and tools, that are available to any developer ;) It's all akka primitives underneat. 

> I don't see how it's possible without keepalive of the tunnel I did

What do you mean by that? Are you talking about death watch between cluster client and receptionist?

PR #2490, opened 2017-01-30
Durable backend for Distributed Data collections

Horusiath:
This PR introduces a durable, persistent backend for ddata. It allows users to specify a list of keys, for which CRDTs should not only be gossiped among cluster nodes but also persisted using durable store. Just like in JVM case, default implementation here uses LMDB (through LigthingDB driver on .NET).

TODO list - my goal here is to make all multinode tests for ddata running before this PR gets merged:

- [ ] ReplicatorSpec
- [ ] ReplicatorPrunningSpec
- [ ] ReplicatorChaosSpec
- [ ] JepsenInspiredInsertSpec
- [ ] DurablePrunningSpec
- [ ] DurableDataSpec


src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
```diff
@@ -0,0 +1,222 @@
...
+using System.IO;
+using System.Text;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.DistributedData.Durable;
+using Akka.Event;
+using Akka.Serialization;
+using LightningDB;
+
+namespace Akka.DistributedData.LightningDB
```
Horusiath: JVM version by default uses LMDB backend for storage. I think, it's better to move that dependency to a separate package however.

Horusiath:
At this point I've already fixed over 5 different bugs while trying to make `ReplicatorSpec` passing. Some of them are critical. Right now I'm struggling to find a next one: for some reason it looks like not all keys gets replicated back to the reconnecting node (scenario: *disconnect&rarr;update while disconnected&rarr;reconnect again and wait for replicas to converge*).

src/contrib/cluster/Akka.DistributedData.Tests.MultiNode/ReplicatorSpec.cs
```diff
@@ -401,22 +421,24 @@ public void Cluster_CRDT_should_converge_after_partition()
...
                 {
-                    var keydn = new GCounterKey("D" + i);
-                    _replicator.Tell(Dsl.Update(keydn, GCounter.Empty, WriteLocal.Instance, x => x.Increment(_cluster.SelfUniqueAddress, i)));
-                    ExpectMsg(new Replicator.UpdateSuccess(keydn, null));
+                    var n = i;
+                    var keydn = new GCounterKey("D" + n);
+                    _replicator.Tell(Dsl.Update(keydn, GCounter.Empty, WriteLocal.Instance, x => x.Increment(_cluster, n)));
+                    ExpectMsg(new UpdateSuccess(keydn, null));
                 }
             }, _config.First);
```
Horusiath: I'm not sure if this is 100% reproducible in each case but it looks like this pattern is a bug in MNTK:

```csharp
EnterBarrier("after-1");

RunOn(() => {
    // this one gets called
}, firstRole, secondRole);

RunOn(() => {
    // this one never gets called
}, firstRole);

EnterBarrier("after-2");
```

Aaronontheweb:
@Horusiath what will it take to get this PR into a state so we can include the bug fixes for 1.2?

src/contrib/cluster/Akka.DistributedData.Tests.MultiNode/ReplicatorSpec.cs
```diff
@@ -505,73 +537,74 @@ public void Cluster_CRDT_should_support_majority_quorum_write_and_read_with_3_no
...
+            }, _first);
 
             EnterBarrier("passThrough-third");
 
             RunOn(() =>
             {
                 _replicator.Tell(Dsl.Get(KeyE, _readMajority));
-                var c155 = ExpectMsg<Replicator.GetSuccess>(g => Equals(g.Key, KeyE)).Get(KeyE);
+
+                var c155 = ExpectMsg<GetSuccess>(g => Equals(g.Key, KeyE)).Get(KeyE);
```
Horusiath: In this place spec fails. Basically what we've done up to this point was:

1. Establish 3-node cluster with replicator instance on each node.
2. Perform some updates on each node.
3. Blackhole 3rd node from the rest of the cluster, making it unreachable.
4. Perform more CRDT updates on each node.
5. Pass through unreachable 3rd node again to the rest of the cluster.
6. While 3rd node is up, replicators should send updates as part of Get request with read majority, and finally converge, but the exception occurs instead, causing node disassociation.

What I've managed to find is that upon marking 3rd node as reachable again - when all nodes have Up state - replicators try to send messages using `Context.ActorSelection(Context.Parent.Path.ToStringWithAddress(address))` (Context.Parent is a replicator instance here). But even when I've confirmed that node under `address` is acknowledged as up, the message never reaches the target. ResolveOne will also cause an exception in this case. I believe that this may be problem with remoting/cluster layer, after unreachable node becomes reachable again.

/cc @Aaronontheweb

alexvaluyskiy: I have the same problem in `SurviveNetworkInstabilitySpec`

Issue #2492, opened 2017-01-31
Cluster.MemberRemoved does not always fire


Aaronontheweb:
Looks like #2455 is a symptom of this issue.

Under some circumstances, `MemberRemoved` is not correctly propagated from the leader to its children.

Some context from an end-user in our Gitter chat earlier today:

> When a node gets into this state it doesn't leave cleanly. It tries to, I have a monitor running in all service discovery nodes (2 of them). They both report the cluster status that they see. When this problem happens the cluster status us everything UP and everything Seen. The leader gets the request that the node is exiting and this is logged every second that it is moving the node to exiting but it never exits. After 15 seconds my windows service will kill the service and failure detection will kick in. I manually down the node although just starting it again causes the cluster to see the new node and remove the old one. The cluster monitors both report the node is removed. The node rejoins and gets stuck.

Not entirely clear what the exact issue is, but this should give us enough information to go looking for it.

maxcherednik:
@Aaronontheweb I was able to reproduce similar case. The setup is really easy:
1 seed node and 4-5 normal nodes. Every node has a graceful shutdown logic, which waits for member removed and then closes the application. I run all the nodes locally and then very fast one by one send Ctrl+C to every node but the seed node. Sometimes 1 node fail to stop.

crucifieddreams:
Apologies for the delay in posting a reply folks. It takes a bit of time to get approval to post logs and screenshots (i'm pretty sure glaciers move faster) as well as gather data that may be useful and illustrate the problem. 

I believe as @maxcherednik has already reproduced that you are correct @Aaronontheweb. I've attached a log as a PDF of our SEQ logs (getting things out of the network is tedious even with approval :), pdf was about the best I could manage apologies) showing an exit from the cluster as well as screenshots of before and after the event as reported in one of our cluster monitor nodes. The monitors are a WCF service hosted inside our process running our service discovery actor systems. The service discovery essentially being lighthouse with a few extra bits (we've yet to move to cluster client which would allow us to do the same thing as the hosted WCF service provides). 

![hurrah](https://cloud.githubusercontent.com/assets/13729156/22528690/4e054f9c-e8cc-11e6-88c1-d86ae8f4f9ff.png)
![oops](https://cloud.githubusercontent.com/assets/13729156/22528694/506053f4-e8cc-11e6-8421-d533528317c7.png)

[...]

PR #2498, opened 2017-02-02
[WIP] possible fix for #2492

maxim-s:


Horusiath:
It sounds super! ... if we actually can verify it somehow :) I've seen, that people from #2492 have shown a possible repro steps for this issue. Can we verify it somehow?

maxim-s:
@Horusiath  in progress of reproducing this case

maxim-s:
also Akka.Cluster.Tests.ClusterSpec.A_cluster_must_complete_LeaveAsync_task_upon_being_removed racy spec

alexvaluyskiy:
@maxim-s Have you tried to implement the whole PR from Akka Scala? https://github.com/akka/akka/pull/21859

maxim-s:
@alexvaluyskiy I didn't see this PR, but looks like we have the same issue with random gossip

maxim-s:
@Aaronontheweb looks like test Akka.Cluster.Tests.ClusterSpec.A_cluster_must_complete_LeaveAsync_task_upon_being_removed is not correct now, could you review it?

src/core/Akka/Util/Internal/Collections/ListExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+
+namespace Akka.Util.Internal.Collections
+{
+    public static class ListExtensions
+    {
```
alexvaluyskiy: Internal should be internal, not public. Use InternalVisibleTo attribute

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -935,7 +936,7 @@ internal class ClusterCoreDaemon : UntypedActor, IRequiresMessageQueue<IUnbounde
         /// TBD
         /// </summary>
         protected readonly UniqueAddress SelfUniqueAddress;
-        private const int NumberOfGossipsBeforeShutdownWhenLeaderExits = 3;
+        private const int NumberOfGossipsBeforeShutdownWhenLeaderExits = 5;
```
Aaronontheweb: What was the purpose of increasing this value? Not questioning the decision to do it, but just want to understand the reason for the change.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1845,6 +1854,21 @@ public bool IsGossipSpeedupNeeded()
...
+
+        /// <summary>
+        /// Sends full gossip to `n` other random members.
+        /// </summary>
+        private void SendGossipRandom(int n)
+        {
+            if (!IsSingletonCluster && n > 0)
+            {
+                var localGossip = _latestGossip;
+                var possibleTargets = new List<UniqueAddress>(localGossip.Members.Where(m => ValidNodeForGossip(m.UniqueAddress)).Select(m => m.UniqueAddress));
```
Aaronontheweb: just call `ToList` here

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -2164,7 +2180,7 @@ public void ReapUnreachableMembers()
                                 _cluster.SelfAddress, nonExiting.Select(m => m.ToString()).Aggregate((a, b) => a + ", " + b), string.Join(",", _cluster.SelfRoles));
 
                         if (!exiting.IsEmpty)
-                            _log.Warning("Marking exiting node(s) as UNREACHABLE [{0}]. This is expected and they will be removed.",
+                            _log.Warning("Cluster Node [{0}] - Marking exiting node(s) as UNREACHABLE [{1}]. This is expected and they will be removed.",
```
Aaronontheweb: Good catch.

src/core/Akka.Cluster/ClusterDaemon.cs
Aaronontheweb: Ah, I get it now - increases the number of random nodes we might gossip to when the leader exits.

src/core/Akka/Util/Internal/Collections/ListExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        public static List<T> Shuffle<T>(this List<T> @this)
```
Aaronontheweb: Which shuffle algorithm are you using here? Fischer-Yates?

src/core/Akka/Util/Internal/Collections/ListExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var list = new List<T>(@this);
+            var rng = new Random();
```
Aaronontheweb: Use `ThreadLocalRandom` here - avoids allocations and also spreads out the seeds (as `ThreadLocalRandom` gets used gradually across available threads)

src/core/Akka.Cluster/ClusterDaemon.cs
maxcherednik: @Aaronontheweb Can we do this one as INFO if it is a normal exit?

src/core/Akka/Util/Internal/Collections/ListExtensions.cs
maxim-s: thank you but I thought to replace with Math.round and Math.random as Java implementation

src/core/Akka/Util/Internal/Collections/ListExtensions.cs
maxim-s: not yet, just extracted class

maxim-s:
@Aaronontheweb racy spec I mean, that some sort of race in the LeaveAsync task, MemberRemoved has been received success, problem with callback and I'm working on it

Aaronontheweb:
@maxim-s oh I see what you mean. I must have misunderstood. So you think it's an issue with how the callback is implemented and not the spec itself?

alexvaluyskiy:
@maxim-s The better way is: implement https://github.com/akka/akka/pull/21859 first in a one PR. And only after merge try to fix https://github.com/akkadotnet/akka.net/issues/2492
We have to very carefully to add something which is not exists in Akka JVM

src/core/Akka.Cluster/ClusterDaemon.cs
Aaronontheweb: @maxcherednik @maxim-s yep, I think that's a good idea.

maxim-s:
@Aaronontheweb yes this PR https://github.com/akkadotnet/akka.net/pull/2501 should solve problem with `Akka.Cluster.Tests.ClusterSpec.A_cluster_must_complete_LeaveAsync_task_upon_being_removed`

src/core/Akka.Cluster/ClusterDaemon.cs
alexvaluyskiy: Akka (JVM) does not have "Cluster Node" parameter in this log entry. Maybe we could just remove the parameter, instead of extending the log message?

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -2164,7 +2183,7 @@ public void ReapUnreachableMembers()
...
                                 _cluster.SelfAddress, exiting.Select(m => m.ToString()).Aggregate((a, b) => a + ", " + b));
```
alexvaluyskiy: We could simplify this code and use something like this
```
_log.Info("Marking exiting node(s) as UNREACHABLE [{0}]. This is expected and they will be removed.",  string.Join(", ", exiting));
```

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1125,6 +1126,7 @@ private void TryingToJoin(object message, Address joinWith, Deadline deadline)
             else if (message is InternalClusterAction.JoinSeedNodes)
             {
                 var js = message as InternalClusterAction.JoinSeedNodes;
+                BecomeUninitialized();
```
alexvaluyskiy: Good catch!

src/core/Akka.Cluster/ClusterDaemon.cs
Aaronontheweb: I think it's fine

Aaronontheweb:
I approve the changes. @alexvaluyskiy how do you feel? Are there changes you still want to see made?

alexvaluyskiy:
@Aaronontheweb I've already reviewed the code and approved it.

Issue #2469, opened 2017-01-20
[MNTK] Log node Role names instead of index


Danthar:
In the MNTK when you start a spec with multiple nodes. The logs log stuff like Node1, Node2, Node3 etc.
This does not help in debugging an MNTK spec. Because it forces you figure out which nodeX is which Role in your Spec.

So an improvement would be to use the names of the Roles which are defined in the Spec itself. That would make the whole thing alot more transparent.

Rattenkrieg:
Well, that one wasn't the easiest.

Finally I've managed to get node role printed everywhere, but I still have some concerns regarding design.
![default](https://cloud.githubusercontent.com/assets/6948307/22408854/63f75a6c-e691-11e6-8bda-5ea44fc41df0.png)





In current implementation node indexes are extracted as follows:

On _Akka.MultiNodeTestRunner_ side Xunit controller walks through assembly and executes _Discovery_ class callback passing metadata regarding test method and its surroundings
Logic in this callback looks for classes that has trailing digits in its names, cuts these digits and treats them as node indexes. These classes should be descendants of `MultiNodeSpec` by convention. Every such class will be instantiated as node in a dedicated process.
https://github.com/akkadotnet/akka.net/blob/v1.1.3/src/core/Akka.MultiNodeTestRunner/Discovery.cs#L65
Then _Akka.NodeTestRunner.exe_ is called for each node with node's index, node's class and test assembly in arguments list.
On _NodeTestRunner_ side node's class is bootstrapped with its own `MultiNodeConfig` instance.
Config provides `RoleName` instances for `MultiNodeSpec` which uses it in `RunOn` calls.

Currently there is no relation between `RoleName`s and node indexes at discovering time.

Thats what I have fixed. Instead of matching classes with trailing digits there is reflection now which builds relation `MultiNodeSpec` <~> `MultiNodeConfig` by looking in `MultiNodeSpec` constructor and extracting `RoleName`s from config
https://github.com/Rattenkrieg/akka.net/blob/mntk-role-names-2469/src/core/Akka.MultiNodeTestRunner/Discovery.cs#L59
Only non-abstract `MultiNodeSpec` descendants are treated as runnable nodes, there will be N instantiations of node in dedicated _Akka.NodeTestRunner_ process where N is number of properties `typeof(RoleName)` in corresponding `MultiNodeConfig` descendant declaration.
_Akka.NodeTestRunner.exe_ is called with additional argument _-Dmultinode.role_
[...]

PR #2502, opened 2017-02-05
[WIP] [MNTK] Logging role names; less boilerplatish specs-coding process

Rattenkrieg:
**Not ready for merge**

Solution for #2469 

- [x]  PoC for discovery-time <_node id - node role_> mapping construction
- [x]  Printing/reporting routines adjustment
- [x]  Refactoring Akka.Remote.Tests.MultiNode
- [x]  Refactoring Akka.Cluster.Tests.MultiNode
- [x]  Refactoring Akka.Cluster.Tools.Tests.MultiNode
- [x]  Solution to avoid depending on assemblies being tested in MultiNodeTestRunner
- [x]  Unit-tests for new approach

### DIFF
--- a/src/core/Akka.MultiNodeTestRunner.Shared/Sinks/ConsoleMessageSinkActor.cs
+++ b/src/core/Akka.MultiNodeTestRunner.Shared/Sinks/ConsoleMessageSinkActor.cs
@@ -42,7 +42,7 @@ namespace Akka.MultiNodeTestRunner.Shared.Sinks
             WriteSpecMessage(string.Format("Start time: {0}", new DateTime(data.StartTime, DateTimeKind.Utc)));
             foreach (var node in data.NodeFacts)
             {
-                WriteSpecMessage(string.Format(" --> Node {0}: {1} [{2} elapsed]", node.Value.NodeIndex,
+                WriteSpecMessage(string.Format(" --> Node {0}:{1} : {2} [{3} elapsed]", node.Value.NodeIndex, node.Value.NodeRole,
                     node.Value.Passed.GetValueOrDefault(false) ? "PASS" : "FAIL", node.Value.Elapsed));
             }
             WriteSpecMessage(string.Format("End time: {0}",
@@ -58,14 +58,14 @@ namespace Akka.MultiNodeTestRunner.Shared.Sinks
                 {
                     if (node.Value.Passed.GetValueOrDefault(false) == false)
                     {
-                        WriteSpecMessage(string.Format("<----------- BEGIN NODE {0} ----------->", node.Key));
+                        WriteSpecMessage(string.Format("<----------- BEGIN NODE {0}:{1} ----------->", node.Key, node.Value.NodeRole));
                         foreach (var resultMessage in node.Value.ResultMessages)
                         {
                             WriteSpecMessage(String.Format(" --> {0}", resultMessage.Message));
                         }
                         if(node.Value.ResultMessages == null || node.Value.ResultMessages.Count == 0)
                             WriteSpecMessage("[received no messages - SILENT FAILURE].");
-                        WriteSpecMessage(string.Format("<----------- END NODE {0} ----------->", node.Key));
+                        WriteSpecMessage(string.Format("<----------- END NODE {0}:{1} ----------->", node.Key, node.Value.NodeRole));
                     }
                 }
             }
@@ -73,7 +73,7 @@ namespace Akka.MultiNodeTestRunner.Shared.Sinks
 
         protected override void HandleNodeSpecFail(NodeCompletedSpecWithFail nodeFail)
         {
-            WriteSpecFail(nodeFail.NodeIndex, nodeFail.Message);
+            WriteSpecFail(nodeFail.NodeIndex, nodeFail.NodeRole, nodeFail.Message);
 
             base.HandleNodeSpecFail(nodeFail);
         }
@@ -132,7 +132,7 @@ namespace Akka.MultiNodeTestRunner.Shared.Sinks
 
         protected override void HandleNodeSpecPass(NodeCompletedSpecWithSuccess nodeSuccess)
         {
-            WriteSpecPass(nodeSuccess.NodeIndex, nodeSuccess.Message);
+            WriteSpecPass(nodeSuccess.NodeIndex, nodeSuccess.NodeRole, nodeSuccess.Message);
 
             base.HandleNodeSpecPass(nodeSuccess);
         }
@@ -151,17 +151,17 @@ namespace Akka.MultiNodeTestRunner.Shared.Sinks
             Console.ResetColor();
         }
 
-        private void WriteSpecPass(int nodeIndex, string message)
+        private void WriteSpecPass(int nodeIndex, string nodeRole, string message)
         {
             Console.ForegroundColor = ConsoleColor.Green;
-            Console.WriteLine("[NODE{0}][{1}]: SPEC PASSED: {2}", nodeIndex, DateTime.UtcNow.ToShortTimeString(), message);
+            Console.WriteLine("[NODE{0}:{1}][{2}]: SPEC PASSED: {3}", nodeIndex, nodeRole, DateTime.UtcNow.ToShortTimeString(), message);
             Console.ResetColor();
         }
 
-        private void WriteSpecFail(int nodeIndex, string message)
+        private void WriteSpecFail(int nodeIndex, string nodeRole, string message)
         {
             Console.ForegroundColor = ConsoleColor.Red;
-            Console.WriteLine("[NODE{0}][{1}]: SPEC FAILED: {2}", nodeIndex, DateTime.UtcNow.ToShortTimeString(), message);
+            Console.WriteLine("[NODE{0}:{1}][{2}]: SPEC FAILED: {3}", nodeIndex, nodeRole, DateTime.UtcNow.ToShortTimeString(), message);
             Console.ResetColor();
         }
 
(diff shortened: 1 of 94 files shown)


Rattenkrieg:
Got rid of remote.tests.multinode references.
[More reflection was employed. ](https://github.com/akkadotnet/akka.net/blob/de510a18de7c1b14f01df5d9f8bdf592866c3373/src/core/Akka.MultiNodeTestRunner/Discovery.cs#L61)
[More command-line arguments were employed.](https://github.com/akkadotnet/akka.net/blob/de510a18de7c1b14f01df5d9f8bdf592866c3373/build.fsx#L322)
Ready for review. Tests on the way.

build.fsx
```diff
@@ -319,6 +319,10 @@ Target "MultiNodeTests" <| fun _ ->
         let args = new StringBuilder()
                 |> append assembly
                 |> append "-Dmultinode.enable-filesink=on"
+                |> append "-Dmultinode.remote-testkit-dll=Akka.Remote.TestKit.dll"
```
Aaronontheweb: So what's this for, telling the MNTR where to find the testkit assembly?

src/core/Akka.MultiNodeTestRunner/Akka.MultiNodeTestRunner.csproj
```diff
@@ -73,14 +73,6 @@
       <Project>{964F0EC5-FBE6-47C5-8AE6-145114D5DB8C}</Project>
       <Name>Akka.MultiNodeTestRunner.Shared</Name>
     </ProjectReference>
-    <ProjectReference Include="..\Akka.NodeTestRunner\Akka.NodeTestRunner.csproj">
```
Aaronontheweb: No need for the `NodeRunner.exe` any more?

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/MessageSink.cs
```diff
@@ -88,27 +88,27 @@ public enum MultiNodeTestRunnerMessageType
...
         private const string NodeFailureReasonRegexString =
-            @"\[(\w){4}(?<node>[0-9]{1,2})\]\[(?<status>(FAIL-EXCEPTION))\]{1}\s(?<message>.*)";
+            @"\[(\w){4}(?<node>[0-9]{1,2})(?<role>:\w+)?\]\[(?<status>(FAIL-EXCEPTION))\]{1}\s(?<message>.*)";
         protected static readonly Regex NodeFailureReasonRegex = new Regex(NodeFailureReasonRegexString);
 
         /*
          * Regular expressions - go big or go home. [Aaronontheweb]
          */
-        private const string NodeLogMessageRegexString = @"\[(\w){4}(?<node>[0-9]{1,2})\]\[(?<level>(\w)*)\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\](?<thread>\[(\w|\s)*\])\[(?<logsource>(\[|\w|:|/|\(|\)|\]|\.|-|\$|%|\+|#|\^|@)*)\]\s(?<message>(\w|\s|:|<|\.|\+|>|,|\[|/|-|]|%|\$|\+|\^|@|\(|\))*)";
+        private const string NodeLogMessageRegexString = @"\[(\w){4}(?<node>[0-9]{1,2})(?<role>:\w+)?\]\[(?<level>(\w)*)\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\](?<thread>\[(\w|\s)*\])\[(?<logsource>(\[|\w|:|/|\(|\)|\]|\.|-|\$|%|\+|#|\^|@)*)\]\s(?<message>(\w|\s|:|<|\.|\+|>|,|\[|/|-|]|%|\$|\+|\^|@|\(|\))*)";
```
Aaronontheweb: We don't really use these `Regex` anymore, now that most node-specific logs are just written directly to their own file

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/ConsoleMessageSinkActor.cs
```diff
@@ -151,17 +151,17 @@ private void WriteSpecMessage(string message)
...
+            Console.WriteLine("[NODE{0}:{1}][{2}]: SPEC PASSED: {3}", nodeIndex, nodeRole, DateTime.UtcNow.ToShortTimeString(), message);
             Console.ResetColor();
         }
 
-        private void WriteSpecFail(int nodeIndex, string message)
+        private void WriteSpecFail(int nodeIndex, string nodeRole, string message)
         {
             Console.ForegroundColor = ConsoleColor.Red;
-            Console.WriteLine("[NODE{0}][{1}]: SPEC FAILED: {2}", nodeIndex, DateTime.UtcNow.ToShortTimeString(), message);
+            Console.WriteLine("[NODE{0}:{1}][{2}]: SPEC FAILED: {3}", nodeIndex, nodeRole, DateTime.UtcNow.ToShortTimeString(), message);
```
Aaronontheweb: Love this change. Having the rolename listed in the log will make things so much easier.

src/core/Akka.MultiNodeTestRunner/Akka.MultiNodeTestRunner.csproj
Rattenkrieg: I have deleted this by accident when was cleaning now unnecessary Akka.Remote.TestKit

Rattenkrieg:
Regarding dependency on Akka.Remote.TestKit:
I have implemented discovery routine using only names of types and loading Akka.Remote.TestKit assembly from file passed through commandline. Yesterday everything worked ok, but now when I'm running tests nothing being discovered, and in debug I'm getting weird exceptions. After investigation I have found that Akka.Remote.TestKit is sitting in my Akka.MultiNodeTestRunner/bin folder and interfering with Akka.Remote.TestKit which I'm looking somwhere in <AssemblyUnderTest>/bin. Guess where it came from? 
This is the line from Akka.MultiNodeTestRunner.csproj:
`<Compile Include="..\Akka.Remote.TestKit\CommandLine.cs">`
So we are already depending on Akka.Remote.TestKit but in less obvious way.
I have decided to bring back reflection based on `typeof(MultiNodeConfig)` and sweeping flooded commandline args [here](https://github.com/akkadotnet/akka.net/pull/2502/files/de510a18de7c1b14f01df5d9f8bdf592866c3373#diff-6074983490330a0aac1bd23c958232d7)
Code will be less fragile that way.

build.fsx
Rattenkrieg: Will remove. For explanation check my comments below.

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/MessageSink.cs
Rattenkrieg: But there is entire _ParsingSpec_ in _Akka.MultiNodeTestRunner.Shared.Tests_ using this regexp.

Rattenkrieg:
Current fails looks transient. Need someone to confirm/re-run.
Everything planned is done, waiting for review.

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/MessageSink.cs
Aaronontheweb: We can remove that spec.

Aaronontheweb:
@Rattenkrieg have a merge conflict with the `ClusterSingletonManagerLeaveSpec`, which was updated in a recent PR.

Aaronontheweb:
OMG, I love this

image

So much easier to figure out WTF is going on now that the role names are included in the per-node and in the aggregate logs.

Aaronontheweb:
@Rattenkrieg so, binary compatibility question: this will require users to upgrade just the `Akka.MultiNodeTestRunner` package they're using, right? Or does the version of the `Akka.Remote.TestKit` they're using also have to be upgraded?

Rattenkrieg:
@Aaronontheweb due to [this](https://github.com/akkadotnet/akka.net/pull/2502/files#diff-c09922238719191ad5e2f7211d0a01e7) change `Akka.Remote.Testkit` is also affected. But this parameter was added mostly for convinience. I can still manipulate with node index on that side though it makes me less confident about order being preserved.

Aaronontheweb:
@Rattenkrieg eh, I don't think that change is the end of the world. Benefits outweigh the costs of upgrading IMHO. 

As part of this PR, could you also update https://github.com/akkadotnet/getakka.net/blob/master/src/docs/testing/multinode-testkit.md ?

Rattenkrieg:
@Aaronontheweb rebased on head of dev, fixed conflicts. Now going to update the docs. As for regex foo you mentioned above is hard to track every caller of `MessageSink` but seems it all ends [here](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.MultiNodeTestRunner.Shared/Sinks/SinkCoordinator.cs#L156) and I can't see any `Tell<string>(...)` to that actor. According to that observations it's safe do delete this regex and everything down the road. But I think this is story for separate PR which I can address after finishing this one.

Horusiath:
@Rattenkrieg My question regarding this PR: what scope of changes has affected visualizer? Will it show logs or barrier entries on the timeline? That would be super useful.

Rattenkrieg:
@Horusiath That html file genereated is the product of "Visualizer"? If so then it was affected as everything where I had found usage of nodeid. Check the screenshot below:
default
Don't hesitate to ask for changes if I have missed some logs / or if you think that some logs could be done better.

PR #2503, opened 2017-02-05
Send terminationMessage to singleton when leaving last

alexvaluyskiy:
Implemented [send terminationMessage to singleton when leaving last](https://github.com/akka/akka/pull/21654) [2.4.12]


alexvaluyskiy:
@Aaronontheweb whats happened with Perf tests?

PR #2504, opened 2017-02-06
removed issuestats and outdated milestone from ReadMe

Aaronontheweb:


Aaronontheweb:
IssueStats no longer appears to be reliable. Images show up as broken in the readme.

PR #2506, opened 2017-02-08
squelched the Akka.Streams benchmarks with the highest flip rate

Aaronontheweb:
Reviewed the past couple of months worth of data from [TeamCity's Flaky Test Report](http://petabridge-ci.cloudapp.net/project.html?projectId=AkkaNet&tab=flakyTests&branch_AkkaNet=__all_branches__&guest=1), and opted to disable assertions for the Akka.Streams benchmarks with the highest flip rate. Some of these tests have a fail rate that's approaching ~50%, and I think it's because all of these specs work with the highest number of flows (import 1000 flows,etc...)

I'm proposing that we squelch them so CI fails less often. The cases these benchmarks are asserting are just the largest in the group (except for maybe `FusedGraphsBenchmark+RepeatTakeSelectAndAggregate`) and there are smaller versions of them all pass with no problem. Alternatively, we can just raise the ceiling on these assertions, since they assume that "if 10 X can be done in N milliseconds, then 100 X can be done in 10 * N milliseconds," which probably doesn't hold true in these cases.


Danthar:
Squelsh them now, and create an seperate issue to raise the ceiling ? That way we can solve the issue _now_. And have someone with to much time contribute by having him find out what the right timings should be.

(or if we fix it in a different way, e.g. with improvements to nbench or something). Either way, that way it would be visible for more then just the people who frequent TC status logs :P

Silv3rcircl3:
Maybe we need some more predictable hardware for those benchmarks, would also give us more trust in failing tests. But I'm fine if we disable the biggest ones for now, performance drops in feature changes are hopefully covered with the smaller ones.

PR #2514, opened 2017-02-11
Customizable serializers

Horusiath:
### Motivation

Some of the users lack an ability to customize used serializers i.e. adding custom converters to json.net serializer. Another time some of them may want to harden serializer constraints on themselves in order to achieve better performance. This is especially a case with new Hyperion serializer, which uses things like preserving object references and version tolerance, which both increase the serialization time and a payload size, causing our default hyperion settings to be actually always a worst case scenario for that library.

### Description

This PR introduces an ability to add custom options to serializers through HOCON config. It can be used as a general purpose mechanism. From now, users may define their own serializer configuration within `akka.actor.serialization-settings` i.e.:

```hocon
akka.actor {
  serializers {
    hyperion = "Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion"
  }
  serialization-bindings {
    "System.Object" = hyperion
  }
  serialization-settings {
    hyperion {
      version-tolerance = false
      preserve-object-references = false
    }
  }
}
```

This way a `Config` object holding an `akka.actor.serialization-settings.hyperion` section will be passed as a second parameter to constructor of a serializer aliased as "hyperion" in `akka.actor.serializers`. From there, a serializer class is free to interpret it. This behavior is done at `Serialization` class level, so it's applied automatically to any serializer, that wants to use it.

This PR introduces this behavior on 2 serializers:

1. `NewtonSoftJsonSerializer` may optionally take a config with a list of custom converter types to be added.
2. `HyperionSerializer` may optionally take a config with 3 settings:
    - `perserve-object-references` used to track references (not only values) in serialized object graph, including cyclic references.
    - `version-tolerance` which can be turned off to reduce a payload manifest size.
    - `known-types-provider` which takes a type name for a class implementing `IKnownTypesProvider` - an interface that will be used at start of an actor system to provide a list of types, that are assumed to be known by all communicating sides using Hyperion serializer. This also can be used to optimize Hyperion usage.

This new feature has its separate test to check if HOCON configs are bound to right serializers when serialization map is being created.

### DIFF
--- a/src/contrib/serializers/Akka.Serialization.Hyperion.Tests/HyperionConfigTests.cs
+++ b/src/contrib/serializers/Akka.Serialization.Hyperion.Tests/HyperionConfigTests.cs
@@ -1,0 +1,125 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="HyperionConfigTests.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Akka.Actor;
+using Akka.Configuration;
+using Xunit;
+
+namespace Akka.Serialization.Hyperion.Tests
+{
+    public class HyperionConfigTests
+    {
+        [Fact]
+        public void Hyperion_serializer_should_have_correct_defaults()
+        {
+            var config = ConfigurationFactory.ParseString(@"
+                akka.actor {
+                    serializers.hyperion = ""Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion""
+                    serialization-bindings {
+                        ""System.Object"" = hyperion
+                    }
+                }
+            ");
+            using (var system = ActorSystem.Create(nameof(HyperionConfigTests), config))
+            {
+                var serializer = (HyperionSerializer)system.Serialization.FindSerializerForType(typeof(object));
+                Assert.Equal(true, serializer.Settings.VersionTolerance);
+                Assert.Equal(true, serializer.Settings.PreserveObjectReferences);
+                Assert.Equal("NoKnownTypes", serializer.Settings.KnownTypesProvider.Name);
+            }
+        }
+
+        [Fact]
+        public void Hyperion_serializer_should_allow_to_setup_custom_flags()
+        {
+            var config = ConfigurationFactory.ParseString(@"
+                akka.actor {
+                    serializers.hyperion = ""Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion""
+                    serialization-bindings {
+                        ""System.Object"" = hyperion
+                    }
+                    serialization-settings.hyperion {
+                        preserve-object-references = false
+                        version-tolerance = false
+                    }
+                }
+            ");
+            using (var system = ActorSystem.Create(nameof(HyperionConfigTests), config))
+            {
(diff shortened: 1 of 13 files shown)


rogeralsing:
Some background here, and I've verified this with your current codebase too.
Version tolerance is not implemented, there is some initial step towards it by storing type information in the payload, but nothing more. there is nothing on the deserializing side to make something out of this information.

By exposing the `version-tolerance` property, it might make people  believe that there is tolerance support.
I've seen some people talking about using Wire/Hyperion for persistence, which I've explicitly warned about in the readme.

src/contrib/serializers/Akka.Serialization.Hyperion/HyperionSerializer.cs
```diff
@@ -87,5 +110,53 @@ public override object FromBinary(byte[] bytes, Type type)
...
+        public static readonly HyperionSerializerSettings Default = new HyperionSerializerSettings(
+            preserveObjectReferences: true,
+            versionTolerance: true,
+            knownTypesProvider: typeof(NoKnownTypes));
+
+        public static HyperionSerializerSettings Create(Config config)
+        {
+            if (config == null) throw new ArgumentNullException(nameof(config), "HyperionSerializerSettings require a config, default path: `akka.serializers.hyperion`");
+
+            var type = Type.GetType(config.GetString("known-types-provider"), true);
```
Aaronontheweb: Need some specs to verify that all of these settings can be loaded by the Hyperion serializer from a valid HOCON file.

src/contrib/serializers/Akka.Serialization.Hyperion/reference.conf
```diff
@@ -0,0 +1,27 @@
+﻿akka.actor {
+	serializers {
+		hyperion = "Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion"
+	}
+	serialization-settings {
+		hyperion {
+			# If true, hyperion serializer will preserve references between objects. 
+			# This is necessary to support things such as cyclic dependencies between objects.
+			preserve-object-references = true
```
Aaronontheweb: Need specs to validate the defaults (protects against regressions where the defaults get changed by accident)

src/core/Akka/Serialization/NewtonSoftJsonSerializer.cs
```diff
@@ -54,10 +89,38 @@ public NewtonSoftJsonSerializer(ExtendedActorSystem system)
...
+                ContractResolver = new AkkaContractResolver()
+            };
+
+            _serializer = JsonSerializer.Create(_settings);
+        }
+
+        public NewtonSoftJsonSerializer(ExtendedActorSystem system, Config config)
+            : base(system)
+        {
+            var serializerSettings = NewtonSoftJsonSerializerSettings.Create(config);
```
Aaronontheweb: Need specs to validate that these settings can be correctly loaded from config

Aaronontheweb: Also: what are the default JSON.NET HOCON settings? Shouldn't these be expressed somewhere also?

Horusiath:
I've added config specs + 2 extra settings for `NewtonsoftJsonSerializer` (in case if it will be used as secondary serializer) + xml docs for settings.

src/contrib/serializers/Akka.Serialization.Hyperion.Tests/HyperionConfigTests.cs
```diff
@@ -0,0 +1,125 @@
...
+using Akka.Actor;
+using Akka.Configuration;
+using Xunit;
+
+namespace Akka.Serialization.Hyperion.Tests
+{
+    public class HyperionConfigTests
+    {
+        [Fact]
+        public void Hyperion_serializer_should_have_correct_defaults()
```
Aaronontheweb: Thank you @Horusiath. Looks great.

PR #2523, opened 2017-02-19
Cluster Sharding with remember-entity enabled fails to recover after restart

ismaelhamed:
We've been having difficulty getting back persistent actors (using Oracle) in cluster sharding on a cold start. I noticed some work has already been done in akka to mitigate it thru the usage of strategies when recovering, so this is a direct port of that work.

Original issue and discussion:
(https://github.com/akka/akka/issues/20744)

Original PR:
(https://github.com/akka/akka/pull/20745)


src/core/Akka/Pattern/FutureTimeoutSupport.cs
```diff
@@ -0,0 +1,40 @@
...
+    public static class FutureTimeoutSupport
+    {
+        /// <summary>
+        /// Returns a Task that will be completed with the success or failure of the provided value after the specified duration.
+        /// </summary>
+        /// <typeparam name="T">TBD</typeparam>
+        /// <param name="value">TBD</param>
+        /// <param name="timeout">TBD</param>
+        /// <param name="scheduler">TBD</param>
+        public static Task<T> After<T>(TimeSpan timeout, IScheduler scheduler, Func<Task<T>> value)
```
Horusiath: I'm not sure if we need to move this as a pattern. If this would be a potential patter it should probably be an extension method with signature like `Task<T> After<T>(this IScheduler scheduler, TimeSpan timeout, Func<Task<T>> call)`. But so far it's going to be used only in one place. I'd keep it encapsulated within entity recovery strategy.

src/contrib/cluster/Akka.Cluster.Sharding/EntityRecoveryStrategy.cs
```diff
@@ -0,0 +1,82 @@
...
+            this.frequency = frequency;
+            this.numberOfEntities = numberOfEntities;
+        }
+
+        public override IImmutableSet<Task<IImmutableSet<EntityId>>> RecoverEntities(IImmutableSet<EntityId> entities)
+        {
+            return entities.Grouped(numberOfEntities).Aggregate(
+                Tuple.Create(frequency, ImmutableHashSet<Task<IImmutableSet<EntityId>>>.Empty),
+                (accumulated, next) => Tuple.Create(accumulated.Item1 + frequency, accumulated.Item2.Add(ScheduleEntities(accumulated.Item1, next.ToImmutableHashSet())))
+            ).Item2;
```
Horusiath: I think, that this logic could be simplified. I know, it's direct reflection of what happens on the scala side, but we could aim for clarity here.  Example:

```csharp
var stamp = this.frequency;
var builder = ImmutableHashSet<Task<IImmutableSet<EntityId>>>.Empty.ToBuilder();
foreach (var bucket in entities.Grouped(numberOfEntities)) 
{
    var scheduled = ScheduleEntities(stamp, bucket.ToImmutableHashSet());
    builder.Add(scheduled);
    stamp += this.frequency;
}
return builder.ToImmutable();
```
I think, this is cleaner. It's also a lot less allocations.

Horusiath:
Great work! It's nice, that you've catch that up 👍 I've added a few remarks, but beside that it looks good.

src/contrib/cluster/Akka.Cluster.Sharding/EntityRecoveryStrategy.cs
ismaelhamed: Should I leave the Grouped extension method where it is right now? I couldn't find a obvious place to put it so I left it there for now.

src/Akka.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+VisualStudioVersion = 14.0.25420.1
```
Aaronontheweb: Not sure if it was this change that caused this, but the mono build failed to execute as a result of some of the changes in this PR.

```
[18:11:15]	[docker]   1) Building /checkout/src/Akka.sln failed with exitcode 1.
[18:11:15]	[docker]   2) CS1061: Transport/AkkaIO/AddressConverters.cs(24,86): Type `System.Net.IPAddress' does not contain a definition for `MapToIPv4' and no extension method `MapToIPv4' of type `System.Net.IPAddress' could be found. Are you missing an assembly reference?
[18:11:15]	[docker]   3) CS2001: CSC(0,0): Source file `Dsl/Flow.cs' could not be found
```

ismaelhamed: Ok, we don't need this anyway. I'll revert it back later when I address the rest of the remarks.

ismaelhamed:
Attended all remarks, but I still need help with one of the tests before you should consider merging it.

src/contrib/cluster/Akka.Cluster.Sharding.Tests/ConstantRateEntityRecoveryStrategySpec.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            strategy = EntityRecoveryStrategy.ConstantStrategy(Sys, TimeSpan.FromSeconds(1), 2);
+        }
+
+        [Fact]
+        public void ConstantRateEntityRecoveryStrategySpec_must_recover_entities()
+        {
+            var entities = ImmutableHashSet.Create<EntityId>("1", "2", "3", "4", "5");
+
+            // TODO: https://github.com/akka/akka/blob/master/akka-cluster-sharding/src/test/scala/akka/cluster/sharding/ConstantRateEntityRecoveryStrategySpec.scala
```
Horusiath: I cannot compile it rigth now, but I think, this scala code:

```scala
val startTime = System.nanoTime()
      val resultWithTimes = strategy.recoverEntities(entities).map(
        _.map(entityIds ⇒ (entityIds → (System.nanoTime() - startTime).nanos)))

      val result = Await.result(Future.sequence(resultWithTimes), 6.seconds)
        .toVector.sortBy { case (_, duration) ⇒ duration }
      result.size should ===(3)

      val scheduledEntities = result.map(_._1)
```

should be equivalent to this:

```csharp
var entities = ImmutableHashSet.Create<EntityId>("1", "2", "3", "4", "5");
var startTime = DateTime.UtcNow;
var resultWithTimes = strategy.RecoverEntities(entities)
    .Select(async bucketTask => new KeyValuePair<,>((await bucketTask), DateTime.UtcNow - startTime))
    .ToArray();

var result = Task.WaitAll(resultWithTimes).OrderBy(pair => pair.Value).ToArray();
result.Length.Should().Be(3);

var scheduledEntities = result.Select(pair => pair.Key).ToArray();
```

Aaronontheweb:
@ismaelhamed lol, funnily enough I added `FutureTimeoutSupport` back as a pattern since I needed it for `CoordinatedShutdown`

Aaronontheweb:
Still have a build failure on Mono:

```
Building /checkout/src/Akka.sln failed with exitcode 1.
[10:42:44]	Status:           Failure
[10:42:44]	  1) Building /checkout/src/Akka.sln failed with exitcode 1.
[10:42:44]	  2) : /usr/lib/mono/xbuild/14.0/bin/Microsoft.CSharp.targets(0,0): Compiler crashed with code: 1.
```

Side note: should definitely get an achievement or something for killing the compiler. That's a pretty epic error message lol.

ismaelhamed:
@Aaronontheweb I shall get another compiler, that actually knows how to compile

Aaronontheweb:
Looks like it's the Sharding.Test.MultiNode project that is causing the trouble:

```
[16:38:57]	[docker] /checkout/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj (Rebuild) ->
[16:38:57]	[docker] /usr/lib/mono/xbuild/14.0/bin/Microsoft.CSharp.targets (CoreCompile target) ->
[16:38:57]	[docker] 
[16:38:57]	[docker] 	/usr/lib/mono/xbuild/14.0/bin/Microsoft.CSharp.targets: error : Compiler crashed with code: 1.
```

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -56,6 +56,11 @@ public ClusterShardingSpecConfig()
                   shard-start-timeout = 5s
                   entity-restart-backoff = 1s
                   rebalance-interval = 2 s
+                  entity-recovery-strategy = ""all""
+                  entity-recovery-constant-rate-strategy {{
```
Aaronontheweb: Have an extra set of `{` here

Aaronontheweb:
@Horusiath I think @ismaelhamed has addressed all of your remarks

Aaronontheweb:
thanks for your contribution @ismaelhamed . We'll roll this into the 1.2 release.

Issue #2516, opened 2017-02-14
Akka.Cluster: Add CoordinatedShutdown


Aaronontheweb:
As part of `LeaveAsync` and some of the other PRs around graceful termination of cluster nodes and other services such as Cluster.Sharding that depend on Akka.Cluster, we should replicate what the JVM has done and include the `CoordinatedShutdown` class: https://github.com/akka/akka/issues/21537

PR #2267, opened 2016-08-24
Implemented ClusterSingletonRestartSpec

alexvaluyskiy:
Implemented https://github.com/akka/akka/issues/20942 and https://github.com/akka/akka/issues/21236
The parts of this changes were implemented here https://github.com/akkadotnet/akka.net/pull/2495

PR #2530, opened 2017-02-24
implemented specs and CoordinatedShutdown 

Aaronontheweb:
Have `CoordinatedShutdown` implemented and specs passing for #2516, but the following still need to be done:

* [x]  Implement CLR exit hooks, which causes `CoordinatedShutdown` to run when the process running the `ActorSystem` attempts to exit.
* [x] Need to add integration with `Akka.Cluster`
* [x] Need to add integration with `Akka.Cluster.Tools`
* [x] Need to add integration with `Akka.Cluster.Sharding`
* [x] Need to add getakka.net documentation page describing "graceful exit" scenarios stemming from this PR

That'll be a bit of work but I can start pulling this in gradually and accept comments and change requests while the work is still being done.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
@@ -227,6 +227,9 @@ namespace Akka.Cluster.Sharding
         private bool _loggedFullBufferWarning = false;
         private const int RetryCountThreshold = 5;
 
+        private readonly CoordinatedShutdown _coordShutdown = CoordinatedShutdown.Get(Context.System);
+        private readonly TaskCompletionSource<Done> _gracefulShutdownProgress = new TaskCompletionSource<Done>();
+
         /// <summary>
         /// TBD
         /// </summary>
@@ -253,6 +256,17 @@ namespace Akka.Cluster.Sharding
             MembersByAge = membersByAgeBuilder.ToImmutable();
 
             _retryTask = Context.System.Scheduler.ScheduleTellRepeatedlyCancelable(Settings.TunningParameters.RetryInterval, Settings.TunningParameters.RetryInterval, Self, Retry.Instance, Self);
+            SetupCoordinatedShutdown();
+        }
+
+        private void SetupCoordinatedShutdown()
+        {
+            var self = Self;
+            _coordShutdown.AddTask(CoordinatedShutdown.PhaseClusterShardingShutdownRegion, "region-shutdown", () =>
+            {
+                self.Tell(GracefulShutdown.Instance);
+                return _gracefulShutdownProgress.Task;
+            });
         }
 
         private ILoggingAdapter _log;
@@ -263,7 +277,7 @@ namespace Akka.Cluster.Sharding
         /// <summary>
         /// TBD
         /// </summary>
-        public bool GracefulShutdownInProgres { get; private set; }
+        public bool GracefulShutdownInProgress { get; private set; }
         /// <summary>
         /// TBD
         /// </summary>
@@ -294,21 +308,18 @@ namespace Akka.Cluster.Sharding
             }
         }
 
-        /// <summary>
-        /// TBD
-        /// </summary>
+        /// <inheritdoc cref="ActorBase.PreStart"/>
         protected override void PreStart()
         {
             Cluster.Subscribe(Self, new[] { typeof(ClusterEvent.IMemberEvent) });
         }
 
-        /// <summary>
-        /// TBD
-        /// </summary>
+        /// <inheritdoc cref="ActorBase.PostStop"/>
         protected override void PostStop()
         {
             base.PostStop();
             Cluster.Unsubscribe(Self);
+            _gracefulShutdownProgress.TrySetResult(Done.Instance);
             _retryTask.Cancel();
         }
 
@@ -339,11 +350,7 @@ namespace Akka.Cluster.Sharding
             }
         }
 
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="message">TBD</param>
-        /// <returns>TBD</returns>
+        /// <inheritdoc cref="ActorBase.Receive"/>
         protected override bool Receive(object message)
         {
             if (message is Terminated) HandleTerminated(message as Terminated);
@@ -507,7 +514,7 @@ namespace Akka.Cluster.Sharding
             else if (command is GracefulShutdown)
             {
                 Log.Debug("Starting graceful shutdown of region and all its shards");
-                GracefulShutdownInProgres = true;
+                GracefulShutdownInProgress = true;
                 SendGracefulShutdownToCoordinator();
                 TryCompleteGracefulShutdown();
             }
@@ -553,16 +560,28 @@ namespace Akka.Cluster.Sharding
             return Task.WhenAll(tasks);
         }
 
+        private List<ActorSelection> GracefulShutdownCoordinatorSelections
+        {
+            get
+            {
+                return
+                    MembersByAge.Take(2)
+                        .Select(m => Context.ActorSelection(new RootActorPath(m.Address) + CoordinatorPath))
+                        .ToList();
+            }
+        }
+
         private void TryCompleteGracefulShutdown()
         {
-            if(GracefulShutdownInProgres && Shards.Count == 0 && ShardBuffers.Count == 0)
+            if(GracefulShutdownInProgress && Shards.Count == 0 && ShardBuffers.Count == 0)
                 Context.Stop(Self);     // all shards have been rebalanced, complete graceful shutdown
         }
 
         private void SendGracefulShutdownToCoordinator()
         {
-            if (GracefulShutdownInProgres && _coordinator != null)
-                _coordinator.Tell(new PersistentShardCoordinator.GracefulShutdownRequest(Self));
+            if (GracefulShutdownInProgress)
+                GracefulShutdownCoordinatorSelections
+                    .ForEach(c => c.Tell(new PersistentShardCoordinator.GracefulShutdownRequest(Self)));
         }
 
         private void HandleCoordinatorMessage(PersistentShardCoordinator.ICoordinatorMessage message)
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestartSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestartSpec.cs
@@ -1,74 +1,57 @@
-﻿//-----------------------------------------------------------------------
-// <copyright file="ClusterSingletonRestartSpec.cs" company="Akka.NET Project">
-//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
-//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
-// </copyright>
-//-----------------------------------------------------------------------
-
+﻿using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Cluster.Tools.Singleton;
 using Akka.Configuration;
 using Akka.TestKit;
+using Akka.TestKit.TestActors;
 using FluentAssertions;
 using Xunit;
 
 namespace Akka.Cluster.Tools.Tests.Singleton
 {
-    public sealed class ClusterSingletonRestartSpec : AkkaSpec
+    public class ClusterSingletonRestartSpec : AkkaSpec
     {
-        public class EchoActor : ActorBase
-        {
-            protected override bool Receive(object message)
-            {
-                Sender.Tell(message);
-                return true;
-            }
+        private readonly ActorSystem sys1;
+        private readonly ActorSystem sys2;
+        private ActorSystem sys3 = null;
 
-            public static Props Props { get; } = Props.Create(() => new EchoActor());
-        }
-
-        private readonly ActorSystem _sys1;
-        private readonly ActorSystem _sys2;
-        private ActorSystem _sys3;
-
-        public ClusterSingletonRestartSpec() : base(GetConfig())
-        {
-            _sys1 = ActorSystem.Create(Sys.Name, Sys.Settings.Config);
-            _sys2 = ActorSystem.Create(Sys.Name, Sys.Settings.Config);
-            _sys3 = null;
-        }
-
-        public static Config GetConfig()
-        {
-            return ConfigurationFactory.ParseString(@"
+        public ClusterSingletonRestartSpec() : base(@"
               akka.loglevel = INFO
-              akka.actor.provider = ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""
-              akka.remote {{
-                dot-netty.tcp {{
+              akka.actor.provider = ""cluster""
+              akka.cluster.auto-down-unreachable-after = 2s
+              akka.remote {
+                dot-netty.tcp {
                   hostname = ""127.0.0.1""
                   port = 0
-                }}
-              }}
-            ");
+                }
+              }")
+        {
+            sys1 = ActorSystem.Create(Sys.Name, Sys.Settings.Config);
+            sys2 = ActorSystem.Create(Sys.Name, Sys.Settings.Config);
         }
 
-        private void Join(ActorSystem from, ActorSystem to)
+        public void Join(ActorSystem from, ActorSystem to)
         {
-            from.ActorOf(ClusterSingletonManager.Props(
-                singletonProps: EchoActor.Props,
-                terminationMessage: PoisonPill.Instance,
-                settings: ClusterSingletonManagerSettings.Create(from)),
-                name: "echo");
+            from.ActorOf(ClusterSingletonManager.Props(Echo.Props,
+                PoisonPill.Instance,
+                ClusterSingletonManagerSettings.Create(from)), "echo");
 
-            Within(10.Seconds(), () =>
+            Within(TimeSpan.FromSeconds(10), () =>
             {
                 AwaitAssert(() =>
                 {
                     Cluster.Get(from).Join(Cluster.Get(to).SelfAddress);
-                    Cluster.Get(from).State.Members.Select(c => c.UniqueAddress).Should().Contain(Cluster.Get(from).SelfUniqueAddress);
-                    Cluster.Get(from).State.Members.Select(x => x.Status).ToImmutableHashSet().Should().Equal(ImmutableHashSet.Create(MemberStatus.Up));
+                    Cluster.Get(from).State.Members.Select(x => x.UniqueAddress).Should().Contain(Cluster.Get(from).SelfUniqueAddress);
+                    Cluster.Get(from)
+                        .State.Members.Select(x => x.Status)
+                        .ToImmutableHashSet()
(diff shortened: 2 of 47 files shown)


Aaronontheweb:
One translation issue worth explaining before diving into the code behind it: detecting the current process exiting on the CLR. The JVM has the ability to add shutdown hooks directly to the Java Virtual Machine instance before it exits, which makes this relatively straightforward in Akka.

In .NET, we have to hook into the [`AppDomain.ProcessExit` event](https://msdn.microsoft.com/en-us/library/system.appdomain.processexit.aspx) and use that to try to run our shutdown sequence. That might cause some issues in environments where Akka.NET is running on medium or low trust potentially, but those are the exception and not the rule. Hooking into this event can be turned on or off via HOCON config, but what it does is causes all of the termination hooks to run and the `ActorSystem` to gracefully terminate when it receives a signal that the process it attempting to shutdown. It then has 2 seconds to execute before the process leaves for good.

On the flipside, we also have the ability to shut down the process when the `ActorSystem` is terminated, so when the `Terminate` task completes an `Environment.Exit(0)` call will be made. This setting can also be enabled via HOCON.

Goal of these settings: make it possible to terminate the process and shutdown the `ActorSystem` and vice-versa in a single call. Worth noting: `AppDomain.ProcessExit` won't pick up a process being killed via TaskManager or `CTRL+C` since that sends a different process kill signal than a normal exit.

I've verified that both of these exist in .NET Core, so this shouldn't cause portability issues there:
* [`AppDomain.ProcessExit` in netcore](https://github.com/dotnet/coreclr/blob/68f72dd2587c3365a9fe74d1991f93612c3bc62a/src/mscorlib/src/System/AppDomain.cs#L832)
* [`Environment.Exit` in netcore](https://github.com/dotnet/coreclr/blob/68f72dd2587c3365a9fe74d1991f93612c3bc62a/src/mscorlib/src/System/Environment.cs#L214)

Aaronontheweb:
Have all Akka.Cluster leave scenarios working correctly. Been able to verify with some of the additional specs ported from the JVM. Waiting to see if the MNTR can confirm.

Aaronontheweb:
Can confirm from the test logs thus far that there's still an issue where `MemberRemoved` isn't correctly published even after `MemberExited` and `MemberLeft` for that node have been published. Will get to the bottom of that as part of this PR.

Aaronontheweb:
Hah, found the issue I think:

```
[docker] [WARNING][02/28/2017 02:03:46][Thread 0057][akka://ClusterSpec2/user/testActor63] DeadLetter from [akka://ClusterSpec2/system/cluster/core/publisher#1794505030] to [akka://ClusterSpec2/user/testActor63#884883929]: <Received dead letter from [akka://ClusterSpec2/system/cluster/core/publisher#1794505030]: Akka.Cluster.ClusterEvent+MemberRemoved(Member=Member(address = akka.tcp://ClusterSpec2@0.0.0.0:37171, status = Removed, role=[], upNumber=1))>
```

Looks like the `MemberRemoved` message is being sent after all, although it's not delivered successfully before the shutdown process kills off all of the Cluster daemon actors. I'll do some more digging.

Aaronontheweb:
Yep, it was the `CoordinatedShutdown` using its new default settings and killing off the `ActorSystem` upon leaving the cluster, so a `TestProbe` declared inside that `ActorSystem` couldn't receive that node's own `MemberRemoved` event because it could possibly die first.

In the JVM they have the ability to override the reference HOCON settings for an entire test assembly, and they have these values passed in for all of the Akka.Cluster.* tests (but not the multi-node specs):

```
akka.coordinated-shutdown.run-by-jvm-shutdown-hook = off
 akka.coordinated-shutdown.terminate-actor-system = off
 akka.cluster.run-coordinated-shutdown-when-down = off
```

We'll need to turn these settings off in those test assemblies as well.

Aaronontheweb:
Had a bunch of Akka.Cluster multi-node tests fail as well on this most recent set of changes. I'll check to see if those defaults also need to be changed.

Aaronontheweb:
Nodes bailed and shutdown early upon leaving the cluster without letting the `TestConductor` know beforehand. I'll add a shim in the `Akka.Cluster.TestKit` configuration to turn off these settings there.

Aaronontheweb:
Got the MNTR suite passing. Just committed the integration bits for Akka.Cluster.Tools.Singleton. Have to add some more specs for it but I wanted to kick off an MNTR run now.

Aaronontheweb:
Cluster.Tools is finished. All of the changes were to Cluster.Singleton.

maxcherednik:
@Aaronontheweb I was just curious - do we really need to have this clr hook? A public method, which work should be fine I guess. Cause overall it might over-complicate things.

src/contrib/cluster/Akka.Cluster.Tools.Tests/packages.config
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="FluentAssertions" version="4.1.0" targetFramework="net45" />
+  <package id="FluentAssertions" version="3.3.0" targetFramework="net45" />
```
alexvaluyskiy: Why?

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestartSpec.cs
```diff
@@ -0,0 +1,155 @@
...
+using Akka.Cluster.Tools.Singleton;
+using Akka.Configuration;
+using Akka.TestKit;
+using Akka.TestKit.TestActors;
+using FluentAssertions;
+using Xunit;
+
+namespace Akka.Cluster.Tools.Tests.Singleton
+{
+    public class ClusterSingletonRestartSpec : AkkaSpec
```
alexvaluyskiy: This spec was implemented here https://github.com/akkadotnet/akka.net/pull/2267

src/contrib/cluster/Akka.Cluster.Tools.Tests/packages.config
Aaronontheweb: @alexvaluyskiy a mistake on my part by the looks of it

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestartSpec.cs
Aaronontheweb: @alexvaluyskiy doh, I totally missed that

Aaronontheweb:
@maxcherednik my thoughts on that were: the CLR hooks will help the node still gracefully exit even if the process was terminated without calling `Cluster.Leave` first.

Aaronontheweb:
I think the code changes here are done. Need to add documentation next.

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -294,21 +308,18 @@ protected object RegistrationMessage
             }
         }
 
-        /// <summary>
-        /// TBD
-        /// </summary>
+        /// <inheritdoc cref="ActorBase.PreStart"/>
```
alexvaluyskiy: FYI: inheritdoc does not work in docfx at the moment

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestartSpec.cs
```diff
@@ -1,74 +1,61 @@
-﻿//-----------------------------------------------------------------------
-// <copyright file="ClusterSingletonRestartSpec.cs" company="Akka.NET Project">
+﻿// <copyright file="ClusterSingletonRestartSpec.cs" company="Akka.NET Project">
```
alexvaluyskiy: Maybe you could exclude this file from the pull request?

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -612,8 +638,7 @@ private void GetNextOldestChange()
         {
             var newOldest = handOverTo?.Path.Address;
             Log.Info("Singleton terminated, hand-over done [{0} -> {1}]", _cluster.SelfAddress, newOldest);
-            handOverTo?.Tell(HandOverDone.Instance);
```
alexvaluyskiy: Why you have remove it?
https://github.com/akka/akka/blob/v2.5-M2/akka-cluster-tools/src/main/scala/akka/cluster/singleton/ClusterSingletonManager.scala#L756

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -488,6 +488,34 @@ namespace Akka.Actor
         public override int CompareTo(Akka.Actor.ActorPath other) { }
         public override Akka.Actor.ActorPath WithUid(long uid) { }
     }
+    public sealed class CoordinatedShutdown : Akka.Actor.IExtension
+    {
+        public const string PhaseActorSystemTerminate = "actor-system-terminate";
```
alexvaluyskiy: should we make these constants as public?

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -244,7 +244,7 @@ public void AddTask(string phase, string taskName, Func<Task<Done>> task)
         /// the Akka.NET internal shutdown hooks execute.
         /// </summary>
         /// <param name="hook">A task that will be executed during shutdown.</param>
-        public void AddClrShutdownHook(Func<Task<Done>> hook)
+        internal void AddClrShutdownHook(Func<Task<Done>> hook)
```
alexvaluyskiy: Why you made it as internal? It is a `public` in JVM

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -488,6 +488,34 @@ namespace Akka.Actor
...
+        public const string PhaseBeforeServiceUnbind = "before-service-unbind";
+        public const string PhaseClusterExiting = "cluster-exiting";
+        public const string PhaseClusterExitingDone = "cluster-exiting-done";
+        public const string PhaseClusterLeave = "cluster-leave";
+        public const string PhaseClusterShardingShutdownRegion = "cluster-sharding-shutdown-region";
+        public const string PhaseClusterShutdown = "cluster-shutdown";
+        public const string PhaseServiceRequestsDone = "service-requests-done";
+        public const string PhaseServiceStop = "service-stop";
+        public const string PhaseServiceUnbind = "service-unbind";
+        public CoordinatedShutdown(Akka.Actor.ExtendedActorSystem system, System.Collections.Generic.Dictionary<string, Akka.Actor.Phase> phases) { }
```
alexvaluyskiy: Should be internal

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1309,6 +1337,16 @@ namespace Akka.Actor
             public Akka.Util.ISurrogated FromSurrogate(Akka.Actor.ActorSystem system) { }
         }
     }
+    public sealed class Phase
```
alexvaluyskiy: Should be internal

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
alexvaluyskiy: FYI: `docfx` does not support `inheritdoc` well, only basic support:
https://github.com/dotnet/docfx/issues/247

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestartSpec.cs
alexvaluyskiy: Maybe you could exclude this file from the pull request?

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
alexvaluyskiy: Why you have remove it?
https://github.com/akka/akka/blob/v2.5-M2/akka-cluster-tools/src/main/scala/akka/cluster/singleton/ClusterSingletonManager.scala#L756

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
alexvaluyskiy: should we make these constants as internal or private?

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -0,0 +1,634 @@
...
+
+        /// <summary>
+        /// Add a shutdown hook that will execute when the CLR process begins
+        /// its shutdown sequence, invoked via <see cref="AppDomain.ProcessExit"/>.
+        /// 
+        /// Added hooks may run in any order concurrently, but they are run before
+        /// the Akka.NET internal shutdown hooks execute.
+        /// </summary>
+        /// <param name="hook">A task that will be executed during shutdown.</param>
+        internal void AddClrShutdownHook(Func<Task<Done>> hook)
```
alexvaluyskiy: Why you made it as internal? It is a public in JVM

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -5155,6 +5201,12 @@ namespace Akka.Util.Internal
         public static System.Text.StringBuilder AppendJoin<T>(this System.Text.StringBuilder sb, string separator, System.Collections.Generic.IEnumerable<T> values) { }
         public static System.Text.StringBuilder AppendJoin<T>(this System.Text.StringBuilder sb, string separator, System.Collections.Generic.IEnumerable<T> values, System.Action<System.Text.StringBuilder, T, int> valueAppender) { }
     }
+    public class static TaskEx
```
alexvaluyskiy: Let's make all internal extensions as internal.

src/core/Akka/Actor/CoordinatedShutdown.cs
Aaronontheweb: There was a discussion about this on Gitter... It seemed extraneous, to some extent, since the notion of a JvmShutdownHook is actually built into the Java VM itself, whereas I invented our own version here that doesn't necessarily have the same level of support out of the box. I thought it best to leave it as `internal` for now, see if there's demand for it / how reliable it is.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
Aaronontheweb: Merge mistake on my part when I rebased this PR after yours had been merged in. I'll undo it.

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
Aaronontheweb: Yeah, I'm hoping they'll pick it up. In the meantime this is less ugly than having all of those TBDs

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestartSpec.cs
Aaronontheweb: yeah, I don't need this as part of the PR anymore since you implemented it.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Aaronontheweb: Sounds good

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Aaronontheweb: Yep

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Aaronontheweb: Sounds good

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Aaronontheweb: The constants should be `public`, as they're useful when an end-user wants to append a task to a built-in phase. The constants in this case are all of the phase names.

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
alexvaluyskiy: Docfx supports only plain `<inheritdoc>` and it could inherit from the base class or from interface. But it could not inherit the documentation from the custom elements

Aaronontheweb:
Docs are here: https://github.com/akkadotnet/getakka.net/pull/235

Aaronontheweb:
I still need to implement @alexvaluyskiy's suggestions, and I've found a bug with the `AppDomain.ProcessExit` hook that affects applications that have multiple `ActorSystem`s in the same process... So I need to verify a fix for that first.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
alexvaluyskiy: We have already had a file `src/core/Akka/Util/Internal/TaskExtensions.cs`, we could use it for those extensions

Aaronontheweb: @alexvaluyskiy that class conflicts with the built-in `System.Threading.Tasks.TaskExtensions`, which required me to use a ton of namespace aliases everywhere. Coming up with a class name with no collisions with a built-in type was the better choice and I'd move any existing methods on the old `TaskExtensions` class into `TaskEx` in the future.

Aaronontheweb:
Ready for review once the perf specs pass

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -553,16 +560,28 @@ private void ReplyToRegionStatsQuery(IActorRef sender)
             return Task.WhenAll(tasks);
         }
 
+        private List<ActorSelection> GracefulShutdownCoordinatorSelections
```
Horusiath: I don't think, using property for work performing some computation + allocations is good choice. Either we should make it lazy/memoized if possible or turn into method if not.

alexvaluyskiy:
@Aaronontheweb I reviewed this PR a couple of days ago. Looks good to me

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
Aaronontheweb: Over thinking it here IMHO. The property allocates a new list because the `MembersByAge` value will be determined at the time this method is called, and it only gets called once during shutdown so allocations / performance isn't an issue. All it's really doing is hiding the implementation details for how the next best coordinators are likely to be picked.

Issue #2320, opened 2016-09-23
Akka.Persistence - fail loudly when the backend store connection cannot be established


eriklagerholm:
In its current state, it seems like the persistence framework doesn't bubble exceptions when opening connections to the backend store. This was encountered using the SqlServer adapter and when providing a bad connection string the exception generated when opening the sql connection is not caught by the actor system logger. This silent error results in persistent actors getting stuck in recovery mode, thus giving the impression of the system just stopping without any apparent reason.

By checking out the SqlServer adapter repo and debugging the code I was able to see that my bad connection string did indeed generate an exception when the adapter tried to open an sql connection, but somewhere the exception seems to be handled without any logging behaviour. As a user I would expect a fatal log entry, clearly stating what wrong.

This might be unique to the sql framework or the sqlserver adapter specifically. I have not had the ability to test this using other adapters.

I've attached some screenshots to more specifically describe the flow of events.

_The occurring exception while debugging the sql adapter:_

![exception being thrown](https://cloud.githubusercontent.com/assets/7983352/18777525/e92f438e-816f-11e6-9b90-57dc37cb1c98.png)

_The behaviour of the logging with the coordinator actor seemingly stuck in recovery_

![logging from the actor system](https://cloud.githubusercontent.com/assets/7983352/18777528/ee48f1b2-816f-11e6-8e3d-e2b2ae2c6e10.png)

malclear:
This has been open for a while. Are there any plans to address this?

malclear:
On further review, it seems that this has been addressed. 
Can the issue be closed?

PR #2536, opened 2017-03-02
Add logging behavior to Akka.Persistence SqlJournal to fail loudly when backend store connection cannot be established

heynickc:
Per Issue #2320, added try/catch to `SqlJournal.CreateDbConnection()` to catch and log failures in overridden calls to `CreateDbConnection(string connectionString)` by Akka.Persistence plugins.  Added new spec called `SqlJournalConnectionFailureSpec` that should be implemented by plugins to ensure that they don't swallow exceptions during establishing connection to Journal store.  Implemented said spec in Akka.Persistence.Sqlite.Tests as an example.  Also added missing xunit dependencies to Akka.Persistence.Sql.Testkit (i.e. xunit, xunit.core, etc)

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
@@ -282,19 +282,26 @@ namespace Akka.Persistence.Sql.Common.Journal
                 .WasHandled;
         }
 
-        private async Task<AllPersistenceIds> Initialize()
+        private async Task<object> Initialize()
         {
-            using (var connection = CreateDbConnection())
+            try
             {
-                await connection.OpenAsync();
-
-                if (_settings.AutoInitialize)
+                using (var connection = CreateDbConnection())
                 {
-                    await QueryExecutor.CreateTablesAsync(connection, _pendingRequestsCancellation.Token);
-                }
+                    await connection.OpenAsync();
 
-                var ids = await QueryExecutor.SelectAllPersistenceIdsAsync(connection, _pendingRequestsCancellation.Token);
-                return new AllPersistenceIds(ids);
+                    if (_settings.AutoInitialize)
+                    {
+                        await QueryExecutor.CreateTablesAsync(connection, _pendingRequestsCancellation.Token);
+                    }
+
+                    var ids = await QueryExecutor.SelectAllPersistenceIdsAsync(connection, _pendingRequestsCancellation.Token);
+                    return new AllPersistenceIds(ids);
+                }
+            }
+            catch (Exception e)
+            {
+                return new Failure { Exception = e };
             }
         }
 
@@ -304,16 +311,8 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// <returns>TBD</returns>
         public DbConnection CreateDbConnection()
         {
-            try
-            {
-                var connectionString = GetConnectionString();
-                return CreateDbConnection(connectionString);
-            }
-            catch (Exception e)
-            {
-                Log.Error(e, "Failure creating DbConnection");
-                throw;
-            }
+            var connectionString = GetConnectionString();
+            return CreateDbConnection(connectionString);
         }
 
         /// <summary>
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
@@ -301,7 +301,7 @@ namespace Akka.Persistence.Sql.Common.Journal
             }
             catch (Exception e)
             {
-                return new Failure { Exception = e };
+                return new Failure {Exception = e};
             }
         }
 
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
@@ -304,8 +304,16 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// <returns>TBD</returns>
         public DbConnection CreateDbConnection()
         {
-            var connectionString = GetConnectionString();
-            return CreateDbConnection(connectionString);
+            try
+            {
+                var connectionString = GetConnectionString();
+                return CreateDbConnection(connectionString);
+            }
+            catch (Exception e)
+            {
+                Log.Error(e, "Failure creating DbConnection");
+                throw;
+            }
         }
 
         /// <summary>
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
@@ -105,13 +105,20 @@ namespace Akka.Persistence.Sql.Common.Snapshot
             _pendingRequestsCancellation.Cancel();
         }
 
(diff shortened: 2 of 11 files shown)


Horusiath:
What is the current behavior? Any exception that happens inside an actor is (or at least should be) logged by default anyway. No special try catch included. If this doesn't happen, it may be a bug somewhere higher in supervision chain.

heynickc:
Thanks @Horusiath.  Yea, logging isn't the right approach then.  The default behavior is that exceptions thrown in a particular plugin's overridden `CreateDbConnection(string connectionString)` method don't bubble up to the Persistent Actor.  This test fails when I force `SqliteJournal.CreateDbConnection` to throw a `SQLiteException`:

```
[Fact]
public void Persistent_actor_should_throw_exception_upon_connection_failure()
{
    EventFilter.Exception<Exception>().ExpectOne(() =>
    {
        var perf = Sys.ActorOf(Props.Create(() => new ReceiveAnyPersistentActor("test-persistent-actor")));
        perf.Tell("save");
    });

    ExpectNoMsg();
}
```
SqliteJournal.cs
```
protected override DbConnection CreateDbConnection(string connectionString)
{
    //return new SQLiteConnection(connectionString);
    throw new SQLiteException();
}
```
I'll continue working on it and revise accordingly.  Thanks.

heynickc:
Updated solution for `SqlJournal`.  Rather than treating the symptom of logs not reflecting `DbConnection ` errors, treated root cause of such exceptions not bubbling up to `PersistenceActor` during initialization.  Still need to apply this to `SqlSnapshot` which has a similar thing going on.

heynickc:
Swallowed DbConnection initialization failure fix has now also been applied to `SqlSnapshotStore` per latest commit.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
```diff
@@ -282,19 +282,26 @@ protected bool WaitingForInitialization(object message)
...
+                        await QueryExecutor.CreateTablesAsync(connection, _pendingRequestsCancellation.Token);
+                    }
+
+                    var ids = await QueryExecutor.SelectAllPersistenceIdsAsync(connection, _pendingRequestsCancellation.Token);
+                    return new AllPersistenceIds(ids);
+                }
+            }
+            catch (Exception e)
+            {
+                return new Failure {Exception = e};
```
heynickc: The result of the `Initialize()` method is passed via `PipeTo()` to `WaitForInitialization(object message)`.  Previous behavior was that it would only pass the result `AllPersistenceIds(ids)`, regardless of the success or failure of the previous calls.  This change adds handling of any possible exceptions and population of a `Failure` message that is passed to `WaitForInitialization`,  which handles the exception and logs it appropriately.

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
```diff
@@ -105,13 +105,20 @@ protected override void PostStop()
...
+                using (var connection = CreateDbConnection())
+                {
+                    await connection.OpenAsync(_pendingRequestsCancellation.Token);
+                    await QueryExecutor.CreateTableAsync(connection, _pendingRequestsCancellation.Token);
+                    return Initialized.Instance;
+                }
+            }
+            catch (Exception e)
+            {
+                return new Failure {Exception = e};
```
heynickc: Similar to the change made to `SqlJournal`, this adds handling of exceptions during `Initialize` and passing a `Failure` message to `WaitForInitialization` so it can be handled and logged.

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
```diff
@@ -123,7 +130,7 @@ private bool WaitingForInitialization(object message) => message.Match()
             })
             .With<Failure>(failure =>
             {
-                _log.Error(failure.Exception, "Error during snapshot store initialization");
+                Log.Error(failure.Exception, "Error during snapshot store initialization");
```
heynickc: This was initially calling `_log`, which may or may not be null at the time.  Changed to `Log` which will instantiate a default if `_log` is null.

src/contrib/persistence/Akka.Persistence.Sql.TestKit/SqlJournalConnectionFailureSpec.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public SqlJournalConnectionFailureSpec(Config config = null, ITestOutputHelper output = null)
+            : base(config)
+        {
+        }
+
+        [Fact]
+        public void Persistent_actor_should_throw_exception_upon_connection_failure()
+        {
+            EventFilter.Exception<Exception>().ExpectOne(() =>
```
heynickc: The test expects one `Exception` to be thrown when the `PersistenceActor` is sent a message that triggers `Persist`.  Since the `PersistenceActor` has been given a bad connection string, it should throw an exception that it cannot initialize because it can't create a new `DbConnection`

src/contrib/persistence/Akka.Persistence.Sql.TestKit/SqlSnapshotConnectionFailureSpec.cs
```diff
@@ -0,0 +1,94 @@
...
+        protected static readonly string DefaultInvalidConnectionString = "INVALID_CONNECTION_STRING";
+
+        public SqlSnapshotConnectionFailureSpec(Config config = null, ITestOutputHelper output = null) : base(config)
+        {
+        }
+
+        [Fact]
+        public void Persistent_actor_should_throw_exception_upon_connection_failure_when_saving_snapshot()
+        {
+            EventFilter.Exception<Exception>().ExpectOne(() =>
```
heynickc: This test expects an exception to be thrown when a `PersistenceActor` is told to take a snapshot, but the actor cannot initialize its `SqlSnapshotStore` because it's been given a bad connection string.

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/SqliteJournalConnectionFailureSpec.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Akka.Configuration;
+using Akka.Persistence.Sql.TestKit;
+using Akka.Util.Internal;
+using Xunit.Abstractions;
+
+namespace Akka.Persistence.Sqlite.Tests
+{
+    public class SqliteJournalConnectionFailureSpec : SqlJournalConnectionFailureSpec
```
heynickc: An example of inheriting the base `SqlJournalConnectionFailureSpec` to make sure a `SqliteJournal` given a bad connection string will throw an error clearly to the user

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
Aaronontheweb: Looks good

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
Aaronontheweb: Good catch.

Aaronontheweb:
@heynickc to close out the original issue we'll need to update the `Akka.Persistence.SqlServer` plugin with this change.

PR #2537, opened 2017-03-04
Improved throughput specs

alexvaluyskiy:
I think, the throughput specs should be improved. We should not only send messages, but also match them.  I've added a couple of match handlers to both specs. Also I've added a third spec - ActorBase + Akka.PatternMatch

## UntypedActor
Gen0: 157.46 collections
Gen1: 65.38 collections
Gen2: 3.77 collections
Average operations: 4,354,511.12 operations per second
TotalBytesAllocated: 15,188,140.31 bytes

## ReceivedActor
Gen0: 160.31 collections
Gen1: 71.08 collections
Gen2: 7.77 collections
Average operations: 4,136,675.05 operations per second
TotalBytesAllocated: 19,469,942.15 bytes

## ActorBase + Akka.PatternMatch
Gen0: 1,534.77 collections
Gen1: 174.62 collections
Gen2: 5.77 collections
Average operations: 2,606,221.09 operations per second
TotalBytesAllocated: 15,284,980.31 bytes

And I have a bit different numbers about allocated bytes on my local PC
## UntypedActor
TotalBytesAllocated: 5 176 788,31 bytes

## ReceivedActor
TotalBytesAllocated: 5 324 963,69 bytes

## ActorBase + Akka.PatternMatch
TotalBytesAllocated: 6 512 699,08 bytes

Locally UntypedActor and ReceiveActor allocate almost the same amount of bytes, but PatternMatch - much more. But on the build machine, PatternMatch and UntypedActor - almost the same, but ReceiveActor - much more


alexvaluyskiy:
@Aaronontheweb what do you think?

PR #2540, opened 2017-03-05
Update AutoDown Api

alexvaluyskiy:
`AutoDown` class should be `public` not `internal`


Horusiath:
Should we really expose it? It's configurable, but I don't expect someone will invoke it directly from the code.

alexvaluyskiy:
@Horusiath I just want to make the API consistent with Scala version.
https://github.com/akka/akka/blob/master/akka-cluster/src/main/scala/akka/cluster/AutoDown.scala#L29

PR #2541, opened 2017-03-05
Deprecate and remove Xunit1 Testkit project

alexvaluyskiy:
XUnit2 was released two years ago, I think it is a time to remove the package for Xunit1


Aaronontheweb:
Don't we need to rename the XUnit2 nuget package then?

Aaronontheweb:
errr, let me rephrase both the XUnit 1 and XUnit2 packages should now have the same content as the XUnit2 one, correct?

alexvaluyskiy:
@Aaronontheweb the last version of XUnit 1.x was released almost 4 years ago. Why we should support it in 2017?

Aaronontheweb:
@alexvaluyskiy what I meant was: just have both packages start serving up XUnit2 instead. So effectively make the Akka.TestKit.XUnit project serve up the same material as XUnit2. Make sense?

PR #2543, opened 2017-03-06
[WIP] Removed Mono workarounds

alexvaluyskiy:
- Removed MonoConcurrentQueue


src/core/Akka/Util/MonotonicClock.cs
```diff
@@ -19,9 +19,6 @@ internal static class MonotonicClock
     {
         private static readonly Stopwatch Stopwatch = Stopwatch.StartNew();
 
-        [DllImport("kernel32")]
-        private static extern ulong GetTickCount64();
```
Aaronontheweb: @alexvaluyskiy why was this removed?

alexvaluyskiy: We have already discussed it few months ago. I can't use it in NetCore, and you recommended to remove it

Aaronontheweb: ah right. That's what I thought.

Issue #2544, opened 2017-03-06
Too many allocations of int[] in Config.GetNode


alexvaluyskiy:
Akka.NET calls `Mailboxes.GetMailboxType` when it tries to create a new actor. And this methods calls `Config.GetNode` every time. However the method produces a lot of allocations, because it is using `string.Split` here 
https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Util/Internal/Extensions.cs#L73

![configgetnode](https://cloud.githubusercontent.com/assets/4232072/23626103/545d1d64-02b4-11e7-8b2c-e0395469c43e.jpg)

I've used Skynet benchmark to profiling: https://gist.github.com/rogeralsing/0b7a923786e0dc6e54c5c82955133c75

alexvaluyskiy:
Also Hocon extensively using LINQ and making a lot of allocations, like 
![hoconvalue concatstring](https://cloud.githubusercontent.com/assets/4232072/23626800/ca157efa-02b6-11e7-9934-777bf46d7ded.jpg)

I think we should avoid LINQ here

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/pull/2549

PR #2547, opened 2017-03-06
Rewrited SplitDottedPathHonouringQuotes using indexof and substr

Rattenkrieg:
Fix for #2544
Should reduce allocations.
Can run benchmarks for exact numbers a bit later.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -5111,7 +5111,6 @@ namespace Akka.Util.Internal
     {
         public static System.Collections.Generic.IDictionary<TKey, TValue> AddAndReturn<TKey, TValue>(this System.Collections.Generic.IDictionary<TKey, TValue> hash, TKey key, TValue value) { }
         public static void AddOrSet<TKey, TValue>(this System.Collections.Generic.IDictionary<TKey, TValue> hash, TKey key, TValue value) { }
-        public static System.Collections.Generic.IEnumerable<TOut> AlternateSelectMany<TIn, TOut>(this System.Collections.Generic.IEnumerable<TIn> self, System.Func<TIn, System.Collections.Generic.IEnumerable<TOut>> evenSelector, System.Func<TIn, System.Collections.Generic.IEnumerable<TOut>> oddSelector) { }
         public static T AsInstanceOf<T>(this object self) { }
         public static string BetweenDoubleQuotes(this string self) { }
         public static System.Collections.Generic.IEnumerable<T> Concat<T>(this System.Collections.Generic.IEnumerable<T> enumerable, T item) { }
@@ -5123,6 +5122,7 @@ namespace Akka.Util.Internal
         public static string Join(this System.Collections.Generic.IEnumerable<string> self, string separator) { }
         public static System.TimeSpan Max(this System.TimeSpan @this, System.TimeSpan other) { }
         public static System.TimeSpan Min(this System.TimeSpan @this, System.TimeSpan other) { }
+        [System.Runtime.CompilerServices.IteratorStateMachineAttribute(typeof(Akka.Util.Internal.Extensions.<SplitDottedPathHonouringQuotes>d__3))]
         public static System.Collections.Generic.IEnumerable<string> SplitDottedPathHonouringQuotes(this string path) { }
         public static System.Collections.Generic.IEnumerable<T> TakeRight<T>(this System.Collections.Generic.IEnumerable<T> self, int n) { }
     }
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -2060,7 +2060,6 @@ namespace Akka.Configuration
         public virtual bool IsEmpty { get; }
         public Akka.Configuration.Hocon.HoconValue Root { get; }
         public System.Collections.Generic.IEnumerable<Akka.Configuration.Hocon.HoconSubstitution> Substitutions { get; set; }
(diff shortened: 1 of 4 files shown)


src/core/Akka/Util/Internal/Extensions.cs
```diff
@@ -72,10 +72,30 @@ public static T Head<T>(this IEnumerable<T> self)
         /// <returns>The path elements</returns>
         public static IEnumerable<string> SplitDottedPathHonouringQuotes(this string path)
         {
-            return path.Split('\"')
-                .AlternateSelectMany(
```
Horusiath: I think, we could get rid of `AlternateSelectMany` from the code base.

alexvaluyskiy:
Need performance results

rogeralsing:
I beleive this does not take HOCON escape codes into account?
https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Configuration/Hocon/HoconTokenizer.cs#L157

src/core/Akka/Util/Internal/Extensions.cs
```diff
@@ -72,10 +72,30 @@ public static T Head<T>(this IEnumerable<T> self)
...
-                    outsideQuote => outsideQuote.Split(new[] { '.' }, StringSplitOptions.RemoveEmptyEntries),
-                    insideQuote => new[] { insideQuote });
+            var i = 0;
+            var j = 0;
+            var result = new List<string>(8);
```
alexvaluyskiy: Why 8? Most of the HOCON string contains 3-4 elements

rogeralsing:
IMO. the string parsing per request should go away.
If possible, the `Config` class could contains a concurrent dict that just takes the entire string as a key to return the hocon object.
if missing, THEN it could fallback to this string parsing.

That would probably have a drastic effect to the cost of spawning actors.

(There is ofcourse the issue fallback configs etc. but it should be possible to reset such cache when a new  fallback is introduced)

Aaronontheweb:
> If possible, the Config class could contains a concurrent dict that just takes the entire string as a key to return the hocon object. if missing, THEN it could fallback to this string parsing.

We have this capability in the `CachingConfig` and it gets used with our dispatchers and other parts of the actor creation process. Turns out we weren't using it in a key area where it's needed for mailboxes though. Addressed that here: https://github.com/akkadotnet/akka.net/pull/2549

Rattenkrieg:
@alexvaluyskiy here comes profiling results.
First, original version:

default

1. `int[]` allocations caused by `String.Split` as in your report
2. Closures
3. Actual payload

50f6cb4 version:

default

1. `string[]` behind the `List<string>`
2. `List<string>` itself

50f6cb4 version with yield-based coroutines:

default

Just iterator.

I ran benchmarks as well (Skynet via BenchmarkDotNet). And there is observable (even in Skynet scenario) improvement over current linq version. Yield-iterator performs a little better than list.

src/core/Akka/Util/Internal/Extensions.cs
Rattenkrieg: I'm just liking that digit. ;) But can change to 4 according to your heuristics.

src/core/Akka/Util/Internal/Extensions.cs
Rattenkrieg: Will move to tests since there is dependence in test-oracle scenario now. Just was hesitating to remove public method.

Rattenkrieg:
@rogeralsing 
> I beleive this does not take HOCON escape codes into account?

Right. Nor my proposed replacement neither current linq version cares about escape codes.
Philosophically speaking we working with user input (_our own input_) here, like [this one](https://github.com/akkadotnet/akka.net/blob/7166307bed5f87e1137518b60d82e57696110bc2/src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs#L155)
So should we care about potential misspelling or even intentional attack?
Lightbend guys do care, so they [dealing](https://github.com/typesafehub/config/blob/f6680a5dad51d992139d45a84fad734f1778bf50/config/src/main/java/com/typesafe/config/impl/PathParser.java#L51) with such strings same way they dealing with entire config - parsing it with same routines they parsing configs. But they have self-contained hocon project...

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -5137,6 +5136,7 @@ namespace Akka.Util.Internal
         public static string Join(this System.Collections.Generic.IEnumerable<string> self, string separator) { }
         public static System.TimeSpan Max(this System.TimeSpan @this, System.TimeSpan other) { }
         public static System.TimeSpan Min(this System.TimeSpan @this, System.TimeSpan other) { }
+        [System.Runtime.CompilerServices.IteratorStateMachineAttribute(typeof(Akka.Util.Internal.Extensions.<SplitDottedPathHonouringQuotes>d__3))]
```
Rattenkrieg: Seems like a bug in approver 🙈

Rattenkrieg:
@rogeralsing @Aaronontheweb 
> If possible, the Config class could contains a concurrent dict that just takes the entire string as a key to return the hocon object.

I had proposed something similar [here](https://github.com/akkadotnet/akka.net/issues/2404#issuecomment-266321385)
The idea is to bootstrap everything upfront in single dictionary, with such design we don't need any (concurrent) cache.

Aaronontheweb:
@Rattenkrieg reason why we don't do that today is that most of the HOCON values don't get used more than once after they're read into a settings object. Materializing an unknowably large object into memory upfront isn't a great trade. The `CachingConfig` is used for the frequently looked up values like dispatchers and mailboxes, and it uses a `ConcurrentDictionary` because that's done lazily. The frequent HOCON lookups are meant to be the exception, not the rule. In the case of the `Deployer` and speeding that up, optimizing the "no hit" path where no matching deployment is found is a better use of time there.

PR #2550, opened 2017-03-07
Fix deadlock when calling ActorSelection.ResolveOne from GUI context

0x53A:
The first test passes, the second test blocks indefinitely (even though I passed a timeout!).

This is probably an issue with a missing ``.ConfigureAwait(false)``.

``SynchronizationContextHelper`` was copied from here: http://stackoverflow.com/questions/14087257/how-to-add-synchronization-context-to-async-test-method

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -3796,8 +3796,8 @@ namespace Akka.Pattern
         public Akka.Pattern.CircuitBreaker OnClose(System.Action callback) { }
         public Akka.Pattern.CircuitBreaker OnHalfOpen(System.Action callback) { }
         public Akka.Pattern.CircuitBreaker OnOpen(System.Action callback) { }
-        public async System.Threading.Tasks.Task<T> WithCircuitBreaker<T>(System.Func<System.Threading.Tasks.Task<T>> body) { }
-        public async System.Threading.Tasks.Task WithCircuitBreaker(System.Func<System.Threading.Tasks.Task> body) { }
+        public System.Threading.Tasks.Task<T> WithCircuitBreaker<T>(System.Func<System.Threading.Tasks.Task<T>> body) { }
+        public System.Threading.Tasks.Task WithCircuitBreaker(System.Func<System.Threading.Tasks.Task> body) { }
         public void WithSyncCircuitBreaker(System.Action body) { }
         public T WithSyncCircuitBreaker<T>(System.Func<T> body) { }
     }
--- a/build.fsx
+++ b/build.fsx
@@ -364,7 +364,7 @@ Target "NBench" <| fun _ ->
         let result = ExecProcess(fun info -> 
             info.FileName <- nbenchTestPath
             info.WorkingDirectory <- (Path.GetDirectoryName (FullName nbenchTestPath))
-            info.Arguments <- args) (System.TimeSpan.FromMinutes 25.0) (* Reasonably long-running task. *)
+            info.Arguments <- args) (System.TimeSpan.FromMinutes 40.0) (* Reasonably long-running task. *)
         if result <> 0 then failwithf "NBench.Runner failed. %s %s" nbenchTestPath args
     
     nbenchTestAssemblies |> Seq.iter (runNBench)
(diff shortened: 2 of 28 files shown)


0x53A:
Just adding this one ConfigureAwait fixes this specific case, but I went through and added it to the whole Akka lib (excluding a few contrib and tests).

IMO this should be everywhere, and adding an await without ConfigureAwait should fail the build ...

0x53A:
The test failed in CI (after my fix) with
image

So i guess I would need to reimplement the scheduler instead of reusing the WPF bits ...

0x53A:
Nice ...
image

0x53A:
Eh, I have set the test to skipped ... should hopefully become green now

0x53A:
@Aaronontheweb I rebased it on dev, and the current failure is
image

which also fails in my other PR.

I assume the CI is just slightly flaky?

Danthar:
Not in this case. The `Akka.Tests.Performance.Actor.Pattern.AskSpec+AskThroughput` test times out. Maybe a deadlock due to your latest changes ?

alexvaluyskiy:
@Danthar No, we have problems with Perf tests in all pull requests

Danthar:
@alexvaluyskiy But nbench tests should be reported normally in TC right? In this case it times out on invoking the nbench runner itself. 
Considering other tests run fine, this leads me to believe there is something going on with the spec its running. ?

0x53A:
The failing part seems to be ``Akka.Tests.Performance.Actor.Pattern.AskSpec+AskThroughput`` (http://petabridge-ci.cloudapp.net/viewLog.html?buildId=22768&buildTypeId=AkkaNet_AkkaNetWindowsPerformanceTests&tab=buildLog#_state=15933&focus=15933), which does sound suspicious. I will try to test it locally, but can only do that in the evening.

0x53A:
The passing test run from here: http://petabridge-ci.cloudapp.net/viewLog.html?buildId=22829&buildTypeId=AkkaNet_AkkaNetWindowsPerformanceTests&tab=buildLog#_state=7528&focus=14984
image

It ran between __[21:48:53]__ and __[22:13:20]__ , so took __24:27__ minutes. So the timeout of 25 minutes is maybe slightly optimistic.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -3796,8 +3796,8 @@ namespace Akka.Pattern
         public Akka.Pattern.CircuitBreaker OnClose(System.Action callback) { }
         public Akka.Pattern.CircuitBreaker OnHalfOpen(System.Action callback) { }
         public Akka.Pattern.CircuitBreaker OnOpen(System.Action callback) { }
-        public async System.Threading.Tasks.Task<T> WithCircuitBreaker<T>(System.Func<System.Threading.Tasks.Task<T>> body) { }
-        public async System.Threading.Tasks.Task WithCircuitBreaker(System.Func<System.Threading.Tasks.Task> body) { }
+        public System.Threading.Tasks.Task<T> WithCircuitBreaker<T>(System.Func<System.Threading.Tasks.Task<T>> body) { }
```
Aaronontheweb: cc @alexvaluyskiy I get that this is an API change, but would removing the `async` keyword here result in a break for something that was `await`-ing on this task? Or is this just goofery from the API printer?

0x53A:
@Aaronontheweb is it possible to compare two runs of nbench? This does touch a lot of places, so in theory i would need to check every performance spec.
Otherwise i could revert the last commit so that only the one await is changed, reducing the diff.

Thank you.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Danthar: async is used as a function modifier to avoid breaking existing code that use await as a variable (and some other reasons). So basically backwards compat reason on a c# feature level. 
Removing or adding it does not influence upstream dependencies at all as far as i know.

Aaronontheweb: thanks @Danthar

Aaronontheweb:
@0x53A this looks good to me, but the do we need to add a license header / third party notice file for the Nito source code? cc @StephenCleary

0x53A:
Just to be clear, the Nito source is only used in the unit test project. I would have used the nuget, if possible ...

StephenCleary:
It's all MIT-licensed. No problems.

Aaronontheweb:
thanks @0x53A - I'm satisfied and it sounds like @StephenCleary is too. Thanks for all of your hard work. I'll merge this in once CI runs. I checked the perf numbers and everything looked fine there.

Issue #2539, opened 2017-03-05
Too many allocation in ActorPath.Elements


alexvaluyskiy:
I think we should cache ActorPath.Elements, because it could be called many and many times, and this getter produces a lot of allocation and garbage pressure.

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/ActorPath.cs#L200
Memory profiling
![actorpath](https://cloud.githubusercontent.com/assets/4232072/23624392/30cb1c1c-02ae-11e7-8352-2fbdafd6e1fa.jpg)

I've used Skynet benchmark to profiling: https://gist.github.com/rogeralsing/0b7a923786e0dc6e54c5c82955133c75

alexvaluyskiy:
Maybe we could use Lazy in ActorPath?
csharp
```
        protected ActorPath(ActorPath parentPath, string name, long uid)
        {
             ...
            _elements = new FastLazy<IReadOnlyList<string>>(FillElements);
        }

        private readonly FastLazy<IReadOnlyList<string>> _elements;
        private IReadOnlyList<string> FillElements()
        {
            var current = this;
            var elements = new List<string>();
            while (!(current is RootActorPath))
            {
                elements.Add(current.Name);
                current = current.Parent;
            }
            elements.Reverse();
            return elements.AsReadOnly();
        }

        public IReadOnlyList<string> Elements => _elements.Value;
```

alexvaluyskiy:
I've checked again the memory consumption. Deployer.Lookup still produces a lot of memory traffic
![updateddeployerlookup](https://cloud.githubusercontent.com/assets/4232072/23742442/6efbc14a-04b5-11e7-9590-50b05b5495c1.jpg)

PR #2554, opened 2017-03-12
ActorPath.Elements optimization

Rattenkrieg:
Fix for #2539
Based on @alexvaluyskiy idea.
The trick is reusing `Parent`'s cached elements.

Can't compare two snapshots in dotMemory, hence two different screenshots.

Baseline:
default

Optimization:
default

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -166,7 +166,6 @@ namespace Akka.Actor
         protected ActorPath(Akka.Actor.ActorPath parentPath, string name, long uid) { }
         public Akka.Actor.Address Address { get; }
         public System.Collections.Generic.IReadOnlyList<string> Elements { get; }
-        public System.Collections.Generic.IReadOnlyList<string> ElementsWithUid { get; }
         public string Name { get; }
         public abstract Akka.Actor.ActorPath Parent { get; }
         public abstract Akka.Actor.ActorPath Root { get; }
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -4902,6 +4902,12 @@ namespace Akka.Util
         public bool IsValueCreated { get; }
         public T Value { get; }
     }
+    public sealed class FastLazy<S, T>
+    {
+        public FastLazy(System.Func<S, T> producer, S state) { }
+        public bool IsValueCreated { get; }
+        public T Value { get; }
+    }
     public interface ILinearSeq<out T> : System.Collections.Generic.IEnumerable<T>, System.Collections.IEnumerable
     {
         T Head { get; }
--- a/src/core/Akka/Actor/ActorPath.cs
+++ b/src/core/Akka/Actor/ActorPath.cs
@@ -142,19 +142,19 @@ namespace Akka.Actor
             return !s.StartsWith("$") && Validate(s.ToCharArray(), s.Length);
         }
 
+        private static bool IsValidChar(char c) => (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || ValidSymbols.Contains(c);
+        private static bool IsHexChar(char c) => (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F') || (c >= '0' && c <= '9');
+
         private static bool Validate(IReadOnlyList<char> chars, int len)
         {
-            Func<char, bool> isValidChar = c => (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || ValidSymbols.Contains(c);
-            Func<char, bool> isHexChar = c => (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F') || (c >= '0' && c <= '9');
-
             var pos = 0;
             while (pos < len)
             {
-                if (isValidChar(chars[pos]))
+                if (IsValidChar(chars[pos]))
                 {
                     pos = pos + 1;
                 }
-                else if (chars[pos] == '%' && pos + 2 < len && isHexChar(chars[pos + 1]) && isHexChar(chars[pos + 2]))
+                else if (chars[pos] == '%' && pos + 2 < len && IsHexChar(chars[pos + 1]) && IsHexChar(chars[pos + 2]))
                 {
                     pos = pos + 3;
                 }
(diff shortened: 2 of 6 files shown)


Rattenkrieg:
Now caching `FillElements` delegate. See dotnet/roslyn#5835 and linked issues for further information.
I think we should elaborate on leveraging this trick in other performance critical places.

Another allocation hotspot was [this invocation](https://github.com/Rattenkrieg/akka.net/commit/05bf8698fc7468da455b769c147cd1ae10b20ce0#diff-0781b7603a7709c61de01d6d02dff2adL39). Have no idea why someone put `ToList()` here.
default

At the end I've managed to reduce `string[]` allocations at 10x factor.
default

src/core/Akka/Util/FastLazy.cs
```diff
@@ -81,5 +81,80 @@ public T Value
...
+    /// but allows executions to occur concurrently.
+    /// 
+    /// Does not cache exceptions. Designed for use with <typeparamref name="T"/> types that are <see cref="IDisposable"/>
+    /// or are otherwise considered to be expensive to allocate. 
+    /// 
+    /// Read the full explanation here: https://github.com/Aaronontheweb/FastAtomicLazy#rationale
+    /// </summary>
+    /// <typeparam name="S">State type</typeparam>
+    /// <typeparam name="T">Value type</typeparam>
+    public sealed class FastLazy<S, T>
```
alexvaluyskiy: Why you have created this type?

src/core/Akka/Actor/ActorPath.cs
```diff
@@ -185,35 +192,72 @@ protected ActorPath(ActorPath parentPath, string name, long uid)
...
+                parentElems.CopyTo(myElems, 0);
+                myElems[myElems.Length - 1] = actorPath.Name;
+                return myElems;
+            }
+
+            var current = actorPath;
+            var elements = new List<string>();
+            while (!(current is RootActorPath))
+            {
+                if (current._elements.IsValueCreated)
```
alexvaluyskiy: Could you describe these lines.

alexvaluyskiy:
Which test/benchmark have you used to analyze memory traffic?

alexvaluyskiy:
Even SkyNet benchmark is not sufficient for that case, because it reads the Elements proty once

alexvaluyskiy:
ElementsWithUid also should be fixed (and become internal IMHO)

src/core/Akka/Util/FastLazy.cs
Rattenkrieg: See https://github.com/Rattenkrieg/akka.net/commit/05bf8698fc7468da455b769c147cd1ae10b20ce0#diff-74f21fe653ff44fb831459d0f3120fc0R181
A caching optimization for delegates. But now I need to somehow pass parameters to cached func. Decided to create separate class to keep original `FastLazy` _struct_ slim in terms of memory (additional space for state) and code hassle (overloaded ctor, etc).

Rattenkrieg:
I have used SkyNet with api controlled profiling (snapshotting at the very end).
> Even SkyNet benchmark is not sufficient for that case, because it reads the Elements proty once

Thats exactly why your implementation didn't show desired drop of allocations. Here I'm reusing `Parent`'s cached `Elements`. So SkyNet appears worst showcase for such optimization and even using it this fix demonstrates 10x improvement.
Will do same for `ElementsWithUid` when we come to agreement on all fixes.

src/core/Akka/Actor/ActorPath.cs
Rattenkrieg: `current._elements.IsValueCreated` will be `true` when someone already has called `ActorPath.Elements`. So we can stop iterating here and reuse `Parent`'s array. If it's not created yet dont touch it (preventing additional allocations).

src/core/Akka/Util/FastLazy.cs
Aaronontheweb: I think this optimization is pretty clever, for the record 👍

Aaronontheweb: Also, might want to consider contributing this change back to the separate repo I setup for the original `FastLazy` implementation in case others have this issue: https://github.com/Aaronontheweb/FastAtomicLazy/

Rattenkrieg: @Aaronontheweb sure, I will send PR soon.

Rattenkrieg:
I think thats all. Decided not to introduce another lazy field for `ElementsWithUid` (wasting space in my opinion) just reusing `_elements` array and changing last entry - only one array allocation per call.

Aaronontheweb:
Based on the NBench data, this should improve the performance of `Group` routers by about 10%, since they re-use the same `ActorSelection` over and over again:

## Before
|[Counter] ActorSelectionOperationCompleted |      operations |       10,000.00 |       10,000.00 |       10,000.00 |            0.00 |

### Per-second Totals
|[Counter] ActorSelectionOperationCompleted |      operations |      483,912.33 |      448,246.81 |      412,187.56 |       22,317.69 |

## After

| [Counter] ActorSelectionOperationCompleted |      operations |       10,000.00 |       10,000.00 |       10,000.00 |            0.00 |

### Per-second Totals
|[Counter] ActorSelectionOperationCompleted |      operations |      552,883.29 |      492,949.58 |      424,529.30 |       30,472.83 |

alexvaluyskiy:
I've just remember a quote

> Code readability and maintenance is more important than performance until proven otherwise. If you find you do need to make deep changes for performance reasons, do it in a way that is as transparent to the code above it as possible. Keep the level above it as clear as possible.
Once you do make the code worse to read in favor of performance, make sure you document in the code why you are doing it so that someone does not come by after you and “clean up” your elegant optimization by making it simpler. (c) Ben Watson

@Rattenkrieg Could you make please a couple of comments inside the code, about optimizations?

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -4902,6 +4901,12 @@ namespace Akka.Util
         public bool IsValueCreated { get; }
         public T Value { get; }
     }
+    public sealed class FastLazy<S, T>
```
alexvaluyskiy: @Aaronontheweb Should we make both `FastLazy` classes as internal?

alexvaluyskiy:
@Rattenkrieg Why you did not optimize `Skip`
skipiterator
 iterator?

alexvaluyskiy:
@Rattenkrieg Could you please replace this line
https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/ActorCell.DeathWatch.cs#L123

to ` if (watchedBy.Count == 0) return;`

anyenumerator

alexvaluyskiy:
Overall result: in the last 3 performance pull requests (including this one) we have made a big progress.
SpawnBenchmark - memory consumption reduced from ~2GB to ~1GB, execution time was decreased by 20-30%.

Good job

Rattenkrieg:
@alexvaluyskiy I've pushed fixes as you asked. Regarding skip iterator, your're talking about [this line](https://github.com/akkadotnet/akka.net/pull/2554/files#diff-eed7b329733b55c2df26f4c59306a09cR63) right? How it could be optimized? I see no way 👀

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Rattenkrieg: Why can't we push https://github.com/Aaronontheweb/FastAtomicLazy to nuget and use it that way?
BTW @Aaronontheweb I've created an issue and PR to FastAtomicLazy, please check it.

Aaronontheweb: @Rattenkrieg we could do what we did for the Helios.DedicatedThreadPool class: have the NuGet package automatically insert a new `internal FastLazy` code file into the application.

For really small packages I like that approach more than taking a dependency on another binary ("micro" packages.) Does that make sense?

PR #2559, opened 2017-03-16
Akka Fsm Update

alexvaluyskiy:
- [x] [additional debug logging for fsm](https://github.com/akka/akka/pull/20954) [2.4.9]
- [x] [per additional test for onTransition behaviour when initialize](https://github.com/akka/akka/pull/17960) [2.4.0]
- [x] [cancel StateTimeout when FSM actor is stopped](https://github.com/akka/akka/pull/16396) [2.4.0]
- [x] [Send transistion on goto(CurrentState) in FSM](https://github.com/akka/akka/pull/15561) [2.4.0]
- [x] Make all messages as sealed
- [ ] [forMax(Inf) can now override stateTimeout](https://github.com/akka/akka/pull/17152) [2.3.10]
- [ ] Implement `FSMActor_must_unlock_the_lock` test
- [ ] Implement `FSMActor_must_log_events_and_transitions_if_asked_to_do_so` test
- [ ] Implement `FSMActor_must_fill_rolling_event_log_and_hand_it_out` test

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -720,6 +720,7 @@ namespace Akka.Actor
         public TState StateName { get; }
         public void CancelTimer(string name) { }
         public Akka.Actor.FSMBase.State<TState, TData> GoTo(TState nextStateName) { }
+        [System.ObsoleteAttribute("This method is obsoleted. Use GoTo(nextStateName).Using(newStateData) [1.2.0]")]
         public Akka.Actor.FSMBase.State<TState, TData> GoTo(TState nextStateName, TData stateData) { }
         public void Initialize() { }
         public bool IsTimerActive(string name) { }
@@ -750,52 +751,59 @@ namespace Akka.Actor
     public abstract class FSMBase : Akka.Actor.ActorBase
     {
         protected FSMBase() { }
-        public class CurrentState<TS>
+        public sealed class CurrentState<TS>
         {
             public CurrentState(Akka.Actor.IActorRef fsmRef, TS state) { }
             public Akka.Actor.IActorRef FsmRef { get; }
             public TS State { get; }
+            public override bool Equals(object obj) { }
+            public override int GetHashCode() { }
+            public override string ToString() { }
         }
-        public class Event<TD> : Akka.Actor.INoSerializationVerificationNeeded
+        public sealed class Event<TD> : Akka.Actor.INoSerializationVerificationNeeded
         {
             public Event(object fsmEvent, TD stateData) { }
             public object FsmEvent { get; }
             public TD StateData { get; }
             public override string ToString() { }
         }
-        public class Failure : Akka.Actor.FSMBase.Reason
+        public sealed class Failure : Akka.Actor.FSMBase.Reason
         {
             public Failure(object cause) { }
             public object Cause { get; }
         }
-        public class LogEntry<TS, TD>
+        public sealed class LogEntry<TS, TD>
         {
             public LogEntry(TS stateName, TD stateData, object fsmEvent) { }
             public object FsmEvent { get; }
             public TD StateData { get; }
             public TS StateName { get; }
+            public override string ToString() { }
         }
-        public class Normal : Akka.Actor.FSMBase.Reason
+        public sealed class Normal : Akka.Actor.FSMBase.Reason
         {
+            [System.ObsoleteAttribute("This constructor is obsoleted. Use Normal.Instance [1.2.0]")]
             public Normal() { }
+            public static Akka.Actor.FSMBase.Normal Instance { get; }
         }
         public abstract class Reason
         {
             protected Reason() { }
         }
-        public class Shutdown : Akka.Actor.FSMBase.Reason
+        public sealed class Shutdown : Akka.Actor.FSMBase.Reason
         {
+            [System.ObsoleteAttribute("This constructor is obsoleted. Use Shutdown.Instance [1.2.0]")]
             public Shutdown() { }
+            public static Akka.Actor.FSMBase.Shutdown Instance { get; }
         }
         public class State<TS, TD> : System.IEquatable<Akka.Actor.FSMBase.State<TS, TD>>
         {
-            public State(TS stateName, TD stateData, System.Nullable<System.TimeSpan> timeout = null, Akka.Actor.FSMBase.Reason stopReason = null, System.Collections.Generic.List<object> replies = null) { }
-            public System.Collections.Generic.List<object> Replies { get; set; }
+            public State(TS stateName, TD stateData, System.Nullable<System.TimeSpan> timeout = null, Akka.Actor.FSMBase.Reason stopReason = null, System.Collections.Generic.IReadOnlyCollection<object> replies = null, bool notifies = True) { }
+            public System.Collections.Generic.IReadOnlyCollection<object> Replies { get; set; }
             public TD StateData { get; }
             public TS StateName { get; }
             public Akka.Actor.FSMBase.Reason StopReason { get; }
             public System.Nullable<System.TimeSpan> Timeout { get; }
-            public Akka.Actor.FSMBase.State<TS, TD> Copy(System.Nullable<System.TimeSpan> timeout, Akka.Actor.FSMBase.Reason stopReason = null, System.Collections.Generic.List<object> replies = null) { }
             public bool Equals(Akka.Actor.FSMBase.State<TS, TD> other) { }
             public override bool Equals(object obj) { }
             public Akka.Actor.FSMBase.State<TS, TD> ForMax(System.TimeSpan timeout) { }
@@ -804,31 +812,36 @@ namespace Akka.Actor
             public override string ToString() { }
             public Akka.Actor.FSMBase.State<TS, TD> Using(TD nextStateData) { }
         }
-        public class StateTimeout
+        public sealed class StateTimeout
         {
+            [System.ObsoleteAttribute("This constructor is obsoleted. Use Shutdown.Instance [1.2.0]")]
             public StateTimeout() { }
+            public static Akka.Actor.FSMBase.StateTimeout Instance { get; }
         }
-        public class StopEvent<TS, TD> : Akka.Actor.INoSerializationVerificationNeeded
+        public sealed class StopEvent<TS, TD> : Akka.Actor.INoSerializationVerificationNeeded
         {
             public StopEvent(Akka.Actor.FSMBase.Reason reason, TS terminatedState, TD stateData) { }
             public Akka.Actor.FSMBase.Reason Reason { get; }
             public TD StateData { get; }
             public TS TerminatedState { get; }
+            public override string ToString() { }
         }
-        public class SubscribeTransitionCallBack
+        public sealed class SubscribeTransitionCallBack
         {
             public SubscribeTransitionCallBack(Akka.Actor.IActorRef actorRef) { }
             public Akka.Actor.IActorRef ActorRef { get; }
         }
-        public class Transition<TS>
+        public sealed class Transition<TS>
         {
             public Transition(Akka.Actor.IActorRef fsmRef, TS from, TS to) { }
             public TS From { get; }
             public Akka.Actor.IActorRef FsmRef { get; }
             public TS To { get; }
+            public override bool Equals(object obj) { }
+            public override int GetHashCode() { }
             public override string ToString() { }
         }
-        public class UnsubscribeTransitionCallBack
+        public sealed class UnsubscribeTransitionCallBack
         {
             public UnsubscribeTransitionCallBack(Akka.Actor.IActorRef actorRef) { }
             public Akka.Actor.IActorRef ActorRef { get; }
@@ -3952,7 +3965,7 @@ namespace Akka.Routing
         protected CustomRouterConfig() { }
         protected CustomRouterConfig(string routerDispatcher) { }
     }
-    public class Deafen : Akka.Routing.ListenerMessage
+    public sealed class Deafen : Akka.Routing.ListenerMessage
     {
         public Deafen(Akka.Actor.IActorRef listener) { }
         public Akka.Actor.IActorRef Listener { get; }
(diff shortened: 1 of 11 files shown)


src/core/Akka/Actor/FSM.cs
```diff
@@ -650,13 +739,7 @@ public void StartWith(TState stateName, TData stateData, TimeSpan? timeout = nul
...
 
-        /// <summary>
-        /// Produce transition to other state. Return this from a state function
-        /// in order to effect the transition.
-        /// </summary>
-        /// <param name="nextStateName">State designator for the next state</param>
-        /// <param name="stateData">Data for next state</param>
-        /// <returns>State transition descriptor</returns>
+        [Obsolete("GoTo method is osboleted. Use GoTo(TState nextStateName) [1.2.0]")]
         public State<TState, TData> GoTo(TState nextStateName, TData stateData)
```
alexvaluyskiy: This method is not present in the Scala API

src/core/Akka.Persistence/Fsm/PersistentFSMBase.cs
```diff
@@ -1059,7 +1059,7 @@ public State ForMax(TimeSpan timeout)
             /// <summary>
             /// TBD
             /// </summary>
-            public List<object> Replies
+            public IReadOnlyCollection<object> Replies
```
alexvaluyskiy: We should prefer `ReadOnly*` and `Immutable` collections in the public API. But this change could break compatibility

src/core/Akka.Tests/Actor/FSMTimingSpec.cs
```diff
@@ -8,223 +8,220 @@
...
-                ExpectMsg(new FSMBase.Transition<State>(fsm, State.TestRepeatedTimer, State.Initial), FSMSpecHelpers.TransitionStateExpector<State>());
-                return true;
+                ExpectMsg(new Transition<State>(fsm, State.TestRepeatedTimer, State.Initial));
             });
         }
 
+        [Fact]
+        public void FSM_must_notify_unhandled_messages()
+        {
+            // EventFilter
```
alexvaluyskiy: EventFilter chaining is not working as expected

src/core/Akka/Actor/FSM.cs
```diff
@@ -151,23 +202,39 @@ public abstract class Reason { }
...
+        }
 
         /// <summary>
         /// Reason given when someone as calling <see cref="Stop"/> from outside;
         /// also applies to <see cref="ActorSystem"/> supervision directive.
         /// </summary>
-        public class Shutdown : Reason { }
+        public sealed class Shutdown : Reason
+        {
+            private Shutdown() { }
```
alexvaluyskiy: This one could break compatibility. Probably I should keep a public constructor with `ObsoletedAttribute`

src/core/Akka/Actor/FSM.cs
```diff
@@ -151,23 +202,39 @@ public abstract class Reason { }
         /// <summary>
         /// Default <see cref="Reason"/> if calling Stop().
         /// </summary>
-        public class Normal : Reason { }
+        public sealed class Normal : Reason
+        {
+            private Normal() { }
```
alexvaluyskiy: This one could break compatibility. Probably I should keep a public constructor with `ObsoletedAttribute`

src/core/Akka/Actor/FSM.cs
```diff
@@ -275,31 +341,20 @@ public Timer(string name, object message, bool repeat, int generation, IActorCon
...
             public void Schedule(IActorRef actor, TimeSpan timeout)
             {
-                var name = Name;
-                var message = Message;
-
-                Action send;
-                if(_debugLog != null)
-                    send = () =>
-                    {
-                        _debugLog.Debug("{0}Timer '{1}' went off. Sending {2} -> {3}",_ref.IsCancellationRequested ? "Cancelled " : "", name, message, actor);
```
alexvaluyskiy: Why this debugging was implemented? It is not present in the scala API

src/core/Akka/Actor/FSM.cs
```diff
@@ -275,31 +341,20 @@ public Timer(string name, object message, bool repeat, int generation, IActorCon
...
+                    ? _scheduler.ScheduleTellRepeatedlyCancelable(timeout, timeout, actor, this, Context.Self) 
+                    : _scheduler.ScheduleTellOnceCancelable(timeout, actor, this, Context.Self);
             }
 
             /// <summary>
             /// TBD
             /// </summary>
             public void Cancel()
             {
-                if (!_ref.IsCancellationRequested)
```
alexvaluyskiy: `IsCancellationRequested` is always `false`

src/core/Akka/Actor/FSM.cs
```diff
@@ -178,22 +245,26 @@ public Failure(object cause)
...
+            public object Cause { get; }
         }
 
         /// <summary>
         /// Used in the event of a timeout between transitions
         /// </summary>
-        public class StateTimeout { }
+        public sealed class StateTimeout
+        {
+            private StateTimeout() { }
```
alexvaluyskiy: This one could break compatibility. Probably I should keep a public constructor with `ObsoletedAttribute`

alexvaluyskiy:
Why we have implemented `LoggingFsm` and an interface with implicit behavior. Instead of creating `LoggingFsm` class?

alexvaluyskiy:
Should I implement this one? 
https://github.com/akka/akka/pull/17152

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -768,52 +769,55 @@ namespace Akka.Actor
     public abstract class FSMBase : Akka.Actor.ActorBase
     {
         protected FSMBase() { }
-        public class CurrentState<TS>
+        public sealed class CurrentState<TS>
```
Aaronontheweb: Sealing all of this off seems good to me.

src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
```diff
@@ -846,14 +846,14 @@ namespace Akka.Persistence.Fsm
         public void WhenUnhandled(Akka.Persistence.Fsm.PersistentFSMBase<TState, TData, TEvent>.StateFunction stateFunction) { }
         public class State<TState, TData, TEvent> : Akka.Actor.FSMBase.State<TState, TData>
         {
-            public State(TState stateName, TData stateData, System.Nullable<System.TimeSpan> timeout = null, Akka.Actor.FSMBase.Reason stopReason = null, System.Collections.Generic.List<object> replies = null, Akka.Util.ILinearSeq<TEvent> domainEvents = null, System.Action<TData> afterTransitionDo = null) { }
+            public State(TState stateName, TData stateData, System.Nullable<System.TimeSpan> timeout = null, Akka.Actor.FSMBase.Reason stopReason = null, System.Collections.Generic.IReadOnlyCollection<object> replies = null, Akka.Util.ILinearSeq<TEvent> domainEvents = null, System.Action<TData> afterTransitionDo = null) { }
```
Aaronontheweb: Like this change too. We should have been using read-only collections from the start here.

src/core/Akka.Persistence/Fsm/PersistentFSMBase.cs
Aaronontheweb: It shouldn't really affect how end-users are consuming it that much, unless they were modifying the collection themselves which would have been "illegal" anyway. Guess this falls into the gray area of "is fixing a bug in the public API really a breaking change?" I'm fine with it.

src/core/Akka.Tests/Actor/FSMTimingSpec.cs
Aaronontheweb: I'm assuming that's unrelated to this PR, and IIRC it's an area where we have some known bugs still anyway.

alexvaluyskiy:
When we are `converting` scala's traits, we should prefer to use Java API experience. Like to create an additional abstract class, like `AbstractLoggingFSM`

src/core/Akka/Actor/FSM.cs
Aaronontheweb: Yeah, I would do that.

src/core/Akka/Actor/FSM.cs
Aaronontheweb: Same as previous. Go ahead and add that back.

src/core/Akka/Actor/FSM.cs
Aaronontheweb: Same as previous. Go ahead and add that back.

alexvaluyskiy:
Why we need `FSMBase` abstract class? All messages should situated inside `FSM` class. `FSMBase` is redundant file. But we could not drop it without breaking changes.
At least we could make it static. And `FSM` actor  will inherit `ActorBase`

src/core/Akka/Actor/FSM.cs
Aaronontheweb: I don't think I've even seen this method get used. It's always been `GoTo().WithState` AFAIK

src/core/Akka/Actor/FSM.cs
```diff
@@ -744,17 +834,18 @@ public StateFunction Using(Func<State<TState, TData>, State<TState, TData>> andT
         /// <param name="repeat">send once if false, scheduleAtFixedRate if true</param>
         public void SetTimer(string name, object msg, TimeSpan timeout, bool repeat = false)
         {
-            if(DebugEvent)
-                _log.Debug("setting " + (repeat ? "repeating" : "") + "timer '{0}' / {1}: {2}", name, timeout, msg);
-            if(_timers.ContainsKey(name))
+            if (DebugEvent)
+            {
+                _log.Debug($"setting {(repeat ? "repeating" : "")} timer {name}/{timeout}: {msg}");
```
Aaronontheweb: Liking the C#6 API here... makes it way easier to read.

src/core/Akka/Actor/FSM.cs
```diff
@@ -795,24 +886,14 @@ public bool IsTimerActive(string name)
         /// <param name="timeout">TBD</param>
         public void SetStateTimeout(TState state, TimeSpan? timeout)
         {
-            if(!_stateTimeouts.ContainsKey(state))
-                _stateTimeouts.Add(state, timeout);
-            else
-                _stateTimeouts[state] = timeout;
+            _stateTimeouts.AddOrSet(state, timeout);
```
Aaronontheweb: Change this to 

`stateTimeouts[state] = timeout`

AddOrSet is something stupid I added to the codebase that we should deprecate and remove at some point. It's an unnecessary abstraction.

src/core/Akka/Actor/FSM.cs
```diff
@@ -1017,73 +1097,82 @@ private static StateFunction OrElse(StateFunction original, StateFunction fallba
...
+            if (timeoutMarker != null)
+            {
+                if (_generation == timeoutMarker.Generation)
                 {
-                    Context.Unwatch(d.Listener);
-                    Listeners.Remove(d.Listener);
-                })
-                .With<InternalActivateFsmLogging>(_=> { DebugEvent = true; })
-                .Default(msg =>
+                    ProcessMsg(StateTimeout.Instance, "state timeout");
```
Aaronontheweb: Way easier to read (the main FSM handling routine) and will have fewer allocations also. Nice job.

src/core/Akka/Actor/FSM.cs
```diff
@@ -1137,43 +1235,38 @@ private string GetSourceString(object source)
...
-                            new Failure(String.Format("Next state {0} does not exist", upcomingState.StateName))));
+                Terminate(Stay().WithStopReason(new Failure($"Next state {nextState.StateName} does not exist")));
             }
             else
             {
-                var replies = upcomingState.Replies;
-                replies.Reverse();
-                foreach (var r in replies) { Sender.Tell(r); }
-                if (!_currentState.StateName.Equals(upcomingState.StateName))
+                nextState.Replies.Reverse().ForEach(r => Sender.Tell(r));
```
Aaronontheweb: Much cleaner.

src/core/Akka/Actor/FSM.cs
alexvaluyskiy: We should avoid of PatternMatch class

src/core/Akka/Actor/FSM.cs
alexvaluyskiy: Hold my beer. I'm planning to do C# 7 refactoring :)

src/core/Akka/Actor/FSM.cs
alexvaluyskiy: I think I should get it back. List.Reverse has better performance than IEnumerable.Reverse

sean-gilliam: instead of reversing the list and then iterating, why not do both like

```csharp
for (int i = nextState.Replies.Count - 1; i >= 0; i--)
{
    Sender.Tell(nextState.Replies[i]);
}
```

src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
```diff
@@ -854,7 +854,7 @@ namespace Akka.Persistence.Fsm
             public System.Action<TData> AfterTransitionHandler { get; }
             public Akka.Util.ILinearSeq<TEvent> DomainEvents { get; }
             public bool Notifies { get; set; }
-            public System.Collections.Generic.IReadOnlyCollection<object> Replies { get; }
+            public System.Collections.Generic.IReadOnlyList<object> Replies { get; }
```
Danthar: Why the switch to IReadOnlyList ?

alexvaluyskiy: Because I implemented this code :)
```csharp
for (int i = nextState.Replies.Count - 1; i >= 0; i--)
{
    Sender.Tell(nextState.Replies[i]);
}
```

Aaronontheweb:
@alexvaluyskiy on the `LoggingFsm` bit, because I thought that preferable to having an explosion of base classes to choose from. Better to be able to decorate an existing one to add a piece of minor functionality like that instead of having to change the class you inherit from.

src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
sean-gilliam: :+1:

PR #2561, opened 2017-03-18
Prepare ClusterTools API to the stable version

alexvaluyskiy:
- Made most of the public classes as sealed
- Replaced public fields by properties
- Changed visibility of serialization members


src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientSettings.cs
```diff
@@ -62,44 +62,44 @@ public static ClusterClientSettings Create(Config config)
         /// <summary>
         /// Actor paths of the <see cref="ClusterReceptionist"/> actors on the servers (cluster nodes) that the client will try to contact initially.
         /// </summary>
-        public readonly IImmutableSet<ActorPath> InitialContacts;
+        public IImmutableSet<ActorPath> InitialContacts { get; }
```
alexvaluyskiy: Should we expose `ImmutableList` instead of some ReadOnly abstraction?

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
```diff
@@ -661,51 +667,21 @@ public override int GetHashCode()
...
-            return string.Format("SendToAll<path:{0}, excludeSelf:{1}, message:{2}>", Path, ExcludeSelf, Message);
+            return $"SendToAll<path:{Path}, excludeSelf:{ExcludeSelf}, message:{Message}>";
         }
     }
 
     /// <summary>
     /// TBD
     /// </summary>
     [Serializable]
-    public sealed class GetTopics : IEquatable<GetTopics>
```
alexvaluyskiy: Why we need an `IEquatable` for singletons?

src/core/Akka.API.Tests/CoreAPISpec.ApproveClusterTools.approved.txt
```diff
@@ -802,22 +165,22 @@ namespace Akka.Cluster.Tools.PublishSubscribe
 {
     public sealed class CurrentTopics : System.IEquatable<Akka.Cluster.Tools.PublishSubscribe.CurrentTopics>
     {
-        public readonly string[] Topics;
         public CurrentTopics(string[] topics) { }
+        public string[] Topics { get; }
```
alexvaluyskiy: I'm not fan of exposing arrays in public API. Could I change it to IReadOnlyCollection<string> ?

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientSettings.cs
Danthar: Exposing the interface is fine imo.

Danthar:
Another note that I want to add. Im not a fan of forcing people to define an `IImmutableSet` for their initial contact list. 
I think its an implementation detail that should be the responsibility of the underlying structure. And not the user.
Same goes for the `WithInitialContacts` function.

We could still provide a private copy constructor that does accept the `ImmutableSet` parameter though.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
Danthar: Agreed. Seems unnecessary

src/core/Akka.API.Tests/CoreAPISpec.ApproveClusterTools.approved.txt
Danthar: In this case I do not think it matters much. Wrapping it in an IReadOnlyCollection in this case only provides an illusion of safety.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
Horusiath: This is serializable class. Singleton messages are not guaranteed to be deserialized back into `Instance` object. Moreover, usually they are guaranteed to be different objects.

Aaronontheweb:
@alexvaluyskiy minor merge conflict

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
alexvaluyskiy: We don't transfer this object via wire. it is local only

Aaronontheweb:
@Danthar @alexvaluyskiy @Horusiath is this ready for merge?

alexvaluyskiy:
@Aaronontheweb ready

PR #2569, opened 2017-03-25
Hide MatchHandler from the public API

alexvaluyskiy:
- Hide MatchHandler from the public API
- Added Xml comments to AutoReceive messages
- Enabled InternalVisibleTo for Persistence and for Streams

### DIFF
--- a/src/core/Akka.Persistence/AtLeastOnceDelivery.cs
+++ b/src/core/Akka.Persistence/AtLeastOnceDelivery.cs
@@ -159,7 +159,7 @@ namespace Akka.Persistence
         /// <param name="receive">TBD</param>
         /// <param name="message">TBD</param>
         /// <returns>TBD</returns>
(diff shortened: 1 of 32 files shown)


src/core/Akka.Persistence/AtLeastOnceDelivery.cs
```diff
@@ -159,7 +159,7 @@ protected override void OnReplaySuccess()
         /// <param name="receive">TBD</param>
         /// <param name="message">TBD</param>
         /// <returns>TBD</returns>
-        protected override bool AroundReceive(Receive receive, object message)
+        protected internal override bool AroundReceive(Receive receive, object message)
```
alexvaluyskiy: It is not affect a public API. But we need it if we want to enable [InternalVisibleTo] for this assembly

Aaronontheweb:
@alexvaluyskiy have a conflict with @sean-gilliam's PR

PR #2574, opened 2017-03-29
close #2570 - ported to Akka.Remote.TestKit to DotNetty

### LINKED ISSUE #2570: Replace Helios to DotNetty in Akka.Remote.TestKit
While Helios was replaced in Remote and in the all test projects, Remote.TestKit still dependes on Helios. It should be replaced there.

The problem is - DotNetty does not expose ILogger interface, something else should be used as a logger

Aaronontheweb:


Aaronontheweb:
I had to disable some of the extra serialization and encoding logging we used to have with Helios since the DotNetty implementation of that stuff depends on implementing a bunch of gnarly Microsoft.Extensions.Logging stuff

There's a pre-built package we can use but that requires a .NET Framework version runtime bump.

I'm thinking we can add this back once we're on .NET Standard or 4.5.2

src/core/Akka.Remote.TestKit/Conductor.cs
```diff
@@ -74,17 +73,6 @@ public IActorRef Controller
         }
 
         /// <summary>
-        /// Obtain the port to which the controller’s socket is actually bound. This
-        /// will deviate from the configuration in `akka.testconductor.port` in case
-        /// that was given as zero.
-        /// </summary>
-        /// <returns>The address of the controller's socket endpoint</returns>
-        public Task<INode> SockAddr()
```
alexvaluyskiy: Have we used this method somewhere?

Aaronontheweb: Not since Akka.NET 1.1; it directly exposed old Helios 1.4* members to the public and stopped being used after we switched to Helios 2.0 last year.

Aaronontheweb:
Looks like this had some portability differences between the Helios 2.x and DotNetty APIs. Addressed those and added standard out logging back.

Aaronontheweb:
Also added an additional spec that can run outside the MNTR to verify that `RemoteConnection` works.

Aaronontheweb:
Bah. Still need to prune the `packages.config` for Akka.Remote.TestKit before we merge this in.

src/core/Akka.Remote.TestKit/packages.config
```diff
@@ -1,9 +1,47 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="DotNetty.Buffers" version="0.4.3" targetFramework="net45" />
+  <package id="DotNetty.Codecs" version="0.4.3" targetFramework="net45" />
+  <package id="DotNetty.Common" version="0.4.3" targetFramework="net45" />
+  <package id="DotNetty.Transport" version="0.4.3" targetFramework="net45" />
   <package id="Google.ProtocolBuffers" version="2.4.1.555" targetFramework="net45" />
-  <package id="Helios" version="2.1.2" targetFramework="net45" />
+  <package id="Microsoft.Extensions.DependencyInjection.Abstractions" version="1.1.0" targetFramework="net45" />
```
Aaronontheweb: Need to prune all of this crap out so it doesn't get pulled into the `.nuspec` metadata for `Akka.Remote.TestKit` as explicit dependencies.

alexvaluyskiy:
LGTM

Issue #2566, opened 2017-03-22
Too many allocation of string on serialization


alexvaluyskiy:
I've used this test to measure memory traffic: https://github.com/alexvaluyskiy/ActorBenchmarks/tree/master/AkkaNet/LocalPingPongWithSerialization

ActorPath.ToStringWithAddress
![localpingpongstringallocations1](https://cloud.githubusercontent.com/assets/4232072/24211133/d86f21f2-0f33-11e7-8596-35e027aca753.png)

alexvaluyskiy:
TypeExtensions.TypeQualifiedName
![localpingpongstringallocations2](https://cloud.githubusercontent.com/assets/4232072/24211181/f326b6e0-0f33-11e7-906e-ddc6541b3f53.png)

alexvaluyskiy:
ActorPath.TryParse
![localpingpongstringallocations3](https://cloud.githubusercontent.com/assets/4232072/24211206/00792846-0f34-11e7-8210-5b7e0ab71699.png)

alexvaluyskiy:
These methods are slow in profiler
`TypeExtensions.TypeQualifiedName`
`Serialization.FindSerializerForType`
`ActorPath.ToStringWithAddress` and `ActorPath.TryParseAddress`
`LocalActorRef.GetChild` and `LocalActorRef.GetSingleChild`

![hotmethods](https://cloud.githubusercontent.com/assets/4232072/24211494/f275a14c-0f34-11e7-8dc9-83f6907953a7.png)

alexvaluyskiy:
Scala version has implemented FindSerializerForType a bit smarter https://github.com/akka/akka/blob/v2.5.0-RC1/akka-actor/src/main/scala/akka/serialization/Serialization.scala#L222

Should work much faster if we try to find a serializer for some type in the second and subsequent times

siudeks:
The link provided in the first comment (LocalPingPongWithSerialization) is not available (404)

PR #2576, opened 2017-03-30
Make FindSerializerForType faster

alexvaluyskiy:
Related https://github.com/akkadotnet/akka.net/issues/2566

I've reused some parts of Scala implementation
**Before**: 1 400 243 ops/s
**After**: 4 551 167 ops/s

### DIFF
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -123,14 +123,25 @@ namespace Akka.Serialization
         /// </summary>
         public ActorSystem System { get; }
 
+        /// <summary>
+        /// TBD
+        /// </summary>
+        /// <param name="serializer">TBD</param>
+        /// <returns>TBD</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal void AddSerializer(Serializer serializer)
+        public void AddSerializer(Serializer serializer)
         {
             _serializers.Add(serializer.Identifier, serializer);
         }
 
+        /// <summary>
+        /// TBD
+        /// </summary>
+        /// <param name="type">TBD</param>
+        /// <param name="serializer">TBD</param>
+        /// <returns>TBD</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal void AddSerializationMap(Type type, Serializer serializer)
+        public void AddSerializationMap(Type type, Serializer serializer)
         {
             _serializerMap[type] = serializer;
         }
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.Serialization;
@@ -60,7 +61,7 @@ namespace Akka.Serialization
 
         private readonly Serializer _nullSerializer;
 
-        private readonly Dictionary<Type, Serializer> _serializerMap = new Dictionary<Type, Serializer>();
+        private readonly ConcurrentDictionary<Type, Serializer> _serializerMap = new ConcurrentDictionary<Type, Serializer>();
         private readonly Dictionary<int, Serializer> _serializers = new Dictionary<int, Serializer>();
 
         /// <summary>
@@ -118,7 +119,7 @@ namespace Akka.Serialization
                     continue;
                 }
 
-                _serializerMap.Add(messageType, serializer);
+                _serializerMap.TryAdd(messageType, serializer);
             }
         }
 
@@ -143,7 +144,7 @@ namespace Akka.Serialization
         /// <param name="serializer">TBD</param>
         public void AddSerializationMap(Type type, Serializer serializer)
         {
-            _serializerMap.Add(type, serializer);
+            _serializerMap.TryAdd(type, serializer);
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


Horusiath:
Looks like DData types have been resolved to NewtonsoftJsonSerializer instead of their custom implementation (which is hyperion-based).

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -229,20 +229,34 @@ public Serializer FindSerializerFor(object obj)
...
-                return _serializerMap[_objectType];
+                //do a final check for the "object" serializer
+                if (_serializerMap.ContainsKey(_objectType) && _objectType.IsAssignableFrom(type))
+                    serializer = _serializerMap[_objectType];
 
-            throw new SerializationException($"Serializer not found for type {objectType.Name}");
+                if (serializer == null)
+                    throw new SerializationException($"Serializer not found for type {objectType.Name}");
+
+                _serializerMap.Add(type, serializer);
```
Rattenkrieg: Shouldn't that become threadsafe collection in presence of updates? In Scala ConcurrentHashMap is used.

Rattenkrieg: Also there will be duplicate exception if we first found object on line 251 and then trying to add it back. What the point of this btw?

alexvaluyskiy: We will gen an exception only if we remove all serializers from all configs

alexvaluyskiy: I've changed it to ConcurrentDictionary

Aaronontheweb: Why not just change it to `_serializerMap.Add[type] = serializer;` ? Becomes an upsert that way.

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -229,20 +230,37 @@ public Serializer FindSerializerFor(object obj)
...
+                        serializer = serializerType.Value;
+                        break;
+                    }
+                }
 
-            //do a final check for the "object" serializer
-            if (_serializerMap.ContainsKey(_objectType) && _objectType.IsAssignableFrom(type))
-                return _serializerMap[_objectType];
+                //do a final check for the "object" serializer
+                if (serializer == null && _serializerMap.ContainsKey(_objectType) && _objectType.IsAssignableFrom(type))
```
alexvaluyskiy: Why we need to check that type is subclass of objec?

Aaronontheweb: lol.... it's literally checking to see if we're assignable from `object` ?

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -6,27 +6,23 @@
...
 using Akka.Util.Internal;
 
 namespace Akka.Serialization
 {
     /// <summary>
-    /// TBD
+    /// Serialization information needed for serializing local actor refs.
     /// </summary>
-    public class Information
+    internal class Information
```
alexvaluyskiy: Made it internal

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -60,7 +56,7 @@ public static T SerializeWithTransport<T>(ActorSystem system, Address address, F
 
         private readonly Serializer _nullSerializer;
 
-        private readonly Dictionary<Type, Serializer> _serializerMap = new Dictionary<Type, Serializer>();
+        private readonly ConcurrentDictionary<Type, Serializer> _serializerMap = new ConcurrentDictionary<Type, Serializer>();
```
alexvaluyskiy: Changed to ConcurrentDictionary

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -294,12 +300,7 @@ public static string SerializedActorPath(IActorRef actorRef)
             }
         }
 
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="serializerId">TBD</param>
-        /// <returns>TBD</returns>
-        public Serializer GetSerializerById(int serializerId)
+        internal Serializer GetSerializerById(int serializerId)
```
alexvaluyskiy: This one should be internal

src/core/Akka/Serialization/Serializer.cs
```diff
@@ -87,22 +88,11 @@ public byte[] ToBinaryWithAddress(Address address, object obj)
...
-        /// </remarks>
-        /// <param name="type">TBD</param>
-        /// <returns>TBD</returns>
-        protected static string TypeQualifiedNameForManifest(Type type)
-        {
-            return type == null ? string.Empty : string.Format("{0},{1}", type.FullName, type.Assembly.GetName().Name);
-        }
+        /// <param name="bytes">The array containing the serialized object</param>
+        /// <returns>The object contained in the array</returns>
+        public T FromBinary<T>(byte[] bytes) => (T)FromBinary(bytes, typeof(T));
```
alexvaluyskiy: I think we should have this method

src/core/Akka/Serialization/Serializer.cs
```diff
@@ -87,22 +88,11 @@ public byte[] ToBinaryWithAddress(Address address, object obj)
...
         /// </summary>
-        /// <remarks>
-        /// See <see cref="Type.GetType(string)"/> for details on how a type is looked up
-        /// from a name. In particular, if the (partial) assembly name is not included
-        /// only the assembly calling <see cref="Type.GetType(string)"/> is searched.
-        /// If the (partial) assembly name is included, it searches in the specified assembly.
-        /// </remarks>
-        /// <param name="type">TBD</param>
-        /// <returns>TBD</returns>
-        protected static string TypeQualifiedNameForManifest(Type type)
```
alexvaluyskiy: We have an extension method for than. We don't need this method anymore

src/core/Akka/Serialization/Serialization.cs
alexvaluyskiy: changed

PR #2578, opened 2017-03-31
[api-docs] Even more documentation

sean-gilliam:
This PR fleshes out more of the documentation labeled as TBD.

### DIFF
--- a/src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
+++ b/src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
@@ -41,22 +41,20 @@ namespace Akka.Persistence
         /// </summary>
         /// <param name="currentDeliveryId">TBD</param>
         /// <param name="unconfirmedDeliveries">TBD</param>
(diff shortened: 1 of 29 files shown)


src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
```diff
@@ -141,31 +154,37 @@ public bool Equals(UnconfirmedWarning other)
...
+        /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table. 
+        /// </returns>
         public override int GetHashCode()
         {
             return (UnconfirmedDeliveries != null ? UnconfirmedDeliveries.GetHashCode() : 0);
         }
 
         /// <summary>
-        /// TBD
+        /// Returns a <see cref="System.String" /> that represents this instance.
```
alexvaluyskiy: We should use inheritdoc here

sean-gilliam: I thought docfx didn't support that. Believe me, I'd love to use it.

PR #2580, opened 2017-04-01
Added GraphDSL.CreateMaterialized method

Horusiath:
It's a small improvement: a `GraphDSL.CreateMaterialized<TShape, TMat>` method. It's more flexible than `GraphDSL.Create<TShape>` method as it allows to construct from scratch a graph, than may have materialized value different than `NotUsed`. 

It's necessary, as there is no other way to build a generic `IGraph<TMat>` version - builder dsl constructors are internal. Personally I need this for building F# API for constructing graphs.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
@@ -1241,6 +1241,8 @@ namespace Akka.Streams.Dsl
             where TShape2 : Akka.Streams.Shape
             where TShape3 : Akka.Streams.Shape
             where TShape4 : Akka.Streams.Shape { }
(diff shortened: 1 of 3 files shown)


Silv3rcircl3:
@Horusiath ApiApproval is missing, also while your on it you could replace the other TBD comments for the other overloads ;-)

alexvaluyskiy:
Is that API comes from Streams JVM?

Horusiath:
@alexvaluyskiy not directly. JVM version of create has inherently generic materializer - however we deliberatelly decided to have it fixed to `NotUsed` type to guarantee, that C# compiler will be able to infere types. Otherwise, a full list of generic parameters would be need.

However problem is that graph API doesn't expose `Create` equivalent with generic materialized value - also graph builder has internal constructor, so this makes it virtually impossible to create such graph with user-specified materialized value without using more complex overloads. 

Personally for me it's problem for Akkling - which replaced `NotUsed` with F# `unit` type - which has direct compiler support - where possible.

PR #2583, opened 2017-04-06
close #2581 - fixed PhiAccrualFailureDetector clock roll over

### LINKED ISSUE #2581: intervalSum must be >= 0, got -2146468163
Got this exception in logs from two Lighthouse instances:

Exception message:
```
intervalSum must be >= 0, got -2146468163
Parameter name: intervalSum
```
Source: `akka://FooBar/system/cluster/core/daemon/heartbeatSender`

Looks like some counter overlapped.
Don't know yet, is it harmless or not, keeping for records for now.

Akka 1.1.3

Aaronontheweb:


src/core/Akka.Remote.Tests/AccrualFailureDetectorSpec.cs
```diff
@@ -245,6 +245,45 @@ public void StatisticsForHeartbeats_must_be_capped_by_the_specified_maxSampleSiz
             ShouldBe(history5.Variance, 688.88888889D, 0.00001D);
         }
 
+
+        // handle an edge case when the Clock rolls over
+        // see https://github.com/akkadotnet/akka.net/issues/2581
+        [Fact]
+        public void PhiAccrualHistory_can_roll_over()
```
Aaronontheweb: Reproduction spec

src/core/Akka.Remote.Tests/PhiAccrualModelBasedSpecs.cs
```diff
@@ -0,0 +1,159 @@
...
+using FluentAssertions;
+using FsCheck;
+using FsCheck.Experimental;
+using FsCheck.Xunit;
+using Microsoft.FSharp.Core;
+using Xunit;
+
+namespace Akka.Remote.Tests
+{
+    public class PhiAccrualModelBasedSpecs
```
Aaronontheweb: Added model-based test here just to make sure all of the math was being done correctly inside the `PhiAccrualFailureDetector`.

src/core/Akka.Remote/FailureDetector.cs
```diff
@@ -34,12 +35,11 @@ public abstract class FailureDetector
...
 
-        //fixed: sign will no longer flip, but the tickcount will go back down to zero every 24.9 days 
         /// <summary>
-        /// TBD
+        /// The default clock implementation used by the <see cref="PhiAccrualFailureDetector"/>
         /// </summary>
-        /// <returns>TBD</returns>
-        public static readonly Clock DefaultClock = () => Environment.TickCount & Int32.MaxValue;
+        /// <returns>A clock instance.</returns>
+        public static readonly Clock DefaultClock = () => MonotonicClock.GetTicks();
```
Aaronontheweb: Key fix to the issue here. Switched to 64 bit timer instead of the previous 32bit implementation with `Environment.TickCount`.

src/core/Akka.Remote/PhiAccrualFailureDetector.cs
```diff
@@ -222,11 +222,14 @@ internal double Phi(long timestamp)
                 return 0.0d; //treat unmanaged connections, e.g. with zero heartbeats, as healthy connections
             else
             {
-                var timeDiff = timestamp - oldTimestamp.Value;
-                var history = oldState.History;
-                var mean = history.Mean;
-                var stdDeviation = EnsureValidStdDeviation(history.StdDeviation);
-                return Phi(timeDiff, mean + AcceptableHeartbeatPauseMillis, stdDeviation);
+                unchecked // in the extremely rare event of a clock roll-over
```
Aaronontheweb: Made this unchecked in the extremely rare event of a clock rollover. Realistically, this will never happen.

Aaronontheweb:
Deadline failure detector was using MS when it should have been using ticks, hence the failure. Fixed in latest commit.

PR #2585, opened 2017-04-06
Revise TestActor used for EventsByPersistenceIdSpec to block until ack of Journal cleanup

heynickc:
PR in response to a set of tests deemed flaky by TeamCity CI.  Example here: http://petabridge-ci.cloudapp.net/viewLog.html?buildId=21833&tab=buildResultsDiv&buildTypeId=AkkaNet_AkkaPersistenceImplementations_AkkaPersistenceSqlServer_PrBuilds#testNameId-488764090331667693

Test failures were due to the `DeleteCommand` being sent to the `TestActor` not completing the Journal cleanup before the assertions were being made against the cleaned-up Journal.

```
Failed: Expected a message of type Akka.Streams.TestKit.TestSubscriber+OnComplete, 
but received {TestSubscriber.OnNext(g1-1)} 
(type Akka.Streams.TestKit.TestSubscriber+OnNext`1[System.Object]) instead  
from [akka://test/user/StreamSupervisor-16/Flow-0-0-select#2080382836]
```

Revised `TestActor` to send acknowledgement to TestKit actor only after it has received the `DeleteMessagesSuccess` message back from the Journal.  Also added `AwaitAssert` to wait for the `ExpectMsg("-deleted")` message or fail if not received within default timespan.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.TestKit/EventsByPersistenceIdSpec.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.TestKit/EventsByPersistenceIdSpec.cs
@@ -91,7 +91,7 @@ namespace Akka.Persistence.Sql.TestKit
             var pref = Setup("g1");
 
             pref.Tell(new TestActor.DeleteCommand(3));
-            ExpectMsg("3-deleted");
+            AwaitAssert(() => ExpectMsg("3-deleted"));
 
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
@heynickc after the nightlies have a chance to run, you should be able to do your integration tests with this against the individual plugins.

PR #2587, opened 2017-04-07
Added more messages which ignores the serialization

alexvaluyskiy:
I've skipped the persistence

### DIFF
--- a/src/core/Akka.Cluster/ClusterDaemon.cs
+++ b/src/core/Akka.Cluster/ClusterDaemon.cs
@@ -38,20 +38,28 @@ namespace Akka.Cluster
         /// </summary>
         internal abstract class BaseClusterUserAction
         {
-            readonly Address _address;
-
-            /// <summary>
-            /// TBD
-            /// </summary>
-            public Address Address { get { return _address; } }
-
             /// <summary>
             /// TBD
             /// </summary>
             /// <param name="address">TBD</param>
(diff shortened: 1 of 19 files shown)


src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -408,21 +408,20 @@ private IInternalActorRef MakeChild(Props props, string name, bool async, bool s
                     {
                         if (argument != null && !(argument is INoSerializationVerificationNeeded))
                         {
-                            var objectType = argument.GetType();
```
alexvaluyskiy: We were serializing the type of object, instead of object  before

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -38,41 +38,41 @@ internal class ClusterUserAction
...
+
+            /// <summary>
+            /// TBD
+            /// </summary>
+            /// <param name="other">TBD</param>
             /// <returns>TBD</returns>
-            public override bool Equals(object obj)
+            protected bool Equals(BaseClusterUserAction other)
             {
-                var baseUserAction = (BaseClusterUserAction)obj;
```
alexvaluyskiy: This could cause an exception in some serializers

Aaronontheweb:
Merge conflicts with the most recent DData changes we pulled in

PR #2592, opened 2017-04-09
Allow success/failure message creation for Tasks that doesn't return …

Silv3rcircl3:
…a value


src/core/Akka/Actor/PipeToSupport.cs
```diff
@@ -35,8 +35,8 @@ public static Task PipeTo<T>(this Task<T> taskToPipe, ICanTell recipient, IActor
             {
                 if (tresult.IsCanceled || tresult.IsFaulted)
                     recipient.Tell(failure != null
-                        ? failure((Exception)tresult.Exception)
-                        : new Status.Failure((Exception)tresult.Exception), sender);
```
Silv3rcircl3: type cast is redundant

PR #2593, opened 2017-04-10
PersistenceFSM with snapshots

alexvaluyskiy:
It was updated to the latest JVM sources. Passes all specs and supports snapshotting.
Unfortunately, it is not compatible with the previous version. Now, every state class should implement `IFsmState` interface.
The main difference from JVM version, it reuses the most of the types from Akka.FSM, while JVM version has created the same classes twice


src/core/Akka.Persistence/Fsm/PersistentFSM.cs
```diff
@@ -5,96 +5,405 @@
...
             }
         }
     }
+
+    public static class PersistentFSM
+    {
+        /// <summary>
+        /// IFsmState interface, makes possible for simple default serialization by conversion to String
+        /// </summary>
+        public interface IFsmState
```
alexvaluyskiy: Each `State` class should implement this interface

src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
```diff
@@ -782,119 +783,73 @@ namespace Akka.Persistence
...
-        protected abstract void OnRecoveryCompleted();
+        protected abstract TData ApplyEvent(TEvent domainEvent, TData currentData);
+        protected override void ApplyState(Akka.Persistence.Fsm.PersistentFSM.State<TState, TData, TEvent> nextState) { }
+        protected virtual void OnRecoveryCompleted() { }
         protected override bool ReceiveRecover(object message) { }
+        public void SaveStateSnapshot() { }
     }
     public abstract class PersistentFSMBase<TState, TData, TEvent> : Akka.Persistence.PersistentActor, Akka.Routing.IListeners
     {
-        protected Akka.Persistence.Fsm.PersistentFSMBase<TState, TData, TEvent>.State _currentState;
```
alexvaluyskiy: I hide a lot of internal stuff.

src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
```diff
@@ -782,119 +783,73 @@ namespace Akka.Persistence
...
         public Akka.Routing.ListenerSupport Listeners { get; }
         public TData NextStateData { get; }
         public TData StateData { get; }
         public TState StateName { get; }
-        protected virtual void ApplyState(Akka.Persistence.Fsm.PersistentFSMBase<TState, TData, TEvent>.State upcomingState) { }
+        protected System.Collections.Generic.IEnumerable<TState> StateNames { get; }
+        protected virtual void ApplyState(Akka.Persistence.Fsm.PersistentFSM.State<TState, TData, TEvent> nextState) { }
         public void CancelTimer(string name) { }
-        public Akka.Persistence.Fsm.PersistentFSMBase<TState, TData, TEvent>.State GoTo(TState nextStateName) { }
-        public Akka.Persistence.Fsm.PersistentFSMBase<TState, TData, TEvent>.State GoTo(TState nextStateName, TData stateData) { }
```
alexvaluyskiy: Removed this method, because it should not exist

src/core/Akka.Persistence/Fsm/PersistentFSM.cs
```diff
@@ -5,96 +5,405 @@
...
+            /// </summary>
+            public TD Data { get; }
+
+            /// <summary>
+            /// FSM state timeout.
+            /// </summary>
+            public TimeSpan? Timeout { get; }
+        }
+
+        public class State<TS, TD, TE>
```
alexvaluyskiy: Should I implement `Equals` here?

src/core/Akka.Persistence.Tests/Fsm/PersistentFSMSpec.cs
```diff
@@ -522,16 +675,63 @@ protected override bool ReceiveCommand(object message)
...
+    public class Paid : IUserState
+    {
+        public static Paid Instance { get; } = new Paid();
+
+        private Paid() { }
+        public override bool Equals(object obj) => !ReferenceEquals(obj, null) && obj is Paid;
+        public override int GetHashCode() => nameof(Paid).GetHashCode();
+        public override string ToString() => nameof(Paid);
+
+        public string Identifier => "Paid";
```
alexvaluyskiy: Each state class should implement `Identifier` property. It is something like `persistenceId`

src/core/Akka.Persistence.Tests/Akka.Persistence.Tests.csproj
```diff
@@ -49,6 +55,7 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
+    <Reference Include="System.Xml.Linq" />
```
Horusiath: Is it needed?

src/core/Akka.Persistence.Tests/Fsm/PersistentFSMSpec.cs
```diff
@@ -189,87 +219,144 @@ public void PersistentFSM_should_execute_the_defined_actions_following_successfu
...
         [Fact]
-        public void PersistentFSM_should_not_trigger_onTransition_for_stay()
+        public void PersistentFSM_must_not_trigger_onTransition_for_stay()
         {
-            var reportActorProbe = CreateTestProbe(Sys);
+            var probe = CreateTestProbe(Sys);
+            var fsmRef = Sys.ActorOf(SimpleTransitionFSM.Props(Name, probe.Ref));
 
-            var fsmRef = Sys.ActorOf(Props.Create(() => new SimpleTransitionFSM(Name, reportActorProbe.Ref)));
+            probe.ExpectMsg("LookingAround -> LookingAround", 3.Seconds()); // caused by initialize(), OK
```
Horusiath: Is this kin of string-based complex type comparison really what we want?

src/core/Akka.Persistence.Tests/Fsm/PersistentFSMSpec.cs
Horusiath: This will be a irritating for people, who are using state types defined in libraries with no Akka dependency. Maybe we could provide something like:

```csharp
interface IIdentity<TState>
{
    string GetIdentifier(TState state);
}
```
It's like `IEqualityComparer<>` (so we can supply our own impl of it even thou each object already implements some GetHashCode and Equals method).

src/core/Akka.Persistence/Fsm/PersistentFSM.cs
Horusiath: Probably not, this class has a mix of custom generic types, `IReadOnlyList<object>` and Action delegates, none of which is guaranteed to provide structural equality.

src/core/Akka.Persistence/Fsm/PersistentFSMBase.cs
```diff
@@ -391,877 +283,400 @@ protected internal void Initialize()
...
 
-            /// <summary>
-            /// TBD
-            /// </summary>
-            public void Cancel()
+            var timer = message as FSMBase.Timer;
+            if (timer != null)
             {
-                if (!_ref.IsCancellationRequested)
+                if (_timers.ContainsKey(timer.Name) && _timers[timer.Name].Generation == timer.Generation)
```
Horusiath: This will run 2 dictionary scans. `_timers.TryGetValue` would be better.

PR #2606, opened 2017-04-12
Reactive behavior trees akka workflows

dpavlenkov:
Translate workflow chart into akka actor. Let me know if there's a chance of this making it into core, and if not, what would you recommend.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/PublishSubscribe/DistributedPubSubMediatorSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/PublishSubscribe/DistributedPubSubMediatorSpec.cs
@@ -37,6 +37,7 @@ namespace Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe
             CommonConfig = ConfigurationFactory.ParseString(@"
                 akka.loglevel = INFO
                 akka.actor.provider = ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""
+                akka.actor.serialize-messages = off
                 akka.remote.log-remote-lifecycle-events = off
                 akka.cluster.auto-down-unreachable-after = 0s
                 akka.cluster.pub-sub.max-delta-elements = 500
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
@@ -15,7 +15,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
     /// <summary>
     /// TBD
     /// </summary>
-    public class DistributedPubSubSettings
+    public sealed class DistributedPubSubSettings : INoSerializationVerificationNeeded
     {
         /// <summary>
         /// Creates cluster publish/subscribe settings from the default configuration `akka.cluster.pub-sub`.
--- a/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxySettings.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxySettings.cs
@@ -14,7 +14,7 @@ namespace Akka.Cluster.Tools.Singleton
     /// <summary>
     /// TBD
     /// </summary>
-    public sealed class ClusterSingletonProxySettings
+    public sealed class ClusterSingletonProxySettings : INoSerializationVerificationNeeded
     {
         /// <summary>
         /// TBD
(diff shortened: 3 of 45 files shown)


Horusiath:
@dpavlenkov could you provide some more description? Things like motivation behind this one, maybe with examples?

I think that this may be nice feature, but I'm not sure if it should be inside core library, maybe as a separate plugin.

dpavlenkov:
This is the alternative to FSM. I've used both in projects. With FSM, state is explicit and logic is hidden. With BT, logic is explicit (in workflows) and state is implicit (in the path that is chosen). I'll come up with examples later.

dpavlenkov:
I like a few things about my implementation:

1) Easy to encode HFSM. Take classical case: when a toy is off, it can't do much, but when it's on, it can do a few things: bark, jump, sleep. But when it's asleep, it can't bark or jump, so under On, we can have two states Awake, or Sleep. But toy must turn off in any of the On states. BT controls this through Parallel/Spawn/Become constructs. Parallel allows two workflows execute simultaneously (conceptually, because everything is still single thread on callback). And Spawn limits the Become scope:

https://github.com/dpavlenkov/akka.net/blob/feature-bt/src/core/Akka.Tests/Actor/BTActorSpec.cs#L540-L582
```csharp
        public class ToyTest : BT<object>
        {
            public abstract class Message
            {
                public class ConnectedMsg : Message { }
                public class DisconnectedMsg : Message { }
                public class DoBark : Message { }
                public class DoSleep : Message { }
                public class DoJump : Message { }
                public class DoWake : Message { }
            }

            public ToyTest()
            {
                StartWith(Disconnected(), null);
            }

            TreeMachine.IWorkflow Disconnected() =>
                Receive<Message.ConnectedMsg>(
                    Become(Connected)); // Become takes in a Func, in order avoid cycles in graph

            TreeMachine.IWorkflow Connected() =>
                Parallel(ss => ss.AllComplete(), // AllComplete irrelevant when Become overwrites the whole workflow
                    Receive<Message.DisconnectedMsg>(
                        Become(Disconnected)),
                    Spawn(Awake())); // Using Spawn to limit Become scope, Becomes under Spawn overwrite Spawn child
                                     // Spawns still belong to their parent, and will be overwritten along with parent
            TreeMachine.IWorkflow Awake() =>
                AllComplete(
                    Execute(_ => Sender.Tell("YAWN")),
                    Forever( // Forever until Become
                        Parallel(ss => ss.AnySucceed(), // Parallel to listen to several message types/conditions
                            Receive<Message.DoBark>(Execute(ctx => Sender.Tell("BARK"))), // Receive is a single callback
                            Receive<Message.DoJump>(Execute(ctx => Sender.Tell("WHEE"))), // use loops to receive again
                            Receive<Message.DoSleep>(Become(Sleeping)))));

            TreeMachine.IWorkflow Sleeping() =>
                AllComplete(
                    Execute(_ => Sender.Tell("ZZZZ")),
                    Receive<Message.DoWake>( // Ignoring all messages besides DoWake
                        Become(Awake))); // But the Parallel in Connected also has a DisconnectedMsg listener
        }
```
https://github.com/dpavlenkov/akka.net/blob/feature-bt/src/core/Akka.Tests/Actor/BTActorSpec.cs#L1031-L1065
```csharp
        [Fact]
        public void BTActor_ToyTest()
        {
            var bt = Sys.ActorOf(Props.Create(() => new ToyTest()));

            Sys.EventStream.Subscribe(TestActor, typeof(UnhandledMessage));

            Action<object> assertUnhandled = s =>
            {
                bt.Tell(s, TestActor);
                ExpectMsg<UnhandledMessage>(m => m.Message.Equals(s));
            };

            Action<object, object> assertReply = (s, r) =>
            {
                bt.Tell(s, TestActor);
                ExpectMsg(r);
            };

            assertUnhandled("Hey");
            assertUnhandled(new ToyTest.Message.DoBark());
            assertReply(new ToyTest.Message.ConnectedMsg(), "YAWN");
            assertUnhandled(new ToyTest.Message.ConnectedMsg());
            assertUnhandled(new ToyTest.Message.DoWake());
            assertReply(new ToyTest.Message.DoBark(), "BARK");
            assertReply(new ToyTest.Message.DoJump(), "WHEE");
            assertReply(new ToyTest.Message.DoSleep(), "ZZZZ");
            assertUnhandled(new ToyTest.Message.DoBark());
            assertReply(new ToyTest.Message.DoWake(), "YAWN");
            assertReply(new ToyTest.Message.DoBark(), "BARK");

            bt.Tell(new ToyTest.Message.DisconnectedMsg(), TestActor);
            assertUnhandled(new ToyTest.Message.DoBark());
            assertReply(new ToyTest.Message.ConnectedMsg(), "YAWN");
        }
```
2) Being reactive, the tree doesn't need to be traversed every time, only active nodes need to be invoked.
3) The workflow interpreter, TreeMachine, can be made agnostic of its host: all it needs is that it's ProcessMessage be called and a scheduling facility. But still, Akka made it easy to test and debug and use, hence the BT actor.

dpavlenkov:
This seems like it's still a work and progress. I'm fine with closing this request, would just like your thoughts.

Horusiath:
@dpavlenkov sorry, but I hadn't much time lately to focus on your proposal. From the first hindsight it looks interesting, but I think it would be better to move it into separate package for time being, to evaluate its potential in context of existing "Become" behavior. If it would incubate and get warm welcome from the community, we could eventually merge it into core.

dpavlenkov:
I put this in https://github.com/dagmatic/akka.net . Actor become and BT become have similar intent, but different signatures.

Issue #2595, opened 2017-04-10
TcpEchoService.Server stops receiving communication if logging is used.


spankr:
We noticed a weird situation where after we closed 2 connections to our tcp server, the 3rd would connect but not handle any of the incoming communication.

I took the TcpEchoService.Server project ( *akka.net/src/examples/TcpEchoService.Server/* from master) and was able to recreate what we are seeing. It isn't a consistent number of times before it stops working.
It seems to be when the ```Console.WriteLine()``` comments are replaced with ```_log.Info()``` in the Echo actors.

I also added a Hocon configuration section to my app.config.

[logging-changes.zip](https://github.com/akkadotnet/akka.net/files/910998/logging-changes.zip)

My input:
```
$ telnet localhost 9001
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
test
test
exit
$ telnet localhost 9001
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
test
test
exit
Connection closed by foreign host.
$ telnet localhost 9001
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
test   <-- This doesn't get echoed back
exit   <-- This doesn't close the connection
```

My log (some repetitive logs trimmed out)
```
TCP server is listening on *:9001
ENTER to exit...
[DEBUG][4/10/2017 6:08:47 PM][Thread 0001][EventStream(echo-server-system)] Logger log1-DefaultLogger [DefaultLogger] started
[DEBUG][4/10/2017 6:08:47 PM][Thread 0001][EventStream(echo-server-system)] StandardOutLogger being removed
[DEBUG][4/10/2017 6:08:47 PM][Thread 0001][EventStream(echo-server-system)] Default Loggers started
[WARNING][4/10/2017 6:08:47 PM][Thread 0001][ActorSystem(echo-server-system)] NewtonSoftJsonSerializer has been detected as a default serializer. It will be obsoleted in Akka.NET starting from version 1.5 in the favor of Hyperion
[INFO][4/10/2017 6:08:55 PM][Thread 0006][[akka://echo-server-system/user/echo-service#150159752]] Remote address [::ffff:127.0.0.1]:61159 connected
[INFO][4/10/2017 6:08:59 PM][Thread 0004][[akka://echo-server-system/user/echo-service/$a#884462836]] Received 'test' from remote address [[::ffff:127.0.0.1]:61159]
[...]

PR #2615, opened 2017-04-13
cleaning up read & write sockets happens consistently when a NoSocket…

spankr:
This is the change I made (locally) to resolve issue #2595 and handle the SocketException cleanup.

The original logic, on exception, simply removes all the non-connected sockets from the dictionaries....only when the SocketException is of NotSocket. (other SocketExceptions just eat the exception)

This proposed change will allow the associated channels to be notified of the socket removal when a SocketException (NotSocket) is thrown. Other SocketExceptions will now bubble up the chain.

I could not find any existing tests for this logic so I'm not entirely sure how to test this under CI.  I do know that this change is working for our team now. We can receive repeated incoming connections with this change. (@jaydeboer is a team member)


Aaronontheweb:
I'm going to go ahead and merge this and get it included as part of the next maintenance release. We might bring in our SocketAsyncEventArgs changes that could wipe these ones out potentially, but this is still a good change as far as I can tell.

spankr:
I was looking through a related issue in the milestone and it looks like the akka.io branch will most likely resolve this issue, once it makes it out. That seems to be a substantial rewrite of the IO code. 

Either way, it looks like our issue will go away and that's a win for us!

Issue #2610, opened 2017-04-13
Random Mono test failure: Akka.Configuration.ConfigurationException : Logger [Akka.Event.DefaultLogger] specified in config cannot be loaded


Aaronontheweb:
Seen this come up on many different tests lately:

```
Akka.Configuration.ConfigurationException : Logger [Akka.Event.DefaultLogger] specified in config cannot be loaded: System.AggregateException: One or more errors occurred. ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   --- End of inner exception stack trace ---
  at System.Threading.Tasks.Task.ThrowIfExceptional (System.Boolean includeTaskCanceledExceptions) [0x00014] in <dca3b561b8ad4f9fb10141d81b39ff45>:0 
  at System.Threading.Tasks.Task`1[TResult].GetResultCore (System.Boolean waitCompletionNotification) [0x00034] in <dca3b561b8ad4f9fb10141d81b39ff45>:0 
  at System.Threading.Tasks.Task`1[TResult].get_Result () [0x0000b] in <dca3b561b8ad4f9fb10141d81b39ff45>:0 
  at Akka.Event.LoggingBus.AddLogger (Akka.Actor.Internal.ActorSystemImpl system, System.Type loggerType, Akka.Event.LogLevel logLevel, System.String loggingBusName, System.TimeSpan timeout) [0x00042] in <2f7572f625f74ec0b489853bc1ea74e6>:0 
  at Akka.Event.LoggingBus.StartDefaultLoggers (Akka.Actor.Internal.ActorSystemImpl system) [0x000a5] in <2f7572f625f74ec0b489853bc1ea74e6>:0 
---> (Inner Exception #0) System.Threading.Tasks.TaskCanceledException: A task was canceled.<---

---- System.AggregateException : One or more errors occurred.
-------- System.Threading.Tasks.TaskCanceledException : A task was canceled.
  at Akka.Event.LoggingBus.StartDefaultLoggers (Akka.Actor.Internal.ActorSystemImpl system) [0x000cd] in <2f7572f625f74ec0b489853bc1ea74e6>:0 
  at Akka.Actor.LocalActorRefProvider.Init (Akka.Actor.Internal.ActorSystemImpl system) [0x000e1] in <2f7572f625f74ec0b489853bc1ea74e6>:0 
[...]

Issue #2589, opened 2017-04-07
Issue running unit tests on Debian CI builds using Mono


heynickc:
We've noticed some test failures running unit tests on Mono in our CI system.  The test failures are inconsistent, with tests failing that have not failed in a long time and likely shouldn't fail.

We're seeing the build logs throwing this error, `W: Conflicting distribution: http://download.mono-project.com wheezy/snapshots/4.6.2.16 InRelease (expected wheezy/snapshots but got wheezy)` which might hint at a root cause in our Docker-based CI, but we're still investigating.

Mostly a heads up to contributors that this seems to be more of an infrastructure issue and we're addressing it.

Aaronontheweb:
@heynickc didn't we determine that this wasn't an issue after all?

heynickc:
We didn't confirm that it was or wasn't.  We don't have a root cause still, but at the time I put this issue in that was one an error noticed in the logs.

heynickc:
@Aaronontheweb I see now what you were asking - whether this entire issue *was in fact an issue.  The test failures are still prevalent, but I'm not sure this is an infrastructure issue.  Maybe a Mono-specific issue but really there's no clear root cause.

alexvaluyskiy:
Right now we have completely different CI. It is not an issue anymore

Issue #2558, opened 2017-03-16
Akka.Testkit.Xunit2 does not support Xunit 2.2.0


jaydeboer:
When using `ExpectMsg` in a test when using Xunit 2.2.0, the test fails with the following message:

```
Message: System.IO.FileLoadException : Could not load file or assembly 'xunit.assert, Version=2.1.0.3179, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```

Downgrading Xunit to 2.1.0 will correct the issue.

Example of a test that will cause an issue.

```
        [Fact]
        public void DoTheRightThings()
        {
            var expectedWidget = new Widget();
            var actor = ActorOfAsTestActorRef<MyActor>(MyActor.Props());
            actor.UnderlyingActor.library = new List<Widget>() { expectedWidget };

            actor.Tell(new FetchWidgets());

            ExpectMsg<Widget[]>(a => a.Length==1 && a[0].Id.ToInt() == expectedWidget .Id.ToInt());
        }
```

spankr:
Would setting the **SpecificVersion** to _False_ for the xunit dependencies in the .csproj resolve this?

Aaronontheweb:
cc @alexvaluyskiy IMHO, we should probably just upgrade our XUnit version as part of 1.2

PR #2617, opened 2017-04-14
[WIP] Update Akka.TestKit.Xunit2 from xUnit 2.1.0 to xUnit 2.2.0

heynickc:
Second attempt at upgrading xUnit from 2.1.0 to 2.2.0.  Requirement for v2.2.0 is .NET 4.5.2 so all test projects and TestKits needed to be upgraded to .NET 4.5.2.  All other production code remains at .NET 4.5.  This is an attempt at resolving a couple issues #2558 [Akka.TestKit.Xunit2 does not support 2.2.0] and our issue lately in CI with running tests on Mono #2610 [Random Mono test failures] and #2589.  This also upgrades the xUnit runner itself to v2.2.0


Aaronontheweb:
@heynickc I'm fine with requiring 4.5.2 for testkits going forward. Anyone else take issue with that?

Issue #2628, opened 2017-04-21
Not invoked IHandle Implementation to Explicit Interface Implementation 


FeyverkSoft:
#AKKA 1.2.0.33
Not invoked IHandle Implementation to Explicit Interface Implementation 
DEMO
```
using System;
using System.Threading.Tasks;
using Akka.Actor;

namespace ConsoleApp1
{
    static class Program
    {
        static void Main(string[] args)
        {
            Console.WriteLine(new ss().De().Result);
            Console.ReadLine();
        }
    }

    public class ss
    {
        private readonly IActorRef act;

        public ss()
        {
            var sys = ActorSystem.Create("demo");
            act = sys.ActorOf<Test>();
        }

        public async Task<string> De()
        {
            return await act.Ask<string>("demo");
        }

    }
    public class Test : TypedActor, IHandle<string>
    {
#warning Not invoked 
        void IHandle<string>.Handle(string message)
        {
            Sender.Tell(message);
        }
    }
}

```

Szer:
The problem is here (TypedActor.cs):
MethodInfo method = GetType().GetMethod("Handle", new[] {message.GetType()});

In case you are implementing IHandle explicitly the line above will return null.
I've made pull request with some tests

FeyverkSoft:
Thank you :-)

Danthar:
Closed, See PR

PR #2632, opened 2017-04-24
fixed bug with non invokable explicit IHandle implementation

Szer:
fixed bug with non invokable explicit IHandle implementation in TypedActor (#2628)
Added few tests for TypedActor.

I maybe miss something but it seems TypedActor not covered with tests at all, is it?

### DIFF
--- a/src/core/Akka.Tests/Actor/ReceiveActorTests_Typed.cs
+++ b/src/core/Akka.Tests/Actor/ReceiveActorTests_Typed.cs
@@ -1,0 +1,89 @@
+using Akka.Actor;
+using Akka.Event;
+using Xunit;
+
+namespace Akka.Tests.Actor
+{
+    public partial class ReceiveActorTests
+    {
+        [Fact]
+        public void Given_typed_actor_with_no_receive_specified_When_receiving_message_Then_it_should_be_unhandled()
+        {
+            //Given
+            var system = ActorSystem.Create("test");
+            var actor = system.ActorOf<NoReceiveTypedActor>("no-receive-specified");
+            system.EventStream.Subscribe(TestActor, typeof(UnhandledMessage));
+
+            //When
+            actor.Tell("Something");
+
+            //Then
+            ExpectMsg<UnhandledMessage>(m => ((string) m.Message) == "Something" && m.Recipient == actor);
+            system.EventStream.Unsubscribe(TestActor, typeof(UnhandledMessage));
+        }
+
+        [Fact]
+        public void
+            Given_typed_actor_with_two_implicit_handle_When_sending_different_messages_Then_correct_handler_should_be_invoked()
+        {
+            //Given
+            var system = ActorSystem.Create("test");
+            var actor = system.ActorOf<ReceiveStringAndIntImplicitTypedActor>("typedImplicitTwoHandle");
+
+            //When
+            actor.Tell(4711, TestActor);
+            actor.Tell("hello", TestActor);
+
+            //Then
+            ExpectMsg("int:4711");
+            ExpectMsg("string:hello");
+        }
+
+        [Fact]
+        public void
+            Given_typed_actor_with_two_explicit_handle_When_sending_different_messages_Then_correct_handler_should_be_invoked()
+        {
+            //Given
+            var system = ActorSystem.Create("test");
+            var actor = system.ActorOf<ReceiveStringAndIntExplicitTypedActor>("typedExplicitTwoHandle");
+
+            //When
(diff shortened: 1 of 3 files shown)


src/core/Akka/Actor/TypedActor.cs
```diff
@@ -34,7 +35,10 @@ public abstract class TypedActor : ActorBase
         /// <returns>TBD</returns>
         protected sealed override bool Receive(object message)
         {
-            MethodInfo method = GetType().GetMethod("Handle", new[] {message.GetType()});
+            Type[] msgTypeArray = new[] {message.GetType()};
+            Type actorType = GetType();
+            MethodInfo method = actorType.GetMethod("Handle", msgTypeArray) ??
+                                actorType.GetMethod($"Akka.Actor.IHandle<{msgTypeArray[0].FullName}>.Handle", BindingFlags.NonPublic | BindingFlags.Instance);
```
0x53A: This depends on the compiler naming explicit interface implementations like this, which is an implementation detail, and for example the F# compiler uses a different naming scheme afaik.

Szer: Didnt know that. I could replace this with only reflection API calls but as always with reflection it could be sloppy

0x53A: I would try something like this:
image

Horusiath: 1. This could potentially fail,  if message type is not inside the same assembly. In some cases fully qualified type name is not enough, you'll need assembly name as well.
2. This is a reflection call on each message received. Aside the fact that we (hopefully) will remove `TypedActor` in near future, its usage is heavily discouraged for this exact reason - reflection makes it order of magnitude slower than using `ReceiveActor`.
3. You could potentially patch it by constructing a dictionary of `Dictionary<Type, MethodInfo>` i.e. inside actor constructor. This way you could use reflection only once when actor is created or even try to move this to an actor static constructor.

src/core/Akka.Tests/Actor/ReceiveActorTests_Typed.cs
```diff
@@ -0,0 +1,89 @@
...
+            //Given
+            var system = ActorSystem.Create("test");
+            var actor = system.ActorOf<ReceiveStringAndIntExplicitTypedActor>("typedExplicitTwoHandle");
+
+            //When
+            actor.Tell(4711, TestActor);
+            actor.Tell("hello", TestActor);
+
+            //Then
+            ExpectMsg((object) "int:4711");
```
Horusiath: You don't need object cast here.

src/core/Akka/Actor/TypedActor.cs
Szer: Yep, already did it in my local branch:
https://github.com/Szer/akka.net/commit/a1e97b9a10f0cb5ff78e4c807a5e1d5e83fc74ec#diff-b4b64856bc2f68b9523d55f258b1ba6a

Szer:
I reworked TypedActor.cs now with global cache.
It worked like Dictionary <Type, Dictionary <Type, MethodInfo>> where outer Dictionary works like cache for each concrete TypedActor type.
Inner dictionary is cache for each message type within one concrete TypedActor type.

To prevent multiple creation of outer cache elements I've used one lock with double checking.
To prevent locking in inner cache dictionary I've used AddOrSet instead locking.
In case of race condition it will just overwrite cache instance once.

Please reveiw this again.

rogeralsing:
This can be written using multi methods.
It will also make sure that any subsequent calls re-use callsite information so that the call is faster than `MethodInfo.Invoke`

```csharp
    public abstract class TypedActor
    {
        public void Receive(object message)
        {
            try
            {
                ((dynamic) this).Handle((dynamic) message);
            }
            catch(RuntimeBinderException _)
            {
                //message of not supported type
            }
        }
    }
```

Horusiath:
@rogeralsing unfortunately `dynamic` doesn't work with explicit interface implementation signatures like `void IHandler<string>.Handle(string)`. This was the initial case for this PR.

Danthar:
Is this still being worked on? @Szer

Szer:
@Danthar Not sure it is needed.
If Akka.Net members say that this fix is something that should be implemented, I'll do it by Monday/Tuesday.

Danthar:
@Horusiath or @Aaronontheweb Since there has been no response from our side for almost a year now. Im closing this, and marking the corresponding issue as wont fix.

PR #2636, opened 2017-04-29
Persistence Clean-up 2

alexvaluyskiy:


src/core/Akka.Persistence/Eventsourced.Recovery.cs
```diff
@@ -9,55 +9,28 @@
...
+using Akka.Persistence.Internal;
 
 namespace Akka.Persistence
 {
-    /// <summary>
-    /// TBD
-    /// </summary>
-    /// <param name="receive">TBD</param>
-    /// <param name="message">TBD</param>
     internal delegate void StateReceive(Receive receive, object message);
```
alexvaluyskiy: No TDB for internal stuff

Issue #1675, opened 2016-01-27
Redesign the Dependency Injection API


Horusiath:
Right now our DI API is not sufficient to fluently operate with actors. The main issues here are:
1. Using `Props.DI().Create<InjectedActor>()` is probably not the nicest way to abstract the fact, that we're working with some kind of DI container. This should be transparent.
2. Current design doesn't allow us to make a selective injections - we either need to inject all or none of the actor parameters. This doesn't fit many real-life cases and should be definitely changed.
3. Think about some way of adding additional scoping rules to containers allowing resolved objects to have lifecycle pinned to a particular actor context/actor or one of it's ancestors.

Some ideas here are: 

**Ad.1)** Abstract DI container inside dedicated `ActorProducer`s, which will be aware of their role without explicit notifying using `Props.DI()`.

**Ad.2)** Allow to make a selective injections - I've actually have an idea here to create a feature inside `Akka.DI.Core` (which is shared by all DI plugins) i.e.: 

``` csharp
Props.Create(() => new MyActor(Context.Parent, DI.Inject<MyService>()))
```

`Context.Parent` is example of pinning existing value to a constructor. Here an `Inject` method could work two ways:
1. Simpler version: returns `null`/`default(T)`. DI-aware actor producer would recognize this as a point, where container should resolve a dependency.
2. More complex: returns a dynamic proxy replaced during actor incarnation with instance resolved by the container. This would be nicer for a reason, that dynamic proxy could convey some additional data informing about things like scoping or named/tagged parameters.

Before somebody mention it: I think, that part we shouldn't speak about scoping of actors themselves - akka has it's own lifecycle rules and none of DI libraries would fit into that.

rogeralsing:
@Horusiath the `Props.Create(() =>` uses Linq Expressions so it would be fairly easy to detect the `DI.Inject` code and replace that with resolved values.
No need for magic proxies or anything like that.

Danthar:
So I like idea number 1 because it makes DI more of an infrastructure concern instead of having it be in your face all the time.The concept of selective injections is great. Its something you will want to use any time your passing an IActorRef and DI injected services in an actors constructor.
Not sure if a proxy is the way to go with this. But anything that keeps DI as transparently as possible is good in my book.

About the scoping of dependencies. In my mind thats easy. You either have it scoped to the lifecycle of the actor its being injected in (which means its disposed/recreated on actor restart/creation). Or its singleton scoped.
[...]

PR #2638, opened 2017-04-30
[WIP] New actor creation pipeline

Horusiath:

This PR brings some initial work on the ideas I've started to talk about in #1675 . Esentially it's simplification of an existing actor creation pipeline: instead of ActorProducer/ActorProducerPipeline/Resolver classes we can have single `IDependencyResolver` class directly under `ActorSystem` itself. 

This also means that Akka.DI.Core is no longer needed. If user want ot setup it's own dependency resolver, he/she can set it up by calling `actorSystem.UseDependencyResolver(dependencyResolverImpl)` right after creating actor system.

This implementation also supports using dependency resolver for non-actor objects i.e. persistence EventAdapters. Ideally it would be great, if we could replace all `Activator.CreateInstance` calls with it.

Dependency resolver can be used for all actors **not** created via `Props.Create(() => new MyActor(x, y, z))` as we know, that those actors are created totally correctly without it. It will be used for other types of props i.e.: you no longer need `Props.DI().Create<MyActor>()` as `Props.Create<MyActor>()` will rely on dependency resolver anyway.

There is also a support for partial DI application, so a situation when user may want to specify some of the arguments, while leaving the rest of them in a hands of a resolver. Example:

```csharp
class MyActor : ActorBase
{
    public MyActor(IRepository injectedRepo, IActorRef manualRef) { }
}

Context.ActorOf(Props.Dynamic<MyActor>(new { manualRef }));
```
This will take anonymous object and by using reflection will try to match parameters provided as object properties by name and type (name and type must be an exact match at this point). Missing constructor parameters will be supplied from resolver.

Only regression at this point is that components created by resolver are not allowed to contain null values, as this is the way DR gets informed that it's expected to fill the missing parameter.

### TODOs

- [x] fix akka.remote, akka.cluster and akka.persistence tests failures
- [ ] apply dependency resolver for actor system extensions construction
- [x] apply dependency resolver for serializers construction
- [ ] copy existing type registrations to new dependency resolver when `UseDependencyResolver` will be used to change it.


Horusiath:
I'll send another PR for that.

Issue #2642, opened 2017-05-03
Update F# version


tjaskula:
Given the .NET Framework version is 4.5, it would be nice to upgrade the current F# version `4.3.0.0` to `4.4.0.0` in F# related projects.

tjaskula:
I can work on that one

tjaskula:
this was solved with the following commit 04b71c6592ec789afb8e00a7f8595aac530d8f32

PR #2643, opened 2017-05-03
Update F# projects to the last FSharp.Core version (4.4.1.0)

tjaskula:
+ FSharp.Core dependency comes from nuget
+ Getting rid of <TargetFSharpCoreVersion>4.3.1.0</TargetFSharpCoreVersion> tag as it's not used inside the `fsproj` and it's making a confusion

`System.ValueTuple` is brought along with the latest FSharp.Core.dll


tjaskula:
Ok, I reproduce the failing test and error locally. I'll work on it from there

Horusiath:
@tjaskula perf test may need more than one run. Since they are sort-of dependent on actual state of the underlying hardware, they can occasionally return false negatives.

tjaskula:
@Horusiath Cool, it seems that it passes now. I thought that the failling test was unrelated to the build error we had above

tjaskula:
#2642

PR #2654, opened 2017-05-10
[WIP] Limits the number of concurrent Persistent actors during recovery

ismaelhamed:
This is a port of the original [#22783](https://github.com/akka/akka/issues/22638). The main idea is to limit the number of concurrent persistent actors to avoid overwhelming both the persistent journal and the backend store.

- [x] ~~core implementation~~
- [ ] unit tests


Danthar:
Looks good. One request though could you make this PR to the  v1.3 branch. 
This is our new 'dev' branch untill we release it. We are doing it this way to ensure all contributions are going to be .netstandard 1.6 compatible.

ismaelhamed:
Sure thing.

PR #2662, opened 2017-05-11
Update Akka.Streams to 2.4.12

Silv3rcircl3:
## 2.4.12 (01.10.2016 -  28.10.2016)
- [x]  [Convert Rechunker to GraphStage](https://github.com/akka/akka/pull/21083)
- [x]  [StageLogging that allows logger access in stages](https://github.com/akka/akka/pull/21696)
- [x]  [Rewrite Sink.ignore as a GraphStage](https://github.com/akka/akka/pull/21728)
- [x]  [improve streams actor integration docs](https://github.com/akka/akka/pull/21700)
- [x]  [IOResult construction exposed, in bincompat way](https://github.com/akka/akka/pull/21070)
- [x]  [Increased timeout in long running test FramingSpec](https://github.com/akka/akka/pull/21686)
- [x]  [Optimized indexOf to speed up framing stage](https://github.com/akka/akka/pull/21539)
- [x]  [Added a flag on takeWhile to allow it to include the final element](https://github.com/akka/akka/pull/21687)
- [x]  [Add expectNextChainingPF method to streams testkit](https://github.com/akka/akka/pull/21617)
- [x]  [Add scanAsync to Akka Streams, similar to scan but taking a Future](https://github.com/akka/akka/pull/21553)
- [x]  [Run fold async does not resolve with zero](https://github.com/akka/akka/pull/21659)
- [x]  [add zipWithIndex to FlowOps](https://github.com/akka/akka/pull/21435)
- [x]  [amending mapAsyncUnordered docs to include parallelism](https://github.com/akka/akka/pull/21609)


src/core/Akka.Streams.TestKit.Tests/StreamTestKitSpec.cs
```diff
@@ -151,6 +151,59 @@ public void TestSink_Probe_ExpectNextOrComplete_with_completion()
...
+        public void TestSInk_Probe_ExpectNextChaining_should_pass_with_right_element()
+        {
+            Source.Single(1)
+                .RunWith(this.SinkProbe<int>(), _materializer)
+                .Request(1)
+                .ExpectNextChaining<int>(i => i == 1);
+        }
+
+        [Fact]
+        public void TestSInk_Probe_ExpectNextChaining_should_allow_to_chain_test_methods()
```
alexvaluyskiy: Bad naming

Silv3rcircl3:
@akkadotnet/core This is now ready for review. Doc changes are handled in #2684

src/core/Akka.Streams/Stage/IStageLogging .cs
```diff
@@ -0,0 +1,26 @@
...
+    /// If used with a different materializer <see cref="NoLogger"/> will be returned.
+    ///
+    /// Make sure to only access `Log` from GraphStage callbacks (such as `Pull`, `Push` or the async-callback).
+    ///
+    /// Note, abiding to <see cref="Attributes.LogLevels"/> has to be done manually,
+    /// the logger itself is configured based on the logSource provided to it. Also, the `Log`
+    /// itself would not know if you're calling it from a "on element" context or not, which is why
+    /// these decisions have to be handled by the stage itself.
+    /// </summary>
+    public interface IStageLogging
```
alexvaluyskiy: Do we have something like Java's `GraphStageLogicWithLogging`?

alexvaluyskiy:
@Silv3rcircl3 I don't see the docs about StageLogging here (only in getakka.net).

src/core/Akka.Streams/Dsl/Sink.cs
```diff
@@ -270,8 +270,7 @@ public static class Sink
         /// </summary>
         /// <typeparam name="TIn">TBD</typeparam>
         /// <returns>TBD</returns>
-        public static Sink<TIn, Task> Ignore<TIn>()
-            => new Sink<TIn, Task>(new SinkholeSink<TIn>(Shape<TIn>("BlackholeSink"), DefaultAttributes.IgnoreSink));
+        public static Sink<TIn, Task> Ignore<TIn>() => FromGraph(new IgnoreSink<TIn>());
```
alexvaluyskiy: This method should return Sink<TIn, Task<Done>>
See https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/scaladsl/Sink.scala#L197

alexvaluyskiy: We don't return `Task<Done>>` here and I should write 
`RunWith(Sink.Ignore<int>(), Materializer) .PipeTo(TestActor, success: () => Done.Instance)` instead of `RunWith(Sink.Ignore<int>(), Materializer) .PipeTo(TestActor)`

src/core/Akka.Streams/Stage/IStageLogging .cs
Silv3rcircl3: Problem with this one is that the trait has a default implementation. So this is implemented by the GraphStageLogic directly and I skipped the GraphStageLogicWithLogging because it's lazy and doesn't do anything until you call the Log property.

src/core/Akka.Streams/Dsl/Sink.cs
Silv3rcircl3: I think Done only exists here because they always need to return something, like in F#, so we skipped it because it doesn't contain any useful information and returning Task without T is common in C#.

Silv3rcircl3:
I will add the docs, haven't noticed that the docs are now in the main repo while writing them for the other repo.

src/core/Akka.Streams/Dsl/Sink.cs
alexvaluyskiy: @Silv3rcircl3 I think we should create an article in the docs about all differences in Streams API. Sometimes I think it is a bug, but it was a design decision :)

Silv3rcircl3: That would probably take a while :-) and it only helps people that are already familiar with the scala API, but I see the point from a maintainers perspective. 
Maybe we could do it incremental, write down every difference we see while working on/with it?

Silv3rcircl3:
@alexvaluyskiy logging docu is already in the core repo, you've created the PR 😄 
see https://github.com/akkadotnet/akka.net/pull/2676/files#diff-b2a35eb7d463827cec074803d02b1635

alexvaluyskiy:
@Silv3rcircl3 Oops

PR #2666, opened 2017-05-11
[WIP] Replace dotnet test with dotnet xunit (v2.3.0)

heynickc:
Try to use consistent xunit runner for Linux and Windows that will also run against `net452;netcoreapp1.1` test assemblies.  This was introduced in https://www.nuget.org/packages/dotnet-xunit/2.3.0-beta2-build3683.

It also allows us to pass xunit arguments to `dotnet xunit` the same way it was being done before, e.g. `-parallel none -teamcity`

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -64,12 +64,25 @@ Target "Build" (fun _ ->
 //--------------------------------------------------------------------------------
 
 Target "RunTests" (fun _ ->
-    let projects = !! "./**/core/**/*.Tests.csproj"
-                   ++ "./**/contrib/**/*.Tests.csproj"
-                   -- "./**/Akka.Streams.Tests.csproj"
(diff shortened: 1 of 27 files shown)


Aaronontheweb:
@heynickc looks like we're down to just issues with the LocalSnapshotStore on Linux?

heynickc:
@Aaronontheweb Probably a bit more than that - based on the ordering of the tests collected by the runner, the LocalSnapshotStore fail after 199 passing tests so there's some more to be run.  E.g. `Akka.MultiNodeTestRunner.Shared.Tests` fail too.  They are at least running with the new runner, so I think that's an improvement.

alexvaluyskiy:
@heynickc `CoreAPISpec` should be ignored on Mono

heynickc:
@alexvaluyskiy will do, thanks

PR #2669, opened 2017-05-12
[api-docs] Update docfx config to support new proj

sean-gilliam:
This PR updates the docfx configuration file to support .netStandard
project files.


docs/docfx.json
```diff
@@ -23,7 +23,10 @@
           "src": "../src"
         }
       ],
-      "dest": "api"
+      "dest": "api",
+	  "properties": {
+        "TargetFramework": "netstandard1.6"
```
alexvaluyskiy: Why `netstandard`? Not net45?

sean-gilliam: Unless I'm missing something, we don't have different public APIs between ```net45``` and ```netstandard```. So I chose the later.

alexvaluyskiy: We have. For example right now we dont have Akka.IO on netcore. We don't have some members related to `System.Configuration`

sean-gilliam:
Changed to TargetFramework net45, included missing projects, and formatted the json.

PR #2678, opened 2017-05-15
adding a basic in-memory snapshot store.

spankr:
Created the _MemorySnapshotStore_ in Akka.Persistence.
Hopefully it feels similar to _MemoryJournal_ when it comes to configuration and extensibility.

While in there, I noticed a small inconsistency within _MemoryJournal.Update()_. It was returning **_messages** when similar methods were returning the **Messages** collection. This inconsistency would impact _SharedMemoryJournal_ and any other derived journals.


spankr:
I am curious as to how or why the [LocalSnapshotStore test](http://petabridge-ci.cloudapp.net/viewLog.html?buildId=26455&buildTypeId=AkkaNet_AkkaNetWindowsUnitTests&guest=1) failed. I don't believe I modified anything that would impact those tests. (test also passes locally)

Danthar:
We have some racy tests at the moment. This could be the cause here as well.

spankr:
I shall hope that its a racy issue until I hear otherwise.

Danthar:
Ok, so the failing unit test on linux-mono is unrelated to this change, i restarted that build.

Question: Is this persistence plugin now automatically used when no other snapshot store is configured ? Seems like a good thing to do considering we do the same thing for the journal-store.

Aaronontheweb:
@spankr thanks for your contribution!

spankr:
@Danthar I'd agree that it would be a good default snapshot store. I felt that decision was better left to the core Akka team and not a new guy. :smile:

PR #2681, opened 2017-05-18
Added NetStandard1.6 support for Sqlite plugin

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/issues/2682

Now it uses `Microsoft.Data.Sqlite 2.0.0-preview1` package instead of `System.Data.SQLite.Core`. It supports NetCore, and much faster
```
System.Data.SQLite.Core 1.0.105 + Journal = 1035 events/sec
System.Data.SQLite.Core 1.0.105 + Batching Journal = 10548 events/sec

Microsoft.Data.Sqlite 1.1.0 + Journal = 4221 events/sec
Microsoft.Data.Sqlite 1.1.0 + Batching Journal = 14415 events/sec

Microsoft.Data.Sqlite 2.0.0-preview1 + Journal = 7028 events/sec
Microsoft.Data.Sqlite 2.0.0-preview1 + Batching Journal = 17550 events/sec
```
But it has a breaking change in the configuration
**Before**
```
FullUri=file:memdb-journal.db?mode=memory&cache=shared;Version=3;
```
**After**
```
Datasource=memdb-journal.db;Mode=Memory;Cache=Shared
```


src/benchmark/PersistenceBenchmark/App.config
```diff
@@ -3,4 +3,7 @@
     <startup> 
         <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5" />
     </startup>
+  <runtime>
+    <gcServer enabled="true"/>
```
Aaronontheweb: this is going to skew the benchmark (possibly) if we weren't using it before.

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Query/SqliteAllPersistenceIdsSpec.cs
```diff
@@ -15,7 +15,7 @@ namespace Akka.Persistence.Sqlite.Tests.Query
 {
     public class SqliteAllPersistenceIdsSpec : AllPersistenceIdsSpec
     {
-        public static readonly AtomicCounter Counter = new AtomicCounter(50);
+        public static readonly AtomicCounter Counter = new AtomicCounter(100);
```
Aaronontheweb: What's with all of the `AtomicCounter` value changes in these specs? Not criticizing, just wondering.

src/contrib/persistence/Akka.Persistence.Sqlite/ConnectionContext.cs
```diff
@@ -17,7 +17,7 @@ namespace Akka.Persistence.Sqlite
     /// </summary>
     internal static class ConnectionContext
     {
-        private static readonly ConcurrentDictionary<string, SQLiteConnection> Remembered = new ConcurrentDictionary<string, SQLiteConnection>();
+        private static readonly ConcurrentDictionary<string, SqliteConnection> Remembered = new ConcurrentDictionary<string, SqliteConnection>();
```
Aaronontheweb: We're breaking binary compatibility here anyway by virtue of switching to a different driver, plus we're trying to get this module out of beta... so I don't have an issue with this.

src/contrib/persistence/Akka.Persistence.Sqlite/README.md
```diff
@@ -41,8 +41,8 @@ In addition, journal configuration specifies additional field:
 
 Akka.Persistence.Sqlite plugin allows to use in-memory databases, however requires to use them in shared mode in order to work correctly. Example connection strings for such configurations are described below:
 
-- `FullUri=file::memory:?cache=shared;` for anonymous in-memory database instances.
-- `FullUri=file:<database-name>.db?mode=memory&cache=shared;` for named in-memory database instances. This way you can provide many separate databases residing in memory.
+- `Datasource=file;Mode=Memory` for anonymous in-memory database instances.
```
Aaronontheweb: Good.

Aaronontheweb: (Meaning, good that these changes appear in the docs somewhere)

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Query/SqliteAllPersistenceIdsSpec.cs
alexvaluyskiy: Just to avoid conflicts in filenames

src/benchmark/PersistenceBenchmark/App.config
alexvaluyskiy: I ran all three benchmarks with GcServer-true

PR #2683, opened 2017-05-18
Rework of Akka.IO (v1.3)

Horusiath:
Continued work started at #2405 - now targeting v1.3 / .NET Core branch. Remaining bugs/features to take a look on - probably they will be implemented in upcoming PRs, they shouldn't delay this one from merging:

- [x] Port https://github.com/akka/akka/pull/21685
- [ ] `WriteFile` support - at the present moment .NET Core doesn't support TransferFile API at all. We could implement that as copying filestream to socket, however this is not what transfer file is all about.
- [x] Harden behavior under high amount of concurrent actors: `HorizontalScaleSpec` shown that at some point (around 550-600 concurrent clients on my laptop) TcpManager is failing. I don't know the exact reason yet, but it's deterministic, maybe related to some limits in configuration. Also this amount is already better that the current one (which caused my machine to boil on that amount of actors).
- [x] Investigate `Socket.SendAsync` method with BufferList - as `ByteString` implements byte buffers as array of `ArraySegment<byte>`, maybe we don't need to send those buffers one by one, maybe we could just supply buffer enumeration up front and leave rest to the .NET runtime.
- [ ] Investigate zero-copy buffers - while writing is copy free from our side, reading from socket is copying operation: we copy `SocketAsyncEventArgs` buffer to new `ByteString` instance, as we cannot guarantee that user will free that resource and when he will do that. Maybe there is some way to avoid that.
- [ ] Investigate [Windows RIO](https://technet.microsoft.com/en-us/library/hh997032(v=ws.11).aspx) - we could provide a windows specific implementation of TCP and UDP connection actors that would take advantage of registered I/O.


build.fsx
```diff
@@ -9,7 +9,22 @@ open Fake
...
 
 // Variables
-let configuration = "Release"
+let product = "Akka.NET"
+let authors = [ "Akka.NET Team" ]
+let copyright = "Copyright © 2013-2017 Akka.NET Team"
+let company = "Akka.NET Team"
+let description = "Akka.NET is a port of the popular Java/Scala framework Akka to .NET"
+let tags = ["akka";"actors";"actor";"model";"Akka";"concurrency"]
+let configuration = if isMono then "Release Mono" else "Release"
```
heynickc: This variable is passed into flags during `dotnet build` and `dotnet test` later in the build script.  I think the only options are `Release` and `Debug` for the `-c` flag on those commands.  `Release Mono` would probably break those commands.

dhwed:
I'm currently evaluating Akka.NET, and the TCP support in 1.2 has an issue.  If I abort the connection, wait for it to be re-established, and then abort it again, my TCP listener actor does not receive any subsequent Tcp.Connected messages (I've confirmed with Wireshark that the TCP connection is being re-established after the second RST).

After some minor refactoring to work against this pull request instead, I haven't seen any issues.  Hopefully this gets merged in soon.

Also, my code is much cleaner now that I can use the `ByteString.Buffers` property to get an `IEnumerable<ArraySegment<byte>>`.

build.fsx
alexvaluyskiy: I think you should take the server version of this file

src/examples/TimeServer/TimeServer/TimeServer.csproj
```diff
@@ -43,8 +43,8 @@
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Include="Program.cs" />
-    <Compile Include="Properties\AssemblyInfo.cs" />
+
```
alexvaluyskiy: You should remove all Compile items only in the new projects. All examples still use the old one

Horusiath:
We cannot do anything with mono test failures: the reason for bug lies inside Mono implementation of `SocketAsyncEventArgs` and `Socket` itself. Those tests should be muted. I can also see some racy test on Windows but it's not related to this PR.

alexvaluyskiy:
@Horusiath could you mute them?

Horusiath:
Any idea why build.fsx on linux fails.

alexvaluyskiy:
@Horusiath look at the artifact for `netcore` http://petabridge-ci.cloudapp.net/repository/download/AkkaNet_AkkaNetLinuxMonoUnitTests/27401:id/TestResults/Akka.Tests_xunit-netcoreapp1.1.xml

Aaronontheweb:
Resolved differences with the upstream `v1.3` stuff

Aaronontheweb:
my bad @Horusiath - used the web UI for doing the conflict resolution and screwed up. Should be fixed now.

Aaronontheweb:
Sorry for the commit spam - even after the merge conflict was resolved there were a few things that still needed cleaning up. Think I've addressed them all now.

Aaronontheweb:
@Horusiath seeing some racy behavior with the TCP specs... I'll try to take a look tomorrow.

Horusiath:
Build is failing (failed to clean tests) X-th time in a row. Anyone knows what's the reason?

heynickc:
@Horusiath see #2848

src/core/Akka.Streams.Tests/IO/TcpSpec.cs
```diff
@@ -189,12 +200,12 @@ public void Outgoing_TCP_stream_must_work_when_remote_closes_write_then_client_c
             }, Materializer);
         }
 
-        [Fact(Skip = "Fix me")]
+        [Fact(Skip = "FIXME: actually this is about half-open connection. No other .NET socket lib supports that")]
```
Aaronontheweb: Yep, fine with skipping this.

src/core/Akka.Streams.Tests/IO/TcpSpec.cs
```diff
@@ -529,134 +536,146 @@ private void ValidateServerClientCommunication(ByteString testData, ServerConnec
...
                     .RunAggregate(0, (i, s) => i + s.Count, Materializer);
 
-                total.Wait(TimeSpan.FromSeconds(3)).Should().BeTrue();
+                total.Wait(TimeSpan.FromSeconds(5)).Should().BeTrue();
                 total.Result.Should().Be(1000);
             }, Materializer);
         }
 
-        [Fact(Skip = "Fix me")]
+        [Fact(Skip="FIXME")]
```
Aaronontheweb: What's the issue here?

Horusiath:
### Some benchmarks

I've tested a simple client-server ping/pong scenarios with 50B messages. A single server and 2000 active concurrent clients talking on localhost (so yes, those numbers are totally artificial). OS: Windows 10 (x64).

| Server | Avg. msg/sec | Baseline |
|-----|-----|----|
| DotNetty | 40215,1 | 1.0 |
| Akka.Streams.IO (.NET) |  36983,65 | 0.92 |
| Akka.Streams.IO (JVM) | 37560 | 0.93 |

Aaronontheweb:
@Horusiath impressive numbers!

PR #2684, opened 2017-05-20
[WIP] Akka.Stream 2.4.12 documentation mirror

Arkatufus:
The parts between //NOT DONE still aludes me because I couldn't find the exact implementation in akka.net (line 84-100)

### DIFF
--- a/docs/articles/streams/integration.md
+++ b/docs/articles/streams/integration.md
@@ -78,7 +78,6 @@ is completed with failure a `Akka.Actor.Status.Failure` message will be sent to 
 >It's often better to use `Sink.ActorRefWithAck` or `Ask` in `SelectAsync`, though. 
 
 #### Source.Queue
-// NOT DONE
 `Source.Queue` can be used for emitting elements to a stream from an actor (or from anything running 
 outside the stream). The elements will be buffered until the stream can process them. You can `Offer`
 elements to  the queue and they will be emitted to the stream if there is demand from downstream, 
@@ -87,15 +86,14 @@ otherwise they will be buffered until request for demand is received.
 Use overflow strategy `Akka.Streams.OverflowStrategy.Backpressure` to avoid dropping of elements 
 if the  buffer is full.
 
-`SourceQueue.Offer` returns `CompletionStage<StreamCallbackStatus<Boolean>>` 
-which completes with `Success(true)` if element was added to buffer or sent downstream. 
-It completes with `Success(false)` if element was dropped. It can also complete with
-`StreamCallbackStatus.Failure` when stream failed or `StreamCallbackStatus.StreamCompleted` 
+`ISourceQueueWithComplete.OfferAsync` returns `Task<IQueueOfferResult>` 
+which completes with `QueueOfferResult.Enqueued` if element was added to buffer or sent downstream. 
+It completes with `QueueOfferResult.Dropped` if element was dropped. It can also complete with
+`QueueOfferResult.Failure` when stream failed or `QueueOfferResult.QueueClosed` 
 when downstream is completed.
 
-When used from an actor you typically `pipe` the result of the `CompletionStage` back to the actor 
+When used from an actor you typically `pipe` the result of the `Task` back to the actor 
 to continue processing.
-// NOT DONE
 
 #### Source.ActorRef
 Messages sent to the actor that is materialized by ``Source.ActorRef`` will be emitted to the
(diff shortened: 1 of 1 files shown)


docs/articles/streams/integration.md
```diff
@@ -4,30 +4,112 @@ title: Integration
...
+// NOT DONE
+`Source.Queue` can be used for emitting elements to a stream from an actor (or from anything running 
+outside the stream). The elements will be buffered until the stream can process them. You can `Offer`
+elements to  the queue and they will be emitted to the stream if there is demand from downstream, 
+otherwise they will be buffered until request for demand is received.
+
+Use overflow strategy `Akka.Streams.OverflowStrategy.Backpressure` to avoid dropping of elements 
+if the  buffer is full.
+
+`SourceQueue.Offer` returns `CompletionStage<StreamCallbackStatus<Boolean>>`
```
Silv3rcircl3: The JVM part is correct, we really return those `QueueOfferRestul`s ;-) See https://github.com/akkadotnet/akka.net/blob/v1.3/src/core/Akka.Streams/QueueOfferResult.cs 
The onyl difference is that we use `OfferAsync` + `Task`

> `ISourceQueueWithComplete.OfferAsync` returns `Task<IQueueOfferResult>`  which completes with `IQueueOfferResult.Enqueued` if element was added to buffer or sent downstream. 
It completes with `IQueueOfferResult.Dropped` if element was dropped. It can also complete with
`IQueueOfferResult.Failure` when stream failed or `IQueueOfferResult.QueueClosed` 
when downstream is completed.

docs/articles/streams/integration.md
```diff
@@ -4,30 +4,112 @@ title: Integration
...
+Use overflow strategy `Akka.Streams.OverflowStrategy.Backpressure` to avoid dropping of elements 
+if the  buffer is full.
+
+`SourceQueue.Offer` returns `CompletionStage<StreamCallbackStatus<Boolean>>` 
+which completes with `Success(true)` if element was added to buffer or sent downstream. 
+It completes with `Success(false)` if element was dropped. It can also complete with
+`StreamCallbackStatus.Failure` when stream failed or `StreamCallbackStatus.StreamCompleted` 
+when downstream is completed.
+
+When used from an actor you typically `pipe` the result of the `CompletionStage` back to the actor
```
Silv3rcircl3: result of the `Task`

docs/articles/streams/integration.md
```diff
@@ -4,30 +4,112 @@ title: Integration
...
+
+`SourceQueue.Offer` returns `CompletionStage<StreamCallbackStatus<Boolean>>` 
+which completes with `Success(true)` if element was added to buffer or sent downstream. 
+It completes with `Success(false)` if element was dropped. It can also complete with
+`StreamCallbackStatus.Failure` when stream failed or `StreamCallbackStatus.StreamCompleted` 
+when downstream is completed.
+
+When used from an actor you typically `pipe` the result of the `CompletionStage` back to the actor 
+to continue processing.
+// NOT DONE
```
Silv3rcircl3: Anything else you need to know here?

docs/articles/streams/integration.md
```diff
@@ -735,3 +518,306 @@ var flow = Flow.FromProcessor(()=> createProcessor(materializer));
...
 
 Please note that a factory is necessary to achieve reusability of the resulting `Flow`.
+
+####Implementing Reactive Streams Publisher or Subscriber
+
+As described above any Akka Streams ``Source`` can be exposed as a Reactive Streams ``Publisher`` 
+and any ``Sink`` can be exposed as a Reactive Streams ``Subscriber``. Therefore we recommend that you 
+implement Reactive Streams integrations with built-in stages or [custom stages](customstreamprocessing.md).
+
+For historical reasons the `ActorPublisher` class and `ActorSubscriber` class traits are
```
Silv3rcircl3: remove `traits` , they are kind of interfaces in .net but aren't relevant here since one would extend the abstract classes instead of the interface.

docs/articles/streams/integration.md
Arkatufus: I think I got it all, can you give it a check?

docs/articles/streams/integration.md
```diff
@@ -4,30 +4,112 @@ title: Integration
 ---
 
 #Integrating with Actors
```
Arkatufus: Do you think I should change the other documentation files as well? Should be easy with a simple grep search/replace.

Silv3rcircl3: Sure, go ahead, but in another PR please

PR #2689, opened 2017-05-21
CommandAsync for persistent actors

zbynek001:
Here is another try on CommandAsync topic.
I's introducing similar support for async/await as in ReceiveAsync.

Remarks:
- As with ReceiveAsync, all other commands to actor are blocked until the command handler task finishes processing the current command, including the Persist and PersistAll callbacks.
- It does NOT introduce async support for Persist callbacks, they work the same way as before

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
@@ -244,6 +244,7 @@ namespace Akka.Persistence
         public void PersistAsync<TEvent>(TEvent @event, System.Action<TEvent> handler) { }
         protected abstract bool ReceiveCommand(object message);
         protected abstract bool ReceiveRecover(object message);
+        protected void RunTask(System.Func<System.Threading.Tasks.Task> action) { }
         public void SaveSnapshot(object snapshot) { }
         protected override void Unhandled(object message) { }
     }
@@ -480,6 +481,11 @@ namespace Akka.Persistence
         protected void Command(System.Type messageType, System.Func<object, bool> handler) { }
         protected void Command(System.Action<object> handler) { }
         protected void CommandAny(System.Action<object> handler) { }
+        protected void CommandAnyAsync(System.Func<object, System.Threading.Tasks.Task> handler) { }
+        protected void CommandAsync<T>(System.Func<T, System.Threading.Tasks.Task> handler, System.Predicate<T> shouldHandle = null) { }
+        protected void CommandAsync<T>(System.Predicate<T> shouldHandle, System.Func<T, System.Threading.Tasks.Task> handler) { }
+        protected void CommandAsync(System.Type messageType, System.Func<object, System.Threading.Tasks.Task> handler, System.Predicate<object> shouldHandle = null) { }
+        protected void CommandAsync(System.Type messageType, System.Predicate<object> shouldHandle, System.Func<object, System.Threading.Tasks.Task> handler) { }
         protected virtual void OnCommand(object message) { }
         protected virtual void OnRecover(object message) { }
         protected void Recover<T>(System.Action<T> handler, System.Predicate<T> shouldHandle = null) { }
--- a/src/core/Akka.Persistence.Tests/PersistentActorSpecAsyncAwait.Actors.cs
+++ b/src/core/Akka.Persistence.Tests/PersistentActorSpecAsyncAwait.Actors.cs
@@ -1,0 +1,1300 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="PersistentActorSpec.Actors.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading;
+using Akka.Actor;
+using Akka.TestKit;
+using Akka.Util.Internal;
+using Akka.Persistence.Internal;
+using System.Threading.Tasks;
+
+namespace Akka.Persistence.Tests
+{
+    public partial class PersistentActorSpecAsyncAwait
+    {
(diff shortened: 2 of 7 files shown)


src/core/Akka.Persistence.Tests/PersistentActorSpecAsyncAwait.Actors.cs
```diff
@@ -0,0 +1,1337 @@
...
+using System.Threading;
+using Akka.Actor;
+using Akka.TestKit;
+using Akka.Util.Internal;
+using Akka.Persistence.Internal;
+using System.Threading.Tasks;
+
+namespace Akka.Persistence.Tests
+{
+    //internal class BehaviorOneActor : PersistentActorSpec.ExamplePersistentActor
```
Horusiath: Why is this part commented?

src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
```diff
@@ -243,6 +243,7 @@ namespace Akka.Persistence
         public void PersistAsync<TEvent>(TEvent @event, System.Action<TEvent> handler) { }
         protected abstract bool ReceiveCommand(object message);
         protected abstract bool ReceiveRecover(object message);
+        protected void RunTask(System.Func<System.Threading.Tasks.Task> action) { }
```
Horusiath: This method is not unique for the persistence. It should be either hidden or included in `ActorBase` class.

src/core/Akka.Persistence/Eventsourced.Recovery.cs
```diff
@@ -39,9 +40,48 @@ public EventsourcedState(string name, bool isRecoveryRunning, StateReceive state
...
+        /// <param name="action">TBD</param>
+        protected void RunTask(Func<Task> action)
+        {
+            Func<Task> wrap = () =>
+            {
+                Task t = action();
+                if (!t.IsCompleted)
+                {
+                    var tcs = new TaskCompletionSource<object>();
+                    t.ContinueWith(r =>
```
Horusiath: You probably won't need `TaskCompletionSource` here. `ContinueWith` will return continuation task, while `TaskContinuationOption.NotOnFaulted | TaskContinuationOption.NotOnCancelled` should allow you to get rid of forwarding cancelations and faults.

There's also one thing to remember - things like actor's `Context` or `Self` are fragile and won't work in task continuations. This also concerns all methods invoked here.

src/core/Akka.Persistence.Tests/PersistentActorSpecAsyncAwait.Actors.cs
```diff
@@ -0,0 +1,1337 @@
...
+
+                if (message is Cmd cmd)
+                {
+                    RunTask(async () =>
+                    {
+                        await Task.Yield();
+                        Persist(new Evt(cmd.Data + "-0"), UpdateStateHandler);
+                        Context.Become(NewBehavior);
+                    });
+                    return true;
```
Horusiath: I don't think it's fully valid way of calling RunTask. As it won't be invoked immediately, the handler will return true, while the func inside may throw an exception. This may lead to a false positives, but I'd need to check how this would look like in other cases.

src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
zbynek001: Actually it is unique to Eventsourced, because of the Persist callbacks.
Non-persistence actors already have other equivalents, e.g. UntypedActor have RunTask, ReceiveActor have ReceiveAsync

src/core/Akka.Persistence.Tests/PersistentActorSpecAsyncAwait.Actors.cs
zbynek001: I had to move it to another namespace. It inherits from PersistentActorSpec.ExamplePersistentActor but i need to use a different actor with the RunTask inside

src/core/Akka.Persistence.Tests/PersistentActorSpecAsyncAwait.Actors.cs
zbynek001: Should be the same behavior as with ReceiveAsync

Aaronontheweb:
@zbynek001 

```
ApprovalTests.Core.Exceptions.ApprovalMismatchException : Failed Approval: Received file D:\work\d26c9d7f36545acd\src\core\Akka.API.Tests\CoreAPISpec.ApprovePersistence.received.txt does not match approved file D:\work\d26c9d7f36545acd\src\core\Akka.API.Tests\CoreAPISpec.ApprovePersistence.approved.txt.
   at ApprovalTests.Approvers.FileApprover.Fail()
   at ApprovalTests.Core.Approver.Verify(IApprovalApprover approver, IApprovalFailureReporter reporter)
   at ApprovalTests.Approvals.Verify(IApprovalWriter writer, IApprovalNamer namer, IApprovalFailureReporter reporter)
   at ApprovalTests.Approvals.Verify(IApprovalWriter writer)
   at ApprovalTests.Approvals.Verify(String text)
   at Akka.API.Tests.CoreAPISpec.ApprovePersistence() in D:\work\d26c9d7f36545acd\src\core\Akka.API.Tests\CoreAPISpec.cs:line 53
```

Need to run the API approval process locally... http://getakka.net/docs/akka-developers/public-api-changes

zbynek001:
@Aaronontheweb  will do that. Btw there is an issue with the approval ApiGenerator, the order of items is based on local culture, InvariantCulture might be better. Will create another pull request for that

src/core/Akka.Persistence/Eventsourced.Recovery.cs
zbynek001: Before I didn't need it, but now i need to react also on failure.
Regarding Context, it works as usual. Context, Sender,.. are accessible inside the command handler. Same for Persist callback except of the Sender, but that's by design

Aaronontheweb:
@zbynek001 rebasing on those changes since we merge in your API approval fix.

Danthar:
Has some minor conflicts that need to be resolved. @Aaronontheweb we are not making this part of release 1.3 ?

Aaronontheweb:
@Danthar yeah, this should definitely go into 1.3

PR #2691, opened 2017-05-21
Enable FsCheck on NetCore

alexvaluyskiy:


Aaronontheweb:
@alexvaluyskiy something blew up in Akka.Remote on your PR here - lots of failures back.

PR #2697, opened 2017-05-24
TBD comments and local function support in Akka core

Aaronontheweb:


src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -297,8 +297,8 @@ public Task<Done> Run(string fromPhase = null)
             if (_runStarted.CompareAndSet(false, true))
             {
                 var debugEnabled = Log.IsDebugEnabled;
-                Func<List<string>, Task<Done>> loop = null;
-                loop = remainingPhases =>
+
+                Task<Done> Loop(List<string> remainingPhases)
```
Aaronontheweb: Local function optimization in `CoordinatedShutdown`, which should have minimal impact if any.

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -483,26 +481,24 @@ internal static List<string> TopologicalSort(Dictionary<string, Phase> phases)
             var unmarked = new HashSet<string>(phases.Keys.Concat(phases.Values.SelectMany(x => x.DependsOn)));
             var tempMark = new HashSet<string>(); // for detecting cycles
 
-            Action<string> depthFirstSearch = null;
-            depthFirstSearch = u =>
+            void DepthFirstSearch(string u)
```
Aaronontheweb: Another local function optimization in `CoordinatedShutdown`. Minimal impact - this code only runs once per application.

src/core/Akka/Dispatch/Mailbox.cs
```diff
@@ -375,8 +375,7 @@ private void ProcessMailbox(int left, long deadlineTicks)
         {
             while (ShouldProcessMessage())
             {
-                Envelope next;
-                if (!TryDequeue(out next)) return;
+                if (!TryDequeue(out var next)) return;
```
Aaronontheweb: Changed mailbox to use new `out var` syntax. Don't expect this to have any real performance impact.

src/core/Akka/Dispatch/Mailboxes.cs
```diff
@@ -277,38 +277,36 @@ public MailboxType GetMailboxType(Props props, Config dispatcherConfig)
                 _mailboxSizeWarningIssued = true;
             }
 
-            Func<MailboxType, MailboxType> verifyRequirements = mailboxType =>
+            MailboxType VerifyRequirements(MailboxType mailboxType)
```
Aaronontheweb: This might be interesting - if the `VerifyRequirements` local function is able to be created using a value type, which the compiler tries to do by default, this might improve the rate at which we're able to spawn actors. Will check benchmark results to verify.

alexvaluyskiy:
@Aaronontheweb Where is a performance comparison?

Aaronontheweb:
# Before

# Akka.Tests.Performance.Actor.ActorMemoryFootprintSpec+ReceiveActor_memory_footprint
__Measures the amount of memory used by 10,000 ReceiveActors__
_5/24/2017 4:56:56 PM_
### System Info
```ini
NBench=NBench, Version=1.0.1.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
```

### NBench Settings
```ini
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
|TotalBytesAllocated |           bytes |   20,841,880.00 |   20,037,162.46 |   17,498,864.00 |      910,118.80 |
|[Counter] ActorCreateThroughput |      operations |       10,000.00 |       10,000.00 |       10,000.00 |            0.00 |

### Per-second Totals
|TotalBytesAllocated |           bytes |  149,566,053.58 |  126,771,473.86 |  112,043,922.28 |   10,138,918.00 |
|[Counter] ActorCreateThroughput |      operations |       71,762.27 |       63,337.81 |       54,208.97 |        5,051.87 |

Aaronontheweb:
# After

# Akka.Tests.Performance.Actor.ActorMemoryFootprintSpec+ReceiveActor_memory_footprint
__Measures the amount of memory used by 10,000 ReceiveActors__
_5/24/2017 3:40:00 PM_
### System Info
```ini
NBench=NBench, Version=1.0.1.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
```

### NBench Settings
```ini
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
|TotalBytesAllocated |           bytes |   20,598,024.00 |   20,204,595.69 |   19,381,792.00 |      307,827.70 |
|[Counter] ActorCreateThroughput |      operations |       10,000.00 |       10,000.00 |       10,000.00 |            0.00 |

### Per-second Totals
|TotalBytesAllocated |           bytes |  148,598,273.64 |  137,075,791.03 |  124,809,017.36 |    6,426,981.38 |
|[Counter] ActorCreateThroughput |      operations |       72,142.00 |       67,823.11 |       62,338.93 |        2,470.39 |

Aaronontheweb:
@alexvaluyskiy ActorCreateThroughput is virtually the same, but this benchmark may not be a reliable measure for any perf changes here since the actor type we're creating is the same each time. The JITer can optimize the code after warm-ups to re-use the same mailbox type and type check each time since it's always the same actor and same mailbox type. For the same reasons I mentioned on the other PR though, there's only upside to moving away from delegates to local functions here. Worst case scenario gets implemented the same way we've done it with `Func<>`, using a delegate on the heap.

sean-gilliam:
Another plus is that local functions are *slightly easier and natural* to read than Func<>.

PR #2701, opened 2017-05-24
Use a pattern matching inside ClusterSingletonManager

alexvaluyskiy:


src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -872,174 +866,150 @@ private void InitializeFSM()
...
-                        else
-                        {
-                            return GoToStopping(wasOldestData.Singleton);
-                        }
+                        return wasOldestData.SingletonTerminated ? Stop() : GoToStopping(wasOldestData.Singleton);
                     }
                     else if (takeOverRetry.Count <= _maxTakeOverRetries)
                     {
-                        Log.Info("Retry [{0}], sending TakeOverFromMe to [{1}]", takeOverRetry.Count, wasOldestData.NewOldest);
+                        Log.Info("Retry [{0}], sending TakeOverFromMe to [{1}]", takeOverRetry.Count, wasOldestData.NewOldest?.Address);
```
alexvaluyskiy: Fixed potential `NullReferenceException`

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -696,53 +696,55 @@ private void InitializeFSM()
...
+                                return GoToOldest();
+                            case UniqueAddress prev:
+                                Peer(prev.Address).Tell(HandOverToMe.Instance);
+                                return GoTo(ClusterSingletonState.BecomingOldest).Using(new BecomingOldestData(prev));
                         }
                     }
                     else
                     {
-                        Log.Info("Younger observed OldestChanged: [{0} -> {1}]", youngerData.Oldest.Address, oldestChanged.Oldest.Address);
+                        Log.Info("Younger observed OldestChanged: [{0} -> {1}]", youngerData.Oldest?.Address, oldestChanged.Oldest?.Address);
```
alexvaluyskiy: Fixed potential `NullReferenceException`

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -696,53 +696,55 @@ private void InitializeFSM()
...
+                        ScheduleDelayedMemberRemoved(memberRemoved.Member);
                         return Stay();
                     }
                 }
-                else if ((removed = e.FsmEvent as DelayedMemberRemoved) != null
-                    && (youngerData = e.StateData as YoungerData) != null
-                    && removed.Member.UniqueAddress.Equals(youngerData.Oldest))
+                else if (e.FsmEvent is DelayedMemberRemoved removed
+                    && e.StateData is YoungerData data
+                    && data.Oldest != null
```
alexvaluyskiy: Added a check for `null`

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -755,58 +757,57 @@ private void InitializeFSM()
...
+                        ScheduleDelayedMemberRemoved(memberRemoved.Member);
                         return Stay();
                     }
                 }
-                else if ((removed = e.FsmEvent as DelayedMemberRemoved) != null
-                    && becomingOldest != null
-                    && removed.Member.UniqueAddress.Equals(becomingOldest.PreviousOldest))
+                else if (e.FsmEvent is DelayedMemberRemoved delayed
+                        && e.StateData is BecomingOldestData becoming
+                        && becoming.PreviousOldest != null
```
alexvaluyskiy: Added a check for `null`

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -816,40 +817,33 @@ private void InitializeFSM()
...
-                                return Stay();
+                                case UniqueAddress previousOldest:
+                                    if (previousOldest.Equals(senderUniqueAddress))
+                                        Sender.Tell(HandOverToMe.Instance);
+                                    else
+                                        Log.Info("Ignoring TakeOver request in BecomingOldest from [{0}]. Expected previous oldest [{1}]", Sender.Path.Address, previousOldest.Address);
+                                    return Stay();
+                                case null:
+                                    Sender.Tell(HandOverToMe.Instance);
+                                    return Stay().Using(new BecomingOldestData(senderUniqueAddress));
```
alexvaluyskiy: Use `senderUniqueAddress` instead of `_selfUniqueAddress`

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -696,53 +696,55 @@ private void InitializeFSM()
...
+                return null;
             });
 
             When(ClusterSingletonState.Younger, e =>
             {
-                DelayedMemberRemoved removed;
-                YoungerData youngerData;
-                if (e.FsmEvent is OldestChangedBuffer.OldestChanged && (youngerData = e.StateData as YoungerData) != null)
+                if (e.FsmEvent is OldestChangedBuffer.OldestChanged oldestChanged
+                    && e.StateData is YoungerData youngerData)
```
Aaronontheweb: Nice, didn't realize you could do `var` binding here the same way you can with `switch`. C# 7 is a godsend.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -816,40 +817,33 @@ private void InitializeFSM()
...
-                        {
-                            if (becomingOldest.PreviousOldest.Equals(senderUniqueAddress))
-                            {
-                                Sender.Tell(HandOverToMe.Instance);
-                            }
-                            else
+                        case null:
+                            // from unknown node, ignore
+                            Log.Info("Ignoring TakeOver request from unknown node in BecomingOldest from [{0}]", senderAddress);
+                            return Stay();
```
alexvaluyskiy: Previously we have returned `null` here

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
alexvaluyskiy: @Aaronontheweb Yes we could, but there is some limitations (or bug) in C# implementation. For example, I could use `case patterns` and use a variable with the same name in more that one place
```C#
string obj = "apple";
switch (obj)
{
    case null:
        Console.WriteLine("not a string");
        break;
    case string s when s.Length > 10:
        Console.WriteLine($"long string: {s}");
        break;
    case string s:
        Console.WriteLine($"string: {s}");
        break;
}
```

But I could not do the same with `if-patterns`
```C#
string obj = "apple";
if (obj == null)
{
	Console.WriteLine("not a string");
}
else if (obj is string s && s.Length > 0)
{
	Console.WriteLine($"long string: {s}");
}
else if (obj is string s) // Compiler error
{
	Console.WriteLine($"string: {b}");
}
```

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
Aaronontheweb: @alexvaluyskiy semantically the same - returning `null` returns `stay()` by default.

PR #2702, opened 2017-05-25
[WIP] Sql persistence update

Horusiath:

This PR aims to bring two major features:

1. A common approach to support JSON data type columns for SQL-based journals.
2. Ability for SQL read journals to query current / live events - not only some of them (like in case of tags), but actually all of them.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Query.Sql/AllEventsPublisher.cs
+++ b/src/contrib/persistence/Akka.Persistence.Query.Sql/AllEventsPublisher.cs
@@ -1,0 +1,215 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="AllEventsPublisher.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using Akka.Actor;
+using Akka.Event;
+using Akka.Persistence.Sql.Common.Journal;
+using Akka.Streams.Actors;
+using System;
+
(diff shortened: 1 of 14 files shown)


src/contrib/persistence/Akka.Persistence.Query.Sql/AllEventsPublisher.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            Buffer.DeliverBuffer(TotalDemand);
+            if (Buffer.IsEmpty && CurrentOffset > ToOffset)
+                OnCompleteThenStop();
+
+            Context.Become(Idle);
+        }
+    }
+
+    internal sealed class CurrentAllEventsPublisher : AbstractAllEventsPublisher
```
alexvaluyskiy: I don't like this naming. Maybe we could use simpler `EventsQuery` and `CurrentEventsQuery`?

src/contrib/persistence/Akka.Persistence.Query.Sql/SqlReadJournal.cs
```diff
@@ -165,9 +165,53 @@ public SqlReadJournal(ExtendedActorSystem system, Config config)
...
+        /// stored events is provided by <see cref="CurrentEvents"/>.
+        /// <para></para>
+        /// The SQL write journal is notifying the query side as soon as events are persisted, but for
+        /// efficiency reasons the query side retrieves the events in batches that sometimes can
+        /// be delayed up to the configured `refresh-interval`.
+        /// <para></para>
+        /// The stream is completed with failure if there is a failure in executing the query in the
+        /// backend journal.
+        /// </summary>
+        public Source<EventEnvelope, NotUsed> AllEvents(long fromOffset = 0, long toOffset = long.MaxValue) =>
```
alexvaluyskiy: Maybe we could use simpler name `Events`?

PR #2703, opened 2017-05-25
sharding snapshot & journal cleanup

zbynek001:
port of
https://github.com/akka/akka/issues/21725
https://github.com/akka/akka/issues/22916

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
@@ -18,7 +18,7 @@ namespace Akka.Cluster.Sharding
     using Msg = Object;
 
     /// <summary>
-    /// This actor creates children entity actors on demand that it is told to be 
+    /// This actor creates children entity actors on demand that it is told to be
     /// responsible for. It is used when `rememberEntities` is enabled.
     /// </summary>
     public class PersistentShard : Shard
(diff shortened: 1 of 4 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
```diff
@@ -69,7 +69,46 @@ public class PersistentShard : Shard
...
+                    * for debugging if something goes wrong in production it's very useful to be able to inspect the events
+                    */
+                    var deleteToSequenceNr = m.Metadata.SequenceNr - Settings.TunningParameters.KeepNrOfBatches * Settings.TunningParameters.SnapshotAfter;
+                    if (deleteToSequenceNr > 0)
+                    {
+                        DeleteMessages(deleteToSequenceNr);
+                    }
+                    break;
+                case SaveSnapshotFailure m:
+                    Log.Warning("PersistentShard snapshot failure: {}", m.Cause.Message);
```
Horusiath: In .NET logging we must provide index for arguments. So `{}` itself will fail, you need to use `{0}`, `{1}` and so on.

zbynek001: Right, completely overlooked that. Fixed

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
```diff
@@ -69,7 +69,46 @@ public class PersistentShard : Shard
...
+                    break;
+
+                case DeleteMessagesFailure m:
+                    Log.Warning("PersistentShard messages to {0} deletion failure: {1}", m.ToSequenceNr, m.Cause.Message);
+                    break;
+                case DeleteSnapshotsSuccess m:
+                    Log.Debug("PersistentShard snapshots matching {0} deleted successfully", m.Criteria);
+                    break;
+                case DeleteSnapshotsFailure m:
+                    Log.Warning("PersistentShard snapshots matching {0} deletion falure: {1}", m.Criteria, m.Cause.Message);
```
alexvaluyskiy: Should be `failure`

alexvaluyskiy:
@zbynek001 You have also implemented this issue
https://github.com/akka/akka/pull/20038
Because we did not handle snapshots before

zbynek001:
I think that's obsolete code.
SaveSnapshotSuccess & SaveSnapshotFailure is already handled inside PersistentShard

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
zbynek001: fixed

PR #2710, opened 2017-05-30
Streams update 2.4.16

Silv3rcircl3:
- [x] [fix SourceQueue docs inconsistencies](https://github.com/akka/akka/pull/21877)
- [x] [Fixed the case when identity flow appended to Source combining materialized values](https://github.com/akka/akka/pull/21956)
- [x] [less noisy logging of idle-timeout (for akka-http specifically](https://github.com/akka/akka/pull/21989)
- [x] [asOutput stream close race](https://github.com/akka/akka/pull/21896)
- [x] [InputStream from StreamConverters.asInputStream respects read() contract](https://github.com/akka/akka/pull/21954)
- [x] [Fix handling of escape sequences in JsonObjectParser](https://github.com/akka/akka/pull/21975)
- [x] [add note on (non)guarantees of materializer SPI](https://github.com/akka/akka/pull/21935)
- [x]  [Lazy source](https://github.com/akka/akka/pull/21081)
- [x]  [ree flattening should be separate from Fusing](https://github.com/kkasravi/akka/commit/73ed75a66967ca3692b9ed86775d7d81b54feb4e)
- [x] [Fixed the bug in Attributes.getFirst(default : T)](https://github.com/akka/akka/pull/21846)

~Once #2683 is merged those two changes need to ported too~

~- https://github.com/akka/akka/pull/21989/files#diff-7104b3bccdee7b3ce9073c6f2179f42f~
~- https://github.com/akka/akka/pull/21989/files#diff-3a67e6d52956dd56d88c954b5d90d201~

~and `asOutput stream close race` probably needs to be fixed, couldn't run the test while implemented it~


src/core/Akka.Streams.Tests/Dsl/GraphMatValueSpec.cs
```diff
@@ -325,6 +325,31 @@ public void A_Graph_with_Identity_Flow_optimization_even_if_port_are_wired_in_an
             var matValue = Source.Single("").Via(nest4).To(Sink.Ignore<Tuple<string, string>>()).Run(mat2);
             matValue.Should().Be(NotUsed.Instance);
         }
+
+        [Fact]
+        public void A_Graph_must_not_ignore_materialized_value_of_identity_flow_which_is_optimized_away()
+        {
+            var mat = Sys.Materializer(ActorMaterializerSettings.Create(Sys).WithAutoFusing(false));
+            var t = Source.Single(1)
+                .ViaMaterialized(Flow.Identity<int>(), Keep.Both)
```
alexvaluyskiy: Should we change `Keep.Both` to ValueTuple in the future?

alexvaluyskiy:
`less noisy logging of idle-timeout (for akka-http specifically)` contains changes for TCP. We should create an issue for that

src/core/Akka.Streams/IMaterializer.cs
```diff
@@ -12,7 +12,16 @@
 namespace Akka.Streams
 {
     /// <summary>
-    /// TBD
+    /// Materializer SPI (Service Provider Interface) 
+    /// 
+    /// Binary compatibility is NOT guaranteed on materializer internals.
```
alexvaluyskiy: We do not provide binary compatibility at the moment. It is redundant sentence

alexvaluyskiy:
Have you implemented this one?
[Fixed the bug in Attributes.getFirst(default : T)](https://github.com/akka/akka/pull/21846)

src/core/Akka.Streams.Tests/Dsl/GraphMatValueSpec.cs
Silv3rcircl3: Good idea, I'll change it right away via nuget package.

Silv3rcircl3:
We've reported this bug on the jvm and it was already implemented correctly so no need to port it.

PR #2719, opened 2017-06-01
[routing] Akka.Routing refactoring

sean-gilliam:
This PR refactors the classes in the Akka.Routing namespace to take
advantage of newer C# features.  It also simplifies some of the logic in
the various routers.

### DIFF
--- a/src/core/Akka/Routing/Broadcast.cs
+++ b/src/core/Akka/Routing/Broadcast.cs
@@ -153,38 +153,28 @@ namespace Akka.Routing
 
         private RouterConfig OverrideUnsetConfig(RouterConfig other)
         {
-            if (other is NoRouter)
+            if (other is Pool pool)
             {
-                return this;
-            }
-            else
-            {
-                var pool = other as Pool;
-                if (pool != null)
+                BroadcastPool wssConf;
+
+                if (SupervisorStrategy != null
+                    && SupervisorStrategy.Equals(DefaultSupervisorStrategy)
+                    && !pool.SupervisorStrategy.Equals(DefaultSupervisorStrategy))
                 {
-                    BroadcastPool wssConf;
-
-                    if (SupervisorStrategy != null
-                        && SupervisorStrategy.Equals(Pool.DefaultSupervisorStrategy)
-                        && !(pool.SupervisorStrategy.Equals(Pool.DefaultSupervisorStrategy)))
-                    {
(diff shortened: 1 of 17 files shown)


src/core/Akka/Routing/Broadcast.cs
```diff
@@ -153,38 +153,28 @@ public override RouterConfig WithFallback(RouterConfig routerConfig)
 
         private RouterConfig OverrideUnsetConfig(RouterConfig other)
         {
-            if (other is NoRouter)
+            if (other is Pool pool)
```
sean-gilliam: everything that isn't a ```Pool``` returned ```this``` , so this is the only real check we need. Same with the other routers.

src/core/Akka/Routing/ConsistentHashRouter.cs
```diff
@@ -185,42 +185,43 @@ public override Routee Select(object message, Routee[] routees)
...
-                    _log.Value.Warning("Couldn't route message with consistent hash key [{0}] due to [{1}]", hashData,
-                        ex.Message);
+                    _log.Value.Warning("Couldn't route message with consistent hash key [{0}] due to [{1}]", hashData, ex.Message);
                     return Routee.NoRoutee;
                 }
-            };
+            }
 
-            if (_hashMapping(message) != null)
+            var hashMapping = _hashMapping(message);
```
sean-gilliam: This delegate was being called twice which should return the same value each time. So I stored it in a variable so it only needed to be called once.

Aaronontheweb: Nice catch

Aaronontheweb:
looks great @sean-gilliam - love seeing the C#7 features cutting down on code bloat

src/core/Akka/Routing/ConsistentHash.cs
```diff
@@ -289,35 +289,39 @@ public ISurrogated FromSurrogate(ActorSystem system)
...
-                return BitConverter.GetBytes((double)obj);
-            if (obj is decimal)
-                return new BitArray(decimal.GetBits((decimal)obj)).ToBytes();
-            if (obj is Guid)
-                return ((Guid)obj).ToByteArray();
-            return obj;
+            switch (obj)
+            {
+                case null:
+                    return new byte[] { 0 };
```
maxcherednik: I don't know the context, and how often this one called and maybe it's super micro, but seems like a constant?

Issue #2575, opened 2017-03-30
Port exhaustion problem with Akka.Cluster


Blind-Striker:
Hi,

We created an Akka Cluster infrastructure for Sms, Email and Push notifications. 3 different kind of nodes are exist in the system, which are client, sender and lighthouse. Client role is being used by Web application and API application(Web and API is hosted at IIS). Lighthouse and Sender roles are being hosted as a Windows service. We are also running 4 more console applications of same windows service that in sender role. 

We've been experiencing port exhaustion problems in our Web Server for about 2 weeks. Our Web Server starting to consume the ports quickly and after a while we can not do any SQL operations. 
Sometimes we have no choice but to do iis reset. This problems occur if there are more than one nodes that in sender role. We diagnosed it and found the source of the problem.

```
---------------
HOST                  OPEN    WAIT
SRV_NOTIFICATION      3429    0
SRV_LOCAL             198     0
SRV_UNDEFINED_IPV4    23      0
SRV_DATABASE          15      0
SRV_AUTH              4       0
SRV_API               6       0
SRV_UNDEFINED_IPV6    19      0
SRV_INBOUND           12347   5

TotalPortsInUse   : 17286
MaxUserPorts      : 64510
TcpTimedWaitDelay : 30
03/23/2017 09:30:10
---------------
```
SRV_NOTIFICATION is server that lighthouse ve sender's nodes running. SRV_INBOUND is our Web Server. After checking this table, we checked what ports on the Web Server were assigned. 
And we got results like table below. In netstat there were more than 12000 connections like this :

```
TCP    192.168.1.10:65531     192.168.1.10:3564      ESTABLISHED     5716   [w3wp.exe]
TCP    192.168.1.10:65532     192.168.1.101:17527    ESTABLISHED     5716   [w3wp.exe]
TCP    192.168.1.10:65533     192.168.1.101:17527    ESTABLISHED     5716   [w3wp.exe]
TCP    192.168.1.10:65534     192.168.1.10:3564      ESTABLISHED     5716   [w3wp.exe]
```

192.168.1.10 Web Server
192.168.1.10:3564 API
192.168.1.101:17527 Lighthouse

The connections are opening but not closing.

[...]

PR #2721, opened 2017-06-02
Cleaning up racy specs part1

Aaronontheweb:
I've looked through some of these and the reasons for the raciness in most of these specs are due to issues where sockets aren't being cleaned up quickly enough or the timeouts are too narrowly defined. There might be some legitimate bugs with the specs here too, so I'm going to go through each one and submit some patches for each in the name of making our build process run more smoothly.

Current list:

* [x] Akka.Cluster.Tests.ActorProviderConfig_should_resolve_remote_alias
* [x] Akka.Cluster.Tests.ActorProviderConfig_should_resolve_cluster_alias
* [x] Akka.Cluster.Tests.ClusterSpec.A_cluster_must_cancel_LeaveAsync_task_if_CancellationToken_fired_before_node_left
* [x] Akka.Cluster.Tests.DowningProviderSpec.Downing_provider_should_stop_the_cluster_if_the_downing_provider_throws_exception_in_props
* [x] Akka.Cluster.Tests.ClusterSpec.A_cluster_must_return_completed_LeaveAsync_task_if_member_already_removed
* [x] Akka.Tests.Routing.TailChoppingSpec.Tail_chopping_group_router_must_throw_exception_if_no_result_will_arrive_within_the_given_time

The common thread with most of these issues is that they use the real socket system underneath Akka.Remote, so I suspect it's a problem with how that's being used / not giving it enough time.

I'm also going to see if these failures might be related to https://github.com/akkadotnet/akka.net/issues/2575 - which I was planning on fixing anyway. Stay tuned.

### DIFF
--- a/src/core/Akka.Cluster.Tests/ActorRefProvidersConfigSpec.cs
+++ b/src/core/Akka.Cluster.Tests/ActorRefProvidersConfigSpec.cs
@@ -40,11 +40,13 @@ namespace Akka.Cluster.Tests
 
         private void ConfigureAndVerify(string alias, Type actorProviderType)
         {
-            var config = ConfigurationFactory.ParseString(@"akka.actor.provider = " + alias);
+            var config = ConfigurationFactory.ParseString(@"akka.actor.provider = " + alias)
+                .WithFallback(ConfigurationFactory.ParseString("akka.remote.dot-netty.tcp.port = 0")); // use a random port to avoid issues with async and parallelization
             using (var system = ActorSystem.Create(nameof(ActorRefProvidersConfigSpec), config))
             {
                 var ext = (ExtendedActorSystem) system;
                 ext.Provider.GetType().ShouldBe(actorProviderType);
+                system.Terminate().Wait(TimeSpan.FromSeconds(3)); // force the system to cleanup and shutdown
             }
         }
     }
--- a/src/core/Akka.Cluster.Tests/ClusterSpec.cs
+++ b/src/core/Akka.Cluster.Tests/ClusterSpec.cs
@@ -20,6 +20,12 @@ namespace Akka.Cluster.Tests
 {
     public class ClusterSpec : AkkaSpec
     {
+
+        /*
+         * Portability note: all of the _cluster.Join(selfAddress) calls are necessary here, whereas they are not in the JVM
+         * because the JVM test suite relies on side-effects from one test to another, whereas all of our tests are fully isolated.
+         */
+
         const string Config = @"    
             akka.cluster {
               auto-down-unreachable-after = 0s
@@ -46,7 +52,7 @@ namespace Akka.Cluster.Tests
             _cluster = Cluster.Get(Sys);
         }
 
-        public void LeaderActions()
+        internal void LeaderActions()
         {
             _cluster.ClusterCore.Tell(InternalClusterAction.LeaderActionsTick.Instance);
         }
@@ -89,7 +95,6 @@ namespace Akka.Cluster.Tests
         {
             try
             {
-                // TODO: this should be removed
                 _cluster.Join(_selfAddress);
                 LeaderActions(); // Joining -> Up
 
@@ -113,7 +118,6 @@ namespace Akka.Cluster.Tests
         [Fact]
         public void A_cluster_must_publish_member_removed_when_shutdown()
         {
-            // TODO: this should be removed
             _cluster.Join(_selfAddress);
             LeaderActions(); // Joining -> Up
 
@@ -139,7 +143,6 @@ namespace Akka.Cluster.Tests
         [Fact]
         public void BugFix_2442_RegisterOnMemberUp_should_fire_if_node_already_up()
         {
-            // TODO: this should be removed
             _cluster.Join(_selfAddress);
             LeaderActions(); // Joining -> Up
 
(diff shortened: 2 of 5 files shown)


src/core/Akka.Cluster.Tests/ActorRefProvidersConfigSpec.cs
```diff
@@ -40,11 +40,13 @@ public void ActorProviderConfig_should_resolve_cluster_alias()
 
         private void ConfigureAndVerify(string alias, Type actorProviderType)
         {
-            var config = ConfigurationFactory.ParseString(@"akka.actor.provider = " + alias);
+            var config = ConfigurationFactory.ParseString(@"akka.actor.provider = " + alias)
+                .WithFallback(ConfigurationFactory.ParseString("akka.remote.dot-netty.tcp.port = 0")); // use a random port to avoid issues with async and parallelization
```
Aaronontheweb: Resolved via port 0 binding (won't have any side-effects on using a local `ActorSystem` since the configuration never gets read.)

src/core/Akka.Cluster.Tests/ActorRefProvidersConfigSpec.cs
```diff
@@ -40,11 +40,13 @@ public void ActorProviderConfig_should_resolve_cluster_alias()
...
             using (var system = ActorSystem.Create(nameof(ActorRefProvidersConfigSpec), config))
             {
                 var ext = (ExtendedActorSystem) system;
                 ext.Provider.GetType().ShouldBe(actorProviderType);
+                system.Terminate().Wait(TimeSpan.FromSeconds(3)); // force the system to cleanup and shutdown
```
Aaronontheweb: IMHO, `Dispose` on an `ActorSystem` should probably call this and block until it's had a chance to run part of its shutdown process.

src/core/Akka.Cluster.Tests/ClusterSpec.cs
```diff
@@ -20,6 +20,12 @@ namespace Akka.Cluster.Tests
 {
     public class ClusterSpec : AkkaSpec
     {
+
+        /*
```
Aaronontheweb: Added portability note explaining some of the semantic differences with the specs.

src/core/Akka.Cluster.Tests/ClusterSpec.cs
```diff
@@ -257,7 +261,7 @@ public void A_cluster_must_cancel_LeaveAsync_task_if_CancellationToken_fired_bef
 
             // Subsequent LeaveAsync() tasks expected to complete immediately (not cancelled)
             var task3 = _cluster.LeaveAsync();
-            task3.Should(t => t.IsCompleted && !t.IsCanceled, "Task should be completed, but not cancelled.");
+            AwaitCondition(() => task3.IsCompleted && !task3.IsCanceled, null, "Task should be completed, but not cancelled.");
```
Aaronontheweb: Needed to be await-ed

src/core/Akka.Cluster.Tests/ClusterSpec.cs
```diff
@@ -214,7 +218,7 @@ public void A_cluster_must_return_completed_LeaveAsync_task_if_member_already_re
             });
 
             // LeaveAsync() task expected to complete immediately
-            _cluster.LeaveAsync().IsCompleted.Should().BeTrue();
+            AwaitCondition(() => _cluster.LeaveAsync().IsCompleted);
```
Aaronontheweb: Needed to be await-ed

src/core/Akka.Cluster.Tests/ClusterSpec.cs
```diff
@@ -237,7 +241,7 @@ public void A_cluster_must_cancel_LeaveAsync_task_if_CancellationToken_fired_bef
 
             // Cancelling the first task
             cts.Cancel();
-            task1.Should(t => t.IsCanceled, "Task should be cancelled.");
+            AwaitCondition(() => task1.IsCanceled, null, "Task should be cancelled");
```
Aaronontheweb: Needed to be await-ed

src/core/Akka.Cluster.Tests/ClusterSpec.cs
```diff
@@ -182,15 +185,16 @@ public void A_cluster_must_complete_LeaveAsync_task_upon_being_removed()
...
+            // A second call for LeaveAsync should complete immediately (should be the same task as before)
             Cluster.Get(sys2).LeaveAsync().IsCompleted.Should().BeTrue();
         }
 
-#if CORECLR
-        [Fact(Skip = "Fails on .NET Core")]
-#else
-        [Fact(Skip = "Fails flakily on .NET 4.5")]
-#endif
+//#if CORECLR
```
Aaronontheweb: Should be able to see this spec pass again

Aaronontheweb:
Going to work on `DowningProviderSpec.Downing_provider_should_stop_the_cluster_if_the_downing_provider_throws_exception_in_props` next

Aaronontheweb:
@akkadotnet/core ready for review.

src/core/Akka.Cluster.Tests/DowningProviderSpec.cs
```diff
@@ -59,7 +59,7 @@ public class DowningProviderSpec : AkkaSpec
             loglevel = WARNING
             actor.provider = ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""
             remote {
-              netty.tcp {
+              dot-netty.tcp {
```
alexvaluyskiy: Good catch

src/core/Akka.Tests/Routing/TailChoppingSpec.cs
```diff
@@ -127,8 +127,8 @@ public void Tail_chopping_group_router_must_return_response_from_second_actor_af
         [Fact]
         public void Tail_chopping_group_router_must_throw_exception_if_no_result_will_arrive_within_the_given_time()
         {
-            var actor1 = Sys.ActorOf(Props.Create(() => new TailChopTestActor(500.Milliseconds())), "Actor3");
-            var actor2 = Sys.ActorOf(Props.Create(() => new TailChopTestActor(500.Milliseconds())), "Actor4");
+            var actor1 = Sys.ActorOf(Props.Create(() => new TailChopTestActor(1500.Milliseconds())), "Actor3");
```
alexvaluyskiy: I think we should document all changes of timeout in the tests

Aaronontheweb: @alexvaluyskiy that's what commit histories are for - in this case the issue was just that the delay `500ms` and the timeout `300ms` were sufficiently close together that a busy system could receive the unwanted "ack" message before the delay. Nothing was technically wrong with the spec or the code it was testing otherwise.

Aaronontheweb:
Mentioned this already on https://github.com/akkadotnet/akka.net/issues/2728, but I'm splitting off some of the other racy specs into their own issue. Figured it'd be better to get some fixes in immediately rather than trying to do it all at once.

PR #2723, opened 2017-06-03
Obsoleted TypedActor

Horusiath:
We already encouraged users away from using `TypedActor` and were talking about removing it from the API but no steps were made. This PR adds obsoletes a `TypedActor` class.

Additional feature - it sets up an `akka.suppress-json-serializer-warning = on` in all tests. I think that having JSON serializer warning is useless in case of tests and only adds an unnecessary noise for test output (as it will be printed at least once per each test).

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -1686,6 +1686,7 @@ namespace Akka.Actor
         public TerminatedProps() { }
         public override Akka.Actor.ActorBase NewActor() { }
     }
+    [System.ObsoleteAttribute("TypedActor in its current shape will be removed in v1.5")]
     public abstract class TypedActor : Akka.Actor.ActorBase
     {
         protected TypedActor() { }
--- a/src/core/Akka.TestKit/Internal/Reference.conf
+++ b/src/core/Akka.TestKit/Internal/Reference.conf
@@ -10,7 +10,7 @@ akka {
 	# to use EventFiltering. If no filter is specified it logs to StdOut just like
 	# DefaultLogger.
   loggers = ["Akka.TestKit.TestEventListener, Akka.TestKit"]
-
+  suppress-json-serializer-warning = true
   test {
     # factor by which to scale timeouts during tests, e.g. to account for shared
     # build system load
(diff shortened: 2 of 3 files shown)


src/core/Akka.TestKit/Internal/Reference.conf
```diff
@@ -10,7 +10,7 @@ akka {
 	# to use EventFiltering. If no filter is specified it logs to StdOut just like
 	# DefaultLogger.
   loggers = ["Akka.TestKit.TestEventListener, Akka.TestKit"]
-
+  suppress-json-serializer-warning = true
```
alexvaluyskiy: I love this change

src/core/Akka/Actor/TypedActor.cs
```diff
@@ -25,6 +26,7 @@ public interface IHandle<in TMessage>
     /// <summary>
     ///     Class TypedActor.
     /// </summary>
+    [Obsolete("TypedActor in its current shape will be removed. Use ReceiveActor instead")]
```
alexvaluyskiy: Why ReceiveActor, not UntypedActor + Pattern matching?

src/core/Akka.TestKit/Internal/Reference.conf
Danthar: +1 for this change for sure

src/core/Akka/Actor/TypedActor.cs
Danthar: ReceiveActor is the default generic actor type most people are using. ReceiveActor has been the default actor type we have been recommending for a while now. It only makes sense to put that in here as well.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1686,6 +1686,7 @@ namespace Akka.Actor
         public TerminatedProps() { }
         public override Akka.Actor.ActorBase NewActor() { }
     }
+    [System.ObsoleteAttribute("TypedActor in its current shape will be removed in v1.5")]
```
Horusiath: Since there are multiple ways to replace TypedActor (the simplest one is using `Handle((dynamic)message)`), I think we can just give some deadline expectation to the users. Version 1.5 will already have breaking changes (change of default serializer, potentially more of them), we could remove this class then.

Issue #2670, opened 2017-05-14
Rewrite "Remote Transports" article


alexvaluyskiy:
This article quite outdated: http://akkanetdocfx.azurewebsites.net/articles/remoting/transports.html

PR #2725, opened 2017-06-05
Bringing Remote transports docs up to date. #2670

Danthar:
Updated the docs to illustrate the use of dot-netty instead of helios. I updated alot of links as well. Made them relative where i can, so the should point to the right document. But i can't be sure untill the site is updated.
Furthermore there is a section in there about how to configure multiple transports. That uses the TCP and UDP transport as an example. However how its written now suggest we actually have an UDP transport available for use. 
As far as i know, this isn't true. But i could be missing something. @Aaronontheweb  ?

Eitherway i'd like to change that part abit, to avoid confusion.

### DIFF
--- a/docs/articles/remoting/transports.md
+++ b/docs/articles/remoting/transports.md
@@ -6,7 +6,7 @@ title: Transports
 # Akka.Remote Transports
 In the [Akka.Remote overview](index.md) we introduced the concept of "transports" for Akka.Remote.
 
->  A"transport" refers to an actual network transport, such as TCP or UDP. By default Akka.Remote uses a [Helios](http://helios-io.github.io/ "Helios - Reactive socket middleware for .NET") TCP transport, but you could write your own transport and use that instead of you wish.
+>  A"transport" refers to an actual network transport, such as TCP or UDP. By default Akka.Remote uses a [DotNetty](https://github.com/Azure/DotNetty") TCP transport, but you could write your own transport and use that instead of you wish.
 
 In this section we'll expand a bit more on what transports are and how Akka.Remote can support multiple transports simultaneously.
 
@@ -19,7 +19,7 @@ Transports in Akka.Remote are abstractions on top of actual network transports, 
 Transports **do not need to care** about:
 * **Serialization** - that's handled by Akka.NET itself;
 * **Connection-oriented behavior** - the assocation process inside Akka.Remote ensures this, even over connectionless transports like UDP;
-* **Reliable delivery** - for system messages this is handled by Akka.Remote and for user-defined messages this is taken care of at the application level through something like the [`AtLeastOnceDeliveryActor` class](http://api.getakka.net/docs/stable/html/2FD30363.htm), part of Akka.Persistence;
+* **Reliable delivery** - for system messages this is handled by Akka.Remote and for user-defined messages this is taken care of at the application level through something like the [`AtLeastOnceDeliveryActor` class](/api/Akka.Persistence.AtLeastOnceDeliveryActor.html), part of Akka.Persistence;
 * **Handling network failures** - all a transport needs to do is forward that information back up to Akka.Remote.
 
 Transports **do need to care** about:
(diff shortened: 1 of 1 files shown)


docs/articles/remoting/transports.md
```diff
@@ -19,7 +19,7 @@ Transports in Akka.Remote are abstractions on top of actual network transports,
 Transports **do not need to care** about:
 * **Serialization** - that's handled by Akka.NET itself;
 * **Connection-oriented behavior** - the assocation process inside Akka.Remote ensures this, even over connectionless transports like UDP;
-* **Reliable delivery** - for system messages this is handled by Akka.Remote and for user-defined messages this is taken care of at the application level through something like the [`AtLeastOnceDeliveryActor` class](http://api.getakka.net/docs/stable/html/2FD30363.htm), part of Akka.Persistence;
+* **Reliable delivery** - for system messages this is handled by Akka.Remote and for user-defined messages this is taken care of at the application level through something like the [`AtLeastOnceDeliveryActor` class](/api/Akka.Persistence.AtLeastOnceDeliveryActor.html), part of Akka.Persistence;
```
alexvaluyskiy: Use `(xref:at-least-once-delivery)` instead

docs/articles/remoting/transports.md
```diff
@@ -77,7 +77,7 @@ akka{
 You'd define a custom HOCON section (`akka.remote.google-quic`) and let Akka.Remote know that it should read that section for a transport definition inside `akka.remote.enabled-transports`.
 
 > [!NOTE]
-> To implement a custom transport yourself, you need to implement the [`Akka.Remote.Transport.Transport` abstract class](http://api.getakka.net/docs/stable/html/F60EF1D2.htm).
+> To implement a custom transport yourself, you need to implement the [`Akka.Remote.Transport.Transport` abstract class](/api/Akka.Remote.Transport.Transport.html).
```
alexvaluyskiy: Add `uid:remoting-transport` and change the link to `(xref:remoting-transport)`

docs/articles/remoting/transports.md
```diff
@@ -111,8 +111,8 @@ akka{
 
 Both TCP and UDP are enabled in this scenario. But how do I know which transport is being used when I send a message to a `RemoteActorRef`? That's indicated by the protocol scheme used in the `Address` of the remote actor:
 
-    akka.tcp://MySystem@localhost:8081/user/actor #helios.tcp
-    akka.udp://MySystem@localhost:8082/user/actor #helios.udp
+    akka.tcp://MySystem@localhost:8081/user/actor #dot-netty.tcp
+    akka.udp://MySystem@localhost:8082/user/actor #dot-netty.udp
```
alexvaluyskiy: We don't support UDP atm

docs/articles/remoting/transports.md
Danthar: Im not linking to the document here, im linking to the api reference. If docfx has a syntax for that........  couldn't find it in their docs.

docs/articles/remoting/transports.md
Danthar: Yep. Im going to change that to a fictional protocol, so we still illustrate the point we are making there. But not give false expectations

docs/articles/remoting/transports.md
Danthar: nvm, figured it out

docs/articles/remoting/transports.md
```diff
@@ -33,20 +33,20 @@ Transports **do need to care** about:
...
 ## Akka.Remote's Built-in Transports
-Out of the box Akka.NET uses a socket-based transport built on top of the [Helios socket library](http://helios-io.github.io/).
+Out of the box Akka.NET uses a socket-based transport built on top of the [DotNetty](https://github.com/Azure/DotNetty).
 
 > [!NOTE]
-> Helios supports both TCP and UDP, but currently only TCP support is included within Akka.NET. TCP is what most Akka.Remote and Akka.Cluster users use.
+> DotNetty supports both TCP and UDP, but currently only TCP support is included within Akka.NET. TCP is what most Akka.Remote and Akka.Cluster users use.
 
-To enable the Helios TCP transport, we need to add a section for it inside our `remote` section in [HOCON configuration](/concepts/configuration.md):
+To enable the DotNetty TCP transport, we need to add a section for it inside our `remote` section in [HOCON configuration](/articles/concepts/configuration.md):
```
alexvaluyskiy: Should be `xref:configuration`

PR #2730, opened 2017-06-06
Serialization reload on settings change

zbynek001:
Serializers configurations from extension default config are not correctly loaded.
At the time extension tries to inject it's default config with InjectTopLevelFallback, serializer configuration was already loaded and it's not reloaded afterwards. Currently only workaround I know about is providing all fallback configurations when creating ActorSystem itself. This pr tries to improve this.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -1811,7 +1811,7 @@ namespace Akka.Actor.Internal
         public void UnstashAll() { }
         public void UnstashAll(System.Func<Akka.Actor.Envelope, bool> predicate) { }
     }
-    public class ActorSystemImpl : Akka.Actor.ExtendedActorSystem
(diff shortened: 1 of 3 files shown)


zbynek001:
Not sure if there is a better way how to handle this scenario.
And maybe it soon won't be needed with the change to persistence serialization

Aaronontheweb:
@zbynek001 which serializers do you have this problem for? User-defined ones?

zbynek001:
I've run into this with singleton & sharding. During startup, they are trying to inject default configs via `InjectTopLevelFallback` which works for most stuff, but not for "akka.actor.serializers" configuration. Those never get updated because they were already loaded

Danthar:
Looks good to me. We do the same thing for the Dispatchers subsystem as well. The change in itself is good.

zbynek001:
Should be fixed by #2952

zbynek001:
reason was that two nodes (those with sharding) were using ClusterShardingMessageSerializer, and the Control node was using hyperion so it was not able to deserialize some messages

PR #2738, opened 2017-06-08
Persistence.Tests bugfixing

heynickc:
Related to #2736, this enables Persistence tests that generally pass, ignoring SnapshotFailureRobustnessSpec.cs temporarily to help surface any other racy specs through CI while working on that class.


heynickc:
Resolved by #2763

PR #2742, opened 2017-06-08
[misc] Fix T4 templates in projects

sean-gilliam:
In older versions of VS, the template was the parent of the .cs file.
The current layout doesn't reflect this making both the T4 template and
the code live side by side. This clutters up the Solution Explorer
slightly.

This PR fixes the visual issue with those templates.


sean-gilliam:
Before
tt_before

After
tt_after

Issue #2743, opened 2017-06-08
Promote Akka.Persistence to stable


alexvaluyskiy:
# Stable.Persistence

## Major changes

- [x] Deprecate PersistenceView (https://github.com/akkadotnet/akka.net/pull/2747)
- [x] Remove an old SQL Query (Hints based) (https://github.com/akkadotnet/akka.net/pull/2747)
- [x] Remove all protobuf2 serializers and Google.ProtocolBuffers dependency from Persistence project. (https://github.com/akkadotnet/akka.net/pull/2748)
- [x] Make all Persistence modules NetStandard compatible without any fallbacks
- [x] Rename Persistence.Testkit to Persistence.TCK (https://github.com/akkadotnet/akka.net/pull/2741)
- [x] Add PersistenceQuery, performance and serialization tests to Persistence TCK (https://github.com/akkadotnet/akka.net/pull/2741)
- [x] Persistence Query become stable
- [x] PersistenceFSM become stable and ready for production (https://github.com/akkadotnet/akka.net/pull/2593)

## Minor changes
- [ ] [Allow persisting events when recovery has completed](https://github.com/akka/akka/pull/21893) (**up-for-grabs**)
- [x] [Limit number of concurrent PersistentActor recoveries](https://github.com/akka/akka/pull/22783) (https://github.com/akkadotnet/akka.net/pull/2655)
- [x] [Recover.none should immediately carry None snapshot](https://github.com/akka/akka/pull/23043)
- [ ] [Optimize highestSequenceNr method in AtomicWrite](https://github.com/akka/akka/pull/23059) (**up-for-grabs**)
- [x] [Use snapshot-after in config for periodical snapshots in PersistentFSM](https://github.com/akka/akka/pull/22605)
- [x] Change `AllPersistenceIds` to `PersistenceIds` query
- [x] Add pluggable offset strategies to EventsByTag query (https://github.com/akkadotnet/akka.net/pull/2851)

## Racy specs
- [ ] AtLeastOnceDelivery_should_not_send_when_actor_crashes (https://github.com/akkadotnet/akka.net/issues/1752)
- [ ] LocalSnapshotStore_configured_with_a_failing_directory_name_should_throw_an_exception... (https://github.com/akkadotnet/akka.net/issues/1751)
- [ ] AtLeastOnceDelivery_must_tolerate_and_recover_from_random_failures (https://github.com/akkadotnet/akka.net/issues/2229)

## Clean-up
[...]

PR #2747, opened 2017-06-10
Removed SQL Persistence Query and deprecated PersistentView

alexvaluyskiy:
Removed SQL Persistence Query
Deprecated PersistentView


src/core/Akka.Persistence/PersistentView.cs
```diff
@@ -1,317 +0,0 @@
...
-    /// to the view with a <see cref="SnapshotOffer"/> message, followed by replayed messages, if any, that are younger
-    /// than the snapshot. Default is to offer the latest saved snapshot.
-    /// 
-    /// By default, a view automatically updates itself with an interval returned by `autoUpdateInterval`.
-    /// This method can be overridden by implementation classes to define a view instance-specific update
-    /// interval. The default update interval for all views of an actor system can be configured with the
-    /// `akka.persistence.view.auto-update-interval` configuration key. Applications may trigger additional
-    /// view updates by sending the view <see cref="Update"/> requests. See also methods
-    /// </summary>
-    public abstract partial class PersistentView : ActorBase, ISnapshotter, IPersistentIdentity, IWithUnboundedStash, IPersistenceRecovery
```
alexvaluyskiy: `ISnapshotter` interface was implemented here and should be implemented in `Eventsourced` class (but not in our case). If we remove `PersistentView`, should we remove this useless interface too?

ismaelhamed: Was this advertised anywhere? This is a breaking change that will prevent us from upgrading to v1.3 right away.

alexvaluyskiy: We mention it in the documentation one year ago: http://getakka.net/docs/persistence/persistent-views

alexvaluyskiy: It is still discussable. I've created this meta-issue https://github.com/akkadotnet/akka.net/issues/2743 to discuss all things

alexvaluyskiy:
@ismaelhamed now PersistentView restored and deprecated

ismaelhamed:
> PersistentView was deprecated and will be removed in the next major version [1.3.0]

Good. But, did you mean version 1.4.0?

alexvaluyskiy:
@ismaelhamed We don't have any date for 1.4 yet. First of all, we should release Persistence Query support for all popular Persistence providers, then to check them in production. After that, we could decide to remove PersistenceView.
It is our fault that we did not deprecate `PersistenceView` before.

PR #2749, opened 2017-06-11
Delta-state CRDTs

Horusiath:
This PR introduces CRDTs upgrade, allowing us to optimize replication process by using deltas instead of sending full state changes across the network.

Code will need further upgrade (branch was originally targetting C# 6). Also currently known problem is the state of Hyperion (CRDTs are using custom Hyperion-based serializer), which seems to have some bugs on .NET Core implementation.


sean-gilliam:
very nice :+1:

Horusiath:
Ok, I think this PR is ready for review. This doesn't mean, that work here is done:

- We need to think about CRDT serialization (especially in context of persistence). Hyperion is not good choice for this, as it's binary format is not stable. The best candidates here would be protobuf or bond. Problem with Bond is that we already have a direct dependency on protobuf and adding the next serializer doesn't seem to be a good idea. Problem with protobuf is lack of support for generics, and CRDTs heavily rely on that.
- I needed to provide a generic-less versions of interfaces in few places (i.e. `IReplicatedData`, `IDeltaReplicatedData`) to have a common way of calling them from Replicator. Tbh I think, we could rework the methods a bit and try to use `dynamic` there - first call will be expensive, but subsequent calls are proven to be fast. We already use similar technique in Hyperion.
- I didn't ported all specs for persistence (those regarding size) and didn't fixed MNTK specs for cluster, as last time those were problematic due to some strange behavior in akka cluster itself.
- We still need performance tests, both for new data structures and replication protocol itself.

PR #2757, opened 2017-06-13
Akka.Cluster.Sharding: public events

### LINKED ISSUE #2756: Akka.Cluster.Sharding: making events public
as discussed in #2743, right now some events & states in sharding are internal.
With the move to custom serializers, they should be public

zbynek001:
fixes #2756


Aaronontheweb:
@zbynek001 needs a rebase

PR #2759, opened 2017-06-13
Place MNTR spec results into correct artifacts folder

heynickc:
Output MultiNode Test Runner results to TestResults/multinode so they show up as artifacts in CI.  Also ignore launchSettings.json in .gitignore


build.fsx
```diff
@@ -21,7 +21,7 @@ let outputTests = __SOURCE_DIRECTORY__ @@ "TestResults"
 let outputPerfTests = __SOURCE_DIRECTORY__ @@ "PerfResults"
 let outputBinaries = output @@ "binaries"
 let outputNuGet = output @@ "nuget"
-let outputMultiNode = output @@ "multinode"
+let outputMultiNode = outputTests @@ "multinode"
```
Aaronontheweb: Looks good

Aaronontheweb:
MNTR failure intermittent, but the logs are in the right place now at least

PR #2760, opened 2017-06-13
FrameworkPathOverride to 4.5.2 instead of 4.5

heynickc:
Supported workaround for running tests on Mono is to point an env variable `FrameworkPathOverride` to the mono installation directory.  This is an attempt to point to 4.5.2 folder (it was originally pointing at usr/lib/mono/4.5/  in case it has an affect on our test runs in CI since our test projects are at .NET 4.5.2


heynickc:
changed build.fsx `RunTests` to pass `--no-appdomain` for consistency with `dev` branch suite

Danthar:
Is this ready ?

heynickc:
@Danthar This didn't really make any difference in terms of CI so I'm going to close it for now and revisit it later if it becomes an actual issue.

PR #2763, opened 2017-06-14
Fixed local snapshot on linux

alexvaluyskiy:


src/core/Akka.Persistence/Snapshot/LocalSnapshotStore.cs
```diff
@@ -279,6 +279,8 @@ private IEnumerable<SnapshotMetadata> GetSnapshotMetadata(string persistenceId,
                 .Select(ExtractSnapshotMetadata)
                 .Where(metadata => metadata != null && criteria.IsMatch(metadata) && !_saving.Contains(metadata)).ToList();
 
+            snapshots.Sort(SnapshotMetadata.Comparer);
```
Aaronontheweb: nice catch

PR #2767, opened 2017-06-16
Sharding updated to v2.5.2

### LINKED ISSUE #1578: [Akka.Cluster.Sharding] ClusterShardingSpec MNTK spec
This is up for grabs task for making [ClusterShardingSpec](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Sharding.Tests/ClusterShardingSpec.cs) work on MultiNode TestKit Runner.

### LINKED ISSUE #1579: [Akka.Cluster.Sharding] ClusterShardingSpec MNTK spec
This is up for grabs task for making [ClusterShardingLeavingSpec](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Sharding.Tests/ClusterShardingLeavingSpec.cs) work on MultiNode TestKit Runner.

### LINKED ISSUE #1580: [Akka.Cluster.Sharding] ClusterShardingFailureSpec MNTK spec
This is up for grabs task for making [ClusterShardingFailureSpec](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Sharding.Tests/ClusterShardingFailureSpec.cs) work on MultiNode TestKit Runner.

### LINKED ISSUE #1581: [Akka.Cluster.Sharding] ClusterShardingGracefulShutdownSpec MNTK spec
This is up for grabs task for making [ClusterShardingGracefulShutdownSpec](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Sharding.Tests/ClusterShardingGracefulShutdownSpec.cs) work on MultiNode TestKit Runner.

zbynek001:
This PR updates sharding to version 2.5.2, including tests and multi-node tests.
All tests are passing on my end.

This also brings support for sharding entities without remembering feature, in the future DData support could be done in a similar way.

FYI:
**This update includes one breaking change**, which was introduced in akka 2.5.2:
If you are using Cluster Sharding with “remembering entities” there is one important change that requires that you make a change to your code when updating to 2.5.2. To solve a critical issue https://github.com/akka/akka/issues/22289 that could result in duplicate entities when the extractShardId function is changed (e.g. between deployments) there is now a new message Shard.StartEntity(entityId) that the extractShardId must understand.

close #1578
close #1579
close #1580
close #1581

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/Akka.Cluster.Sharding.Tests.MultiNode.csproj
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/Akka.Cluster.Sharding.Tests.MultiNode.csproj
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\serializers\Akka.Serialization.Hyperion\Akka.Serialization.Hyperion.csproj" />
     <ProjectReference Include="..\Akka.Cluster.Sharding\Akka.Cluster.Sharding.csproj" />
     <ProjectReference Include="..\..\..\core\Akka.Cluster.TestKit\Akka.Cluster.TestKit.csproj" />
     <ProjectReference Include="..\..\..\core\Akka.Tests.Shared.Internals\Akka.Tests.Shared.Internals.csproj" />
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
@@ -86,7 +86,7 @@ namespace Akka.Cluster.Sharding.Tests
     /// <summary>
     /// A journal that delegates actual storage to a target actor. For testing only.
     /// </summary>
-    public abstract class AsyncWriteProxy : AsyncWriteJournal, IWithUnboundedStash
+    public abstract class AsyncWriteProxyEx : AsyncWriteJournal, IWithUnboundedStash
     {
         private bool _isInitialized;
         private bool _isInitTimedOut;
@@ -95,7 +95,7 @@ namespace Akka.Cluster.Sharding.Tests
         /// <summary>
         /// TBD
         /// </summary>
-        protected AsyncWriteProxy()
+        protected AsyncWriteProxyEx()
         {
             _isInitialized = false;
             _isInitTimedOut = false;
@@ -210,9 +210,6 @@ namespace Akka.Cluster.Sharding.Tests
             }, TaskContinuationOptions.ExecuteSynchronously);
 
             return result.Task;
-
-
-            //return _store.AskEx(sender => new DeleteMessagesTo(persistenceId, toSequenceNr, sender), Timeout);
         }
 
         /// <summary>
@@ -255,18 +252,21 @@ namespace Akka.Cluster.Sharding.Tests
             if (_store == null)
                 return StoreNotInitialized<long>();
 
-            return _store.AskEx<object>(sender => new ReplayMessages(0, 0, 0, persistenceId, sender), Timeout)
+            var result = new TaskCompletionSource<long>();
+
+            _store.AskEx<object>(sender => new ReplayMessages(0, 0, 0, persistenceId, sender), Timeout)
                 .ContinueWith(t =>
                 {
-                    if (t.IsCanceled || t.IsFaulted)
-                        return 0L;
-
-                    if (t.Result is RecoverySuccess rs)
-                    {
-                        return rs.HighestSequenceNr;
-                    }
-                    return 0L;
-                }/*, TaskContinuationOptions.OnlyOnRanToCompletion*/, TaskContinuationOptions.ExecuteSynchronously);
+                    if (t.IsFaulted)
+                        result.TrySetException(t.Exception);
+                    else if (t.IsCanceled)
+                        result.TrySetException(new TimeoutException());
+                    else if (t.Result is RecoverySuccess rs)
+                        result.TrySetResult(rs.HighestSequenceNr);
+                    else
+                        result.TrySetException(new InvalidOperationException());
+                }, TaskContinuationOptions.ExecuteSynchronously);
+            return result.Task;
         }
 
         private Task<T> StoreNotInitialized<T>()
@@ -342,7 +342,6 @@ namespace Akka.Cluster.Sharding.Tests
                 //rm.Persistent
                 _replayCallback(rm.Persistent);
             }
-            //else if (message is IPersistentRepresentation) _replayCallback(message as IPersistentRepresentation);
             else if (message is RecoverySuccess)
             {
                 _replayCompletionPromise.SetResult(new object());
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
@@ -304,13 +304,13 @@ namespace Akka.Cluster.Sharding.Tests
                 }, _config.First);
                 RunOn(() =>
                 {
-                    LastSender.Path.Should().Be(Node(_config.First) / "user" / "sharding" / "Entity" / "2" / "2");
+                    LastSender.Path.Should().Be(Node(_config.First) / "system" / "sharding" / "Entity" / "2" / "2");
                 }, _config.Second);
                 EnterBarrier("second-started");
 
                 RunOn(() =>
                 {
-                    Sys.ActorSelection(Node(_config.Second) / "user" / "sharding" / "Entity").Tell(new Identify(null));
+                    Sys.ActorSelection(Node(_config.Second) / "system" / "sharding" / "Entity").Tell(new Identify(null));
                     var secondRegion = ExpectMsg<ActorIdentity>().Subject;
                     _allocator.Value.Tell(new UseRegion(secondRegion));
                     ExpectMsg<UseRegionAck>();
@@ -326,7 +326,7 @@ namespace Akka.Cluster.Sharding.Tests
 
                 RunOn(() =>
                 {
-                    LastSender.Path.Should().Be(Node(_config.Second) / "user" / "sharding" / "Entity" / "3" / "3");
+                    LastSender.Path.Should().Be(Node(_config.Second) / "system" / "sharding" / "Entity" / "3" / "3");
                 }, _config.First);
 
                 EnterBarrier("after-2");
@@ -348,7 +348,7 @@ namespace Akka.Cluster.Sharding.Tests
                         _region.Value.Tell(2, p.Ref);
                         p.ExpectMsg(2, TimeSpan.FromSeconds(2));
 
-                        p.LastSender.Path.Should().Be(Node(_config.Second) / "user" / "sharding" / "Entity" / "2" / "2");
+                        p.LastSender.Path.Should().Be(Node(_config.Second) / "system" / "sharding" / "Entity" / "2" / "2");
                     });
 
                     _region.Value.Tell(1);
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
@@ -20,6 +20,7 @@ using Akka.Event;
 using Akka.TestKit.TestActors;
 using System.Collections.Immutable;
 using System.Threading.Tasks;
+using FluentAssertions;
 
 namespace Akka.Cluster.Sharding.Tests
 {
@@ -259,7 +260,7 @@ namespace Akka.Cluster.Sharding.Tests
             {
                 Sys.ActorSelection(Node(_config.First) / "system" / "akka.persistence.journal.MemoryJournal").Tell(new Identify(null));
                 var sharedStore = ExpectMsg<ActorIdentity>(TimeSpan.FromSeconds(10)).Subject;
-                Assert.NotNull(sharedStore);
+                sharedStore.Should().NotBeNull();
 
                 MemoryJournalShared.SetStore(sharedStore, Sys);
             }, _config.First, _config.Second);
@@ -288,7 +289,7 @@ namespace Akka.Cluster.Sharding.Tests
                     ExpectMsg<UseRegionAck>();
                     _region.Value.Tell(1);
                     ExpectMsg(1);
-                    Assert.Equal(_region.Value.Path / "1" / "1", LastSender.Path);
+                    LastSender.Path.Should().Be(_region.Value.Path / "1" / "1");
                 }, _config.First);
                 EnterBarrier("first-started");
 
@@ -299,11 +300,11 @@ namespace Akka.Cluster.Sharding.Tests
 
                 RunOn(() =>
                 {
-                    Assert.Equal(_region.Value.Path / "2" / "2", LastSender.Path);
+                    LastSender.Path.Should().Be(_region.Value.Path / "2" / "2");
                 }, _config.First);
                 RunOn(() =>
                 {
-                    Assert.Equal(Node(_config.First) / "user" / "sharding" / "Entity" / "2" / "2", LastSender.Path);
+                    LastSender.Path.Should().Be(Node(_config.First) / "user" / "sharding" / "Entity" / "2" / "2");
                 }, _config.Second);
                 EnterBarrier("second-started");
 
@@ -320,12 +321,12 @@ namespace Akka.Cluster.Sharding.Tests
                 ExpectMsg(3);
                 RunOn(() =>
                 {
-                    Assert.Equal(_region.Value.Path / "3" / "3", LastSender.Path);
+                    LastSender.Path.Should().Be(_region.Value.Path / "3" / "3");
                 }, _config.Second);
 
                 RunOn(() =>
                 {
-                    Assert.Equal(Node(_config.Second) / "user" / "sharding" / "Entity" / "3" / "3", LastSender.Path);
+                    LastSender.Path.Should().Be(Node(_config.Second) / "user" / "sharding" / "Entity" / "3" / "3");
                 }, _config.First);
 
                 EnterBarrier("after-2");
@@ -347,12 +348,12 @@ namespace Akka.Cluster.Sharding.Tests
                         _region.Value.Tell(2, p.Ref);
                         p.ExpectMsg(2, TimeSpan.FromSeconds(2));
 
-                        Assert.Equal(Node(_config.Second) / "user" / "sharding" / "Entity" / "2" / "2", p.LastSender.Path);
+                        p.LastSender.Path.Should().Be(Node(_config.Second) / "user" / "sharding" / "Entity" / "2" / "2");
                     });
 
                     _region.Value.Tell(1);
                     ExpectMsg(1);
-                    Assert.Equal(_region.Value.Path / "1" / "1", LastSender.Path);
+                    LastSender.Path.Should().Be(_region.Value.Path / "1" / "1");
                 }, _config.First);
                 EnterBarrier("after-2");
             });
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
@@ -1,0 +1,361 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ClusterShardingCustomShardAllocationSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.IO;
+using System.Linq;
+using Akka.Actor;
+using Akka.Cluster.TestKit;
+using Akka.Cluster.Tests.MultiNode;
+using Akka.Configuration;
+using Akka.Persistence.Journal;
+using Akka.Remote.TestKit;
+using Akka.Remote.Transport;
+using Xunit;
+using Akka.Event;
+using Akka.TestKit.TestActors;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+
+namespace Akka.Cluster.Sharding.Tests
+{
+    public class ClusterShardingCustomShardAllocationSpecConfig : MultiNodeConfig
+    {
+        public RoleName First { get; private set; }
+
+        public RoleName Second { get; private set; }
+
+        public ClusterShardingCustomShardAllocationSpecConfig()
+        {
+            First = Role("first");
+            Second = Role("second");
+
+            CommonConfig = DebugConfig(false)
+                .WithFallback(ConfigurationFactory.ParseString(@"
+                    akka.actor {
+                        serializers {
+                            hyperion = ""Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion""
+                        }
+                        serialization-bindings {
+                            ""System.Object"" = hyperion
+                        }
+                    }
+
+                    akka.persistence.snapshot-store.plugin = ""akka.persistence.snapshot-store.inmem""
+                    akka.persistence.journal.plugin = ""akka.persistence.journal.memory-journal-shared""
+
+                    akka.persistence.journal.MemoryJournal {
+                        class = ""Akka.Persistence.Journal.MemoryJournal, Akka.Persistence""
+                        plugin-dispatcher = ""akka.actor.default-dispatcher""
+                    }
+
+                    akka.persistence.journal.memory-journal-shared {
+                        class = ""Akka.Cluster.Sharding.Tests.MemoryJournalShared, Akka.Cluster.Sharding.Tests.MultiNode""
+                        plugin-dispatcher = ""akka.actor.default-dispatcher""
+                        timeout = 5s
+                    }
+                "))
+                .WithFallback(MultiNodeClusterSpec.ClusterConfig());
+        }
+    }
+
+    public class ClusterShardingCustomShardAllocationSpec : MultiNodeClusterSpec
+    {
+        #region setup
+
+        internal class Entity : ActorBase
+        {
+            protected override bool Receive(object message)
+            {
+                switch (message)
+                {
+                    case int id:
+                        Sender.Tell(id);
+                        return true;
+                }
+                return false;
+            }
+        }
+
+        internal class AllocateReq
+        {
+            public static readonly AllocateReq Instance = new AllocateReq();
+
+            private AllocateReq()
+            {
+            }
+        }
+
+        internal class UseRegion
+        {
+            public readonly IActorRef Region;
+
+            public UseRegion(IActorRef region)
+            {
+                Region = region;
+            }
+        }
+
+        internal class UseRegionAck
+        {
+            public static readonly UseRegionAck Instance = new UseRegionAck();
+
+            private UseRegionAck()
+            {
+            }
+        }
+
+        internal class RebalanceReq
+        {
+            public static readonly RebalanceReq Instance = new RebalanceReq();
+
+            private RebalanceReq()
+            {
+            }
+        }
+
+        internal class RebalanceShards
+        {
+            public readonly IImmutableSet<string> Shards;
+
+            public RebalanceShards(IImmutableSet<string> shards)
+            {
+                Shards = shards;
+            }
+        }
+
+        internal class RebalanceShardsAck
+        {
+            public static readonly RebalanceShardsAck Instance = new RebalanceShardsAck();
+
+            private RebalanceShardsAck()
+            {
+            }
+        }
+
+        internal class Allocator : ActorBase
+        {
+            IActorRef UseRegion;
+            IImmutableSet<string> Rebalance = ImmutableHashSet<string>.Empty;
+
+            protected override bool Receive(object message)
+            {
+                switch (message)
+                {
+                    case UseRegion r:
+                        UseRegion = r.Region;
+                        Sender.Tell(UseRegionAck.Instance);
+                        return true;
+                    case AllocateReq _:
+                        if (UseRegion != null)
+                            Sender.Tell(UseRegion);
+                        return true;
+                    case RebalanceShards rs:
+                        Rebalance = rs.Shards;
+                        Sender.Tell(RebalanceShardsAck.Instance);
+                        return true;
+                    case RebalanceReq _:
+                        Sender.Tell(Rebalance);
+                        Rebalance = ImmutableHashSet<string>.Empty;
+                        return true;
+                }
+                return false;
+            }
+        }
+
+        class TestAllocationStrategy : IShardAllocationStrategy
+        {
+            public readonly IActorRef Ref;
+
+            public TestAllocationStrategy(IActorRef @ref)
+            {
+                Ref = @ref;
+            }
+
+            public Task<IActorRef> AllocateShard(IActorRef requester, string shardId, IImmutableDictionary<IActorRef, IImmutableList<string>> currentShardAllocations)
+            {
+                return Ref.Ask<IActorRef>(AllocateReq.Instance);
+            }
+
+            public Task<IImmutableSet<string>> Rebalance(IImmutableDictionary<IActorRef, IImmutableList<string>> currentShardAllocations, IImmutableSet<string> rebalanceInProgress)
+            {
+                return Ref.Ask<IImmutableSet<string>>(RebalanceReq.Instance);
+            }
+        }
+
+        internal IdExtractor extractEntityId = message => message is int ? Tuple.Create(message.ToString(), message) : null;
+
+        internal ShardResolver extractShardId = message => message is int ? ((int)message).ToString() : null;
+
+        private Lazy<IActorRef> _region;
+        private Lazy<IActorRef> _allocator;
+
+        private readonly ClusterShardingCustomShardAllocationSpecConfig _config;
+
+        public ClusterShardingCustomShardAllocationSpec()
+            : this(new ClusterShardingCustomShardAllocationSpecConfig())
+        {
+        }
+
+        protected ClusterShardingCustomShardAllocationSpec(ClusterShardingCustomShardAllocationSpecConfig config)
+            : base(config)
+        {
+            _config = config;
+
+            _region = new Lazy<IActorRef>(() => ClusterSharding.Get(Sys).ShardRegion("Entity"));
+
+            _allocator = new Lazy<IActorRef>(() => Sys.ActorOf(Props.Create<Allocator>(), "allocator"));
+        }
+
+        protected override int InitialParticipantsValueFactory { get { return Roles.Count; } }
+
+        #endregion
+
+        private void Join(RoleName from, RoleName to)
+        {
+            RunOn(() =>
+            {
+                Cluster.Join(GetAddress(to));
+                StartSharding();
+            }, from);
+            EnterBarrier(from.Name + "-joined");
+        }
+
+        private void StartSharding()
+        {
+            ClusterSharding.Get(Sys).Start(
+                typeName: "Entity",
+                entityProps: Props.Create<Entity>(),
+                settings: ClusterShardingSettings.Create(Sys),
+                idExtractor: extractEntityId,
+                shardResolver: extractShardId,
+                allocationStrategy: new TestAllocationStrategy(_allocator.Value),
+                handOffStopMessage: PoisonPill.Instance);
+        }
+
+        [MultiNodeFact]
+        public void Cluster_sharding_with_custom_allocation_strategy_specs()
+        {
+            Cluster_sharding_with_custom_allocation_strategy_should_setup_shared_journal();
+            Cluster_sharding_with_custom_allocation_strategy_should_use_specified_region();
+            Cluster_sharding_with_custom_allocation_strategy_should_rebalance_specified_shards();
+        }
+
+        public void Cluster_sharding_with_custom_allocation_strategy_should_setup_shared_journal()
+        {
+            // start the Persistence extension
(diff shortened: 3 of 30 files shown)


zbynek001:
If anybody's interested, this is now ready for review.

alexvaluyskiy:
@zbynek001 I will take it. have you fixed all Sharding tests?

zbynek001:
yes, all tests were updated and created if missing

src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
```diff
@@ -88,7 +88,12 @@ public virtual object EntityMessage(object message)
         /// <returns>TBD</returns>
         public virtual string ShardId(object message)
         {
-            return (Math.Abs(EntityId(message).GetHashCode())%MaxNumberOfShards).ToString();
+            EntityId id;
+            if (message is ShardRegion.StartEntity se)
```
zbynek001: ShardRegion.StartEntity now needs to be handled in all shard extractors

src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
```diff
@@ -19,11 +19,44 @@ namespace Akka.Cluster.Sharding
...
+        }
+    }
+
+
     //TODO: figure out how not to derive from persistent actor for the sake of alternative ddata based impl
     /// <summary>
     /// TBD
     /// </summary>
-    public abstract class Shard : PersistentActor
+    public class Shard
```
zbynek001: Shard is now extracted from actor, and only using actor context inside

src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
```diff
@@ -19,11 +19,44 @@ namespace Akka.Cluster.Sharding
     using EntityId = String;
     using Msg = Object;
 
+    public class ShardActor : ActorBase
```
zbynek001: proxy actor delegating messages to Shard logic

src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
```diff
@@ -306,11 +303,29 @@ public override int GetHashCode()
...
+            ClusterShardingSettings settings,
+            IdExtractor extractEntityId,
+            ShardResolver extractShardId,
+            object handOffStopMessage)
+        {
+            if (settings.RememberEntities)
+                return Actor.Props.Create(() => new PersistentShardActor(typeName, shardId, entityProps, settings, extractEntityId, extractShardId, handOffStopMessage))
+                .WithDeploy(Deploy.Local);
+            else
+                return Actor.Props.Create(() => new ShardActor(typeName, shardId, entityProps, settings, extractEntityId, extractShardId, handOffStopMessage))
```
zbynek001: support for non-persistent shards

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
```diff
@@ -10,29 +10,77 @@
...
+            return _shardSemantic.ReceiveCommand(message);
+        }
+
+        protected override bool ReceiveRecover(object message)
+        {
+            return _shardSemantic.ReceiveRecover(message);
+        }
+    }
+
+    public interface IPersistentActorContext
```
zbynek001: Ideally something like this should be in persistence directly

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
```diff
@@ -10,29 +10,77 @@
...
 using Akka.Util.Internal;
+using System.Threading.Tasks;
 
 namespace Akka.Cluster.Sharding
 {
     using ShardId = String;
     using EntryId = String;
     using Msg = Object;
 
+    public class PersistentShardActor : PersistentActor, IPersistentActorContext
```
zbynek001: proxy actor for PersistentShard

src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
```diff
@@ -220,14 +217,14 @@ public sealed class CurrentShardState
             /// <summary>
             /// TBD
             /// </summary>
-            public readonly string[] EntityIds;
+            public readonly IImmutableSet< string> EntityIds;
```
zbynek001: most of the things were changed to immutable

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/MemoryJournalShared.cs
```diff
@@ -0,0 +1,25 @@
...
+using Akka.Persistence.Journal;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Akka.Cluster.Sharding.Tests
+{
+    public class MemoryJournalShared : AsyncWriteProxyEx
```
zbynek001: MemoryJournal which can be shared between nodes

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
```diff
@@ -0,0 +1,461 @@
...
+        /// <summary>
+        /// TBD
+        /// </summary>
+        public readonly IActorRef Store;
+    }
+
+    /// <summary>
+    /// A journal that delegates actual storage to a target actor. For testing only.
+    /// </summary>
+    public abstract class AsyncWriteProxyEx : AsyncWriteJournal, IWithUnboundedStash
```
zbynek001: Enhanced version of AsyncWriteProxy, since the one in Persistence didn't work out.
It's simulating AsyncWriteJournal with sending messages to the read actor via Ask with timeout handling

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
```diff
@@ -0,0 +1,461 @@
...
+                var timeoutException = new AsyncReplayTimeoutException($"Replay timed out after {_replayTimeout.TotalSeconds}s of inactivity");
+                _replayCompletionPromise.SetException(timeoutException);
+                Context.Stop(Self);
+            }
+            else return false;
+            return true;
+        }
+    }
+
+    internal static class FuturesEx
```
zbynek001: Enhanced version of Ask, which provide the real sender address. Needed this since persistence is using Sender from the commands

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/Akka.Cluster.Sharding.Tests.MultiNode.csproj
```diff
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\serializers\Akka.Serialization.Hyperion\Akka.Serialization.Hyperion.csproj" />
```
zbynek001: Doesn't work without hyperion, messages for MemoryJournalShared are not serialized/deserialized properly

Aaronontheweb:
@zbynek001 conflicts here from merging your previous PR

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/Akka.Cluster.Sharding.Tests.MultiNode.csproj
zbynek001: Hyperion is not used for sharding itself, only for MemoryJournalShared messages

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -202,23 +239,96 @@ private void UpdateState(CounterChanged e)
...
+        {
+        }
+
+        protected ClusterShardingSpecRecoveryAll(ClusterShardingSpecConfig config)
+            : base(config)
+        {
+        }
+    }
+
+    //public class ClusterShardingSpecRecoveryConstant : ClusterShardingSpec
```
zbynek001: didn't find a way yet how to run one spec against multiple configs.
Since the MultiNodeTestRunner is using Activator.CreateInstance to create new instance of config objects, it's not possible to have parameters on the ClusterShardingSpecConfig constructor. Is there a way around? I didn't want to duplicate whole config class.

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -387,21 +505,18 @@ public void ClusterSharding_should_work_in_single_node_cluster()
...
             });
         }
 
-        [MultiNodeFact(Skip = "TODO")]
         public void ClusterSharding_should_use_second_node()
         {
-            ClusterSharding_should_work_in_single_node_cluster();
-
-            Within(TimeSpan.FromSeconds(20), () =>
+            Within(TimeSpan.FromSeconds(30), () =>
```
zbynek001: probably a leftover from my tests, will fix

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -738,11 +825,20 @@ public void ClusterSharding_should_be_easy_to_use_with_extensions()
...
                     ClusterSharding.Get(Sys).Start(
                         typeName: "AnotherCounter",
-                        entityProps: Props.Create<Counter>(),
+                        entityProps: Props.Create<AnotherCounter>(),
                         settings: ClusterShardingSettings.Create(Sys),
                         idExtractor: Counter.ExtractEntityId,
                         shardResolver: Counter.ExtractShardId);
-                }, _third, _fourth, _fifth, _sixth);
+
+                    //#counter-supervisor-start
```
zbynek001: I know, but was like this in akka. Will remove it

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -156,14 +191,15 @@ private Stop()
         public static readonly ShardResolver ExtractShardId = message =>
         {
             if (message is EntityEnvelope)
-                return (((EntityEnvelope)message).Id % ShardsCount).ToString();
+                return (((EntityEnvelope)message).Id % NumberOfShards).ToString();
```
zbynek001: i didn't want to use any c# 7 in existing code to keep the # of changes minimal.
Was planning to update this in another pr later on. But if you want, I can change everything to c# 7, but this will increase changes significantly. Shall I change it then in this PR or later?

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -557,20 +656,17 @@ public void ClusterSharding_should_failover_shards_on_crashed_node()
...
         public void ClusterSharding_should_use_third_and_fourth_node()
         {
-            ClusterSharding_should_failover_shards_on_crashed_node();
-
             Within(TimeSpan.FromSeconds(15), () =>
             {
-                Join(_third, _first);
-                Join(_fourth, _first);
+                Join(_config.Third, _config.First);
+                Join(_config.Fourth, _config.First);
```
zbynek001: right

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -869,22 +968,19 @@ public void Persistent_cluster_shards_should_recover_entities_upon_restart()
...
         }
 
-        [MultiNodeFact(Skip = "TODO")]
-        public void Persistent_cluster_shards_should_permanently_stop_entities_which_passivate()
+        public void PersistentClusterShards_should_permanently_stop_entities_which_passivate()
         {
-            Persistent_cluster_shards_should_recover_entities_upon_restart();
-
-            Within(TimeSpan.FromSeconds(15), () =>
+            Within(TimeSpan.FromSeconds(20), () =>
```
zbynek001: right, should be 15

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -830,7 +922,7 @@ public void Persistent_cluster_shards_should_recover_entities_upon_restart()
                     var shard = Sys.ActorSelection(LastSender.Path.Parent);
                     var region = Sys.ActorSelection(LastSender.Path.Parent.Parent);
 
-                    // stop shard
+                    //Stop the shard cleanly
                     region.Tell(new PersistentShardCoordinator.HandOff("1"));
                     ExpectMsg<PersistentShardCoordinator.ShardStopped>(s => s.Shard == "1", TimeSpan.FromSeconds(10));
```
zbynek001: right, it's the hint stuff. This is probably on more places, will fix

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -738,11 +825,20 @@ public void ClusterSharding_should_be_easy_to_use_with_extensions()
...
                         shardResolver: Counter.ExtractShardId);
-                }, _third, _fourth, _fifth, _sixth);
+
+                    //#counter-supervisor-start
+                    ClusterSharding.Get(Sys).Start(
+                      typeName: "SupervisedCounter",
+                      entityProps: Props.Create<CounterSupervisor>(),
+                      settings: ClusterShardingSettings.Create(Sys),
+                      idExtractor: Counter.ExtractEntityId,
+                      shardResolver: Counter.ExtractShardId);
```
zbynek001: Originally I have rename idExtractor & shardResolver to match akka, but it increased number of changes quite a lot. So I've reverted it to keep the # of changes minimal. If you want, I can rename them

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -429,16 +544,16 @@ public void ClusterSharding_should_use_second_node()
                     r.Tell(new Counter.EntityEnvelope(2, Counter.Increment.Instance));
                     r.Tell(new Counter.Get(2));
                     ExpectMsg(3);
-                    Assert.Equal(Node(_second) / "user" / "counterRegion" / "2" / "2", LastSender.Path);
+                    Assert.Equal(Node(_config.Second) / "user" / "counterRegion" / "2" / "2", LastSender.Path);
```
zbynek001: ok, will change

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
alexvaluyskiy: @zbynek001 If the original implementation of `AsyncWriteProxy` inside Persistence package has some bugs, we should fix them there first

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
alexvaluyskiy: I got you point. Yes, we can make some code clean-up in the subsequent pull requests

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
zbynek001: The original implementation in Akka.Persistence is using AsyncWriteTarget messages and can only wrap jornals, which implements them. But there is no journal supporting them.
 
AsyncWriteProxyEx is using standard messages from JournalProtocol and is able to wrap journals implementing AsyncWriteJournal

alexvaluyskiy: How was it implemented on Scala?

zbynek001: it's using AsyncWriteProxy together with SharedLeveldbStore which supports AsyncWriteTarget

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -349,32 +433,66 @@ private IActorRef CreateRegion(string typeName, bool rememberEntities)
...
+                var sharedStore = ExpectMsg<ActorIdentity>(TimeSpan.FromSeconds(10)).Subject;
+                Assert.NotNull(sharedStore);
 
+                MemoryJournalShared.SetStore(sharedStore, Sys);
+            }, _config.First, _config.Second, _config.Third, _config.Fourth, _config.Fifth, _config.Sixth);
             EnterBarrier("after-1");
+
+            RunOn(() =>
+            {
+                //check persistence running
```
zbynek001: yes, it's not there. It's checking if the shared journal actor is accessible from other nodes

Horusiath:
>[NODE2:first][2:23 PM]: SPEC FAILED: Akka.Cluster.Sharding.Tests.ClusterShardingSpecRecoveryAll.ClusterSharding_specs
 FAIL
[NODE4:third][2:23 PM]: SPEC FAILED: Akka.Cluster.Sharding.Tests.ClusterShardingSpecRecoveryAll.ClusterSharding_specs
 FAIL
[NODE5:fourth][2:23 PM]: SPEC FAILED: Akka.Cluster.Sharding.Tests.ClusterShardingSpecRecoveryAll.ClusterSharding_specs
 FAIL
[NODE1:controller][2:23 PM]: SPEC FAILED: Akka.Cluster.Sharding.Tests.ClusterShardingSpecRecoveryAll.ClusterSharding_specs
 FAIL
[NODE6:fifth][2:24 PM]: SPEC FAILED: Akka.Cluster.Sharding.Tests.ClusterShardingSpecRecoveryAll.ClusterSharding_specs
 FAIL
[NODE7:sixth][2:24 PM]: SPEC FAILED: Akka.Cluster.Sharding.Tests.ClusterShardingSpecRecoveryAll.ClusterSharding_specs
 FAIL

It looks like cluster sharding multi node test kit is broken. Maybe an update there is needed as well. @zbynek001 could you rebase the PR?

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
```diff
@@ -627,22 +698,45 @@ private void HandleRegister(Register message)
...
+        private void HandleGetClusterShardingStats(GetClusterShardingStats message)
+        {
+            var sender = Sender;
+            Task.WhenAll(
+                _aliveRegions.Select(regionActor => regionActor.Ask<ShardRegionStats>(GetShardRegionStats.Instance, message.Timeout).ContinueWith(r => Tuple.Create(regionActor, r.Result), TaskContinuationOptions.ExecuteSynchronously))
+                ).ContinueWith(allRegionStats =>
+                {
+                    if (allRegionStats.IsFaulted || allRegionStats.IsCanceled)
+                    {
+                        return new ClusterShardingStats(ImmutableDictionary<Address, ShardRegionStats>.Empty);
```
Arkatufus: Shouldn't we return an empty `ClusterShardingStats` on a timeout only, and not on all fault case?

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
```diff
@@ -627,22 +698,45 @@ private void HandleRegister(Register message)
...
+                            return address;
+                        }, j => j.Item2);
+
+                        return new ClusterShardingStats(regions);
+                    }
+                }, TaskContinuationOptions.ExecuteSynchronously).PipeTo(sender);
+        }
+
+
         private void SaveSnapshotIfNeeded()
```
Arkatufus: This should be renamed to `SaveSnapshotWhenNeeded()`

src/contrib/cluster/Akka.Cluster.Sharding/ClusterShardingGuardian.cs
```diff
@@ -160,49 +160,77 @@ public ClusterShardingGuardian()
...
+
+                    if (Equals(shardRegion, ActorRefs.Nobody))
+                    {
+
+                        shardRegion = Context.ActorOf(ShardRegion.ProxyProps(
+                            typeName: startProxy.TypeName,
+                            settings: settings,
+                            coordinatorPath: coordinatorPath,
+                            extractEntityId: startProxy.ExtractEntityId,
+                            extractShardId: startProxy.ExtractShardId), encName);
```
zbynek001: right, fixed

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
```diff
@@ -45,29 +93,39 @@ public class PersistentShard : Shard
         /// <param name="extractShardId">TBD</param>
         /// <param name="handOffStopMessage">TBD</param>
         public PersistentShard(
+            IActorContext context,
+            IPersistentActorContext persistentContext,
+            Action<object> unhandled,
```
zbynek001: this is not a public api, only internal

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
```diff
@@ -253,7 +290,12 @@ internal static Props Props(string typeName, ClusterShardingSettings settings, I
         /// <summary>
         /// TBD
         /// </summary>
-        public readonly TimeSpan DownRemovalMargin;
+        public readonly TimeSpan RemovalMargin;
+
+        public readonly int MinMembers;
```
zbynek001: not really. I left access modifiers as they were

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
zbynek001: renamed

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -429,59 +545,55 @@ public void ClusterSharding_should_use_second_node()
...
                     ExpectMsg(3);
-                    Assert.Equal(Node(_second) / "user" / "counterRegion" / "2" / "2", LastSender.Path);
+                    LastSender.Path.Should().Be(Node(_config.Second) / "user" / "counterRegion" / "2" / "2");
 
                     r.Tell(new Counter.Get(11));
                     ExpectMsg(1);
-                    // local on first
-                    Assert.Equal(r.Path / "11" / "11", LastSender.Path);
+                    var path11 = LastSender.Path;
+                    LastSender.Path.ToStringWithoutAddress().Should().Be((r.Path / "11" / "11").ToStringWithoutAddress());
```
zbynek001: had to change the spec here.
When the shard "11" is being created, the sharding status is: 
- first node: ["1"]
- second node: ["2"]
Since LeastShardAllocationStrategy has no ordering between nodes with the same number of shards, the next shard could be placed to either of them, right? 
Not sure why akka assumes that shard "11" should be local.

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
zbynek001: Because now there are two different actors with the similar logic. 
There is 
- `ShardActor` - for shards without remembering entities
- `PersistentShardActor` - for shards with remembering entities using persistence
- maybe in the future also `DDataShardActor` - for shards with remembering entities using DData

The logic is now extracted into `Shard` and `PersistentShard` with inheritance and those actors only sends incoming messages into the logic object

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
zbynek001: What should happen on other faults then? Not returning anything?

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
Arkatufus: Yes, but it breaks code compatibility between .net and java Akka, making future code port harder. And even if it is internal, someone might be using it since it is public.

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
Arkatufus: I think it is better if we change this to private for better encapsulation.

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
Arkatufus: well, the original code is
```scala
.recover {
  case x: AskTimeoutException ⇒ ShardRegion.ClusterShardingStats(Map.empty)
}
```
so it should do nothing on other fault cases.

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
zbynek001: This is similar to how AtLeastOnceDelivery is implemented.
If you have better idea how to get around scala traits, without code duplication, i'm open to suggestions.

Horusiath: I'll take a look at it. I've already forked @zbynek001 branch in order to continue porting DData shards from there. Maybe I'll propose some changes to the API - I think those changes will have to be taken into account before release. 

In general we try to achieve scala traits trough composition, when inheritance is not enough. For persistence this was somewhat mitigated in the past, but I guess we cannot delay it if we want to move forward.

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
zbynek001:
```
.recover {
    case x: AskTimeoutException ⇒ ShardRegion.ClusterShardingStats(Map.empty)
}.pipeTo(sender())
```
In case of different exception, doesn't it pipe the exception with the PipeTo?
I'm not scala friend so not sure :)

Arkatufus: For now, can you try rethrowing with a todo line and see if it fails the tests or not? If they do fail, you can let it silently swallow the exception but log an error that it shouldn't actually do that so we can track it down in the future?

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
```diff
@@ -0,0 +1,362 @@
...
+                _region.Value.Tell(2);
+                ExpectMsg(2);
+
+                RunOn(() =>
+                {
+                    LastSender.Path.Should().Be(_region.Value.Path / "2" / "2");
+                }, _config.First);
+                RunOn(() =>
+                {
+                    LastSender.Path.Should().Be(Node(_config.First) / "user" / "sharding" / "Entity" / "2" / "2");
```
alexvaluyskiy: It is has a path `/user/sharding/...` while Akka JVM has `/system/sharding`

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/Akka.Cluster.Sharding.Tests.MultiNode.csproj
alexvaluyskiy: Why we can't use the default serializer for it? Json.NET?

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
```diff
@@ -180,35 +181,51 @@ private void StartSharding()
...
         }
 
-        [MultiNodeFact(Skip = "TODO")]
-        public void ClusterSharding_with_flaky_journal_should_join_cluster()
+        public void ClusterSharding_with_flaky_journal_network_should_join_cluster()
         {
-            ClusterSharding_with_flaky_journal_should_setup_shared_journal();
-
-            Within(TimeSpan.FromSeconds(20), () =>
+            Within(TimeSpan.FromSeconds(30), () =>
```
alexvaluyskiy: Why 30 sec, not 20? If 20 sec does not work, please leave a comment about it in the code

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingGracefulShutdownSpec.cs
```diff
@@ -137,80 +129,126 @@ private void StartSharding()
...
+                        entityProps: Props.Create<Entity>(),
+                        settings: ClusterShardingSettings.Create(Sys),
+                        idExtractor: extractEntityId,
+                        shardResolver: extractShardId,
+                        allocationStrategy: allocationStrategy,
+                        handOffStopMessage: StopEntity.Instance);
+
+                    Watch(regionEmpty);
+                    regionEmpty.Tell(GracefulShutdown.Instance);
+                    ExpectTerminated(regionEmpty);
```
alexvaluyskiy: Should be `expectTerminated(regionEmpty, 5.seconds)` by default it is 3 seconds

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingMinMembersSpec.cs
```diff
@@ -0,0 +1,243 @@
...
+                EnterBarrier("shards-allocated");
+
+                _region.Value.Tell(new GetClusterShardingStats(Remaining));
+
+                var stats = ExpectMsg<ClusterShardingStats>();
+                var firstAddress = Node(_config.First).Address;
+                var secondAddress = Node(_config.Second).Address;
+                var thirdAddress = Node(_config.Third).Address;
+
+                stats.Regions.Keys.Should().BeEquivalentTo(new Address[] { firstAddress, secondAddress, thirdAddress });
```
alexvaluyskiy: @Aaronontheweb Scala tests contain `withClue`. How to translate it to C#/Xunit properly?
```
withClue(stats) {
    stats.regions.keySet should ===(Set(firstAddress, secondAddress, thirdAddress))
    stats.regions(firstAddress).stats.valuesIterator.sum should ===(1)
}
```

src/contrib/cluster/Akka.Cluster.Sharding.Tests/LeastShardAllocationStrategySpec.cs
```diff
@@ -57,20 +58,35 @@ public void LeastShardAllocationStrategy_should_reallocate_from_region_with_most
...
+            {
+                {_regionA, new []{"shard1"}.ToImmutableList() },
+                {_regionB, new []{ "shard2", "shard3", "shard4", "shard5", "shard6" }.ToImmutableList() },
+                {_regionC, ImmutableList<string>.Empty}
+            }.ToImmutableDictionary();
+
+            var r1 = _allocationStrategy.Rebalance(allocations, ImmutableHashSet<string>.Empty).Result;
+            r1.Should().BeEquivalentTo(new[] { "shard2", "shard3" });
+
+            var r2 = _allocationStrategy.Rebalance(allocations, ImmutableHashSet<string>.Empty.Add("shard2").Add("shard3")).Result;
```
alexvaluyskiy: You can use simplier construction here `ImmutableHashSet.Create<string>("shard2", "shard3")`

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
alexvaluyskiy: Good point

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
alexvaluyskiy: @Arkatufus we definitely have to do it, but not necessary in this pull request

alexvaluyskiy:
@zbynek001 Sorry for the delay, but now I've reviewed the pull request line-by-line. Looks pretty solid and ready to merge after addressing all remarks.
But I kindly ask @Horusiath to review the new version of `Shard` class and `AsyncWriteProxyEx`, because I'm not sure what to do with it.

Also this PR fixes
https://github.com/akkadotnet/akka.net/issues/1578
https://github.com/akkadotnet/akka.net/issues/1579
https://github.com/akkadotnet/akka.net/issues/1580
https://github.com/akkadotnet/akka.net/issues/1581

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
zbynek001: because `ClusterShardingGuardian` was created as user actor and not as system actor. Shall I change this?

Aaronontheweb:
Looks like @Horusiath still needs to sign off the changes @alexvaluyskiy called out

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
Horusiath: Lets keep this aligned with JVM version. This kind of dissonance between implementations may lead to confusion later (in docs and in solving issues).

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
```diff
@@ -15,46 +15,64 @@
...
+                    }
+
+                    akka.cluster.auto-down-unreachable-after = 0s
+                    akka.cluster.roles = [""backend""]
+                    akka.cluster.sharding {
+                        coordinator-failure-backoff = 3s
+                        shard-failure-backoff = 3s
+                    }
+                    akka.persistence.snapshot-store.plugin = ""akka.persistence.snapshot-store.inmem""
+                    akka.persistence.journal.plugin = ""akka.persistence.journal.memory-journal-shared""
```
Horusiath: Does this really work? MNTK tests run in separate processes: shared in-memory journal won't help here, as those nodes don't share memory.

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
```diff
@@ -180,35 +181,51 @@ private void StartSharding()
...
+
+        public void ClusterSharding_with_flaky_journal_network_should_setup_shared_journal()
         {
             // start the Persistence extension
             Persistence.Persistence.Instance.Apply(Sys);
             RunOn(() =>
             {
-                Sys.ActorOf(Props.Create<MemoryJournal>(), "store");
-            }, _controller);
+                Persistence.Persistence.Instance.Apply(Sys).JournalFor("akka.persistence.journal.MemoryJournal");
```
Horusiath: You may want to extract journal path into single constant field - it will make eventual changes easier.

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingMinMembersSpec.cs
Horusiath: Scala test clue works in similar way to hint string. It will prepend error message with `stats.ToString`

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
zbynek001: unfortunately it's quite important, not for the sharding itself, but for most of the tests :(

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
zbynek001: Yes, it works, it's not sharing any memory.
The journal actor is created at the beginning of each test on one node, usually the "controller":

```
RunOn(() =>
{
    Persistence.Persistence.Instance.Apply(Sys).JournalFor("akka.persistence.journal.MemoryJournal");
}, _config.Controller);

```

and other nodes are configured to use `MemoryJournalShared` pointing to the real journal actor:

```
RunOn(() =>
{
    Sys.ActorSelection(Node(_config.Controller) / "system" / "akka.persistence.journal.MemoryJournal").Tell(new Identify(null));
    var sharedStore = ExpectMsg<ActorIdentity>(TimeSpan.FromSeconds(10)).Subject;
    MemoryJournalShared.SetStore(sharedStore, Sys);
}, _config.First, _config.Second);

```
`MemoryJournalShared` is the `AsyncWriteProxyEx` actor, and it's basically forwarding persistent api to the real journal actor as messages

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingMinMembersSpec.cs
zbynek001: Still not sure how yo use it in this context, since hint is only supported for ExpectMsg or similar. 
Here i can only provide "because" parameter, is it the same?

PR #2768, opened 2017-06-16
[WIP] Performance tests for Akka, Remote, Cluster in .NET Core

heynickc:
Enables performance tests for .NET Core versions of Akka.Tests.Performance, Akka.Remote.Tests.Performance, and Akka.Tests.Cluster.Performance.  Limits the installation of NBench.Runner.exe to only runs of target: `NBench` since NBench.Runner v1.0.4 includes .NET Core standalone executables which include the .NET Core 1.1 runtime.


build.fsx
```diff
@@ -61,7 +61,8 @@ Target "Build" (fun _ ->
             (fun p -> 
                 { p with
                     Project = project
-                    Configuration = configuration })
+                    Configuration = configuration
+                    AdditionalArgs = ["--no-incremental"]}) // i.e. "Rebuild"
```
heynickc: Additional change - we should be calling "Rebuild" to be consistent with previous build script.

heynickc:
If this PR is merged, we can take out the BuildRelease step from "Windows Performance Tests" in CI since this only selectively builds the necessary projects to run PerfTests

build.fsx
```diff
@@ -131,16 +132,70 @@ Target "MultiNodeTests" (fun _ ->
...
+            info.FileName <- nbenchTestPathNet452
+            info.WorkingDirectory <- (Path.GetDirectoryName (FullName nbenchTestPathNet452))
+            info.Arguments <- args) (System.TimeSpan.FromMinutes 45.0) (* Reasonably long-running task. *)
+        if result <> 0 then failwithf "NBench.Runner failed. %s %s" nbenchTestPathNet452 args
+    
+    nbenchTestAssembliesNet452 |> Seq.iter (runNBenchNet452)
+
+    // .NET Core ---------------------------------------
+    // Publish netcoreapp1.1 performance test assemblies
+    let performanceTestProjects = !! "./**/*.Tests.Performance.csproj"
```
alexvaluyskiy: Each test run for performance tests take 45 minutes right now, we can't just double it. We have to run them separately

PR #2771, opened 2017-06-19
Split test runs for .NET Core and .NET 4.5.2

heynickc:
Initial changes to build.fsx to split test runs so it's easier to distinguish between .NET Core and .NET 4.5 test suites.


build.fsx
```diff
@@ -93,7 +93,36 @@ Target "RunTests" (fun _ ->
...
+                  -- "./**/Akka.Remote.TestKit.Tests.csproj"
+                  -- "./**/Akka.MultiNodeTestRunner.Shared.Tests.csproj"
+                  -- "./**/serializers/**/*Wire*.csproj"
+        // Linux/Mono
+        | _ -> !! "./**/core/**/*.Tests.csproj"
+                  ++ "./**/contrib/**/*.Tests.csproj"
+                  -- "./**/serializers/**/*Wire*.csproj"
+                  -- "./**/Akka.Remote.TestKit.Tests.csproj"
+                  -- "./**/Akka.MultiNodeTestRunner.Shared.Tests.csproj"
+                  -- "./**/Akka.API.Tests.csproj"
```
alexvaluyskiy: It cannot be run on NetCore on both platforms

build.fsx
```diff
@@ -93,7 +93,36 @@ Target "RunTests" (fun _ ->
...
+
+    let runSingleProject project =
+        DotNetCli.RunCommand
+            (fun p -> 
+                { p with 
+                    WorkingDir = (Directory.GetParent project).FullName
+                    TimeOut = TimeSpan.FromMinutes 30. })
+                (sprintf "xunit -parallel none -teamcity -noappdomain -f netcoreapp1.1 -c Release -xml %s_xunit.xml" (outputTests @@ fileNameWithoutExt project))
```
alexvaluyskiy: Should we change a name for `%s_xunit.xml`?
`noappdomain` is not necessary for netcore

build.fsx
heynickc: Can you explain?  We have been running .NET Core on both platforms, just 1 Platform + 2 Frameworks.  My goal is to run 1 Platform + 1 Framework for our intended supported Frameworks.

So instead of
```
Windows
  .NET Core
  .NET 4.5
Linux/Mono
  .NET 4.5
  .NET Core
```
We'll be doing
```
Windows
  .NET Core
Windows
  .NET 4.5
Linux/Mono
  .NET 4.5
Linux
  .NET Core
```

build.fsx
heynickc: Updated to `%s_xunit_netcore.xml`

heynickc:
I've incorporated the change from #2782 "Enable Remote.TestKit.Tests" into this one so that the change doesn't get lost in the shuffle.

heynickc:
Disregard passing windows unit tests, throwing exceptions but build is showing success.

heynickc:
Previous comment resolved

Aaronontheweb:
nice! see the netcore specs are in place - is this still WIP @heynickc ?

Aaronontheweb:
@heynickc mind if we postpone this until after the 1.2.2 release gets out the door tomorrow? Don't want any surprises :p

heynickc:
@Aaronontheweb yes totally

PR #2778, opened 2017-06-21
Added integration source stage for .NET events

Horusiath:
This PR brings 2 extra methods (versions of `FromEvent`, similar to another approach that can be seen in Rx.NET Observables). 

It's not ported from the JVM, however I think we should add it nonetheless: events are pretty essential part of the .NET ecosystem. They exist in many libraries, therefore to make it easier to integrate, we should add native support for them in Akka.Streams.

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/EventSourceSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/EventSourceSpec.cs
@@ -92,7 +92,7 @@ namespace Akka.Streams.Tests.Dsl
         }
 
         [Fact]
-        public void EventSource_must_should_fail_downstream_in_Fail_overflow_mode()
+        public void EventSource_must_fail_downstream_in_Fail_overflow_mode()
         {
             var s = this.CreateManualSubscriberProbe<int>();
             Source.FromEvent<int>(h => _event += h, h => _event -= h, 1, OverflowStrategy.Fail)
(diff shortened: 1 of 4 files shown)


src/core/Akka.Streams.Tests/Dsl/EventSourceSpec.cs
```diff
@@ -0,0 +1,129 @@
...
+            _event(this, 123);
+
+            s1.ExpectNext(123);
+            s2.ExpectNext(123);
+            sub1.Cancel();
+            sub2.Cancel();
+        }
+
+        [Fact]
+        public void EventSource_must_should_fail_downstream_in_Fail_overflow_mode()
```
Silv3rcircl3: Either must or should ;)

src/core/Akka.Streams/Implementation/Sources.cs
```diff
@@ -749,4 +750,150 @@ public UnfoldResourceSourceAsync(Func<Task<TSource>> create, Func<TSource, Task<
...
+            private readonly LinkedList<TEventArgs> _buffer;
+            private readonly TDelegate _handler;
+            private readonly Action<TEventArgs> _onOverflow;
+
+            public Logic(EventSourceStage<TDelegate, TEventArgs> stage) : base(stage.Shape)
+            {
+                _stage = stage;
+                _buffer = new LinkedList<TEventArgs>();
+                var bufferCapacity = stage._maxBuffer;
+                var onEvent = GetAsyncCallback<TEventArgs>(e =>
```
Silv3rcircl3: We could use (local) functions to avoid the allocs... but it doesn't really matter

Horusiath: This one will be converted into delegate anyway.

Silv3rcircl3: Right, my fault, didn't know that the function would be wrapped by the delegate, anyway lesson learned ;-)

Aaronontheweb:
@Horusiath have a test failure on Mono related to these changes - been able to reproduce it a few times.

PR #2773, opened 2017-06-20
fix node stuck at joining issues

Aaronontheweb:
close #2584 
allowed SurviveNetworkInstabilitySpec to run all the way

PR #2785, opened 2017-06-22
v1.2.1 initial release notes

Aaronontheweb:
Not the full notes, but just enough to ship a nightly for users affected by #2773


Aaronontheweb:
Just needed the version bump, really

PR #2787, opened 2017-06-22
Add RunTestsNetCore to dev branch build.fsx

heynickc:
Add `RunTestsNetCore  DoNothing` to dev branch build.fsx per #2786

PR #2788, opened 2017-06-22
Fixed node transition spec

### LINKED ISSUE #2014: Port Akka.Cluster MultiNodeSpec: TransitionSpec
https://github.com/akka/akka/blob/master/akka-cluster/src/multi-jvm/scala/akka/cluster/TransitionSpec.scala

Aaronontheweb:
closes #2014

### DIFF
--- a/src/core/Akka.Cluster.Tests.MultiNode/TransitionSpec.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/TransitionSpec.cs
@@ -61,7 +61,7 @@ namespace Akka.Cluster.Tests.MultiNode
 
         private MemberStatus MemberStatus(Address address)
         {
-            var status = ClusterView.Members.Concat(ClusterView.UnreachableMembers)
+            var status = ClusterView.Members.Union(ClusterView.UnreachableMembers)
                 .Where(m => m.Address == address)
                 .Select(m => m.Status)
                 .ToList();
@@ -158,7 +158,7 @@ namespace Akka.Cluster.Tests.MultiNode
             }, Roles.Where(r => r != fromRole && r != toRole).ToArray());
         }
 
-        //[MultiNodeFact(Skip = "Race conditions that are difficult to reproduce locally")]
+        [MultiNodeFact]
         public void TransitionSpecs()
         {
             A_Cluster_must_start_nodes_as_singleton_clusters();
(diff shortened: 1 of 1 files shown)


src/core/Akka.Cluster.Tests.MultiNode/TransitionSpec.cs
```diff
@@ -51,7 +52,9 @@ protected TransitionSpec(TransitionSpecConfig config) : base(config)
 
         private RoleName Leader(params RoleName[] roles)
         {
-            return roles.First();
+            // sorts the addresses and provides the address of the node with the lowest port number
+            // as that node will be the leader
+            return roles.Select(x => Tuple.Create(x, GetAddress(x).Port)).OrderBy(x => x.Item2).First().Item1;
```
Aaronontheweb: This was the fix - needed to be able to correctly pick the leader from the remaining nodes and not just assume that the first node in the list == the leader.

src/core/Akka.Cluster.Tests.MultiNode/TransitionSpec.cs
```diff
@@ -61,7 +64,7 @@ private RoleName[] NonLeader(params RoleName[] roles)
 
         private MemberStatus MemberStatus(Address address)
         {
-            var status = ClusterView.Members.Concat(ClusterView.UnreachableMembers)
+            var status = ClusterView.Members.Union(ClusterView.UnreachableMembers)
```
Aaronontheweb: Used the `Union` method here to filter out duplicates and be consistent with the JVM.

src/core/Akka.Cluster.Tests.MultiNode/TransitionSpec.cs
```diff
@@ -256,6 +259,7 @@ private void A_Cluster_must_perform_correct_transitions_when_third_joins_second(
             EnterBarrier("convergence-joining-3");
 
             var leader12 = Leader(_config.First, _config.Second);
+            Log.Debug("Leader: {0}", leader12);
```
Aaronontheweb: Used this to help expose the bug; won't log anything by default since we run with `DebugConfig = false`

Aaronontheweb:
FYI, the change that is missing is just this FAKE backport for once we split .NET Core tests on `v1.3` https://github.com/akkadotnet/akka.net/pull/2787

Doesn't affect the Akka.NET code itself.

PR #2791, opened 2017-06-24
log actors by absolute path when using Akka.Remote or Cluster

### LINKED ISSUE #2368: Akka.Cluster: change default system logger to always include the Cluster.SelfAddress in the path of the actor who logged message

Aaronontheweb:
closes #2368 

Staging this just in case I have to do more Akka.Cluster debugging over the weekend ;) cc @nvivo

### DIFF
--- a/src/core/Akka/Event/Logging.cs
+++ b/src/core/Akka/Event/Logging.cs
@@ -89,7 +89,10 @@ namespace Akka.Event
         /// <returns>The newly created logging adapter.</returns>
         public static ILoggingAdapter GetLogger(this IActorContext context, ILogMessageFormatter logMessageFormatter = null)
         {
-            var logSource = context.Self.ToString();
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
End result of this PR is seeing log messages like this:

```
[INFO][6/24/2017 1:47:48 AM][Thread 0028][akka.tcp://MultiNodeClusterSpec@localhost:58385/system/cluster/core/daemon] No seed-nodes configured, manual cluster join required
```

Aaronontheweb:
Should also help make it **much** easier to debug multi-node tests.

Aaronontheweb:
Ah, current IMPL leaves the UID off of the end... Need to fix that. Raised some issues in the test suite otherwise.

Issue #2584, opened 2017-04-06
Akka.Cluster: nodes stuck at joining


Aaronontheweb:
Seen an issue be reported on clusters that have at least 20ish nodes of new members being stuck in the Joining stage. However, those new members are able to join when:

* Another node joins (who then gets stuck)
* The leader gets restarted

It looks to me like the issue, overall, is that there's an edge case where the leader stops responding to some part of the join operation and gets stuck. Been looking into this for a little while but I have more data now and thought to report an issue for other users who've observed this happening.

nvivo:
I'm experiencing the same issue. A basic test so far showed that until 15 nodes all went up correctly. On the 16th it got stuck at JOINING. Once I started another node, the previous one went UP and the last one got stuck at JOINING.

nvivo:
@Aaronontheweb Is there any dev build with this fixed, maybe a 1.2.1 preview? I'd be glad to run it here to see if this is fixed.

oeaoaueaa:
As spoken previously, this bug is stopping us from _enabling_ Akka.net usage in production, so we will be very happy to test the fix in our environments with 25-50 nodes.

nvivo:
I'm waiting patiently for 1.3, but for now I decided to stop relying on akka cluster for communication between nodes for critical stuff, as waiting forever for nodes to join was causing us problems. I'm currently back to db for work syncronization until cluster is more reliable. I'm able to test pre-releases, but I guess 1.3 will be a big one so I'll need to check if there are other impacts before trying it.

Aaronontheweb:
@oeaoaueaa @nvivo my apologies guys, I've been working on this issue but it's been very difficult to reproduce. If you have any additional logs that you can send my way that include any information about gossip conflicts, that would be super helpful as that's where I've been testing and looking thus far.

nvivo:
@Aaronontheweb I'm not sure what to provide, I'm not seeing any logs related to Akka problems. Is there a way to enable logs for this without enabling heartbeat debug logs all the time?

Aaronontheweb:
@nvivo one bit of data that might be helpful; if you have [Petabridge.Cmd](https://cmd.petabridge.com/) or Akka.Cluster.Monitor, can you send me a readout of what the cluster looks like according to that when you're connected to another node that's currently in the cluster when the issue happens? 

[...]

PR #2794, opened 2017-06-24
Fix cluster node stuck joining

Aaronontheweb:
Adding some model-based tests and other tools to try to reproduce #2584 

Currently exploring:
* [x] Issues with membership comparison
* [x] Multiple nodes joining at same time to an existing cluster, one of whom will bump off the current leader

### DIFF
--- a/src/core/Akka.Cluster.Tests.MultiNode/LeaderLeavingSpec.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/LeaderLeavingSpec.cs
@@ -99,7 +99,7 @@ akka.cluster.publish-stats-interval = 25 s")
                         EnterBarrier("leader-left");
 
                         // verify that the LEADER is EXITING
-                        exitingLatch.Ready(TestLatch.DefaultTimeout);
+                        exitingLatch.Ready(TestKitSettings.DefaultTimeout);
 
                         EnterBarrier("leader-shutdown");
                         MarkNodeAsUnavailable(oldLeaderAddress);
--- a/src/core/Akka.Cluster.Tests/Akka.Cluster.Tests.csproj
+++ b/src/core/Akka.Cluster.Tests/Akka.Cluster.Tests.csproj
@@ -43,6 +43,10 @@
       <HintPath>..\..\packages\FsCheck.2.5.0\lib\net45\FsCheck.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="FsCheck.Xunit, Version=2.5.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\FsCheck.Xunit.2.5.0\lib\net45\FsCheck.Xunit.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="FSharp.Core, Version=4.3.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <HintPath>..\..\packages\FSharp.Core.3.1.2.5\lib\net40\FSharp.Core.dll</HintPath>
       <Private>True</Private>
--- a/src/core/Akka.Cluster.Tests/Akka.Cluster.Tests.csproj
+++ b/src/core/Akka.Cluster.Tests/Akka.Cluster.Tests.csproj
@@ -39,6 +39,14 @@
       <HintPath>..\..\packages\FluentAssertions.3.3.0\lib\net45\FluentAssertions.Core.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="FsCheck, Version=2.5.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\FsCheck.2.5.0\lib\net45\FsCheck.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="FSharp.Core, Version=4.3.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\FSharp.Core.3.1.2.5\lib\net40\FSharp.Core.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
     <Reference Include="Google.ProtocolBuffers, Version=2.4.1.555, Culture=neutral, PublicKeyToken=55f7125234beb589, processorArchitecture=MSIL">
       <HintPath>..\..\packages\Google.ProtocolBuffers.2.4.1.555\lib\net40\Google.ProtocolBuffers.dll</HintPath>
       <Private>True</Private>
@@ -94,12 +102,14 @@
     <Compile Include="ClusterDeployerSpec.cs" />
     <Compile Include="ClusterDomainEventPublisherSpec.cs" />
     <Compile Include="ClusterDomainEventSpec.cs" />
+    <Compile Include="ClusterGenerators.cs" />
     <Compile Include="ClusterHeartBeatSenderStateSpec.cs" />
     <Compile Include="ClusterSpec.cs" />
     <Compile Include="ClusterSpecBase.cs" />
     <Compile Include="DowningProviderSpec.cs" />
     <Compile Include="GossipSpec.cs" />
     <Compile Include="HeartbeatNodeRingSpec.cs" />
+    <Compile Include="MemberOrderingModelBasedTests.cs" />
     <Compile Include="MemberOrderingSpec.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Proto\ClusterMessageSerializerSpec.cs" />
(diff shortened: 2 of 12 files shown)


src/core/Akka.Cluster.Tests.MultiNode/LeaderLeavingSpec.cs
```diff
@@ -99,7 +99,7 @@ public void
                         EnterBarrier("leader-left");
 
                         // verify that the LEADER is EXITING
-                        exitingLatch.Ready(TestLatch.DefaultTimeout);
+                        exitingLatch.Ready(TestKitSettings.DefaultTimeout);
```
Aaronontheweb: Just wanted to get rid of an `OBSOLETE` warning while I was poking around in the test suite

src/core/Akka.Cluster.Tests/Akka.Cluster.Tests.csproj
```diff
@@ -39,6 +39,18 @@
       <HintPath>..\..\packages\FluentAssertions.3.3.0\lib\net45\FluentAssertions.Core.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="FsCheck, Version=2.5.0.0, Culture=neutral, processorArchitecture=MSIL">
```
Aaronontheweb: Installed FsCheck for some model-based tests

src/core/Akka.Cluster.Tests/ClusterGenerators.cs
```diff
@@ -0,0 +1,46 @@
...
+using FsCheck;
+
+namespace Akka.Cluster.Tests
+{
+    /// <summary>
+    /// INTERNAL API.
+    /// 
+    /// FsCheck data generators for Akka.Cluster types.
+    /// </summary>
+    public class ClusterGenerators
```
Aaronontheweb: Data generators for FsCheck specs

src/core/Akka.Cluster.Tests/MemberOrderingModelBasedTests.cs
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        [Property]
+        public Property MemberOrderingMustObeyModel()
+        {
+            return new MembershipMachine().ToProperty();
+        }
+
+        [Property(MaxTest = 1000)]
+        public Property DistinctMemberAddressesMustCompareDifferently(Address[] addresses)
```
Aaronontheweb: Sanity check for `Member.AddressOrdering`

src/core/Akka.Cluster.Tests/MemberOrderingModelBasedTests.cs
```diff
@@ -0,0 +1,250 @@
...
+                    return false.Classify(true, "all distinct").Classify(false, "empty set")
+                        .Label($"Expected {a1} and {a2} to compare as different, but did not. Compared as same");
+                a1 = a2; // next member
+            }
+
+            return true.Classify(true, "all distinct").Classify(false, "empty set"); ;
+        }
+    }
+
+    public class MembershipMachine : Machine<MembershipState, MembershipModel>
```
Aaronontheweb: Heavy-duty model-based test for squishing the issues with collections and `Member` comparisons I outlined here: https://github.com/akkadotnet/akka.net/issues/2584#issuecomment-310974051

src/core/Akka.Cluster.Tests/MemberOrderingModelBasedTests.cs
```diff
@@ -0,0 +1,250 @@
...
+            {
+                return $"{GetType()}(Members = [{string.Join(",", _members.Select(x => x.ToString()))}]";
+            }
+        }
+
+        #endregion
+
+        #region Operations
+
+        public class ChangeMemberStatus : Operation<MembershipState, MembershipModel>
```
Aaronontheweb: Verifies that if we change the status of a current member of the cluster to a new `MembershipStatus`, as long as that transition is valid the new instance of the `Member` will be merged correctly into the `ImmutableSortedSet<Member>` collection.

src/core/Akka.Cluster.Tests/MemberOrderingModelBasedTests.cs
```diff
@@ -0,0 +1,250 @@
...
+                NewStatus = newStatus;
+            }
+
+            public override bool Pre(MembershipModel model)
+            {
+                var containsMember = model.AllMembers.ContainsKey(TargetedMember);
+                if (!containsMember) return false; // short-circuit
+
+                var m = model.AllMembers[TargetedMember];
+                return Member.AllowedTransitions[m.Status].Contains(NewStatus);
```
Aaronontheweb: Verifies that the randomly generated transition is valid. I.e. a node can't go from `Joining --> Up`.

src/core/Akka.Cluster.Tests/MemberOrderingModelBasedTests.cs
```diff
@@ -0,0 +1,250 @@
...
+            }
+
+            public override Property Check(MembershipState actual, MembershipModel model)
+            {
+                var members = actual.Members;
+
+                var mergedMembers = Member.PickNextTransition(members, model.AllMembers.Values).ToImmutableSortedSet();
```
Aaronontheweb: This is the fix - simulates what the `ClusterDaemon` is going to on the leader node when it comes to merging in new changes.

src/core/Akka.Cluster.Tests/MemberOrderingModelBasedTests.cs
```diff
@@ -0,0 +1,250 @@
...
+
+                actual.Members = mergedMembers;
+                return
+                    mergedMembers.SetEquals(model.AllMembers.Values)
```
Aaronontheweb: Verifies that the sum of the parts equals the whole and that the modified member ended up with its new status, rather than its original status (which was the bug.)

src/core/Akka.Cluster.Tests/MemberOrderingSpec.cs
```diff
@@ -94,6 +94,42 @@ public void MemberOrdering_must_consistent_ordering_and_equals()
         }
 
         [Fact]
+        public void MemberOrdering_must_work_pick_member_with_legal_transition()
```
Aaronontheweb: Smaller sanity check spec to verify that two collections of `Member`s with both different addresses and statuses for the same address can be merged correctly.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -2204,10 +2204,9 @@ public void LeaderActionsOnConvergence()
                 _exitingConfirmed.Where(x => localGossip.GetMember(x).Status == MemberStatus.Exiting)
                 .ToImmutableHashSet();
 
+            var upNumber = 0;
```
Aaronontheweb: Possible fix for https://github.com/akkadotnet/akka.net/issues/1784 - turns out were were using `upNumber` incorrectly before and could result in multiple nodes having the same upNumber. So the aging mechanism wasn't implemented properly at all, due to what I believe is somewhat confusing syntax on the JVM: https://github.com/akka/akka/blob/master/akka-cluster/src/main/scala/akka/cluster/ClusterDaemon.scala#L1079

This has been fixed. upNumber will now increment correctly. Do you think this could have resolved the cluster singleton issue @Horusiath @zbynek001 ? Any easy way to check with the existing tests?

src/core/Akka.Cluster/Member.cs
```diff
@@ -264,11 +263,62 @@ public int Compare(Member x, Member y)
         }
 
         /// <summary>
+        /// Combines and sorts two lists of <see cref="Member"/> into a single list ordered by Member's valid transitions
+        /// </summary>
+        /// <param name="a">The first collection of members.</param>
+        /// <param name="b">The second collection of members.</param>
+        /// <returns>An immutable hash set containing the members with the next logical transition.</returns>
+        public static ImmutableSortedSet<Member> PickNextTransition(IEnumerable<Member> a, IEnumerable<Member> b)
```
Aaronontheweb: Implemented a pair-wise merging mechanism for two collections containing some of the same members, but with different statuses. Used in the MBT from earlier and the other `MemberOrderingSpec` I added.

src/core/Akka.Cluster/Member.cs
```diff
@@ -264,11 +263,62 @@ public int Compare(Member x, Member y)
...
+
+        /// <summary>
+        /// Compares two copies OF THE SAME MEMBER and returns whichever one is a valid transition of the other.
+        /// </summary>
+        /// <param name="a">First member instance.</param>
+        /// <param name="b">Second member instance.</param>
+        /// <returns>If a and b are different members, this method will return <c>null</c>. 
+        /// Otherwise, will return a or b depending on which one is a valid transition of the other.
+        /// If neither are a valid transition, we return <c>null</c></returns>
+        public static Member PickNextTransition(Member a, Member b)
```
Aaronontheweb: Comments on the code are somewhat self-explanatory, but we pick the right version of the same `Member` based on which one is a transition of the other.

PR #2799, opened 2017-06-26
akka.singleton: small fixes

zbynek001:


src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -630,8 +630,8 @@ private void GetNextOldestChange()
 
         private State<ClusterSingletonState, IClusterSingletonData> GoToOldest()
         {
-            Log.Info("Singleton manager [{0}] starting singleton actor", _cluster.SelfAddress);
             var singleton = Context.Watch(Context.ActorOf(_singletonProps, _settings.SingletonName));
+            Log.Info("Singleton manager starting singleton actor [{0}] ", singleton.Path);
```
Horusiath: Word "starting" is continuous not perfect ;) This means that it didn't finished yet. Please either change it to "started" or move upward.

zbynek001: It's like this in akka. Cannot move it upward, since the path is not yet known there. So will change it to started

PR #2802, opened 2017-06-26
no longer deploy local routees when clustered pool router is on wrong role type

### LINKED ISSUE #2783: When using clustered pool, if use-local-routees is on, it seems to ignore the use-role property
I'm attempting to implement a simple two node cluster (example: https://github.com/jrlost/AkkaNetRoutingTest). I'm seeing that when using a clustered pool, if use-local-routees is on, it seems to ignore the use-role property. Similar to what's outlined here: https://github.com/akka/akka/issues/15042

Aaronontheweb:
close #2783

### DIFF
--- a/src/core/Akka.Cluster.Tests.MultiNode/Routing/UseRoleIgnoredSpec.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/Routing/UseRoleIgnoredSpec.cs
@@ -131,7 +131,7 @@ namespace Akka.Cluster.Tests.MultiNode.Routing
             A_cluster_must_start_cluster();
             A_cluster_must_pool_local_off_roles_off();
             A_cluster_must_group_local_off_roles_off();
-            //A_cluster_must_pool_local_on_role_b();
+            A_cluster_must_pool_local_on_role_b();
             A_cluster_must_group_local_on_role_b();
             A_cluster_must_pool_local_on_role_a();
             A_cluster_must_group_local_on_role_a();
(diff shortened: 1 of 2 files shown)


src/core/Akka.Cluster.Tests.MultiNode/Routing/UseRoleIgnoredSpec.cs
```diff
@@ -131,7 +131,7 @@ public void UseRoleIgnoredSpecs()
             A_cluster_must_start_cluster();
             A_cluster_must_pool_local_off_roles_off();
             A_cluster_must_group_local_off_roles_off();
-            //A_cluster_must_pool_local_on_role_b();
+            A_cluster_must_pool_local_on_role_b();
```
Aaronontheweb: Allowed to test to run

src/core/Akka/Routing/RoutedActorCell.cs
```diff
@@ -160,14 +160,7 @@ public override void Start()
             if (RouterConfig is Pool)
             {
                 var pool = (Pool)RouterConfig;
-                // must not use pool.GetNrOfInstances(system) for old (not re-compiled) custom routers
-                // for binary backwards compatibility reasons
-                var deprecatedNrOfInstances = pool.NrOfInstances;
-
-                var nrOfRoutees = deprecatedNrOfInstances < 0
```
Aaronontheweb: any routers that are compiled without `NrOfInstances(ActorSystem)` can't run on 1.1 or later anyway.

heynickc:
@Aaronontheweb looks like a spec failed on Windows: ` Akka.Remote.Tests.ActorsLeakSpec.Remoting_must_not_leak_actors`.  I'm going to kick off the Windows Unit Tests again.

Aaronontheweb:
@heynickc yeah that one needs some hardening, but that spec failure is unrelated to this

PR #2806, opened 2017-06-27
Akka.Persistence: automatically start snapshot store correctly

### LINKED ISSUE #2770: akka.persistence: snapshot-store auto-start
Looks like snapshot stores which should auto-start are red from "akka.persistence.journal" config, instead of  "akka.persistence.snapshot-store"

Aaronontheweb:
close #2770


Aaronontheweb:
Added a reproduction spec as well

heynickc:
Looks good to me

PR #2814, opened 2017-06-29
Akka.NET v1.2.1 and v1.2.2 rebase without the pain

Aaronontheweb:
Let's see if this is easier...


Aaronontheweb:
Let the rebase drop

PR #2817, opened 2017-06-30
investigate DotNetty clean shutdown

Aaronontheweb:
Related to #2575 - verifying the all manner of Akka.Remote association disposals "do the right thing"


Aaronontheweb:
Alright, this is ready for merge. Going to include it as part of the resolution for 1.2.3.

PR #2818, opened 2017-06-30
Incremental testing and build script refactoring

heynickc:
Implements "incremental" test running for Unit, MNTR, and Perf tests.

Uses Git to determine the `merge-base` point where a branch was forked from v1.3 and then calculates altered files between fork point and PR branch's tip.  Then, it determines which test projects have a dependency on projects that contain the altered files calculated in the first step.  It does this recursively so to account for indirect dependencies.  If any build scripts are altered in the PR, then it will run the entire test suite.  It should not run tests if there are only changes to files outside of the `src` folder, such as docs.

It also includes some refactoring to our existing test project filters, such as only run `Akka.API.Tests.csproj` on Windows.  It puts that logic into the `buildIncremental.fsx` module too.

This PR will use a feature flag so that we can watch the potential effects of turning this on in CI.  We can include the argument `incremental=experimental` so that the build logs will show which tests would be run if we were to turn the feature on, but it will still run as usual.  If everything looks good, we can then turn on the feature with `incremental=true`.


Aaronontheweb:
@heynickc in order to see if this works... I guess we need to start actually using it, right? :p

heynickc:
@Aaronontheweb yes, we can turn the experimental feature on and monitor the effects on other PRs

PR #2833, opened 2017-07-05
Skip Akka.Remote.TestKit.Tests test that causes deadlock

heynickc:
Skip Akka.Remote.TestKit.Tests test that causes deadlocking: `RemoteConnection_should_send_and_decode_Done_message`


alexvaluyskiy:
@heynickc @Danthar Guys, let's create an issue for each skipped tests. We should track all of them

Danthar:
Or at least have a list in a single issue. Agreed.

PR #2836, opened 2017-07-06
Add NetCore support to MultiNodeTestRunner

Arkatufus:
Fix for #2828 
Things that are changed:
- Missing net452 extension methods and classes need to be added:
    - DateTime.ToShortTimeString() extension
    - Type.GetMethod() extension
    - CompileErrorCollection class
    - CompileError class
- Some CORECLR compiler directive needs to be added because net452 Assembly static functions are missing and needs to be changed with core replacement functions.
- MultiNodeSpec.GetCallerName need to be removed, it is impossible to get function caller type in NetCore.

NOTES: 
- Assembly.GetExecutingAssembly() is currently missing in core 1.1, but planned to be re-added in core 2.0. Need to change these back later.

CONCERN:
VisualizerRuntimeTemplate.cs is an auto-generated file, but needs tweaking to work under core. This might cause a build fail later.


src/core/Akka.MultiNodeTestRunner.Shared.Tests/MultiNodeTestRunnerDiscovery/DiscoverySpec.cs
```diff
@@ -46,7 +46,11 @@ public void Public_props_and_fields_are_considered()
 
         private static Dictionary<string, List<NodeTest>> DiscoverSpecs()
         {
+#if CORECLR
+            using (var controller = new XunitFrontController(AppDomainSupport.IfAvailable, new System.Uri(typeof(DiscoveryCases).GetTypeInfo().Assembly.CodeBase).LocalPath))
```
alexvaluyskiy: Could we use the same code for both platforms?

src/core/Akka.MultiNodeTestRunner.Shared.Tests/ParsingSpec.cs
```diff
@@ -78,7 +79,11 @@ public void MessageSink_should_parse_Runner_log_message_correctly()
         [Fact]
         public void MessageSink_should_parse_Node_SpecPass_message_correctly()
         {
+#if CORECLR
+            var specPass = new SpecPass(1, "super_role_1", GetType().GetTypeInfo().Assembly.GetName().Name);
```
alexvaluyskiy: We could use this code for `.NET4.5` too

src/core/Akka.MultiNodeTestRunner.Shared.Tests/TestRunShutdownSpec.cs
```diff
@@ -22,7 +23,7 @@ public class TestRunShutdownSpec : AkkaSpec
         [Fact]
         public void TestCoordinatorEnabledMessageSink_should_receive_TestRunTree_when_EndTestRun_is_received()
         {
-            var consoleMessageSink = Sys.ActorOf(Props.Create(() => new ConsoleMessageSinkActor(false)));
+            var consoleMessageSink = Sys.ActorOf(Props.Create(() => new ConsoleMessageSinkActor(true)));
```
alexvaluyskiy: What changed?

src/core/Akka.MultiNodeTestRunner.Shared/CompilerErrorCollection.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+#if CORECLR
+namespace System.CodeDom.Compiler
+{
+    public class CompilerErrorCollection : List<CompilerError>
```
alexvaluyskiy: Who is using this class?

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -378,7 +378,11 @@ public static Config BaseConfig
         private static string GetCallerName()
         {
             var @this = typeof(MultiNodeSpec).Name;
+#if CORECLR
+            var trace = new StackTrace(new Exception(), false);
```
alexvaluyskiy: Is this works find on `NetCore`? StackTrace is using to get the proper name for test suite, not for exceptions

src/core/Akka/Akka.csproj
```diff
@@ -53,7 +53,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard1.6' ">
-    <DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
+    <DefineConstants>$(DefineConstants);CORECLR;AKKAIO</DefineConstants>
```
alexvaluyskiy: Have you checked Akka.IO unit tests? Because right now all of them disabled (no AkkaIO constant in Akka.Tests.csproj)

alexvaluyskiy:
@Arkatufus DiscoverySpec is failing on Windows

src/core/Akka.MultiNodeTestRunner.Shared.Tests/MultiNodeTestRunnerDiscovery/DiscoverySpec.cs
Arkatufus: No, because Assembly.GetExecutingAssembly() isn't available in core 1.1, it's supposed to be back in 2.0, but we have to wait.

src/core/Akka.MultiNodeTestRunner.Shared.Tests/TestRunShutdownSpec.cs
Arkatufus: The spec said TestCoordinatorEnabled, but it is disabled in the original code (false), so I turn it on.

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
Arkatufus: I know, but you can't call StackTrace() without parameters in `NetCore` anymore.

src/core/Akka.MultiNodeTestRunner.Shared/CompilerErrorCollection.cs
Arkatufus: The only class that uses this is VisualizerRuntimeTemplate

src/core/Akka/Akka.csproj
Arkatufus: No, not yet. It would have to be the next thing to fix/port. MultiNodeTestRunner wouldn't work without NodeTestRunner, and NodeTestRunner requires Akka.IO

alexvaluyskiy: Why?

Arkatufus:
Yes, it is failing on the console runner, it passes in visual studio runner, I don't exactly know why, it fails when it instantiates XunitFrontController. I'm guessing that it is caused by AppDomain problem.
Note though that it fails on Net452 only, it passes NetCore tests.

src/core/Akka/Akka.csproj
Arkatufus: Both NodeTestRunner and MultiNodeTestRunner uses the Tcp class in each Program.cs

Arkatufus:
DiscoverySpec runs fine if `dotnet-xunit` is run using `-noappdomain`, but `Newtonsoft.Json` de-serialization failed with `-noappdomain`

Arkatufus:
On hold. Waiting for Akka.IO NetCore support.

Arkatufus:
Removed Akka.MultiNodeTestRunner and Akka.NodeTestRunner dependancy from Akka,MultiNodeTestRunner.Shared, this decouples it from Akka.IO so it should work now.

Arkatufus:
Very simple cross platform spec test for multi node test runner.
All it does is alternate the node test runner binary used to run each node role in each spec between Net and Net Core binaries

Arkatufus:
Note that cross platform test support is only available on Net Core build.

heynickc:
@Arkatufus want to put that fix from #2920 into this PR since it seems to be showing up in .NET Core unit tests?  CI won't test #2920 on .NET Core.

heynickc:
@Arkatufus Put this PR in on your branch: https://github.com/Arkatufus/akka.net/pull/1.  Based on local tests seems to have mitigated the BarrierSpec failures consistently once you fix the merge conflicts here.

Arkatufus:
Alright, I'll put those in

heynickc:
@alexvaluyskiy are you still requesting changes to this PR or do you approve?

PR #2844, opened 2017-07-10
Add cancellation to ActorSelection.ResolveOne

0x53A:
Add a new overload with a cancellation token.

Note: a new overload instead of adding it as an optional parameter is required to keep binary compatibility.


src/core/Akka/Actor/ActorSelection.cs
```diff
@@ -120,11 +120,31 @@ public void Tell(object message, IActorRef sender = null)
...
+        /// <param name="timeout">
+        /// The amount of time to wait while resolving the selection before terminating the operation and generating an error.
+        /// </param>
+        /// <param name="ct">
+        /// TBD
+        /// </param>
+        /// <exception cref="ActorNotFoundException">
+        /// This exception is thrown if no such actor exists or the identification didn't complete within the supplied <paramref name="timeout"/>.
+        /// </exception>
+        /// <returns>TBD</returns>
```
Danthar: Would you mind describing this TBD as well.

src/core/Akka/Actor/ActorSelection.cs
```diff
@@ -120,11 +120,31 @@ public void Tell(object message, IActorRef sender = null)
         /// <returns>TBD</returns>
```
Danthar: And while your at it. This one as well ?

Horusiath:
I'm not convinced to this solution: having to provide both timeout and cancellation token looks like terrible hack. I'd love to see more cancellation tokens in Akka.NET asynchronous API... but this must be reflected in the actual underlying implementation and not glued to existing methods.

Horusiath:
I'm not convinced to this solution: having to provide both timeout and cancellation token looks like terrible hack. I'd love to see more cancellation tokens in Akka.NET asynchronous API... but this must be reflected in the actual underlying implementation and not glued to existing methods.

0x53A:
The reason I sent the PR is that I was auditing our startup code for cancellation, and this was one of the few potentially long running (until the timeout) operations that did not yet have explicit cancellation support.

In my opinion, cancellation and timeout are orthogonal concepts.

* The timeout is just how long I want to wait for a network response.
* The cancellation is, most of the time, independent of the timeout and in response to some event from a logical parent, e.g. if the user clicks on the cancel button.

I'm also not a big fan of having to pass-through the tokens everywhere, but that is just part of the "design" of async-await.

With F# async, you would have the ambient cancellation token and so only needed to take care of the timeout.

Horusiath:
@0x53A I'd like to replace timeouts with cancellation tokens (as you've noticed, you can emulate first with the latter). Unfortunately, Akka.NET have a burden of backward compatibility and performance expectations. This kind of replacement had to be cautious rewrite of big parts of the library (and agreement of the whole team :P ).

heynickc:
@0x53A Just a heads up, the build failures showing on CI aren't due to the code introduced in your PR.  We're having issues with a process not shutting down properly on the build agent running your build due to issue #2848.  I'll be trying to kick off the build on another build agent to get this one to run properly.

heynickc:
@0x53A While your PR is sometimes affected by #2848, you also have an issue with building the project that seems unrelated, indicating that Akka.csproj can't be built.  It's showing as a timeout in CI, but if you look through the build logs you'll see messages like this:
```
...
[23:46:47][Build] Actor\ActorSelection.cs(141,61): error CS0246: The type or namespace name 'CancellationToken' could not be found (are you missing a using directive or an assembly reference?) [D:\work\d26c9d7f36545acd\src\core\Akka\Akka.csproj]
[23:46:47][Build] Actor\ActorSelection.cs(143,73): error CS0246: The type or namespace name 'CancellationToken' could not be found (are you missing a using directive or an assembly reference?) [D:\work\d26c9d7f36545acd\src\core\Akka\Akka.csproj]
...
[23:46:53][Build]     21 Warning(s)
[23:46:53][Build]     26 Error(s)
```

0x53A:
@heynickc Ooops. My work firewall had blocked the git client (until today, when I finally got IT to unblock it), so the PR was typed in the github ui.

Should now be fixed, local build was green and ``RunTests`` is currently running.

Thanks!

Aaronontheweb:
Thanks @0x53A !

PR #2845, opened 2017-07-10
Port JVM Akka's watchWith for custom termination messages

quartusdev:
I ported [PR #22794](https://github.com/akka/akka/pull/22794) from Akka that allows custom termination messages. The old deathwatch tests all pass and I added an additional one testing the custom message.
I think this is a really nice feature and worthwhile having, because this allows giving termination messages some more meaning and integrate them better in the message protocol of an actor.


quartusdev:
The Akka.API.Tests obviously don't pass anymore, because of the addition of the WatchWith method.
How are these API text files updated?

Aaronontheweb:
Restarted this build. Had an issue with the TC agent where it wasn't cleaning files for some reason.

Aaronontheweb:
@quartusdev thanks for your contribution!

PR #2851, opened 2017-07-11
Add pluggable offset strategies to EventsByTag query

alexvaluyskiy:
- Add support for Time-based UUID offset into EventsByTag query
- Make an offset as exclusive
- Added PersistenceQuery to API.Tests

### DIFF
--- a/src/core/Akka.Persistence.Query/Interfaces.cs
+++ b/src/core/Akka.Persistence.Query/Interfaces.cs
@@ -58,7 +58,7 @@ namespace Akka.Persistence.Query
         /// Corresponding query that is completed when it reaches the end of the currently
         /// stored events is provided by <see cref="ICurrentEventsByTagQuery.CurrentEventsByTag"/>.
         /// </summary>
-        Source<EventEnvelope, NotUsed> EventsByTag(string tag, long offset);
+        Source<EventEnvelope, NotUsed> EventsByTag(string tag, Offset offset);
     }
 
     /// <summary>
@@ -71,7 +71,7 @@ namespace Akka.Persistence.Query
         /// is completed immediately when it reaches the end of the "result set". Events that are
         /// stored after the query is completed are not included in the event stream.
         /// </summary>
-        Source<EventEnvelope, NotUsed> CurrentEventsByTag(string tag, long offset);
+        Source<EventEnvelope, NotUsed> CurrentEventsByTag(string tag, Offset offset);
     }
 
     /// <summary>
--- a/src/core/Akka.Persistence.Query/Offset.cs
+++ b/src/core/Akka.Persistence.Query/Offset.cs
@@ -11,11 +11,40 @@ namespace Akka.Persistence.Query
 {
     public abstract class Offset
     {
+        /// <summary>
+        /// Used when retrieving all events.
+        /// </summary>
         public static Offset NoOffset() => Query.NoOffset.Instance;
+
+        /// <summary>
+        /// Corresponds to an ordered sequence number for the events.Note that the corresponding
+        /// offset of each event is provided in the <see cref="EventEnvelope"/>,
+        /// which makes it possible to resume the stream at a later point from a given offset.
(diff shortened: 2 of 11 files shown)


alexvaluyskiy:
@Aaronontheweb @heynickc `ActorsLeakSpec.Remoting_must_not_leak_actors` test should be muted on all  four configuration, not only on `.NET +Win`

alexvaluyskiy:
@Horusiath The problem is that TimeBased UUID offset was designed for Cassandra. I've made these changes right know because we are planning to make Persistence Query as stable. And we can't make breaking changes after that

alexvaluyskiy:
@Horusiath Each provider could decide which type of offset should be implemented, but sequence-based is a default. 

I think a should rename this PR to `Add pluggable offsets strategies to Persistence Query`

src/core/Akka.Persistence.Query/Interfaces.cs
```diff
@@ -58,7 +58,7 @@ public interface IEventsByTagQuery : IReadJournal
         /// Corresponding query that is completed when it reaches the end of the currently
         /// stored events is provided by <see cref="ICurrentEventsByTagQuery.CurrentEventsByTag"/>.
         /// </summary>
-        Source<EventEnvelope, NotUsed> EventsByTag(string tag, long offset);
+        Source<EventEnvelope, NotUsed> EventsByTag(string tag, Offset offset);
```
Aaronontheweb: Looks good.

src/core/Akka.Persistence.Query/Offset.cs
```diff
@@ -0,0 +1,115 @@
...
+    /// The `offset` is exclusive, i.e.the event with the exact same sequence number will not be included
+    /// in the returned stream.This means that you can use the offset that is returned in <see cref="EventEnvelope"/>
+    /// as the `offset` parameter in a subsequent query.
+    /// </summary>
+    public sealed class TimeBasedGuid : Offset
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TimeBasedGuid"/> class.
+        /// </summary>
+        public TimeBasedGuid(Guid value)
```
Aaronontheweb: Correct me if I'm wrong.... But how is any of this a time-based `Guid`? Where do we generate the `Guid` based on the current time value? Don't we need a data structure that actually does this? Am I missing something?

alexvaluyskiy: JVM versions uses `UUID` type https://github.com/akka/akka/blob/master/akka-persistence-query/src/main/scala/akka/persistence/query/Offset.scala#L42

Each record in their Cassandra DB has a `timeuuid` column. And it's possible to use this field in offsets, like this 
https://github.com/akka/akka-persistence-cassandra/blob/master/core/src/test/scala/akka/persistence/cassandra/query/EventsByTagSpec.scala#L279

alexvaluyskiy: And I forget to add `IComparable` to all Offset implementations

alexvaluyskiy: I could temporarily remove this class. Right now it is useless

Aaronontheweb: @alexvaluyskiy as long as the IMPL can be overridden by a persistence plugin like the Cassandra one, then I'd go ahead and remove this. I think the `timeuuid` in the DataStax driver is a custom type.

Aaronontheweb: @alexvaluyskiy it is, but it can be safely and implicitly cast to a `Guid`: http://docs.datastax.com/en/drivers/csharp/3.3/html/T_Cassandra_TimeUuid.htm

alexvaluyskiy: @Aaronontheweb Ok, I will remove it

Aaronontheweb:
Looks good now!

Issue #2728, opened 2017-06-06
Fix racy unit tests part 2


Aaronontheweb:
Splitting these out from https://github.com/akkadotnet/akka.net/pull/2721, so we can get those fixes merged immediately and help eliminate some of the noise from our CI system.

* [x] Akka.Cluster.Tests.ClusterSpec.A_cluster_must_terminate_ActorSystem_via_leave_CoordinatedShutdown
* [x] Akka.Cluster.Tests.ClusterSpec.A_cluster_must_terminate_ActorSystem_via_Down_CoordinatedShutdown
* [x] Akka.Cluster.Tests.ClusterSpec.A_cluster_must_leave_via_CoordinatedShutdownRun 
* [ ] Akka.Remote.Tests.RemoteDeathWatchSpec.Must_quarantine_systems_after_unsuccessful_system_message_delivery_if_have_not_communicated_before 
* [ ] Akka.Remote.DotNettyTransportDnsResolutionSpec.HeliosTransport_Should_Resolve_DNS
* [ ] Akka.Persistence.Tests.LocalSnapshotStoreSpec.SnapshotStore_should_save_and_overwrite_snapshot_with_same_sequence_number
* [x]  Akka.Remote.Tests.RemotingSpec.Remoting_must_lookup_actors_across_node_boundaries
* [ ] Akka.Remote.Tests.RemotingTerminatorSpecs.RemotingTerminator_should_shutdown_properly_without_exception_logging_while_graceful_shutdown
* [x] DowningProviderSpec.Downing_provider_should_stop_the_cluster_if_the_downing_provider_throws_exception_in_props (**still having problems**)

Arkatufus:
- Akka.Cluster.Tests.ClusterSpec.A_cluster_must_terminate_ActorSystem_via_leave_CoordinatedShutdown => Done (#2843)

Arkatufus:
- Akka.Remote.DotNettyTransportDnsResolutionSpec.HeliosTransport_Should_Resolve_DNS => Couldn't reproduce

Aaronontheweb:
The rest of these appear to be time-out sensitive. Haven't seen much in the way of consistent failures from any of them in a while. Closing out this issue.

PR #2869, opened 2017-07-14
Fix RemotingSpec.Remoting_must_lookup_actors_across_node_boundaries

Arkatufus:
Related to #2728
Increase timeout duration to TestKitBase.ExpectMsg<T>()
Test failed because TestKitBase.ExpectMsg<T>() timed out on slow machines.


Aaronontheweb:
Yeah fair enough to expect this would have some issues on our under-powered Docker containers. Looks like a good solution in the short-run.

Arkatufus:
Still need looking into though, since it still timed out even with 3 second wait time. Something is really slow underneath.

Arkatufus:
This fix only addresses a small percentage of the problem, the ActorSystem still shuts down sporadically.

PR #2874, opened 2017-07-17
Increment build number with TeamCity number for -beta releases

heynickc:
For nightlies, an environment variable is set by TeamCity `BUILD_NUMBER` that should then be passed into the Assembly Version in `common.props` so that the assembly version is `1.3.0.*`.  Nightly builds pass `nugetprerelease=dev` to enable this.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -16,7 +16,7 @@ open Fake.Git
 let configuration = "Release"
 let solution = "./src/Akka.sln"
 let versionSuffix = 
-    match (getBuildParamOrDefault "nugetprerelease" "") with
+    match (getBuildParam "nugetprerelease") with
     | "dev" -> "beta"
     | _ -> ""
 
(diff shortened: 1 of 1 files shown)


alexvaluyskiy:
@heynickc @Aaronontheweb We should change `common.props` if we could set a version via argyuments? Like `dotnet restore|pack /p:VersionSuffix=alpha-2134'

heynickc:
@alexvaluyskiy when I tried to set it for `dotnet-pack` command, it automatically added a dash before the version suffix, so it would look like `1.0.3-beta-1234`.  wanted to keep it consistent with our nightlies with the build number `1.0.3.1234-beta` but if we're ok with changing the format I'll change it.

alexvaluyskiy:
@heynickc the new tooling support only semantic versioning.

heynickc:
@alexvaluyskiy ok, just let me know and I'll change it as long as there is consensus to do so.

PR #2876, opened 2017-07-17
Actorpath performance improvements

Aaronontheweb:
Also going to try to address this race condition that keeps coming up on Linux....

```
[22:14:14]	[docker] [ERROR][07/17/2017 22:14:15][Thread 0150][LocalActorRefProvider(akka://AkkaSpec)] guardian [akka://AkkaSpec/] failed, shutting down!
[22:14:14]	[docker] Cause: System.NullReferenceException: Object reference not set to an instance of an object.
[22:14:14]	[docker]    at Akka.Actor.ActorPath.FillElements(ActorPath actorPath) in /checkout/src/core/Akka/Actor/ActorPath.cs:line 207
[22:14:14]	[docker]    at Akka.Actor.ActorPath.<>c.<.cctor>b__56_0(ActorPath actorPath) in /checkout/src/core/Akka/Actor/ActorPath.cs:line 156
[22:14:14]	[docker]    at Akka.Util.FastLazy`2.get_Value() in /checkout/src/core/Akka/Util/FastLazy.cs:line 149
[22:14:14]	[docker]    at Akka.Actor.ActorPath.get_Elements() in /checkout/src/core/Akka/Actor/ActorPath.cs:line 255
[22:14:14]	[docker]    at Akka.Actor.ActorPath.Equals(ActorPath other) in /checkout/src/core/Akka/Actor/ActorPath.cs:line 306
[22:14:14]	[docker]    at Akka.Actor.ActorRefBase.Equals(IActorRef other) in /checkout/src/core/Akka/Actor/ActorRef.cs:line 331
[22:14:14]	[docker]    at Akka.Actor.ActorCell.AddWatcher(IActorRef watchee, IActorRef watcher) in /checkout/src/core/Akka/Actor/ActorCell.DeathWatch.cs:line 223
[22:14:14]	[docker]    at Akka.Actor.ActorCell.SysMsgInvokeAll(EarliestFirstSystemMessageList messages, Int32 currentState) in /checkout/src/core/Akka/Actor/ActorCell.DefaultMessages.cs:line 262
```

### DIFF
--- a/src/core/Akka/Actor/ActorPath.cs
+++ b/src/core/Akka/Actor/ActorPath.cs
@@ -618,64 +618,60 @@ namespace Akka.Actor
                 var acc = new Stack<string>();
                 while (true)
                 {
-                    switch (p)
-                    {
-                        case RootActorPath r:
-                            return acc.ToList();
-                        default:
-                            acc.Push(p.Name);
-                            p = p.Parent;
-                            continue;
-                    }
+                    if (p is RootActorPath)
+                        return acc.ToList();
+                    acc.Push(p.Name);
+                    p = p.Parent;
                 }
             }
-        }
-
-        /// <inheritdoc/>
-        public override ActorPath Root
-        {
-            get
-            {
-                var current = _parent;
-                while (current is ChildActorPath)
-                {
-                    current = ((ChildActorPath)current)._parent;
-                }
-                return current.Root;
-            }
-        }
-
-        /// <summary>
-        /// Creates a copy of the given ActorPath and applies a new Uid
-        /// </summary>
-        /// <param name="uid"> The uid. </param>
-        /// <returns> ActorPath. </returns>
-        public override ActorPath WithUid(long uid)
-        {
-            if (uid == Uid)
-                return this;
-            return new ChildActorPath(_parent, _name, uid);
-        }
-
-        /// <inheritdoc/>
-        public override int CompareTo(ActorPath other)
-        {
-            return InternalCompareTo(this, other);
-        }
-
-        private int InternalCompareTo(ActorPath left, ActorPath right)
-        {
-            if (ReferenceEquals(left, right)) return 0;
-            var leftRoot = left as RootActorPath;
-            if (leftRoot != null)
-                return leftRoot.CompareTo(right);
-            var rightRoot = right as RootActorPath;
-            if (rightRoot != null)
-                return -rightRoot.CompareTo(left);
-            var nameCompareResult = Compare(left.Name, right.Name, StringComparison.Ordinal);
-            if (nameCompareResult != 0)
-                return nameCompareResult;
-            return InternalCompareTo(left.Parent, right.Parent);
         }
     }
+
+    /// <inheritdoc/>
+    public override ActorPath Root
+    {
+        get
+        {
+            var current = _parent;
+            while (current is ChildActorPath)
+            {
+                current = ((ChildActorPath)current)._parent;
+            }
+            return current.Root;
+        }
+    }
+
+    /// <summary>
+    /// Creates a copy of the given ActorPath and applies a new Uid
+    /// </summary>
+    /// <param name="uid"> The uid. </param>
+    /// <returns> ActorPath. </returns>
+    public override ActorPath WithUid(long uid)
+    {
+        if (uid == Uid)
+            return this;
+        return new ChildActorPath(_parent, _name, uid);
+    }
+
+    /// <inheritdoc/>
+    public override int CompareTo(ActorPath other)
+    {
+        return InternalCompareTo(this, other);
+    }
+
+    private int InternalCompareTo(ActorPath left, ActorPath right)
+    {
+        if (ReferenceEquals(left, right)) return 0;
+        var leftRoot = left as RootActorPath;
+        if (leftRoot != null)
+            return leftRoot.CompareTo(right);
+        var rightRoot = right as RootActorPath;
+        if (rightRoot != null)
+            return -rightRoot.CompareTo(left);
+        var nameCompareResult = Compare(left.Name, right.Name, StringComparison.Ordinal);
+        if (nameCompareResult != 0)
+            return nameCompareResult;
+        return InternalCompareTo(left.Parent, right.Parent);
+    }
+}
 }
--- a/src/core/Akka/Actor/ActorPath.cs
+++ b/src/core/Akka/Actor/ActorPath.cs
@@ -8,9 +8,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Collections.ObjectModel;
 using System.Linq;
-using Akka.Actor.Dsl;
 using Akka.Util;
 using Newtonsoft.Json;
 using static System.String;
@@ -152,9 +150,6 @@ namespace Akka.Actor
             return true;
         }
 
-        private static readonly Func<ActorPath, IList<string>> FillElementsFunc =
-            actorPath => FillElements(actorPath);
-
         /// <summary>
         /// Initializes a new instance of the <see cref="ActorPath" /> class.
         /// </summary>
@@ -164,7 +159,6 @@ namespace Akka.Actor
         {
             Name = name;
             Address = address;
-            _elements = new FastLazy<ActorPath, IList<string>>(FillElementsFunc, this);
         }
 
         /// <summary>
@@ -178,7 +172,6 @@ namespace Akka.Actor
             Address = parentPath.Address;
             Uid = uid;
             Name = name;
-            _elements = new FastLazy<ActorPath, IList<string>>(FillElementsFunc, this);
         }
 
         /// <summary>
@@ -186,73 +179,16 @@ namespace Akka.Actor
         /// </summary>
         /// <value> The uid. </value>
         public long Uid { get; }
-        
-        private readonly FastLazy<ActorPath, IList<string>> _elements;
 
-        private static readonly string[] _emptyElements = { };
-        private static readonly string[] _systemElements = { "system" };
-        private static readonly string[] _userElements = { "user" };
-
-        /// <summary>
-        /// This method pursuits optimization goals mostly in terms of allocations.
-        /// We're computing elements chain only once and storing it in <see cref="_elements" />.
-        /// Computed chain meant to be reused not only by calls to <see cref="Elements" /> 
-        /// but also during chain computation of children actors.
-        /// </summary>
-        private static IList<string> FillElements(ActorPath actorPath)
-        {
-            // fast path next three `if`
-            if(actorPath is RootActorPath)
-                return _emptyElements;
-            if (actorPath.Parent is RootActorPath)
-            {
-                if (actorPath.Name.Equals("system", StringComparison.Ordinal))
-                    return _systemElements;
-                if (actorPath.Name.Equals("user", StringComparison.Ordinal))
-                    return _userElements;
-                return new [] {actorPath.Name};
-            }
-            // if our direct parent has computed chain we can skip list (for intermediate results) creation and resizing
-            if (actorPath.Parent._elements.IsValueCreated)
-            {
-                var parentElems = actorPath.Parent._elements.Value;
-                var myElems = new string[parentElems.Count + 1];
-                parentElems.CopyTo(myElems, 0);
-                myElems[myElems.Length - 1] = actorPath.Name;
-                return myElems;
-            }
-
-            // walking from `this` instance up to root actor
-            var current = actorPath;
-            var elements = new List<string>();
-            while (!(current is RootActorPath))
-            {
-                // there may be already computed elements chain for some of our parents, so reuse it!
-                if (current._elements.IsValueCreated)
-                {
-                    var parentElems = current._elements.Value;
-                    var myElems = new string[parentElems.Count + elements.Count];
-                    parentElems.CopyTo(myElems, 0);
-                    // parent's chain already in order, we need to reverse values collected so far
-                    for (int i = elements.Count - 1; i >= 0; i--)
-                    {
-                        myElems[parentElems.Count + (elements.Count - 1 - i)] = elements[i];
-                    }
-                    return myElems;
-                }
-                elements.Add(current.Name);
-                current = current.Parent;
-            }
-            // none of our parents have computed chain (no calls to Elements issued)
-            elements.Reverse();
-            return elements;
-        }
+        internal static readonly string[] EmptyElements = { };
+        internal static readonly string[] SystemElements = { "system" };
+        internal static readonly string[] UserElements = { "user" };
 
         /// <summary>
         /// Gets the elements.
         /// </summary>
         /// <value> The elements. </value>
-        public IReadOnlyList<string> Elements => new ReadOnlyCollection<string>(_elements.Value);
+        public abstract IReadOnlyList<string> Elements { get; }
 
         /// <summary>
         /// INTERNAL API.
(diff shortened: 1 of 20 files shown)


alexvaluyskiy:
@Aaronontheweb Is `GetHashCode` and `Equals` are slow? Do you have any real-world benchmarks?
My benchmarks show that `ActorPathToStringWithAddress` and `ActorPath.TryParse` are slow
https://github.com/akkadotnet/akka.net/issues/2566

alexvaluyskiy:
```
BenchmarkDotNet=v0.10.8, OS=Windows 10 Redstone 1 (10.0.14393)
Processor=Intel Core i5-3570 CPU 3.40GHz (Ivy Bridge), ProcessorCount=4
Frequency=3320338 Hz, Resolution=301.1742 ns, Timer=TSC
  [Host]     : Clr 4.0.30319.42000, 64bit RyuJIT-v4.7.2101.1
  Job-QFFVFK : Clr 4.0.30319.42000, 64bit RyuJIT-v4.7.2101.1

Jit=RyuJit  Platform=X64  Runtime=Clr  
Server=True  
```
 |                                   Method |         Mean |  Gen 0 | Allocated |
 |----------------------------------------- |-------------:|-------:|----------:|
 |            ActorPath_Parse_local_address | 1,506.281 ns | 0.0412 |    1272 B |
 |           ActorPath_Parse_remote_address | 1,713.906 ns | 0.0436 |    1344 B |
 |         ActorPath_TryParse_local_address | 1,520.568 ns | 0.0411 |    1272 B |
 |        ActorPath_TryParse_remote_address | 1,690.888 ns | 0.0439 |    1344 B |
 |  ActorPath_TryParseAddress_local_address |   663.254 ns | 0.0148 |     464 B |
 | ActorPath_TryParseAddress_remote_address |   908.522 ns | 0.0170 |     536 B |
 |                       ActorPath_Elements |     7.098 ns | 0.0010 |      32 B |
 |                ActorPath_Concat_Operator |    74.415 ns | 0.0093 |     288 B |
 |            ActorPath_ToStringWithAddress |   304.458 ns | 0.0086 |     272 B |
 |          ActorPath_ToSerializationFormat |   306.284 ns | 0.0086 |     272 B |
 |                         ActorPath_Equals |   135.046 ns | 0.0047 |     144 B |
 |                    ActorPath_GetHashCode |   337.701 ns | 0.0086 |     272 B |

Aaronontheweb:
@alexvaluyskiy I added benchmarks to this PR

Aaronontheweb:
# Akka.Tests.Performance.Actor.ActorPathEqualitySpec+ActorPath_GetHashCode_throughput
__Tests how quickly ActorPath.Equals can perform against two equal paths (worst case performance)__
_7/17/2017 11:43:25 PM_
### System Info
```ini
NBench=NBench, Version=1.0.1.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
```

### NBench Settings
```ini
RunMode=Throughput, TestMode=Measurement
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
|TotalCollections [Gen0] |     collections |           78.00 |           78.00 |           78.00 |            0.00 |
|TotalCollections [Gen1] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|TotalCollections [Gen2] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|[Counter] EqualsOp |      operations |    1,958,817.00 |    1,958,817.00 |    1,958,817.00 |            0.00 |

### Per-second Totals
|TotalCollections [Gen0] |     collections |          106.08 |          103.78 |          100.91 |            1.65 |
|TotalCollections [Gen1] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|TotalCollections [Gen2] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|[Counter] EqualsOp |      operations |    2,663,869.46 |    2,606,135.79 |    2,534,264.28 |       41,541.19 |

Aaronontheweb:
@alexvaluyskiy Equals is at about 5m op /s. 

TryParse is tricky, since most of the performance overhead comes from the built-in Url parsing... We've hacked around at trying to improve the perf of that in the past, but it's tough to trade away the reliability of that method given how many edge cases there are. Hence why most just try to cache the `ActorPath`s or `IActorRef`s instead.

The reason why `GetHashCode` is so slow is because it stringifies the entire `ActorPath` - it just calls `ToString().GetHashCode()`

Aaronontheweb:
And reason I started digging into it's performance is because....

This is the code for `ActorRefBase.GetHashCode()`
```
/// <inheritdoc/>
        {
            unchecked
            {
                var hash = 17;
                hash = hash * 23 + Path.Uid.GetHashCode();
                hash = hash * 23 + Path.GetHashCode();
                return hash;
            }
        }
```

We're going to incur that performance hit any time we have a collection of `IActorRef`s, such as the child container collection.

Aaronontheweb:
Ahh... All the way down the food chain, `GetHashCode` calls `ToStringWithAddress` each time.

Horusiath:
@Aaronontheweb my idea: make `ActorRefBase.GetHashCode()`:

1. Cache its hash code.
2. Ensure, that it returns a consistent hash code (as we already know string hashes are not consistent across machines).

This way we would gain on both performance and ability to have unique `int` identifier of an actor ref in the future (which could be possibly used in compression tables for actor refs deserialization optimization).

Aaronontheweb:
@Horusiath yeah, I think that's a good idea....

What do you think about:

```
Murmur3HashOf(actorPath) ^ UID
```

That will be stable so long as the actor doesn't restart (which will force the UID to change).

Aaronontheweb:
@rogeralsing the PID concept isn't something I'm going to try to implement on this PR. Needs a real design. 

We'd probably be better off combining the `ActorSystem` UID and hashing that with the ActorRef's UID + path instead of worrying about the address information anyway, since the Address information isn't available until after remoting has started... Or at the very least, computing the PID of each actorref inside the remoting system after it has started (so we can use the address.) and separating that concern entirely from the hashcode of the object.

rogeralsing:
>or at the very least, computing the PID of each actorref inside the remoting system after it has started 

The main issue here is Akka's ability to use multiple transports, I've never heard of anybody using multiple at the same time, but that design makes things problematic.
You would pretty much need to keep a PID cache per transport layer

Horusiath:
> I've never heard of anybody using multiple at the same time, but that design makes things problematic.

@rogeralsing  All memberlist lib users, including Consul ;)

rogeralsing:
I meant multiple Akka transports

Aaronontheweb:
# Akka.Tests.Performance.Actor.ActorPathEqualitySpec+ActorPath_GetHashCode_throughput
__Tests how quickly ActorPath.Equals can perform against two equal paths (worst case performance)__
_7/18/2017 5:08:53 PM_
### System Info
```ini
NBench=NBench, Version=1.0.1.0, Culture=neutral, PublicKeyToken=null
ProcessorCount=2
```

### NBench Settings
```ini
RunMode=Throughput, TestMode=Measurement
Concurrent=True
Tracing=True
```

## Data
-------------------

### Totals
|TotalCollections [Gen0] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|TotalCollections [Gen1] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|TotalCollections [Gen2] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|[Counter] EqualsOp |      operations |    9,234,765.00 |    9,234,765.00 |    9,234,765.00 |            0.00 |

### Per-second Totals
|TotalCollections [Gen0] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|TotalCollections [Gen1] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|TotalCollections [Gen2] |     collections |            0.00 |            0.00 |            0.00 |            0.00 |
|[Counter] EqualsOp |      operations |  105,326,509.08 |  104,169,896.74 |  101,776,645.55 |      878,254.06 |

Added a basic lazy cache to the hash code. Now it runs at 100m op/s but with an extra 4 bytes per `ActorPath`. That's going to speed up the slowest part of `ActorRefBase.GetHashCode`  by 50x. 

If we don't want to go with that, the other approach I used - just computing the value from the sum of its parts without stringifying it, gets a 4x increase in performance but no additional memory footprint beyond what we have now.

Where I'm hoping this has an improvement is on set and collection operations involving large lists of actors, such as heavily loaded `DistributedPubSub` systems. Should also see some nominal improvements around things like terminating really wide actor hierarchies.

There's a clear throughline in the improvement in performance from our `Akka.Tests.Performance.Actor.ActorSystemShutdown100000Spec+ActorSystemShutdownTime` spec:

## Spec results with no perf improvements
### Totals
|    Elapsed Time |              ms |        2,615.00 |        2,485.00 |        2,399.00 |           60.35 |

### Per-second Totals
|    Elapsed Time |              ms |        1,000.10 |          999.92 |          999.73 |            0.12 |

## Spec results with better `GetHashCode` impl
### Totals
|    Elapsed Time |              ms |        2,343.00 |        2,184.46 |        2,117.00 |           61.49 |

### Per-second Totals
|    Elapsed Time |              ms |        1,000.34 |        1,000.14 |          999.95 |            0.11 |

## Spec results with `GetHashCode` caching
|    Elapsed Time |              ms |        2,069.00 |        1,998.54 |        1,912.00 |           43.54 |

### Per-second Totals
|    Elapsed Time |              ms |        1,000.20 |        1,000.02 |          999.89 |            0.10 |

That's a 500ms (20%) improvement in the runtime by the end. The `GetHashCode` calls aren't going to show up on a performance profiler (tried that earlier today) since the instrumentation is going to hit on much more expensive stuff like `Props.Create` or the child container deep copying, but that doesn't mean their slowness doesn't have an impact. Had I more time, I'd write an ETW trace to collect the call site counts but I don't think this merits that.

Running the NBench suite for this a few more times just to validate the findings at the end (since we have some hardware variation to account for with the auto-spawned VMs...) - but either way we should pick one of the two new implementations and roll with that.

Aaronontheweb:
personally, I don't think 4bytes is a bad trade for having 0-allocations for this method - if we want to save on memory footprint with actors, there are much better culprits for doing that around the `ActorCell` and `Mailbox`. But to put it in real terms, that `int?` is going to cost 4mb for every million actors, and we do have users running around with 3-4 million per machine... Ehh.... I think we can live with that.

Just my 0.02

Aaronontheweb:
Made a major set of changes to `ActorPath`'s design around how it handles `Elements`, mostly to get around the issue we have on .NET Core on Linux, but also to improve overall memory footprint. Going to see the results and make some more changes - will probably end up having to add back a simple cache of some kind depending on the results.

Aaronontheweb:
FYI, going to also need to add some special overrides to `RootActorPath`'s string-ification methods in order to make its behavior consistent with its previous self again - haven't forgotten, just not worrying about it until I'm happier with the `Elements` design. 

Also, removed the `GetHashCode` cache - wasn't being called often enough to merit the extra field.

Aaronontheweb:
```
Cause: Akka.Remote.EndpointException: Failed to write message to the transport ---> System.AggregateException: One or more errors occurred. ---> System.ArgumentException: No association present
```

Welp, there go the Akka.Remote perf numbers for this round. Going to have to finish the changes before we can complete this.

src/core/Akka/Actor/ActorPath.cs
```diff
@@ -676,14 +607,31 @@ public ChildActorPath(ActorPath parentPath, string name, long uid)
...
-        /// </summary>
+        public override IReadOnlyList<string> Elements
+        {
+            get
+            {
+                ActorPath p = this;
+                var acc = new Stack<string>();
+                while (true)
+                {
+                    switch (p)
```
sean-gilliam: what if instead of the switch change it to

```csharp
if (p is RootActorPath _)
   return acc.ToList();
   
acc.Push(p.Name);
p = p.Parent;
```

Aaronontheweb: yep, that's worth trying out...

Aaronontheweb:
This ready to go. Also added `ActorPath` and `Address` parse caching, which 2x-ed the performance of Akka.Remote again.

rogeralsing:
What is the intention of the threadlocal's for the caches?

If I read this right, each thread will have it's own cache with entries that expire over time.
_but_ in the case of multiple transports active at the same time, the cache would only contain data for one of the transports.
(see my previous comment)

So if a serialize call for the second transport happens to be scheduled on a thread already containing data for the same actorref but for the first transport, the data would be incorrect.

My interpretation of this issue is that either: drop support for multiple transports, or avoid caches for the actorrefs.

or, drop the thread local caches and use a shared cache and include the current (threadstatic) transport information in the hash key.

Aaronontheweb:
> If I read this right, each thread will have it's own cache with entries that expire over time.
but in the case of multiple transports active at the same time, the cache would only contain data for one of the transports.

Not how it works. Threads in this case are bound to the remoting dispatcher, which are transport-agnostic. The caches are kept thread-local to avoid contention between cache instances when multiple endpoints are running in parallel. 

The addresses and strings being parsed are the *full remote path* from the actor sending the message, so one remote actor communicating with this node across two different transports would have two different entries in the cache.

rogeralsing:
👍 good good, I see that it's only on the parsing side, my bad.

Aaronontheweb:
@rogeralsing I haven't been brave enough to start work on the write-side yet 🐉

alexvaluyskiy:
## before
 |                                   Method |         Mean |  Gen 0 | Allocated |
 |----------------------------------------- |-------------:|-------:|----------:|
 |            ActorPath_Parse_local_address | 1,443.007 ns | 0.0378 |    1272 B |
 |           ActorPath_Parse_remote_address | 1,496.700 ns | 0.0417 |    1344 B |
 |         ActorPath_TryParse_local_address | 1,384.291 ns | 0.0394 |    1272 B |
 |        ActorPath_TryParse_remote_address | 1,499.314 ns | 0.0409 |    1344 B |
 |  ActorPath_TryParseAddress_local_address |   709.234 ns | 0.0141 |     464 B |
 | ActorPath_TryParseAddress_remote_address |   913.149 ns | 0.0160 |     536 B |
 |                       ActorPath_Elements |     6.293 ns | 0.0010 |      32 B |
 |                ActorPath_Concat_Operator |    80.371 ns | 0.0093 |     288 B |
 |            ActorPath_ToStringWithAddress |   298.788 ns | 0.0084 |     272 B |
 |          ActorPath_ToSerializationFormat |   294.988 ns | 0.0083 |     272 B |
 |                         ActorPath_Equals |   135.920 ns | 0.0044 |     144 B |
 |                    ActorPath_GetHashCode |   330.021 ns | 0.0081 |     272 B |
 
## after
 |                                   Method |        Mean |  Gen 0 | Allocated |
 |----------------------------------------- |------------:|-------:|----------:|
 |            ActorPath_Parse_local_address | 1,240.88 ns | 0.0334 |    1104 B |
 |           ActorPath_Parse_remote_address | 1,431.29 ns | 0.0340 |    1176 B |
 |         ActorPath_TryParse_local_address | 1,195.19 ns | 0.0338 |    1104 B |
 |        ActorPath_TryParse_remote_address | 1,426.48 ns | 0.0351 |    1176 B |
 |  ActorPath_TryParseAddress_local_address |   716.53 ns | 0.0137 |     464 B |
 | ActorPath_TryParseAddress_remote_address |   918.62 ns | 0.0158 |     536 B |
 |                       ActorPath_Elements |   145.93 ns | 0.0073 |     232 B |
 |                ActorPath_Concat_Operator |    49.57 ns | 0.0057 |     176 B |
 |            ActorPath_ToStringWithAddress |   447.63 ns | 0.0146 |     472 B |
 |          ActorPath_ToSerializationFormat |   451.84 ns | 0.0144 |     472 B |
 |                         ActorPath_Equals |   430.04 ns | 0.0170 |     544 B |
 |                    ActorPath_GetHashCode |   241.26 ns | 0.0079 |     272 B |

`ActorPath.Elements`, `Equals`, `ToStringWithAddress` and `ToSerializationFormat` became slower

PR #2880, opened 2017-07-19
[WIP] Indiscriminate non optimize -- TEST ONLY

Arkatufus:
Indiscriminately turns off all code optimization on all projects
For testing .Net Core JIT and Xunit stability, removing code optimization as a possible variable.


Aaronontheweb:
Worthy experiment, bug this is a performance killer. Appreciate your help @Arkatufus

PR #2884, opened 2017-07-19
GraphInterpreter: C# 7 style + const IsDebug

Horusiath:
Small micro optimizations around `GraphInterpreter` and `ActorGraphInterpreter`.


src/core/Akka.Streams/Implementation/Fusing/GraphInterpreter.cs
```diff
@@ -270,7 +270,7 @@ public sealed class Connection
         /// <summary>
         /// TBD
         /// </summary>
-        public static readonly bool IsDebug = false;
+        public const bool IsDebug = false;
```
Horusiath: Putting this as **const** is actually important. Otherwise compiler won't optimize away all branches in akka.streams interpreter hot path.

Silv3rcircl3:
API aaproval is missing

PR #2889, opened 2017-07-20
Incremental tests - fix platform-based filtering and logging

heynickc:
Fix formatting for log output in build logs for test discovery.  Filtering wasn't being applied per platform when incremental tests were enabled, fixed that.


Aaronontheweb:
@heynickc got incremental builds working on Linux?

heynickc:
@Aaronontheweb Yes, on the TeamCity side. I might have to add an additional filter for projects that aren't ported to .NET Core yet, but once this PR gets merged I'll know for sure.

PR #2890, opened 2017-07-20
Remove building of sample project in DocFx step

heynickc:
Remove the building of the example project from the `DocFx` build step.  It was intermingling with the DocFx generation making the build log confusing and possibly stopping the step too early.


alexvaluyskiy:
@heynickc that was the main idea. If the samples do not build that we should not publish our docs

heynickc:
@alexvaluyskiy fair enough

PR #2902, opened 2017-07-24
Akka.Remote LRU cache

Aaronontheweb:
Haven't started on the cache just yet; wrote the Akka.Remote benchmarks because our previously written benchmark had historically had some trouble revealing fine-grained changes. So I've rewritten our current one to follow a ping-pong strategy more closely.

Submitting this PR now so we can get some numbers early on, and then contrast them with the numbers following the addition of the cache.


src/core/Akka.Remote/Serialization/LruBoundedCache.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// </summary>
+    /// <remarks>
+    /// Fast hash based on the 128 bit Xorshift128+ PRNG. Mixes in character bits into the random generator state.
+    /// </remarks>
+    public static class FastHash
+    {
+        public static int OfString(string s)
+        {
+            var chars = s.ToCharArray();
+            var s0 = 391408L;
```
Danthar: Where do these random numbers come from ? Or are they chosen by random dice roll ?

rogeralsing: https://github.com/akka/akka/blob/master/akka-remote/src/main/scala/akka/remote/artery/LruBoundedCache.scala#L19

src/core/Akka.Remote/Serialization/LruBoundedCache.cs
```diff
@@ -0,0 +1,110 @@
...
+            var s1 = 601258L;
+            var i = 0;
+            unchecked
+            {
+                while (i < chars.Length)
```
rogeralsing: Cosmetic, but why a while loop here and a for loop in OfStringFast?

src/core/Akka.Remote/Serialization/LruBoundedCache.cs
Danthar: Ah. Well actually that still does not explain it :P Did **they** chose it by random dice roll ?

Horusiath: I agree, that this sort of magic numbers should be described before merging to avoid confusion.

Aaronontheweb: The magic numbers are just seed values, since the hash won't work properly if it's seeded at 0 initially: https://en.wikipedia.org/wiki/Xorshift 

In terms of why these exact seed numbers were chosen, no idea. Nothing special about them.

src/core/Akka.Remote/Serialization/LruBoundedCache.cs
Aaronontheweb: The for syntax was just an easier way to express the pointer arithmetic I needed to traverse the character array and I wrote that function second.

Aaronontheweb:
This is ready to go. Yielded about a 10% improvement, although a change I'm submitting in another PR will dwarf this one numbers-wise.

src/core/Akka.Remote/Serialization/LruBoundedCache.cs
```diff
@@ -107,13 +115,13 @@ public CacheStatistics(int entries, int maxProbeDistance, double averageProbeDis
...
     /// <typeparam name="TValue">The type of value used in the cache.</typeparam>
-    internal abstract class LruBoundedCache<TKey, TValue> where TValue:class
+    internal abstract class LruBoundedCache<TKey, TValue> where TValue : class
     {
         protected LruBoundedCache(int capacity, int evictAgeThreshold)
         {
             if (capacity <= 0)
                 throw new ArgumentOutOfRangeException(nameof(capacity), "Capacity must be larger than zero.");
-            if((capacity & (capacity - 1)) != 0) throw new ArgumentOutOfRangeException(nameof(capacity), "Capacity must be a power of two.");
+            if ((capacity & (capacity - 1)) != 0) throw new ArgumentOutOfRangeException(nameof(capacity), "Capacity must be a power of two.");
```
sean-gilliam: maybe simplify the check to
```csharp
(capacity & -capacity) != capacity
```

sean-gilliam: though the compiler will probably compile down to the same.

Aaronontheweb: Is that simpler? :p

sean-gilliam: two's complement :)

PR #2909, opened 2017-07-27
added RemotePingPong benchmark for .NET Core and NETFX

Aaronontheweb:
I wanted to get .NET Core figures and also test how scaling the number of concurrent actors affects remoting throughput, and the numbers tend to hold pretty steady once we have more than a few actors running. I'm hoping this gives me some better readings than what I was getting from NBench, although I have to admit that these benchmarks can still vary a lot too depending on how busy the system is in the background.

As a side note, I want to move our benchmarks onto dedicated hardware and off of Azure in the not too distant future.


Aaronontheweb:
@heynickc this should not trigger any tests to run per the incremental build rules, FYI

src/benchmark/RemotePingPong.NET46/Program.cs
```diff
@@ -0,0 +1,282 @@
...
+            var bindingConfig =
+                ConfigurationFactory.ParseString(@"akka.remote.dot-netty.tcp.hostname = """ + ipOrHostname + @"""")
+                    .WithFallback(ConfigurationFactory.ParseString(@"akka.remote.dot-netty.tcp.port = " + port));
+
+            return bindingConfig.WithFallback(baseConfig);
+        }
+
+        private static void Main(params string[] args)
+        {
+            Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.High;
```
Aaronontheweb: Same as we do with NBench - give this process priority with the OS scheduler while it runs. Doesn't guarantee much on a busy system but I find it helps with noise reduction.

src/benchmark/RemotePingPong.NET46/Program.cs
```diff
@@ -0,0 +1,282 @@
...
+            var testReady = (bool)rsp;
+            if (!testReady)
+            {
+                throw new Exception("Received report that 1 or more remote actor is unable to begin the test. Aborting run.");
+            }
+
+            var sw = Stopwatch.StartNew();
+            receivers.ForEach(c =>
+            {
+                for (var i = 0; i < 50; i++) // prime the pump so EndpointWriters can take advantage of their batching model
```
Aaronontheweb: Kick off a batch of messages locally at the start of each test - allows the `EndpointWriter` batching system to run the same way it would in a production environment.

src/benchmark/RemotePingPong.NET46/Program.cs
```diff
@@ -0,0 +1,282 @@
...
+                redCount = 0;
+            }
+            else
+            {
+                redCount++;
+                Console.ForegroundColor = ConsoleColor.Red;
+            }
+
+            Console.ForegroundColor = foregroundColor;
+            Console.WriteLine("{0,10},{1,8},{2,10},{3,11}", numberOfClients, totalMessagesReceived, throughput, sw.Elapsed.TotalMilliseconds.ToString("F2", CultureInfo.InvariantCulture));
```
Aaronontheweb: I gave up formatting this stuff correctly, but could fix it in the future.

src/benchmark/RemotePingPong.NET46/RemotePingPong.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+<PropertyGroup>
+  <Description>Akka.Remote x-plat performance benchmark</Description>
+  <Copyright>Copyright (c) Akka.NET Team</Copyright>
+  <AssemblyTitle>RemotePingPong</AssemblyTitle>
+  <AssemblyName>RemotePingPong</AssemblyName>
+  <VersionPrefix>1.3.0</VersionPrefix>
+  <Authors>Petabridge</Authors>
```
alexvaluyskiy: Should be Akka. NET team

src/benchmark/RemotePingPong.NET46/RemotePingPong.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+<PropertyGroup>
+  <Description>Akka.Remote x-plat performance benchmark</Description>
```
alexvaluyskiy: You have to enable ServerGc for NetCore too in csproj file

src/benchmark/RemotePingPong.NET46/RemotePingPong.csproj
Aaronontheweb: Whoops, sorry about that - copied config from the `.csproj` for the NBench runner.

src/benchmark/RemotePingPong.NET46/RemotePingPong.csproj
Aaronontheweb: Ok cool, I was wondering how to do that for .NET Core

Aaronontheweb:
@alexvaluyskiy addressed your comments

PR #2916, opened 2017-07-28
Update NuGet.config to use live nightlies

heynickc:
Updates NuGet.config to point to nightlies that are live and pushed daily to MyGet feed.


Aaronontheweb:
@heynickc this can't run at all on Linux. Bombs out on build.

heynickc:
@Aaronontheweb yea it's because the new nightly feed labels Hyperion as version `0.9.5-beta` instead of `0.9.5` and the one on NuGet doesn't have .NET Core support yet.  I'll point Akka.DistributedData to Hyperion `0.9.5-beta` instead if that's ok.

Aaronontheweb:
should be fine

PR #2919, opened 2017-08-01
Add cross-platform spec test to MNTR

Arkatufus:


Arkatufus:
Very simple cross platform spec test for multi node test runner.
All it does is alternate the node test runner binary used to run each node role in each spec between Net and Net Core binaries

Arkatufus:
Note that cross platform test support is only available on Net Core build.

Aaronontheweb:
@Arkatufus just so I understand correctly, all this PR provides right now are some switches for enabling X-plat multi-node tests. Not any actual x-plat multi-node tests themselves, correct?

Aaronontheweb:
Derp, nevermind:

> Very simple cross platform spec test for multi node test runner.
All it does is alternate the node test runner binary used to run each node role in each spec between Net and Net Core binaries

PR #2920, opened 2017-08-01
[WIP] fix BarrierSpec fail 

Arkatufus:


Arkatufus:
Better fix from @heynickc  already merged into #2836

PR #2922, opened 2017-08-02
Reduced allocations for serializers with IncludeManifest

alexvaluyskiy:
Slightly reduced allocations for serializers with Includemanifest=true

|                                         Method |      Mean |     Error |    StdDev |  Gen 0 | Allocated |
|----------------------------------------------- |----------:|----------:|----------:|-------:|----------:|
|          MessageSerializer_Serialize_Primitive | 47.544 us | 0.3309 us | 0.2583 us | 0.1261 |    6577 B |
| MessageSerializer_SerializeOptimized_Primitive | 48.715 us | 0.3919 us | 0.3474 us | 0.1221 |    6289 B |
|            MessageSerializer_Serialize_Complex | 63.856 us | 1.2648 us | 1.3533 us | 0.2360 |   10601 B |
|   MessageSerializer_SerializeOptimized_Complex | 64.427 us | 1.2760 us | 1.2532 us | 0.2360 |   10601 B |

### DIFF
--- a/src/core/Akka.Remote/MessageSerializer.cs
+++ b/src/core/Akka.Remote/MessageSerializer.cs
@@ -8,6 +8,7 @@
 using System;
 using Akka.Actor;
 using Akka.Serialization;
+using Akka.Util;
 using Google.Protobuf;
 using SerializedMessage = Akka.Remote.Serialization.Proto.Msg.Payload;
 
@@ -41,7 +42,8 @@ namespace Akka.Remote
         /// <returns>SerializedMessage.</returns>
         public static SerializedMessage Serialize(ActorSystem system, Address address, object message)
         {
-            Serializer serializer = system.Serialization.FindSerializerFor(message);
+            var objectType = message.GetType();
+            Serializer serializer = system.Serialization.FindSerializerForType(objectType);
 
             var serializedMsg = new SerializedMessage
             {
@@ -49,8 +51,7 @@ namespace Akka.Remote
                 SerializerId = serializer.Identifier
             };
 
-            var serializer2 = serializer as SerializerWithStringManifest;
-            if (serializer2 != null)
+            if (serializer is SerializerWithStringManifest serializer2)
             {
                 var manifest = serializer2.Manifest(message);
                 if (!string.IsNullOrEmpty(manifest))
@@ -61,7 +62,7 @@ namespace Akka.Remote
             else
             {
                 if (serializer.IncludeManifest)
-                    serializedMsg.MessageManifest = ByteString.CopyFromUtf8(message.GetType().AssemblyQualifiedName);
+                    serializedMsg.MessageManifest = ByteString.CopyFromUtf8(objectType.TypeQualifiedName());
             }
 
             return serializedMsg;


src/core/Akka.Remote/MessageSerializer.cs
```diff
@@ -41,16 +42,16 @@ public static object Deserialize(ActorSystem system, SerializedMessage messagePr
         /// <returns>SerializedMessage.</returns>
         public static SerializedMessage Serialize(ActorSystem system, Address address, object message)
         {
-            Serializer serializer = system.Serialization.FindSerializerFor(message);
+            var objectType = message.GetType();
+            Serializer serializer = system.Serialization.FindSerializerForType(objectType);
 
             var serializedMsg = new SerializedMessage
             {
                 Message = ByteString.CopyFrom(serializer.ToBinaryWithAddress(address, message)),
```
alexvaluyskiy: Why we are using `ToBinaryWithAddress` here? Akka JVM uses simple `ToBinary`

alexvaluyskiy: If I change it to `ToBinary` I could save another 128 bytes of allocations per call

|                                  Method |      Mean |     Error |    StdDev |  Gen 0 | Allocated |
|---------------------------------------- |----------:|----------:|----------:|-------:|----------:|
|   MessageSerializer_Serialize_Primitive | 46.153 us | 0.2032 us | 0.1901 us | 0.1343 |    6161 B |
|     MessageSerializer_Serialize_Complex | 61.302 us | 0.3956 us | 0.3089 us | 0.2441 |   10473 B |

alexvaluyskiy:
@Aaronontheweb In Akka 1.2.3 we allocated 6577 bytes per each message. After you commit with Lazy caching for Serialirer, we have reduced allocations by 2636 bytes per call. If we merge this PR, and will change `ToBinaryWithAddress` to `ToBinary` we will allocate only 192 bytes per call. 32 time better than before :)

src/core/Akka.Remote/MessageSerializer.cs
alexvaluyskiy: Also `ByteString.CopyFrom` make a copy of an array. But unfortunately `Google.Protobuf` does not have `CopyFromUnsafe`. Actually it has this feature, but it is internal https://github.com/google/protobuf/blob/master/csharp/src/Google.Protobuf/ByteString.cs#L66

src/core/Akka.Remote/MessageSerializer.cs
```diff
@@ -41,16 +42,16 @@ public static object Deserialize(ActorSystem system, SerializedMessage messagePr
         /// <returns>SerializedMessage.</returns>
         public static SerializedMessage Serialize(ActorSystem system, Address address, object message)
         {
-            Serializer serializer = system.Serialization.FindSerializerFor(message);
```
alexvaluyskiy: The `FindSerializerFor` has null check, but `FindSerializerForType` don't. I don't know how it will behave with nulls

Aaronontheweb: IMHO, throwing an exception if we try to serialize a null message is fine by me. We can't send it anyway :p

src/core/Akka.Remote/MessageSerializer.cs
Aaronontheweb: Related issue: https://github.com/google/protobuf/issues/3431

PR #2933, opened 2017-08-04
Persisence serializer setting

Danthar:
Work in progress for the Persistence serializer proposal. See https://github.com/akkadotnet/akka.net/issues/2743#issuecomment-314790660


src/contrib/persistence/Akka.Persistence.Sql.Common/Settings.cs
```diff
@@ -130,6 +135,7 @@ public SnapshotStoreSettings(Config config)
             SchemaName = config.GetString("schema-name");
             TableName = config.GetString("table-name");
             AutoInitialize = config.GetBoolean("auto-initialize");
+            DefaultSerializer = config.GetString("serializer");
```
Danthar: Even if this setting is not specified by the plugin explicitly, it should still be available. See changes in persistence.conf

src/core/Akka.Persistence/persistence.conf
```diff
@@ -115,6 +115,9 @@ akka.persistence {
         # Dispatcher for message replay.
         replay-dispatcher = "akka.persistence.dispatchers.default-replay-dispatcher"
 
+        # Default serializer used as manifest serializer when applicable and payload serializer when no specific binding overrides are specified
+        serializer = "json"
+
```
Danthar: Here i defined the default serializer setting in the `journal-plugin-fallback`. Which should be used by default as the fallback plugin for every journal plugin, by the persistence subsystem. But I still need to verify this is actually true

src/core/Akka.Persistence/persistence.conf
```diff
@@ -171,6 +174,9 @@ akka.persistence {
         # Dispatcher for the plugin actor.
         plugin-dispatcher = "akka.persistence.dispatchers.default-plugin-dispatcher"
 
+        # Default serializer used as manifest serializer when applicable and payload serializer when no specific binding overrides are specified
+        serializer = "json"
+
```
Danthar: Same goes for this one, only then defined in the ` snapshot-store-plugin-fallback`

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -119,6 +119,26 @@ public Serialization(ExtendedActorSystem system)
             }
         }
 
+        private Serializer GetSerializerByName(string name)
+        {
```
Danthar: So this is the helper method for translating the serializer name to an actual instance. 
Downside, having to translate it here. Upside, being able to leverage caching once this is done.

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -119,6 +119,26 @@ public Serialization(ExtendedActorSystem system)
...
+            if (name == null)
+                return null;
+
+            var serializersConfig = System.Settings.Config.GetConfig("akka.actor.serializers").AsEnumerable().ToList();
+            foreach (var kvp in serializersConfig)
```
Danthar: Because this config is not saved in the Serialization subsystem, I have to retrieve this value again. Leaning on the assumption of hocon config caching here.

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -119,6 +119,26 @@ public Serialization(ExtendedActorSystem system)
...
+            {
+                if (kvp.Key.Equals(name))
+                {
+                    var serializerTypeName = kvp.Value.GetString();
+                    var serializerType = Type.GetType(serializerTypeName);
+
+                    var serializerId = SerializerIdentifierHelper.GetSerializerIdentifierFromConfig(serializerType, (ExtendedActorSystem)System);
+                    return GetSerializerById(serializerId);
```
Danthar: Because I want to leverage existing infrastructure as much as possible, without having to resort to creating a new map from name to id (which is possibly easier, but costs more memory, although small). I resolve the name to the serializer ID and use the existing map to return the instance based on that ID.

Im still not completely set on this implementation though. I might just go with a seperate dict that maps name -> id which is initialised in the constructor like the other mappings are.

Impact of this method should be small though, as its only called once per type, after which caching takes over.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -507,7 +515,7 @@ public RequestChunk(int chunkId, IJournalRequest[] requests)
         private readonly HashSet<IActorRef> _allIdsSubscribers;
         private readonly HashSet<string> _allPersistenceIds;
 
-        private readonly Func<Type, Serializer> _getSerializer;
+        private readonly Func<Type,string, Serializer> _getSerializer;
```
Danthar: @Horusiath why did you used this pattern? Why not simply call `Context.System.Serialization.FindSerializerFor` When where you need it? Why store a reference in a func ?

Danthar:
Needs ApiApproval. But im going to hold off with that untill i get some feedback.

Danthar:
Ready for review. Should probably update the DData integration as well. Which should not be hard to do. But want to get some feedback first.

heynickc:
@Danthar can you add a quick spec or two for the key parts of this?

src/core/Akka.Persistence.Tests/Serialization/SerializerSpec.cs
```diff
@@ -209,5 +209,14 @@ public void MessageSerializer_should_serialize_manifest_provided_by_EventAdapter
 
             back.Manifest.ShouldBe(p1.Manifest);
         }
+
+        [Fact]
+        public void Serialization_respects_default_serializer_parameter()
```
Aaronontheweb: Add a spec to demonstrate what happens if the default serializer can't be found

src/core/Akka/Serialization/Serialization.cs
Aaronontheweb: > Impact of this method should be small though, as its only called once per type, after which caching takes over.

IMHO, should be fine.

src/core/Akka.Persistence.Tests/Serialization/MessageSerializerRemotingSpec.cs
```diff
@@ -23,12 +23,12 @@ internal class Configs
 akka.actor {
   serializers {
     my-payload = ""Akka.Persistence.Tests.Serialization.MyPayloadSerializer, Akka.Persistence.Tests""
-    my-payload2 = ""Akka.Persistence.Tests.Serialization.MyPayload2Serializer, Akka.Persistence.Tests""
     old-payload = ""Akka.Persistence.Tests.Serialization.OldPayloadSerializer, Akka.Persistence.Tests""
+    testserializer = ""Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion""
```
Danthar: In order to fix these tests. I switched system default serializer for these tests to Hyperion.
But im not sure if thats a good idea.

src/core/Akka.Persistence.Tests/Serialization/SerializationSpec.cs
```diff
@@ -40,7 +40,7 @@ public void Serialization_respects_default_serializer_parameter()
         {
             var message = new TestMessage("this is my test message");
             var serializer = _serialization.FindSerializerFor(message, "json");
-            Assert.True(serializer.Identifier == 1); //by default configuration the serializer id for json == newtonsoft == 1
+            Assert.Equal(1, serializer.Identifier); //by default configuration the serializer id for json == newtonsoft == 1
```
Danthar: Upside is that this test is more explicit.

src/core/Akka.Persistence.Tests/Serialization/SerializationSpec.cs
```diff
@@ -50,7 +50,7 @@ public void Serialization_falls_back_to_system_default_if_unknown_serializer_par
             var serializer = _serialization.FindSerializerFor(message, "unicorn");
             //since unicorn is an unknown serializer, the system default will be used
             //which incedentally is JSON at the moment
-            Assert.True(serializer.Identifier == 1); //by default configuration the serializer id for json == newtonsoft == 1
+            Assert.Equal(-5, serializer.Identifier); //system serializer is currently configured to be hyperion which is -5
```
Danthar: Same goes for this one

src/core/Akka.Persistence.Tests/Serialization/MessageSerializerRemotingSpec.cs
Danthar: In fact. After talking this over with @alexvaluyskiy we are not sure why we have this Spec at all. It was disabled before. But it does not test any real-world scenario as far as we know. 
What are the real-world scenario's for sending the `Persistent` or `AtomicWrite` types across the wire.

Aaronontheweb: None that I can think of

src/core/Akka.Persistence.Tests/Serialization/SerializationSpec.cs
```diff
@@ -0,0 +1,56 @@
+﻿// <copyright file="SerializationSpec.cs" company="Copaco B.V.">
+//        Copyright (c) 2015 - 2017 All Right Reserved
+//        Author: Arjen Smits
```
Aaronontheweb: Copyright header needs to be changed

Aaronontheweb: @Danthar indicated in Gitter chat that this was a mistake with his ReSharper template, so no worries.

PR #2944, opened 2017-08-09
ClusterSingletonManager remove ClusterEvent.MemberExited events

zbynek001:
this removes some unhandled events warnings


Danthar:
Im not sure if the bug is that we are subscribing to the `MemberExited` event or if the bug is, that we are not handling this event. /cc @Aaronontheweb  ?

zbynek001:
at least scala subscribes only to `MemberRemoved`, and not to `MemberExited`. And there is nothing handling `MemberExited` in ClusterSingletonManager

alexvaluyskiy:
@zbynek001 looks good

Aaronontheweb:
@Danthar `MemberRemoved` gets fired when the node is officially gone from the cluster.

PR #2945, opened 2017-08-09
Issue 2649

### LINKED ISSUE #2649: Actor implementing IWithUnboundedStash could not use custom priority mailbox
After upgrade to 1.2 from 1.0
i noticed if actor is using stash, whenever i try to use custom mailbox type. I got exception:

`ArgumentException: produced message queue type of [Akka.Dispatch.MessageQueues.UnboundedPriorityMessageQueue] does not fulfill requirement for actor class [xxx.ReceiverActor].Must be a subclass of [Akka.Dispatch.MessageQueues.IMessageQueue]
`

However if i use either of them, it compiles and works as expected:
i.e.:
Actor uses priority mailbox without implementing IWithUnboundedStash. 
Or. Actors implementing IWithUnboundedStash without using priority mailbox.

Danthar:
close #2649 

Fix for stashing support in priority mailbox.

### DIFF
--- a/src/core/Akka.Tests/Dispatch/MailboxesSpec.cs
+++ b/src/core/Akka.Tests/Dispatch/MailboxesSpec.cs
@@ -58,6 +58,67 @@ namespace Akka.Tests.Dispatch
         }
     }
 
+    public class StashingActor : ReceiveActor, IWithUnboundedStash
+    {
+        private readonly TestKitBase _testkit;
+        private readonly bool _echoBackToSenderAsWell;
+        public IStash Stash { get; set; }
+
+        /// <summary>
+        /// TBD
+        /// </summary>
+        /// <param name="testkit">TBD</param>
+        /// <param name="echoBackToSenderAsWell">TBD</param>
+        public StashingActor(TestKitBase testkit, bool echoBackToSenderAsWell = true)
+        {
+            _testkit = testkit;
+            _echoBackToSenderAsWell = echoBackToSenderAsWell;
+            StashAll();
+        }
+
+        private void StashAll()
+        {
+            Receive<Start>(_ =>
+            {
+                Become(Process);
+            });
+            ReceiveAny(msg =>
+            {
+                Stash.Stash();
+            });
+        }
+
+        private void Process()
+        {
+            ReceiveAny(msg =>
+            {
+                var sender = Sender;
+                var testActor = _testkit.TestActor;
+                if (_echoBackToSenderAsWell && testActor != sender)
+                    sender.Forward(msg);
+                testActor.Tell(msg, Sender);
+            });
+        }
+
+        /// <summary>
+        /// Returns a <see cref="Props"/> object that can be used to create an <see cref="EchoActor"/>.
+        /// The  <see cref="EchoActor"/> echoes whatever is sent to it, to the
+        /// TestKit's <see cref="TestKitBase.TestActor"/>.
+        /// By default it also echoes back to the sender, unless the sender is the <see cref="TestKitBase.TestActor"/>
+        /// (in this case the <see cref="TestKitBase.TestActor"/> will only receive one message) or unless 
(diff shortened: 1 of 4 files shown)


src/core/Akka.Tests/Dispatch/MailboxesSpec.cs
```diff
@@ -58,6 +59,68 @@ public IntPriorityMailbox(Settings settings, Config config) : base(settings, con
...
+                Stash.Stash();
+            });
+        }
+
+        private void Process()
+        {
+            ReceiveAny(msg =>
+            {
+                var sender = Sender;
+                var testActor = _testkit.TestActor;
```
Aaronontheweb: I'd just pass in an `IActorRef` instead of doing this stuff with the `TestKit` in the constructor.

src/core/Akka/Dispatch/MessageQueues/UnboundedPriorityMessageQueue.cs
```diff
@@ -14,9 +15,11 @@ namespace Akka.Dispatch.MessageQueues
     /// <summary> 
     /// Base class for a message queue that uses a priority generator for messages 
     /// </summary>
-    public class UnboundedPriorityMessageQueue : BlockingMessageQueue
+    public class UnboundedPriorityMessageQueue : BlockingMessageQueue, IUnboundedDequeBasedMessageQueueSemantics
```
Aaronontheweb: On the fence about this - it definitely adds deque semantics to the mailbox, but I'm now left wondering if we should be allowing stashing to work with the priority mailboxes in the first place.

If I think about it, the entire point of the prioritymailbox is to process messages in the order of the priority generator. FIFO no longer takes hold once you allow prioritization. So what good is stashing then? If I unstash a message I couldn't previously process, it either gets pushed back onto the queue in a totally different order (possibly not getting processed next) or if we stick with our normal double-ended queue semantics, it doesn't get processed in its normal priority order again.

I dunno, I'm leaning towards telling the users "don't use priority mailboxes with stashing." Either we just re-add messages we can't process back into the priority queue and re-prioritize them, which doesn't let us preserve message order (the re-added message will appear to be newer) or maybe we have to rewrite the prepend buffer to behave like a priority queue as well and process that first.

src/core/Akka.Tests/Dispatch/MailboxesSpec.cs
Danthar: This test actor was a quick copy paste from the EchoActor in the Testkit. But your right. It could be done simpler

src/core/Akka/Dispatch/MessageQueues/UnboundedPriorityMessageQueue.cs
Danthar: You are absolutely right. It does add some wierd semantics to the whole thing. I asked around on jvm akka channel about this. But haven't gotten any response as of yet.

The way this works now, is that messages are enqueued in the priority queue first. Then are handled by the actor, and then are stashed. Which means that messages are stashed in the order of priority as initially determined. When you unstash all, they are fed to the actor in that order as well.

This makes for some wierd behavior. Because it would seem to work when you simply stash all messages, and unstash all at a certain point. (see the spec i added as an example). But ofcourse this breaks down if you only intermittently stash and unstash a message.

Changing the prepend buffer to a priority queue wont solve this either. As that would mean the order would only be guarenteed within the context of the stashed messages, and not the mailbox in itself (the buffer of messages waiting will not be taken into account when determining priority)

Aaronontheweb: Welp, I think my final thoughts on the subject are: let the users worry about their message orders. If they want to use stashing and priority mailboxes where the results aren't 100% predictable or stable, that's up to them.

PR #2947, opened 2017-08-09
Implement cross-platform json serializer and spec

Arkatufus:
NOTE: 
- Only merge this when there are requests from users for net framework - net core cross platform support. 
- The serializer would have lower performance than the regular one.

### DIFF
--- a/src/core/Akka/Serialization/CompatJsonSerializer.cs
+++ b/src/core/Akka/Serialization/CompatJsonSerializer.cs
@@ -1,0 +1,302 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="NewtonSoftJsonSerializer.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Util;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
+
+namespace Akka.Serialization
+{
+    /// <inheritdoc />
+    /// <summary>
+    /// This is a special <see cref="P:Akka.Serialization.CompatJsonSerializer.Serializer" /> that serializes and deserializes javascript objects only.
+    /// These objects need to be in the JavaScript Object Notation (JSON) format.
+    /// </summary>
+    public class CompatJsonSerializer : Serializer
+    {
+        private static readonly string CoreAssemblyName = typeof(object).GetTypeInfo().Assembly.GetName().Name;
+
+        private readonly JsonSerializer _serializer;
+
+        /// <summary>
+        /// TBD
+        /// </summary>
+        public JsonSerializerSettings Settings { get; }
+
+        /// <summary>
+        /// TBD
+        /// </summary>
+        public object Serializer => _serializer;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NewtonSoftJsonSerializer" /> class.
+        /// </summary>
+        /// <param name="system">The actor system to associate with this serializer. </param>
+        public CompatJsonSerializer(ExtendedActorSystem system)
+            : this(system, NewtonSoftJsonSerializerSettings.Default)
+        {
+        }
+
+        public CompatJsonSerializer(ExtendedActorSystem system, Config config)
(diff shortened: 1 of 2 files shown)


src/core/Akka/Serialization/CompatJsonSerializer.cs
```diff
@@ -0,0 +1,302 @@
...
+        /// <summary>
+        /// Serializes the given object into a byte array
+        /// </summary>
+        /// <param name="obj">The object to serialize </param>
+        /// <returns>A byte array containing the serialized object</returns>
+        public override byte[] ToBinary(object obj)
+        {
+            string data = JsonConvert.SerializeObject(obj, Formatting.None, Settings);
+#if CORECLR
+            data = data.Replace("System.Private.CoreLib", "%CORE%");
```
alexvaluyskiy: it is not enough. There is more `System.Private.*` assemblies, like `System.Private.Uri`

ITHedgeHog:
Hi, we are potentially looking at doing this as we have certain back end services which will take 6 to 12 months to convert to .Net Core.

One question: when you say lower performance on the serializer, do you have any feel for how much of a performance hit we would be talking? Are we talking 10-20% hit or 80-90% hit?

paulinfrancis:
We are also interested in a hybrid architecture consisting of some legacy systems on full framework, and new development with .NET Core.

notour:
I copy this serializer and it work for me. (Client .net 4.7.1 -> Cluster .net Core 2.0)

I have to Add a specific Identifier and replace the Preprocess condition "#if CORECLR" by a runtime check if i'm runing in .net core or .net classic.

For now the only solution i found to check if i run in .net core is : 

        /// <summary>
        /// Gets the net core version.
        /// </summary>
        public static string GetNetCoreVersion()
        {
            var assembly = typeof(System.Runtime.GCSettings).GetTypeInfo().Assembly;
            var assemblyPath = assembly.CodeBase.Split(new[] { '/', '\\' }, StringSplitOptions.RemoveEmptyEntries);
            int netCoreAppIndex = Array.IndexOf(assemblyPath, "Microsoft.NETCore.App");
            if (netCoreAppIndex > 0 && netCoreAppIndex < assemblyPath.Length - 2)
                return assemblyPath[netCoreAppIndex + 1];
            return null;
        }

Aaronontheweb:
@notour would you like to send an updated PR with that code change? Appreciate the info!

angrybug:
We're looking for this capability also.

notour:
I made a pull request :

https://github.com/Arkatufus/akka.net/pull/2

leo12chandu:
I have an implementation for Hyperion to work across .Net Core to .NET Full framework. Currently it supports collections that switch between mscorlib to System.Private.CorLib as well as ExpandoObjects going from System.Dynamic to System.Linq.

https://github.com/leo12chandu/akka.net/blob/dev/src/core/Akka/Serialization/HyperionByteHelper.cs

john-cullen:
Hello,

What is the current state of this?

We are (unfortunately) still having to target .NET Framework as the WPF on .NET Core will not be ready for us in time and have just hit this issue.

Regards
JC

Aaronontheweb:
@john-cullen we should probably implement @leo12chandu's changes

leo12chandu:
@Aaronontheweb - I would be happy to branch out from akka.net master, implement it and create a pull request for you to review if it helps merge so others can use it. I've performance tested it 6 months back with reasonable size payloads and its does not add more than 10-20 milliseconds for our objects with over a 1000 properties per object.

Usage defined at https://github.com/akkadotnet/akka.net/issues/3587

awostenberg:
Any developments? 

We have a situation of a legacy .net framework app, new actors in .net core core... 

and I was pleasantly surprised today to find there is some level of interoperability out of the box (it formed a cluster, an actor in .net core taking strings from actor in .net framework worked fine)
but quickly hit deserialization problems,this PR would fix, e.g. "Error [Error resolving type specified in JSON 'System.Collections.Generic.List`1[[Shared.Messages.MergeOneString, Shared]], System.Private.CoreLib'."

john-cullen:
@awostenberg I went with modifying the serialised text for now. I've not had any problems with it (but my app is still in testing / pre-release)

https://gist.github.com/john-cullen/c39d56efcb15e60b66cd4d05149316ed

edvinasz:
@john-cullen you saved me.  It took me 3 days of banging my head against why our system is just not working.  Long and behold once we connected our WPF .NET framework clients they don't communicate messages with arrays in them (with cryptic dissociation errors).  Once I discovered this thread I was concerned that this is going to be a showstopper for our effort to port our WCF to akka.net (months of effort), since our WPF apps can't be converted to .net core anytime soon.  It would be really nice, going forward, to integrate @leo12chandu solution for planed 1.5 release when akka.net switches to Hyperion.  Thanks.

Aaronontheweb:
@Arkatufus I know you're OOF, but maybe we should just merge this in - I can think of several support customers who are going to run into this as they execute their migrations off of .NET Framework this year.

Aaronontheweb:
Probably need a feature toggle for this.

PR #3003, opened 2017-08-17
fix Akka.Serialization.Hyperion dependency

Aaronontheweb:
cc @Ralf1108 - this will fail to compile if the build server is the issue.


src/contrib/serializers/Akka.Serialization.Hyperion/HyperionSerializer.cs
```diff
@@ -70,7 +70,8 @@ public HyperionSerializer(ExtendedActorSystem system, HyperionSerializerSettings
                     preserveObjectReferences: settings.PreserveObjectReferences,
                     versionTolerance: settings.VersionTolerance,
                     surrogates: new[] { akkaSurrogate },
-                    knownTypes: provider.GetKnownTypes()));
+                    knownTypes: provider.GetKnownTypes(), 
+                    ignoreISerializable:true));
```
Aaronontheweb: IMHO, I'm fine with leaving this setting on. `ISerializable` is dangerous and shouldn't be used for anything other than local serialization, not cross-network.

alexvaluyskiy: @Aaronontheweb But we don't have any alternatives for Exceptions. We can use only `ISerializable` serialization

Ralf1108: old behavior would be to ignore it because ISerializable support in Hyperion was introduced only in latest release.
Maybe you can just release a fixed version for "Akka.Serialization.Hyperion" without ISerialization support and discuss this afterwards ?

Danthar: I agree, this PR is not about using ignoreISerializable or not. Its about fixing the package. Default value is false, so set it at false. And lets have the discussion on when to turn this option on, a different time.

Aaronontheweb: > old behavior would be to ignore it because ISerializable support in Hyperion was introduced only in latest release.

Agreed. We weren't using it before. Status quo.

PR #3004, opened 2017-08-17
[api-docs][misc] Fix various issues with api-docs / copyright headers

sean-gilliam:
This PR fixes various issues with api-doc tags. It also fixes / adds missing copyright headers.


Aaronontheweb:
@sean-gilliam for the copyright headers, are you using a R# template?

sean-gilliam:
I just have a dumb script that does it. Though it would be awesome if someone had one to contribute.

Issue #2995, opened 2017-08-15
Udp sending issue (Akka.Net 1.3.0)


blackclavus:
Hello, Guys. 
There an issue when sending a data via UDP if ByteString's payload has buffers with length more than 1. UdpSender will only write the first part of buffers and drop the rest then unable to write anything.

`
Example Code:
```
using Akka.Actor;
using Akka.Configuration;
using Akka.IO;
using System;
using System.Net;
using System.Net.Sockets;

namespace UdpSendTest
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var system = ActorSystem.Create("udp-send-test", 
                ConfigurationFactory.ParseString(@"
                    akka.suppress-json-serializer-warning = on
                    akka.io.udp.max-channels = unlimited
                    akka.io.udp.nr-of-selectors = 1
                    akka.io.udp.direct-buffer-pool-limit = 100
                    akka.io.udp.direct-buffer-size = 1024
                    akka.io.udp.trace-logging = on
                    akka.loglevel = DEBUG")))
            {
                IPEndPoint serverAddress = TemporaryServerAddress();

                var handler = system.ActorOf(Props.Create(() => new Handler()));
                system.ActorOf(Props.Create(() => new TestActor(handler, serverAddress)));

                Console.ReadLine();
            }
        }

        class TestActor : ReceiveActor
        {
            private IActorRef _server;
            private IPEndPoint _address;
            
            public TestActor(IActorRef handler, IPEndPoint address)
            {
                _address = address;

                Receive<Udp.Bound>(bound =>
                {
                    _server = Sender;

                    Context.System.Udp().Tell(Udp.SimpleSender.Instance);
                });

                Receive<Udp.SimpleSenderReady>(_ =>
                {
                    //Create ByteString from multiple instance
                    var data = ByteString.FromString("this") + ByteString.FromString(" is ") + 
                        ByteString.FromString("string");

                    Sender.Tell(Udp.Send.Create(data, _address));
                });

                Context.System.Udp().Tell(new Udp.Bind(handler, address));
            }
        }

        class Handler : ReceiveActor
        {
            public Handler()
            {
                Receive<Udp.Received>(x =>
                {
                    //Target will only receive first part of buffer
                    string received = x.Data.ToString();
                    Console.WriteLine($"Received ({x.Data.Count}): {received}");
                });
            }
        }

        public static IPEndPoint TemporaryServerAddress()
        {
[...]

PR #3005, opened 2017-08-17
Updates and fixes for Akka.IO UDP support

Horusiath:
This should fix the issue reported in #2995 (PR contains tests which reproduced the issue). It also upgrades the codebase to C# 7.

### DIFF
--- a/src/core/Akka.Tests/IO/UdpIntegrationSpec.cs
+++ b/src/core/Akka.Tests/IO/UdpIntegrationSpec.cs
@@ -28,7 +28,8 @@ namespace Akka.Tests.IO
                     akka.io.udp.nr-of-selectors = 1
                     akka.io.udp.direct-buffer-pool-limit = 100
                     akka.io.udp.direct-buffer-size = 1024
-                    akka.loglevel = INFO", output)
+                    #akka.io.udp.trace-logging = on
+                    akka.loglevel = DEBUG", output)
         {
             _addresses = TestUtils.TemporaryServerAddresses(6, udp: true).ToArray();
         }
@@ -55,6 +56,20 @@ namespace Akka.Tests.IO
             var serverAddress = _addresses[0];
             var server = BindUdp(serverAddress, TestActor);
             var data = ByteString.FromString("To infinity and beyond!");
+            SimpleSender().Tell(Udp.Send.Create(data, serverAddress));
+
+            ExpectMsg<Udp.Received>(x => x.Data.ShouldBe(data));
+        }
+
+        [Fact]
+        public void The_UDP_Fire_and_Forget_implementation_must_be_able_to_send_multipart_ByteString_without_binding()
+        {
+            var serverAddress = _addresses[0];
+            var server = BindUdp(serverAddress, TestActor);
+            var data = ByteString.FromString("This ") 
+                + ByteString.FromString("is ") 
+                + ByteString.FromString("multiline ") 
+                + ByteString.FromString(" string!");
             SimpleSender().Tell(Udp.Send.Create(data, serverAddress));
 
             ExpectMsg<Udp.Received>(x => x.Data.ShouldBe(data));
--- a/src/core/Akka/IO/SocketEventArgsPool.cs
+++ b/src/core/Akka/IO/SocketEventArgsPool.cs
@@ -11,6 +11,7 @@ using System.Collections.Concurrent;
 using System.Net.Sockets;
 using System.Runtime.CompilerServices;
 using Akka.IO.Buffers;
+using Akka.Util;
 
 namespace Akka.IO
 {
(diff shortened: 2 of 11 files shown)


Aaronontheweb:
@Horusiath some UDP specs on Mono failing following this change

Danthar:
Because you also refactored to use new patterns and stuff. Its hard to review. Could you point out what you changed to fix the problem ?

Horusiath:
Essentially the problem was that I didn't updated the UDP handler code in line to the changes I was making in `TcpConnection` during Akka.IO rework. In previous versions we were sending `ByteString` via socket chunk by chunk (waiting for the previous chunk to be sent before sending next one). New implementation (that already is present in `TcpConnection`) makes smart distinction between compact and multi-chunk `ByteString`s:

- Compact byte string consist only of single chunk, so it's simply set as `SocketAsyncEventArgs.SetBuffer(buffer.Array, buffer.Offset, buffer.Count)` before sending.
- However multi-chunk byte string can also be send in a single call, thanks to `SocketAsyncEventArgs.BufferList` property.

I've extracted logic responsible for recognizing which method to use (`SetBuffer`/`BufferList`) from `TcpConnection` into single extension method [SocketAsyncEventArgs.SetBuffer(ByteString)](https://github.com/akkadotnet/akka.net/pull/3005/files#diff-c1896e208456b262362b837e0f76cd05), which now allows us to handle socket send request via TCP/UDP in quite similar way.

alexvaluyskiy:
@Horusiath you made a lot of API changes, it will be impossible to ship it in 1.3.x. Is it possible to bring back some constructors and properties and make them obsoleted?

Horusiath:
@alexvaluyskiy 90% of API changes is about sealing message classes (which should be sealed). There's a one change in the `Send` constructor, but the problem is more that you shouldn't even build it via constructor - which should be hidden - but rather via `Send.Create(ByteString)`. I can make an obsoleted constructor, but `Send.Payload` should be `ByteString` anyway.

Horusiath:
The reason of failure on Mono found: *Mono doesn't support using BufferList at this point.* This probably means, that the same problem concerns TCP, we just didn't have tests to detect that. 

As I definitely don't want to fallback the whole ability to send multiple buffers at once or build a separate (and very complex) buffer send awaiting logic just to cover Mono, my proposal is: If we detect, that we're on Mono, we use `ByteString.Compact()` method, to squash multi-buffer `ByteString` into single-buffer one. This is copying operation. it sucks for performance and memory usage, but it will cripple only some of the cases (multi-buffer `ByteString`s) and only on Mono.

Are we fine with that? /cc @akkadotnet/core

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -80,4 +81,38 @@ private SocketAsyncEventArgs CreateSocketAsyncEventArgs()
...
+                args.SetBuffer(buffer.Array, buffer.Offset, buffer.Count);
+            }
+            else
+            {
+                if (RuntimeDetector.IsMono)
+                {
+                    // Mono doesn't support BufferList - falback to compacting ByteString
+                    var compacted = data.Compact();
+                    var buffer = compacted.Buffers[0];
+                    args.SetBuffer(buffer.Array, buffer.Offset, buffer.Count);
```
Horusiath: If we detect a Mono runtime, we'll try to fallback and compact multiple byte buffers into one (this operations copies buffers onto newly allocated array).

Aaronontheweb:
> If we detect, that we're on Mono, we use ByteString.Compact() method, to squash multi-buffer ByteString into single-buffer one. 

Fine with me. Not our job to implement the missing parts of the Mono runtime.

src/core/Akka.Tests/IO/UdpIntegrationSpec.cs
```diff
@@ -61,6 +62,20 @@ public void The_UDP_Fire_and_Forget_implementation_must_be_able_to_send_without_
         }
 
         [Fact]
+        public void The_UDP_Fire_and_Forget_implementation_must_be_able_to_send_multipart_ByteString_without_binding()
```
Aaronontheweb: Looks good to me

src/core/Akka/IO/SocketEventArgsPool.cs
Aaronontheweb: looks good to me

src/core/Akka/IO/UdpConnected.cs
```diff
@@ -152,44 +152,48 @@ public NoAck(object token)
         /// </summary>
         public sealed class Send : Command
         {
+            [Obsolete("Akka.IO.Udp.Send public constructors are obsolete. Use `Send.Create` or `Send(ByteString, EndPoint, Event)` instead.")]
+            public Send(IEnumerator<ByteBuffer> payload, Event ack)
```
Aaronontheweb: Looks good

src/core/Akka/Util/ByteString.cs
```diff
@@ -35,6 +35,16 @@ public sealed class ByteString : IEquatable<ByteString>, IEnumerable<byte>
         #region creation methods
 
         /// <summary>
+        /// INTERNAL API: remove this method once <see cref="Udp.Send"/> public constructor will be removed.
+        /// </summary>
+        internal static ByteString FromBuffers(IEnumerator<ByteBuffer> buffers)
```
Aaronontheweb: Looks good

PR #3011, opened 2017-08-19
Address & UniqueAddres -> IComparable interface

### LINKED ISSUE #3010: Akka.DistributedData - failing to propagate deltas
#### Package

Akka.DistributedData version 1.3

#### How to reproduce the problem

1. Create an Akka cluster with (in my case) at least one distributed data aware node. 
2. Make some updates on that node, using i.e. `await DistributedData.Get(system).UpdateAsync(key, crdt, new WriteAll(timeout))`.
3. After performing some of the updates start next 1 or 2 distributed data aware nodes.
4. At least one of them will catch following error during sync.

#### Stack trace

> [ERROR][19.08.2017 18:31:02][Thread 0004][akka://crdt-demo/user/ddataReplicator] At least one object must implement IComparable.
Cause: System.ArgumentException: At least one object must implement IComparable.
   at System.Collections.Comparer.Compare(Object a, Object b)
   at System.Linq.EnumerableSorter`2.CompareAnyKeys(Int32 index1, Int32 index2)
   at System.Linq.EnumerableSorter`1.CompareKeys(Int32 index1, Int32 index2)
   at System.Linq.EnumerableSorter`1.QuickSort(Int32[] map, Int32 left, Int32 right)
   at System.Linq.EnumerableSorter`1.Sort(TElement[] elements, Int32 count)
   at System.Linq.OrderedEnumerable`1.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at System.Collections.Immutable.ImmutableArray.CreateRange[T](IEnumerable`1 items)
   at Akka.DistributedData.Replicator.ReplicatorDeltaPropagationSelector.get_AllNodes()
   at Akka.DistributedData.DeltaPropagationSelector.CollectPropagations()
   at Akka.DistributedData.Replicator.ReceiveDeltaPropagationTick()
   at Akka.DistributedData.Replicator.<NormalReceive>b__46_7(DeltaPropagationTick _)
   at lambda_method(Closure , Object , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Object[] )
   at Akka.Tools.MatchHandler.PartialHandlerArgumentsCapture`16.Handle(T value)
   at Akka.Actor.ReceiveActor.OnReceive(Object message)

This stack trace points exactly here: https://github.com/akkadotnet/akka.net/blob/c12d08582bbeda6bb4d024df4f3eb992f4a00e0f/src/contrib/cluster/Akka.DistributedData/Replicator.cs#L1409-L1410

The reason is that we cannot compare two `Address`es in our current implementation. On the scala side, `Address`es can be sorted using implicit ordering: 

```scala
  implicit val addressOrdering: Ordering[Address] = Ordering.fromLessThan[Address] { (a, b) ⇒
    if (a eq b) false
    else if (a.protocol != b.protocol) a.system.compareTo(b.protocol) < 0
    else if (a.system != b.system) a.system.compareTo(b.system) < 0
    else if (a.host != b.host) a.host.getOrElse("").compareTo(b.host.getOrElse("")) < 0
    else if (a.port != b.port) a.port.getOrElse(0) < b.port.getOrElse(0)
    else false
  }
````

### PR DESCRIPTION
**Pull request by Horusiath**
This PR introduces an `IComparable<>`/`IComparable` interface implementations for both `Address`/`UniqueAddress`, which should help to fix #3010 and allow us to keep our implementations in line with Akka JVM code.

**WARNING: There is a difference in `UniqueAddress.CompareTo(UniqueAddress)` implementation!** Previously to compare address parts, a `Member.AddressOrdering` was used. This was different from JVM implementation, which by default uses `Address.addressOrdering`. The difference between the two is that the first one ignores the protocol and system name part of the address during comparison.

### DIFF
--- a/src/core/Akka.Cluster/Member.cs
+++ b/src/core/Akka.Cluster/Member.cs
@@ -420,7 +420,7 @@ namespace Akka.Cluster
     /// The `uid` is needed to be able to distinguish different
     /// incarnations of a member with same hostname and port.
     /// </summary>
-    public class UniqueAddress : IComparable<UniqueAddress>, IEquatable<UniqueAddress>
+    public class UniqueAddress : IComparable<UniqueAddress>, IEquatable<UniqueAddress>, IComparable
     {
         /// <summary>
         /// The bound listening address for Akka.Remote.
@@ -468,16 +468,21 @@ namespace Akka.Cluster
         /// <summary>
         /// TBD
         /// </summary>
-        /// <param name="that">TBD</param>
+        /// <param name="uniqueAddress">TBD</param>
         /// <returns>TBD</returns>
-        public int CompareTo(UniqueAddress that)
+        public int CompareTo(UniqueAddress uniqueAddress)
         {
-            var result = Member.AddressOrdering.Compare(Address, that.Address);
-            if (result == 0)
-                if (Uid < that.Uid) return -1;
-                else if (Uid == that.Uid) return 0;
-                else return 1;
-            return result;
+            if (uniqueAddress == null) throw new ArgumentNullException(nameof(uniqueAddress));
+
+            var result = Address.Comparer.Compare(Address, uniqueAddress.Address);
+            return result == 0 ? Uid.CompareTo(uniqueAddress.Uid) : result;
+        }
+
+        int IComparable.CompareTo(object obj)
+        {
+            if (obj is UniqueAddress address) return CompareTo(address);
+
+            throw new ArgumentException($"Cannot compare {nameof(UniqueAddress)} with instance of type '{obj?.GetType().FullName ?? "null"}'.");
         }
 
         /// <inheritdoc cref="object.ToString"/>
--- a/src/core/Akka.Cluster/Member.cs
+++ b/src/core/Akka.Cluster/Member.cs
@@ -21,7 +21,7 @@ namespace Akka.Cluster
     /// NOTE: <see cref="GetHashCode"/> and <see cref="Equals"/> are solely based on the underlying <see cref="Address"/>, 
     /// not its <see cref="MemberStatus"/> and roles.
     /// </remarks>
-    public class Member : IComparable<Member>
+    public class Member : IComparable<Member>, IComparable
     {
         /// <summary>
         /// TBD
@@ -112,9 +112,13 @@ namespace Akka.Cluster
         }
 
         /// <inheritdoc cref="IComparable.CompareTo"/>
-        public int CompareTo(Member other)
+        public int CompareTo(Member other) => Ordering.Compare(this, other);
+
+        int IComparable.CompareTo(object obj)
         {
-            return Ordering.Compare(this, other);
+            if (obj is Member member) return CompareTo(member);
+
+            throw new ArgumentException($"Cannot compare {nameof(Member)} to an instance of type '{obj?.GetType().FullName ?? "null"}'");
         }
 
         /// <inheritdoc cref="object.ToString"/>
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/core/Akka.Cluster/Member.cs [hunk 1]
```diff
@@ -468,16 +468,21 @@ public override int GetHashCode()
...
         /// </summary>
-        /// <param name="that">TBD</param>
+        /// <param name="uniqueAddress">TBD</param>
         /// <returns>TBD</returns>
-        public int CompareTo(UniqueAddress that)
+        public int CompareTo(UniqueAddress uniqueAddress)
+        {
+            if (uniqueAddress == null) throw new ArgumentNullException(nameof(uniqueAddress));
+
+            var result = Address.Comparer.Compare(Address, uniqueAddress.Address);
```
Horusiath: This changes how `UniqueAddress`es are compared - previously a protocol and system name parts were ignored. I think it was a bug, and it's not like the JVM version was working. It should be correct now.

src/core/Akka.Cluster/Member.cs
```diff
@@ -192,10 +196,13 @@ internal class AddressComparer : IComparer<Address>
             /// <inheritdoc cref="IComparer{Address}.Compare"/>
             public int Compare(Address x, Address y)
             {
-                if (x.Equals(y)) return 0;
-                if (!x.Host.Equals(y.Host)) return String.Compare(x.Host.GetOrElse(""), y.Host.GetOrElse(""), StringComparison.Ordinal);
-                if (!x.Port.Equals(y.Port)) return Nullable.Compare(x.Port.GetOrElse(0), (y.Port.GetOrElse(0)));
```
Horusiath: There's no sense is all those Equals comparisons, they are redundant in context of comparison (especially `x.Equals(y)` which basically is comparison made twice here).

Aaronontheweb:
@Horusiath same MNTR spec failed on both .NET Core and .NET full.

Horusiath:
So far I've found the failing assertion: https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs#L1166

src/core/Akka/Actor/Address.cs
```diff
@@ -45,7 +45,7 @@ public int Compare(Address x, Address y)
                 if (result != 0) return result;
                 result = string.CompareOrdinal(x.Host ?? "", y.Host ?? "");
                 if (result != 0) return result;
-                result = Nullable.Compare(x.Port, x.Port);
+                result = (x.Port ?? 0).CompareTo(y.Port ?? 0);
```
Horusiath: It looks like this might have been source of failing cluster specs. From what I've investigated, `Nullable.Compare<int>(int?, int?)` is not reliable and may return 0 for two different numbers. Waiting for CI to confirm.

Danthar: Nasty. Good find!

Horusiath:
All tests have passed. I think it's ready. /cc @Danthar @Aaronontheweb

Danthar:
In theory this could break backwards compat for cluster users. Meaning if they update to the next release (if we merge this in now) they would have to update their entire cluster. 

I say in theory, because the member comparison bugfix, may cause members to be compared differently as it was before. Causing subtle bugs if your running different versions in one cluster.

But we can't do much about that now. This is a bug that needs to be fixed. 
So im merging it.

PR #3022, opened 2017-08-24
Improving Support for OSX & Rider

joshgarnett:
* Rider/OSX appears to be less forgiving around the csproj format and requires that the nuget dependencies use Version and not version.
* Updated the build.sh script to only install dotnet once and also to pull down protoc from nuget.
* The FAKE build has been updated to support building the Protobuf target on OSX & Linux

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -51,12 +51,13 @@ if [ ! -f "$DOTNET_EXE" ]; then
     fi
     curl -Lsfo "$SCRIPT_DIR/.dotnet/dotnet-install.sh" $DOTNET_INSTALLER_URL
     bash "$SCRIPT_DIR/.dotnet/dotnet-install.sh" --version $DOTNET_VERSION --install-dir .dotnet --no-path
-    export PATH="$SCRIPT_DIR/.dotnet":$PATH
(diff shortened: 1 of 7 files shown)


Arkatufus:
Looks good :smile:

Aaronontheweb:
I'm not sure what the issue is here exactly, but the Mono test runner is hitting the deck when it starts scanning for test assemblies to run... What could be causing that?

joshgarnett:
It doesn't seem like the failure was related to the change from what I can tell.  I'll look at getting the branch caught up to dev.

build.sh
```diff
@@ -42,17 +44,19 @@ fi
 # INSTALL .NET CORE CLI
 ###########################################################################
 
-echo "Installing .NET CLI..."
```
Danthar: This change for not installing .NET CLI is whats causing the issue here. Revert this section and it should solve your problem

build.sh
```diff
@@ -87,6 +91,30 @@ if [ ! -f "$FAKE_EXE" ]; then
 fi
 
 ###########################################################################
+# INSTALL PROTOC
+###########################################################################
+
+unameOut="$(uname -s)"
```
Danthar: What does this line do ?

Issue #3021, opened 2017-08-24
Akka.IO.Tcp fails to write some outgoing messages [1.3.0]


Aaronontheweb:
Working on upgrading Petabridge.Cmd and hit this issue:
```
[DEBUG][8/24/2017 1:53:25 AM][Thread 0004][[akka://akkacmd/user/petabridge.cmd/127.0.0.1%3A51015#756879318]] Sending [RESPONSE][FINAL=False]: [INFO][8/24/2017 1:53:24 AM][Thread 0021][[akka://akkacmd/user/petabridge.cmd#2122656557]] petabridge.cmd host bound to [0.0.0.0:9110] to client [277 bytes]
[DEBUG][8/24/2017 1:53:25 AM][Thread 0004][[akka://akkacmd/user/petabridge.cmd/127.0.0.1%3A51015#756879318]] Sending [RESPONSE][FINAL=False]: [INFO][8/24/2017 1:53:24 AM][Thread 0021][Petabridge.Cmd.Host.Tests.PetabridgeCmdConnectionSpecs] 1 to client [242 bytes]
[DEBUG][8/24/2017 1:53:25 AM][Thread 0004][[akka://akkacmd/user/petabridge.cmd/127.0.0.1%3A51015#756879318]] Sending [RESPONSE][FINAL=True]: [ERROR][8/24/2017 1:53:24 AM][Thread 0021][Petabridge.Cmd.Host.Tests.PetabridgeCmdConnectionSpecs] 2
Cause: Unknown to client [259 bytes]
```
sending a total of 778 bytes

```
 [DEBUG][8/24/2017 1:53:25 AM][Thread 0018][[akka://akkacmd/user/$a#660563327]] Received [256 bytes] from host
[DEBUG][8/24/2017 1:53:25 AM][Thread 0018][[akka://akkacmd/user/$a#660563327]] Received [21 bytes] from host
[DEBUG][8/24/2017 1:53:25 AM][Thread 0018][[akka://akkacmd/user/$a#660563327]] Deserialized [[RESPONSE][FINAL=False]: [INFO][8/24/2017 1:53:24 AM][Thread 0021][[akka://akkacmd/user/petabridge.cmd#2122656557]] petabridge.cmd host bound to [0.0.0.0:9110]] from host
[DEBUG][8/24/2017 1:53:25 AM][Thread 0018][[akka://akkacmd/user/$a#660563327]] Received [256 bytes] from host
[...]

PR #3023, opened 2017-08-24
fix 3021 - returned Akka.IO.Tcp semantics to use synchronous write

Aaronontheweb:
TL;DR;, the design of this class as it currently stands is a port of the JVM and all of its semantics around queueing pending writes, backpressure, etc, all depend on synchronous socket sends. We still use async socket reads - those work fine. But all writes are performed synchronously as of this PR.

This resolves the issue I outlined in #3021 and makes it possible to queue writes with Akka.IO.Tcp again.

### DIFF
--- a/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
+++ b/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
@@ -17,6 +17,7 @@ using Akka.TestKit;
 using Akka.Util.Internal;
 using Xunit;
 using Xunit.Abstractions;
+using FluentAssertions;
 
 namespace Akka.Tests.IO
 {
@@ -151,8 +152,8 @@ namespace Akka.Tests.IO
             {
                 var actors = x.EstablishNewClientConnection();
 
-                // create a 256 byte string
-                var str = Enumerable.Repeat("f", 128).Join("");
+                // create a large-ish byte string
+                var str = Enumerable.Repeat("f", 567).Join("");
                 var testData = ByteString.FromString(str);
 
                 // queue 3 writes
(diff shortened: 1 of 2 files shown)


src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -144,6 +145,31 @@ public void The_TCP_transport_implementation_should_properly_handle_connection_a
             });
         }
 
+        [Fact]
+        public void BugFix_3021_Tcp_Should_not_drop_large_messages()
```
Aaronontheweb: Reproduction spec. Now passes.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -829,28 +830,40 @@ public override PendingWrite DoWrite(ConnectionInfo info)
             {
                 try
                 {
-                    if (!data.IsEmpty)
+                    PendingWrite writeToChannel(ByteString data)
```
Aaronontheweb: Using a local function to recursively perform writes until the current write is fully flushed. 

This actor will need to be redesigned to support async writes and will probably need to include a gather --> batch mechanism similar to what Helios 2.* and DotNetty use.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -829,28 +830,40 @@ public override PendingWrite DoWrite(ConnectionInfo info)
...
                     {
-                        connection.SetStatus(ConnectionStatus.Sending);
-
-                        sendArgs.SetBuffer(data);
-
-                        if (!connection.Socket.SendAsync(sendArgs))
-                            self.Tell(SocketSent.Instance);
+                        var bytesWritten = _connection.Socket.Send(data.Buffers);
+                        if (_connection.traceLogging)
```
Aaronontheweb: Added support for output trace logging while I was at it

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -829,28 +830,40 @@ public override PendingWrite DoWrite(ConnectionInfo info)
...
                 }
                 catch (SocketException e)
                 {
-                    connection.HandleError(info.Handler, e);
+                    _connection.HandleError(info.Handler, e);
                     return this;
                 }
             }
 
             public override void Release() { }
```
Aaronontheweb: Really should find a way to use buffer pooling here the same way the JVM does, unless we bolt that onto `SocketAsyncEventArgs` pooling.

PR #3025, opened 2017-08-24
Add TestKitBase.IgnoreMessages<TMsg>

ayoung:
Just thought this would be convenient.

Also filled in some tests for non-generic IgnoreMessages().


Danthar:
lgtm 👍 needs to be updated though

Issue #3019, opened 2017-08-23
Persistence plugins need to support SerializerWithStringManifest


Danthar:
Right now it seems that the Sqlite and the SqlServer plugins does not support `SerializerWithStringManifest`. 

See: https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Serialization/SqliteJournalSerializationSpec.cs#L51 For Sqlite.

The current SqlServer plugin is being updated to 1.3. See: https://github.com/akkadotnet/Akka.Persistence.SqlServer/pull/73
However that PR still does not support `SerializerWithStringManifest`. 

Support for `SerializerWithStringManifest` is important. Because cluster sharding demands it. So right now using cluster sharding in combination with the Sql plugins will cause problems.

Danthar:
Solved via: https://github.com/akkadotnet/akka.net/pull/3029 and https://github.com/akkadotnet/akka.net/pull/3030

PR #3028, opened 2017-08-24
Fix Akka.Persistence with string manifest

joshgarnett:
This addresses #3019 

* Ported over MessageSerializer & SnapshotSerializer from Akka.Scala
* Switched Akka.Persistence.Sql.Common to use the Akka.Persistence serializers
* LocalSnapshotStore automatically picks up the new serializer and should work better with protobuf based snapshots


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -628,8 +628,8 @@ protected DbCommand GetCommand(DbConnection connection, string sql)
         protected virtual void WriteEvent(DbCommand command, IPersistentRepresentation e, IImmutableSet<string> tags)
         {
             var manifest = string.IsNullOrEmpty(e.Manifest) ? e.Payload.GetType().TypeQualifiedName() : e.Manifest;
-            var serializer = Serialization.FindSerializerFor(e.Payload);
-            var binary = serializer.ToBinary(e.Payload);
+            var serializer = Serialization.FindSerializerForType(typeof(IPersistentRepresentation));
```
alexvaluyskiy: The main idea of all SQL plugins is to store payload only into `Payload` column, not the whole message. We are using columns to store the header information, like `PersistenceId`, `SequenceNr`. So, we don't need any Protobuf serializers there, because those serializers don't serialize a payload.

joshgarnett: The SQL columns should only be used for storing fields that need to be queried on.  This is the pattern that JVM Akka follows for their Akka plugins.  All of their supported plugins just use the PersistentRepr protobuf serializer, mainly because it does the heavy lifting of ensuring that the actual payload is properly handled.

joshgarnett: This actually makes it much simpler for the writers of plugins as they don't have to worry about it.  This is a pattern I strongly suggest we adopt from the JVM Akka project.

joshgarnett: One problem a co-worker pointed out though, is this will break existing users of the sql plugin on an upgrade.  I still think the serialization changes are good to pick up and leverage moving forward, but I'll go ahead and rework the SQL changes specifically.  Moving forward it'd be worth considering using the new model and perhaps we can create a plugin that has breaking changes.

Aaronontheweb: Existing users since 1.3.0? No one has been able to use Akka.Persistence on that version yet

Aaronontheweb: Existing users since 1.3.0? No one has been able to use Akka.Persistence on that version yet

Horusiath:
While idea of keeping `SerializerId` as a column field is neat (I was thinking about the same for some time), there's a problem with message serializers - it will break backward compatibility. There's a migration path needed here: probably SerializerId could be used to serialize old message types using old serializer, while still allowing for new serialization semantics.

joshgarnett:
I'm going to break this out into two separate PRs

PR #3029, opened 2017-08-25
Protobuf Serialization for Akka.Persistence.IMessage

joshgarnett:
* Ported over MessageSerializer & SnapshotSerializer from JVM Akka
* Custom serializers are now being used as expected.  Previously, as is evident by the unit test changes, custom serializers were not being used when serializing Akka.Persistence.IMessage.
* LocalSnapshotStore automatically picks up the new serializer and should work better with protobuf based snapshots


src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
```diff
@@ -0,0 +1,167 @@
...
+            message.Payload = GetPersistentPayload(persistent.Payload);
+            message.SequenceNr = persistent.SequenceNr;
+            message.Deleted = persistent.IsDeleted;
+
+            return message;
+        }
+
+        private PersistentPayload GetPersistentPayload(object obj)
+        {
+            Serializer serializer = system.Serialization.FindSerializerFor(obj);
```
Danthar: Needs DefaultSerializer persistence setting support.

Danthar: But i dont consider it a blocking issue. We can fix it later. Its not a thing that will break any existing stuff

joshgarnett: As we talked about in chat, the current dev branch does not support specifying the default serializer for persistence serialization, it just uses whatever has been defined for the objects in the serialization-bindings.

Currently the dev branch doesn't actually support custom serializers at all for the payload within the IPersistentRepresentation, this was fixed and is reflected in the unit test changes.

PR #3030, opened 2017-08-25
Added support for SerializerWithStringManifest to Akka.Persistence.Sql.Common

joshgarnett:
This addresses #3019

* Added a serializer id to the sql columns
* Updated how the manifest is set to support SerializerWithStringManifest
* ReadEvent should be backwards compatible

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
@@ -6,7 +6,6 @@
 //-----------------------------------------------------------------------
 
 using System;
-using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Data;
 using System.Data.Common;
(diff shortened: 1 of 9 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -627,8 +641,21 @@ protected DbCommand GetCommand(DbConnection connection, string sql)
         /// <returns>TBD</returns>
         protected virtual void WriteEvent(DbCommand command, IPersistentRepresentation e, IImmutableSet<string> tags)
         {
-            var manifest = string.IsNullOrEmpty(e.Manifest) ? e.Payload.GetType().TypeQualifiedName() : e.Manifest;
             var serializer = Serialization.FindSerializerFor(e.Payload);
```
Danthar: Should use Configuration.DefaultSerializer

joshgarnett: Fixed

Danthar:
@Horusiath we cannot put this off untill a later release. Supporting SerializerWithStringManifest is crucial for cluster sharding to work. Because since 1.3 its broken.

PR #3032, opened 2017-08-27
Updating ProtobufSerializer to automatically lookup the MessageParser

joshgarnett:
The existing implementation of ProtobufSerializer requires RegisterFileDescriptor to be called with the file descriptor of every protobuf message that will be serialized.  This adds a significant amount of boilerplate to serialize protobuf messages automatically.  To work around this, I've updated the fromBinary method to lookup the MessageParser at runtime based on the type.  This brings it inline with how the JVM Akka protobuf serializer works.


alexvaluyskiy:
Have you benchmarked it?

joshgarnett:
I have not benchmarked this specific change, but benchmarks of Activator.CreateInstance with the default constructor, which we are using in this case, show that the overhead is minimal, plus we are caching the lookup, so at that point we'll only pay the cost of getting the key from the concurrent dictionary.

joshgarnett:
TryGetValue is lockfree for ConcurrentDictionary btw

joshgarnett:
Correct, the idea is that if a user wants to send IMessage over the wire, the remoting protobuf serializer would automatically pick this up without any additional work from the user.

src/core/Akka.Remote/Serialization/ProtobufSerializer.cs
```diff
@@ -58,9 +48,15 @@ public override byte[] ToBinary(object obj)
         public override object FromBinary(byte[] bytes, Type type)
         {
             if (TypeLookup.TryGetValue(type.FullName, out var parser))
+            {
                 return parser.ParseFrom(bytes);
-
-            throw new ArgumentException("Need a protobuf message class to be able to serialize bytes using protobuf");
+            }
+            // MethodParser is not in the cache, look it up with reflection
+            IMessage msg = Activator.CreateInstance(type) as IMessage;
```
Aaronontheweb: That's clever.

Aaronontheweb:
Going to run a local ping pong once I pull this down. the NBench numbers were in the 20k range.

Aaronontheweb:
Benchmark numbers while I was running ~8 instances of Visual Studio on my machine:

```
ProcessorCount:                    8
Actor Count:                       16
Messages sent/received per client: 20000  (2e4)

         1,   20000,     42373,     472.50
         5,  100000,     44884,    2228.36
        10,  200000,     52070,    3841.58
        15,  300000,     55361,    5419.78
        20,  400000,     50814,    7872.45
        25,  500000,     52643,    9498.59
```

It's right on with top numbers I recorded before 1.3.0. Good to go on the perf IMHO.

alexvaluyskiy:
@Aaronontheweb I'm not sure that we are talking about the same things. Because we don't have this `ProtobufSerializer` before 1.3.0
https://github.com/akkadotnet/akka.net/blob/v1.2.3/src/core/Akka.Remote/Serialization/ProtobufSerializer.cs

Aaronontheweb:
@alexvaluyskiy I'm talking about the work done on the 1.3.0 branch, when I added the benchmark.

PR #3033, opened 2017-08-27
sharding c# 7

zbynek001:


Horusiath:
Quick question: Have you changed anything in the implementation itself, you have you just upgraded code style to new C# 7 features?

zbynek001:
nothing else was changed, only adding c# 7 features

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -219,20 +220,36 @@ protected override void PostStop()
...
-            return message.Match()
-                .With<CounterChanged>(UpdateState)
-                .WasHandled;
+            switch (message)
+            {
+                case CounterChanged cc:
+                    UpdateState(cc);
+                    return true;
+            }
+            return false;
```
Danthar: @Horusiath Is this a valid replacement for the message.Match().With<> semantics ?

alexvaluyskiy: Yes

Danthar: ok

Danthar:
Mono build fails with 21 build errors. Relevant log starts around here: 
http://petabridge-ci.cloudapp.net/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=39173#_focus=1907

zbynek001:
I've checked the errors, it's related to a bug in Roslyn, which was fixed some time ago:
https://github.com/dotnet/roslyn/issues/16935
So maybe there is old sdk on build server or something similar? Not sure

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
sean-gilliam: why the switch statement on only one case?  it would be better expressed as an if statement.

Aaronontheweb:
@zbynek001 we just upgraded to the .NET 2.0 SDK today

zbynek001:
@Aaronontheweb I know, that's why I tried updating this. And seems it helped

PR #3047, opened 2017-08-29
Protobuf serialization support for FSM types

joshgarnett:
To bring the Akka.Persistence module more in line with the JVM Akka implementation protobuf serialization of PersistentFSMSnapshot and StateChangeEvent classes.  As PersistentFSMSnapshot uses generics, reflection is needed to properly serialize and deserialize the class.

### DIFF
--- a/src/core/Akka.Persistence/Fsm/PersistentFSM.cs
+++ b/src/core/Akka.Persistence/Fsm/PersistentFSM.cs
@@ -10,6 +10,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Akka.Actor;
 using Akka.Configuration;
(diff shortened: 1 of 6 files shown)


Aaronontheweb:
@joshgarnett missing API approvals

src/core/Akka.Persistence/Fsm/PersistentFSM.cs
```diff
@@ -173,7 +174,7 @@ public interface IFsmState
         /// Persisted on state change
         /// TODO: should implement IMessage
         /// </summary>
-        public class StateChangeEvent
+        public class StateChangeEvent : IMessage
```
Aaronontheweb: Looks good

Issue #3042, opened 2017-08-28
Add migration script for Akka.Persistence.Sql 1.2.3 --> 1.3.0 or revert schema changes.


Aaronontheweb:
If w're keeping the schema changes from 1.3.0 thus far, then we need to add a forward-migration script.

IMHO, should probably just drop the schema changes and just use the data from https://github.com/akkadotnet/akka.net/pull/3029 to do the work.

PR #3052, opened 2017-08-30
[WIP] Persistence v1.3.1 backwards compatible with v1.2.3

heynickc:
addresses #3042 and supersedes #3048.  the additional column `serializer_id` introduced in #3030 is not backwards compatible with v1.2.3 persistence schema (Sqlite used as an example here).  this is an attempt to remove the column but keep the functionality introduced by #3030

- spec added to read all events from a v1.2.3 sqlite db
- remove all uses of `serializer_id` column

TODO:

- the spec to validate backwards compatibility passes in isolation, but not when run inside the suite
- the `JournalSerializationSpec` and `SnapshotSerializationSpec` both fail now, I haven't addressed this yet - maybe due to the removal of the `serializer_id` column.  specs fail when serializer implements `SerializerWithStringManifest`


heynickc:
error thrown from `JournalSerializationSpec`

```
    Akka.Persistence.Sqlite.Tests.Serialization.SqliteJournalSerializationSpec.Journal_should_serialize_Persistent_with_string_manifest [FAIL]
      Failed: Expected a message of type Akka.Persistence.ReplayedMessage, but received {ReplayMessagesFailure<cause: System.TypeLoadException: Could not load type 'MyPayload-V2' from assembly 'Akka.Persistence.S
ql.Common, Version=1.3.0.0, Culture=neutral, PublicKeyToken=null'.
         at System.RuntimeTypeHandle.GetTypeByName(String name, Boolean throwOnError, Boolean ignoreCase, Boolean reflectionOnly, StackCrawlMarkHandle stackMark, IntPtr pPrivHostBinder, Boolean loadTypeFromPartia
lName, ObjectHandleOnStack type)
         at System.RuntimeTypeHandle.GetTypeByName(String name, Boolean throwOnError, Boolean ignoreCase, Boolean reflectionOnly, StackCrawlMark& stackMark, IntPtr pPrivHostBinder, Boolean loadTypeFromPartialName
)
         at System.RuntimeType.GetType(String typeName, Boolean throwOnError, Boolean ignoreCase, Boolean reflectionOnly, StackCrawlMark& stackMark)
         at System.Type.GetType(String typeName, Boolean throwOnError)
         at Akka.Persistence.Sql.Common.Journal.AbstractQueryExecutor.ReadEvent(DbDataReader reader)
         at Akka.Persistence.Sql.Common.Journal.AbstractQueryExecutor.<SelectByPersistenceIdAsync>d__44.MoveNext()
         at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
         at Akka.Persistence.Sql.Common.Journal.SqlJournal.<ReplayMessagesAsync>d__32.MoveNext()>} (type Akka.Persistence.ReplayMessagesFailure) instead  from [akka://SqliteJournalSerializationSpec/deadLetters]
      Expected: True
      Actual:   False
```

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -651,10 +644,7 @@ protected virtual void WriteEvent(DbCommand command, IPersistentRepresentation e
             }
             else
             {
-                if (serializer.IncludeManifest)
```
Danthar: Why did you remove this ?

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
```diff
@@ -350,13 +348,9 @@ protected virtual void SetManifestParameters(object snapshot, DbCommand command)
             }
             else
             {
-                if (serializer.IncludeManifest)
```
Danthar: Same here

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
heynickc: If this is set to false for any serializer, we can't `ReadEvent` because there's no type name in the `manifest` field and all the specs fail.  Not sure of another way around it. https://github.com/akkadotnet/akka.net/pull/3052/files/702218b1cb29718e86cb7d6c1547cc61217c900d#diff-cb3053db11018a282911ee665d86db6eR686

PR #3053, opened 2017-08-30
Updated UdpConnection to match TcpConnection send semantics

Horusiath:
This PR aims to provide the same send semantics over UDP connection, as the one used by TcpConnection, that were introduced by @Aaronontheweb in #3023


Aaronontheweb:
@Horusiath oh snap, I was in the middle of implementing this myself. I'll review yours instead

src/core/Akka/IO/UdpConnectedManager.cs
```diff
@@ -34,7 +34,7 @@ protected override bool Receive(object message)
             {
                 case UdpConnected.Connect connect:
                     {
-                        var commander = Sender;
+                        var commander = Sender; // NOTE: Aaronontheweb (9/1/2017) this should probably be the Handler...
```
Aaronontheweb: This seems like a bug to me; the `.Handler` property on the `Connect` message is never actually used... But changing it now would be a breaking change for anyone who has ever used this feature. Decided to just take note of it and punt on it until a later time in the interests of fixing sending behavior first.

Issue #3055, opened 2017-09-01
The type of the materialised value in the Flow returned by Bidiflow JoinMat() is incorrect


ianclegg:
The generic type parameters of the flow returned from the current implementation of Bidiflow's JoinMat method are incorrect.

The type of the materialised value in the new flow should reflect the return type of the user supplied `combine` functor. By using the wrong type, the current implementations forces the combine function to always return the same type as the a Keep.Left.

Current method signature:
```public Flow<TIn1, TOut2, TMat> JoinMat<TMat2, TMat3>(Flow<TOut1, TIn2, TMat2> flow, Func<TMat, TMat2, TMat3> combine)```

Correct method signature:
```public Flow<TIn1, TOut2, TMat3> JoinMat<TMat2, TMat3>(Flow<TOut1, TIn2, TMat2> flow, Func<TMat, TMat2, TMat3> combine)```

Some minor changes are required in the method to ensure the type is propagated.

Aaronontheweb:
closed by #3056

PR #3056, opened 2017-09-01
Fix the materialised value of the Flow returned by Bidiflow JoinMat() #3055

ianclegg:
The generic type parameters of the flow returned from the current implementation of Bidiflow's JoinMat method are incorrect as discussed in issue #3055.
This patch corrects the type, an equivalent PR is also being raised with Lightbend for the Akka Scala project for consistency.


Aaronontheweb:
@ianclegg need to approve public API changes to streams: http://getakka.net/community/public-api-changes.html

PR #3062, opened 2017-09-02
Update Xunit to 2.3.0-beta5

alexvaluyskiy:
https://xunit.github.io/releases/2.3-beta3
https://xunit.github.io/releases/2.3-beta4
https://xunit.github.io/releases/2.3-beta5


Aaronontheweb:
good catch @alexvaluyskiy

PR #3063, opened 2017-09-02
fix DNS resolution and add IPV6 support to Akka.IO.Tcp

Aaronontheweb:
* restored DNS support with specs to cover IPV6 and IPV4 scenarios
* added IPV6 support to TCP sockets on Akka.IO

Still working on UDP

### DIFF
--- a/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
+++ b/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
@@ -14,10 +14,14 @@ using System.Text;
 using Akka.Actor;
 using Akka.IO;
 using Akka.TestKit;
+using Akka.Util;
 using Akka.Util.Internal;
 using Xunit;
 using Xunit.Abstractions;
 using FluentAssertions;
+#if CORECLR
+using System.Runtime.InteropServices;
+#endif
 
 namespace Akka.Tests.IO
 {
@@ -150,6 +154,15 @@ namespace Akka.Tests.IO
         [Theory]
         public void The_TCP_transport_implementation_should_properly_support_connecting_to_DNS_endpoints(AddressFamily family)
         {
+            // Aaronontheweb, 9/2/2017 - POSIX-based OSES are still having trouble with IPV6 DNS resolution
+#if CORECLR
+            if(!System.Runtime.InteropServices.RuntimeInformation
+                .IsOSPlatform(OSPlatform.Windows))
+            return;
+#else
+            if (RuntimeDetector.IsMono && family == AddressFamily.InterNetworkV6) // same as above
+                return;
+#endif
             var serverHandler = CreateTestProbe();
             var bindCommander = CreateTestProbe();
             bindCommander.Send(Sys.Tcp(), new Tcp.Bind(serverHandler.Ref, new IPEndPoint(family == AddressFamily.InterNetwork ? IPAddress.Loopback 
--- a/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
+++ b/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
@@ -145,6 +145,39 @@ namespace Akka.Tests.IO
             });
         }
 
+        [InlineData(AddressFamily.InterNetworkV6)]
+        [InlineData(AddressFamily.InterNetwork)]
+        [Theory]
+        public void The_TCP_transport_implementation_should_properly_support_connecting_to_DNS_endpoints(AddressFamily family)
+        {
+            var serverHandler = CreateTestProbe();
+            var bindCommander = CreateTestProbe();
+            bindCommander.Send(Sys.Tcp(), new Tcp.Bind(serverHandler.Ref, new IPEndPoint(family == AddressFamily.InterNetwork ? IPAddress.Loopback 
+                : IPAddress.IPv6Loopback, 0)));
+            var boundMsg = bindCommander.ExpectMsg<Tcp.Bound>();
+
+            // setup client to connect 
+            var targetAddress = new DnsEndPoint("localhost", boundMsg.LocalAddress.AsInstanceOf<IPEndPoint>().Port);
+            var clientHandler = CreateTestProbe();
+            Sys.Tcp().Tell(new Tcp.Connect(targetAddress), clientHandler);
+            clientHandler.ExpectMsg<Tcp.Connected>(TimeSpan.FromMinutes(10));
+            var clientEp = clientHandler.Sender;
+            clientEp.Tell(new Tcp.Register(clientHandler));
+            serverHandler.ExpectMsg<Tcp.Connected>();
+            serverHandler.Sender.Tell(new Tcp.Register(serverHandler));
+
+            var str = Enumerable.Repeat("f", 567).Join("");
+            var testData = ByteString.FromString(str);
+            clientEp.Tell(Tcp.Write.Create(testData, Ack.Instance), clientHandler);
+            clientHandler.ExpectMsg<Ack>();
+            var received = serverHandler.ReceiveWhile<Tcp.Received>(o =>
+            {
+                return o as Tcp.Received;
+            }, RemainingOrDefault, TimeSpan.FromSeconds(0.5));
+
+            received.Sum(s => s.Data.Count).Should().Be(testData.Count);
+        }
+
         [Fact]
         public void BugFix_3021_Tcp_Should_not_drop_large_messages()
         {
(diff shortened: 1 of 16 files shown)


Aaronontheweb:
Sitting there thinking about it... UDP isn't going to be able to take advantage of the same design I added for TCP here with falling back between IPV4 / IPV6 upon resolution. We'd need to get an explicit connection failure notification back, which we won't under UDP.

Aaronontheweb:
Had a spec failure with IPV6 on Linux .NET Core... Re-running to see if it's a timeout or something off with DNS on Linux .NET Core...

Aaronontheweb:
Have a failure with IPV6 support on Mono on Linux now too... Looks like this is something at a lower level on the stack than the .NET runtime being used. Might be an issue with the version of Ubuntu we're using not supporting IPV6 properly on DNS (it's come up in the past.)

Going to mute those specs on our Linux distributions.

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -145,6 +149,48 @@ public void The_TCP_transport_implementation_should_properly_handle_connection_a
...
         }
 
+        [InlineData(AddressFamily.InterNetworkV6)]
+        [InlineData(AddressFamily.InterNetwork)]
+        [Theory]
+        public void The_TCP_transport_implementation_should_properly_support_connecting_to_DNS_endpoints(AddressFamily family)
+        {
+            // Aaronontheweb, 9/2/2017 - POSIX-based OSES are still having trouble with IPV6 DNS resolution
+#if CORECLR
+            if(!System.Runtime.InteropServices.RuntimeInformation
```
Aaronontheweb: Check for POSIX operating systems on .NET Core

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -145,6 +149,48 @@ public void The_TCP_transport_implementation_should_properly_handle_connection_a
...
+                .IsOSPlatform(OSPlatform.Windows))
+            return;
+#else
+            if (RuntimeDetector.IsMono && family == AddressFamily.InterNetworkV6) // same as above
```
Aaronontheweb: Check for Mono and bail.

src/core/Akka/IO/TcpOutgoingConnection.cs
```diff
@@ -57,7 +58,7 @@ private void ReportConnectFailure(Action thunk)
             }
             catch (Exception e)
             {
-                Log.Debug("Could not establish connection to [{0}] due to {1}", _connect.RemoteAddress, e);
+                Log.Error(e, "Could not establish connection to [{0}].", _connect.RemoteAddress);
```
Aaronontheweb: Log connection failures as errors.

src/core/Akka/IO/TcpOutgoingConnection.cs
```diff
@@ -24,7 +25,7 @@ internal sealed class TcpOutgoingConnection : TcpConnection
         private readonly Tcp.Connect _connect;
         
         public TcpOutgoingConnection(TcpExt tcp, IActorRef commander, Tcp.Connect connect)
-            : base(tcp, new Socket(connect.RemoteAddress.AddressFamily, SocketType.Stream, ProtocolType.Tcp) { Blocking = false }, connect.PullMode)
```
Aaronontheweb: This was the change that caused the most issues with DNS. Passing in a `DnsEndpoint` with an unspecified `AddressFamily` caused DNS to not work at all, which is how most people use it.  We never specified the `AddressFamily` in Akka.IO prior to 1.3.0 except on Mono to work around an issue where binding to an IPV6 socket was unsupported. That issue appears to be fixed in the more recent versions of Mono, although as I noted in this PR there appears to be issues at the OS level with DNS resolution and IPV6 still.

src/core/Akka/IO/TcpOutgoingConnection.cs
```diff
@@ -73,12 +74,14 @@ protected override void PreStart()
                     var resolved = Dns.ResolveName(remoteAddress.Host, Context.System, Self);
                     if (resolved == null)
                         Become(Resolving(remoteAddress));
-                    else
-                        Register(new IPEndPoint(resolved.Addr, remoteAddress.Port));
+                    else if(resolved.Ipv4.Any() && resolved.Ipv6.Any()) // one of both families
+                        Register(new IPEndPoint(resolved.Ipv4.FirstOrDefault(), remoteAddress.Port), new IPEndPoint(resolved.Ipv6.FirstOrDefault(), remoteAddress.Port));
```
Aaronontheweb: Added a fallback scheme here that only gets invoked when using DNS endpoints. If using DNS, first attempt to connect to a socket is done with IPV4 (default from earlier versions of Akka.IO)... If that fails, second attempt will swap and use IPV6. Third attempt will swap back to IPV4. Fourth will swap back to IPV6... Etc... Idea is to provide some help given the ambiguity of DNS resolution and the address family being used.

As I noted here https://github.com/akkadotnet/akka.net/issues/3064, this could be improved to use the explicit `AddressFamily` provided on the `DnsEndPoint` when it's available. We could also extend this in the future to support falling back through multiple resolved addresses until the right one is found. But I decided to leave well-enough alone for now.

src/core/Akka/IO/UdpConnection.cs
```diff
@@ -228,7 +227,7 @@ private void ReportConnectFailure(Action thunk)
             }
             catch (Exception e)
             {
-                Log.Debug("Failure while connecting UDP channel to remote address [{0}] local address [{1}]: {2}", _connect.RemoteAddress, _connect.LocalAddress, e);
+                Log.Error(e, "Failure while connecting UDP channel to remote address [{0}] local address [{1}]", _connect.RemoteAddress, _connect.LocalAddress);
```
Aaronontheweb: Log connection failures as errors.

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -145,6 +149,48 @@ public void The_TCP_transport_implementation_should_properly_handle_connection_a
...
+            if(!System.Runtime.InteropServices.RuntimeInformation
+                .IsOSPlatform(OSPlatform.Windows) && family == AddressFamily.InterNetworkV6)
+            return;
+#else
+            if (RuntimeDetector.IsMono && family == AddressFamily.InterNetworkV6) // same as above
+                return;
+#endif
+            var serverHandler = CreateTestProbe();
+            var bindCommander = CreateTestProbe();
+            bindCommander.Send(Sys.Tcp(), new Tcp.Bind(serverHandler.Ref, new IPEndPoint(family == AddressFamily.InterNetwork ? IPAddress.Loopback
```
Aaronontheweb: Bind to an IP address which can be resolved via local DNS.

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -145,6 +149,48 @@ public void The_TCP_transport_implementation_should_properly_handle_connection_a
...
+                return;
+#endif
+            var serverHandler = CreateTestProbe();
+            var bindCommander = CreateTestProbe();
+            bindCommander.Send(Sys.Tcp(), new Tcp.Bind(serverHandler.Ref, new IPEndPoint(family == AddressFamily.InterNetwork ? IPAddress.Loopback 
+                : IPAddress.IPv6Loopback, 0)));
+            var boundMsg = bindCommander.ExpectMsg<Tcp.Bound>();
+
+            // setup client to connect 
+            var targetAddress = new DnsEndPoint("localhost", boundMsg.LocalAddress.AsInstanceOf<IPEndPoint>().Port);
```
Aaronontheweb: Connect via DNS.

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -145,6 +149,48 @@ public void The_TCP_transport_implementation_should_properly_handle_connection_a
...
+            var clientEp = clientHandler.Sender;
+            clientEp.Tell(new Tcp.Register(clientHandler));
+            serverHandler.ExpectMsg<Tcp.Connected>();
+            serverHandler.Sender.Tell(new Tcp.Register(serverHandler));
+
+            var str = Enumerable.Repeat("f", 567).Join("");
+            var testData = ByteString.FromString(str);
+            clientEp.Tell(Tcp.Write.Create(testData, Ack.Instance), clientHandler);
+            clientHandler.ExpectMsg<Ack>();
+            var received = serverHandler.ReceiveWhile<Tcp.Received>(o =>
```
Aaronontheweb: received data from connected DNS endpoint following resolution.

Horusiath:
Is this still work in progress @Aaronontheweb or is it ready to merge?

Danthar:
I think it was ready. Additionally this PR added some behavior that is worth mentioning in the docs.
On that note, the akka.io docs could really use some love imho.

Aaronontheweb:
@Horusiath this is all finished

@Danthar I agree that the docs need some love for Akka.IO.... IMHO most important thing to add is the workflow for how you set up a connection in the first place.

Issue #3067, opened 2017-09-04
PersistenceMessageSerializer fails with wrong exception if non-supported type provided


alexvaluyskiy:
If I send a string into `PersistenceMessageSerializer` I except that I get an error `ArgumentException`, but it fails on this line instead:

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs#L35

Danthar:
Plugins are expected to use the `IPersistentRepresentation`

alexvaluyskiy:
Serializer expects `IMessage`, like `IPersistenceRepresentation` or `AtLeastOnceDeliverySnapshot`. But if we provide some different type, it should be handled properly. We should not call `GetGenericTypeDefinition()` on non-generic types

alexvaluyskiy:
We should change it to
```
if (obj.GetType().IsGenericType && obj.GetType().GetGenericTypeDefinition() == typeof(PersistentFSM.PersistentFSMSnapshot<>))
{
}
```

PR #3069, opened 2017-09-05
Updated PersistenceMessageSerializer

alexvaluyskiy:
Fix https://github.com/akkadotnet/akka.net/issues/3067

### DIFF
--- a/src/core/Akka.Persistence.Tests/Serialization/MessageSerializerRemotingSpec.cs
+++ b/src/core/Akka.Persistence.Tests/Serialization/MessageSerializerRemotingSpec.cs
@@ -11,7 +11,6 @@ using System.Collections.Immutable;
 using System.Linq;
 using Akka.Actor;
 using Akka.Configuration;
-using Akka.Persistence.Fsm;
 using Akka.TestKit;
 using FluentAssertions;
 using Xunit;
@@ -201,45 +200,6 @@ akka {
                 ExpectMsg("p.a.");
                 ExpectMsg("p.b.");
             });
-        }
-
-        [Fact]
-        public void MessageSerializer_should_serialize_manifest_provided_by_EventAdapter()
-        {
-            var p1 = new Persistent(new MyPayload("a"), sender: TestActor).WithManifest("manifest");
-            var serializer = _serialization.FindSerializerFor(p1);
-            var bytes = serializer.ToBinary(p1);
-            var back = (Persistent)serializer.FromBinary(bytes, typeof (Persistent));
-
-            back.Manifest.ShouldBe(p1.Manifest);
-        }
-
-        [Fact]
-        public void MessageSerializer_should_serialize_state_change_event()
-        {
-            var p1 = new Persistent(new PersistentFSM.StateChangeEvent("a", TimeSpan.FromSeconds(10)), sender: TestActor);
-            var serializer = _serialization.FindSerializerFor(p1);
-            var bytes = serializer.ToBinary(p1);
-            var back = (Persistent)serializer.FromBinary(bytes, typeof (Persistent));
-            var payload = back.Payload as PersistentFSM.StateChangeEvent;
-            payload.ShouldNotBe(null);
-            payload.StateIdentifier.ShouldBe("a");
-            payload.Timeout.ShouldBeEquivalentTo(TimeSpan.FromSeconds(10));
-        }
-
-        [Fact]
-        public void MessageSerializer_should_serialize_fsm_snapshot()
-        {
-            var snapshot = new PersistentFSM.PersistentFSMSnapshot<MyPayload>("a", new MyPayload("b"), TimeSpan.FromSeconds(10));
-            var p1 = new Persistent(snapshot, sender: TestActor);
-            var serializer = _serialization.FindSerializerFor(p1);
-            var bytes = serializer.ToBinary(p1);
-            var back = (Persistent)serializer.FromBinary(bytes, typeof (Persistent));
-            var backSnapshot = back.Payload as PersistentFSM.PersistentFSMSnapshot<MyPayload>;
-            backSnapshot.ShouldNotBe(null);
-            backSnapshot.StateIdentifier.ShouldBe("a");
-            backSnapshot.Data.Data.ShouldBe(".b.");    // custom MyPayload serializer prepends and appends .
-            backSnapshot.Timeout.ShouldBeEquivalentTo(TimeSpan.FromSeconds(10));
         }
     }
 }
--- a/src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
+++ b/src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
@@ -9,30 +9,32 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Reflection;
+using System.Runtime.Serialization;
 using Akka.Actor;
 using Akka.Persistence.Fsm;
 using Akka.Persistence.Serialization.Proto.Msg;
 using Akka.Serialization;
+using Akka.Util;
 using Google.Protobuf;
 
 namespace Akka.Persistence.Serialization
 {
-    public class PersistenceMessageSerializer : Serializer
+    public sealed class PersistenceMessageSerializer : Serializer
     {
         public PersistenceMessageSerializer(ExtendedActorSystem system) : base(system)
         {
-            IncludeManifest = true;
         }
 
-        public override bool IncludeManifest { get; }
+        public override bool IncludeManifest { get; } = true;
 
         public override byte[] ToBinary(object obj)
         {
-            if (obj is IPersistentRepresentation) return GetPersistentMessage((IPersistentRepresentation)obj).ToByteArray();
-            if (obj is AtomicWrite) return GetAtomicWrite((AtomicWrite)obj).ToByteArray();
-            if (obj is AtLeastOnceDeliverySnapshot) return GetAtLeastOnceDeliverySnapshot((AtLeastOnceDeliverySnapshot)obj).ToByteArray();
-            if (obj is PersistentFSM.StateChangeEvent) return GetStateChangeEvent((PersistentFSM.StateChangeEvent)obj).ToByteArray();
-            if (obj.GetType().GetGenericTypeDefinition() == typeof(PersistentFSM.PersistentFSMSnapshot<>)) return GetPersistentFSMSnapshot(obj).ToByteArray();
+            if (obj is IPersistentRepresentation repr) return GetPersistentMessage(repr).ToByteArray();
+            if (obj is AtomicWrite aw) return GetAtomicWrite(aw).ToByteArray();
+            if (obj is AtLeastOnceDeliverySnapshot snap) return GetAtLeastOnceDeliverySnapshot(snap).ToByteArray();
+            if (obj is PersistentFSM.StateChangeEvent stateEvent) return GetStateChangeEvent(stateEvent).ToByteArray();
+            if (obj.GetType().GetTypeInfo().IsGenericType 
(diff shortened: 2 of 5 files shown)


src/core/Akka.Persistence.Tests/Serialization/MessageSerializerRemotingSpec.cs
```diff
@@ -202,44 +201,5 @@ public void MessageSerializer_should_custom_serialize_AtomicWrite_messages_durin
...
-            var back = (Persistent)serializer.FromBinary(bytes, typeof (Persistent));
-
-            back.Manifest.ShouldBe(p1.Manifest);
-        }
-
-        [Fact]
-        public void MessageSerializer_should_serialize_state_change_event()
-        {
-            var p1 = new Persistent(new PersistentFSM.StateChangeEvent("a", TimeSpan.FromSeconds(10)), sender: TestActor);
-            var serializer = _serialization.FindSerializerFor(p1);
```
alexvaluyskiy: We should not test all our serialization infrastructure here, just a concrete serializer.

src/core/Akka.Persistence.Tests/Serialization/MessageSerializerRemotingSpec.cs
```diff
@@ -202,44 +201,5 @@ public void MessageSerializer_should_custom_serialize_AtomicWrite_messages_durin
...
-            payload.ShouldNotBe(null);
-            payload.StateIdentifier.ShouldBe("a");
-            payload.Timeout.ShouldBeEquivalentTo(TimeSpan.FromSeconds(10));
-        }
-
-        [Fact]
-        public void MessageSerializer_should_serialize_fsm_snapshot()
-        {
-            var snapshot = new PersistentFSM.PersistentFSMSnapshot<MyPayload>("a", new MyPayload("b"), TimeSpan.FromSeconds(10));
-            var p1 = new Persistent(snapshot, sender: TestActor);
```
alexvaluyskiy: PersistentFSMSnapshot should not be put into `Persistent` object

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
```diff
@@ -9,30 +9,32 @@
...
-            if (obj is IPersistentRepresentation) return GetPersistentMessage((IPersistentRepresentation)obj).ToByteArray();
-            if (obj is AtomicWrite) return GetAtomicWrite((AtomicWrite)obj).ToByteArray();
-            if (obj is AtLeastOnceDeliverySnapshot) return GetAtLeastOnceDeliverySnapshot((AtLeastOnceDeliverySnapshot)obj).ToByteArray();
-            if (obj is PersistentFSM.StateChangeEvent) return GetStateChangeEvent((PersistentFSM.StateChangeEvent)obj).ToByteArray();
-            if (obj.GetType().GetGenericTypeDefinition() == typeof(PersistentFSM.PersistentFSMSnapshot<>)) return GetPersistentFSMSnapshot(obj).ToByteArray();
+            if (obj is IPersistentRepresentation repr) return GetPersistentMessage(repr).ToByteArray();
+            if (obj is AtomicWrite aw) return GetAtomicWrite(aw).ToByteArray();
+            if (obj is AtLeastOnceDeliverySnapshot snap) return GetAtLeastOnceDeliverySnapshot(snap).ToByteArray();
+            if (obj is PersistentFSM.StateChangeEvent stateEvent) return GetStateChangeEvent(stateEvent).ToByteArray();
+            if (obj.GetType().GetTypeInfo().IsGenericType
```
alexvaluyskiy: `obj.GetType().GetTypeInfo().IsGenericType` - it prevents us from unhandled exceptions

Danthar: It will still cause an exception down the line. Namely the ArgumentException. But that will be for more descriptive.

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
```diff
@@ -56,19 +58,18 @@ private PersistentMessage GetPersistentMessage(IPersistentRepresentation persist
...
+                string manifest = serializer2.Manifest(obj);
                 payload.PayloadManifest = ByteString.CopyFromUtf8(manifest);
             }
             else
             {
                 if (serializer.IncludeManifest)
                 {
-                    var payloadType = obj.GetType();
-                    payload.PayloadManifest = ByteString.CopyFromUtf8(payloadType.AssemblyQualifiedName);
+                    payload.PayloadManifest = ByteString.CopyFromUtf8(obj.GetType().TypeQualifiedName());
```
Danthar: Type qualified name is without the assemblyname right ?

src/core/Akka.Persistence.Tests/Serialization/MessageSerializerRemotingSpec.cs
Danthar: You moved this and the test above to a seperate spec file. But the tests remain the same. So what do you mean ?

src/core/Akka.Persistence.Tests/Serialization/MessageSerializerRemotingSpec.cs
Danthar: Correct. There is a special case scenario in the `PersistenceMessageSerializer` to handle it. Putting it in the Persistent envelope is wrong.

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
```diff
@@ -9,30 +9,32 @@
...
-            if (obj is PersistentFSM.StateChangeEvent) return GetStateChangeEvent((PersistentFSM.StateChangeEvent)obj).ToByteArray();
-            if (obj.GetType().GetGenericTypeDefinition() == typeof(PersistentFSM.PersistentFSMSnapshot<>)) return GetPersistentFSMSnapshot(obj).ToByteArray();
+            if (obj is IPersistentRepresentation repr) return GetPersistentMessage(repr).ToByteArray();
+            if (obj is AtomicWrite aw) return GetAtomicWrite(aw).ToByteArray();
+            if (obj is AtLeastOnceDeliverySnapshot snap) return GetAtLeastOnceDeliverySnapshot(snap).ToByteArray();
+            if (obj is PersistentFSM.StateChangeEvent stateEvent) return GetStateChangeEvent(stateEvent).ToByteArray();
+            if (obj.GetType().GetTypeInfo().IsGenericType 
+                && obj.GetType().GetGenericTypeDefinition() == typeof(PersistentFSM.PersistentFSMSnapshot<>)) return GetPersistentFSMSnapshot(obj).ToByteArray();
 
             throw new ArgumentException($"Can't serialize object of type [{obj.GetType()}] in [{GetType()}]");
```
joshgarnett: Should this also be updated to SerializationException?

Danthar: good point

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
joshgarnett: It looks like this is an extension method in the Akka.Util package, which will attempt to serialize without the assembly name if it is part of the core assembly.  I'm a little bit nervous on how well this works across runtime versions and if there are going to be an issues during upgrades.  The change can be found at https://github.com/akkadotnet/akka.net/commit/d3328cab47e471783f8c7a4b3379188a5c9ac02a

joshgarnett: Yeah, so I just did a quick comparison.  A mono 4.8 app has a core assembly name of mscorlib, while a .net core standard app has a core assembly name of System.Private.CoreLib.  This could cause issues with people that are trying to switch runtimes.  I'll create an issue.

heynickc: @joshgarnett is this an issue with this specific change, or also with the previous code?

src/core/Akka.Persistence.Tests/Serialization/MessageSerializerRemotingSpec.cs
alexvaluyskiy: I stopped using `FindSerializerFor` and added a test for AtLeastOnceDeliverySnapshot
We have integrations tests for all standard persistence messages in Persistence.TCK

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
alexvaluyskiy: `ArgumentException` is fine, by design

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
joshgarnett: Actually I take that back, in theory the default should always deserialize properly for types with no assembly included?

alexvaluyskiy: @joshgarnett Have you found some bugs? Because I've covered this scenario https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Util/TypeExtensions.cs#L63

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
alexvaluyskiy: I sync it with JVM version. They have `NonSerializableException` in `FromBinary` and they don't treat this exception as a fatal

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
heynickc: Just so I have it straight - the issue here is when someone previously was serializing messages with a mono 4.8 app, then switch to a .net core app, they wouldn't be able to deserialize the messages based on the type specified in the manifest, because it will not be `System.String, System.Private.CoreLib`, but instead be `System.String, mscorlib`

heynickc: @joshgarnett, also related to this block of code - what if the serializer is not set to `serializer.IncludeManifest = true`?  How would we be able to `ReadEvent`?

alexvaluyskiy: @heynickc Anyway, we have used the similar method in 1.2.x

joshgarnett: Yeah, I think we are fine, I retracted my comment right after yours @heynickc I was concerned at first, but all this is doing is grouping together objects that would be under the core assembly, which should be the same set of objects regardless of the actual name of the core assembly.

heynickc: @alexvaluyskiy ok sorry, I was in the wrong vacinity.  this is where I was curious about such an issue: https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs#L656

joshgarnett: @heynickc as for IncludeManifest, that is how serializers should work, right?  If they do not set IncludeManifest to true, then the assumption is that the serializer does not need the manifest in order to deserialize the data properly.

heynickc: @joshgarnett gotcha, thanks.  I ran into the issue testing backwards compatibility, but I may have been testing an edge case that wasn't possible.

PR #3085, opened 2017-09-07
Improved Streams Error-Handling docs

alexvaluyskiy:
Scala JVM has `RestartSource`, `RestartSink` and `RestartFlow`, which is not implemented in Akka.NET at the moment
Fixed https://github.com/akkadotnet/akka.net/issues/3086

### DIFF
--- a/docs/articles/streams/error-handling.md
+++ b/docs/articles/streams/error-handling.md
@@ -1,31 +1,108 @@
 ---
-layout: docs.hbs
-title: Error Handling
+uid: error-handling
+title: Error Handling in Streams
 ---
+# Error Handling in Streams
 
-Strategies for how to handle exceptions from processing stream elements can be defined when
-materializing the stream. The error handling strategies are inspired by actor supervision
-strategies, but the semantics have been adapted to the domain of stream processing.
+When a stage in a stream fails this will normally lead to the entire stream being torn down. Each of the stages downstream gets informed about the failure and each upstream stage sees a cancellation.
 
-> [!WARNING]
-> *ZipWith*, *GraphStage* junction, *ActorPublisher* source and *ActorSubscriber* sink 
-components do not honour the supervision strategy attribute yet.
+In many cases you may want to avoid complete stream failure, this can be done in a few different ways:
 
-# Supervision Strategies
+- `Recover` to emit a final element then complete the stream normally on upstream failure
+- `RecoverWithRetries` to create a new upstream and start consuming from that on failure
+- Using a supervision strategy for stages that support it
+
+In addition to these built in tools for error handling, a common pattern is to wrap the stream inside an actor, and have the actor restart the entire stream on failure.
+
+## Recover
+`Recover` allows you to emit a final element and then complete the stream on an upstream failure. Deciding which exceptions should be recovered is done through a `delegate`. If an exception does not have a matching case the stream is failed.
+
+Recovering can be useful if you want to gracefully complete a stream on failure while letting downstream know that there was a failure.
+
+```C#
+Source.From(Enumerable.Range(0, 6)).Select(n =>
+    {
+        if (n < 5)
+            return n.ToString();
+
+        throw new ArithmeticException("Boom!");
+    })
+    .Recover(exception =>
+    {
+        if (exception is ArithmeticException)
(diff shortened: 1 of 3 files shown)


docs/articles/streams/error-handling.md
```diff
@@ -1,78 +1,155 @@
...
+        if (n < 5)
+            return n.ToString();
+
+        throw new ArithmeticException("Boom!");
+    })
+    .Recover(exception =>
+    {
+        if (exception is ArithmeticException)
+            return new Option<string>("stream truncated");
+        return Option<string>.None;
```
alexvaluyskiy: @Silv3rcircl3 I wondered that we have such API. Why not just string?

Horusiath: `Option<>` is like `Nullable<>`, but it can also work with non-value types. We needed a universal primitive that will notify about value absence and will work with both (classes and structs).

docs/articles/streams/error-handling.md
```diff
@@ -46,33 +123,33 @@ var result = source.RunWith(Sink.Aggregate<int, int>(0, (sum, i) => sum + i), ma
 // result here will be a Task completed with Success(228)
 ```
 
-Here you can see that all ``DivideByZeroException`` will resume the processing, i.e. the
+Here you can see that all `DivideByZeroException` will resume the processing, i.e. the
 elements that cause the division by zero are effectively dropped.
 
 > [!NOTE]
-> Be aware that dropping elements may result in deadlocks in graphs with cycles, as explained in [Graph cycles, liveness and deadlocks](workingwithgraphs#graph-cycles-liveness-and-deadlocks).
+> Be aware that dropping elements may result in deadlocks in graphs with cycles, as explained in [Graph cycles, liveness and deadlocks](xref:working-with-graphs#graph-cycles-liveness-and-deadlocks).
```
alexvaluyskiy: fixed the link

docs/articles/streams/integration.md
```diff
@@ -518,7 +518,7 @@ Please note that a factory is necessary to achieve reusability of the resulting
 
 As described above any Akka Streams ``Source`` can be exposed as a Reactive Streams ``Publisher`` 
 and any ``Sink`` can be exposed as a Reactive Streams ``Subscriber``. Therefore we recommend that you 
-implement Reactive Streams integrations with built-in stages or [custom stages](customstreamprocessing.md).
+implement Reactive Streams integrations with built-in stages or [custom stages](xref:custom-stream-processing).
```
alexvaluyskiy: Fixed the link

docs/articles/streams/workingwithgraphs.md
```diff
@@ -509,7 +509,7 @@ public static IMessage FromBytes(ByteString bytes)
 
 In this way you could easily integrate any other serialization library that turns an object into a sequence of bytes.
 
-The other stage that we talked about is a little more involved since reversing a framing protocol means that any received chunk of bytes may correspond to zero or more messages. This is best implemented using a `GraphStage` (see also [Custom processing with GraphStage](customstreamprocessing.md#custom-processing-with-graphstage)).
+The other stage that we talked about is a little more involved since reversing a framing protocol means that any received chunk of bytes may correspond to zero or more messages. This is best implemented using a `GraphStage` (see also [Custom processing with GraphStage](xref:custom-stream-processing#custom-processing-with-graphstage)).
```
alexvaluyskiy: fixed the link

docs/articles/streams/error-handling.md
Silv3rcircl3: Also note that NULL is not allowed as element, so return null would fail the stream.

alexvaluyskiy: @Silv3rcircl3 Is null is accepted in `RecoveryWithRetries`?

Silv3rcircl3: Yup, because there you return a complete new source and null indicates that you don't want to recover but rather fail the stream. By element I mean a single message that is processed by the stream, i.e. Select(x => null as string) would fail the stream

alexvaluyskiy: Ok, thanks

Issue #3087, opened 2017-09-07
ClusterShardingMessageSerializer Exception after upgrade from 1.2.0 to 1.3.1 


amichel:
After upgrading my project with sharded cluster from 1.2.0 to 1.3.1 and switching serializer from wire to Hyperion, I get this exception at ClusterSharding.Get(system) :

**> Attempt by method 'Akka.Cluster.Sharding.Serialization.ClusterShardingMessageSerializer..ctor(Akka.Actor.ExtendedActorSystem)' to access method 'Akka.Serialization.SerializerIdentifierHelper.GetSerializerIdentifierFromConfig(System.Type, Akka.Actor.ExtendedActorSystem)' failed.**

My config looks like:
provider = "Akka.Cluster.ClusterActorRefProvider, Akka.Cluster"
serializers {
wire = "Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion"
}
serialization-bindings {
"System.Object" = wire
}
no special entries or custom fallbacks for sharding

amichel:
When I reference assemblies built from dev branch it doesn't happen. When I switch back to nuget I get same exception again.

Aaronontheweb:
This is a confirmed issue with the way we've using the `SerializerId` for the dedicated Cluster.Sharding serializer.... We'll see about getting it fixed in 1.3.2

Aaronontheweb:
@amichel this should be fixed in tonight's nightly build; would you mind giving that a try and let us know if it resolves your issue? http://getakka.net/community/getting-access-to-nightly-builds.html

Aaronontheweb:
Nightlies run at midnight UTC usually, so it should be ready for you guys by tomorrow AM.

amichel:
Great. Will give it a try in the morning. Actually like I mentioned, it
did't happen to me when I referenced debug mode assemblies from dev branch,
only with nuget packages.
So I used this as kind of workaround to avoid downgrading solution, but of
course I can't checkin like that.

amichel:
I can confirm that this bug was solved when using nightly build (beta426).

I have other issues now preventing cluster from starting related to SQL persistence changes. Will try to dig more info on them.

Aaronontheweb:
@amichel are you using the latest Akka.Persistence.Sql driver? We just pushed out an update for those packages last week

amichel:
I updated all packages from nightly build. Will try test with empty db to
isolate replaying of events.

PR #3094, opened 2017-09-12
Fixed Sharding serialization bug

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/issues/3087
Also converted Sharding example to .NET Core 1.1


src/core/Akka/Serialization/Serializer.cs
```diff
@@ -149,12 +149,9 @@ public sealed override object FromBinary(byte[] bytes, Type type)
     /// <summary>
     /// INTERNAL API.
     /// </summary>
-    internal static class SerializerIdentifierHelper
+    public static class SerializerIdentifierHelper
```
alexvaluyskiy: I made this class as internal by accident. It should be public

Aaronontheweb:
@alexvaluyskiy I don't get it, where is the actual fix for the issue in this PR?

PR #2941, opened 2017-08-08
WIP Add Akka.FSharp to v1.3 build

heynickc:
Builds Akka.FSharp and Akka.FSharp.Tests using the build script, but does not add it to the .sln because VS 2017 15.2 doesn't support the new format of MSBuild 15 projects for .fsproj per this issue: https://github.com/Microsoft/visualfsharp/issues/2316

**Note**: 2 specs fail and I'm having difficulty grokking why

```
    Akka.FSharp.Tests.InfrastructureTests.IActorRef should be possible to use as a Key [FAIL]
      System.ArgumentException : Expression of type 'Microsoft.FSharp.Core.FSharpFunc`2[Microsoft.FSharp.Core.Unit,Akka.FSharp.Actors+FunActor`2[System.Object,System.Object]]' cannot be used for return type 'Akka.
FSharp.Actors+FunActor`2[System.Object,System.Object]'
      Stack Trace:
           at System.Linq.Expressions.Expression.ValidateLambdaArgs(Type delegateType, Expression& body, ReadOnlyCollection`1 parameters)
           at System.Linq.Expressions.Expression.Lambda(Type delegateType, Expression body, String name, Boolean tailCall, IEnumerable`1 parameters)
           at System.Linq.Expressions.Expression.Lambda(Type delegateType, Expression body, IEnumerable`1 parameters)
           at Microsoft.FSharp.Linq.RuntimeHelpers.LeafExpressionConverter.ConvExprToLinqInContext(ConvEnv env, FSharpExpr inp)
           at Microsoft.FSharp.Linq.RuntimeHelpers.LeafExpressionConverter.QuotationToLambdaExpression[T](FSharpExpr`1 e)
           at Akka.FSharp.Spawn.spawnOpt[Message,Returned](IActorRefFactory actorFactory, String name, FSharpFunc`2 f, FSharpList`1 options)
           at Akka.FSharp.Tests.InfrastructureTests.IActorRef should be possible to use as a Key()
    Akka.FSharp.Tests.ApiTests.actor that accepts _ will receive string message [FAIL]
      System.ArgumentException : Expression of type 'Microsoft.FSharp.Core.FSharpFunc`2[Microsoft.FSharp.Core.Unit,Akka.FSharp.Actors+FunActor`2[System.Object,System.Object]]' cannot be used for return type 'Akka.
FSharp.Actors+FunActor`2[System.Object,System.Object]'
      Stack Trace:
[...]

PR #3095, opened 2017-09-12
Fix FSharp projects. Part One

alexvaluyskiy:
It is a replacement for https://github.com/akkadotnet/akka.net/pull/2941 but based on .NET Core SDK 2.0.
It still has two failed tests. If we merge it, then @Horusiath could start work on it


alexvaluyskiy:
Ok. Mono 5.0 can't build the solution with the new FSharp project. I should try 5.2 or 5.4

PR #3099, opened 2017-09-13
Member status: Weakly up

Horusiath:


Horusiath:
One note: if we're going to release v1.3.2, don't merge it yet. It's too serious to land in patch release, it's meant to for v1.4.

Aaronontheweb:
1.4 aint happen for a while. Need a feature branch then.

Horusiath:
Got some Hyperion serialization exceptions of MNTK for windows .NET core - something related to cluster sharding.

Aaronontheweb:
Yeah, this is a breakage in the .NET Core Tooling - nothing to do with your PR.

Horusiath:
I've turned this feature off by default - this way it won't affect current cluster protocol, and it still can be set manually. I think it could be merged directly.

Aaronontheweb:
I need to check:

* [ ] How `WeaklyMemberUp` factors into leader election
* [ ] How `WeaklyMemberUp` impacts the bug fixes introduced in solving #2584

Aaronontheweb:
Looks like we're in the clear on the node leader transition stuff, since you added `WeaklyUp` to the `AllowedTransitions` table, which is what our leader picking algorithm uses to determine what's the correct "next state" for each node in transition.

Aaronontheweb:
Doesn't affect leader elections either:

```csharp
   private static readonly ImmutableHashSet<MemberStatus> LeaderMemberStatus =
            ImmutableHashSet.Create(MemberStatus.Up, MemberStatus.Leaving);
```

Aaronontheweb:
Looks good to me. We'll get this in for v1.3.3

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -2138,6 +2142,39 @@ public void LeaderActions()
...
+
+            var changedMembers = localMembers
+                .Where(IsJoiningToWeaklyUp)
+                .Select(m => m.Copy(MemberStatus.WeaklyUp))
+                .ToImmutableSortedSet();
+
+            if (!changedMembers.IsEmpty)
+            {
+                // replace changed members
+                var newMembers = changedMembers.Union(localMembers);
```
Aaronontheweb: Son of a bitch... I totally missed this when I originally reviewed the PR. It's going to great the same problems here that happened in https://github.com/akkadotnet/akka.net/issues/2584. This is a bug. I'll PR it per https://github.com/akkadotnet/akka.net/issues/3274

PR #3101, opened 2017-09-14
InternalApi and ApiMayChange annotations

Horusiath:
Just like in case of Akka, I've added attribute annotations for parts of the API which are not stable (`ApiMayChange`) or shouldn't be treated as public even thou they are (because of .NET limitations). Regarding latter, I've annotated only those members/types which have either public/protected/not-set encapsulation. I didn't bother to add `InternalApi` to internal types, as this would be redundant.

Those attributes could be possibly used and leveraged in future by static analysis tools to make sure users are not unintentionally using private parts of Akka.NET API.


alexvaluyskiy:
I like this change.
We have a repository for static analyzers: https://github.com/AkkaNetContrib/AkkaAnalyzer

I think we should not use `InternalApi` for internal types. If you can make an internal type - just do it. If you can't - use this attribute

Issue #3090, opened 2017-09-08
Implement Restart Flow/Source/Sink


alexvaluyskiy:
https://github.com/akka/akka/pull/23367

PR #3103, opened 2017-09-15
Added RestartSource/Flow/Sink

alexvaluyskiy:
Implemented https://github.com/akkadotnet/akka.net/issues/3090

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs
@@ -1,0 +1,196 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.Streams.Dsl;
+using Akka.Streams.TestKit;
+using Akka.Util.Internal;
+using Xunit;
+using FluentAssertions;
+using Xunit.Abstractions;
+
+namespace Akka.Streams.Tests.Dsl
+{
+    public class RestartSpec : Akka.TestKit.Xunit2.TestKit
+    {
+        private ActorMaterializer Materializer { get; }
+
+        public RestartSpec(ITestOutputHelper output) : base("", null, output)
+        {
+            Materializer = Sys.Materializer();
+        }
+
+        [Fact]
+        public void A_restart_with_backoff_source_should_run_normally()
+        {
+            var created = new AtomicCounter(0);
+            var probe = RestartSource.WithBackoff(() =>
+            {
+                created.IncrementAndGet();
+                return Source.Repeat("a");
+            }, TimeSpan.FromMilliseconds(10), TimeSpan.FromMilliseconds(20), 0).RunWith(this.SinkProbe<string>(), Materializer);
+
+            probe.RequestNext("a");
+            probe.RequestNext("a");
+            probe.RequestNext("a");
+            probe.RequestNext("a");
+            probe.RequestNext("a");
+
+            created.Current.Should().Be(1);
+
+            probe.Cancel();
+        }
+
+        [Fact]
+        public void A_restart_with_backoff_source_should_restart_on_completion()
+        {
+            var created = new AtomicCounter(0);
+            var probe = RestartSource.WithBackoff(() =>
+            {
+                created.IncrementAndGet();
+                return Source.From(new List<string> { "a", "b" });
+            }, TimeSpan.FromMilliseconds(10), TimeSpan.FromMilliseconds(20), 0).RunWith(this.SinkProbe<string>(), Materializer);
+
+            probe.RequestNext("a");
+            probe.RequestNext("b");
+            probe.RequestNext("a");
+            probe.RequestNext("b");
+            probe.RequestNext("a");
+
+            created.Current.Should().Be(3);
+
+            probe.Cancel();
+        }
+
+        [Fact]
+        public void A_restart_with_backoff_source_should_restart_on_failure()
+        {
+            var created = new AtomicCounter(0);
+            var probe = RestartSource.WithBackoff(() =>
+            {
+                created.IncrementAndGet();
+                var enumerable = new List<string> { "a", "b", "c" }.Select(c =>
+                {
+                    if (c == "c")
+                        throw new ArgumentException("failed");
+                    return c;
+                });
+                return Source.From(enumerable);
+            }, TimeSpan.FromMilliseconds(10), TimeSpan.FromMilliseconds(20), 0).RunWith(this.SinkProbe<string>(), Materializer);
+
+            probe.RequestNext("a");
+            probe.RequestNext("b");
+            probe.RequestNext("a");
+            probe.RequestNext("b");
+            probe.RequestNext("a");
+
+            created.Current.Should().Be(3);
+
+            probe.Cancel();
+        }
+
+        [Fact]
+        public void A_restart_with_backoff_source_should_backoff_before_restart()
+        {
+            var created = new AtomicCounter(0);
+            var probe = RestartSource.WithBackoff(() =>
+            {
+                created.IncrementAndGet();
+                return Source.From(new List<string> { "a", "b" });
+            }, TimeSpan.FromMilliseconds(200), TimeSpan.FromMilliseconds(1000), 0).RunWith(this.SinkProbe<string>(), Materializer);
+
+            probe.RequestNext("a");
+            probe.RequestNext("b");
+            probe.Request(1);
+            // There should be a delay of at least 200ms before we receive the element, wait for 100ms.
+            var deadline = TimeSpan.FromMilliseconds(100).FromNow();
+            // But the delay shouldn't be more than 300ms.
+            probe.ExpectNext(TimeSpan.FromMilliseconds(300), "a");
+            deadline.IsOverdue.Should().Be(true);
+
+            created.Current.Should().Be(2);
+
+            probe.Cancel();
+        }
+
+        [Fact]
+        public void A_restart_with_backoff_source_should_reset_exponential_backoff_back_to_minimum_when_source_runs_for_at_least_minimum_backoff_without_completing()
+        {
+            var created = new AtomicCounter(0);
+            var probe = RestartSource.WithBackoff(() =>
+            {
+                created.IncrementAndGet();
+                return Source.From(new List<string> { "a", "b" });
+            }, TimeSpan.FromMilliseconds(200), TimeSpan.FromMilliseconds(2000), 0).RunWith(this.SinkProbe<string>(), Materializer);
+
+            probe.RequestNext("a");
+            probe.RequestNext("b");
+            // There should be a 200ms delay
+            probe.RequestNext("a");
+            probe.RequestNext("b");
+            probe.Request(1);
+            // The probe should now be backing off for 400ms
+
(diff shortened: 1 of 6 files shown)


src/core/Akka.Streams/Dsl/Restart.cs
```diff
@@ -0,0 +1,490 @@
...
+        /// When that happens, the wrapped <see cref="Source"/>, if currently running will be cancelled, and it will not be restarted.
+        /// This can be triggered simply by the downstream cancelling, or externally by introducing a <see cref="IKillSwitch"/> right
+        /// after this <see cref="Source"/> in the graph.
+        /// This uses the same exponential backoff algorithm as <see cref="Akka.Pattern.Backoff"/>.
+        /// </summary>
+        /// <param name="sourceFactory">A factory for producing the <see cref="Source"/> to wrap.</param>
+        /// <param name="minBackoff">Minimum (initial) duration until the child actor will started again, if it is terminated</param>
+        /// <param name="maxBackoff">The exponential back-off is capped to this duration</param>
+        /// <param name="randomFactor">After calculation of the exponential back-off an additional random delay based on this factor is added, e.g. `0.2` adds up to `20%` delay. In order to skip this additional delay pass in `0`.</param>
+        public static Source<T, NotUsed> WithBackoff<T>(Func<Source<T, NotUsed>> sourceFactory, TimeSpan minBackoff, TimeSpan maxBackoff, double randomFactor)
```
alexvaluyskiy: `sourceFactory` should have a signature `Creator[Source[T, _]]`, I don't know how to do it on C#

src/core/Akka.Streams/Dsl/Restart.cs
```diff
@@ -0,0 +1,490 @@
...
+        /// The restart process is inherently lossy, since there is no coordination between cancelling and the sending of
+        /// messages. When the wrapped <see cref="Sink"/> does cancel, this <see cref="Sink"/> will backpressure, however any elements already
+        /// sent may have been lost.
+        /// This uses the same exponential backoff algorithm as <see cref="Akka.Pattern.Backoff"/>.
+        /// </summary>
+        /// <param name="sinkFactory">A factory for producing the <see cref="Sink"/> to wrap.</param>
+        /// <param name="minBackoff">Minimum (initial) duration until the child actor will started again, if it is terminated</param>
+        /// <param name="maxBackoff">The exponential back-off is capped to this duration</param>
+        /// <param name="randomFactor">After calculation of the exponential back-off an additional random delay based on this factor is added, e.g. `0.2` adds up to `20%` delay. In order to skip this additional delay pass in `0`.</param>
+        public static Sink<T, NotUsed> WithBackoff<T>(Func<Sink<T, NotUsed>> sinkFactory, TimeSpan minBackoff, TimeSpan maxBackoff, double randomFactor)
```
alexvaluyskiy: `sinkFactory` should have a signature `Creator[Sink[T, _]]`, I don't know how to do it in C#

src/core/Akka.Streams/Dsl/Restart.cs
```diff
@@ -0,0 +1,490 @@
...
+        /// messages. A termination signal from either end of the wrapped <see cref="Flow"/> will cause the other end to be terminated,
+        /// and any in transit messages will be lost. During backoff, this <see cref="Flow"/> will backpressure.
+        /// This uses the same exponential backoff algorithm as <see cref="Akka.Pattern.Backoff"/>.
+        /// </summary>
+        /// <param name="flowFactory">A factory for producing the <see cref="Flow"/>] to wrap.</param>
+        /// <param name="minBackoff">Minimum (initial) duration until the child actor will started again, if it is terminated</param>
+        /// <param name="maxBackoff">The exponential back-off is capped to this duration</param>
+        /// <param name="randomFactor">After calculation of the exponential back-off an additional random delay based on this factor is added, e.g. `0.2` adds up to `20%` delay. In order to skip this additional delay pass in `0`.</param>
+        public static Flow<TIn, TOut, NotUsed> WithBackoff<TIn, TOut>(Func<Flow<TIn, TOut, NotUsed>> flowFactory, TimeSpan minBackoff, TimeSpan maxBackoff, double randomFactor)
```
alexvaluyskiy: `flowFactory` should have a signature `Creator[Flow[In, Out, _]`, I don't know how to do it in C#

src/core/Akka.Streams/Dsl/Restart.cs
```diff
@@ -0,0 +1,490 @@
...
+        {
+            StartGraph();
+            _resetDeadline = _minBackoff.FromNow();
+        }
+
+        // When the stage starts, start the source
+        public override void PreStart() => StartGraph();
+    }
+
+    internal sealed class Deadline
```
alexvaluyskiy: We have a similar class in Akka.Remote, but it accepts `DateTime`, not `TimeSpan`

src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs
```diff
@@ -0,0 +1,672 @@
...
+
+                // Now wait for the 400ms delay to pass, then it will start the new source, we also want to wait for the
+                // subsequent 200ms min backoff to pass, so it resets the restart count
+                Thread.Sleep(700);
+                probe.ExpectNext("a");
+                probe.RequestNext("b");
+
+                // We should have reset, so the restart delay should be back to 200ms, ie we should definitely receive the
+                // next element within 300ms
+                probe.RequestNext().Should().Be("a"); // TODO: why I can't set timeout here?
```
alexvaluyskiy: Can't setup timeouts here

src/core/Akka.Streams/Dsl/Sink.cs
```diff
@@ -241,7 +241,12 @@ public static class Sink
             => FromGraph(new LastOrDefault<TIn>()).WithAttributes(DefaultAttributes.LastOrDefaultSink);
 
         /// <summary>
-        /// TBD
+        /// A <see cref="Sink{TIn,TMat}"/> that keeps on collecting incoming elements until upstream terminates.
+        /// As upstream may be unbounded, `Flow.Create{T}().Take` or the stricter `Flow.Create{T}().Limit` (and their variants)
+        /// may be used to ensure boundedness.
+        /// Materializes into a <see cref="Task"/> of <see cref="Seq{TIn}"/> containing all the collected elements.
+        /// `Seq` is limited to <see cref="int.MaxValue"/> elements, this Sink will cancel the stream
+        /// after having received that many elements.
```
alexvaluyskiy: I've added a XML doc for Sink.Seq

src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs
```diff
@@ -416,18 +416,20 @@ public void A_restart_with_backoff_sink_should_not_restart_the_sink_when_complet
...
+                    var snk = Flow.Create<string>()
+                        .TakeWhile(s =>
+                        {
+                            return s != "cancel";
+                        })
+                        .To(Sink.ForEach<string>(c => flowInSource.SendNext(c))
+                            .MapMaterializedValue(task => task.ContinueWith(
+                                t1 =>
+                                {
+                                    if (t1.IsFaulted || t1.IsCanceled)
```
alexvaluyskiy: Good catch

src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs
```diff
@@ -479,129 +485,123 @@ public void A_restart_with_backoff_flow_should_run_normally()
         [Fact]
         public void A_restart_with_backoff_flow_should_restart_on_cancellation()
         {
-            this.AssertAllStagesStopped(() =>
```
alexvaluyskiy: What was the problem with `AssertAllStagesStopped`

src/core/Akka.Streams/Dsl/Restart.cs
alexvaluyskiy: @Aaronontheweb What to do with this class?

src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs
Silv3rcircl3: Not all specs are designed in a way that all stages are stopped at this point so the spec fails.

src/core/Akka.Streams/Dsl/Restart.cs
```diff
@@ -0,0 +1,490 @@
...
+        /// <param name="maxBackoff">The exponential back-off is capped to this duration</param>
+        /// <param name="randomFactor">After calculation of the exponential back-off an additional random delay based on this factor is added, e.g. `0.2` adds up to `20%` delay. In order to skip this additional delay pass in `0`.</param>
+        public static Source<T, NotUsed> WithBackoff<T, TMat>(Func<Source<T, TMat>> sourceFactory, TimeSpan minBackoff, TimeSpan maxBackoff, double randomFactor) 
+            => Source.FromGraph(new RestartWithBackoffSource<T, TMat>(sourceFactory, minBackoff, maxBackoff, randomFactor));
+    }
+
+    internal sealed class RestartWithBackoffSource<T, TMat> : GraphStage<SourceShape<T>>
+    {
+        public Func<Source<T, TMat>> SourceFactory { get; }
+        public TimeSpan MinBackoff { get; }
```
Silv3rcircl3: Why public properties?

src/core/Akka.Streams/Dsl/Restart.cs
```diff
@@ -0,0 +1,490 @@
...
+                    activeOutIn = null;
+                }
+            }
+        }
+    }
+
+    /// <summary>
+    /// Shared logic for all restart with backoff logics.
+    /// </summary>
+    internal abstract class RestartWithBackoffLogic<S> : TimerGraphStageLogic where S : Shape
```
Silv3rcircl3: You could implement IInAndOutHandler and use SetHandler(Out/In, this) to save a few allocations, we’ve done that in many other stages.

alexvaluyskiy: I'm working on microbenchmark set for Akka.Streams, and I will try to reduce some allocations in the future. I don;t like to do some unobvious performance improvements without any tests.

src/core/Akka.Streams/Dsl/Restart.cs
alexvaluyskiy: I want to access to those properties from GraphStageLogic. Anyway I can't see these properties because the class is internal

Silv3rcircl3: But the Logic class is inside the Source so you can access them even if they are private

alexvaluyskiy: It is true. But what the different? Scala converts all constructor's parameters (without additional `val` or `val`) to public fields. Why can't we do it?

Silv3rcircl3: No real difference, was just curious if there is a reason because in the base logic they’re private

alexvaluyskiy: We have a manual `How to convert a default Scala constructor to C#`. If I am not right, you can change this manual 
https://github.com/akkadotnet/akka.net/wiki/Scala-to-C%23-Conversion-Guide#class-constructors

Aaronontheweb:
The error on Mono is persistent

PR #3104, opened 2017-09-15
Obsoleted actor producer pipeline extensions

Horusiath:
As proposed in #3050 and #3057 , if we're going to modernize the way how we do create actors (including native support for dependency injection) an existing actor producer pipeline API will need to be removed - this isn't actually a great pain, as AFAIK it's not a very popular or well-known part of the API.

This PR obsoletes relevant types and fields with *Actor producer pipeline API will be removed in v1.5.* message, so that eventual users will be informed about this beforehand.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -202,6 +202,7 @@ namespace Akka.Actor
             public override int GetHashCode() { }
         }
     }
+    [System.ObsoleteAttribute("Actor producer pipeline API will be removed in v1.5.")]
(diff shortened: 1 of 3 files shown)


src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -202,6 +202,7 @@ namespace Akka.Actor
             public override int GetHashCode() { }
         }
     }
+    [System.ObsoleteAttribute("Actor producer pipeline API will be removed in v1.5.")]
```
alexvaluyskiy: Why v1.5? We could remove it in 1.4 or in 1.6. Actually in some next major version

Horusiath: I think it won't hurt if we give users more time for adoption. I hoped to bring native DI with v1.5, there is no immediate need to close this part of an API before that.

Issue #2354, opened 2016-10-26
Allow Akka.Streams to utilize an IObservable<T> as a source and sink


SeanFarrow:
It would be nice to allow Akka.Streams to use an IObservable<T> as both a source and a Sink.

I'm happy to do this, but the only question I have is whilst IObservable<T> is in the core framework assemblies there are no implementations available. Given this, do we want to take a dependency on RX, which would make it easier to implement the IObservable<T> interface as they provide base classes rather than implementing it ourselves?

Thoughts appreciated.
Cheers
Sean.

Silv3rcircl3:
I wouldn't like to add a RX dependency to Akka.Streams directly but instead create another library that only has a dependency to RX and Reactive.Streams. This way any other library that implements Reactive.Streams can use your library to interop with RX.

SeanFarrow:
Would you add a dependency to Akka.streams as well, there is a contrib project on the jvm, I wonder whether we should do something similar?
So have a separate akka.streams.contrib?

Silv3rcircl3:
We don't need a reference to Akka.Streams because we already have overloads to create a source from a publisher, it would kinda look like this

```csharp
var pub = rx.ToPublisher(Strategy.DropHead);
var src = Source.FromPublisher(pub);
```
same for the other way. 

So the library would be independent from streams. 

Yup we have already planed to create a similar repository, so this could be the first contribution ;-) 
I'll talk to the rest of the team and let you know once the repository is ready.

Silv3rcircl3:
@SeanFarrow we now have a repository for akka streams connectors, see https://github.com/akkadotnet/Alpakka . 

So if you are working on it please submit a PR there, if not, please tell us and I set it as `up for grabs`

SeanFarrow:
@silv3rcircl3,
I will be working on this in the next few weeks, so hopefully there will be a pr early December.
Cheers

PR #3112, opened 2017-09-16
[WIP] Interop betweek Akka.Streams and IObservable

Horusiath:

#### Motivation

Currently Akka.Streams is pretty self-contained - Rx.NET Observables adoption is much wider: partially probably due to fact that `IObservable<T>` is part of standard library and reactive streams are much younger and not so well-known. As it doesn't bring any extra dependencies, and may help with interop with existing technologies (i.e. SignalR for .NET Core 2.0 supports observables natively), I've decided to make this PR.

#### API 

This PR aims to introduce 2 new methods:

- `Sink.AsObservable<T>()` (already implemented with tests) - allows to materialize stream as `IObservable<T>`, which can be used for multi-observers scenarios (internally it's using immutable structures + interlocked, so it should be thread-safe as well, but that's harder to check).
- `Source.FromObservable<T>(IObservable<T> observable, int maxCapacity, OverflowStrategy strategy)` (in progress) it works in a similar way to how `Source.FromEvent` works right now - actually I want to merge those implementations together as C# events can be described as observables easily.


src/core/Akka.Streams/Implementation/Sources.cs
```diff
@@ -1003,7 +1003,7 @@ private Action<TEventArgs> SetupOverflowStrategy(OverflowStrategy overflowStrate
         /// <param name="addHandler">Action which attaches given event handler to the underlying .NET event.</param>
         /// <param name="removeHandler">Action which detaches given event handler to the underlying .NET event.</param>
         /// <param name="maxBuffer">Maximum size of the buffer, used in situation when amount of emitted events is higher than current processing capabilities of the downstream.</param>
-        /// <param name="overflowStrategy">Overflow strategy used, when buffer (size specified by <paramref name="maxBuffer"/>) has been overflown.</param>
+        /// <param name="overflowStrategy">Overflow strategy used, when b{uffer (size specified by <paramref name="maxBuffer"/>) has been overflown.</param>
```
alexvaluyskiy: Are you sure?

alexvaluyskiy:
Related https://github.com/akkadotnet/akka.net/issues/2354

alexvaluyskiy:
@Horusiath Probably we need an integration with `System.Threading.Tasks.Channels` in the future too

Horusiath:
When building `Source.FromObservable` I've run into following problem - since observables don't have any notion of demand/backpressure, they got pretty straightforward way of working order of `OnNext`/`OnError`/`OnCompleted` signals:

```csharp
var subject = new Subject<int>();
subject.Subscribe(
    onNext: x => Console.WriteLine($"Event: {x}"), 
    onError: e => Console.WriteLine($"Error: {e.Message}"), 
    onCompleted: () => Console.WriteLine("Completed"));

subject.OnNext(1);
subject.OnError(new Exception("hello"));
subject.OnNext(2);

// this will produce: 
//  - Event: 1
//  - Error: hello
```

When using reactive streams, we do have a backpressure - this means that no new events will be send downstream until an explicit demand will be signalized using `subscription.Request(n)`. *However backpressure works only on events!* This means that errors and completion signals are being send using a side lane, as soon as they come. This means that errors and completions may came out of order.

What does it mean in the current implementation?:

```csharp
var subject = new Subject<int>();
var probe = this.CreateManualSubscriberProbe<int>();
Source.FromObservable(subject).To(Sink.FromSubscriber(probe)).Run(_materializer);

var subscription = probe.ExpectSubscription();

subject.OnNext(1);
subject.OnError(new Exception("hello"));
subject.OnNext(2);

subscription.Request(2);

// what user may expect:
subscription.ExpectNext(1); // this won't happen as error is not backpressured and will be delivered immediatelly
subscription.ExpectError();
```

Ofc, there's an alternative approach - we could buffer errors/completions just like we buffer normal events - but this would complicate buffering and overflow strategy.

I think, that we should allow current behavior, simply because `subscription.Request` should happen immediately after subscribing to an observable.

alexvaluyskiy:
I think a documentation for this integration should be a part of the PR

Aaronontheweb:
@Horusiath is this still in a "do not merge" state or is it ready for review?

Horusiath:
@Aaronontheweb it's ready to be reviewed. There are some issues with Mono runtime (as usual).

Aaronontheweb:
> There are some issues with Mono runtime (as usual).

Grrr... I'll take a look at this and see what I can dig up on those Mono issues.

docs/articles/streams/integration.md
```diff
@@ -406,6 +406,55 @@ The numbers in parenthesis illustrates how many calls that are in progress at
...
+```
+
+In order to use an observable as an input source to Akka graph, you may want to use `Source.FromObservable<T>` method. Example:
+
+```csharp
+await Source.FromObservable(observable, maxBufferCapacity: 128, overflowStrategy: OverflowStrategy.DropHead)
+    .RunForEach(Console.WriteLine, materializer);
+```
+
+You may notice two extra parameters here. One of the advantages of Akka.Streams (and reactive streams in general) over Reactive Extensions in notion of backpressure - absent in Rx.NET. This puts a constraint of rate limiting the events incoming form upstream. If an observable will be producing events faster, than downstream is able to consume them, source stage will start to buffer them up to a provided `maxBufferCapacity` limit. Once that limit is reached, an overflow strategy will be applied. There are several different overflow strategies to choose from:
```
Silv3rcircl3: Is “in notion” correct? Sounds a bit strange to me or should it have been “is notion” ?

src/core/Akka.Streams/Implementation/Sinks.cs
```diff
@@ -1118,4 +1120,96 @@ public override ILogicAndMaterializedValue<Task<TMat>> CreateLogicAndMaterialize
...
+
+                        Pull(stage.Inlet);
+                    },
+                    onUpstreamFinish: () =>
+                    {
+                        var old = Interlocked.Exchange(ref _observers, ImmutableDictionary<IObserver<T>, ObserverDisposable>.Empty);
+                        foreach (var disposer in old.Values) disposer.Dispose(unregister: false);
+                    },
+                    onUpstreamFailure: e =>
+                    {
```
Silv3rcircl3: Don’t we need to remove observers in case of failure too?

src/core/Akka.Streams/Implementation/Sources.cs
```diff
@@ -962,60 +1014,39 @@ private Action<TEventArgs> SetupOverflowStrategy(OverflowStrategy overflowStrate
...
-        public EventSourceStage(Action<TDelegate> addHandler, Action<TDelegate> removeHandler, Func<Action<TEventArgs>, TDelegate> conversion, int maxBuffer, OverflowStrategy overflowStrategy)
+        public ObservableSourceStage(IObservable<T> observable, int maxBufferCapacity, OverflowStrategy overflowStrategy)
         {
-            _conversion = conversion ?? throw new ArgumentNullException(nameof(conversion));
-            _addHandler = addHandler ?? throw new ArgumentNullException(nameof(addHandler));
-            _removeHandler = removeHandler ?? throw new ArgumentNullException(nameof(removeHandler));
-            _maxBuffer = maxBuffer;
-            _overflowStrategy = overflowStrategy;
-            Shape = new SourceShape<TEventArgs>(Out);
+            this._observable = observable;
```
Silv3rcircl3: Maybe want to remove “this”, haven’t used it anywhere else so...

src/core/Akka.Streams/Implementation/Sinks.cs
Horusiath: Are we 100% sure that upstream failure should cause removal? What about `Rescue`?

Silv3rcircl3: Even if the observable catches the failure the Sink is dead so it wouldn’t produce any new items

Horusiath:
I couldn't debug this shit to identify the Mono issue. Any ideas?

src/core/Akka.Streams.Tests/Dsl/ObservableSourceSpec.cs
```diff
@@ -0,0 +1,298 @@
...
+            public IObserver<T> Observer { get; private set; }
+
+            public IDisposable Subscribe(IObserver<T> observer)
+            {
+                Observer = observer;
+                Subscribed = true;
+                return new TestDisposer(this);
+            }
+
+            public void Event(T element) => Observer.OnNext(element);
```
alexvaluyskiy: Try to put `[MethodImpl(MethodImplOptions.NoInlining)] here. It could on Mono

nesc58:
Something new? This interop would be very nice.

Aaronontheweb:
@neekgreen we need to disable Mono compilation on these builds. I'm sure the Mono runtime can run with these changes just fine, but the Mono compiler cannot.

vasily-kirichenko:
Any plans to include it to 1.4?

Horusiath:
I'll rebase it and push again - the only reason why it didn't land so far was some unrecognized issue on linux/mono.

PR #3120, opened 2017-09-22
Fix Ask deadlock #2

0x53A:
https://github.com/akkadotnet/akka.net/pull/2548

### DIFF
--- a/src/core/Akka/Actor/Futures.cs
+++ b/src/core/Akka/Actor/Futures.cs
@@ -128,10 +128,18 @@ namespace Akka.Actor
             return null;
         }
 
+        private const int RunContinuationsAsynchronously = 64;
+        private static readonly bool isRunContinuationsAsynchronouslyAvailable = Enum.IsDefined(typeof(TaskCreationOptions), RunContinuationsAsynchronously);
+
+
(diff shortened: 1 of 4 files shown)


src/core/Akka/Actor/Futures.cs
```diff
@@ -128,10 +128,18 @@ internal static IActorRefProvider ResolveProvider(ICanTell self)
             return null;
         }
 
+        private const int RunContinuationsAsynchronously = 64;
+        private static bool isRunContinuationsAsynchronouslyAvailable = Enum.IsDefined(typeof(TaskCreationOptions), RunContinuationsAsynchronously);
```
Aaronontheweb: @0x53A mark as `readonly`

Aaronontheweb:
@0x53A missing a public API approval: http://getakka.net/community/public-api-changes.html

Aaronontheweb:
@0x53A and the other PR we re-opened, should still look at merging that in addition to this one, correct?

0x53A:
This one supersedes the other one - you can decide which one to take (this branch is even based on the other one, but I would rebase it before you merge)

Issue #3109, opened 2017-09-16
Add Streams IO docs


alexvaluyskiy:
Akka.NET documentation does not have any mention of Streams IO (Tcp and Udp).
http://getakka.net/articles/streams/workingwithstreamingio.html

Horusiath:
I'm working on this one - already got some draft prepared.

PR #3127, opened 2017-09-26
Improve Streams IO docs

alexvaluyskiy:
Fixed: https://github.com/akkadotnet/akka.net/issues/3109


docs/examples/DocsExamples/Streams/StreamTcpDocTests.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            Materializer = Sys.Materializer();
+        }
+
+        [Fact]
+        public async Task Simple_server_connection_must_bind_and_unbind()
+        {
+
+            #region echo-server-simple-bind
+            var tcp = new TcpExt(Sys.AsInstanceOf<ExtendedActorSystem>());
```
Horusiath: This is ugly API (it should be hidden IMHO). `Sys.TcpStream()` looks a lot better.

alexvaluyskiy: Ok :)

alexvaluyskiy:
@Horusiath We also missed .NET Streams (MemoryStream, FileStream, etc)

alexvaluyskiy:
@Horusiath I've added images added an example with `BindAndHandle`. And have tested the first example with `netcat` on Linux Subsystem for Windows.

Still working on the rest two examples

src/core/Akka.Streams/Dsl/Tcp.cs
```diff
@@ -144,6 +145,7 @@ public class TcpExt : IExtension
         /// TBD
         /// </summary>
         /// <param name="system">TBD</param>
+        [InternalApi]
```
alexvaluyskiy: I've made it internal

docs/articles/streams/workingwithstreamingio.md
```diff
@@ -4,6 +4,61 @@ title: Working with streaming IO
...
+
+Notice that while most building blocks in Akka Streams are reusable and freely shareable, this is not the case for the incoming connection Flow, since it directly corresponds to an existing, already accepted connection its handling can only ever be materialized once.
+
+Closing connections is possible by cancelling the incoming connection `Flow` from your server logic (e.g. by connecting its downstream to a `Sink.Cancelled` and its upstream to a `Source.Empty`). It is also possible `to shut down the server’s socket by cancelling the `IncomingConnection` source `connections`.
+
+[!code-csharp[StreamTcpDocTests.cs](../../examples/DocsExamples/Streams/StreamTcpDocTests.cs?name=close-incoming-connection)]
+
+We can then test the TCP server by sending data to the TCP Socket using `netcat` (on Windows it is possible to use Linux Subsystem for Windows):
+```
+echo -n "Hello World" | netcat 127.0.0.1 8888
```
Horusiath: netcat is not available on Windows by default (or by Git bash). Have you tried to use it?

alexvaluyskiy: yes

docs/examples/DocsExamples/Streams/StreamTcpDocTests.cs
```diff
@@ -146,6 +115,52 @@ public void Simple_server_must_initial_server_banner_echo_server()
                 connection.HandleWith(serverLogic, Materializer);
             }, Materializer);
             #endregion
+
+            var input = new AtomicReference<List<string>>(new List<string> { "Hello world", "What a lovely day", null });
+
+            string ReadLine(string prompt)
+            {
+                // TODO: implement it
```
alexvaluyskiy: Have no idea how to convert it to C#
```scala
val input = new AtomicReference("Hello world" :: "What a lovely day" :: Nil)
def readLine(prompt: String): String = {
    input.get() match {
    case all @ cmd :: tail if input.compareAndSet(all, tail) => cmd
    case _ => "q"
    }
}
```

Horusiath: This is a very convoluted way to having concurrent queue with 2 elements in it ("Hello world", "What a lovely day") and picking those elements in thread safe fashion. Once a `readLine` hits an empty queue a `"q"` string is returned instead of normal content.

Horusiath:
Approvals API needs to be updated.

docs/examples/DocsExamples/Streams/StreamTcpDocTests.cs
alexvaluyskiy: Thanks, It works now

Aaronontheweb:
still missing API approval

Issue #3129, opened 2017-09-27
DistributedPubSub does not Send messages to remotely deployed actors


sbacquet:
Provided I have 2 actors running in a cluster :
- 1 locally deployed on node1, path = /user/MyActor
- 1 remotely deployed from node1 onto node2, path = /user/MyActor

When I do : 
`mediator.Tell(new SendToAll("/user/MyActor", "Hello"));`

I expect "Hello" to be sent on both actors. But the actor deployed on node1 is the only one to receive the message.

After debugging, it appears that the actor remotely deployed onto node2 is registered in the DistributedPubSub's mediator registry with path "/remote/.../user/MyActor", hence not matching the search key "/user/MyActor".

To fix it, I suggest to remove the "/remote/..." part from the path key when registering the actor ref in DistributedPubSub mediator's registry.

Akka.Net version : 1.3.1
Platform : Windows 7 Pro 64b
.NET : 4.5

sbacquet:
Added to the PR a unit test revealing the issue

Aaronontheweb:
Sorry for the delay on this, @sbacquet - going to take a look at your PR again for this soon

Aaronontheweb:
closed via #4207

PR #3132, opened 2017-09-28
Fix proposal for issue #3129

sbacquet:
Issue #3129 : remotely deployed actors were registered in DistributedPubSubMediator's register with key = "/remote/...", hence did not receive messages from mediator sent with Send("/user/...") or SendToAll("/user/...").
Now the key is the same for all actors = the path beginning with /user/.


Aaronontheweb:
@sbacquet not entirely sure I understand the outcome of the fix here. Could you please add a unit test to demonstrate?

sbacquet:
@Aaronontheweb the case is pretty simple and fully described in issue #3129.
As this issue involves remote deployment, I'm not sure I'd be able to write a unit test.
Will investigate...

sbacquet:
I've just pushed a unit test to highlight the issue.
Revert the modifications in Topics.cs and the test _DistributedPubSubMediator_must_publish_remote_ will fail.
In order to be able to remotely deploy actors, I've had to refactor a little bit.

sbacquet:
I can't figure out why the tests are racy (as usual, "it works fine on my computer").
I guess the issue is around : 
               var testActor = DistributedPubSubMediatorSpec.Current.TestActor;
but I don't know how to make it safe.
Any idea ?

It was an attempt to make a MultiNodeSpec compatible with remotely-deployed actors.
But obviously it's not the right way to do it...

Aaronontheweb:
closed via #4207

Issue #3123, opened 2017-09-24
Broken .NET Core Linux build


alexvaluyskiy:
```
[11:31:37][docker] ﻿It was not possible to find any compatible framework version
[11:31:37][docker] The specified framework 'Microsoft.NETCore.App', version '1.1.2' was not found.
[11:31:37][docker]   - Check application dependencies and target a framework version installed at:
[11:31:37][docker]       /
[11:31:37][docker]   - Alternatively, install the framework version '1.1.2'.
[11:31:37][docker] xUnit2 reported an error (Error Code 131)
```

marcpiechura:
Linux Build is working, so closing this one

PR #3135, opened 2017-09-29
Fix in resolving serializer id

Danthar:
Serializer id is now only determined once, upon initialisation. And its
the serializer implementation which dictates where that id comes from.
We no longer assume it always comes from the hocon config.

### DIFF
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -124,10 +124,8 @@ namespace Akka.Serialization
         {
             if (name == null)
                 return null;
-
-            if (!_serializersByName.TryGetValue(name, out Serializer serializer))
-                throw new ArgumentException($"Couldn't find serializer for [{name}] make sure you correctly defined.");
-
+           
+            _serializersByName.TryGetValue(name, out Serializer serializer);
             return serializer;
         }
 
@@ -271,7 +269,7 @@ namespace Akka.Serialization
                 }
             }
 
-            if (serializer == null)
+            if (serializer == null)  
                 serializer = GetSerializerByName(defaultSerializerName);
 
             // do a final check for the "object" serializer
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -71,7 +71,7 @@ namespace Akka.Serialization
             System = system;
 
             _nullSerializer = new NullSerializer(system);
-            AddSerializer(_nullSerializer);
+            AddSerializer("null",_nullSerializer);
 
             var serializersConfig = system.Settings.Config.GetConfig("akka.actor.serializers").AsEnumerable().ToList();
             var serializerBindingConfig = system.Settings.Config.GetConfig("akka.actor.serialization-bindings").AsEnumerable().ToList();
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -59,8 +59,8 @@ namespace Akka.Serialization
         private readonly Serializer _nullSerializer;
 
         private readonly ConcurrentDictionary<Type, Serializer> _serializerMap = new ConcurrentDictionary<Type, Serializer>();
-        private readonly Dictionary<int, Serializer> _serializers = new Dictionary<int, Serializer>();
-        private readonly Dictionary<string, int> _serializerIdByName = new Dictionary<string, int>();
+        private readonly Dictionary<int, Serializer> _serializersById = new Dictionary<int, Serializer>();
+        private readonly Dictionary<string, Serializer> _serializersByName = new Dictionary<string, Serializer>();
 
         /// <summary>
         /// TBD
@@ -76,7 +76,7 @@ namespace Akka.Serialization
             var serializersConfig = system.Settings.Config.GetConfig("akka.actor.serializers").AsEnumerable().ToList();
             var serializerBindingConfig = system.Settings.Config.GetConfig("akka.actor.serialization-bindings").AsEnumerable().ToList();
             var serializerSettingsConfig = system.Settings.Config.GetConfig("akka.actor.serialization-settings");
-            var namedSerializers = new Dictionary<string, Serializer>();
+            
             foreach (var kvp in serializersConfig)
             {
                 var serializerTypeName = kvp.Value.GetString();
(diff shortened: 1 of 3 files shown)


src/core/Akka/Serialization/Serialization.cs
```diff
@@ -145,14 +146,15 @@ private Serializer GetSerializerByName(string name)
...
-        /// <param name="serializer">TBD</param>
-        /// <returns>TBD</returns>
+        /// <param name="name">Configuration name of the serializer</param>
+        /// <param name="serializer">Serializer instance</param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void AddSerializer(Serializer serializer)
+        public void AddSerializer(string name, Serializer serializer)
         {
             _serializers.Add(serializer.Identifier, serializer);
+            _serializerIdByName.Add(name, serializer.Identifier);
```
Danthar: Ok so technically, the Identifier is retrieved twice. But since Hocon caching is a thing. I found this to be more expressive then first storing the id in a temp var and using that.

Aaronontheweb:
Going to do a review of this today

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -91,7 +92,8 @@ public Serialization(ExtendedActorSystem system)
                 var serializer = serializerConfig != null
                     ? (Serializer)Activator.CreateInstance(serializerType, system, serializerConfig)
                     : (Serializer)Activator.CreateInstance(serializerType, system);
-                AddSerializer(serializer);
+
+                AddSerializer(kvp.Key, serializer);
                 namedSerializers.Add(kvp.Key, serializer);
```
Aaronontheweb: @Danthar so aren't we doing the same thing with ` _serializerIdByName` and the `namedSerializers` local here? Why not just use that?

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -129,10 +131,9 @@ private Serializer GetSerializerByName(string name)
             {
                 if (kvp.Key.Equals(name))
                 {
-                    var serializerTypeName = kvp.Value.GetString();
-                    var serializerType = Type.GetType(serializerTypeName);
+                    if (!_serializerIdByName.TryGetValue(name, out int serializerId))
```
Aaronontheweb: I'd remove the `serializersConfig` line here - no need to read in HOCON if we're just using a cached map.

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -145,14 +146,15 @@ private Serializer GetSerializerByName(string name)
...
-        /// TBD
+        /// Adds the serializer to the internal state of the serialization subsystem
         /// </summary>
-        /// <param name="serializer">TBD</param>
-        /// <returns>TBD</returns>
+        /// <param name="name">Configuration name of the serializer</param>
+        /// <param name="serializer">Serializer instance</param>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void AddSerializer(Serializer serializer)
+        public void AddSerializer(string name, Serializer serializer)
```
Aaronontheweb: This is a breaking API change since we're modifying a public method. IMHO, add a new overload and mark the old one as deprecated. That way we won't have binary compatibility issues on upgrade.

src/core/Akka/Serialization/Serialization.cs
Aaronontheweb: I'd take a look at my previous comment and consider just maintaining two `<TKey, Serializer>` dictionaries - the ID-based one and the name based one, rather than go name --> id --> serializer you can just go name --> serializer. Does that make sense?

src/core/Akka/Serialization/Serialization.cs
Danthar: Because we already have a dictionary mapping id => serializer instance. I could have promoted the `namedSerializers` collection to a private member var. But that would mean we would store a reference to the same instance twice.

Not sure what my rationale was when i was writing this, but i wanted to avoid that. Hence the separate map `_serializerIdByName`

src/core/Akka/Serialization/Serialization.cs
Danthar: Yeah your right. The extra check on line 132 serves no purpose.

src/core/Akka/Serialization/Serialization.cs
Danthar: You are correct. However seeing as its being aggressively inlined. Im wondering why it was public in the first place. But ok.

Aaronontheweb: @Danthar 
> Im wondering why it was public in the first place. 

I'm right there with you.

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -124,19 +125,10 @@ private Serializer GetSerializerByName(string name)
...
 
-            var serializersConfig = System.Settings.Config.GetConfig("akka.actor.serializers").AsEnumerable().ToList();
-            foreach (var kvp in serializersConfig)
-            {
-                if (kvp.Key.Equals(name))
-                {
-                    var serializerTypeName = kvp.Value.GetString();
-                    var serializerType = Type.GetType(serializerTypeName);
+            if (!_serializersByName.TryGetValue(name, out Serializer serializer))
+                throw new ArgumentException($"Couldn't find serializer for [{name}] make sure you correctly defined.");
```
Danthar: I kept this check. Because it still possible to misconfigure for example serializer setting in persistence. Which is not checked during initialization.

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -145,14 +137,26 @@ private Serializer GetSerializerByName(string name)
         public ActorSystem System { get; }
 
         /// <summary>
-        /// TBD
+        /// Adds the serializer to the internal state of the serialization subsystem
         /// </summary>
-        /// <param name="serializer">TBD</param>
-        /// <returns>TBD</returns>
+        /// <param name="serializer">Serializer instance</param>
+        [Obsolete("No longer supported. Use the AddSerializer(name, serializer) overload instead.", true)]
```
Danthar: Left out a "deadline" version here. Or would you rather have i put in something like. will be removed in 1.4 or something ?

Aaronontheweb: This is fine as-is.

src/core/Akka/Serialization/Serialization.cs
Aaronontheweb: Sounds good

Aaronontheweb:
@Danthar looks like a NRE in your new spec

Danthar:
:D its interesting that the linux-netcore/unit-tests does complete.

However another spec failed as well. Ill be fixing both.

alexvaluyskiy:
@Danthar because it does not work https://github.com/akkadotnet/akka.net/issues/3123

PR #3138, opened 2017-10-03
Fixed performance degradation if recovering with big numbers of snaps…

planerist:
1) If your actor has a big number of snapshots then system experiences performance degradation. 

The reason for this degradation is SQL-driver rows pre-fetch. Furthermore, the current implementation of SelectSnapshotAsync reads only one row, so there is no reason to query more than one snapshot row.

2) DbCommand has memory leaks in ExecuteNonQueryAsync and ExecuteDbDataReaderAsync methods. So you should provide new cancellation token for each SQL command and dispose it (more here: https://github.com/dotnet/corefx/issues/16439).

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
@@ -273,7 +273,8 @@ namespace Akka.Persistence.Sql.Common.Snapshot
                 WHERE {Configuration.PersistenceIdColumnName} = @PersistenceId 
                     AND {Configuration.SequenceNrColumnName} <= @SequenceNr
                     AND {Configuration.TimestampColumnName} <= @Timestamp
-                ORDER BY {Configuration.SequenceNrColumnName} DESC";
+                ORDER BY {Configuration.SequenceNrColumnName} DESC
+                LIMIT 1";
 
             DeleteSnapshotSql = $@"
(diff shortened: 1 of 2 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
```diff
@@ -273,7 +273,8 @@ protected AbstractQueryExecutor(QueryConfiguration configuration, Akka.Serializa
                 WHERE {Configuration.PersistenceIdColumnName} = @PersistenceId 
                     AND {Configuration.SequenceNrColumnName} <= @SequenceNr
                     AND {Configuration.TimestampColumnName} <= @Timestamp
-                ORDER BY {Configuration.SequenceNrColumnName} DESC";
+                ORDER BY {Configuration.SequenceNrColumnName} DESC
+                LIMIT 1";
```
Danthar: 👍

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
```diff
@@ -110,9 +110,10 @@ protected override void PostStop()
             try
             {
                 using (var connection = CreateDbConnection())
+                using (var nestedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(_pendingRequestsCancellation.Token))
                 {
```
Danthar: Good solution on using CreateLinkedTokenSource

PR #3139, opened 2017-10-03
Fixed Akka.Cluster.TestKit nuget package (Fixes #3131)

### LINKED ISSUE #3131: Nuget Akka.Cluster.Testkit 1.3+ has a dependency on Akka.Tests.Shared.Internals (>= 1.0.0)
https://www.nuget.org/packages/Akka.Cluster.TestKit/1.3.1

View dependencies.

 Nuget Akka.Tests.Shared.Internals doesn't exist and I would imagine shouldn't be there?

derrickcrowne:
Removed Akka.Cluster.TestKit's dependency on Akka.Test.Shared.Internals which is not on nuget (and according to it's readme not intended to be public).

Replaced the usages of "Should" extension assertions with ITestKitAssertions as they were used elsewhere in the file, and the Shared.Internals extensions weren't really providing that much value.

### DIFF
--- a/src/core/Akka.Cluster.TestKit/MultiNodeClusterSpec.cs
+++ b/src/core/Akka.Cluster.TestKit/MultiNodeClusterSpec.cs
@@ -362,14 +362,14 @@ namespace Akka.Cluster.TestKit
                     AwaitAssert(() =>
                     {
                         foreach (var a in canNotBePartOfMemberRing)
-                            ClusterView.Members.Select(m => m.Address).Contains(a).ShouldBeFalse();
+                            _assertions.AssertFalse(ClusterView.Members.Select(m => m.Address).Contains(a));
                     });
-                AwaitAssert(() => ClusterView.Members.Count.ShouldBe(numbersOfMembers));
-                AwaitAssert(() => ClusterView.Members.All(m => m.Status == MemberStatus.Up).ShouldBeTrue("All members should be up"));
+                AwaitAssert(() => _assertions.AssertEqual(numbersOfMembers, ClusterView.Members.Count));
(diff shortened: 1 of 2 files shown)


derrickcrowne:
This can be considered a breaking change since Akka.Cluster.TestKit version 1.2.3 and earlier was exposing the AkkaSpecExtensions class because it was a public linked file. Not sure if that was intentional or not.

Aaronontheweb:
@derrickcrowne thanks! No, this was not an intentional breaking change. An oversight on our part mostly due to the change in MSBUILD15 and how we handled linked files.

src/core/Akka.Cluster.TestKit/MultiNodeClusterSpec.cs
```diff
@@ -362,14 +362,14 @@ public void AssertLeaderIn(ImmutableList<RoleName> nodesInCluster)
                     AwaitAssert(() =>
                     {
                         foreach (var a in canNotBePartOfMemberRing)
-                            ClusterView.Members.Select(m => m.Address).Contains(a).ShouldBeFalse();
+                            _assertions.AssertFalse(ClusterView.Members.Select(m => m.Address).Contains(a));
                     });
-                AwaitAssert(() => ClusterView.Members.Count.ShouldBe(numbersOfMembers));
-                AwaitAssert(() => ClusterView.Members.All(m => m.Status == MemberStatus.Up).ShouldBeTrue("All members should be up"));
+                AwaitAssert(() => _assertions.AssertEqual(numbersOfMembers, ClusterView.Members.Count));
```
Aaronontheweb: Please add the "All members should be up" error message back here.

Aaronontheweb: Ah whoops, nevermind. It's there on the line below.

Aaronontheweb:
Closes #3131

Issue #3141, opened 2017-10-04
Why does Akka.TestKit.Xunit depend on a prerelease version of xUnit.net?


stijnherreman:
The [current stable version](https://www.nuget.org/packages/Akka.TestKit.Xunit/1.3.1) of Akka.TestKit.Xunit depends on a prerelease version of xUnit.net (>= 2.3.0-beta5-build3769).

Any reason you need to use a prerelease version instead of the stable xUnit.net 2.2.0? Is it feasible to to relax the version restriction?

Aaronontheweb:
cc @alexvaluyskiy on that one

PR #3145, opened 2017-10-08
Updated Xunit to the stable version

alexvaluyskiy:
Fixed https://github.com/akkadotnet/akka.net/issues/3141


src/common.props
```diff
@@ -2,14 +2,14 @@
   <PropertyGroup>
     <Copyright>Copyright © 2013-2017 Akka.NET Team</Copyright>
     <Authors>Akka.NET Team</Authors>
-    <VersionPrefix>1.3.1</VersionPrefix>
+    <VersionPrefix>1.3.2</VersionPrefix>
```
alexvaluyskiy: Bumped a version number

alexvaluyskiy: @Aaronontheweb I think we should re-run all tests if `common.props` was changed. But our incremental testing did not do it

Aaronontheweb: @alexvaluyskiy that's a good call - I'll look into adding it

Issue #3147, opened 2017-10-11
ByteString.ToString is sometimes broken for Unicode encoding


tintoy:
Hi.

Currently, `ByteString.ToString(Encoding)` calls `Encoding.GetString` [for each underlying segment](https://github.com/akkadotnet/akka.net/blob/c12d08582bbeda6bb4d024df4f3eb992f4a00e0f/src/core/Akka/Util/ByteString.cs#L581-L590) that comprises the `ByteString`. Unfortunately, this breaks when you use an encoding such as `System.Text.Encoding.Unicode` and the underlying segments do not represent entire characters.

Here's a quick repro that works on Akka.NET 1.3.1 (tested on Windows, but it's probably the same everywhere):

```csharp
[Fact]
public void UnicodeByteString_Failure()
{
    const string expected = "ABC";
    Encoding encoding = Encoding.Unicode;

    byte[] rawData = encoding.GetBytes(expected);

    ByteString data = ByteString.Empty;
    data += ByteString.CopyFrom(rawData, 0, 3); // 1 and a half characters
    data += ByteString.CopyFrom(rawData, 3, 3); // 1 and a half characters
    Assert.Equal(rawData.Length, data.Count);

    string actual = data.ToString(encoding);
    Assert.Equal(expected, actual);
}
```

Note that if you compact the `ByteString` before calling ToString then it works fine.

```csharp
[Fact]
public void UnicodeByteString_Success()
{
    const string expected = "ABC";
    Encoding encoding = Encoding.Unicode;

    byte[] rawData = encoding.GetBytes(expected);

    ByteString data = ByteString.Empty;
    data += ByteString.CopyFrom(rawData, 0, 3);
    data += ByteString.CopyFrom(rawData, 3, 3);
    Assert.Equal(rawData.Length, data.Count);

    data = data.Compact();

    string actual = data.ToString(encoding);
    Assert.Equal(expected, actual);
}
```

I'd suggest that perhaps `ToString` should either call `Compact` or manually assemble a byte array before calling `Encoding.GetString`.

I'm happy to open a PR for this but wanted to discuss it with you, first, since it technically represents a change in behaviour.

Aaronontheweb:
@tintoy yep, looks right to me. Please add the repro specs and a PR to fix por favor.

tintoy:
Cheers, will do!

tintoy:
Going to add 1 more spec to cover the original mistake I made when trying to fix this.

PR #3148, opened 2017-10-12
Ensure ByteStrings are compact when calling ToString.

tintoy:
Fixes akkadotnet/akka.net#3147.

### DIFF
--- a/src/core/Akka/Util/ByteString.cs
+++ b/src/core/Akka/Util/ByteString.cs
@@ -581,9 +581,11 @@ namespace Akka.IO
         public string ToString(Encoding encoding)
         {
             if (IsCompact)
-                return encoding.GetString(_buffers[0].Array);
+                return encoding.GetString(_buffers[0].Array, _buffers[0].Offset, _buffers[0].Count);
 
-            return Compact().ToString(encoding);
+            byte[] buffer = ToArray();
+
+            return encoding.GetString(buffer);
         }
 
         public static bool operator ==(ByteString x, ByteString y) => Equals(x, y);
--- a/src/core/Akka/Util/ByteString.cs
+++ b/src/core/Akka/Util/ByteString.cs
@@ -580,13 +580,10 @@ namespace Akka.IO
 
         public string ToString(Encoding encoding)
         {
-            var builder = new StringBuilder(_count);
-            foreach (var buffer in _buffers)
-            {
-                var part = encoding.GetString(buffer.Array, buffer.Offset, buffer.Count);
-                builder.Append(part);
-            }
-            return builder.ToString();
+            if (IsCompact)
+                return encoding.GetString(_buffers[0].Array);
+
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
@tintoy looks like some of the JSON framing specs are failing for Akka.Streams, but the Akka.Persistence issues might be related to this: https://github.com/akkadotnet/akka.net/pull/3149 - should find out once CI has a change to run, but you should take a look at the failing specs for streams.

tintoy:
Yep, if I revert my change, then that test passes (but of course the other tests that I added fail). As far as I can tell, it's the `UnfoldResource` test that seems to be wrong (although I'm not sure why since I find the code a little hard to follow). It specifies a chunk size of `50`, but then expects a string that's only `10` long, which doesn't sound right.

tintoy:
Ok, found the bug - I didn't realise `ByteBuffer` is an alias for `ArraySegment<byte>`, so I was ignoring the other properties of the `ArraySegment` in the case that `IsCompact == true`. I'm just running the remaining streams tests then I'll push a fix.

src/core/Akka/Util/ByteString.cs
```diff
@@ -580,13 +580,12 @@ public IEnumerator<byte> GetEnumerator()
...
-            foreach (var buffer in _buffers)
-            {
-                var part = encoding.GetString(buffer.Array, buffer.Offset, buffer.Count);
-                builder.Append(part);
-            }
-            return builder.ToString();
+            if (IsCompact)
+                return encoding.GetString(_buffers[0].Array, _buffers[0].Offset, _buffers[0].Count);
+
+            byte[] buffer = ToArray();
```
Horusiath: What advantage does that give over using `StringBuilder`? I'm just curious.

tintoy: Hi! The problem is that when you have a `ByteString` composed of several `ByteString`s and some of those `ByteStrings` represent incomplete Unicode sequences (e.g. `"AB"` in Unicode is 4 bytes long, and if you split the `ByteString` after 3 bytes, resulting in a `ByteString` representing 1-and-a-half characters), then the resulting string will be incorrect because some of its components will be incorrect.

tintoy: Most of the existing code and tests used UTF-8, which encodes characters from the ASCII character set using only a single byte (so the issue may not have come up before). As soon as I switched to Unicode it gave wrong results unless I called `.Compact()` before calling `.ToString()`.

tintoy: [Example](https://github.com/akkadotnet/akka.net/pull/3148/files#diff-07edeee541b3c2f1fbe2a9aab5458214R135)

Aaronontheweb:
I like that there are multiple test cases for UTF8, Unicode, et al here. Looks like a good change to me. We'll get this into 1.3.2

Issue #3150, opened 2017-10-12
Build: .NET Core builds fail to copy some dependencies into appropriate release folder


Aaronontheweb:
Within the past couple of days, all of the Akka.Persistence.Sqlite specs on .NET Core on Windows have been failing with the following error:

```
[RunTestsNetCore] [INFO][10/12/2017 4:13:41 AM][Thread 0009][akka://SqliteSnapshotStoreSpec/system/akka.persistence.snapshot-store.sqlite] Message SaveSnapshot from akka://SqliteSnapshotStoreSpec/system/testActor11 to akka://SqliteSnapshotStoreSpec/system/akka.persistence.snapshot-store.sqlite was not delivered. 1 dead letters encountered.
[04:13:44]    [Akka.Persistence.Sqlite.Tests.SqliteSnapshotStoreSpec.SnapshotStore_should_not_load_a_snapshot_given_non_matching_timestamp_criteria] Failed: Timeout 00:00:03 while waiting for a message of type Akka.Persistence.SaveSnapshotSuccess 
Expected: True
Actual:   False
   at Akka.TestKit.Xunit2.XunitAssertions.Fail(String format, Object[] args) in D:\work\edbce350daf6a08d\src\contrib\testkits\Akka.TestKit.Xunit2\XunitAssertions.cs:line 27
   at Akka.TestKit.TestKitBase.InternalExpectMsgEnvelope[T](Nullable`1 timeout, Action`2 assert, String hint, Boolean shouldLog) in D:\work\edbce350daf6a08d\src\core\Akka.TestKit\TestKitBase_Expect.cs:line 241
   at Akka.TestKit.TestKitBase.InternalExpectMsgEnvelope[T](Nullable`1 timeout, Action`1 msgAssert, Action`1 senderAssert, String hint) in D:\work\edbce350daf6a08d\src\core\Akka.TestKit\TestKitBase_Expect.cs:line 224
   at Akka.Persistence.TCK.Snapshot.SnapshotStoreSpec.<WriteSnapshots>d__13.MoveNext() in D:\work\edbce350daf6a08d\src\core\Akka.Persistence.TCK\Snapshot\SnapshotStoreSpec.cs:line 80
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
[...]

PR #3151, opened 2017-10-12
[WIP] resolve issues with .NET Core dependency globbing

Aaronontheweb:
Attempt at resolving #3150 

Looks like something in the tooling we use changed suddenly, unrelated to our own changes, which has affected the way .NET Core projects copy their output to their respective bin folders from dependent DLLs.

switched to LTS .NET SDK channel to see if this helps.

### DIFF
--- a/src/core/Akka.Streams.Tests/Akka.Streams.Tests.csproj
+++ b/src/core/Akka.Streams.Tests/Akka.Streams.Tests.csproj
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <AssemblyName>Akka.Streams.Tests</AssemblyName>
-    <TargetFrameworks>net452;netcoreapp1.1.1</TargetFrameworks>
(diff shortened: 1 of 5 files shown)


src/core/Akka.Streams.Tests/Akka.Streams.Tests.csproj
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <AssemblyName>Akka.Streams.Tests</AssemblyName>
-    <TargetFrameworks>net452;netcoreapp1.1</TargetFrameworks>
+    <TargetFrameworks>net452;netcoreapp1.1.1</TargetFrameworks>
```
Aaronontheweb: Testing a theory on the recent issues with .NET Core on Linux not being able to run specifications, based on what I read here: https://stackoverflow.com/a/43924403/377476

Danthar:
@Aaronontheweb this thing is obsolete now right? Im closing it.

PR #3164, opened 2017-10-19
Fix an inconstant ToString on ConsistentRoutee when the node is remot…

annymsMthd:
…e vs local

Signed-off-by: Joshua Benjamin <benjamin@syncromatics.com>

### DIFF
--- a/src/core/Akka.Remote.Tests/RemoteConsistentHashingRouterSpec.cs
+++ b/src/core/Akka.Remote.Tests/RemoteConsistentHashingRouterSpec.cs
@@ -45,6 +45,44 @@ namespace Akka.Remote.Tests
             var result2 = keys.Select(k => consistentHash2.NodeFor(k).Routee);
             Assert.Equal(result2,result1);
         }
+
+        /// <summary>
+        /// This test creates two nodes each with a local routee to themselves and a remote routee to the other node.
+        /// When using ToString on ConsistentRoutee the local and remote routees are treated differently.
+        /// This test ensures the two are hashed the same.
+        /// </summary>
+        [Fact]
+        public void ConsistentHashingGroup_must_use_same_hash_ring_independent_of_local_and_remote_nodes()
+        {
+            var a1 = new Address("akka.tcp", "RemoteConsistentHashingRouterSpec-1", "client1", 2552);
+            var a2 = new Address("akka.tcp", "RemoteConsistentHashingRouterSpec-1", "client2", 2552);
+            var localActor = Sys.ActorOf(Props.Empty, "a");
+
+            var s1 = new ActorRefRoutee(localActor);
+            var s2 = new ActorSelectionRoutee(Sys.ActorSelection("akka.tcp://RemoteConsistentHashingRouterSpec-1@client2:2552/user/a"));
+            var nodes1 = new List<ConsistentRoutee>(new[] { new ConsistentRoutee(s1, a1), new ConsistentRoutee(s2, a1) });
+
+            var s4 = new ActorSelectionRoutee(Sys.ActorSelection("akka.tcp://RemoteConsistentHashingRouterSpec-1@client1:2552/user/a"));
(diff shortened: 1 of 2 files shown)


annymsMthd:
This bug is pretty nasty and is affecting us when we have a router that includes a local routee with Consistent hashing. Is there a way we can get this one fast tracked into a release?

Aaronontheweb:
@annymsMthd oh wow.. that IS nasty. Might have one bit of feedback though.

src/core/Akka.Remote.Tests/RemoteConsistentHashingRouterSpec.cs
```diff
@@ -45,6 +45,36 @@ public void ConsistentHashingGroup_must_use_same_hash_ring_independent_of_self_a
...
         }
+
+        [Fact]
+        public void ConsistentHashingGroup_must_use_same_hash_ring_independent_of_local_and_remote_nodes()
+        {
+            var a1 = new Address("akka.tcp", "RemoteConsistentHashingRouterSpec-1", "client1", 2552);
+            var a2 = new Address("akka.tcp", "RemoteConsistentHashingRouterSpec-1", "client2", 2552);
+            var localActor = Sys.ActorOf(Props.Empty, "a");
+
+            var s1 = new ActorRefRoutee(localActor);
```
Aaronontheweb: should add a comment here / something that makes it obvious how this reproduces the bug

src/core/Akka.Remote.Tests/RemoteConsistentHashingRouterSpec.cs
```diff
@@ -45,6 +45,36 @@ public void ConsistentHashingGroup_must_use_same_hash_ring_independent_of_self_a
...
+
+            var result1 = keys
+                .Select(k => consistentHash1.NodeFor(k))
+                .Select(routee => routee.ToString());
+
+            var result2 = keys
+                .Select(k => consistentHash2.NodeFor(k))
+                .Select(routee => routee.ToString());
+
+            Assert.Equal(result2, result1);
```
Aaronontheweb: This does a `.SequenceEqual` call under the covers?

src/core/Akka/Routing/ConsistentHashRouter.cs
```diff
@@ -292,7 +292,7 @@ public override string ToString()
                 case ActorRefRoutee actorRef:
                     return ToStringWithFullAddress(actorRef.Actor.Path);
                 case ActorSelectionRoutee selection:
-                    return ToStringWithFullAddress(selection.Selection.Anchor.Path) +
+                    return ToStringWithFullAddress(selection.Selection.Anchor.Path).TrimEnd('/') +
```
Aaronontheweb: So it was just the trailing slash that created this issue?

Aaronontheweb:
@annymsMthd does this need to be tested on clustered consistent hash routers also? Or is the test case here sufficient?

src/core/Akka.Remote.Tests/RemoteConsistentHashingRouterSpec.cs
annymsMthd: I think so. I can do a FluentAssertion here that's much clearer.

src/core/Akka/Routing/ConsistentHashRouter.cs
annymsMthd: Yup. It results in double slashes. When everything is remote it works because they all have it. If you have a local routee you start seeing a little weirdness.

annymsMthd:
Since the Clustered router uses the in it's internals I think it should be fine just testing it at this level. I'm on a bus but once I get into the office I'll update this PR with your suggestions.

src/core/Akka.Remote.Tests/RemoteConsistentHashingRouterSpec.cs
annymsMthd: added it on the method

PR #3178, opened 2017-10-27
Avoid NullReferenceException in DistributedPubSubMediator.PublishMessage

### LINKED ISSUE #3177: NullReferenceException in DistributedPubSubMediator.PublishMessage
-Akka.net 1.2.3 
-.net 4.5
-Windows Server 2012
```csharp
  at Akka.Cluster.Tools.PublishSubscribe.DistributedPubSubMediator.PublishMessage(String path, Object message, Boolean allButSelf)
```
This exception sometimes appears when an actor Publish a message, but in differents nodes that use PubSub.

https://github.com/akkadotnet/akka.net/blob/28cff8c3bd69fa42f18ab524799b48445f42e14a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs#L516

pubsubmediator

```csharp
private void Handler(SomeMessage message)
{
	var mediator = DistributedPubSub.Get(Context.System).Mediator;	  
	mediator.Tell(new Publish("topic", new SomeOtherMessage() ));	
}
```

ismaelhamed:
Fixes #3177

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
@@ -475,19 +475,23 @@ namespace Akka.Cluster.Tools.PublishSubscribe
 
         private void PublishMessage(string path, object message, bool allButSelf = false)
         {
-            foreach (var entry in _registry)
+            IEnumerable<IActorRef> Refs()
             {
-                var address = entry.Key;
-                var bucket = entry.Value;
-
-                if (!(allButSelf && address == _cluster.SelfAddress) && bucket.Content.TryGetValue(path, out var valueHolder))
+                foreach (var entry in _registry)
                 {
-                    if (valueHolder != null && !valueHolder.Ref.Equals(ActorRefs.Nobody))
-                        valueHolder.Ref.Forward(message);
-                    else
-                        SendToDeadLetters(message);
+                    var address = entry.Key;
+                    var bucket = entry.Value;
+
+                    if (!(allButSelf && address == _cluster.SelfAddress) && bucket.Content.TryGetValue(path, out var valueHolder))
+                    {
+                        if (valueHolder != null && !Equals(valueHolder.Ref, ActorRefs.Nobody))
+                            yield return valueHolder.Ref;
+                    }
                 }
             }
+
+            if (!Refs().Any()) SendToDeadLetters(message);
+            else Refs().ForEach(r => r?.Forward(message));
         }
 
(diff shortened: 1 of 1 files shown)


src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -475,19 +475,23 @@ private void SendToDeadLetters(object message)
...
+
+                    if (!(allButSelf && address == _cluster.SelfAddress) && bucket.Content.TryGetValue(path, out var valueHolder))
+                    {
+                        if (valueHolder != null && !Equals(valueHolder.Ref, ActorRefs.Nobody))
+                            yield return valueHolder.Ref;
+                    }
                 }
             }
+
+            if (!Refs().Any()) SendToDeadLetters(message);
```
alexvaluyskiy: @ismaelhamed Have you tested performance here? LINQ could slow-down the code

ismaelhamed: Nop, are there any performance benchmark in place that can be use?

Horusiath: I guess you can at least be sure that yield generator will be called twice in most scenarios (1st time inside if statement, 2nd inside `Refs.ForEach`.

Aaronontheweb: @ismaelhamed no benchmarks for DPubSub at the moment, but best thing you could do here is to make sure that the list only gets evaluated once. Maybe just do a `foreach` loop over the `IEnumerable` and include a counter variable at the end. If the counter was not incremented, send it to dead letters.

ismaelhamed: Not pretty though

```csharp
var counter = 0;
foreach (var r in Refs())
{
    if (r == null) continue;
    r.Forward(message);
    counter++;
}

if (counter == 0) SendToDeadLetters(message);
```

Danthar:
@ismaelhamed if you make that last change. We can wrap this PR up.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Aaronontheweb: @ismaelhamed high performance code often isn't :(

Issue #3162, opened 2017-10-18
Akka.Status.Failure.ToString() throws when Cause is null


HenryTonnison:
**Akka.Actor.Futures.Ask(this ICanTell self, object message, TimeSpan? timeout = null)**
will cancel the Task if the Timeout is reached.

The cancelled task is then handled by PipeTo
**Akka.Actor.PipeToSupport.PipeTo<T>(this Task<T> taskToPipe, ICanTell recipient, IActorRef sender = null, Func<T, object> success = null, Func<Exception, object> failure = null)**

_if (tresult.IsCanceled || tresult.IsFaulted)
{
          recipient.Tell(failure != null ? failure((Exception) tresult.Exception) : **(object) new Status.Failure((Exception) tresult.Exception)**, sender);
 }_

The new Status.Failure() is created when task is cancelled the Exception is null.
**Akka.Actor.Failure.ToString()**
{
       return "Failure: " + this.Cause.ToString();
}

Then blows up, in akka's own logging.

System.NullReferenceException: Object reference not set to an instance of an object.
   at **Akka.Actor.Status.Failure.ToString()**
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(IFormatProvider provider, String format, Object[] args)
   at Akka.Event.LoggingBus.UnhandledMessageForwarder.ToDebug(UnhandledMessage message)
   at Akka.Event.LoggingBus.UnhandledMessageForwarder.Receive(Object message)
   at Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message)
   at Akka.Actor.ActorCell.ReceiveMessage(Object message)
   at Akka.Actor.ActorCell.Invoke(Envelope envelope)

--------------------------------------
My opinion....

1. Failure.ToString() handles null.
2. In PipeTo, if the task is cancelled, fail with a TaskCanceledException
3. The Ask doesn't cancel a task on timeout but raises a TimeoutException

Aaronontheweb:
> Failure.ToString() handles null. 

Should definitely do that.

> In PipeTo, if the task is cancelled, fail with a TaskCanceledException

[...]

PR #3182, opened 2017-11-01
When a task is cancelled signal and when timed out raise exception

HenryTonnison:
Fix for https://github.com/akkadotnet/akka.net/issues/3162

I was genuinely spooked that we didn't get a TimeoutException on timeouts and TaskCanceledException on cancellations.

I appreciate this is a bit of a fundamental change, I think it's an improvement, but my understanding probably isn't good enough to see possible repercussions.


Horusiath:
This PR changes a timeout contract a little (from `AskTimeoutException` &rarr; `AggregateException` with inner exn). This causes `ScatterGatherFirstCompletedSpec.Scatter_gather_group_must_handle_failing_timeouts` to fail.
 Anyone in @akkadotnet/core has some pros/cons to say here?

Aaronontheweb:
Changes are going to break a whole bunch of other people's code who depend on the current exceptions being thrown as-is.

src/core/Akka.Tests/Actor/AskSpec.cs
```diff
@@ -141,6 +141,23 @@ public void Can_Ask_actor_inside_receive_loop()
             waitActor.Tell("ask");
             ExpectMsg("bar");
         }
+
+        [Fact]
+        public void Cancelled_ask_with_timeout_should_raise_timeout_exception()
```
maxcherednik: Can xunit supports async await. Can we please use it? This will help to avoid wrapping/unwrapping exceptions.

maxcherednik:
@Horusiath Why do we need to change the contract of the method?
Why do we want to use this AggregatedException?

PR #3186, opened 2017-11-07
ported Chat sample to MSBUILD 15

Aaronontheweb:


src/examples/Chat/ChatMessages/Messages.cs
```diff
@@ -1,59 +1,52 @@
-﻿//-----------------------------------------------------------------------
```
Aaronontheweb: Derp... Going to need to add the copyright header back.

PR #3199, opened 2017-11-19
DData-based cluster sharding

Horusiath:
This is PR bringing DistributedData-based mode for Akka.Cluster.Sharding. On the Akka JVM it's already a default option for sharding actors. It allows to use sharding without need for persistent journals.


Aaronontheweb:
Looking forward to this!

Horusiath:
This one is done. The only downside when compared to existing persistence mode (which still is the default one) is that DData doesn't support remember-entities option. The reason here is that ddata durable store is not yet ready, as it requires well-defined data format - so far we serialize ddata structures with Hyperion.

src/contrib/cluster/Akka.Cluster.Sharding/DDataShard.cs
```diff
@@ -0,0 +1,229 @@
...
+        public IStash Stash { get; set; }
+
+        private readonly IReadConsistency _readConsistency;
+        private readonly IWriteConsistency _writeConsistency;
+        private const int MaxUpdateAttempts = 3;
+
+        // The default maximum-frame-size is 256 KiB with Artery.
+        // When using entity identifiers with 36 character strings (e.g. UUID.randomUUID).
+        // By splitting the elements over 5 keys we can support 10000 entities per shard.
+        // The Gossip message size of 5 ORSet with 2000 ids is around 200 KiB.
```
Aaronontheweb: These values are the same in .NET Desktop, slightly smaller in .NET Core since strings are UTF-8 instead of Unicode there. But either way, we're still in the clear.

src/contrib/cluster/Akka.Cluster.Sharding/DDataShard.cs
```diff
@@ -0,0 +1,229 @@
...
+        }
+        private Receive WaitingForState(ImmutableHashSet<int> gotKeys) => message =>
+        {
+            void ReceiveOne(int i)
+            {
+                var newGotKeys = gotKeys.Add(i);
+                if (newGotKeys.Count == NrOfKeys)
+                    RecoveryCompleted();
+                else
+                    Context.Become(WaitingForState(newGotKeys));
```
Aaronontheweb: Just a comment and asking because I don't know the answer, but what's the allocation impact of these delegate calls that rely on argument closures? Just a new delegate and a long-lived reference to the closure itself?

src/contrib/cluster/Akka.Cluster.Sharding/DDataShard.cs
```diff
@@ -0,0 +1,229 @@
...
+        // When using entity identifiers with 36 character strings (e.g. UUID.randomUUID).
+        // By splitting the elements over 5 keys we can support 10000 entities per shard.
+        // The Gossip message size of 5 ORSet with 2000 ids is around 200 KiB.
+        // This is by intention not configurable because it's important to have the same
+        // configuration on each node.
+        private const int NrOfKeys = 5;
+
+        private readonly ImmutableArray<ORSetKey<EntryId>> _stateKeys;
+
+        public DDataShard(
```
Aaronontheweb: Looks good to me; went through all of it.

src/core/Akka.MultiNodeTestRunner/Discovery.cs
```diff
@@ -92,10 +92,10 @@ private IEnumerable<RoleName> RoleNames(Type specType)
             var args = ConfigConstructorParamValues(configType);
             var configInstance = Activator.CreateInstance(configType, args);
             var roleType = typeof(RoleName);
-            var configProps = configType.GetProperties(BindingFlags.DeclaredOnly | BindingFlags.Instance | BindingFlags.Public);
-            var roleProps = configProps.Where(p => p.PropertyType == roleType).Select(p => (RoleName)p.GetValue(configInstance));
-            var configFields = configType.GetFields(BindingFlags.DeclaredOnly | BindingFlags.Instance | BindingFlags.Public);
-            var roleFields = configFields.Where(f => f.FieldType == roleType).Select(f => (RoleName)f.GetValue(configInstance));
+            var configProps = configType.GetProperties(BindingFlags.Instance | BindingFlags.Public);
+            var roleProps = configProps.Where(p => p.PropertyType == roleType && p.Name != "Myself").Select(p => (RoleName)p.GetValue(configInstance));
```
Aaronontheweb: @Horusiath errp.... Looks like these were debug lines added when you were testing this stuff. Going to need to remove that.

Horusiath: No actually those changes are valid. Without them you cannot make inherited role names i.e. if you defined roles First, Second in some base `MultiNodeConfig` class, then inherit from it (like in the case of  `ClusterShardingSpecConfig`&rarr;`PersistentClusterShardingSpecConfig` and `ClusterShardingSpecConfig`&rarr;`DDataClusterShardingSpecConfig`), then MNTK runner won't detect those role names in actual specs, making the test cluster empty.

Aaronontheweb: What about the ` p.Name != "Myself"` - that looks like a one-off to me? No?

Horusiath: By default, every `MultiNodeConfig` contains a property called Myself, which is also a `RoleName`. While we may want to spin cluster nodes based on properties like Controller, First, Second, Third etc., I guess we don't want to spin a cluster node called Myself ;) Especially since it has some specific meaning in context of a config.

I was thinking about excluding properties to be used as test cluster nodes by attaching some special attribute to them, but it seem to be too much ceremony - so far it's used only in one case.

Aaronontheweb: @Horusiath got it. Ok then, objection overruled. This is good to go.

Issue #3203, opened 2017-11-29
Recursive call in Akka.IO


andreichuk:
Akka 1.3.2

[This method](https://github.com/akkadotnet/akka.net/blob/74ab3ce8aff710585b28586cd688220e2fae000a/src/core/Akka/IO/TcpConnection.cs#L833-L849) in Akka.IO uses recursive calls of itselft and that may lead to StackOverflowException because C# does not care about tail-call optimization. The code can be converted to while-loop -- any objections?

Aaronontheweb:
Resolved via https://github.com/akkadotnet/akka.net/pull/3206

Issue #3188, opened 2017-11-10
Akka.IO & WSAEWOULDBLOCK socket error


andreichuk:
Akka.NET version 1.3.2
Windows 10

Akka.IO does not handle non-fatal socket errors like **WSAEWOULDBLOCK** (System.Net.Sockets.SocketError.WouldBlock) which lead to closing of a connection.
From [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668(v=vs.85).aspx#WSAEWOULDBLOCK): _Resource temporarily unavailable. This error is returned from operations on nonblocking sockets that cannot be completed immediately, for example recv when no data is queued to be read from the socket. It is a nonfatal error, and the operation should be retried later._

When I'm trying to send a large (1 Mb or more) amount of data over the wire, I split it in chunks ~64kb and send one by one:
1. Get a chunk of data (if anything left)
2. Send it to Akka.IO
3. Receive an acknowledgement
4. Go to 1.
With default Akka.IO tcp settings sending a file from a client to a server (both localhost) usually fails after few cycles because a socket throws WSAEWOULDBLOCK socket exception. It looks like the acknowledgement is sent by Akka.IO when data has been added to the socket's send buffer, not when it's actually has been sent. So overflowing the send buffer leads to the error. Decreasing chunk size to ~10 Kb helps but there is no guarantee that:
a) there is only one "chunk sender" on the client at the moment;
b) the server is able read/process even those 10 Kb fast enough.

I checked [this code](https://github.com/akkadotnet/akka.net/blob/74ab3ce8aff710585b28586cd688220e2fae000a/src/core/Akka/IO/TcpConnection.cs#L833-L849). It does not catch any exceptions, just checks a number of bytes written. Maybe there should be a catcher for WSAEWOULDBLOCK that waits for N milliseconds and then call the method again? Or you have better ideas? Help is appreciated! :)

Horusiath:
Some notes from talk about this feature in DotNetty (also see code examples: [1](https://github.com/Azure/DotNetty/blob/dev/src/DotNetty.Transport/Channels/Sockets/TcpSocketChannel.cs#L204) and [2](https://github.com/Azure/DotNetty/blob/dev/src/DotNetty.Transport/Channels/Sockets/TcpSocketChannel.cs#L286)):

> Basically it attempts sync ops and switches to async when it would block otherwise
now that there's true support for sync completions in .net core it might make sense to forego sync path completely for writes

[...]

PR #3206, opened 2017-12-02
Akka.IO bug fixes

Horusiath:
This PR fixes 2 bugs:

- One is #3203 - a write request was made as part of recursive function call. This could result with StackOverflowException when ratio of written bytes to TCP buffer chunk size was to big (either the bigger the payload or the smaller the direct byte buffer pool could cause this).
- Another isssue is #3188 - when using non-blocking IO, occasionally we could run into non-fatal `SocketError.WouldBlock`. According to winsock docs, in this case we should try to make a send later. However we could also switch flag to `Socket.Blocking = true` and fallback to blocking send. This is not ideal, but for sure it's better than the current state.

### DIFF
--- a/src/core/Akka/IO/TcpConnection.cs
+++ b/src/core/Akka/IO/TcpConnection.cs
@@ -830,7 +830,8 @@ namespace Akka.IO
             {
                 try
                 {
-                    PendingWrite WriteToChannel(ByteString data)
+                    var data = _buffer;
+                    while(true)
                     {
                         var bytesWritten = _connection.Socket.Send(data.Buffers);
                         if (_connection.traceLogging)
@@ -838,7 +839,7 @@ namespace Akka.IO
                         if (bytesWritten < data.Count)
                         {
                             // we weren't able to write all bytes from the buffer, so we need to try again later
-                            return WriteToChannel(data.Slice(bytesWritten));
+                            data = data.Slice(bytesWritten);
                         }
                         else // finished writing
                         {
@@ -847,8 +848,6 @@ namespace Akka.IO
                             return _connection.CreatePendingWrite(Commander, _tail, info);
                         }
                     }
-
-                    return WriteToChannel(_buffer);
 
                 }
                 catch (SocketException e)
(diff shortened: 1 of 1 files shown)


src/core/Akka/IO/TcpConnection.cs
```diff
@@ -857,6 +858,26 @@ public override PendingWrite DoWrite(ConnectionInfo info)
...
+                try
+                {
+                    return _connection.Socket.Send(data.Buffers);
+                }
+                catch (SocketException e) when (e.SocketErrorCode == SocketError.WouldBlock)
+                {
+                    try
+                    {
+                        _connection.Socket.Blocking = true;
+                        return _connection.Socket.Send(data.Buffers);
```
Aaronontheweb: Looks good to me

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -830,15 +830,16 @@ public override PendingWrite DoWrite(ConnectionInfo info)
...
+                    while(true)
                     {
                         var bytesWritten = _connection.Socket.Send(data.Buffers);
                         if (_connection.traceLogging)
                             _connection.Log.Debug("Wrote [{0}] bytes to channel", bytesWritten);
                         if (bytesWritten < data.Count)
                         {
                             // we weren't able to write all bytes from the buffer, so we need to try again later
-                            return WriteToChannel(data.Slice(bytesWritten));
+                            data = data.Slice(bytesWritten);
```
Aaronontheweb: Looks good

src/core/Akka/IO/TcpConnection.cs
shersh: Guys. Does it mean that connection after that will be "blocking" forever ?

Aaronontheweb: No, only for this operation - on the line below we turn the socket back to non-blocking afterwards in the `finally` block. Unless the semantics of how this is implemented by the CLR are different than expected...

PR #3209, opened 2017-12-11
Fix unicode error

samneirinck:
Fixes a small error I noticed while reading the Streams documentation


Aaronontheweb:
Thanks @samneirinck - not sure what the CI errors are griping about. Has nothing to do with your changes. I'll take a look at it!

Issue #3211, opened 2017-12-13
TcpOutgoingConnection does not dispose properly - memory leak


yoPCix:
Experienced on Akka.Net 1.3.2. On Windows 10 64-bit
Steps to recreate:
1. Open Tcp connections
2. Close all Tcp connections (stop all Tcp connection actors)
Even though all actual Sockets are being closed and disposed, some SocketAsyncEventArgs instances are not being disposed. They contain reference to IActorRef so the ActorCell instance is also not getting GC'ed.

I debugged through source code and found out that when opening connection, TcpOutgoingConnection acquires 3 SocketAsyncEventArgs instances from PreallocatedSocketEventAgrsPool, however when connection closes, only 2 instances are getting released.
Also if an InvalidOperationException exception occurs during release, the SocketAsyncEventArgs instance will probably never get neither back to the pool nor get disposed because the connection actor is already stopped.

Horusiath:
Thanks @yoPCix for pointing that out. I'll try to take a look at this, but given the amount of tasks, I've got on me, this may take a while. If you're willing to help, I'd appreciate that ;)

yoPCix:
I already fixed it. I'll create a pull request soon.

Aaronontheweb:
Closed via #3212

PR #3212, opened 2017-12-13
Fix TcpOutgoingConnection memory leak (#3211)

yoPCix:
TcpOutgoingConnection saves reference to SocketAsyncEventArgs it acquires during connection in a protected field on base TcpConnection class. This field is then being disposed when closing connection.

Also PreallocatedSocketEventAgrsPool forces dispose of SocketAsyncEventArgs during release even if InvalidOperationException is caught. This also lead to some memory leak because this exception gets raised quite often when system is under load.

### DIFF
--- a/src/core/Akka/IO/TcpConnection.cs
+++ b/src/core/Akka/IO/TcpConnection.cs
@@ -622,8 +622,10 @@ namespace Akka.IO
             if (IsWritePending)
             {
                 pendingWrite.Release(); // we should release ConnectionInfo event args (if they're not released already)
-                ReleaseSocketAsyncEventArgs();
             }
+
+            // always try to release SocketAsyncEventArgs to avoid memory leaks
+            ReleaseSocketAsyncEventArgs();
 
             if (closedMessage != null)
             {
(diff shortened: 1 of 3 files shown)


src/core/Akka/IO/TcpConnection.cs
```diff
@@ -560,6 +561,12 @@ private void ReleaseSocketAsyncEventArgs()
                 Tcp.SocketEventArgsPool.Release(SendArgs);
                 SendArgs = null;
             }
+
+            if (ConnectArgs != null)
+            {
+                Tcp.SocketEventArgsPool.Release(ConnectArgs);
```
Horusiath: I'm fine with this solution, however the initial intention here was to recycle connectArgs after connection has been established. After that point it won't be used anymore, so it doesn't need to be keep alive until actor stops.

yoPCix:
The fix was not complete. In case of connection timeout it calls Stop() method on TcpOutgoingConnection and then StopWith(...) of TcpConnection that does not release socket args. So the memory leak is still there.
Indeed it looks like it would be best to release right after successful or failed connect. I will modify it accordingly.
Also if server closes connection, TcpConnection does not call CloseConnection at all, instead it goes to HandleError(..) directly which only stops the actor. It subsequently calls PostStop but it has a condition for socket args release which may not trigger. So in this case SocketAsyncEventArgs are not released as well.
I modified the fix to address all these issues.

Issue #3210, opened 2017-12-13
Udp.Received give incorrect ByteString when client send several packets at once


blackclavus:
Akka version: 1.3.2
Platform: .net core 2.0

When client sending several packets at once (within a loop) like this:
```
a
bb
ccc
ddd
eeee
fffff
gggggg
hhhhhhh
iiiiiiii
jjjjjjjjj
```

Server will receive incorrect bytes like this:
```
i
jj
iii
jjjj

...
```

**The length of the bytes remain correct but the content is replaced by other bytes.**

Here code sample to reproduce result above:

Server:
```csharp
using Akka.Actor;
using Akka.Configuration;
using Akka.Event;
using Akka.IO;
using System;
using System.Net;

class Program
{
    static void Main(string[] args)
    {
        var config = @"akka 
        {
            stdout-loglevel = INFO
            loglevel = DEBUG
            log-config-on-start = off
            io 
            {
                udp 
                {
                    trace-logging = on
                }
            }
            actor 
            {
                serializers 
                {
                    hyperion = ""Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion""
                }
                serialization-bindings 
                {
                    ""System.Object"" = hyperion
                }
            }
        }";

        using (var system = ActorSystem.Create("test", ConfigurationFactory.ParseString(config)))
        {
            system.ActorOf(Props.Create(() => new UdpServer(new IPEndPoint(IPAddress.Any, 32500))));

            Console.Read();
        }
    }

    public class UdpServer : ReceiveActor
    {
        private readonly ILoggingAdapter _log = Context.GetLogger();
        private readonly IPEndPoint _serverAddr;

        private IActorRef _server;

        public UdpServer(IPEndPoint serverAddr)
        {
            _serverAddr = serverAddr;

            Receive<Udp.Bound>(b => _server = Sender);
            Receive<Udp.Received>(rec => _log.Debug($"{rec.Sender} {rec.Data.ToString()}"));
        }

        protected override void PreStart()
        {
            Context.System.Udp().Tell(new Udp.Bind(Self, _serverAddr));
        }
    }
}
```

Client:
```csharp
using Akka.Actor;
using Akka.Configuration;
using Akka.Event;
using Akka.IO;
using System;
using System.Net;
using System.Text;

class Program
{
    static void Main(string[] args)
    {
        var config = @"akka 
        {
            stdout-loglevel = INFO
            loglevel = DEBUG
            log-config-on-start = off
[...]

PR #3218, opened 2017-12-20
Fixed problem with byte buffer overridden inside UDP listener 

Horusiath:
This fixes a bug reported by @blackclavus in #3210 . Also contains new test case for verification.

### DIFF
--- a/src/core/Akka/IO/UdpListener.cs
+++ b/src/core/Akka/IO/UdpListener.cs
@@ -27,18 +27,13 @@ namespace Akka.IO
     {
         private readonly IActorRef _bindCommander;
         private readonly Bind _bind;
-
         protected readonly ILoggingAdapter Log = Context.GetLogger();
-
-        private IActorRef _selector;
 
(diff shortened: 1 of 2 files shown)


src/core/Akka/IO/UdpListener.cs
```diff
@@ -89,46 +84,41 @@ protected override bool Receive(object message)
...
-            }
             return false;
         }
 
         private void DoReceive(SocketAsyncEventArgs e, IActorRef handler)
         {
             try
             {
-                handler.Tell(new Received(ByteString.FromBytes(e.Buffer, e.Offset, e.BytesTransferred), e.RemoteEndPoint));
+                handler.Tell(new Received(ByteString.CopyFrom(e.Buffer, e.Offset, e.BytesTransferred), e.RemoteEndPoint));
```
Horusiath: The bug itself was here - we were creating `ByteString` directly from buffer assigned to a socket and the we continued to receive more bytes (no copying). When they arrived, they were overriding internal content of byte string we already passed to the user.

Aaronontheweb:
nice work @Horusiath

PR #3219, opened 2017-12-21
Logging performance optimization + ease of use fix

Aaronontheweb:
The performance optimization is simple: don't dynamically create a new "ERROR", "INFO" string every time. Just cache one for each log level and re-use it.

The main thing here is a bug I've run into in my own code and in some end-user code I've reviewed: `FormatException`s thrown when the logging string itself isn't formatted correctly. When this error occurs here's the stack trace you get:

```
Cause: System.FormatException: Input string was not in a correct format.
   at System.Text.StringBuilder.FormatError()
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(String format, Object[] args)
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(String format, Object[] args)
   at Akka.Event.LogEvent.ToString()
   at Akka.TestKit.Xunit2.Internals.TestOutputLogger.<>c__DisplayClass0_0.<.ctor>b__0(Debug e)
   at lambda_method(Closure , Object , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 )
   at Akka.Tools.MatchHandler.PartialHandlerArgumentsCapture`6.Handle(T value)
```

Not terribly helpful. Doesn't tell you anything about the log message that caused this error, thus you either have to download a bunch of debug symbols and set breakpoints inside Akka.NET or inside the `System.String` internals (I did the latter) and try to determine which log message actually caused this exception in the first place.

In this PR I introduce some code that isn't exactly beautiful, but it untangles the log message that threw the exception into an understandable error message and prints that instead of this same stack trace.

### DIFF
--- a/src/core/Akka/Event/LogEvent.cs
+++ b/src/core/Akka/Event/LogEvent.cs
@@ -6,11 +6,38 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using Akka.Actor;
 
 namespace Akka.Event
 {
+    /// <summary>
+    /// INTERNAL API.
+    /// 
+    /// Avoids redundant parsing of log levels and other frequently-used log items
+    /// </summary>
+    internal static class LogFormats
+    {
+        public static readonly IReadOnlyDictionary<LogLevel, string> PrettyPrintedLogLevel;
+
+        static LogFormats()
+        {
+            var dict = new Dictionary<LogLevel, string>();
+            foreach(LogLevel i in Enum.GetValues(typeof(LogLevel)))
+            {
+                dict.Add(i, Enum.GetName(typeof(LogLevel), i).Replace("Level", "").ToUpperInvariant());
+            }
+            PrettyPrintedLogLevel = dict;
+        }
+
+        public static string PrettyNameFor(this LogLevel level)
+        {
+            return PrettyPrintedLogLevel[level];
+        }
+    }
+
     /// <summary>
     /// This class represents a logging event in the system.
     /// </summary>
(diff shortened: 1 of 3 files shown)


src/core/Akka/Event/LogEvent.cs
```diff
@@ -62,7 +89,7 @@ protected LogEvent()
         /// <returns>A <see cref="System.String" /> that represents this LogEvent.</returns>
         public override string ToString()
         {
-            return string.Format("[{0}][{1}][Thread {2}][{3}] {4}", LogLevel().ToString().Replace("Level", "").ToUpperInvariant(), Timestamp, Thread.ManagedThreadId.ToString().PadLeft(4, '0'), LogSource, Message);
+            return string.Format("[{0}][{1}][Thread {2}][{3}] {4}", LogLevel().PrettyNameFor(), Timestamp, Thread.ManagedThreadId.ToString().PadLeft(4, '0'), LogSource, Message);
```
Horusiath: Given 4 log levels, it will be much faster to make a switch statement on that enum, than trying to load ldf for dictionary field, then get enum hash, find the bucket inside the hash, then do enum equality comparison, and eventually load the string itself (hopefully still lying on the same cache line)

Aaronontheweb: That's a good call; we haven't changed the number of log levels ever so it's safe to do that. I've been in the midst of writing code like this for a brand new project where the enums are being developed regularly so that's where my head has been at.

PR #3221, opened 2017-12-21
Update DeviceGroupQuery.cs

nsanitate:
Fixed timing of the first change of behaviour, moved from OnReceive to Constructor method

### DIFF
--- a/docs/examples/Tutorials/Tutorial4/DeviceGroupQuery.cs
+++ b/docs/examples/Tutorials/Tutorial4/DeviceGroupQuery.cs
@@ -18,6 +18,7 @@ namespace Tutorials.Tutorial4
         {
             private ICancelable queryTimeoutTimer;
 
+            #region query-state
             public DeviceGroupQuery(Dictionary<IActorRef, string> actorToDeviceId, long requestId, IActorRef requester, TimeSpan timeout)
             {
                 ActorToDeviceId = actorToDeviceId;
@@ -26,6 +27,8 @@ namespace Tutorials.Tutorial4
                 Timeout = timeout;
 
                 queryTimeoutTimer = Context.System.Scheduler.ScheduleTellOnceCancelable(timeout, Self, CollectionTimeout.Instance, Self);
+                
+                Become(WaitingForReplies(new Dictionary<string, ITemperatureReading>(), new HashSet<IActorRef>(ActorToDeviceId.Keys)));
             }
 
             protected override void PreStart()
@@ -47,12 +50,6 @@ namespace Tutorials.Tutorial4
             public long RequestId { get; }
             public IActorRef Requester { get; }
             public TimeSpan Timeout { get; }
-
-            #region query-state
-            protected override void OnReceive(object message)
-            {
-                Become(WaitingForReplies(new Dictionary<string, ITemperatureReading>(), new HashSet<IActorRef>(ActorToDeviceId.Keys)));
-            }
 
             public UntypedReceive WaitingForReplies(
                 Dictionary<string, ITemperatureReading> repliesSoFar,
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Thanks @nsanitate !

nsanitate:
@Aaronontheweb you can try the patch cloning [https://github.com/nsanitate/iot-app](https://github.com/nsanitate/iot-app) and starting tests. If you follow the tutorial changing behaviour in OnResolve method instead of in Constructor, all tests in DeviceGroupQueryTest class will fail.

docs/examples/Tutorials/Tutorial4/DeviceGroupQuery.cs
```diff
@@ -26,6 +27,8 @@ public DeviceGroupQuery(Dictionary<IActorRef, string> actorToDeviceId, long requ
                 Timeout = timeout;
 
                 queryTimeoutTimer = Context.System.Scheduler.ScheduleTellOnceCancelable(timeout, Self, CollectionTimeout.Instance, Self);
+                
+                Become(WaitingForReplies(new Dictionary<string, ITemperatureReading>(), new HashSet<IActorRef>(ActorToDeviceId.Keys)));
```
Aaronontheweb: You can just change this to `WaitingForReplies` without the `Become` syntax here

docs/examples/Tutorials/Tutorial4/DeviceGroupQuery.cs
```diff
@@ -48,12 +51,6 @@ protected override void PostStop()
             public IActorRef Requester { get; }
             public TimeSpan Timeout { get; }
 
-            #region query-state
-            protected override void OnReceive(object message)
-            {
-                Become(WaitingForReplies(new Dictionary<string, ITemperatureReading>(), new HashSet<IActorRef>(ActorToDeviceId.Keys)));
```
Aaronontheweb: Yeah, this is wrong. Good catch.

docs/examples/Tutorials/Tutorial4/DeviceGroupQuery.cs
```diff
@@ -116,7 +113,12 @@ protected override void OnReceive(object message)
                     Context.Become(WaitingForReplies(repliesSoFar, stillWaiting));
                 }
             }
-            #endregion
+            #endregion   
+
+            protected override void OnReceive(object message)
```
Aaronontheweb: We should probably just refactor this actor to do all of its actual message handling inside `OnReceive` instead of having this behavior-switching. We could just make the state private members. Although this is a cool way of showing how you can use closures to build receive methods, which is a valid technique in practice.

Aaronontheweb:
@nsanitate what do you think of my comments?

docs/examples/Tutorials/Tutorial4/DeviceGroupQuery.cs
nsanitate: I think the purpose of this tutorial was to teach `using the Context.Become() trick instead of just making the repliesSoFar and stillWaiting structures mutable fields of the actor`

docs/examples/Tutorials/Tutorial4/DeviceGroupQuery.cs
nsanitate: The output of `WaitingForReplies` is a factory method that will be used as input of `Context.Become` method. The author use it in the same way at row 116.

docs/examples/Tutorials/Tutorial4/DeviceGroupQuery.cs
Aaronontheweb: Got it. In that case, we should leave it as-is. Want to teach the concept.

nsanitate: So, if we change strategy in this actor, we have to change the documentation too. I think this is a good way for the tutorial.

PR #3227, opened 2017-12-27
Throw on generic Ask if response was Status.Failure

Horusiath:
Right now, when we're using a generic variant of `Ask<>` method, we are solving it by simply casting between Task results (from `object` to our expected `TResult` type). 

The problem here is that, by default Akka.NET solves failure cases to be returned to sender using `Status.Failure` message. In this case, if we'll try to cast `Failure` to generic type `TResult`, we'll end with `InvalidCastException`, swallowing the actual error (carried as a payload of `Failure` message) that happened underneath.

What this PR does, is simplifying casting method (I'm using `async` instead of manual cross task casting), and it checks for `Status.Failure` message - if it was returned and the `TResult` itself is not `Failure`, then we are simply rethrowing that error (while still preserving the stack trace). Additionally I've added 2 tests to verify that behavior. For non-generic variant of `Ask` method everything works as previously - since we cannot guarantee, that user didn't expected `Failure` message to be returned.

This should help an error readability a lot.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -1648,13 +1648,13 @@ namespace Akka.Actor
     public abstract class Status
     {
         protected Status() { }
-        public class Failure : Akka.Actor.Status
+        public sealed class Failure : Akka.Actor.Status
         {
             public readonly System.Exception Cause;
             public Failure(System.Exception cause) { }
             public override string ToString() { }
         }
-        public class Success : Akka.Actor.Status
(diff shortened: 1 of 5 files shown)


maxcherednik:
@Horusiath @Aaronontheweb I was wondering if we are working on the same thing. I was hoping to remove this CastTask at all - since this is another thing which is hiding real exceptions. Async/Await makes our code 'sync' for the readers.

Horusiath:
@maxcherednik sorry, but I might be out of track. Do you have some PR for that already?

maxcherednik:
@Horusiath nearly. Apart from fixing, was about to add some documentation. It's this guy: #3220

Horusiath:
Closing this one. Eventually it will become part of max PR.

PR #3235, opened 2017-12-30
Retry stage

vasily-kirichenko:


src/core/Akka.Streams/Dsl/Retry.cs
```diff
@@ -0,0 +1,370 @@
...
+        /// </summary>
+        /// <param name="flow">the flow to retry</param>
+        /// <param name="retryWith">if output was failure, we can optionaly recover from it,
+        /// and retry with a new pair of input and new state we get from this function.</param>
+        /// <typeparam name="TI">input elements type</typeparam>
+        /// <typeparam name="TS">state to create a new `(I,S)` to retry with</typeparam>
+        /// <typeparam name="TO">output elements type</typeparam>
+        /// <typeparam name="TM">materialized value type</typeparam>
+        [ApiMayChange]
+        public static IGraph<FlowShape<Tuple<TI, TS>, Tuple<Result<TO>, TS>>, TM> Create<TI, TS, TO, TM>(
```
Horusiath: I know that his signature is quite long, but maybe we should stick to convention for generic type parameters (TIn, TState, TOut, TMat)?

Aaronontheweb:
@marcpiechura looks like we have a conflict in the `Framing` DSL?

marcpiechura:
fixed conflicts

Aaronontheweb:
@marcpiechura have hard failures on all of the framing specs now. Definitely something wrong there.

src/core/Akka.Streams/Dsl/Framing.cs
```diff
@@ -328,15 +328,7 @@ public DelimiterFramingStage(ByteString separatorBytes, int maximumLineBytes, bo
...
                 _allowTruncation = allowTruncation;
-
-                Shape = new FlowShape<ByteString, ByteString>(_in, _out);
             }
-            
-            private readonly Inlet<ByteString> _in = new Inlet<ByteString>("DelimiterFraming.in");
-
-            private readonly Outlet<ByteString> _out = new Outlet<ByteString>("DelimiterFraming.in");
-
-            public override FlowShape<ByteString, ByteString> Shape { get; }
```
marcpiechura: Not sure why I added those properties since they're already provided by the base class...

PR #3237, opened 2017-12-31
Improve type safety for custom graph stages

### LINKED ISSUE #3231: Make writing custom stream stages more type safe
As pointed out in #3225 it's quite easy to skip the type checking by using one of the overloads which take an `Inlet` instead of an `Inlet<T>` , e.g.
```csharp
var outlet = new Outlet<int>("someOutlet");
Push(outlet, "hello world");
``` 

Proposal would be to restrict the public api to methods which only take  `Inlet<T>` or `Outlet<T>` and keep the ones without type argument internal, if we need them at all.

cc @Horusiath

marcpiechura:
In order to make the API consistent I've used `Inlet<T>` and `Outlet<T>` on all public methods, even if they doesn't really need it, e.g. `IsAvailable` . 
For those methods which are used internally and where no type argument could be provided, I've created a private version without type argument, e.g. `Pull`.

Since this is quite a breaking change we need to decide if we wait for the next major version or release it with the next minor since it would prevent some ugly runtime exceptions. 

closes #3231


src/core/Akka.Streams/KillSwitch.cs
```diff
@@ -179,25 +176,21 @@ public Logic(Task terminationSignal, UniqueBidiKillSwitchStage<TIn1, TOut1, TIn2
...
+            private Outlet<TIn> Out1 { get; } = new Outlet<TIn>("KillSwitchBidi.out1");
 
-            private Inlet<TIn2> In2 { get; } = new Inlet<TIn2>("KillSwitchBidi.in2");
+            private Inlet<TOut> In2 { get; } = new Inlet<TOut>("KillSwitchBidi.in2");
 
-            private Outlet<TOut2> Out2 { get; } = new Outlet<TOut2>("KillSwitchBidi.out2");
+            private Outlet<TOut> Out2 { get; } = new Outlet<TOut>("KillSwitchBidi.out2");
 
-            public override BidiShape<TIn1, TOut1, TIn2, TOut2> Shape { get; }
+            public override BidiShape<TIn, TIn, TOut, TOut> Shape { get; }
```
Horusiath: Are you sure, this is right? It looks like less generic version.

marcpiechura: Yup, the previous more generic version was actually a „bug“, only prevented from happening due to the fact that the public api already restricts it to two different types. 
I got a compilation error at the following line 
„Push(killSwitch.Out2, Grab(killSwitch.In2)),“
which makes it obvious that both, the inlets and outlets, have to be of the same type.

src/core/Akka.Streams/KillSwitch.cs
```diff
@@ -50,7 +50,7 @@ public static class KillSwitches
         /// <typeparam name="TOut1">TBD</typeparam>
         /// <returns>TBD</returns>
         public static IGraph<BidiShape<TIn1, TIn1, TOut1, TOut1>, UniqueKillSwitch> SingleBidi<TIn1, TOut1>
-            () => UniqueBidiKillSwitchStage<TIn1, TIn1, TOut1, TOut1>.Instance;
+            () => UniqueBidiKillSwitchStage<TIn1, TOut1>.Instance;
```
marcpiechura: This is the public api I’m referring to in the comment below

src/core/Akka.Streams/KillSwitch.cs
marcpiechura: The type names are a bit confusing, T1 instead of TIn and T2 instead of TOut would make more  sense, so I’ll update them when if fix the other issue with the performance test project.

Aaronontheweb:
@marcpiechura looks like we have some merge conflicts with the API approval

marcpiechura:
fixed the conflicts

Aaronontheweb:
@marcpiechura 

```
NBench.NBenchException: Error occurred during $Akka.Streams.Tests.Performance.InterpreterBenchmark+Graph_interpreter_100k_elements_with_1_identity RUN. ---> System.Reflection.TargetException: Non-static field requires a target.
   at System.Reflection.RtFieldInfo.CheckConsistency(Object target)
   at System.Reflection.RtFieldInfo.InternalSetValue(Object obj, Object value, BindingFlags invokeAttr, Binder binder, CultureInfo culture, StackCrawlMark& stackMark)
   at System.Reflection.RtFieldInfo.SetValue(Object obj, Object value, BindingFlags invokeAttr, Binder binder, CultureInfo culture)
   at Akka.Streams.Tests.Performance.InterpreterBenchmark.GraphDataSource`1..ctor(String toString, T[] data)
   at Akka.Streams.Tests.Performance.InterpreterBenchmark.<>c__DisplayClass4_0.<Execute>b__0(TestSetup setup, Func`1 lastEvents)
   at NBench.Sdk.Benchmark.WarmUp()
```

Aaronontheweb:
Error was from `Performance.InterpreterBenchmark+Graph_interpreter_100k_elements_with_1_identity`

PR #3246, opened 2018-01-03
extended timeframe to fix Akka.Remote.Ask_does_not_deadlock spec

Aaronontheweb:
Issue with this spec failure is that the timeframe to establish a brand new connection and complete the task was just too short before (1.5s;) it could randomly fail depending on what else was running on the build server at the time. Upped it to 5s.

### DIFF
--- a/src/core/Akka.Remote.Tests/RemotingSpec.cs
+++ b/src/core/Akka.Remote.Tests/RemotingSpec.cs
@@ -152,7 +152,7 @@ namespace Akka.Remote.Tests
         public void Remoting_must_support_remote_lookups()
         {
             here.Tell("ping", TestActor);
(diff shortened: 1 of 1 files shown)


maxcherednik:
@Aaronontheweb it's actually god that this test is failing from time to time.
I believe it's not the timeout here, but rather race condition - it will be failing even with relatively big timeouts(10, 20, 30 seconds).

ps. You did not change the Ask_does_not_deadlock timeout.

On top of this, I don't really like this test case. We probably need to review the internals of the implementation.

Aaronontheweb:
> ps. You did not change the Ask_does_not_deadlock timeout.

I did - changed it to 5.0s. Need to expand the source code view to see it. But even after looking at it, the spec still failed. So no, it's not the timeout that's the cause. It's probably some other issue with the code.

Issue #3247, opened 2018-01-03
Correct errors in Akka.NET Serilog Documentation


Aaronontheweb:
The [Serilog integration page](http://getakka.net/articles/utilities/serilog.html) has a number of small issues with it that make the code un-runnable right out of the gate:

* [x] Need to install a specific logging sink to get the example running
* [x] Need to install another Serilog package to read in the configuration settings from file

And other minor issues like that. We should address all of these in order to provide a "it just works" experience right out of the box.

wtarr:
Hi, I see what you mean about the example not running out of the box.  Serilog has various sinks available on nuget for depending on what you want to do.  In this case it is using a _colored console sink_ as a dependency in the code sample.

So the following is required: Install-Package Serilog.Sinks.ColoredConsole

As for the configuration, you also need the "Serilog.Settings.AppSettings" nuget package, there is a link on the page to the serilog documentation, but it should probably state this package requirement up front.

Install-Package Serilog.Settings.AppSettings

If no one else (first time contributor) wants to pick this up, I can correct it as I put allot of that page together originally.

Danthar:
Go for it

stijnherreman:
The default Console logger uses this format when using :

```
.WriteTo.Console(
    outputTemplate: "[{Timestamp:HH:mm:ss} {Level:u3}] {Message:lj}{NewLine}{Exception}")
```

Akka.NET users will probably want something like this instead:

Aaronontheweb:
closed via https://github.com/akkadotnet/akka.net/pull/3254

PR #3254, opened 2018-01-06
Address Serilog documentation issues raised as per #3247

wtarr:
- Address serilog colored console requirement.
- Remove requirement for "SerilogLogMessageFormatter" as code examples
appear to run without specifying the formatter.
- Add an example of how an output template is configured for the
extensions demo.
- Remove use of serilog application configuration in code from the hocon
example, but mention it as a feature of serilog.


docs/articles/utilities/serilog.md
```diff
@@ -6,14 +6,20 @@ title: Serilog
...
 PM> Install-Package Akka.Logger.Serilog
 ```
 
-This will also install the required Serilog packages.
+## Example
+
+The following example uses Serilogs __Colored Console__ sink available via nuget, there are wide range of other sinks available depending on your needs, for example a rolling log file sink.  See serilogs documentation for details on these.
+
+```
+PM> Install-Package Serilog.Sinks.ColoredConsole
```
Aaronontheweb: Good

docs/articles/utilities/serilog.md
```diff
@@ -35,9 +43,9 @@ var log = Context.GetLogger();
 log.Info("The value is {0}", counter);
 ```
 
-To log using Serilog syntax you need to use the `SerilogLogMessageFormatter`:
+Or alternatively
 ```csharp
-var log = Context.GetLogger(new SerilogLogMessageFormatter());
```
Aaronontheweb: Why remove this? Isn't that necessary to get the semantic Serilog-style formatting to work?

wtarr: Hi, I found when I ran the examples, that it did not seem to make a difference in output whether the formatter was included or not.  I felt if it was not necessary or no longer necessary, it was probably best to remove from the example.  It seems allot of this structuring of output is handled in serilog - https://github.com/serilog/serilog/wiki/Structured-Data#default-behaviour - some of this page seems to apply to 2.6.0 but the following seems to apply for 2.5.0

```csharp
Log.Information("The value is {counter}", counter);
```

wtarr: funnily, the following works too

```csharp

private void Handle(int count)
{
	log.Info("Count is {a} {b} {c}", count, 2, true);
}
```
> 2018-01-08 18:23:12 INF Count is 1 2 True

Aaronontheweb: I see.

Danthar: Correct. The SerilogLogMessageFormatter is no longer needed.

Danthar:
+1 on @stijnherreman comment. And it is exactly the reason why serilog plugins have not been mentioned in the docs. I am not a fan of pointing people to a specific nuget package in this case. Because its an external project, and we don't want to update our docs everytime one of their plugin's moves somewhere else.

Mention the plugin/extensions name, sure. But don't specify an exact nuget package, imho.

wtarr:
If you wish I can replace the colored console with the standard console sink as a sink is necessary for output.  I would hope that the console sink would be considered a standard sink.

Danthar:
Thats also an option. Not sure if the normal console sink is built in though, i never use it :P Either way im ok with either option.

wtarr:
To the best of my knowledge, one sink is necessary.  There own page mention these two packages as part of there basic setup (https://github.com/serilog/serilog/wiki/Getting-Started).

`PM> Install-Package Serilog` and `PM> Install-Package Serilog.Sinks.Console`

I use console myself mostly for testing, but rolling log file in production, it has nice retention handling built in.

PR #3263, opened 2018-01-11
ClusterSingletonManager should ignore FSM events during shutdown

Aaronontheweb:
Port of https://github.com/akka/akka/pull/24236 - solves the same issue in `ClusterSingletonManager`.

Note, we should also keep an eye on https://github.com/akka/akka/issues/24113, which addresses a related race condition inside the `ShardCoordinator`.


Aaronontheweb:
Dang, need to run API approvals and add a copyright header to the new spec I added. Totally forgot to do that.

Aaronontheweb:
Looks like there's an issue with `CoordinatedShutdown` timing out, which is an unrelated issue - but nonetheless might be a genuine bug. Going to take a look at the logs.

Aaronontheweb:
closed via #3551

PR #3266, opened 2018-01-11
Update KeepMajority SplitBrainResolver strategy to avoid exception

mrrd:
* Fix exception that can occur when remaining and unreachable variables both have zero elements
* tidy up if else logic as per Refactor recommendations

This fixes a bug which has been experienced in production where remaining and unreachable have zero elements which causes an exception when trying to call .First() to get the oldest in the else (which obviously does not exist).

Unfortunately I do not have the time right now to write a test for this due to work constraints, do with this as you will.


Aaronontheweb:
Not sure what happened to TeamCity there, but the windows-unit-test stage for this PR did actually pass.

PR #3282, opened 2018-01-20
Use value tuples in Streams

vasily-kirichenko:


Aaronontheweb:
@vasily-kirichenko not sure why exactly, but it looks like the Mono compilation stage is having some trouble with these changes. I think we're running an older version of Mono 5.0* - should we upgrade the Docker image we're using for compilation there to see if that solves the issue? Windows compilation appears to be working just fine.

Aaronontheweb:
.NET Core compilation on Linux works fine too.

vasily-kirichenko:
> I think we're running an older version of Mono 5.0* - should we upgrade the Docker image we're using for compilation there to see if that solves the issue?

I've never used Mono, and I've not found any mentions of Docker in build.fsx. How to do the upgrade?

Aaronontheweb:
> I've never used Mono, and I've not found any mentions of Docker in build.fsx. How to do the upgrade?

We run the `build.fsx` inside the Docker container built using the Dockerfile here for Mono builds: https://github.com/akkadotnet/dockerfiles

Those images eventually get pushed up here: https://hub.docker.com/r/akkadotnet/mono-base/

We use Docker for a few things in our CI system, mostly hosting the Mono environment and running Akka.Persistence integration tests, but that stuff mostly happens invisibly on our TeamCity agents and isn't exposed directly in the build scripts.

vasily-kirichenko:
@Aaronontheweb https://github.com/akkadotnet/dockerfiles/pull/7

marcpiechura:
@vasily-kirichenko as @Aaronontheweb mentioned in the issue, this is quite a breaking change and should therefore target the v1.4 branch. 

BTW, thanks a lot for taking this one 😄

vasily-kirichenko:
@marcpiechura done

src/core/Akka.Streams/Implementation/ActorMaterializerImpl.cs
```diff
@@ -123,51 +126,47 @@ public ActorMaterializerSession(ActorMaterializerImpl materializer, IModule topL
...
+                    AssignPort(stage1.Out, UntypedPublisher.FromTyped(processor));
                     materializedValues.Add(atomic, materialized);
                 }
                 //else if (atomic is TlsModule)
                 //{
                 //})
                 else if (atomic is GraphModule)
                 {
-                    var graph = (GraphModule) atomic;
+                    var graph = (GraphModule)atomic;
```
marcpiechura: Why not also use assignment in the if statement ?

src/core/Akka.Streams/Implementation/ActorMaterializerImpl.cs
```diff
@@ -123,51 +126,47 @@ public ActorMaterializerSession(ActorMaterializerImpl materializer, IModule topL
...
                     MaterializeGraph(graph, effectiveAttributes, materializedValues);
                 }
                 else if (atomic is GraphStageModule)
                 {
-                    var stage = (GraphStageModule) atomic;
+                    var stage = (GraphStageModule)atomic;
```
marcpiechura: As above

src/core/Akka.Persistence.TCK/PluginSpec.cs
```diff
@@ -43,19 +43,12 @@ public void Subscribe<T>(IActorRef subscriber)
             Sys.EventStream.Subscribe(subscriber, typeof (T));
         }
 
-        /// <inheritdoc/>
-        public void Dispose()
-        {
-            Dispose(true);
-            GC.SuppressFinalize(this);
-        }
-
```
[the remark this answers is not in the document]
vasily-kirichenko: Yes. The same code is in the base class's Dispose method.

vasily-kirichenko:
@marcpiechura all done.

Aaronontheweb:
Going to pull the plug on our Mono builds, I think; been talking with @Horusiath about dropping support for it since it affects PRs like this one and our `IObservable` integration. Going to restore our .NET Core on Linux builds and use that as our canary for x-plat compatibility on POSIX operating systems.

marcpiechura:
@Aaronontheweb any timeframe for removing mono? Would like to merge this one and rebase my work for Streams 2.5 on it

Aaronontheweb:
I'll look at resolving the conflicts here and getting this pulled into v1.4 now that v1.4 has been updated with the latest sources.

Aaronontheweb:
Need to give this another look for the next Akka.NET v1.4.0 beta

Aaronontheweb:
Retargeting the `dev` branch here...

Aaronontheweb:
The merge conflicts are a little nasty - I think I'll need to do this offline.

Aaronontheweb:
closed via https://github.com/akkadotnet/akka.net/pull/3973

Issue #3261, opened 2018-01-10
Multi-Data Center Support in Akka.Cluster


Aaronontheweb:
This is something that's been in the works on the JVM for about 6-7 months and we've had some requests for it over the past couple of years on the Akka.NET project. I'm interested in making this happen as part of the Akka.NET v1.4 or v1.5 milestones.

feitzi:
is there any target date for data-center support in akka.net (maybe during 2019?)

Aaronontheweb:
We dropped it for the 1.4.0 release coming up because no one other than @Horusiath (the author) has had the bandwidth to review the code / finish verifying that the multi-node specifications work and we're not in a position to do that immediately, but we should be able to do it this year.

feitzi:
thanks aaron, this year sounds good 👍  
Can we support you in this topic?

Horusiath:
This PR is pretty much done - it's working and tests on .net full framework are passing. The issue is in .NET Core multi-node tests, which are failing for some reason, but since MNTK runner is not working locally for me I'm unable to find the issue.

tcs-devel:
Do you have any plans for a 1.5 release, or any release including this feature?
Thanks!

tcs-devel:
Any plans when 1.5.0 / 2.0.0 with this feature should be available?

tcs-devel:
Just checking if there are any further plans for this?
Thank you!

Aaronontheweb:
@tcs-devel yep, we do plan on implementing this - I'm not sure if it's in-scope for v1.6 or not at this rate simply because there hasn't been a huge amount of demand for it and it is a fairly significant, complicated-to-test feature.

I personally think it's a good idea for disaster recovery scenarios though, so I would like to see it added and evangelized. I think it's one of those things where demand for it will increase once it's available as a viable option.

PR #3284, opened 2018-01-21
[WIP] Multi data center support

Horusiath:
This is port of Multi-DC support from https://doc.akka.io/docs/akka/2.5/cluster-dc.html


Aaronontheweb:
Issue https://github.com/akkadotnet/akka.net/issues/3261

Horusiath:
Can we rerun test for this PR somehow?

Aaronontheweb:
Issue might be the age of the PR - added before Azure DevOps was added. If updating it with the latest doesn't trigger it, might need to close this PR and just create a new one. Not 100% sure.

ismaelhamed:
@Horusiath @Aaronontheweb **Multi-DC Persistence** is a commercial addon, so are we going to have a custom implementation then?

Aaronontheweb:
I wasn't aware of that. 

Putting on my Petabridge hat: at the moment, I'm not sure - we haven't been trying to follow Lightbend's roadmap when it comes to proprietary technology because the way Akka is consumed in .NET looks quite different than it does on the JVM. Our technology is consumed more by enterprises (energy, finance, healthcare, transportation) and whereas Lightbend also has to support large-scale ecommerce, social media, advertising, etc clients who really do need similarly large-scale features like multi-DC persistence. We'd love to have that type of problem too as adoption of Akka.NET continues to expand rapidly ;)

In terms of what Petabridge will support, that comes down entirely to economics - features that are expensive to support and are only needed by a small number of very high-utilization users should absolutely be proprietary, as those users should foot the bill for the engineering and QA time that goes into maintaining those. 

On principle I don't want to co-mingle what Petabridge and the Akka.NET project, but - necessarily, one of the reasons why people adopt Akka.NET in the first place is precisely because _we do support it_ and we are often the ones who foot the maintenance costs for the project. 

Ultimately we have to strike a balance between what's sustainable for both the company + project and the users' best interests. That's a balancing act that we're still working on figuring out - so far Phobos is the only proprietary component but that will probably change in the future. 

As far as implementation goes, we'll need to get the serialization system fully standardized onto the Protobuf format, per https://github.com/akkadotnet/akka.net/issues/3811, so we can introduce features like this in a forward-compatible manner downstream. Obviously this will need to be merged in too before we can start on that.

ismaelhamed:
It makes total sense. I was just wondering, given you guys have put a lot of effort into this PR, what would be the use cases for **Multi-DC** support without **Multi-DC Persistence**? And this is just my ignorance talking now.

Horusiath:
@ismaelhamed @Aaronontheweb  Multi-DC Persistence uses different technique as it allows to have mutliple incarnations of the same logical entity in different data centers. It uses state representation as CRDTs for conflict-avoidance (similar to our DistributedData module).

Mutli-DC persistence is not necessary tho - IMHO the most important part is to have ensure that cluster sharding (ddata-based one) and split-brain resolvers are multi-dc aware. You don't need multi-dc persistence features in many scenarios.

Another thing, that I'm working over in very slow pace is port of [Eventuate](https://github.com/horusiath/eventuate.net) - it's a persistence library (written by original Akka.Persistence author) that introduces features like replicated event journals and gives comparable features, that multi-dc persistence provides - in fact it was inspiration for multi-dc persistence on the first place.

PR #3287, opened 2018-01-21
Implement PartitonHub

### LINKED ISSUE #3111: Implement PartitionHub
http://doc.akka.io/docs/akka/current/scala/stream/stream-dynamic.html#using-the-partitionhub

https://github.com/akka/akka/pull/23102

marcpiechura:
closes #3111

### DIFF
--- a/src/core/Akka.Streams/Dsl/Hub.cs
+++ b/src/core/Akka.Streams/Dsl/Hub.cs
@@ -13,7 +13,6 @@ using System.Linq;
 using System.Threading.Tasks;
 using Akka.Annotations;
 using Akka.Streams.Stage;
-using Akka.Streams.Util;
 using Akka.Util;
 using Akka.Util.Internal;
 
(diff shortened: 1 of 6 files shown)


src/core/Akka.Streams/Dsl/Hub.cs
```diff
@@ -1169,4 +1169,707 @@ public BroadcastHub(int bufferSize)
...
+                if (_queues.TryRemove(id, out var queue))
+                    _totalSize.AddAndGet(-queue.Size);
+            }
+        }
+
+        #endregion
+
+        #region internal classes
+
+        private interface IConsumerEvent { }
```
Horusiath: I don't consider this a mistake, but I'm starting to think, that in the future, we should favor tagged unions over this approach - no in every case, but for sure in cases like several singletons and one type with a error field. Example:

```csharp
struct ConsumerEvent 
{
    enum ConsumerEventType { Wakeup, Initialize, HubCompleted }
    
    readonly ConsumerEventType Type;
    readonly Exception Failure;
    ....
}
```

Horusiath:
@marcpiechura It's good, we can merge it. Right now I'm at work, so I cannot restart tests failing tests, but they are not related to the changes in this PR.

PR #3298, opened 2018-01-26
sharding StartProxy replicator used based on configuration

### LINKED ISSUE #3297: DData replicator is always assigned
As @zbynek001 pointed out, after 1.3.3, the Akka.Cluster.Sharding update is no longer binary compatible - the reason for that is that we initialize DistributeData replicator even thou we didn't use it.

The line which causes an issue: https://github.com/akkadotnet/akka.net/blob/e607088d00e60b318897da3f01b8982666a5f6cc/src/contrib/cluster/Akka.Cluster.Sharding/ClusterShardingGuardian.cs#L225

The problem is that I've missed the initialization code. Replicator should not be initialized via explicit `DistributedData.DistributedData.Get(Context.System).Replicator` call, but rather a `Replicator` method defined few lines below: https://github.com/akkadotnet/akka.net/blob/e607088d00e60b318897da3f01b8982666a5f6cc/src/contrib/cluster/Akka.Cluster.Sharding/ClusterShardingGuardian.cs#L252

You can see that error only refers to `StartProxy` command. Standard `Start` command works as designed.

zbynek001:
fixes #3297

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ClusterShardingGuardian.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ClusterShardingGuardian.cs
@@ -222,7 +222,7 @@ namespace Akka.Cluster.Sharding
                     var coordinatorSingletonManagerName = CoordinatorSingletonManagerName(encName);
                     var coordinatorPath = CoordinatorPath(encName);
                     var shardRegion = Context.Child(encName);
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Looks like our new Docker container for Linux had some problems. Going to re-run these builds using the previous one.

PR #3299, opened 2018-01-26
AkkaPduCodec performance fixes [remoting]

Aaronontheweb:
Noticed some low-hanging fruit for performance in the `AkkaPduCodec` fixes while I was working on implementing a custom serializer for Akka.Remote today.

This is the class responsible for serializing 100% of message traffic that passes over each Akka.Remote connection.

### DIFF
--- a/src/core/Akka.Remote/Transport/AkkaPduCodec.cs
+++ b/src/core/Akka.Remote/Transport/AkkaPduCodec.cs
@@ -381,24 +381,20 @@ namespace Akka.Remote.Transport
             }
         }
 
-        private static ByteString _heartbeatPdu;
+        /*
+         * Since there's never any ActorSystem-specific information coded directly
+         * into the heartbeat messages themselves (i.e. no handshake info,) there's no harm in caching in the
+         * same heartbeat byte buffer and re-using it.
+         */
+        private static readonly ByteString HeartbeatPdu = ConstructControlMessagePdu(CommandType.Heartbeat);
 
         /// <summary>
-        /// TBD
+        /// Creates a new Heartbeat message instance.
         /// </summary>
-        /// <returns>TBD</returns>
+        /// <returns>The Heartbeat message.</returns>
         public override ByteString ConstructHeartbeat()
         {
-            /*
-             * Since there's never any ActorSystem-specific information coded directly
-             * into the heartbeat messages themselves (i.e. no handshake info,) there's no harm in caching in the
-             * same heartbeat byte buffer and re-using it.
-             */
-            if (_heartbeatPdu == null)
-            {
-                _heartbeatPdu = ConstructControlMessagePdu(CommandType.Heartbeat);
-            }
-            return _heartbeatPdu;
+            return HeartbeatPdu;
         }
 
         /// <summary>
@@ -549,7 +545,7 @@ namespace Akka.Remote.Transport
             get { return ConstructControlMessagePdu(CommandType.DisassociateQuarantined); }
         }
 
-        private ByteString ConstructControlMessagePdu(CommandType code, AkkaHandshakeInfo handshakeInfo = null)
+        private static ByteString ConstructControlMessagePdu(CommandType code, AkkaHandshakeInfo handshakeInfo = null)
         {
             var controlMessage = new AkkaControlMessage() { CommandType = code };
             if (handshakeInfo != null)
--- a/src/core/Akka.Remote/Transport/AkkaPduCodec.cs
+++ b/src/core/Akka.Remote/Transport/AkkaPduCodec.cs
@@ -228,14 +228,19 @@ namespace Akka.Remote.Transport
         /// <returns>TBD</returns>
         public virtual ByteString EncodePdu(IAkkaPdu pdu)
         {
-            ByteString finalBytes = null;
-            pdu.Match()
-                .With<Associate>(a => finalBytes = ConstructAssociate(a.Info))
-                .With<Payload>(p => finalBytes = ConstructPayload(p.Bytes))
-                .With<Disassociate>(d => finalBytes = ConstructDisassociate(d.Reason))
-                .With<Heartbeat>(h => finalBytes = ConstructHeartbeat());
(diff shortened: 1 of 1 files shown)


src/core/Akka.Remote/Transport/AkkaPduCodec.cs
```diff
@@ -228,14 +228,19 @@ protected AkkaPduCodec(ActorSystem system)
         /// <returns>TBD</returns>
         public virtual ByteString EncodePdu(IAkkaPdu pdu)
         {
-            ByteString finalBytes = null;
-            pdu.Match()
```
Aaronontheweb: Get rid of the old `.Match` extension method, which allocates. Replace with a C# 7 `switch` statement instead. Does not allocate.

src/core/Akka.Remote/Transport/AkkaPduCodec.cs
```diff
@@ -228,14 +228,19 @@ protected AkkaPduCodec(ActorSystem system)
...
-                .With<Associate>(a => finalBytes = ConstructAssociate(a.Info))
-                .With<Payload>(p => finalBytes = ConstructPayload(p.Bytes))
-                .With<Disassociate>(d => finalBytes = ConstructDisassociate(d.Reason))
-                .With<Heartbeat>(h => finalBytes = ConstructHeartbeat());
-
-            return finalBytes;
+            switch (pdu)
+            {
+                case Payload p:
```
Aaronontheweb: Change the order in which we filter for events in Akka.Remote serialization; `Payload` makes up the vast majority of network traffic, since that's all of the messages that aren't part of the Akka.Remote association protocol itself. `Payload` was previously handled second, so we always had one case we needed to fall through each time.

`Heartbeat` is the second most frequently used message type, since these are emitted once per second.

src/core/Akka.Remote/Transport/AkkaPduCodec.cs
```diff
@@ -376,13 +381,24 @@ public override ByteString ConstructDisassociate(DisassociateInfo reason)
...
         /// </summary>
         /// <returns>TBD</returns>
         public override ByteString ConstructHeartbeat()
         {
-            return ConstructControlMessagePdu(CommandType.Heartbeat);
+            /*
+             * Since there's never any ActorSystem-specific information coded directly
+             * into the heartbeat messages themselves (i.e. no handshake info,) there's no harm in caching in the
+             * same heartbeat byte buffer and re-using it.
+             */
```
Aaronontheweb: As explained in the comment above: `Heartbeat` messages don't contain any handshake information, therefore the same "heartbeat" message can be used across many different associations and even many different `ActorSystems` without having to be created from scratch each time. We cache this value in a static mutable field on the serializer now. We could probably make this immutable....

src/core/Akka.Remote/Transport/AkkaPduCodec.cs
```diff
@@ -376,13 +381,20 @@ public override ByteString ConstructDisassociate(DisassociateInfo reason)
             }
         }
 
+        /*
+         * Since there's never any ActorSystem-specific information coded directly
+         * into the heartbeat messages themselves (i.e. no handshake info,) there's no harm in caching in the
+         * same heartbeat byte buffer and re-using it.
+         */
+        private static readonly ByteString HeartbeatPdu = ConstructControlMessagePdu(CommandType.Heartbeat);
```
Aaronontheweb: Was able to make this static and immutable as well.

Aaronontheweb:
I'll let the NBench stuff do its thing here, but I'm also going to run a PingPong benchmark locally for comparison. These optimizations are all fairly "micro," but they should noticeably reduce allocations per-remote-message sent by at least 1.

Aaronontheweb:
Bearing in mind that these values aren't scientific because I'm ran them on my personal computer, which has tons of stuff running in the background:


## Before

λ  dotnet run .\RemotePingPong --framework netcoreapp1.1
Messages sent/received per client: 20000  (2e4)

         1,   20000,     60423,     331.72
         5,  100000,     63573,    1573.50
        10,  200000,     64957,    3079.72
        15,  300000,     64103,    4680.37
        20,  400000,     64382,    6213.41
        25,  500000,     63630,    7858.75
        30,  600000,     63379,    9467.96
Done..

λ  dotnet run .\RemotePingPong --framework net461
Messages sent/received per client: 20000  (2e4)

         1,   20000,     60607,     330.69
         5,  100000,     71995,    1389.54
        10,  200000,     70722,    2828.55
        15,  300000,     73530,    4080.33
        20,  400000,     69554,    5751.79
        25,  500000,     70902,    7052.97
        30,  600000,     69590,    8622.68
Done..

## After

λ  dotnet run .\RemotePingPong --framework netcoreapp1.1
Messages sent/received per client: 20000  (2e4)

         1,   20000,     57804,     346.07
         5,  100000,     63654,    1571.40
        10,  200000,     64600,    3096.74
        15,  300000,     62138,    4828.39
        20,  400000,     63715,    6278.23
        25,  500000,     64070,    7804.18
        30,  600000,     62559,    9591.81
Done.

λ  dotnet run .\RemotePingPong --framework net461
Messages sent/received per client: 20000  (2e4)

         1,   20000,     64517,     310.58
         5,  100000,     73153,    1367.91
        10,  200000,     74627,    2680.02
        15,  300000,     73350,    4090.98
        20,  400000,     72993,    5480.39
        25,  500000,     70732,    7069.06
        30,  600000,     71600,    8380.05
Done..

IMHO, looks to me like the change made a material difference on .NET 4.6.1 but none at all on .NET Core 1.1. I'll wait and see what the build server reports from NBench though.

Aaronontheweb:
@Horusiath looks like there was one in the .NET 4.6.1 pipeline but not .NET Core 1.1. Guess in this instance the amount of objects being allocated wasn't enough to make a material difference to the frequency at which the garbage collector runs. In other words, there's bigger bottlenecks than the `AkkaPduCodec`.

PR #3304, opened 2018-01-30
expose RemoteActorRef APIs for extensibility

Aaronontheweb:
Implementing this change for reasons similar to https://github.com/akkadotnet/akka.net/pull/3300

Also, all of the other `IActorRef` implementations such as `LocalActorRef` and `RepointablActorRef` are similarly extensible.


src/core/Akka.Remote/RemoteActorRefProvider.cs
```diff
@@ -403,13 +403,25 @@ internal IInternalActorRef ResolveActorRefWithLocalAddress(string path, Address
...
         }
 
+        
+        /// <summary>
+        /// Used to create <see cref="RemoteActorRef"/> instances upon deserialiation inside the Akka.Remote pipeline.
+        /// </summary>
+        /// <param name="actorPath">The remote path of the actor on its physical location on the network.</param>
+        /// <param name="localAddress">The local path of the actor.</param>
+        /// <returns>An <see cref="IInternalActorRef"/> instance.</returns>
+        protected virtual IInternalActorRef CreateRemoteRef(ActorPath actorPath, Address localAddress)
```
Aaronontheweb: Needed to expose a factory method here inside the `RemoteActorRefProvider` that for changing the underlying `RemoteActorRef` implementation to a version that can be instrumented. This is stuff we're using for remote propagation of tracing information mostly.

src/core/Akka.API.Tests/CoreAPISpec.ApproveRemote.approved.txt
```diff
@@ -121,6 +121,20 @@ namespace Akka.Remote
         void Remove(T resource);
         void Reset();
     }
+    [Akka.Annotations.InternalApiAttribute()]
+    public interface IRemoteActorRefProvider : Akka.Actor.IActorRefProvider
```
Aaronontheweb: Biggest change in this PR is the introduction of the  `IRemoteActorRefProvider` interface and its adoption throughout the internals of Akka.Remote. This makes it possible to extend Akka.Remote's capabilities (such as, introducing instrumentation) without having to re-implement all of the internals.

This is an `InternalApi` and may be subject to change, but most of the consumption of this interface all occurs internally within Akka.Remote, hence why some of the changes to `Endpoints.cs` et al don't show up in the API diff. This change is binary compatible with previous versions of Akka.NET.

PR #3313, opened 2018-02-03
Discard timers from a previous (FSM) incarnation after restart

ismaelhamed:
Fixes "... a small bug in FSM timers, which could result in that a timer from a previous incarnation was let through to new incarnation after restart".

Original PR: [#22873](https://github.com/akka/akka/pull/22873)


Horusiath:
Do we actually have a tests to verify that behavior? AFAIK Timers from Akka on the JVM have not been ported to .NET.

Aaronontheweb:
@Horusiath talking about the new behavior traits? No, not as far as I can tell.

PR #3322, opened 2018-02-10
Fixing premature pruning of Topics

joshgarnett:
The DistributedPubSubMediator wasn't checking if the TopicActor was actually terminated before pruning it from the bucket.  This can cause problems if a TopicActor is re-suscribed to before being stopped.  The Subscribe message only checks Context.Child, but does not check if the bucket is still valid.  So it was possible to get in a state where subscribes/unsubscribes were succeeding, but any publishes to the topic where being dropped on the floor.

I've also switched from null to ActorRefs.Nobody.  Previously, if a Topic actor had terminated and a publish for that topic was received before the DistributedPubSubMediator did a prune, the publish would throw an exception.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
@@ -220,7 +220,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
             {
                 if (_registry.TryGetValue(_cluster.SelfAddress, out var bucket))
                 {
-                    if (bucket.Content.TryGetValue(remove.Path, out var valueHolder) && valueHolder.Ref != null)
+                    if (bucket.Content.TryGetValue(remove.Path, out var valueHolder) && !Equals(valueHolder.Ref, ActorRefs.Nobody))
                     {
                         Context.Unwatch(valueHolder.Ref);
                         PutToRegistry(remove.Path, ActorRefs.Nobody);
@@ -371,7 +371,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
             Receive<ClusterEvent.IMemberEvent>(_ => { /* ignore */ });
             Receive<Count>(_ =>
             {
-                var count = _registry.Sum(entry => entry.Value.Content.Count(kv => kv.Value.Ref != null));
+                var count = _registry.Sum(entry => entry.Value.Content.Count(kv => !Equals(kv.Value.Ref, ActorRefs.Nobody)));
                 Sender.Tell(count);
             });
             Receive<DeltaCount>(_ =>
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
@@ -223,7 +223,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
                     if (bucket.Content.TryGetValue(remove.Path, out var valueHolder) && !valueHolder.Ref.IsNobody())
                     {
                         Context.Unwatch(valueHolder.Ref);
-                        PutToRegistry(remove.Path, ActorRefs.Nobody);
+                        PutToRegistry(remove.Path, null);
                     }
                 }
             });
@@ -329,7 +329,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
 
                 if (_registry.TryGetValue(_cluster.SelfAddress, out var bucket))
                     if (bucket.Content.TryGetValue(key, out var holder) && terminated.ActorRef.Equals(holder.Ref))
-                        PutToRegistry(key, ActorRefs.Nobody); // remove
+                        PutToRegistry(key, null); // remove
 
                 _buffer.RecreateAndForwardMessagesIfNeeded(key, () => NewTopicActor(terminated.ActorRef.Path.Name));
             });
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
@@ -223,7 +223,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
                     if (bucket.Content.TryGetValue(remove.Path, out var valueHolder) && valueHolder.Ref != null)
                     {
                         Context.Unwatch(valueHolder.Ref);
-                        PutToRegistry(remove.Path, null);
+                        PutToRegistry(remove.Path, ActorRefs.Nobody);
                     }
                 }
             });
@@ -329,7 +329,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
 
                 if (_registry.TryGetValue(_cluster.SelfAddress, out var bucket))
                     if (bucket.Content.TryGetValue(key, out var holder) && terminated.ActorRef.Equals(holder.Ref))
-                        PutToRegistry(key, null); // remove
+                        PutToRegistry(key, ActorRefs.Nobody); // remove
 
                 _buffer.RecreateAndForwardMessagesIfNeeded(key, () => NewTopicActor(terminated.ActorRef.Path.Name));
             });
@@ -549,7 +549,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
                 var bucket = entry.Value;
 
                 var oldRemoved = bucket.Content
-                    .Where(kv => (bucket.Version - kv.Value.Version) > _settings.RemovedTimeToLive.TotalMilliseconds)
+                    .Where(kv => Equals(kv.Value.Ref, ActorRefs.Nobody) && (bucket.Version - kv.Value.Version) > _settings.RemovedTimeToLive.TotalMilliseconds)
                     .Select(kv => kv.Key);
 
                 if (oldRemoved.Any())
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
@@ -220,7 +220,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe
             {
                 if (_registry.TryGetValue(_cluster.SelfAddress, out var bucket))
                 {
-                    if (bucket.Content.TryGetValue(remove.Path, out var valueHolder) && !Equals(valueHolder.Ref, ActorRefs.Nobody))
+                    if (bucket.Content.TryGetValue(remove.Path, out var valueHolder) && !valueHolder.Ref.IsNobody())
                     {
                         Context.Unwatch(valueHolder.Ref);
                         PutToRegistry(remove.Path, ActorRefs.Nobody);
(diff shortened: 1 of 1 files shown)


joshgarnett:
My guess is whomever did the initial port misunderstood how Scala pattern matching works.  From https://github.com/akka/akka/blob/master/akka-cluster-tools/src/main/scala/akka/cluster/pubsub/DistributedPubSubMediator.scala#L835

```
  def prune(): Unit = {
    registry foreach {
      case (owner, bucket) ⇒
        val oldRemoved = bucket.content.collect {
          case (key, ValueHolder(version, None)) if (bucket.version - version > removedTimeToLiveMillis) ⇒ key
        }
        if (oldRemoved.nonEmpty)
          registry += owner → bucket.copy(content = bucket.content -- oldRemoved)
    }
  }
```

`case (key, ValueHolder(version, None))` is actually doing a pattern match on ValueHolder.Ref == None.  If that value is not None it does not do the version compare and will not prune the topic from the bucket.

Aaronontheweb:
Looks like we have a failure in the multi-node `DistributedPubSub.Mediator` specs as a result of this change:

```
[Node1:first][FAIL] Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.DistributedPubSubMediatorSpecs
[Node1:first][FAIL-EXCEPTION] Type: Xunit.Sdk.EqualException
--> [Node1:first][FAIL-EXCEPTION] Message: Assert.Equal() Failure
Expected: 4
Actual:   5
--> [Node1:first][FAIL-EXCEPTION] StackTrace:    at Xunit.Assert.Equal[T](T expected, T actual, IEqualityComparer`1 comparer)
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.<>c__DisplayClass24_0.<AwaitCount>b__0() in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 286
   at Akka.TestKit.TestKitBase.AwaitAssert(Action assertion, Nullable`1 duration, Nullable`1 interval)
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.AwaitCount(Int32 expected) in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 287
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.<DistributedPubSubMediator_must_remove_terminated_users>b__30_0() in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 440
   at Akka.TestKit.TestKitBase.<>c__DisplayClass141_0.<Within>b__0()
   at Akka.TestKit.TestKitBase.Within[T](TimeSpan min, TimeSpan max, Func`1 function, String hint, Nullable`1 epsilonValue)
   at Akka.TestKit.TestKitBase.Within(TimeSpan min, TimeSpan max, Action action, String hint, Nullable`1 epsilonValue)
   at Akka.TestKit.TestKitBase.Within(TimeSpan max, Action action, Nullable`1 epsilonValue)
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.DistributedPubSubMediator_must_remove_terminated_users() in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 442
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.DistributedPubSubMediatorSpecs() in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 299
```

I'll take a look at the source in a moment here and see what's up

Aaronontheweb:
(Could be that the spec itself implements the same "misinterpretation" of the Scala code as the underlying source did)

Aaronontheweb:
This appears to be the offending line of code: 

https://github.com/akkadotnet/akka.net/blob/72d281d55878ee5ec623766de8db6102076abfe9/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/PublishSubscribe/DistributedPubSubMediatorSpec.cs#L430-L442

Should see the number of publishers drop from 5 to 4 with this change, but that's no longer the case.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -220,10 +220,10 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
             {
                 if (_registry.TryGetValue(_cluster.SelfAddress, out var bucket))
                 {
-                    if (bucket.Content.TryGetValue(remove.Path, out var valueHolder) && valueHolder.Ref != null)
+                    if (bucket.Content.TryGetValue(remove.Path, out var valueHolder) && !Equals(valueHolder.Ref, ActorRefs.Nobody))
```
Aaronontheweb: Use the `IsNobody` extension method here IMHO. It'll perform a null check and this reference comparison to `ActorRefs.Nobody` as well.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -371,7 +371,7 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
             Receive<ClusterEvent.IMemberEvent>(_ => { /* ignore */ });
             Receive<Count>(_ =>
             {
-                var count = _registry.Sum(entry => entry.Value.Content.Count(kv => kv.Value.Ref != null));
+                var count = _registry.Sum(entry => entry.Value.Content.Count(kv => !Equals(kv.Value.Ref, ActorRefs.Nobody)));
```
Aaronontheweb: `IsNobody` again

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -329,7 +329,7 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
 
                 if (_registry.TryGetValue(_cluster.SelfAddress, out var bucket))
                     if (bucket.Content.TryGetValue(key, out var holder) && terminated.ActorRef.Equals(holder.Ref))
-                        PutToRegistry(key, null); // remove
+                        PutToRegistry(key, ActorRefs.Nobody); // remove
```
Aaronontheweb: Like the change

Aaronontheweb:
Looks like the issue occurs earlier in the spec, when node 2 goes down as a result of a normal `Remove` command to a set of subscribers not being propagated:

```
[Node2:second][FAIL-EXCEPTION] Type: Xunit.Sdk.EqualException
--> [Node2:second][FAIL-EXCEPTION] Message: Assert.Equal() Failure
Expected: 5
Actual:   6
--> [Node2:second][FAIL-EXCEPTION] StackTrace:    at Xunit.Assert.Equal[T](T expected, T actual, IEqualityComparer`1 comparer)
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.<>c__DisplayClass24_0.<AwaitCount>b__0() in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 286
   at Akka.TestKit.TestKitBase.AwaitAssert(Action assertion, Nullable`1 duration, Nullable`1 interval)
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.AwaitCount(Int32 expected) in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 287
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.<DistributedPubSubMediator_must_keep_track_of_removed_users>b__29_0() in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 426
   at Akka.TestKit.TestKitBase.<>c__DisplayClass141_0.<Within>b__0()
   at Akka.TestKit.TestKitBase.Within[T](TimeSpan min, TimeSpan max, Func`1 function, String hint, Nullable`1 epsilonValue)
   at Akka.TestKit.TestKitBase.Within(TimeSpan min, TimeSpan max, Action action, String hint, Nullable`1 epsilonValue)
   at Akka.TestKit.TestKitBase.Within(TimeSpan max, Action action, Nullable`1 epsilonValue)
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.DistributedPubSubMediator_must_keep_track_of_removed_users() in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 428
   at Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe.DistributedPubSubMediatorSpec.DistributedPubSubMediatorSpecs() in D:\work\edbce350daf6a08d\src\contrib\cluster\Akka.Cluster.Tools.Tests.MultiNode\PublishSubscribe\DistributedPubSubMediatorSpec.cs:line 298
```

Offending code:

https://github.com/akkadotnet/akka.net/blob/72d281d55878ee5ec623766de8db6102076abfe9/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/PublishSubscribe/DistributedPubSubMediatorSpec.cs#L420-L424

Aaronontheweb:
On a whim, going to re-run this and see if the issue is a race condition or not. I don't think so though: it failed in both the .NET 4.6.1 and .NET Core 1.1 implementations.

Aaronontheweb:
Node 3 failed in the same spot as node 2. So only node 1's count was accurate after the `Remove` event after ~10 seconds.

Aaronontheweb:
@joshgarnett sure looks like the issue was the `ActorRefs.Nobody` comparison... Tests pass now.

Aaronontheweb:
Looks like the reason why the `ActorRefs.Nobody` substitution didn't work is because of the way the built-in `DistributedPubSub` serializer works:

https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Serialization/DistributedPubSubMessageSerializer.cs#L272

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/ActorRef.cs#L593

Haven't written a test for it, but I'm pretty sure that calling `system.Provider.ResolveActorRef(path);` on the serialized path of `Nobody.Path` won't resolve `Nobody.Instance`, which will cause the subsequent equality comparison to fail. Weird.

Aaronontheweb:
Going to merge this to get the bugfix in; don't think the juice is worth the squeeze in trying to make the serialization / deserialization of `ActorRefs.Nobody` consistent as part of this PR. Sounds like an Akka core issue. I'll file a separate issue for this.

PR #3321, opened 2018-02-10
StreamRefs

Horusiath:
This is work in progress over porting [StreamRefs](https://doc.akka.io/docs/akka/snapshot/stream/stream-refs.html) feature to .NET.

TODOs:

- [x] Ported settings, and all stream-ref related types
- [x] StreamRefSpec (except serialization)
- [x] Serialization (in progress) - this is the most problematic part, thanks to awesome support of generic data types by Protobuf. Also I've noticed, that JVM implementation gave up on serializing exceptions. I think we could go a little more ambitious here.
- [x] Documentation

This PR may require akka.streams 2.5 to be merged first (cc: @marcpiechura).

PR #3324, opened 2018-02-14
FunctionRef

Horusiath:
This is port of https://github.com/akka/akka/pull/19487

It allows to use Akka.Stream stage actor refs to be location transparent. It's necessary step before #3321


Horusiath:
I see that some of the @Aaronontheweb changes from rebased commits have leaked during rebase. I'll fix them later.

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -1619,21 +1619,35 @@ protected Action GetAsyncCallback(Action handler)
...
 
-            return _stageActorRef;
+            return _stageActor;
         }
 
+        /// <summary>
+        /// Override and return a name to be given to the StageActor of this stage.
+        /// 
+        /// This method will be only invoked and used once, during the first <see cref="GetStageActor"/>
+        /// invocation whichc reates the actor, since subsequent `getStageActors` calls function
```
marcpiechura: "which creates"

src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
```diff
@@ -4400,21 +4402,17 @@ namespace Akka.Streams.Stage
         protected PushStage() { }
         public virtual Akka.Streams.Stage.ISyncDirective OnPull(Akka.Streams.Stage.IContext<TOut> context) { }
     }
-    public sealed class StageActorRef : Akka.Actor.MinimalActorRef
+    public sealed class StageActor
```
marcpiechura: Should we move this to 1.4 for the breaking changes?

src/core/Akka.Tests/Actor/FunctionRefSpec.cs
```diff
@@ -0,0 +1,253 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="FunctionRefSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2018 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2018 Akka.NET project <https://github.com/akkadotnet/akka.net>
```
marcpiechura: needs the year range for Akka.NET too.

src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
Horusiath: Depends on how do we threat a `[ApiMayChange]` attribute - only methods with that API have been changed. We can put it to v1.4 - from my perspective it's necessary for purposes of StreamRefs, which probably will land somewhere in v1.4 as well.

src/core/Akka.Tests/Actor/FunctionRefSpec.cs
Aaronontheweb: @marcpiechura going to be submitting an update to all of the headers soon

src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
marcpiechura: Except for the `StageActorRef` property itself

Horusiath:
Ok @akkadotnet/core  , do we want to merge those changes? I know that we have some breaking changes in area around StageActorRefs, so my question is should we merge it to dev or v1.4 branch?

Issue #3320, opened 2018-02-09
Ensure that PipeTo supports Task cancellations


Horusiath:
At this moment our `PipeTo` support scenarios of failed or succeed task execution. However it turned out, that it may not work when a Task is being cancelled - the code in theory should work, however in practice it may not be the case.

We need tests in [PipeToSupportSpec](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Tests/Actor/PipeToSupportSpec.cs) which will confirm that task cancellation is handled correctly.

Also another matter - something to discuss about - is what the correct behavior should be in case if this case is not handled at this moment. 

1. As @OnurGumus in case of cancellation `task.Exception` can be null. I this is the only case (I haven't heard about the case when `task.IsFailed` is true and no exception has been provided), we could propagate null exception and i.e. add `IsCancelled` property to `Status.Failure`.
2. We could explicitly convert null exceptions to `OperationCancelledException` and forward it to failure case of `PipeTo` method. Afaik this is the closest to actual JVM Akka behavior - *Akka Futures don't have cancellations, but if future times out a `Status.Failure` with `TimeOutException` is returned*.
3. We can also add third case (like `Status.Cancelled`), however I'm afraid that this will increase complexity of the overall solution.

Aaronontheweb:
Believe this has been resolved via https://github.com/akkadotnet/akka.net/pull/5123

PR #3329, opened 2018-02-19
Ask/PipeTo improvements

Horusiath:
This PR introduces:

1. Several tests in `PipeTo` over cancelled tasks to verify correct behavior: see #3320
2. Similar tests for `Ask`.
2. `IsCancelled` property in `Status.Failure` message - by default failures for cancelled task never have exceptions.
3.  The biggest improvement is `Status.Failure` propagation inside `Ask<>` - previously if i.e. Ask was expecting a string but `Status.Failure` was returned, ask was failing with InvalidCastException. Original failure reason was lost. Now the exception wrapped by `Status.Failure` will be re-thrown with a stack trace from origin.

### DIFF
--- a/src/core/Akka.Tests/Actor/AskSpec.cs
+++ b/src/core/Akka.Tests/Actor/AskSpec.cs
@@ -265,7 +265,7 @@ namespace Akka.Tests.Actor
         {
             var actor = Sys.ActorOf<SomeActor>();
             var result = await actor.Ask<Status.Failure>("return-cancelled", timeout: TimeSpan.FromSeconds(3));
-            result.IsCancelled.ShouldBe(true);
+            result.Cause.ShouldBe(null);
         }
     }
 }
--- a/src/core/Akka.Tests/Actor/AskSpec.cs
+++ b/src/core/Akka.Tests/Actor/AskSpec.cs
@@ -21,19 +21,39 @@ namespace Akka.Tests.Actor
             : base(@"akka.actor.ask-timeout = 3000ms")
         { }
 
+        public class ExpectedTestException : Exception
+        {
+            public ExpectedTestException(string message) : base(message)
+            {
+            }
+        }
+        
         public class SomeActor : UntypedActor
         {
             protected override void OnReceive(object message)
             {
-                if (message.Equals("timeout"))
+                switch (message)
                 {
-                    Thread.Sleep(5000);
+                    case "timeout":
+                        Thread.Sleep(5000);
+                        break;
+                    case "answer":
+                        Sender.Tell("answer");
+                        break;
+                    case "throw":
+                        Task.Run(ThrowNested).PipeTo(Sender); 
+                        break;
+                    case "return-cancelled":
+                        var token = new CancellationToken(canceled: true);
+                        new Task(() => { }, token).PipeTo(Sender);
+                        break;
                 }
-
-                if (message.Equals("answer"))
-                {
-                    Sender.Tell("answer");
-                }
+            }
+            
+            internal async Task ThrowNested()
+            {
+                await Task.Delay(1);
+                throw new ExpectedTestException("BOOM!");
             }
         }
 
@@ -202,6 +222,50 @@ namespace Akka.Tests.Actor
             var waitActor = Sys.ActorOf(Props.Create(() => new WaitActor(replyActor, TestActor)));
             waitActor.Tell("ask");
             ExpectMsg("bar");
+        }
+        
+        [Fact]
+        public void Generic_Ask_when_Failure_is_returned_should_throw_error_payload_and_preserve_stack_trace()
+        {
+            var actor = Sys.ActorOf<SomeActor>();
+            var aggregateException = Assert.Throws<AggregateException>(() =>
+            {
+                var result = actor.Ask<string>("throw", timeout: TimeSpan.FromSeconds(3)).Result;
+            });
+            var exception = aggregateException.Flatten().InnerException;
+            exception.GetType().ShouldBe(typeof(ExpectedTestException));
+            exception.Message.ShouldBe("BOOM!");
+            exception.StackTrace.Contains(nameof(SomeActor.ThrowNested)).ShouldBeTrue("stack trace should be preserved");
+        }
+        
+        [Fact]
+        public async Task Generic_Ask_when_Failure_is_and_Failure_was_expected_should_not_throw()
+        {
+            var actor = Sys.ActorOf<SomeActor>();
+            var result = await actor.Ask<Status.Failure>("throw", timeout: TimeSpan.FromSeconds(3));
+            var exception = ((AggregateException)result.Cause).Flatten().InnerException;
+            exception.GetType().ShouldBe(typeof(ExpectedTestException));
+            exception.Message.ShouldBe("BOOM!");
+        }
+        
+        [Fact]
+        public async Task Generic_Ask_when_asker_task_was_cancelled_and_should_fail()
+        {
+            var actor = Sys.ActorOf<SomeActor>();
+            var aggregateException = Assert.Throws<AggregateException>(() =>
+            {
+                var result = actor.Ask<string>("return-cancelled", timeout: TimeSpan.FromSeconds(3)).Result;
+            });
+            var exception = aggregateException.Flatten().InnerException;
+            exception.GetType().ShouldBe(typeof(TaskCanceledException));
+        }
+        
+        [Fact]
+        public async Task Generic_Ask_when_asker_task_was_cancelled_and_Failure_was_expected_should_return_a_Failure()
+        {
+            var actor = Sys.ActorOf<SomeActor>();
+            var result = await actor.Ask<Status.Failure>("return-cancelled", timeout: TimeSpan.FromSeconds(3));
+            result.IsCancelled.ShouldBe(true);
         }
     }
 }
--- a/src/core/Akka.Tests/Actor/AskSpec.cs
+++ b/src/core/Akka.Tests/Actor/AskSpec.cs
@@ -225,17 +225,19 @@ namespace Akka.Tests.Actor
         }
         
         [Fact]
-        public void Generic_Ask_when_Failure_is_returned_should_throw_error_payload_and_preserve_stack_trace()
+        public async Task Generic_Ask_when_Failure_is_returned_should_throw_error_payload_and_preserve_stack_trace()
         {
             var actor = Sys.ActorOf<SomeActor>();
-            var aggregateException = Assert.Throws<AggregateException>(() =>
+
+            try
             {
-                var result = actor.Ask<string>("throw", timeout: TimeSpan.FromSeconds(3)).Result;
-            });
-            var exception = aggregateException.Flatten().InnerException;
-            exception.GetType().ShouldBe(typeof(ExpectedTestException));
-            exception.Message.ShouldBe("BOOM!");
-            exception.StackTrace.Contains(nameof(SomeActor.ThrowNested)).ShouldBeTrue("stack trace should be preserved");
+                await actor.Ask<string>("throw", timeout: TimeSpan.FromSeconds(3));
+            }
+            catch (ExpectedTestException exception)
+            {
+                exception.Message.ShouldBe("BOOM!");
+                exception.StackTrace.Contains(nameof(SomeActor.ThrowNested)).ShouldBeTrue("stack trace should be preserved");
+            }
         }
         
         [Fact]
@@ -252,12 +254,10 @@ namespace Akka.Tests.Actor
         public async Task Generic_Ask_when_asker_task_was_cancelled_and_should_fail()
         {
             var actor = Sys.ActorOf<SomeActor>();
-            var aggregateException = Assert.Throws<AggregateException>(() =>
+            await Assert.ThrowsAsync<TaskCanceledException>(async () =>
             {
-                var result = actor.Ask<string>("return-cancelled", timeout: TimeSpan.FromSeconds(3)).Result;
+                var result = await actor.Ask<string>("return-cancelled", timeout: TimeSpan.FromSeconds(3));
             });
-            var exception = aggregateException.Flatten().InnerException;
-            exception.GetType().ShouldBe(typeof(TaskCanceledException));
         }
         
         [Fact]
--- a/src/core/Akka.Tests/Actor/PipeToSupportSpec.cs
+++ b/src/core/Akka.Tests/Actor/PipeToSupportSpec.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Event;
@@ -29,7 +30,7 @@ namespace Akka.Tests.Actor
         }
 
         [Fact]
-        public void Should_by_default_send_task_result_as_message()
+        public void PipeTo_should_by_default_send_task_result_as_message()
         {
             _task.PipeTo(TestActor);
             _taskCompletionSource.SetResult("Hello");
@@ -37,7 +38,7 @@ namespace Akka.Tests.Actor
         }
 
         [Fact]
-        public void Should_by_default_not_send_a_success_message_if_the_task_does_not_produce_a_result()
+        public void PipeTo_should_by_default_not_send_a_success_message_if_the_task_does_not_produce_a_result()
         {
             _taskWithoutResult.PipeTo(TestActor);
             _taskCompletionSource.SetResult("Hello");
@@ -45,7 +46,7 @@ namespace Akka.Tests.Actor
         }
 
         [Fact]
-        public void Should_by_default_send_task_exception_as_status_failure_message()
+        public void PipeTo_should_by_default_send_task_exception_as_status_failure_message()
         {
             _task.PipeTo(TestActor);
             _taskWithoutResult.PipeTo(TestActor);
@@ -55,7 +56,7 @@ namespace Akka.Tests.Actor
         }
 
         [Fact]
-        public void Should_use_success_handling_to_transform_task_result()
+        public void PipeTo_should_use_success_handling_to_transform_task_result()
         {
             _task.PipeTo(TestActor, success: x => "Hello " + x);
             _taskWithoutResult.PipeTo(TestActor, success: () => "Hello");
@@ -65,13 +66,23 @@ namespace Akka.Tests.Actor
         }
 
         [Fact]
-        public void Should_use_failure_handling_to_transform_task_exception()
+        public void PipeTo_should_use_failure_handling_to_transform_task_exception()
         {
             _task.PipeTo(TestActor, failure: e => "Such a " + e.InnerException.Message);
             _taskWithoutResult.PipeTo(TestActor, failure: e => "Such a " + e.InnerException.Message);
             _taskCompletionSource.SetException(new Exception("failure..."));
             ExpectMsg("Such a failure...");
             ExpectMsg("Such a failure...");
+        }
+
+        [Fact]
+        public void PipeTo_should_by_default_inform_about_task_cancellation()
+        {
+            _task.PipeTo(TestActor);
+            _taskWithoutResult.PipeTo(TestActor, failure: e => $"Exception found: [{e}]" );
+            _taskCompletionSource.SetCanceled();
+            ExpectMsg<Status.Failure>(x => x.IsCancelled && x.Cause == null);
+            ExpectMsg("Exception found: []");
         }
     }
 }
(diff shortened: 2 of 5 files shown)


src/core/Akka.Tests/Actor/PipeToSupportSpec.cs
```diff
@@ -65,13 +66,23 @@ public void Should_use_success_handling_to_transform_task_result()
...
         {
             _task.PipeTo(TestActor, failure: e => "Such a " + e.InnerException.Message);
             _taskWithoutResult.PipeTo(TestActor, failure: e => "Such a " + e.InnerException.Message);
             _taskCompletionSource.SetException(new Exception("failure..."));
             ExpectMsg("Such a failure...");
             ExpectMsg("Such a failure...");
         }
+
+        [Fact]
+        public void PipeTo_should_by_default_inform_about_task_cancellation()
```
Aaronontheweb: Looks good

src/core/Akka/Actor/ActorBase.cs
```diff
@@ -57,11 +57,13 @@ public Failure(Exception cause)
                 Cause = cause;
             }
 
+            /// <summary>
+            /// Checks if a provided failure message was a result of cancelled operation.
+            /// </summary>
+            public bool IsCancelled => Cause == null;
```
Aaronontheweb: I approve of these API changes, although we'll need to start telling people in the docs to look for `Status.Failure` messages if they're not handling task failures / cancellations themselves inside their own continuations at the end of a `Task`

Aaronontheweb: Although, this check for determining if the task was cancelled... `Cause == null`.... IMHO, this value shouldn't be inferred by a loose convention like this. Rather, we should add a constructor overload to `Status.Failed` where we explicitly set this value to true or false and mark the old constructor as `Obsolete`. 

Right now we're assuming that having a null exception means a task cancellation - might be right most of the time but that's not good enough. Needs to be intentional. We should change the API to add an explicit overload and handle that inside the `PipeTo` implementation itself.

Horusiath: I think, that we could remove that field and simply let users know (through docs) that in case of cancellation exception should be null (+ keep `TaskCancellationException` filler in Ask<>). Setting up separate cancellation handler in PipeTo is bad - it breaks backwards compatibility (currently a null exception is already passed to PipeTo's failure handler on cancelled task).

src/core/Akka.Tests/Actor/AskSpec.cs
```diff
@@ -203,6 +223,50 @@ public void Can_Ask_actor_inside_receive_loop()
...
             ExpectMsg("bar");
         }
+        
+        [Fact]
+        public void Generic_Ask_when_Failure_is_returned_should_throw_error_payload_and_preserve_stack_trace()
+        {
+            var actor = Sys.ActorOf<SomeActor>();
+            var aggregateException = Assert.Throws<AggregateException>(() =>
+            {
+                var result = actor.Ask<string>("throw", timeout: TimeSpan.FromSeconds(3)).Result;
```
maxcherednik: Guys, please let's not do result - it makes things complicated. We start doing AggregateException unwrapping, flattening and so on. Just by reading code I can't understand the level of nested exception...

src/core/Akka.Tests/Actor/AskSpec.cs
```diff
@@ -203,6 +223,50 @@ public void Can_Ask_actor_inside_receive_loop()
...
+            exception.Message.ShouldBe("BOOM!");
+        }
+        
+        [Fact]
+        public async Task Generic_Ask_when_asker_task_was_cancelled_and_should_fail()
+        {
+            var actor = Sys.ActorOf<SomeActor>();
+            var aggregateException = Assert.Throws<AggregateException>(() =>
+            {
+                var result = actor.Ask<string>("return-cancelled", timeout: TimeSpan.FromSeconds(3)).Result;
```
maxcherednik: Same here. There is ThrowsAsync api

src/core/Akka/Actor/Futures.cs
```diff
@@ -108,7 +111,17 @@ public static Task<T> Ask<T>(this ICanTell self, object message, CancellationTok
...
+            
+            // if failure was delivered and not expected, rethrow it
+            if (result is Status.Failure f && typeof(T) != FailureType)
+            {
+                // if failure was result of cancelled task, exception will be null
+                var cause = f.Cause ?? new TaskCanceledException();
+                ExceptionDispatchInfo.Capture(cause).Throw();
+                return default(T);
+            }
+            else return (T) result;
```
maxcherednik: I am wondering if it's a good idea in general. If something failed somewhere - would not it be better not to reply at all? Basically we are speaking of unhandled exception... Maybe it's better let it die there?

ps. will it work across remote ?

Horusiath: @maxcherednik it will work across remote connections as long as serializer is able to correctly serialize/deserialize exceptions (I'm not sure what's the status of it on .NET core). Right now if something has failed and `Status.Failure` is returned, it's the worst scenario of all: the InvalidCastExcepton is throw (as we try to cast Status.Failure to T), while the original reason is lost. I've already tried so fix that back in #3227 , but I've thought that your work over `Ask` upgrade will cover that.

Horusiath:
I've made some final polishing of the underlying logic. Example:

```csharp
class CustomActor : UntypedActor
{
    override void OnReceive(object message)
    { 
        // spawn some async job, that will throw CustomException inside
        Task.Run(SomeAsyncJob).PipeTo(Sender); 
    }
}

async Task CustomCallAsync()
{
    try 
    {

        // the CustomException is thrown somewhere inside of an `actor` side-process
        // before this, this await would throw InvalidCastException and primary reason would be unknown
        var result = await system.ActorOf<CustomActor>().Ask<Reply>(new Request());
    } 
    catch (CustomException e)
    {
        // here it works with async/await and has a stack trace preserved
    }
}
```

src/core/Akka/Actor/Futures.cs
maxcherednik: I see. No, did not cover that case. InvalidCastException is ugly for sure.

But do we need to bring some deep exception, which arrived across async message boundary?  As an option - we don't catch unhandled exception at all and asking part will just see that nothing was sent back for whatever reason(network, slowness, db issue). Anything which is not public API of that remote service should not be sent back. IMHO.

Horusiath: I think, this case (ignore sending failures back to the user) should be handled on the sender side as opt-in. I.e. if `PipeTo(Sender, failure: ex => null)`, then don't send any response back.

maxcherednik: @Horusiath @Aaronontheweb what I am trying to say, Status.Failure is not any actor's default API in case of error, right? And it should not be a special case in Ask method. I remember, I did propagate a general Exception somewhere - that time I cringe and thought, I probably should not catch it at all, it should fail and die.

Horusiath:
@maxcherednik I'll move the discussion here - it's not visible in code comments.

The starting point for us is something I'd call a broken behavior - throwing inappropriate exception inside `Ask` task. Since the current behavior is not satisfactory, I think we're free to define our own. What is my goal here is to make easier for developers to reason about the eventual failure in case of `Ask<>`. Given assumptions:

1. `Ask` represents a request-response pattern, where response can be either success or failure.
2. `Ask<>` is used mainly in external communication (3rd party code uses Ask to communication with internal akka actors). It's not natural to see a response pattern matching outside actor system (like in case of custom success/failure messages). We are still free to do so, but have a task, that already uses success/failure states.
3. We don't want to introduce heavy breaking changes - this means that in case of `Status.Failure` we still want to throw exception, just like before. AFAIK nobody make a special try/catch case in `Ask<>` to catch `InvalidCastException`.
4. `Status.Failure` is default fault message, unless explicitly specified otherwise.

I think it's good to give a `Status.Failure` an extra rule - we throw an exception anyway, so why not to throw a meaningful one? People often start to looking for errors at caller side, this will make them some good points to begin with - error message + stack trace notifying about the origin actor/method where the error was thrown.

What I was thinking about:

- Is it good idea to rethrow unwrapped callee-side exceptions at caller side? Maybe they should be wrapped in some custom exception i.e. `AskFailureException` and available as inner exception. Stack trace should still be preserved.
- In this PR, an exception inside `Status.Failure` is being unwrapped and captured before rethrown. I was thinking about putting that logic as a virtual method into `Status.Failure` itself - I know that messages should not contain this kind of logic, but this way we could override a `Status.Failure` with our own message to customize an error that would be thrown by `Ask<>` in case of failure.
- Make `Ask<>` rethrow rule dependent on some marker interface ie. `IFailure`, rather than `Status.Failure` class (which would implement it) - this would allow users to define their own message types, if they would like to use for Ask rethrow mechanics.

Aaronontheweb:
I agree with most of what @Horusiath has summarized here. Definitely better to send SOMETHING back and cause the Ask to blow-up quickly rather than slowly. "Fail fast" et al.

>In this PR, an exception inside Status.Failure is being unwrapped and captured before rethrown. I was thinking about putting that logic as a virtual method into Status.Failure itself - I know that messages should not contain this kind of logic, but this way we could override a Status.Failure with our own message to customize an error that would be thrown by Ask<> in case of failure

This sounds like a bottomless pit of small, subtle bugs accrued over time. The "don't be cute" rule applies here. Just keep the behavior simple and leave it up to the end-user to infer "meaning" from the underlying exception. As long as we pass some sort of exception into the `Status.Failure` upon cancellation, which is what this PR was originally trying to accomplish, that should be sufficient. Stuff a `TaskCancellationException` in there instead of letting it be null and call it a day IMHO.

Horusiath:
@Aaronontheweb problem with filling `TaskCancellationException` on callee side is that it changes current behavior, which passes `Status.Failure` with null cause on cancellation. While this normally is not a problem, when we're talking about incremental cluster upgrade, it could be irritating i.e:

1. Node A (already upgraded) `Ask`s node B (not yet upgraded)
2. Node B task is cancelled with old behavior (so `Status.Failure` with null cause)
3. Node A receives cancelled task and breaks in `ExceptionDispatchInfo.Capture` which cannot accept nulls.

While this is nothing super important (we'll throwing exception anyway), I think it's worth noticing.

Aaronontheweb:
@Horusiath depends. You wanted to pass along information about when a task is cancelled, no? If not, then don't worry about it. I think that data would be helpful - more clear than `null` == `TaskCancelled`.

maxcherednik:
@Horusiath @Aaronontheweb I will also try to summarize.
1. Ask<> pattern is an edge between .net world and actor model
2. It's known that by using Ask we are sending a message to an actor(remote or local - we don't know)

Having these 2 facts, lets just take the second one - sending a message to an actor. If the actor fails - would you send Status.Failure by default? I guess the answer is no. Because this is the nature of the  actor model approach(as I see it, correct me if I am wrong). 

Now let's get back to the the first point. Ask should definitely bring back any exception happened in the implementation of the Ask itself(we are not swallowing anything).
The behavior we currently observe(InvalidCastException) it is a bug, caused by another "hack", which is PipeTo. If "normal" actor will never send you something like Status.Failure or whatever.

`var res = serviceActor.Ask<string>(serviceRequest);`

if I receive InvalidCastException here - the only reason for this is that I made a mistake. It should have been this:

`var res = serviceActor.Ask<ServiceResponse>(serviceRequest);`

@Aaronontheweb mentioned "fail fast". So in my opinion failing fast - would be not to send it back from the place it was thrown.

Sorry if it looks cumbersome, I hope described it well.

ps. let's not break actor model - my main point.

Aaronontheweb:
@maxcherednik `Status.Failure` and `PipeTo` aren't hacks; the latter is just a bridge between the TPL and the actors themselves. It's accomplishing the same goal as `Ask`, but just from a different direction: TPL --> Actor instead of Actor --> TPL. It eliminates the need to write your own boilerplate continuation code to the do the same thing, something that thousands of users would need to do otherwise. 

So as far as I can tell, this change just unpacks the exception from a `PipeTo` operation that may have been sent back to the `FutureActorRef` running the `Ask`.  In essence, @Horusiath is trying to force the original thing that caused the `PipeTo` to fail to get flushed up to the end-user.

You're arguing that in addition to being breaking behavior for those edge-cases specifically, we probably shouldn't be littering a special case inside the infrastructure code used to power `Ask`. Do I understand that correctly? If not, then I don't understand what your objection to this change is.

maxcherednik:
> You're arguing that in addition to being breaking behavior for those edge-cases specifically, we probably shouldn't be littering a special case inside the infrastructure code used to power Ask. Do I understand that correctly? 

Yes. 

I know it might be not popular, but I really think that PipeTo is a hack. Ask is a well - designed bridge, there is nothing to argue. It sends and waits. All clean.

But PipeTo was introduced so that to "unblock" the Actor which did some async operation inside the actor. It's just another operation sync, async, it does not matter - it takes time and yes, you actor will be blocked. Async/await stuff introduced just to make resource usage more efficient - it's not breaking any other rules - you read and execute code sequentially. That's why async code inside the actor should not be treated differently from the sync one - cause they are the same. If we have a long running blocking operation inside the actor, we are not trying to unblock it somehow be PipeTo sync or whatever. It takes what it takes, we just spin another actor next to for another similar task, cause this is our concurrency primitive. IMHO.

Aaronontheweb:
Ok, got it @Horusiath what edge case motivated you to handle the `Status.Failure` stuff inside `Ask`? If it's a relatively uncommon / avoidable case maybe @maxcherednik is right and we shouldn't be handling it there.

maxcherednik:
@Horusiath @Aaronontheweb I am not saying that I am 100% right, it's a discussion. 

Let's imagine a situation. I send a message with Ask to a remote actor and somehow it returned me something scary OutOfMemoryException or whatever. What do I need to think of this on my side? Where it happened? Local, remote? I can't say it from the exception been thrown, right?

Aaronontheweb:
@maxcherednik generally speaking, on all pull requests I use the following thought process?

1. Is this change implementing something the framework should be responsible for?
2. Does this change help more users than it hurts?
3. Does this change actually work?

Answering "no" to any question means I'm not going to approve the PR. At the moment, the discussion has made me move my answer on questions 1 and 2 from "yes" to "no." That's why I asked @Horusiath for some clarification.

maxcherednik:
@Aaronontheweb np with that.

Horusiath:
@Aaronontheweb 

1. **Is this change implementing something the framework should be responsible for?** - notifying end user code about failures is part of the framework responsibilities. + Ideally it should also inform not only about the failure happening but also a reason.
2. **Does this change help more users than it hurts?** - yes, more on that below.
3. **Does this change actually work?** - yes, I've detailed 3 concerns of mine above, but they don't change the mechanics.

Imagine following scenario:

A programmer is interacting with actor system from the outside, i.e. ASP.NET controller action. Now he wants to handle an Ask. What is **minimum necessary code** to do it right now?

```csharp
try
{
    var response = await SystemReferences.Actor.Ask<IResponse>(new Request(), timeout: TimeSpan.FromSeconds(30));
    switch (response)
    {
        case SuccessfulResponse success: ...
        case FailureResponse failure: 
            // if you want to pattern match on the exceptions, feel free to go or rethrow them here
            switch(failure.Cause?.InnerException)
            {
                case CustomException e: ...
                case null: break; ignore
                default var ex: 
                    // rethrow other types of exceptions
                    ExceptionDispatchInfo.Capture(ex).Throw();
                    break;
            }
    }
}
catch (AskTimeoutException e)
{
    // since Ask may fail anyway, we still may need to handle failures
}
```

That's not all:

- `IResponse`, `SuccessfulResponse` and `FailureResponse` are not defined there. They need to be defined potentially for every interaction. This not Scala, you cannot do that in 3 lines of code.
- Since we use custom response, if the Ask response was result of `PipeTo`, then we need to declare failure handle for that PipeTo as well. Of course we could not, but then using a generic `Ask<>` is totally pointless, because the only case when it would work is when we call `Ask<object>()`.
- Since C# doesn't have exhaustive pattern matching (like F# or Scala) we don't know if our switch statement handled every case. Also IDE/compiler won't tell us what are valid response types, we need to figure them out by ourselves or be a little "smarter" when defining them.

This is a lot of boilerplate. It's also totally unidiomatic for the people using C# on daily basis. What I've proposed, doesn't break existing messaging semantics of the framework. What it allows to do, is to make them **opt-out**, when people are not interested in that.

Aaronontheweb:
> IResponse, SuccessfulResponse and FailureResponse are not defined there. They need to be defined potentially for every interaction. This not Scala, you cannot do that in 3 lines of code.

This PR doesn't address this at all. Nor point 3 near as I can tell.

> Since we use custom response, if the Ask response was result of PipeTo, then we need to declare failure handle for that PipeTo as well. Of course we could not, but then using a generic Ask<> is totally pointless, because the only case when it would work is when we call Ask<object>().

Still seems like a relatively rare edge case to me. Why should we check for it on *every* Ask?

maxcherednik:
@Horusiath 
> IResponse, SuccessfulResponse and FailureResponse are not defined there. They need to be defined potentially for every interaction. This not Scala, you cannot do that in 3 lines of code.

In fact we never need to do this. We define a Response, which contains the result of the operation, but we never need to convey all the bad things happened. They are out of scope, actor failed and died and never managed to reply - we should not bring this info of unhandled exception back - it is not our contract.

maxcherednik:
> if a users calls a var task = actorRef.Ask<Reply>(new Request()), and receives a Status.Failure back, which is incompatible, what in your opinion should be behavior of Ask task?

@Horusiath @Aaronontheweb  it's actually a really good question. So that to speculate on this topic, can we go a bit more abstract. 
Let's say, we have an actor. We design it in a way that if it receives message A, it should respond with message B. If we use: 
`var response = actorRef.Ask<B>(new A())`
, then everything looks good so far.
But this construction does not let us design an actor in a way, that with message A, it can reply with message B(let's say correct result) and message C(incorrect). In this way there is no way to use Ask pattern with strongly specified response type, cause now we have 2 different response types. How can we overcome this? Either we design the message B, which contains a field which represents correct and incorrect result; or we  do:
`var response = actorRef.Ask<object>(new A())` 
and deal with all the possible response types on the "client" side. Which one is better or more canonical? I don't know.

I guess if we can answer this generic question, we will be able to understand what to do with Ask, which can all of a sudden receive Status.Failure. 

My opinion on this, if we expect to receive more than one type of responses, we need to go explicit:
```
var response = actorRef.Ask<object>(new A())
if(response is B)
{

{
else if( response is Status.Failure) // we are telling explicitly, that this is part of our communication protocol
{

}
```

Horusiath:
@maxcherednik what you propose is keep current status quo, which IMO is a broken design.

1. You cannot prevent Ask result from a potential failure. Moreover, when we got `Status.Failure` we essentially have all of the necessary information to give user a feedback about the cause of failure. We already made (a significant) cost of serializing and deserializing exception with stack trace over the network. The only issue is that on the last mile we essentially say: "fuck you, we know what happened but we won't tell you, go to your logs (if you have any) and figure it out by yourself".
2. In current design the result of failure is `InvalidCastException`. The natural reaction of almost every developer is to start looking from there for what was the reason of failure. I'm sure they will be happy to see, that the actual reason for failure has nothing to do with received exception (which is only symptom) - neither Ask nor PipeTo API mentions about `Status.Failure`, and even if they did, the actual correlation between Ask and PipeTo may not be obvious, easy to track or to document.
3. Current API cannot - and won't be able to - prohibit users from making this kind of mistakes. Moreover this is a default behavior (when used with `PipeTo`, which defaults failures to `Status.Failure`), which means that this will be probably a problem for many users, including majority of newcomers.

My proposal in addition to this PR is to:

- If `Ask` results with `Status.Failure` (and an expected generic response was of another type), we'll throw something like `AskFailedException($"Actor {Sender} failed while processing Ask request")` + failure.Cause as its inner exception.

Why?

- `AskFailedException` will inform about the source of an error, while inner exception will have details about the cause (including stack trace). 
- It's also pretty clear about that the failure is not in current processing thread, but somewhere else. 
- Also it's easy to mute or log (via try/catch), while catch still can be specialized (via `catch (AskFailedException e) when (e.InnerException is MyCustomException`).

Aaronontheweb:
> what you propose is keep current status quo, which IMO is a broken design

Strongly disagree. Works fine for tens of thousands of users.

I'm closing this PR. `PipeTo` and `Ask` don't have a direct relationship and it's unclear how often they're being used in combination with each other. Putting code for handling `PipeTo` failures inside `Ask` doesn't add any value, IMHO.

SamEmber:
So is the fact a `Sender.Tell(new Failure { Exception = e }, Self)` throws a Unable to cast object of type 'Akka.Actor.Failure' to type Exception going to be ignored? 

I think this should either be fixed to give the correct exception included in the Failure or removed form the documentation as its currently incorrect and is confusing people.

PR #3331, opened 2018-02-20
Added .NET Foundation Copyright headers

Aaronontheweb:
Also updated the dates on all files. 
Updated Powershell script to have it always replace headers on files when run. Powershell script assumes that first seven lines of all files belong to the copyright holder.


Aaronontheweb:
Need to let build server run compile stage for this. Failed locally on my machine when running build script, but succeeded in VS. Not sure why. Probably an edge-case where a file had something functional in the first seven lines or someone wrapped the copyright header in a region.

Issue #3276, opened 2018-01-17
Use ValueTuple instead of Tuple  


marcpiechura:
We should switch to `ValueTuple`s to reduces allocations and improve readability, since this would be quite a big chance if introduced at once we split it up in several chunks. 

The first place would be [Keep.Both](https://github.com/akkadotnet/akka.net/blob/a478c5e8c8d5de97dcf7bcea7e2a80fbaa5e6cc8/src/core/Akka.Streams/Dsl/Keep.cs#L47) .

Horusiath:
You have my sword :)

vasily-kirichenko:
Yes!

vasily-kirichenko:
https://github.com/akkadotnet/akka.net/pull/3282

Aaronontheweb:
@marcpiechura since this would constitute a new package dependency and possibly some API changes, going to propose that we dovetail this into the 1.4 branch.

NickPolyder:
Is this issue resolved ?

marcpiechura:
@NickPolyder yup, the PR https://github.com/akkadotnet/akka.net/pull/3282 is not merged but the work is done, at least for Akka.Streams. Out of my head I don’t know if we have other places where it would make sense, I’ll try to find some and create a new issue you could work on 😉

NickPolyder:
@marcpiechura  Thank you for your quick response 👍

marcpiechura:
@NickPolyder After a quick look the "other places" are everywhere :stuck_out_tongue: 
So if you want to start with it you need to target the V1.4.0 branch and reference  System.ValueTuple version 4.4.0.
You can pick any of the projects you want except for Akka.Streams.XXX since those are already done. 
If you need any help let us know, of course you could also pick a different task if you want.

NickPolyder:
@marcpiechura i have done most of the job on my branch [here](https://github.com/NickPolyder/akka.net/tree/issue-3276) although i have some kind of knock back on the compilation due to error on the compilation of the docs [error CS1569: Error writing to XML documentation file: Object reference not set to an instance of an object.](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs1569) did you ever had this problem ?

Thanks for your time.

marcpiechura:
@NickPolyder haven't seen this before and your branch builds just fine after a clean checkout. 
[...]

PR #3332, opened 2018-02-20
ValueTuple for all projects except for streams 

NickPolyder:
This pull request serves to close the issue: #3276 

I tried to find as many Tuples as i could find. Using the visual studio search capabilities.
Although i did not change every occurrence of `System.Tuple` and that's because i thought that on those rare occurrences the cost of boxing the `ValueTuple` to an object reference was not worth the conversion and the extra code. 
If you think that this is not the case let me know and i will change the remaining occurrences.

Thanks for your time.

### DIFF
--- a/src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
+++ b/src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
@@ -40,9 +40,8 @@ namespace Akka.Remote.Tests
         {
             protected override void OnReceive(object message)
             {
-                if (message is ValueTuple<Props, string>)
+                if (message is ValueTuple<Props, string> tuple)
                 {
-                    var tuple = ((Props, string))message;
                     Sender.Tell(Context.ActorOf(tuple.Item1, tuple.Item2));
                 }
             }
--- a/src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
+++ b/src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
@@ -40,9 +40,9 @@ namespace Akka.Remote.Tests
         {
             protected override void OnReceive(object message)
             {
-                var tuple = message as Tuple<Props, string>;
-                if (tuple != null)
+                if (message is ValueTuple<Props, string>)
                 {
+                    var tuple = ((Props, string))message;
                     Sender.Tell(Context.ActorOf(tuple.Item1, tuple.Item2));
                 }
             }
@@ -343,7 +343,7 @@ namespace Akka.Remote.Tests
             // it's used for the pool of the SimpleDnsManager "/IO-DNS/inet-address"
             var probe = CreateTestProbe(masterSystem);
             var parent = ((ExtendedActorSystem)masterSystem).SystemActorOf(FromConfig.Instance.Props(Props.Create<Parent>()), "sys-parent");
-            parent.Tell(Tuple.Create(FromConfig.Instance.Props(EchoActorProps), "round"), probe);
+            parent.Tell((FromConfig.Instance.Props(EchoActorProps), "round"), probe);
             var router = probe.ExpectMsg<IActorRef>();
             var replies = CollectRouteePaths(probe, router, 10);
             var children = new HashSet<ActorPath>(replies);
--- a/src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
+++ b/src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
@@ -40,7 +40,7 @@ namespace Akka.Remote.Tests
         {
             protected override void OnReceive(object message)
             {
-                if (message is ValueTuple<Props, string> tuple)
+                if (message != null && message is ValueTuple<Props, string> tuple)
                 {
                     Sender.Tell(Context.ActorOf(tuple.Item1, tuple.Item2));
                 }
--- a/src/core/Akka/Actor/Inbox.Actor.cs
+++ b/src/core/Akka/Actor/Inbox.Actor.cs
@@ -25,7 +25,7 @@ namespace Akka.Actor
 
         private object _currentMessage;
         private Select? _currentSelect;
-        private Tuple<TimeSpan, ICancelable> _currentDeadline;
+        private (TimeSpan, ICancelable) _currentDeadline;
 
         private readonly int _size;
         private readonly ILoggingAdapter _log = Context.GetLogger();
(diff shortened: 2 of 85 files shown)


src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
```diff
@@ -40,9 +40,9 @@ private class Parent : UntypedActor
         {
             protected override void OnReceive(object message)
             {
-                var tuple = message as Tuple<Props, string>;
-                if (tuple != null)
+                if (message is ValueTuple<Props, string>)
                 {
+                    var tuple = ((Props, string))message;
```
marcpiechura: this can be removed 
```
  if (message is ValueTuple<Props, string> tuple)
```

src/core/Akka/Actor/Inbox.Actor.cs
```diff
@@ -185,29 +185,21 @@ protected override bool Receive(object message)
 
             if (_clients.Count == 0)
             {
-                if (_currentDeadline != null)
-                {
-                    _currentDeadline.Item2.Cancel();
-                    _currentDeadline = null;
-                }
+                _currentDeadline.Item2.Cancel();
```
marcpiechura: While `_currentDeadline` can't be null anymore `Item2` can, so we should still check if it's not null before cancelling

src/core/Akka/Actor/Inbox.Actor.cs
```diff
@@ -185,29 +185,21 @@ protected override bool Receive(object message)
...
             {
                 var next = _clientsByTimeout.FirstOrDefault();
                 if (next != null)
                 {
-                    if (_currentDeadline != null)
-                    {
-                        _currentDeadline.Item2.Cancel();
-                        _currentDeadline = null;
-                    }
+                    _currentDeadline.Item2.Cancel();
```
marcpiechura: as above

src/core/Akka/Routing/ConsistentHash.cs
```diff
@@ -47,10 +47,10 @@ public ConsistentHash(SortedDictionary<int, T> nodes, int virtualNodesFactor)
             _virtualNodesFactor = virtualNodesFactor;
         }
 
-        private Tuple<int[], T[]> _ring = null;
-        private Tuple<int[], T[]> RingTuple
+        private (int[], T[])? _ring = null;
+        private (int[], T[]) RingTuple
         {
-            get { return _ring ?? (_ring = Tuple.Create(_nodes.Keys.ToArray(), _nodes.Values.ToArray())); }
+            get { return ((int[], T[]))(_ring ?? (_ring = (_nodes.Keys.ToArray(), _nodes.Values.ToArray()))); }
```
marcpiechura: Do we need this explicit cast?

src/core/Akka.Remote.Tests/RemoteRouterSpec.cs
NickPolyder: yes i tried to not force the latest semantics of the language in case you did not wanted too.

src/core/Akka/Routing/ConsistentHash.cs
NickPolyder: If we proceed with Nullable ValueTuple then yes. Otherwise is a compilation error

NickPolyder:
@marcpiechura i was concerned with the change to default values too. We could make most of them Nullable. But this makes them reference types and i think that this is what we trying to avoid here. Am i right?

marcpiechura:
@NickPolyder I agree, Nullable isn’t ideal but it keeps the semantics from normal Tuple and it’s in fact a struct too. So I would propose to change all places where we used null before to Nullable. 

CC @akkadotnet/core

NickPolyder:
@marcpiechura  Okay. I will procceed with this change request and will ping you here when its ready.

Thanks

marcpiechura:
Nice @NickPolyder, looks good now 👍

We can’t merge this PR right now, because the Mono compiler doesn’t likes ValueTuples, hence that’s the reason why the Mono build always fails. 
We’ve planned to drop Mono support in the 1.4 branch around next week, so until then, you’re “done” here and take the next task 😜

NickPolyder:
@marcpiechura okay. 

Thank you very much. If anything comes up it will be my pleasure to help.

marcpiechura:
@NickPolyder we have a lot of thinks that could be done, basically everything which is marked with “up for grabs” , or some low hangining fruits for “beginners”, feel free to pick whatever you like 😉

https://github.com/akkadotnet/akka.net/labels/good%20for%20first-time%20contributors

Danthar:
Marked on hold because this PR can only be finished when we axe Mono support in the next major release.
Reviewed with Remarks, because of the nullable valuetuple question.

On that subject. I think its a better alternative then using the `default()` construct.

marcpiechura:
@Danthar @NickPolyder has already replaced all `default()` cases to Nullable ValueTuple, so from my point of view the remark was fixed in the last commit, or have I missed some places?

NickPolyder:
@marcpiechura maybe some of the new pr's made use of `default()`. 
Basically i think he means that he agree's with the way we solved it.

Danthar:
@NickPolyder Yes that is exactly what i mean :-)

marcpiechura:
@Aaronontheweb while your working on #3282 might also want to look at this one

Aaronontheweb:
Need to revisit this for v1.4 - might just be a matter of sorting through the merge conflicts here

NickPolyder:
@Aaronontheweb I can have a look over the weekend to fix the conflicts if you are okay with it :)

NickPolyder:
@Aaronontheweb I have resolved the conflicts please can you review the latest changes and conflict resolutions ?

Aaronontheweb:
Will do! Not sure why CI didn't run

Aaronontheweb:
Changed the base branch to `dev` so CI should run properly now.

Aaronontheweb:
Looks like CI still didn't run properly, probably due to the old build script. Going to update the branch and see how that affects things.

Aaronontheweb:
Have a couple of issues here:

- Looks like Incrementalist generated a bad output (only Akka.csproj)
- Looks like there's some issue with the reference assemblies for .NET Core on Linux.

I'll look into it.

Aaronontheweb:
https://github.com/petabridge/Incrementalist/issues/63 - found the issue; yet another Roslyn quirk.

Aaronontheweb:
Ok, going to re-run this with #3906 merged and see if that solves the issue

PR #3333, opened 2018-02-21
fixed API call inside TransportAdapters exentsion to support third-party RemoteActorRefProviders

Aaronontheweb:
An oversight from some of the work I did leading up to 1.3.4. Being able to use the `TransportAdapter` system as an injection point for doing things like slapping a byte / s monitor in front of an Akka.Remote transport is pretty handy. But, needed to change the cast in the constructor to target the `IRemoteActorRefProvider` interface rather than the concrete class. Also added some more comments to some of the public parts of this code while I was there.

### DIFF
--- a/src/core/Akka.Remote/Transport/TransportAdapters.cs
+++ b/src/core/Akka.Remote/Transport/TransportAdapters.cs
@@ -13,6 +13,7 @@ using Akka.Actor;
 using Akka.Actor.Internal;
 using Akka.Event;
 
+
 namespace Akka.Remote.Transport
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


src/core/Akka.Remote/Transport/TransportAdapters.cs
```diff
@@ -69,7 +70,7 @@ internal class TransportAdapters : IExtension
         public TransportAdapters(ExtendedActorSystem system)
         {
             System = system;
-            Settings = ((RemoteActorRefProvider)system.Provider).RemoteSettings;
+            Settings = ((IRemoteActorRefProvider)system.Provider).RemoteSettings;
```
Aaronontheweb: This is the only real change here. This call happens once at startup per `ActorSystem` the first time this `ActorSystemExtension` is used.

PR #3366, opened 2018-03-17
Allow persisting events when recovery has completed

ismaelhamed:
~~**DO NOT MERGE** before #3372~~

> This PR adds support for persisting when recovery is completed, before any commands are received. It's still not possible to persist new events when replaying events. Replaying events should typically be free of external side effects so it doesn't make much sense to support persisting during the actual replay, but there might be some use cases for doing it right afterwards.

Original issues: [#21736](https://github.com/akka/akka/issues/21736) and [#22609](https://github.com/akka/akka/issues/22609)
Original PR: [#21893](https://github.com/akka/akka/pull/21893)

- [x] ~~core implementation~~
- [x] ~~unit tests~~

### DIFF
--- a/src/core/Akka.Persistence.Tests/PersistentActorSpec.cs
+++ b/src/core/Akka.Persistence.Tests/PersistentActorSpec.cs
@@ -611,5 +611,21 @@ namespace Akka.Persistence.Tests
             persistentActor.Tell("boom");
             ExpectMsg("failed with TestException while processing boom");
         }
+
+        [Fact]
+        public void PersistentActor_should_be_able_to_persist_events_that_happen_during_recovery()
+        {
+            var persistentActor = ActorOf(Props.Create(() => new PersistInRecovery(Name)));
+            persistentActor.Tell(GetState.Instance);
+            ExpectMsgInOrder("a-1", "a-2", "rc-1", "rc-2");
+            persistentActor.Tell(GetState.Instance);
+            ExpectMsgInOrder("a-1", "a-2", "rc-1", "rc-2", "rc-3");
+            persistentActor.Tell(new Cmd("invalid"));
+            persistentActor.Tell(GetState.Instance);
+            ExpectMsgInOrder("a-1", "a-2", "rc-1", "rc-2", "rc-3", "invalid");
+            Watch(persistentActor);
+            persistentActor.Tell("boom");
+            ExpectTerminated(persistentActor);
+        }
     }
 }
--- a/src/core/Akka.Persistence/Eventsourced.Recovery.cs
+++ b/src/core/Akka.Persistence/Eventsourced.Recovery.cs
@@ -17,7 +17,7 @@ namespace Akka.Persistence
 
     internal class EventsourcedState
     {
-        public EventsourcedState(string name, bool isRecoveryRunning, StateReceive stateReceive)
+        public EventsourcedState(string name, Func<bool> isRecoveryRunning, StateReceive stateReceive)
         {
             Name = name;
             IsRecoveryRunning = isRecoveryRunning;
@@ -25,9 +25,7 @@ namespace Akka.Persistence
         }
 
         public string Name { get; }
-
-        public bool IsRecoveryRunning { get; }
-
+        public Func<bool> IsRecoveryRunning { get; }
         public StateReceive StateReceive { get; }
 
         public override string ToString() => Name;
@@ -47,7 +45,7 @@ namespace Akka.Persistence
         /// </summary>
         private EventsourcedState WaitingRecoveryPermit(Recovery recovery)
         {
-            return new EventsourcedState("waiting for recovery permit", true, (receive, message) =>
+            return new EventsourcedState("waiting for recovery permit", () => true, (receive, message) =>
             {
                 if (message is RecoveryPermitGranted)
                     StartRecovery(recovery);
@@ -59,7 +57,7 @@ namespace Akka.Persistence
         /// <summary>
         /// Processes a loaded snapshot, if any. A loaded snapshot is offered with a <see cref="SnapshotOffer"/>
         /// message to the actor's <see cref="ReceiveRecover"/>. Then initiates a message replay, either starting
-        /// from the loaded snapshot or from scratch, and switches to <see cref="ReplayStarted"/> state.
+        /// from the loaded snapshot or from scratch, and switches to <see cref="RecoveryStarted"/> state.
         /// All incoming messages are stashed.
         /// </summary>
         /// <param name="maxReplays">Maximum number of messages to replay</param>
@@ -81,7 +79,7 @@ namespace Akka.Persistence
                 else return false;
             };
 
-            return new EventsourcedState("recovery started - replay max: " + maxReplays, true, (receive, message) =>
+            return new EventsourcedState("recovery started - replay max: " + maxReplays, () => true, (receive, message) =>
             {
                 try
                 {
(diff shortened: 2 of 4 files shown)


src/core/Akka.Persistence/Eventsourced.Recovery.cs
```diff
@@ -17,17 +17,15 @@ namespace Akka.Persistence
 
     internal class EventsourcedState
     {
-        public EventsourcedState(string name, bool isRecoveryRunning, StateReceive stateReceive)
+        public EventsourcedState(string name, Func<bool> isRecoveryRunning, StateReceive stateReceive)
```
ismaelhamed: Is this the best option?

src/core/Akka.Persistence/Eventsourced.Recovery.cs
```diff
@@ -157,83 +150,85 @@ private EventsourcedState Recovering(Receive recoveryBehavior, TimeSpan timeout)
             // protect against event replay stalling forever because of journal overloaded and such
             var timeoutCancelable = Context.System.Scheduler.ScheduleTellRepeatedlyCancelable(timeout, timeout, Self, new RecoveryTick(false), Self);
             var eventSeenInInterval = false;
+            var recoveryRunning = true;
 
-            return new EventsourcedState("replay started", true, (receive, message) =>
+            return new EventsourcedState("replay started", () => recoveryRunning, (receive, message) =>
```
ismaelhamed: Implicitly capture closure

Aaronontheweb:
@ismaelhamed still waiting on more changes first?

ismaelhamed:
@Aaronontheweb code is complete, but I'd appreciate comments on those remarks. Until then, this is still WIP

Aaronontheweb:
Thanks, I'll take a look

src/core/Akka.Persistence/Eventsourced.Recovery.cs
```diff
@@ -17,17 +17,15 @@ namespace Akka.Persistence
...
             Name = name;
             IsRecoveryRunning = isRecoveryRunning;
             StateReceive = stateReceive;
         }
 
         public string Name { get; }
-
-        public bool IsRecoveryRunning { get; }
-
+        public Func<bool> IsRecoveryRunning { get; }
```
ismaelhamed: This breaks the public API

src/core/Akka.Persistence.Tests/PersistentActorSpec.cs
```diff
@@ -611,5 +611,21 @@ public void PersistentActor_should_brecover_the_message_which_caused_the_restart
...
+        public void PersistentActor_should_be_able_to_persist_events_that_happen_during_recovery()
+        {
+            var persistentActor = ActorOf(Props.Create(() => new PersistInRecovery(Name)));
+            persistentActor.Tell(GetState.Instance);
+            ExpectMsgInOrder("a-1", "a-2", "rc-1", "rc-2");
+            persistentActor.Tell(GetState.Instance);
+            ExpectMsgInOrder("a-1", "a-2", "rc-1", "rc-2", "rc-3");
+            persistentActor.Tell(new Cmd("invalid"));
+            persistentActor.Tell(GetState.Instance);
+            ExpectMsgInOrder("a-1", "a-2", "rc-1", "rc-2", "rc-3", "invalid");
```
Aaronontheweb: @ismaelhamed shouldn't there be an `invalid-recovery` event recovered here?

src/core/Akka.Persistence/Eventsourced.Recovery.cs
Aaronontheweb: Nah, this is an `internal` API. None of the third party Akka.Persistence plugins have access to this so we're able to preserve binary compatibility here. It's only the Akka.Persistence engine itself that consumes this directly so I think you're fine.

src/core/Akka.Persistence/Eventsourced.Recovery.cs
```diff
@@ -159,85 +152,90 @@ private EventsourcedState Recovering(Receive recoveryBehavior, TimeSpan timeout)
...
             var eventSeenInInterval = false;
+            var recoveryRunning = true;
 
-            return new EventsourcedState("replay started", true, (receive, message) =>
+            return new EventsourcedState("replay started", () => recoveryRunning, (receive, message) =>
             {
                 try
                 {
-                    if (message is ReplayedMessage)
+                    switch (message)
```
Aaronontheweb: I like this change

src/core/Akka.Persistence/Eventsourced.Recovery.cs
Aaronontheweb: Question: how often is the `EventsourcedState` object allocated? From a functional standpoint it should be fine, but if we allocate a new `EventsourcedState` instance regularly then anonymous method allocations could add up. I'm not familiar enough with Akka.Persistence's internals to know off-hand. Do you or @Horusiath have some idea?

Aaronontheweb:
Left some review notes for you @ismaelhamed

src/core/Akka.Persistence.Tests/PersistentActorSpec.cs
```diff
@@ -611,5 +611,21 @@ public void PersistentActor_should_brecover_the_message_which_caused_the_restart
...
         }
+
+        [Fact]
+        public void PersistentActor_should_be_able_to_persist_events_that_happen_during_recovery()
+        {
+            var persistentActor = ActorOf(Props.Create(() => new PersistInRecovery(Name)));
+            persistentActor.Tell(GetState.Instance);
+            ExpectMsgInOrder("a-1", "a-2", "rc-1", "rc-2");
+            persistentActor.Tell(GetState.Instance);
+            ExpectMsgInOrder("a-1", "a-2", "rc-1", "rc-2", "rc-3");
```
ismaelhamed: There's some race condition (at least in the JVM) between the GetState and the PersistAsync, so they've simplified the test like this:

```expectMsgAnyOf(List("a-1", "a-2", "rc-1", "rc-2"), List("a-1", "a-2", "rc-1", "rc-2", "rc-3"))```

But I'm not sure this is possible with the current ExpectMsgAnyOf in Akka.NET

src/core/Akka.Persistence/Eventsourced.Recovery.cs
ismaelhamed: @Aaronontheweb not sure about this one, honestly. Maybe @Horusiath wants to weigh in.

Aaronontheweb:
Bleh, I forgot to come back and revisit this - will do

src/core/Akka.Persistence.Tests/PersistentActorSpec.cs
ismaelhamed: @Aaronontheweb about this?

Aaronontheweb: I saw the comment - the test passes currently, right? If it's going to be racy, do we need to implement some new features in the `TestKit` in order to do this correctly? What's the issue?

ismaelhamed: The test runs fine, at least every time I've ran it. But I noticed they (JVM) changed it since the original PR for the reason I mentioned, just so you know.

Aaronontheweb:
Thanks for your contribution @ismaelhamed

PR #3373, opened 2018-03-26
Add an Ask 'overload' that allows message generation with `IActorRef` inside

boekabart:
Change the inner `Ask` implementation to accept a message factory instead of a message, so the message can optionally encapsulate the sending `IActorRef` - the address of the light actor inside `Ask`. Only the most complex version of the public `Ask` extension (with Timeout and Cancellation params) also exposes this Factory option. Existing `Ask` extension methods proxy through this new one.

The reasoning behind this:
In the support framework of our application we sometimes need to encapsulate the 'requester' of a message explicitly into the payload, rather than rely on 'Sender'. Two cases stand out: Forwarding (where the generic-ness of the helper prevents us from using `ref.Forward` rather than `ref.Tell`, and the other case is merging/grouping requests -> multiple intended recipients of the answer.

When unit testing actors that expect such requests, we currently have to 'proxy' the message injected into Akka by `.Ask()` via a proxy actor that only inserts `Sender` into the correct field - a big hassle that makes tests harder to read and maintain.

We want to be able to just `sut.Ask<Response>( askRef => new Request( "question", respondTo = askRef) )`

The change is fairly trivial, and given the cost of `Ask` in general, the cost of the extra indirection to use the current `Ask` is negligible.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -883,7 +883,8 @@ namespace Akka.Actor
         public static System.Threading.Tasks.Task<object> Ask(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout, System.Threading.CancellationToken cancellationToken) { }
         public static System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout = null) { }
         public static System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Threading.CancellationToken cancellationToken) { }
-        public static async System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout, System.Threading.CancellationToken cancellationToken) { }
+        public static System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout, System.Threading.CancellationToken cancellationToken) { }
+        public static async System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, System.Func<Akka.Actor.IActorRef, object> messageFactory, System.Nullable<System.TimeSpan> timeout, System.Threading.CancellationToken cancellationToken) { }
     }
     public class static GracefulStopSupport
     {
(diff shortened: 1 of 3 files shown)


src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -883,7 +883,8 @@ namespace Akka.Actor
         public static System.Threading.Tasks.Task<object> Ask(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout, System.Threading.CancellationToken cancellationToken) { }
         public static System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout = null) { }
         public static System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Threading.CancellationToken cancellationToken) { }
-        public static async System.Threading.Tasks.Task<T> Ask<T>(this Akka.Actor.ICanTell self, object message, System.Nullable<System.TimeSpan> timeout, System.Threading.CancellationToken cancellationToken) { }
```
boekabart: IMO `async` keyword should *not* be part of the public API description - it's a hint for the compiler of the implementation, and not part of the public API. line 886 is therefore not an API change at all.

maxcherednik:
@boekabart I did no quite get the use case of it. Could you please give an example?

boekabart:
@maxcherednik added to description

Danthar:
What is it you are trying achieve here? Because to me, it seems like you are trying to expose the same functionality as the Context.Sender property that the receiving actor already provides(?).
Because in the receiving actor, the Context.Sender will be the lightweight Ask actor that send the message.

And in the case you don't want to respond back to the sender, but another Actor. Passing along the IActorRef in the message your sending, is the standard pattern for this.
And if you are calling Ask from inside an actor, and want to have the response go back to that actor, use the `PipeTo` construct.

boekabart:
@Danthar apparently my description skills aren't as good as I thought.

> And in the case you don't want to respond back to the sender, but another Actor. Passing along the IActorRef in the message your sending, is the standard pattern for this.

That's exactly what we're trying to do, in fact.
I was under the impression that the description explained the 2 main use cases where using `Context.Sender` as the mechanism isn't feasible for the Generic Framework Actors we're building (not because we couldn't access it, but because of the need to compose and/or forward requests. If in some cases we'd have to use Sender, and in some cases a message-encapsulated 'sender', we'd either need to duplicate the message classes to one with, and one without sender, OR build in 'fallback logic' and copy rather than forward messages. 
We wish to do neither.

Our use cases for `Ask`, by the way, are restricted to Outside-Of-Akka -2- Akka (e.g. API Controller), and of course Unit Tests.

So, what we WANT to do is be able to send requests with a message that contains the intended recipient. And in SOME cases (some unit tests, and some Controller implementations) it's just much more efficient if we can use `Ask` in combination with that pattern - but for that, we need to be able to send a message with the `Ask`ers IActorRef embedded

Danthar:
LGTM

PR #3380, opened 2018-04-04
Clarifying companion object documentation.

nathvi:


nathvi:
What was wrong with this code?

Danthar:
Nothing, MNTK testrunner can be racy at times, ill reschedule.

PR #3384, opened 2018-04-08
Updated example on how to send message to actors behind router

m-wilczynski:
Updated example on how to send message to actors behind router.
This might be trivial for those who already use Akka.NET but for newcomers it's not so obvious (wasn't for me), that you simply communicate with it via its `IActorRef` and then it sends the messages to actors' mailboxes according to its routing strategy.

### DIFF
--- a/docs/articles/actors/routers.md
+++ b/docs/articles/actors/routers.md
@@ -96,6 +96,27 @@ Routers are implemented as actors, so a router is supervised by it's parent, and
 
 By default, pool routers use a custom strategy that only returns `Escalate` for all exceptions, the router supervising the failing worker will then escalate to it's own parent, if the parent of the router decides to restart the router, all the pool workers will also be recreated as a result of this.
 
+## Sending messages to actors behind router
+
+As routers are actors, we send over messages to actors they manage the same way we communicate to other actors via `ActorRef`:
+
+```cs
+var router = system.ActorOf(Props.Create<Worker>().WithRouter(new RoundRobinPool(3)), "some-pool");
+foreach (var element in Enumerable.Range(0, 6))
+{
+  router.Tell(new Message(element));
+}
+```
+
+Router will send over messages according to its strategy.
+In this particular case, we have `RoundRobinPool` configuration that will:
+* create 3 instances (because it's Pool strategy - see above) of `Worker` actor 
+* enqueue<sup>1</sup> total of 6 messages of type `Message` to actors **mailboxes** it manages according to its RoundRobin routing strategy
+
+Final outcome should be 2 messages per each actor's mailbox (see how RoundRobin strategy works below).
+
+<sup>1</sup> *assuming that `Worker` actor is configured to `Receive<Message>`*
+
 ## Routing Strategies
 
 These are the routing strategies provided by Akka.NET out of the box.


Danthar:
I don't understand why this is relevant information for a normal user? Sure, its not explicitly explained, but its being implied almost everywhere. Since at the beginning of the article, its being explained that routers are Actors themselves. And because in the base documentation for Akka it is already stated that Actors can only communicate with each other through `IActorRef`. 
Your addition seems superfluous to me.  And even then, the documentation you added contains some incorrect statements. I'm not against explaining this detail more explicitly, but then lets at least get it right ;).

Ill add some inline comments.

docs/articles/actors/routers.md
```diff
@@ -96,6 +96,27 @@ Routers are implemented as actors, so a router is supervised by it's parent, and
 
 By default, pool routers use a custom strategy that only returns `Escalate` for all exceptions, the router supervising the failing worker will then escalate to it's own parent, if the parent of the router decides to restart the router, all the pool workers will also be recreated as a result of this.
 
+## Sending messages to actors behind router
```
Danthar: This headline is wrong. It implies that you are explaining how to directly send messages to a Router's `Routees` . This is not true.
Its better to change it to something like: "How routers communicate with their routee's"

docs/articles/actors/routers.md
```diff
@@ -96,6 +96,27 @@ Routers are implemented as actors, so a router is supervised by it's parent, and
...
+
+As routers are actors, we send over messages to actors they manage the same way we communicate to other actors via `ActorRef`:
```
Danthar: Same problem as above. It would be better if it said: "As routers are actors, they communicate with their `Routees` the same way normal `Actors` would communicate with each other.

docs/articles/actors/routers.md
```diff
@@ -96,6 +96,27 @@ Routers are implemented as actors, so a router is supervised by it's parent, and
...
+```cs
+var router = system.ActorOf(Props.Create<Worker>().WithRouter(new RoundRobinPool(3)), "some-pool");
+foreach (var element in Enumerable.Range(0, 6))
+{
+  router.Tell(new Message(element));
+}
+```
+
+Router will send over messages according to its strategy.
+In this particular case, we have `RoundRobinPool` configuration that will:
```
Danthar: In this particular case, we **use the** `RoundRobinPool` configuration that will:

docs/articles/actors/routers.md
```diff
@@ -96,6 +96,27 @@ Routers are implemented as actors, so a router is supervised by it's parent, and
...
+* create 3 instances (because it's Pool strategy - see above) of `Worker` actor
```
Danthar: create 3 instances (because it's **a** Pool strategy - see above) of **the** `Worker` actor

docs/articles/actors/routers.md
```diff
@@ -96,6 +96,27 @@ Routers are implemented as actors, so a router is supervised by it's parent, and
...
+foreach (var element in Enumerable.Range(0, 6))
+{
+  router.Tell(new Message(element));
+}
+```
+
+Router will send over messages according to its strategy.
+In this particular case, we have `RoundRobinPool` configuration that will:
+* create 3 instances (because it's Pool strategy - see above) of `Worker` actor 
+* enqueue<sup>1</sup> total of 6 messages of type `Message` to actors **mailboxes** it manages according to its RoundRobin routing strategy
```
Danthar: ..... **Tell** (not enqueue)  **a** total of 6 messages ....... to its `routees` according to its RoundRobin routing strategy.

Dont use words like Enqueue and Mailboxes here. Because its a detail of the IActorRef implementation. Routee's are nothing more then IActorRef's. Whether or not messages are enqueued or not, is a detail of the IActorRef implementation you should not mention here, because the Router Actor has no knowledge of these details.

Same goes for mailboxes. What if the IActoRef is actually a RemoteActorRef because the routees being used are actually remote actors ? A lot more stuff is happening then. Again, its an implementation detail the Router Actor has no knowledge about so don't describe that here.

m-wilczynski:
Thanks for the review.
I too found some of the statements in my commit 'slippery' but couldn't find an appropriate way to articulate them correctly (probably due to only basic experience with actor model).
Main reason for this PR is to make docs more accessible for reader to actual use (in code) of the concept from article.
Think docs.microsoft.com example on .NET articles - you have general concept (the "why" part), code sample on how the usage looks like (the "how" part) and result and either screenshot or written description on what happens afterwards (the "outcome" part).
I guess we have great stuff on "why" part, great visual part and written part of "outcome" part (visualization of strategies makes them really approachable), yet we miss the "how" (or plumbing) part.

I'll fix parts you mentioned as I agree to all of them.
I'd also add link to code example for Routing on GitHub (https://github.com/akkadotnet/akka.net/tree/dev/src/examples/Routing). 
I almost missed it and found it only because other article mentiones Chat example with direct link to src/examples directory on GitHub.

Danthar:
👍 Adding the link is a good idea.

Danthar:
@m-wilczynski :nudge nudge: ;)

m-wilczynski:
Sry, I've got lots of stuff going on recently. :(
I'll try to push my fixes on Sat/Sun.

Issue #3318, opened 2018-02-07
Floating point to long conversion in BackoffSupervisor::CalculateDelay could overflow


ianclegg:
While porting replyWhileStopped in #3317 for I noticed a risky double to long conversion using an explicit cast. With a large value the long will overflow and we'll have a negative timespan.

```c#
var max = Math.Min(maxBackoff.Ticks, minBackoff.Ticks * Math.Pow(2, restartCount)) * rand;
if (max >= double.MaxValue)
{
   return maxBackoff;
}
else
{
    return new TimeSpan((long)max);
}
```

Ending up with a timespan large enough to overflow is probably a user error in itself, but it shouldn't be exacerbated by Akka.

It's a minor fix, happy to raise a PR for it if your interested

marcpiechura:
@ianclegg sure thing, go ahead !

sean-gilliam:
Closed via #3678

PR #3385, opened 2018-04-10
Adding fix for issue #3318 (Possible negative TimeSpan in CalculateDelay Function)

nathvi:
Adding a fix for the possible negative TimeSpan issue references in issue #3318 (Possible negative TimeSpan).


src/core/Akka/Pattern/BackoffSupervisor.cs
```diff
@@ -227,7 +227,10 @@ public static Props Props(BackoffOptions options)
                 }
                 else
                 {
-                    return new TimeSpan((long)max);
+                    long maxAsLong = (long)max; 
+                    bool maxAsLongIsnegative = maxAsLong >= 0? true: false;
+                    if (maxAsLongIsnegative) {throw new ArgumentOutOfRangeException($"{nameof(maxAsLong)} cannot be negative.");}
```
marcpiechura: I‘m not sure why we should throw an exception, a simple
 `new TimeSpan( max > Long.MaxValue ? Long.MaxValue : max ) `
Should be enough.

marcpiechura: And when I look at the code a little bit closer I don’t think the fix is neccassary at all and the if conditions will also never be evaluate to true :-) 
`Math.Min`  checks `TimeSpan.Tick` which is already a `Long` so `max` can never be greater then `Long.MaxValue` and therefore never be `>= Double.MaxValue`

marcpiechura: So I would propose to remove the if else branch and simply return `TimeSpan((long)max)` as before , maybe with a comment that `Math.Min` already checks for possible overflows. 

In case I‘m missing something CC @akkadotnet/core

nathvi: @marcpiechura , I think you are missing something, as it is possible to have a negative value.

If you make the following call:
`TimeSpan delay  = BackoffSupervisor.CalculateDelay(1, 
                TimeSpan.FromSeconds(10), 
                TimeSpan.FromSeconds(10), 10);`

You will get a negative result for `maxAsLongIsnegative`

nathvi: Ok, giving the correct way to compare a negative... 
`bool maxAsLongIsnegative = maxAsLong < 0? true: false;`

You can still get in a situation where the result of `maxAsLong` is negative.
Consider the situation when you call:
`TimeSpan delay = BackoffSupervisor.CalculateDelay(1,
                TimeSpan.FromSeconds(10),
                TimeSpan.FromSeconds(10), double.MinValue);`

Danthar:
Wouldn't it be quicker to replace: ` if (max >= double.MaxValue)` with ` if (max >= long.MaxValue)`

The the cast `(long)max` will be guaranteed to not fail.

Aaronontheweb:
@marcpiechura have you had a chance to go back through these changes?

marcpiechura:
@nathvi I would propose to do the following, 
- check whether `randomFactor` is negativ, if so throw an `InvalidArgumentException`
- change the `if (max >= double.MaxValue)` check as suggested by @Danthar to `if (max >= long.MaxValue)`
- leave the `else` branch as is, i.e. `return new Timespan((long)max)`

This would give one the correct error if you use a negative random factor and prevents the overflow.

Danthar:
@nathvi have you seen @marcpiechura comments?

nathvi:
@Danthar , I have, just been busy with other stuff.

ismaelhamed:
Closed via #3678

sean-gilliam:
Thank you for your contribution @nathvi, but it looks like #3678 bet you to the punch. Closing.

PR #3391, opened 2018-04-14
[WIP] getting .NET Core on Linux Specs working

Aaronontheweb:


Aaronontheweb:
Digging into this locally... Fails because if this bit of code:

https://github.com/akkadotnet/akka.net/blob/3f924fda8d6ca55c5af90a710d23be8ce36d5c7a/src/core/Akka/Util/Internal/NetCorePolyfills.cs#L8-L19

.NET Core 2.0 brings the `SerializableAttribute` back, so we'll need to adjust the pre-compilation headers we use for backwards compatibility in order to get these specs to compile under .NET Core 2.0.

Aaronontheweb:
closing in favor of #3400

PR #3400, opened 2018-04-18
changed version of .NET Core SDK being used on Linux

Aaronontheweb:
Trying to solve the same problem as #3391, but from a different angle.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -11,7 +11,7 @@ NUGET_URL=https://dist.nuget.org/win-x86-commandline/v4.3.0/nuget.exe
 FAKE_VERSION=4.63.0
 FAKE_EXE=$TOOLS_DIR/FAKE/tools/FAKE.exe
 DOTNET_EXE=$SCRIPT_DIR/.dotnet/dotnet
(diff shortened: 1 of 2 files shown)


Danthar:
@Aaronontheweb Is this still a thing ?

Aaronontheweb:
@Danthar yeah, for the 1.3.* branches at least. Gets us some Linux tests again now that we're no longer doing Mono. You should still move forward with https://github.com/akkadotnet/akka.net/pull/3452 for the v1.4.0 rollout though.

Aaronontheweb:
@Danthar I think I have a fix in-place here to resolve that `ClusterSpec` that failed on Linux... Issue is that a cancelled task wasn't actually throwing a `TaskCancellationException`... Has to be a platform difference issue, I think. We'll see ❓

Aaronontheweb:
Threw in the towel on that spec -had to mute it on Mono for the exact same reason. Something in the underlying code doesn't throw on Linux when the cancellation token is invoked. So, guess this is good to go now.

PR #3406, opened 2018-04-19
Unit test Akka.FSharp.System.create with extensions

aaronpalmer:
Add a unit test to demonstrate a bug with Akka.FSharp.System.create when creating with a configuration that has akka.extensions specified.


AaronEshbach:
I think your unit test is wrong.  It needs to end with an XUnit.Assert.NotEqual rather than returning a boolean value.

aaronpalmer:
Based on Aaron Eshbach's comment I am closing this and will submit an updated PR

PR #3407, opened 2018-04-19
Unit test Akka.FSharp.System.create with extensions

aaronpalmer:
Add a unit test to demonstrate a bug with Akka.FSharp.System.create when creating with a configuration that has akka.extensions specified.

### DIFF
--- a/src/core/Akka.FSharp.Tests/Akka.FSharp.Tests.fsproj
+++ b/src/core/Akka.FSharp.Tests/Akka.FSharp.Tests.fsproj
@@ -16,6 +16,7 @@
 
   <ItemGroup>
     <Content Include="App.config" />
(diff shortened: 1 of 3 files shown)


AaronEshbach:
I did a little research, and there is a conflicting serializer ID.  Both the Cluster Tools DistributedPubSubMessageSerializer and the FSharp ExprSerializer use the same identifier, 9.  I think changing the ID for one of these serializers would fix this unit test.

Danthar:
Thx for your contribution

Aaronontheweb:
@AaronEshbach did that issue with the colliding serializer IDs get resolved? Or should we file a new issue for that?

AaronEshbach:
We tested Akka.FSharp with Akka.Cluster.Tools at version 1.3.8, and the issue appears to be resolved.

Issue #3409, opened 2018-04-20
Incompatibility with DotNetty 0.4.7


vasily-kirichenko:
```
13:31:11 FATAL [DomainUnhandled, 1] Domain unhandled exception of type AggregateException occured (System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> System.AggregateException: One or more errors occurred. ---> System.MissingMethodException: Method not found: 'Void DotNetty.Buffers.UnpooledByteBufferAllocator..ctor()'.
   at Akka.Remote.Transport.DotNetty.DotNettyTransport.ServerFactory()
   at Akka.Remote.Transport.DotNetty.DotNettyTransport.<NewServer>d__20.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Akka.Remote.Transport.DotNetty.DotNettyTransport.<Listen>d__21.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Akka.Remote.Transport.DotNetty.DotNettyTransport.<Listen>d__21.MoveNext()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Akka.Remote.Transport.AbstractTransportAdapter.<>c__DisplayClass11_0.<<Listen>b__0>d.MoveNext()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Akka.Remote.EndpointManager.<>c__DisplayClass48_1.<get_Listens>b__4(Task`1 result)
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.Tasks.Task.Execute()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at Akka.Remote.EndpointManager.<>c.<get_Listens>b__48_5(Task`1 transportResults)
[...]

PR #3410, opened 2018-04-20
Pin DotNetty version to 0.4.6

vasily-kirichenko:
This fixes https://github.com/akkadotnet/akka.net/issues/3409


Aaronontheweb:
We really need DotNetty v0.5.0 to ship, but I have no idea when they're going to get around to that.

Aaronontheweb:
Thanks @vasily-kirichenko

Issue #265, opened 2014-08-15
InboxActor: EnqueueQuery and other methods should be private. Rewrite Receive


HCanber:
### Public to private

`InboxActor` has many public methods (although the class is internal) that should not be called externally (and they are not currently).
For example, if `EnqueueQuery` was to be called externally it can introduce race conditions (`_clients` is updated but not `_clientsByTimeout`).
`MessagePredicate` and `EnqueueMessage` have concurrent problems as well.
Make all public methods private.
### Receive

Core actors should not be using `PatternMatch` (or `ReceiveActor`) due to performance and GC overhead. Rewrite `Receive` to raw C# by using `is` or `as`:

``` c#
var getMessage=message as Get;
if(getMessage != null)
{
   ...
}
```

nbparrell:
I will be working on this.

nbparrell:
@HCanber Just want to clarify, where you are asking for the rewrite of receive. Is that in the InboxActor or in the receiveActor that you are asking for the change?

Aaronontheweb:
@nbparrell `InboxActor`

Aaronontheweb:
nice @nbparrell ! Send in a PR - easier to comment there than on an issue.

nbparrell:
where is the PR...... lol
Alittle new to the contributing scene.

Aaronontheweb:
@nbparrell all good!

https://help.github.com/articles/about-pull-requests/ - this explains how to send in a pull request along with a little walkthrough. That should help!

nagytech:
Ready for review in #4073

Aaronontheweb:
closed via #4073

PR #3412, opened 2018-04-21
Inbox.Actor and String Like refactoring

nbparrell:
These are the changes asked for by the ticket https://github.com/akkadotnet/akka.net/issues/265

Made all public methods private and also rewrote the Receive so that it is using is with if else instead of a switch.

### DIFF
--- a/src/core/Akka/Actor/Inbox.Actor.cs
+++ b/src/core/Akka/Actor/Inbox.Actor.cs
@@ -105,37 +105,32 @@ namespace Akka.Actor
         /// <returns>TBD</returns>
         protected override bool Receive(object message)
         {
-            /*var getMessage = message is Get;
-            var selectMessage = message is Select;
-            var startWatch = message is StartWatch;
-            var stopWatch = message is StopWatch;
-            var kick = message is Kick;*/
-            //var 
-            //Console.WriteLine("Entered the switch");
-            if(message != null){
-                if(message is Get){
+            if(message != null)
+            {
+                if(message is Get get)
+                {
                     if (_messages.Count == 0)
                     {
-                        EnqueueQuery((Get)message);
+                        EnqueueQuery(get);
                     }
                     else
                     {
                         Sender.Tell(_messages.Dequeue());
                     }
                 }
-                else if(message is Select){
-                    var selectMessage = (Select)message;
+                else if(message is Select select)
+                {
                     if (_messages.Count == 0)
                     {
-                        EnqueueQuery(selectMessage);
+                        EnqueueQuery(select);
                     }
                     else
                     {
-                        _currentSelect = selectMessage;
+                        _currentSelect = select;
                         var firstMatch = _messages.DequeueFirstOrDefault(MessagePredicate);
                         if (firstMatch == null)
                         {
-                            EnqueueQuery(selectMessage);
+                            EnqueueQuery(select);
                         }
                         else
                         {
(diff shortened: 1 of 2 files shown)


src/core/Akka/Actor/Inbox.Actor.cs
```diff
@@ -105,49 +105,57 @@ public bool MessagePredicate(object message)
         /// <returns>TBD</returns>
         protected override bool Receive(object message)
         {
-            switch (message)
-            {
-                case Get get:
+            /*var getMessage = message is Get;
+            var selectMessage = message is Select;
+            var startWatch = message is StartWatch;
+            var stopWatch = message is StopWatch;
```
marcpiechura: Please remove comments

src/core/Akka/Actor/Inbox.Actor.cs
```diff
@@ -105,49 +105,57 @@ public bool MessagePredicate(object message)
...
+            var kick = message is Kick;*/
+            //var 
+            //Console.WriteLine("Entered the switch");
+            if(message != null){
```
marcpiechura: No need for null check, „is“ handles it automatically.

src/core/Akka/Actor/Inbox.Actor.cs
```diff
@@ -105,49 +105,57 @@ public bool MessagePredicate(object message)
...
+                if(message is Get){
                     if (_messages.Count == 0)
                     {
-                        EnqueueQuery(get);
+                        EnqueueQuery((Get)message);
```
marcpiechura: Please use the new syntax and define a variable name for the casted get directly in the if statement, I.e. „ if ( message is Get get)“

src/core/Akka/Actor/Inbox.Actor.cs
marcpiechura: And please don’t use java style for the curlybraces, put them in the next line.

src/core/Akka/Actor/Inbox.Actor.cs
```diff
@@ -5,69 +5,69 @@
...
 {
     /// <summary>
     /// TBD
     /// </summary>
-    internal class InboxActor : ActorBase
+    internal class InboxActor:ActorBase 
     {
-        private readonly InboxQueue<object> _messages = new InboxQueue<object>();
-        private readonly InboxQueue<IQuery> _clients = new InboxQueue<IQuery>();
+        private readonly InboxQueue < object > _messages = new InboxQueue < object > ();
```
marcpiechura: Something is wrong with your auto formatting rules, we don't want to have a space in the generic type parameter ;)

src/core/Akka/Util/StringLike.cs
```diff
@@ -26,7 +25,15 @@ public static bool Like(this string text,string pattern, bool caseSensitive = fa
...
-            pattern = pattern.Replace("*", ".*?");
+            if(pattern[0] == '*' && pattern[pattern.Length-1]!='*'){
+                pattern = pattern.Replace("*", "(.*?)");
+                pattern = pattern + "(?!.)";
+            }else if(pattern[0]!='*' && pattern[pattern.Length-1]=='*'){
+                pattern = pattern.Replace("*", "(.*?)");
+                pattern = "^"+pattern;
+            }else{
+                pattern = pattern.Replace("*", "(.*?)");
+            }
```
marcpiechura: Could you please move this into its own PR ? Makes its easier to reason about the changes.

nbparrell: Not sure what you mean by PR? Not completely up on the short hand.

src/core/Akka/Actor/Inbox.Actor.cs
nbparrell: Will do.

src/core/Akka/Util/StringLike.cs
marcpiechura: Soory, PR means Pull Request. In the end it would be good to have a pull request per issue it handles.

marcpiechura:
@nbparrell just wanted to know if you’re still working on this one.

nbparrell:
@marcpiechura Yeah, I will be finishing this, as the summer started things got kinda hectic. Will make changes asap.

Danthar:
@nbparrell minor poke ;)

Aaronontheweb:
Replaced via #4073

Issue #3422, opened 2018-04-26
Persistence: Slow reading of big snapshots


TietoOliverKurowski:
We are seeing very slow performance when reading big snapshots on SQL Server. 

Snapshots of some of your actors are in the range of over 100MB. Reading that snapshot takes around 15 **minutes** and causes 30% CPU usage.
I traced the issue down to the way the data reader is created:

`using (var reader = await command.ExecuteReaderAsync(cancellationToken))`

when I use the option `CommandBehavior.SequentialAccess` reading only takes around 2 **seconds** and no CPU usage.

Would it be possible to add that parameter?

- Which Akka.Net version you are using: 1.3.6
- Akka.Persistence.SqlServer: 1.3.2
- On which platform you are using Akka.Net: .Net 4.7.1, targeting 4.5.2

In parallel I'm working in getting the snapshot smaller.

Aaronontheweb:
Thanks for reporting this! And yep, we can send in a PR to https://github.com/akkadotnet/Akka.Persistence.SqlServer to get that fixed.

Aaronontheweb:
Although actually, that might be an issue with the `Persistence.Sql.Common` middleware in this project

Horusiath:
I can do it, but it may take me some time (I'm still working on other stuff). I think, it's something that could be even picked up as first contribution. Essentially changes could be closed inside Akka.Persistence.Sql.Common project. 

IMO sequential access should be a configurable setting (example path: `akka.persistence.snapshot-store.<sql-plugin>.use-sequential-access = true|false`). If not provided, it should default to false to keep existing behavior intact.

[AbstractQueryExecutor.SelectSnapshotAsync](https://github.com/akkadotnet/akka.net/blob/9c1bc9fbe980335a1bc34d61a02ab4afd4203a37/src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs#L469) is the place for applying command behavior. Probably an additional parameter will be necessary.

Also this change most probably will hit all existing Sql providers in terms of backward compatibility.

Aaronontheweb:
@TietoOliverKurowski, would you be interested in contributing this change? Otherwise one of us on the @akkadotnet/core team can pick it up.

TietoOliverKurowski:
Yes, I want to give it a try. By end of next Friday I should have something working.

PR #3429, opened 2018-05-02
#3422 SQL persistence providers: Data reader SequentialAccess

TietoOliverKurowski:
Configuration option for SQL persistence providers: Data readers can use Sequential access, which improves reading speed for large payloads.

- Added for Snapshot query
- Added for journal queries

This change is not backwards compatible.

As far as I can see there are no tests for the Akka.Persistence.SQL.Common project, so I didn't add any new test.
I also prepared a change for the SQL Server provider, but I wanted to wait for your review comments before publishing it.
Please let me know of any thinks I still need to do. This is my first pull request, so everything is quite new ;)


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -171,6 +171,11 @@ public class QueryConfiguration
         /// </summary>
         public string DefaultSerializer { get; }
 
+        /// <summary>
+        /// Uses the CommandBehavior.SequentialAccess when creating the command, providing a performance improvement for reading large BLOBS.
+        /// </summary>
+        public readonly bool UseSequentialAccess;
```
Horusiath: We use auto-get properties.

TietoOliverKurowski:
Thanks @Horusiath for the quick code review. I updated according to you suggestions.

Aaronontheweb:
Related issue: https://github.com/akkadotnet/akka.net/issues/3422

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
```diff
@@ -147,6 +154,7 @@ public class QueryConfiguration
             SerializerIdColumnName = serializerIdColumnName;
             Timeout = timeout;
             DefaultSerializer = defaultSerializer;
+            UseSequentialAccess = useSequentialAccess;
```
Aaronontheweb: What's the default value of this? Is it left up to the plugin implementation? Should we add it to some of the built-in HOCON files?

Aaronontheweb:
@TietoOliverKurowski did you see my comments?

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
TietoOliverKurowski: Sorry for the late response.
Default value is false = current behavior
As the configuration is read by the plugin implementations it is basically up to the plugins. Therefore I'm not sure if it makes sense to put into some HOCO file on Akka level.
So I'm not sure if further changes are needed.

Aaronontheweb:
Ok, got it - so this is a breaking change from an API perspective and will require us to re-compile any Akka.Persistence plugins that use SQL. Not the end of the world. We'll go ahead and do that.

TietoOliverKurowski:
Great, thanks a lot!

Issue #3431, opened 2018-05-03
Akka.Persistence LocalSnapshotStore file open race conditions


Aaronontheweb:
Version: Akka.NET v1.3.6

Looks like, based on some persistence IDs that are used and the presence of special characters like `/` or `\` might throw `System.IO.FileNotFoundException: Unable to find the specified file.` exceptions inside the `LocalSnapshotStore`.

I'm thinking that we need to either come up with some way of encoding the PersistenceID safely or throw a more explicit exception if that's the case. Going to add a spec to reproduce first.

Horusiath:
If I remember correctly Peristent ID is escaped using URI data escape method. Maybe we need something stronger?

Aaronontheweb:
@Horusiath yeah, I need to rename the issue - after looking at the logs looks like the issue is actually a bit of a race condition - tried to write to the file before it was open.

Aaronontheweb:
https://github.com/akkadotnet/akka.net/blob/9c1bc9fbe980335a1bc34d61a02ab4afd4203a37/src/core/Akka.Persistence/Snapshot/LocalSnapshotStore.cs#L202-L214

Looks like the `MoveTo` command is failing for some reason.

Aaronontheweb:
some possibilities I am considering here:

* Some other process is mangling the temp files before they have a chance to be moved;
* There are multiple competing snapshot stores running on the same machine;
* Etc...

I do see an error that looks like this elsewhere in the same set of logs:

```
2018-04-23 18:06:11 [Error] (akka://someSys/someActor) Unable to save ReliableMailmanActor snapshot with id "foo"
System.IO.IOException: The process cannot access the file 'C:\Program Files (x86)\BD\MMS Fk\Mms.Fk.UpstreamProxy\snapshots\snapshot-foo-0-636601323711494050.tmp' because it is being used by another process.
```

I'm going to see if I can rule out that this is an issue with one isolated `LocalSnapshotStore` happily doing its thing on its own and see if we can determine if it was an issue with the end-user environment that caused this.

Aaronontheweb:
The issue is that since we’re not incrementing the sequence number (using the Journal) in Akka.Persistence, two snapshots taken close together are going to have the same filename since the timestamp will also match. Given that the `LocalSnapshotStore` processes snapshot save requests concurrently, we’re going to run into trouble if the filenames are identical, and that’s what is happening here.

[...]

PR #3432, opened 2018-05-03
Akka.Persistence XML-DOC and cosmetic cleanup

Aaronontheweb:
Working on #3431


Aaronontheweb:
@Horusiath ok, got this test to fail pretty easily.

```
Expected True because expected all snapshots to be saved, but found 21 that were not. Output: SaveSnapshotFailure<meta: SnapshotMetadata<pid: $a, seqNr: 0, timestamp: 0001/01/01>, cause: System.IO.IOException: The process cannot access the file 'C:\Repositories\olympus\akka.net\src\core\Akka.Persistence.Tests\bin\Debug\net452\snapshots\snapshot-%24a-0-636609739579544082.tmp' because it is being used by another process.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy)
   at System.IO.FileStream..ctor(String path, FileMode mode)
   at Akka.Persistence.Snapshot.LocalSnapshotStore.WithOutputStream(SnapshotMetadata metadata, Action`1 p)
   at Akka.Persistence.Snapshot.LocalSnapshotStore.Save(SnapshotMetadata metadata, Object snapshot)
   at Akka.Persistence.Snapshot.LocalSnapshotStore.<>c__DisplayClass10_0.<SaveAsync>b__0()
   at Akka.Persistence.Snapshot.LocalSnapshotStore.<>c__DisplayClass28_0`1.<RunWithStreamDispatcher>b__0()
```

Worst part is that this stuff still returns `SaveSnapshotSuccess` messages even when this stuff fails. So it looks like the issue here is multiple async tasks trying to write to the same temporary file at the same time. There should be something inside the journal that can synchronize writes for the same snapshot.

Aaronontheweb:
Going to clean up this PR and just pass in my XML-DOC and code cleanups. I don't think #3431 is a bug. I think it's an unsupported edge case. I'll pass along a writeup for why on that issue momentarily.

Aaronontheweb:
Reverted the spec that was added to test #3431. Don't need it.

PR #3436, opened 2018-05-05
Akka.Remote memory leak fix: use singleton allocators for DotNetty channels.

### LINKED ISSUE #3430: Increase Server Memory When Client program repeats ActorSelection and Terminated
Akka version: 1.3.6
Akka.Remote: 1.3.6 (with DotNetty 0.4.6)
Platform: .Net Framework 4.7
Project link: https://github.com/iris4865/IssueStorages/tree/master/AkkaRemote_1.3.6/ActorSelection

Server memory usage has increased When Client program connects and disconnects. At the start of the program, the memory was 35.85MB. ActorSelection test result increased to 1.88GB. I did profiling with JetBrains DotMemory 2018.1. I think that Dotnetty.Buffers is not garbage collected.

1. Start Server program with Akka.Remote.
2. Client program connects to Server through Akka.Remote.
3. Client program sends the message to Server.
4. Server program reply to the Client program after processing the message.
5. Client program prints the result and exit.
6. Multiple clients repeat 2~5.
7. **Memory usage has increased.**

- dotMemory profiling: https://github.com/iris4865/IssueStorages/raw/master/AkkaRemote_1.3.6/ActorSelection/ActorSelection-Akka1.3.6-DotNetty0.4.6.dmw

image

- server code
```
namespace AkkaSelectionNettyEnvServer
{
    class Program
    {
        {
            ActorSystem system = ActorSystem.Create("Server");

            system.ActorOf(Props.Create(() => new HelloActor()), "HelloActor");

            system.WhenTerminated.Wait();
        }
    }

    public class HelloActor : ReceiveActor
    {
        public HelloActor()
        {
            Receive<string>(value =>
            {
                Console.WriteLine(value);
            });
        }
    }
}
```

- client code
```
namespace AkkaSelectionNettyEnvClient
{
    class Program
    {
        {
            try
            {

                ActorSystem system = ActorSystem.Create("Client");

                var actor = system.ActorSelection("akka.tcp://Server@192.168.100.99:8083/user/HelloActor")
                    .ResolveOne(TimeSpan.FromSeconds(10))
                    .Result;

                actor.Tell("Hello");

                system.Terminate();
            }
            catch(Exception)
            {

            }
        }
    }
}
```

- client executer code
```
    class Program
    {
        {
            if (args.Length == 0)
                ExecuteLoop("AkkaSelectionNettyEnvClient.exe");
            else
                ExecuteLoop(args[0]);
        }

        public static void ExecuteLoop(string filename)
        {
            ProcessStartInfo info = new ProcessStartInfo()
            {
                FileName = filename,
                CreateNoWindow = true,
                WindowStyle = ProcessWindowStyle.Hidden,
            };
            while (true)
            {
                List<Process> pList = new List<Process>
                {
                    Process.Start(info),
                };

                foreach (var p in pList)
                {
                    p.WaitForExit();
                    p.Dispose();
                }

                pList = null;
            }
        }
    }
```

**p.s** 
I think Akka version 1.1.3 (with Akka.Remote 1.1.3 and Helios 2.1.2) is no problem.
so I am using version 1.1.3 temporarily.

Aaronontheweb:
close #3430 

So this was a bug I introduced in https://github.com/akkadotnet/akka.net/pull/3395

When manually configuring the allocator for use inside the `DotNettyTransport`, I was `new`-ing up an allocator each time. Instead going forward we'll use the singleton default built into DotNetty.

I'm not 100% sure this eliminates any possibility of buffers leaking on all platforms, since it seems like there's some major runtime differences that could affect this on .NET Core vs. .NET and on Linux / Windows, but this is definitely one possible source of the problem.

### DIFF
--- a/src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
+++ b/src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
@@ -125,8 +125,8 @@ namespace Akka.Remote.Transport.DotNetty
         protected volatile Address LocalAddress;
         protected internal volatile IChannel ServerChannel;
 
-        private readonly IEventLoopGroup serverEventLoopGroup;
-        private readonly IEventLoopGroup clientEventLoopGroup;
+        private readonly IEventLoopGroup _serverEventLoopGroup;
+        private readonly IEventLoopGroup _clientEventLoopGroup;
 
         protected DotNettyTransport(ActorSystem system, Config config)
         {
@@ -141,8 +141,8 @@ namespace Akka.Remote.Transport.DotNetty
 
             Settings = DotNettyTransportSettings.Create(config);
             Log = Logging.GetLogger(System, GetType());
-            serverEventLoopGroup = new MultithreadEventLoopGroup(Settings.ServerSocketWorkerPoolSize);
-            clientEventLoopGroup = new MultithreadEventLoopGroup(Settings.ClientSocketWorkerPoolSize);
+            _serverEventLoopGroup = new MultithreadEventLoopGroup(Settings.ServerSocketWorkerPoolSize);
+            _clientEventLoopGroup = new MultithreadEventLoopGroup(Settings.ClientSocketWorkerPoolSize);
             ConnectionGroup = new ConcurrentSet<IChannel>();
             AssociationListenerPromise = new TaskCompletionSource<IAssociationEventListener>();
 
@@ -161,8 +161,7 @@ namespace Akka.Remote.Transport.DotNetty
             if (InternalTransport != TransportMode.Tcp)
                 throw new NotImplementedException("Haven't implemented UDP transport at this time");
 
-            var dns = listenAddress as DnsEndPoint;
-            if (dns != null)
+            if (listenAddress is DnsEndPoint dns)
             {
                 listenAddress = await DnsToIPEndpoint(dns).ConfigureAwait(false);
             }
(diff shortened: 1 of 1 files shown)


src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
```diff
@@ -379,13 +378,13 @@ private ServerBootstrap ServerFactory()
...
             var server = new ServerBootstrap()
-                .Group(serverEventLoopGroup)
+                .Group(_serverEventLoopGroup)
                 .Option(ChannelOption.SoReuseaddr, Settings.TcpReuseAddr)
                 .Option(ChannelOption.SoKeepalive, Settings.TcpKeepAlive)
                 .Option(ChannelOption.TcpNodelay, Settings.TcpNoDelay)
                 .Option(ChannelOption.AutoRead, false)
                 .Option(ChannelOption.SoBacklog, Settings.Backlog)
-                .Option(ChannelOption.Allocator, Settings.EnableBufferPooling ? (IByteBufferAllocator)new PooledByteBufferAllocator() : new UnpooledByteBufferAllocator())
+                .Option(ChannelOption.Allocator, Settings.EnableBufferPooling ? (IByteBufferAllocator)PooledByteBufferAllocator.Default : UnpooledByteBufferAllocator.Default)
```
Aaronontheweb: This would definitely solve the allocator issue for clients... but the server is where we detected this memory leak and IIRC, this `ServerBootstrap` gets re-used to recreate client connections over and over again. Thus, I'm not sure this would actually impact the server. Bleh.

PR #3439, opened 2018-05-05
upgraded to DotNetty v0.4.8

### LINKED ISSUE #3437: Upgrade to DotNetty v0.4.8
It's now available on NuGet. Going to run the memory leak reproduction suite that users created for #3430 and make sure there aren't any outstanding issues there before upgrade.

Release notes for DotNetty v0.4.8:
---
#### 0.4.8 April 24, 2018
- Unsafe direct buffers
- HTTP 1.1 codec
- FlowControlHandler
- Channel pool
- Better Buffer-String integration
- Better shutdown handling for sockets
- Realigned Redis codec
- Fixes to LenghtFieldPrepender, LengthFieldBasedDecoder
- Fixes to libuv-based transport
- Fixes to buffer management on flush for .NET Core
- Fixes to ResourceLeakDetector

Aaronontheweb:
close #3437

### DIFF
--- a/src/core/Akka.Remote/Transport/DotNetty/TcpTransport.cs
+++ b/src/core/Akka.Remote/Transport/DotNetty/TcpTransport.cs
@@ -209,17 +209,18 @@ namespace Akka.Remote.Transport.DotNetty
                 var handler = (TcpClientHandler)associate.Pipeline.Last();
                 return await handler.StatusFuture.ConfigureAwait(false);
             }
+            catch (ConnectException c)
+            {
+                throw HandleConnectException(remoteAddress, c, null);
+            }
             catch (AggregateException e) when (e.InnerException is ConnectException)
             {
                 var cause = (ConnectException)e.InnerException;
-                var socketException = cause?.InnerException as SocketException;
-
-                if (socketException?.SocketErrorCode == SocketError.ConnectionRefused)
-                {
-                    throw new InvalidAssociationException(socketException.Message + " " + remoteAddress);
-                }
-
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
Looks like the behavior is different upon shutdown in .NET Core for DotNetty v0.4.8 - going to need to adjust the specs here. I know they've changed some stuff with trying to prevent resource leaks inside DotNetty itself in this patch and it looks like that's what is showing up in the specs here.

src/core/Akka.Remote/Transport/DotNetty/TcpTransport.cs
```diff
@@ -209,17 +209,18 @@ public TcpTransport(ActorSystem system, Config config) : base(system, config)
                 var handler = (TcpClientHandler)associate.Pipeline.Last();
                 return await handler.StatusFuture.ConfigureAwait(false);
             }
+            catch (ConnectException c)
```
Aaronontheweb: Fixed the failing test - simple change, really. Issue just was that DotNetty v0.4.8 doesn't necessarily throw an `AggregateException` with a `ConnectException` anymore; we'll just get the `ConnectException` itself in the `catch` block. So, modified those methods to handle that case explicitly.

Aaronontheweb:
This is ready to go - once it's merged in I'm going to run a long-running "soak test" for about 48 hours on another machine to see if we have any problems with resource leaks. Reports I've received from end-users running custom builds indicate that this isn't an issue, but I'd like to be sure.

If we run into trouble with v0.4.8 I'll institute a rollback before we push Akka.NET v1.3.8.

PR #3444, opened 2018-05-13
Override suppressed csharp warnings in FSharp projects

ivog:
These csharp warnings are currently imported through the common.props file but are not understood by the fsharp compiler. This is now fixed by overriding the NoWarn property in the FSharp projects.


ivog:
@Aaronontheweb This PR fixes the "FSC : warning FS0203: Invalid warning number 'CS1591'"

Aaronontheweb:
Thanks for your contribution @ivog

PR #3451, opened 2018-05-16
Minor fix to the Persistence Query sample code

ismaelhamed:
Currently the code returns NULL which is not only not allowed but will silently fail if the faulted Task is not dealt with --and will probably confuse people, not just me :)

Let's instead return the actual event to keep the flow going.


Danthar:
Thx for your contribution :)

PR #3458, opened 2018-05-22
Akka.Cluster: don't let SplitBrainResolver log 'partition detected' messages when there are no unreachable nodes

### LINKED ISSUE #3450: Akka.Cluster: SplitBrainResolver logs "network partition detected" after change in cluster membership, even when no unreachable nodes
Akka.NET Version: 1.3.7

Aaronontheweb:
close #3450 

Purely cosmetic fix. Keep seeing these messages during normal cluster startup and it caused a room full of Akka.NET developers, including me, to freak out every time we saw it. Figured changing it to only log this message only when there were unreachable nodes detected by the time the cluster stabilized would be a good idea.

@Horusiath do you have any opinions on this?

### DIFF
--- a/src/core/Akka.Cluster/SplitBrainResolver.cs
+++ b/src/core/Akka.Cluster/SplitBrainResolver.cs
@@ -317,7 +317,7 @@ namespace Akka.Cluster
 
         private void HandleStabilityReached()
         {
-            if (Log.IsInfoEnabled)
+            if (Log.IsInfoEnabled && _unreachable.Any())
             {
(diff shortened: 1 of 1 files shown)


src/core/Akka.Cluster/SplitBrainResolver.cs
```diff
@@ -317,7 +317,7 @@ protected override void OnReceive(object message)
 
         private void HandleStabilityReached()
         {
-            if (Log.IsInfoEnabled)
+            if (Log.IsInfoEnabled && _unreachable.Any())
```
Horusiath: Please use `IsEmpty` - `Any()` is really bad way of handling that.

PR #3464, opened 2018-05-23
Fix ActorSelection to remotely deployed actors

Aaronontheweb:


src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -316,6 +316,7 @@ namespace Akka.Actor
         public Akka.Actor.SelectionPathElement[] Elements { get; }
         public object Message { get; }
         public bool WildCardFanOut { get; }
+        public Akka.Actor.ActorSelectionMessage Copy(object message = null, Akka.Actor.SelectionPathElement[] elements = null, System.Nullable<bool> wildCardFanOut = null) { }
```
Aaronontheweb: Just ported the `Copy` method over from Scala for the `ActorSelectionMessage` type. No other public API changes to speak of.

src/core/Akka.Remote.Tests/RemotingSpec.cs
```diff
@@ -205,15 +205,15 @@ public void Resolve_does_not_deadlock_GuiApplication()
         [Fact]
         public void Remoting_must_not_send_remote_recreated_actor_with_same_name()
```
Aaronontheweb: All of the original assertions that @marcpiechura commented out due to #1544 being an issue have since been uncommented and now pass, which is what I'm using for verification on this issue.

src/core/Akka.Remote/Endpoint.cs
```diff
@@ -109,20 +109,17 @@ public DefaultMessageDispatcher(ActorSystem system, IRemoteActorRefProvider prov
                     var msgLog = $"RemoteMessage: {payload} to {recipient}<+{originalReceiver} from {sender}";
                     _log.Debug("received local message [{0}]", msgLog);
                 }
-                if (payload is ActorSelectionMessage)
+                if (payload is ActorSelectionMessage sel)
                 {
-                    var sel = (ActorSelectionMessage)payload;
-
-                    var actorPath = "/" + string.Join("/", sel.Elements.Select(x => x.ToString()));
```
Aaronontheweb: A minor performance improvement: we were allocating a new `actorPath` string every single time we received an `ActorSelectionMessage` in Akka.Remote even though its value is only used when `_settings.UntrustedMode` returns `true`, which it doesn't by default.

So instead, we no longer allocate this object unless that setting is enabled - we do have to allocate the string twice now, however, due to the way I moved that code into its own function. Not a huge deal given that untrusted mode isn't used all that much, but I wanted to call out this change on the PR.

src/core/Akka.Remote/RemoteActorRef.cs
```diff
@@ -112,7 +113,16 @@ public override IActorRefProvider Provider
         /// <exception cref="System.NotImplementedException">TBD</exception>
         public override IActorRef GetChild(IEnumerable<string> name)
         {
-            throw new NotImplementedException();
+            var items = name.ToList();
+            switch (items.FirstOrDefault())
```
Aaronontheweb: Needed to resolve `ActorSelection`s on remotely deployed actors - matches the corresponding JVM code here: https://github.com/akka/akka/blob/84966a2ca54f96360a3996e1c670684e2948f870/akka-remote/src/main/scala/akka/remote/RemoteActorRefProvider.scala#L513

src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -124,7 +125,53 @@ protected override void TellInternal(object message, IActorRef sender)
                 Log.Debug("Received command [{0}] to RemoteSystemDaemon on [{1}]", message, Path.Address);
                 if (message is DaemonMsgCreate) HandleDaemonMsgCreate((DaemonMsgCreate)message);
             }
+            else if (message is ActorSelectionMessage sel)
```
Aaronontheweb: Most of the heavy lifting for #1544 occurs here inside the `RemoteDaemon` - we never programmed it to handle `ActorSelectionMessage` instances before. 

Opted not to use `ValueTuple` here since we're adding that dependency in 1.4. Didn't want to introduce it now for such a small change.

Issue #3205, opened 2017-12-02
DeleteMessageFailure fails 


jalchr:
Sometimes ... I get the following error. Why it is "inserting" on a delete operation ?
```
2017-12-02 13:25:24,757 [36] ERROR Archive.Video.Importer.Actors.FileHandler - [DeleteMessagesFailure]: shutting down !
System.Data.SqlClient.SqlException (0x80131904): Violation of PRIMARY KEY constraint 'PK_Metadata'. Cannot insert duplicate key in object 'dbo.Metadata'. The duplicate key value is (\\nas20\d$\archive\incoming\aptn\cctv025151_russia-scochinese premier.xml, 1).
The statement has been terminated.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString, Boolean isInternal, Boolean forDescribeParameterEncryption)
   at System.Data.SqlClient.SqlCommand.CompleteAsyncExecuteReader(Boolean isInternal, Boolean forDescribeParameterEncryption)
   at System.Data.SqlClient.SqlCommand.InternalEndExecuteNonQuery(IAsyncResult asyncResult, String endMethod, Boolean isInternal)
   at System.Data.SqlClient.SqlCommand.EndExecuteNonQueryInternal(IAsyncResult asyncResult)
   at System.Data.SqlClient.SqlCommand.EndExecuteNonQueryAsync(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
[...]

PR #3468, opened 2018-05-24
When deleting SQL journal entries also delete corresponding metadata entries

hirzraimund:
When deleting persistence journal entries through the common sql implementation also delete the entries in the metadata table. This solves #3205 (violation of primary key constraint when the value already exists) and ensures that there exists at most one entry per PersistenceId value.

Added @ in front of the parameters for the DeleteBatchSql command to be in line with the other usage of AddParameter.


Aaronontheweb:
Good to know.

hirzraimund:
I am pretty sure that it still works as intended since in the methods [`AbstractQueryExecutor.DeleteBatchAsync`](https://github.com/hirzraimund/akka.net/blob/22c98c0a466e40747805881f9aca53c51ac8ef1d/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs#L599) and  in [`BatchingSqlJournal.HandleDeleteMessagesTo`](https://github.com/hirzraimund/akka.net/blob/22c98c0a466e40747805881f9aca53c51ac8ef1d/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs#L900) , which are the only places where the delete query is used, the highest sequence number is insert into metadata table if it is inside the deleted range. The insert is done inside the same transaction as the delete and therefore it is ensured that both or none of these commands are executed.

hirzraimund:
@Horusiath

Aaronontheweb:
@hirzraimund @Horusiath I'll reopen the PR, but ya'll need to get to the bottom of what's really going on under the covers here before I close it again or merge it :p 

Unit test it, IMHO.

ismaelhamed:
In the current state, maybe it should be advisable for plugins to resolve this with the proper upsert (i.e, a MERGE in Sql Server and Oracle).

Another option would be to get rid of the `Metadata` table all together, and make a soft delete on the last event instead (the jdbc plugin way). By keeping just the last event for reference, this keeps track of the highest sequence number in case the user cleans up the journal while remaining as an internal implementation detail.

Horusiath:
@ismaelhamed We could do a proper upsert in metadata table as a separate PR (only there, as event journal is append only). Regarding removal of Metadata table. It could be done' but I'm not sure if we wouldn't break backward compatibility this way.

Aaronontheweb:
LGTM

PR #3474, opened 2018-05-27
Syntactic sugar to make the "create if not exist" pattern more idiomatic

ismaelhamed:
So we can now do something like this:
```
var child = Context.Child(actorName).GetOrElse(() => Context.ActorOf(...));
```
In the JVM this method is part of the OptionVal object, but I thought it made sense to it put along with the IsNobody extension method.

### DIFF
--- a/src/core/Akka/Actor/ActorRef.Extensions.cs
+++ b/src/core/Akka/Actor/ActorRef.Extensions.cs
@@ -5,6 +5,8 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System;
+
 namespace Akka.Actor
 {
     /// <summary>
@@ -22,6 +24,17 @@ namespace Akka.Actor
         public static bool IsNobody(this IActorRef actorRef)
         {
             return actorRef == null || actorRef is Nobody || actorRef is DeadLetterActorRef;
+        }
+
+        /// <summary>
+        /// Returns the <paramref name="actorRef"/>'s value if it's not <see langword="null"/>, <see cref="Nobody"/>, 
+        /// or <see cref="DeadLetterActorRef"/>. Otherwise return the result of evaluating `elseValue`.
+        /// </summary>
+        /// <param name="actorRef">The actor that is being tested.</param>
+        /// <param name="elseValue">TBD</param>
+        public static IActorRef GetOrElse(this IActorRef actorRef, Func<IActorRef> elseValue)
+        {
+            return actorRef.IsNobody() ? elseValue() : actorRef;
         }
     }
 }
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ClusterShardingGuardian.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ClusterShardingGuardian.cs
@@ -172,10 +172,9 @@ namespace Akka.Cluster.Sharding
                     var encName = Uri.EscapeDataString(start.TypeName);
                     var coordinatorSingletonManagerName = CoordinatorSingletonManagerName(encName);
                     var coordinatorPath = CoordinatorPath(encName);
-                    var shardRegion = Context.Child(encName);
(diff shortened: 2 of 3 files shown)


src/core/Akka/Actor/ActorRef.Extensions.cs
```diff
@@ -23,6 +25,17 @@ public static bool IsNobody(this IActorRef actorRef)
...
+
+        /// <summary>
+        /// Returns the <paramref name="actorRef"/>'s value if it's not <see langword="null"/>, <see cref="Nobody"/>, 
+        /// or <see cref="DeadLetterActorRef"/>. Otherwise return the result of evaluating `elseValue`.
+        /// </summary>
+        /// <param name="actorRef">The actor that is being tested.</param>
+        /// <param name="elseValue">TBD</param>
+        public static IActorRef GetOrElse(this IActorRef actorRef, Func<IActorRef> elseValue)
+        {
+            return actorRef.IsNobody() ? elseValue() : actorRef;
```
ismaelhamed: Akka actually checks if it's not `isEmpty`, but I'm not sure if we should check `actorRef.IsNobody()`, `actorRef == null || actorRef is Nobody`, or just `actorRef == null`.

Horusiath:
Maybe instead of allocating lambdas, this approach would be better?:

```csharp
if (!Context.TryGetChild(actorName, out var child))
    child = Context.ActorOf(props, actorName);
```

ismaelhamed:
Close enough, actually :)

I didn't think this would make a dent since most probably that Func<> will end up being a delegate or static function (and cached if I'm not mistaken) unless the user do something fancy with it and capture some variable.

Aaronontheweb:
given that we're going to be parameterizing things like the child's name for each child actor created, I think it'd probably be fine if we just went with a lambda that returns an `IActorRef` here. It shouldn't be scoped for long and won't be referenced by anything IIRC, since we're not passing it around. This approach looks fine.

Horusiath:
@Aaronontheweb I think that `GetOrElse` is very vague name for method that is actually super specific - if `GetChild` returned null or option, this would be fine, but not here. On the other side `TryGetChild` is very intuitive for any person, that has ever worked with .NET collections and can do the same job.

src/core/Akka/Actor/ActorRef.Extensions.cs
Aaronontheweb: `IsNobody` performs all of those checks internally - does a null check, `is Nobody`, and also checks if the ref is a `DeadLetterActorRef`

ismaelhamed: Sure, what I meant was that in the JVM they just check for isEmpty (which I think is just NULL).

ismaelhamed:
@Aaronontheweb @Horusiath Refactored a couple of places in the cluster sharding code to show off usage.

Aaronontheweb:
I like it, but you'll need to run the API approval code to make changes to public APIs: http://getakka.net/community/public-api-changes.html

ismaelhamed:
@Aaronontheweb, done.

Aaronontheweb:
@Horusiath you had some input on this still?

ismaelhamed:
@Aaronontheweb should we close this PR then? Anyone could add an extension like this one to his project anyway.

Danthar:
@Aaronontheweb @Horusiath what are we going to do with this ? Yay or Nay

Danthar:
ClusterSingletonManagerStartupSpec seems to be failing consistently now. Might have something to do with my merge....

Aaronontheweb:
@Danthar I think that one has historically had some issues

ismaelhamed:
Changes to the `Receive<StartProxy>` where lost in the merge. I'm going to rebase and squash to keep it clean.

Aaronontheweb:
Nice catch

PR #3475, opened 2018-05-28
Introducing BenchmarkDotNet benchmarks

Horusiath:
This PR introduces new benchmarking project using [BenchmarkDotNet](https://benchmarkdotnet.org/). So far following cases have been included:

- `Akka.IO.ByteString` benchmarks including byte string creation, concatenation and slicing.
- Consistent hash benchmarks where I'm comparing our Murmur implementation with Jenkins hash used in Orleans.
- `Akka.Util.FastLazy` value materialization comparison with native `System.Lazy`.
- `Type.TypeQualifiedName()` compared to `Type.FullName`.
- `ActorPath` and `Address` parsing, concat, comparison and equivalence benchmarks.

There are also two monitored benchmarks, that shouldn't count as *microbenchmarks*, but BenchmarkDotNet can do them too:

- Spawn actor measurements.
- Local-only 1-to-1 actor ping pong (I'm measuring only 2 communicating actors here). 

Those 2 checks can potentially make PingPong.csproj and SpawnBenchmark.csproj obsolete.

I'm hoping to introduce more benchmarks in future PRs.


Aaronontheweb:
Where does the output from these get stored?

Horusiath:
Default location is `/bin/Release/netcoreapp2.0/BenchmarkDotNet.Artifacts/results`

Aaronontheweb:
@Horusiath got a warning back that the API specs had failed for Streams and Core. No idea why. Rebased on the latest `dev` - did you make any changes that would have affected those? I couldn't see any during my review, but I was mostly looking at the project metadata itself.

Horusiath:
Beside adding `[assembly:InternalsVisibleTo("Akka.Benchmarks")]` nothing has been changed in any other project.

Aaronontheweb:
We'll see what it says on this run.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1,4 +1,5 @@
-﻿[assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("Akka.Cluster")]
```
Aaronontheweb: Ah I see.. It was the internals stuff

PR #3476, opened 2018-05-30
Akka.Streams: add PreMaterialize support for Sources

Aaronontheweb:
Port of https://github.com/akka/akka/pull/24497

### DIFF
--- a/docs/articles/streams/basics.md
+++ b/docs/articles/streams/basics.md
@@ -389,7 +389,7 @@ RunnableGraph<Tuple<TaskCompletionSource<int>, ICancelable, Task<int>>> r12 =
 
 ### Source pre-materialization
 There are situations in which you require a `Source` materialized value **before** the `Source` gets hooked up to the rest of the graph.
-+This is particularly useful in the case of "materialized value powered" `Source`s, like `Source.Queue`, `Source.ActorRef` or `Source.Maybe`.
+This is particularly useful in the case of "materialized value powered" `Source`s, like `Source.Queue`, `Source.ActorRef` or `Source.Maybe`.
 
 By using the `PreMaterialize` operator on a `Source`, you can obtain its materialized value and another `Source`. The latter can be used to consume messages from the original `Source`. Note that this can be materialized multiple times.
 
--- a/docs/articles/streams/basics.md
+++ b/docs/articles/streams/basics.md
@@ -393,12 +393,7 @@ There are situations in which you require a `Source` materialized value **before
 
 By using the `PreMaterialize` operator on a `Source`, you can obtain its materialized value and another `Source`. The latter can be used to consume messages from the original `Source`. Note that this can be materialized multiple times.
 
-```csharp
-var source = Source.Maybe<int>;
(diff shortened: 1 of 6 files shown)


docs/articles/streams/basics.md
```diff
@@ -386,7 +386,19 @@ RunnableGraph<Tuple<TaskCompletionSource<int>, ICancelable, Task<int>>> r12 =
 > [!NOTE]
 > In Graphs it is possible to access the materialized value from inside the stream processing graph.
 > For details see [Accessing the materialized value inside the Graph](xref:streams-working-with-graphs#accessing-the-materialized-value-inside-the-graph).
-   
+
+### Source pre-materialization
+There are situations in which you require a `Source` materialized value **before** the `Source` gets hooked up to the rest of the graph.
++This is particularly useful in the case of "materialized value powered" `Source`s, like `Source.Queue`, `Source.ActorRef` or `Source.Maybe`.
```
marcpiechura: A + too much

docs/articles/streams/basics.md
```diff
@@ -386,7 +386,19 @@ RunnableGraph<Tuple<TaskCompletionSource<int>, ICancelable, Task<int>>> r12 =
...
+
+By using the `PreMaterialize` operator on a `Source`, you can obtain its materialized value and another `Source`. The latter can be used to consume messages from the original `Source`. Note that this can be materialized multiple times.
+
+```csharp
+var source = Source.Maybe<int>;
+Tuple<TaskCompletionSource<int>, Source<int>> materialized = matValPoweredSource.PreMaterialize(Sys.Materializer());
+var tcs = materialized.Item1;
+var nextSource = materialized.Item2;
+```
```
marcpiechura: It’s probably fine to include this code directly in the .md since all the other code examples are here as well, but in general haven’t we agreed on using the document projects for code examples ?!

docs/articles/streams/basics.md
Aaronontheweb: Whoops, good catch.

docs/articles/streams/basics.md
Aaronontheweb: @marcpiechura yeah, we try to do that for the most part going forward but we don't have anything organized for that in Akka.Streams itself at the moment. I'd personally prefer to do something that is compile-time checked and there referred to via DocFx's source code embedding, akin to https://github.com/akka/akka/pull/24497/files#diff-96bb38a1bfc1933567a890af8f3d0263

But starting the infrastructure for that is a bigger project than adding just this one feature. So question for you is: start building the infrastructure now or throw yet one more thing on the future technical debt pile for later? I'm opted for the latter (obviously) but, if there's buy-in from other folks to go down the route of doing something similar to what the JVM does then I'd be all for it.

Aaronontheweb: Welp, I was wrong about this - we do have support for this built into the project already. Just uploaded some changes which incorporate it.

Aaronontheweb:
@marcpiechura I think I addressed all of your comments in the newest commit.

marcpiechura:
@Aaronontheweb there is still the ++ in the readme ;-)

Aaronontheweb:
@marcpiechura damn, thought I got it - just fixed... again :p

Issue #3471, opened 2018-05-25
SqlCommands with connections are not garbagecollected, just disposed.


jarlrasm:
`Akka.Persistence.Sql.Common.Journal.SqlJournal` is using a single `CancellationToken` for all db-operations, that is only disposed on shutdown, if I understand the code correcly. This indirectly keeps references to all `SqlCommands`, which again have references to connections and so on, preventing GC. I believe this can be fixed with creating a new `CancellationTokenSource` with `CreateLinkedTokenSource` for each command, however I am not quite sure why the implementation is as it is, so I would like to have some feedback from the developers before I fix it and create a pullrequest.

Horusiath:
@jarlrasm If I remember correctly, the idea was to interrupt all in-progress commands, when the journal is going to die.

Aaronontheweb:
closed via https://github.com/akkadotnet/akka.net/pull/3494

PR #3494, opened 2018-06-07
Create a new linked CancellationToken for each database operation(#3471)

jarlrasm:
This fixes a memory leak with SqlCommands being referenced indefinitely in the SqlJournal's long living CancellationToken.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
@@ -168,7 +168,8 @@ namespace Akka.Persistence.Sql.Common.Journal
                     }
 
                     var batch = new WriteJournalBatch(eventToTags);
-                    await QueryExecutor.InsertBatchAsync(connection, _pendingRequestsCancellation.Token, batch);
+                    using(var cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(_pendingRequestsCancellation.Token))
+                        await QueryExecutor.InsertBatchAsync(connection, cancellationToken.Token, batch);
                 }
             }).ToArray();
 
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
@jarlrasm did you see a memory leak crop up during profiling or something?

jarlrasm:
@Aaronontheweb yes. I noticed hundreds of thousands of uncollected sql-related objects when I was profiling our app with some load. Tracked it down to the cancellationToken.

maxcherednik:
@jarlrasm Could you please explain the reason behind? Where is the issue itself? The reason I am asking is that I don't remember CancellationToken behaves this way.

jarlrasm:
@maxcherednik DbCommand registers a callback on the CancellationToken which is never removed, thus preventing garbagecollection.

maxcherednik:
@jarlrasm Yeah, this change is going to fix the issue. But the root cause here is that we misuse cancelationTokenSource

jarlrasm:
@maxcherednik I agree that it is strange. But since I don't know the background for why the design is as it is, I didn't want to change it in a fundamental way.

maxcherednik:
@jarlrasm  btw there are 2 issues here:
1. there is a bug in the DbCommand - they dont dispose the registration: https://github.com/Microsoft/referencesource/blob/3b1eaf5203992df69de44c783a3eda37d3d4cd10/System.Data/System/Data/Common/DBCommand.cs#L213
2. CancellationTokenSource is misused here. For example, post stop method cancels the CTS and then what? Who will make sure that everything was complete? 
Cancellation should always be called from the places where the initial call was made and should be done within the context of the method.
```
Task DoSomething(int param, CancellationToken ct);
```

Maybe we can the core team if this class is public.

The fix of yours will probably remove the leak, but it will also introduce some allocations, potentially heavy. There could be more subtle issues, which are not visible right now.

jarlrasm:
1. Looks like the upstream bug is fixed in dotnet-core:
https://github.com/dotnet/corefx/blob/master/src/System.Data.Common/src/System/Data/Common/DbCommand.cs#L144
2. I agree here, but I think someone who understand the intention behind the design should look into it, rather than me just guessing.

I don't think the allocations introduced by my fix are at all heavy, but I agree with the sentiment that the root problems should be fixed rather than patching on something that is flawed.

Aaronontheweb:
@maxcherednik Max, do you feel like you have a strong enough handle on the Akka.Persistence stuff to take a look at this? I'm neck deep in fixing some issues with a different OSS project today (NBench) but I might be able to take a look at it this weekend. Otherwise, @Horusiath might be able to take a look and see what's up.

maxcherednik:
@Aaronontheweb will try to have a look in a couple of days.

Aaronontheweb:
@maxcherednik thanks bud, I owe you one.

maxcherednik:
For the reference: https://github.com/dotnet/corefx/issues/16439

Aaronontheweb:
Looks good to me. Read up on it some more and couldn't see any harm in it.

PR #3497, opened 2018-06-08
[WIP] CoordinatedShutdown and ClusterDaemon fixes

### LINKED ISSUE #3413: Bug: ClusterDaemon doesn't stop upon CoordinatedShutdown
Verison: 1.3.6

Expected behavior: when `CoordinatedShutdown` is executed we expect the `ClusterDaemon` actor to terminate during the `cluster-shutdown` phase.

Actual behavior: `ClusterDaemon.PostStop` is never called.

Logs to verify this: https://gist.github.com/Horusiath/22c040af3c5f431d509163f7eb332405

Aaronontheweb:
Designed to reproduce and fix #3413


src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -534,7 +534,7 @@ internal static void InitPhaseActorSystemTerminate(ActorSystem system, Config co
                         // We must spawn a separate Task to not block current thread,
                         // since that would have blocked the shutdown of the ActorSystem.
                         var timeout = coord.Timeout(PhaseActorSystemTerminate);
-                        return Task.Run(() =>
+                        Task.Run(() =>
```
Aaronontheweb: Not supposed to return a `Task` here. Supposed to just let the process run asynchronously and invoke once the `ActorSystem` has finished terminating.

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -351,12 +351,12 @@ Task<Done> Loop(List<string> remainingPhases)
                                     return TaskEx.FromException<Done>(ex);
                                 }
                             }))
-                            .ContinueWith(tr =>
+                            .ContinueWith(async tr =>
                             {
                                 // forces downstream error propagation if recover is disabled
-                                var force = tr.Result;
```
Aaronontheweb: Let the task run asynchronously, but still propagate failures downstream.

Aaronontheweb:
@Horusiath thanks for the suggestions; was able to reproduce the issue with #3413 - I think I can TDD it from here.

Aaronontheweb:
Looks like it's the reproduction spec for this that's failing.

PR #3499, opened 2018-06-09
Non-blocking task completion source helpers

### LINKED ISSUE #3498: Akka.Streams.IO TcpStage deadlocks on server binding
It turned out, that in some cases TCP stage from Akka.Streams can hang infinitely - this is related to the fact that `TaskCompletionSource<T>.TrySetResult` and family is [blocking operation](https://stackoverflow.com/questions/39296587/in-which-case-does-taskcompletionsource-setresult-run-the-continuation-synchro?rq=1) which causes [this line](https://github.com/akkadotnet/akka.net/blob/11985622b0e4748d1be5742b6ac4109d4c18a228/src/core/Akka.Streams/Implementation/IO/TcpStages.cs#L127) to deadlock and never end, effectively blocking TCP server from accepting new connections.

Horusiath:
Fixes #3498 but we need to find a better way to address those issues. Potentially moving to netstandard 2+ compatibility (where async task completion source flags are available by default).

### DIFF
--- a/src/core/Akka.Streams/Implementation/IO/TcpStages.cs
+++ b/src/core/Akka.Streams/Implementation/IO/TcpStages.cs
@@ -124,17 +124,19 @@ namespace Akka.Streams.Implementation.IO
                         _listener.Tell(new Tcp.ResumeAccepting(1), StageActorRef);
 
                     var thisStage = StageActorRef;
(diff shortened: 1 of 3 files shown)


0x53A:
You could also do a runtime check: https://github.com/akkadotnet/akka.net/pull/3120/files#diff-087cfe232c174bfb770e0de44390a722R132

PR #3503, opened 2018-06-12
[api-docs] fix a few xref/parameter issues

sean-gilliam:
This PR fixes some xrefs issues wrt to referencing base class
fields/methods. It also adds some missing method parameter tags.


Aaronontheweb:
Looks good to me

PR #3504, opened 2018-06-12
upgraded to XUnit 2.3.1 and VsTest SDK 15.7.2

Aaronontheweb:


Aaronontheweb:
LOL... apparently this broke 100% of the ApiApproval specs. Must be because it affected something in the API.Approvals test assembly itself. I'll double check and see what's up.

Danthar:
This update is also needed to support netstandard 2.0 apparently.

Aaronontheweb:
Got that fixed - had to adjust the output copying of the stored API approval file

Aaronontheweb:
@Danthar I'm probably going to need to merge this into the v1.4 branch as well then :p

PR #3505, opened 2018-06-13
fix concurrency bug in CircuitBreaker

zbynek001:
Fixes concurrency issue in CircuitBreaker
Migrated from: https://github.com/akka/akka/issues/17415

Didn't include the CircuitBreakerStressSpec, because I was not able to reproduce the issue with it


Aaronontheweb:
Thanks for your contribution @zbynek001

PR #3509, opened 2018-06-14
Sharding - RestartShard escapes into userspace

zbynek001:
there is a possibility that the `ShardRegion.RestartShard` message escapes into userspace. This can cause issues in custom `ExtractEntityId` implementations.

This can happen, when the `RestartShard` message gets into ShardBuffers here: https://github.com/akkadotnet/akka.net/blob/d732274b0fa9cb6863fabbdf9d1d2c447d61d6f3/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs#L560
and later is send to `Shard`


Aaronontheweb:
Just double checked this against the JVM code base and it looks like they have the same bug. Nice catch. LGTM.

PR #3452, opened 2018-05-16
WIP Netstandard2.0upgrade

Danthar:
Work in Progress to update everything to netstandard 2.0 and netcore2.0.
First stage was updating TargetFrameworks in all the project files. Which is partly done.

- [X] Akka Core projects update
- [x] Contrib projects update
   - [X] Akka.Cluster.x
   - [X] Akka.DistributedData
   - [x] The rest
- [ ] Benchmark projects should also target and run netcore suite
- [ ] Check and verify project dependencies
   - [x] Update XUnit dependencies (needs XUnit core) 
   - [x] Update to newtonsoft.json 110.2
   - [x] Update to System.Collections.Immutable 1.5.0
- [x] Possibly update CI infrastructure (scripts and such)
- [X] Delete netcorepolyfills, aka, get rid of SerializableAttribute shims
- [ ] ApiApproval project needs updating
- [ ] Example projects needs to directly reference akka projects instead of assemblies, and be .netcore compatible instead of targetting net452. Possibly leave some projects to be net452 on purpose as a canary
   - [ ] Fsharp examples needs to be fixed
- [x] Update akka.Fsharp references to Fsharp.core 4.5.2

This work will be the enabler for some other improvements we have in mind. Feedback is welcome

Note; Lots of the example projects are still in old style csproj. And could use updating. I might do that in here if it gets in the way. But it might be better to make it a seperate issue. Could be an good for first contributors thingy.

PR #3510, opened 2018-06-14
V1.4 Test SDK Upgrades

Aaronontheweb:
Normally wouldn't do another re-sync so quickly, but need to get this in here to support https://github.com/akkadotnet/akka.net/pull/3452 - part of the V1.4 roadmap.


Aaronontheweb:
Going to go with "rebase and merge" for this PR provided that we get all the green check marks.

Issue #3417, opened 2018-04-24
ClusterClientReception Ask missing response to Client after a few thousand


leo12chandu:
When creating a new issue, please make sure the following information is part of your issue description. (if applicable). Thank You!

- Which Akka.Net version you are using
1.3.4
- On which platform you are using Akka.Net
Windows 7 & Visual Studio 2017
- A list of steps to reproduce the issue. Or an gist or github repo which can be easily used to reproduce your case.

Is there other ways to send/receive/ask messages to a cluster apart from ClusterClientReceptionist from outside the cluster?

We are currently registering actors with ClusterClientReceptionist and use ClusterClient to "ask" messages to cluster from outside. However, when we test with 10,000 (small) messages after a few thousand even though the server node processes those messages, the client does not receive a reply for them through Ask. Is there a way to flush the messages back to cluster after Client.Tell?

ClusterClientReceptionist with 2 seed and 2 worker (non-seed) nodes seem to evenly distribute the load between both the worker nodes. I am curious if ClusterClientReceptionist is the recommended way to distribute load across nodes or if there is a better way to handle it (including the ability to send/ask messages from outside of cluster).

leo12chandu:
I have additional details on this. This is reproducible only if 
1) the seed node is separated out from the actual worker node.
2) have custom HyperionSerializer.

When we make the actual worker node a seed node onto itself, we are unable to reproduce this even though we continue to use custom HyperionSerializer. I.E, we are not missing any message responses back to cluster client when all our nodes are seed nodes by themselves.

If it helps to know, when seed nodes are separated out, due to custom serializer the message is actually bouncing around between seed nodes and worker nodes twice. Once when the message arrives, it first goes to seed node, deserializes and then is sent to the actual worker node which is again deserialized. And then when the result is sent back through ClusterClientReceptionist, the message is sent first to the seed node, deserializes and then seed node inturn sends it to the clusterclient. My thought is, this is probably because the initial contacts we defined are for seed nodes. Hence I am guessing the message has to always pass through the seed node. But i donno if this leads to dropped responses (back to client).

leo12chandu:
More information:-
When seed node and worker node are the same, the cluster clients that are in the same geography as the seed/worker nodes are working fine. They are able to receive all the messages back even when I try to send 10,000 messages at once. However, when we try have cluster clients reside in a different geography than the seed/worker nodes that do the work, after 1000 requests, we are missing some message responses.

Horusiath:
[...]

PR #3512, opened 2018-06-14
fixed NRE issue with ClusterClientSettings.Copy

Aaronontheweb:
mentioned in #3417 

Added spec to reproduce and verify fix as well.


src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientSettings.cs
```diff
@@ -211,7 +211,7 @@ public ClusterClientSettings WithReconnectTimeout(TimeSpan? reconnectTimeout)
             TimeSpan? reconnectTimeout = null)
         {
             return new ClusterClientSettings(
-                initialContacts,
+                initialContacts ?? InitialContacts,
```
Aaronontheweb: Forgot to copy over any previously set `InitialContacts`

PR #3524, opened 2018-06-24
Sharding update

zbynek001:
Contains following changes:

- Provide access to known shard types
https://github.com/akka/akka/issues/23912

- Separate sharding regions and proxies
https://github.com/akka/akka/issues/23472
Fix lookup of coordinator for sharding proxies
https://github.com/akka/akka/pull/23995

- Fix race in ClusterShardingFailureSpec
AFAICT there was nothing ensuring the order of messages when sent to the
shard and the region so first checkthat the passivation has happened
before sending another add in the test
https://github.com/akka/akka/issues/24013

- Better warning message on cluster sharding registration
https://github.com/akka/akka/pull/24906

- entityId => Behavior in ClusterSharding API
mixture of
https://github.com/akka/akka/issues/24053
https://github.com/akka/akka/issues/21809
https://github.com/akka/akka/issues/24470

- ClusterSharding: automatically choose start or startProxy by a node role
https://github.com/akka/akka/issues/23934

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/GetShardTypeNamesSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/GetShardTypeNamesSpec.cs
@@ -22,7 +22,7 @@ namespace Akka.Cluster.Sharding.Tests
 
         public static Config GetConfig()
         {
-            return ConfigurationFactory.ParseString("akka.actor.provider = \"Akka.Cluster.ClusterActorRefProvider, Akka.Cluster\"")
+            return ConfigurationFactory.ParseString("akka.actor.provider = cluster")
 
                 .WithFallback(Sharding.ClusterSharding.DefaultConfig())
                 .WithFallback(DistributedData.DistributedData.DefaultConfig())
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/GetShardTypeNamesSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/GetShardTypeNamesSpec.cs
@@ -1,0 +1,69 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ClusterShardingConfigSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2018 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2018 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


Aaronontheweb:
@zbynek001 sure you don't want that tshirt?

Aaronontheweb:
Seriously though, I was about to go through Cluster.Sharding this weekend to look at some of the open issues we have on that front :p - guess I'll start by reviewing this monster 👍

src/contrib/cluster/Akka.Cluster.Sharding.Tests/GetShardTypeNamesSpec.cs
```diff
@@ -0,0 +1,69 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ClusterShardingConfigSpec.cs" company="Akka.NET Project">
```
Aaronontheweb: Copyright headers have the wrong file names

src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
```diff
@@ -567,6 +569,297 @@ public static Config DefaultConfig()
...
+        /// The message that will be sent to entities when they are to be stopped for a rebalance or
+        /// graceful shutdown of a <see cref="Sharding.ShardRegion"/>, e.g. <see cref="PoisonPill"/>.
+        /// </param>
+        /// <exception cref="IllegalStateException">
+        /// This exception is thrown when the cluster member doesn't have the role specified in <paramref name="settings"/>.
+        /// </exception>
+        /// <returns>The actor ref of the <see cref="Sharding.ShardRegion"/> that is to be responsible for the shard.</returns>
+        public IActorRef Start(
+            string typeName,
+            Func<string, Props> entityPropsFactory,
```
Aaronontheweb: So I think this API change is fine, but we should have some documentation about it on the cluster-sharding Akka.NET documentation similar to what they did on the JVM in that PR: https://github.com/akka/akka/pull/24058/files#diff-d0e085b72595aa4dfa9e30261feecd93

Aaronontheweb:
@zbynek001 I'm in the process of ordering some more tshirts, might take a bit to ship them but it's in the works. Send me an email with your shipping info :p

Horusiath:
@zbynek001 ActorSystem contains logger instance as well.

zbynek001:
also included this one:
ClusterSharding: automatically choose start or startProxy by a node role
akka/akka#23934

And that will be all in this PR

Aaronontheweb:
Great work as usual @zbynek001

PR #3527, opened 2018-06-25
Fix MemorySnapshotStore invalid operation

### LINKED ISSUE #3526: Akka.Persistence.MemorySnapshotStore: System.InvalidOperation exception
Version: Akka.NET v1.3.8

Error should be relatively easy to find:

```
'18-06-21 08:36:52: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException_InvalidOperation_EnumFailedVersion()
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.WhereListIterator`1.MoveNext()
   at System.Linq.OrderedEnumerable`1.TryGetFirst(Boolean& found)
   at System.Linq.OrderedEnumerable`1.TryGetElementAt(Int32 index, Boolean& found)
   at System.Linq.OrderedPartition`1.TryGetFirst(Boolean& found)
   at System.Linq.Enumerable.SelectIPartitionIterator`2.TryGetFirst(Boolean& found)
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Boolean& found)
   at Akka.Persistence.Snapshot.MemorySnapshotStore.<>c__DisplayClass5_0.<LoadAsync>b__0()
   at System.Threading.Tasks.Task`1.InnerInvoke()
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
```

Aaronontheweb:
Fixes #3526 and includes a spec for verification


src/core/Akka.Persistence.TCK.Tests/MemorySnapshotStoreSpec.cs
```diff
@@ -23,5 +32,129 @@ public MemorySnapshotStoreSpec(ITestOutputHelper output)
             Initialize();
         }
 
+        [Fact]
+        public void MemorySnapshotStore_is_threadsafe()
+        {
+            EventFilter.Error().Expect(0, () =>
```
Aaronontheweb: This comments describe most of what's going on inside this spec, but the gist of it is: spin up several `SnapshotStore` calls from persistent actors and force it to barf. The cause of the original `InvalidOperationException` is because the collection was being modified concurrently from multiple threads.

src/core/Akka.Persistence/Snapshot/MemorySnapshotStore.cs
```diff
@@ -9,69 +9,73 @@
...
         protected override Task DeleteAsync(SnapshotMetadata metadata)
         {
-            Func<SnapshotEntry, bool> pred = x => x.PersistenceId == metadata.PersistenceId &&
-            (metadata.SequenceNr <= 0 || metadata.SequenceNr == long.MaxValue || x.SequenceNr == metadata.SequenceNr) &&
-            (metadata.Timestamp == DateTime.MinValue || metadata.Timestamp == DateTime.MaxValue || x.Timestamp == metadata.Timestamp.Ticks);
+            bool Pred(SnapshotEntry x) => x.PersistenceId == metadata.PersistenceId && (metadata.SequenceNr <= 0 || metadata.SequenceNr == long.MaxValue || x.SequenceNr == metadata.SequenceNr)
+                                                                                    && (metadata.Timestamp == DateTime.MinValue || metadata.Timestamp == DateTime.MaxValue || x.Timestamp == metadata.Timestamp.Ticks);
+            
 
-            return Task.Run(() =>
```
Aaronontheweb: The root cause of the issues the `MemorySnapshotStore` had were that we're using a `List<T>`, which is not thread safe, and were kicking off all of our operations inside the `SnapshotStore` asynchronously via `Task.Run`. 

My solution was to just get rid of the `Task.Run` calls and have all of the work done to the in-memory collection occur synchronously inside the `SnapshotStore` actor. This caused the test to go from fail to pass.

Aaronontheweb:
Gonna need to fix the API approval spec for Akka.Persistence back at home - my work computer isn't picking up the Git diff when I run it locally :(

Danthar:
So this basically serializes access to the MemorySnapshotStore. So potentially less performance. But more stability. While keeping the code simple. I think thats a fair tradeoff, considering this thing is not meant for production use cases.

Aaronontheweb:
>  So potentially less performance. But more stability. While keeping the code simple. I think thats a fair tradeoff, considering this thing is not meant for production use cases.

Correct. It's intended for dev & test purposes.

PR #3531, opened 2018-06-29
File publisher allows opening the same file for reading multiple times

Sam13:
As discussed in Gitter: Fixes bug in `FileIO.FromFile` that the same file can be opened multiple times at the same time

### DIFF
--- a/src/core/Akka.Streams.Tests/IO/FileSourceSpec.cs
+++ b/src/core/Akka.Streams.Tests/IO/FileSourceSpec.cs
@@ -202,6 +202,31 @@ namespace Akka.Streams.Tests.IO
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


src/core/Akka.Streams.Tests/IO/FileSourceSpec.cs
```diff
@@ -201,6 +201,31 @@ public void FileSource_should_complete_only_when_all_contents_of_a_file_have_bee
             }, _materializer);
         }
 
+        [Fact]
+        public void FileSource_should_open_file_in_shared_mode_for_reading_multiple_times()
```
marcpiechura: 👍

marcpiechura:
Thanks a lot @Sam13

Issue #2378, opened 2016-11-15
Akka.Remote is really slow


florisch:
Hi,

The company I work at is working on two new projects involving Akka.Net. We designed our application around Akka clustering capabilities, now we see big performance issues with Akka cluster/remote application.

In our applications we have many (more than 20) nodes. Each node sending more than 30K msg/sec to a central server. We have a simulation mode where we generate all the messages from the same server/process in wich case everythings works fine. As soon as we generate the messages using Akka.Remote our performances issues appears (and even cause a "BufferIsEmpty" inside Sockets/Helios exception on the linux arm mono nodes after some time wich just kill the process).

In order to understand the performance limitations of Akka, on a modern I7 computer, we ran simple pingpong benchmark in two scenarios.

- With one single application running receivers and senders I get a good 25 Million messages per seconds.

- Running receiver and sender actors in two processes on a local machine communicating through Akka.Remote (127.0.0 with wire serializer). We get 54K messages per seconds with 100% of cpu usage.

Of course we see that in this condition our application will never be able to run on with current Akka.net remote performances. We understand that remote involves a lot more processing than local message passing because of serialization and network but here after a quick profiling session it appears that most of the time is spent doing some internal akka/remote stuff and not in helios/wire/protobuf.

Is this performance considered normal or is there any hope to have better performances by tweaking options or any scheduled remote improvement ?

rogeralsing:
The new Stream transport bumps this up to ~100k messages per second.

That being said, there are some design issues we have inherited from the Scala implementation.
e.g. every message have a sender, the sender, the sender is an actorref and that has to be serialized, deserialized and then resolved on the remote system.
This is one of the major bottlenecks.

So even if the receiving side don't use the Sender, it will still be there and there is nothing we can do except use some form of cache for the sender lookup.

In JVM Akka, they have recently introduced a new transport "Artery" built on the new Aeron protocol.
In this transport they now can store actorrefs by ID, thus only passing the actorref as an integer between two nodes. 
Using protobuf and the new transport they are able to do ~300 messages per second.

There are no on plans right now to port this transport to .NET, it would be possible as Aeron also exist for .NET now.

So I dont really have a good answer for you. this is how it is right now.
Project Orleans also have a throughput of ~100k msg/sec, but AFAIK, without ordering.

florisch:
Thanks for your answer. From my perspective this is a major deal-breaker for large scale distributed applications witch at first really look like the big selling point of Akka with clustering.

[...]

PR #3532, opened 2018-06-29
[WIP] Akka.Serializer performance fixes

Aaronontheweb:
Noticed some unnecessary allocations inside the `Serializer` class while I was doing some other work today (working on an Akka.Persistence) plugin. Redesigned those portions to prevent garbage collection.

This code affects not just Akka.Remote, but also the Akka.Persistence and any other serializer that might need to serialize an `IActorRef`.

When I ran the RemotePingPong numbers I didn't see a significant material difference - not sure why yet.

WIP-ing this because I want to see some of the build server numbers and because I think it may be worth benchmarking this specific method independently from the entire Akka.Remote messaging pipeline.

### DIFF
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -12,6 +12,7 @@ using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Reflection;
 using System.Runtime.Serialization;
+using System.Threading;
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
Related to https://github.com/akkadotnet/akka.net/issues/2378

Horusiath:
API approvals needs to be fixed.

Issue #3517, opened 2018-06-20
Akka.Streams - port Flow.ask stage


Horusiath:
[Flow.ask](https://doc.akka.io/docs/akka/2.5/stream/operators/Source-or-Flow/ask.html) is another kind of interop stage between streams and actors.

Horusiath:
Done by #3539 as part of future release v1.4.

PR #3539, opened 2018-07-06
Reintroduce stream refs

Horusiath:
This PR reintroduces changes for `FunctionRef` and `StreamRefs`, that were accidentally reverted. 
Also introduces `Flow.Watch` and `Flow.Ask` (see #3517).

/cc @Aaronontheweb


Horusiath:
No idea why, but I cannot attach updated api approvals - git just doesn't pick them up.

Aaronontheweb:
@Horusiath I started having the same problem recently too - wonder if it's something we added to .GITIGNORE that is doing it. I had this issue long before we did https://github.com/akkadotnet/akka.net/pull/3535 though

pshrosbree:
You could edit your gitignore and see if you still have the problem...

Horusiath:
It was my fault - when test failure was happening, I got approved/received files and edited them directly - therefore I was editing copy represented in bin instead of the original file being copied.

Aaronontheweb:
`FlowAskSpec.Flow_with_ask_must_signal_ask_failure` - legit failure

PR #3551, opened 2018-07-15
Coordinated downing improvements

zbynek001:
contains 

https://github.com/akkadotnet/akka.net/pull/3263
- ClusterSingletonManager should ignore FSM events during shutdown

https://github.com/akka/akka/issues/24048
- Run all CoordinatedShutdown phases also when downing
- add Reason to CoordinatedShutdown

https://github.com/akka/akka/issues/24140
- Allow member to leave a cluster via CoordinatedShutdown.run when MemberStatus is Joining/WeaklyUp/Up.

https://github.com/akka/akka/issues/23449
- ClusterSpec, race between MemberRemoved and MemberExited

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/CoordinatedShutdownShardingSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/CoordinatedShutdownShardingSpec.cs
@@ -1,14 +1,17 @@
-﻿using System;
-using System.Collections.Generic;
+﻿//-----------------------------------------------------------------------
+// <copyright file="CoordinatedShutdownShardingSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2016 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2016 Akka.NET project <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
 using System.Linq;
-using System.Text;
 using System.Threading.Tasks;
 using Akka.Actor;
-using Akka.Cluster.Sharding.Serialization;
 using Akka.Cluster.Tools.Singleton;
 using Akka.Configuration;
 using Akka.TestKit;
-using Akka.TestKit.TestActors;
 using FluentAssertions;
 using Xunit;
 using Xunit.Abstractions;
@@ -17,17 +20,17 @@ namespace Akka.Cluster.Sharding.Tests
 {
     public class CoordinatedShutdownShardingSpec : AkkaSpec
     {
-        private ActorSystem sys1;
-        private ActorSystem sys2;
-        private ActorSystem sys3;
+        private readonly ActorSystem _sys1;
+        private readonly ActorSystem _sys2;
+        private readonly ActorSystem _sys3;
 
-        private IActorRef region1;
-        private IActorRef region2;
-        private IActorRef region3;
+        private readonly IActorRef _region1;
+        private readonly IActorRef _region2;
+        private readonly IActorRef _region3;
 
-        private TestProbe _probe1;
-        private TestProbe _probe2;
-        private TestProbe _probe3;
+        private readonly TestProbe _probe1;
+        private readonly TestProbe _probe2;
+        private readonly TestProbe _probe3;
 
         private static readonly Config SpecConfig;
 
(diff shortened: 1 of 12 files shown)


Horusiath:
To be honest I think that #3284 may have covered this PR.

zbynek001:
It might, but I guess that's still far away.
Just wanted to share it when I already migrated it. But it's up to you whether you merge it or not. If not, will just add it to my custom build.

src/contrib/cluster/Akka.Cluster.Sharding.Tests/CoordinatedShutdownShardingSpec.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        protected override void BeforeTermination()
+        {
+            Shutdown(_sys1);
+            Shutdown(_sys2);
+        }
+
+        /// <summary>
+        /// Using region 2 as it is not shutdown in either test.
```
Aaronontheweb: Good to know (the comment was helpful during my review :p)

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -379,8 +379,15 @@ private void SetupCoordinatedShutdown()
             var self = Self;
             _coordShutdown.AddTask(CoordinatedShutdown.PhaseClusterShardingShutdownRegion, "region-shutdown", () =>
             {
-                self.Tell(GracefulShutdown.Instance);
-                return _gracefulShutdownProgress.Task;
+                if (Cluster.IsTerminated || Cluster.SelfMember.Status == MemberStatus.Down)
```
Aaronontheweb: Looks good

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -152,6 +152,76 @@ public static CoordinatedShutdown Get(ActorSystem sys)
...
 
+
+
+        /// <summary>
+        /// Reason for the shutdown, which can be used by tasks in case they need to do
+        /// different things depending on what caused the shutdown. There are some
+        /// predefined reasons, but external libraries applications may also define
+        /// other reasons.
+        /// </summary>
+        public class Reason
```
Aaronontheweb: I like this!

Horusiath: Actually this should be an enum.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1668,7 +1668,7 @@ public void Welcome(Address joinWith, UniqueAddress from, Gossip gossip)
         public void Leaving(Address address)
         {
             // only try to update if the node is available (in the member ring)
-            if (_latestGossip.Members.Any(m => m.Address.Equals(address) && m.Status == MemberStatus.Up))
+            if (_latestGossip.Members.Any(m => m.Address.Equals(address) && (m.Status == MemberStatus.Joining || m.Status == MemberStatus.WeaklyUp || m.Status == MemberStatus.Up)))
```
Aaronontheweb: Resolves https://github.com/akkadotnet/akka.net/pull/2347/ while we're at it.

src/core/Akka/Actor/CoordinatedShutdown.cs
Aaronontheweb: Disagree. If it needs properties and data that can be overridden in the future, for whatever reason, you can't have that extensibility with an `Enum`

Aaronontheweb: I.E. the `CustomReason` example used in the spec earlier illustrates one such case for that.

Horusiath: Then you can easily change it into tagged union or just change from enum to class. It's time to stop putting enum-like types into singletons:

1. They blow up the code base.
2. They are hardly maintainable. You need a lot of boilerplate to keep things like equality, hash codes and string formatting in place.
3. They are not safe - you never know if you have covered all switch cases and no tool on earth will help you with that.
4. They are slow and memory-consuming (this don't matter much in this particular case).

If you can model some cases as enums, you should do that. This is not JVM, they won't be optimized automatically.

Aaronontheweb: None of the above is true in this particular case - it's a class that gets used once on shutdown. I don't think these are even true generally.

zbynek001: I actually started this as an enum, but then I got to the custom reason and I got stuck.
To support custom reasons with enum, we'd need to cast it to another enum extended with the custom reasons and I didn't like that. Honestly, the current approach I also don't like so much

PR #3555, opened 2018-07-22
Fixed ReceiveTimeout not triggered in some case when combined with NotInfluenceReceiveTimeout messages

ismaelhamed:
Port [#24927](https://github.com/akka/akka/pull/24927)

### DIFF
--- a/src/core/Akka/Actor/ActorCell.ReceiveTimeout.cs
+++ b/src/core/Akka/Actor/ActorCell.ReceiveTimeout.cs
@@ -10,7 +10,7 @@ using System;
 namespace Akka.Actor
 {
     /// <summary>
-    /// TBD
+    /// Marker interface to indicate that a message should not reset the receive timeout.
     /// </summary>
     public interface INotInfluenceReceiveTimeout
     {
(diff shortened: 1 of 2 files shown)


src/core/Akka/Actor/ActorCell.ReceiveTimeout.cs
```diff
@@ -47,7 +47,7 @@ public void SetReceiveTimeout(TimeSpan? timeout=null)
         public void CheckReceiveTimeout()
         {
             CancelReceiveTimeout();
-            if (_receiveTimeoutDuration != null && !Mailbox.HasMessages)
+            if (_receiveTimeoutDuration != null)
```
ismaelhamed: There were some performance concerns about this re-scheduling always now, but the PR was approved anyway.

Aaronontheweb: > I agree that this is the most straight forward way to implement it. I'm somewhat concerned about the performance implication of always rescheduling, but other solutions I can think of are too complicated and have other drawbacks.

I agree with what Patrik said.

PR #3558, opened 2018-07-29
Sequence number should not be set if snapshot is unhandled

ismaelhamed:
> The Eventsourced trait does update the sequence number to the snapshoted even if the snapshot is not handled by the actor and then continues recovering from there.
> 
> This may lead to only partially recovered actors. If the snapshot is not handled the actor should recover from scratch to prevent such corrupted states.

Port [#23469](https://github.com/akka/akka/pull/23469)

### DIFF
--- a/src/core/Akka.Persistence/Eventsourced.Recovery.cs
+++ b/src/core/Akka.Persistence/Eventsourced.Recovery.cs
@@ -88,10 +88,14 @@ namespace Akka.Persistence
                         timeoutCancelable.Cancel();
                         if (res.Snapshot != null)
                         {
-                            var snapshot = res.Snapshot;
-                            LastSequenceNr = snapshot.Metadata.SequenceNr;
-                            // Since we are recovering we can ignore the receive behavior from the stack
-                            base.AroundReceive(recoveryBehavior, new SnapshotOffer(snapshot.Metadata, snapshot.Snapshot));
+                            var offer = new SnapshotOffer(res.Snapshot.Metadata, res.Snapshot.Snapshot);
+                            var seqNr = LastSequenceNr;
+                            LastSequenceNr = res.Snapshot.Metadata.SequenceNr;
+                            if (!base.AroundReceive(recoveryBehavior, offer))
+                            {
+                                LastSequenceNr = seqNr;
+                                Unhandled(offer);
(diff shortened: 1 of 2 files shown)


src/core/Akka.Persistence/Eventsourced.Recovery.cs
```diff
@@ -88,10 +88,15 @@ private EventsourcedState RecoveryStarted(long maxReplays)
                         timeoutCancelable.Cancel();
                         if (res.Snapshot != null)
                         {
-                            var snapshot = res.Snapshot;
-                            LastSequenceNr = snapshot.Metadata.SequenceNr;
-                            // Since we are recovering we can ignore the receive behavior from the stack
-                            base.AroundReceive(recoveryBehavior, new SnapshotOffer(snapshot.Metadata, snapshot.Snapshot));
+                            var offer = new SnapshotOffer(res.Snapshot.Metadata, res.Snapshot.Snapshot);
+                            if (base.AroundReceive(recoveryBehavior, offer))
```
ismaelhamed: In scala you can check whether a partial function can handle a message (i.e, `recoveryBehavior.isDefinedAt(offer)`) before actually calling it, but I don't think we can do anything like that in C#, except to actually invoke it.

Horusiath: Yes, we cannot do that in C#, but also we cannot allow to do something like this (executing AroundReceive before setting a snapshot) - as this is a change of behavior.

What we could do is to optimistically set `LastSequenceNr`  and call `base.AroundReceive`. Then when returned value was false, we can retroactivelly set `LastSequenceNr` back to the previous value:

```csharp
var seqnr = LastSequenceNr;
LastSequenceNr = res.Snapshot.Metadata.SequenceNr;
if (!base.AroundReceive(recoveryBehavior, offer) 
{
    LastSequenceNr = seqNr;
    Unhandled(offer);
}
```

ismaelhamed: I'll push a fix later today.

ismaelhamed: @Horusiath just in case you missed my last push.

PR #3559, opened 2018-07-30
Fix small typo

fdbeirao:
I don't believe `Selet` is a valid method


Aaronontheweb:
thanks for your contribution @fdbeirao

Issue #3234, opened 2017-12-30
Port remaining stages from akka.streams.contrib repository


marcpiechura:
We've ported retry stage in #3233 but there are a few more left, this would be a good opportunity for beginners since most of those stages are quite simple. 

https://github.com/akka/akka-stream-contrib/tree/master/contrib/src/main/scala/akka/stream/contrib

KarolKalbarczyk:
I would like to try and implement at least some of those stages, is it still up for grabs?

Arkatufus:
@KarolKalbarczyk yes, they're still up for grabs

PR #3560, opened 2018-07-30
Port KeepAliveConcat and UnfoldFlow

alex-bogomaz:
Hi, 

please review this port of ```KeepAliveConcat``` and ```UnfoldFlow``` streams stages. 
This is for issue https://github.com/akkadotnet/akka.net/issues/3234.


alex-bogomaz:
@marcpiechura - I added 1 tests that checks that ```KeepAliveConcat```  still emits buffered elements after upstream completes.  Elements go through ```KeepAliveConcat``` with sufficiently large time-out to keep all them in buffer, than upstream completes and test check if elements are still emitted after that.

marcpiechura:
@alex-bogomaz thanks for verifying it !

PR #3565, opened 2018-08-05
_lastTemperatureReading should be updatable

nyolamike:
private readonly double? _lastTemperatureReading = null;
on line 61 would make this field not editable because of the red only modifier


Danthar:
this is true. but to be fair. The property isn't updated anywhere anyway :) its just used as a poc. :) But ill approve either way.

PR #3570, opened 2018-08-15
Port PagedSource & IntervalBasedRateLimiter

alex-bogomaz:
Hi 

please review port of these streams stages:
https://github.com/akka/akka-stream-contrib/blob/master/contrib/src/main/scala/akka/stream/contrib/PagedSource.scala
and
https://github.com/akka/akka-stream-contrib/blob/master/contrib/src/main/scala/akka/stream/contrib/IntervalBasedRateLimiter.scala

This is for issue #3234


Aaronontheweb:
thanks for your contribution @alex-bogomaz - forgot to say that earlier!

PR #3579, opened 2018-08-24
Add ToString override to TestProbe.

basbossinkdivverence:
This minimal change adds a `ToString` `override` to `TestProbe` to be able to use it in debug logging of tests.


Aaronontheweb:
Thanks! @basbossinkdivverence - we'll merge this in.

PR #3580, opened 2018-08-25
[build system] fix NuGet package dependency version overrides

Aaronontheweb:
resolves issue with production beta packages like Cluster.Sharding depending on the wrong version, i.e.:

image

All of those packages, except for DData, should be references to the stable version. Not beta packages. This PR resolves that issue for all beta packages we still publish.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -298,15 +298,6 @@ Target "NBench" <| fun _ ->
 // Nuget targets 
 //--------------------------------------------------------------------------------
 
-let overrideVersionSuffix (project:string) =
-    match project with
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Going forward, if we need to release any new beta packages, we just need to modify the `.CSPROJ` like this:

https://github.com/akkadotnet/akka.net/blob/5a346f276d8ddc9cce3e1050619946e1be667792/src/contrib/cluster/Akka.Cluster.Sharding/Akka.Cluster.Sharding.csproj#L7

PR #3582, opened 2018-08-27
Port TimeWindow streams stage

alex-bogomaz:
Hi, please review port of this streams stage:
https://github.com/akka/akka-stream-contrib/blob/master/contrib/src/main/scala/akka/stream/contrib/TimeWindow.scala

This is for issue #3234


Horusiath:
This method is in essence combination of 2 different stages (no custom functions, no extra classes) - is it really so common to use, that it should be a new method in an API? Maybe showing it as pattern or sample would be enough?

alex-bogomaz:
Maybe you are right. Stage is really a combo of 2 other. 
If this should be a part of documentation - where this should be added?

Closing PR for now.

Issue #3128, opened 2017-09-27
ActorSelection wildcard matching rules


spirit11:
```cs
var sys = ActorSystem.Create("sys");
var worker = sys.ActorOf<Worker>("worker");
sys.ActorSelection("akka://sys/user/ker*").Tell(42);
```
Why in this case worker receives the message? I'd expected that "akka://sys/user/**\*ker\***" matches "worker" but "akka://sys/user/**ker\***" should not. If it is not a proper behaviour it can be fixed in "Util/StringLike.cs"

marcpiechura:
It’s indeed a bug within the Regex in `StringLike`. The problem is that it matches `worker` with `ker.*?` ( `*` is replaced with `.*?` ) so it will return true for any apperance of `ker` no matter if it starts with `ker` or not. I don’t have much Regex skills but I think the solution would be to add `^` at the beginning of the pattern, i.e. `^ker.*?`. This will only match if the string starts with `ker`. The old behavior could still be archived by using the wildcard `*ker*`.

nbparrell:
I will be working on this.

nbparrell:
Suggested solution:
if(pattern[0] == '*' && pattern[pattern.Length-1]!='*'){
                pattern = pattern.Replace("*", "(.*?)");
                pattern = pattern + "(?!.)";
            }else if(pattern[0]!='*' && pattern[pattern.Length-1]=='*'){
                pattern = pattern.Replace("*", "(.*?)");
                pattern = @"\A" + pattern;
            }else{
                pattern = pattern.Replace("*", "(.*?)");
            }

marcpiechura:
@nbparrell 
I think it’s even simpler, replace 
`Replace("*", ".*?");`
With
`Replace("*", "^.*?");`

Would also be good to add a unit test to verify the new behavior.

nbparrell:
In either case we would have to figure out the placement of the \* in order to make sure that we are placing  the new regex expression in the correct place. That being said I used \A because is clarifies that the particular pattern has to be at the beginning of a string where as ^ is the beginning of a line. 

 The issue with using ^ is that when you place ^\.\*? it will pick up any characters before the given pattern.


So given:
  pattern = ker\*
  Actor kernal
  Actor worker
 The resulting pattern will be \^\*?ker.\*? which would pick up both kernal and worker since the given pattern starts at the beginning of the line. In this instance we want it only to select kernal.

[...]

PR #3601, opened 2018-09-25
Actor selection #3128

pshrosbree:
Addressing [up for grabs #3128](https://github.com/akkadotnet/akka.net/issues/3128)

Stops the following from matching the selection:

```var sys = ActorSystem.Create("sys");
var worker = sys.ActorOf<Worker>("worker");
sys.ActorSelection("akka://sys/user/ker*").Tell(42);```


Aaronontheweb:
Thanks for your contribution, @pshrosbree

PR #3605, opened 2018-09-29
Don't automatically down quarantined nodes

ismaelhamed:
Currently, a quarantined node will be automatically downed and removed from the cluster without the downing provider being aware of the situation.

Port [#25678](https://github.com/akka/akka/pull/25678)

### DIFF
--- a/src/core/Akka.Cluster.Tests.MultiNode/SurviveNetworkInstabilitySpec.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/SurviveNetworkInstabilitySpec.cs
@@ -157,7 +157,7 @@ namespace Akka.Cluster.Tests.MultiNode
             A_Network_partition_tolerant_cluster_must_heal_after_one_isolated_node();
             A_Network_partition_tolerant_cluster_must_heal_two_isolated_islands();
             A_Network_partition_tolerant_cluster_must_heal_after_unreachable_when_ring_is_changed();
-            A_Network_partition_tolerant_cluster_must_down_and_remove_quarantined_node();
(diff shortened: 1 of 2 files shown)


src/core/Akka.Cluster.Tests.MultiNode/SurviveNetworkInstabilitySpec.cs
```diff
@@ -426,11 +425,36 @@ public void A_Network_partition_tolerant_cluster_must_down_and_remove_quarantine
...
 
+                RunOn(() =>
+                {
+                    Thread.Sleep(2000.Milliseconds());
+
+                    // not be down
+                    var secondUnique = Cluster.State.Members.SingleOrDefault(m => m.Address == GetAddress(_config.Second) && m.Status == MemberStatus.Up);
+                    if (secondUnique == null)
+                    {
+                        // TODO: fail("Unexpected removal of quarantined node")
```
ismaelhamed: This should be some sort of assertion fail. Couldn't figure out the api for that in Akka.NET, yet.

Horusiath: You can just use Fluent Assertions. Lookin at the code from original test:

```csharp
var secondUniqueAddress = Cluster.State.Members.SingleOrDefault(m => m.Address == GetAddress(_config.Second));
secondUniqueAddress.Should().Not.BeNull(because: "2nd node should stay visible");
secondUniqueAddress.Status.Should().Be(MemberStatus.Up, because: "2nd node should be Up");
```

Aaronontheweb:
Thanks @ismaelhamed !

PR #3618, opened 2018-10-13
Update Azure PaaS Worker Role example to use the provided `CancellationToken`

caioproiete:
Update the [Azure PaaS Worker Role example](https://getakka.net/articles/intro/use-case-and-deployment-scenarios.html#azure-paas-worker-role) to use the provided `CancellationToken` when _awaiting_.


Aaronontheweb:
Thanks for your contribution!

PR #3622, opened 2018-10-15
fix failed when run test on Tutorials.Tutorial4

torresng:
Run dotnet test failed when using Tutorials.Tutorial4 on main Program.cs


Aaronontheweb:
Thanks, @torresng - we'll get this merged in.

PR #3628, opened 2018-10-17
[WIP] ddata protobuf

Horusiath:
**WORK IN PROGRESS**

This PR will introduce new serializer for DData messsages and data types - based on Protocol Buffers definitions send by JVM version.

The major pain point is about serialization (or actually deserialization) of ddata keys: `ORSetKey<T>`, `LWWDictionaryKey<T>` etc. It's hard since, the generic param is not encoded in anyway. In JVM this is not a problem (because erased generics). Here, we don't have that comfort. 

1. We cannot just add extra field to store generic type, as this may be persisted on disk and it's not version tolerant: any change made on type name will cause its key to be irrecoverably unserializable.
2. For this reason my proposal is simply to **drop typed keys and use plain strings instead**. This ofc means, that the API will change.


Aaronontheweb:
> For this reason my proposal is simply to drop typed keys and use plain strings instead. This ofc means, that the API will change.

Makes sense to me.

Aaronontheweb:
On second thought, I want to take a look at this before we go through with changing the API.

src/contrib/cluster/Akka.DistributedData/reference.conf
```diff
@@ -98,4 +98,36 @@ akka.actor {
 	"Akka.DistributedData.Serialization.ReplicatedDataSerializer, Akka.DistributedData" = 11
     "Akka.DistributedData.Serialization.ReplicatorMessageSerializer, Akka.DistributedData" = 12
   }
+  serialization-settings {
+    akka-replicated-data {
+        mappings {
+            "System.Object" = 0
```
Horusiath: This is a general idea to keep things working. While this design works with mostly the same protobuf definitions as core akka ddata ones, there was a serious problem around notion of generics: protobuf doesn't support generic types. In akka it's not the problem, as JVM erases them anyway and their solution is basically *fuck it, just assume they don't exist*.

This however is not the case in .NET - we need to know all generic type parameters before we instantiate generic types using them. My solution here was to add extra `uint32` fields in protobuf definitions where they were necessary - those fields provide basic tagging of core types. 

Since we may need to declare tagging for types that are defined in assemblies, users don't have control over, I've decided to move them to be defined in HOCON as you can see here.

Horusiath:
Ok, to shortly describe generics problem - protobuf doesn't support them, so we need some way to be able to deserialize a generic message of eg. `LWWDictionary<TKey, TVal>` where `TKey` and `TVal` are not known prior the deserialization. This means, that the type must be encoded into the protobuf payload itself. 

The most efficient thing seems to be using a varint fields used as generic message descriptors, and this is what I'm doing in this PR. These fields are extended beyond original JVM akka protobuf definition, so in the future if we want to have some sort of JVM interop, the corresponding protobuf fields will be 0 - for this reason I've used 0 number as descriptor for System.Object.

Now, var ints let us to describe types in compact way and don't require to encode type name into payload - it's important property, as we're persisting that data. If we'd have to persist type name, it means it would never be allowed to change. Since it's a number, we can change it by changing type<->id mapping, at the cost of having the mapping itself.

For now, I'm using an `akka.actor.serialization-settings` feature to supply our ddata serializer with a list of mappings. It's pre-filled with some well known akka types. This means that any user custom class needs to be defined in the configuration to work with new ddata serializer.

Some other solutions (that can work together with current one):

### Tag messages with attributes

Basically we could define some custom attribute like `[TypeTag(123)]` to be used on the messages and just read all referenced assemblies at the actor system start to fill the mapping table.

Pros:

- easy
- opinionated

Cons:

- it requires user to have dependency on Akka.DistributedData at the user message level, which is not always desired
- assemblies loaded dynamically at runtime can also be a small problem

### Assembly mapping resolver

We could extend configuration to define i.e. some specific class that's supposed to supply us with the necessary mappings, like:

```csharp
// This could be the same Startup class, that ASP.NET users use to configure their system. 
class Startup
{
    public IDictionary<Type, int> GetTypeMappings() { }
}
```

Pros:

- highly configurable - it's up to users how they want to supply our mappings. They could define their own attributes, reuse attributes from existing serializer libraries, or even use some naming conventions.

Cons:

- you still need to define startup class in the config - you just need to define it once, instead of having a config for every single type you want to use

Issue #3506, opened 2018-06-13
DotNetty Remote Transport Issues with .NET Core 2.1


DaKaLKa:
I have a test application with 1 seed node and two service nodes.
After migrating to .NET Core 2.1, I have some  "Can't join seed-node" issues using Akka.net 1.3.6 / 1.3.8.  The software is build and executed in docker with official Microsoft .NET Core 2.1 Images (microsoft/dotnet:2.1-sdk and microsoft/dotnet:2.1-runtime). Sometimes the nodes get connected without problems, sometimes I have the "can't join seed-nodes"-problems without any changes. 
The problem appears on my local laptom and on our docker-server (and some other test-Systems).

The same application caused no problems running it on windows through Visual Studio. 
Migration back to ".NET Core 2.0" solved the problem for now. (Project-Settings and other microsoft/dotnet:2.0-docker-Images).

Aaronontheweb:
Technically, everything we do should be working on .NET Core 2.1 since .NET Standard 1.6 can be consumed upstream. We'll investigate. Might be a DotNetty issue since v0.4.8 started using some of the unsafe APIs.

xlegalles:
It may be also valid for Windows as our implementation does not work anymore since we moved to .NET Core 2.1. We still need to investigate but the problem is exactly the same: remote actors cannot join anymore. We tested both in Docker and out of container.

xlegalles:
Forget my comment: it was not related to AKKA. So it means that we are able to use AKKA.Net with 2.1 under Windows.

annymsMthd:
We are currently running a 1.3.8 cluster with .net core 2.1 in k8s with only a few minor issues.

Lutando:
@annymsMthd  what were those minor issues?

annymsMthd:
Sometimes we have a node that starts kicking out other nodes after we deploy even though it's not leader. It just starts marking nodes as down and even though they restart it just targets them again and marks them. When we reset the node that is doing the marking everything works. We only see this when we deploy though so it isn't a show stopper.

Lutando:
We have a similar behavior. Might be worth investigating.

Cephei:
Experiencing the same issue. Net core 2.1 on Linux. The problem is random ; restating often works

Aaronontheweb:
Based on some of the quality issues we've had with DotNetty recently, I've been exploring a replacement for it built on Akka.Streams.

[...]

PR #3633, opened 2018-10-24
upgraded to DotNetty v0.6.0 and System.Collections.Immutable v1.5.0

Aaronontheweb:


Aaronontheweb:
Doing this to address some of the underlying causes of https://github.com/akkadotnet/akka.net/issues/3506

Aaronontheweb:
In case anyone asks: DotNetty depends on System.Collections.Immutable v1.5.0 so I decided to just upgrade both everywhere.

Aaronontheweb:
cc @Horusiath looks like Hyperion is blowing up using the latest System.Collections.Immutable - that's what all of the failed DData specs are about, right?

Aaronontheweb:
Looks like it's only an issue on .NET Core

Aaronontheweb:
Issue may have been caused by an older version of Hyperion

Aaronontheweb:
Looks like this is the issue I'm running into: https://github.com/akkadotnet/Hyperion/issues/60

Aaronontheweb:
So I can get these tests to pass in Visual Studio, but not on the command line on the build server or on my development machines. Something weird is going on with binary tolerance inside Hyperion.

Aaronontheweb:
These XUnit failures have nothing to do with Hyperion directly - it's an issue caused by how `dotnet xunit` resolves assemblies. Since `dotnet xunit` has been deprecated, going to move onto using `dotnet test` - which allows these specs to pass without issue. I was able to reproduce this issue related to `dotnet xunit` by just running the Hyperion test suite on its on - I have no idea why the XUnit runner loads two different versions of the same assembly during these .NET Core unit tests when executed via `dotnet xunit` but not when `dotnet test`.

Going to need to upgrade the build system first 🙄

Aaronontheweb:
Need to also take a look at https://github.com/akkadotnet/Hyperion/issues/100 as a potential source of the issue here

Aaronontheweb:
Been able to reproduce this Hyperion issue with https://github.com/akkadotnet/Hyperion/pull/105

src/contrib/cluster/Akka.DistributedData/Serialization/ReplicatedDataSerializer.cs
```diff
@@ -29,7 +29,7 @@ public ReplicatedDataSerializer(ExtendedActorSystem system) : base(system)
             _serializer =
                 new Hyperion.Serializer(new SerializerOptions(
                     preserveObjectReferences: true,
-                    versionTolerance: true,
+                    versionTolerance: false,
```
Aaronontheweb: I need to roll back the changes to this file

PR #3637, opened 2018-11-01
[Akka.Cluster] cleanup and fix event logging

### LINKED ISSUE #3635: Debug Logging of GossipStatus
1.3.9

Windows 10 64Bit

When I have loglevel set to DEBUG, but with actor and cluster debug options set to off, I still continuously receive some cluster gossip debug log messages, like:

``` 
 [Uninitialized] Received Akka.Cluster.InternalClusterAction+Subscribe

 [TryingToJoin] Received Akka.Cluster.InternalClusterAction+Welcome

 [Initialized] Received Akka.Cluster.GossipEnvelope

 [Initialized] Received GossipStatus(from=UniqueAddress: (akka.tcp://migrationarchitect@127.0.0.1:4053, 410406129), version=VectorClock(E5C3551E85FDF0DE5C54487E6B71E9A9->6))
 [Initialized] Received GossipStatus(from=UniqueAddress: (akka.tcp://system@127.0.0.1:4053, 410406129), version=VectorClock(E5C3551E85FDF0DE5C54487E6B71E9A9->6))

 [Initialized] Received Akka.Cluster.InternalClusterAction+ExitingConfirmed
```

The GossipStatus message repeats every second or so, from each node in the cluster, while the others are just at startup / shutdown.

My hocon is as follows:

```
	akka {
		loggers=["Akka.Logger.Serilog.SerilogLogger, Akka.Logger.Serilog"]
		log-config-on-start = off
		loglevel = DEBUG
		stdout-loglevel = INFO

		actor {
			provider = cluster			
			debug {
				receive = off
				autoreceive = off
				lifecycle = off
				event-stream = off
				unhandled = off
			}			
		}
	
		remote {
			log-remote-lifecycle-events = INFO
			dot-netty.tcp {
				hostname = "127.0.0.1"
				port = 0
			}
		}            

		cluster {
			seed-nodes = ["akka.tcp://system@127.0.0.1:4053"]			
			debug {
				verbose-heartbeat-logging = off
				verbose-receive-gossip-logging = off
			}                    
		}
	}

```

I would expect the GossipStatus messages to be covered by the 'verbose-receive-gossip-logging' setting in the config and not be sent when it is set to 'off'.

Is there some other config setting that would disable these messages?

Aaronontheweb:
Resolves #3635 and implements a bunch of more helpful, useful log messages inside the cluster which properly attribute the NODE ADDRESS to each logged message. Should be a nice usability improvement to what we have already.

### DIFF
--- a/src/core/Akka.Cluster/Cluster.cs
+++ b/src/core/Akka.Cluster/Cluster.cs
@@ -82,6 +82,11 @@ namespace Akka.Cluster
         public UniqueAddress SelfUniqueAddress { get; }
 
         /// <summary>
+        /// Used to retain the <see cref="InfoLogger"/> instance that decorates the cluster.
+        /// </summary>
+        internal InfoLogger CurrentInfoLogger { get; }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="Cluster"/> class.
         /// </summary>
         /// <param name="system">The actor system that hosts the cluster.</param>
@@ -100,12 +105,14 @@ namespace Akka.Cluster
 
             _log = Logging.GetLogger(system, "Cluster");
 
+            CurrentInfoLogger = new InfoLogger(_log, Settings, SelfAddress);
+
             LogInfo("Starting up...");
 
-            _failureDetector = new DefaultFailureDetectorRegistry<Address>(() => FailureDetectorLoader.Load(Settings.FailureDetectorImplementationClass, Settings.FailureDetectorConfig,
+            FailureDetector = new DefaultFailureDetectorRegistry<Address>(() => FailureDetectorLoader.Load(Settings.FailureDetectorImplementationClass, Settings.FailureDetectorConfig,
                 system));
 
-            _scheduler = CreateScheduler(system);
+            Scheduler = CreateScheduler(system);
 
             // it has to be lazy - otherwise if downing provider will init a cluster itself, it will deadlock
             _downingProvider = new Lazy<IDowningProvider>(() => Akka.Cluster.DowningProvider.Load(Settings.DowningProviderType, system), LazyThreadSafetyMode.ExecutionAndPublication);
@@ -481,7 +488,7 @@ namespace Akka.Cluster
         /// </summary>
         public ExtendedActorSystem System { get; }
 
-        private Lazy<IDowningProvider> _downingProvider;
+        private readonly Lazy<IDowningProvider> _downingProvider;
         private readonly ILoggingAdapter _log;
         private readonly ClusterReadView _readView;
 
@@ -490,12 +497,10 @@ namespace Akka.Cluster
         /// </summary>
         internal ClusterReadView ReadView { get { return _readView; } }
 
-        private readonly DefaultFailureDetectorRegistry<Address> _failureDetector;
-
         /// <summary>
         /// The set of failure detectors used for monitoring one or more nodes in the cluster.
         /// </summary>
-        public DefaultFailureDetectorRegistry<Address> FailureDetector { get { return _failureDetector; } }
+        public DefaultFailureDetectorRegistry<Address> FailureDetector { get; }
 
         /// <summary>
         /// TBD
@@ -506,11 +511,10 @@ namespace Akka.Cluster
         // ===================== WORK DAEMONS =====================
         // ========================================================
 
-        readonly IScheduler _scheduler;
         /// <summary>
         /// TBD
         /// </summary>
-        internal IScheduler Scheduler { get { return _scheduler; } }
+        internal IScheduler Scheduler { get; }
 
         private static IScheduler CreateScheduler(ActorSystem system)
         {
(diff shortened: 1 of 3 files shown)


src/core/Akka.Cluster/Cluster.cs
```diff
@@ -81,6 +81,11 @@ internal static bool IsAssertInvariantsEnabled
         /// </summary>
         public UniqueAddress SelfUniqueAddress { get; }
 
+        /// <summary>
+        /// Used to retain the <see cref="InfoLogger"/> instance that decorates the cluster.
+        /// </summary>
+        internal InfoLogger CurrentInfoLogger { get; }
```
Aaronontheweb: I don't have a nice clean `object` abstraction I can use to hide this stuff the same way we do on the JVM, and `static` members would have some unintended consequences here. Decided to expose the `InfoLogger` as an `internal` field, but I mostly just ended up using the `LogInfo` methods that were already built into the `Cluster` class instead.

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -561,13 +565,65 @@ internal IActorRef ClusterCore
             }
         }
 
+        /// <summary>
+        /// INTERNAL API.
+        ///
+        /// Used for logging important messages with the cluster's address appended.
+        /// </summary>
+        internal class InfoLogger
```
Aaronontheweb: Created an internal abstraction for the `InfoLogger` in order to try to follow roughly what the JVM does, but also because I imagined some use cases where we might want to pass it around in the future.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1121,7 +1118,7 @@ protected override void PreStart()
 
             if (_seedNodes.IsEmpty)
             {
-                _log.Info("No seed-nodes configured, manual cluster join required");
+                _cluster.LogInfo("No seed-nodes configured, manual cluster join required");
```
Aaronontheweb: Made this change in about 45 places inside the `ClusterCoreDaemon` - allows the `akka.cluster.log-info` setting to toggle outputting these messages on and off and also appends the `SelfAddress` of the cluster to each of the inbound and outbound messages, which should improve the quality of the data we get from the logs. In the future the `InfoLogger` will also include the ID of the datacenter this node is participating in.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1243,35 +1233,27 @@ private void Uninitialized(object message)
...
             }
-            else if (message is InternalClusterAction.JoinSeedNodes)
+            else if (message is InternalClusterAction.JoinSeedNodes js)
             {
-                var js = message as InternalClusterAction.JoinSeedNodes;
                 BecomeUninitialized();
                 JoinSeedNodes(js.SeedNodes);
             }
-            else if (message is InternalClusterAction.ISubscriptionMessage)
+            else if (message is InternalClusterAction.ISubscriptionMessage isub)
```
Aaronontheweb: Implementing C# 7 pattern matching as I pass through these classes, mostly for stylistic reasons and brevity.

src/core/Akka.Cluster/ClusterSettings.cs
```diff
@@ -57,8 +57,7 @@ public ClusterSettings(Config config, string systemName)
 
             Roles = cc.GetStringList("roles").ToImmutableHashSet();
             MinNrOfMembers = cc.GetInt("min-nr-of-members");
-            //TODO:
-            //_minNrOfMembersOfRole = cc.GetConfig("role").Root.GetArray().ToImmutableDictionary(o => o. )
```
Aaronontheweb: Already implemented this feature a long time ago, so the TODO was no longer needed.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1243,35 +1233,27 @@ private void Uninitialized(object message)
 
         private void TryingToJoin(object message, Address joinWith, Deadline deadline)
         {
-            if (_logInfo && !(message is InternalClusterAction.ITick))
-            {
-                _log.Debug("[TryingToJoin] Received {0}", message);
```
Danthar: This was intentionally removed?

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1311,21 +1293,14 @@ private void BecomeInitialized()
 
         private void Initialized(object message)
         {
-            if (_logInfo && !(message is InternalClusterAction.ITick))
-            {
-                _log.Debug("[Initialized] Received {0}", message);
```
Danthar: Same goes for this ?

Danthar:
Left some questions.

Aaronontheweb:
Yep, removed those on purpose

PR #3639, opened 2018-11-01
made it easier to generate local beta packages of Akka.NET

Aaronontheweb:
If TeamCity isn't available to pass in a build number environment variable, beta NuGet packages will now be named using the following convention:

```
[PackageName].[VerisonInReleaseNotes]-beta[DateTime.UtcNow.Ticks]
```

If the `BUILD_NUMBER` environment variable is available, NuGet packages will be built the same way they are now:

```
[PackageName].[VerisonInReleaseNotes]-beta[BUILD_NUMBER]
```

Ported the strategy over from NBench and Hyperion: https://github.com/akkadotnet/Hyperion/pull/103


Aaronontheweb:
Ok, nightlies look good - still using their TeamCity-assigned numbers

PR #3649, opened 2018-11-14
Stream ref fixes

### LINKED ISSUE #3620: StreamRef deserialization doesn't work for serializers without manifest
In Akka 1.4 beta 4, when trying to deserialize a message with an associated serializer that does not have a string manifest (Hyperion, Newtonsoft), the serializer crashes in Akka.Streams/Serialization/StreamRefSerializer.cs at:
158                    var type = Type.GetType(manifest, throwOnError: true);

The reason is manifest == null.

Removing the if condition at line 214 - 219 seems to solve the issue:
214 if (serializer.IncludeManifest)
215            {
216                manifest = serializer is SerializerWithStringManifest s
217                    ? s.Manifest(payload)
218                    : payload.GetType().TypeQualifiedName();
219            }

When removing the if condition, manifest is no longer null.

### LINKED ISSUE #3621: StreamRef subscription time out does not get canceled after handshake
In Akka 1.4 beta 4, when using stream refs, the stream receives a timeout after as much time as is defined in  stream.materializer.stream-ref.subscription-timeout. The OnSubscribeHandshake message is never received and as a result the CancelTimer(SubscriptionTimeoutKey) call (line 630 of Akka.Streams/Dsl/StreamRefs.cs) is never made. As a result the stream can only last as long as the subscription timeout.

Horusiath:
Fixes #3621
Fixes #3620

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
@@ -1829,12 +1829,13 @@ namespace Akka.Streams.Dsl
     }
     public sealed class StreamRefSettings
     {
-        public StreamRefSettings(int bufferCapacity, System.TimeSpan demandRedeliveryInterval, System.TimeSpan subscriptionTimeout) { }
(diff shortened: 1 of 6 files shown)


Aaronontheweb:
Looks like approvals failed on Windows

Aaronontheweb:
nice work @Horusiath

PR #3651, opened 2018-11-15
Mention a correct test framework

okobylianskyi:


Horusiath:
@okobylianskyi thanks! :)

PR #3653, opened 2018-11-15
Added full Akka.Cluster.Tools.ClusterClient example

Aaronontheweb:
Going to update the documentation on the website too, but for the time being - I've added a full example that uses three node types:

* Client
* Seed (receptionist)
* Worker

Shows how to start and configure the `ClusterClient` on both ends of the connection and how work can be distributed from the receptionist to other nodes inside the cluster that aren't necessarily hosting a receptionist themselves.


Aaronontheweb:
Unfortunately, the script I used for adding the copyright headers touched a bunch of other files - hence why the count is so large.

Issue #3654, opened 2018-11-18
PersistentFSM: Data passed to AndThen is not the latest version


aloker:
Akka version: 1.3.10
Platform : .NET Core 2.1/Windows

When defining a side effect in PersistentFSM, the data object passed to the AndThen callback is NOT the version returned from ApplyEvent. 

Consider this actor. When it receives a `Command`, it updates the state data (`Data`) by applying the `CommandReceived` event and uses the AndThen function to reply the current `Data` to the sender.
```
public class AndThenTestActor : PersistentFSM<AndThenTestActor.IState, AndThenTestActor.Data, AndThenTestActor.IEvent>
{
    public override string PersistenceId => "PersistentFSMSpec.AndThenTestActor";

    public AndThenTestActor()
    {
        StartWith(Init.Instance, new Data());
        When(Init.Instance, (evt, state) =>
        {
            switch (evt.FsmEvent)
            {
                case Command cmd:
                    return Stay()
                        .Applying(new CommandReceived(cmd.Value))
                        .AndThen(data =>
                        {
                            // NOTE At this point, I'd expect data to be the value returned by ApplyEvent
                            Sender.Tell(data, Self);
                        });
                default:
                    return Stay();
            }
        });
    }

    protected override Data ApplyEvent(IEvent domainEvent, Data currentData)
    {
        switch (domainEvent)
        {
            case CommandReceived cmd:
                // Data is immutable; a new object with the applied event is returned
                return new Data(cmd.Value);
            default:
                return currentData;
        }
    }


    public interface IState : PersistentFSM.IFsmState
    {
    }

    public class Init : IState
    {
        public static readonly Init Instance = new Init();
        public string Identifier => "Init";
    }

    public class Data
    {
        public Data()
        {
        }

        public Data(string value)
        {
            Value = value;
        }

        public string Value { get; }
    }

    public interface IEvent
    {
    }

    public class CommandReceived : IEvent
    {
        public CommandReceived(string value)
        {
            Value = value;
        }

        public string Value { get; }
    }

    public class Command
    {
        public Command(string value)
        {
            Value = value;
        }

        public string Value { get; }
    }
}
```
Let's say I ask the actor like so, I'd expect the returned data to reflect the latest changes.
```
var response = await actor.Ask<AndThenTestActor.Data>(new AndThenTestActor.Command("test"));
Assert.Equal("test", response.Value);
```
However, it does not! Instead I get the `Data` object that was the state data *before* applying the event.

[...]

Issue #136, opened 2014-05-29
AkkaSpec.expectMsg do not fail on timeouts


HCanber:
When `queue.TryTake` in `AkkaSpec.expectMsg` times out the test does not fail.

The reason for this is that the check is done if `TryTake` returns `true` which it does if _has_ received a message.

``` cs
if (queue.TryTake(out t, timeout))
{
    Assert.IsNotNull(t, "exected message {0} but timed out after {1}", expected, timeout);
    Assert.AreEqual(expected, t);
}
```

The fix (a pull request is coming) is simply to move the failure:

``` cs
if (queue.TryTake(out t, timeout))
{
    Assert.AreEqual(expected, t);
}
else
{
    Assert.Fail("exected message {0} but timed out after {1}", expected, timeout);
}
```

PR #3655, opened 2018-11-18
fix: data provided to AndThen (PersistentFSM) is not the latest version

aloker:
Fixes the issue https://github.com/akkadotnet/akka.net/issues/3654


src/core/Akka.Persistence/Fsm/PersistentFSM.cs
```diff
@@ -131,7 +131,8 @@ void ApplyStateOnLastHandler()
                     handlersExecutedCounter++;
                     if (handlersExecutedCounter == eventsToPersist.Count)
                     {
-                        base.ApplyState(nextState.Using(nextData));
+                        nextState = nextState.Using(nextData);
```
ismaelhamed: Isn't this a change in behavior, since nextState is actually set downstream in the MakeTransition method? Looking a the JVM, what we're missing is:

```csharp
base.ApplyState(nextState.Copy(stateData: nextData));
```

aloker: I must be missing something. Here's what's happening now in ApplyStateOnLastHandler which is called after the persistence is completed.

https://github.com/akkadotnet/akka.net/blob/d9f6c573afe1ba59c3d6733dda4bc1f7dd089efc/src/core/Akka.Persistence/Fsm/PersistentFSM.cs#L129-L143


With my comments:
```
 void ApplyStateOnLastHandler()
{
    handlersExecutedCounter++;
    if (handlersExecutedCounter == eventsToPersist.Count)
    {
        // Using creates a COPY of nextState with stateData set to nextData
        // nextState still refers to the OLD version WITHOUT the NEW nextData being set
        base.ApplyState(nextState.Using(nextData));

        // uses OLD nextState, but that doesn't matter, as Timeout hasn't changed        
        CurrentStateTimeout = nextState.Timeout;

        // HERE's the issue and the actual bug
        // The "AfterTransitionDo" callback (ie, AndThen) - if present - is
        // called with the StateData Property of the OLD, UNCHANGED state.
        // That is, this StateData is always lagging one version behind.        
        nextState.AfterTransitionDo?.Invoke(nextState.StateData);

        // the rest is not relevant
        if (doSnapshot)
        {
            Log.Info($"Saving snapshot, sequence number [{SnapshotSequenceNr}]");
            SaveStateSnapshot();
        }
    }
}
```

This issue was not discovered using the current set of unit tests, but it is very easily to show (eg. in the shopping cart example) that the current version is *broken*.

Please try it for yourself. Here's a snippet from the specs:
```
When(LookingAround.Instance, (evt, state) =>
{
    // NB: When in this state, the cart is empty
    if (evt.FsmEvent is AddItem addItem)
    {
        // visitor adds an item to the EMPTY cart
        return GoTo(Shopping.Instance)
            .Applying(new ItemAdded(addItem.Item))
            .ForMax(TimeSpan.FromSeconds(1))
            .AndThen(cart => /* what to you expect cart to be? Empty or non-empty? */);
    }
    [...]
});
```

The issue is at the "/* what to you expect cart to be? Empty or non-empty? */" part. You'd expect the cart to be non-empty, but it isn't. It's still the OLD, EMPTY version.

ismaelhamed: So I checked, and line [#136](https://github.com/akkadotnet/akka.net/blob/ad59e30ac3058a0c22301214b526a00011486c8c/src/core/Akka.Persistence/Fsm/PersistentFSM.cs#L137) needs to be fixed too:

```csharp
nextState.AfterTransitionDo?.Invoke(StateData);
```
With these two changes, it works as expected.

ismaelhamed:
Actually @aloker , the whole thing seems to be documented in this [issue](https://github.com/akka/akka/issues/23739), so IMO I'd be great if you could bring this [PR](https://github.com/akka/akka/pull/23740) too, and also check that we're not missing any tests for the PersistentFMS while you're at it ;)

cc: @Aaronontheweb @Horusiath @Danthar

Danthar:
+1 from me

Op za 24 nov. 2018 12:57 schreef Ismael Hamed <notifications@github.com:

src/core/Akka.Persistence/Fsm/PersistentFSM.cs
aloker: It should actually work without. To me, it's either-or. My PR is one solution. The other one is `nextState.AfterTransitionDo?.Invoke(nextData);` which fixes the issue, too, but for my personal taste is the less clean approach. Ymmv.

aloker:
@ismaelhamed  I'll have a look at it. Thanks for your time!

src/core/Akka.Persistence/Fsm/PersistentFSM.cs
aloker: Ok, the jvm version uses the second approach. I'll stick to that and will update the PR asap.

Aaronontheweb:
:+1: thanks for the review @ismaelhamed - when this is up to date, I'd be happy to review it for merge

aloker:
Sorry for the delay. My daily work took all of my time. I have the rest of the year off so I'll be able to take care of this next week.

Aaronontheweb:
@aloker I'd be happy to merge this in once the PR is updated - or you and @ismaelhamed resolve your conversation from earlier. Sound good?

ismaelhamed:
@aloker Is there any chance of this making it into v1.4? :wink:

Aaronontheweb:
@ismaelhamed yep - let me re-run CI and check the results

Aaronontheweb:
@ismaelhamed is this change good as-is? Or do we need to make some additional updates to it?

ismaelhamed:
@Aaronontheweb  I think there're a couple of changes still pending

PR #3660, opened 2018-11-26
Changed copyright and added CSS to fix links issue. #3393

### LINKED ISSUE #3393: Akka.NET home page broken link and copyright
A partial regression of #3106. 

Link non-functional: "Learn about Akka for the JVM **here**." The "here" is not active.

Copyright in footer: Year should be 2018. Could add some JavaScript to make it roll over each year, unless there's a reason for not having JS in footer?

rubenmamo:
Fixes #3393 

I've changed the copyright to 2018.

The issue with the links was due to the h4 having ```height: 100px; margin-top: -100px;```. There was a comment in the css file as follows ```/* workaround for leave space for fixed navbar with # anchor url*/```.

I wasn't sure how the css is being retrieve as it doesn't seem to be in the source files. I added a specific overrid for h4 below a featured-box-minimal which removes the heigh and margin-top. This seems to have solved the issue.


sean-gilliam:
Thank you for your contribution @rubenmamo!

rubenmamo:
Welcome

Issue #2640, opened 2017-05-03
Memory leak when disposing actor system with non default ActorRefProvider


Ralf1108:
using  Akka 1.2.0

If local actor system is created and disposed repeatedly then everything is fine.
If same is done with cluster actor system then there seems to be a memory leak after disposing.

Check tests:
- IfLocalActorSystemIsStartedAndDisposedManyTimes_ThenThereShouldBeNoMemoryLeak
Output:
Got ActorIdentity: 42
After first run - MemoryUsage: 1mb
Iteration: 2 - MemoryUsage: 1mb
Got ActorIdentity: 42
Got ActorIdentity: 42
Iteration: 4 - MemoryUsage: 1mb
Got ActorIdentity: 42
Got ActorIdentity: 42
Iteration: 6 - MemoryUsage: 1mb
...
Got ActorIdentity: 42
Got ActorIdentity: 42
Iteration: 98 - MemoryUsage: 1mb
Got ActorIdentity: 42
Got ActorIdentity: 42
Iteration: 100 - MemoryUsage: 1mb
Got ActorIdentity: 42

- IfClusterActorSystemIsCreatedAndDisposedManyTimes_ThenThereShouldBeNoMemoryLeak
Output:
Got ActorIdentity: 42
After first run - MemoryUsage: 35mb
Iteration: 2 - MemoryUsage: 35mb
Got ActorIdentity: 42
Got ActorIdentity: 42
Iteration: 4 - MemoryUsage: 102mb
Got ActorIdentity: 42
Got ActorIdentity: 42
Iteration: 6 - MemoryUsage: 169mb
System.InvalidOperationException : There seems to be a memory leak!

```csharp
using System;
using Akka.Actor;
using Akka.Cluster.Tools.Client;
using Akka.Configuration;
using NUnit.Framework;

namespace StressTests
{
    [TestFixture]
    public class AkkaTests
    {
        private const string ClusterServerConfig = @"
akka {   
    actor {
        provider = ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""
    }
    remote {
        helios.tcp {
            hostname = ""127.0.0.1""
            port = 3000
        }
    }
    cluster {
        seed-nodes = [""akka.tcp://ClusterServer@127.0.0.1:3000""]
    }  
}
";

        private const string ClusterClientConfig = @"
akka {  
    actor {
        provider = ""Akka.Cluster.ClusterActorRefProvider, Akka.Cluster""   
    }
[...]

PR #3668, opened 2018-12-03
Move Build System to .NET Core 2.1, projects to .NET Standard 2.0 and .NET 4.5.2

Aaronontheweb:
This PR introduces the following changes:

1. Standardize most .NET Framework / Standard / Core versions used inside the core / contrib libraries, their unit tests, MNTR, and so forth. These versions are now set inside `common.props` and referenced via build parameters.
2. Remove `dotnet-xunit`, given that it's deprecated and won't be supported any further in the future. 
3. Replace `dotnet-xunit` with `dotnet test` inside our build system, which has future-facing support for the foreseeable future. All of our current build parameters (such as `-parallel none`) that we rely on inside XUnit are still fully supported.
4. Fix the long-standing issues with VSTest and Visual Studio locking up the solution whenever certain test suites are run - turns out this is caused by .NET 4.5.2 and possibly older versions of .NET and some of the changes introduced into more recent versions of the VSTest adapter system. Upgrading our unit tests to target .NET 4.6.1 at a minimum fixes the issue.
5. Update some of our plugins that were supporting really random versions of .NET, like .NET 4.5.1. 

All testkits now target .NET 4.5.2 - consistent with what XUnit supports.
All libraries still target .NET 4.5 and .NET Standard 1.6, same as before.
All unit tests, MNTR, perf tests, and so forth target .NET 4.6.1 and .NET Core 1.1.

I have another set of changes planned to get us running on the .NET Core 2.1.500 SDK, but I want to restrict the number of changes made at anyone time and this PR already touches a lot of different projects.


Aaronontheweb:
Based on build system output thus far:
* Windows build - ok
* MNTR .NET Core - ok
* MNTR .NET full - ok
* Unit Tests .NET Core / Windows - ok but muted test failures cause process to abort; need to add back the smarter try...catch error handling we had in FAKE
* Unit Tests .NET Full / Windows - **hung** on Akka.Cluster.Tests after 30 minutes. Same issue with VSTest adapters in Windows. Upgrading from .NET 4.5.2 was supposed to address this. Might also have to address the .NET Core SDK and XUnit versions to fix this.
* Unit Tests .NET Core / Linux - **compilation failure** due to missing F# dependencies on Akka.FSharp.
* Performance tests Windows / .NET full - ok but still racy (NBench issue)
* NuGet pack - **fails** on MNTR (still looking .NET 4.5.2 directory)

Aaronontheweb:
Going to work the problem starting with the hung .NET Framework tests, since that's been the single biggest pain in the ass for everyone who's been complaining about ReSharper issues inside Visual Studio. I need to do some analysis and see which projects it is that hangs just in-case I can't fix it by upgrading the build + test runtime we're using.

Essentially, in the span of a year all of the .NET Core 1.1 infrastructure has byte-rotted immensely and this is what we're now paying for.

Aaronontheweb:
https://github.com/Microsoft/msbuild/issues/2275

Aaronontheweb:
Ok, yeah - can confirm the hang in the VSTest adapter was caused by FsCheck. Unfortunately, upgrading to the latest (2.13.0) didn't solve the issue. But removing the FSCHECK compilation symbol from .NET 4.6.1 builds did. 

Going to do that and have us skip FsCheck specs for the time being until we can get that regression fixed.

Aaronontheweb:
I think I got all of the FsCheck instances in this case. Just need to add back the `FSCHECK` compilation symbol for .NET 4.6.1 builds to restore this functionality again.

Aaronontheweb:
Akka.Persistence specs OOM-ing on .NET 4.6.1

Aaronontheweb:
Wonder if this is caused by the `EventMemoryJournal` filling up and never leaving memory

Aaronontheweb:
OOM caused

```
Failed   Akka.Persistence.Tests.AtLeastOnceDeliveryReceiveActorSpec.PersistentReceive_must_deliver_messages_in_order_whe
n_nothing_is_lost                                                                                                       
Error Message:                                                                                                          
 System.OutOfMemoryException : Exception of type 'System.OutOfMemoryException' was thrown.                              
Stack Trace:                                                                                                            
   at System.String.Concat(String str0, String str1, String str2)                                                       
   at Akka.Configuration.Hocon.Parser.ParseObject(HoconValue owner, Boolean root, String currentPath)                   
   at Akka.Configuration.Hocon.Parser.ParseKeyContent(HoconValue value, String currentPath)                             
   at Akka.Configuration.Hocon.Parser.ParseObject(HoconValue owner, Boolean root, String currentPath)                   
   at Akka.Configuration.Hocon.Parser.ParseKeyContent(HoconValue value, String currentPath)                             
   at Akka.Configuration.Hocon.Parser.ParseObject(HoconValue owner, Boolean root, String currentPath)                   
   at Akka.Configuration.Hocon.Parser.ParseKeyContent(HoconValue value, String currentPath)                             
   at Akka.Configuration.Hocon.Parser.ParseObject(HoconValue owner, Boolean root, String currentPath)                   
   at Akka.Configuration.Hocon.Parser.ParseKeyContent(HoconValue value, String currentPath)                             
   at Akka.Configuration.Hocon.Parser.ParseObject(HoconValue owner, Boolean root, String currentPath)                   
   at Akka.Configuration.Hocon.Parser.ParseText(String text, Func`2 includeCallback)                                    
   at Akka.Configuration.ConfigurationFactory.FromResource(String resourceName, Assembly assembly)                      
   at Akka.Configuration.ConfigurationFactory.FromResource(String resourceName)                                         
   at Akka.Actor.Settings..ctor(ActorSystem system, Config config)                                                      
   at Akka.Actor.Internal.ActorSystemImpl..ctor(String name, Config config)                                             
   at Akka.TestKit.TestKitBase.InitializeTest(ActorSystem system, Config config, String actorSystemName, String testActo
rName)                                                                                                                  
   at Akka.TestKit.TestKitBase..ctor(ITestKitAssertions assertions, ActorSystem system, Config config, String actorSyste
mName, String testActorName)                                                                                            
   at Akka.TestKit.AkkaSpec..ctor(Config config, ITestOutputHelper output)                                              
   at Akka.Persistence.Tests.PersistenceSpec..ctor(Config config, ITestOutputHelper output)                             
   at Akka.Persistence.Tests.AtLeastOnceDeliveryReceiveActorSpec..ctor()                                                
```

Aaronontheweb:
I suspect this issue is related to this: https://github.com/akkadotnet/akka.net/issues/2640

Looks like it's an issue with the `ForkJoinDispatcher`.

Aaronontheweb:
Since #3734 has been merged in, re-testing this to see if the OOM exception in the `Akka.Persistence.Tests` library still persists. Tried it locally on my machine and it appears as though the issue was resolved.

Aaronontheweb:
Looks like an FSharp-related compilation failure on Linux

Aaronontheweb:
Looks like also a config issue with the way the `Akka.Streams` testkit specs are setup. Looking into that next.

Aaronontheweb:
Upgraded to the 2.1 NET Core SDK after all - resolves issue with `fsproj`s not being compiled when built from the `.sln` file on Linux.

Aaronontheweb:
Can't bump to 2.1 SDK without dealing with the usual .NET Standard 2.0 issues -- going to see if I can just hack the FSharp issues for now.

Aaronontheweb:
If this doesn't work, we'll simply drop the `.fsproj` from solution file until 1.3.12 is finished, then immediately add them back once we start the 1.4 train as soon as the 1.3.12 bits are out the door. Can't hold up progress forever working around bugs on old tools.

Aaronontheweb:
Build system now works but there are some flakey specifications at work again. Going to double check to make sure that there's nothing systematically wrong (i.e. the build system is broken) but I feel pretty good about the results thus far.

Aaronontheweb:
Looks like we have some Akka.Streams.Dsl specs failing consistently on .NET Core - going to take a look at those.

Aaronontheweb:
Was able to verify that the issue with the Streams specs hitting this error:

https://github.com/akkadotnet/akka.net/blob/41c116d2a3d280481e584401608ec848dc147632/src/core/Akka.Streams.TestKit.Tests/ScriptedTest.cs#L249

"Too many idle rounds"

Is not caused by changing Akka.Streams from .NET Standard 1.6 to 2.0. If anything, it appears to be an issue solely related to changing the test suite from .NET Core 1.1 to .NET Core 2.1. The issue doesn't occur on .NET Framework 4.6.1 at all, but it does appear on both Linux and Windows on .NET Core 2.1 but not 1.1.

I suspect it's a timing issue related to performance changes in the underlying platform; .NET Core 2.1 is significantly faster than .NET 4.6.1 and .NET Core 1.1.

I can't point to anything specific but that's the best explanation so far.

Aaronontheweb:
```
Akka.Tests.Util.ByteStringSpec.A_ByteString_must_be_sequential_when_slicing_from_index] System.TypeInitializationException : The type initializer for '<StartupCode$FsCheck>.$Arbitrary' threw an exception.
---- System.IO.FileLoadException : Could not load file or assembly 'FSharp.Core, Version=4.4.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)
-------- System.IO.FileLoadException : Could not load file or assembly 'FSharp.Core, Version=4.4.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. Strong name signature could not be verified.  The assembly may have been tampered with, or it was delay signed but not fully signed with the correct private key. (Exception from HRESULT: 0x80131045)
```

Fun with strong naming between FsCheck and FSharp.Core...

Aaronontheweb:
```
Unable to load DLL 'sqlite3': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
[19:36:34] :		 [RunTests] Cause: [akka://SqliteSnapshotStoreSpec/system/akka.persistence.snapshot-store.sqlite#893941320]: Akka.Actor.ActorInitializationException: Exception during creation ---> System.DllNotFoundException: Unable to load DLL 'sqlite3': The specified module could not be found. (Exception from HRESULT: 0x8007007E)
```

Looks like DLL issues with the SQLite drivers

Aaronontheweb:
Things are looking MUCH better in the .NET Framework test suite now. Had to drop Akka.Persistence.Sqlite specs for the time being due to a native DLL loading exception that doesn't occur on .NET Core - going to test some of our SQLite samples though and see if they still work prior to signing off on that work.

src/core/Akka.Streams.TestKit.Tests/ScriptedTest.cs
```diff
@@ -200,7 +200,7 @@ public void Request(int demand)
 
             public bool ShakeIt()
             {
-                var oneMilli = TimeSpan.FromMilliseconds(1);
+                var oneMilli = TimeSpan.FromMilliseconds(10);
```
Aaronontheweb: So the issue, near as I can tell based on the additional debug logging I added: the `ReceiveWhile` never even had a chance to run. Bails out immediately in some small cases. I can't figure out what caused this to happen between .NET Core 1.1 and 2.1 but I'm not going to wait forever to get to the bottom of it.

Aaronontheweb:
Not quite ready to merge this in yet - going to try to push out a v1.3.13 with some bug fixes first. Once this gets merged in, we're on the v1.4.0 train for good.

Aaronontheweb:
Pulling in the latest Akka.NET v1.3.13 bits before we merge this into the `dev` branch and "burn the boats" on any more 1.3.* changes.

heatonmatthew:
@Aaronontheweb I just pushed a commit that removes some of the deficiencies of the 1.x netstandard api which are now available in the 2.0 netstandard api. Serialization is a big one.

I'll take a look into it, but this might help with getting #3725 running on DotNet Core.

heatonmatthew:
@Aaronontheweb I cherry-picked your changes from #3725 on top of my two commits above and both the Framework and Core tests appear to work on windows. I'll spin it up in a Linux docker container tomorrow at work and given them a run there too, but I think having the serialization attribute be the platform one (rather than the Akka polyfil) has helped. 

This is incidentally one of the core factors that caused me to start down this netstandard2.0 path. I was doing testing with the 3.0 preview 5 and things were getting a bit confused with the significant differences to the 1.x standard api. :)

Aaronontheweb:
@heatonmatthew great! I need to work on that ASAP as that bug is having a live impact on some of our customers - would you mind sending in a pull request so I can see the changes?

Aaronontheweb:
@heatonmatthew as for those commits you want to put on top of this PR - would love to see a PR for that too. Would be very much appreciated!

heatonmatthew:
@Aaronontheweb PRs created: 

- #3790 removes the netstandard1.x restrictions
- #3791 cherry-picks your Akka.Cluster and Akka.Remote changes on top of them

Aaronontheweb:
@heatonmatthew thanks Matt! reviewed #3790 - I'll take a look at #3791 soon too. Really appreciate your help.

PR #3677, opened 2018-12-16
Fixes DistributedPubSubMediator not unsubscribing actors when they te…

### LINKED ISSUE #3658: Actor still receiving messages from mediator after termination
**Akka.Net Version:** 1.3.9
**Platform:** Windows 10 - Microsoft Service Fabric

According to the docs:

> Actors are automatically removed from the registry when they are terminated, or you can explicitly remove entries with DistributedPubSubMediator.Unsubscribe.

https://getakka.net/articles/clustering/distributed-publish-subscribe.html

But after terminating the subscriber actor, messages published to the subscribed topic keep being redirected to the subscriber's dead letters.
Demo: https://github.com/pedroyan/MediatorBug

ismaelhamed:
Fixes #3658

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/PublishSubscribe/DistributedPubSubMediatorSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/PublishSubscribe/DistributedPubSubMediatorSpec.cs
@@ -286,6 +286,15 @@ namespace Akka.Cluster.Tools.Tests.MultiNode.PublishSubscribe
             });
         }
 
+        private void AwaitCountSubscribers(int expected, string topic)
+        {
+            AwaitAssert(() =>
+            {
+                Mediator.Tell(new CountSubscribers(topic));
+                Assert.Equal(expected, ExpectMsg<int>());
+            });
+        }
+
         #endregion
 
         [MultiNodeFact]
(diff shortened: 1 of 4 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/PublishSubscribe/DistributedPubSubMediatorSpec.cs
```diff
@@ -820,5 +830,23 @@ public void DistributedPubSubMediator_must_get_topics_after_simple_publish()
                 EnterBarrier("after-get-topics");
             });
         }
+
+        public void DistributedPubSubMediator_must_remove_topic_subscribers_when_they_terminate()
```
Aaronontheweb: Looks good

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
```diff
@@ -233,6 +237,7 @@ protected override bool Business(object message)
                 var terminated = (Terminated)message;
                 var key = Utils.MakeKey(terminated.ActorRef);
                 _buffer.RecreateAndForwardMessagesIfNeeded(key, () => NewGroupActor(terminated.ActorRef.Path.Name));
+                Remove(terminated.ActorRef);
```
Aaronontheweb: Good

PR #3693, opened 2019-01-07
Add "Install-Package" instruction for Akka.Streams

warrenfalk:
The Getting Started guide lacks crucial knowledge that is necessary to actually get started, which is the nuget package.  So above the ```using Akka```s paragraph, I've added the ```Install-Package Akka.Streams``` instruction.


Aaronontheweb:
@warrenfalk thanks Warren! We'll merge this in.

PR #3695, opened 2019-01-10
Support ClusterSharding passivation in BackoffSupervisor

ismaelhamed:
> Can be used to allow cluster sharding to work with a back off on stop supervisor which is typically what you want for sharded persistent actors

[#25933](https://github.com/akka/akka/pull/25933)
[#26142](https://github.com/akka/akka/pull/26142)


ismaelhamed:
@Aaronontheweb this feature has been way overdue. The BackoffSupervisor API is starting to look a little cluttered now, but there's a proposal to simplify it, so I'll keep an eye on it.

Danthar:
Cant. I think its because we had a TC meltdown if i understood @Aaronontheweb correctly. You need to push a new commit. Or rebase to force a refresh.

Aaronontheweb:
@ismaelhamed @Danthar yeah, TeamCity hit the deck here. I'll fix that.

PR #3699, opened 2019-01-19
Add MaxRestarts to RestartWithBackoff

ismaelhamed:
Port [#24328](https://github.com/akka/akka/pull/24328)

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
@@ -1529,16 +1529,21 @@ namespace Akka.Streams.Dsl
     public class static RestartFlow
     {
         public static Akka.Streams.Dsl.Flow<TIn, TOut, Akka.NotUsed> OnFailuresWithBackoff<TIn, TOut, TMat>(System.Func<Akka.Streams.Dsl.Flow<TIn, TOut, TMat>> flowFactory, System.TimeSpan minBackoff, System.TimeSpan maxBackoff, double randomFactor) { }
+        public static Akka.Streams.Dsl.Flow<TIn, TOut, Akka.NotUsed> OnFailuresWithBackoff<TIn, TOut, TMat>(System.Func<Akka.Streams.Dsl.Flow<TIn, TOut, TMat>> flowFactory, System.TimeSpan minBackoff, System.TimeSpan maxBackoff, double randomFactor, int maxRestarts) { }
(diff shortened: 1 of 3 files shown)


src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
```diff
@@ -1528,17 +1528,21 @@ namespace Akka.Streams.Dsl
     }
     public class static RestartFlow
     {
-        public static Akka.Streams.Dsl.Flow<TIn, TOut, Akka.NotUsed> OnFailuresWithBackoff<TIn, TOut, TMat>(System.Func<Akka.Streams.Dsl.Flow<TIn, TOut, TMat>> flowFactory, System.TimeSpan minBackoff, System.TimeSpan maxBackoff, double randomFactor) { }
+        public static Akka.Streams.Dsl.Flow<TIn, TOut, Akka.NotUsed> OnFailuresWithBackoff<TIn, TOut, TMat>(System.Func<Akka.Streams.Dsl.Flow<TIn, TOut, TMat>> flowFactory, System.TimeSpan minBackoff, System.TimeSpan maxBackoff, double randomFactor, int maxRestarts) { }
         public static Akka.Streams.Dsl.Flow<TIn, TOut, Akka.NotUsed> WithBackoff<TIn, TOut, TMat>(System.Func<Akka.Streams.Dsl.Flow<TIn, TOut, TMat>> flowFactory, System.TimeSpan minBackoff, System.TimeSpan maxBackoff, double randomFactor) { }
```
ismaelhamed: The JVM doesn't have an overload. Should we just make a breaking change here in order to keep the API in sync?

ismaelhamed:
> So I found out why TeamCity hates you - the dates on your commits are so old that TeamCity won't treat them like fresh targets for builds.

Is that a feature? :)

Aaronontheweb:
> Is that a feature? :)

It just means that TeamCity is playing hard to get

src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
Aaronontheweb: @ismaelhamed IMHO, no - I think we should just add a new method. Better to not break the API if we can help it. Would you mind making that change?

ismaelhamed: Done.

PR #3701, opened 2019-01-23
Include manifest or class in missing serializer failure if possible

ismaelhamed:
Got one of this exceptions early today when switching to a very old branch. Figured out we were missing something [#26225](https://github.com/akka/akka/pull/26225)

### DIFF
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -183,7 +183,7 @@ namespace Akka.Serialization
         {
             if (!_serializersById.TryGetValue(serializerId, out var serializer))
                 throw new SerializationException(
-                    $"Cannot find serializer with id [{serializerId}]. The most probable reason" +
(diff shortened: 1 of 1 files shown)


src/core/Akka/Serialization/Serialization.cs
```diff
@@ -183,7 +183,7 @@ public object Deserialize(byte[] bytes, int serializerId, Type type)
         {
             if (!_serializersById.TryGetValue(serializerId, out var serializer))
                 throw new SerializationException(
-                    $"Cannot find serializer with id [{serializerId}]. The most probable reason" +
+                    $"Cannot find serializer with id [{serializerId}] (class [{type?.Name}]). The most probable reason" +
```
Aaronontheweb: Yes, this is a significant usability improvement. I love it.

Aaronontheweb:
thanks @ismaelhamed !

PR #3703, opened 2019-02-02
Fix concat example.

OnurGumus:


Aaronontheweb:
thanks for your contribution @OnurGumus !

Issue #3724, opened 2019-02-26
Akka: serialize-messages barfs on built-in Akka.Cluster / Akka.Remote types


Aaronontheweb:
Version: Akka.NET v1.3.11
Issue:

When `akka.actor.serialize-messages = on`, built-in Akka.Cluster and (possibly) Akka.Remote get serialized with JSON.NET (the default `object` serializer) instead of their specific serializer registrations, and thus this exception gets thrown:

```
Cause: Newtonsoft.Json.JsonSerializationException: Unable to find a constructor to use for type Akka.Cluster.InternalClusterAction+PublishChanges. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute. Path 'NewGossip', line 1, position 97.
```

Reproduction from @Dan-Albrecht: https://github.com/Dan-Albrecht/AkkaRepro

Suspect that the issue is that the underlying `serialize-messages` code isn't correctly looking up the customer serializer correctly.

Aaronontheweb:
You know what... The issue here might be that some of the internal messages we pass around, internally, inside the `Gossip` system aren't inherently serializable AND they aren't meant to be sent across the network either. In which case, we need to mark those messages with [`INoSerializationVerificationNeeded`](https://getakka.net/api/Akka.Actor.INoSerializationVerificationNeeded.html)

ctrlaltdan:
Just thought I'd contribute another serialization bug using `1.4.4`.

```
[14:17:25 ERR] Swallowing exception during message send
Newtonsoft.Json.JsonSerializationException: Unable to find a constructor to use for type Akka.Remote.Ack. A class should either have a default constructor, one constructor with arguments or a constructor marked with the JsonConstructor attribute. Path 'Ack.CumulativeAck', line 1, position 145.
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateNewObject(JsonReader reader, JsonObjectContract objectContract, JsonProperty containerMember, JsonProperty containerProperty, String id, Boolean& createdFromNonDefaultCreator)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
[...]

Issue #3414, opened 2018-04-22
Exception in PersistentShardCoordinator ReceiveRecover


joshgarnett:
Akka 1.3.5

This morning while making some provisioning changes, we ended up in a state where two single node clusters were running that pointed to the same database.  After fixing the error and starting only a single node, the underlying akka code was failing to recover.  

```
2018-04-22 15:28:45.148 ERROR PersistentShardCoordinator Exception in ReceiveRecover when replaying event type [Akka.Cluster.Sharding.PersistentShardCoordinator+ShardHomeAllocated] with sequence number [40650] for persistenceId [/system/sharding/zCoordinator/singleton/coordinator]
System.ArgumentException: Region [akka://AkkaCluster/system/sharding/z#673731278] not registered
Parameter name: e
   at Akka.Cluster.Sharding.PersistentShardCoordinator.State.Updated(IDomainEvent e)
   at Akka.Cluster.Sharding.PersistentShardCoordinator.ReceiveRecover(Object message)
   at Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message)
   at Akka.Persistence.Eventsourced.<>c__DisplayClass92_0.<Recovering>b__0(Receive receive, Object message)

2018-04-22 15:28:45.438 ERROR PersistentShardCoordinator Exception in ReceiveRecover when replaying event type [Akka.Cluster.Sharding.PersistentShardCoordinator+ShardHomeAllocated] with sequence number [50633] for persistenceId [/system/sharding/oCoordinator/singleton/coordinator]
System.ArgumentException: Shard 78 is already allocated
Parameter name: e
   at Akka.Cluster.Sharding.PersistentShardCoordinator.State.Updated(IDomainEvent e)
   at Akka.Cluster.Sharding.PersistentShardCoordinator.ReceiveRecover(Object message)
   at Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message)
   at Akka.Persistence.Eventsourced.<>c__DisplayClass92_0.<Recovering>b__0(Receive receive, Object message

2018-04-22 15:29:07.714 ERROR PersistentShardCoordinator Exception in ReceiveRecover when replaying event type [Akka.Cluster.Sharding.PersistentShardCoordinator+ShardHomeAllocated] with sequence number [11774] for persistenceId [/system/sharding/pCoordinator/singleton/coordinator]
System.ArgumentException: Region [akka://AkkaCluster/system/sharding/p#1560991559] not registered
[...]

PR #3725, opened 2019-02-26
Fix serialize-messages for Akka.Cluster and Akka.Remote

Aaronontheweb:
Working on fixing #3724


Aaronontheweb:
Reproduction spec failed on both Windows and Linux, as is expected.

src/core/Akka.Cluster.Tests/Serialization/BugFix3724Spec.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            _cluster.Subscribe(TestActor, ClusterEvent.SubscriptionInitialStateMode.InitialStateAsEvents,
+                typeof(ClusterEvent.MemberUp));
+            Within(TimeSpan.FromSeconds(10), () =>
+            {
+                EventFilter.Exception<Exception>().Expect(0, () =>
+                {
+                    // wait for a singleton cluster to fully form and publish a member up event
+                    _cluster.Join(_selfAddress);
+                    var up = ExpectMsg<ClusterEvent.MemberUp>();
```
Aaronontheweb: The test case here is simple: test that end-to-end, the Akka.Cluster self-join process works and all manner of internal event publication goes off without a hitch. Caught two different major serialization errors with this and fixed both.

Aaronontheweb: If the `MemberUp` event makes it all the way to this actor, we can safely assume that the messages necessary to run Akka.Cluster's internals made it intact.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -754,41 +754,46 @@ public SendCurrentClusterState(IActorRef receiver)
...
+        /// INTERNAL API.
+        ///
+        /// Marker interface for publication events from Akka.Cluster.
         /// </summary>
-        interface IPublishMessage { }
+        /// <remarks>
+        /// <see cref="INoSerializationVerificationNeeded"/> is not explicitly used on the JVM,
+        /// but without it we run into serialization issues via https://github.com/akkadotnet/akka.net/issues/3724
+        /// </remarks>
+        private interface IPublishMessage : INoSerializationVerificationNeeded { }
```
Aaronontheweb: First fix: made it so `IPublishMessage` doesn't get checked by `akka.actor.serialize-messages` - did this because designating a single `Gossip` class constructor with `JsonConstructor` wasn't going to work due to the fact that we make use of all three different constructors in different situations. The JVM gets away with this by using the built-in `Serializable` trait in Scala, which is roughly equivalent to our binary formatter, which .NET no longer has in .NET Core. So, I figured this would be the best substitute here.

Aaronontheweb: Relevant link on Scala behavior: https://alvinalexander.com/scala/how-to-use-serialization-in-scala-serializable-trait

src/core/Akka.Cluster/Member.cs
```diff
@@ -92,6 +93,20 @@ internal Member(UniqueAddress uniqueAddress, int upNumber, MemberStatus status,
...
 
+        /// <summary>
+        /// Used when `akka.actor.serialize-messages = on`.
+        /// </summary>
+        /// <param name="uniqueAddress">The address of the member.</param>
+        /// <param name="upNumber">The upNumber of the member, as assigned by the leader at the time the node joined the cluster.</param>
+        /// <param name="status">The status of this member.</param>
+        /// <param name="roles">The roles for this member. Can be empty.</param>
+        [JsonConstructor]
+        internal Member(UniqueAddress uniqueAddress, int upNumber, MemberStatus status, IEnumerable<string> roles)
```
Aaronontheweb: The issue with serializing the `Member` class is an old, but familiar one: JSON.NET is unable to serialize the `System.Collections.Immutable.ImmutableHashSet<string>` taken for the `roles` parameter on the normal constructor this actor uses. Thus, I created a second `internal` constructor that just takes an `IEnumerable<string>` instead and marked that as the `JsonConstructor` - ideally once Hyperion is done we will no longer need this.

Aaronontheweb:
Looks like there's still a hard failure on Linux with this.

Aaronontheweb:
Failed on Windows too - passes on .NET Framework locally for me, but looks like it doesn't for .NET Core.

Dan-Albrecht:
I pulled this and it seems to fix my problems in my sample and our actual real code.

Aaronontheweb:
@Dan-Albrecht my guess is that it's an issue related to the old version of .NET Core we're running on (1.1.) I haven't had a chance to test that just yet as I've been neck-deep in working on another issue for most of this afternoon (#3414) but I have a plan for getting all of these projects moved onto .NET Core 2.2 pronto as part of an upcoming sprint.

Glad to hear the fix is working though - I'll merge this in as part of 1.3.12, _somehow_ :p

PR #3709, opened 2019-02-07
Fix actor dependency injector link

jdsartori:
Original link target was ambiguous.

I did some research and figured the author meant this: AkkaNetContrib/Akka.DI.SimpleInjector

PR #3727, opened 2019-02-27
Update dependency-injection.md

jdsartori:
As far as I can tell, this is a better reference for that link.
I could find no other related document in the Akka articles.


sean-gilliam:
As I mentioned in #3709, linking to SimpleInjector is incorrect. Someone needs to port over https://github.com/akkadotnet/getakka.net/blob/master/src/docs/DI%20Core.md for the link to be correct.

jdsartori:
D'oh! Sorry about that. I'll close this PR.

PR #3734, opened 2019-03-11
Fix ForkJoinExecutor / Dispatcher Leak

Aaronontheweb:
Have a reproduction spec for #2640 in here, but not a fix yet.

### DIFF
--- a/src/core/Akka.Tests/Actor/Dispatch/Bug2640Spec.cs
+++ b/src/core/Akka.Tests/Actor/Dispatch/Bug2640Spec.cs
@@ -1,8 +1,12 @@
-﻿using System;
+﻿// //-----------------------------------------------------------------------
+// // <copyright file="Bug2640Spec.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2019 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2019 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
 using System.Collections.Generic;
-using System.Collections.Immutable;
 using System.Linq;
-using FluentAssertions;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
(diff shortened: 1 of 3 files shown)


src/core/Akka.Tests/Actor/Dispatch/Bug2640Spec.cs
```diff
@@ -0,0 +1,112 @@
...
+            Watch(actor);
+            for (var i = 0; i < 100; i++)
+                actor.Tell(GetThread.Instance);
+
+            threads = ReceiveN(100).Cast<Thread>().GroupBy(x => x.ManagedThreadId)
+                .ToDictionary(x => x.Key, grouping => grouping.First());
+            threads.Count.Should().Be(4, "Expected 4 distinct threads in this example");
+
+            await Sys.Terminate();
+            AwaitAssert(() => threads.Values.All(x => x.IsAlive == false).Should().BeTrue("All threads should be stopped"));
```
Aaronontheweb: This assertion currently fails - the `Task` from `Terminate()` completes without killing off any of the `ForkJoinExecutor` threads.

Aaronontheweb:
Looks like the real fix to this bug is completing all outstanding work items inside the `IScheduler` upon shutdown in order to guarantee that dispatchers are terminated.

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
```diff
@@ -387,7 +387,28 @@ public void Dispose()
                 return;
             }
 
-            // todo: log unprocessed scheduler registrations?
+            // Execute all outstanding work items
+            foreach (var task in stopped.Result)
+            {
+                try
+                {
+                    task.Action.Run();
```
Aaronontheweb: So this was the major fix - the JVM has equivalent code too; need to make sure everything gets processed upon `IScheduler` shutdown.

src/core/Akka.Tests/Actor/Dispatch/Bug2640Spec.cs
```diff
@@ -0,0 +1,123 @@
...
+            Watch(actor);
+            for (var i = 0; i < 100; i++)
+                actor.Tell(GetThread.Instance);
+
+            threads = ReceiveN(100).Cast<Thread>().GroupBy(x => x.ManagedThreadId)
+                .ToDictionary(x => x.Key, grouping => grouping.First());
+            threads.Count.Should().Be(4, "Expected 4 distinct threads in this example");
+
+            await Sys.Terminate();
+            AwaitAssert(() =>
```
izavala: Just to make sure I'm completely clear on this, before the fix was implemented this test would have failed due to threads not correctly shutting down?

Aaronontheweb: I wasn't able to get this to fail on the build server because it's a racy specification - the memory leak only occurs if the host system is able to terminate the `ActorSystem` faster than 1 second (the default timeout parameter for shutting down dispatchers via the `IScheduler`,) which happened reliably (100% of the time) on my local system during testing. 

However, given that our build servers are running on relatively low-powered Windows Azure VMs, it's possible that the `IScheduler` can win the race. But all we needed to prove that this was the issue was for this test to fail once, which it did.

PR #3741, opened 2019-03-20
Added stages responsible for streams context propagation

Horusiath:
Port of feature from: https://github.com/akka/akka/pull/25951

This allows to implicitly pass a some object as an implicit context of execution between stages, without exposing it on the majority of the stages. IMHO the major use cases are all kinds of persistent queues, which require explicit commit, after message was processed throughout the stages. In that case we'd like to pass message context/id (necessary for commit), but without seeing it in our business domain stages, that have no interest with it.

Example:

```csharp
AmazonSQS.Source()
   .Select(msg => (msg.MessageRef, msg.Body))
   .AsSourceWithContext(tuple => tuple.Item1))
       .Select(str => JsonConvert.DeserializeObject<MyMessage>(str))
       // do some other processing 
   .AsSource()
   .SelectAsync(4, tuple => AmazonSQS.CommitAsync(tuple.Item2))
   .To(Sink.Ignore<Tuple<MyMessage, string>>())
   .Run(materializer);
```


Aaronontheweb:
I've very excited about this feature, btw

Horusiath:
I think naming needs to be updated - some time after I've ported this feature Lightbend decided to change names in the API.

Aaronontheweb:
Ok, want to do that now then?

Horusiath:
I'll try, it should be just a naming convention change + maybe few new stages added there.

Aaronontheweb:
@Horusiath sounds good, we'll get that out in the next beta.

Aaronontheweb:
@Horusiath I'm going to work on the API renaming in a new PR this week - along with updating the APIs for the `StreamRefs`. I'll address those separately shortly.

PR #3744, opened 2019-03-22
Fix sharding recovery error and WithTransport serialization

### LINKED ISSUE #3414: Exception in PersistentShardCoordinator ReceiveRecover

Aaronontheweb:
Should resolve #3414 and related issues.


Aaronontheweb:
Got massively interrupted (airport) while I was working on this, so it's not in a compile-able state. Will prioritize this ASAP though. Think there's some possibilities for Akka.Remote performance improvements too.

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Batching/BatchingSqliteJournalSpec.cs
```diff
@@ -42,5 +42,8 @@ class = ""Akka.Persistence.Sqlite.Journal.BatchingSqliteJournal, Akka.Persistenc
                     }
                 }");
         }
+
+        // TODO: hack. Replace when https://github.com/akkadotnet/akka.net/issues/3811
+        protected override bool SupportsSerialization => false;
```
Aaronontheweb: Another reason for https://github.com/akkadotnet/akka.net/issues/3811 - none of the Akka.Persistence.Sql journals support `WriterGuid` at the moment due to their hand-rolled serialization layers, so the serialization specs will not pass.

Aaronontheweb:
This PR is finished - complete port of https://github.com/akka/akka/pull/25068 sans the Artery bits.

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -120,6 +179,47 @@ public Serialization(ExtendedActorSystem system)
...
+        {
+            var info = system.Provider.SerializationInformation;
+            if (CurrentTransportInformation == info)
+            {
+                // already set
+                return action();
+            }
+            else
+            {
+                CurrentTransportInformation = info;
```
Horusiath: Previous transport information is never restored after calling this method - is that how it's supposed to work?

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1205,29 +1203,35 @@ protected virtual void WriteEvent(TCommand command, IPersistentRepresentation pe
             var payloadType = persistent.Payload.GetType();
             var serializer = _serialization.FindSerializerForType(payloadType, Setup.DefaultSerializer);
 
-            string manifest = "";
-            if (serializer is SerializerWithStringManifest)
+            // TODO: hack. Replace when https://github.com/akkadotnet/akka.net/issues/3811
+            Akka.Serialization.Serialization.WithTransport(_serialization.System, () =>
```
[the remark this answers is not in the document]
Aaronontheweb: We don't pass it in because other libraries may not directly have access to when it's set via Akka.Remote - rather it gets parked ambiently in the local thread's storage and then unset when it's done being used. Won't be an issue for multiple actor systems due to the latter - the reference always gets unset and not leaked from the thread. There's more comments on the JVM issue that I ported this from: https://github.com/akka/akka/pull/25068

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
```diff
@@ -57,26 +57,42 @@ private PersistentMessage GetPersistentMessage(IPersistentRepresentation persist
...
                 {
-                    payload.PayloadManifest = ByteString.CopyFromUtf8(obj.GetType().TypeQualifiedName());
+                    var manifest = serializer2.Manifest(obj);
+                    payload.PayloadManifest = ByteString.CopyFromUtf8(manifest);
+                }
+                else
+                {
+                    if (serializer.IncludeManifest)
+                    {
+                        payload.PayloadManifest = ByteString.CopyFromUtf8(obj.GetType().TypeQualifiedName());
```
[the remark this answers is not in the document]
Aaronontheweb: This is the same method we called before - I just moved it into a different place.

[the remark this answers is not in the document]
Aaronontheweb: I.e. I kept it consistent with its previous behavior as of 1.3.13

src/core/Akka/Serialization/Serialization.cs
Aaronontheweb: Yeah that might be a mistake - let me double check

src/core/Akka.Persistence/Snapshot/LocalSnapshotStore.cs
```diff
@@ -199,8 +199,8 @@ private Serialization.Snapshot Deserialize(Stream stream)
             var buffer = new byte[stream.Length];
             stream.Read(buffer, 0, buffer.Length);
             var snapshotType = typeof(Serialization.Snapshot);
-            var serializer = _serialization.FindSerializerForType(snapshotType, _defaultSerializer);
-            var snapshot = (Serialization.Snapshot)serializer.FromBinary(buffer, snapshotType);
+            var serializerId = _serialization.FindSerializerForType(snapshotType).Identifier;
+            var snapshot = (Serialization.Snapshot)_serialization.Deserialize(buffer, serializerId, snapshotType);
```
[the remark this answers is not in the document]
Aaronontheweb: Done, going to be pushing this up in the next set of commits

src/core/Akka/Serialization/Serialization.cs
Aaronontheweb: So I think this is some weird scala syntax:

```scala
val info = system.provider.serializationInformation
    if (Serialization.currentTransportInformation.value eq info)
      f() // already set
    else
      Serialization.currentTransportInformation.withValue(info) {
        f()
      }
  }
```

I'm going to interpret the scoping operating there like a `using` block - the `currentTransportInformation` only has that value for the duration of that call. Therefore, this is probably a bug in the way I translated the code. I'll change it so it unsets the `CurrentTransportInformation` at the end. Good catch.

Aaronontheweb: More explanation on `DynamicVariable` in Scala: https://stackoverflow.com/questions/5116352/when-we-should-use-scala-util-dynamicvariable

src/core/Akka.Persistence/Snapshot/LocalSnapshotStore.cs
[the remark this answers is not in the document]
Aaronontheweb: This was fixed

src/core/Akka/Serialization/Serialization.cs
Aaronontheweb: This has been resolved.

PR #3754, opened 2019-04-07
Ensure that BatchingSqlJournal will propagate sql connection opening failure

### LINKED ISSUE #3753: RestartSource + PersistenceQuery don't behave like expected
Up until now, I have taken for granted that the `RestartSource` would work with a `PersistenceQuery` just like it does with any other type of stream. But after digging a little deeper this doesn't seem to be the case.

Given some code like this, I'd expect that any exceptions in the `PersistenceQuery` (including transient disconnections from the database) would result in the stream being restarted automatically:

```csharp
var mat = ActorMaterializer.Create(system);
var backoffSource = RestartSource.WithBackoff(() =>
    {
        Console.WriteLine("Starting stream...");

        return PersistenceQuery.Get(system)
            .ReadJournalFor<SqlReadJournal>(SqlReadJournal.Identifier)
            .EventsByPersistenceId("a-0", 0L, long.MaxValue);
    },
    minBackoff: TimeSpan.FromSeconds(3),
    maxBackoff: TimeSpan.FromSeconds(30),
    randomFactor: 0.2);

backoffSource.RunWith(Sink.Ignore<EventEnvelope>(), mat);
```

What I've noticed however is this:

...when using the `BatchingSqlJournal`:

- If I drop the database connection while the stream is running, I can see in the logs that the `ActorPublisherSource` stops requesting replays for this persistenceId, so the stream comes to a halt (because of backpressure from the downstream journal?). Anyways, I'd expect the `RestartSource` to restart the stream at this point.

...when using the `SqlJournal`:

- If I drop the database connection while the stream is running, the `RestartSource` does restart the stream in this case, but only up until the circuit breaker from the journal kicks in. At that point the circuit breaker fails fast and the stream comes to a halt just like with the `BatchingSqlJournal`.

I also wanted to see how the `RestartSource` will behave with a different type of exceptions coming from the journal, so I intentionally inserted some events serialized with a different serializer than configured to make the journal throw during the replay of the events. But in this case the `RestartSource` worked as expected with both journals, restarting the stream over and over again. So, it may seems that due to how the journals are implemented internally some exceptions are prevented from bubbling up.

Horusiath:
Should fix #3753.

I think, that the issue lies in combination of [ExecuteChunk](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs#L850) which doesn't have guard around `connection.OpenAsync` (it was sending failures to subscribers only when the SQL statements themselves were failing) and [ExecuteChunk method call](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs#L840) (which send failure as `Status.Failure` to `Self` but was never handled).

This PR solves this case by adding custom `ChunkExecutionFailure` message that is used to catch those cases and inform awaiting requesters about failure.


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -837,7 +875,8 @@ private void TryProcess()
 
                 var chunk = DequeueChunk(_remainingOperations);
                 var context = Context;
-                _circuitBreaker.WithCircuitBreaker(() => ExecuteChunk(chunk, context)).PipeTo(Self);
+                _circuitBreaker.WithCircuitBreaker(() => ExecuteChunk(chunk, context))
```
ismaelhamed: Halfway there, but still when the circuit breaker throws an `OpenCircuitException` the restarts won't happen again. Wrapping this piece in a try...catch... and self issuing a `new ChunkExecutionFailure(ex, chunk.Requests)` seems to do the trick.

Horusiath: Are you saying that `OpenCircuitException` is not propagated to `PipeTo`?

Aaronontheweb: The `PipeTo` will work

Aaronontheweb: Unless @ismaelhamed means that the `_circuitBreaker` itself throws instead of injecting the exception into its `Task`, in which case his feedback is correct.

ismaelhamed: Seems like the `_circuitbreaker` is throwing instead of propagating to `PipeTo`, hence why I had to wrap in a try...catch... that call to `_circuitBreaker.WithCircuitBreaker()` in order for it to work. 

I´ll do some more tests tomorrow to confirm it.

Aaronontheweb: > Seems like the _circuitbreaker is throwing instead of propagating to PipeTo, hence why I had to wrap in a try...catch... that call to _circuitBreaker.WithCircuitBreaker() in order for it to work.

Not to get side-tracked, but if that's the case then isn't that a bug with the way the `CircuitBreaker` is designed?

Horusiath: I've updated `CircuitBreaker` implementation to raise exceptions withing returned `Task`s (simply by marking corresponding methods as async).

src/core/Akka/Pattern/CircuitBreakerState.cs
```diff
@@ -106,13 +106,13 @@ public HalfOpen(CircuitBreaker breaker)
...
         /// <returns><see cref="Task"/> containing result of protected call</returns>
-        public override Task<T> Invoke<T>(Func<Task<T>> body)
+        public override async Task<T> Invoke<T>(Func<Task<T>> body)
         {
             if (!_lock.CompareAndSet(true, false))
             {
                 throw new OpenCircuitException();
             }
-            return CallThrough(body);
+            return await CallThrough(body);
```
Aaronontheweb: LGTM - we probably should have done this all along

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Batching/BatchingSqliteJournalConnectionFailureSpec.cs
```diff
@@ -0,0 +1,40 @@
...
+// </copyright>
+//-----------------------------------------------------------------------
+
+using Akka.Configuration;
+using Akka.Persistence.Sql.TestKit;
+using Xunit.Abstractions;
+
+namespace Akka.Persistence.Sqlite.Tests.Batching
+{
+    public class BatchingSqliteJournalConnectionFailureSpec : SqlJournalConnectionFailureSpec
```
Aaronontheweb: LGTM

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -658,10 +670,36 @@ protected sealed override bool Receive(object message)
...
         }
 
+        private void FailChunkExecution(ChunkExecutionFailure message)
+        {
+            var cause = message.Cause;
+            Log.Error(cause, "Failed to execute chunk for {0} requests", message.Requests.Length);
+
+            foreach (var req in message.Requests)
+            {
+                switch (req)
```
Aaronontheweb: LGTM

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -351,6 +351,18 @@ public abstract class BatchingSqlJournal<TConnection, TCommand> : WriteJournalBa
         where TCommand : DbCommand
     {
         #region internal classes
+        
+        private sealed class ChunkExecutionFailure : IDeadLetterSuppression
```
Aaronontheweb: LGTM

ismaelhamed:
@Horusiath thank, working like a charm now.

PR #3763, opened 2019-04-24
OnPersistRejected now logs an error with the complete stacktrace

ismaelhamed:
As an user I want the complete stacktrace when persistence of an event is rejected. Plus IMO any issue wih persistence is a serious enough to log an error instead. I checked, and the JVM is logging an error too.


Aaronontheweb:
Nice work @ismaelhamed

Issue #2026, opened 2016-06-01
Akka.Persistence journal fails to start in tests


kantora:
I've found the strange problem. In tests if I set `akka.test.calling-thread-dispatcher` as default dispatcher the default `inmem` journal (along with any other) will fail to start.

Even if I use the default dispatcher but will send to journal some custom message for example like this:
`Persistence.Instance.CreateExtension((ExtendedActorSystem)this.Sys).JournalFor(null).Tell("hello")` as the first test instruction it would fail to start.

I see this in logs:

```
2016-06-01 11:25:41.963 +03:00 [Debug] "Started (Akka.TestKit.Internal.InternalTestActor)"
2016-06-01 11:25:42.017 +03:00 [Debug] "now supervising akka://test/system/akka.persistence.journal.inmem"
2016-06-01 11:25:42.036 +03:00 [Error] "Object reference not set to an instance of an object."
[akka://test/system/akka.persistence.journal.inmem]: Akka.Actor.ActorInitializationException: Exception during creation ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Akka.Actor.Props.NewActor()
   at Akka.Actor.ActorCell.CreateNewActorInstance()
   at Akka.Actor.ActorCell.<>c__DisplayClass113_0.<NewActor>b__0()
   at Akka.Actor.ActorCell.UseThreadContext(Action action)
   at Akka.Actor.ActorCell.NewActor()
   at Akka.Actor.ActorCell.Create(Exception failure)
   --- End of inner exception stack trace ---
   at Akka.Actor.ActorCell.Create(Exception failure)
   at Akka.Actor.ActorCell.SystemInvoke(Envelope envelope)
```

gengle:
Is there more context on this bug?  I have the same issue with testkit 1.2 & akka.persistence 1.2.0.36-beta

Here's a failing test.

```
    [TestFixture]
    public class PersistenceTestFixture : TestKit
    {
        [Test]
        public void inmem_should_not_crash()
        {
            var sut = Sys.ActorOf(Props.Create<FooActor>(), "foo1");

            sut.Tell("hello, why am I failing?");

            Assert.IsNotNull(sut);
            // [ERROR][6/9/2017 10:35:28 PM][Thread 0011][akka://test/system/akka.persistence.journal.inmem] Object reference not set to an instance of an object.
[...]

PR #3770, opened 2019-04-30
Ensure Arguments are never null

basbossinkdivverence:
While running some unit tests using a PersistentActor a `NullReferenceException` was encountered. The culprit seems to be that when `producer.Produce()` throws an exception, the creation of the `TypeLoadException` throws a `NullReferenceException` when `Arguments` is `null`. The relevant excerpt from `Props` lines 568-577

```csharp
        public virtual ActorBase NewActor()
        {
            var type = Type;
            var arguments = Arguments;
            try {
                return producer.Produce();
            } catch (Exception e) {
                throw new TypeLoadException($"Error while creating actor instance of type {type} with {arguments.Length} args: ({StringFormat.SafeJoin(",", arguments)})", e);
            }
        }
```

This fix ensures that no matter how the  `Props`  is constructed, `Arguments` will not be `null` fixing the above issue. This problem might be related to issue #2026.

### DIFF
--- a/src/core/Akka/Actor/Props.cs
+++ b/src/core/Akka/Actor/Props.cs
@@ -293,7 +293,7 @@ namespace Akka.Actor
         {
             Deploy = deploy;
             inputType = type;
-            Arguments = args;
+            Arguments = args ?? noArgs;
             producer = CreateProducer(inputType, Arguments);
         }
 


Aaronontheweb:
Test failure was unrelated - re-running it now.

PR #3778, opened 2019-05-04
Timers for self scheduled messages added, FSM timers fixes

zbynek001:
migrated from https://github.com/akka/akka/pull/22873

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -1985,6 +1985,7 @@ namespace Akka.Actor.Scheduler
         void CancelAll();
         bool IsTimerActive(string key);
         void StartPeriodicTimer(string key, object msg, System.TimeSpan interval);
+        void StartPeriodicTimer(string key, object msg, System.TimeSpan initialDelay, System.TimeSpan interval);
         void StartSingleTimer(string key, object msg, System.TimeSpan timeout);
     }
 }
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -47,6 +47,7 @@ namespace Akka.Actor
         protected virtual Akka.Actor.SupervisorStrategy SupervisorStrategy() { }
         protected void UnbecomeStacked() { }
         protected virtual void Unhandled(object message) { }
+        protected void WithTimers() { }
     }
     [System.Diagnostics.DebuggerDisplayAttribute("{Self,nq}")]
     public class ActorCell : Akka.Actor.IActorContext, Akka.Actor.IActorRefFactory, Akka.Actor.ICanWatch, Akka.Actor.ICell, Akka.Actor.IUntypedActorContext
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -1981,12 +1981,12 @@ namespace Akka.Actor.Scheduler
 {
     public interface ITimerScheduler
     {
-        void Cancel(string key);
+        void Cancel(object key);
         void CancelAll();
-        bool IsTimerActive(string key);
-        void StartPeriodicTimer(string key, object msg, System.TimeSpan interval);
-        void StartPeriodicTimer(string key, object msg, System.TimeSpan initialDelay, System.TimeSpan interval);
-        void StartSingleTimer(string key, object msg, System.TimeSpan timeout);
+        bool IsTimerActive(object key);
+        void StartPeriodicTimer(object key, object msg, System.TimeSpan interval);
+        void StartPeriodicTimer(object key, object msg, System.TimeSpan initialDelay, System.TimeSpan interval);
+        void StartSingleTimer(object key, object msg, System.TimeSpan timeout);
     }
 }
 namespace Akka.Annotations
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -30,6 +30,7 @@ namespace Akka.Actor
         protected static Akka.Actor.Receive EmptyReceive { get; }
         protected Akka.Actor.IActorRef Self { get; }
         protected Akka.Actor.IActorRef Sender { get; }
+        public Akka.Actor.Scheduler.ITimerScheduler Timers { get; }
         public virtual void AroundPostRestart(System.Exception cause, object message) { }
         public virtual void AroundPostStop() { }
         public virtual void AroundPreRestart(System.Exception cause, object message) { }
@@ -78,7 +79,7 @@ namespace Akka.Actor
         protected void AddWatcher(Akka.Actor.IActorRef watchee, Akka.Actor.IActorRef watcher) { }
         protected void AddressTerminated(Akka.Actor.Address address) { }
         public virtual Akka.Actor.IActorRef AttachChild(Akka.Actor.Props props, bool isSystemService, string name = null) { }
-        protected virtual void AutoReceiveMessage(Akka.Actor.Envelope envelope) { }
+        protected internal void AutoReceiveMessage(Akka.Actor.Envelope envelope) { }
         public void Become(Akka.Actor.Receive receive) { }
         public void BecomeStacked(Akka.Actor.Receive receive) { }
         public void CheckReceiveTimeout(bool reschedule = True) { }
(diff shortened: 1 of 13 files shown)


ismaelhamed:
@zbynek001 Nice!! I ported just the fix a while back (#3313) but the rest of  that PR has been in my backlog since then.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2049,6 +2050,18 @@ namespace Akka.Actor.Internal
         public UnboundedStashImpl(Akka.Actor.IActorContext context) { }
     }
 }
+namespace Akka.Actor.Scheduler
+{
+    public interface ITimerScheduler
+    {
+        void Cancel(string key);
+        void CancelAll();
+        bool IsTimerActive(string key);
```
Horusiath: Original impleentation uses `object` keys. Any particular reason to use strings?

src/core/Akka/Actor/ActorBase.cs
```diff
@@ -147,6 +147,17 @@ protected static IActorContext Context
             }
         }
 
+        private Scheduler.TimerScheduler timers;
+        public Scheduler.ITimerScheduler Timers
+        {
+            get
+            {
+                if (timers == null)
+                    timers = new Scheduler.TimerScheduler(Context);
```
Horusiath: This can be also written as `get => timers ?? (timers = new Scheduler.TimerScheduler(context))`

src/core/Akka/Actor/ActorBase.cs
```diff
@@ -155,6 +166,31 @@ protected static IActorContext Context
         /// <returns>TBD</returns>
         protected internal virtual bool AroundReceive(Receive receive, object message)
         {
+            if (message is Scheduler.TimerScheduler.ITimerMsg tm)
```
Horusiath: This is hot path of message execution. I'd check if this somehow affects actor benchmarks.

src/core/Akka.Tests/Actor/TimerSpec.cs
```diff
@@ -0,0 +1,563 @@
...
+            dilatedInterval = Dilated(TimeSpan.FromSeconds(interval));
+        }
+
+        [Fact]
+        public void Must_schedule_non_repeated_ticks()
+        {
+            var probe = CreateTestProbe();
+            var _ref = this.Sys.ActorOf(TargetProps(probe.Ref, TimeSpan.FromMilliseconds(10), false));
+
+            probe.ExpectMsg<Tock>(m => m.N == 1);
```
Horusiath: Most of the cases (and also my preference) is to add structural equality of messages (like `Tock` in this case) and compare directly to expected message instance. But this is only small note, not any real issue.

src/core/Akka.Tests/Actor/TimerSpec.cs
```diff
@@ -0,0 +1,563 @@
...
+
+            _ref.Tell(End.Instance);
+            probe.ExpectMsg<GotPostStop>(m => m.TimerActive == false);
+        }
+
+        [Fact]
+        public void Must_schedule_repeated_ticks()
+        {
+            var probe = CreateTestProbe();
+            var _ref = this.Sys.ActorOf(TargetProps(probe.Ref, dilatedInterval, true));
```
Horusiath: In general, try to avoid underscoring variables. A lot of code uses R# naming convention, and _-prefixed names are usually present in context of fields.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
zbynek001: changed to `object`

src/core/Akka/Actor/ActorBase.cs
zbynek001: changed

src/core/Akka.Tests/Actor/TimerSpec.cs
zbynek001: changed

src/core/Akka.Tests/Actor/TimerSpec.cs
zbynek001: renamed to `actor`, wanted to keep the name similar to scala version

src/core/Akka/Actor/ActorBase.cs
zbynek001: not sure if there is a better place for this to work in general.
Other option is do this only in specific actors where you want to use `Timers`, but this would be easy to forget

zbynek001: What about this approach?
To use timers, you'll have to enable them in actor constructor via this method:
https://github.com/akkadotnet/akka.net/blob/16932f4561d46c75d68908b8b9bc97880daacb8f/src/core/Akka/Actor/ActorBase.cs#L162

and then we can avoid type check in actors not using timers, and only do a null check first:
https://github.com/akkadotnet/akka.net/blob/16932f4561d46c75d68908b8b9bc97880daacb8f/src/core/Akka/Actor/ActorBase.cs#L175-L177

ismaelhamed:
@zbynek001 @Aaronontheweb any news on this one?

src/core/Akka/Actor/ActorBase.cs
Aaronontheweb: Why not use an interface similar to Scala's original trait?

```
if(this is IWithTimers wt){
     WithTimers();
}
```

Having users manually have to call that method is going to be a pain point - decorating the actor with an interface like how they do with stash is a good compromise. Signals the intent that they want to use this feature via the type system rather than an internal method call.

Aaronontheweb:
@zbynek001 updated the review - I think this needs some minor API changes and some documentation, but otherwise looks like a great improvement.

zbynek001:
@Aaronontheweb Should it be done via the Actor producer pipeline or directly in `ActorCell.CreateNewActorInstance()`?
Or some other way, when the Actor producer pipeline is marked as obsolete?

Aaronontheweb:
@zbynek001 I was going to suggest that, but ultimately deleted it. I think that's probably a better way to go though - have it be done via the way actors are created. I think that'd be a good compromise for now. We're going to have to replace that pipeline with something else eventually and I get the feeling that default interfaces probably aren't going to be it.

src/core/Akka/Actor/FSM.cs
```diff
@@ -304,7 +304,7 @@ internal class Timer : INoSerializationVerificationNeeded
             /// <param name="generation">TBD</param>
             /// <param name="owner">TBD</param>
             /// <param name="context">TBD</param>
-            public Timer(string name, object message, bool repeat, int generation, IActorRef owner, IActorContext context)
+            public Timer(string name, object message, bool repeat, int generation, ActorBase owner, IActorContext context)
```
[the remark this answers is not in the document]
zbynek001: i think it's because IActorRef which was used (Self) was the same after actor restart

zbynek001:
at the end had to initialize Timers in `ActorBase` constructor. Because often you want to configure them in your constructor

zbynek001:
this should be ready. I guess the failed test is unrelated to this pr and at least locally it's all passing

src/core/Akka/Actor/ActorBase.cs
```diff
@@ -103,6 +103,10 @@ protected ActorBase()
         {
             if (ActorCell.Current == null)
                 throw new ActorInitializationException("Do not create actors using 'new', always create them using an ActorContext/System");
+
+            if (this is IWithTimers withTimers)
+                withTimers.Timers = new Scheduler.TimerScheduler(Context);
```
zbynek001: timers initialization needs to be here support scenarios, when times are used in actor constructor

src/core/Akka/Actor/ActorBase.cs
zbynek001: updated, now actors which want to use timers need to implement `IWithTimers` interface

src/core/Akka/Actor/ActorBase.cs
```diff
@@ -155,6 +159,38 @@ protected static IActorContext Context
         /// <returns>TBD</returns>
         protected internal virtual bool AroundReceive(Receive receive, object message)
         {
+            if (message is Scheduler.TimerScheduler.ITimerMsg tm)
+            {
+                if (this is IWithTimers withTimers && withTimers.Timers is Scheduler.TimerScheduler timers)
```
zbynek001: now with the `IWithTimers` we end up here with two more type checks, but only for timer messages

src/core/Akka/Actor/ActorBase.cs
Aaronontheweb: Thanks, looks great

Aaronontheweb:
We're going to need to document these changes - for the time being I've created an open issue here: https://github.com/akkadotnet/akka.net/issues/4357

PR #3779, opened 2019-05-05
BatchingSqlJournal now preserves Sender in PersistCallback

ismaelhamed:
Currently, you have to capture Sender before calling Persist() when you use the BatchingSqlJournal, otherwise you end up with the wrong Sender in the PersistCallback.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
@@ -112,7 +112,7 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// Maximum number of failures that can happen before the circuit opens.
         /// </summary>
         public int MaxFailures { get; }
-        
+
         /// <summary>
         /// Maximum time available for operation to execute before 
         /// <see cref="CircuitBreaker"/> considers it a failure.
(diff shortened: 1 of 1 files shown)


ismaelhamed:
@Horusiath mind validating this, or suggesting and alternative solutions?

StephSanchez:
OK
Le dim. 5 mai 2019 à 15:29, Ismael Hamed <notifications@github.com> a
écrit :

src/contrib/persistence/Akka.Persistence.Sql.Common/Akka.Persistence.Sql.Common.csproj
```diff
@@ -10,6 +10,10 @@
     <NetStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netstandard1.6' ">1.6.1</NetStandardImplicitPackageVersion>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
Horusiath: While I'd like to move to value tuples, I think this is a change to be done directly on Akka core package ;)

ismaelhamed: Makes sense

Aaronontheweb:
You good with the updates @Horusiath ?

PR #3785, opened 2019-05-15
Fix repeatable failure of a "Tail_chopping_group_router" test case

heatonmatthew:
I've found that the Tail_chopping_group_router_must_throw_exception_if_no_result_will_arrive_within_the_given_time test case in the Akka.Tests project consistently fails.

While I may be missing something, I think that the reason is because the timeout of an Ask operation throws an AskTimeoutException rather than TaskCanceledException. The latter is only for when the request is explicitly cancelled using the token.

Adding the extra catch clause fixes that test.

### DIFF
--- a/src/core/Akka/Routing/TailChopping.cs
+++ b/src/core/Akka/Routing/TailChopping.cs
@@ -122,7 +122,14 @@ namespace Akka.Routing
                     try
                     {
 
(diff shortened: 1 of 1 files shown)


heatonmatthew:
These changes have been rolled into #3790. Will close this in favour of that one.

Issue #3786, opened 2019-05-16
Eliminate or fix racy unit tests


Aaronontheweb:
Version: Akka.NET v1.3.13 or later.

The time has come, once again, to clean up the mess that impedes our work on Akka.NET as a result of some racy tests.

* [x] [DistributedPubSubMessageSerializerSpec.Can_serialize_SendToOneSubscriber](http://petabridge-ci.cloudapp.net/viewLog.html?buildId=56858&buildTypeId=AkkaNet_AkkaNetWindowsUnitTestsNetCore&guest=1) - .NET Core Windows
* [x] [DistributedPubSubMessageSerializerSpec.Can_serialize_Status](http://petabridge-ci.cloudapp.net/viewLog.html?buildId=56858&buildTypeId=AkkaNet_AkkaNetWindowsUnitTestsNetCore&guest=1) - .NET Core Windows
* [ ] [AkkaProtocolSpec.ProtocolStateActor_must_in_inbound_mode_accept_payload_after_Associate_PDU_received](http://petabridge-ci.cloudapp.net/viewLog.html?buildId=56858&buildTypeId=AkkaNet_AkkaNetWindowsUnitTestsNetCore&guest=1) - .NET Core Windows
* [x] [FlowDelaySpec.A_Delay_must_pass_elements_with_delay_through_normally_in_backpressured_mode ](http://petabridge-ci.cloudapp.net/viewLog.html?buildId=56858&buildTypeId=AkkaNet_AkkaNetWindowsUnitTestsNetCore&guest=1) - .NET Core Windows
* [x] [BackoffSupervisorSpec.BackoffSupervisor_must_stop_restarting_the_child_if_final_stop_message_received_using_BackOff_OnStop](http://petabridge-ci.cloudapp.net/viewLog.html?buildId=56859&buildTypeId=AkkaNet_AkkaNetWindowsUnitTests&guest=1) - .NET Framework
* [x] [BackoffSupervisorSpec.BackoffSupervisor_must_stop_restarting_the_child_after_reaching_maxNrOfRetries_limit_using_BackOff_OnFailure](http://petabridge-ci.cloudapp.net/viewLog.html?buildId=56859&buildTypeId=AkkaNet_AkkaNetWindowsUnitTests&guest=1) - .NET Framework
[...]

PR #3790, opened 2019-05-16
Remove restrictions required by netstandard 1.x but available in 2.0

heatonmatthew:
The serialization attribute, appdomains and cloneable interfaces are now supported in the netstandard2.0 libraries so can be re-enabled for dotnet core targets.

Also noticed that I included a fix in the TailChopping Router to catch AskTimeoutException and handle that in the same way as a TaskCanceledException. The associated unit test was routinely breaking without this change for me.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -25,8 +25,8 @@ let outputPerfTests = __SOURCE_DIRECTORY__ @@ "PerfResults"
 let outputBinaries = output @@ "binaries"
 let outputNuGet = output @@ "nuget"
 let outputMultiNode = outputTests @@ "multinode"
-let outputBinariesNet45 = outputBinaries @@ "net45"
-let outputBinariesNetStandard = outputBinaries @@ "netstandard1.6"
+let outputBinariesNet45 = outputBinaries @@ "net452"
+let outputBinariesNetStandard = outputBinaries @@ "netstandard2.0"
 
 let buildNumber = environVarOrDefault "BUILD_NUMBER" "0"
 let hasTeamCity = (not (buildNumber = "0")) // check if we have the TeamCity environment variable for build # set
@@ -313,7 +313,7 @@ open Fake.TemplateHelper
 Target "PublishMntr" (fun _ ->
     let executableProjects = !! "./src/**/Akka.MultiNodeTestRunner.csproj"
 
-    // Windows .NET 4.5.2
+    // Restore
     executableProjects |> Seq.iter (fun project ->
         DotNetCli.Restore
             (fun p -> 
@@ -322,7 +322,7 @@ Target "PublishMntr" (fun _ ->
                     AdditionalArgs = ["-r win7-x64"; sprintf "/p:VersionSuffix=%s" versionSuffix] })
     )
 
-    // Windows .NET 4.5.2
+    // Windows .NET 4.6.1
     executableProjects |> Seq.iter (fun project ->  
         DotNetCli.Publish
             (fun p ->
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
@@ -43,7 +43,6 @@ namespace Akka.Cluster.Sharding.Tests
         {
         }
 
-#if SERIALIZATION
         /// <summary>
         /// Initializes a new instance of the <see cref="AsyncReplayTimeoutException"/> class.
         /// </summary>
@@ -53,7 +52,6 @@ namespace Akka.Cluster.Sharding.Tests
             : base(info, context)
         {
         }
-#endif
     }
 
     /// <summary>
--- a/src/contrib/cluster/Akka.Cluster.Sharding/Akka.Cluster.Sharding.csproj
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/Akka.Cluster.Sharding.csproj
@@ -23,10 +23,6 @@
     <Reference Include="System.Data.DataSetExtensions" />
   </ItemGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
-    <DefineConstants>$(DefineConstants);SERIALIZATION</DefineConstants>
-  </PropertyGroup>
-
   <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetStandardLibVersion)' ">
     <DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
   </PropertyGroup>
--- a/src/contrib/cluster/Akka.DistributedData/Akka.DistributedData.csproj
+++ b/src/contrib/cluster/Akka.DistributedData/Akka.DistributedData.csproj
@@ -20,10 +20,6 @@
     <PackageReference Include="Hyperion" Version="$(HyperionVersion)" />
   </ItemGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
-    <DefineConstants>$(DefineConstants);SERIALIZATION;CLONABLE</DefineConstants>
-  </PropertyGroup>
-
   <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetStandardLibVersion)' ">
     <DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
   </PropertyGroup>
--- a/src/contrib/persistence/Akka.Persistence.Query.Sql/Akka.Persistence.Query.Sql.csproj
+++ b/src/contrib/persistence/Akka.Persistence.Query.Sql/Akka.Persistence.Query.Sql.csproj
@@ -22,10 +22,6 @@
     <Reference Include="System.Xml.Linq" />
   </ItemGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
-    <DefineConstants>$(DefineConstants);SERIALIZATION</DefineConstants>
-  </PropertyGroup>
-
   <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetStandardLibVersion)' ">
     <DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
   </PropertyGroup>
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Akka.Persistence.Sql.Common.csproj
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Akka.Persistence.Sql.Common.csproj
@@ -23,7 +23,7 @@
   </ItemGroup>
 
   <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
-    <DefineConstants>$(DefineConstants);SERIALIZATION;CONFIGURATION</DefineConstants>
+    <DefineConstants>$(DefineConstants);CONFIGURATION</DefineConstants>
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetStandardLibVersion)' ">
--- a/src/contrib/serializers/Akka.Serialization.Hyperion/Akka.Serialization.Hyperion.csproj
+++ b/src/contrib/serializers/Akka.Serialization.Hyperion/Akka.Serialization.Hyperion.csproj
@@ -21,9 +21,6 @@
     <Reference Include="System" />
     <Reference Include="Microsoft.CSharp" />
   </ItemGroup>
-  <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
-    <DefineConstants>$(DefineConstants);SERIALIZATION</DefineConstants>
-  </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
     <DefineConstants>$(DefineConstants);RELEASE</DefineConstants>
   </PropertyGroup>
--- a/src/contrib/serializers/Akka.Serialization.TestKit/Akka.Serialization.TestKit.csproj
+++ b/src/contrib/serializers/Akka.Serialization.TestKit/Akka.Serialization.TestKit.csproj
@@ -14,10 +14,6 @@
     <ProjectReference Include="..\..\..\core\Akka.Tests.Shared.Internals\Akka.Tests.Shared.Internals.csproj" />
   </ItemGroup>
 
-  <PropertyGroup Condition=" '$(TargetFramework)' == 'net452' ">
-    <DefineConstants>$(DefineConstants);SERIALIZATION</DefineConstants>
-  </PropertyGroup>
-
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
     <DefineConstants>$(DefineConstants);RELEASE</DefineConstants>
   </PropertyGroup>
--- a/src/core/Akka.MultiNodeTestRunner/Discovery.cs
+++ b/src/core/Akka.MultiNodeTestRunner/Discovery.cs
@@ -19,11 +19,7 @@ using Xunit.Sdk;
 
 namespace Akka.MultiNodeTestRunner
 {
-#if CORECLR
-    public class Discovery : IMessageSink, IDisposable
-#else
     public class Discovery : MarshalByRefObject, IMessageSink, IDisposable
-#endif
     {
         public Dictionary<string, List<NodeTest>> Tests { get; set; }
         public List<ErrorMessage> Errors { get; } = new List<ErrorMessage>();
@@ -48,12 +44,8 @@ namespace Akka.MultiNodeTestRunner
                 case ITestCaseDiscoveryMessage testCaseDiscoveryMessage:
                     var testClass = testCaseDiscoveryMessage.TestClass.Class;
                     if (testClass.IsAbstract) return true;
-#if CORECLR
-                    var specType = testCaseDiscoveryMessage.TestAssembly.Assembly.GetType(testClass.Name).ToRuntimeType();
-#else
                     var testAssembly = Assembly.LoadFrom(testCaseDiscoveryMessage.TestAssembly.Assembly.AssemblyPath);
                     var specType = testAssembly.GetType(testClass.Name);
-#endif
                     var roles = RoleNames(specType);
 
                     var details = roles.Select((r, i) => new NodeTest
@@ -106,20 +98,11 @@ namespace Akka.MultiNodeTestRunner
             var current = configUser;
             while (current != null)
             {
-
-#if CORECLR
-                var ctorWithConfig = current
-                    .GetConstructors(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance)
-                    .FirstOrDefault(c => null != c.GetParameters().FirstOrDefault(p => p.ParameterType.GetTypeInfo().IsSubclassOf(baseConfigType)));
-            
-                current = current.GetTypeInfo().BaseType;
-#else
                 var ctorWithConfig = current
                     .GetConstructors(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance)
                     .FirstOrDefault(c => null != c.GetParameters().FirstOrDefault(p => p.ParameterType.IsSubclassOf(baseConfigType)));
 
                 current = current.BaseType;
-#endif
                 if (ctorWithConfig != null) return ctorWithConfig;
             }
 
@@ -131,15 +114,9 @@ namespace Akka.MultiNodeTestRunner
             var ctors = configType.GetConstructors(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance);
             var empty = ctors.FirstOrDefault(c => !c.GetParameters().Any());
 
-#if CORECLR
-            return empty != null
-                ? new object[0]
-                : ctors.First().GetParameters().Select(p => p.ParameterType.GetTypeInfo().IsValueType ? Activator.CreateInstance(p.ParameterType) : null).ToArray();
-#else
             return empty != null
                 ? new object[0]
                 : ctors.First().GetParameters().Select(p => p.ParameterType.IsValueType ? Activator.CreateInstance(p.ParameterType) : null).ToArray();
-#endif
         }
 
         /// <inheritdoc/>
--- a/src/core/Akka.Remote.Tests/Transport/DotNettySslSupportSpec.cs
+++ b/src/core/Akka.Remote.Tests/Transport/DotNettySslSupportSpec.cs
@@ -194,12 +194,8 @@ namespace Akka.Remote.Tests.Transport
             }
             finally
             {
-#if NET452 //netstandard 1.6 doesn't have close on store
                 store.Close();
-#else
-#endif
             }
-
         }
 
 
@@ -220,10 +216,7 @@ namespace Akka.Remote.Tests.Transport
             }
             finally
             {
-#if NET452 //NetStandard1.6 doesn't have close on store.
                 store.Close();
-#else
-#endif
             }
         }
         public class Echo : ReceiveActor
--- a/src/core/Akka.Remote/Serialization/ExceptionSupport.cs
+++ b/src/core/Akka.Remote/Serialization/ExceptionSupport.cs
@@ -12,9 +12,7 @@ using Akka.Actor;
 using Akka.Util;
 using Akka.Util.Internal;
 using Google.Protobuf;
-#if SERIALIZATION
 using System.Runtime.Serialization;
-#endif
 
 namespace Akka.Remote.Serialization
 {
@@ -53,11 +51,7 @@ namespace Akka.Remote.Serialization
 
         internal Proto.Msg.ExceptionData ExceptionToProto(Exception exception)
         {
-#if SERIALIZATION
             return ExceptionToProtoNet(exception);
-#else
-            return ExceptionToProtoNetCore(exception);
-#endif
         }
 
         public Exception DeserializeException(byte[] bytes)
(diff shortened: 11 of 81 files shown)


Aaronontheweb:
Thanks @heatonmatthew ! Just an FYI - you might get hit with some of these https://github.com/akkadotnet/akka.net/issues/3786 in the CI system. Known issue - not your fault. Going to be dealt with by https://github.com/akkadotnet/akka.net/pull/3788 and some other work our team is doing to clean up the test suite.

I'll look over this ASAP.

build.fsx
```diff
@@ -25,8 +25,8 @@ let outputPerfTests = __SOURCE_DIRECTORY__ @@ "PerfResults"
 let outputBinaries = output @@ "binaries"
 let outputNuGet = output @@ "nuget"
 let outputMultiNode = outputTests @@ "multinode"
-let outputBinariesNet45 = outputBinaries @@ "net45"
-let outputBinariesNetStandard = outputBinaries @@ "netstandard1.6"
+let outputBinariesNet45 = outputBinaries @@ "net452"
+let outputBinariesNetStandard = outputBinaries @@ "netstandard2.0"
```
Aaronontheweb: build.fsx changes look fine.

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/AsyncWriteProxyEx.cs
```diff
@@ -43,7 +43,6 @@ public AsyncReplayTimeoutException(string message)
         {
         }
 
-#if SERIALIZATION
```
Aaronontheweb: I can't tell you how thrilled I am to get rid of this crap @heatonmatthew - so far, according to the CI system, the compiler and the test runtimes seem pretty happy with these changes so I don't have much to add on each individual change here.

src/core/Akka.MultiNodeTestRunner/Discovery.cs
```diff
@@ -106,20 +98,11 @@ internal static ConstructorInfo FindConfigConstructor(Type configUser)
             var current = configUser;
             while (current != null)
             {
-
-#if CORECLR
-                var ctorWithConfig = current
```
Aaronontheweb: I think this is probably fine, but I'll need to see the feedback from the MNTR for certain

src/core/Akka.Remote.Tests/Transport/DotNettySslSupportSpec.cs
```diff
@@ -194,12 +194,8 @@ private void InstallCert()
             }
             finally
             {
-#if NET452 //netstandard 1.6 doesn't have close on store
                 store.Close();
```
Aaronontheweb: nice

src/core/Akka.Streams.TestKit/TestSubscriber.cs
```diff
@@ -34,7 +34,7 @@ public OnSubscribe(ISubscription subscription)
             public override string ToString() => $"TestSubscriber.OnSubscribe({Subscription})";
         }
 
-        public struct OnNext<T> : ISubscriberEvent
+        public struct OnNext<T> : ISubscriberEvent, IEquatable<OnNext<T>>
```
Aaronontheweb: What's the purpose of this change?

src/core/Akka.Tests/Actor/Dispatch/Bug2640Spec.cs
```diff
@@ -46,6 +46,12 @@ public class ThreadReporterActor : ReceiveActor
         {
             public ThreadReporterActor()
             {
+                // Improve reliability of ForkJoinExecutorShouldShutdownUponAllActorsTerminating
+                // test which intermittently can fail because all messages are actually serviced 
+                // by a single thread from the pool. Happens under load, presumably when the processor
+                // finds it more efficient to keep thread locality of the execution.
+                Thread.Sleep(1);
```
Aaronontheweb: Thanks for the explanation - makes sense.

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -224,7 +222,7 @@ internal sealed class ThreadPoolExecutorServiceFactory : ExecutorServiceConfigur
         /// <returns>TBD</returns>
         public override ExecutorService Produce(string id)
         {
-#if APPDOMAIN
+#if UNSAFE_THREADING
```
Aaronontheweb: :+1:

src/core/Akka/Routing/TailChopping.cs
```diff
@@ -122,7 +122,14 @@ public override void Send(object message, IActorRef sender)
                     try
                     {
 
-                        completion.TrySetResult(await (_routees[currentIndex].Ask(message, _within)).ConfigureAwait(false));
+                        completion.TrySetResult(
+                            await (_routees[currentIndex].Ask(message, _within)).ConfigureAwait(false));
+                    }
+                    catch (AskTimeoutException)
+                    {
+                        completion.TrySetResult(
```
Aaronontheweb: Ah, is this your fix from your other PR @heatonmatthew ?

src/core/Akka.Streams.TestKit/TestSubscriber.cs
heatonmatthew: I accidentally brought this across from a branch in which the FluentAssertions library had been upgraded. There were one or two tests that were failing in the assertion because there was no default comparer that worked correctly for these messages.

Probably best for me to remove this change for the time being?

src/core/Akka/Routing/TailChopping.cs
heatonmatthew: Yes it is. Happy to remove it if you want them separated?

src/core/Akka.Streams.TestKit/TestSubscriber.cs
Aaronontheweb: I don't think it's hurting anything, so no - you can leave it. Just wanted to make sure I wasn't overlooking something more gruesome (i.e. the entire Akka.Streams test suite exploded because equality was broken by some inexplicable platform change and this is the workaround.) But based on your explanation this sounds innocent enough. We'll keep it in.

src/core/Akka/Routing/TailChopping.cs
Aaronontheweb: Nope, it's fine to keep it in. Just wanted to make sure I understood what it was.

heatonmatthew:
@Aaronontheweb I've been doing some testing of the Akka.Persistence.SqlServer project building against a version of this branch. I've found an assumption in the Akka.Persistence.TCK.Journal.JournalSpec regarding JSON serialization that are not universally true that I think are worth mentioning.

The test is attempting to force JSON serialization to fail by implementing a test event as shown below:
```
        /// <summary>
        /// JSON serializer should fail on this
        /// </summary>
        private class NotSerializableEvent : ISerializable
        {
            public NotSerializableEvent(bool foo) { }

            public void GetObjectData(SerializationInfo info, StreamingContext context)
            {
                throw new NotImplementedException();
            }
        }
```

Now that we are running this against a netcoreapp2.1 target, it fails consistently. Thinking it through, I'm a little dubious that the comment about the class failing JSON serialization makes complete sense (because JSON serialization is different to .NET runtime serialization). After playing around with this a bit, I think that the Newtonsoft library author actually agrees with me because the behavior differs between versions as follows:

Runtime    | Newtonsoft 9.x | Newtonsoft 12.x
------------| ----------------- | ------------------
net452       |      Exception    |         OK
netcore2.1 |         OK           |           OK

I can't think of a scenario in which anyone would be relying on this behaviour in a real non-test application (i.e. deliberately trying to break JSON serialization) but I thought it's worth a broader ponder.

Now the tests can be easily fixed by changing the test class to the following.

```
        /// <summary>
        /// JSON serializer should fail on this
        /// </summary>
        private class NotSerializableEvent : ISerializable
        {
            public NotSerializableEvent(bool foo) { }

            public string Data => throw new NotImplementedException();

            public void GetObjectData(SerializationInfo info, StreamingContext context)
            {
                throw new NotImplementedException();
            }
        }
```

Do you have any strong opinion about whether this is likely to be a broader concern?

Aaronontheweb:
I need to look at this again once I'm finished with the changes in https://github.com/akkadotnet/akka.net/pull/3744

Aaronontheweb:
@heatonmatthew regarding your earlier comment, can you link me to the test classes where this should fail? Trying to understand your comment from earlier.

heatonmatthew:
@Aaronontheweb sorry for the slow response.  The test in question is [Journal_optionally_may_reject_non_serializable_events](https://github.com/akkadotnet/akka.net/blob/33aea3684f695cef2e6ff0650cece3a701fd782c/src/core/Akka.Persistence.TCK/Journal/JournalSpec.cs#L280).

In the master branch, the entire test including the NotSerializableEvent is not built for Dotnet Core because of a [#if !CORECLR](https://github.com/akkadotnet/akka.net/blob/33aea3684f695cef2e6ff0650cece3a701fd782c/src/core/Akka.Persistence.TCK/Journal/JournalSpec.cs#L265) block.  It was probably switched out for netcore originally because serialization wasn't supported at that time.

I removed the compiler directive in my branch which is probably why I'm seeing the error but the mainline tests aren't yet.

Aaronontheweb:
Got it, thanks @heatonmatthew - I'll see about getting this merged in ASAP. Going to re-run CI now with many of the other changes we've made.

Aaronontheweb:
For some reason our incremental build system isn't detecting any changes here - wonder if that has to do with `build.fsx`'s changes. Let me investigate that.

src/contrib/cluster/Akka.DistributedData/Akka.DistributedData.csproj
```diff
@@ -14,6 +14,10 @@
     <ProjectReference Include="..\..\..\core\Akka.Cluster\Akka.Cluster.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
Aaronontheweb, line 17: Remove this - not needed

src/contrib/cluster/Akka.Cluster.Sharding/Akka.Cluster.Sharding.csproj
```diff
@@ -18,6 +18,12 @@
     <ProjectReference Include="..\Akka.DistributedData\Akka.DistributedData.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
```
Aaronontheweb, line 21: Not needed

src/contrib/persistence/Akka.Persistence.Query.Sql/Akka.Persistence.Query.Sql.csproj
```diff
@@ -16,6 +16,11 @@
     <ProjectReference Include="..\Akka.Persistence.Sql.Common\Akka.Persistence.Sql.Common.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
```
Aaronontheweb, line 19: Not needed

src/contrib/persistence/Akka.Persistence.Sql.Common/Akka.Persistence.Sql.Common.csproj
```diff
@@ -21,6 +21,10 @@
     <PackageReference Include="System.Data.Common" Version="4.3.0" />
   </ItemGroup>
 
+  <PropertyGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
```
Aaronontheweb, line 24: Not needed

src/contrib/serializers/Akka.Serialization.Hyperion/Akka.Serialization.Hyperion.csproj
```diff
@@ -12,6 +12,14 @@
     <PackageReference Include="Hyperion" Version="$(HyperionVersion)" />
     <ProjectReference Include="..\..\..\core\Akka\Akka.csproj" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == '$(NetStandardLibVersion)' ">
```
Aaronontheweb, line 15: Not needed

src/contrib/serializers/Akka.Serialization.Hyperion/Akka.Serialization.Hyperion.csproj
```diff
@@ -12,6 +12,14 @@
...
+    <PackageReference Include="System.Reflection" Version="4.3.0" />
+    <PackageReference Include="System.Runtime" Version="4.3.1" />
+  </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
```
Aaronontheweb, line 19: Not needed

src/contrib/serializers/Akka.Serialization.TestKit/Akka.Serialization.TestKit.csproj
```diff
@@ -13,4 +13,9 @@
   <ItemGroup>
     <ProjectReference Include="..\..\..\core\Akka.Tests.Shared.Internals\Akka.Tests.Shared.Internals.csproj" />
   </ItemGroup>
+
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
```
Aaronontheweb, line 17: Not needed

Aaronontheweb (changes requested):
LGTM but we need to fix compliation errors and remove some of the additional stuff added to the `.csproj` files

src/core/Akka.Streams.Tests/Dsl/StageActorRefSpec.cs
```diff
@@ -178,15 +178,15 @@ public async Task A_Graph_stage_ActorRef_must_ignore_and_log_warnings_for_Poison
             warn.Message.ToString()
                 .Should()
                 .MatchRegex(
-                    "<PoisonPill> message sent to StageActor\\(akka\\://AkkaSpec/user/StreamSupervisor-[0-9]+/\\$\\$[a-z]+\\) will be ignored, since it is not a real Actor. Use a custom message type to communicate with it instead.");
+                    "<PoisonPill> message sent to StageActor\\(akka\\://StageActorRefSpec-1/user/StreamSupervisor-[0-9]+/\\$\\$[a-z]+\\) will be ignored, since it is not a real Actor. Use a custom message type to communicate with it instead.");
```
Arkatufus, line 181: Need to change this because it was hardwired as "AkkaSpec" before the cleanup.

src/core/Akka/Util/MatchHandler/CachedMatchCompiler.cs
```diff
@@ -71,27 +71,6 @@ private Delegate CompileToDelegate(IReadOnlyList<TypeHandler> handlers, IReadOnl
             delegateArguments = result.Arguments;
             return compiledLambda;
         }
-
-#if !CORECLR
```
Arkatufus, line 75: Byterot, these codes were never compiled before the cleanup (since we moved to single `netstandard2.0` platform target)

src/core/Akka/Util/MatchHandler/ILambdaExpressionCompiler.cs
```diff
@@ -22,15 +22,6 @@ internal interface ILambdaExpressionCompiler
         /// <param name="expression">The expression to compile</param>
         /// <returns>A delegate containing the compiled version of the lambda.</returns>
         Delegate Compile(LambdaExpression expression);
-
-#if !CORECLR
```
Arkatufus, line 26: Byterot, these codes were never compiled before the cleanup (since we moved to single `netstandard2.0` platform target)

src/core/Akka/Util/MatchHandler/IMatchCompiler.cs
```diff
@@ -25,20 +25,6 @@ internal interface IMatchCompiler<in T>
...
-        /// TBD
-        /// </summary>
-        /// <param name="handlers">TBD</param>
-        /// <param name="capturedArguments">TBD</param>
-        /// <param name="signature">TBD</param>
-        /// <param name="typeBuilder">TBD</param>
-        /// <param name="methodName">TBD</param>
-        /// <param name="methodAttributes">TBD</param>
-        /// <returns>TBD</returns>
-        void CompileToMethod(IReadOnlyList<TypeHandler> handlers, IReadOnlyList<Argument> capturedArguments, MatchBuilderSignature signature, TypeBuilder typeBuilder, string methodName, MethodAttributes methodAttributes = MethodAttributes.Public | MethodAttributes.Static);
```
Arkatufus, line 40: Byterot, these codes were never compiled before the cleanup (since we moved to single `netstandard2.0` platform target)

src/core/Akka/Util/MatchHandler/LambdaExpressionCompiler.cs
```diff
@@ -25,19 +25,6 @@ public Delegate Compile(LambdaExpression expression)
...
         }
-
-#if !CORECLR
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="expression">TBD</param>
-        /// <param name="method"></param>
-        /// <returns>TBD</returns>
-        public void CompileToMethod(LambdaExpression expression, MethodBuilder method)
```
Arkatufus, line 36: Byterot, these codes were never compiled before the cleanup (since we moved to single `netstandard2.0` platform target)

src/core/Akka/Util/MatchHandler/MatchBuilder.cs
```diff
@@ -171,21 +171,6 @@ public PartialAction<TItem> Build()
...
 
-#if !CORECLR
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="typeBuilder">TBD</param>
-        /// <param name="methodName">TBD</param>
-        /// <param name="attributes">TBD</param>
-        /// <returns>TBD</returns>
-        public void BuildToMethod(TypeBuilder typeBuilder, string methodName, MethodAttributes attributes = MethodAttributes.Public | MethodAttributes.Static)
```
Arkatufus, line 182: Byterot, these codes were never compiled before the cleanup (since we moved to single `netstandard2.0` platform target)

src/core/Akka/Util/MatchHandler/ILambdaExpressionCompiler.cs
```diff
@@ -22,15 +22,6 @@ internal interface ILambdaExpressionCompiler
...
-        /// <summary>
-        /// Compiles the lambda into a method definition.
-        /// </summary>
-        /// <param name="expression">The expression to compile</param>
-        /// <param name="method">A <see cref="MethodBuilder"/> which will be used to hold the lambda's IL.</param>
-        void CompileToMethod(LambdaExpression expression, MethodBuilder method);
```
Arkatufus, line 32: Byterot, these codes were never compiled before the cleanup (since we moved to single `netstandard2.0` platform target)

src/core/Akka/Util/MatchHandler/CachedMatchCompiler.cs
```diff
@@ -71,27 +71,6 @@ private Delegate CompileToDelegate(IReadOnlyList<TypeHandler> handlers, IReadOnl
...
-        /// TBD
-        /// </summary>
-        /// <param name="handlers">TBD</param>
-        /// <param name="capturedArguments">TBD</param>
-        /// <param name="signature">TBD</param>
-        /// <param name="typeBuilder">TBD</param>
-        /// <param name="methodName">TBD</param>
-        /// <param name="methodAttributes">TBD</param>
-        /// <returns>TBD</returns>
-        public void CompileToMethod(IReadOnlyList<TypeHandler> handlers, IReadOnlyList<Argument> capturedArguments, MatchBuilderSignature signature, TypeBuilder typeBuilder, string methodName, MethodAttributes methodAttributes = MethodAttributes.Public | MethodAttributes.Static)
```
Arkatufus, line 86: Byterot, these codes were never compiled before the cleanup (since we moved to single `netstandard2.0` platform target)

Arkatufus:
Most of the changes were related to .csproj cleanup and some `CORECLR`, `SERIALIZATION`, and `CONFIGURATION` compiler directive cleanup.
I've put notes on the important changes

Arkatufus:
Some test still fails, I'm still looking into that

src/core/Akka.Streams.Tests/Dsl/StageActorRefSpec.cs
Arkatufus: Further changed by using string interpolation, injecting `ActorSystem.Name` instead. No more hard wiring.

Arkatufus:
Everything should pass now

src/core/Akka.Remote/Serialization/ExceptionSupport.cs
```diff
@@ -104,22 +113,25 @@ public Exception ExceptionFromProtoNet(Proto.Msg.ExceptionData proto)
...
-            serializationInfo.AddValue("ExceptionMethod", string.Empty);
+            serializationInfo.AddValue("ExceptionMethod", null);
             serializationInfo.AddValue("HResult", int.MinValue);
 
             foreach (var field in proto.CustomFields)
             {
-                serializationInfo.AddValue(field.Key, _wrappedPayloadSupport.PayloadFrom(field.Value));
+                var payload = _wrappedPayloadSupport.PayloadFrom(field.Value);
+                if (payload is ExceptionData exception)
+                    payload = ExceptionFromProto(exception);
```
Arkatufus, line 133: Old code would fail if it encounters an `Exception` property inside an `Exception`, have to make sure that the serializer recurse through these

src/core/Akka.Remote/Serialization/ExceptionSupport.cs
```diff
@@ -137,5 +149,8 @@ public Exception ExceptionFromProtoNet(Proto.Msg.ExceptionData proto)
 
             return obj;
         }
+
+        private string ValueOrNull(string value)
+            => string.IsNullOrEmpty(value) ? null : value;
```
Arkatufus, line 154: Default value for strings inside an `Exception` is null, have to convert proto empty strings to null to make a proper `Exception` object

src/core/Akka.Remote/Serialization/ExceptionSupport.cs
```diff
@@ -84,14 +85,22 @@ public Proto.Msg.ExceptionData ExceptionToProtoNet(Exception exception)
...
 
             foreach (var info in serializationInfo)
             {
-                if (DefaultProperties.Contains(info.Name)) continue;
-                var preparedValue = _wrappedPayloadSupport.PayloadToProto(info.Value);
-                message.CustomFields.Add(info.Name, preparedValue);
+                if (_defaultProperties.Contains(info.Name)) continue;
+                if (info.Value is Exception exceptionValue)
+                {
+                    var exceptionPayload = ExceptionToProto(exceptionValue);
```
Arkatufus, line 96: Old code would fail if it encounters an Exception property inside an Exception, have to make sure that the serializer recurse through these

Arkatufus:
Still need to test that exceptions in `Akka.Cluster` are serialized properly

src/core/Akka.Tests/Actor/Dispatch/Bug2640Spec.cs
```diff
@@ -76,6 +82,7 @@ public async Task ForkJoinExecutorShouldShutdownUponActorSystemTermination()
 
             threads = ReceiveN(100).Cast<Thread>().GroupBy(x => x.ManagedThreadId)
                 .ToDictionary(x => x.Key, grouping => grouping.First());
+            threads.Count.Should().Be(4, "Expected 4 distinct threads in this example");
```
Arkatufus, line 85: This change actually makes the test really flaky. In CI/CD environment, `Count` can either be 4 or 3, it looks like there is no guarantee that the system can provide all the requested thread count.

src/core/Akka.Tests/Actor/Dispatch/Bug2640Spec.cs
```diff
@@ -82,7 +82,6 @@ public async Task ForkJoinExecutorShouldShutdownUponActorSystemTermination()
 
             threads = ReceiveN(100).Cast<Thread>().GroupBy(x => x.ManagedThreadId)
                 .ToDictionary(x => x.Key, grouping => grouping.First());
-            threads.Count.Should().Be(4, "Expected 4 distinct threads in this example");
```
Arkatufus: We don't actually care about the amount of thread that is assigned to the test, we're testing if they're all terminated correctly.

Aaronontheweb: We did care about that early on - I haven't seen this spec act up in many years though. Why is it having problems now?

Arkatufus: This is a new addition from this PR, we don't have this test before

Arkatufus: See my explanation comment just before this

Aaronontheweb: Ah I see

Aaronontheweb:
Going to move this into 1.4.20 just because 1.4.19 is already quite a big release, but this looks good to me thus far

Aaronontheweb (approved):
LGTM

Aaronontheweb:
@Arkatufus this test is now failing on every run (racily) on this PR since your changes:

```
ForkJoinExecutor should terminate all threads upon all attached actors shutting down
```

Please revert

Aaronontheweb:
@Arkatufus seeing this show up in the MNTR for .NET 5 / .NET Core 3.1

```
Failed to load test details for [Akka.Cluster.Tests.MultiNode.SingletonClusterWithAccrualFailureDetectorMultiNode] test class: System.IO.FileNotFoundException: Could not load file or assembly 'D:\a\1\s\src\core\Akka.MultiNodeTestRunner\bin\Release\net5.0\win10-x64\publish\Akka.Cluster.Tests.MultiNode.dll'. The system cannot find the file specified.
File name: 'D:\a\1\s\src\core\Akka.MultiNodeTestRunner\bin\Release\net5.0\win10-x64\publish\Akka.Cluster.Tests.MultiNode.dll'
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromPath(IntPtr ptrNativeAssemblyLoadContext, String ilPath, String niPath, ObjectHandleOnStack retAssembly)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String assemblyPath)
   at System.Reflection.Assembly.LoadFrom(String assemblyFile)
```

That's why everything is running in 15 minutes. Any ideas on this?

Aaronontheweb:
Looks like it's working now - must have been an intermittent issue.

PR #3791, opened 2019-05-16
Fix serialize-messages for Akka.Cluster and Akka.Remote (netstandard2.0)

heatonmatthew:
Applied changes from #3725 on top of netstandard 2.0 targeted upgrade with the serialization restriction of dotnet core projects lifted (#3790).

### DIFF
--- a/src/core/Akka.Cluster/Member.cs
+++ b/src/core/Akka.Cluster/Member.cs
@@ -11,6 +11,7 @@ using System.Collections.Immutable;
 using System.Linq;
 using Akka.Actor;
 using Akka.Util.Internal;
+using Newtonsoft.Json;
 
 namespace Akka.Cluster
 {
@@ -90,6 +91,20 @@ namespace Akka.Cluster
             UpNumber = upNumber;
             Status = status;
             Roles = roles;
+        }
+
+        /// <summary>
+        /// Used when `akka.actor.serialize-messages = on`.
+        /// </summary>
+        /// <param name="uniqueAddress">The address of the member.</param>
+        /// <param name="upNumber">The upNumber of the member, as assigned by the leader at the time the node joined the cluster.</param>
+        /// <param name="status">The status of this member.</param>
+        /// <param name="roles">The roles for this member. Can be empty.</param>
+        [JsonConstructor]
+        internal Member(UniqueAddress uniqueAddress, int upNumber, MemberStatus status, IEnumerable<string> roles)
+            : this(uniqueAddress, upNumber, status, roles.ToImmutableHashSet())
+        {
+
         }
 
         /// <summary>
--- a/build.fsx
+++ b/build.fsx
@@ -25,8 +25,8 @@ let outputPerfTests = __SOURCE_DIRECTORY__ @@ "PerfResults"
 let outputBinaries = output @@ "binaries"
 let outputNuGet = output @@ "nuget"
 let outputMultiNode = outputTests @@ "multinode"
-let outputBinariesNet45 = outputBinaries @@ "net45"
-let outputBinariesNetStandard = outputBinaries @@ "netstandard1.6"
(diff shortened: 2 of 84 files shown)


src/core/Akka.Cluster/Member.cs
```diff
@@ -92,6 +93,20 @@ internal Member(UniqueAddress uniqueAddress, int upNumber, MemberStatus status,
...
         }
 
+        /// <summary>
+        /// Used when `akka.actor.serialize-messages = on`.
+        /// </summary>
+        /// <param name="uniqueAddress">The address of the member.</param>
+        /// <param name="upNumber">The upNumber of the member, as assigned by the leader at the time the node joined the cluster.</param>
+        /// <param name="status">The status of this member.</param>
+        /// <param name="roles">The roles for this member. Can be empty.</param>
+        [JsonConstructor]
```
zbynek001: isn't this serialized via protobuff ClusterMessageSerializer? Why json serialization?

heatonmatthew: @zbynek001 A good question to ask. You're correct that there are a lot of protobuf encoded Cluster Messages (in ClusterMessages.g.cs), however this particular Member.cs class is actually being referenced in the messages defined under ClusterEvent.XYZ which are messages published to the EventBus (and in the context of the BugFix3724Spec test at least, are destined for the Event.LoggingBus).

I am new to the akka.net code base so take the following comments as very tentative:

- The messages under ClusterEvent.XYZ are JSON serialized by design at this point
- My gut says that protobuf would be better (even if we are waiting for [Hyperion](https://github.com/akkadotnet/akka.net/pull/3725#discussion_r260521727)) but that would be a breaking change
- I'm keen to circle back to [this comment](https://github.com/akkadotnet/akka.net/pull/3725#discussion_r260521054) in the context of netstandard2.0 to see whether #3790 will enable the INoSerializationVerificationNeeded restriction to be lifted

@Aaronontheweb Do you have any input on this point?

heatonmatthew: So after a quick tour through the code, I think that the underlying Gossip and state management messages for the cluster do use the protobuf messages, but the observable messages published by the ClusterDomainEventPublisher on the EventBus are JSON serialized.

I'm interested to know if this is a specific design decision and/or if there would be resistance to migrating to a consistent protobuf wire format for both?

Aaronontheweb:
@heatonmatthew so I just merged in https://github.com/akkadotnet/akka.net/pull/3725 - but you have some additional code in this PR beyond my changes. What's the best way to reconcile this?

src/core/Akka.Cluster/Member.cs
Aaronontheweb: @zbynek001 @heatonmatthew so this might be an issue with `serialize-messages` inside Akka core where we don't use the correct serializer based on its manifest type.... I'm going to see if I can fix that in https://github.com/akkadotnet/akka.net/pull/3744 - I added a `JsonConstructor` call similar to this one for the `Member` class and that may no longer be necessary once the serialization fixes are added to serialize-messages in that PR.

heatonmatthew:
@Aaronontheweb this one builds on my changes from #3790 which is waiting to be merged. The majority of the extra changes will be from that, so it would be helpful to progress that one first. I've just updated that branch now.

Once #3790 is merged, I can review this against the combined set of changes. Since this whole branch is intended to be a cherry-pick of #3724 on top of #3790 (mainly so that I can use it in pre-production for myself), there might be nothing more to do with this specific task.

src/core/Akka.Cluster/Member.cs
heatonmatthew: @Aaronontheweb let me know if there's anything specific you could do with a hand on. It's quite interesting code to be working on.

Aaronontheweb:
No needed - this is already fixed.

PR #3780, opened 2019-05-06
Batch update with singleton/sharding fixes

zbynek001:
Mostly singleton & sharding fixes, plus some dependencies
Will split it into smaller parts if needed

Included:
https://github.com/akka/akka/pull/25854
https://github.com/akka/akka/issues/25274
https://github.com/akka/akka/pull/24442
https://github.com/akka/akka/issues/25639
https://github.com/akka/akka/issues/26336
https://github.com/akka/akka/issues/26793
https://github.com/akka/akka/pull/25648
https://github.com/akka/akka/pull/26061
https://github.com/akka/akka/issues/26012
https://github.com/akka/akka/issues/26214
https://github.com/akka/akka/issues/26451
https://github.com/akka/akka/pull/26819
& persistent shard passivate-idle fix

PR #3801, opened 2019-05-21
[SpecFix] fixed Passivation_of_inactive_entities spec

Aaronontheweb:
#3786

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/InactiveEntityPassivationSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/InactiveEntityPassivationSpec.cs
@@ -99,9 +99,13 @@ namespace Akka.Cluster.Sharding.Tests
         public void Passivation_of_inactive_entities_must_passivate_entities_when_they_have_not_seen_messages_for_the_configured_duration()
         {
             // Single node cluster
-            Cluster.Get(Sys).Join(Cluster.Get(Sys).SelfAddress);
+            Within(TimeSpan.FromSeconds(5), () =>
+            {
+                Cluster.Get(Sys).Join(Cluster.Get(Sys).SelfAddress);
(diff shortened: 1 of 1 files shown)


src/contrib/cluster/Akka.Cluster.Sharding.Tests/InactiveEntityPassivationSpec.cs
```diff
@@ -132,7 +132,7 @@ public void Passivation_of_inactive_entities_must_passivate_entities_when_they_h
 
             // Make sure "1" hasn't seen a message in 3 seconds and passivates
             var timeSinceOneSawAMessage = DateTime.Now.Ticks - timeOneSawMessage;
-            probe.ExpectNoMsg(TimeSpan.FromSeconds(3) - TimeSpan.FromTicks(timeSinceOneSawAMessage));
+            probe.ExpectNoMsg(TimeSpan.FromSeconds(3) - TimeSpan.FromTicks(timeSinceOneSawAMessage) - TimeSpan.FromMilliseconds(40));
```
Aaronontheweb: The fix here was simple - this spec, when it'd fail on the build server, would be for reasons like this:

```
Failed: Expected no messages during 00:00:00.9951057, instead we received <1 passivating> from [akka://AkkaSpec/system/sharding/myType/1/1#392767797] after 00:00:00.9946593
Expected: True
Actual: False
```
The issue in this case is that the time we were setting during the `ExpectNoMsg` period is more precise than what's allowed by the server's clock resolution. Thus, I padded the time by an additional 40ms - which should be plenty enough to make sure the test is still meaningful but a big enough margin to avoid having the test too tightly fitted.

Aaronontheweb:
Race conditions still aren't quite resolved yet even though the last build passed - have some more hardening to do on this spec.

Aaronontheweb:
Covered via #3780

PR #3805, opened 2019-05-21
fix PersistentActor_should_be_able_to_persist_events_that_happen_during_recovery

Aaronontheweb:
#3786 - fix PersistentActor_should_be_able_to_persist_events_that_happen_during_recovery by porting https://github.com/akka/akka/pull/22232

### DIFF
--- a/src/core/Akka.Persistence.Tests/PersistentActorRecoveryTimeoutSpec.cs
+++ b/src/core/Akka.Persistence.Tests/PersistentActorRecoveryTimeoutSpec.cs
@@ -11,14 +11,16 @@ using Akka.Configuration;
 using Akka.Event;
 using Akka.Persistence.Tests.Journal;
 using Akka.TestKit;
+using Akka.Util.Internal;
 using Xunit;
 
 namespace Akka.Persistence.Tests
 {
-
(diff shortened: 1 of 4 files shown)


src/core/Akka.Persistence.Tests/PersistentActorRecoveryTimeoutSpec.cs
```diff
@@ -11,14 +11,16 @@
...
 using Xunit;
 
 namespace Akka.Persistence.Tests
 {
-
+    [Collection("PersistentActorRecoveryTimeout")] // force tests to run sequentially
     public class PersistentActorRecoveryTimeoutSpec : PersistenceSpec
     {
-        private const string JournalId = "persistent-actor-recovery-timeout-spec";
+        private static readonly AtomicCounter JournalIdNumber = new AtomicCounter(0);
```
Aaronontheweb: Race conditions in this test were caused by multiple parallel test cases accessing the same `SteppingMemoryJournal` - forced this test case to run sequentially (via the `Collection` attribute on the class) and made it so each `SteppingMemoryJournal` instance was isolated to each test case.

PR #3806, opened 2019-05-22
[SpecFix] working on fixing AtLeastOnceDeliveryCrashSpec

Aaronontheweb:
#3786 - added some debug logging to see why this test is racy. Looks like it's either an issue with Recovery, the timing configuration of the `AtLeastOnceDeliveryActor` in this spec, or the `AtLeastOnceDeliverySemantic` not handling recoveries properly.


src/core/Akka.Persistence.Tests/ReceivePersistentActorAsyncAwaitSpec.cs
```diff
@@ -574,11 +574,13 @@ public AsyncPipeToDelayActor(string persistenceId)
...
                 CommandAsync<string>(async msg =>
                 {
+                    var sender = Sender;
+                    var self = Self;
                     Task.Run(() =>
                     {
                         Thread.Sleep(10);
                         return msg;
-                    }).PipeTo(Sender, Self); //LogicalContext is lost?!?
+                    }).PipeTo(sender, self);
```
Aaronontheweb: Fixed this spec too - would sometimes fail because we didn't close over `Sender` .

Aaronontheweb: This isn't the final fix here - still having issues.

PR #3808, opened 2019-05-22
sharding PassivateIdle fix

zbynek001:
PassivateIdleTask was not cancelled for persistent shards on stop leading to deadletters


Aaronontheweb:
LGTM

PR #3810, opened 2019-05-25
Allow AtLeastOnceDelivery parameters to be set from deriving classes (as intended)

ondrejpialek:
Currently all AtLeastOnceDelivery parameters are read from settings passed via `ctor` when the semantic is created. The documentation says that properties that expose these parameters may be overwritten in derived classes to provide different values. This however does not work. The semantic never reads from the derived actors themselves (it has no reference to the owning actor to even read the properties).

This PR makes the owning actors' properties non-virtual and write through, effectively encapsulating the ones of the semantic. The semantic is additionally updated to handle changing values even as the actor runs (so they can be modified even outside the actor's `ctor` and the new values will be picked up by the semantic).

The values from settings are still the default.

I also fixed the redelivery interval being fired at double the intended rate.

This might be a potentially breaking change, as the properties are no longer virtual. But anyone overwriting these was not getting what they wanted anyway - so perhaps a good thing they will get a compiler error and thus a chance to update their code to set the interval and other properties properly?


src/core/Akka.Persistence.Tests/AtLeastOnceDeliveryFailureSpec.cs
```diff
@@ -144,9 +144,9 @@ public ChaosSender(IActorRef destination, IActorRef probe)
                 _config = Context.System.Settings.Config.GetConfig("akka.persistence.sender.chaos");
                 _liveProcessingFailureRate = _config.GetDouble("live-processing-failure-rate");
                 _replayProcessingFailureRate = _config.GetDouble("replay-processing-failure-rate");
-            }
 
-            public override TimeSpan RedeliverInterval { get { return TimeSpan.FromMilliseconds(500); } }
+                RedeliverInterval = TimeSpan.FromMilliseconds(500);
```
ondrejpialek: This is a good example how these properties are now intended to be used; just set them anywhere from within the actor instead of overwriting.

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
```diff
@@ -380,8 +432,8 @@ public AtLeastOnceDeliverySemantic(IActorContext context, PersistenceSettings.At
         private void StartRedeliverTask()
         {
             if (_redeliverScheduleCancelable != null) return;
-            var interval = new TimeSpan(RedeliverInterval.Ticks / 2);
-            _redeliverScheduleCancelable = _context.System.Scheduler.ScheduleTellRepeatedlyCancelable(interval, interval, _context.Self,
+            var delay = new TimeSpan(RedeliverInterval.Ticks / 2);
+            _redeliverScheduleCancelable = _context.System.Scheduler.ScheduleTellRepeatedlyCancelable(delay, RedeliverInterval, _context.Self,
```
ondrejpialek: Was this a bug? The first parameter is the initial delay of the very first callback. The second parameter is then the interval this will fire in, so this should be the whole `RedeliverInterval`, not half?

Aaronontheweb:
@ondrejpialek need to approve the public API changes: https://getakka.net/community/public-api-changes.html

Aaronontheweb:
That'll get the failing unit test to pass.

ondrejpialek:
Sorry I haven't read that part of the docs. That explains that popup I saw and the files I had to exclude from the commit :) Will add it and update the PR.

ondrejpialek:
PR ready @Aaronontheweb

Aaronontheweb:
Thank you!

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
```diff
@@ -326,18 +328,37 @@ public sealed class RedeliveryTick : INotInfluenceReceiveTimeout, IDeadLetterSup
...
-        /// non-default values.
+        /// configuration key.
         /// </summary>
-        public virtual TimeSpan RedeliverInterval => _settings.RedeliverInterval;
+        public TimeSpan RedeliverInterval
+        {
+            get => _redeliverInterval;
+            set
+            {
+                if (_redeliverInterval != value)
```
Horusiath: Doesn't this looks too surprising? I mean, using setter is usually considered to be simple operation with no side effect (except changing value type). The fact, that it resets schedule may be the source of problems, for example in debugging, in the future.

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
```diff
@@ -346,31 +367,62 @@ public AtLeastOnceDeliverySemantic(IActorContext context, PersistenceSettings.At
...
+                {
+                    _warnAfterNumberOfUnconfirmedAttempts = value;
+
+                    var deliveriesAboveThreshold = _unconfirmed.Where(u => u.Value.Attempt >= value)
+                        .Select(u => new UnconfirmedDelivery(u.Key, u.Value.Destination, u.Value.Message))
+                        .ToArray();
+
+                    if (deliveriesAboveThreshold.Length > 0)
+                    {
+                        _context.Self.Tell(new UnconfirmedWarning(deliveriesAboveThreshold));
```
Horusiath: Same here, the last thing I would expect from using (`this.WarnAfterNumberOfUnconfirmedAttempts = 5`) is that it may send messages to actors and allocate extra arrays.

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
ondrejpialek: It's here so that the semantic behaves "correctly", e.g. if you set a different interval it should send things in a different interval. I agree that having too much logic in setters is not great though (thought I thought it's acceptable and tried to limit the resets by checking for a change).

What do you propose? Have a method instead for changing this? Or have this read only and only accept the value via `ctor`? Or making this smarter by calculating the correct delay between the elapsed time and new interval?

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
ondrejpialek: Again - this is to make the logic correct - if you start with 10, have 6 unconfirmed, then switch to 5 then you should receive the warning.

What do you think is best? Have this read only (and accept the value only via `ctor`?) or not trigger immediately but wait for next unconfirmed message?

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
Horusiath: AFAIK those configuration values should be defined exactly once - when an actor is created. I'd probably need verify it somehow, but we're having a state machines here, so changing this value in different actor "state" may result in different side effects - and I'm not sure if all of them are acceptable.

Simplest way I can think of ATM to make it break the actor would be to call `RedeliverInterval = x` in actor's PostStop method after calling `base.PostStop()`, which is cancelling all scheduled messages. Any call like this one made afterwards would trigger it to schedule the message back again, but never actually cancel.

ondrejpialek: That would not happen the way it's currently written as it checks for `_redeliverScheduleCancelable ` being active before restarting it.

```
if (_redeliverScheduleCancelable != null)
{
	_redeliverScheduleCancelable.Cancel();
	_redeliverScheduleCancelable = null;
	StartRedeliverTask();
}
```

But if you prefer not to have these mutable I can redo this PR in a different way - the values would be set via `ctor` only (at the cost of having to read the redelivery settings in two places (both AtLeastOnce actors that use this semantic).

The problem with that approach is that the `AtLeastOnceDeliverySemantic` is created in the actor's `ctor` and you should not access virtual fields inside a `ctor`, so the only way that comes in mind is specifying two `ctors` for each actor so that you can specify these parameters calling the right base `ctor`.

ondrejpialek: Actually there is already a `ctor` that takes the redelivery settings. So perhaps we make the method non-virtual in order not to confuse the developers (as overwriting has no effect currently) and we fix the docs to say to use the overloaded `ctor` instead of overwriting?

Horusiath: It would be nice, but the problem is backwards compatibility. I think, we can stay with your current approach. Any thoughts @Aaronontheweb ?

Aaronontheweb:
@Horusiath saw that you tagged me here - my apologies for the delay! I'll take a look at this today.

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
Aaronontheweb: @ondrejpialek @Horusiath 

> Actually there is already a ctor on the actors that takes the redelivery settings. So perhaps we make the properties non-virtual in order not to confuse the developers (as overwriting has no effect currently) and we fix the docs to say to use the overloaded ctor instead of overwriting?

Since this is going into the v1.4 branch and we can afford to take some more liberties with the API there - if overriding the properties already has no effect, I don't see an issue with removing the `virtual` modifier on those properties and just having developers use the constructor overload instead. I think making these settings immutable and settable through the constructor is probably the best way to go, in order to avoid having those side effects during the processing of the state machine as @Horusiath mentioned.

Can you go ahead with those changes @ondrejpialek ?

src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
```diff
@@ -16,11 +16,11 @@ namespace Akka.Persistence
     {
         protected AtLeastOnceDeliveryActor() { }
         protected AtLeastOnceDeliveryActor(Akka.Persistence.PersistenceSettings.AtLeastOnceDeliverySettings settings) { }
-        public int MaxUnconfirmedMessages { get; }
-        public virtual System.TimeSpan RedeliverInterval { get; }
```
Aaronontheweb: I'm ok with this change.

src/core/Akka.Persistence/AtLeastOnceDeliveryReceiveActor.cs
```diff
@@ -52,31 +55,40 @@ protected AtLeastOnceDeliveryReceiveActor(PersistenceSettings.AtLeastOnceDeliver
...
         /// The default value can be configure with the 'akka.persistence.at-least-once-delivery.max-unconfirmed-messages'
-        /// configuration key. This method can be overridden by implementation classes to return
-        /// non-default values.
+        /// configuration key.
         /// </summary>
-        public int MaxUnconfirmedMessages => _atLeastOnceDeliverySemantic.MaxUnconfirmedMessages;
+        public int MaxUnconfirmedMessages
+        {
+            get => _atLeastOnceDeliverySemantic.MaxUnconfirmedMessages;
+            set => _atLeastOnceDeliverySemantic.MaxUnconfirmedMessages = value;
```
Aaronontheweb: All of these settings should be immutable and overloadable via the CTOR, as you mentioned @ondrejpialek

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
Aaronontheweb: Agree with @Horusiath - this should be immutable I think. Pick your strategy and stick with it at the time the actor starts.

src/core/Akka.Persistence/AtLeastOnceDeliverySemantic.cs
Aaronontheweb: yeah, must have been a bug. Nice catch.

src/core/Akka.Persistence/AtLeastOnceDelivery.cs
```diff
@@ -61,14 +61,24 @@ protected AtLeastOnceDeliveryActor(PersistenceSettings.AtLeastOnceDeliverySettin
             _atLeastOnceDeliverySemantic = new AtLeastOnceDeliverySemantic(Context, settings);
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AtLeastOnceDeliveryActor"/> class.
+        /// </summary>
+        /// <param name="overrideSettings">A lambda to tweak the default AtLeastOnceDelivery settings.</param>
+        protected AtLeastOnceDeliveryActor(Func<PersistenceSettings.AtLeastOnceDeliverySettings, PersistenceSettings.AtLeastOnceDeliverySettings> overrideSettings)
```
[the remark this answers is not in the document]
Aaronontheweb: I think that's a reasonable compromise and a clean API.

Aaronontheweb:
@Horusiath I'm good with this - any objections?

PR #3817, opened 2019-06-04
fixed issue with multiple specs all trying to use same Akka.Remote port at the same time

Aaronontheweb:
part of #3786


Aaronontheweb:
I think I addressed this inside https://github.com/akkadotnet/akka.net/pull/3744 using `Collection` attributes instead of changing the port numbers to 0. Might be a cleaner way of handling this.

Aaronontheweb:
This is a more heavy-duty change, but I think it's the right way to handle this: if we're going to run the TestSuite concurrently, have everything use port 0 by default. Any HOCON supplied on an individual test will override this.

Issue #3830, opened 2019-06-26
Akka.Cluster.Sharding GetShardRegionStats serialization issue


Aaronontheweb:
Version: Akka.NET v1.4.0 current `dev` branch

Error occurs when attempting to send a  `GetClusterShardingStats` message from a non-coordinator `ShardRegion` to the `PersistentShardRegion` coordinator, in this case via the [Petabridge.Cmd.Cluster.Sharding command palette](https://cmd.petabridge.com/articles/commands/cluster-sharding-commands.html), a deserialization error occurs in JSON.NET when deserializing the `ClusterShardingStats` response sent back by the coordinator:

```
ka.tcp://ShardFight@7061a2bcb4e7:5110: Error [Could not convert string 'akka.tcp://ShardFight@8fb41e2d2e27:5110' to dictionary key type 'Akka.Actor.Address'. Create a TypeConverter to convert from the string to the key type object. Path 'Regions['akka.tcp://ShardFight@8fb41e2d2e27:5110']', line 1, position 324.] [   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateDictionary(IDictionary dictionary, JsonReader reader, JsonDictionaryContract contract, JsonProperty containerProperty, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ResolvePropertyAndCreatorValues(JsonObjectContract contract, JsonProperty containerProperty, JsonReader reader, Type objectType)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObjectUsingCreatorWithParameters(JsonReader reader, JsonObjectContract contract, JsonProperty containerProperty, ObjectConstructor`1 creator, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
[...]

PR #3832, opened 2019-06-26
ShardStats and ClusterShardingStats serialization fix

Aaronontheweb:


Aaronontheweb:
This is ready to go - will fix #3830

src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs
```diff
@@ -69,9 +77,9 @@ public ClusterShardingMessageSerializer(ExtendedActorSystem system) : base(syste
         {
             _fromBinaryMap = new Dictionary<string, Func<byte[], object>>
             {
-                {EntityStateManifest, EntityStateFromBinary},
-                {EntityStartedManifest, EntityStartedFromBinary},
-                {EntityStoppedManifest, EntityStoppedFromBinary},
+                {EntityStateManifest, bytes => EntityStateFromBinary(bytes) },
```
Horusiath: I prefer using method straight away instead of lambda syntax - I don't know it lambda syntax will trigger generation of hidden class representing it, but this way I'm sure I don't give a compiler reasons to think it should try so.

src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs
```diff
@@ -140,6 +152,10 @@ public override byte[] ToBinary(object obj)
                 case ShardRegion.StartEntityAck o: return StartEntityAckToProto(o).ToByteArray();
                 case Shard.GetShardStats o: return new byte[0];
                 case Shard.ShardStats o: return ShardStatsToProto(o).ToByteArray();
+                case GetShardRegionStats o: return new byte[0];
```
Horusiath: I think, .NET standard offers `Array.Empty<byte>()` which is static and doesn't allocate new object every time.

src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs
```diff
@@ -378,14 +398,14 @@ private IActorRef ActorRefMessageFromBinary(byte[] binary)
         // Shard.ShardState
         //
 
-        private Proto.Msg.EntityState EntityStateToProto(Shard.ShardState entityState)
+        private static Proto.Msg.EntityState EntityStateToProto(Shard.ShardState entityState)
```
Horusiath: Static methods are always nicer :+1:

src/contrib/cluster/Akka.Cluster.Sharding/ShardingMessages.cs
```diff
@@ -141,42 +142,73 @@ public CurrentRegions(IImmutableSet<Address> regions)
...
+        /// Creates a new GetClusterShardingStats message instance.
         /// </summary>
-        /// <param name="timeout">TBD</param>
+        /// <param name="timeout">The amount of time to allow this operation to run.</param>
         public GetClusterShardingStats(TimeSpan timeout)
         {
             Timeout = timeout;
         }
+
+        private bool Equals(GetClusterShardingStats other)
```
Horusiath: If you created equality rules, please make that type implement IEquatable explicitly.

src/contrib/cluster/Akka.Cluster.Sharding/ShardingMessages.cs
```diff
@@ -141,42 +142,73 @@ public CurrentRegions(IImmutableSet<Address> regions)
...
+        /// Creates a new ClusterShardingStats message.
         /// </summary>
-        /// <param name="regions">TBD</param>
+        /// <param name="regions">The set of sharding statistics per-node.</param>
         public ClusterShardingStats(IImmutableDictionary<Address, ShardRegionStats> regions)
         {
             Regions = regions;
         }
+
+        private bool Equals(ClusterShardingStats other)
```
Horusiath: Same thing here about IEquatable.

src/contrib/cluster/Akka.Cluster.Sharding/ShardingMessages.cs
```diff
@@ -242,24 +274,40 @@ public CurrentShardRegionState(IImmutableSet<ShardState> shards)
...
+        /// Creates a new ShardRegionStats instance.
         /// </summary>
-        /// <param name="stats">TBD</param>
+        /// <param name="stats">The set of shardId / entity count pairs</param>
         public ShardRegionStats(IImmutableDictionary<string, int> stats)
         {
             Stats = stats;
         }
+
+        private bool Equals(ShardRegionStats other)
```
Horusiath: Again: IEquatable.

src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs
Aaronontheweb: Not supported in .NET 4.5.2, which we are still multi-targeting. I'm thinking about dropping multi-targeting altogether but I don't want to byte that off in the first beta of v1.4.0 just yet.

src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs
Aaronontheweb: The existing serializer does a mix of both already and these messages aren't sent very often.

src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs
Aaronontheweb: Agree!

Aaronontheweb:
Addressed comments

PR #3833, opened 2019-06-27
Don't passivate idle for remembering entities

ismaelhamed:
Port [#27177](https://github.com/akka/akka/pull/27177)


ismaelhamed:
Some tests seem to be failing, I'll look into it.

ismaelhamed:
So we had one outdated test, and another one that is not needed anymore because having remember entities enabled will now disable passivate idle.

This is now ready to go.

PR #3835, opened 2019-07-01
Add Epsilon within Within in TestKitSampleTest.cs

jg11jg:
I'm new to Akka and spent ages trying to figure out why the above test kept intermittently failing despite me copying it directly into visual studio.  Turned out it was because of this issue: https://github.com/akkadotnet/akka.net/issues/2130.  So, I suggest including an Epsilon into the TestKit example.


Danthar:
Good catch.

PR #3839, opened 2019-07-07
disable some racy RemotingSpecs [Akka.Remote.Tests]

Aaronontheweb:
These specs seemed to be especially sensitive to timeouts due to some of them being long-running - forcing them to run serially to see if it helps.


Aaronontheweb:
Per https://github.com/akkadotnet/akka.net/issues/3786

PR #3841, opened 2019-07-07
disable XUnit parallel execution for Akka.Remote.Tests

Aaronontheweb:


src/xunitSettings.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+  <!-- use the xunit JSON settings file -->
+  <ItemGroup>
```
Aaronontheweb: This can be used in other test suites too - the Akka.Remote ones have the most flaky behavior aside from a handful of Akka.Streams specifications.

Aaronontheweb:
Looks like some of our specs using remote deployments appear to be the common thread here - something about the `ResolveActor` calls there can be racy over the in-memory transport.

PR #3842, opened 2019-07-07
reduce allocations in RemoteSystemDaemon ActorPath resolution

Aaronontheweb:
Also cleaned up the API.

### DIFF
--- a/src/core/Akka.Remote/RemoteSystemDaemon.cs
+++ b/src/core/Akka.Remote/RemoteSystemDaemon.cs
@@ -304,44 +304,25 @@ namespace Akka.Remote
         /// <returns>ActorRef.</returns>
         public override IActorRef GetChild(IEnumerable<string> name)
         {
-            var elements = name.ToArray();
-            var path = elements.Join("/");
+            var path = name.Join("/");
             var n = 0;
             while (true)
             {
                 var nameAndUid = ActorCell.SplitNameAndUid(path);
-                var child = GetChild(nameAndUid.Name);
-                if (child == null)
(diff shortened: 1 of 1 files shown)


src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -304,45 +304,26 @@ private void HandleDaemonMsgCreate(DaemonMsgCreate message)
         /// <returns>ActorRef.</returns>
         public override IActorRef GetChild(IEnumerable<string> name)
         {
-            var elements = name.ToArray();
-            var path = elements.Join("/");
```
Aaronontheweb: No need for redundant .ToArray + Join call.  All of the objects passed into this method are `IReadOnlyList<string>`

src/core/Akka.Remote/RemoteSystemDaemon.cs
```diff
@@ -304,45 +304,26 @@ private void HandleDaemonMsgCreate(DaemonMsgCreate message)
...
                 }
-                if (nameAndUid.Uid != ActorCell.UndefinedUid && nameAndUid.Uid != child.Path.Uid)
-                    return Nobody.Instance;
-
-                return n == 0 ? child : child.GetChild(elements.TakeRight(n));
-            }
-        }
-
 
-        private IInternalActorRef GetChild(string name)
```
Aaronontheweb: No need for extra `GetChild` method, which also calls `ActorCell.SplitNameAndUid` redundantly

PR #3843, opened 2019-07-08
NBench standardization

Aaronontheweb:
This is the first in a few pull requests aimed at offering better performance testing support for Akka.NET - this upgrades the solution to use NBench 1.2.2 and uses the `dotnet-nbench` command, but I will need to make some changes to NBench in order to make `dotnet-nbench` fully compatible with `common.props` before it can be run officially.

I'll also be submitting another pull request aimed at introducing .NET Core tests for Akka.NET, which NBench 1.2.2 is a pre-requisite for.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -345,55 +345,28 @@ Target "MultiNodeTestsNetCore" (fun _ ->
 
 Target "NBench" <| fun _ ->
     if not skipBuild.Value then
-        CleanDir outputPerfTests
-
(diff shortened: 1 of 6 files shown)


Aaronontheweb:
Should also note: all NBench tests support incremental builds.

Aaronontheweb:
This should also speed up most local builds - no need to install `NBench.Runner` anymore.

PR #3845, opened 2019-07-12
enable CoordinatedShutdown to run on CLR exit in .NET Core

### LINKED ISSUE #3711: CoordinatedShutdown should run automatically upon SIGTERM in .NET Core
Verison: Akka.NET v1.3.11

When I originally implemented `CoordinatedShutdown`, I left this little "TODO" in the source because .NET Standard 1.6 didn't expose an easy way for me to hook into process shutdown events, so I thought at least:

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Actor/CoordinatedShutdown.cs#L682

Found this issue while working on https://github.com/petabridge/Cluster.WebCrawler/issues/7

So, is there a way to implement this in .NET Standard 1.6 today? Or do we need to look at doing this within the confines of .NET Standard 2.0+?

Aaronontheweb:
close #3711


Aaronontheweb:
.NET Standard 2.0 supports the `AppDomain.ProcessExit` event, so we can just re-use that. Worked fine on .NET Framework.

Aaronontheweb:
failures thus far are a bunch of unrelated Akka.Streams specs that have been consistently racy - might give them the banhammer over the weekend.

Aaronontheweb:
Doesn't look like this is the "one true fix" - looks like there might be an issue with the `HashedWheelTimerScheduler` closing over the `ActorSystem` and not getting properly disposed, based on the output from dotMemory

PR #3846, opened 2019-07-12
unregister AppDomain.ProcessExit event handler after ActorSystem terminates

### LINKED ISSUE #3735: CoordinatedShutdown Memory Leak
Version: Akka.NET v1.3.11

Based on the investigation I did for issue #2640, I found that while the lion's share of leaked memory came from the `ForkJoinExecutor` and `PinnedDispatcher`s not being shutdown correctly as a result of an `IScheduler` bug, there was also a small amount of memory leaked as a result of some of the closures over the `ActorSystem` used inside the `CoordinatedShutdown` class as well.

I haven't fully investigated this yet, but I intend to make some headway on that ASAP in order to ensure that `ActorSystem`s don't hang around after being shutdown inside an Akka.NET process.

Aaronontheweb:
close #3735 - going to measure this and see if it worked


Aaronontheweb:
Doesn't look like this is the "one true fix" - looks like there might be an issue with the `HashedWheelTimerScheduler` closing over the `ActorSystem` and not getting properly disposed, based on the output from dotMemory

PR #3850, opened 2019-07-13
force racy Akka.Persistence specs to run serially

Aaronontheweb:


Aaronontheweb:
Ran the suite with these changes about 6-7 times now. Looks stable. Going to pull it in.

Issue #3851, opened 2019-07-14
Target .NET Standard 2.0 and newer only


Aaronontheweb:
Version: 1.4.0 beta and beyond

As part of the v1.4.0 release we are now targeting .NET Standard 2.0, instead of .NET Standard 1.6, and long-term we plan to stick with .NET Standard as the "guts" of Akka.NET.

However, we are also still dual-targeting .NET 4.5.2 - a version of .NET that is getting quite old and is missing some features (i.e. some of the `TaskCompletionSource` flags) that we currently have to hack around.

Going forward, I would like to propose the following roadmap for how we manage the versions of the .NET runtimes that Akka.NET targets:

- v1.4.0 - .NET Standard 2.0 only
- v1.4.* - .NET Standard 2.0 & .NET Core 3.0 - this would be done as a part of "spanifying" parts of the Akka.IO, Akka.Persistence, Akka Serialization, and Akka.Remote systems in order to improve performance. .NET Core 3.0 would enter into the picture as a means of gradually moving our "spanification" efforts down into the serialization and I/O layers, which require access to APIs that are not available in .NET Standard at the moment.
- v1.5.0 and beyond - .NET 5.0? Who knows.

Here's the strategic question we have to wrangle as maintainers: how do we balance getting the best possible experience for Akka.NET users with supporting our enterprise users, many of whom are stuck with .NET Framework for the time being?

Technical purity is usually the losing argument here - Akka.NET needs to create as wide of a tent as possible, but supporting .NET Framework 4.5 in perpetuity isn't sustainable long-term either. I might answer this question by polling the Petabridge mailing list, but I wanted to invite some open discussion on Github first.

poke:
I don't use Akka but I think your roadmap looks good. However, whenever targeting .NET Standard 2.0, I would also suggest to keep dual targeting the .NET Framework since that makes the consumption from the framework easier. You can push the version up to 4.7.2 or even 4.8 though.

From a business perspective, supporting one of the latest versions of the .NET Framework will also allow customers that still have to rely on the framework to plan a long term migration strategy without locking them out immediately.

netclectic:
I would have though there's no point explicitly supporting anything from 4.7.2 and above, as they are fully .net standard 2 compliant.

So, the question has to be how many existing Akka.Net users are there out there stuck somewhere between .NET 4.5.2 and  .NET 4.7.2? Find out the actual extent of the problem before coming to any conclusions.

[...]

PR #3862, opened 2019-07-19
Native support for dependency injection

Horusiath:
This is second attempt to add native dependency injection support for akka actors. This one is implemented end-to-end.

### How does it work?

In order to configure dependencies, you need to create your own instance of class called as `DependencyResolver` (it has to have a default constructor):

```csharp
public sealed class MyDependencyResolver : DependencyResolver
{
    public override void ConfigureDependencies(ExtendedActorSystem system, IServiceCollection services)
    {
        base.ConfigureDependencies(system, services);

        // configure your dependencies here...
        services.AddTransient<IService, MyService>();
        services.AddTransient<MyActor>();
    }
}

{
    public MyActor(IService service) // service to be injected
    {
          // actor configuration
    }
}

var config = ConfigurationFactory.ParseString( "akka.dependency-resolver-class = \"MyNamespace.MyDependencyResolver, MyAssembly\");
using (var system = ActorSystem.Create("system", config)))
{
    var props = Props.Create<MyActor>();
    var actor = system.ActorOf(props);
}
```

All, what's necessary is to configure it to be used by your actor system on startup via HOCON configuration:

```hocon
akka.dependency-resolver-class = "MyNamespace.MyDependencyResolver, MyAssembly"
```

With that, you can create actors with automatically resolved dependencies via `Props.Create<MyActor>()`. This also works in remote deployment, cluster router scenarios and cluster sharding scenarios.

#### Scopes

Microsoft dependency injection system defines 3 scope types: transient, scoped, singleton. While they may have some sense in stateless scenario - I guess design was targetting primarly the ASP.NET Core pipeline - they need to be interpreted separatelly in stateful apps:

- **Singleton** scope means that instance is singleton in scope of particular `ActorSystem`. Actor systems don't share their singletons, so each of them will have its own "singleton". This is a deliberate decision - actor systems are not supposed to share any data.
- **Scoped** means, its scope is bounded to ActorContext lifetime of a particular actor. That means, that the same instance will be kept alive even between actor restarts. May be useful for things like metrics tracking etc. Worth noting is that scoped won't survive over actor context - it's disposed on actor termination, so it won't survive reincrarnation of eg. persistent actors or entity migration between shards in cluster.sharding module.
- **Transient** scope is bound to an incarnation of a given actor - if actor will die, each time upon restart it will receive a new instance of transient dependency.

### Open questions

1. How to integrate this with existing frameworks? If I got it right if you configure deps for actor system and for eg. ASP.NET core runtime, they will live in two separate containers. But is this even a problem?


frenklin:
Might be a problem if someone needs a shared singleton between akka and something (WebHost for example). I can give you an example from prod app with 1 container shared between akka and WebHost. 

**How it works now.** 
1. create ActorSystem
2. add ActorSystem and other items into ServiceCollection
3. build service provider
4. register akka di resolver with service provider
5. get IWebHost from service provider and run it
6. get ActorSystem from service provider and run it

From my perspective perfect DI scenario is that ActorSystem and internals should be added to ServiceCollection by extension methods (for example .AddAkka(), .AddAkkaStreams(), .AddAkkaCluster() etc). And ActorSystem should read configuration from IOptions<T>. Then we can get ActorSystem from ServiceProvider and call Run(). I hope this may help you.

Horusiath:
While I understand `.AddAkka()` approach, I'm very much against `IOptions<T>`:

1. IMHO they are badly designed - are needlessly, require tons of dependencies.
2. IOption operates on static objects, while Akka operates on dynamic extensions: in practice is not possible to build a production-ready object type for akka, as it simply changes over time and can be even send over the network.
3. IOption (and ASP.NET) requires configuration class with default `new()` and getters/setters on all properties - this is exactly against design principles of Akka configs: which are immutable and extensible via `WithXYZ` methods with deep copy mechanics - this allows you to create many different configuration objects, which differ only in 1-few places in a single line of code.

ismaelhamed:
> Also IMHO primary use case of Akka.NET is not to be embedded inside ASP.NET core applications - its one of the use cases but honestly not the major one.

With the new GenericHost both ASP.NET and non-ASP.NET (Console/Windows Services) are converging on a single API for bootstrapping applications. So eventually, user will expect to be able to configure DI following the same guidelines everywhere (same goes with logging).

frenklin:
@Horusiath 

1. If I recall correctly IOptions has only one dep lib(Microsoft.Extensions.Options)
2. for dynamically changing configs you can try IOptionsMonitor or IOptionsSnapshot depends on use case. In common case scenario it's easy serializable POCOs and I don't see why it can't be send over the network.
3. I understand your concern. 

another cons it's not very friendly to another DI containers.

anyway even .AddAkka() huge step forward.

One possible broad use case it's a healthcheck for container orchestration for example api hc endpoint can send a message to some actor and check a response.

Aaronontheweb:
I'm on mobile and can't leave PR comments, but I have many.

Big picture though:

> With the new GenericHost both ASP.NET and non-ASP.NET (Console/Windows Services) are converging on a single API for bootstrapping applications.

https://blog.ploeh.dk/2014/05/19/conforming-container/

DI inside actors is something we've historically tried to discourage, because it adds unnecessary complexity in an environment where you have direct control over how objects are instantiated. If the goal is to simplify bootstrapping entire actor systems, we can solve that workout rewriting how actors work.

This is approach is also going to inevitably have bugs and historically it's been me that has to spend tens of hours fixing them -  I'm totally unconvinced that this is a good idea and therefore will not be supporting it unless my mind is changed.

Horusiath:
>If I recall correctly IOptions has only one dep lib(Microsoft.Extensions.Options)

Deps taken from [Nuget](https://www.nuget.org/packages/Microsoft.Extensions.Options/3.0.0-preview6.19304.6):

- Microsoft.Extensions.DependencyInjection.Abstractions (>= 3.0.0-preview6.19304.6)
- Microsoft.Extensions.Primitives (>= 3.0.0-preview6.19304.6)
    - System.Memory (>= 4.5.1)
    - System.Runtime.CompilerServices.Unsafe (>= 4.5.1)
- System.ComponentModel.Annotations (>= 4.6.0-preview6.19303.8)

Quite a lot for something that's supposed to give you an equivalent of `Func<T>`.
We have a long and painful history of having dependencies on projects that didn't take a proper care on their transitive dependencies, ultimatelly resulting in having dozens of libraries necessary to achieve the simples minimal program.

> For dynamically changing configs you can try IOptionsMonitor or IOptionsSnapshot depends on use case. In common case scenario it's easy serializable POCOs and I don't see why it can't be send over the network.

If POCOs are not always easily serializable. Fields can change while running two different versions of application in distribued cluster, you may want to have a validation on object creation etc. 

While this is not the topic here, I'd love to remove HOCON config dependency in actor system in the future and replace it with [IConfigurationRoot](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.iconfigurationroot?view=aspnetcore-2.2) which is able abstract over many different config formats (JSON, XML, HOCON, even command line args and environment variables).

src/contrib/cluster/Akka.Cluster.Sharding/Akka.Cluster.Sharding.csproj
```diff
@@ -5,9 +5,9 @@
     <AssemblyTitle>Akka.Cluster.Sharding</AssemblyTitle>
     <Description>Sharded actors with managed lifecycle for Akka.NET cluster</Description>
     <VersionSuffix>beta</VersionSuffix>
-    <TargetFrameworks>$(NetFrameworkLibVersion);$(NetStandardLibVersion)</TargetFrameworks>
```
Aaronontheweb: Need to resolve https://github.com/akkadotnet/akka.net/issues/3851 before doing this

src/contrib/cluster/Akka.Cluster.Sharding/Akka.Cluster.Sharding.csproj
```diff
@@ -5,9 +5,9 @@
...
     <PackageTags>$(AkkaPackageTags);network;cluster;sharding</PackageTags>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
Aaronontheweb: Need to use the parameterized values from common.props, and the attribute will still need to be `<TargetFrameworks>` since we'll likely be targeting multiple versions of .NET Standard or .NET Standard + .NET Core + .NET Framework, depending on https://github.com/akkadotnet/akka.net/issues/3851

src/core/Akka.Cluster.Tests/Akka.Cluster.Tests.csproj
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <AssemblyTitle>Akka.Cluster.Tests</AssemblyTitle>
-    <TargetFrameworks>$(NetFrameworkTestVersion);$(NetCoreTestVersion)</TargetFrameworks>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
Aaronontheweb: Same feedback as above - needs to be parameterized from `common.props` and kept with `<TargetFrameworks>` for all tests.

src/core/Akka.NodeTestRunner/Akka.NodeTestRunner.csproj
```diff
@@ -2,9 +2,9 @@
   <Import Project="..\..\common.props" />
   <PropertyGroup>
     <AssemblyTitle>Akka.NodeTestRunner</AssemblyTitle>
-    <TargetFrameworks>$(NetFrameworkTestVersion);$(NetCoreTestVersion)</TargetFrameworks>
     <OutputType>Exe</OutputType>
     <RuntimeIdentifiers>win7-x64</RuntimeIdentifiers>
+    <TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>
```
Aaronontheweb: Needs to be parameterized

src/core/Akka.Remote.Tests/RemoteDependencyResolverSpec.cs
```diff
@@ -0,0 +1,118 @@
+#region copyright
```
Aaronontheweb: Don't need a region for this

src/core/Akka.Tests/Configuration/ConfigurationSpec.cs
```diff
@@ -63,16 +63,16 @@ public void The_default_configuration_file_contain_all_configuration_properties(
...
-        {
-            var section = (AkkaConfigurationSection)System.Configuration.ConfigurationManager.GetSection("akka");
-            Assert.NotNull(section);
-            Assert.False(string.IsNullOrEmpty(section.Hocon.Content));
-            var akkaConfig = section.AkkaConfig;
-            Assert.NotNull(akkaConfig);
-        }
+#if NET461
+//        [Fact]
+//        public void Deserializes_hocon_configuration_from_net_config_file()
```
Aaronontheweb: Why is this spec commented out for .NET Framework?

src/core/Akka/Actor/ActorCell.FaultHandling.cs
```diff
@@ -67,9 +67,10 @@ private void FaultRecreate(Exception cause)
                     
                     failedActor.AroundPreRestart(cause, optionalMessage);
 
-                    // run actor pre-incarnation plugin pipeline
-                    var pipeline = _systemImpl.ActorPipelineResolver.ResolvePipeline(failedActor.GetType());
-                    pipeline.BeforeActorIncarnated(failedActor, this);
+                    if (failedActor is IActorStash stashed)
+                    {
+                        stashed.Stash.UnstashAll();
```
Aaronontheweb: Is the idea here to include automatic stash / unstashing upon actor restart?

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -452,7 +449,6 @@ protected void ClearActor(ActorBase actor)
                     }
                 }
 
-                ReleaseActor(actor);
```
Aaronontheweb: In the event of injected dependencies inside an actor, what's responsible for cleaning up those dependencies? The `scope.Dispose()` call earlier?

src/core/Akka/Actor/DependencyResolver.cs
```diff
@@ -0,0 +1,40 @@
...
+            return new DefaultServiceProviderFactory();
+        }
+
+        /// <summary>
+        /// Extension point, where user can configure service dependencies, which will be injected during actor creation
+        /// via dynamic Props model.
+        /// </summary>
+        public virtual void ConfigureDependencies(ExtendedActorSystem system, IServiceCollection services)
+        {
+            services.AddSingleton<ActorSystem>(system);
```
Aaronontheweb: So this will limit each `IServiceCollection` to a single `ActorSystem`, correct?

src/core/Akka/Actor/ActorCell.FaultHandling.cs
Horusiath: This is not my idea - I'm just keeping the original behavior ;)

src/core/Akka/Actor/ActorCell.cs
Horusiath: Yes.

src/core/Akka/Actor/DependencyResolver.cs
Horusiath: Yes, each ActorSystem gets its own unique `IServiceCollection` instance.

src/contrib/cluster/Akka.Cluster.Sharding/Akka.Cluster.Sharding.csproj
Horusiath: I did it mostly because Microsoft.Extensions.DependencyInjection is not compatible with .NET Framework pre 4.6.1. I can set it back up.

Aaronontheweb: My recommendation: if we're going to drop .NET framework 4.5.2 (good chance) definitely do it in a separate PR - don't couple it in this one

src/core/Akka/Actor/ActorCell.cs
Aaronontheweb: Ok, and that's implemented by the DI container itself right?

Horusiath: I can write tests to ensure it is.

src/core/Akka/Actor/ActorCell.FaultHandling.cs
Aaronontheweb: Not in the Akka.NET codebase - open a new PR for this.

Horusiath: See: https://github.com/akkadotnet/akka.net/pull/3862/files#diff-a2c304463bb7ef467388df75c265bbfbL63

Aaronontheweb: I stand corrected! You're right.

https://github.com/akkadotnet/akka.net/blob/cf0cb8c1daffc68b8673a491759416cae47f6af1/src/core/Akka.Tests/Actor/Stash/ActorWithStashSpec.cs#L90-L111

Horusiath:
I've reverted target framework changes, but now there's an issue:

>   ActorSystemImpl.cs(457, 68): [CS0246] The type or namespace name 'ServiceCollection' could not be found (are you missing a using directive or an assembly reference?)
  DependencyResolver.cs(27, 24): [CS0246] The type or namespace name 'DefaultServiceProviderFactory' could not be found (are you missing a using directive or an assembly reference?)

It looks like `Microsoft.Extensions.DependencyInjection` even tough has compilation targets for both .NET Standard and .NET Framework, in the latter case has undefined some classes.

Horusiath:
@Aaronontheweb I've addressed your remarks. The build fails because I had to increase .NET Framework version (AFAIK new microsoft stuff doesn't support 4.6.1) while some CI scripts explicitly run on 4.6.1 (if I'm right).

Cephei:
Is this pullreq dead? I'm curious what the prevailing thoughts concerning Akka.net DI are.

Aaronontheweb:
@Cephei this isn't really a high priority for our team right now. Actually doing this and working through all of the bugs is a significant undertaking and I haven't been convinced that initial and ongoing maintenance costs are justified.

andersstorhaug:
What's the recommended approach for remote deployment of actors, and DI with non-serializable dependencies (IoC container or not)?

Current approach of mine is to use a static factory..

```csharp
public class RemoteActor : ReceiveActor
{
    public static Func<(ISomeDependency, ISomeOtherDependency)> Inject;
}
```

```csharp
// In Program.cs / composition root

RemoteActor.Inject = () => (someDependency, someOtherDependency);
```

Is there a better pattern for this, besides using a service-locator pattern?

Aaronontheweb:
Got about as far as we should for this via #4590

PR #3863, opened 2019-07-20
Add timeout to abort joining of seed nodes

ismaelhamed:
Port [#23551](https://github.com/akka/akka/pull/23551)

### DIFF
--- a/src/core/Akka.Cluster/ClusterDaemon.cs
+++ b/src/core/Akka.Cluster/ClusterDaemon.cs
@@ -13,7 +13,6 @@ using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Dispatch;
 using Akka.Event;
-using Akka.Pattern;
 using Akka.Remote;
 using Akka.Util;
 using Akka.Util.Internal;
@@ -1003,6 +1002,7 @@ namespace Akka.Cluster
         private ImmutableList<Address> _seedNodes;
         private IActorRef _seedNodeProcess;
         private int _seedNodeProcessCounter = 0; //for unique names
+        private Deadline _joinSeedNodesDeadline;
 
         private readonly IActorRef _publisher;
         private int _leaderActionCounter = 0;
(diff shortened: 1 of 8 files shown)


ismaelhamed:
@Aaronontheweb there's some documentation related to these new settings, but the Akka.NET documentation is so different in this topic (cluster) that I'm not sure where to put it.

Aaronontheweb:
@ismaelhamed do the best you can to add it to this PR, and I'll help you out if need be

ismaelhamed:
@Aaronontheweb it's part of the [Joining To Seed Nodes](https://doc.akka.io/docs/akka/current/cluster-usage.html#joining-to-seed-nodes) section in the **Cluster Usage** document. But the whole thing is missing. Maybe we're a better off adding another PR with all the missing documentation related to the cluster later on.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1003,6 +1002,7 @@ internal static string VclockName(UniqueAddress node)
         private ImmutableList<Address> _seedNodes;
         private IActorRef _seedNodeProcess;
         private int _seedNodeProcessCounter = 0; //for unique names
+        private Deadline _joinSeedNodesDeadline;
```
Horusiath: Call me old fashioned, but I'd prefer it to be a `CancellationToken` instead - this way we could also use it inside `await Cluster.JoinSeedNodesAsync(nodes, token)`

Horusiath: Just like any other argument passed there - as a field in the message. This would need a shift from the canonical Scala implementation, but IMO they don't support any cancellation of calls whatsoever, while this is:

1. Pretty much obvious in .NET.
2. Allows you to specify more interesting patterns like retrying to join with widening/narrowing timeout window - ofc this can be done with current design too, but it would require recreating ActorSystem with different config parameter value each time.

ismaelhamed:
@Aaronontheweb not sure where all those errors are coming from, because the unit tests are passing just fine in my machine.

Aaronontheweb:
@ismaelhamed it's usually due to all of the blocking calls inside the `TestKit` and that we're running many different specs concurrently. It's a regular nuisance

src/core/Akka.Cluster/ClusterDaemon.cs
ismaelhamed: Maybe this could be improved upon a later PR.

ismaelhamed:
@Aaronontheweb  LOL... I thought I'd never see this PR green.

Aaronontheweb:
weird, got a compilation failure this time? wtf

ismaelhamed:
> @ismaelhamed it's usually due to all of the blocking calls inside the `TestKit` and that we're running many different specs concurrently. It's a regular nuisance

Would automerge be an option for approved pull requests?

Aaronontheweb:
@ismaelhamed I have to review stuff sometimes when these specs fail just to be sure we didn't make the problem worse. We have one PR open where the MNTR has failed in the same location every time we've run it, even though there's nothing obvious in the PR that would have affected it. That's why we're a but judicious about combing through these before merging.

ismaelhamed:
> @ismaelhamed I have to review stuff sometimes when these specs fail just to be sure we didn't make the problem worse. We have one PR open where the MNTR has failed in the same location every time we've run it, even though there's nothing obvious in the PR that would have affected it. That's why we're a but judicious about combing through these before merging.

Agree. It's just that the build system seems to frequently fail (regular nuisance like you said), and that makes some PR stale for a long time --I opened this PR in July and have force-pushed 8 times since then.

Aaronontheweb:
Yeah, that's a nuisance - we need to get some changes implemented in the `TestKit` and in the way our test suites themselves run in order to alleviate this.

PR #3864, opened 2019-07-21
Akka.NET v1.3.14 backport

Aaronontheweb:
An attempt to take the non-breaking, but useful changes from the v1.4.0-beta1 milestone and release them into a new stable release of Akka.NET based on the v1.3.13 runtime.


Aaronontheweb:
Removed some reverted commits from the history here - hence the force-push

Aaronontheweb:
Have compilation and NuGet packaging back again. That's a good start.

Aaronontheweb:
Ok, compile + run for all testing stages should work now minus .NET Core on Linux. Don't know what the issue is there but I'm not going to keep chasing it. Just want to see a full test run and make sure none of these changes are dependent on .NET Standard 2.0.

Aaronontheweb:
MNTR looks like it's working

PR #3869, opened 2019-07-26
V1.3.14 backport2

Aaronontheweb:
Replaces https://github.com/akkadotnet/akka.net/pull/3864


Aaronontheweb:
Need to bring https://github.com/akkadotnet/akka.net/pull/3864/commits/a946059abaa51a2aa5e10816bb25a6a77f5cd891 into this PR

Aaronontheweb:
So the entire test suite is running in parallel due to `dotnet test` and XUnit not receiving the flag to disable it. I can fix this via an `xunit.runner.json` file, but for the time being I'm going to try to run the entire test suite end to end.

But everything looks safe for release despite that - I don't want to port over all of my racy test fixes since I think that will generate some noise.

Aaronontheweb:
Squashed all of my "edit" commits that it took in order to get the build system to run on .NET Core 1.1 / .NET Framework 4.5 - hence the force-push.

Aaronontheweb:
Ok, this should do it - https://github.com/akkadotnet/akka.net/pull/3872 - this will get `dev` and `master` back into clean states once this is merged.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
```diff
@@ -373,11 +374,12 @@ public void AddPersistenceIdSubscriber(IActorRef subscriber, string persistenceI
         {
             if (!_persistenceIdSubscribers.TryGetValue(persistenceId, out var subscriptions))
             {
-                subscriptions = new HashSet<IActorRef>();
-                _persistenceIdSubscribers.Add(persistenceId, subscriptions);
+                _persistenceIdSubscribers = _persistenceIdSubscribers.Add(persistenceId, ImmutableHashSet.Create(subscriber));
+            }
+            else
+            {
+                _persistenceIdSubscribers = _persistenceIdSubscribers.Add(persistenceId, subscriptions.Add(subscriber));
```
tris2481, line 381: I think this needs to be SetItem or you get errors when adding more than one subscriber, see AddTagSubscriber where SetItem is used for this reason.

PR #3872, opened 2019-07-30
V1.3.14 sync

Aaronontheweb:


Aaronontheweb:
Looks like I messed something up with the API Specs - NRE issue reading the files. I'll need to troubleshoot that before we can sync `dev` with `master` on this PR.

Aaronontheweb:
Had some racy test failure, but otherwise things look good. Going to let the MNTR finish on .NET Framework first.

PR #3878, opened 2019-08-06
Update IotApp.cs sample code for tutorial 1 documentation.

ondrelogin:
edit the code sample of the main app to properly utilize the "...recommended creational pattern for actors; define a static Props() method".
i.e. using the static Props method instead of Props.Create.

I read the Contributing guide for Akka.NET after I made the edits in the browser so I'm not sure if the edits are inline with the guide. If not, I can issue a new, more correct, pull request.


Danthar:
@ondrelogin it looks fine. Thx for the contribution :)

Issue #3877, opened 2019-08-06
Akka.Persistence.TestKit implementation


Aaronontheweb:
Ported from https://github.com/petabridge/Akka.Persistence.Extras/issues/58

# Akka.Persistence.TestKit Requirements
This is an outline for an "Akka.Persistence.TestKit" NuGet package that can be used by Akka.NET developers to help test for possible errors that may come up in the course of using Akka.Persistence, such as:

1. Rejected messages due to serialization errors;
2. Write failures due to the journal or snapshot store being offline; and
3. Recovery errors as a result of serialization or connectivity issues.

These fake journal and snapshot store implementations should be programmable, so if we take a typical [Arrange-Act-Assert pattern](http://wiki.c2.com/?ArrangeActAssert) and factor our "FailingJournal" or "FailingSnapshotStore" into it:

```
public class MyTestKitClass : Akka.TestKit.Xunit2.Testkit{
	private readonly FailingJournal _failingJournal

	public MyTestKitClass() 
		: base(FailingJournal.DefaultConfig()) // need FailingJournal HOCON 
	{ 
		// need to grab an instance of the FailingJournal for programming purposes
		_failingJournal = PersistenceExt.Get(Sys);
	}

	public async Task MyFakeUnitTest(){
		// arrange
		var myPersistentActor = Sys.ActorOf(Props.Create(() => new MyPersistActor()), "p");
		myPersistentActor.Tell("write"); // write one message to journal while it's still working
		ExpectMsg("ack"); // message gets processed

		// terminate actor
		await myPersistentActor.GracefulStop();		

		// act

		await _failingJournal.BeginFailingWrites(); // simulate journal network disconnect		

		// assert
			
		// re-create a persistent actor - should fail
		EventFilter.Exception<RecoveryException>().ExpectOne(() => {
			var myPersistentActor2 = Sys.ActorOf(Props.Create(() => new MyPersistActor()), "p");
			
			// actor should stop as a result of recovery failure
			Watch(myPersistentActor);
			ExpectTerminated(myPersistentActor);
		});
	}
}

```

That's the general idea - make it easy to progam the behavior the journal and shapshot store in Akka.Persistence prior to actually using it.

## Lexicon and Terminology
To make it easier to understand the requirements going forward, here are some technical terms we're going to use throuhout this document:

* `FailingJournal` - this is the journal that can be programmed to fail altogether or reject persistent messages.
[...]

Issue #3854, opened 2019-07-16
API proposal: rewrite blocking elements of TestKit to return awaitable Tasks


Aaronontheweb:
Version: Akka.NET 1.5.0 and later

As part of the Akka.NET v1.4..0 sprint I've spent _a lot_ of time going through our test suite and disabling parts of it that are especially timing-sensitive. I think the reasons for most of these races boils down to the following:

1. On Azure DevOps we can't be sure of the size and load of the hosted build agent pool. I haven't been able to find out for certain looking through the docs but I assume agents are single tenant when they're actually running a live job (not running jobs for multiple tenants in parallel.)
2. Many of the operations inside the testkit explicitly block threads, i.e. `TestLatch.Ready`, `ExpectMsg`, `AwaitCondition`, etc.
3. We do want to block _the flow of execution_ in our tests, for sure - that's how we make it sane and easy to write them. However, the fact that there are blocked threads means that an assertion like `ExpectMsg<T>(TimeSpan.FromSeconds(300))` can succeed or fail depending on how many cores are available and wait the blocking state of the work queue looks like.
4. If these were written to use `async` and `await` instead, it could work with XUnit's ability to take `Task` off of the test method and would allow for background threads with work to do to run while foreground threads that are waiting on results are also unblocked.
5. I think this would help make our test suite much less racy overall and make the TestKit more reliable.

However, this comes with some flies in the ointment:

1. More cruft to deal with inside the writing of each test - now we have to propagate `Task` everywhere.
2. XUnit supports returning `Task` off the top of each of its test methods, but what about NUnit, MsTest, and other test frameworks that we've supported historically? Where does that leave them?
3. It's still possible that we could make all of these changes and not see much improvement in the test suite - because there are likely other factors at work that I'm discounting, such as how much work other processes are plying onto the build agent during our test runs.

If we were going to introduce this, here is what I'd propose doing:

1. Leave all existing public TestKit APIs as-is for backwards compatibility, mark them as deprecated.
2. Add Async versions that return `Task` - don't need to rename the method necessarily with the `Async` prefix, just need to add some overloads that return them. 
3. Replace the default method implementations to all call down to the async versions of these.

Either way, wanted to throw this idea out there for consideration in Akka.NET 1.5.0 and later. What are your thoughts?

Horusiath:
Honestly, I think we could *greatly* simplify the testkit API, like eg.

```csharp
// equivalent of probe.ExpectNext
[...]

PR #3671, opened 2018-12-14
upgraded to JSON.NET 12.0.2

Aaronontheweb:

PR #3881, opened 2019-08-08
TestJournal and TestSnaphostStore

valdisz:
PR addresses issue #3877

Added 3 new projects:
* `Akka.Persistence.TestKit`
* `Akka.Persistence.TestKit.Tests`
* `Akka.Persistence.TestKit.Xunit2`

It adds a new journal (`TestJournal`) and a new snapshot store (`TestSnaphostStore`). It is possible to intercept calls to them and apply user-defined behavior, i.e. to simulate network problems or anything else.

It adds new test base class `PersistenceTestKit` (Xunit2 based) to simplify usage. API loos like this:

``` csharp
public class PersistentActorSpec : PersistenceTestKit
{
    [Fact]
    public async Task actor_must_fail_when_journal_will_fail_saving_message()
    {
        await WithJournalWrite(write => write.Fail(), () =>
        {
            var actor = ActorOf(() => new PersistActor());
            Watch(actor);

            actor.Tell("write", TestActor);
            ExpectTerminated(actor);
        });
    }
}
```


valdisz:
This is related to #3877

Aaronontheweb:
Build failed due to `TestJournalSpec.cs(28,13): error CS0165: Use of unassigned local variable 'journal' [D:\a\1\s\src\core\Akka.Persistence.TestKit.Tests\Akka.Persistence.TestKit.Tests.csproj]`

src/Akka.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2010
+# Visual Studio Version 16
```
Aaronontheweb: Should probably bump the version of Visual Studio tooling we're using in a separate PR, but I'm open to doing that.

src/core/Akka.Persistence.TestKit.Tests/Akka.Persistence.TestKit.Tests.csproj
```diff
@@ -0,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\common.props" />
+
+  <PropertyGroup>
+    <AssemblyTitle>Akka.Persistence.TestKit.Tests</AssemblyTitle>
+    <TargetFrameworks>$(NetFrameworkTestVersion);$(NetCoreTestVersion)</TargetFrameworks>
```
Aaronontheweb: Good use of `common.props` here - appreciate that.

src/core/Akka.Persistence.TestKit.Tests/test-journal.conf
```diff
@@ -0,0 +1,13 @@
+akka {
+  persistence {
+    journal {
+      plugin = "akka.persistence.journal.test"
+      auto-start-journals = ["akka.persistence.journal.test"]
+
+      test {
+        class = "Akka.Persistence.TestKit.TestJournal, Akka.Persistence.TestKit"
```
Aaronontheweb: Looks good

src/core/Akka.Persistence.TestKit/IJournalWriteInterceptor.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Akka.Persistence.TestKit
+{
+    using System.Threading.Tasks;
+
+    public interface IJournalWriteInterceptor
```
Aaronontheweb: I like this idea - having a generic action that can be used to intercept a common journal operation, like a Recover or Write.

src/Akka.sln
valdisz: I will revert that line. I committed it accidentally.

src/core/Akka.Persistence.TestKit/JournalInterceptors.cs
```diff
@@ -0,0 +1,99 @@
+﻿namespace Akka.Persistence.TestKit
+{
+    using System;
+    using System.Threading.Tasks;
+
+    internal static class JournalInterceptors
+    {
+        internal class Noop : IJournalInterceptor
```
Aaronontheweb: Would you mind adding some XML-DOC comments to explain what each of these do just to help other OSS users know in the future?

src/core/Akka.Persistence.TestKit/JournalInterceptors.cs
```diff
@@ -0,0 +1,99 @@
+﻿namespace Akka.Persistence.TestKit
+{
+    using System;
+    using System.Threading.Tasks;
+
+    internal static class JournalInterceptors
```
Aaronontheweb: Overall, I think the set of these are elegantly simple - I like the look of it and I like the idea that they're extensible in the future. I think the design is good @valdisz

src/core/Akka.Persistence.TestKit/JournalInterceptors.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        internal class Rejection : IJournalInterceptor
+        {
+            public static readonly IJournalInterceptor Instance = new Rejection();
+
+            public Task InterceptAsync(IPersistentRepresentation message) => throw new TestJournalRejectionException(); 
+        }
+        
+        internal class Delay : IJournalInterceptor
```
Aaronontheweb: Is the goal of this `IJournalInterceptor` to simulate latency?

src/core/Akka.Persistence.TestKit/TestJournal.cs
```diff
@@ -0,0 +1,138 @@
...
+    using System.Threading.Tasks;
+
+    public sealed class TestJournal : MemoryJournal
+    {
+        private IJournalInterceptor _writeInterceptor = JournalInterceptors.Noop.Instance;
+        private IJournalInterceptor _recoveryInterceptor = JournalInterceptors.Noop.Instance;
+
+        protected override bool ReceivePluginInternal(object message)
+        {
+            switch (message)
```
Aaronontheweb: Makes sense

src/core/Akka.Persistence.TestKit/TestJournal.cs
```diff
@@ -0,0 +1,138 @@
...
+                    _recoveryInterceptor = use.Interceptor;
+                    Sender.Tell(Ack.Instance);
+                    return true;
+                
+                default:
+                    return base.ReceivePluginInternal(message);
+            }
+        }
+
+        protected override async Task<IImmutableList<Exception>> WriteMessagesAsync(IEnumerable<AtomicWrite> messages)
```
Aaronontheweb: Looks good to me

src/core/Akka.Persistence.TestKit/TestJournal.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            var highest = HighestSequenceNr(persistenceId);
+            if (highest != 0L && max != 0L)
+            {
+                var messages = Read(persistenceId, fromSequenceNr, Math.Min(toSequenceNr, highest), max);
+                foreach (var p in messages)
+                {
+                    try
+                    {
+                        await _recoveryInterceptor.InterceptAsync(p);
```
Aaronontheweb: LGTM

src/core/Akka.Persistence.TestKit/JournalInterceptors.cs
valdisz: Yes, like if there is some retry strategy which will make several attempts before fail.

It was eady to add...

src/core/Akka.Persistence.TestKit/JournalInterceptors.cs
valdisz: Definitely. I want to finish first with snapshots
 Because most probably I will refactor something so that it all looks as complete picture.

valdisz:
@Aaronontheweb I have implemented `PersistenceTestKit`. Is it what you were expecting?

valdisz:
One test in a totally different project (which was not modified) is failing. No idea why.

Aaronontheweb:
@valdisz we have a number of test failures that are due to concurrent tests with hard deadlines blocking threads, We have an open proposal to fix this by moving everything to an `async` / `await` model in https://github.com/akkadotnet/akka.net/issues/3854  - we're going to have to address it in a future release. Don't worry about it for the time being.

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
```diff
@@ -0,0 +1,35 @@
...
+//     Copyright (C) 2009-2019 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2019 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+namespace Akka.Persistence.TestKit
+{
+    using Actor;
+    using Akka.TestKit;
+    using Akka.TestKit.Xunit;
```
Aaronontheweb: For the benefit of other Akka.NET OSS users: we're taking an explicit dependency on XUnit here. This means that we're not going to be supporting NUnit as part of this feature. If this creates an issue for you, speak now or forever hold your peace.

src/core/Akka.Persistence.TestKit.Tests/TestJournalSpec.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        [Fact]
+        public async Task during_recovery_by_setting_fail_will_cause_recovery_failure()
+        {
+            var actor = Sys.ActorOf<PersistActor>();
+            await actor.Ask("write");
+            await actor.GracefulStop(TimeSpan.FromSeconds(3));
+
+            WithFailingJournalRecovery(() =>
```
Aaronontheweb: @valdisz so this is an example of what I'm looking for - I've abstracted over the `ITestJournal` and made it so the end-users who are trying to simulate failures on the TestKit have to know the bare minimum about the underlying implementation. This is what I'd like to see more of - does that make sense?

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            JournalActorRef = GetJournalRef(Sys);
+            Journal = TestJournal.FromRef(JournalActorRef);
+        }
+
+        public IActorRef JournalActorRef { get; }
+
+        public ITestJournal Journal { get; }
+
+        public void WithFailingJournalRecovery(Action execution)
```
Aaronontheweb: @valdisz here's an example of this method on the implementation side -we should expose some more of these for both the TestJournal and the TestSnapshotstore. We can implement the latter in a separate pull request - but does this make sense?

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
```diff
@@ -0,0 +1,64 @@
...
+                // restore normal functionality
+                Journal.OnRecovery.Pass();
+            }
+        }
+
+        public void WithFailingJournalWrites(Action execution)
+        {
+            try
+            {
+                Journal.OnWrite.Fail();
```
Aaronontheweb: IMHO, we should probably get into the habit of returning `Task` here instead, in order to avoid the types of issues I dsecribe in https://github.com/akkadotnet/akka.net/issues/3854 going forward

src/core/Akka.Persistence.TestKit/JournalWriteBehavior.cs
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        public void RejectIf(Func<IPersistentRepresentation, bool> predicate)
+        {
+            if (predicate is null)
+            {
+                throw new ArgumentNullException(nameof(predicate));
+            }
+
+            SetInterceptor(new JournalInterceptors.OnCondition(predicate, JournalInterceptors.Rejection.Instance));
```
Aaronontheweb: Let's have `SetInterceptor` return a `Task` and have all of these methods `await` on it

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
Aaronontheweb: Likewise, this method should return `Task` too given the previous comment

src/core/Akka.Persistence.TestKit/JournalWriteBehavior.cs
valdisz: This make sense...

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
valdisz: This is clear, I agree that it will be cleaner.

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
valdisz: I will add a base class with assertions `ITestKitAssertions assertions` available for modification, but assembly will still reference `Akka.TestKit.Xunit2` unless there is 3rd assembly with integration.

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
valdisz: End-user always must do their assertions inside `execution` callback. Otherwise, it could lead to unexpected results when Journal behavior is reverted, but actor under test have not to even tried to persist anything into Journal.

valdisz: @Aaronontheweb how about such API?

``` csharp
public async Task WithJournalRecovery(Func<JournalRecoveryBehavior, Task> behaviorSelector, Func<Task> execution)
{
    try
    {
        await behaviorSelector(Journal.OnRecovery);
        await execution();
     }
    finally
    {
        await Journal.OnRecovery.Pass();
    }
}
```

and use it like this:

``` csharp
[Fact]
public async Task new_api()
{
    await WithJournalRecovery(recovery => recovery.Fail(), () =>
    {
        // code to trigger actor under test
        // and assertions
        return Task.CompletedTask;
    });
}
```

src/core/Akka.Persistence.TestKit/JournalRecoveryBehavior.cs
```diff
@@ -0,0 +1,194 @@
...
+    public class JournalRecoveryBehavior
+    {
+        internal JournalRecoveryBehavior(IJournalBehaviorSetter setter)
+        {
+            this.Setter = setter;
+        }
+
+        protected readonly IJournalBehaviorSetter Setter;
+
+        public Task SetInterceptorAsync(IJournalInterceptor interceptor) => Setter.SetInterceptorAsync(interceptor);
```
valdisz: `SetInterceptorAsync` now is `public` so that it is possible to have custom interceptors.

Aaronontheweb: Nice idea

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
Aaronontheweb: > End-user always must do their assertions inside execution callback. Otherwise, it could lead to unexpected results when Journal behavior is reverted, but actor under test have not to even tried to persist anything into Journal.

Yep, I agree - that's what we should teach them to do in the documentation.

Aaronontheweb: I like the API design too - if our goal is to avoid a blocking testkit going forward, returning `Task` is the way to go.

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
Aaronontheweb: Sounds like a great compromise.

Aaronontheweb:
@valdisz this is looking good - what are next steps for implementing this? More overloads on the `PersistenceTestKit` class?

valdisz:
@Aaronontheweb more unit tests for existing code (some are still needed), then SnaphsotStore and then XML DOC.

src/core/Akka.Persistence.TestKit/PersistenceTestKit.cs
valdisz: `Akka.Persistence.TestKit.Xunit2`... but it will have just one class (`PersistenceTestKit`) 🤔

src/Akka.sln
Danthar: For the record. This line does nothing. Its only a hint/marker for visual studio. If you commit this, and you open the sln in vs2015 it will change this line back automatically. I opened multiple bug reports with the VS team about this annoyance, and for some reason, its reintroduced again and again after a few VS updates.

valdisz: good to know, tnx

src/core/Akka.Persistence.TestKit.Tests/TestSnapshotStoreSpec.cs
```diff
@@ -0,0 +1,94 @@
...
+        [Fact]
+        public async Task after_load_behavior_was_executed_store_is_back_to_pass_mode()
+        {
+            // create snapshot
+            var actor = ActorOf(() => new SnapshotActor(_probe));
+            actor.Tell("save");
+            _probe.ExpectMsg<SaveSnapshotSuccess>();
+            await actor.GracefulStop(TimeSpan.FromSeconds(3));
+
+            await WithSnapshotLoad(load => load.Fail(), () =>
```
Aaronontheweb: This is exactly the sort of API I was looking for - excellent

src/core/Akka.Persistence.TestKit.Xunit2/Akka.Persistence.TestKit.Xunit2.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\common.props" />
+
+  <PropertyGroup>
+    <AssemblyTitle>Akka.Persistence.TestKit.Xunit2</AssemblyTitle>
+    <Description>TestKit for writing tests for Akka.NET Persistance module using xUnit</Description>
+    <TargetFrameworks>$(NetFrameworkLibVersion);$(NetStandardLibVersion)</TargetFrameworks>
+    <PackageTags>$(AkkaPackageTags);testkit;persistance;xunit</PackageTags>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NetStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netstandard1.6' ">1.6.1</NetStandardImplicitPackageVersion>
```
Aaronontheweb: You can delete this line - we don't need it

src/core/Akka.Persistence.TestKit.Xunit2/Akka.Persistence.TestKit.Xunit2.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <ProjectReference Include="..\Akka.TestKit\Akka.TestKit.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <DefineConstants>$(DefineConstants);RELEASE</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- WORKAROUND: for some reason starting at Akka.NET 1.3.2 this package was determined as "unpackable" by default via DOTNET CLI -->
+    <IsPackable>true</IsPackable>
```
Aaronontheweb: Good, need that for any publish-able NuGet package that takes a dependency on XUnit.

src/core/Akka.Persistence.TestKit.Xunit2/PersistenceTestKit.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Akka.Persistence.TestKit
+{
+    using Akka.TestKit.Xunit2;
+
+    /// <summary>
+    /// This class represents an Akka.NET Persistence TestKit that uses <a href="https://xunit.github.io/">xUnit</a>
+    /// as its testing framework.
+    /// </summary>
+    public abstract class PersistenceTestKit : PersistenceTestKitBase
```
Aaronontheweb: Need to add a constructor overload here which takes an `ITestOutputHelper` for XUnit - that's how we're able to inject the XUnit output capture system into the Akka.NET logging infrastructure. Here's an example of a unit test where a user would do that:

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote.Tests/RemotingSpec.cs#L31

src/core/Akka.Persistence.TestKit/Akka.Persistence.TestKit.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\common.props" />
+
+  <PropertyGroup>
+    <AssemblyTitle>Akka.Persistence.TestKit</AssemblyTitle>
+    <Description>TestKit for writing tests for Akka.NET Persistance module.</Description>
+    <TargetFrameworks>$(NetFrameworkLibVersion);$(NetStandardLibVersion)</TargetFrameworks>
+    <PackageTags>$(AkkaPackageTags);testkit;persistance</PackageTags>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NetStandardImplicitPackageVersion Condition=" '$(TargetFramework)' == 'netstandard1.6' ">1.6.1</NetStandardImplicitPackageVersion>
```
Aaronontheweb: Can get rid of this line here too

src/core/Akka.Persistence.TestKit/Journal/JournalRecoveryBehavior.cs
```diff
@@ -0,0 +1,419 @@
...
+        /// <param name="interceptor">User defined interceptor which implements <see cref="IJournalInterceptor"/> interface.</param>
+        /// <exception cref="ArgumentNullException">When <paramref name="interceptor"/> is <c>null</c>.</exception>
+        public Task SetInterceptorAsync(IJournalInterceptor interceptor)
+        {
+            if (interceptor == null) throw new ArgumentNullException(nameof(interceptor));
+
+            return Setter.SetInterceptorAsync(interceptor);
+        }
+
+        /// <summary>
```
Aaronontheweb: XML-DOC comments look good throughout this file.

src/core/Akka.Persistence.TestKit/PersistenceTestKitBase.cs
```diff
@@ -0,0 +1,290 @@
...
+    using System;
+    using System.Threading.Tasks;
+    using Actor;
+    using Akka.TestKit;
+    using Configuration;
+
+    /// <summary>
+    /// This class represents an Akka.NET Persistence TestKit base class which is testing framework agnostic.
+    /// </summary>
+    public abstract class PersistenceTestKitBase : TestKitBase
```
Aaronontheweb: The API looks good and simple. Nice job.

src/core/Akka.Persistence.TestKit.Xunit2/PersistenceTestKit.cs
valdisz: To correctly handle `ITestOutputHelper` in test kit I changed the base class of `PersistenceTestKit` to `TestKit`. This means that `PersistenceTestKitBase` no longer exists other test frameworks will not have Test Kit class with persistence testing API out of the box for now.

valdisz:
PR is ready, all work is done. If any issues are found, please comment, otherwise, merge.

Danthar:
We got some errors in Tests, in the MNTK tests related to persistence are failing with casting errors. However as far as i can tell that should not be code this PR changed. 

Did we change an IList to an SortedSet somewhere in this PR ?

valdisz:
> We got some errors in Tests, in the MNTK tests related to persistence are failing with casting errors. However as far as i can tell that should not be code this PR changed.
> 
> Did we change an IList to an SortedSet somewhere in this PR ?

This PR does not change anything except adding 3 new projects to the solution.

Danthar:
Yeah I could not find it either. Hence my question. Just checking if I'm
not crazy. So this means a PR was merged which caused problems. Odd that ci
didn't catch it. I got some time tomorrow ill see if I can find the
offending PR. That is if no one else beats me to it.

Op zo 25 aug. 2019 22:09 schreef Valdis Zobēla <notifications@github.com>:

Danthar:
Think i found the issue. Problem happens during json serialization/deserialization. Im thinking this PR is the culprit https://github.com/akkadotnet/akka.net/pull/3671 Because i cannot find anything else that might cause this.

valdisz:
Surprisingly my next PR does not fail any builds.

Danthar:
That is because we have a system. Which detects what code is affected and
then only builds and runs the projects with the affected code. Cuts back on
ci time. Since your other PR is a docs only PR. Only the docs project is
checked.

Op ma 26 aug. 2019 12:50 schreef Valdis Zobēla <notifications@github.com>:

PR #3883, opened 2019-08-12
standardized Google.Protobuf version to 3.9.1

Aaronontheweb:


src/common.props
```diff
@@ -14,6 +14,7 @@
     <HyperionVersion>0.9.8</HyperionVersion>
     <NewtonsoftJsonVersion>12.0.2</NewtonsoftJsonVersion>
     <NBenchVersion>1.2.2</NBenchVersion>
+    <ProtobufVersion>3.9.1</ProtobufVersion>
```
Aaronontheweb: Added new entry in `common.props` to standardize Google.Protobuf versions across Akka.Persistence, Akka.Remote, and Akka.Streams.

Akka.Cluster and other packages that depend on Google.Protobuf draw their versions from one of these three dependencies.

Aaronontheweb:
Azure Pipelines appears to be having trouble picking up the latest changes - seen that on a couple of PRs this morning. Re-running this via manual kickoff.

Danthar:
So we went from 3.3.0 to 3.9.1. Went through the changes for the library, in so far i could find them in their official docs, i didn't do a git based comparison. And so far i found the following changes for the c# library:

  * Implemented IReadOnlyDictionary<K,V> in MapField<K,V>
  * Added TryUnpack method for Any message in addition to Unpack.
  * Extension ranges now accept options and are customizable.
  * "reserve" keyword now supports “max” in field number ranges,
    e.g.  reserve 1000 to max;
  * Added unknown field support in JsonParser.
  * Fixed oneof message field merge.
  * Simplify parsing messages from array slices.
  * Unknown fields are now preserved by default.
  * Floating point values are now bitwise compared, affecting message equality
    check and Contains() API in map and repeated fields.
  * (protoc) Fixed a bug introduced in 3.5.0 and protoc in Windows now accepts non-ascii
    characters in paths again.
  * Unknown fields are now parsed and then sent back on the wire. They can be
    discarded at parse time via a CodedInputStream option.
  * Expression trees are no longer used
  * Floating point values are now compared bitwise (affects NaN value
    comparisons)
  * The default size limit when parsing is now 2GB rather than 64MB
  * MessageParser now supports parsing from a slice of a byte array
  * JSON list parsing now accepts null values where the underlying proto
    representation does
  * More support for FieldMask include merge, intersect and more.
  * Increasing the default recursion limit to 100.
  * Support loading FileDescriptors dynamically.
  * Provide access to comments from descriptors.
  * Added Any.Is method.
  * Compatible with C# 6
  * Added IComparable and comparison operators on Timestamp.
 * Added System.Memory dependency.

Excerpt from this file: https://github.com/protocolbuffers/protobuf/blob/master/CHANGES.txt

So there are some things which could possibly break some stuff. Especially the change in floating point comparison might bite us, although im unsure if we use floating points in our gossip messages anywhere.

All in all the standardization work is solid and a much better way to do things then to have each project define its own version. But even though tests are passing, i would not be surprised if we run into some comparison bugs that we missed the first time around.

Aaronontheweb:
@Danthar 

> Added System.Memory dependency.

That's the only big thing I can see being a change of any significance - but ultimately not a big deal because we're planning on adding that dependency inside Akka core in the not too distant future anyway. I should probably run a performance benchmark locally too to see if there's been any major performance changes with and without this change, but I think it's inevitable that we're going to need to upgrade anyway.

Issue #3886, opened 2019-08-13
`Config.ToString` doesn't include `WithFallback` config


boekabart:
On Akka.Net 1.3.14 (and we think we've seen this before), the following is true:

Given 2 `Config` instances `a` and `b`:

`a`
```
  akka : {
    some-key : value
  }
``` 
`b`
```
  akka : {
    other-key : 42
  }
```

When we merge them with the 'fallback' constructor:
`c3 = new Config(c1,c2).ToString(true)`
```
  akka : {
    some-key : value
    other-key : 42
  }
```
`new Config(a,b).ToString(false)`
```
  akka : {
    some-key : value
  }
```
They seem to be merged, with knowledge of what's the fallback - the second dump which asks to exclude it, does so. Good.

However, behold what happens when we use the `.WithFallback()` method on `Config`:
`a.WithFallback(b).ToString(true)`
```
  akka : {
    some-key : value
  }
``` 
`a.WithFallback(b).ToString(false)`
```
  akka : {
    some-key : value
  }
``` 
Regardless of how we ask to `ToString` the resulting instance, it refuses to dump our fallback configuration. We can, however, prove that the config does contain the knowledge of the fallback config:
`c1.WithFallback(c2).GetString('akka.some-key')`
```
value 
```
`c1.WithFallback(c2).GetInt('akka.other-key')`
```
42 
```
So, the `42` is there.

This issue is observed when the configuration value `akka.log-config-on-start` is set to `on` : when the config is built using the `WithFallback` method, only a partial Hocon is dumped.

So there seems to be a weird glitch in the merge logic which is used in `WithFallback` - this method doesn't actually use the `Fallback` property of `Config` - where the mentioned constructor does.

boekabart:
Repro (Linqpad script):
```
var c1 = ConfigurationFactory.ParseString("akka.some-key = value");
var c2 = ConfigurationFactory.ParseString("akka.other-key = 42");

c1.ToString().Dump("c1");
c2.ToString().Dump("c2");

new Config(c1, c2).ToString(true).Dump("c3 = new Config(c1,c2).ToString(true)");

c1 = ConfigurationFactory.ParseString("akka.some-key = value");
c2 = ConfigurationFactory.ParseString("akka.other-key = 42");
new Config(c1, c2).ToString(false).Dump("new Config(a,b).ToString(false)");

c1 = ConfigurationFactory.ParseString("akka.some-key = value");
[...]

PR #3893, opened 2019-08-27
Hocon toString() with correct fallback  behavior

feitzi:
As described in #3886 the `Config.ToString(true)` doesn't include all fallback values. 
The reason for the missing value is, that the toString() methods of the underlying hocon classes do skip all fallback values. 

```csharp
if (kvp.Value.AdoptedFromFallback) continue;
```

With this PR I adapted this behavior with an internal toString() that also expects an `includeFallback` boolean. The public toString() uses  this internal overload with 'includeFallback = false`.

Furthermore,  the `toString()` of `config`, changes the internal representation. That isn't expected (at least for me). Therefore I changed this to an immutable merge and execute the merging on this temporarily new object.


Aaronontheweb:
@IgorFedchenko did you already implement this in the stand-alone HOCON project?

IgorFedchenko:
> @IgorFedchenko did you already implement this in the stand-alone HOCON project?

@Aaronontheweb Yup, this is already implemented there.

Aaronontheweb:
Implemented in https://github.com/akkadotnet/HOCON/pull/181

Aaronontheweb:
This would actually be really useful in debugging some of the issues I'm running into with https://github.com/akkadotnet/akka.net/pull/4128

PR #3897, opened 2019-08-29
increase Akka.Remote.TestKit TestConductor timeout to 30s by default

### LINKED ISSUE #3868: MNTR: need to increase TestConductor timeout
Version: 1.3.13 and later

Noticed this issue while working on https://github.com/akkadotnet/akka.net/pull/3866

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote.TestKit/Internals/Reference.conf#L16

This timeout window is used by tools like the `TestConductor` to determine how long we're willing to wait while a node is being blackholed, exited, etc... 5 seconds is the current default value and it's not quite long enough for some scenarios - specifically, when we're trying to `TestConductor.Exit` a node that is also doing work concurrently inside its `RunOn` method - strong chance that this node might have actors who are doing blocking work and can't be shutdown right away.

I recommend changing this value to something greater than or equal to the longest `Within` blocks you see in MNTR specs - typically about 30 seconds. 

On the JVM they don't use at timeout for this at all - I added one years ago because we ran into issues with MNTR deadlock periodically.

Aaronontheweb:
close #3868

### DIFF
--- a/src/core/Akka.Remote.TestKit/Internals/Reference.conf
+++ b/src/core/Akka.Remote.TestKit/Internals/Reference.conf
@@ -13,7 +13,7 @@ akka {
     barrier-timeout = 30s
     
     # Timeout for interrogation of TestConductor’s Controller actor
-    query-timeout = 5s
+    query-timeout = 30s
     
     # Threshold for packet size in time unit above which the failure injector will
     # split the packet and deliver in smaller portions; do not give value smaller


Aaronontheweb:
@Danthar looks like we found an [Incrementalist](https://github.com/petabridge/Incrementalist) bug here - changing a value in an embedded non-code resource doesn't trigger a re-run of that tests' dependency track. I'll see about patching this and re-running it.

Aaronontheweb:
Looks like I should have this fixed shortly: https://github.com/petabridge/Incrementalist/pull/57

Danthar:
Cool. Found myself browsing the code this afternoon. Should not be to hard
to fix. Code looks pretty straightforward. 😊

Op vr 30 aug. 2019 22:02 schreef Aaron Stannard <notifications@github.com>:

Aaronontheweb:
@Danthar I would love some help cleaning up the Incrementalist code - it's in dire need of refinement / cleanup.

Aaronontheweb:
Turns out that the reason for this issue was that the Roslyn analyzer doesn't discover non-code files in the MSBuild Workspace

Aaronontheweb:
Re-running using the latest Incrementalist version.

Aaronontheweb:
Bah, detected the Remote.TestKit change and compiled it, but not any of its dependencies... I'm going to need to look at this again.

Aaronontheweb:
New culprit is an old one: https://github.com/petabridge/Incrementalist/issues/55

Aaronontheweb:
That's more like it - took 65 minutes to run. Validated that this didn't break the MNTR suite in any meaningful way.

PR #3901, opened 2019-09-02
Add MNTR documentation

Aaronontheweb:
Work in progress - want to add some docs that explain how end-users can leverage the multi-node testkit.

These docs will likely change if we get `dotnet tool` support implemented, but not in any meaningful way - the way tests are written and how the syntax works will largely remain the same.


Aaronontheweb:
Whoops, this was meant to be a draft

Aaronontheweb:
This took an enormous amount of work, but I'm finally finished with it.

Issue #3855, opened 2019-07-16
Get MNTR to format messages for Azure DevOps


Aaronontheweb:
Version: 1.4.0 beta

Need to get the MNTR to be able to format its test output into a TRX file or something else that Azure DevOps can parse. Pluggable outputs may not be a bad idea.

Aaronontheweb:
closed via #3904

PR #3904, opened 2019-09-05
MultiNode test reporter for Azure DevOps

valdisz:
Test reporter which is using Azure DevOps API to report MNTR test progress.

PR is related to #3855


valdisz:
Right now PR exists to verify that something will be reported to DevOps during a build.

Please do not look into it yet :)

valdisz:
@Aaronontheweb Incrementalist is failing :( Can you direct into a possible issue? Before I will spend some hours on it.

Aaronontheweb:
Yep, I'd be happy to take a look

valdisz:
@Aaronontheweb seems that build related token does not allow to publish test results. I can verify that this is really so, then we have 2 possibilities:

1. Create access token with needed permissions
2. Generate xunit test report style instead of direct reporting to devops

What you will choose?

Aaronontheweb:
@valdisz I'd just emit an XUnit-style report and I can come along later and output those TRX files up to Azure DevOps, which is what we do here:

https://github.com/akkadotnet/akka.net/blob/dev/build.fsx#L232

This forces XUnit to emit a TRX file, which can be read and understood by Azure DevOps.

And here is where we tell Azure DevOps how to consume it:

https://github.com/akkadotnet/akka.net/blob/37be6a5624367d43bf46ae04334387fcdc47ce94/build-system/azure-pipeline.template.yaml#L34-L39

So all we really need is just a way to output a .TRX file - does that make sense?

src/core/Akka.MultiNodeTestRunner.AzureDevOps/Akka.MultiNodeTestRunner.AzureDevOps.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\common.props" />
+  <PropertyGroup>
+    <TargetFrameworks>$(NetFrameworkLibVersion);$(NetStandardLibVersion)</TargetFrameworks>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.TeamFoundationServer.Client" Version="16.153.0" />
```
Aaronontheweb: I'd avoid the route of creating another NuGet package here. We just need TRX output formatting, not necessarily a tool for connecting directly to Azure DevOps.

valdisz:
@Aaronontheweb you can look on this.

Aaronontheweb:
@valdisz will do

build.fsx
```diff
@@ -289,7 +289,7 @@ Target "MultiNodeTests" (fun _ ->
 
             let args = StringBuilder()
                     |> append assembly
-                    |> append (sprintf "-Dmultinode.teamcity=%b" hasTeamCity)
+                    |> append (sprintf "-Dmultinode.reporter=%s" (if hasTeamCity then "teamcity" else "trx"))
```
Aaronontheweb: LGTM

src/core/Akka.MultiNodeTestRunner.TrxReporter.Tests/ResultSummaryTests.cs
```diff
@@ -0,0 +1,84 @@
+// -----------------------------------------------------------------------
+//  <copyright file="ResultSummaryTests.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2019 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2019 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+namespace Akka.MultiNodeTestRunner.AzureDevOps.Tests
+{
+    using System.Collections.Generic;
```
Aaronontheweb: Nitpick: we typically put `using` statements above the namespace.

src/core/Akka.MultiNodeTestRunner.TrxReporter/Akka.MultiNodeTestRunner.TrxReporter.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <Import Project="..\..\common.props" />
+  <PropertyGroup>
+    <TargetFrameworks>$(NetFrameworkLibVersion);$(NetStandardLibVersion)</TargetFrameworks>
+    <LangVersion>latest</LangVersion>
```
Aaronontheweb: Should set this to a specific value so we don't accidentally jump onto C# 8, which includes some breaking changes (`static interface`) - although I don't think targeting C# 8 by itself will create a breaking change with our DLLs. It's only if we use `static interface` that this will come up.

src/core/Akka.MultiNodeTestRunner.TrxReporter/Models/ResultSummary.cs
```diff
@@ -0,0 +1,101 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="ResultSummary.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2019 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2019 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+namespace Akka.MultiNodeTestRunner.AzureDevOps.Models
```
Aaronontheweb: Need to adjust the namespace - should be `Akka.MultiNodeTestRunner.TrxReporter`.

src/core/Akka.MultiNodeTestRunner.TrxReporter/Models/UnitTest.cs
```diff
@@ -0,0 +1,57 @@
...
+        public Identifier Id { get; } = Identifier.Create();
+        public Identifier TestListId { get; }
+        public Identifier ExecutionId { get; } = Identifier.Create();
+        
+        public string Name { get; }
+        public string Storage { get; }
+        public string AdapterTypeName => "executor://xunit/MulitNodeTestRunner";
+        public string ClassName { get; }
+
+        public List<UnitTestResult> Results = new List<UnitTestResult>();
```
Aaronontheweb: So I think I understand this - these are all of the sub-results for each individual node's test.

I.e. in a multi-node test with 4 nodes, we would expect 5 items in this collection: 1 for each node and 1 for the summary of the results. Is that correct? Just want to make sure I understand how it works.

src/core/Akka.MultiNodeTestRunner.TrxReporter/TrxSinkActor.cs
```diff
@@ -0,0 +1,219 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Xml.Linq;
+    using MultiNodeTestRunner.AzureDevOps.Models;
+    using Reporting;
+    using Sinks;
+
+    public class TrxSinkActor : TestCoordinatorEnabledMessageSink
```
Aaronontheweb: Actor design looks good

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -123,15 +125,30 @@ static void Main(string[] args)
...
 
             var listenAddress = IPAddress.Parse(CommandLine.GetPropertyOrDefault("multinode.listen-address", "127.0.0.1"));
             var listenPort = CommandLine.GetInt32OrDefault("multinode.listen-port", 6577);
             var listenEndpoint = new IPEndPoint(listenAddress, listenPort);
             var specName = CommandLine.GetPropertyOrDefault("multinode.spec", "");
             var platform = CommandLine.GetPropertyOrDefault("multinode.platform", "net");
+            var reporter = CommandLine.GetPropertyOrDefault("multinode.reporter", "trx");
+
+            Props coordinatorProps;
+            switch (reporter.ToLowerInvariant())
```
Aaronontheweb: Need to have an option for the console formatter, which is what would occur in the case that this value was never provided.

src/core/Akka.MultiNodeTestRunner.TrxReporter/Models/UnitTest.cs
valdisz: `UnitTest` is a test container. In usual cases, there will be just one `UnitTestResult`, but when, for example, data-driven tests are executed, there could be more results.

Each `UnitTestResult` could have inner test results and each inner test result could have its own inner results too... as much as you need it.

I implemented that each multi-node test will have one `UnitTest` with one `UnitTestResult`, but it will container inner results for each individual node.

```
Multi-node test (UnitTest)
|
+-- Overall test outcome, a summary of all individual nodes (UnitTestResult)
     |
     +-- Node 1 test result (UnitTestResult in InnerResults)
     +-- Node 2 test result (UnitTestResult in InnerResults)
     +-- Node n test result (UnitTestResult in InnerResults)
```

Aaronontheweb: Got it, that makes sense

Aaronontheweb:
@valdisz thanks for answering my question about the test structure - will review this again once the other changes are implemented.

PR #3911, opened 2019-09-10
have DocsSamples.sln get ValueTuple version from Akka.NET project

Aaronontheweb:


Aaronontheweb:
Oversight on my part - we really need to merge these solutions together when it's convenient.

PR #3913, opened 2019-09-10
Merge doc samples into main solution

Aaronontheweb:
Merged `DocSamples.sln` into the main solution.

There's still some more work to be done here, mainly moving the `Akka.Docs.Tests` project into the `/src` folder so the tests can be run on every build - but that also requires finding all of the DocFx references and updating them too. This is a good starting point, which ensures at the very least that compilation and test runs from Visual Studio hit any snags these tests detect.


Aaronontheweb:
Looks like an issue with the `Tutorials` project - exactly the type of issue I want to find.

Aaronontheweb:
Replaced via https://github.com/akkadotnet/akka.net/pull/4005

PR #3916, opened 2019-09-17
ClusterRouter no routees on startup 

Zetanova:
The ClusterRouterActor is initializing the nodes on startup but not adding any routees. 

The startup should have the same behavior as the ClusterEvent.CurrentClusterState message.


Zetanova:
It did not realy fix the problem. 

Because a router is used "directly" there is a short window after the router got created and the cstr() and PreStart() gets executed.

The real problem is that it is unclear for the creator of the routerActor when the router is initialized and ready to use. 

My current workaround (blocking) is:
```
var router = Context.ActorOf(MyActor.Props().WithRouter(routerConfig), $"my-remote-pool");
router.Ask<ActorIdentity>(new Identify("")).GetAwaiter().GetResult();
//now router is initialzed 
```

Zetanova:
Maybe this PR fixed it.

I am trying to make a Akka.Cluster.Tests.MultiNode.Routing test for it but struggle to get the Multinode testkit running.

This bug is happening for sure with a ClusterRouterPool router 
with settings allowLocalRoutees=false and/or a Role that is not on the local node.

What i think is happening is that the base of RoutedActorRef is RepointableActorRef
and it makes on async init a hot swap of UnstartedCell to the RoutedActorCell. 
The UnstartedCell is buffering messages until the real RoutedActorCell is initialized
but after the initializaion no routees where added (the PR fixed this) 
and the buffered messages ignored.

With the settings allowLocalRoutees=true and a Role that is valid for the local node.
then the local node gets taken automatically on an empty routee list.

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
```diff
@@ -94,7 +92,7 @@ public static ClusterRouterGroupSettings FromConfig(Config config)
         {
             return new ClusterRouterGroupSettings(
                 GetMaxTotalNrOfInstances(config),
-                ImmutableHashSet.Create(config.GetStringList("routees.paths").ToArray()),
+                ImmutableHashSet.CreateRange(config.GetStringList("routees.paths")),
```
Aaronontheweb: Good

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
```diff
@@ -266,7 +258,7 @@ internal static int GetMaxTotalNrOfInstances(Config config)
             }
             else
             {
-                return number; ;
```
Aaronontheweb: Good

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
```diff
@@ -801,39 +750,32 @@ public virtual void RemoveMember(Member member)
         /// <param name="message">TBD</param>
         protected override void OnReceive(object message)
         {
-            if (message is ClusterEvent.CurrentClusterState)
+            switch(message)
```
Aaronontheweb: LGTM (the entire `switch` block)

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
```diff
@@ -744,20 +691,22 @@ public ImmutableSortedSet<Address> AvailableNodes
         /// <returns>TBD</returns>
         public Address FullAddress(Routee routee)
         {
-            Address a = null;
-            if (routee is ActorRefRoutee)
+            Address a;
+            switch(routee)
```
Aaronontheweb: LGTM

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
```diff
@@ -879,34 +814,28 @@ public ClusterRouterGroupActor(ClusterRouterGroupSettings settings) : base(setti
         /// <summary>
         /// TBD
         /// </summary>
-        public ImmutableDictionary<Address, ImmutableHashSet<string>> UsedRouteePaths { get; private set; }
+        public ImmutableDictionary<Address, ImmutableHashSet<string>> UsedRouteePaths { get; private set; } = ImmutableDictionary<Address, ImmutableHashSet<string>>.Empty;
```
Aaronontheweb: LGTM

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
```diff
@@ -855,20 +797,13 @@ internal class ClusterRouterGroupActor : ClusterRouterActor
...
-            }
 
-            UsedRouteePaths = Settings.AllowLocalRoutees
-                ? ImmutableDictionary<Address, ImmutableHashSet<string>>.Empty.Add(Cluster.SelfAddress, settings.RouteesPaths.ToImmutableHashSet())
-                : ImmutableDictionary<Address, ImmutableHashSet<string>>.Empty;
+            _group = Cell.RouterConfig as Group 
+                ?? throw new ActorInitializationException(
+                    $"ClusterRouterGroupActor can only be used with group, not {Cell.RouterConfig.GetType()}"); ;
+
+            if (Settings.AllowLocalRoutees)
```
Aaronontheweb: Cleaner than what we had before

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
```diff
@@ -1054,29 +978,28 @@ public Address SelectDeploymentTarget()
...
         {
-            // Moved from RouterPoolActor
-            var poolSize = message as AdjustPoolSize;
-            if (poolSize != null)
+            switch(message)
             {
-                if (poolSize.Change > 0)
-                {
+                // Moved from RouterPoolActor
+                case AdjustPoolSize poolSize when poolSize.Change > 0:
```
Aaronontheweb: Did we not support this properly before? @Zetanova

Aaronontheweb:
See #3901 for MNTR docs - should be publishing that to https://getakka.net/ soon.

src/core/Akka.Cluster/Routing/ClusterRoutingConfig.cs
Zetanova: What property? 
There should be no behavior changes in this commit. 
Only some cleanup.

Aaronontheweb: Oh I see what you mean - duh. I read the diff wrong.

Aaronontheweb:
@Zetanova do you want a separate PR to address those bugs? Can I merge in this one to get the improvements?

Zetanova:
I think https://github.com/akkadotnet/akka.net/pull/3916/commits/1e6469aaa697f72595f6a649673d97fb3a7304f1 fixed it

What is missing is a test to test for the bug. 
I will try to make the MultiNode test for it and add it do this PR.

Test setup would be:
```
//required remote nodes with Role: RemoteNodeRole are UP
var pool = [some pool]
var settings = new ClusterRouterPoolSettings(10000, 1, true, "RemoteNodeRole"); 
var routerConfig = new ClusterRouterPool(pool, settings);
var router = Context.ActorOf(MyActor.GetProps()
                        .WithRouter(routerConfig), $"myremotedeployactor");

//following message would be silently ignored without the fix
router.Tell(new TestMessage("1")); //disappears without log or warning

//the current Workaround does work like following :
router.Tell(Cluster.Get(Context.System).State);
router.Ask<ActorIdentity>(new Identify(null)).GetAwaiter().GetResult(); //lock
router.Tell(new TestMessage("2")); //success
```

Aaronontheweb:
MNTR docs are here: https://getakka.net/articles/testing/multi-node-testing.html

Aaronontheweb:
Must have broken this thing on rebase. I'll fix it.

PR #3922, opened 2019-09-20
Inject Address into LogSource

### LINKED ISSUE #3867: Inject Address into LogSource for Actors
Version: 1.3.13 and later

I tried to do this a number of years ago with limited success and ultimately had to roll these changes back, but it looks like the JVM eventually come up with a way of doing this too:

https://github.com/akka/akka/blob/e188b1f653e2c8382312eb587749bda9955bdd5b/akka-actor/src/main/scala/akka/event/Logging.scala#L298-L319

When Akka.Remote is running, this will inject the address of the current node into all _local_ logs for that node - this should significantly help improve the debugging experience of working with Akka.NET.

Aaronontheweb:
closes #3867

### DIFF
--- a/src/core/Akka/Actor/Internal/ActorSystemImpl.cs
+++ b/src/core/Akka/Actor/Internal/ActorSystemImpl.cs
@@ -28,7 +28,7 @@ namespace Akka.Actor.Internal
     }
 
     /// <summary>
-    /// TBD
+    /// INTERNAL API
     /// <remarks>Note! Part of internal API. Breaking changes may occur without notice. Use at own risk.</remarks>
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
Still needs some unit test coverage before it's ready for primetime.

Aaronontheweb:
This broke some tests that were expecting `LogSource` to appear a specific way - that's to be expected. Will adjust the `LogSource` or the tests, depending on what looks more useful.

Aaronontheweb:
Still need to do API approval, but I fixed the failing `LogSource` specs in the `TestKit.Tests` package.

PR #3924, opened 2019-09-21
Fix build script to be able to run Fake from a path that contains spaces

cptjazz:
When using a checkout directory that contains spaces the PowerShell build script was unable to execute the Fake command. This PR fixes this issue by quoting the path to Fake in the `Invoke-Expression` statement.


build.ps1
```diff
@@ -171,6 +171,6 @@ $Arguments = @{
 
 # Start Fake
 Write-Host "Running build script..."
-Invoke-Expression "$FakeExePath `"build.fsx`" $ScriptArgs $Arguments"
-
+Invoke-Expression "& `"$FakeExePath`" `"build.fsx`" $ScriptArgs $Arguments"
```
Aaronontheweb: LGTM - I'll review the build system output and make sure that's germane.

PR #3925, opened 2019-09-21
Support paths containing spaces in 'build.fsx'

cptjazz:
The `dotnet pack` commands in `build.fsx` were struggling with an `-OutputPath` that contains spaces, e. g. when the checkout directory of Akka.NET contains spaces.
This PR fixes this issue by quoting the output path.


Aaronontheweb:
@cptjazz looks like the build system hung - were you able to get this to run locally?

cptjazz:
@Aaronontheweb I re-tested this and I can confirm that the `build build` and `build nuget` commands run normally. Since only the NuGet parts were affected in the PR I did not specifically test other commands when I filed this PR.
I now tested with more commands and the `build runtests` seems to have a spaces-in-path issue too. I will eventually file a new PR for specifically the problem in `build runtests`.

Can the build be triggered again? Maybe something else went wrong.

cptjazz:
Despite stating otherwise above: Since the changes are only small, I did not create a new PR but I just updated this PR and fixed another path problem regarding the XML output path used in the xUnit command in `build.fsx`.

Aaronontheweb:
@cptjazz have some merge conflicts here

cptjazz:
Seems I based off an older `dev` branch.
I now merged in the newest `upstream/dev` which should resolve the conflicts.

Aaronontheweb:
@cptjazz looks like it's working now.

PR #3930, opened 2019-09-23
V1.3.15 backport release

Aaronontheweb:
Release notes forthcoming...


Aaronontheweb:
Damn, looks like we may not be able to include the Akka.Persistence.TestKit in a backport release after all - the pre-.NET Standard 2.0 polyfills are creating an issue there. Going to give this another try without it and just ship the Akka.Remote fix instead.

PR #3940, opened 2019-09-27
[xunit] Fix xunit warnings

sean-gilliam:
This PR fixes a few xunit warnings.

Most of these warnings were in the form of Assert.Equals(<bool>, <some value>) where xunit suggests the more readable Assert.<bool>(<some value>).

Also removed a duplicate InlineData entry that wasn't testing duplicates.


Aaronontheweb:
Going to merge https://github.com/akkadotnet/akka.net/pull/3942 in to help with this PR. Only test that failed.

Issue #599, opened 2015-02-07
Make SupervisorStrategy serializable


Aaronontheweb:
One of the outstanding dependent issues we need to resolve before we can consider #545 resolved is a question @rogeralsing  raised in #581:

> The currently largest issue is the implementation of SupervisorStrategy, we use a delegate ('Decider') to deal with exception evaluation, this does not serialize.
>
> The only option I can think of is to replace this with a KVP -> exception type + directive and serialize that instead.

To which I added:

> For the `SupervisorStrategy` bit, we should still allow the use of delegates *but not as the default option*. There are some capabilities that supervisors need to do such as publish child failures to another subscriber (this gets used into the Akka.Remote internals for listening for transport errors.) Can't get rid of the delegate option entirely.
> 
> We should warn the end user that any supervisor strategy that uses a delegate is not serializable. Kind of like how we do with `DynamicProps` (which may no longer be needed, now that we have `IndirectActorProducer` support.)
> 
> For 80% of cases, using a `Decider` DSL that just does pattern matching around Exception types will get the job done and we can guarantee that that is serializable. We should make that the safe default and allow advanced users access to the delegate method, with the warning that you can't serialize this stuff over the wire.

Conclusion: we need a wire-friendly DSL for building `Deciders`. A `PatternMatch` equivalent for handling exceptions, issuing directives, and optionally being able to send a message to an `ActorRef` should be included in the design.

rogeralsing:
What if we make the decider completely value based, but based on outcome, call some lifecycle events, where you could trigger various actions, instead of triggering them from within the decider itself. 
Would that work?

Aaronontheweb:
I think so - depends on whether or not those actions need to be network-serializable, right?

rogeralsing:
What if those actions belongs to the actor, not the decider?

decider Outcome -> actor.onXxxx(..)

Aaronontheweb:
So here's a real-world example from one of our unit tests:

```csharp
 public void RemoteRouter_must_set_supplied_SupervisorStrategy()
        {
            var probe = CreateTestProbe(masterActorSystem);
            var escalator = new OneForOneStrategy(ex =>
            {
                probe.Ref.Tell(ex);
                return Directive.Escalate;
            });

            var router =
                masterActorSystem.ActorOf(
                    new RemoteRouterConfig(new RoundRobinPool(1, null, escalator, null),
                        new[] { new Address("akka.tcp", sysName, "localhost", port) })
[...]

PR #3941, opened 2019-09-27
[build] Fix CS0162 warnings

sean-gilliam:
This PR fixes a number of warnings related to CS0162 (Unreachable code detected).

Most of these are related to conditional debugging flags that aren't necessarrily set during a normal compilation. There are some however that are there as to clear any compile errors that would happen due to ambigious references.

The pragmas are added to the function boundaries so as not to pollute the logic in the function.

### DIFF
--- a/src/core/Akka/Actor/Props.cs
+++ b/src/core/Akka/Actor/Props.cs
@@ -105,11 +105,13 @@ namespace Akka.Actor
             return Deploy.Equals(other.Deploy);
         }
 
+#pragma warning disable CS0162 // Disabled because it's marked as a TODO
         private bool CompareSupervisorStrategy(Props other)
         {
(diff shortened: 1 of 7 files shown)


src/core/Akka/Actor/Props.cs
```diff
@@ -105,11 +105,13 @@ private bool CompareDeploy(Props other)
             return Deploy.Equals(other.Deploy);
         }
 
+#pragma warning disable CS0162 // Disabled because it's marked as a TODO
```
Aaronontheweb: We should probably fix this issue by closing #599 just comparing if the two supervisor strategies are equal by reference. Too complicated to do it via anything else. What do you think @sean-gilliam ?

Aaronontheweb: Also, we have a serializable supervisor strategy but no one uses it because the syntax is extremely awkward.

PR #3944, opened 2019-10-01
use .NET Standard 2.0 and later as only build target

### LINKED ISSUE #3851: Target .NET Standard 2.0 and newer only

Aaronontheweb:
close #3851 - dropping .NET Framework 4.5.2 support.

Draft PR because we need to get rid of the old Helios transport package first.


Aaronontheweb:
https://github.com/akkadotnet/akka.net/pull/3945 needs to be merged into `dev` first before this can run successfully.

Aaronontheweb:
Welp, here's where this build hit the wall:

```
https://github.com/akkadotnet/akka.net/blob/0a8c0390ee24927e3c381a99bca079ebc7e5b8e9/src/core/Akka/Configuration/Hocon/AkkaConfigurationSection.cs
```

... @Arkatufus - so what should we do with this given that we're moving HOCON as a stand-alone feature into Akka.NET? If we get rid of `AkkaConfigurationSection`?

Aaronontheweb:
I think what we're going to end up doing here, so as to not introduce too many behavior changes at once, is wait and see how the new stand-alone HOCON integration goes and see if that library can offer a good solution.

Arkatufus:
> 
> 
> Welp, here's where this build hit the wall:
> 
> https://github.com/akkadotnet/akka.net/blob/0a8c0390ee24927e3c381a99bca079ebc7e5b8e9/src/core/Akka/Configuration/Hocon/AkkaConfigurationSection.cs
> 
> ... @Arkatufus - so what should we do with this given that we're moving HOCON as a stand-alone feature into Akka.NET? If we get rid of `AkkaConfigurationSection`?

This should fix it: https://github.com/Arkatufus/akka.net/commit/3e08c21a41a732d3dfcb70bf169c45375129fdf7

Arkatufus:
We should mark the old `System.Configuration` implementation as `deprecated` to prepare for standalone HOCON integration.

Aaronontheweb:
> We should mark the old `System.Configuration` implementation as `deprecated` to prepare for standalone HOCON integration.

Have some API ideas for how we can do this...

Aaronontheweb:
@Arkatufus retrying this PR with your fix

Aaronontheweb:
Looks like we're good to go here. Going to merge this in and set the groundwork for stand-alone HOCON.

PR #3949, opened 2019-10-04
[docs] Move/update Sandcastle project

sean-gilliam:
This PR moves the Sandcastle project to under the docs directory. It also updates the project to the current version that supports .NET Standard.

Although we're using DocFx now, this can be used for people who prefer the old Sandcastle generated docs. It's also useful in helping to determine potential documentation issues.


Aaronontheweb:
@sean-gilliam related issue - should we add build script support for sandcastle again? I haven't looked at it in a couple of years.

sean-gilliam:
@Aaronontheweb I would say probably not. I haven't heard anyone complaining about it; albeit I've been away from this project for a while. If someone REALLY wants them, then we can revisit it. Or they can use this script to generate the docs themselves.

Does that make sense?

PR #3950, opened 2019-10-04
Akka.Persistence.Query: made Offset IComparable to itself

### LINKED ISSUE #3947: Akka.Persistence.Query: need to make Offset implement IComparable
Version: Akka.NET v1.3.13

At the moment, it's not easy to verify during testing that all of the events came back in a particular order when testing `EventsByTag` queries in Akka.Persistence.Query. Implementing `IComparable` on `Offset`, which operates off of `long` integers internally, should make this feasible in future releases.

Aaronontheweb:
close #3947

### DIFF
--- a/src/core/Akka.Persistence.Query/Offset.cs
+++ b/src/core/Akka.Persistence.Query/Offset.cs
@@ -9,7 +9,13 @@ using System;
 
 namespace Akka.Persistence.Query
 {
-    public abstract class Offset
+    /// <summary>
+    /// Used in <see cref="IEventsByTagQuery"/> implementations to signal to Akka.Persistence.Query
+    /// where to begin and end event by tag queries.
+    ///
+    /// For concrete implementations, see <see cref="Sequence"/> and <see cref="NoOffset"/>.
+    /// </summary>
+    public abstract class Offset : IComparable<Offset>
     {
         /// <summary>
         /// Used when retrieving all events.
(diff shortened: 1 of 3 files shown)


src/core/Akka.Persistence.Query/Offset.cs
```diff
@@ -59,6 +71,16 @@ public override bool Equals(object obj)
...
         public override int GetHashCode() => Value.GetHashCode();
+
+        public override int CompareTo(Offset other)
+        {
+            if (other is Sequence seq)
+            {
+                return CompareTo(seq);
+            }
+
+            throw new InvalidOperationException($"Can't compare offset of type {GetType()} to offset of type {other.GetType()}");
```
Aaronontheweb: Only allowed to compare to other `Offset` implementations of the same type

src/core/Akka.Persistence.Query/Offset.cs
```diff
@@ -71,5 +93,15 @@ public sealed class NoOffset : Offset
...
         private NoOffset() { }
+
+        public override int CompareTo(Offset other)
+        {
+            if (other is NoOffset no)
+            {
+                return 0;
+            }
+
+            throw new InvalidOperationException($"Can't compare offset of type {GetType()} to offset of type {other.GetType()}");
```
Aaronontheweb: Same here

Aaronontheweb:
Bah, forgot the API approval tests.

Issue #3955, opened 2019-10-04
Deprecate PersistentFSM+State.Using to deter usage in user code


aloker:
PersistentFSM+State.Using is meant for internal usage only (unlike FSMBase+State.Using).

https://github.com/akkadotnet/akka.net/blob/ca9916097b5efc06b7e9fbd408306f2df456a3da/src/core/Akka.Persistence/Fsm/PersistentFSM.cs#L381-L390

Internally, calls to it can be replaced with Copy, so the method itself can be dropped or marked internal, but as a non-breaking change it should be marked [Obsolete] to discourage usage in user code.

JVM equivalent https://github.com/akka/akka/pull/23740

PR: #3956

Aaronontheweb:
closed via #3956

PR #3956, opened 2019-10-04
Deprecate PersistentFSM+State.Using (#3955)

aloker:
Also replaces all internal calls with calls to Copy, so Using can safely be removed in the future.


Aaronontheweb:
@aloker ah, need to approve the API: https://getakka.net/community/public-api-changes.html

aloker:
Oh, didn't know about that. Let me check.

Aaronontheweb:
It's ok, we keep it in there so people don't make breaking API changes by accident.

aloker:
So I basically diff the related .received.txt and and the .approved.txt file to check that only the intended changes are in there (here: the added attribute) and then copy .received.txt over to .approved.txt, right?

Aaronontheweb:
@aloker all you should need to do is just commit the new `approved.txt` after running it through a diff / merge tool. Should launch automatically when the test fails. Otherwise, you could copy it over manually and rename it.

aloker:
I didn't notice P4Merge pop up in the background. Nice :-)

PR #3958, opened 2019-10-04
Revert "Fix nightly-build template"

Danthar:
Reverts akkadotnet/akka.net#3957

It now runs on merge as well. Template is probably correct and im seriously suspecting an AzureDevops bug


Aaronontheweb:
@Danther the things we do for science

PR #3962, opened 2019-10-05
Fixed singleton issue when leaving several nodes

ismaelhamed:
Issue [#27487](https://github.com/akka/akka/issues/27487)
Port [#27488](https://github.com/akka/akka/pull/27488)

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Runtime.Serialization;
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Using this to help guinea pig the new TRX reporter for the MNTR

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -632,13 +634,11 @@ private ActorSelection Peer(Address at)
...
-        private void GetNextOldestChange()
+        private void GetNextOldestChanged()
         {
-            if (_oldestChangedReceived)
-            {
-                _oldestChangedReceived = false;
-                _oldestChangedBuffer.Tell(OldestChangedBuffer.GetNext.Instance);
-            }
+            if (!_oldestChangedReceived) return;
+            _oldestChangedReceived = false;
```
Aaronontheweb: So if I understand this correctly, this code can never run inside the `ClusterSingletonState.Start`, right?

ismaelhamed: This code hasn't change, I just inverted the "if" statement.

Aaronontheweb: Got it

PR #3963, opened 2019-10-06
Fixed scheduling and cancelling receive timeout

ismaelhamed:
Issue: [#25340](https://github.com/akka/akka/issues/25340)
Port: [#25347](https://github.com/akka/akka/pull/25347)

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -78,7 +78,7 @@ namespace Akka.Actor
         protected virtual void AutoReceiveMessage(Akka.Actor.Envelope envelope) { }
         public void Become(Akka.Actor.Receive receive) { }
         public void BecomeStacked(Akka.Actor.Receive receive) { }
-        public void CheckReceiveTimeout() { }
+        public void CheckReceiveTimeout(bool reschedule = True) { }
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
@ismaelhamed missing an API approval https://getakka.net/community/public-api-changes.html

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -78,7 +78,7 @@ namespace Akka.Actor
         protected virtual void AutoReceiveMessage(Akka.Actor.Envelope envelope) { }
         public void Become(Akka.Actor.Receive receive) { }
         public void BecomeStacked(Akka.Actor.Receive receive) { }
-        public void CheckReceiveTimeout() { }
+        public void CheckReceiveTimeout(bool reschedule = True) { }
```
Aaronontheweb: @ismaelhamed so this is a breaking API change that will require all plugins et al to be recompiled in the event that they target Akka.NET v1.3.15 and earlier. I'm not worried about this specific change, however, because it's part of the `ActorCell` which is meant to be an internal-use only set of APIs and isn't called very often by end-users.

Aaronontheweb: Just commenting to add some information, not requesting a change. I think this is fine as-is.

Issue #3954, opened 2019-10-04
SourceLink and Symbol Packages - are they working for you?


Aaronontheweb:
https://github.com/akkadotnet/akka.net/pull/3953 - @mmisztal1980 working on a repro spec now.

For reasons unknown, source link was letting me see where the error was happening. It may be set up incorrectly or our symbol package publishing may not be working correctly.

_Originally posted by @Aaronontheweb in https://github.com/akkadotnet/akka.net/issues/3943#issuecomment-538531518_

Aaronontheweb:
Version: Akka v1.3.15, v1.4.0-beta2+

Not seeing the SourceLink we added back in 1.3.13 and earlier work out in the wild. Has anyone else been able to get it working?

Aaronontheweb:
https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink

aloker:
I *think* you need to add a portable PDB file to the NuGet package in order to make it work, but I'm not 100% sure. That's also mentioned on the page you linked. The Akka NuGet packages don't appear to have any pdb files included. Also, the v1.4.0 beta package is the first one to have the required <repository> element in its spec.

Aaronontheweb:
@akoker so, we haven't been publishing PDBs correctly then by the sound of it? IIRC we do publish separate symbol packages, but we haven't been pushing those correctly. 

From our 1.3.15 release build logs:

```

2019-09-23T08:24:05.9254751Z 
2019-09-23T08:24:07.2654719Z   Successfully created package 'D:\a\1\s\bin\nuget\Akka.TestKit.1.3.15.nupkg'.
2019-09-23T08:24:07.2775640Z C:\Program Files\dotnet\sdk\2.2.105\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(202,5): warning NU5125: The 'licenseUrl' element will be deprecated. Consider using the 'license' element instead. [D:\a\1\s\src\core\Akka.TestKit\Akka.TestKit.csproj]
2019-09-23T08:24:07.2869748Z   Successfully created package 'D:\a\1\s\bin\nuget\Akka.TestKit.1.3.15.snupkg'.
2019-09-23T08:24:07.3194336Z Finished Target: CreateNuget
2019-09-23T08:24:07.3195734Z Starting Target: PublishNuget (==> CreateNuget)
[...]

PR #3966, opened 2019-10-08
fix: NuGet symbols not published (#3954)

aloker:
Updates the build script to handle the newer snupkg format.

Please try and check if this works - can't test it with the official nuget packages obviously.

refs #3954

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -486,17 +486,15 @@ Target "PublishNuget" (fun _ ->
     if (shouldPushNugetPackages || shouldPushSymbolsPackages) then
         printfn "Pushing nuget packages"
         if shouldPushNugetPackages then
-            let normalPackages= 
-                !! (outputNuGet @@ "*.nupkg") 
-                -- (outputNuGet @@ "*.symbols.nupkg") |> Seq.sortBy(fun x -> x.ToLower())
+            let normalPackages= !! (outputNuGet @@ "*.nupkg") |> Seq.sortBy(fun x -> x.ToLower())
             for package in normalPackages do
                 try
-                    publishPackage (getBuildParamOrDefault "nugetpublishurl" "") (getBuildParam "nugetkey") 3 package
+                    publishPackage (getBuildParamOrDefault "nugetpublishurl" "https://api.nuget.org/v3/index.json") (getBuildParam "nugetkey") 3 package
                 with exn ->
                     printfn "%s" exn.Message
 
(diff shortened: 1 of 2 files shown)


build.fsx
```diff
@@ -486,17 +486,15 @@ Target "PublishNuget" (fun _ ->
     if (shouldPushNugetPackages || shouldPushSymbolsPackages) then
         printfn "Pushing nuget packages"
         if shouldPushNugetPackages then
-            let normalPackages= 
-                !! (outputNuGet @@ "*.nupkg") 
-                -- (outputNuGet @@ "*.symbols.nupkg") |> Seq.sortBy(fun x -> x.ToLower())
+            let normalPackages= !! (outputNuGet @@ "*.nupkg") |> Seq.sortBy(fun x -> x.ToLower())
```
aloker: No .symbols.nupkg files are created anymore, just .nupkg and .snupkg.

build.fsx
```diff
@@ -486,17 +486,15 @@ Target "PublishNuget" (fun _ ->
...
             for package in normalPackages do
                 try
-                    publishPackage (getBuildParamOrDefault "nugetpublishurl" "") (getBuildParam "nugetkey") 3 package
+                    publishPackage (getBuildParamOrDefault "nugetpublishurl" "https://api.nuget.org/v3/index.json") (getBuildParam "nugetkey") 3 package
```
aloker: This one's crucial: auto pushing related snupkg files only works with the V3 API.

Aaronontheweb: Been meaning to make that change myself - we should probably also update the `build-system/windows-release.yaml` file  too.

aloker: Good point, PR updated

Aaronontheweb:
@aloker one last change:

```
let shouldPushSymbolsPackages = (hasBuildParam "symbolspublishurl") && (hasBuildParam "symbolskey")
```

Is there ever a condition where we wouldn't want to try to push debug symbols? Is this check necessary any more?

aloker:
If auto-publishing the related snupkg files works, the check and the whole second part (publishing snupkg files) isn't necessary. I'm not sure if auto-publishing always works for private package servers or if this hardcoded for nuget.org only. I don't have a private package server at hand that's uses the V3 API.

Personally I'd leave it in just in case someone publishes to a private package server and needs to push the symbols separately.

Issue #3967, opened 2019-10-08
Build script fails when dotnet core v3 SDK is installed


aloker:
Running the build script will fail if dotnet code v3 SDK (3.0.100) is installed on the machine.

As soon as Akka.FSharp.fsproj is packed, an error occus:

```
dotnet "pack" "C:\projects\Privat\akka.net\src\core\Akka.FSharp\Akka.FSharp.fsproj" --configuration Release --output "C:\projects\Privat\akka.net\bin\nuget" --include-symbols --no-build
Microsoft (R)-Build-Engine, Version 16.3.0+0f4c62fea für .NET Core
Copyright (C) Microsoft Corporation. Alle Rechte vorbehalten.

C:\projects\Privat\akka.net\src\core\Akka.FSharp\Akka.FSharp.fsproj : warning NU1605: Detected package downgrade: System.ValueTuple from 4.5.0 to 4.4.0. Reference the package directly from the project to select a different version.
C:\projects\Privat\akka.net\src\core\Akka.FSharp\Akka.FSharp.fsproj : warning NU1605:  Akka.FSharp -> Akka -> System.ValueTuple (>= 4.5.0)
C:\projects\Privat\akka.net\src\core\Akka.FSharp\Akka.FSharp.fsproj : warning NU1605:  Akka.FSharp -> System.ValueTuple (>= 4.4.0)
C:\Program Files\dotnet\sdk\3.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.targets(133,5): error NETSDK1085: Die Eigenschaft "NoBuild" wurde auf TRUE festgelegt, aber das Ziel "Build" wurde aufgerufen. [C:\projects\Privat\akka.net\src\core\Akka\Akka.csproj]
Running build failed.
Error:
System.Exception: Pack failed on "pack" "C:\projects\Privat\akka.net\src\core\Akka.FSharp\Akka.FSharp.fsproj" --configuration Release --output "C:\projects\Privat\akka.net\bin\nuget" --include-symbols --no-build
   bei Fake.DotNetCli.Pack@372-3.Invoke(String message) in D:\code\fake\src\app\FakeLib\DotNetCLIHelper.fs:Zeile 372.
   bei Fake.DotNetCli.Pack(FSharpFunc`2 setPackParams) in D:\code\fake\src\app\FakeLib\DotNetCLIHelper.fs:Zeile 372.
[...]

PR #3968, opened 2019-10-08
fix: build fails when SDK 3.0.100 is installed (#3967)

aloker:
Disables the --no-build flag when packing NuGet packages of F# projects as that causes build errors when .net core SDK 3.0.100 is installed.

refs #3967

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -451,7 +451,7 @@ Target "CreateMntrNuget" (fun _ ->
                     { p with
                         Project = project
                         Configuration = configuration
-                        AdditionalArgs = ["--include-symbols"]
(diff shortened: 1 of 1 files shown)


build.fsx
```diff
@@ -451,7 +451,7 @@ Target "CreateMntrNuget" (fun _ ->
                     { p with
                         Project = project
                         Configuration = configuration
-                        AdditionalArgs = ["--include-symbols"]
+                        AdditionalArgs = ["--include-symbols"] @ if System.IO.Path.GetExtension(project) = ".fsproj" then [] else ["--no-build"]
```
aloker: I'm no F# master, so if there's a more elegant way, let me know.

PR #3970, opened 2019-10-09
fix: build fails when SDK 3.0.100 is installed, second attempt (#3967)

aloker:
Removes --no-build flag from all NuGet builds to make sure F# projects can be packed when .net Core 3 SDK is used.

refs #3967 
refs #3968
refs #3969

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -392,7 +392,7 @@ Target "CreateNuget" (fun _ ->
                     { p with
                         Project = project
                         Configuration = configuration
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
@aloker I'll test this locally

Aaronontheweb:
Just tested this locally - it works and yep, solved the problems I ran into trying to package Akka.NET locally on .net core 3

Issue #3920, opened 2019-09-19
Can't compile code that uses ValueTuple on Linux while targeting .NET Framework 4.5.2


Aaronontheweb:
Version: v1.4.0-beta1

Ran into this on https://github.com/akkadotnet/akka.net/pull/3914 and the other ValueTuple PRs - can't use them on Linux without compilation errors while we also cross-compile for .NET Framework. Going to need to look into a fix for this or just drop .NET Framework altogether and stick with .NET Standard 2.0 as a baseline moving forward.

PR #3973, opened 2019-10-10
Moving to ValueTuples

IgorFedchenko:
@Aaronontheweb After updating only few projects others won't compile because of changing signatures.

Should I continue creating separate PRs per project? In this case we could create some "value-tuples" branch in your repo and safely merge PRs until all projects will be updated and will compile, and then merge to dev with possible conflicts resolving.

Because we can not merge this PR to dev anyway until solution will compile, right?

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -451,7 +451,7 @@ Target "CreateMntrNuget" (fun _ ->
                     { p with
                         Project = project
                         Configuration = configuration
-                        AdditionalArgs = ["--include-symbols"] @ if System.IO.Path.GetExtension(project) = ".fsproj" then [] else ["--no-build"]                        
+                        AdditionalArgs = ["--include-symbols"]
                         VersionSuffix = versionSuffix
                         OutputPath = "\"" + outputNuGet + "\"" } )
         )
--- a/build.fsx
+++ b/build.fsx
@@ -451,7 +451,7 @@ Target "CreateMntrNuget" (fun _ ->
                     { p with
                         Project = project
                         Configuration = configuration
-                        AdditionalArgs = ["--include-symbols"]
+                        AdditionalArgs = ["--include-symbols"] @ if System.IO.Path.GetExtension(project) = ".fsproj" then [] else ["--no-build"]                        
                         VersionSuffix = versionSuffix
                         OutputPath = "\"" + outputNuGet + "\"" } )
         )
--- a/src/benchmark/PingPong/Program.cs
+++ b/src/benchmark/PingPong/Program.cs
@@ -164,7 +164,7 @@ namespace PingPong
             }
         }
 
-        private static async Task<Tuple<bool, long, int>> Benchmark<TActor>(int factor, int numberOfClients, long numberOfRepeats, PrintStats printStats, long bestThroughput, int redCount) where TActor : ActorBase
+        private static async Task<(bool, long, int)> Benchmark<TActor>(int factor, int numberOfClients, long numberOfRepeats, PrintStats printStats, long bestThroughput, int redCount) where TActor : ActorBase
         {
             var totalMessagesReceived = GetTotalMessagesReceived(numberOfRepeats);
             //times 2 since the client and the destination both send messages
@@ -197,7 +197,7 @@ namespace PingPong
             if (!countdown.Wait(TimeSpan.FromSeconds(10)))
             {
                 Console.WriteLine("The system did not start in 10 seconds. Aborting.");
-                return Tuple.Create(false, bestThroughput, redCount);
+                return (false, bestThroughput, redCount);
             }
             var setupTime = totalWatch.Elapsed;
             var sw = Stopwatch.StartNew();
@@ -237,7 +237,7 @@ namespace PingPong
             }
             Console.ForegroundColor = foregroundColor;
 
-            return Tuple.Create(redCount <= 3, bestThroughput, redCount);
+            return (redCount <= 3, bestThroughput, redCount);
         }
 
         private static long GetTotalMessagesReceived(long numberOfRepeats)
--- a/src/benchmark/RemotePingPong/Program.cs
+++ b/src/benchmark/RemotePingPong/Program.cs
@@ -141,7 +141,7 @@ namespace RemotePingPong
             return numberOfClients * numberOfRepeats * 2;
         }
 
-        private static async Task<Tuple<bool, long, int>> Benchmark(int numberOfClients, long numberOfRepeats, long bestThroughput, int redCount)
+        private static async Task<(bool, long, int)> Benchmark(int numberOfClients, long numberOfRepeats, long bestThroughput, int redCount)
         {
             var totalMessagesReceived = GetTotalMessagesReceived(numberOfClients, numberOfRepeats);
             var system1 = ActorSystem.Create("SystemA", CreateActorSystemConfig("SystemA", "127.0.0.1", 0));
@@ -211,7 +211,7 @@ namespace RemotePingPong
 
             Console.ForegroundColor = foregroundColor;
             Console.WriteLine("{0,10},{1,8},{2,10},{3,11}", numberOfClients, totalMessagesReceived, throughput, sw.Elapsed.TotalMilliseconds.ToString("F2", CultureInfo.InvariantCulture));
-            return Tuple.Create(redCount <= 3, bestThroughput, redCount);
+            return (redCount <= 3, bestThroughput, redCount);
         }
 
         private class AllStartedActor : UntypedActor
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
@@ -181,7 +181,7 @@ namespace Akka.Cluster.Sharding.Tests
             }
         }
 
-        internal ExtractEntityId extractEntityId = message => message is int ? Tuple.Create(message.ToString(), message) : null;
+        internal ExtractEntityId extractEntityId = message => message is int ? (message.ToString(), message) : null;
 
         internal ExtractShardId extractShardId = message => message is int ? message.ToString() : null;
 
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
@@ -12,6 +12,7 @@ using Akka.Configuration;
 using Akka.Remote.TestKit;
 using System.Collections.Immutable;
 using System.Threading.Tasks;
+using Akka.Util;
 using FluentAssertions;
 
 namespace Akka.Cluster.Sharding.Tests
@@ -181,7 +182,7 @@ namespace Akka.Cluster.Sharding.Tests
             }
         }
 
-        internal ExtractEntityId extractEntityId = message => message is int ? (message.ToString(), message) : null;
+        internal ExtractEntityId extractEntityId = message => message is int ? (message.ToString(), message) : Option<(string, object)>.None;
 
         internal ExtractShardId extractShardId = message => message is int ? message.ToString() : null;
 
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
@@ -152,9 +152,9 @@ namespace Akka.Cluster.Sharding.Tests
             switch (message)
             {
                 case Get msg:
-                    return Tuple.Create(msg.Id, message);
+                    return (msg.Id, message);
                 case Add msg:
-                    return Tuple.Create(msg.Id, message);
+                    return (msg.Id, message);
             }
             return null;
         };
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
@@ -14,6 +14,7 @@ using Akka.Cluster.TestKit;
 using Akka.Configuration;
 using Akka.Remote.TestKit;
 using Akka.Remote.Transport;
+using Akka.Util;
 using FluentAssertions;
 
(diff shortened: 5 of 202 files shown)


IgorFedchenko:
Some `Tuple`s are still there in updated projects, because sometimes `class` types are required instead of `struct`s. Method constraints like `where T: class`, or `Interlocked.Exchange(ref object obj, object val)` also requires reference types.
But this is not going to keep `Tuple`s in public APIs anyway.

I am using `Search with Pattern` ReSharper's feature, and find/replace with regex when I need to replace types when this feature does not help. And then resolving compile errors manually.

Aaronontheweb:
> Should I continue creating separate PRs per project? In this case we could create some "value-tuples" branch in your repo and safely merge PRs until all projects will be updated and will compile, and then merge to dev with possible conflicts resolving.

I'd work backwards - do separate PRs per-project, but do higher level projects that don't have any downstream dependencies (i.e. Cluster.Sharding, Akka.Persistence.Query, Streams) first, the move onto the heavy-duty projects like Akka.Remote, Akka, Akka.Cluster...

IgorFedchenko:
All right, now this is going to be single huge PR that will replace `Tuples` with `ValueTuples` where possible.

IgorFedchenko:
@Aaronontheweb Have a question about Akka.Streams update.
Our update is going to be breaking change anyway, that is fine. Basically, when moving from `Tuple` to `ValueTuple`, we do not allow passing `null` values anymore. 
While in other projects this may be fine (I did not find much issues with this, or even heavy `Tuple` class usage at all), Akka.Streams project uses tuples a lot, and there are lots of cases when `null` value is used as a tuple value. 

Here is an example, where all big change starts:
```csharp
public static FlowWithContext<TCtx, TIn, TCtx, TOut2, TMat> Collect<TCtx, TIn, TOut, TOut2, TMat>(
	this FlowWithContext<TCtx, TIn, TCtx, TOut, TMat> flow, Func<TOut, TOut2> fn) where TOut2 : class
{
	var stage = new Collect<(TOut, TCtx), (TOut2, TCtx)?>(func: x =>
	{
		var result = fn(x.Item1);
		return ReferenceEquals(result, null) ? ((TOut2, TCtx)?)null : (result, x.Item2);
	});
	return flow.Via(Flow.FromGraph(stage));
}
```

The `flow.Via` method used here is:
```csharp
public FlowWithContext<TCtxIn, TIn, TCtx2, TOut2, TMat> Via<TCtx2, TOut2, TMat2>(
	IGraph<FlowShape<(TOut, TCtxOut), (TOut2, TCtx2)>, TMat2> viaFlow) =>
	FlowWithContext.From(Flow.FromGraph(Inner).Via(viaFlow));
```

So it requires `(TOut2, TCtx2)` tuple, which can not be `null`, while expression `ReferenceEquals(result, null) ? ((TOut2, TCtx)?)null : (result, x.Item2)` may return null in some cases.

The solution of making this work without logical change is change `(TOut2, TCtx2)` to `(TOut2, TCtx2)?` (make it nullable). And this works well, but after making lots of changes like that turned out that almost all tuples, used in `Akka.Streams` public API, should be replaced with `nullable` `ValueTuples`.

While this can be done, I am getting lots of places where code like `Select(t => t.Item1)` should be replaced with `Select(t => t.Value.Item1)` and get `Possible null reference exception` warning in addition. Lots of places like that, but there is no null check in original code which makes me think that nullable tuples are not expected there.

Additionally, there is lots of code like this:
```csharp
Source.From(Enumerable.Range(1, 5))
	.MapMaterializedValue<(Task<IImmutableList<int>>, Task<IImmutableList<int>>, Task<IImmutableList<int>>)?>(_ => null);
```

This is a part of test method, and `null` is used as materialization value - so I have to cast to nullable `ValueTuple` here, and in lots of other places. And use `.Value` property to get result. While obviously this source then chained with other flows and result will not produce `null`-tuples, compiler forces usage nullable tuples and `.Value` property.

Moving to nullable `ValueTuple` may break lots of code because of this. The other option would be to not use `null` as a result of `Akka.Stream` methods that are expected to return tuple.

So, other suggestion is - should I change `null` in such code to `default(<tuple_type>)` for example? It seems like null is actually not expected anywhere in this methods anyway, so it can be any value. 

What do you think?

Aaronontheweb:
@IgorFedchenko thanks for a detailed and thoughtful analysis. I'd also take a look at https://github.com/akkadotnet/akka.net/pull/3282 and see what @vasily-kirichenko did there when he worked on this problem a while back.

> Moving to nullable ValueTuple may break lots of code because of this. The other option would be to not use null as a result of Akka.Stream methods that are expected to return tuple.

Nullable value tuples seem like hot garbage in this API, and generally speaking Akka.Streams shouldn't be returning `null` - what it sounds like we really want is a result that indicates "no results," such as `NotUsed`. 

I think the best approach here is to move away from nullable `ValueTuple` output, but I say that without knowing how feasible that is - the devil is in the details of the implementation.

Can you make a list, briefly, of the areas where there will be some complex refactoring as a result of losing `null`ability?

Aaronontheweb:
A second though: null items inside a `ValueTuple` is fine, just not a null `ValueTuple` itself.

IgorFedchenko:
@Aaronontheweb About null items inside `ValueTuple` - sure, the issue is with null `ValueTuple` itself.

In general, I suspect that any time something is returning `null` tuple value in stream this means that the result is not going to be used. If we need to report special "no result", `Option` class could be used, otherwise there is no hint that `null` is possible value of a tuple and everybody would be getting `NullReferenceException`, all the time. So, my idea is that we can put anything instead of `null` and this is not going to be used anyway. Only if somewhere there is a special `if (tuple == null) DoOneThing() else DoAnother()` check, I guess `Option` class would be used in that case.

This idea is somewhat approved with PR you have referenced. In this PR, some classes that I have issues with are missing (like [this](https://github.com/akkadotnet/akka.net/blob/e4bef6e9b077661161b5fac65b9581ae973f0df7/src/core/Akka.Streams/Dsl/FlowWithContextOperations.cs#L43-L52) - but I know that `FlowWithContext` is recent feature, or [this](https://github.com/akkadotnet/akka.net/blob/3d471b15253cb39b21f429fb75ac2a6243e68655/src/core/Akka.Streams/Dsl/PagedSource.cs#L62)), but anyway whenever there was returning `null` in tests it is changed to `default(<tuple_type>)`. 

So I am going to use that way and keep going with non-nullable `ValueTuples` `default` values. The only place I will appreciate help with so far is [this one](https://github.com/akkadotnet/akka.net/blob/e4bef6e9b077661161b5fac65b9581ae973f0df7/src/core/Akka.Streams/Dsl/FlowWithContextOperations.cs#L43-L52) , because I am not sure about code semantics. Seems like we are converting output elements with preserving context. But why to return null tuple value when conversion function returns null?.. Ok, I see why - this is a case I was talking about, there is a special check [here](https://github.com/akkadotnet/akka.net/blob/0155f322aa5580f5db40b3537e4ca8586e1a5a49/src/core/Akka.Streams/Implementation/Fusing/Ops.cs#L480) that changes behavior depending on `null` or `non-null` values. @Aaronontheweb I am suggesting to change this to use `Option` class instead of `null`. Or just check for `Item1.HasValue`. 

What do you think?

IgorFedchenko:
About my last question: I have just tried to use `Option` in that case, and this will change signature of the method and will break more code, so change `Collect` class check to check only `Item1.HasValue` condition would be preferable.

Aaronontheweb:
cc @Horusiath - see @IgorFedchenko's post above about `FlowWithContext` API changes. Might need some help figuring out the best way to rewrite that without using `null`.

IgorFedchenko:
This function, that returns null if conversion result is null, is used on `Collect` class in single place. It is stored in `_func` field and used like this:

```csharp
var result = WithSupervision(() => _stage._func(Grab(_stage.In)));
if (result.HasValue)
{
	if (result.Value.IsDefaultForType())
		Pull(_stage.In);
	else
		Push(_stage.Out, result.Value);
}
```

`WithSupervision` executes passed lambda, `_stage._func(Grab(_stage.In))` returns `null` without exceptions, so `result` is going to be `Option` with value `null` and `result.HasValue = true`... All right, I see now. `IsDefaultForType()` method is performing check I was looking for, and this will work with `ValueTuple`'s `default` value also. 

Seems like I got it, but please correct me if I am wrong. 

All in all, changing `null` to `default` looks like best solution for all cases I was talking about.

Aaronontheweb:
@IgorFedchenko that looks like a good solution.

IgorFedchenko:
I will add Public API approval, and there is one test that is failing in my branch but is passing on dev:
`Akka.Cluster.Sharding.Tests.PersistentShardSpec.Persistent_Shard_must_remember_entities_started_with_StartEntity`

Aaronontheweb:
@IgorFedchenko looks like we ran into this issue on Linux too: https://github.com/akkadotnet/akka.net/issues/3920

IgorFedchenko:
@Aaronontheweb In this case, this may be a part of [this known issue](https://github.com/dotnet/roslyn/issues/12897). 

For those who will be interested in what is going on, here is an exception I was getting in test:
```
System.ArgumentException : Cannot resolve method Void .ctor(System.String, System.Object) because the declaring type of the method handle System.ValueTuple`2[T1,T2] is generic. Explicitly provide the declaring type to GetMethodFromHandle.
   at System.Reflection.MethodBase.GetMethodFromHandle(RuntimeMethodHandle handle)
   at Akka.Cluster.Sharding.Tests.PersistentShardSpec.Persistent_Shard_must_remember_entities_started_with_StartEntity() in C:\Projects\Upwork\Aaron\Akka.net\src\contrib\cluster\Akka.Cluster.Sharding.Tests\PersistentShardSpec.cs:line 60
```

While this was looking like some issue with parsing expression with value tuple constructor, I have just refactored test to define lambda as a function object, and then passing this object as a parameter - that worked.

IgorFedchenko:
@Aaronontheweb Seems like this PR is ready for review now.

Aaronontheweb:
@IgorFedchenko excellent work. I'll get about reviewing it today.

Aaronontheweb:
Looks like my other PR didn't totally fix the issue on Linux - have some other build failures stemming from NuGet.exe failing to download... wonder if there's some azure connectivity problems, given that we had issues with pipelines on that other PR too.

IgorFedchenko:
@Aaronontheweb In error report I still see errors like this:
```
Routing/ConsistentHash.cs(51,17): error CS8356: Predefined type 'System.ValueTuple`2' is declared in multiple referenced assemblies: 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' and 'System.ValueTuple, Version=4.0.3.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' [/home/vsts/work/1/s/src/core/Akka/Akka.csproj]
Routing/ConsistentHash.cs(50,17): error CS8356: Predefined type 'System.ValueTuple`2' is declared in multiple referenced assemblies: 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' and 'System.ValueTuple, Version=4.0.3.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' [/home/vsts/work/1/s/src/core/Akka/Akka.csproj]
```

But did not noticed anything related to connectivity issues, at least at that log... Should I take some time to resolve this?

IgorFedchenko:
Hmm, can not reproduce this errors locally, when trying to build in inside `mcr.microsoft.com/dotnet/core/sdk` docker container all is building fine with `dotnet build -f netstandard2.0`. @Aaronontheweb Do you have some hints to reproduce this issue outside of CI? I guess this is some environment configuration difference

Aaronontheweb:
My money is on the Mono compiler struggling with ValueTuple in some of the projects

Aaronontheweb:
Don't have this issue with .NET Core compilation on Linux

IgorFedchenko:
@Aaronontheweb Seems like you were right - disabling `ValueTuple` package reference under mono fixed the problem.

But we get another one, during `Akka.MultiNodeTestRunner.Shared.Tests.csproj` project build:
```
AsyncContext/AsyncContext.TaskQueue.cs(39,32): error CS8356: Predefined type 'System.ValueTuple`2' is declared in multiple referenced assemblies: 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' and 'System.ValueTuple, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' [/home/vsts/work/1/s/src/core/Akka.Tests.Shared.Internals/Akka.Tests.Shared.Internals.csproj]
```

I could not quickly find where does this project get that reference - seems like it is inside one of the referenced packages... Anyway, this does not belong to public API, so I just rolled back `Tuple<T1,T2>` for this class, and it did not break anything.

**Update**: one more fix like that in test project
```
ImmutableMessage.cs(43,33): error CS8356: Predefined type 'System.ValueTuple`2' is declared in multiple referenced assemblies: 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' and 'System.ValueTuple, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' [/home/vsts/work/1/s/src/contrib/serializers/Akka.Serialization.TestKit/Akka.Serialization.TestKit.csproj]
```

IgorFedchenko:
@Aaronontheweb Build issues under linux are resolved now, but I had to make 2 more small fixes (rolling back to `Tuple<T1, T2>`) in two more places in test assemblies.

While this should not be a problem to the users, we will not be able to use `ValueTuple` in that projects because of mono issues. This projects do not reference `System.ValueTuple` directly, so I suspect that some of the referenced packages are referencing it. So this may be a problem to resolve such conflicts without modifying their sources. 
But if we will detect which reference is a source of the problem, possibly we'll be able to fix that (or at least send PR for fixing that in the future). Not sure if this should be a part of this PR, so maybe better to put it in separate issue?

Aaronontheweb:
> But if we will detect which reference is a source of the problem, possibly we'll be able to fix that (or at least send PR for fixing that in the future). Not sure if this should be a part of this PR, so maybe better to put it in separate issue?

I'd keep a separate PR for that - this will probably be fixed as a result of changing which version of .NET Framework / .NET Core / .NET Standard we ultimately support as part of Akka.NET v1.4.0.

Aaronontheweb:
@IgorFedchenko 

Akka.Persistence.TestKit.Tests.CounterActorTests.CounterActor_internal_state_will_be_lost_if_underlying_

Failure on all 3 platforms

IgorFedchenko:
There may be race conditions because of creating new actor with same name as terminated actor from outside of actor's parent. Committed naming fix to avoid such conflict in this test.

Aaronontheweb:
Couldn't get that last check to re-run, but it was a few racy specs that have a high flip rate. We should probably disable them. Nice work @IgorFedchenko .

PR #3974, opened 2019-10-13
Provide static GetRoutees.Instance property that holds a singleton in…

cptjazz:
…stance of this message.

This mimics the behaviour of `PoisonPill.Instance` and `Kill.Instance` and feels a bit more natural to use than `RouterMessage.GetRoutees`.


Aaronontheweb:
@cptjazz need to approve the API changes: https://getakka.net/community/public-api-changes.html

Aaronontheweb:
Ah nevermind, you already did!

PR #3988, opened 2019-10-17
Implement UnfoldInfinite + update Unfold to use Option class

### LINKED ISSUE #3977: Akka.Streams.Dsl.Source.UnfoldInfinite - not implemented method in Akka.Streams
In Akka.Streams Akka.Streams.Dsl.Source.UnfoldInfinite always throws NotImplemented exception.
I'm using version 1.3.15.
Having not implemented methods in API can be confusing and I think they shouldn't make it to stable releases.

IgorFedchenko:
This is going to close #3977, because implements `UnfoldInfinite` method.

The point of this method in scala is to skip `Option` class usage when returning `Unfold` conversion function result, which makes code simpler. Useful for sequences that are not going to finish or when sequence is going to be finished with exception.

One more thing this PR adds is updating `Unfold` and `UnfoldAsync` methods to use `Option<Tuple>` instead of `Tuple` as a conversion function result. We are trying to avoid using `null` values as tuple values (this is also useful from the #3973 point of view), so the source should be finished when it gets `Option.None` value.

As a useful addition, `.AsOption()` extension method is implemented in `Akka.Stream.Utils` namespace, which helps to update existing code that uses `Unfold`/`UnfoldAsync`. 

Close #3977

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/LastElementSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/LastElementSpec.cs
@@ -56,7 +56,7 @@ namespace Akka.Streams.Tests.Dsl
         [Fact]
         public void A_stream_via_LastElement_should_materialize_to_the_last_element_emitted_by_a_source_before_it_failed()
         {
-            var t = Source.Unfold(1, n => n >= 3 ? throw new Exception() : Tuple.Create(n + 1, n + 1))
+            var t = Source.UnfoldInfinite(1, n => n >= 3 ? throw new Exception() : Tuple.Create(n + 1, n + 1))
                 .ViaMaterialized(new LastElement<int>(), Keep.Right)
                 .ToMaterialized(Sink.Aggregate<int, Option<int>>(Option<int>.None, (_, o) => new Option<int>(o)), Keep.Both)
                 .Run(Sys.Materializer());
--- a/src/core/Akka.Streams.Tests/Dsl/SourceSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/SourceSpec.cs
@@ -13,6 +13,7 @@ using System.Threading.Tasks;
 using Akka.Streams.Dsl;
 using Akka.Streams.TestKit;
 using Akka.Streams.TestKit.Tests;
+using Akka.Streams.Util;
 using Akka.TestKit;
 using FluentAssertions;
 using Xunit;
@@ -323,7 +324,8 @@ namespace Akka.Streams.Tests.Dsl
                 var a = tuple.Item1;
                 var b = tuple.Item2;
                 if (a > 10000000)
-                    return null;
+                    return Option<Tuple<Tuple<int, int>, int>>.None;
+                
                 return Tuple.Create(Tuple.Create(b, a + b), a);
             }).RunAggregate(new LinkedList<int>(), (ints, i) =>
             {
@@ -337,12 +339,13 @@ namespace Akka.Streams.Tests.Dsl
         {
             EventFilter.Exception<Exception>(message: "expected").ExpectOne(() =>
             {
-                var task = Source.Unfold(Tuple.Create(0, 1), tuple =>
+                var task = Source.Unfold<Tuple<int, int>, int>(Tuple.Create(0, 1), tuple =>
                 {
                     var a = tuple.Item1;
                     var b = tuple.Item2;
                     if (a > 10000000)
                         throw new Exception("expected");
+                    
                     return Tuple.Create(Tuple.Create(b, a + b), a);
                 }).RunAggregate(new LinkedList<int>(), (ints, i) =>
                 {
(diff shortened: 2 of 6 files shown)


IgorFedchenko:
Have to admit that this is going to be breaking change for most part of `Unfold` method usages. But it is easy to fix, and is required anyway for #3973

IgorFedchenko:
Some tests are failing, but it seems like the same tests are failing in original `dev` branch.

src/core/Akka.Streams.Tests/Dsl/LastElementSpec.cs
```diff
@@ -56,7 +56,7 @@ public void A_stream_via_LastElement_should_materialize_to_materialize_to_None_f
         [Fact]
         public void A_stream_via_LastElement_should_materialize_to_the_last_element_emitted_by_a_source_before_it_failed()
         {
-            var t = Source.Unfold(1, n => n >= 3 ? throw new Exception() : Tuple.Create(n + 1, n + 1))
+            var t = Source.UnfoldInfinite(1, n => n >= 3 ? throw new Exception() : Tuple.Create(n + 1, n + 1))
```
IgorFedchenko: This is one way to fix compilation error after adding `Option` requirement for `Unfold` method. Also, this allows to verify that `UnfoldInfinite` works well.

src/core/Akka.Streams.Tests/Dsl/SourceSpec.cs
```diff
@@ -337,12 +339,13 @@ public void Unfold_Source_must_terminate_with_a_failure_if_there_is_an_exception
         {
             EventFilter.Exception<Exception>(message: "expected").ExpectOne(() =>
             {
-                var task = Source.Unfold(Tuple.Create(0, 1), tuple =>
+                var task = Source.Unfold<Tuple<int, int>, int>(Tuple.Create(0, 1), tuple =>
```
IgorFedchenko: Another way to resolve compilation error after update with `Option` requirement

src/core/Akka.Streams/Dsl/Source.cs
```diff
@@ -550,7 +550,7 @@ public static class Source
         public static Source<T, NotUsed> Repeat<T>(T element)
         {
             var next = new Tuple<T, T>(element, element);
-            return Unfold(element, _ => next).WithAttributes(DefaultAttributes.Repeat);
+            return Unfold(element, _ => new Option<Tuple<T, T>>(next)).WithAttributes(DefaultAttributes.Repeat);
```
IgorFedchenko: `UnfoldInfinite` could be used here too, by the way. Maybe even more convenient...

src/core/Akka.Streams/Util/ObjectExtensions.cs
```diff
@@ -21,5 +21,10 @@ public static class ObjectExtensions
         /// <param name="obj">TBD</param>
         /// <returns>TBD</returns>
         public static bool IsDefaultForType<T>(this T obj) => EqualityComparer<T>.Default.Equals(obj, default(T));
+        
+        /// <summary>
+        /// Wraps object to the <see cref="Option{T}"/> monade
+        /// </summary>
+        public static Option<T> AsOption<T>(this T obj) => new Option<T>(obj);
```
IgorFedchenko: This is something I am really missing very often. But maybe there also exist some method like that? Did not find simple way to convert object to optional value - `new Option(obj)` requires specifying target type.

src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
```diff
@@ -1760,8 +1760,8 @@ namespace Akka.Streams.Dsl
         public static Akka.Streams.SourceShape<T> Shape<T>(string name) { }
         public static Akka.Streams.Dsl.Source<T, Akka.NotUsed> Single<T>(T element) { }
         public static Akka.Streams.Dsl.Source<T, Akka.Actor.ICancelable> Tick<T>(System.TimeSpan initialDelay, System.TimeSpan interval, T tick) { }
-        public static Akka.Streams.Dsl.Source<TElem, Akka.NotUsed> Unfold<TState, TElem>(TState state, System.Func<TState, System.Tuple<TState, TElem>> unfold) { }
-        public static Akka.Streams.Dsl.Source<TElem, Akka.NotUsed> UnfoldAsync<TState, TElem>(TState state, System.Func<TState, System.Threading.Tasks.Task<System.Tuple<TState, TElem>>> unfoldAsync) { }
+        public static Akka.Streams.Dsl.Source<TElem, Akka.NotUsed> Unfold<TState, TElem>(TState state, System.Func<TState, Akka.Streams.Util.Option<System.Tuple<TState, TElem>>> unfold) { }
```
Aaronontheweb: This is an API change, but I know it's being done ultimately in service of https://github.com/akkadotnet/akka.net/pull/3973 - I'm fine with it.

PR #3998, opened 2019-10-18
only reference System.ValueTuple NuGet package when building against .NET Framework

### LINKED ISSUE #3920: Can't compile code that uses ValueTuple on Linux while targeting .NET Framework 4.5.2

Aaronontheweb:
close #3920

### DIFF
--- a/src/core/Akka/Akka.csproj
+++ b/src/core/Akka/Akka.csproj
@@ -16,11 +16,11 @@
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonVersion)" />
     <PackageReference Include="System.Collections.Immutable" Version="1.6.0" />
-    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == '$(NetFrameworkLibVersion)' ">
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
@IgorFedchenko compilation didn't barf on Linux immediately. I think this fixed the issue - problem was that .NET Standard 2.0 already includes a reference to `System.ValueTuple` and therefore the package introduces a conflict, whereas .NET Framework 4.6.2 and earlier requires an explicit reference on the NuGet package. So I just changed it such that the `System.ValueTuple` NuGet package only gets included when we build against the .NET Framework target.

I have no idea why this worked without any issues on Windows but failed on Linux - differences in the compiler implementations I guess?

IgorFedchenko:
Hmm, great catch! Indeed, maybe some difference in package conflict resolution algorithms... It's nice to see Linux tests passing now.

Aaronontheweb:
Looks like Azure Pipelines is having an issue - can't get it to re-run some MNTR specs that timed out at startup.

PR #4003, opened 2019-10-24
Singleton - missing state change fix

zbynek001:
introduced in #3962


Aaronontheweb:
@zbynek001 nice catch

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -728,7 +728,7 @@ private void InitializeFSM()
 
                         if (youngerData.Oldest.All(m => _removed.ContainsKey(m)))
                         {
-                            GoToOldest();
+                            return GoToOldest();
```
Aaronontheweb: LGTM

PR #4009, opened 2019-10-29
upgraded to Incrementalist v0.2.0

Aaronontheweb:
Should resolve issues with F# projects and `common.props` changes not being tested properly.


Aaronontheweb:
The output has looked good when I've been testing Incrementalist, but for whatever reason the latest version of Roslyn now produces a ton of failure / warning spam in the logs - which during our integration testing on [Incrementalist](https://github.com/petabridge/Incrementalist) itself turned out not to mean anything.

I'll keep an eye on it as I re-run some of the dependabot PRs that couldn't be properly tested prior to this.

Aaronontheweb:
cc @akkadotnet/core on that last bit

PR #4013, opened 2019-10-30
V1.3.16 backport

Aaronontheweb:
#### 1.3.16 September 23 2019 ####
**Maintenance Release for Akka.NET 1.3**

1.3.16 is a minor change, brought about by the introduction of [Hyperion v0.9.10](https://github.com/akkadotnet/Hyperion/releases/tag/0.9.10). This fixes a breaking change introduced by .NET Core 3.0, which could cause Akka.Cluster.Sharding users some problems when running apps on .NET Core 3.0.

Hyperion v0.9.10 also allows for full x-plat communciation between .NET Framework and .NET Core on all supported versions of each.

| --- | --- | --- | --- |
| 2 | 10 | 40 | Aaron Stannard |

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,14 +1,10 @@
-#### 1.3.15 September 23 2019 ####
+#### 1.3.16 October 30 2019 ####
 **Maintenance Release for Akka.NET 1.3**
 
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
Looks like there won't be a 1.3.16 - newest Hyperion is not compatible with v1.3.* build targets. Stick to v1.4.0-beta* if you need .NET Core 3.0 support.

Issue #3328, opened 2018-02-16
System.IO.Pipelines as Akka.NET remote layer


Horusiath:
As it looks that System.IO.Pipelines will soon reach .NET corefx repository, we could take a look at it and check if we could use for our remote layer.

Quick recap: System.IO.Pipelines is a low-alloc socket library which originates from Kestrel - Asp.NET high-performance HTTP layer. There's a lot of work which happened around in .NET to provide a low-alloc/alloc-free access used here, including heavy use of `Span<byte>`, `Memory<byte>` and `ByteSegment`s (something similar to our impl of `ByteString`, but mutable). Pipelines also have ability to compose custom user behavior by exposing `IPipe` interfaces, see:

- An experimental Redis client on top of pipelines: https://github.com/mgravell/RedisPing
- Websocket protocol implementation: https://github.com/mgravell/Pipelines.WebSockets

Another thing I'm working aside is a zero-cost abstraction layer over Hyperion, which could potentially work together with `IOutput` and `WritableBuffer` - pipelines API for reading/writing binary data. Hopefully other serialization libraries will go this path as well.

All of this would fit well with integration of Akka.NET with new .NET ecosystem i.e.:

- Let a `ByteString` use `ReadOnlySpan<byte>` instead of `ArraySegment<byte>`
- Replace/extend our `IBufferPool` with global `ByteArrayPool` from System.Memory namespace.
- Replace/extend our serializers API to work with new low-alloc APIs.

This is nothing sure atm, but I think that ADO.NET soon also will make use of those as well.

Aaronontheweb:
@Horusiath question: think it'd be easier to prove this as a concept inside of Akka.IO initially? Fewer moving parts than Akka.Remote.

But otherwise, I'm open to the idea. 

One of the problems our serialization / networking layer has right now is that all of the components involved have their own separate concepts of things like buffers, pooling, etc... Protobuf, DotNetty, JSON.NET, Hyperion. If these separate components adopt some unified abstraction like `Span`, it'll make it less expensive for us to keep our high degree of configurability without having to make the same performance compromises we have to today.

Anyway, I see this as a long term project since it'll inevitably depend on other third party libraries adopting these features as they are RTMed, bugs are found, etc - so the time horizon for this might be quite long before it can be release-worthy. Still worth doing.

Aaronontheweb:
[...]

PR #4015, opened 2019-10-31
Add Akka.IO.Tcp benchmark

IgorFedchenko:
To start work on issue #3328 we need to be able to measure `Akka.IO.Tcp` throughput before and after migration to `System.IO.Pipelines` and compare.

This PR adds benchmark as a simple client-server communication, exchanging with different number of messages with different payload sizes.


IgorFedchenko:
I am making 100 iterations per parameters combination, and two values per parameter. As you can see increasing message count gives linear increase of time consumed (I have also tried bigger values), and payload almost does not change time (which is pretty understandable in localhost networking).

Here is what I am getting on my local machine right now:
```
                    Method | MessageCount | MessageLength |      Mean |     Error |    StdDev |    Median | Allocated |
-------------------------- |------------- |-------------- |----------:|----------:|----------:|----------:|----------:|
 ClientServerCommunication |          100 |            10 |  42.54 ms | 2.7769 ms |  8.188 ms |  40.55 ms |       0 B |
 ClientServerCommunication |          100 |           100 |  40.98 ms | 0.6515 ms |  1.921 ms |  40.31 ms |       0 B |
 ClientServerCommunication |         1000 |            10 | 388.77 ms | 8.4383 ms | 24.881 ms | 380.52 ms |       0 B |
 ClientServerCommunication |         1000 |           100 | 377.30 ms | 3.9668 ms | 11.696 ms | 373.36 ms |       0 B |
```

Aaronontheweb:
So ~380ms for 1000 round trips?

IgorFedchenko:
> So ~380ms for 1000 round trips?

Yes

PR #4019, opened 2019-11-04
Akka.IO.Tcp benchmark with parallel clients

IgorFedchenko:
This is improvement of #4015 benchmark, which adds support for multiple parallel clients sending requests to server actor.

Also, I had to increase `BenchmarkDotNet` package version to `0.12.0`, because current version `0.10.14` is a year old and does not contain some issue fixes (like it was failing to perform benchmark if there was any output to a console).

Issue #3921, opened 2019-09-20
Our Tcp connections appear to be sharing their internal(to Akka) ByteBuffers


spankr:
# Intro
We are using Akka.NET to manage short-lived communication from our array of IoT devices.
The connect, give us their status, we respond with "OK\r\n" and then close the connection.
We are running v1.3.12 on a .NET Core application within a Docker container. (so Linux platform).

Over the past year we've been seeing communication that looks like it is over-writing itself. This is a recent example: `OK\r\ntle 4.1 BufferGet Loaded\r\nd 2\r\n`

Valid communication examples:
* `OK\r\n`
* `Shuttle 4.1 BufferGet Loaded\r\n`
* `Buffer 4.1 ShuttleAlign Aligned 2\r\n`

For the past year, we've just chalked it up as a glitch in the few IoT devices we've had live and asked our IoT team to hunt it down and fix it.

About 3 months ago, we've been scaling the number of IoT devices and we've been seeing more of these messages(about 2-4 per day). Not only that but the messages seem to be amalgamations of communication from different devices. There is no way a single IoT device can construct/send some of these messages.

What appears to be happening is that on our side (the Akka system-side), the same memory is being shared for multiple TCP sockets.

The example I provided above, conveniently aligns to the valid examples (whitespace added to fill for non-printable characters):
```
OK\r\ntle 4.1 BufferGet Loaded\r\nd 2\r\n
OK\r\n
Shut  tle 4.1 BufferGet Loaded\r\n
Buff  er 4.1 ShuttleAlign Aligne  d 2\r\n
```

When researching this, we came across an interesting **_TODO_** in the **ReleaseSocketAsyncEventArgs()** of [Akka.IO.TcpConnection.cs](https://github.com/akkadotnet/akka.net/blob/a0c44f1ffff794139dcb65cba65c9e204ec7e9e8/src/core/Akka/IO/TcpConnection.cs):
```
        private void ReleaseSocketAsyncEventArgs()
        {
            if (ReceiveArgs != null)
            {
                var buffer = new ByteBuffer(ReceiveArgs.Buffer, ReceiveArgs.Offset, ReceiveArgs.Count);
                Tcp.SocketEventArgsPool.Release(ReceiveArgs);
                //TODO: there is potential risk, that socket was working while released. In that case releasing buffer may not be safe.
                Tcp.BufferPool.Release(buffer);
                ReceiveArgs = null;
            }

            if (SendArgs != null)
            {
[...]

PR #4022, opened 2019-11-05
Update Akka.IO.TcpConnection implementation to use Socket.SendAsync method

### LINKED ISSUE #4018: Use Socket.SendAsync instead of blocking Send in Akka.IO.TcpConnection
Right now, `TcpConnection` actor is blocked while sending data, and unable to handle other messages (like read commands). Need to use async writes - more details [here](https://github.com/akkadotnet/akka.net/pull/4017#issuecomment-548940172)

IgorFedchenko:
Close #4018

Keeping in draft state until it will pass test with big messages


IgorFedchenko:
Added `akka.io.tcp.write-commands-queue-max-size` setting to limit size of the internal write commands queue in bytes.
Basically, if before update implementation was allowing only one pending write at a time (and it was possible to use only with `Tcp.CompountWrite` message), and if it was already filled the next write was ending with failure message - now it allows buffering of any number of messages, but limits the buffered writes total size - which actually seems more convenient. Setting this buffer to 0 will fallback to the previous behavior, because I have left the `CompountWrite` message handling unmodified.

So it is getting write request, puts it on the wire, then get another request, tries to buffer it.
Also, if write request was compound write, it basically stores it in special `pendingWrite` object and sends parts one by one, without touching the buffer.

In other words, the behavior is the same as it was before, but now if you are not comfortable with sending writes only after getting their acks, you can just use buffering (with unlimited size by default) and keep sending write commands. You will still need to keep track of their acks in case if you are going to close a connection for example - you need to be sure that all buffered commands are sent.

IgorFedchenko:
Before moving this to review, I want to make some refactoring over the code. Because now it got even more complicated then before (with this dual buffering - ComposedWrite handling and internal writes queue).

IgorFedchenko:
This is actually might be ready for review now.

In the last part, before it, there was buffering in ComposedWrite (because it could contain another ComposedWrite, and etc) on the one hand, and our internal queue for `SimpleWriteCommand`'s on another. Both of them were exactly for one same thing: preserve order of incoming `SimpleWriteCommand` requests (`ComposedWrite` is a way to batch them in one message). Nice functional style, but when we added a queue inside, this became much easier way to store pending writes.

So, I removed obsolete pieces and rewritten this parts with internal queue usage. When getting `ComposedWrite`, we are disassembling it onto `SimpleWriteCommand`'s and put them into the queue. Also, there is a special case when we have buffer size constraint and getting composed message with too big first `SimpleWriteCommand` - in that case, if this is possible, we are not putting that piece into the queue and sending it instantly. 
Well, just an implementation detail.

I will try running benchmark tomorrow - this should work little bit faster in a few ways. Might be 2-4% gain, but still.

IgorFedchenko:
@Aaronontheweb other thing that could be done is sending bytes from several messages at once.

I mean: suppose we have several `Write` commands, 1st and 2nd without ack and 3d with ack required. Or one `ComposedWrite` containing this sequence, does not matter.
Right now, it is sending them one by one - send first, wait until sent, then send second, and after this is done send 3rd (and send ack to it's sender once done).
When messages do not require ack, we could concatenate them into single `Socket.SendAsync`. In case above, we could send them all as one message. and once this is done send ack for the third message sender.
This will reduce buffering, ant speed-up the process as a whole. 

I have already changed the send-related code pretty much, so it differs from scala version. If this is ok, I could make this implementation also. 
Because if I get it right when sending i.e. 10Kb message, receiver can easily receive for example 2 messages, with 7K and 3K size. So, why not to make it possible to send 7K + 3K and get single 10K (or separate 9K + 1K messages) on receiver end? This is something one should expect when working with sockets.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -646,51 +685,34 @@ protected override void PostRestart(Exception reason)
             throw new IllegalStateException("Restarting not supported for connection actors.");
         }
 
-        private PendingWrite CreatePendingWrite(IActorRef commander, WriteCommand write, ConnectionInfo info)
+        private Option<PendingWrite> GetNextWrite(IEnumerable<(SimpleWriteCommand Command, IActorRef Sender)> headCommands = null)
         {
-            var head = write;
-            WriteCommand tail = Write.Empty;
-            while (true)
+            headCommands = headCommands ?? new List<(SimpleWriteCommand Command, IActorRef Sender)>();
```
Aaronontheweb: If there's a way we can use a static `List<(SimpleWriteCommand Command, IActorRef Sender)>.Empty` or something here instead of `new`-ing up a list each time, I think that would help reduce allocations here.

Aaronontheweb: `ImmutableList<(SimpleWriteCommand Command, IActorRef Sender)>.Empty` would do it.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -646,51 +685,34 @@ protected override void PostRestart(Exception reason)
...
+            foreach (var commandInfo in headCommands.Concat(_writeCommandsQueue.DequeueAll()))
             {
-                if (head == Write.Empty)
+                switch (commandInfo.Command)
```
Aaronontheweb: LGTM

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -815,73 +827,169 @@ private sealed class PendingBufferWrite : PendingWrite
...
-                            Release();
-                            return _connection.CreatePendingWrite(Commander, _tail, info);
-                        }
-                    }
-
+                    _sendArgs.SetBuffer(_buffer);
+                    if (!_connection.Socket.SendAsync(_sendArgs))
+                        _self.Tell(SocketSent.Instance);
+                    
+                    Release();
```
Aaronontheweb: Is the idea behind the `Release` method here to make the `PendingBufferWrite`s re-usable in a pool, so we don't have to frequently allocate and GC new ones? CC @Horusiath

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -815,73 +827,169 @@ private sealed class PendingBufferWrite : PendingWrite
...
                 }
                 catch (SocketException e)
                 {
                     _connection.HandleError(info.Handler, e);
-                    return this;
                 }
             }
 
-            private int Send(ByteString data)
+            public override void Release() { }
```
Aaronontheweb: Because if that's the case, we need to clear away the old content here and make this object re-usable again.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -3587,7 +3587,7 @@ namespace Akka.IO
     }
     public class TcpSettings
     {
-        public TcpSettings(string bufferPoolConfigPath, int initialSocketAsyncEventArgs, bool traceLogging, int batchAcceptLimit, System.Nullable<System.TimeSpan> registerTimeout, int receivedMessageSizeLimit, string managementDispatcher, string fileIoDispatcher, int transferToLimit, int finishConnectRetries, bool outgoingSocketForceIpv4) { }
+        public TcpSettings(string bufferPoolConfigPath, int initialSocketAsyncEventArgs, bool traceLogging, int batchAcceptLimit, System.Nullable<System.TimeSpan> registerTimeout, int receivedMessageSizeLimit, string managementDispatcher, string fileIoDispatcher, int transferToLimit, int finishConnectRetries, bool outgoingSocketForceIpv4, int writeCommandsQueueMaxSize) { }
```
Aaronontheweb: Breaking API change here, but it should be fine since that constructor isn't usually called - most of these values come from HOCON.

Aaronontheweb:
> I mean: suppose we have several Write commands, 1st and 2nd without ack and 3d with ack required. Or one ComposedWrite containing this sequence, does not matter.
> Right now, it is sending them one by one - send first, wait until sent, then send second, and after this is done send 3rd (and send ack to it's sender once done).
> When messages do not require ack, we could concatenate them into single Socket.SendAsync. In case above, we could send them all as one message. and once this is done send ack for the third message sender.
> This will reduce buffering, ant speed-up the process as a whole.

What I would do:

1. If we don't have a pending write, send the first message immediately and don't worry about buffering (best response time for low-frequency messaging)
2. If we __do__ have a pending write, buffer pending writes together into a `ComposedWrite` up until we reach the maximum allowable size in the socket, and then flush those all at once. 

How you handle the ACK / NON-ACK part of this.... I would try to separate the concerns and ensuring that buffering works regardless of the ACK / NON-ACK property of each `Tcp.Write`. Might mean having to maintain a list of pending ACKs that need to be sent once the `Socket.SendAsync` call completes. Does that make sense?

Horusiath:
Just to make sure - in the past IO actors where implemented using SendAsync, but it caused serious bugs in the library. Therefore @Aaronontheweb removed them to synchronous Send few years ago. I'm don't know if we have a way to guarantee that this issue will not reappear.

IgorFedchenko:
> I'm don't know if we have a way to guarantee that this issue will not reappear.

I really hope that we have some tests to verify old issues will not get back.
I will take a look on why some tests are failing right now.

> How you handle the ACK / NON-ACK part of this.... I would try to separate the concerns and ensuring that buffering works regardless of the ACK / NON-ACK property of each Tcp.Write. Might mean having to maintain a list of pending ACKs that need to be sent once the Socket.SendAsync call completes. Does that make sense?

@Aaronontheweb  Ah, yes, great idea. 
When buffering is required (there is `SendAsync `in progress), I am accumulating incoming writes in internal queue. And you are right - no need to distinguish between ACK / NON-ACK writes - I will just put them all from queue to the socket at once (as much as socket max send size allows), and will keep list of pending acks from them to send once `SendAsync `is finished. This should increase throughput .

> Do you have any updated benchmark figure since making these changes @IgorFedchenko ?

I did one, but there was no much difference for 1 client - that is definitely because one client does not generate much parallel writes, so our optimization here does not give anything. But still need to run benchmark with multiple parallel clients... Will do that after making en-queued writes sent at one shot.

Aaronontheweb:
@IgorFedchenko 

> I really hope that we have some tests to verify old issues will not get back.

Running the benchmarks with concurrent writes against a single connection should do it.

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: Good point, thanks!

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: @Aaronontheweb Looking at scala's implementation, here is what they do: they are taking `ByteString` that should be transferred, then allocating buffer from buffer pool, and make copy of the bytestring to that buffer. And after `Send` is complete, they are using `Release` method to return this buffer back to the pool.

Our `ByteString` is build on top of `ArraySegment<byte>`, which in turn can be set as `SocketAsyncEventArgs` source buffer for sending in `SendAsync`, without any allocations. I do not see any point is making copy to the buffer and releasing it that (maybe this is done due to some side-effect of scala's socket `Send` implementation?). 

So, I am suggesting to just remove this Release method, because we are not allocating anything here, and nothing to be released.

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: See my reply above.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
IgorFedchenko: Yup, and there is an "infinite" fallback value for HOCON config loading.

Aaronontheweb:
@IgorFedchenko build failure on Linux:

```
IO/TcpConnection.cs(961,27): error CS8107: Feature 'default literal' is not available in C# 7.0. Please use language version 7.1 or greater. [/home/vsts/work/1/s/src/core/Akka/Akka.csproj]
```

Might need to adjust the `<LangVersion>` property inside Akka.csproj

IgorFedchenko:
> but those tests are pretty easy to spot when they fail
Done

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: Update: Just realized that to make buffered writes sent all at once I will need to allocate buffer to store all of them in single array - so maybe Release method will still be relevant. Will see.

Aaronontheweb: Yeah it sounds like object pooling is what they're doing there. We should probably take a look at making `ByteString` castable to something like `Span<T>` or `Memory<T>` in a future release.

IgorFedchenko:
Added batching of buffered writes as we discussed, plus added tests for concurrent writes (no messages should be lost), ordered writes with buffering (messages should arrive in order), and receiving acks for concurrent writes (just to verify none of them are lost while writes are buffered).

@Aaronontheweb Now this PR should be ready for review.

Aaronontheweb:
@IgorFedchenko thanks Igor - we'll see how CI does with it, and then I'll take a look.

Aaronontheweb:
@IgorFedchenko looks like one of the new specs failed:

https://dev.azure.com/dotnet/Akka.NET/_build/results?buildId=24678&view=ms.vss-test-web.build-test-results-tab&runId=1070084&resultId=101443&paneView=debug

Aaronontheweb:
MNTR spec failed due to https://github.com/akkadotnet/akka.net/issues/4035 - probably unrelated to your changes

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -210,14 +226,123 @@ public void BugFix_3021_Tcp_Should_not_drop_large_messages()
...
+                }, RemainingOrDefault, TimeSpan.FromSeconds(2));
 
                 serverMsgs.Sum(s => s.Data.Count).Should().Be(testData.Count*3);
             });
         }
 
-        class Aye : Tcp.Event { public static readonly Aye Instance = new Aye();}
-        class Yes : Tcp.Event { public static readonly Yes Instance = new Yes();}
+        [Fact]
+        public void When_multiple_concurrent_writing_clients_Should_not_loose_messages()
```
Aaronontheweb: Minor typo here - should be "lose," not "loose"

Aaronontheweb: Otherwise the test looks good

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -210,14 +226,123 @@ public void BugFix_3021_Tcp_Should_not_drop_large_messages()
...
+                Parallel.ForEach(clients, client =>
+                {
+                    var msg = ByteString.FromBytes(new byte[1]);
+                    client.Probe.Send(actors.ClientConnection, Tcp.Write.Create(msg, AckWithValue.Create(client.Index)));
+                });
+                
+                // All acks should be received
+                clients.ForEach(client =>
+                {
+                    client.Probe.ExpectMsg<AckWithValue>(ack => ack.Value.ShouldBe(client.Index), TimeSpan.FromSeconds(10));
```
Aaronontheweb: Good

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -116,5 +118,27 @@ public static void SetBuffer(this SocketAsyncEventArgs args, ByteString data)
                 }
             }
         }
+        
+        public static void SetBuffer(this SocketAsyncEventArgs args, IEnumerable<ByteString> dataCollection)
```
Aaronontheweb: Any chance that these changes might help address https://github.com/akkadotnet/akka.net/issues/3921 @IgorFedchenko ?

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -271,34 +286,64 @@ private Receive HandleWriteMessages(ConnectionInfo info)
...
                     case SocketSent _:
+                        // Send ack to sender
                         AcknowledgeSent();
-                        if (IsWritePending)
+                        
+                        // If there is something to send - send it
+                        DoWrite(info, GetAllowedPendingWrite());
+                        
+                        // If message is fully sent, notify sender who sent ResumeWriting command
+                        if (!IsWritePending && _interestedInResume != null)
```
Aaronontheweb: Why do we need to check `IsWritePending` here? Is that how we know that `DoWrite` started processing a new write?

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -271,34 +286,64 @@ private Receive HandleWriteMessages(ConnectionInfo info)
...
+
+                        try
                         {
-                            if (traceLogging) Log.Debug("Dropping write because queue is full");
-                            Sender.Tell(write.FailureMessage);
-                            if (info.UseResumeWriting) SetStatus(ConnectionStatus.WritingSuspended);
+                            if (HasStatus(ConnectionStatus.Sending))
+                            {
+                                // If we are sending something right now, just enqueue incoming write
+                                _writeCommandsQueue.EnqueueSimpleWrites(write, Sender);
```
Aaronontheweb: Good

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -778,110 +818,191 @@ public WriteFileFailed(SocketException cause)
...
+            /// </exception>
+            public void EnqueueSimpleWrites(WriteCommand command, IActorRef sender)
+            {
+                foreach (var writeInfo in ExtractFromCommand(command))
                 {
-                    return _connection.Socket.Send(data.Buffers);
+                    var sizeAfterAppending = _totalSizeInBytes + writeInfo.DataSize;
+                    if (_maxQueueSizeInBytes.HasValue && _maxQueueSizeInBytes.Value < sizeAfterAppending)
+                    {
+                        throw new InternalBufferOverflowException($"Could not receive write command of size {writeInfo.DataSize} bytes, " +
```
Aaronontheweb: If we're going to reject a write because of the buffer being full, we might be better off just having this method return a `bool` - returns `true` if the write can fit, `false` if it can't. Throwing exceptions is expensive (from a throughput standpoint) and I think this is an area where we can avoid doing it. What do you think @IgorFedchenko @Horusiath ?

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -778,110 +818,191 @@ public WriteFileFailed(SocketException cause)
...
+                        first = writeInfo.Command;
+                        continue;
                     }
-                    finally
+                    
+                    var sizeAfterAppending = _totalSizeInBytes + writeInfo.DataSize;
+                    if (_maxQueueSizeInBytes.HasValue && _maxQueueSizeInBytes.Value < sizeAfterAppending)
                     {
-                        _connection.Socket.Blocking = false;
+                        throw new InternalBufferOverflowException($"Could not receive write command of size {writeInfo.DataSize} bytes, " +
```
Aaronontheweb: I wouldn't throw here - I'd just skip any additional writes once the buffer is full enough. Pick it up on the next pass.

The only time we should throw is if a write is so large that it could never possibly fit in the socket buffer in the first place - does that make sense?

Aaronontheweb: Or am I misunderstanding the code here?

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: One more update: turned out that `SocketAsyncEventArgs` are using same `IList<ArraySegment<byte>>` as our `ByteString` does, so no more allocations were required to pass data into socket. Nice.

src/core/Akka/IO/SocketEventArgsPool.cs
IgorFedchenko: I do not think so. My update here is related to `SocketAsyncEventArgs` buffers initialization before sending data to socket. And in referenced issue they are not writing, but reading corrupted messages (also, since it is opened for 24 days since last commit, I guess they do not reproduce issue anymore). So, no, this one should not be related.

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: `IsWritePending` means "do we have any writes buffered and are we allowed to send them?". 

Somewhere there lives actor, that is waiting when `TcpConnection` actor will resume paused writing - it is waiting for `WritingResumed` message to start sending his `Write` messages. 
So, before letting him know that he is allowed to send anything, we want to make buffers empty. 
Generally speaking, write will be suspended only when there is buffer overflow and we can not receive writes anymore.
Actually, we could not to check `IsWritePending` in updated implementation, because usually if `SendAsync` finished and write was suspended means that all buffer is already batched to the wire and no more writes were received since `SendAsync` was started (writing was paused). But on some abstract layer checking that buffer is empty before allowing to send more after buffer was overfilled may be convenient. 

I do not fully like this `IsWritePending` naming here, it is taken from akka implementation. Maybe change it to `IsWritePendingAndAllowed`...?

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: Indeed, you are right. Will refactor this right now.

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
IgorFedchenko: Thanks! Fixed

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: Mm... The idea is that if (this is optional) `TcpConnection`'s internal writes queue is limited in size, we should start dropping incoming writes. 
As akka's docs says, there are few strategies do deal with such overflows. Basically, your actor either is waiting for ack messages before send next write, either use auto-write-suspending thing and just write until getting `CommandFailed` response, after which wait for `WritingResumed` message and go on.

So, here I was throwing exception to catch it with `TcpConnection` actor and send `CommandFailed` (+ suspend writing). Now it is just returning `bool` instead of throwing, you are absolutely write, no reason to throw here.

Also, if incoming write is too large, but we can send it immediately, we will not give an error because there is no particular limit for buffers in `SocketAsyncEventArgs`. But yes, if we have to buffer write and it is too big and hits the limit - we will... not throw now, but just get `false` response from internal `PendingSimpleWritesQueue` and report failure to the sender.

Did I got your point here?

IgorFedchenko:
@Aaronontheweb Added changes you requested - and will submit updated benchmark results soon.

src/core/Akka/IO/TcpConnection.cs
Aaronontheweb: Very nice

src/core/Akka/IO/TcpConnection.cs
Aaronontheweb: > Also, if incoming write is too large, but we can send it immediately, we will not give an error because there is no particular limit for buffers in SocketAsyncEventArgs. But yes, if we have to buffer write and it is too big and hits the limit - we will... not throw now, but just get false response from internal PendingSimpleWritesQueue and report failure to the sender.

Yeah, that sounds right - we're never going to be able to fit a 4GB message into a 256k buffer, so we always need to reject that. But otherwise we can buffer and wait. Sounds good.

IgorFedchenko:
@Aaronontheweb 
> do you have any updated performance benchmarks with this code?

Here is an updated table. Table before this PR is [here](https://github.com/akkadotnet/akka.net/pull/4019#issuecomment-549269987)

```
Intel Core i7-4770 CPU 3.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100
  [Host]     : .NET Core 2.1.11 (CoreCLR 4.6.27617.04, CoreFX 4.6.27617.02), X64 RyuJIT
  Job-MWFOPX : .NET Core 2.1.11 (CoreCLR 4.6.27617.04, CoreFX 4.6.27617.02), X64 RyuJIT

InvocationCount=1  IterationCount=100  LaunchCount=1  
RunStrategy=Monitoring  UnrollFactor=1  WarmupCount=1  

|                    Method | MessageCount | MessageLength | ClientsCount |        Mean |     Error |     StdDev |      Median |     Gen 0 |     Gen 1 | Gen 2 | Allocated |
|-------------------------- |------------- |-------------- |------------- |------------:|----------:|-----------:|------------:|----------:|----------:|------:|----------:|
| ClientServerCommunication |          100 |            10 |            1 |    43.27 ms |  1.151 ms |   3.395 ms |    42.11 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |            3 |    24.63 ms |  0.851 ms |   2.509 ms |    23.85 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |            5 |    21.70 ms |  0.429 ms |   1.264 ms |    21.39 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |            7 |    22.13 ms |  0.378 ms |   1.115 ms |    21.82 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |           10 |    24.52 ms |  1.023 ms |   3.018 ms |    23.96 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |           20 |    37.59 ms |  1.273 ms |   3.752 ms |    37.12 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |           30 |    49.73 ms |  1.443 ms |   4.255 ms |    48.69 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |           40 |    66.67 ms |  1.695 ms |   4.998 ms |    65.81 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |            1 |    43.80 ms |  0.883 ms |   2.603 ms |    43.17 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |            3 |    22.54 ms |  0.442 ms |   1.303 ms |    22.16 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |            5 |    21.69 ms |  0.410 ms |   1.210 ms |    21.24 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |            7 |    22.84 ms |  0.604 ms |   1.782 ms |    22.30 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |           10 |    23.87 ms |  0.356 ms |   1.051 ms |    23.60 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |           20 |    34.25 ms |  0.253 ms |   0.746 ms |    34.16 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |           30 |    46.88 ms |  0.470 ms |   1.385 ms |    46.56 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |           40 |    64.10 ms |  1.189 ms |   3.504 ms |    62.85 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |            1 |   405.63 ms | 11.219 ms |  33.079 ms |   395.01 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |            3 |   190.53 ms |  6.840 ms |  20.167 ms |   184.19 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |            5 |   165.34 ms |  8.664 ms |  25.547 ms |   154.21 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |            7 |   136.30 ms |  3.306 ms |   9.747 ms |   132.45 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |           10 |   134.90 ms |  3.416 ms |  10.072 ms |   131.21 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |           20 |   145.29 ms |  6.544 ms |  19.296 ms |   138.41 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |           30 |   149.52 ms |  4.165 ms |  12.282 ms |   145.17 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |           40 |   153.16 ms |  3.160 ms |   9.318 ms |   150.23 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |            1 |   411.42 ms |  8.725 ms |  25.725 ms |   405.61 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |            3 |   180.59 ms |  3.926 ms |  11.577 ms |   177.37 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |            5 |   149.23 ms |  3.330 ms |   9.820 ms |   145.82 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |            7 |   136.76 ms |  2.668 ms |   7.867 ms |   133.38 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |           10 |   134.08 ms |  3.845 ms |  11.338 ms |   130.03 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |           20 |   139.23 ms |  3.499 ms |  10.316 ms |   136.87 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |           30 |   147.10 ms |  5.524 ms |  16.286 ms |   141.99 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |           40 |   160.41 ms |  4.223 ms |  12.452 ms |   157.92 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |            1 | 3,910.47 ms | 50.201 ms | 148.020 ms | 3,861.83 ms | 7000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |            3 | 1,701.91 ms | 23.092 ms |  68.087 ms | 1,669.61 ms | 7000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |            5 | 1,389.60 ms | 18.536 ms |  54.653 ms | 1,365.81 ms | 7000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |            7 | 1,346.76 ms | 42.246 ms | 124.562 ms | 1,320.29 ms | 7000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |           10 | 1,244.45 ms | 51.119 ms | 150.724 ms | 1,183.69 ms | 7000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |           20 | 1,153.84 ms | 23.224 ms |  68.477 ms | 1,119.97 ms | 7000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |           30 | 1,158.50 ms | 16.651 ms |  49.095 ms | 1,136.47 ms | 8000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |           40 | 1,170.24 ms | 36.331 ms | 107.122 ms | 1,123.18 ms | 8000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |            1 | 4,001.42 ms | 61.434 ms | 181.140 ms | 3,940.78 ms | 8000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |            3 | 1,737.26 ms | 30.539 ms |  90.045 ms | 1,701.13 ms | 8000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |            5 | 1,461.54 ms | 32.706 ms |  96.435 ms | 1,437.93 ms | 8000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |            7 | 1,333.24 ms | 41.302 ms | 121.781 ms | 1,300.18 ms | 8000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |           10 | 1,242.71 ms | 25.839 ms |  76.187 ms | 1,218.65 ms | 8000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |           20 | 1,144.46 ms | 28.467 ms |  83.936 ms | 1,117.14 ms | 8000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |           30 | 1,128.19 ms | 20.453 ms |  60.307 ms | 1,106.81 ms | 8000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |           40 | 1,161.90 ms | 20.751 ms |  61.185 ms | 1,139.57 ms | 9000.0000 | 1000.0000 |     - |   1.58 KB |
```

Almost no improvements, some numbers are even bigger... But my CPU was loaded for 100%, so any other processes running in parallel might have impact on the results. 
I will try to run benchmark on `dev` branch again.

Also, maybe our improvements are for more high-loaded scenario, then request-reply from 40 actors to one server actor.

Aaronontheweb:
@IgorFedchenko what those benchmark numbers tell me is that the socket performance over localhost is probably not the bottleneck. We should take a look at the dispatcher and maybe the actors themselves.

Aaronontheweb:
But this change is still good - in areas where I/O might really be the bottleneck (over a real network) these changes will certainly help.

src/core/Akka/IO/TcpConnection.cs
Aaronontheweb: @IgorFedchenko I think it's probably fine - but up to you.

IgorFedchenko:
> @IgorFedchenko what those benchmark numbers tell me is that the socket performance over localhost is probably not the bottleneck. We should take a look at the dispatcher and maybe the actors themselves.

I agree with you.

Aaronontheweb:
@IgorFedchenko we've been getting a persistent MNTR failure here, and I believe that's not showing up in the test results due to https://github.com/akkadotnet/akka.net/issues/4035

IgorFedchenko:
@Aaronontheweb We are still getting error - and I am able to reproduce this locally:
```
[ERROR][22.11.2019 14:40:28][Thread 0041][akka://TestRunnerLogging/system/IO-TCP/$a] Monitored actor [[akka://TestRunnerLogging/user/TcpLogger#73156647]] terminated
Cause: Akka.Actor.DeathPactException: Monitored actor [[akka://TestRunnerLogging/user/TcpLogger#73156647]] terminated
   в Akka.Actor.ActorBase.Unhandled(Object message)
   в Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message)
   в Akka.Actor.ActorCell.ReceiveMessage(Object message)
   в Akka.Actor.ActorCell.ReceivedTerminated(Terminated t)
   в Akka.Actor.ActorCell.AutoReceiveMessage(Envelope envelope)
   в Akka.Actor.ActorCell.Invoke(Envelope envelope)
```

Will try to fix this.

IgorFedchenko:
@Aaronontheweb  Added #4058 for this issue - will merge my branch here and see if this will help

Aaronontheweb:
Looks like there's a persistent failure with the `ClusterClientSpecs` in the MNTR.

Aaronontheweb:
Good chance they are not related to these changes, but we need to investigate.

IgorFedchenko:
> Good chance they are not related to these changes, but we need to investigate.

I am almost sure that MNTR failures are not related, because the same failures are happening in #4057 , and nothing from this PR is implemented there.

@Aaronontheweb if otherwise this PR looks good it can be merged? I am going to make separate PR for MNTR issue for this failure anyway.

Aaronontheweb:
Sounds good.

PR #4032, opened 2019-11-12
HashedWheelTimerScheduler waiting overhead fix

### LINKED ISSUE #4031: HashedWheelTimerScheduler is spending 35% of execution time in sleep
I used Akka.IO `TcpOperationBenchmark` as a sample to apply profiling, and it turns out that it is spending ~35% of time in sleep (`Sleep` function is my wrapper over `Thread.Sleep` that is used in scheduler). See the picture:

image

So, 1/3 of execution time scheduler is waiting for the next tick. This may slow down actors messaging pretty much. I am going to open PR for it.

IgorFedchenko:
Close #4031 

As stated in original issue, scheduler is waiting more then it should.

For this PR, I have added simple test, based on one of Akka.IO benchmarks - it is performing some communication via Akka.IO.Tcp, and uses some counters in `HashedWheelTimerScheduler` to track, how much ticks it should be waiting each time before performing scheduled actions, and how much it is actually working.

Then, there are few simple assertions:
```csharp
msActual.ShouldBeLessThan(msRequired * 1.1M, "We absolutelly do not want scheduler to have more then 10% sleep overhead");
msActual.ShouldBeLessThan(msRequired * 1.05M, "We do not want scheduler to have more then 5% sleep overhead");
msActual.ShouldBeLessThan(msRequired * 1.01M, "Would be really nice for scheduler to have less then 1% sleep overhead");
```

This allows to keep track of how much overhead do we have.

Original implementation was ceiling required ticks to closest milliseconds value, and was making `Thread.Sleep` on it. This was giving much overhead in cases when sleep is going to be short (less then ~10-15 ms on Windows).

After some experiments, I ended up with pretty simple solution - using empty loop with `Stopwatch.ElapsedTicks` condition gives much more accurate results.
This is what I was getting in test output with initial implementation:
```
We do not want scheduler to have more then 5% sleep overhead
Required wait time is 19,0089815ms, and actual is 20,5070609ms

We absolutelly do not want scheduler to have more then 10% sleep overhead
Required wait time is 18,2111832ms, and actual is 20,6246174ms
```

And this is what I am getting with updated one (test passes of course):
```
Required wait time is 21,7370645ms, and actual is 21,7728415ms

Required wait time is 21,1151001ms, and actual is 21,1228956ms

Required wait time is 22,147157ms, and actual is 22,1707158ms

Required wait time is 22,0152161ms, and actual is 22,0428943ms
```

Time loss is much smaller.

P.S. With that said, `Sleep` is still for good when we need to wait more then 10-15 ms. Not sure how often this can be the case, but still, I made `Sleep` function smart enough to use `Thread.Sleep` instead of looping when required delay is big enough (scala's implementation calls "big enough" 1ms under linux and 10ms under windows).

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj
@@ -1,12 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\..\common.props" />
-    
-    <!-- use the xunit JSON settings file to disable parallel tests running-->
-    <ItemGroup>
-        <Content Include="..\..\..\xunit.runner.json" Link="xunit.runner.json">
-            <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-        </Content>
-    </ItemGroup>
+  <Import Project="..\..\..\xunitSettings.props" />
 
   <PropertyGroup>
     <AssemblyTitle>Akka.Cluster.Sharding.Tests</AssemblyTitle>
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj
@@ -1,5 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\..\common.props" />
+    
+    <!-- use the xunit JSON settings file to disable parallel tests running-->
+    <ItemGroup>
+        <Content Include="..\..\..\xunit.runner.json" Link="xunit.runner.json">
+            <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+        </Content>
+    </ItemGroup>
 
   <PropertyGroup>
     <AssemblyTitle>Akka.Cluster.Sharding.Tests</AssemblyTitle>
--- a/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
+++ b/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
@@ -148,7 +148,7 @@ namespace Akka.Actor
         /// <summary>
         /// Scheduler thread entry method
         /// </summary>
-        private void Run()
+        private async Task Run()
         {
             // Initialize the clock
             _startTime = HighResMonotonicClock.Ticks;
@@ -162,7 +162,7 @@ namespace Akka.Actor
 
             do
             {
-                var deadline = WaitForNextTick();
+                var deadline = await WaitForNextTick();
                 if (deadline > 0)
                 {
                     var idx = (int)(_tick & _mask);
@@ -202,7 +202,7 @@ namespace Akka.Actor
             _rescheduleRegistrations.Clear();
         }
 
-        private long WaitForNextTick()
+        private async Task<long> WaitForNextTick()
         {
             var deadline = _tickDuration * (_tick + 1);
             unchecked // just to avoid trouble with long-running applications
@@ -225,7 +225,7 @@ namespace Akka.Actor
 #if UNSAFE_THREADING
                     try
                     {
-                        Sleep(ticksToSleep);
+                        await Sleep(ticksToSleep);
                     }
                     catch (Exception ex) when (ex is TaskCanceledException || ex is OperationCanceledException)
                     {
@@ -233,7 +233,7 @@ namespace Akka.Actor
                             return long.MinValue;
                     }
 #else             
-                    Sleep(ticksToSleep);
+                    await Sleep(ticksToSleep);
 #endif
                     
                     stopWatch.Stop();
@@ -243,16 +243,10 @@ namespace Akka.Actor
             }
         }
         
-        private void Sleep(long ticks)
+        private async Task Sleep(long ticks)
         {
-            var remainingTicks = (int)Math.Min(int.MaxValue, ticks);
-            _sleepWatch.Restart();
-            while (remainingTicks > 0)
-            {
-                Thread.SpinWait(remainingTicks);
-                remainingTicks = (int)(ticks - _sleepWatch.ElapsedTicks);
-            }
-            _sleepWatch.Stop();
+            var ms = (ticks + TimeSpan.TicksPerMillisecond - 1) / TimeSpan.TicksPerMillisecond;
+            await Task.Delay(TimeSpan.FromMilliseconds(ms));
         }
 
         private void TransferRegistrationsToBuckets()
@@ -289,7 +283,6 @@ namespace Akka.Actor
             var bucket = _wheel[stopIndex];
             bucket.AddRegistration(reg);
         }
-
 
         /// <summary>
         /// TBD
--- a/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
+++ b/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
@@ -252,6 +252,11 @@ namespace Akka.Actor
             }
             else
             {
+                // Sleep some ms anyway, if there is at least one tick
+                sleepMs = (ticks + TimeSpan.TicksPerMillisecond - 1) / TimeSpan.TicksPerMillisecond;
+                await Task.Delay(TimeSpan.FromMilliseconds(sleepMs));
+                return;
+                
                 // If there is no much time to sleep, let's be more accurate
                 _sleepWatch.Restart();
                 while (_sleepWatch.ElapsedTicks < ticks) { /*waiting*/ }
--- a/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
+++ b/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
@@ -8,12 +8,14 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Configuration;
 using Akka.Dispatch;
 using Akka.Event;
 using Akka.Util;
+using Akka.Util.Internal;
 
 // ReSharper disable NotResolvedInText
 
@@ -36,6 +38,10 @@ namespace Akka.Actor
     {
         private readonly TimeSpan _shutdownTimeout;
         private readonly long _tickDuration; // a timespan expressed as ticks
+
+        public static AtomicCounter TotalTicksRequiredToWaitStrict = new AtomicCounter(0);
+        public static AtomicCounter TotalTicksRequiredToWaitRounded = new AtomicCounter(0);
+        public static AtomicCounter TotalTicksActual = new AtomicCounter(0);
 
         /// <summary>
         /// TBD
@@ -110,6 +116,7 @@ namespace Akka.Actor
 
         private readonly HashSet<SchedulerRegistration> _unprocessedRegistrations = new HashSet<SchedulerRegistration>();
         private readonly HashSet<SchedulerRegistration> _rescheduleRegistrations = new HashSet<SchedulerRegistration>();
+        private readonly Stopwatch _sleepWatch = new Stopwatch();
 
         private Thread _worker;
 
@@ -216,31 +223,48 @@ namespace Akka.Actor
                 for (;;)
                 {
                     long currentTime = HighResMonotonicClock.Ticks - _startTime;
-                    var sleepMs = ((deadline - currentTime + TimeSpan.TicksPerMillisecond - 1) / TimeSpan.TicksPerMillisecond);
+                    var sleepMs = RuntimeDetector.IsWindows
+                        ? ((deadline - currentTime + TimeSpan.TicksPerMillisecond * 5 - 1) / (TimeSpan.TicksPerMillisecond * 10) * 10)
+                        : ((deadline - currentTime + TimeSpan.TicksPerMillisecond - 1) / TimeSpan.TicksPerMillisecond);
 
                     if (sleepMs <= 0) // no need to sleep
                     {
                         if (currentTime == long.MinValue) // wrap-around
                             return -long.MaxValue;
                         return currentTime;
-
                     }
 
+                    var stopWatch = Stopwatch.StartNew();
+                    
 #if UNSAFE_THREADING
                     try
                     {
-                        Thread.Sleep(TimeSpan.FromMilliseconds(sleepMs));
+                        Sleep(deadline - currentTime);
+                        // Thread.Sleep(TimeSpan.FromMilliseconds(sleepMs));
                     }
                     catch (ThreadInterruptedException)
                     {
                         if (_workerState == WORKER_STATE_SHUTDOWN)
                             return long.MinValue;
                     }
-#else
-                    Thread.Sleep(TimeSpan.FromMilliseconds(sleepMs));
+#else             
+                    Sleep(deadline - currentTime);
+                    // Thread.Sleep(TimeSpan.FromMilliseconds(sleepMs));
 #endif
+                    
+                    stopWatch.Stop();
+                    TotalTicksRequiredToWaitStrict.AddAndGet((int)Math.Min(deadline - currentTime, int.MaxValue));
+                    TotalTicksRequiredToWaitRounded.AddAndGet((int)Math.Min(sleepMs * TimeSpan.TicksPerMillisecond, int.MaxValue));
+                    TotalTicksActual.AddAndGet((int)Math.Min(stopWatch.ElapsedTicks, int.MaxValue));
                 }
             }
+        }
+        
+        private void Sleep(long ticks)
+        {
+            _sleepWatch.Restart();
+            while (_sleepWatch.ElapsedTicks < ticks) { /*waiting*/ }
+            _sleepWatch.Stop();
         }
 
         private void TransferRegistrationsToBuckets()
--- a/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
+++ b/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
@@ -236,7 +236,7 @@ namespace Akka.Actor
 #endif
                     
                     stopWatch.Stop();
-                    TotalTicksRequiredToWaitStrict.AddAndGet((int)Math.Min(deadline - currentTime, int.MaxValue));
+                    TotalTicksRequiredToWaitStrict.AddAndGet((int)Math.Min(ticksToSleep, int.MaxValue));
                     TotalTicksActual.AddAndGet((int)Math.Min(stopWatch.ElapsedTicks, int.MaxValue));
                 }
             }
@@ -244,9 +244,13 @@ namespace Akka.Actor
         
         private void Sleep(long ticks)
         {
-            // If there is no much time to sleep, let's be more accurate
+            var remainingTicks = (int)Math.Min(int.MaxValue, ticks);
             _sleepWatch.Restart();
-            while (_sleepWatch.ElapsedTicks < ticks) { /*waiting*/ }
+            while (remainingTicks > 0)
+            {
+                Thread.SpinWait(remainingTicks);
+                remainingTicks = (int)(ticks - _sleepWatch.ElapsedTicks);
+            }
             _sleepWatch.Stop();
         }
 
--- a/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
+++ b/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
@@ -117,20 +117,17 @@ namespace Akka.Actor
         private readonly HashSet<SchedulerRegistration> _rescheduleRegistrations = new HashSet<SchedulerRegistration>();
         private readonly Stopwatch _sleepWatch = new Stopwatch();
 
-        private Thread _worker;
-
         private void Start()
         {
             if (_workerState == WORKER_STATE_STARTED) { } // do nothing
             else if (_workerState == WORKER_STATE_INIT)
             {
-                _worker = new Thread(Run) { IsBackground = true };
 #pragma warning disable 420
                 if (Interlocked.CompareExchange(ref _workerState, WORKER_STATE_STARTED, WORKER_STATE_INIT) ==
 #pragma warning restore 420
                     WORKER_STATE_INIT)
                 {
-                    _worker.Start();
+                    Task.Run(Run);
                 }
             }
 
@@ -143,24 +140,14 @@ namespace Akka.Actor
                 throw new InvalidOperationException($"Worker in invalid state: {_workerState}");
             }
 
-            while (_startTime == 0)
-            {
-#if UNSAFE_THREADING
-                try
-                {
-                    _workerInitialized.Wait();
-                }
-                catch (ThreadInterruptedException) { }
-#else
-                _workerInitialized.Wait();
-#endif
-            }
+            // Wait until _startTime is initialized in Run task
+            _workerInitialized.Wait();
         }
 
         /// <summary>
         /// Scheduler thread entry method
         /// </summary>
-        private void Run()
+        private async Task Run()
         {
             // Initialize the clock
             _startTime = HighResMonotonicClock.Ticks;
@@ -174,7 +161,7 @@ namespace Akka.Actor
 
             do
             {
-                var deadline = WaitForNextTick();
+                var deadline = await WaitForNextTick();
                 if (deadline > 0)
                 {
                     var idx = (int)(_tick & _mask);
@@ -214,7 +201,7 @@ namespace Akka.Actor
             _rescheduleRegistrations.Clear();
         }
 
-        private long WaitForNextTick()
+        private async Task<long> WaitForNextTick()
         {
             var deadline = _tickDuration * (_tick + 1);
             unchecked // just to avoid trouble with long-running applications
@@ -237,15 +224,15 @@ namespace Akka.Actor
 #if UNSAFE_THREADING
                     try
                     {
-                        Sleep(ticksToSleep);
+                        await Sleep(ticksToSleep);
                     }
-                    catch (ThreadInterruptedException)
+                    catch (Exception ex) when (ex is TaskCanceledException || ex is OperationCanceledException)
                     {
                         if (_workerState == WORKER_STATE_SHUTDOWN)
                             return long.MinValue;
                     }
 #else             
-                    Sleep(ticksToSleep);
+                    await Sleep(ticksToSleep);
 #endif
                     
                     stopWatch.Stop();
@@ -255,13 +242,13 @@ namespace Akka.Actor
             }
         }
         
-        private void Sleep(long ticks)
+        private async Task Sleep(long ticks)
         {
             var sleepMs = ticks / TimeSpan.TicksPerMillisecond;
             if (sleepMs >= 1 && !RuntimeDetector.IsWindows || sleepMs >= 10) // Windows clocks are too slow
             {
                 // If sleep is going to be long enough, let it release unused CPU resources for that time
-                Thread.Sleep(TimeSpan.FromMilliseconds(sleepMs));
+                await Task.Delay(TimeSpan.FromMilliseconds(sleepMs));
             }
             else
             {
--- a/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
+++ b/src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
@@ -38,19 +38,45 @@ namespace Akka.Actor
     /// </summary>
     public class HashedWheelTimerScheduler : SchedulerBase, IDateTimeOffsetNowTimeProvider, IDisposable
     {
+        /// <summary>
+        /// When there are too small number of ticks to wait, use SpinWait instead
+        /// </summary>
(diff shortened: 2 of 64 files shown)


Aaronontheweb:
Good idea - not sure why the Cluster.Sharding specs all failed the first time around. Might be that these changes with the timer may have affected how long certain things happen - i.e. tasks that used to take ~n seconds now take < n.

Aaronontheweb:
Good idea regarding `async` / `await` - we need to get into the habit of not blocking threads when we're waiting for work.

Aaronontheweb:
@IgorFedchenko looks like we're having issues with these sharding specs again, but only on Windows - doesn't look like that issue came up at all on Linux. Would you mind looking into it?

IgorFedchenko:
> Would you mind looking into it?

@Aaronontheweb Sure

Aaronontheweb:
Looks like the multi-node specs don't start properly or hang with these changes too.... Weird.

IgorFedchenko:
@Aaronontheweb Sharding tests pass well on my local machine. And all of them are failing due to timeouts, which makes me thing that they can be racy too.

Let me set larger timeout for few of them, just for experiment.

IgorFedchenko:
@Aaronontheweb Seems like Sharding tests are pretty heavy... Should we set then into xunit test collection to run one-by-one?

Aaronontheweb:
@IgorFedchenko yeah, we can disable parallelization there by using collections or by doing what we did here:

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote.Tests/Akka.Remote.Tests.csproj#L3

That will use the `xunit.runner.json` settings to disable parallelization for the entire DLL.

IgorFedchenko:
> That will use the xunit.runner.json settings to disable parallelization for the entire DLL.

Awesome

src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj
```diff
@@ -1,5 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\..\common.props" />
+    
+    <!-- use the xunit JSON settings file to disable parallel tests running-->
```
IgorFedchenko: Could not just reference `<Import Project="..\..\xunitSettings.props" />` here, because this `props` file references `..\..\xunit.runner.json` relative path, which is not valid here (need one more `..\`). Possibly there is some cleaner war do do that.

Aaronontheweb: Ah... I'm sure there is using some MsBuild variables, but good point. May as well just use collection fixtures then.

IgorFedchenko:
It helped pretty well - most of failing tests passed. But still, few tests failing due timings, and pass locally. I will try to increase timeouts...

IgorFedchenko:
@Aaronontheweb All right, seems like the only result I can get is occasionally get all tests passed single time, before some of them will hit some timeouts again. Some tests are racy, and there is nothing I can do with this (quickly). One time they pass, and the next time they do not.

I will rollback timeout changes, since they are not very helpful.

Aaronontheweb:
Why does the MNTR not run for this PR?

IgorFedchenko:
Updated `pr-validation.yaml` file to have each job in separate stage to make them run sequentially. For me, this racy issues and failing to run (and even to cancel) MNTR tests may indicate some serious resource starvation somewhere there on CI build server. In theory this may be a root of the issue. So, as an experiment, will try to run all jobs sequentially, and if issues will still be there, will also try to run all tests per assembly sequentially (with adding xunit runner settings into common.props).

IgorFedchenko:
Failing again on Sharding tests, with some timeouts... Does not look like resources problem now.

Maybe this looping instead of sleep in scheduler takes too much resources then? Let me check if making it use "Task.Delay" again will fix issues with this tests

IgorFedchenko:
So, without looping instead of sleep MNTR works...

IgorFedchenko:
And with looping instead of sleep it does not run again... Very weird.

Zetanova:
If possible, Thread.Sleep should be omitted. On busy windows boxes it can wait 10x longer as u want
(+>10 seconds for one Thread.Sleep)

Instead of ```while (_sleepWatch.ElapsedTicks < ticks) { /*waiting*/ }``` a ```System.Threading.SpinWait``` should be used. It takes Intel hyperthreading and other stuff into account.

IgorFedchenko:
@Zetanova Did not know that `Thread.Sleep` can be so inaccurate - thanks for tip! 

Let's try using `Threading.SpinWait` then. I was thinking about that, but decided to move on with looping because it does not accept `ticks` - it is working with `iterations` instead, and the actual time may depend on the processor speed (as stated in the [docs](https://docs.microsoft.com/ru-ru/dotnet/api/system.threading.spinwait.spinuntil?view=netframework-4.8)).

But lets try to do something like this:
```csharp
private void Sleep(long ticks)
{
	var remainingTicks = (int)Math.Min(int.MaxValue, ticks);
	_sleepWatch.Restart();
	while (remainingTicks > 0)
	{
		Thread.SpinWait(remainingTicks);
		remainingTicks = (int)(ticks - _sleepWatch.ElapsedTicks);
	}
	_sleepWatch.Stop();
}
``` 

This still gives great results in terms of saving time (passes all tests), but should take additional stuff into account.

IgorFedchenko:
Seems like MNTR at least started :)

Aaronontheweb:
> If possible, Thread.Sleep should be omitted. On busy windows boxes it can wait 10x longer as u want
> (+>10 seconds for one Thread.Sleep)
> 
> Instead of `while (_sleepWatch.ElapsedTicks < ticks) { /*waiting*/ }` a `System.Threading.SpinWait` should be used. It takes Intel hyperthreading and other stuff into account.

I had no idea. Really useful to know!

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
```diff
@@ -110,21 +115,19 @@ private static int NormalizeTicksPerWheel(int ticksPerWheel)
...
             else if (_workerState == WORKER_STATE_INIT)
             {
-                _worker = new Thread(Run) { IsBackground = true };
 #pragma warning disable 420
                 if (Interlocked.CompareExchange(ref _workerState, WORKER_STATE_STARTED, WORKER_STATE_INIT) ==
 #pragma warning restore 420
                     WORKER_STATE_INIT)
                 {
-                    _worker.Start();
+                    Task.Run(Run);
```
Aaronontheweb: I think this might be why we're having issues on the MNTR - the design before called for the `Scheduler` to run on a dedicated thread and periodically sleep / wait itself. This was intentional - we wanted the dedicated thread so other items piling up inside the `TaskScheduler`'s queue couldn't affect time-sensitive jobs performed by the scheduler.

So here's my suggestion: - use a custom `TaskScheduler` that runs on a single, dedicated thread (I think you can find an example inside the `ForkJoinExecutor` code inside the Akka.Dispatch namespace) and have the scheduler run on that. No more work queue issues - the only competition will be for the CPU itself as the operating system scheduler level. Moreover, we can also use `await` and `Task.Delay` instead of sleep to see if that reduces CPU utilization coming from the scheduler (wasted cycles on `Thread.Sleep`) - and if that doesn't work, we can go back to using `Thread.SpinWait`.

Does that sound reasonable?

IgorFedchenko: The first part is clear, and that's really good idea, I will work on this tomorrow.

But did not get why should be get back to `Task.Delay`? Most of the time scheduler needs to wait less then `10ms` for the next tick - and `Task.Delay` will always give us more sleep time (the same +10% as it was). 
What can be done is still use `Task.Delay` for pauses more then `10ms` - they are possible if scheduler tick duration will be set to something >10ms via `akka.scheduler.tick-duration` setting.
@Aaronontheweb Or we should get back to `Task.Delay` because of some other benefits?

Aaronontheweb: > Or we should get back to Task.Delay because of some other benefits?

Ah, nah, nevermind - for some reason I assumed that `Task.Delay` could use an I/O completion port and not use CPU when idling... You'd need to look at the reference source to be sure. Thread.SpintWait is probably best.

IgorFedchenko:
So, MNTR is finally running (thanks @Aaronontheweb for the great idea about tasks starvation and custom `TaskScheduler` usage). 
But it is failing. Will take a look, why.

Zetanova:
@IgorFedchenko Maybe u are fighting the wheel? 
To use a synchronisation Context (TaskScheduler) that limit to one concurrency is will not work. 
Because to many libs assume multi-concurrency support at least single threaded. 

The algo should be:
x) Add the work into buckets by supposed schedule time, but never to an obsolete bucket.  
x) Run a single thread in a loop that schedules all work from all obsolete buckets and sleeps for bucket period after (with a system timer and waithandle).

If the bucket-period is 10ms  then new work will be scheduled 1x-2x of the period even when a Schedule time of 1ms is requested. 

Because the scheduler is using a dedicated thread the period can lowered from 10ms to 2ms or less.
The schedule work exact will never work. after all the scheduler has no control over the ActorCell Dispatcher.  SpinWaits should only be for fraction of much less then 1ms. 
They are used for lock-free sync between CPU-cores/cache. Else u will generate 1-5% CPU load in the idle process.

IgorFedchenko:
@Zetanova So, you think that there is almost nothing to do with this sleeping overhead? We have added custom scheduler, that will use single thread but allow to use `Task.Delay` instead of `Thread.Sleep` to save some resources (and maybe this will also reduce sleep time), but this is all we can do here? Maybe you are right, in terms that there will always be trade-off between having scheduler sleeping less time and CPU load. 

Existing scheduling algorithm that was already implemented here before my PR already looks good for me - and, it is organized the same way as in [scala](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/LightArrayRevolverScheduler.scala), so even if some optimizations are possible I do not think it may be any kind of bottleneck.

After some local testing of latest version, I have also noticed much higher CPU load. So I replaced `Thread.SpanWait` to `Task.Delay` under new custom `TaskScheduler` that will execute it on single thread. If this will help with MNTR failures, I will run Akka.IO benchmark to see if `async/await` usage on single thread will give any improvements. But not sure about that, since we are releasing thread that is not used by anyone else.

IgorFedchenko:
So now it is working, mostly as before, and I do not expect much throughput improvements... But feeling that something is not right is still there :) 
We want to wait for 10ms or less between each bucket registrations execution. And each time we will wait for at least ~15ms. Each time. My sleep overhead test shows sleeping 28 seconds instead of requested 15 seconds now (I have commented out this check for now, but maybe will bring it back).
Yes, we can not afford to spend so much CPU cycles on `Thread.SpinWait`, but I'll try some more ideas to change this... 
If one thread should either sleep for 15+ms or not sleep at all, maybe worth trying to start two threads with some initial time shift to make them both trying to handle buckets scheduling. I.e. the one who has woken up closer to weel tick will execute bucket tasks and sleep again. sometimes this may give us more accurate pauses between weel ticks, with the cost of two threads.
Also, we can make number of threads configurable, because obviously the more threads are trying to catch tick moment, the more accurate it will be scheduled.
A kind of crazy idea, but might work? I.e. two threads with two loops, each is waiting for tick, and once awaken first of them marks tick as handled and executes bucket tasks, and second starts sleeping for next tick immediately (and therefore has more chances to wake up earlier next time).
Is it worth trying? Or I am missing something?

Zetanova:
You are running in the normal wrong assumptions:
1) 1ms is a very short time.  
=> It is not for the CPU, SpinWaits are for a few tousend interations (less then 10k). 
Thats on a 3.4Ghz CPU how long? 

2) A process can tell the CPU to wait for a time.
=> It can not, it can do work or tell the OS not to require the thread anymore.

Aaronontheweb:
@IgorFedchenko @Zetanova so here's what I'd propose:

- Let's start with ditching the `TaskScheduler` and stick with a dedicated thread in order to keep things simple
- Let's focus on improving the CPU utilization of that thread, when at all possible. 

For instance, `Thread.Sleep` may not be that big of a performance hit in the grand scheme of things:

* https://docs.microsoft.com/en-us/dotnet/api/system.threading.thread.sleep?view=netframework-4.8
* https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep?redirectedfrom=MSDN - the underlying Windows system call that `Thread.Sleep` invokes

> This function causes a thread to relinquish the remainder of its time slice and become unrunnable for an interval based on the value of dwMilliseconds. The system clock "ticks" at a constant rate. If dwMilliseconds is less than the resolution of the system clock, the thread may sleep for less than the specified length of time. If dwMilliseconds is greater than one tick but less than two, the wait can be anywhere between one and two ticks, and so on. To increase the accuracy of the sleep interval, call the timeGetDevCaps function to determine the supported minimum timer resolution and the timeBeginPeriod function to set the timer resolution to its minimum.

We can't worry about being super-precise with the `Scheduler` - the underlying operating system doesn't. We need something that is good enough and that's what the hashed wheel timer mechanism does - predicts the number of revolutions around the wheel that are needed before the task is due based on a constant "tick" rate. What we can probably do, however, is make the way we wait for those ticks to elapse more efficient - I'm open to suggestions on that, but I'd stick with running things inside a dedicated thread and using threading primitives like `SpinWait` and `Sleep`

SpinWait might be advantageous to use in scenarios where we can predict that the next bucket will be coming up _very_ soon - i.e. when the scheduler is using either a very short interval for a recurring task or there's a large number of scheduled tasks queued. In scenarios where the workload is more infrequent, Sleeping is probably ok. I'd stick to working within the structure we have and focusing in making it so we waste less of the CPU's time when the scheduler is idle.

IgorFedchenko:
@Zetanova 
> You are running in the normal wrong assumptions

Well, not exactly... Of course 1ms is a huge amount of time for almost any modern CPU, and there is no way to ask CPU to wait (only to ask OS to release resources for a while). 

The initial idea was to make some useless work on current thread but make more accurate pauses between weel ticks. And see if this will improve overall performance, since scheduler is a critical place and maybe giving "some" more CPU would worth it.
But as tests showing, this is too high cost when we are running multiple `ActorSystem`'s in parallel. Several MNTR tests (each is starting several child processes with their own schedulers) load CPU too much, so we have to drop initial idea.

@Aaronontheweb So, indeed, does not seem like `async/await` will give us much improvement when using `TaskScheduler` over single thread - maybe `Thread.Sleep` will do the same job for us. 

I will revert back changes, and will add small optimization to use `Thread.SpinWait` when there are only few hundreds of ticks to wait for (something close to what @Zetanova said about `SpinWait` usage).

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
```diff
@@ -215,31 +235,58 @@ private long WaitForNextTick()
...
+                    if (ticksToSleep <= 0) // no need to sleep
                     {
                         if (currentTime == long.MinValue) // wrap-around
                             return -long.MaxValue;
+                        
                         return currentTime;
-
                     }
 
+                    var stopWatch = Stopwatch.StartNew();
```
Aaronontheweb: Instead of starting a new `Stopwatch` here, if you use `HighRestMonotonicClock.Ticks` that will give you the current `Stopwatch.Elapsed` value since Akka.NET started and you can acquire a second comparison value where you have `Stopwatch.Sleep` instead - just to save on allocations.

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
```diff
@@ -34,8 +38,45 @@ namespace Akka.Actor
...
+        /// <summary>
+        /// 0 - init, 1 - started, 2 - shutdown
+        /// </summary>
+        private volatile int _workerState = WORKER_STATE_INIT;
+
+        /// <summary>
+        /// API for internal usage
+        /// </summary>
+        [InternalApi]
+        public static readonly AtomicCounter TotalTicksRequiredToWaitStrict = new AtomicCounter(0);
```
Aaronontheweb: What are these being used for exactly? Testing?

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
IgorFedchenko: Ah, actually had to move this `stopwatch` to private field instantiated once - but missed this. And sure, your suggestion is even better.

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
IgorFedchenko: Indeed, some more comments required here... Yes, for testing - I will update summary now.

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
IgorFedchenko: Used the same thing for sleep time tracking when making `Thread.SpinWait`, instead of separate `Stopwatch` instance usage.

Aaronontheweb:
@IgorFedchenko looks like we need an API approval here.

IgorFedchenko:
Benchmark results (compare to [this](https://github.com/akkadotnet/akka.net/pull/4019#issuecomment-549269987) one) - I was just waiting for it to finish:

```
Intel Core i7-4770 CPU 3.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100
  [Host]     : .NET Core 2.1.11 (CoreCLR 4.6.27617.04, CoreFX 4.6.27617.02), X64 RyuJIT
  Job-ESBJQR : .NET Core 2.1.11 (CoreCLR 4.6.27617.04, CoreFX 4.6.27617.02), X64 RyuJIT

InvocationCount=1  IterationCount=100  LaunchCount=1  
RunStrategy=Monitoring  UnrollFactor=1  WarmupCount=1  

|                    Method | MessageCount | MessageLength | ClientsCount |        Mean |     Error |     StdDev |      Median |     Gen 0 |     Gen 1 | Gen 2 | Allocated |
|-------------------------- |------------- |-------------- |------------- |------------:|----------:|-----------:|------------:|----------:|----------:|------:|----------:|
| ClientServerCommunication |          100 |            10 |            1 |    57.73 ms |  2.044 ms |   6.026 ms |    57.76 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |            3 |    29.01 ms |  1.106 ms |   3.262 ms |    28.57 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |            5 |    23.37 ms |  0.742 ms |   2.188 ms |    22.69 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |            7 |    26.15 ms |  1.859 ms |   5.480 ms |    24.66 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |           10 |    24.30 ms |  0.561 ms |   1.653 ms |    23.67 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |           20 |    34.56 ms |  0.319 ms |   0.940 ms |    34.32 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |           30 |    48.56 ms |  0.599 ms |   1.767 ms |    48.45 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |            10 |           40 |    60.99 ms |  0.968 ms |   2.854 ms |    60.02 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |            1 |    43.49 ms |  0.578 ms |   1.706 ms |    43.08 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |            3 |    22.63 ms |  0.531 ms |   1.564 ms |    22.14 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |            5 |    21.61 ms |  0.462 ms |   1.363 ms |    21.17 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |            7 |    22.17 ms |  0.535 ms |   1.577 ms |    21.54 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |           10 |    23.79 ms |  0.394 ms |   1.161 ms |    23.54 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |           20 |    35.11 ms |  0.446 ms |   1.315 ms |    34.92 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |           30 |    47.21 ms |  0.674 ms |   1.989 ms |    46.43 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |          100 |           100 |           40 |    61.06 ms |  0.859 ms |   2.531 ms |    60.52 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |            1 |   403.88 ms |  5.701 ms |  16.808 ms |   399.22 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |            3 |   179.15 ms |  5.153 ms |  15.195 ms |   173.13 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |            5 |   146.48 ms |  3.080 ms |   9.081 ms |   141.66 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |            7 |   130.54 ms |  1.972 ms |   5.813 ms |   128.32 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |           10 |   127.79 ms |  2.292 ms |   6.759 ms |   125.80 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |           20 |   136.60 ms |  3.864 ms |  11.392 ms |   134.19 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |           30 |   136.68 ms |  2.551 ms |   7.521 ms |   134.19 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |            10 |           40 |   144.96 ms |  2.349 ms |   6.925 ms |   143.74 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |            1 |   402.41 ms |  3.585 ms |  10.570 ms |   399.46 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |            3 |   174.00 ms |  4.007 ms |  11.813 ms |   167.71 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |            5 |   142.37 ms |  2.376 ms |   7.006 ms |   139.87 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |            7 |   139.31 ms |  3.900 ms |  11.498 ms |   137.67 ms |         - |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |           10 |   129.31 ms |  2.730 ms |   8.050 ms |   126.67 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |           20 |   129.00 ms |  1.351 ms |   3.983 ms |   127.82 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |           30 |   136.91 ms |  2.293 ms |   6.760 ms |   134.40 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |         1000 |           100 |           40 |   150.63 ms |  3.591 ms |  10.588 ms |   147.31 ms | 1000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |            1 | 3,943.60 ms | 12.900 ms |  38.036 ms | 3,934.15 ms | 7000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |            3 | 1,648.77 ms | 20.125 ms |  59.340 ms | 1,627.00 ms | 7000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |            5 | 1,355.34 ms | 17.621 ms |  51.957 ms | 1,332.90 ms | 7000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |            7 | 1,244.47 ms | 20.884 ms |  61.578 ms | 1,218.32 ms | 7000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |           10 | 1,182.45 ms | 21.427 ms |  63.179 ms | 1,154.13 ms | 7000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |           20 | 1,069.44 ms | 12.701 ms |  37.450 ms | 1,062.56 ms | 7000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |           30 | 1,099.14 ms | 13.493 ms |  39.783 ms | 1,085.21 ms | 8000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |            10 |           40 | 1,091.91 ms | 15.335 ms |  45.217 ms | 1,078.84 ms | 8000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |            1 | 3,945.08 ms | 14.754 ms |  43.501 ms | 3,938.04 ms | 8000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |            3 | 1,658.57 ms | 23.353 ms |  68.858 ms | 1,625.38 ms | 8000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |            5 | 1,348.68 ms | 14.412 ms |  42.493 ms | 1,331.29 ms | 8000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |            7 | 1,238.80 ms | 21.984 ms |  64.820 ms | 1,212.24 ms | 8000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |           10 | 1,289.90 ms | 40.027 ms | 118.019 ms | 1,270.24 ms | 8000.0000 |         - |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |           20 | 1,179.72 ms | 29.509 ms |  87.009 ms | 1,164.57 ms | 8000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |           30 | 1,155.36 ms | 27.313 ms |  80.533 ms | 1,136.48 ms | 9000.0000 | 1000.0000 |     - |   1.58 KB |
| ClientServerCommunication |        10000 |           100 |           40 | 1,236.05 ms | 32.779 ms |  96.649 ms | 1,230.72 ms | 9000.0000 | 1000.0000 |     - |   1.58 KB |
``` 

Looks almost the same - sometimes little bit faster, sometimes slower.
May need to play with `TicksToUseThreadSpinWait` value in scheduler - right now it is set to `100` ticks, which means that when we have 101 tick to wait we will sleep for at least `10 000 * 10` ticks (`10ms`). Or even better - get the list of sleep times during some simple benchmark and try to set this ticks treashold to less "random" value.

IgorFedchenko:
So, about sleep times.

I started one sample from benchmark, which is sending 10_000 messages (10 bytes each) in request-reply manner from 3 concurrent clients (that is ~3333 messages per client) to single server via `Akka.IO.Tcp` (this is what my benchmark from previous comment is doing), and have done this for 100 iterations (pass 10k messages to this 3 clients 100 times). Each client is sending next message once server responded with reply for the previous message sent by this client.

There were `19284` sleeps. `18950` from them were in `[80_000, 90_000]` ticks range (`10k` is `1ms` as you probably know). `157` sleeps below 80K, and overall it looks similar to normalized distribution. 

Well, here is the file (initially created for me, without formatting, but will share it with you, guys). Sleep numbers are in ascending order.
[waitings.txt](https://github.com/akkadotnet/akka.net/files/3917653/waitings.txt)

So it is obvious that my `100` ticks threshold has nothing to do this this. And actually does not seem that we can successfully use `Thread.SpinWait` here, because delays by the most part are near to the wheel tick value of 100K ticks, and none of them are below 20K - which is pretty big value for `SpinWait` if I understand right.

So here are two questions:
1. Is it too low load for scheduler to make any conclusions? I mean, how can I (and is it possible) to make sleep times closer to 1K or less? If so, I could make some sample run with such scenario and see what threshold might be used for `Thread.SpinWait` in that case.

2. If performing all scheduled actions for 1-2 ms is find and something typical for scheduler, that we just can not optimize it with `Thread.SpinWait` and just have to wait for 15+ ms each time instead of 8-9ms requested.

Aaronontheweb:
@IgorFedchenko so, it sounds like this PR won't add any value to the existing scheduler performance then?

IgorFedchenko:
@Aaronontheweb If in most use cases scheduler does not have much work to do (as in my benchmark) - then yes, there is nothing to optimize. Well, I wish we could make scheduler sleep for more accurate interval, but looks like the only way to do this (and keep things simple enough) is using `Thread.SpinWait`, which requires too much CPU for the timings we have.

IgorFedchenko:
@Aaronontheweb Should we close it?

Zetanova:
I looked at the current implementation and it seams to be well done.
My test was always on point with the tick schedule less then one ms off (sometimes)

The only improvement what i can see is:
1) Maybe to use a readonly ValueType for the arrays in the buckets. This will speed up the iterations over them. But the effect will be only minimal and only noticeable for very large entry counts (>5k) 
2) Maybe to sleep longer then it is known that the next ticks will be empty.

PR #4037, opened 2019-11-14
V1.3.16 Release

Aaronontheweb:
Putting this into draft mode for now - need to see CI pass before we try releasing this against the v1.3.* branch.


Aaronontheweb:
Also, I still need to add release notes

PR #4038, opened 2019-11-14
Backport v1.3.16 into dev

Aaronontheweb:
Backporting changes from the `master` branch back into `dev`


Aaronontheweb:
Rebased everything - will see how CI goes, but otherwise should be good to go.

**Important: need to use a normal merge commit here - squashing will force us to go through and rebase again in the future**

PR #4042, opened 2019-11-16
Stop DeadLetterListener on terminate if LogDeadLettersDuringShutdown is disabled

ismaelhamed:
Looks like we never had a chance of disabling `log-dead-letters-during-shutdown`

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
@@ -106,6 +106,26 @@ namespace Akka.Tests.Actor
         }
 
         [Fact]
+        public void Log_dead_letters()
+        {
+            var sys = ActorSystem.Create("LogDeadLetters", ConfigurationFactory.ParseString("akka.loglevel=INFO")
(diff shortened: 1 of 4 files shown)


src/core/Akka.Tests/Actor/ActorSystemSpec.cs
```diff
@@ -105,6 +105,26 @@ public void Allow_valid_names()
...
+                var eventFilter = new EventFilterFactory(new TestKit.Xunit2.TestKit(sys));
+                eventFilter.Info(contains: "not delivered").Expect(1, () =>
+                {
+                    a.Tell("run");
+                    a.Tell("boom");
+                });
+            }
+            finally { Shutdown(sys); }
+        }
+
```
ismaelhamed: Just a missing test

Aaronontheweb:
Failure on two specs here related to these changes:

`Akka.Persistence.Tests.Serialization.MessageSerializerRemotingSpec.MessageSerializer_should_custom_serialize_Persistent_messages_during_remoting`

Might be an issue with the way that test is written.

Also, needs an API approval.

PR #4047, opened 2019-11-18
Context.WatchWith memory leak fix

### LINKED ISSUE #4026: Memory leak when using Context.WatchWith
Hi, I encountered a memory leak with ``Context.WatchWith``. Here is a repro code:
``` C#
using System;
using System.Collections.Generic;
using Akka.Actor;

namespace MemoryLeakRepro
{
    public class LoadHandler : ReceiveActor
    {
        private readonly List<IActorRef> _subjects;
        private readonly ICancelable _cancel;

        public LoadHandler()
        {
            _subjects = new List<IActorRef>();
            _cancel = Context.System.Scheduler.ScheduleTellRepeatedlyCancelable(
                initialDelay: TimeSpan.FromSeconds(1),
                interval: TimeSpan.FromSeconds(1),
                receiver: Self,
                message: Iteration.Instance,
                sender: ActorRefs.NoSender);

            Receive<Iteration>(
                _ =>
                {
                    // stop actors created on previous iteration
                    _subjects.ForEach(Context.Stop);
                    _subjects.Clear();

                    // create a set of actors and start watching them
                    for (var i = 0; i < 10_000; i++)
                    {
                        var subject = Context.ActorOf(Props.Create<Subject>());
                        _subjects.Add(subject);
                        Context.WatchWith(subject, new Stopped(subject));
                    }
                });

            Receive<Stopped>(_ => { });
        }

        private class Iteration
        {
            public static readonly Iteration Instance = new Iteration();
            private Iteration() { }
        }

        private class Stopped
        {
            public IActorRef ActorRef { get; }

            public Stopped(IActorRef actorRef)
            {
                ActorRef = actorRef;
            }
        }

        private class Subject : ReceiveActor
        {
            // simulate internal state
            private byte[] _state = new byte[1000];
        }

        protected override void PostStop() => _cancel.Cancel();
    }

    public static class Program
    {
        public static void Main(string[] args)
        {
            using (var actorSystem = ActorSystem.Create("repro"))
            {
                actorSystem.ActorOf(Props.Create<LoadHandler>());
                Console.ReadKey();
            }
        }
    }
}
```
```XML
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.1</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Akka" Version="1.3.15" />
  </ItemGroup>

</Project>
```

In my understanding this code should not leak: ``Subject`` actor references must be released as soon as ``LoadHandler`` processed ``Stopped`` message. However, looking at profiler - it doesn't look like this happens:
image
image

I also found a very similar issue in Akka repository which is already fixed, but I don't think this was ported to Akka.NET: https://github.com/akka/akka/issues/26625

IgorFedchenko:
Close #4026 

In draft until fix will be implemented. Only reproducing project and unit test for now.


IgorFedchenko:
.NET implementation turned out to be organized in somewhat different way then in scala, so my attempt to directly translate their PR to our code did not fix the issue (maybe I missed something).
So will try to understand issue more deeply and commit working fix.

IgorFedchenko:
All right, actually, scala's PR was just fine.

For those, who loves pretty puctures:
Before fix:
image
After fix:
image

Basically, the issue was that we were removing custom termination message from internal state only after getting `Terminated` message - but we were never sending `Terminated`, we were sending that custom message. So stored custom message was never removed from internal set, and was taking memory.

Aaronontheweb:
@IgorFedchenko need to approve the public API changes: https://getakka.net/community/public-api-changes.html

IgorFedchenko:
> @IgorFedchenko need to approve the public API changes: https://getakka.net/community/public-api-changes.html

Done

gamith55:
> All right, actually, scala's PR was just fine.
> 
> For those, who loves pretty puctures:
> Before fix:
> image
> After fix:
> image
> 
> Basically, the issue was that we were removing custom termination message from internal state only after getting `Terminated` message - but we were never sending `Terminated`, we were sending that custom message. So stored custom message was never removed from internal set, and was taking memory.

hi

gamith55:
hi guys,
Could you please help me to overcome memory leak issue in my Akka.net service.
 
I have uprated my AKKA.net version as below,
  
 FROM "Akka" Version="1.3.11"  TO  "Akka" Version="1.4.13" 
AND
    FROM "Akka.Streams" Version="1.3.11"   TO "Akka.Streams" Version="1.4.13"

but still my memory consuming looks like here.
image

IgorFedchenko:
@gamith55 So in your case, first or all, we need to find out is this memory leak happening somewhere on Akka.NET side, or probably something in your service's code is not fully cleaned up.

What I would suggest to do:
1. Try to isolate the problem. Remove pieces of code, one after another, until you will see that memory is not leaking. 
2. Once you found the leaking code, try to make sure that there is noting outside of Akka.NET that may grow in memory. I.e. some user actor having dictionary property that may be growing and never reduced in size, or something like that. Or - maybe this code does not use Akka.NET at all.
3. If it is clear that nothing in your code could give memory leaks - create new issue here, with detailed description of what your code is going, so that we could reproduce the problem. Sharing the code that reproduces the issue would be just awesome.

So if there is something that Akka.NET team can help you with - feel free to create new issue and describe your case there.

Aaronontheweb:
Do you have any profiler data you can share? That would be very helpful

PR #4053, opened 2019-11-20
Added standard logger print hook.

djm132:


Aaronontheweb:
Yeah, this is a bit of a hack. I don't want to add this to our API. Let's take a look at your issue and see if there's a better way of solving this.

PR #4056, opened 2019-11-21
MNTR NRE fix for #4035

### LINKED ISSUE #4035: MNTR: TRX reporter fails with NRE
Version: 1.4.0-beta3

```
Cause: System.NullReferenceException: Object reference not set to an instance of an object.
   at Akka.MultiNodeTestRunner.Shared.AzureDevOps.TrxSinkActor.ReportFailure(SpecSession session, Dictionary`2 nodeResults)
   at Akka.MultiNodeTestRunner.Shared.AzureDevOps.TrxSinkActor.ReportSpec(SpecSession session, TestRun testRun, String computerName)
   at Akka.MultiNodeTestRunner.Shared.AzureDevOps.TrxSinkActor.HandleEndSpec(EndSpec endSpec)
   at lambda_method(Closure , Object , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 )
   at Akka.Tools.MatchHandler.PartialHandlerArgumentsCapture`11.Handle(T value)
```

Failed during https://dev.azure.com/dotnet/Akka.NET/_build/results?buildId=24678&view=logs&j=dc4e9172-2a93-56ab-8572-2079bd522eec&t=7d446c1b-f726-5de4-2784-d7d7d79a78d2

IgorFedchenko:
Close #4035 

Here is the code that is failing with NRE:
```csharp
private static void ReportFailure(SpecSession session, Dictionary<int, UnitTestResult> nodeResults)
{
	foreach (var (time, message) in session.Fails)
	{
		var result = nodeResults[message.NodeIndex];
		result.Outcome = TestOutcome.Failed;
		result.EndTime = time;
		result.Output = new Output();

		result.Output.StdErr.Add(message.Message);
		result.Output.DebugTrace.Add(message.Message);
		result.Output.ErrorInfo.Message = message.Message;
	}
}
```

Since we do not have exact like, there are few candidates to throw NRE (expressions that could be `null` in this case): 
`session`
`noreResults
`message` from `var (time, message) in session.Fails`
`nodeResults[message.NodeIndex]`
`result.Output.StdErr` or `result.Output.DebugTrace` or `result.Output.ErrorInfo`

Luckily, the code is pretty simple here and all of this is initialized in this class or somewhere near in the call tree. So after some careful check, I have verified that nobody could be `null` except  `result.Output.ErrorInfo` (either because it is just impossible, or it should throw somewhere before this `ReportFailure` invocation). And `result.Output.ErrorInfo` is just not initialized, so the fix is simple.

### DIFF
--- a/src/core/Akka.MultiNodeTestRunner.Shared/TrxReporter/TrxSinkActor.cs
+++ b/src/core/Akka.MultiNodeTestRunner.Shared/TrxReporter/TrxSinkActor.cs
@@ -133,7 +133,7 @@ namespace Akka.MultiNodeTestRunner.Shared.AzureDevOps
 
                 result.Output.StdErr.Add(message.Message);
                 result.Output.DebugTrace.Add(message.Message);
-                result.Output.ErrorInfo.Message = message.Message;
+                result.Output.ErrorInfo = new ErrorInfo() { Message = message.Message };
             }
         }
 


IgorFedchenko:
P.S. Did not want to initialize it in `Output` class directly, because in that case after output serialization we would see non-empty `ErrorInfo` value with empty properties, which does not look nice. Let it be `null` if no errors occurred.

PR #4058, opened 2019-11-22
Fix actor terminated exception from TcpListener actor

IgorFedchenko:
When running CI in #4022, we were getting errors during MNTR run:

```
[ERROR][22.11.2019 14:40:28][Thread 0041][akka://TestRunnerLogging/system/IO-TCP/$a] Monitored actor [[akka://TestRunnerLogging/user/TcpLogger#73156647]] terminated
Cause: Akka.Actor.DeathPactException: Monitored actor [[akka://TestRunnerLogging/user/TcpLogger#73156647]] terminated
   в Akka.Actor.ActorBase.Unhandled(Object message)
   в Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message)
   в Akka.Actor.ActorCell.ReceiveMessage(Object message)
   в Akka.Actor.ActorCell.ReceivedTerminated(Terminated t)
   в Akka.Actor.ActorCell.AutoReceiveMessage(Envelope envelope)
   в Akka.Actor.ActorCell.Invoke(Envelope envelope)
```

Turned out that it is due the fact that `TcpLoggingServer` actor is binding via Akka.IO to the socket, but does not unbind before shutting down the whole actor system. 
In theory, if internal `Akka.IO.TcpListener` actor will shutdown first, we will not get this error, but I could stably reproduce this issue on my local machine.
So basically I am just asking `TcpLoggingServer` to unbind from socket (an therefore `TcpListener` will shutdown and stop watching logger), and after that shutting down the system.

Even if MNTR is failing for some another reason, this PR will at least fix warning about `Tcp.Bound` message was not delivered (now it is handling by `TcpLoggingServer` to ask for `Unbind` at the end).

### DIFF
--- a/src/benchmark/Akka.Benchmarks/IO/TcpOperationsBenchmarks.cs
+++ b/src/benchmark/Akka.Benchmarks/IO/TcpOperationsBenchmarks.cs
@@ -44,7 +44,7 @@ namespace Akka.Benchmarks
             _system = ActorSystem.Create("system");
             _message = new byte[MessageLength];
 
-            var port = new Random().Next(18000, 19000);
+            var port = GetFreeTcpPort();
             _server = _system.ActorOf(Props.Create(() => new EchoServer(port)));
             _clientCoordinator = _system.ActorOf(Props.Create(() => new ClientCoordinator("127.0.0.1", port, ClientsCount)));
         }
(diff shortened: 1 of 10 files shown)


Aaronontheweb:
> Even if MNTR is failing for some another reason, this PR will at least fix warning about Tcp.Bound message was not delivered (now it is handling by TcpLoggingServer to ask for Unbind at the end).

I agree, this will clean up a very annoying error message that shows up routinely in the MNTR logs.

IgorFedchenko:
@Aaronontheweb Fixed tests skipping - now should be good to merge.

PR #4062, opened 2019-11-23
Logging and buffering Tcp.Write before Tcp.Register received

### LINKED ISSUE #3775: Akka.IO: Tcp.Write messages sent to TcpConnection prior to Tcp.Register are silently dropped
Version: 1.3.13 and earlier

Found this issue while debugging an issue in another project earlier today:

https://github.com/akkadotnet/akka.net/blob/1b35f594411a7450f8121520858a77bb33c8c6e8/src/core/Akka/IO/TcpConnection.cs#L137-L178

If a `Tcp.Write` message is sent to the `TcpConnection` actor prior to receiving a `Tcp.Register` message, those `Tcp.Write` messages are dropped - in our system this was a minor race condition that only occurred during unit testing (registration and writes were sent by two separate actors) but this could absolutely happen in a production environment too if someone used a similar setup to us.

In order to make debugging easier for those users, I'd strongly recommend that we add an explicit warning message if a `Tcp.Write` is sent prior to `Tcp.Register` - that way I don't have to go looking inside the `akka.actor.debug.unhandled = on` box to find the issue. If we wanted to go a step further, buffering messages prior to receiving a `Tcp.Register` may not be a bad idea but I think logging is a good "first step" in making these types of issues easier to discover and debug.

IgorFedchenko:
Close #3775

Since internal writes buffering is something built-in now, it is pretty simple to handle this inside of `TcpConnection` actor. So both logging and buffering are implemented.

Also, internal buffer overflow should be handled.

### DIFF
--- a/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
+++ b/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
@@ -256,6 +256,20 @@ namespace Akka.Tests.IO
                 serverMsgs.Should().HaveCount(1).And.Subject.Should().Contain(m => m.Data.Equals(msg));
             });
         }
+        
+        [Fact]
+        public void Write_before_Register_should_Be_dropped_if_buffer_is_full()
+        {
+            new TestSetup(this).Run(x =>
+            {
+                var actors = x.EstablishNewClientConnection(registerClientHandler: false);
+
+                var overflowData = ByteString.FromBytes(new byte[InternalConnectionActorMaxQueueSize + 1]);
+
+                actors.ClientHandler.Send(actors.ClientConnection, Tcp.Write.Create(overflowData));
+                actors.ClientHandler.ExpectMsg<Tcp.CommandFailed>(TimeSpan.FromSeconds(10));
+            });
+        }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -232,6 +233,55 @@ public void BugFix_3021_Tcp_Should_not_drop_large_messages()
             });
         }
 
+        [Fact]
+        public void Write_before_Register_should_not_be_silently_dropped()
```
Aaronontheweb: :+1:

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -170,6 +170,15 @@ private Receive WaitingForRegistration(IActorRef commander)
 
                         Context.SetReceiveTimeout(null);
                         Context.Become(Connected(registerInfo));
+
+                        // If there is something buffered before we got Register message - put it all to the socket
```
Aaronontheweb: Good

IgorFedchenko:
@Aaronontheweb Does not look like any of failed tests are related to this update (looks like this is racing again).

Issue #3765, opened 2019-04-25
Issue with StreamsRef over remote system


pm-mw:
I have build a small test case for the beta StreamsRef feature (1.4 branch), very similar to the unit test, which is used in the docs for StreamsRef but using two (remote) Akka systems.

The relevant code is here https://gist.github.com/pm-mw/6df9e955412ae5c1087ad10e741d166e
When the source actor receives the  Receive<StartListening> message and pipes the stream to the sink (receiving actor) I get the exception

```
[ERROR][25.04.2019 06:45:15][Thread 0011][remoting] Die angegebene Methode wird nicht unterstützt.
Cause: System.NotSupportedException: Die angegebene Methode wird nicht unterstützt.
   bei System.Collections.Immutable.ImmutableArray`1.System.Collections.Generic.ICollection<T>.Add(T item)
   bei Newtonsoft.Json.Utilities.CollectionWrapper`1.Add(T item)
   bei Newtonsoft.Json.Utilities.CollectionWrapper`1.System.Collections.IList.Add(Object value)
   bei Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObjectUsingCreatorWithParameters(JsonReader reader, JsonObjectContract contract, JsonProperty containerProperty, ObjectConstructor`1 creator, String id)
   bei Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateNewObject(JsonReader reader, JsonObjectContract objectContract, JsonProperty containerMember, JsonProperty containerProperty, String id, Boolean& createdFromNonDefaultCreator)
   bei Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   bei Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateValueInternal(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   bei Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ResolvePropertyAndCreatorValues(JsonObjectContract contract, JsonProperty containerProperty, JsonReader reader, Type objectType)
[...]

PR #4063, opened 2019-11-23
StreamRef can be sent over wire in POCO envelope

valdisz:
Fixes issue #3765 
Surrogate was missing for StreamRefImpl


valdisz:
@Aaronontheweb You can look into this PR.
Looks like there are problems with the test or unrelated code.

valdisz:
@Aaronontheweb to resolve the issue I need to add `ISurrogated` to `public interface ISourceRef<TOut>` which breaks the test `CoreAPISpec.ApproveStreams`. Any suggestions for how to move on in this situation?

Aaronontheweb:
@valdisz https://getakka.net/community/public-api-changes.html - you can follow that guide there.

Issue #4059, opened 2019-11-22
Improve MNTR output readability


IgorFedchenko:
There are actually two points here: 
- make MNTR output simpler to use by other programmers, who are using nuget package
- make it simpler to review failure when debugging Azure Pipelines output for PR validation in Akka.NET

Right now, there is a nice [tool](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.MultiNodeTestRunner.Shared/Persistence/VisualizerPersistentTestRunStore.cs) that allows to review some basic information about test results on html page:
![image](https://user-images.githubusercontent.com/9683660/69457623-88c6ed00-0d7e-11ea-8b09-86b95d5aa62a.png)

This html file is generated near the folders with logs of each test/node.
There is no much details here, and when some of the tests are failing, you have to go into folder, open several text log files with parallel timelines and spend some time on it.
What would be nice is if we could get some aggregated logs with single timeline somewhere on the page above, so that we could easily review failed test logs here, maybe with some visualization of any critical places (i.e. logged exception might be displayed on the graph). Right now it seems like failed test is not showing it's timeline at all.

The second part of the improvement could be to add MNTR nodes logs into Azure Pipelines job output. So that even before downloading any of build artifacts you had a chance to make quick search there for some info. And, of course, would be nice to be able to see what assertion failed during multi-node test - right now we can only see the fact that some particular node failed:
![image](https://user-images.githubusercontent.com/9683660/69458186-ddb73300-0d7f-11ea-9281-c07610ed9e5e.png)

Maybe some other improvements could be done here?

Aaronontheweb:
I'm happy with our progress here. I'll mark this as closed.

PR #4067, opened 2019-11-25
MNTR logs representation improvements

IgorFedchenko:
Working on #4059 here.

So far, added aggregation of all log messages into single `aggregated.txt` file, ordered by timestamp. So, now it should be easier to review what is happening across the nodes.

I guess we will also need to output this aggregated timeline to Azure Pipelines output. This should be as simple as writing all of this logs into a console.


Aaronontheweb:
Thanks @IgorFedchenko - I'll check some of the MNTR output with this change.

Aaronontheweb:
[aggregated.txt](https://github.com/akkadotnet/akka.net/files/3888061/aggregated.txt) - this is already a major usability improvement. Much easier to see what's happening over the course of the test this way.

PR #4069, opened 2019-11-26
Fix MNTR logs datetime parsing

IgorFedchenko:
Related to #4059

Looks like in my #4067 PR I missed the fact that logger will format log record date and time depending on current locale. Current code was trying to parse single format (the once that is used on my local machine) - so need to fix this to support current locate. This will also fix parsing logs from CI run.

P.S. By the way, actually even if it will fail to parse the format, current implementation takes date and time of receiving the log for sorting. So, it will still give some more or less valid record order (but obviously some surprises may come up in that case).


Aaronontheweb:
Weird, wonder why this didn't run....

PR #4070, opened 2019-11-26
Possible correction in one of the intro documents. Changed to 'user guardian'.

chrillejb:
I was reading the documentation and stumbled upon the description of the user guardian which caught my attention since it seemed a bit too generic.

By looking at the description of the other guardians I came to the conclusion that the description probably meant to say: _user guardian_ instead of just: _guardian_.

What do you think? :smiley:


Aaronontheweb:
Yep, I think this is a good clarifying change.

PR #4071, opened 2019-11-27
TypeExtensions.TypeQualifiedName like in Hyperion

### LINKED ISSUE #3767: TypeExtensions.TypeQualifiedName includes strict assembly references for generic arguments
- Akka.NET 1.3.12
- Windows 10 64-bit (.NET Core 2.2).

The following code:
```
using Akka.Util;
using System;

namespace QuickTest
{
    public sealed class OtherClassA { }

    public sealed class OtherClassB { }

    public sealed class OtherClassC { }

    public sealed class ParentClass<T1, T2, T3>
    {
        public sealed class ChildClass
        {

        }
    }

    public sealed class Program
    {
        public static void Main(string[] args)
        {
            Console.WriteLine(TypeExtensions.TypeQualifiedName(typeof(ParentClass<OtherClassA, OtherClassB, OtherClassC>.ChildClass)));
            Console.ReadLine();
        }
    }
}
```

Produces the following result:

```
QuickTest.ParentClass`3+ChildClass[[QuickTest.OtherClassA, QuickTest, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null],[QuickTest.OtherClassB, QuickTest, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null],[QuickTest.OtherClassC, QuickTest, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]], QuickTest
```

This, as seen, includes fairly strict assembly references for generic arguments (`, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null`), while it does not for non-generic types (and this appears to be very much intentional given the [current source](https://github.com/akkadotnet/akka.net/blob/5ca783fef57a23f9e0ee3d4c826c77e8b9155d6b/src/core/Akka/Util/TypeExtensions.cs#L67)).

This method is used when describing a type for the serialization manifest, and I do not feel it should ever be making a strict (versioned) assembly reference.

valdisz:
Ported type name extraction from Hyperion into Akka's `TypeExtensions.TypeQualifiedName`.
This PR resolves #3767

### DIFF
--- a/src/core/Akka.Tests/Serialization/SerializationSpec.cs
+++ b/src/core/Akka.Tests/Serialization/SerializationSpec.cs
@@ -6,6 +6,10 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Concurrent;
+using System.IO;
+using System.Reflection;
+using System.Text.RegularExpressions;
 using Akka.Actor;
 using Akka.Configuration;
 using Akka.Dispatch.SysMsg;
@@ -13,6 +17,9 @@ using Akka.Routing;
 using Akka.Serialization;
 using Akka.TestKit;
 using Akka.TestKit.TestActors;
+using Akka.Util;
+using Akka.Util.Reflection;
+using FluentAssertions;
 using Xunit;
 
 namespace Akka.Tests.Serialization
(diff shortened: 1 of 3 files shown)


valdisz:
Working on review comments right now

valdisz:
@Aaronontheweb added 2 tests:
1. checks that assembly qualified type name does not have unwanted components like `Version`
2. checks that type manifest from old and new code is resolved to the identical type object

2nd point is to check serialization - there are many serializers, some of them are not using that extension method at all and are not affected. IMHO for me this looks better than storing old and new serialized messages somewhere and comparing them.

But, I can still make such a test if it is needed.

valdisz:
Strange that tests are failing, locally they are passing...

src/core/Akka.Tests/Serialization/SerializationSpec.cs
```diff
@@ -552,6 +559,29 @@ public void Can_apply_a_config_based_serializer_by_the_binding()
...
+            string coreAssemblyName = typeof(object).GetTypeInfo().Assembly.GetName().Name;
+            var assemblyName = type.GetTypeInfo().Assembly.GetName().Name;
+            var shortened = assemblyName.Equals(coreAssemblyName)
+                ? type.GetTypeInfo().FullName
+                : $"{type.GetTypeInfo().FullName}, {assemblyName}";
+            return shortened;
+        }
+
+        [Fact]
+        public void Legacy_and_shortened_types_names_are_equivalent()
```
Aaronontheweb: Good

src/core/Akka.Tests/Util/TypeExtensionsTests.cs
```diff
@@ -33,6 +35,42 @@ public void Test_nongeneric_implements()
...
+                yield return new object[] { typeof(ParentClass<OtherClassA, OtherClassB, OtherClassC>.ChildClass) };
+                yield return new object[] { typeof(string) };
+                yield return new object[] { typeof(object) };
+                yield return new object[] { typeof(TypeExtensionsTests) };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(TargetTypes))]
+        public void Type_qualified_name_includes_only_typename_and_assemblyname(Type targetType)
```
Aaronontheweb: Good

Aaronontheweb:
@valdisz the specs were racy tests from our suite, a major nuisance that we're working on fixing through a number of changes - but your code looks good and I've approved it. Waiting on a re-run of the test suite.

PR #4073, opened 2019-11-30
Fix/inbox actor reference

nagytech:
Rebasing nbparrell/akka.net:dev and addressing the outstanding comments in PR #3412


nagytech:
In my rebase, I've dropped the StringLike changes as noted here: https://github.com/akkadotnet/akka.net/pull/3412#discussion_r183769797

Additionally, the changes made to StringLike.cs appear to be superseded in the latest on dev: https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Util/StringLike.cs

nagytech:
I missed the fact that this isn't building.  Will look and fix it.

Aaronontheweb:
@nagytech nah, I think you're in good shape - the build failure is an Akka.Streams test.

@IgorFedchenko is working on some changes that we think will reduce the frequency of racy test failures in our suite, as that's the number one source of frustration for regular Akka.NET contributors.  But no worries, this failure is unrelated to your changes and I'll re-run it to verify.

PR #4079, opened 2019-12-03
SqliteSnapshotStore is not getting table name from config

maratoss:
Table name was hardcoded. Seems that is breaking changes :(

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sqlite/Snapshot/ConflictNameResolver.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sqlite/Snapshot/ConflictNameResolver.cs
@@ -1,0 +1,53 @@
+// //-----------------------------------------------------------------------
+// // <copyright file="ConflictNameResolver.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2019 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2019 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using System;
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
Should close #4080

Aaronontheweb:
Tests look good - will review soon

src/contrib/persistence/Akka.Persistence.Sqlite/Snapshot/ConflictNameResolver.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// </summary>
+    internal class ConflictNameResolver
+    {
+        /// <summary>
+        /// Tries to resolve the conflict of snapshot table names
+        /// </summary>
+        /// <param name="connection">sqlite connection</param>
+        /// <param name="configuration">sqlite query configuration</param>
+        /// <returns>Returns true if has the conflict</returns>
+        public async Task<bool> Resolve(SqliteConnection connection, SqliteQueryConfiguration configuration)
```
Aaronontheweb: ... Why is this class necessary? I don't understand the need for the rest of the changes in this PR. I thought it was just a HOCON fallback?

maratoss: Did not get the point with HOCON fallback :)
The idea was to do not get db corrupted for people who upgrade akka library.
This fix gets akka still use old hardcoded table 'snapshot' despite of 'table-name' setting value.

Aaronontheweb: I'll take another look at it.

PR #4084, opened 2019-12-05
Remove string interpolation from cluster logs

ismaelhamed:
Usage of string interpolation prevents some loggers (i.e., Serilog) from applying the correct format to the output template.


Aaronontheweb:
@ismaelhamed LGTM - thanks!

PR #4086, opened 2019-12-09
Fix PR #3132 for pubsub mediator

nagytech:
Fixing up the content in https://github.com/akkadotnet/akka.net/pull/3132 relating to issue https://github.com/akkadotnet/akka.net/issues/3129.


Aaronontheweb:
Looks like we had a racy-ish failure the first time this ran on MNTR for Windows... Re-running the specs now to see if this occurs again.

Aaronontheweb:
@nagytech `DistributedPubSubMediatorSpec` has failed a couple of times now - could you take a look at the detailed logs and see what's going on with it? I can't tell at a glance if these are new issues with it or not, but I don't remember it having this high of a flip rate before.

nagytech:
Thanks for the comments, I've been looking more into the base of the problem and it looks like something we can solve in a future major release.  

As for this PR, I think that mediator registration in the constructor is likely to be the cause of our intermittent failures.  As we're not seeing this issue locally, I'm going to run the PR Validations and see what happens.

Aaronontheweb:
Re-running this with the latest MNTR output + DevOps integration.

Aaronontheweb:
Going to see if #4204 helped here or not.

Aaronontheweb:
`DistributedPubSubMediatorSpec` is still failing frequently here - I'll need to take a look at the test changes and logs to understand why.

Aaronontheweb:
closed via #4207

PR #4090, opened 2019-12-12
Convert to ImmutableHashSet for consistency in comparison

nagytech:
This test was giving me a fairly high flip rate locally - and the fix is quite straight forward.

Previous code was equating an IEnumerable<T> to an ImmutableHashSet<T> which yields inconsistent results due to the way a ImmutableHashSet<T> is structured.  For consistency, we now convert to a ImmutableHashSet<T> before the comparison.


nagytech:
Contributes to #3786

PR #4092, opened 2019-12-12
Add dynamic .NET Framework linking for Mono build

nagytech:


Aaronontheweb:
Might need to adjust our `build-system/*.yaml` files in order to get full Mono builds to run as part of this.

PR #4097, opened 2019-12-15
Increase default number of nodes for cluster failure detection

ismaelhamed:
> Default number of nodes that each node is observing for failure detection has increased from 5 to 9. The reason is to have better coverage and unreachability information for downing decisions.

Port: [#27173](https://github.com/akka/akka/pull/27173)


Aaronontheweb:
Azure DevOps having issues again... not marking the final step as complete... Seems like they break something every other week

PR #4118, opened 2019-12-21
detect and run F# projects during incremental builds

### LINKED ISSUE #4117: Build System: need Incremental testing to run F# tests when only F# libraries are affected
Version: current `dev` branch (1.4.0-beta3)

A PR like https://github.com/akkadotnet/akka.net/pull/4046 currently doesn't generate any test activity. This is a bug - we need to run the F# projects once a change comes in. Currently none of the tests are run.

Aaronontheweb:
close #4117

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -224,8 +224,8 @@ module internal ResultHandling =
 Target "RunTests" (fun _ ->    
     let projects = 
         let rawProjects = match (isWindows) with 
-                            | true -> !! "./src/**/*.Tests.csproj"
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
Looks like we have the tests passing on the full .NET framework, but not running at all on .NET Core. That must be a result of the way the test projects themselves are coded.

Aaronontheweb:
Above issue should be resolved via https://github.com/akkadotnet/akka.net/pull/4119

PR #4119, opened 2019-12-21
modify F# tests to also target .NET Core

Aaronontheweb:


Aaronontheweb:
Looks like issues with `App.config` loading in .NET Core... Going to need to modify the Akka.FSharp.Tests project some.

Issue #2432, opened 2016-12-31
Attempted to divide by zero


diegolinan:
So, I'm trying to use Akka on Xamarin on my macbook. I am using the .Net framework 4.6.1 and I'm getting this nice error when trying to create a new actor system:

```
[ERROR][12/31/2016 9:12:32 PM][Thread 0005][akka://UserProfileActorSystem/system/log1-DefaultLogger] Attempted to divide by zero.
Cause: System.DivideByZeroException: Attempted to divide by zero.
  at System.TermInfoDriver.ChangeColor (System.String format, System.ConsoleColor color) [0x00016] in /private/tmp/source-mono-4.6.0/bockbuild-mono-4.6.0-branch/profiles/mono-mac-xamarin/build-root/mono-x86/mcs/class/corlib/System/TermInfoDriver.cs:398 
  at System.TermInfoDriver.set_ForegroundColor (System.ConsoleColor value) [0x00011] in /private/tmp/source-mono-4.6.0/bockbuild-mono-4.6.0-branch/profiles/mono-mac-xamarin/build-root/mono-x86/mcs/class/corlib/System/TermInfoDriver.cs:432 
  at System.ConsoleDriver.set_ForegroundColor (System.ConsoleColor value) [0x0001f] in /private/tmp/source-mono-4.6.0/bockbuild-mono-4.6.0-branch/profiles/mono-mac-xamarin/build-root/mono-x86/mcs/class/corlib/System/ConsoleDriver.cs:134 
  at System.Console.set_ForegroundColor (System.ConsoleColor value) [0x00000] in /private/tmp/source-mono-4.6.0/bockbuild-mono-4.6.0-branch/profiles/mono-mac-xamarin/build-root/mono-x86/mcs/class/corlib/System/Console.cs:587 
  at Akka.Util.StandardOutWriter.WriteToConsole (System.String message, System.Nullable`1[T] foregroundColor, System.Nullable`1[T] backgroundColor, System.Boolean line) [0x00034] in <9e39de83b9544f988b2007a135264702>:0 
  at Akka.Util.StandardOutWriter.WriteLine (System.String message, System.Nullable`1[T] foregroundColor, System.Nullable`1[T] backgroundColor) [0x00000] in <9e39de83b9544f988b2007a135264702>:0 
  at Akka.Event.StandardOutLogger.PrintLogEvent (Akka.Event.LogEvent logEvent) [0x0006a] in <9e39de83b9544f988b2007a135264702>:0 
[...]

PR #4133, opened 2019-12-31
Add initializer for console properties

nagytech:
So long as we are looking to support Mono, this patch will be necessary to help mitigate race conditions within Mono's TermInfoDriver.  I've already got an active PR to fix Mono, but I'm not sure if it's going to be back-merged into other streams.

Please speak up if there's concerns about the way this hack works.  I don't like it, but can't quite think of any less-worse way to do it.


nagytech:
Fixes #2432

PR #4134, opened 2020-01-03
Doc improvement. add Partition in Fan-out stage

chnlkw:
Copy the description of Partition from https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Streams/Dsl/Graph.cs#L1057


Aaronontheweb:
Thanks @chnlkw !

PR #4135, opened 2020-01-04
Refactor TaskEx API to deprecate pre-4.6.1 methods

Aaronontheweb:
Still need to:

1. [x] Add API approvals
2. [x] Possibly mark some of these APIs are obsolete and / or remove them, since they're no longer needed.

### DIFF
--- a/src/core/Akka/Util/Internal/TaskEx.cs
+++ b/src/core/Akka/Util/Internal/TaskEx.cs
@@ -18,9 +18,6 @@ namespace Akka.Util.Internal
     /// </summary>
     internal static class TaskEx
     {
-        private const int RunContinuationsAsynchronously = 64;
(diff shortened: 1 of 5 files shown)


src/core/Akka/Util/Internal/TaskEx.cs
```diff
@@ -31,14 +28,7 @@ internal static class TaskEx
...
         {
-            if (IsRunContinuationsAsynchronouslyAvailable)
-            {
-                return new TaskCompletionSource<T>((TaskCreationOptions)RunContinuationsAsynchronously);
-            }
-            else
-            {
-                return new TaskCompletionSource<T>();
-            }
+            return new TaskCompletionSource<T>(TaskCreationOptions.RunContinuationsAsynchronously);
```
Aaronontheweb: All TCS instances can now run with `TaskCreationOptions.RunContinuationsAsynchronously` by default in .NET Standard 2.0+

Aaronontheweb:
Decided not to bother deprecating the APIs - they're not doing any harm as-is and actually help standardize the way we create `TaskCompletionSource` objects.

PR #4140, opened 2020-01-11
Remove/disable pooling by default when working with Tcp SAEA

### LINKED ISSUE #4127: Get rid of SAEA pools in Akka.IO.Tcp
Since we are trying to share pool of `SocketAsyncEventArgs` instances between Akka.IO connections, we may get issues like #3572, #3921 and others.

So as @Aaronontheweb suggested let's simplify get rid of this code and have two dedicated SAEA instances per connection.

IgorFedchenko:
Close #4127

To things changed here:

1. There is no pooling for `SocketAsyncEventArgs` used at all (in `Tcp`-related classes, not related to `Udp`). Each time we are creating incoming/outcoming connection (child of `TcpConnection` actor), we are creating two new SAEA instances dedicated to this actor. Once connection is closed, GC will collect them.

2. SAEA requires bytes buffer to be provided. Theoretically, there is no much people who are having issues with buffers shared between different SAEA instances, and buffers pooling may really save time when connections are frequently opened/closed. In contrast to SAEA pooling, where this does not have notable impact on performance.
So instead of just removing the idea of pools from code and configuration, I have added another pool implementation, `DisabledBufferPool` class, which is used by default (see updated in `Pigeon.conf` file). This will allow to simply switch back to pooling if this is required, also allows to set allocated buffer sizes (and possibly any other settings in future) in convenient way. 
If we are sure that we want to completely remove possibility of using buffer pools, then we can get rid of this configurations at `akka.tcp` and just have `akka.tcp.allocated-buffer-size` setting.


Aaronontheweb:
cc @IgorFedchenko missing https://getakka.net/community/public-api-changes.html

IgorFedchenko:
@Aaronontheweb Ah, this is where API approval is really useful. `DisabledBufferPool` should be internal of course, so fixed this, and no public API changed required now.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -619,31 +619,84 @@ protected void AcquireSocketAsyncEventArgs()
...
 
             if (SendArgs != null)
             {
-                Tcp.SocketEventArgsPool.Release(SendArgs);
+                ReleaseSocketEventArgs(SendArgs);
                 SendArgs = null;
             }
         }
 
+        protected SocketAsyncEventArgs CreateSocketEventArgs(IActorRef onCompleteNotificationsReceiver)
```
Aaronontheweb: LGTM

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -619,31 +619,84 @@ protected void AcquireSocketAsyncEventArgs()
...
 
         private void ReleaseSocketAsyncEventArgs()
         {
             if (ReceiveArgs != null)
             {
                 var buffer = new ByteBuffer(ReceiveArgs.Buffer, ReceiveArgs.Offset, ReceiveArgs.Count);
-                Tcp.SocketEventArgsPool.Release(ReceiveArgs);
+                ReleaseSocketEventArgs(ReceiveArgs);
                 //TODO: there is potential risk, that socket was working while released. In that case releasing buffer may not be safe.
                 Tcp.BufferPool.Release(buffer);
```
Aaronontheweb: Is this line still necessary?

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -42,12 +40,9 @@ public PreallocatedSocketEventAgrsPool(int initSize, EventHandler<SocketAsyncEve
 
         public SocketAsyncEventArgs Acquire(IActorRef actor)
         {
-            SocketAsyncEventArgs e;
-            if (!_pool.TryPop(out e))
-            {
+            if (!_pool.TryPop(out var e))
```
Aaronontheweb: Should we just get rid of this class altogether, and the HOCON that goes with it?

Aaronontheweb:
@IgorFedchenko have some additional questions on this PR

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: This depends on `Tcp.BufferPool` implementation used. If we will remove pooling completely and will allocate new byte buffers each time when new connection is opened (that's what `DisabledBufferPool` is doing) - that we will be able to remove this line. 

For now just updated to reference DirectBufferPool class

src/core/Akka/IO/SocketEventArgsPool.cs
IgorFedchenko: We wanted to keep pooling for UDP, right? That's why there is still some configuration for this in HOCON, and this class itself. 
I did not look into UDP implementation yet, so not sure we should touch it (and if it will improve anything).

src/core/Akka/IO/TcpConnection.cs
IgorFedchenko: cc @Aaronontheweb

IgorFedchenko:
@Aaronontheweb Answered questions. 
So, the main question is do we want to keep byte buffer (not SAEA) pools to be available as an option (not default). Current implementation allows to switch buffer pools back (but not SAEA), because it may slow down processing in some cases. But we can also strip this off completely. From TCP. And we are keeping it for UDP, right?

Aaronontheweb:
@IgorFedchenko I'm tempted to say ["YAGNI"](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) here - having each connection manage its own two SAEA objects and re-use those makes sense for TCP. Doing it for UDP too.... Eh, I'm not sure to be honest.

I appreciate the conservatism of your approach here - keeping our options open to change again int he future. The best thing to do in the short run is probably to stick with your plan and wait and see if that feature ever gets used.

IgorFedchenko:
> I appreciate the conservatism of your approach here - keeping our options open to change again int he future. The best thing to do in the short run is probably to stick with your plan and wait and see if that feature ever gets used.

Yup, just lets verify that nobody will have any issues with all pool types disabled (for SAEA and for byte array buffers) - and then will make cleanup of TCP pools at all.

PR #4147, opened 2020-01-16
Added missing brackets in HOCON test configs

IgorFedchenko:
Turned out that few tests have bad HOCON configuration specified - fixed this.


Aaronontheweb:
Nice catch

PR #4148, opened 2020-01-20
#2892 - TestKit: create child actors via TestProbe

### LINKED ISSUE #2892: Port TestKit parent actor utilities
See: https://github.com/akka/akka/pull/21201.

We are currently missing these TestKit facilities, And they should be ported.

nagytech:
Closes #2892


Aaronontheweb:
@nagytech needs API approval: https://getakka.net/community/public-api-changes.html

docs/articles/actors/testing-actor-systems.md
```diff
@@ -179,6 +179,11 @@ The first option is to avoid use of the `context.parent` function and create a c
 
 [!code-csharp[DependentChild](../../../src/core/Akka.Docs.Tests/Testkit/ParentSampleTest.cs?range=44-57)]
 
+###Create the child using the TestProbe
+The `TestProbe` class can directly create child actors using the `ChildActorOf` methods.  
+
+[!code-csharp[TestProbeChild](../../../src/core/Akka.Docs.Tests/Testkit/ParentSampleTest.cs?range=60-64)]
```
Aaronontheweb: I'd change this to use `#region` targeting instead, i.e. https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Docs.Tests/Streams/StreamRefsDocTests.cs#L20 - you can actually target the region name using DocFx: https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html#code-snippet-sample

src/core/Akka.TestKit/TestActor.cs
```diff
@@ -118,5 +119,53 @@ public class SetAutoPilot : INoSerializationVerificationNeeded
...
             public AutoPilot AutoPilot { get { return _autoPilot; } }
         }
+
+        /// <summary>
+        /// Message which is intended to allow TestKit to spawn a child actor
+        /// </summary>
+        public class Spawn : INoSerializationVerificationNeeded
+        {
+            /// <summary>
+            /// TBD
```
Aaronontheweb: I'd just leave off the `TBDs` rather than add new ones.

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -450,6 +452,54 @@ protected virtual void Shutdown(ActorSystem system, TimeSpan? duration = null, b
...
         }
 
+        /// <summary>
+        /// Spawns an actor as a child of this test actor, and returns the child's IActorRef
+        /// </summary>
+        /// <param name="props">Child actor props</param>
+        /// <param name="name">Child actor name</param>
+        /// <param name="supervisorStrategy">Supervisor strategy for the child actor</param>
+        /// <returns></returns>
+        public IActorRef ChildActorOf(Props props, string name, SupervisorStrategy supervisorStrategy)
```
Aaronontheweb: These all look good to me

docs/articles/actors/testing-actor-systems.md
Aaronontheweb: Reason why I recommend it: re-organizing the file doesn't break documentation as easily.

docs/articles/actors/testing-actor-systems.md
```diff
@@ -187,7 +192,7 @@ If you prefer to avoid modifying the parent or child constructor you can create
 ###Externalize child making from the parent
 Alternatively, you can tell the parent how to create its child. There are two ways to do this: by giving it a `Props` object or by giving it a function which takes care of creating the child actor:
 
-[!code-csharp[FabrikatedParent](../../../src/core/Akka.Docs.Tests/Testkit/ParentSampleTest.cs?range=84-103)]
+[!code-csharp[FabrikatedParent](../../../src/core/Akka.Docs.Tests/Testkit/ParentSampleTest.cs?range=94-110)]
```
Aaronontheweb: Same feedback as above.

docs/articles/actors/testing-actor-systems.md
nagytech: The latest changes should be using DocFX named references now.  

To help convert the remaining docs, I've written a program to go through all the .md files and automatically convert line references to C# regions.  I think there's a few references that use `#` syntax, so might need to be adapted in the future. 

https://gist.github.com/nagytech/bb07f1c46cda7b88ed8c3a5f4df7e00b

Aaronontheweb:
@nagytech still need to commit your API approvals - last thing we need before I can merge this in.

nagytech:
@Aaronontheweb I just noticed this was pending for '1 change requested'.  But I can't see any outstanding comments that I would need to resolve.

Aaronontheweb:
@nagytech I thought it was because the API approvals were missing, but maybe I'm mistaken about that

PR #4150, opened 2020-01-20
harden NodeChurnSpec

Aaronontheweb:
One of the worst offenders for racy tests... Taking a look at it now.


Aaronontheweb:
Issue was the the nodecount was pretty large - 18 nodes in total. Depending on how the random gossip went and the start of the three processes, it's possible that some of the nodes would be slower to become members than others. 

I also noticed that we weren't as diligent about shutting down nodes cleanly as we should have, and that might have contributed to the problem too (since failures tended to occur later on during the test.)

So rather than run 3 rounds with 18 nodes each, decided to run 5 rounds with 12 nodes each. Tests the same thing (gossip growth and cluster stability) but with fewer nodes to manage during each iteration.

PR #4157, opened 2020-01-23
Added xunitSettings.props to all test assemblies

### LINKED ISSUE #4076: Try to reduce number of failing time-sensitive tests with running all tests in sequence
Need to reference `xunitRunner.props` from all test specs projects and see if this will help.

IgorFedchenko:
Close #4076

Let's just see how this will help with running racy tests in this PR

### DIFF
--- a/src/core/Akka.Remote.Tests.MultiNode/Akka.Remote.Tests.MultiNode.csproj
+++ b/src/core/Akka.Remote.Tests.MultiNode/Akka.Remote.Tests.MultiNode.csproj
@@ -1,6 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\common.props" />
-  <Import Project="..\..\xunitSettings.props" />
 
   <PropertyGroup>
     <AssemblyTitle>Akka.Remote.Tests.MultiNode</AssemblyTitle>
--- a/src/core/Akka.Remote.Tests.MultiNode/Akka.Remote.Tests.MultiNode.csproj
+++ b/src/core/Akka.Remote.Tests.MultiNode/Akka.Remote.Tests.MultiNode.csproj
@@ -1,5 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\common.props" />
+  <Import Project="..\..\xunitSettings.props" />
 
   <PropertyGroup>
     <AssemblyTitle>Akka.Remote.Tests.MultiNode</AssemblyTitle>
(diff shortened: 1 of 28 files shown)


src/xunitSettings.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <!-- use the xunit JSON settings file -->
   <ItemGroup>
-    <Content Include="..\..\xunit.runner.json" Link="xunit.runner.json">
+    <Content Include="$(SolutionDir)xunit.runner.json" Link="xunit.runner.json">
```
Aaronontheweb: Good change.

src/core/Akka.Remote.Tests.MultiNode/Akka.Remote.Tests.MultiNode.csproj
```diff
@@ -1,5 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\common.props" />
+  <Import Project="..\..\xunitSettings.props" />
```
Aaronontheweb: Won't have any impact on the MNTR - those tests are already single threaded.

IgorFedchenko: Will remove this entries once CI will finish current run. It is going well so far

IgorFedchenko: Will remove this entries once CI will finish current run. It is going well so far

Aaronontheweb:
NodeChurn spec failed, but we already know that one has issues: https://github.com/akkadotnet/akka.net/pull/4150

Aaronontheweb:
Overall I agree with you @IgorFedchenko - these results look promising. Kicked off another build with the latest `dev` changes.

IgorFedchenko:
Another thing we could make here is running PR Validation steps sequentially instead of running in parallel. But on the other hand, tasks should be running in separate containers (they even use different docker images) - so they should be executed independently anyway.

IgorFedchenko:
So this run has NodeChurn  and 2 more specs that are known to be racy. Looks like any other CI run on other PRs at this time.

All right, seems like we may need this PR merged, but only after some nice work on #3786. After tests will be less racy (if not fixed at all) they still may fail when running with big concurrency, but without xunit collections parallelization may behave well.

@Aaronontheweb What do you think?

IgorFedchenko:
Actually, let's make some more runs to verify that there is any more or less stable improvement.

IgorFedchenko:
Ok, agree with you. Let me just remove those `xunitSettings.props` from MNTR test projects

PR #4158, opened 2020-01-23
Fix racy test: Akka.Cluster.Sharding.Tests.CoordinatedShutdownShardingSpec.Sharding_and_CoordinatedShutdown_must_run_successfully

IgorFedchenko:
Related to #3786

This PR should improve/fix flip rate of this test:  Akka.Cluster.Sharding.Tests.CoordinatedShutdownShardingSpec.Sharding_and_CoordinatedShutdown_must_run_successfully

It is time sensitive and failing due to timeouts (because of remote address gating). Seems like the test is correct by itself, so lets start with using `async` test APIs and increasing timeouts.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/CoordinatedShutdownShardingSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/CoordinatedShutdownShardingSpec.cs
@@ -103,9 +103,9 @@ namespace Akka.Cluster.Sharding.Tests
         /// <summary>
         /// Using region 2 as it is not shutdown in either test.
         /// </summary>
-        private void PingEntities()
+        private async Task PingEntities()
         {
-            AwaitAssert(() =>
+            await AwaitAssertAsync(() =>
             {
                 _region2.Tell(1, _probe2.Ref);
                 _probe2.ExpectMsg<int>(1.Seconds()).Should().Be(1);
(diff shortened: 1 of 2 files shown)


IgorFedchenko:
I guess we need to run this 10-20 times to make sure it is not failing anymore, since it is passing in 87% cases (according to [this report](https://dev.azure.com/dotnet/Akka.NET/_test/analytics?definitionId=84&contextType=build)

IgorFedchenko:
Mm, also most likely we need to start all the Akka tests to make CI loaded... So will add some spaces to the core project.

src/contrib/cluster/Akka.Cluster.Sharding.Tests/CoordinatedShutdownShardingSpec.cs
```diff
@@ -103,36 +103,36 @@ protected override void BeforeTermination()
...
         /// Using region 2 as it is not shutdown in either test.
         /// </summary>
-        private void PingEntities()
+        private async Task PingEntities()
         {
-            AwaitAssert(() =>
+            await AwaitAssertAsync(() =>
             {
                 _region2.Tell(1, _probe2.Ref);
-                _probe2.ExpectMsg<int>(1.Seconds()).Should().Be(1);
```
Aaronontheweb: I'd leave these at 1 second each, but just expand the total wait time.

Aaronontheweb: (which you already did)

PR #4168, opened 2020-01-28
Add documentation for Akka.Cluster.Metrics

IgorFedchenko:
Added article documenting how Akka.Cluster.Metrics extension works and how it can be used.

Also, added some minor changes in code which I noticed while was writing documentation.


docs/articles/clustering/cluster-metrics.md
```diff
@@ -0,0 +1,391 @@
...
+Different collector implementations may provide different subsets of metrics published to the cluster. 
+Metrics currently supported are defined in `Akka.Cluster.Metrics.StandardMetrics` class:
+* `MemoryUsed` - total memory allocated to the currently running process
+* `MemoryAvailable` - memory, available for the process
+* `MaxMemoryRecommended` - if set, memory limit recommended for current process
+* `Processors` - number of available processors
+* `CpuProcessUsage` - CPU usage by current process
+* `CpuTotalUsage` - total CPU usage
+
+> Note: currently, due to some .NET Core limitations `CpuTotalUsage` is the same as `CpuProcessUsage` metrics,
```
Aaronontheweb: I'd include a link to the Github issue where we discuss fixing this.

docs/articles/clustering/cluster-metrics.md
```diff
@@ -0,0 +1,391 @@
...
+
+The collected metrics values are smoothed with [exponential weighted moving average](https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average). 
+In the cluster configuration you can adjust how quickly past data is decayed compared to new data.
+
+Let’s take a look at this router in action. What can be more demanding than calculating factorials?
+
+The backend worker that performs the factorial calculation:
+
+```c#
+public class FactorialBackend : ReceiveActor
```
Aaronontheweb: We should include this type of material inside https://github.com/akkadotnet/akka.net/tree/dev/src/core/Akka.Docs.Tests and use DocFx syntax to link from the code in that project, i.e.:

https://raw.githubusercontent.com/akkadotnet/akka.net/dev/docs/articles/streams/stream-dynamic.md

```
[!code-csharp[KillSwitchDocTests.cs](../../../src/core/Akka.Docs.Tests/Streams/KillSwitchDocTests.cs?name=unique-shutdown)]
```

This guarantees that in the event that we refactor / change these APIs, the example included in the documentation will break - thus if we do things this way we can guarantee that all examples are statically type checked... And since they're modeled as unit tests, they also have to be functionally accurate as well. 

This is also designed to help prevent code samples in the documentation from rotting over time, since most contributors (myself included) don't go back over them and check them as often. So going forward we should implement this best practice throughout the Akka.NET documentation. Does that make sense?

IgorFedchenko: Sure, makes sense. I was thinking about something like that, but checked out cluster sharding sample and saw that code was just pasted there, so made in the same manner.

Thanks for samples and links, I will update

docs/articles/clustering/cluster-metrics.md
IgorFedchenko: Sure

docs/articles/clustering/cluster-metrics.md
Aaronontheweb: Would you mind opening an issue to fix that with our Cluster.Sharding documentation too? We should, ideally, go back and replace these static code samples with the dynamically linked ones everywhere.

src/core/Akka.Docs.Tests/Clustet.Metrics/RoutersUsageSample.cs
```diff
@@ -97,6 +100,6 @@ private void SendJobs()
             }
         }
     }
-
+    // </FactorialFrontend>
```
Aaronontheweb: Oh wow, this works without using `#region`s ?

docs/articles/toc.yml
```diff
@@ -166,6 +166,8 @@
     href: clustering/cluster-client.md
   - name: Cluster Sharding
     href: clustering/cluster-sharding.md
+  - name: Cluster Metrics
+    href: clustering/cluster-metrics.md
```
Aaronontheweb: LGTM

docs/articles/clustering/cluster-metrics.md
```diff
@@ -27,7 +27,7 @@ Metrics currently supported are defined in `Akka.Cluster.Metrics.StandardMetrics
 * `CpuTotalUsage` - total CPU usage
 
 > Note: currently, due to some .NET Core limitations `CpuTotalUsage` is the same as `CpuProcessUsage` metrics, 
-> but this is something to be fixed in near future.
+> but this is something to be fixed in near future (see [this issue](https://github.com/akkadotnet/akka.net/issues/4142) for details).
```
Aaronontheweb: Thanks for adding that

IgorFedchenko:
@Aaronontheweb Updated docs file to use code references instead of directly pasted code.

The only issue is that I could not check it with `serve-docs.cmd` , still getting same NRE error I had few months ago. You said that you have same error on one of your machines - did you manage to resolve this?

src/core/Akka.Docs.Tests/Clustet.Metrics/RoutersUsageSample.cs
IgorFedchenko: Yup, it should, according to this description: https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html#tag-name-representation-in-code-snippet-source-file

IgorFedchenko: As I said below, could not check if all this references work properly...

Aaronontheweb: Guess we'll find out - would love to get rid of regions elsewhere.

Aaronontheweb:
@IgorFedchenko not sure how I fixed it - might have been updating my Visual Studio version that did it.

docs/articles/clustering/cluster-metrics.md
IgorFedchenko: Sure, but it is not only Cluster.Sharding - there are few more files to be fixed. 

Here is an issue: https://github.com/akkadotnet/akka.net/issues/4169

Issue #3615, opened 2018-10-09
Rewrite MultiNodeTestRunner as .NET Core tool


Aaronontheweb:
Having done this recently for NBench: https://github.com/petabridge/NBench/pull/240/ 

I think that it'd be an end-user experience improvement to leverage .NET Core's built-in tooling for the MNTR. We can go down the global tool route or use a downloadable package. But what I like about this idea is that it makes it possible for us to target the `.csproj` file used for the multi-node project rather than the output assemblies. This will help simplify build scripts significantly and in particular will make projects that need to support multi-targeting, like Akka.NET itself, easier to work with.

igormcoelho:
Perhaps related to: https://github.com/akkadotnet/akka.net/issues/3696

IgorFedchenko:
@Aaronontheweb So something like this?
```
<ItemGroup>
  <DotNetCliToolReference Include="dotnet-mntr" />
</ItemGroup>
```

Aaronontheweb:
Yeah I'd either do it via:

1. The way you suggested;
2. Writing a Visual Studio plugin so `dotnet test` and the VS Test Runner can pick up MNTR specifications and we can attach a debugger to them. You can see some examples of how to do this here: https://github.com/microsoft/VSSDK-Extensibility-Samples

Aaronontheweb:
I've had a lot of issues getting `dotnet nbench` to perform consistently well across multiple versions of .NET Core - ultimately ran into the same kinds of problems the xUnit project did with `dotnet xunit` - that's why they deprecated it after xUnit 2.3.1 and change their guidance to recommend that all developers start using `dotnet test` instead - that was Visual Studio becomes responsible for finding all of the right framework / runtime binaries instead of your program.

IgorFedchenko:
@Aaronontheweb So seems like we have so things to implement here:
1. Rewrite MNTR to include custom test runner supported by `dotnet test`, so that running `dotnet test` after installing our nuget packages (like `xunit` and `xunit.runner.visualstudio`) would run MNTR tests for given project.
This does not seem to be related to Visual Studio, since it is just using `dotnet test` command with custom test runner (as described [here](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test?tabs=netcore21#description))

2. Write VS extension that will detect MNTR specifications. Speaking of that, JetBrains Rider already suggests me to run MNTR spec:
[...]

Issue #4146, opened 2020-01-15
Add async overloads to MultiNode test APIs, and update existing tests


IgorFedchenko:
We should add overloads like `EnterBarrierAsync` for all blocking methods. And also would be nice to rewrite existing tests to use this APIs, this may help with racy issues on CI.

Aaronontheweb:
:1:

PR #4177, opened 2020-01-30
Implement custom VisualStudio test runner for multi-node tests discovery and execution

IgorFedchenko:
This PR is work on #3615

The goal is to be able to include something like thins into multi-node specs project:
```
<ItemGroup>
      <PackageReference Include="Akka.MultiNodeTestRunner.VisualStudio" Version="1.0.0" />
  </ItemGroup>
```

And then running `dotnet test` should discover all MNTR tests and use MNTR to execute those tests.


IgorFedchenko:
Will move on with small steps. 

#### First step
Since we have our multi-node specs using `[MultiNodeFact]` attributes derived from xunit `[Fact]` attribute, looks like we can reuse some part of [xunit.runner.visualstudio](https://github.com/xunit/visualstudio.xunit) package source. Ideally would be to reference this package and inherit from those types , but I will start from copying their source and making it work as a `Akka.MultiNodeTestRunner.VisualStudio` package.

Actually, right now I already have this done - put this new package source in separate solution (because in Akka.NET solution it was building with some weird nuget warnings, so I decided to use separate solution - at the end, we are going to move it to separate repo anyway). I am publishing package to local folder nuget feed, and using `Akka.Remoting.Tests.MultiNode` project as a sample tests project. 
Right now my package does not discover those tests, but I will fix this. I guess this is related to package name that is hardcoded somewhere in runner sources, as assembly name.

#### Second step
After I will have my own working `Akka.MultiNodeTestRunner.VisualStudio` package, I will try to modify tests discovery to skip usual `[Fact]` tests and only find `[MultiNodeFact]`.

#### Third step
Here is where the actual fun starts: I will need to rewrite test executor to use our MNTR code (possibly modified in some degree) to start `Akka.NodeTestRunner`'s. 

Currently xunit [runs all tests in same process](https://github.com/xunit/xunit/issues/1044#issuecomment-265347025), to enable debugging feature. So most likely **we still will not be able to debug code on nodes** during tests. @Aaronontheweb - do we really need this process isolation on MNTR? Or we can run nodes in separate app domains to give managed code isolation, but be able to debug node code execution?

Anyway, looking at [this doc](https://xunit.net/docs/running-tests-in-parallel) looks like tests execution may be delegated to some other packages like `xunit.execution.dll` - so maybe will need to make forks of those as well.

In addition to existing test runner sources (thank you, opensource!) this resources give some light on how this test runners are structured:
https://docs.microsoft.com/en-us/archive/blogs/bhuvaneshwari/authoring-a-new-visual-studio-unit-test-adapter
http://matthewmanela.com/blog/anatomy-of-the-chutzpah-test-adapter-for-vs-2012-rc/
https://stackoverflow.com/a/38206731/3094849

Aaronontheweb:
> Or we can run nodes in separate app domains to give managed code isolation, but be able to debug node code execution?

We have tried doing this in the past without much success, but you can always try again. I'm open

IgorFedchenko:
> We have tried doing this in the past without much success, but you can always try again. I'm open

Ok, got it. The main reason why I am suggesting this is because xunit should be already doing this in their runner, so I can try to reuse it

Aaronontheweb:
The issues we had in the past amounted to just issues with race conditions being worse and so forth. If we're migrating many of the MNTR methods to be asynchronous (and, in fact, if we implemented the `Main` method in the `NodeRunner.exe` to be `async Task main`) that might help alleviate some of those issues caused by having many threads all blocking / waiting on each other in each individual node.

IgorFedchenko:
Hmm, I see. We shell try then.
Maybe I should also work on closing [this issue](https://github.com/akkadotnet/akka.net/issues/4146) as well.

IgorFedchenko:
Closing this PR because MNTR moved to separate repository. Further work on this PR is here: https://github.com/akkadotnet/Akka.MultiNodeTestRunner/pull/8

PR #4181, opened 2020-01-31
Migrate entire build system and test runtime to .NET Core 3.1

Aaronontheweb:
Doesn't impact the framework itself, only the environment in which it is built and tested.


Aaronontheweb:
closed via https://github.com/akkadotnet/akka.net/pull/4216

PR #4196, opened 2020-02-10
enabled DData MNTR ReplicatorPruningSpec

Aaronontheweb:
close #3080 
close #4198
close #4199
close #4200

PR #4201, opened 2020-02-11
added Akka.DistributedData to Akka.Cluster friends list

Aaronontheweb:


Aaronontheweb:
Part of https://github.com/akkadotnet/akka.net/pull/4196 - but should be in its own PR since it triggers a much bigger set of changes than what we're actually trying to address in the other PR.

PR #4207, opened 2020-02-12
Fix/3129 pubsub - reworking MNTR specifications

Aaronontheweb:
MNTR specs for https://github.com/akkadotnet/akka.net/pull/4086 needed some work. Opening a PR for them here.


Aaronontheweb:
Reverted back to the original MNTR implementation for this spec. Not up for a lot of debugging to see why things got racy in the other PR.

PR #4216, opened 2020-02-13
Migrate entire build system and test runtime to .NET Core 3.1

IgorFedchenko:
Continuation of https://github.com/akkadotnet/akka.net/pull/4181


IgorFedchenko:
@Aaronontheweb They have changed dotnet-install script url - old one returns 404 code. Hope this will make CI pass now

IgorFedchenko:
Oops, there is no `3.1.0` version of sdk - it should be `3.1.100` (need strict version match)

Aaronontheweb:
> Oops, there is no 3.1.0 version of sdk - it should be 3.1.100 (need strict version match)

https://media.giphy.com/media/KcW2FVEJnzKeiZkVI1/giphy.gif

IgorFedchenko:
Sure. Will need to also update `netcoreapp2.1` to `3.1` in `build.fsx`

Aaronontheweb:
@IgorFedchenko looks like tests never ran on .NET Core Linux

IgorFedchenko:
@Aaronontheweb Indeed - incrementalist did not find any changes for some reason...

Aaronontheweb:
@IgorFedchenko platform changes may have affected Incrementalist perhaps

Aaronontheweb:
Although only on Linux? 😢

IgorFedchenko:
Just took a quick glance to the Incrementalist source - it is targeting netcoreapp2.1, so maybe has some issues with loading projects under linux... If I will not get better idea I'll just run it under docker container and see what is going on there.

IgorFedchenko:
@Aaronontheweb The issues with NRE were due to some weird code that was trying to get `System.Runtime.Serialization.FormatterServices` class via reflection from `typeof(string)..GetTypeInfo().Assembly` assembly. Maybe this was a hack for some old .net core support, because now this type is available in compile time.

I guess this `System.Runtime.Serialization.FormatterServices` was just moved to separate nuget package after some netcore update, so it is no longer found near to `string` type.

IgorFedchenko:
But it looks like Incrementalist still ignores changes in all projects under linux/netcore3.1, except the one I have just updated to fix NRE exception.
So will try to reproduce issue, and post either issue to Incrementalist repo, or fix it here

IgorFedchenko:
Hmm, actually right now both Windows and Linux incrementalist showing 
```
Detected 1 affected projects in solution - writing out to D:\a\1\s\bin\incrementalist.txt
```

So it's hard to say now why it was ignoring `common.props` change that time.

@Aaronontheweb  I guess better to create separate issue and investigate this behavior after this PR will be merged. Because `netcoreapp3.1` runtime looks working (so far only racing specs failed), and this edge case with `common.props` need to be checked separately. 

Can not repro issue locally. Tried to setup incrementalist in docker, but it uses `libgit2sharp` which does not provide native binaries for all platforms (like for `debian.10-x64` which is used in official `sdk:3.1` image)... So need additional efforts, and sounds like a separate issue, possibly related to incrementalist itself.

IgorFedchenko:
So confirmed bug in Incrementalist: https://github.com/petabridge/Incrementalist/issues/106

@Aaronontheweb Should we wait for this bug to be fixed in Incrementalist before merging this PR? On the one hand, build system works. On the other hand, if we will have another PR that changes common.props versions (like from @dependabot), they will be skipped under Linux.

Aaronontheweb:
@IgorFedchenko I think for the time being it's probably ok - we haven't done an Incrementalist update since November and since then we've updated the underlying Roslyn version and some other tools. But, worst case scenario, we can manually re-run the Linux test suite until we get this resolved. Better to have the technical debt paid off on this repo by running the latest .NET Core version going forward.

PR #4219, opened 2020-02-14
Fix/check common.props change detection under linux

IgorFedchenko:
Possible issue found at  https://github.com/akkadotnet/akka.net/pull/4216

When only `common.props` file is changed and we are targeting netcoreapp3.1 in CI, no project are changed according to Incrementalist tool under linux.

So first of all need to confirm the issue exists, and either resolve it here or create issue in Incrementalist repo.


IgorFedchenko:
So can see this issue again here.

IgorFedchenko:
All right, this is going to be Incrementalist issue anyway.
Let's just verify that any change in common.props is ignored, even less important (like description update). In current implementation Incrementalist should mark all referencing projects as updated even in such case.

IgorFedchenko:
So created this issue: https://github.com/petabridge/Incrementalist/issues/106

Closing this PR as a finished repro.

PR #4226, opened 2020-02-17
Fix config key lookup performance regression

### LINKED ISSUE #4221: ActorSpawn performance regression (HOCON)
Version: Akka.NET v1.4.0-RC1

After merging https://github.com/akkadotnet/akka.net/pull/4128, which introduced stand-alone HOCON integration into Akka.NET, I ran our [SpawnBenchmark](https://github.com/akkadotnet/akka.net/tree/dev/src/benchmark/SpawnBenchmark)

Some figures - running on an 8-core 3.0 Ghz Ryzen architecture.

### Akka.NET v1.4.0-beta4 (no stand-alone HOCON)

```
Is Server GC True
Start run 5
Run 5 result: 499999500000 in 5110 ms
Start run 4
Run 4 result: 499999500000 in 4856 ms
Start run 3
Run 3 result: 499999500000 in 5757 ms
Start run 2
Run 2 result: 499999500000 in 5596 ms
Start run 1
Run 1 result: 499999500000 in 5845 ms
```

### Akka.NET v1.4.0-RC1 (with stand-alone HOCON)

```
Is Server GC True
Start run 5
Run 5 result: 499999500000 in 7866 ms
Start run 4
Run 4 result: 499999500000 in 7522 ms
Start run 3
Run 3 result: 499999500000 in 7394 ms
Start run 2
Run 2 result: 499999500000 in 7396 ms
Start run 1
Run 1 result: 499999500000 in 7365 ms
```

That's a performance regression of about 50% for actor spawning - and I suspect the issue is the speed at which we're able to look up values from HOCON inside the `Deployer` during `IActorRef` creation.

Related issue: https://github.com/akkadotnet/HOCON/issues/200

Arkatufus:
close #4221

### DIFF
--- a/src/core/Akka/Configuration/FlatConfig.cs
+++ b/src/core/Akka/Configuration/FlatConfig.cs
@@ -30,7 +30,8 @@ namespace Akka.Configuration
         {
             if (_cache.TryGetValue(path, out var result))
                 return result;
-            result = base.GetNode(path);
+
+            result = Root.GetObject().GetValue(HoconPath.Parse(path));
             _cache[path] = result;
             return result;
         }
@@ -40,7 +41,8 @@ namespace Akka.Configuration
             var fullPath = path.ToString();
             if (_cache.TryGetValue(fullPath, out var result))
                 return result;
-            result = base.GetNode(path);
+
+            result = Root.GetObject().GetValue(path);
             _cache[fullPath] = result;
             return result;
         }
@@ -50,11 +52,15 @@ namespace Akka.Configuration
             if (_cache.TryGetValue(path, out result))
                 return true;
 
-            if (!base.TryGetNode(path, out result))
-                return false;
+            if(HoconPath.TryParse(path, out var hoconPath))
+                if (Root.TryGetObject(out var obj))
+                    if (obj.TryGetValue(hoconPath, out result))
+                    {
+                        _cache[path] = result;
+                        return true;
+                    }
 
-            _cache[path] = result;
-            return true;
+            return false;
         }
 
         protected override bool TryGetNode(HoconPath path, out HoconValue result)
@@ -63,11 +69,13 @@ namespace Akka.Configuration
             if (_cache.TryGetValue(fullPath, out result))
                 return true;
 
-            if (!base.TryGetNode(path, out result))
-                return false;
-
-            _cache[fullPath] = result;
-            return true;
+            if(Root.TryGetObject(out var obj))
+                if(obj.TryGetValue(path, out result))
+                {
+                    _cache[fullPath] = result;
+                    return true;
+                }
+            return false;
         }
     }
 }
--- a/src/core/Akka/Configuration/FlatConfig.cs
+++ b/src/core/Akka/Configuration/FlatConfig.cs
@@ -1,0 +1,73 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Text;
+using Hocon;
+
+namespace Akka.Configuration
+{
+    public class FlatConfig:Config
+    {
+        private ConcurrentDictionary<string, HoconValue> _cache = new ConcurrentDictionary<string, HoconValue>();
+
+        public FlatConfig(HoconRoot root, Config fallback) : base(root, fallback)
+        {
+            if (root is FlatConfig caching)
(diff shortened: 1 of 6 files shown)


src/core/Akka/Configuration/FlatConfig.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Text;
+using Hocon;
+
+namespace Akka.Configuration
+{
+    public class FlatConfig:Config
```
Aaronontheweb: Do we want to mark this as `internal` and `sealed`?

src/core/Akka/Configuration/FlatConfig.cs
```diff
@@ -0,0 +1,73 @@
...
+            return result;
+        }
+
+        protected override HoconValue GetNode(HoconPath path)
+        {
+            var fullPath = path.ToString();
+            if (_cache.TryGetValue(fullPath, out var result))
+                return result;
+            result = base.GetNode(path);
+            _cache[fullPath] = result;
```
Aaronontheweb: So we shouldn't do this - it'll create a large memory footprint for systems that have to deploy a huge number of actors. We want to avoid storing a cached object for _every single actorpath_ we resolve. Ideally, we want to only store cached objects for paths that get used.

This is what the `CachingConfig` does - we have to strike a balance between memory and throughput here and this goes too far in the direction of trading away memory for throughput.

src/core/Akka/Configuration/FlatConfig.cs
Arkatufus: I think `sealed` is better, yes

src/core/Akka/Configuration/FlatConfig.cs
Arkatufus: Its not, really... it only caches the hits and ignores the misses, the memory hit shouldn't be that bad.

Arkatufus:
> 1 - Improve the performance of `CachingConfig` by reducing duplicate HOCON lookups on cache misses here:

We can't reduce duplicate cache misses unless we store them, which costs memory :(

>     1. Improve the performance of the underlying HOCON methods themselves - this is where the performance regression originally came from.  This will take more time and effort to do though as it'll require benchmarking and improving the HOCON lookup methods individually.

I think the biggest performance hit on using standalone HOCON is that each time a `string` path is passed on into a function, it has to be parsed by the full blown HOCON parser tokenizer to ensure that the path is in the proper HOCON format. One way to circumvent this is by using `HoconPath` constructor directly on time critical codes.
 
>     2. Merging fallbacks and configs on startup inside the `CachingConfig` - this might help improve the speed of cache misses, since we have the entire `Config` merged together upon creation of the `CachingConfig` object. This means fewer fallback traversals on cache misses, in particular.

This is a good idea, we should've done this earlier, we already have a cache of the merged object on hand, there is no need to use the fallbacks to resolve any values.

Aaronontheweb:
Before

``` ini

.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT
  Job-LYHBZY : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT

InvocationCount=100000  IterationCount=20  RunStrategy=Throughput  
WarmupCount=5  

```
|      Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------ |---------:|---------:|---------:|-------:|-------:|------:|----------:|
| Actor_spawn | 34.62 us | 1.082 us | 1.158 us | 2.7400 | 0.6700 |     - |  16.43 KB |

src/core/Akka/Configuration/FlatConfig.cs
Aaronontheweb: Got it - based on our conversation from earlier it sounds like I misunderstood.

Aaronontheweb:
Ah shit, this benchmark dot net benchmark won't capture the right thing - can't program it to wait until all actors have started and checked in.

Aaronontheweb:
Superseded by #4231

PR #4250, opened 2020-02-25
fix RemotePingPong benchmark HOCON

Aaronontheweb:


src/benchmark/RemotePingPong/Program.cs
```diff
@@ -58,7 +58,7 @@ public static Config CreateActorSystemConfig(string actorSystemName, string ipOr
                     hostname = ""localhost""
                 }
               }
-            ");
+            }");
```
Aaronontheweb: New stand-alone HOCON library detected invalid HOCON here

PR #4269, opened 2020-02-27
harden DData.Tests.WriteAggregatorSpecs

Aaronontheweb:
per #3786

### DIFF
--- a/src/contrib/cluster/Akka.DistributedData.Tests/WriteAggregatorSpec.cs
+++ b/src/contrib/cluster/Akka.DistributedData.Tests/WriteAggregatorSpec.cs
@@ -162,9 +162,14 @@ namespace Akka.DistributedData.Tests
             // no reply
             probe.ExpectMsg<Write>();
             // no reply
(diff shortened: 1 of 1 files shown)


src/contrib/cluster/Akka.DistributedData.Tests/WriteAggregatorSpec.cs
```diff
@@ -162,9 +162,14 @@ public void WriteAggregator_must_timeout_when_less_than_required_ACKs()
             // no reply
             probe.ExpectMsg<Write>();
             // no reply
-            ExpectMsg(new UpdateTimeout(KeyA, null));
```
Aaronontheweb: We've set the `WriteTimeout` in all of the various write consistencies used in this spec to _at least_ 3 seconds depending on the consistency level used and the dilation factor, without updating any of the assertions - which wait _at most_ 3 seconds. Thus we can prematurely blow up the spec before the `WriteAggregator` has had a chance to actually generate its timeout.

PR #4273, opened 2020-02-28
Remove 1x netcore restriction

Aaronontheweb:
Updated version of https://github.com/akkadotnet/akka.net/pull/3790


Aaronontheweb:
This PR is too big and too dangerous to do so close to the release of 1.4.0.

PR #4296, opened 2020-03-05
update build YAML for v1.3

Aaronontheweb:


Aaronontheweb:
There we go - that finally got it working

PR #4299, opened 2020-03-06
Reintroduce old HOCON, lay groundwork for gradual introduction of HOCON stand-alone

Aaronontheweb:
Thinking about the breaking changes we have in the current Akka.NET v1.4.1-RC1 and how we can introduce the changes to HOCON more gradually, rather than doing it all at once, in order to make the changes more gradual and more side-by-side with the existing HOCON APIs that are built into Akka.NET.

Primary innovation here is:

- Keep original Akka.Configuration.Config code
- Have it implement an interface defined inside HOCON NuGet package, `Hocon.IHoconConfig` - non-breaking change since it already implements all of those methods;
- Gradually add new method overloads which consume `Hocon.IHoconConfig` rather than `Akka.Configuration.Config`;
- Gradually deprecate methods that take `Akka.Configuration.Config` dependencies;
- Begin using `Hocon.IHoconConfig` methods, powered by stand-alone HOCON, exclusively as the library matures.


src/core/Akka/Actor/Settings.cs
```diff
@@ -46,6 +46,75 @@ public void InjectTopLevelFallback(Config config)
             RebuildConfig();
         }
 
+        public Settings(ActorSystem system, IHoconConfig config)
```
Aaronontheweb: Example of adding new overload - without changing old one

src/core/Akka/Configuration/IHoconConfig.cs
```diff
@@ -0,0 +1,222 @@
...
+// //     Copyright (C) 2013-2020 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+
+namespace Hocon
+{
+    public interface IHoconConfig
```
Aaronontheweb: The interface - doesn't expose any underlying HOCON object graph implementation.

Issue #4304, opened 2020-03-09
Actor reference leak in AddressTerminatedTopic


petrikero:
In case where actor A, that is being watched by remote actor B, terminates immediately after being unwatched by B, there is a chance the AddressTerminatedTopic leaks a reference to A.

The issue arises due to TellWatchersWeDied() clearing the incorrect "watching" list when it should clear the watchedBy list in its finally block. Also, the clearing should be wrapped in a MaintainAddressTerminatedSubscription().

For reference, original Akka already implements the proposed change.

I have a pending fix for this and will make a PR.

Using Akka.Net version 1.3.17, though bug seems to have been there for years. Bug appears on Windows and dockerized Linux. I don't have a simple repro, but am seeing it in a larger system.

petrikero:
Here's a link to the Akka implementation in tellWatchersWeDied(): https://github.com/akka/akka/blob/6bf20f4117a8c27f8bd412228424caafe76a89eb/akka-actor/src/main/scala/akka/actor/dungeon/DeathWatch.scala#L133

petrikero:
Here's a link to the PR as well: https://github.com/akkadotnet/akka.net/pull/4305

PR #4305, opened 2020-03-09
Fixes actor ref leak in AddressTerminatedTopic (#4304)

petrikero:
I made the PR against v1.3 as that's what we're using and where I could confirm the bug and the fix.

### DIFF
--- a/src/core/Akka/Actor/ActorCell.DeathWatch.cs
+++ b/src/core/Akka/Actor/ActorCell.DeathWatch.cs
@@ -173,7 +173,10 @@ namespace Akka.Actor
             }
             finally
             {
-                _state = _state.ClearWatching();
+                MaintainAddressTerminatedSubscription(() =>
+                {
+                    _state = _state.ClearWatchedBy();
+                });
             }
         }
 
--- a/src/core/Akka/ActorState.cs
+++ b/src/core/Akka/ActorState.cs
@@ -57,6 +57,11 @@ namespace Akka.Actor
         /// <returns>TBD</returns>
         IActorState ClearWatching();
         /// <summary>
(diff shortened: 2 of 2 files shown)


petrikero:
The failing .NET Framework Unit Tests seem to me to be unrelated, but I'm not sure I'm reading the outputs correctly.

Aaronontheweb:
@petrikero the v1.3 branch is missing a bunch of racy spec fixes we introduced in the v1.4 / dev branch. I'll take a look at these.

petrikero:
Thanks! The issues are crashing our systems after a few days of running, so they're pretty critical for us. I did manage to inject a custom-built Akka.dll (with the two PRs included) into our build pipeline to solve the bleeding for now.

If there's anything more I can do to help get them merged, please let me know. I'll be happy to help, but am a bit at a loss with the size of the project. :)

Aaronontheweb:
failed specs appear to be largely the result of:
- Multiple bind on port errors (fixed in v1.4 branch)
- Specs that are tightly temporally coupled - most of them have been disabled in v1.4 branch.

src/core/Akka/Actor/ActorCell.DeathWatch.cs
```diff
@@ -173,7 +173,10 @@ protected void TellWatchersWeDied()
             }
             finally
             {
-                _state = _state.ClearWatching();
+                MaintainAddressTerminatedSubscription(() =>
```
Aaronontheweb: Wow, this is an old bug. Looks right though.

Aaronontheweb:
I'll see about having us to a v1.3.18 release to fix these.

PR #4313, opened 2020-03-09
FIX: #4079 sqlite table name was hardcoded

Arkatufus:
Closes #4079 

Extra bug fix: Sqlite plugin default config was not applied as a fallback to populate `JournalSettings` instance in `SqlJournal` constructor

### NOTE ###
This might be a breaking change for users that copy pasted sqlite journal configuration from a sample that contains a `table-name = not_called_snapshot` entry because it will start to use the property declaration instead of using the hard coded `snapshot` table.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Settings.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Settings.cs
@@ -68,14 +68,14 @@ namespace Akka.Persistence.Sql.Common
             if (config.IsNullOrEmpty())
                 throw ConfigurationException.NullOrEmptyConfig<JournalSettings>();
 
-            ConnectionString = config.GetString("connection-string", null);
-            ConnectionStringName = config.GetString("connection-string-name", null);
-            ConnectionTimeout = config.GetTimeSpan("connection-timeout", null);
+            ConnectionString = config.GetString("connection-string");
+            ConnectionStringName = config.GetString("connection-string-name");
+            ConnectionTimeout = config.GetTimeSpan("connection-timeout");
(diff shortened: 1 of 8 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Settings.cs
```diff
@@ -14,7 +14,7 @@ namespace Akka.Persistence.Sql.Common
     /// <summary>
     /// Configuration settings representation targeting Sql Server journal actor.
     /// </summary>
-    public class JournalSettings
+    public sealed class JournalSettings
```
Aaronontheweb: Good fix

src/contrib/persistence/Akka.Persistence.Sqlite/Snapshot/SqliteSnapshotStore.cs
```diff
@@ -102,25 +102,26 @@ public class SqliteSnapshotStore : SqlSnapshotStore
...
         /// <param name="snapshotConfig">TBD</param>
-        public SqliteSnapshotStore(Config snapshotConfig) : base(snapshotConfig)
+        public SqliteSnapshotStore(Config snapshotConfig) : 
+            base(snapshotConfig.WithFallback(Extension.DefaultSnapshotConfig))
         {
             var config = snapshotConfig.WithFallback(Extension.DefaultSnapshotConfig);
             QueryExecutor = new SqliteSnapshotQueryExecutor(new QueryConfiguration(
                 schemaName: null,
-                snapshotTableName: "snapshot",
+                snapshotTableName: config.GetString("table-name"),
```
Aaronontheweb: Good fix

PR #4317, opened 2020-03-10
DData Replicator deserialize ORMultiValueDictionary as ORDictionary causing invalid cast exception

### LINKED ISSUE #4302: DData Replicator Changed Message value type for ORMultiValueDictionary is sent as ORDictionary
When creating a new issue, please make sure the following information is part of your issue description. (if applicable). Thank You!

Akka Version: 1.4 RC1
Platform: Windows 10, .Net 4.7

We use Akka Clustering with singleton node performing all operations.  All nodes in the cluster register with the Replicator for change log by calling. On the singleton node we create a ORMultiValueDictionary`2[System.String,System.String] and also register with the Replicator for the changes. using the below code. All nodes in the clsuter register to get the changes for DDData.

var reg = new ORMultiValueDictionaryKey<string, string>("someregistry");
DistributedData.Get(Context.System).Replicator.Tell(Dsl.Subscribe(reg, Self));

On all non singleton nodes the Changed message sent by the replicator has invalid type, the value sent back if of type 'Akka.DistributedData.ORDictionary`2[System.String,Akka.DistributedData.ORSet`1[System.String]]'.

This leads to a type casting error 
{"Unable to cast object of type 'Akka.DistributedData.ORDictionary`2[System.String,Akka.DistributedData.ORSet`1[System.String]]' to type 'Akka.DistributedData.ORMultiValueDictionary`2[System.String,System.String]'."}

Arkatufus:
closes #4302

### DIFF
--- a/src/contrib/cluster/Akka.DistributedData/ORMultiValueDictionary.cs
+++ b/src/contrib/cluster/Akka.DistributedData/ORMultiValueDictionary.cs
@@ -34,7 +34,6 @@ namespace Akka.DistributedData
     /// </summary>
     internal interface IORMultiValueDictionaryDeltaOperation
     {
-        bool WithValueDeltas { get; }
         ORDictionary.IDeltaOperation Underlying { get; }
     }
 
@@ -305,13 +304,12 @@ namespace Akka.DistributedData
 
         #region delta
 
-        internal sealed class ORMultiValueDictionaryDelta : ORDictionary<TKey, ORSet<TValue>>.IDeltaOperation, IReplicatedDeltaSize, IORMultiValueDictionaryDeltaOperation
+        public sealed class ORMultiValueDictionaryDelta : ORDictionary<TKey, ORSet<TValue>>.IDeltaOperation, IReplicatedDeltaSize, IORMultiValueDictionaryDeltaOperation
         {
             internal readonly ORDictionary<TKey, ORSet<TValue>>.IDeltaOperation Underlying;
 
-            public ORMultiValueDictionaryDelta(ORDictionary<TKey, ORSet<TValue>>.IDeltaOperation underlying, bool withValueDeltas)
+            public ORMultiValueDictionaryDelta(ORDictionary<TKey, ORSet<TValue>>.IDeltaOperation underlying)
             {
-                WithValueDeltas = withValueDeltas;
                 Underlying = underlying;
                 if (underlying is IReplicatedDeltaSize s)
                 {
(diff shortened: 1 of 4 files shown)


src/contrib/cluster/Akka.DistributedData/ORMultiValueDictionary.cs
```diff
@@ -27,6 +27,16 @@ internal interface IORMultiValueDictionaryKey
         Type ValueType { get; }
     }
 
+    /// <summary>
+    /// INTERNAL API.
+    ///
+    /// For serialization purposes.
+    /// </summary>
+    internal interface IORMultiValueDictionaryDeltaOperation
```
Aaronontheweb: Good, similar to what we had to do with `PNCounterDictionary`

src/contrib/cluster/Akka.DistributedData/Serialization/ReplicatedDataSerializer.cs
```diff
@@ -166,6 +168,7 @@ public override string Manifest(object o)
                 case IPNCounterDictionary _: return PNCounterMapManifest;
                 case IPNCounterDictionaryDeltaOperation _: return PNCounterMapDeltaOperationManifest;
                 case IORMultiValueDictionary _: return ORMultiMapManifest;
+                case IORMultiValueDictionaryDeltaOperation _: return ORMultiMapDeltaOperationManifest;
```
Aaronontheweb: Don't we also need an equivalent `FromBinary` operation here?

Arkatufus: I think I missed that, let me check

src/contrib/cluster/Akka.DistributedData/ORMultiValueDictionary.cs
```diff
@@ -27,6 +27,16 @@ internal interface IORMultiValueDictionaryKey
...
+    {
+        bool WithValueDeltas { get; }
```
Aaronontheweb: This property is not needed - it's a property of the `ORMultiValueDictionary`, not its delta operations.

src/contrib/cluster/Akka.DistributedData/ORMultiValueDictionary.cs
```diff
@@ -295,10 +305,85 @@ public override string ToString()
...
+                return Underlying.Equals(other);
+            }
+
+            public override int GetHashCode()
+            {
+                return Underlying.GetHashCode();
+            }
+
+            public int DeltaSize { get; }
+            public bool WithValueDeltas { get; }
```
Aaronontheweb: Same feedback here - not needed.

src/contrib/cluster/Akka.DistributedData/Serialization/ReplicatedDataSerializer.cs
Arkatufus: Done

src/contrib/cluster/Akka.DistributedData/ORMultiValueDictionary.cs
Arkatufus: Done

PR #4324, opened 2020-03-11
IActorRef Equals and CompareTo behavior change

### LINKED ISSUE #3781: ActorRefBase.Equals(IActorRef) and ActorRefBase.CompareTo(IActorRef) do not handle null parameters properly
https://github.com/akkadotnet/akka.net/blob/bc5cc65a3dff829cc5844f921826586f828510f5/src/core/Akka/Actor/ActorRef.cs#L347-L361
Both methods do not handle null parameter properly and if the parameter is null, System.NullReferenceException will be thrown on the other.Path property access.

Arkatufus:
Closes #3781

### DIFF
--- a/src/core/Akka/Actor/ActorRef.cs
+++ b/src/core/Akka/Actor/ActorRef.cs
@@ -302,9 +302,10 @@ namespace Akka.Actor
         /// <inheritdoc/>
         public override bool Equals(object obj)
         {
-            var other = obj as IActorRef;
-            if (other == null) return false;
-            return Equals(other);
+            if (obj is IActorRef other)
+                return Equals(other);
+
+            return false;
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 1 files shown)


src/core/Akka/Actor/ActorRef.cs
```diff
@@ -302,9 +302,10 @@ public override string ToString()
         /// <inheritdoc/>
         public override bool Equals(object obj)
         {
-            var other = obj as IActorRef;
-            if (other == null) return false;
-            return Equals(other);
+            if (obj is IActorRef other)
```
Aaronontheweb: LGTM

src/core/Akka/Actor/ActorRef.cs
```diff
@@ -325,21 +326,37 @@ public override int GetHashCode()
...
+        /// <param name="other"></param>
+        /// <returns>
+        /// <c>true</c> if this <see cref="IActorRef"/> instance have the same reference 
+        /// as the <paramref name="other"/> instance, if this <see cref="ActorPath"/> of 
+        /// this <see cref="IActorRef"/> instance is equal to the <paramref name="other"/> instance, 
+        /// and <paramref name="other"/> is not <c>null</c>; otherwise <c>false</c>.
+        /// </returns>
         public bool Equals(IActorRef other)
         {
+            if (other is null) return false;
```
Aaronontheweb: LGTM

Issue #4314, opened 2020-03-09
Akka.Pattern: Pass in clearer error message into OpenCircuitException


Aaronontheweb:
Akka.NET Version: 1.4.1-rc1 and later

When an Akka.Persistence journal begins experiencing failures on the database-side of the operation, users begin seeing lots of `OpenCircuitExceptions` appear inside their logs - these errors don't do much to tell the human operators of Akka.NET clusters and applications what is really going on under the covers and what the root cause of the issue is.

If we could add some method overloads to `CircuitBreaker` to make it possible to provide a hint as to what's going on, that would be very useful for operators.

ismaelhamed:
I've had this conversation so many time. Because only the very first exception will actually give you some insight, and everything after that is just failing fast. 

One possible solution could be to make the `CircuitBreaker` remember the last exception (which made the circuit to break) and set it as the `InnerException` of the `OpenCircuitException`

PR #4339, opened 2020-03-13
Expose success and fail methods in CircuitBreaker

ismaelhamed:
> Current implementation works great when the call is wrapping a future with direct response. However, if the actor is expecting a message back or ReceiveTimeout, then using the circuit breaker is awkward -- there's no way to access the circuit breaker's current state, or signal a failure or success directly.

PR [#21074](https://github.com/akka/akka/pull/21074)

### DIFF
--- a/docs/articles/utilities/circuit-breaker.md
+++ b/docs/articles/utilities/circuit-breaker.md
@@ -74,3 +74,12 @@ dangerousActor.Tell("is my middle name");
 // My CircuitBreaker is now closed
 // This really isn't that dangerous of a call after all
 ```
+
+### Tell Pattern
+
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
Does this relate to https://github.com/akkadotnet/akka.net/issues/4314 at all @ismaelhamed ?

ismaelhamed:
No, this just makes it easier to use the `CircuitBreaker` when the callee is an actor but you can't use the `Ask` pattern, you are not actually working with a `Task`. See [#18347](https://github.com/akka/akka/issues/18347)

This is considered a power-user API, but might come in handy for some people.

PR #4344, opened 2020-03-17
Lease API & integration

zbynek001:
Note: this depends on Timers (#3778), will rebase it when it's merged

- [x] Lease API
- [x] Singleton Integration
- [x] Sharding Integration

Open quesions:
- where to put the api, right now it's directly in Akka project, should it be moved somewhere else?

migrated from https://github.com/akka/akka/pull/26629


zbynek001:
sharding integration would be better to do after Timers API (https://github.com/akkadotnet/akka.net/pull/3778)

Aaronontheweb:
@zbynek001 is #3778 good to go?

zbynek001:
@Aaronontheweb  in theory yes. There is only the open question if the code in `ActorBase.AroundReceive` will have any performance implication

Aaronontheweb:
@zbynek001 ok, got it - I'll review the PR then benchmark it.

Aaronontheweb:
@zbynek001 reviewed #3778 - do you need to address those before this PR gets finished?

Aaronontheweb:
Working on reviewing this now that #3778 has been merged. Might take a bit as I'm familiarizing myself with the API as I go.

Going to need to add documentation for this too if the API is public.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -917,38 +1089,88 @@ private void InitializeFSM()
...
+            When(ClusterSingletonState.AcquiringLease, e =>
             {
-                if (e.FsmEvent is OldestChangedBuffer.OldestChanged oldestChanged && e.StateData is OldestData oldestData)
+                if (e.FsmEvent is AcquireLeaseResult alr)
                 {
-                    _oldestChangedReceived = true;
-                    Log.Info("Oldest observed OldestChanged: [{0} -> {1}]", _cluster.SelfAddress, oldestChanged.Oldest?.Address);
-                    switch (oldestChanged.Oldest)
+                    Log.Info("Acquire lease result {0}", alr.HoldingLease);
+                    if (alr.HoldingLease)
```
Aaronontheweb: So, removing a lot of this business logic from the cluster singleton implementation itself and moving it to the leasing system - ok, that makes sense.

Aaronontheweb: Oh I see - it's been moved to the state rather than the event-handling methods themselves.

src/core/Akka.API.Tests/CoreAPISpec.ApproveClusterSharding.approved.txt
```diff
@@ -45,17 +45,19 @@ namespace Akka.Cluster.Sharding
...
         public readonly System.TimeSpan PassivateIdleEntityAfter;
         public readonly bool RememberEntities;
         public readonly string Role;
         public readonly string SnapshotPluginId;
         public readonly Akka.Cluster.Sharding.StateStoreMode StateStoreMode;
         public readonly Akka.Cluster.Sharding.TunningParameters TunningParameters;
-        public ClusterShardingSettings(string role, bool rememberEntities, string journalPluginId, string snapshotPluginId, System.TimeSpan passivateIdleEntityAfter, Akka.Cluster.Sharding.StateStoreMode stateStoreMode, Akka.Cluster.Sharding.TunningParameters tunningParameters, Akka.Cluster.Tools.Singleton.ClusterSingletonManagerSettings coordinatorSingletonSettings) { }
+        public ClusterShardingSettings(string role, bool rememberEntities, string journalPluginId, string snapshotPluginId, System.TimeSpan passivateIdleEntityAfter, Akka.Cluster.Sharding.StateStoreMode stateStoreMode, Akka.Cluster.Sharding.TunningParameters tunningParameters, Akka.Cluster.Tools.Singleton.ClusterSingletonManagerSettings coordinatorSingletonSettings, Akka.Coordination.LeaseUsageSettings leaseSettings) { }
```
Aaronontheweb: This is a breaking API change - need to add a new method overload instead.

src/core/Akka.API.Tests/CoreAPISpec.ApproveClusterTools.approved.txt
```diff
@@ -342,14 +342,16 @@ namespace Akka.Cluster.Tools.Singleton
     }
     public sealed class ClusterSingletonManagerSettings : Akka.Actor.INoSerializationVerificationNeeded
     {
-        public ClusterSingletonManagerSettings(string singletonName, string role, System.TimeSpan removalMargin, System.TimeSpan handOverRetryInterval) { }
+        public ClusterSingletonManagerSettings(string singletonName, string role, System.TimeSpan removalMargin, System.TimeSpan handOverRetryInterval, Akka.Coordination.LeaseUsageSettings leaseSettings) { }
```
Aaronontheweb: Same issue was with sharding - need to add a method overload

src/core/Akka.API.Tests/CoreAPISpec.ApproveClusterTools.approved.txt
```diff
@@ -373,20 +375,21 @@ namespace Akka.Cluster.Tools.Singleton
         public static Akka.Cluster.Tools.Singleton.ClusterSingletonProxySettings Create(Akka.Configuration.Config config) { }
         public Akka.Cluster.Tools.Singleton.ClusterSingletonProxySettings WithBufferSize(int bufferSize) { }
         public Akka.Cluster.Tools.Singleton.ClusterSingletonProxySettings WithRole(string role) { }
-        public Akka.Cluster.Tools.Singleton.ClusterSingletonProxySettings WithSingletonIdentificationInterval(string singletonIdentificationInterval) { }
+        public Akka.Cluster.Tools.Singleton.ClusterSingletonProxySettings WithSingletonIdentificationInterval(System.TimeSpan singletonIdentificationInterval) { }
         public Akka.Cluster.Tools.Singleton.ClusterSingletonProxySettings WithSingletonName(string singletonName) { }
     }
     public enum ClusterSingletonState
```
Aaronontheweb: If I recall correctly, none of these values are transmitted over the wire - correct?

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2329,6 +2329,71 @@ namespace Akka.Configuration.Hocon
         public char Take() { }
     }
 }
+namespace Akka.Coordination
```
Aaronontheweb: Need to move all of this into its own project - `Akka.Coordination`

Aaronontheweb:
One question I have that wasn't totally clear to me based on what I reviewed - do any of these leasing messages go over the wire? Didn't look that way when I first looked but it might be implementation-specific for things like Akka.Cluster.Tools.Singleton

Aaronontheweb:
@zbynek001 this will be a great addition to Akka.NET - happy to do it once we can address these comments.

src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj
```diff
@@ -12,6 +12,7 @@
     <ProjectReference Include="..\..\..\core\Akka.Persistence\Akka.Persistence.csproj" />
     <ProjectReference Include="..\..\persistence\Akka.Persistence.Sql.Common\Akka.Persistence.Sql.Common.csproj" />
     <ProjectReference Include="..\..\..\core\Akka.Tests.Shared.Internals\Akka.Tests.Shared.Internals.csproj" />
+    <ProjectReference Include="..\Akka.Cluster.Tools.Tests\Akka.Cluster.Tools.Tests.csproj" />
```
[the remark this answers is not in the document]
zbynek001: there are some common classes for the lease testing (TestLease.cs)
It's used in both singleton and sharding tests

src/contrib/cluster/Akka.Cluster.Sharding/DDataShard.cs
```diff
@@ -51,6 +53,12 @@ internal sealed class DDataShard : ActorBase, IShard, IWithUnboundedStash
         public ICancelable PassivateIdleTask { get; }
 
         private EntityRecoveryStrategy RememberedEntitiesRecoveryStrategy { get; }
+
+        public ITimerScheduler Timers { get; set; }
+        public Lease Lease { get; }
+        public TimeSpan LeaseRetryInterval { get; } = TimeSpan.FromSeconds(5); // won't be used
```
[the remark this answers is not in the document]
zbynek001: It's needed there, just the default value of 5 seconds should never get used.
When it should be used, it'll have value from config

src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
```diff
@@ -357,6 +368,39 @@ public sealed class PassivateIdleTick : INoSerializationVerificationNeeded
             private PassivateIdleTick() { }
         }
 
+
+        [Serializable]
+        public sealed class LeaseAcquireResult : IDeadLetterSuppression
```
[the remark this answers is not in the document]
zbynek001: not needed i think, it's all local messages.
Maybe we can also add INoSerializationVerificationNeeded

src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
```diff
@@ -430,12 +497,98 @@ internal static class Shards
...
+        }
+
+        public static void TryGetLease<TShard>(this TShard shard, Lease lease) where TShard : IShard
+        {
+            shard.Log.Info("Acquiring lease {0}", lease.Settings);
+
+            var self = shard.Self;
+            lease.Acquire(reason =>
+            {
+                self.Tell(new Shard.LeaseLost(reason));
```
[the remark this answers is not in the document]
zbynek001: this callback should be called e.g. in following situation:
* we successfully acquire lease initially
* lease is doing internally some renewal of the lease
* because of e.g. network issue the renewal failed
* lease can send notification via the callback, that it was lost

src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
```diff
@@ -509,6 +667,15 @@ internal static class Shards
             }
         }
 
+        //private static void HandleLeaseLost<TShard>(this TShard shard, Shard.LeaseLost message) where TShard : IShard
```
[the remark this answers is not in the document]
zbynek001: yup, some leftover there

src/core/Akka.API.Tests/CoreAPISpec.ApproveClusterTools.approved.txt
zbynek001: yes, don't think it's send over the wire.
It was more like a bug there, because the string value was not used for anything there. Instead it was using the value, which was already there 

```
return Copy(singletonIdentificationInterval: SingletonIdentificationInterval);
```
vs
```
return Copy(singletonIdentificationInterval: singletonIdentificationInterval);
```
'S' vs 's'

can also add a "dummy" overload for that if needed

Aaronontheweb: Man that's a nasty bug. Yeah, add an overload for good measure there.

src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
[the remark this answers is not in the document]
Aaronontheweb: I think this should be fine as-is.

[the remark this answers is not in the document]
Aaronontheweb: But yeah, add the `INoSerializationVerificationNeeded` marker to these as well

src/contrib/cluster/Akka.Cluster.Sharding.Tests/Akka.Cluster.Sharding.Tests.csproj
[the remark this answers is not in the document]
Aaronontheweb: I see.

src/contrib/cluster/Akka.Cluster.Sharding/DDataShard.cs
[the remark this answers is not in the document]
Aaronontheweb: Ah I see. Ok, I misunderstood the purpose of the comment then.

zbynek001:
@Aaronontheweb basically no messages here go over wire, this is just adding API to create your own `Lease` implementation.
And the specific `Lease` implementation would then internally communicate with some external system to create the lease, e.g. redis cache.

zbynek001:
@Aaronontheweb I think all changes should be done. Not sure about the new projects, if there is something else to change in build script or somewhere else

Aaronontheweb:
@zbynek001 I'll check that now.

src/core/Akka.Coordination.Tests/Akka.Coordination.Tests.csproj
```diff
@@ -0,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
Aaronontheweb: Test CSPROJ looks right

src/core/Akka.Coordination/Akka.Coordination.csproj
```diff
@@ -0,0 +1,32 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\common.props" />
+
+  <PropertyGroup>
+    <AssemblyTitle>Akka.Coordination</AssemblyTitle>
+    <Description></Description>
```
Aaronontheweb: Need to add a real description here, but otherwise it looks good

src/core/Akka.Coordination/Lease.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using System.Threading.Tasks;
+
+namespace Akka.Coordination
+{
+    public abstract class Lease
```
Aaronontheweb: Should add some API docs to here

src/core/Akka.Coordination/LeaseException.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+namespace Akka.Coordination
+{
+    public class LeaseException : Exception
```
Aaronontheweb: Needs an API doc

src/core/Akka.Coordination/LeaseProvider.cs
```diff
@@ -0,0 +1,134 @@
+﻿using System;
+using System.Collections.Concurrent;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Event;
+
+namespace Akka.Coordination
+{
+    public class LeaseProviderExtensionProvider : ExtensionIdProvider<LeaseProvider>
```
Aaronontheweb: Same as before - needs API docs and copyright headers

src/core/Akka.Coordination/LeaseSettings.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using Akka.Configuration;
+
+namespace Akka.Coordination
+{
+    public sealed class LeaseSettings
```
Aaronontheweb: Same as above

src/core/Akka.Coordination/LeaseTimeoutException.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+namespace Akka.Coordination
+{
+    public class LeaseTimeoutException : LeaseException
```
Aaronontheweb: Same as above

src/core/Akka.Coordination/LeaseUsageSettings.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace Akka.Coordination
+{
+    public sealed class LeaseUsageSettings
```
Aaronontheweb: Same as above

src/core/Akka.Coordination/TimeoutSettings.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using Akka.Configuration;
+
+namespace Akka.Coordination
+{
+    public sealed class TimeoutSettings
```
Aaronontheweb: Same as above

Aaronontheweb:
Nice work @zbynek001

Lutando:
General question about this, this a timed lease that works similar to Azure Blob leases? ie Only one actor/process can hold a reference to an actor in a bounded time period?

Aaronontheweb:
@Lutando that's correct - you could implement a lease that is backed by Azure blobs, a Redis object, etc...

Issue #4353, opened 2020-03-23
Akka.Remote starts listening on different port than specified


mmisztal1980:
Hi guys,

I've returned to one of my pet projects (lighthouse alternative), after some time off. In the project I have a bootstrapper, that generates HOCON fallbacks, from which the configuration is being built.

```csharp
var config = ConfigurationFactory.Empty;

foreach (var cfg in fallbacks)
{
  config = config.WithFallback(cfg);
}

return ActorSystem.Create("System", config);
```
Upon startup, I'm noticing that:
- The actorsystem only logs the 1st fallback of the configuration rather than the entire HOCON
- Akka REMOTE attempts to start listening on a different port than specified in HOCON, appears to be a default value from `Remote.conf` (2552)
- Akka.Cluster seed nodes' fallback config appears to have been ignored and "No seed-nodes configured" message was logged

HOCON, pseudocode and log outputs located [here](https://gist.github.com/mmisztal1980/4f3c8ac7baf58979e0d525a90d374788)

- Which Akka.Net version you are using: 
  - 1.4.3 (bump from 1.3.15)
- On which platform you are using Akka.Net
  - Windows
  - Linux containers

mmisztal1980:
I've went back in time, and checked out a repro #3943 that @Aaronontheweb was trying out the other day. I've applied the fallback configs and the cluster was self-formed... :/ 

```csharp
public class Bugfix3943Specs
    {
        public static string[] Hocons =
             {
                 "akka : {\r\n actor : {\r\n provider : cluster }}",
                 "akka : {\r\n stdout-loglevel : INFO\r\n loglevel : INFO\r\n log-config-on-start : on\r\n loggers : [\"Akka.Logger.Serilog.SerilogLogger, Akka.Logger.Serilog\"] \r\nactor : {\r\n debug : {\r\n receive : on\r\n autoreceive : on\r\n lifecycle : on\r\n event-stream : on\r\n unhandled : on\r\n } } }",
                 "akka : {\r\n remote : {\r\n dot-netty : {\r\n tcp : {\r\n log-transport : true\r\n transport-class : \"Akka.Remote.Transport.DotNetty.TcpTransport, Akka.Remote\"\r\n transport-protocol : tcp\r\n hostname : 0.0.0.0\r\n public-hostname : localhost\r\n port : 9000\r\n } } } }",
[...]

PR #4358, opened 2020-03-25
#4353 Config.WithFallback is acting inconsistently

Arkatufus:
Possible fix for #4353, there might be other factors involved.

### DIFF
--- a/src/core/Akka/Actor/Settings.cs
+++ b/src/core/Akka/Actor/Settings.cs
@@ -309,7 +309,7 @@ namespace Akka.Actor
         /// <inheritdoc/>
         public override string ToString()
         {
-            return Config.ToString();
+            return Config.Root.ToString();
         }
     }
 }
--- a/src/core/Akka/Configuration/Config.cs
+++ b/src/core/Akka/Configuration/Config.cs
@@ -37,6 +37,7 @@ namespace Akka.Configuration
             if (root.Value == null)
                 throw new ArgumentNullException(nameof(root), "The root value cannot be null.");
 
+            Value = root.Value;
             Root = root.Value;
             Substitutions = root.Substitutions;
         }
@@ -52,6 +53,7 @@ namespace Akka.Configuration
             if (source == null)
                 throw new ArgumentNullException(nameof(source), "The source configuration cannot be null.");
 
+            Value = source.Value;
             Root = source.Root;
             Fallback = fallback;
         }
@@ -68,6 +70,8 @@ namespace Akka.Configuration
         {
             get { return Root == null || Root.IsEmpty; }
         }
+
+        private HoconValue Value { get; set; }
 
         /// <summary>
         /// The root node of this configuration section
@@ -90,6 +94,7 @@ namespace Akka.Configuration
             {
                 Fallback = Fallback != null ? Fallback.Copy(fallback) : fallback,
                 Root = Root,
+                Value = Value,
                 Substitutions = Substitutions
             };
         }
@@ -423,10 +428,10 @@ namespace Akka.Configuration
         /// <returns>A string containing the current configuration.</returns>
         public override string ToString()
         {
-            if (Root == null)
+            if (Value == null)
                 return "";
 
-            return Root.ToString();
+            return Value.ToString();
         }
 
         /// <summary>
--- a/src/core/Akka/Configuration/Config.cs
+++ b/src/core/Akka/Configuration/Config.cs
@@ -467,7 +467,7 @@ namespace Akka.Configuration
                 throw new ArgumentException("Config can not have itself as fallback", nameof(fallback));
             if (fallback == null)
                 return this;
-            var mergedRoot = Root.GetObject().MergeImmutable(fallback.Root.GetObject());
+            var mergedRoot = fallback.Root.GetObject().MergeImmutable(Root.GetObject());
             var newRoot = new HoconValue();
             newRoot.AppendValue(mergedRoot);
             var mergedConfig = Copy(fallback);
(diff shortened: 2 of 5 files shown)


Aaronontheweb:
cc @mmisztal1980

Aaronontheweb:
@Arkatufus what exactly was the issue and how does this fix it? Asking because I'm in the dark about this

Arkatufus:
there was some inconsistency in accessing Config and keys missing during fallback merge due to keys not being copied to the new config object

mmisztal1980:
interesting :)

src/core/Akka/Configuration/Hocon/HoconObject.cs
```diff
@@ -196,6 +196,8 @@ public void Merge(HoconObject other)
                     //if both values are objects, merge them
                     if (thisItem.IsObject() && otherItem.Value.IsObject())
                         thisItem.GetObject().Merge(otherItem.Value.GetObject());
+                    else
```
mmisztal1980: ok, so if I'm reading this right, if both items are objects, they are to be merged, but how is the alternative supposed to work? what is otherItem's type?

Arkatufus: HOCON have 3 basic datatypes, `Object`, `Array`, and `Literal`. 
`Array` and `Literal` will always override merged object fields, but when an `Object` met another `Object`, they got merged instead.

Arkatufus: The full specification for HOCON can be read here: https://github.com/akkadotnet/hocon

mmisztal1980: ah ok, I see it now

mmisztal1980: let's give it a shot then

Aaronontheweb:
@mmisztal1980 I'll go over this and if it's in good shape, merge it and kick it off in a nightly build. You keen to give it a try once that goes through?

src/core/Akka/Actor/Settings.cs
```diff
@@ -309,7 +309,7 @@ private static string GetProviderClass(string provider)
         /// <inheritdoc/>
         public override string ToString()
         {
-            return Config.ToString();
+            return Config.Root.ToString();
```
Aaronontheweb: Is this going to skip fallback values during debugging?

src/core/Akka/Configuration/Config.cs
```diff
@@ -467,7 +458,10 @@ public virtual Config WithFallback(Config fallback)
                 throw new ArgumentException("Config can not have itself as fallback", nameof(fallback));
             if (fallback == null)
                 return this;
-            var mergedRoot = Root.GetObject().MergeImmutable(fallback.Root.GetObject());
+            if (IsEmpty)
```
Aaronontheweb: Makes sense

src/core/Akka/Configuration/Hocon/HoconObject.cs
```diff
@@ -159,7 +159,6 @@ public string ToString(int indent)
             var sb = new StringBuilder();
             foreach (var kvp in Items)
             {
-                if (kvp.Value.AdoptedFromFallback) continue;
```
Aaronontheweb: Was this causing the print out to miss items or something?

src/core/Akka/Configuration/Hocon/HoconObject.cs
```diff
@@ -228,6 +241,8 @@ internal HoconObject MergeImmutable(HoconObject other)
                         mergedValue.AppendValue(mergedObject);
                         thisItems[otherItem.Key] = mergedValue;
                     }
+                    else
+                        thisItems[otherItem.Key] = otherItem.Value;
```
Aaronontheweb: So for non-`HoconObject` types, we just ignored them on merge before?

src/core/Akka/Configuration/Hocon/HoconObject.cs
Arkatufus: Yes, I have no idea why they are there, some items are marked as non printable, which doesn't make any sense, especially since codes depends on string comparison to compare 2 config instances.

src/core/Akka/Configuration/Hocon/HoconObject.cs
Arkatufus: yeah, it causes a lot of missing object fields, I'm surprised that this bug went through the test suite

Aaronontheweb: Is there a simple test case we can add for this to the HOCON test suite? Thinking about it mostly as a regression test.

src/core/Akka/Configuration/Hocon/HoconObject.cs
Aaronontheweb: Understood, thanks

src/core/Akka/Actor/Settings.cs
Arkatufus: No, `Root` is a cumulative merged object, it already includes all of the fallbacks in the `Config`.

src/core/Akka/Configuration/Hocon/HoconObject.cs
Arkatufus: We do have rigorous testing suite in the standalone HOCON project, I think we used to have a limited HOCON test suite in Akka, they might be accidentally removed.

mmisztal1980:
> @mmisztal1980 I'll go over this and if it's in good shape, merge it and kick it off in a nightly build. You keen to give it a try once that goes through?

of course

PR #4359, opened 2020-03-26
Add hocon unit test back

Arkatufus:
Hocon unit test was accidentally removed,


src/core/Akka.Tests/Configuration/HoconTests.cs
```diff
@@ -0,0 +1,921 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="HoconTests.cs" company="Akka.NET Project">
```
Aaronontheweb: Can you change these to 2020?

PR #4371, opened 2020-04-07
Update tutorial-1.md

PiotrJustyna:
Added a standalone, working code example for the `Structure of an IActorRef and Paths of Actors` section. Looking at the documentation trend, it should ideally be made a into a reference to a github-hosted piece of code. Looking at how tutorials are organised at the moment, the structure doesn't really encourage individual contributions - the linked examples are a part of a larger project and a larger group of examples, which are not standalone. To contribute, one would have to build the entire solution and modify the entire `Akka.Docs.Tutorials.csproj` project. What I would advise would encourage community contributions/engagement are more standalone, solution-independent examples that illustrate sections/problems clearer.

### DIFF
--- a/docs/articles/intro/tutorial-1.md
+++ b/docs/articles/intro/tutorial-1.md
@@ -113,8 +113,68 @@ Creating a non-top-level actor is possible from any actor, by invoking
 `Context.ActorOf()` which has the exact same signature as its top-level
 counterpart. This is how it looks like in practice:
 
-[!code-csharp[Main](../../../src/core/Akka.Docs.Tutorials/Tutorial1/ActorHierarchyExperiments.cs?name=print-refs)]
-[!code-csharp[Main](../../../src/core/Akka.Docs.Tutorials/Tutorial1/ActorHierarchyExperiments.cs?name=print-refs2)]
+```code-csharp
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace AkkaSandbox
+{
+  public class Program
+  {
+    static void Main(string[] args)
+    {
+      CreateHostBuilder(args).Build().Run();
+    }
+
+    public static IHostBuilder CreateHostBuilder(string[] args) =>
+    Microsoft.Extensions.Hosting.Host.CreateDefaultBuilder(args)
+    .ConfigureServices((hostContext, services) =>
+    {
+      services.AddHostedService<HelloWorldHostedService>();
+    });
+  }
+
+  public class HelloWorldHostedService : IHostedService
+  {
+    public Task StartAsync(CancellationToken cancellationToken)
+    {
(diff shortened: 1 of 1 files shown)


docs/articles/intro/tutorial-1.md
```diff
@@ -113,8 +113,68 @@ Creating a non-top-level actor is possible from any actor, by invoking
 `Context.ActorOf()` which has the exact same signature as its top-level
 counterpart. This is how it looks like in practice:
 
-[!code-csharp[Main](../../../src/core/Akka.Docs.Tutorials/Tutorial1/ActorHierarchyExperiments.cs?name=print-refs)]
```
Aaronontheweb, line 116: So we need to follow this technique - this allows all of the code we use for examples to be hosted, type checked, and tested at runtime. Can't just go embedding code directly into the docs - that's what leads to a lot of this byterot in the first place.

Stick your sample code here: https://github.com/akkadotnet/akka.net/tree/dev/src/core/Akka.Docs.Tests as a unit test. It'll get run on every build.

PiotrJustyna: @Aaronontheweb Sure, can do it, but how do you see this example getting converted into a unit test? Looks like we’re trying to work around a technique that’s not really working all that well. Maybe it’s a good time to change it?

If you look at my sandbox repo: https://github.com/PiotrJustyna/akkasandbox, you’ll notice that all examples are also independently hosted, type checked and not bound by any solution file. The way they are structured is perfect for people wanting to experiment and to quickly evaluate the library with minimum amount of fuss. Not to mention it would make community contributions easier.

I’m not proposing you follow my pattern specifically, but at the moment the code examples provided in the documentation are incomplete and misleading.

Aaronontheweb: The technique works great when they're modeled as a unit test, because those tests run whenever the code included in the example is modified.

PiotrJustyna: @Aaronontheweb Sure, that makes sense in principle, but still not clear how you see that example getting converted into a unit test. That is not the first thing I would think of doing.

Aaronontheweb: Yeah modeling this exact bit of code as a unit test would be cumbersome, but do you see how embedding code directly inside the `.md` file rather than an C# file is a degradation? We lose compile-time checking and refactoring support on the samples themselves.

PiotrJustyna: @Aaronontheweb Ah yeah, I even mentioned that in the PR description, so not really surprised changes are being requested. Wanted to open a discussion with my PR, though - looks like something has to be done with the examples and with how tricky it is for the community to make changes there.

PiotrJustyna: It’s getting late in my time zone, but I’ll get back to this tomorrow.

Aaronontheweb: Thanks! I appreciate you contributing and discussing this with me - and I hope to see this issue get merged in and maybe some new ones from you too ;)

PR #4372, opened 2020-04-08
Add double wildcard support for deployment

### LINKED ISSUE #4136: Support double wildcards in actor deployment config
Akka supports double wildcards in the actor deployment configuration. This is documented [here](https://doc.akka.io/docs/akka/2.5.4/scala/general/configuration.html#actor-deployment-configuration) (it's missing in the current docs):

> you can use double-wildcards in the last position to match all child actors and their children recursively: `/someParent/**`

The relevant PR for this feature is https://github.com/akka/akka/pull/20525. See https://github.com/akka/akka/blob/6278968af56f36b8ff3fd4a0a443e5218bceb838/akka-actor/src/main/scala/akka/util/WildcardIndex.scala for current version of this file.

It would be great to have this available in Akka.NET too. This will allow simplifying configuration from
```
"/foo/*" {
	dispatcher = bar
}
"/foo/*/*" {
	dispatcher = bar
}
"/foo/*/*/*" {
	dispatcher = bar
}
```
to
```
"/foo/**" {
	dispatcher = bar
}
```

Arkatufus:
Related to #4136
Closes #4136

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -4893,17 +4893,6 @@ namespace Akka.Util
     {
         public static System.Func<System.Func<T>, System.Collections.Generic.IList<T>> Fill<T>(int number) { }
     }
-    public sealed class WildcardIndex<T> : System.IEquatable<Akka.Util.WildcardIndex<T>>
-        where T :  class
-    {
-        public WildcardIndex() { }
-        public bool IsEmpty { get; }
-        public bool Equals(Akka.Util.WildcardIndex<T> other) { }
-        public override bool Equals(object obj) { }
-        public T Find(System.Collections.Generic.IEnumerable<string> elems) { }
-        public override int GetHashCode() { }
-        public Akka.Util.WildcardIndex<T> Insert(System.Collections.Generic.IEnumerable<string> elems, T data) { }
-    }
     public class static WildcardMatch
     {
         public static bool Like(this string text, string pattern, bool caseSensitive = False) { }
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -628,7 +628,6 @@ namespace Akka.Actor
         public Deployer(Akka.Actor.Settings settings) { }
         public Akka.Actor.Deploy Lookup(Akka.Actor.ActorPath path) { }
         public Akka.Actor.Deploy Lookup(System.Collections.Generic.IEnumerable<string> path) { }
-        public Akka.Actor.Deploy Lookup(System.Collections.Generic.IEnumerator<string> path) { }
         public virtual Akka.Actor.Deploy ParseConfig(string key, Akka.Configuration.Config config) { }
         public void SetDeploy(Akka.Actor.Deploy deploy) { }
     }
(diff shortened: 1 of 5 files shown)


Arkatufus:
Still need Akka.API.Tests modification

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -628,7 +628,6 @@ namespace Akka.Actor
         public Deployer(Akka.Actor.Settings settings) { }
         public Akka.Actor.Deploy Lookup(Akka.Actor.ActorPath path) { }
         public Akka.Actor.Deploy Lookup(System.Collections.Generic.IEnumerable<string> path) { }
-        public Akka.Actor.Deploy Lookup(System.Collections.Generic.IEnumerator<string> path) { }
```
Aaronontheweb, line 631: Did this just appear in the API approval twice for some reason?

src/core/Akka/Util/WildcardTree.cs
```diff
@@ -15,8 +15,10 @@ namespace Akka.Util
     /// A searchable nested dictionary, represents a searchable tree structure underneath
     /// </summary>
     /// <typeparam name="T">TBD</typeparam>
-    public sealed class WildcardTree<T> where T:class
```
Aaronontheweb, line 18: Meh.... This is a breaking change but this really never should have been part of the public API

src/core/Akka/Util/WildcardIndex.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            if (obj is null) return false;
+            if (ReferenceEquals(obj, this)) return true;
+            if (!(obj is WildcardIndex<T> other)) return false;
+            return Equals(other);
+        }
+
+        public override int GetHashCode()
+        {
+            var hashCode = -872755323;
```
Aaronontheweb, line 71: needs to be wrapped inside an `unchecked` block

src/core/Akka/Util/WildcardIndex.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Akka.Util
+{
+    public sealed class WildcardIndex<T> : IEquatable<WildcardIndex<T>> where T : class
```
Aaronontheweb, line 8: Shouldn't this be made `internal` too? Any reason why not?

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Arkatufus: no, one of them is for `IEnumerable`, the other is for `IEnumerator`, we don't need the `IEnumerator` one anymore.

src/core/Akka/Util/WildcardIndex.cs
Arkatufus: The original `WildcardTree<T>` was a public sealed class, I was just preserving that, though I don't see why it shouldn't be an internal class, since nothing uses it except for `Deployer`

src/core/Akka/Util/WildcardTree.cs
Arkatufus: The original class from the scala implementation was private, it was never ment to be public in the first place.

src/core/Akka/Util/WildcardIndex.cs
Arkatufus: Done

src/core/Akka/Util/WildcardIndex.cs
Arkatufus: Changed to internal

PR #4377, opened 2020-04-11
Mark Member.AgeOrdering as public

ismaelhamed:
Recently, I needed an AgeComparer to calculate the oldest member of the cluster. `Member` class already has a comparer that could be used, but unlike the `Member.AddressOrdering` public field, the `Member.AgeOrdering` is marked internal. 

I also noticed there's another [AgeComparer](https://github.com/akkadotnet/akka.net/blob/735aafbc7d70948d8f58d770bbbed50908745bc8/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs#L240) inside the `ShardRegion` class that should be safe to remove and replace with the above.


Aaronontheweb:
If you want to take this out of draft state, it looks good to me

PR #4387, opened 2020-04-24
    Added constructor that takes a configuration for PersistenceTestKit

Bogdan-Rotund:
Added constructor for PersistenceTestKit class that takes a configuration. Added this so we can pass in custom configurations similar to what is possible for the TestKit class.


Aaronontheweb:
@Bogdan-Rotund great catch

PR #4393, opened 2020-04-27
Bug fix #4376

### LINKED ISSUE #4376: Akka.Net actors freezing
### Environment
- OS: Windows 10.0.17134.885 (1803/April2018Update/Redstone4)
- CPU: Intel Core i5-8600K CPU 3.60GHz (Coffee Lake), 1 CPU, 6 logical and 6 physical cores
- .Net Core SDK: 3.1.102; .NET Core 3.1.2 (CoreCLR 4.700.20.6602, CoreFX 4.700.20.6702), X64 RyuJIT
- Akka.Net: 1.4.4 or older
- Build configuration: Release x64

### Symptoms 
After several restarts, actors behind a pool router stop working.

### How to reproduce
Either use the code from [the gist](https://gist.github.com/Nk185/1ca82463bb1709728a23b4aee6a50180) or follow the following:
1. Create a simple actor that can either do something (e.g., write to stdout) or throw an exception.
2. Instantiate ActorSystem and spawn the actor from step 1 behind a pool router with, say, 10 routees.
3. In a loop (the lowest number of iterations I tried was 5) force the actor from step 1 to throw an exception.
4. Optionally wait a bit after the loop ends.
5. In a loop with a number of iterations equal to the number of routees try to force the actor from step 1 to do something
6. Observe that not all of the actors are processing the message

**Side note:** you may need to play with a number of thrown exceptions

### Gist overview 
In the Gist I'm trying in different ways to make the `SimpleActor` to throw an exception and then print a text.
Six main scenarios tried:
* an actor behind a round-robin pool router without routees supervision strategy specified (i.e., the default on - to escalate issue), 
* same but with routees supervision strategy "Restart", 
* same but with routees supervision strategy "Resume", 
* same but with routees supervision strategy "Escalate" (actually the same as the first way), 
* custom round-robin router based on an actor with supervision strategy "Escalate"
* single actor without router

For each of the scenarios with a router (so all except the last one), I'm forcing the `SimpleActor` to throw an exception. Then I'm getting all routees and "asking" them to print a text with their address to stdout both with direct `Tell` and indirect via a router.

### Main observations based on Gist code
1. In all cases where routees supervision strategy **is not** "Escalate" I'm able to see that they are printing a text to stdout both through direct and indirect `Tell`s.
2. In cases where routees supervision strategy **is** "Escalate" one or more routees were not printing out any text to stdout (or at least reading the message from their mailbox) neither by direct nor indirect `Tell`s.
3. All actors that got stuck had `MailboxStatus` equal to 4 (`SuspendUnit`) and `SuspendMask` set.
4. The issue reproduces with custom `ReceiveActor`-based router which means that issue is not in (or is not only in) implementation of `RoutedActorCell` and `RouterActor`.
5. To "catch" the issue you need to play with a number of thrown exception - if not enough exceptions thrown or there was a delay between them issue will not reproduce.

### Assumptions
Taking into account that the number of exceptions and their frequency has an effect and each time number of frozen routees is different and that I wasn't able to reproduce the issue with non-Escalate supervision strategy, I can presume that there is a race condition during escalations in `Dispatcher` or in `Mailbox` that prevents removal of `SuspendUnit` from actor's mailbox.

#### Option 1
The easiest way to work this out is to specify a needed supervision strategy at router level like `Props.Create<YourAwesomeActor>().WithRouter(new RoundRobinPool(routeesNr, null, new OneForOneStrategy(Decider.From(Directive.Restart)), Dispatchers.DefaultDispatcherId));` as it leads to same default behaviour - restarts your routee (unless you overrode it in configs).

#### Option 2
If you have to have more complicated logic that based on router parent's state, you have to write your own router actor but be aware that this will lead to a performance impact.

Arkatufus:
Closes #4376


Aaronontheweb:
Definitely looks like there's a bug with the supervision code when the parent actor is a pool router.

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -246,90 +246,66 @@ private void SendAllToDeadLetters(EarliestFirstSystemMessageList messages)
...
+                HandleInvokeFailure(cause);
+            }
+
+            var newState = CalculateState();
+            // As each state accepts a strict subset of another state, it is enough to unstash if we "walk up" the state
+            // chain
+            var todo = newState < currentState ? rest + UnstashAll() : rest;
 
+            if (IsTerminated) SendAllToDeadLetters(todo);
+            else if (!todo.IsEmpty) SysMsgInvokeAll(todo, newState);
```
Aaronontheweb, line 308: IMHO, let's put the `while` loop back in place so we can save on recursion costs - shouldn't be a big deal either way since system messages are pretty rare but I'd feel more comfortable only changing one thing at a time here.

Arkatufus: done

PR #4401, opened 2020-05-01
Bugfix #4400 LWWDictionary Delta serialization

### LINKED ISSUE #4400: Akka.DistributedData.ORDictionary Unable to cast object of type
Which Akka.Net version you are using
1.4.5

.Net 4.7

Akka logs are producing the below error with the latest release of akka, and no ddata is being replicated to any other node in the cluster.

We use akka clustering with multiple nodes and a singleton, singleton makes most of the changes to the ddata.

Akka.Actor.OneForOneStrategy Unable to cast object of type 'Akka.DistributedData.ORDictionary`2[System.ValueTuple`2[System.Guid,System.UInt16],Akka.DistributedData.LWWRegister`1[Ouroboros.IStreamSessionPayloadProgress]]' to type 'Akka.DistributedData.LWWDictionary`2[System.ValueTuple`2[System.Guid,System.UInt16],Ouroboros.IStreamSessionPayloadProgress]'.

Arkatufus:
Closes #4400


Aaronontheweb:
@Arkatufus looks like the reproduction spec is failing - might be racy

Arkatufus:
Yes, it is, added a slight delay to make it work

src/contrib/cluster/Akka.DistributedData.Tests/LWWDictionarySpec.cs
```diff
@@ -131,12 +132,14 @@ public void LWWDictionary_must_be_able_to_work_with_deltas()
...
         /// </summary>
         [Fact]
-        public void Bugfix_4400_LWWDictionary_Deltas_must_merge_other_LWWDictionary()
+        public async Task Bugfix_4400_LWWDictionary_Deltas_must_merge_other_LWWDictionary()
         {
             var m1 = LWWDictionary<string, string>.Empty
                 .SetItem(_node1, "a", "A")
                 .SetItem(_node1, "b", "B1");
 
+            await Task.Delay(200);
```
Aaronontheweb, line 141: There isn't anything `async` happening here though - why would this work?

Arkatufus: I just added a slight delay between setting the "b" item so the second set always have a higher timestamp compared to the first step.
I don't know how, but in net461, the second set can have an earlier timestamp if you leave out the delay.

Aaronontheweb: Ah I see

PR #4409, opened 2020-05-12
fix: improve doc for persistenceId for sharded entity actor

### LINKED ISSUE #4406: Is the suggested PersistenceId in Cluster Sharding Documentation right?
In Akka Cluster Sharding documentation for persistent actors
https://getakka.net/articles/clustering/cluster-sharding.html#integrating-cluster-sharding-with-persistent-actors

It says each actor path will follow the `/system/{typeName}/{shardId}/{entityId}` pattern
It then suggests that the PersistenceId could be created from the actor path as

`PersistenceId = Context.Parent.Path.Name + "-" + Self.Path.Name;`

which will use `shardId` and `entityId` as the persistent id.

I think this is an error and it should be

`PersistenceId = Context.Parent.Path.Parent.Name + "-" + Context.Self.Path.Name;`

which will use `typeName` and `entityId` as the persistent id (which will be unchanged if you ever need to change the `maxNumberOfShards`).

If I'm right please point me at the doc in source control and I'll raise a PR.

chrisbarker:
Closes #4406

### DIFF
--- a/docs/articles/clustering/cluster-sharding.md
+++ b/docs/articles/clustering/cluster-sharding.md
@@ -132,7 +132,7 @@ This can be resolved by getting information about shard/entity ids directly from
 >
 > Therefore, this design gives the sharding system a chance to hand over all of the sharded entity actors running on the terminating node over to the other remaining nodes in the cluster.
 
-Given these values we can build consistent, unique `PersistenceId`s on the fly like on the following example:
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Thanks for your contribution @chrisbarker !

Aaronontheweb:
And for being our 200th contributor!

chrisbarker:
Woop!

PR #4422, opened 2020-05-20
enable loggers to start asynchronously

### LINKED ISSUE #4054: ActorSystem.Create() blocking indefinitely. [Current status of Akka.NET .NET Standard Support]
I am using Akka.Net Version 1.3.16 on .NET Standard 2.0 and have tested with multiple other versions of Akka.NET.

Steps to reproduce problem.

1. Create new Xamarin forms project using Visual Studio 2019
2. Add the following code on application start.

```
protected override void OnStart()
{
    using (var system = ActorSystem.Create("client"))
    {
    }
}

```

3. Run application.
4. Note that the line using (var system = ActorSystem.Create("client")) is entered into and blocks indefinitely.

I have checked to ensure:

1. Internet permission is enabled.
2. Akka and Akka.Remote is referenced from both the core and android project.

 What else could be wrong here?

Aaronontheweb:
close #4054

This is an experiment to see if we can avoid thread starvation issues starting up the logging system on machines with low core counts.

### DIFF
--- a/src/core/Akka/Event/LoggingBus.cs
+++ b/src/core/Akka/Event/LoggingBus.cs
@@ -85,7 +85,7 @@ namespace Akka.Event
         /// <exception cref="LoggerInitializationException">
         /// This exception is thrown if the logger doesn't respond with a <see cref="LoggerInitialized"/> message when initialized.
         /// </exception>
-        internal void StartDefaultLoggers(ActorSystemImpl system)
+        internal async void StartDefaultLoggers(ActorSystemImpl system)
         {
             var logName = SimpleName(this) + "(" + system.Name + ")";
             var logLevel = Logging.LogLevelFor(system.Settings.LogLevel);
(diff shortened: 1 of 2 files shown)


src/core/Akka/Event/LoggingBus.cs
```diff
@@ -85,7 +85,7 @@ protected override Type GetClassifier(object @event)
         /// <exception cref="LoggerInitializationException">
         /// This exception is thrown if the logger doesn't respond with a <see cref="LoggerInitialized"/> message when initialized.
         /// </exception>
-        internal void StartDefaultLoggers(ActorSystemImpl system)
+        internal async void StartDefaultLoggers(ActorSystemImpl system)
```
Aaronontheweb, line 88: Had to use an `async void` here to stop `Task` contamination from reaching the top of the API. Don't need `Task` error handling here either.

Open to suggestions on how to do this better if needed.

Aaronontheweb:
Looks like these changes hung the test suite

PR #4424, opened 2020-05-21
Make loggers to not block ActorSystem creation

IgorFedchenko:
Continuation of #4422 , trying to make loggers to start without blocking and not blow up the test suite

PR #4430, opened 2020-05-22
Added WindowsForms sample

IgorFedchenko:
During work on #4054 wanted to verify that `ActorSystem` creation goes well, and `Ask` is not hanging when invoked from UI thread with single-threaded synchronization context.


IgorFedchenko:
@Aaronontheweb Looks like system starts well even without `logger-async-start` setting from #4424. Also, `Ask` call can be awaited without issues - I am getting message box on form loading.

So, two conclusions:
1. When we have serious thread starvation on the machine (like running big test suite), and lots of `ActorSystem`s are started in the same time, we may have loggers initialization taking really long time, and that's why `logger-async-start` was introduced in another PR - just to speed-up ActorSystem creation without loggers started.

2. The case when `ActorSystem.Create` blocks indefinitely (because loggers `Ask` call blocks indefinitely) seems to be related to some Xamarin-specific issues. Or probably some other configuration stuff not related to Akka.NET.

Not sure if we need this WindowsForms sample to be included to the main repo, but just wanted to share this sample with conclusions.

IgorFedchenko:
I will close this since this does not show any issues we need to reproduce.

PR #4438, opened 2020-05-26
Fix/4434 idle cpu - removing unused #if compile time conditionals

### LINKED ISSUE #4434: Akka v1.4 Idle CPU usage increased comparing v1.3
**Info**:
.Net Core 3.1
Akka 1.3.18 and Akka 1.4.6    
Platforms: 
* Docker for Windows (Alpine images). Docker version: 19.03.8   
* AWS EKS   

**Issue**

After migration from 1.3.18 to 1.4.6 we've noticed that usage of Akka clustering added more CPU overhead to our deployments. After small investigation and comparison, we've found out that in newer versions Akka brings significantly more CPU usage in Idle state of the system. 
Docker For Windows relative usage values:   
akka-stats

AWS EKS absolute values (m5.large instance):
image

As you can see almost 4x in absolute values in EKS.

**Repro**:
Use code from [repository](https://github.com/mralexes/Akka.Net-versions-issue), run via docker-compose, and check docker stats.
Repro contains three seed nodes with next code:
```
class Program
    {
        private static readonly string Config = @"
        {
            ""akka"": {
              ""loglevel"": ""Warning"",
              ""actor"": {
                ""provider"": ""cluster"",
              },
               ""remote.dot-netty.tcp.batching.enabled"":""false"" 
            }
          }
        ";


        static async Task Main(string[] args)
        {
            var bootstrapFromDocker = DockerBootstrap.BootstrapFromDocker(Config);

            using (var actorSystem = ActorSystem.Create(Environment.GetEnvironmentVariable("ACTOR_SYSTEM"), bootstrapFromDocker))
            {
                await actorSystem.WhenTerminated;
            }
        }
    }
```

Aaronontheweb:
close #4434

Testing a theory for idle CPU performance that ultimately hasn't borne any fruit yet - that we have unresolved clutter as a result of cleaning up the .NET Framework / .NET Core `#if` conditionals

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,17 @@
-#### 1.4.7 May 12 2020 ####
+#### 1.4.8 May 26 2020 ####
 **Placeholder for nightlies**
+
+#### 1.4.7 May 26 2020 ####
+**Maintenance Release for Akka.NET 1.4**
+
+Akka.NET v1.4.7 is a fairly sizable path that introduces some significant new features and changes:
+
+* [Akka: added `akka.logger-async-start` HOCON setting to allow loggers to start without blocking `ActorSystem` creation](https://github.com/akkadotnet/akka.net/pull/4424) - useful for developers running Akka.NET on Xamarin or in other environments with low CPU counts;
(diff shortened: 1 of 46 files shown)


Aaronontheweb:
This isn't the full fix for #4434 but we ought to merge it in anyway since we're no longer using these `#ifdef` directives.

PR #4442, opened 2020-05-27
Move LMDB into release version

Arkatufus:
- Remove `Akka.DistributedData.LightningDB` beta status
- Add reference to `Akka.DistributedData.LightningDB` in `Akka.Cluster.Sharding` for easier end user implementation.


Arkatufus:
Any user who needs to access LMDB from outside of `Akka.Cluster.Sharding` will still need to reference it manually, we can't include LMDB inside `Akka.DistributedData` unless we move the whole code into it (circular reference issue).
This is actually the case in scala Akka, LMDB is part of the `Akka.DistributedData` package, instead of its own standalone package.

Aaronontheweb:
> Any user who needs to access LMDB from outside of `Akka.Cluster.Sharding` will still need to reference it manually, we can't include LMDB inside `Akka.DistributedData` unless we move the whole code into it (circular reference issue).
> This is actually the case in scala Akka, LMDB is part of the `Akka.DistributedData` package, instead of its own standalone package.

Should we just do that in a new release in the future? Pretty easy to deprecate packages on NuGet now.

Arkatufus:
It should be trivial to do, `Akka.DistributedData.LightningDB` only contains 1 class file and 1 HOCON configuration file

Issue #4436, opened 2020-05-26
Implement Akka.Remote.Artery transport


Arkatufus:
This is a very large implementation project, it will be split into separate port/update projects, and each project has to work/compatible with previous Akka.NET implementation.
We need to make sure that all projects still work with current test implementation with no/very little modification.

Current work list:
- [ ] Port Akka.Remote.Artery:
  - [ ] Port/upgrade Akka.Streams.Attributes
    - [ ] New Attributes internal list is reversed from previous versions
  - [ ] Port/upgrade Akka.Streams.ActorMaterializer

Aaronontheweb:
Based on our conversation in chat, not doing Aeron ports - correct? @Arkatufus

Arkatufus:
Correct, since there are no such thing as a proper port of Aeron in C#. Unless we want to port that over too. not.

Arkatufus:
An issue raised by @to11mtm, will have to keep an eye on TCS creation to avoid deadlocks.
https://devblogs.microsoft.com/premier-developer/the-danger-of-taskcompletionsourcet-class/

Aaronontheweb:
@Arkatufus adding an issue for this to the current project.

Here's what I really want out of this:

1. Take all of the items on your checklist, convert them into individual Github issues.
2. For each Github issue, link to the other "critical dependency" issues that have to be completed first. I.E. if we have to add `CompressionTable` before can we add `CompressionProtocol`, we should document that in the `CompressionProtocol` github issue and link to the `CompressionTable` implementation issue.
3. Each issue should also link to a bit of documentation describing how to contribute towards Artery - i.e. which branch to use, porting notes from Scala, etc. We don't want this stuff going into `dev` just yet.
4. Arrange all of this into a project on Github, so we can have visibility into the progress of how far along the port is.

Think you can do this next week or this week?

Arkatufus:
# Current port status

# ArteryTransport.scala
One of the main class of artery, this file touches almost everything in the akka.remote.artery 
namespace. Some of the Ported Content have already been ported (marked with a check), while others are 
still in the stub method form.

Note that while there is an _Association.scala_ file, the `AssociationState` class belongs to
this file.
## Dependency 
### Classes
#### Artery
- akka.remote.artery.ArterySettings
- akka.remote.artery.Association
- akka.remote.artery.AssociationRegistry --> akka/remote/artery/Association.scala
- akka.remote.artery.Decoder --> akka/remote/artery/Codecs.scala
[...]

PR #4443, opened 2020-05-27
ARTERY: Port `Akka.Remote.Artery.ArterySettings`

Arkatufus:
Related to #4436


Aaronontheweb:
I'm fine merging this type of stuff into the v1.4 branch since it's all `internal`. If it doesn't show up on the public API I don't think it's an issue to lay the groundwork in `dev`. 

Otherwise, we'd need to open this up onto the v1.5 branch.

Arkatufus:
I'm putting these `Artery` work into PRs just for review right now, just in case I did something wrong, and for progress documentation purposes. 
There is no need to merge it right now, I wouldn't want to risk unforeseen consequences.

PR #4453, opened 2020-06-05
Akka.Remote.Artery port - Base code changes

Arkatufus:
Needed changes in the base code (outside of `Akka.Remote.Artery`) needed by `Artery` to make `Artery` port work.


Aaronontheweb:
@Arkatufus do we need to create a `1.5` branch to push these to?

Arkatufus:
I think so, just to be safe, though I've made sure that only the essentials are included here, and none of the public APIs are changed

Aaronontheweb:
Did you mean to delete this?

PR #4455, opened 2020-06-07
Backport of the feature called ClusterDistribution in Lagom

ismaelhamed:
> This is a feature backported from Lagom which essentially works like a sharded singleton. A number of actor instances are kept alive and balanced across the cluster.
The main motivation to bring it into Akka is the use case where workers that update projections. By tagging akka-persistence events with N different tags and then running N workers, one for each tag which consumes that partition of the events and updates a projection.

This feature was implemented in Akka typed, not yet available in Akka.NET. But given that users will either borrow Lagom's implementation anyway or roll their own, IMO it's worthwhile having this on classical Akka.

- [X] Implementation
- [X] Unit tests
- [X] Multinode tests

### DIFF
--- a/docs/articles/clustering/cluster-sharded-daemon-process.md
+++ b/docs/articles/clustering/cluster-sharded-daemon-process.md
@@ -1,0 +1,31 @@
+# Sharded Daemon Process
+
+> [!WARNING]
+>This module is currently marked as [may change](../utilities/may-change.md) because it is a new feature that
+>needs feedback from real usage before finalizing the API. This means that API or semantics can change without
+>warning or deprecation period. It is also not recommended to use this module in production just yet.
+
+## Introduction
+
+Sharded Daemon Process provides a way to run `N` actors, each given a numeric id starting from `0` that are then kept alive
+and balanced across the cluster. When a rebalance is needed the actor is stopped and, triggered by a keep alive running on 
+all nodes, started on a new node (the keep alive should be seen as an implementation detail and may change in future versions).
+
+The intended use case is for splitting data processing workloads across a set number of workers that each get to work on a subset
+of the data that needs to be processed. This is commonly needed to create projections based on the event streams available
+from all the [Persistent Actors](../persistence/event-sourcing.md) in a CQRS application. Events are tagged with one out of `N` tags
+used to split the workload of consuming and updating a projection between `N` workers.
+
+For cases where a single actor needs to be kept alive see [Cluster Singleton](cluster-singleton.md)
+
+## Basic example
+
+To set up a set of actors running with Sharded Daemon process each node in the cluster needs to run the same initialization
+when starting up:
+
+[!code-csharp[ShardedDaemonProcessSpec.cs](../../../src/contrib/cluster/Akka.Cluster.Sharding.Tests/ShardedDaemonProcessSpec.cs?name=tag-processing)]
+
+## Scalability  
+
+This cluster tool is intended for small numbers of consumers and will not scale well to a large set. In large clusters 
+it is recommended to limit the nodes the sharded daemon process will run on using a role.
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ShardedDaemonProcessSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ShardedDaemonProcessSpec.cs
@@ -1,0 +1,174 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ClusterShardingSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2020 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2020 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Linq;
+using Akka.Actor;
+using Akka.Cluster.TestKit;
+using Akka.Cluster.Tools.Singleton;
+using Akka.Configuration;
+using Akka.Remote.TestKit;
+using FluentAssertions;
+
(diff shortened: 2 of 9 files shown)


Aaronontheweb:
@ismaelhamed we should add some documentation for this feature too - honestly that'll help me review it. We were just looking at `akka-projections` in the Gitter chat room the other day too. Seems like a useful feature in the long run.

ismaelhamed:
[This is](https://doc.akka.io/docs/akka/current/typed/cluster-sharded-daemon-process.html) pretty much the only documentation you'll find, and is already included in this PR. But because this feature was borrowed from Lagom, you could also read [its documentation](https://www.lagomframework.com/documentation/1.6.x/java/ReadSide.html#Event-tags) for a little more of background (though this implementation is way more straightforward).

Actually, [the recommended way](https://doc.akka.io/docs/akka-projection/current/running.html) to distribute the projections in `akka-projections` is by leveraging this new ShardedDaemonProcess.

src/contrib/cluster/Akka.Cluster.Sharding/reference.conf
```diff
@@ -268,6 +268,20 @@ akka.cluster.sharding {
 }
 # //#sharding-ext-config
 
+akka.cluster.sharded-daemon-process {
+  # Settings for the sharded dameon process internal usage of sharding are using the akka.cluste.sharding defaults.
+  # Some of the settings can be overriden specifically for the sharded daemon process here. For example can the
+  # `role` setting limit what nodes the daemon processes and the keep alive pingers will run on.
+  # Some settings can not be changed (remember-entitites and related settings, passivation, number-of-shards),
+  # overriding those settings will be ignored.
+  sharding = ${akka.cluster.sharding}
```
ismaelhamed, line 277: This seems to work, but `akka.cluster.sharding.coordinator-singleton = ${akka.cluster.singleton}` does not. I guess the current hocon implementation cannot load a config section from another file!?

Aaronontheweb: I don't think so - cc @Arkatufus

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ShardedDaemonProcessSpec.cs
```diff
@@ -0,0 +1,174 @@
...
+        [MultiNodeFact]
+        public void ShardedDaemonProcess_Specs()
+        {
+            ShardedDaemonProcess_Should_Init_Actor_Set();
+        }
+
+        public void ShardedDaemonProcess_Should_Init_Actor_Set()
+        {
+            // HACK
+            RunOn(() => FormCluster(_config.First, _config.Second, _config.Third), _config.First);
```
ismaelhamed, line 74: Test does not pass unless I put `FormCluster()` inside a `RunOn()`. I'm missing something.

Aaronontheweb: I'll take a look during review

src/core/Akka.API.Tests/CoreAPISpec.ApproveClusterSharding.approved.txt
```diff
@@ -190,7 +190,7 @@ namespace Akka.Cluster.Sharding
             public override bool Equals(object obj) { }
             public override int GetHashCode() { }
         }
-        public sealed class StartEntityAck : Akka.Cluster.Sharding.IClusterShardingSerializable
+        public sealed class StartEntityAck : Akka.Cluster.Sharding.IClusterShardingSerializable, Akka.Event.IDeadLetterSuppression
```
Aaronontheweb, line 193: LGTM

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ShardedDaemonProcessSpec.cs
Aaronontheweb: Not sure, off the top of my head, what the issue might be here - what type of error do you get?

docs/articles/clustering/cluster-sharded-daemon-process.md
```diff
@@ -0,0 +1,51 @@
...
+used to split the workload of consuming and updating a projection between `N` workers.
+
+For cases where a single actor needs to be kept alive see [Cluster Singleton](cluster-singleton.md)
+
+## Basic example
+
+To set up a set of actors running with Sharded Daemon process each node in the cluster needs to run the same initialization
+when starting up:
+
+```csharp
```
Aaronontheweb, line 26: Use a DocFx snippet here: https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html#code-snippet

docs/articles/clustering/cluster-sharded-daemon-process.md
```diff
@@ -0,0 +1,51 @@
+# Sharded Daemon Process
```
Aaronontheweb, line 1: Need to reference this article from https://github.com/akkadotnet/akka.net/blob/dev/docs/articles/toc.yml in order for it to appear on the website

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ShardedDaemonProcessSpec.cs
ismaelhamed: I only get one `ProcessActorEvent` instead of the 4 expected (one per `ProcessActor`), so the test fails with timeout.

Aaronontheweb:
Is there a reason why I can't update this PR myself? I don't see an "Update Branch" button :\

ismaelhamed:
"Allow edits by maintainers" check was disabled. Is this option disabled now by default?

Aaronontheweb:
@ismaelhamed maybe - I'll need to check!

PR #4459, opened 2020-06-08
Update Akka.NET source code to support Akka.Remote.Artery port.

Arkatufus:
Several changes needed in the base Akka.NET source code to support Akka.Remote.Artery port.

### DIFF
--- a/src/core/Akka.Remote/UniqueAddress.cs
+++ b/src/core/Akka.Remote/UniqueAddress.cs
@@ -1,0 +1,118 @@
+﻿using System;
+using System.Collections.Generic;
+using Akka.Actor;
+
+namespace Akka.Remote
+{
+    // ARTERY: Document that UniqueAddress is a copy of the same class from Akka.Cluster.Member
+    // ARTERY: Temporarily use internal for now, so we don't break API approval list
+    /// <summary>
+    /// Member identifier consisting of address and random `uid`.
+    /// The `uid` is needed to be able to distinguish different
+    /// incarnations of a member with same hostname and port.
+    /// </summary>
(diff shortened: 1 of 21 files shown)


src/core/Akka.Remote/UniqueAddress.cs
```diff
@@ -0,0 +1,118 @@
...
+namespace Akka.Remote
+{
+    // ARTERY: Document that UniqueAddress is a copy of the same class from Akka.Cluster.Member
+    // ARTERY: Temporarily use internal for now, so we don't break API approval list
+    /// <summary>
+    /// Member identifier consisting of address and random `uid`.
+    /// The `uid` is needed to be able to distinguish different
+    /// incarnations of a member with same hostname and port.
+    /// </summary>
+    internal class UniqueAddress : IComparable<UniqueAddress>, IEquatable<UniqueAddress>, IComparable
```
Arkatufus, line 14: This code duplication is the same as in the scala code, UniqueAddress code is duplicated in Akka.Cluster and Akka.Remote

src/core/Akka.Streams/Properties/AssemblyInfo.cs
```diff
@@ -15,6 +15,7 @@
 [assembly: InternalsVisibleTo("Akka.Streams.Tests")]
 [assembly: InternalsVisibleTo("Akka.Streams.TestKit")]
 [assembly: InternalsVisibleTo("Akka.Benchmarks")]
+[assembly: InternalsVisibleTo("Akka.Remote")]
```
Arkatufus, line 18: Akka.Stream internals access are needed in Akka.Remote.Artery

PR #4460, opened 2020-06-08
Update Akka.NET source code to support Akka.Remote.Artery port.

Arkatufus:
Related to #4436
Several changes needed in the base Akka.NET source code to support Akka.Remote.Artery port.

### DIFF
--- a/src/core/Akka.Remote/Akka.Remote.csproj
+++ b/src/core/Akka.Remote/Akka.Remote.csproj
@@ -9,6 +9,7 @@
   </PropertyGroup>
   <ItemGroup>
     <EmbeddedResource Include="Configuration\Remote.conf" />
+    <ProjectReference Include="..\Akka.Streams\Akka.Streams.csproj" />
     <ProjectReference Include="..\Akka\Akka.csproj" />
   </ItemGroup>
   <ItemGroup>
--- a/src/core/Akka.Remote/UniqueAddress.cs
+++ b/src/core/Akka.Remote/UniqueAddress.cs
@@ -1,0 +1,118 @@
+﻿using System;
+using System.Collections.Generic;
+using Akka.Actor;
+
+namespace Akka.Remote
+{
+    // ARTERY: Document that UniqueAddress is a copy of the same class from Akka.Cluster.Member
+    // ARTERY: Temporarily use internal for now, so we don't break API approval list
+    /// <summary>
+    /// Member identifier consisting of address and random `uid`.
+    /// The `uid` is needed to be able to distinguish different
+    /// incarnations of a member with same hostname and port.
+    /// </summary>
+    internal class UniqueAddress : IComparable<UniqueAddress>, IEquatable<UniqueAddress>, IComparable
+    {
+        /// <summary>
+        /// The bound listening address for Akka.Remote.
+        /// </summary>
+        public Address Address { get; }
+
+        /// <summary>
+        /// A random long integer used to signal the incarnation of this cluster instance.
+        /// </summary>
+        public int Uid { get; }
+
+        /// <summary>
+        /// Creates a new unique address instance.
+        /// </summary>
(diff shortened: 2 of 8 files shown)


src/core/Akka.Remote/UniqueAddress.cs
```diff
@@ -0,0 +1,118 @@
...
+namespace Akka.Remote
+{
+    // ARTERY: Document that UniqueAddress is a copy of the same class from Akka.Cluster.Member
+    // ARTERY: Temporarily use internal for now, so we don't break API approval list
+    /// <summary>
+    /// Member identifier consisting of address and random `uid`.
+    /// The `uid` is needed to be able to distinguish different
+    /// incarnations of a member with same hostname and port.
+    /// </summary>
+    internal class UniqueAddress : IComparable<UniqueAddress>, IEquatable<UniqueAddress>, IComparable
```
Arkatufus, line 14: This code duplication is the same as in the scala code, UniqueAddress code is duplicated in Akka.Cluster and Akka.Remote

src/core/Akka.Remote/Akka.Remote.csproj
```diff
@@ -9,6 +9,7 @@
   </PropertyGroup>
   <ItemGroup>
     <EmbeddedResource Include="Configuration\Remote.conf" />
+    <ProjectReference Include="..\Akka.Streams\Akka.Streams.csproj" />
```
Arkatufus, line 12: Akka.Remote.Artery is coupled to Akka.Streams

src/core/Akka/Util/Helpers.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Akka.Util
+{
+    public static class Helpers
+    {
+        public static T Requiring<T>(this T obj, Func<T, bool> cond, string msg)
```
Arkatufus, line 9: This is a helper fluent build pattern function to do an "assert" in fluent manner.
Usage: `myConfig.GetInt("hocon.path.value").Requiring(i => i > 0, "value must be greater than zero.")`

src/core/Akka.Remote/Akka.Remote.csproj
Aaronontheweb: Makes sense

src/core/Akka.Remote/UniqueAddress.cs
Aaronontheweb: Got it - so we'll eventually move this type from Akka.Cluster to Akka.Remote? Is that the idea?

src/core/Akka/Util/Helpers.cs
Aaronontheweb: This is meant to be used in the main Akka.NET code base too then I take it? I.e. not just for testing?

src/core/Akka/Util/WildcardIndex.cs
```diff
@@ -7,21 +7,21 @@ namespace Akka.Util
 {
     internal sealed class WildcardIndex<T> : IEquatable<WildcardIndex<T>> where T : class
     {
-        private readonly WildcardTree<T> _wildcardTree;
-        private readonly WildcardTree<T> _doubleWildcardTree;
+        public WildcardTree<T> WildcardTree { get; }
```
Aaronontheweb, line 10: Is this used for something inside Artery?

Arkatufus: yeah, they're accessed directly from one of the classes. I forgot which one...

Aaronontheweb: Got it.

Aaronontheweb:
@Arkatufus CI didn't run, but the changes LGTM

PR #4462, opened 2020-06-08
Port Akka.Remote.Artery settings classes

Arkatufus:
Related to #4436 
Implementation of ArterySettings and all of its support classes.


Arkatufus:
I'm making that now.

PR #4472, opened 2020-06-15
Changed Akka.Streams backoff error logging to include full exception message

Horusiath:
Sometimes exceptions thrown from inside of a stage don't have any stack traces. Other times the actual meaningful content lies within error message itself. This PR changes the logging behavior of Backoff akka.streams stage to log exception using its default `ToString()` method, which includes both error message and stack trace.

### DIFF
--- a/src/core/Akka.Streams/Dsl/Restart.cs
+++ b/src/core/Akka.Streams/Dsl/Restart.cs
@@ -494,7 +494,7 @@ namespace Akka.Streams.Dsl
                         Fail(Out, ex);
                     else
                     {
-                        Log.Warning($"Restarting graph due to failure. Stacktrace: {ex.StackTrace}");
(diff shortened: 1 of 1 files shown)


src/core/Akka.Streams/Dsl/Restart.cs
```diff
@@ -494,7 +494,7 @@ protected SubSinkInlet<TOut> CreateSubInlet(Outlet<TOut> outlet)
                         Fail(Out, ex);
                     else
                     {
-                        Log.Warning($"Restarting graph due to failure. Stacktrace: {ex.StackTrace}");
+                        Log.Warning("Restarting graph due to failure: {0}", ex);
```
Aaronontheweb, line 497: The `Warning` method can now take the `Exception` as an input parameter. Please change to use that method.

PR #4473, opened 2020-06-15
added spec to validate programmatic ActorSystem override

Aaronontheweb:
Minor PR to address a gap in testing - needed to validate that `ActorSystemSetup` can successfully override the `ActorRefProvider` under the covers.

Added the spec to Akka.Cluster.Tests so we could override using the `ClusterActorRefProvider`.


src/core/Akka/Actor/Settings.cs
```diff
@@ -80,7 +80,7 @@ public Settings(ActorSystem system, Config config, ActorSystemSetup setup)
 
             var providerSelectionSetup = Setup.Get<BootstrapSetup>()
                 .FlatSelect(_ => _.ActorRefProvider)
-                .Select(_ => _.Identifier)
+                .Select(_ => _.Fqn)
```
Aaronontheweb, line 83: Fixed a bug here - caused us not to load any `ProviderSelection.Custom` instances correctly

src/core/Akka/Actor/ActorSystem.cs
```diff
@@ -33,12 +33,13 @@ private ProviderSelection(string identifier, string fqn, bool hasCluster)
         internal string Fqn { get; }
         internal bool HasCluster { get; }
 
+        internal const string LocalActorRefProvider = "Akka.Actor.LocalActorRefProvider, Akka";
         internal const string RemoteActorRefProvider = "Akka.Remote.RemoteActorRefProvider, Akka.Remote";
         internal const string ClusterActorRefProvider = "Akka.Cluster.ClusterActorRefProvider, Akka.Cluster";
 
         public sealed class Local : ProviderSelection
         {
-            private Local() : base("local", typeof(LocalActorRefProvider).FullName, false)
```
Aaronontheweb, line 41: Fixed the FQN here to match the expected Akka.NET format

Issue #4485, opened 2020-06-19
Akka.Persistence.Sql.Common can not subscribe to persistence id twice


kpko:
I'm having the following issue with Akka.net version 1.4.7 on Win10. I'm using the builtin Sqlite persistence module and I get an error when trying to use persistency query to query by the same persistence id from two different places.

```csharp
var readJournal = PersistenceQuery.Get(Context.System).ReadJournalFor<SqlReadJournal>(SqlReadJournal.Identifier);
readJournal.EventsByPersistenceId("flowManager", _currentOffset, long.MaxValue)...
```

Error:

![image](https://user-images.githubusercontent.com/835010/85142622-f5d52f00-b248-11ea-8eb1-5f4e8339647b.png)

As a first glance this might be due to a bug in SqlJournal.AddPersistenceIdSubscriber at https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs#L381. When a dictionary entry of a persistence id already exists, it gets added to the dictionary again which causes the error above. Am I wrong or should the line use _persistenceIdSubscribers.SetItem?

Thanks!

kpko:
Fixed through #4486

PR #4486, opened 2020-06-19
SqlJournal: Fixed updating the _persistenceIdSubscribers dictionary when adding a subscriber

kpko:
Fix for issue #4485

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
@@ -378,7 +378,7 @@ namespace Akka.Persistence.Sql.Common.Journal
             }
             else
             {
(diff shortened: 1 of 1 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
```diff
@@ -378,7 +378,7 @@ public void AddPersistenceIdSubscriber(IActorRef subscriber, string persistenceI
             }
             else
             {
-                _persistenceIdSubscribers = _persistenceIdSubscribers.Add(persistenceId, subscriptions.Add(subscriber));
+                _persistenceIdSubscribers = _persistenceIdSubscribers.SetItem(persistenceId, subscriptions.Add(subscriber));
```
Arkatufus, line 381: Good catch 👍

Arkatufus:
Looks good, I'm merging this in.

PR #4488, opened 2020-06-20
Cluster.Join fails when called asynchronically when SerializeAllMessages is set

### LINKED ISSUE #4481: ClusterUpEvent seems to have issues with Serialiazation
- Which Akka.Net version you are using: 1.4.8
- On which platform you are using Akka.Net: Win10

We are using the following implementation to check when a node is up and running within an Akka cluster

```
  protected void WaitUntilClusterIsUp() {
            var cluster = Akka.Cluster.Cluster.Get(ActorSystemInfo.ActorSystem);
            cluster.RegisterOnMemberUp(ClusterMemberAdded);

            _clusterUpEvent.WaitOne();
            _clusterUpEvent.Reset();
        }
       
        private void ClusterMemberAdded() {
            _clusterUpEvent.Set();
        }
```

This works fine.

In our integration tests we want to set the following option in the akkaconfig, to check that all messages are serializable:

`actor
    {
        provider=cluster
        serialize-messages = on
}`

If the serialize-message option is on, the call in the above method of    _clusterUpEvent.WaitOne(); does not return.

### LINKED ISSUE #4484: Cluster.JoinAsync may never complete indefinetly
`Cluster.JoinAsync()` `Task` may never complete, depending on 2 conditions:
- `akka.actor.serialize-messages` is set to true.
- The cluster already joined, the `RegisterOnMemberUp` callback never gets called, `Task` will never be set.

Arkatufus:
Closes #4484
Closes #4481

Serializer failed to serialize CurrentClusterState when SerializeAllMessages was set, causing MemberUp state to fail to propagate properly.


src/core/Akka.Cluster/ClusterEvent.cs
```diff
@@ -69,11 +70,16 @@ public interface IClusterDomainEvent : IDeadLetterSuppression { }
         /// </summary>
         public sealed class CurrentClusterState
         {
+            [JsonProperty]
```
Aaronontheweb, line 73: Just mark this class with [`INoSerializationVerificationNeeded`](https://getakka.net/api/Akka.Actor.INoSerializationVerificationNeeded.html)

Arkatufus: Done

PR #4491, opened 2020-06-23
Fix CPU utilization regression due to .NET SDK regression (addedum)

Arkatufus:
Related to #4434 and #4475

Forgot to upgrade .NET SDK in the Linux build file to 3.1.105


JessieWadman:
When do you think we can we expect this to be available on NuGet? Our Kubernetes is running so hot it's on fire :')   As in, will this be an immediate release, or queued for a milestone?

Aaronontheweb:
@JessieWadman which version are you running right now?

JessieWadman:
> @JessieWadman which version are you running right now?

We are running 1.4.8 currently

Aaronontheweb:
1.4.8 should have the most important CPU fixes if you have akka.remote batching disabled already. Appears to have been a DotNetty-related problem.

JessieWadman:
Oh, okay! I thought this PR would fix it for Linux runtimes.

Aaronontheweb:
All of the binaries we ship are compiled on Windows, which already has this fix. You might want to upgrade the version of the .NET Core SDK you're running on though.

JessieWadman:
Thanks! I'll check that 👍

PR #4492, opened 2020-06-24
Fix one typo in Akka Streams configuration doc

tometchy:
I've read your contributing guidelines, but decided not to do create feature branch only for one word typo fix.
If for some reason it's not acceptable, for example because of some _CI/CD_ rules, I don't mind rejecting this pull request and fixing typo some other way :)


Aaronontheweb:
@tometchy thank you!

PR #4499, opened 2020-07-03
Log rebalance progress at INFO per rebalance info

ismaelhamed:
> For cases where rebalance takes a long time due to a relatively large cluster change
> e.g. when doubling number of nodes (common for 1-2, 3-6 etc) there is
> little insignt into what is happening unless DEBUG logging is enabled.
> 
> This adds an INFO log per rebalance-interval (default is 10s) to show
> what is in progress and which new shards are starting rebalancing.


src/contrib/cluster/Akka.Cluster.Sharding/ShardCoordinator.cs
```diff
@@ -462,6 +462,13 @@ private static bool ShuttingDown(object message)
 
         private static void ContinueRebalance<TCoordinator>(this TCoordinator coordinator, IImmutableSet<ShardId> shards) where TCoordinator : IShardCoordinator
         {
+            if (coordinator.Log.IsInfoEnabled && (shards.Count > 0 || !coordinator.RebalanceInProgress.IsEmpty))
```
Aaronontheweb, line 465: LGTM

PR #4506, opened 2020-07-08
Implement Murmur2 hashing algorithm

Arkatufus:


Arkatufus:
Closed because port is canceled.

Issue #4419, opened 2020-05-18
How force akka system run on its own threadpool and mixed used of cluster client


Ralf1108:
Hi,

we are running Akka.Net v1.4.5 on Windows and experiencing random network partitions (sometime twice an hour, sometimes after 3 to 4 ours).
Occasionally we see missing heartbeats in the log:

> Scheduled sending of heartbeat was delayed. Previous heartbeat was sent [19170,2226] ms ago, expected interval is [1000] ms. This may cause failure detection to mark members as unreachable. The reason can be thread starvation, e.g. by running blocking tasks on the default dispatcher, CPU overload, or GC.

Sometimes everything works but otherwise we see:

```
[31] WARN  - 2020-05-18 01:39:45,690: Cluster Node [akka.tcp://So3ClusterServer@10.112.10.181:13178] - Marking node(s) as UNREACHABLE [Member(address = akka.tcp://So3ClusterServer@10.112.10.171:51747, Uid=1251870918 status = Up, role=[Web], upNumber=2)]. Node roles [So3]
[33] INFO  - 2020-05-18 01:39:53,925: A network partition detected - unreachable nodes: [akka.tcp://So3ClusterServer@10.112.10.171:51747], remaining: [akka.tcp://So3ClusterServer@10.112.10.181:13178]
[31] INFO  - 2020-05-18 01:39:53,972: A network partition has been detected. KeepMajority(role: 'So3') decided to down following nodes: [Member(address = akka.tcp://So3ClusterServer@10.112.10.171:51747, Uid=1251870918 status = Up, role=[Web], upNumber=2)]
[24] INFO  - 2020-05-18 01:39:53,987: Cluster Node [akka.tcp://So3ClusterServer@10.112.10.181:13178] - Marking unreachable node [akka.tcp://So3ClusterServer@10.112.10.171:51747] as [Down]
[24] INFO  - 2020-05-18 01:39:54,362: Cluster Node [akka.tcp://So3ClusterServer@10.112.10.181:13178] - Leader is moving node [akka.tcp://So3ClusterServer@10.112.10.171:51772] to [Up]
[24] INFO  - 2020-05-18 01:39:54,362: Cluster Node [akka.tcp://So3ClusterServer@10.112.10.181:13178] - Leader is removing unreachable node [akka.tcp://So3ClusterServer@10.112.10.171:51747]
[...]

PR #4511, opened 2020-07-10
Port scala akka PR #26816 to Akka.NET

Arkatufus:
Related to #4419
Port of https://github.com/akka/akka/pull/26816


Arkatufus:
I'm having problem with this port, the port passes a scala `ExecutionContext` into the `ActorSystem` factory, and this `ExecutionContext` trickles down into the `MessageDispatcher` to be used inside the thread pool.
Problem is, I can't really emulate this with C#.
Any help will be greatly appreciated.

Arkatufus:
ActorSystem.Create() --> ActorSystemImpl() --> DefaultDispatcherPrerequisites().
Prerequisites object is then injected into the thread executor factory to be used as a configurator for each thread instantiation.

Arkatufus:
Should `Akka.Cluster.Metric" and "Akka.Remote" use internal dispatcher too?

src/core/Akka/Configuration/Pigeon.conf
```diff
@@ -315,7 +315,23 @@ akka {
...
+        # Max number of threads to cap factor-based parallelism number to
+        parallelism-max = 64
+
+        # Setting to "FIFO" to use queue like peeking mode which "poll" or "LIFO" to use stack
+        # like peeking mode which "pop".
+        task-peeking-mode = "FIFO"
+
+
+        dedicated-thread-pool{ #settings for Helios.DedicatedThreadPool
             thread-count = 3 #number of threads
```
Aaronontheweb, line 335: We might need to adjust this too.... The dedicated threadpool sized might interfere with these settings on the `ForkJoinExecutor`

src/core/Akka/Helios.Concurrency.DedicatedThreadPool.cs
```diff
@@ -94,11 +94,16 @@ internal class DedicatedThreadPoolSettings
                 throw new ArgumentOutOfRangeException(nameof(numThreads), $"numThreads must be at least 1. Was {numThreads}");
         }
 #else
-        public DedicatedThreadPoolSettings(int numThreads, string name = null, TimeSpan? deadlockTimeout = null)
-            : this(numThreads, DefaultThreadType, name, deadlockTimeout)
+        public DedicatedThreadPoolSettings(int numThreads, string name = null, TimeSpan? deadlockTimeout = null, SynchronizationContext synchronizationContext = null)
+            : this(numThreads, DefaultThreadType, name, deadlockTimeout, synchronizationContext)
```
Aaronontheweb, line 98: Go ahead and remove `SynchronizationContext` from this. Not used.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -4823,6 +4826,7 @@ namespace Akka.Util
         public static readonly Akka.Util.Option<T> None;
         public Option(T value) { }
         public bool HasValue { get; }
+        public bool IsEmpty { get; }
```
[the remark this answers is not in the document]
Arkatufus: Its a convenience property, it saves from typing an awkward `if(!HasValue)` and it reads better.

src/core/Akka/Helios.Concurrency.DedicatedThreadPool.cs
Arkatufus: I missed that one, good catch

Aaronontheweb:
LGTM

src/core/Akka/Actor/ActorRefProvider.cs
```diff
@@ -296,7 +296,7 @@ private RootGuardianActorRef CreateRootGuardian(ActorSystemImpl system)
                 return Directive.Stop;
             });
             var props = Props.Create<GuardianActor>(rootGuardianStrategy);
-            var rootGuardian = new RootGuardianActorRef(system, props, DefaultDispatcher, _defaultMailbox, supervisor, _rootPath, _deadLetters, _extraNames);
+            var rootGuardian = new RootGuardianActorRef(system, props, InternalDispatcher, _defaultMailbox, supervisor, _rootPath, _deadLetters, _extraNames);
```
Aaronontheweb, line 299: So the only reason I haven't merged this PR yet - we are now guaranteed to have to create threads in the `InternalDispatcher` tool each and every time we create an `ActorSystem`, even for these `RootGuardianActorRef` actors who are very infrequently used.... I need to check the `DedicatedThreadPool` code to see if we allow those threads to terminate if there's not enough work for them to do. That'd be ideal and a better use of CPU resources rather than having those threads idly polling for work all the time.

Aaronontheweb:
Putting this on hold until we've completed https://github.com/akkadotnet/akka.net/issues/4537

Aaronontheweb:
> Should `Akka.Cluster.Metric" and "Akka.Remote" use internal dispatcher too?

I need to check if Akka.Remote should use the same internal dispatcher - that's critical enough where it might merit having its own threads still.

src/core/Akka/Actor/ActorRefProvider.cs
Aaronontheweb: Did this check - concluded that it is pretty safe.

PR #4512, opened 2020-07-11
DData: Use SetItem() instead of Add() for InitRemovedNodePruning

to11mtm:
Seeing some "An Item with the same key but different value already exists" exceptions in an App on InitRemovedNodePruning. Usually see this happen after rotating nodes during a deployment.

No tests added, but all existing tests pass. Also This change appears to be in line with Scala codebase, they use `.updated`: https://github.com/akka/akka/blob/142a63f600af6b8b805a10f0f401a4615237be48/akka-distributed-data/src/main/scala/akka/cluster/ddata/Replicator.scala#L937


Aaronontheweb:
We've had a number of bugs in the Akka.NET codebase stemming from `Dictionary.Add` - we should add a Roslyn analyzer that requires `Dictionary.SetItem` everywhere. Idempotent operations are almost always what we want.

PR #4513, opened 2020-07-12
Re-enable dead letter logging after specified duration

ismaelhamed:
This could be added to the release notes:

>#### Logging of dead letters
>
>When the number of dead letters have reached configured `akka.log-dead-letters` value it didn't log
>more dead letters in previous version. In Akka.NET 1.4.9 the count is reset after configured `akka.log-dead-letters-suspend-duration`.
>
>`akka.log-dead-letters-during-shutdown` default configuration changed from `on` to `off`.

### DIFF
--- a/src/core/Akka/Event/DeadLetterListener.cs
+++ b/src/core/Akka/Event/DeadLetterListener.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Threading;
 using Akka.Actor;
 
 namespace Akka.Event
@@ -16,6 +17,7 @@ namespace Akka.Event
     public class DeadLetterListener : ActorBase
     {
         private readonly EventStream _eventStream = Context.System.EventStream;
+        private readonly bool _isAlwaysLoggingDeadLetters = Context.System.Settings.LogDeadLetters == int.MaxValue;
         private readonly int _maxCount = Context.System.Settings.LogDeadLetters;
         private int _count;
 
@@ -38,7 +40,7 @@ namespace Akka.Event
         /// </summary>
         protected override void PreStart()
         {
-            _eventStream.Subscribe(Self, typeof (DeadLetter));
+            _eventStream.Subscribe(Self, typeof(DeadLetter));
         }
 
         /// <summary>
@@ -49,6 +51,19 @@ namespace Akka.Event
             _eventStream.Unsubscribe(Self);
         }
 
+        private void IncrementCount()
+        {
+            if (_count == int.MaxValue)
+            {
+                Logging.GetLogger(Context.System, this).Info("Resetting DeadLetterListener counter after reaching Int.MaxValue.");
+                _count = 1;
+            }
+            else
+            {
+                _count++;
+            }
+        }
+
         /// <summary>
         /// TBD
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/core/Akka/Event/DeadLetterListener.cs
```diff
@@ -49,35 +51,235 @@ protected override void PostStop()
...
         /// <param name="message">TBD</param>
         /// <returns>TBD</returns>
         protected override bool Receive(object message)
         {
-            var deadLetter = (DeadLetter)message;
-            var snd = deadLetter.Sender;
-            var rcp = deadLetter.Recipient;
+            if (_isAlwaysLoggingDeadLetters)
+            {
+                return ReceiveWithAlwaysLogging()(message);
```
ismaelhamed, line 76: This is the correct way of calling a Receive delegate, right?

src/core/Akka/Event/DeadLetterListener.cs
```diff
@@ -49,35 +51,235 @@ protected override void PostStop()
...
+                $"Message [{message.GetType().Name}] {origin} to {recipient.Path} was not delivered. [{_count}] dead letters encountered{doneMsg}. " +
+                $"If this is not an expected behavior then {recipient.Path} may have terminated unexpectedly. " +
+                "This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' " +
+                "and 'akka.log-dead-letters-during-shutdown'."));
+        }
+
+        /// <summary>
+        /// This class represents the latest date or time by which an operation should be completed.
+        /// </summary>
+        private class Deadline
```
ismaelhamed, line 180: Temporary copied this from `Akka.Remote`. Maybe we should move it to `Akka.Utils` or something.

src/core/Akka/Event/DeadLetterListener.cs
Aaronontheweb: Rather than return a `Receive` delegate here, I'd probably just feed the message directly into the handler. Fewer allocations this way.

src/core/Akka/Event/DeadLetterListener.cs
Aaronontheweb: Any reason why we shouldn't make this a `struct`? Any `deafult(Deadline)` calls anywhere in the main Akka.Remote code base? Or `null` checks?

src/core/Akka/Event/DeadLetterListener.cs
ismaelhamed: But then I need to pass a Receive delegate to the `Context.Become`. See line [133](https://github.com/akkadotnet/akka.net/blob/59fe3556ef1cf07fd50b32b39b92a1a849eb096e/src/core/Akka/Event/DeadLetterListener.cs#L133) for instance.

Aaronontheweb: Got it - that makes sense. I'm just worried about the memory footprint of returning new delegates for an actor that lives forever and might be receiving `DeadLetter`s regularly in a busy system. What would be a reasonable way of testing that? I just want to make sure that nothing gets rooted inside the `delegate`s so they can be GCed regularly.

ismaelhamed: I'll look into it. Do you have an alternative for this scenario, anyway?

Aaronontheweb: Eh, there probably isn't one. We should just merge this in then.

src/core/Akka/Event/DeadLetterListener.cs
ismaelhamed: @Aaronontheweb I made the exercise of converting Deadline into a readonly struct, since this one is private to the DeadLetterListener. Take a look and let me know.

We could see whether this could also be done in `Akka.Remote`  in another PR.

PR #4515, opened 2020-07-14
Change type constraints for graph builder

### LINKED ISSUE #4505: Forward Opts Type Constraints backwards?
I am using Akka.Net 1.4.0-beta3 and I am also using the documentation here: https://getakka.net/api/Akka.Streams.Dsl.ForwardOps.html

From what I can see, the ForwardOpts methods always constraint TIn : TOut. This allows this to compile:

`
GraphDsl.Create<ClosedShape>(
    builder =>
        {
                 builder.From(Source.Single<object>(""))
                      .To(builder.Add(Sink.Create<string>(null)));

                 return ClosedShape.Instance;
         });
`

This to me is wrong since a `string` sink can get things that are not strings as the object Source can basically emit whatever it wants. Furthermore, the compiler would reject this:

`
GraphDsl.Create<ClosedShape>(
    builder =>
        {
                 builder.From(Source.Single<string>(""))
                      .To(builder.Add(Sink.Create<object>(null)));

                 return ClosedShape.Instance;
         });
`

Which in my opinion is perfectly fine since a object sink should be able to be provided with strings.

Sorry if this issue is a dupe and I hope it helps. Thanks for the good work and great library.

IgorFedchenko:
Close #4505

### DIFF
--- a/src/core/Akka.Streams/Dsl/Internal/InternalFlowOperations.cs
+++ b/src/core/Akka.Streams/Dsl/Internal/InternalFlowOperations.cs
@@ -1999,7 +1999,7 @@ namespace Akka.Streams.Dsl.Internal
         /// <param name="segmentSize">TBD</param>
         /// <returns>TBD</returns>
         public static IFlow<T2, TMat> Interleave<T1, T2, TMat>(this IFlow<T1, TMat> flow,
-            IGraph<SourceShape<T2>, TMat> graph, int segmentSize) where T1 : T2
+            IGraph<SourceShape<T2>, TMat> graph, int segmentSize) where T1 : T1
         {
             return flow.Via(InterleaveGraph<T1, T2, TMat>(graph, segmentSize));
         }
--- a/src/core/Akka.Streams.Tests/Dsl/BidiFlowSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/BidiFlowSpec.cs
@@ -179,7 +179,7 @@ namespace Akka.Streams.Tests.Dsl
 
                 var right = Flow.FromGraph(GraphDsl.Create(Sink.First<List<long>>(), (b, sink) =>
                 {
-                    var flow = b.Add(Flow.Create<long>().Grouped(10));
+                    var flow = b.Add(Flow.Create<long>().Grouped(10).Select(collection => collection.ToList()));
                     var source = b.Add(Source.Single(ByteString.FromString("10")));
                     b.From(flow).To(sink);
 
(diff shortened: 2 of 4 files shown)


IgorFedchenko:
Have issues with `Interleave` method and few others like this.

Before PR, this was valid:
```
// Can do
Source.From(new List<Child>()).Interleave(Source.From(new List<Base>()), 2);
// Can not do
 Source.From(new List<Base>()).Interleave(Source.From(new List<Child>()), 2);
```

So method assumes first source can be child of second, but not vice versa. This is generally fine, it is just taking elements from first and converts to base class - but now we require another order of type arguments, and need to change the way this graphs works internally, I guess...

src/core/Akka.Streams/Dsl/Internal/InternalFlowOperations.cs
```diff
@@ -1999,7 +1999,7 @@ public static IFlow<(T1, long), TMat> ZipWithIndex<T1, TMat>(this IFlow<T1, TMat
         /// <param name="segmentSize">TBD</param>
         /// <returns>TBD</returns>
         public static IFlow<T2, TMat> Interleave<T1, T2, TMat>(this IFlow<T1, TMat> flow,
-            IGraph<SourceShape<T2>, TMat> graph, int segmentSize) where T1 : T2
+            IGraph<SourceShape<T2>, TMat> graph, int segmentSize) where T1 : T1
```
netogallo, line 2002: I think "T1 : T1" should be "T2 : T1"

:)

IgorFedchenko: Yeah, I thought so too, but this line will not compile

https://github.com/akkadotnet/akka.net/blob/0948525ff24f68fa939142f9d41f59744c0cd2ff/src/core/Akka.Streams/Dsl/Internal/InternalFlowOperations.cs#L2045

seems like we somehow use this "base to child" connection. Probably we are missing some core idea here...

netogallo: Does it have any value leaving the constraint?

Ie. will it not compile or cause other issues if the constraint is simply removed?

IgorFedchenko: Will not compile. Feel free to continue my work here by the way - could not change this code quickly enough to have this inheritance constraints swapped... Have a feeling that it works here due to some hack in graph architecture, but could not catch it.

netogallo: Thank you! I will give it a go!

netogallo: I saw this in the build logs: 

https://dev.azure.com/dotnet/Akka.NET/_build/results?buildId=35831&view=logs&j=d83928d8-8bfe-51de-721e-9025f72ee732&t=21d77f07-2bef-56ac-91e9-8618e6456479&l=1178

I just want to confirm if this is fine and if we need some compiler flag for T1 : T1 to be accepted. I got the same in my local build. Otherwise, I can try removing the constraint and see what I manage.

Let me know!

Issue #4518, opened 2020-07-15
Akka.Persistence.Query: need to make it clear that PersistenceIds queries only work on what the current JournalActor has observed. 


Aaronontheweb:
Version: 1.4.8

Common issue that I run into all the time myself: Akka.Persistence.Query `PersistenceIds` or `CurrentPersistenceIds` queries _only query what actually passes through the journal actors on their live, running nodes locally_ - not on the entire content of everything stored inside the database.

Need to make that clear going forward.

Aaronontheweb:
Need to implement https://github.com/akkadotnet/akka.net/issues/4519 to make some kind of "all persistence Ids" streaming query feasible.

PR #4521, opened 2020-07-15
Add AllEvents persistent query command

### LINKED ISSUE #4519: Akka.Persistence.Query: add "all events" query
In order to make it feasible to do things like continuously query all events, or query events that have never been tagged previously, we should add an "All Events" query that steps through the entire journal sequentially.

This query should have pagingation and live query semantics that are identical to the `EventsByTag` query. 

This query will not guarantee global order - it's going to read whatever was stored inside the journal as-is in that order.

Arkatufus:
Closes #4519

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Query.Sql/AllEventsPublisher.cs
+++ b/src/contrib/persistence/Akka.Persistence.Query.Sql/AllEventsPublisher.cs
@@ -1,0 +1,133 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Akka.Actor;
+using Akka.Event;
+using Akka.Persistence.Sql.Common.Journal;
+using Akka.Streams.Actors;
+
+namespace Akka.Persistence.Query.Sql
+{
+    internal sealed class AllEventsPublisher : ActorPublisher<EventEnvelope>
+    {
+        [Serializable]
+        public sealed class Continue
+        {
+            public static readonly Continue Instance = new Continue();
+
+            private Continue() { }
+        }
+
+        public static Props Props(long fromOffset, int maxBufferSize, string writeJournalPluginId)
+            => Actor.Props.Create(() => new AllEventsPublisher(fromOffset, maxBufferSize, writeJournalPluginId));
+
+        private readonly ILoggingAdapter _log;
+
+        private readonly DeliveryBuffer<EventEnvelope> _buffer;
+        private readonly IActorRef _journalRef;
+        private readonly int _maxBufferSize;
+        private bool _completed = false;
+
+        private readonly long _fromOffset;
+        private long _currentOffset;
+
+        public AllEventsPublisher(long fromOffset, int maxBufferSize, string writeJournalPluginId)
+        {
+            _currentOffset = _fromOffset = fromOffset;
+            _maxBufferSize = maxBufferSize;
+            _buffer = new DeliveryBuffer<EventEnvelope>(OnNext);
+            _journalRef = Persistence.Instance.Apply(Context.System).JournalFor(writeJournalPluginId);
+
+            _log = Context.GetLogger();
+        }
+
+        private bool IsTimeForReplay => (_buffer.IsEmpty || _buffer.Length <= _maxBufferSize / 2) && !_completed;
+
+        protected override bool Receive(object message) => message.Match()
+            .With<Request>(_ => Replay())
+            .With<Continue>(() => { })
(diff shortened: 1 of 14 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -556,6 +585,46 @@ public virtual async Task<ImmutableArray<string>> SelectAllPersistenceIdsAsync(D
...
+                AddParameter(command, "@Ordering", DbType.Int64, fromOffset);
+                AddParameter(command, "@Take", DbType.Int64, max);
+
+                var commandBehavior = Configuration.UseSequentialAccess ? 
+                    CommandBehavior.SequentialAccess : 
+                    CommandBehavior.Default;
+
+                using (var reader = await command.ExecuteReaderAsync(commandBehavior, cancellationToken))
+                {
+                    var rowCounter = 0;
```
Aaronontheweb, line 607: Might want to make this a `long`

src/contrib/persistence/Akka.Persistence.Query.Sql/AllEventsPublisher.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            var limit = _maxBufferSize - _buffer.Length;
+            _log.Debug("replay all events request from [{0}], limit [{1}]", _currentOffset, limit);
+            _journalRef.Tell(new ReplayAllEvents(_currentOffset, limit, Self));
+            Context.Become(Replaying());
+        }
+
+        private Receive Replaying()
+        {
+            return message => message.Match()
```
Aaronontheweb, line 75: Probably should get rid of our `PatternMatch` usage here and just go with a `switch` statement

Aaronontheweb: With the `default` returning `false` so the system knows it was unhandled

src/core/Akka.Persistence.TCK/Query/AllEventsSpec.cs
```diff
@@ -0,0 +1,134 @@
...
+            var eventSrc2 = queries.AllEvents(offs);
+            var probe2 = eventSrc2.RunWith(this.SinkProbe<EventEnvelope>(), Materializer);
+            probe2.Request(10);
+            // hello and world is not included, since exclusive offset
+            probe2.ExpectNext<EventEnvelope>(p => p.PersistenceId == "b" && p.SequenceNr == 1L && p.Event.Equals("test"));
+            probe2.Cancel();
+        }
+
+        [Fact]
+        public virtual void ReadJournal_query_AllEvents_should_see_all_150_events()
```
Aaronontheweb, line 108: Is 150 below the pagination threshold?

Arkatufus:
Forgot to make the appropriate changes to BatchingSqlJournal

src/core/Akka.Persistence.TCK/Query/AllEventsSpec.cs
Arkatufus: No, the command uses the max buffer size as the pagination size, it is set to 3 in the test.

Aaronontheweb: Ok good, so just so I understand correctly - this spec forces the journal to work through several pages worth of results in separate queries, right?

Aaronontheweb:
cc @ismaelhamed - see anything in this query that scares you? Wanted to add support for querying all existing messages in the journal with proper pagination and offset support, just like `EventsByTag` query. Everything looks good to me so far.

ismaelhamed:
LGTM too

src/core/Akka.Persistence.TCK/Query/AllEventsSpec.cs
Arkatufus: yes, it retrieves the events through separate queries, using "WHERE ordering > lastindex" and "LIMIT x"

Arkatufus:
#4518 in documentation, right? will do.

Aaronontheweb:
@Arkatufus yep, that's correct

Arkatufus:
PS: These changes still need to be propagated to the rest of Akka.Persistence.Sql plugin family

Aaronontheweb:
whoops, merged this in before you added docs - could you do that in a separate PR?

Arkatufus:
Documentation and code improvement is in #4523

PR #4534, opened 2020-07-30
Enhance OpenCircuitException error reporting

### LINKED ISSUE #4314: Akka.Pattern: Pass in clearer error message into OpenCircuitException

Arkatufus:
Closes #4314

- `CircuitBreaker` remembers the last exception that passes it inside the `LastCaughtException` property. 
  This exception is passed to the `OpenCircuitException` whenever it is thrown.
- Changed the `OpenCircuitException` message when it is thrown because of a half-open circuit breaker state.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -3891,10 +3891,11 @@ namespace Akka.Pattern
         public bool IsClosed { get; }
         public bool IsHalfOpen { get; }
         public bool IsOpen { get; }
+        public System.Exception LastCaughtException { get; }
         public int MaxFailures { get; }
         public System.TimeSpan ResetTimeout { get; }
         public static Akka.Pattern.CircuitBreaker Create(int maxFailures, System.TimeSpan callTimeout, System.TimeSpan resetTimeout) { }
-        public void Fail() { }
+        public void Fail(System.Exception cause = null) { }
         public Akka.Pattern.CircuitBreaker OnClose(System.Action callback) { }
         public Akka.Pattern.CircuitBreaker OnHalfOpen(System.Action callback) { }
         public Akka.Pattern.CircuitBreaker OnOpen(System.Action callback) { }
@@ -3917,6 +3918,7 @@ namespace Akka.Pattern
     public class OpenCircuitException : Akka.Actor.AkkaException
     {
         public OpenCircuitException() { }
+        public OpenCircuitException(System.Exception cause) { }
         public OpenCircuitException(string message) { }
         public OpenCircuitException(string message, System.Exception cause) { }
     }
@@ -5056,8 +5058,9 @@ namespace Akka.Util.Internal
     public interface IAtomicState
     {
         bool HasListeners { get; }
+        System.Exception LastCaughtException { get; }
         void AddListener(System.Action listener);
-        void Enter();
+        void Enter(System.Exception lastFailureCause);
         System.Threading.Tasks.Task<T> Invoke<T>(System.Func<System.Threading.Tasks.Task<T>> body);
     }
     public class static InterlockedSpin
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -3895,7 +3895,7 @@ namespace Akka.Pattern
         public int MaxFailures { get; }
         public System.TimeSpan ResetTimeout { get; }
         public static Akka.Pattern.CircuitBreaker Create(int maxFailures, System.TimeSpan callTimeout, System.TimeSpan resetTimeout) { }
-        public void Fail(System.Exception cause = null) { }
+        public void Fail() { }
         public Akka.Pattern.CircuitBreaker OnClose(System.Action callback) { }
         public Akka.Pattern.CircuitBreaker OnHalfOpen(System.Action callback) { }
         public Akka.Pattern.CircuitBreaker OnOpen(System.Action callback) { }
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
cc @ismaelhamed

Arkatufus:
Would this be enough information? I'm open for adding more if anyone can think of any,

src/core/Akka/Pattern/CircuitBreakerState.cs
```diff
@@ -37,9 +37,9 @@ public Open(CircuitBreaker breaker)
         /// <param name="body">N/A</param>
         /// <exception cref="OpenCircuitException">This exception is thrown automatically since the circuit is open.</exception>
         /// <returns>N/A</returns>
-        public override async Task<T> Invoke<T>(Func<Task<T>> body)
```
ismaelhamed, line 40: Watch out, @Horusiath introduced these as part of #3754 , though I've been willing to ask how that works actually.

src/core/Akka/Pattern/CircuitBreaker.cs
```diff
@@ -274,7 +276,7 @@ private void Transition(AtomicState fromState, AtomicState toState)
             if (SwapState(fromState, toState))
             {
                 Debug.WriteLine("Successful transition from {0} to {1}", fromState, toState);
-                toState.Enter();
+                toState.Enter(fromState.LastCaughtException);
```
ismaelhamed, line 279: Do we need to pass the old exception during the transition? Shouldn't the new transtion start with a clean slate?

src/core/Akka/Util/Internal/AtomicState.cs
```diff
@@ -196,6 +199,8 @@ public void Enter()
     /// </summary>
     public interface IAtomicState
     {
+        Exception LastCaughtException { get; }
```
ismaelhamed, line 202: Since the `CircuitBreaker._currentState` is of type `AtomicState` (not `IAtomicState`) and this is `internal` anyway, we could just make `LastCaughtException` a protected internal field of `AtomicState` instead. Just thinking out loud...

ismaelhamed:
@Arkatufus @Aaronontheweb nice one, this is going to help a lot.

src/core/Akka/Pattern/CircuitBreakerState.cs
Aaronontheweb: I'll need to check references to see if this actually gets used anywhere.

src/core/Akka/Pattern/CircuitBreaker.cs
Arkatufus: No, the last caught exception need to be passed from one state to another, for example, exception caught in open state needs to be passed to the closed state, so that failing fast can report the exception that caused the state to close, same thing with closed to half open.

src/core/Akka/Pattern/CircuitBreakerState.cs
Arkatufus: Its fine, the async keyword is just a syntactic sugar, it just flags the compiler that the function uses await and it should optimize it to use them. Since the method doesn't use await, we don't need one.

src/core/Akka/Util/Internal/AtomicState.cs
Arkatufus: LastCaughtException is accessed by the circuit breaker, so it needs to be public, but yeah, it can be moved to `AtomicState`. Actually, `IAtomicState` is an unnescessary abstraction, since the abstract `AtomicState` is enough.

src/core/Akka/Pattern/CircuitBreaker.cs
Arkatufus: I guess another way of doing this by making `OnFail` and `OnSuccess` callback and store/clear the last caught exception in the circuit breaker class instead, do you think that'll be cleaner?

src/core/Akka/Pattern/CircuitBreakerState.cs
Arkatufus: By the way, you can use the async version by passing a `Func` that produces a task into a circuit breaker like this:
```
await _circuitBreaker.WithCircuitBreaker(() => MyTask);
```
currently, it accepts `Func<Task>` and `Func<Task<T>`

src/core/Akka/Pattern/CircuitBreaker.cs
Aaronontheweb: > I guess another way of doing this by making `OnFail` and `OnSuccess` callback and store/clear the last caught exception in the circuit breaker class instead, do you think that'll be cleaner?

Yes, I think that's a cleaner way of doing this. Let's do that.

Aaronontheweb:
Thanks @Arkatufus - I'll review this ASAP

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -3891,10 +3891,11 @@ namespace Akka.Pattern
         public bool IsClosed { get; }
         public bool IsHalfOpen { get; }
         public bool IsOpen { get; }
+        public System.Exception LastCaughtException { get; }
         public int MaxFailures { get; }
         public System.TimeSpan ResetTimeout { get; }
         public static Akka.Pattern.CircuitBreaker Create(int maxFailures, System.TimeSpan callTimeout, System.TimeSpan resetTimeout) { }
-        public void Fail() { }
+        public void Fail(System.Exception cause = null) { }
```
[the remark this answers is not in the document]
Arkatufus: OK, I'll figure out another way

PR #4541, opened 2020-08-11
Clear language on clustering conditions

PatLeong:
`You should use clustering in any scenario where you have....Non-Trivial to perform.` didn't make sense to me. I'm actually taking a guess as to what this condition is, but I propose that you're saying that we should use clustering when there is a non-trivial task to perform (because you want complex tasks to be fault-tolerant?).


Aaronontheweb:
LGTM - thanks for your contribution

PR #4545, opened 2020-08-19
Add docs about routing management messages

eisendle:
Currently the router documentation is missing information about management messages.
These message are supported according to the Routing test spec.
This change adds information about these messages


Aaronontheweb:
@eisendle this is a great addition. Thank you.

PR #4552, opened 2020-08-21
Fix bug in BatchingSqlJournal CurrentAllEvents and AllEventsPublisher.

Arkatufus:
Fixed bugs:
- CurrentAllEvents uses improper highest sequence number instead of the highest ordering number.
- AllEventsPublisher, the CurrentAllEvents publisher may overshoot the intended ToOffset target.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
@@ -1069,14 +1069,22 @@ namespace Akka.Persistence.Sql.Common.Journal
 
             try
             {
-                var maxSequenceNr = 0L;
+                var toOffset = req.ToOffset;
                 var fromOffset = req.FromOffset;
+                var max = req.Max;
+
+                var take = Math.Min(toOffset - fromOffset, max);
+
+                command.CommandText = HighestOrderingSql;
+                command.Parameters.Clear();
+
+                var maxOrdering = (await command.ExecuteScalarAsync()) as long? ?? 0L;
 
                 command.CommandText = AllEventsSql;
                 command.Parameters.Clear();
 
                 AddParameter(command, "@Ordering", DbType.Int64, fromOffset);
-                AddParameter(command, "@Take", DbType.Int64, req.Max);
+                AddParameter(command, "@Take", DbType.Int64, take);
 
                 using (var reader = await command.ExecuteReaderAsync())
                 {
@@ -1084,7 +1092,6 @@ namespace Akka.Persistence.Sql.Common.Journal
                     {
                         var persistent = ReadEvent(reader);
                         var ordering = reader.GetInt64(OrderingIndex);
-                        maxSequenceNr = Math.Max(maxSequenceNr, persistent.SequenceNr);
 
                         foreach (var adapted in AdaptFromJournal(persistent))
                         {
@@ -1093,7 +1100,7 @@ namespace Akka.Persistence.Sql.Common.Journal
                     }
                 }
 
-                replyTo.Tell(new EventReplaySuccess(maxSequenceNr));
+                replyTo.Tell(new EventReplaySuccess(maxOrdering));
             }
             catch (Exception cause)
             {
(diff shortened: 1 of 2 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1069,22 +1069,29 @@ protected virtual async Task HandleReplayAllMessages(ReplayAllEvents req, TComma
...
+                command.Parameters.Clear();
+
+                var maxOrdering = (await command.ExecuteScalarAsync()) as long? ?? 0L;
 
                 command.CommandText = AllEventsSql;
                 command.Parameters.Clear();
 
                 AddParameter(command, "@Ordering", DbType.Int64, fromOffset);
-                AddParameter(command, "@Take", DbType.Int64, req.Max);
+                AddParameter(command, "@Take", DbType.Int64, take);
```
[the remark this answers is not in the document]
Arkatufus: That was part of it, th other was using the wrong number (sequence number) instead of the highest ordering number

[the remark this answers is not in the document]
Aaronontheweb: Where was the change for switching the number?

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1094,7 +1101,7 @@ protected virtual async Task HandleReplayAllMessages(ReplayAllEvents req, TComma
                     }
                 }
 
-                replyTo.Tell(new EventReplaySuccess(maxSequenceNr));
+                replyTo.Tell(new EventReplaySuccess(maxOrdering));
```
Aaronontheweb, line 1104: Ah I see, the change was here

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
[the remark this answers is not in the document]
Aaronontheweb: NVM, I found it

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1070,22 +1070,29 @@ protected virtual async Task HandleReplayAllMessages(ReplayAllEvents req, TComma
...
+                var toOffset = req.ToOffset;
                 var fromOffset = req.FromOffset;
+                var max = req.Max;
+
+                var take = Math.Min(toOffset - fromOffset, max);
+
+                command.CommandText = HighestOrderingSql;
+                command.Parameters.Clear();
+
+                var maxOrdering = (await command.ExecuteScalarAsync()) as long? ?? 0L;
```
Aaronontheweb, line 1082: So we have to query the max ordering number every time? Can't we just add an offset to the previous value on the command?

Arkatufus: These are legacy steps taken by old codes that does paging against a database. 
Basically, the publisher would have a variable called ToOffset which is initialized as long.MaxValue. The first executor result will contain the highest sequence number at that query. 
The publisher compares the value with ToOffset, if it is less than ToOffset, then store it in ToOffset.
Use the value of ToOffset in subsequent requests as a limit.

I don't know why it was done this way, I was just avoiding possible run-time behavior changes

Aaronontheweb: So do we only query that value once or do we have to do it for each piece of pagination?

Arkatufus: Basically, that MaxOrderingNr return value is only used in the `Current` query variant to limit returned events to only the last event row when the query was first executed

Aaronontheweb: Ah got it - that makes perfect sense

PR #4556, opened 2020-08-25
Fix premature reset of the 'writeInProgress' flag in case of persistence failure

ismaelhamed:
A fix for:

> Possible race condition in Akka Persistence that can lead to lost events when using persistAsync() [#28629](https://github.com/akka/akka/issues/28629)

The idea is not to reset the writeInProgress flag prematurely before the actor gets stopped by WriteMessageFailure.

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
@@ -709,8 +709,12 @@ namespace Akka.Persistence
     }
     public sealed class WriteMessagesFailed : Akka.Actor.INoSerializationVerificationNeeded, Akka.Persistence.IJournalMessage, Akka.Persistence.IJournalResponse, Akka.Persistence.IPersistenceMessage, System.IEquatable<Akka.Persistence.WriteMessagesFailed>
     {
+        [System.ObsoleteAttribute("Deprecated since Akka 1.4.11, use the overloaded one which accepts the number of " +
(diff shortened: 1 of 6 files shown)


src/core/Akka.API.Tests/CoreAPISpec.ApprovePersistence.approved.txt
```diff
@@ -709,8 +709,9 @@ namespace Akka.Persistence
     }
     public sealed class WriteMessagesFailed : Akka.Actor.INoSerializationVerificationNeeded, Akka.Persistence.IJournalMessage, Akka.Persistence.IJournalResponse, Akka.Persistence.IPersistenceMessage, System.IEquatable<Akka.Persistence.WriteMessagesFailed>
     {
-        public WriteMessagesFailed(System.Exception cause) { }
+        public WriteMessagesFailed(System.Exception cause, int writeCount) { }
```
Aaronontheweb, line 712: Need to change this to a new overload that gets called from the original constructor with a default value of some kind - otherwise this will break all journal implementations that aren't upgraded to 1.4.11+ right away.

ismaelhamed: I added back the previous constructor and marked it as obsolete --feel free to sugest a better message. `WriteCount` will always be zero in this case, which will make the code behave as before.

Aaronontheweb: Sounds like a good solution

PR #4559, opened 2020-09-04
Remove NetFrameworkLibVersion support

### LINKED ISSUE #4558: Cleanup NetFrameworkLibVersion variable
The variable [`NetFrameworkLibVersion`](https://github.com/akkadotnet/akka.net/blob/dev/src/common.props#L21) seems no longer used as a target, only in conditions. Is it safe to remove it from conditions and remove related things like the `#if UNSAFE_THREADING` pre-compiler blocks?

image

huysentruitw:
Removed NetFrameworkLibVersion

Let's first see what the build pipeline tells us, after that, we need to figure out what we will do with the following defines:

- UNSAFE_THREADING
- SERIALIZATION
- CLONEABLE
- APPDOMAIN
- AKKAIO
- CORECLR
- FSCHECK

We can also remove some additional left-overs from netstandard1.6

Should fix #4558


Aaronontheweb:
I'll review this today

Aaronontheweb (approved):
LGTM

PR #4561, opened 2020-09-07
Remove UNSAFE_THREADING blocks

huysentruitw:


src/core/Akka.Tests/Actor/Dispatch/ActorModelSpec.cs
```diff
@@ -219,11 +219,7 @@ public DispatcherActor()
                 Receive<CountDownNStop>(countDown => { Ack(); countDown.Latch.Signal(); Context.Stop(Self); _busy.SwitchOff(); });
                 Receive<Restart>(restart => { Ack(); _busy.SwitchOff(); throw new Exception("restart requested"); }, restart => true); // had to add predicate for compiler magic
                 Receive<Interrupt>(interrupt => { Ack(); Sender.Tell(new Status.Failure(new ActorInterruptedException(cause: new Exception(Ping)))); _busy.SwitchOff(); throw new Exception(Ping); }, interrupt => true);
-#if UNSAFE_THREADING
-                Receive<InterruptNicely>(interrupt => { Ack(); Sender.Tell(interrupt.Expect); _busy.SwitchOff(); Thread.CurrentThread.Interrupt(); });
```
huysentruitw, line 223: This is the only place where the UNSAFE_THREADING is still getting used (when testing Akka.Tests targetting `NetFrameworkTestVersion`). So not sure if this will be breaking or not.

Aaronontheweb: This is fine - it's an obscure error case that I've never seen happen in the wild.

PR #4562, opened 2020-09-08
Print TypeName in ShardRegion logs

ismaelhamed:
Port: [#26611](https://github.com/akka/akka/pull/26611)

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
@@ -474,7 +474,9 @@ namespace Akka.Cluster.Sharding
         private void LogPassivateIdleEntities()
         {
             if (Settings.ShouldPassivateIdleEntities)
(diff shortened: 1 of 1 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -1019,9 +1027,9 @@ private void HandleClusterEvent(ClusterEvent.IClusterDomainEvent e)
                 case ClusterEvent.MemberDowned md:
                     if (md.Member.UniqueAddress == Cluster.SelfUniqueAddress)
                     {
+                        Log.Info("{0}: Self downed, stopping ShardRegion [{1}]", TypeName, Self.Path);
                         Context.Stop(Self);
                     }
-                    Log.Info("Self downed, stopping ShardRegion [{0}]", Self.Path);
```
ismaelhamed, line 1024: This was always logging, whether it was the same node or not.

Aaronontheweb: Good catch

Aaronontheweb (approved):
LGTM

Issue #4567, opened 2020-09-10
Akka Should Allow Separate Read and Write Event Adapter Bindings


briansain:
- Akka.Net Version: 1.3

We are currently using Akka.Net and another 3rd party library ([Akkatecture](https://github.com/Lutando/Akkatecture)) in our application. Akkatecture has their own class that inherits from IReadEventAdapter that is used for event upgrading, and we have built our own IWriteEventAdapter that will tag our events according to our specifications. We made changes to some of our events, so we need to use IReadEventAdapter in order to upgrade those events to the new version, as well as use our own event tagging IWriteEventAdapter. We're not able to bind multiple read event adapters and a single write adapter, to a single event type. 

**Example/Reproduce** (Taken from .\src\core\Akka.Persistence.Tests\Journal\MemoryEventAdaptersSpec.cs):

In the constructor of MemoryEventAdaptersSpec, I add this configuration to the in-line hocon file:

      `""" + typeof(ReadWriteEvent).FullName + @", Akka.Persistence.Tests"" = [reader, writer]`

And then below, I add this class:
```
    public class ReadWriteEvent
    {
        public override string ToString()
        {
            return "ReadWriteEvent()";
        }
    }
```

I should expect that the Event Adapters that are bound to ReadWriteEvent would be an instance of the reader event adapter and the write event adapters that are defined in the test class constructor. However, the writer adapter isn't being used, but a NoopWriteEventAdapter is being set instead.

**Proposed Solution**: 
I will have a pull request out shortly that will add another event adapter that will accept multiple read event adapters and a single write event adapter.

ismaelhamed:
@briansain @Aaronontheweb 

You can only have multiple event adapters if they're of type `read`. Otherwise, any `write` event adapter will be wrapped in a `CombinedReadEventAdapter` for safety, I guess.

Now, as to why it's not allow to have multiple `read` event adapters with just once `write` event adapter, I'm not quite sure. But you can easily overcome this by creating a full-fledged `EventAdapter`:

```
public class ReaderWriterAdapter : IEventAdapter
{
    public string Manifest(object evt) => string.Empty;

    public object ToJournal(object evt) => "to-" + evt;

    public IEventSequence FromJournal(object evt, string manifest) => EventSequence.Single("from-" + evt);
}

```

briansain:
[...]

PR #4568, opened 2020-09-10
Create ReadWriteEventAdapter and set if event is bound to separate re…

briansain:
Fix for issue [#4567](https://github.com/akkadotnet/akka.net/issues/4567)

Created a new EventAdapter that will take in 1 write adapter, and multiple read adapters.

If the event is bound to mulitple read adapters, and no write adapters, then return a NoopWriteEventAdapter that has a CombindReadEventAdapter

If the event is bound to one or more read adapters, and a write adapter, then use the read adapters and write adapters when handling journal events

Throw an exception if there are mulitple write adapters for a single adapter binding.


Aaronontheweb:
cc @ismaelhamed @to11mtm

to11mtm:
I don't know a whole lot about real world use of Persistent Query... Here's what L2DB Journal Does (might be useful for background info...)

```
public Linq2DbReadJournal(ExtendedActorSystem system,
            Configuration.Config config, string configPath)
            //probably not related, but configpath used here 
            //to help with naming things, need to add support in ADN
        {
            //If User specified nondefault write-plugin, get it's name.
            writePluginId = config.GetString("write-plugin");

            //IDK Why we need this Register call, but we do or tests fail.
            system.RegisterExtension(Persistence.Instance);
            var persist = Persistence.Instance.Get(system);
            eventAdapters = persist
                .AdaptersFor(writePluginId);  //Find adapters for either default write plugin or specified
```

Then configuration is handled: 

```
akka {
  persistence {
    query {
      journal {
        myLinq2DbJournal { 
           class = "Akka.Persistence.Sql.Linq2Db.Query.Linq2DbReadJournalProvider, Akka.Persistence.Sql.Linq2Db"
           write-plugin = "akka.persistence.journal.myLinq2DbJournal"
           connection-string = "" # Connection String is Required!
           provider-name = "SqlServer.2012"
        }
      }
    }
  }
}
```

Aaronontheweb (approved):
LGTM

PR #4572, opened 2020-10-09
Fix ToString for KeepOldest strategy

pirrmann:
And then the logs look better!


Aaronontheweb:
Looks good to me! Thanks for your contribution.

PR #4577, opened 2020-10-13
Cluster - Add app-version to the Member information

zbynek001:
Add app-version to the Member information, (Migrated from https://github.com/akka/akka/pull/29546)
will be used in rolling update features
configured with akka.cluster.app-version


build.fsx
```diff
@@ -545,6 +545,7 @@ Target "Protobuf" <| fun _ ->
                 |> append (sprintf "-I=%s" (__SOURCE_DIRECTORY__ @@ "/src/protobuf/common") )
                 |> append (sprintf "--csharp_out=internal_access:%s" (__SOURCE_DIRECTORY__ @@ destinationPath))
                 |> append "--csharp_opt=file_extension=.g.cs"
+                |> append "--experimental_allow_proto3_optional"
```
zbynek001, line 548: needed for `optional` proto members

build.ps1
```diff
@@ -35,7 +35,7 @@ $DotNetVersion = "3.1.105";
 $DotNetInstallerUri = "https://dot.net/v1/dotnet-install.ps1";
 $NugetVersion = "4.3.0";
 $NugetUrl = "https://dist.nuget.org/win-x86-commandline/v$NugetVersion/nuget.exe"
-$ProtobufVersion = "3.4.0"
+$ProtobufVersion = "3.13.0"
```
zbynek001, line 38: needed for `optional` proto members

src/core/Akka/Util/AkkaVersion.cs
```diff
@@ -0,0 +1,251 @@
...
+    ///  plain alphanumeric if it couldn't be parsed as a number.
+    ///
+    ///  It may also have a qualifier at the end for 2 or 3 digit version numbers such as "1.2-RC1".
+    ///  For 1 digit with qualifier, 1-RC1, it is interpreted as plain alphanumeric.
+    ///
+    ///  It has support for https://github.com/dwijnand/sbt-dynver format with `+` or
+    ///  `-` separator. The number of commits from the tag is handled as a numeric part.
+    ///  For example `1.0.0+3-73475dce26` is less than `1.0.10+10-ed316bd024` (3 < 10).
+    /// </summary>
+    public class AkkaVersion : IComparable<AkkaVersion>, IEquatable<AkkaVersion>
```
zbynek001, line 32: Not sure about the name. Should be called just `Version`, but that would conflict everywhere with `System.Version`

build.ps1
Aaronontheweb: Got it

build.fsx
Aaronontheweb: LGTM

src/core/Akka.API.Tests/CoreAPISpec.ApproveCluster.approved.txt
```diff
@@ -73,6 +73,7 @@ namespace Akka.Cluster
             public CurrentClusterState() { }
             public CurrentClusterState(System.Collections.Immutable.ImmutableSortedSet<Akka.Cluster.Member> members, System.Collections.Immutable.ImmutableHashSet<Akka.Cluster.Member> unreachable, System.Collections.Immutable.ImmutableHashSet<Akka.Actor.Address> seenBy, Akka.Actor.Address leader, System.Collections.Immutable.ImmutableDictionary<string, Akka.Actor.Address> roleLeaderMap) { }
             public System.Collections.Immutable.ImmutableHashSet<string> AllRoles { get; }
+            public bool HasMoreThanOneAppVersion { get; }
```
Aaronontheweb, line 76: Ran into this issue the other day with a large user cluster - looks like a good addition

src/core/Akka.Cluster.Tests/ClusterConfigSpec.cs
```diff
@@ -45,6 +46,7 @@ public void Clustering_must_be_able_to_parse_generic_cluster_config_elements()
             settings.MinNrOfMembers.Should().Be(1);
             settings.MinNrOfMembersOfRole.Should().Equal(ImmutableDictionary<string, int>.Empty);
             settings.Roles.Should().BeEquivalentTo(ImmutableHashSet<string>.Empty);
+            settings.AppVersion.Should().Be(AkkaVersion.Zero);
```
Aaronontheweb, line 49: LGTM

src/core/Akka.Cluster/Configuration/Cluster.conf
```diff
@@ -77,6 +77,23 @@ akka {
...
+    # plain alphanumeric if it couldn't be parsed as a number.
+    #
+    # It may also have a qualifier at the end for 2 or 3 digit version numbers such
+    # as "1.2-RC1".
+    # For 1 digit with qualifier, 1-RC1, it is interpreted as plain alphanumeric.
+    #
+    # It has support for https://github.com/dwijnand/sbt-dynver format with `+` or
+    # `-` separator. The number of commits from the tag is handled as a numeric part.
+    # For example `1.0.0+3-73475dce26` is less than `1.0.10+10-ed316bd024` (3 < 10).
+    app-version = "0.0.0"
```
Aaronontheweb, line 95: So the App Version is used to specify the version of the user's software, not Akka.NET... so to take advantage of this, users will need to populate it inside their HOCONs. We'll need to add this to our cluster documentation.

src/core/Akka/Util/AkkaVersion.cs
Aaronontheweb: I can rename it `AppVersion` after merge - that way it wouldn't be confused with the Akka.NET version, which is what I initially thought when I saw the name.

Aaronontheweb (approved):
LGTM

PR #4581, opened 2020-10-15
Akka.Streams tcp transport

to11mtm:
This PR Contains:

 - An Akka.Remote transport utilizing Akka.Streams

 - Added IO.SocketOptions for:
   - TLSConnectionOption (WIP)
   - ByteBufferPoolSize to override default settings for Fetch Size (Transport works better at larger buffer sizes and we dont want people defaults in akka.io to trip up stream transport.

 - WIP TLS Support:
   - TLSConnection
   - TLSConnectionListener
   - TLSOutgoingConenection
   - Changes to TCPManager such that these will be created when TLSConnectionOption is opened.

This approach minimizes new code effort to get a working IO.TLS implementation up and running. In theory on newer net standard platforms (2.1+), the approach used for read/write will have SSLStream->NetworkStream utilize SocketAsyncEventArgs as best it can. There may be better potential for deupe here.
 
 - Changes to TcpManager such that when it sees a TLSConnection Option it will create TLS actors instead of TCP Actors
 
 - Changes to TCPConnection such that it will utilize ByteBufferPoolSize

 - Addition of COMPRESS/DECOMPRESS Flows for Akka Streams.
   - This was done while trying to see if pipeline could be optimized. It didn't help but since JVM Akka has it, might be useful.

### Status:
  - On local machine, Remotepingpong is at or ahead of DotNetty for both smaller and larger messages.
  - Dissociation is unclean, you see error logs dumped from stream actors terminating
  - Most of the remote unit tests pass, but there are some that do not, i.e. injected transport tests looking for gremlin in identifier fail, actorleak, etc.
 

This will DROP Dotnetty as a dependency and hopefully solve a lot of other issues we have seen.

### Needs:
  - Tests around SSL Concept
  - Any adaptations necessary for SSL in transport (I think this will actually be minimal thanks to the pattern of it just being an extra Inet.SocketOption)
  - Hardcoded entries moved into configuration (I'll try to work on this if I have time.)
  - Potentially testing/hardening around public/private configuration scenarios. _TCP is not my forte._
  - We do some tricks against `Google.Protobuf.ByteString` to eliminate a copy on write to transport. This should probably have some form of safety switch behind it; even if DepBot catches the issue, it sounds better to users to let them flip a switch rather than force immediate upgrade if they have a problem. We could either make it a config value, or 'test' run it before we decide to use it.
    - This is also a concern because I need to think about whether the way it is initialized could cause problems in that case regardless. Will ponder and if so fix.
  - Checks against dotnetty/helios compatibility. When I squint I'm hopeful the Akka.Streams framing is the same. If it's not we would have to make framing stages for that as well as helios. But it certainly seems simple.

### Nice to have:
  - Someone who knows streams better than I may know a few tricks to make the pipeline better. The places I put Async stages at were based on repeated RemotePingPong benchmarks as well as some debug checks (i.e. checking how many bytes were being pushed in the final `bufferedSelectFlow` stage.)

### Future ideas:
  - Once this is stable... this would not be backwards compatible. But since we are using streams, that would be only thing stopping us from in near future setting up a multi link with partition/mergehub. At that point you could interleave and/or have a channel where smaller messages are prioritized. Put a source.single handshake bytestring on pipe and merge with batching pump, set 100ms delay on batching pump so we know handshake goes first, listener will getoradd `StreamingTcpassociationhandle`/partition/mergehubs based on handshake match and wire up new connection... But, one step at a time. :)

### DIFF
--- a/src/core/Akka.Remote/Transport/Streaming/StreamingTcpTransport.cs
+++ b/src/core/Akka.Remote/Transport/Streaming/StreamingTcpTransport.cs
@@ -35,11 +35,17 @@ namespace Akka.Remote.Transport.Streaming
         {
             _queue = queue;
         }
+        public StreamingTcpAssociationHandle(Address localAddress,
+            Address remoteAddress,
+            TaskCompletionSource<ISourceQueueWithComplete<IO.ByteString>> queueTask) : base(localAddress, remoteAddress)
+        {
+            queueTask.Task.ContinueWith(r => _queue = r.Result);
+        }
 
         public override bool Write(ByteString payload)
         {
             
-            return _queue.OfferAsync(IO.ByteString.FromImmutable(payload.ToByteArray())).Result is QueueOfferResult
+            return _queue?.OfferAsync(IO.ByteString.FromImmutable(payload.ToByteArray())).Result is QueueOfferResult
                 .Enqueued;
         }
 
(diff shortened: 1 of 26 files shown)


to11mtm:
I've added some additional notes based on profiling and observations on some of the more 'experimental' branches.

Aaronontheweb:
@to11mtm this has just the Akka.Streams transport built into it?

to11mtm:
> 
> 
> @to11mtm this has just the Akka.Streams transport built into it?

Yes, Just the streams transport with minimal changes elsewhere, only the tiniest bit of manual serialization in this one.

I actually had the changes for threadpool locally, and have pushed them to this branch.

src/core/Akka/Util/ThreadLocalRandom.cs
```diff
@@ -28,7 +29,13 @@ public static Random Current
         {
             get
             {
-                return _rng.Value;
+                if (_rng == null)
```
Aaronontheweb, line 32: Why this change?

to11mtm: Paranoia change, but Reason is documented here:

https://ayende.com/blog/189761-A/production-postmortem-the-slow-slowdown-of-large-systems

src/core/Akka.Remote/Transport/Streaming/StreamingTcpTransport.cs
```diff
@@ -0,0 +1,384 @@
...
+
+            TransportSettings = StreamingTcpTransportSettings.Create(config);
+
+            SocketOptions = ImmutableList.Create<Inet.SocketOption>(
+                new Inet.SO.ReceiveBufferSize(
+                    TransportSettings.SocketReceiveBufferSize),
+                new Inet.SO.SendBufferSize(
+                    TransportSettings.SocketSendBufferSize),
+                new Inet.SO.ByteBufferPoolSize(TransportSettings.TransportReceiveBufferSize),
+                new Inet.SO.WorkerDispatcher("akka.remote.default-remote-dispatcher"));
```
to11mtm, line 138: Should the IO.Tcp Actors be running on Remote Dispatcher? or on their own (perhaps same pool as materializer?)

Aaronontheweb: Yes they should IMHO - I'd have them share the remote dispatcher

Aaronontheweb:
Looks like we are still having some wire issues with data being dropped

to11mtm:
Converting this back to a Draft since we need to fix issues with buffering inbound messages while associating or MNTRs fail

PR #4584, opened 2020-10-19
DO NOT MERGE: Streams TCP Transport with Manual Protobuf Serialization

to11mtm:
This branch contains a version of #4581 that uses Hand-wired Protobuf serialization for the main AkkaPDU/AkkaProtocolTransport path.

Doing so lowers certain copy-allocations in the current serialization strategy.

This implementation currently also uses LRU Caches for Serializer IDs, Manifest Strings, as well as sender/receiver paths to avoid re-creating the bytestrings.

Based on profiling this improves our write speed, however this does not reflect much in benchmarks, based on the data I think Reads are our bottleneck at this point.

Advantages:
  - Reduced GC Pressure compared to current serialization
  - Improved throughput in lower client messaging scenarios

Disadvantages:
  - Increased memory usage with LRU Caches

There are other improvements to be had here, basically we should be trying to pre-allocate arrays wherever possible and fill them, use the arraysegments at the end; lots of small bytestrings are bad for the socket buffer as are too many ByteString .Concats()


to11mtm:
Closing this for now as we are focusing on Artery/etc. Will keep branch live as the protobuf deserialization may be useful

PR #4587, opened 2020-10-23
akka.cluster - split brain resolver

zbynek001:
Migration of the split brain resolver from akka, since it's now open source. And has more features than the one currently in akka.net.
If somebody is up for a review :)


Aaronontheweb:
Thanks @zbynek001 - we'll take a look. Need to make sure things are backwards compatible with the current configuration settings - it looks like you've done that based on my first pass here but I'll need to take a deeper look.

src/core/Akka.Cluster/SBR/SplitBrainResolverProvider.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Akka.Cluster.SBR
+{
+    /// <summary>
+    /// Enabled with configuration:
+    /// {
+    /// akka.cluster.downing-provider-class = "Akka.Cluster.SBR.SplitBrainResolverProvider"
+    /// }
+    /// </summary>
+    public class SplitBrainResolverProvider : IDowningProvider
```
Aaronontheweb, line 15: I think we ought to make this backwards compatible with what we currently have in our documentation:

```
downing-provider-class = "Akka.Cluster.SplitBrainResolver, Akka.Cluster"
```

Can't really force people to update their configurations during a minor update. We can either resolve this with a branch inside the configuration classes for Akka.Cluster or we can do it via changing this type name. Thoughts?

Aaronontheweb (changes requested):
Handful of changes and discussion points here

Aaronontheweb:
@zbynek001 what do you think about my backwards compatibility remark? Is that handled somewhere and did I just miss it?

src/core/Akka.API.Tests/CoreAPISpec.ApproveCoordination.approved.txt
```diff
@@ -32,7 +32,6 @@ namespace Akka.Coordination
         public LeaseSettings(string leaseName, string ownerName, Akka.Coordination.TimeoutSettings timeoutSettings, Akka.Configuration.Config leaseConfig) { }
         public Akka.Configuration.Config LeaseConfig { get; }
         public string LeaseName { get; }
-        public System.Type LeaseType { get; }
```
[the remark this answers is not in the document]
zbynek001: type resolving was moved to `LeaseProvider` to support some test scenarios. And it's more in line how it's handled in akka

src/core/Akka.Cluster/SBR/SplitBrainResolverProvider.cs
zbynek001: I didn't replace the old resolver, so user can choose, which resolver to use. Because there is one scenario (`keep-referee`) which is not supported in the new resolver.
If you want to completely replace the old provider, we can either change the name & namespace or add some wrapper there or something. And make obsolete `keep-referee` scenario, which is also no longer supported in akka

Aaronontheweb: Got it - I'll noodle on that some.

Aaronontheweb:
I think what I'd like to do here is merge this in, make another PR to "merge" this implementation in with the existing SplitBrainResolver, and keep some form of the deprecated "keep referee" system. That makes more sense to me than keeping two different implementations.

Aaronontheweb:
Going to merge this in and work on backwards-compat stuff in a subsequent PR.

Aaronontheweb (approved):

Issue #2838, opened 2017-07-07
[Documentation] Remoting: clarify disassociation, reassociation and actor lifecycle


0x53A:
It would be great to extend the documentation at http://getakka.net/docs/Remoting on how the system behaves with regard to disassociation / reassociation and IActorRefs.

Quick test:

1) Connect two systems on two different computers and get a remote IActorRef through an ActorSelection.
1) Send a few messages between them, everything works.
1) Cut the network connection and wait a few seconds for the red/yellow error texts.
1) Reconnect the network cable.
1) Send a message to the previously obtained IActorRef -> everything works!

So it looks like the system auto-heals:
* After disassociation, messages are sent to deadletters.
* After reassociation, existing IActorRefs are still valid, and everything can resume.
* Awesome!

I also have two questions (but these may be more suited to stackoverflow):
* Is there a way to query whether a remote IActorRef is disassociated? I know you can ``Watch`` an actor, and will get a ``Terminated`` on death / disassociation, but I would like to just check if it is known-disassociated now, and don't care about the future.
* Is there an equivalent event to ``Terminated`` for re-association?

Aaronontheweb:
To help close out this issue - we need to add some documentation around this, but it is possible to track changes in association / disassociation while just using Akka.Remote:

```
Context.System.EventStream.Subscribe(Self, typeof(AssociatedEvent));
Context.System.EventStream.Subscribe(Self, typeof(DisassociatedEvent));
```

We use this inside [Petabridge.Cmd.Remote `pbm remote tail` for tailing live remoting events](https://cmd.petabridge.com/articles/commands/remote-commands.html#remote-tail)

>Is there a way to query whether a remote IActorRef is disassociated? I know you can Watch an actor, and will get a Terminated on death / disassociation, but I would like to just check if it is known-disassociated now, and don't care about the future.

`Context.Watch`-ing an `IActorRef` on a disassociated connection right now will result in an immediate `Terminated` message.

It's probably best to use Akka.Cluster for this sort of thing though, since you can query the global connectivity state via `Cluster.State` and see all of the unreachable members at any point in time.

PR #4592, opened 2020-10-27
Support for Own Actor to Subscribe `Disassociated` Event

gsw945:
related issues: https://github.com/akkadotnet/akka.net/issues/2838

### DIFF
--- a/src/core/Akka.Remote/Endpoint.cs
+++ b/src/core/Akka.Remote/Endpoint.cs
@@ -1941,7 +1941,14 @@ namespace Akka.Remote
                     }
                 }
             });
-            Receive<Disassociated>(disassociated => HandleDisassociated(disassociated.Info));
+            Receive<Disassociated>(disassociated =>
+            {
+                if (Context.System.EventStream.IsSubscribing(typeof(Disassociated)))
+                {
+                    Context.System.EventStream.Publish(disassociated);
+                }
+                HandleDisassociated(disassociated.Info);
+            });
             Receive<EndpointWriter.StopReading>(stop =>
             {
                 SaveState();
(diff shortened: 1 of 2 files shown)


src/core/Akka/Event/EventBus.cs
```diff
@@ -225,5 +225,10 @@ private List<TSubscriber> UpdateCacheForEventClassifier(TEvent @event, TClassifi
                 return list;
             }
         }
+
+        public bool IsSubscribing(TClassifier eventClass)
```
Aaronontheweb, line 229: So this checks to see if anyone is subscribed to a given topic? Can you rename this method to `HasSubscribers` ?

Aaronontheweb: Also, you will probably need to approve this change to the public API here via: https://getakka.net/community/public-api-changes.html

src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1941,7 +1941,14 @@ private void Reading()
                     }
                 }
             });
-            Receive<Disassociated>(disassociated => HandleDisassociated(disassociated.Info));
+            Receive<Disassociated>(disassociated =>
```
Aaronontheweb, line 1944: So this is actually a duplicate of functionality that we have inside the `ActorSystem` right now:

```
Context.System.EventStream.Subscribe(Self, typeof(AssociatedEvent));
Context.System.EventStream.Subscribe(Self, typeof(DisassociatedEvent));
```

We use this inside [Petabridge.Cmd.Remote `pbm remote tail` for tailing live remoting events](https://cmd.petabridge.com/articles/commands/remote-commands.html#remote-tail)

Aaronontheweb (changes requested):
No need for publishing the `Disassociated` event via the `EventStream` as `DisassociatedEvent` is already publicly available, but I like having the ability to check for subscribers on the `EventStream` though - would still accept an open PR for that.

PR #4593, opened 2020-10-27
Fix ByteString.Empty throws when ToArray was called

### LINKED ISSUE #4580: Writing an empty ByteString to an OutputStreamSink closes the underlying stream
- Which Akka.Net version you are using: 1.4.10
- On which platform you are using Akka.Net: Windows 10

This is either a bug of ByteString or OutputStreamSubscriber.
I create a Stream that should handle ByteStrings, the Sink is a named pipe wrapped in an OutputStreamSubscriber:

```
[TestMethod]
public async Task EmptyByteStringCausesPipeToBeClosed()
{
	NamedPipeServerStream serverPipe = new NamedPipeServerStream("uniquepipename", PipeDirection.Out, 1, PipeTransmissionMode.Byte, PipeOptions.Asynchronous, 20, 20);
	Task serverPipeConnectionTask = serverPipe.WaitForConnectionAsync();

	NamedPipeClientStream clientPipe = new NamedPipeClientStream(".", "uniquepipename", PipeDirection.In, PipeOptions.Asynchronous);
	clientPipe.Connect();

	await serverPipeConnectionTask;
	int cnt = 0;

	Task<IOResult> writeToStreamTask = Source.From(Enumerable.Range(0, 100))
		.Select(i => ByteString.FromString(i.ToString()))                          
		.Select(bs => cnt++ == 10 ? ByteString.Empty : bs)
		.ToMaterialized(StreamConverters.FromOutputStream(() => serverPipe), Keep.Right)
		.Run(materializer);

	Task readFromStreamTask = StreamConverters.FromInputStream(() => clientPipe, 1)
		.RunForeach(bs => TestContext.WriteLine(bs.ToString(Encoding.ASCII)), materializer);

	await Task.WhenAll(writeToStreamTask, readFromStreamTask);                       
}
```

The line `.Select(bs => cnt++ == 10 ? ByteString.Empty : bs)` causes the stream to be finished after the 10th byte.
This is because ByteString.Empty.ToArray() throws an Exception which is caught inside class OutputStreamSubscriber, and in the exception handler, the stage is Cancelled and the IOResult is set to failed.

I didn't expect this behaviour, because I would think writing an empty ByteString to a stream should be ok. Is there a reason, why ByteString.Empty.ToArray() throws instead of returning an empty array ? Or should the OutputStreamSubscriber handle this case better ?
The obvious workaround I use is filtering all Empty ByteStrings before sending them to the Sink...

Best regards,
Detlef

Arkatufus:
Closes #4580

The problem was an empty byte array buffer passed into the `CopyTo` function. `CopyTo` failed immediately with an `ArgumentOutOfRangeException`.
Fixed by immediately returning `Array.Empty<byte>()` when `ToArray()` is called on an empty `ByteString`.

### DIFF
--- a/src/core/Akka/Util/ByteString.cs
+++ b/src/core/Akka/Util/ByteString.cs
@@ -443,6 +443,9 @@ namespace Akka.IO
         /// <returns>TBD</returns>
         public byte[] ToArray()
         {
+            if (_count == 0)
+                return Array.Empty<byte>();
+
             var copy = new byte[_count];
             this.CopyTo(copy, 0, _count);
             return copy;
--- a/src/core/Akka.Streams.Tests/BugSpec.cs
+++ b/src/core/Akka.Streams.Tests/BugSpec.cs
@@ -1,0 +1,62 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO.Pipes;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Akka.IO;
+using Akka.Streams.Dsl;
+using Akka.Streams.IO;
+using Akka.TestKit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Streams.Tests
+{
+    public class BugSpec : AkkaSpec
+    {
+        private ActorMaterializer Materializer { get; }
+
+        public BugSpec(ITestOutputHelper helper) : base(helper)
+        {
+            Materializer = ActorMaterializer.Create(Sys);
+        }
+
+        [Fact]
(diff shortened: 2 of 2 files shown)


src/core/Akka/Util/ByteString.cs
```diff
@@ -443,6 +443,9 @@ public bool HasSubstring(ByteString other, int index)
         /// <returns>TBD</returns>
         public byte[] ToArray()
         {
+            if (_count == 0)
```
Aaronontheweb, line 446: LGTM

Aaronontheweb (approved):
LGTM

PR #4594, opened 2020-10-28
Streams TCP Transport with Hand-Written Protobuf Deserialization

to11mtm:
Last One, Promise. :)

This branch has a few more tweaks:

 - Hand written Deserializer for Remote Protobuf payloads to minimize copies on read.
 - `LRUSegmentCache` for using `Span<byte>` for Input (i.e. UTF8 Actor Paths, only copying to `string` if needed)
 - Switching to `ThreadStatic` from `ThreadLocal` where possible
 - Moving Stream Materializer for Streaming transport to it's own dispatcher config. At the moment it is using thread pool which lowers performance a bit, but hopefully can be brought up with tweaks to batch settings or thread pool throughput.


Aaronontheweb:
I'll take a look!

to11mtm:
# What This needs, copied/adapted from #4581 

Needs AKA "I need help!" :

   - Tests around SSL Concept
   - Any adaptations necessary for SSL in transport (I think this will actually be minimal thanks to the pattern of it just being an extra Inet.SocketOption)
   - Hardcoded entries moved into configuration (I'll try to work on this if I have time.)
     -   Config entries are no longer hard coded, but we should decide whether to fill in certain parts or fallback to Dotnetty Settings.
 -  Hardening around using no delay in batch grouping. RemotePingpong becomes unstable in this scenario.
 -  Potentially testing/hardening around public/private configuration scenarios. TCP is not my forte.
 -   Checks against dotnetty/helios compatibility. When I squint at DotNetty I'm hopeful the Akka.Streams framing is the same. If it's not we would have to make framing stages for that as well as helios. But it certainly seems simple.
 - **MUST Check/Fix** Not entirely certain the protobuf varint128 code handles negative numbers correctly. I borrowed it from the google repo but might have picked something that didn't do negative numbers
 - **This is a very strict and opinionated parser**. We may want to provide a fallback mode in case the way older clients encode messages is not able to be decoded. Google's protobuf serializer does NOT guarantee deterministic order of fields, this parser expects it.

Nice to have:

 -  Someone who knows streams better than I may know a few tricks to make the pipeline better. The places I put Async stages at were based on repeated RemotePingPong benchmarks as well as some debug checks (i.e. checking how many bytes were being pushed in the final bufferedSelectFlow stage.)

to11mtm:
@Aaronontheweb  I added what this needs if anyone in the community or elsewhere wishes to help!

If some of this is too extreme, we can move the important parts (scheduler, batch tweaks,etc) and still do a couple more zero-copy tricks on one of the other branches.

FWIW I'd say this is the most 'extreme' branch, The one where we are only *serializing* the protobuf payloads I would say is pretty dang safe.

Aaronontheweb:
@to11mtm have any before / after performance numbers you can share?

to11mtm:
These are round-about numbers ATM, will try to run more real benchmarks later.

But these numbers are fair in that they do show some scale:

RemotePingPong, I7 8750H, 36GB Ram.

### Dedicated Thread Pool:
 - Streaming, No special Protobuf Serialization (i.e. just minimizing copies from ByteString to ByteArray for serialization/deserialization:
   - ~170,000-180,000 messages per sec for numclients>1

 - Streaming, Hand-Written Protobuf Serializer/Deserializer (this branch)
    - 220,000-240,000 messages per sec for numclients >1

 - DotNetty, TCP
    - 220,000-240,000 Messages per sec for numclients >1 

### Running streaming all streaming transport bits (i.e. Materializer, remoting, IO.TCP actors) on `akka.actor.default-dispatcher`:
  - Streaming, No special Protobuf Serialization
    - ~210,000-240,000 messages per sec for numclients >1

  - Streaming, Hand written Protobuf Serializer/Deserializer (This branch)
    - ~240,000-300,000 Messages per sec for numclients >1

to11mtm:
I added one more bench number to above, since I think it does a decent job of showing that the more threadpools we have involved the more things seem to hurt.

That said, I think these are some good baselines for everyone to see where things are at.

Aaronontheweb:
@to11mtm started work reviewing this PR - been working my way through it. I'll go ahead and try to run these benchmarks locally too.

Aaronontheweb:
> I added one more bench number to above, since I think it does a decent job of showing that the more threadpools we have involved the more things seem to hurt.

The mutual exclusion benefits don't really show up well under single connection benchmarks - but we definitely need to be judicious about limiting the number of independent threadpools we use. Returns diminish quickly on each additional one.

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
```diff
@@ -0,0 +1,313 @@
...
+        {
+            _listener?.Notify(inboundPayload);
+        }
+    }
+    class StreamingTcpTransport : Transport
+    {
+        public sealed override string SchemeIdentifier { get; protected set; } =
+            "tcp";
+        private ActorMaterializer _mat;
+        private DotNettyTransportSettings Settings;
```
Aaronontheweb, line 82: nice, for making this backwards compatible...

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
```diff
@@ -0,0 +1,313 @@
...
+                    {   
+                        var otherListener = s.Result;
+                        handle.RegisterListener(otherListener);
+                    }, TaskContinuationOptions.ExecuteSynchronously);
+                    var outQueue = ic.HandleWith(buildFlow(handle)
+                        .Recover(
+                            ex =>
+                            {
+                                handle.Notify(
+                                    new Disassociated(DisassociateInfo.Unknown));
```
Aaronontheweb, line 187: Good way to error-handle this

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
```diff
@@ -0,0 +1,313 @@
...
+                                return IO.ByteString.Empty;
+                            }), _mat);
+                    queuePromise.SetResult(outQueue);
+                    listener.Notify(new InboundAssociation(handle));
+                });
+                return NotUsed.Instance;
+            }).ToMaterialized(Sink.Ignore<NotUsed>(),Keep.Left).Run(_mat);
```
Aaronontheweb, line 194: We need to pass in some hints to the materializer here as to which dispatcher the materialized actors should run on, IIRC

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        
+        private Source<IO.ByteString, ISourceQueueWithComplete<IO.ByteString>> sendFlow()
+        {
+            //TODO: Improve batching voodoo in pipeline.
+            return Source
+                .Queue<IO.ByteString>(512, OverflowStrategy.DropNew)
+                .Via(
+                    Framing.SimpleFramingProtocolEncoder(Settings.MaxFrameSize))
+                .GroupedWithin(256, TimeSpan.FromMilliseconds(40))
```
Aaronontheweb, line 254: Nice, a much more concise way of doing the "batching" than what our current transport does. We'll still need a flag to disable the batching stage via config for connections with low traffic (i.e. a system like Lighthouse)

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
```diff
@@ -0,0 +1,313 @@
...
+            queuePromise.SetResult(run.sq);
+            return handle;
+        }
+
+        
+        private Source<IO.ByteString, ISourceQueueWithComplete<IO.ByteString>> sendFlow()
+        {
+            //TODO: Improve batching voodoo in pipeline.
+            return Source
+                .Queue<IO.ByteString>(512, OverflowStrategy.DropNew)
```
Aaronontheweb, line 251: Probably better for us if we just reject the new write to the queue, so the `EndpointWriter` can backoff

Aaronontheweb (commented):
Leaving my initial comments up - not done yet, but best I could get to this evening

src/benchmark/RemotePingPong/Program.cs
```diff
@@ -69,7 +73,14 @@ public static Config CreateActorSystemConfig(string actorSystemName, string ipOr
 
         private static void Main(params string[] args)
         {
-            Process.GetCurrentProcess().PriorityClass = ProcessPriorityClass.High;
+            AppDomain.CurrentDomain.UnhandledException += (sender, eventArgs) =>
+            {
+                Console.WriteLine(eventArgs.ExceptionObject as Exception);
+                Console.WriteLine("STACKOVERFLOW");
```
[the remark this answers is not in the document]
to11mtm: Uhhhh, it rather involved some Spans and FastHash while I was debugging... :)

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
to11mtm: @Aaronontheweb my bad on this, you want to look at /src/core/akka.remote/  for the most up to date bits.

Moving everything into its own /contrib/ will be a bit more work because of all the `internal` things a transport needs

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
to11mtm: the /core/akka.remote/ code has dispatcher hints :)

Aaronontheweb: Ah ok - I was wondering if I was missing something

src/benchmark/RemotePingPong/Program.cs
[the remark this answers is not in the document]
Aaronontheweb: LOL

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
```diff
@@ -0,0 +1,313 @@
...
+        {
+            queueTask.Task.ContinueWith(r => _queue = r.Result);
+        }
+
+        public override bool Write(ByteString payload)
+        {
+
+            try
+            {
+                return _queue?.OfferAsync(IO.ByteString.FromBytes(payload.ToByteArray())).Result is QueueOfferResult
```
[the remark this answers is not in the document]
to11mtm: (Question still valid on live code so I'll just answer here:) 
It proably wouldn't hurt. I was suprised at how much time `.OfferAsync` was taking in the profiler.

IIRC the challenge was the actors calling it aren't ReceiveActors and I don't know how to make other ones work with Async :)

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
to11mtm: Not gonna lie, I want to FRAME that block of code. While I know there's room for improvement it just feels so elegant.

src/contrib/transports/Akka.Remote.Transport.Streams.Tcp/Class1.cs
Aaronontheweb: I think this is a great start - I’m very impressed with your hard work. I think this will be an amazing improvement over DotNetty

Aaronontheweb: Looking forward to reading the rest of it

to11mtm:
# What This needs:



> 
> 
> Leaving my initial comments up - not done yet, but best I could get to this evening

No problem. In fact, I apologize in advance for AkkaPduCodec.cs . You may want to skip that for the moment; right now it is a poster-child for "Why Projects have Contributor Guidelines".

src/core/Akka/IO/Inet.cs
```diff
@@ -221,6 +222,39 @@ public override void AfterConnect(Socket s)
                     //TODO: What is the .NET equivalent
                 }
             }
+
+            public class ByteBufferPoolSize : SocketOption
```
to11mtm, line 226: The intent of this class is to explicitly set a Byte Buffer Pool size for an IO.TCP Connection without disrupting what is configured in HOCON as default.

src/core/Akka/IO/Inet.cs
```diff
@@ -221,6 +222,39 @@ public override void AfterConnect(Socket s)
...
+            {
+                public int ByteBufferPoolSizeBytes { get; }
+
+                public ByteBufferPoolSize(int sizeBytes)
+                {
+                    ByteBufferPoolSizeBytes = sizeBytes;
+                }
+            }
+
+            public class TlsConnectionOption : SocketOption
```
to11mtm, line 236: Idea here is this is sent as SocketOption to configure TLS

src/core/Akka/IO/Inet.cs
```diff
@@ -221,6 +222,39 @@ public override void AfterConnect(Socket s)
...
+                public bool ClientAsServer { get; set; }
+
+                public TlsConnectionOption(X509Certificate2 cert, bool suppressValidation)
+                {
+                    Certificate = cert;
+                    SuppressValidation = suppressValidation;
+                }
+            }
+
+            public class WorkerDispatcher : SocketOption
```
to11mtm, line 249: Similar to ByteBuffer setting, this lets a User have a TCP Connection run on a specific Dispatcher instead of Akka.IO Defaults in HOCON

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -135,7 +135,10 @@ public static void SetBuffer(this SocketAsyncEventArgs args, IEnumerable<ByteStr
             }
             else
             {
-                args.SetBuffer(null, 0, 0);
+                if (args.Buffer != null)
```
to11mtm, line 138: SetBuffer(Null) has a lot going on in it. Checking if the buffer is null, on the other hand, is a simple Get in `SocketAsyncEventArgs`

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -104,7 +104,7 @@ enum ConnectionStatus
         private CloseInformation _closedMessage;  // for ConnectionClosed message in postStop
 
         private IActorRef _watchedActor = Context.System.DeadLetters;
-
+        protected abstract IBufferPool BufferPool { get; }
```
to11mtm, line 107: Connections may now have explicit buffer pools set, so we need this

src/core/Akka/IO/TcpManager.cs
```diff
@@ -100,5 +119,21 @@ protected override bool Receive(object message)
...
+            IEnumerable<Inet.SocketOption> options)
+        {
+            if (options.FirstOrDefault(o => o is Inet.SO.WorkerDispatcher) is
+                Inet.SO.WorkerDispatcher opt)
+            {
+                return props.WithDispatcher(opt.Dispatcher);
+            }
+
+            return props;
+        }
```
to11mtm, line 137: Since I can't fathom how making TCP actors would be on the Hot path, went this route to keep things simple and DSL-y

src/core/Akka/IO/TlsConnection.cs
```diff
@@ -0,0 +1,1183 @@
+﻿// //-----------------------------------------------------------------------
```
to11mtm, line 1: This is basically a Stab-in-the-dark at using `SslStream` in place of `Socket`.

src/core/Akka/IO/TlsConnection.cs
```diff
@@ -0,0 +1,1183 @@
...
+using Akka.Actor;
+using Akka.Dispatch;
+using Akka.Event;
+using Akka.IO.Buffers;
+using Akka.Pattern;
+using Akka.Util;
+
+namespace Akka.IO
+{
+    internal class SendReceiveArgsFake
```
to11mtm, line 29: SslStream doesn't have `SocketAsyncEventArgs`, so to keep this class as similar as possible to Tcp we use this `SendReceiveArgsFake` to mimic the abstraction

src/core/Akka/IO/TlsConnection.cs
```diff
@@ -0,0 +1,1183 @@
...
+                        {
+                            foreach (var buffer in byteString.Buffers)
+                            {
+                                await _connection.SslStream.WriteAsync(
+                                    buffer.Array,
+                                    buffer.Offset, buffer.Count);
+                            }
+                        }
+
+                        return NotUsed.Instance;
```
to11mtm, line 1003: This is probably Bad and Wrong. There's definitely some ugly going on:
  - Task.Run is a smell
  - Lots of back-and-forth here with awaits when we have small buffers (i.e. after manual serialization/framing/batching)

src/core/Akka/Util/ByteString.cs
```diff
@@ -191,6 +192,12 @@ public static ByteString FromString(string str, Encoding encoding)
         private readonly int _count;
         private readonly ByteBuffer[] _buffers;
 
+        public ByteString(IList<ByteString> buffers)
+        {
+            _count = buffers.Sum(s => s.Count);
+            _buffers = buffers.SelectMany(r=>r._buffers).ToArray();
+        }
+
```
to11mtm, line 200: So, this is useful for some cases, but looking at it I can't help but wish we had a nice way to pack this and fetch lazily (so we don't have to do this SelectMany+ToArray

src/core/Akka/Util/ByteString.cs
```diff
@@ -451,6 +458,12 @@ public byte[] ToArray()
             return copy;
         }
 
+        internal ByteBuffer ReadOnlyCompacted()
+        {
+            var c = this.Compact();
+            return c._buffers[0];
+        }
```
to11mtm, line 465: The idea here is to get a Single Contiguous Arraysegment back; For smaller messages, this will most likely not copy/move anything in `.Compact()` step.

 However in either case, the ArraySegment(ByteBuffer) can then be used directly either by 'Proper' Protobuf deserializer or custom deserializer.

src/core/Akka/Util/ThreadLocalRandom.cs
```diff
@@ -19,7 +19,9 @@ public static class ThreadLocalRandom
     {
         private static int _seed = Environment.TickCount;
 
-        private static ThreadLocal<Random> _rng = new ThreadLocal<Random>(() => new Random(Interlocked.Increment(ref _seed)));
+        [ThreadStatic] 
+        private static Random _rng2;
+        //private static ThreadLocal<Random> _rng = new ThreadLocal<Random>(() => new Random(Interlocked.Increment(ref _seed)));
```
to11mtm, line 24: Avoiding ThreadLocal wherever we don't need it for some special reason.

Reasoning here: https://ayende.com/blog/189761-A/production-postmortem-the-slow-slowdown-of-large-systems

src/core/Akka.Streams/Implementation/Fusing/GraphInterpreter.cs
```diff
@@ -344,7 +344,8 @@ public sealed class Connection
         // Using an Object-array avoids holding on to the GraphInterpreter class
         // when this accidentally leaks onto threads that are not stopped when this
         // class should be unloaded.
-        private static readonly ThreadLocal<object[]> CurrentInterpreter = new ThreadLocal<object[]>(() => new object[1]);
+        [ThreadStatic]
+        private static object[] CurrentInterpreter;
```
to11mtm, line 348: Changing to ThreadStatic because of https://ayende.com/blog/189761-A/production-postmortem-the-slow-slowdown-of-large-systems

src/core/Akka.Remote/Transport/AkkaPduCodec.cs
```diff
@@ -487,12 +759,240 @@ private AcknowledgementInfo AckBuilder(Ack ack)
...
+                    protoLengthDelimitedHeader(1, senderPathUtf8.Length))
+                .Concat(IO.ByteString.FromBytes(senderPathUtf8));
+            
+            return EncodeMessageAsField(4, mySender);
+        }
+
+        IO.ByteString CreateAckInfo(Ack ackOption)
+        {
+            return EncodeMessageAsField(1,WriteSingleLongAs64Bit(1, (ulong)ackOption.CumulativeAck.RawValue)
+                .Concat(WriteRepeatedLongsAs64Bit(2, ackOption.Nacks)));
```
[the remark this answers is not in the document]
Aaronontheweb: Nice, when can I see the new code?

PR #4598, opened 2020-10-29
akka.cluster - disseminate downing decisions faster

zbynek001:
disseminate downing decisions faster (Migrated from  https://github.com/akka/akka/pull/29640)

* when SBR downs the reachable side (minority) it's important
  to quickly inform everybody to shutdown
* send gossip directly to downed node, STONITH signal
* gossip to a few random immediately when self is downed, which
  is always the last from the SBR downing
* enable gossip speedup when there are downed members
* adjust TransitionSpect to the new behavior

* update ValidNodeForGossip behavior


Aaronontheweb:
looks excellent - I'll be eager to dive in and take a look.

Aaronontheweb (approved):
LGTM

PR #4600, opened 2020-10-31
akka.cluster.sharding - batch update

zbynek001:
Mostly Sharding allocation / rebalancing changes making use of the newly introduced AppVersion.
Let me know if you want to split it into smaller parts.

* Holistic shard allocation strategy,  (migrated from https://github.com/akka/akka/pull/29555)

  *  The rebalance in the LeastShardAllocationStrategy is only comparing the region
  with most shards with the one with least shards. Makes the rebalance rather
  slow. By default it's only rebalancing 1 shard at a time.
  * This new strategy looks at all current allocations to find the optimal
  number of shards per region and tries to adjust towards that value.
  Picking from all regions with more shards than the optimal.
  * Absolute and relative limit on how many shards that can be rebalanced
  in one round.
  * It's also not starting a new rebalance round until the previous has
  completed.
  * unit tests
  * second phase for fine grained rebalance, due to rounding it will not be perfect in the first phase
  * randomized unit test
  * configuration settings
  * docs

* Cluster aware shard allocation and rolling updates (migrated from https://github.com/akka/akka/pull/29548)
  * Adds some level of cluster awareness to both the LeastShardAllocationStrategy implementations:
  * prefer shard allocations on new nodes during rolling updates
  * don't rebalance during rolling update
  * don't rebalance when there are joining nodes
  * don't allocate to leaving, downed, exiting and unreachable nodes
  * When allocating when there are joining, unreachable, are leaving are de-prioritized to decrease the risk that a shard is allocated just to directly need to be re-allocated on a different node.

* Handle when there are no known regions yet (migrated from https://github.com/akka/akka/pull/29712)

* ShardRegion ignores HostShard when shutting down (migrated from https://github.com/akka/akka/pull/29707)

* Adds ActorRef.Ignore (Migrated from https://github.com/akka/akka/pull/28630)

* Allocate rebalanced shards immediately (migrated from https://github.com/akka/akka/pull/29551)
  * Otherwise it will continue to deallocate when there is no traffic
  that triggers allocations

* Forward terminated from ShardCoordinator to RebalanceWorker (migrated from https://github.com/akka/akka/pull/29463)
  * Avoiding the need for rebalance workers to watch shard regions which is
expensive as there is one rebalance worker per shard
  * fix race in rebalance worker (partially migrated from https://github.com/akka/akka/pull/29502)

* ClusterSharding: increasing retry interval for registration call (migrated from https://github.com/akka/akka/pull/26756)

* Make a distinction between hand-offs for rebalance and region shutdown (migrated from https://github.com/akka/akka/pull/29579)

* improve logging of shard region shutdown timeout (migrated from https://github.com/akka/akka/pull/29747)

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/MultiNodeClusterShardingConfig.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/MultiNodeClusterShardingConfig.cs
@@ -64,7 +64,6 @@ namespace Akka.Cluster.Sharding.Tests
                 ConfigurationFactory
                     .ParseString($@"""
                         akka.actor.provider = ""cluster""
-                        akka.cluster.downing-provider-class = akka.cluster.testkit.AutoDowning
                         akka.cluster.testkit.auto-down-unreachable-after = 0s
                         akka.cluster.sharding.state-store-mode = ""{mode}""
                         akka.cluster.sharding.remember-entities = {rememberEntities}
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/MultiNodeClusterShardingConfig.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/MultiNodeClusterShardingConfig.cs
@@ -1,0 +1,92 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="MultiNodeClusterShardingConfig.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2020 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2020 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System.Text.RegularExpressions;
+using Akka.Cluster.TestKit;
+using Akka.Configuration;
+using Akka.Remote.TestKit;
+
+namespace Akka.Cluster.Sharding.Tests
+{
+    public abstract class MultiNodeClusterShardingConfig : MultiNodeConfig
+    {
+        protected static Config PersistenceConfig()
+        {
+            return ConfigurationFactory.ParseString($@"""
+                akka.actor {{
+                    serializers {{
+                        hyperion = ""Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion""
+                    }}
+                    serialization-bindings {{
+                        ""System.Object"" = hyperion
+                    }}
+                }}
+                akka.persistence.snapshot-store.plugin = ""akka.persistence.snapshot-store.inmem""
+                akka.persistence.journal.plugin = ""akka.persistence.journal.memory-journal-shared""
+
+                akka.persistence.journal.MemoryJournal {{
+                    class = ""Akka.Persistence.Journal.MemoryJournal, Akka.Persistence""
+                    plugin-dispatcher = ""akka.actor.default-dispatcher""
+                }}
+                akka.persistence.journal.memory-journal-shared {{
+                    class = ""Akka.Cluster.Sharding.Tests.MemoryJournalShared, Akka.Cluster.Sharding.Tests.MultiNode""
+                    plugin-dispatcher = ""akka.actor.default-dispatcher""
+                    timeout = 5s
+                }}
+                """);
+        }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="mode">mode the state store mode</param>
+        /// <param name="rememberEntities">rememberEntities defaults to off</param>
+        /// <param name="additionalConfig">additionalConfig additional config</param>
+        /// <param name="loglevel">loglevel defaults to INFO</param>
+        protected MultiNodeClusterShardingConfig(
+            string mode = ClusterShardingSettings.StateStoreModeDData,
+            bool rememberEntities = false,
+            string additionalConfig = "",
+            string loglevel = "INFO")
+        {
+            Mode = mode;
+            RememberEntities = rememberEntities;
+            AdditionalConfig = additionalConfig;
+            Loglevel = loglevel;
+
+            Config persistenceConfig = (mode == ClusterShardingSettings.StateStoreModeDData) ? ConfigurationFactory.Empty : PersistenceConfig();
+
+            Config common =
+                ConfigurationFactory
+                    .ParseString($@"""
+                        akka.actor.provider = ""cluster""
+                        akka.cluster.downing-provider-class = akka.cluster.testkit.AutoDowning
+                        akka.cluster.testkit.auto-down-unreachable-after = 0s
+                        akka.cluster.sharding.state-store-mode = ""{mode}""
+                        akka.cluster.sharding.remember-entities = {rememberEntities}
+                        akka.cluster.sharding.fail-on-invalid-entity-state-transition = on
+                        akka.loglevel = {loglevel}
+                        akka.remote.log-remote-lifecycle-events = off
+
+                        akka.cluster.sharding {{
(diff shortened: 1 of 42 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -378,8 +440,11 @@ public ShardRegion(string typeName, Func<string, Props> entityProps, ClusterShar
             _replicator = replicator;
             _majorityMinCap = majorityMinCap;
 
-            _retryTask = Context.System.Scheduler.ScheduleTellRepeatedlyCancelable(Settings.TunningParameters.RetryInterval, Settings.TunningParameters.RetryInterval, Self, Retry.Instance, Self);
             SetupCoordinatedShutdown();
+
+            _retryInterval = Settings.TuningParameters.RetryInterval;
+            _initRegistrationDelay = TimeSpan.FromMilliseconds(100).Max(new TimeSpan(_retryInterval.Ticks / 2 / 2 / 2));
```
Aaronontheweb, line 446: why `_retryInterval.Ticks / 2 / 2 / 2` ?

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -718,25 +809,63 @@ private void HandleShardRegionCommand(IShardRegionCommand command)
...
 
+                    // we depend on the coordinator each time, if empty we need to register
+                    // otherwise we can try to deliver some buffered messages
                     if (_coordinator == null) Register();
                     else
                     {
-                        RequestShardBufferHomes();
+                        // Note: we do try to deliver buffered messages even in the middle of
+                        // a graceful shutdown every message that we manage to deliver is a win
+                        TryRequestShardBufferHomes();
```
Aaronontheweb, line 825: LGTM

src/core/Akka/Actor/ActorPath.cs
```diff
@@ -445,7 +445,7 @@ public string ToStringWithoutAddress()
         /// <inheritdoc/>
         public override string ToString()
         {
-            return ToStringWithAddress();
+            return $"{Address}{Join()}";
```
Aaronontheweb, line 448: Might need to check performance impact here - making a note

src/core/Akka/Actor/ActorRef.cs
```diff
@@ -569,6 +569,68 @@ public override bool IsLocal
         public override bool IsTerminated { get { return false; } }
     }
 
+
+    /// <summary>
+    /// An ActorRef that ignores any incoming messages.
```
Aaronontheweb, line 574: Very cool

src/core/Akka/Actor/ActorRefProvider.cs
```diff
@@ -53,6 +53,8 @@ public interface IActorRefProvider
         /// <summary>Gets the dead letters.</summary>
         IActorRef DeadLetters { get; }
 
+        IActorRef IgnoreRef { get; }
```
Aaronontheweb, line 56: LGTM

src/core/Akka.Tests/Actor/ActorRefIgnoreSpec.cs
```diff
@@ -0,0 +1,132 @@
...
+using Akka.Serialization;
+using Akka.TestKit;
+using Akka.TestKit.TestActors;
+using Xunit;
+using Akka.Util.Internal;
+using FluentAssertions;
+
+namespace Akka.Tests.Actor
+{
+    public class ActorRefIgnoreSpec : AkkaSpec, INoImplicitSender
```
Aaronontheweb, line 21: LGTM

src/core/Akka.Cluster.Tests/ActorRefIgnoreSerializationSpec.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        protected override void AfterAll()
+        {
+            base.AfterAll();
+            system1.Terminate();
+            system2.Terminate();
+        }
+
+        [Fact]
+        public void ActorSystem_IgnoreRef_should_return_a_serializable_ActorRef_that_can_be_sent_between_two_ActorSystems_using_remote()
```
Aaronontheweb, line 54: LGTM

Aaronontheweb (changes requested):
Have a handful of minor changes and comments

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/MultiNodeClusterShardingConfig.cs
```diff
@@ -0,0 +1,92 @@
...
+            AdditionalConfig = additionalConfig;
+            Loglevel = loglevel;
+
+            Config persistenceConfig = (mode == ClusterShardingSettings.StateStoreModeDData) ? ConfigurationFactory.Empty : PersistenceConfig();
+
+            Config common =
+                ConfigurationFactory
+                    .ParseString($@"""
+                        akka.actor.provider = ""cluster""
+                        akka.cluster.downing-provider-class = akka.cluster.testkit.AutoDowning
```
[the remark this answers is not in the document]
zbynek001: removed, leftover from akka code

src/contrib/cluster/Akka.Cluster.Sharding/Internal/AbstractLeastShardAllocationStrategy.cs
```diff
@@ -0,0 +1,147 @@
...
+{
+    using static Akka.Cluster.ClusterEvent;
+    using ShardId = String;
+
+    /// <summary>
+    /// Common logic for the least shard allocation strategy implementations
+    /// </summary>
+    public abstract class AbstractLeastShardAllocationStrategy : IActorSystemDependentAllocationStrategy
+    {
+        private static ImmutableHashSet<MemberStatus> JoiningCluster = ImmutableHashSet.Create(MemberStatus.Joining, MemberStatus.WeaklyUp);
```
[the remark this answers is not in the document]
zbynek001: changed

src/contrib/cluster/Akka.Cluster.Sharding/Internal/AbstractLeastShardAllocationStrategy.cs
```diff
@@ -0,0 +1,147 @@
...
+        private static ImmutableHashSet<MemberStatus> LeavingClusterStatuses = ImmutableHashSet.Create(MemberStatus.Leaving, MemberStatus.Exiting, MemberStatus.Down);
```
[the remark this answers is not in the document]
zbynek001: changed

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
zbynek001: this is to make the registration retry process faster at the beginning and then the retry interval is doubled until it matches the normal `_retryInterval` over 3 iterations

src/core/Akka/Actor/ActorPath.cs
zbynek001: should not, the previous `ToStringWithAddress` is doing the same inside

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/MultiNodeClusterShardingConfig.cs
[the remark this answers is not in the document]
Aaronontheweb: Ah, thanks for clarifying

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
Aaronontheweb: I see

PR #4603, opened 2020-11-03
Fix TestKitBase and Option<T>

### LINKED ISSUE #4496: Multiple tests run fails with Akka.TestKit 1.4.8 (worked fine with 1.4.7)
Hi,

Akka: 1.4.8, NUnit3 package, Windows 10 Pro, .NET Framework 4.7.2, using Resharper to run tests.

Just upgraded from 1.4.7 to 1.4.8 and noticed that tests are failing with:
```
System.NullReferenceException : Object reference not set to an instance of an object.
TearDown : System.NullReferenceException : Object reference not set to an instance of an object.
   at Akka.TestKit.TestKitBase.<>c.<InitializeTest>b__10_1(Config c)
   at Akka.Util.Option`1.Select[TNew](Func`2 selector)
   at Akka.TestKit.TestKitBase.InitializeTest(ActorSystem system, ActorSystemSetup config, String actorSystemName, String testActorName)
   at Akka.TestKit.NUnit3.TestKit.InitializeActorSystemOnSetUp()
--TearDown
   at Akka.TestKit.TestKitBase.Dilated(TimeSpan duration)
   at Akka.TestKit.TestKitBase.Shutdown(ActorSystem system, Nullable`1 duration, Boolean verifySystemShutdown)
   at Akka.TestKit.TestKitBase.Shutdown(Nullable`1 duration, Boolean verifySystemShutdown)
   at Akka.TestKit.NUnit3.TestKit.AfterAll()
```
tests pass if I run one at a time, but run fails with several tests selected within such run.

If Akka.TestKit is rolled back to 1.4.7 then everything's back to normal.

Thank you!

Arkatufus:
Closes #4496.

Original problem was that 
- `Option<T>` failed to report that it has no value when passed a null value.
- `TestKitBase` did not check for `Option<Config>.None`

### DIFF
--- a/src/core/Akka/Util/Option.cs
+++ b/src/core/Akka/Util/Option.cs
@@ -16,7 +16,7 @@ namespace Akka.Util
     /// Useful where distinguishing between null (or zero, or false) and uninitialized is significant.
     /// </summary>
     /// <typeparam name="T">TBD</typeparam>
-    public struct Option<T>
+    public readonly struct Option<T>
     {
         /// <summary>
         /// None.
@@ -30,7 +30,7 @@ namespace Akka.Util
         public Option(T value)
         {
             Value = value;
-            HasValue = true;
+            HasValue = value != null;
         }
 
         /// <summary>
@@ -81,16 +81,15 @@ namespace Akka.Util
             return mapper(Value);
         }
 
-        /// <inheritdoc/>
         public bool Equals(Option<T> other)
             => HasValue == other.HasValue && EqualityComparer<T>.Default.Equals(Value, other.Value);
 
         /// <inheritdoc/>
         public override bool Equals(object obj)
         {
-            if (ReferenceEquals(null, obj))
+            if (obj is null)
                 return false;
-            return obj is Option<T> && Equals((Option<T>)obj);
+            return obj is Option<T> opt && Equals(opt);
         }
         
         /// <inheritdoc/>
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.approved.txt
@@ -1292,7 +1292,9 @@ namespace Akka.Streams.Dsl
         public static Akka.Streams.Dsl.Flow<TIn, TOut2, TMat> Batch<TIn, TOut, TOut2, TMat>(this Akka.Streams.Dsl.Flow<TIn, TOut, TMat> flow, long max, System.Func<TOut, TOut2> seed, System.Func<TOut2, TOut, TOut2> aggregate) { }
         public static Akka.Streams.Dsl.Flow<TIn, TOut2, TMat> BatchWeighted<TIn, TOut, TOut2, TMat>(this Akka.Streams.Dsl.Flow<TIn, TOut, TMat> flow, long max, System.Func<TOut, long> costFunction, System.Func<TOut, TOut2> seed, System.Func<TOut2, TOut, TOut2> aggregate) { }
         public static Akka.Streams.Dsl.Flow<TIn, TOut, TMat> Buffer<TIn, TOut, TMat>(this Akka.Streams.Dsl.Flow<TIn, TOut, TMat> flow, int size, Akka.Streams.OverflowStrategy strategy) { }
+        [System.ObsoleteAttribute("Deprecated. Please use Collect(isDefined, collector) instead")]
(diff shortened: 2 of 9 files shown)


src/core/Akka/Util/Option.cs
```diff
@@ -30,7 +30,7 @@ public struct Option<T>
         public Option(T value)
         {
             Value = value;
-            HasValue = true;
+            HasValue = value != null;
```
IgorFedchenko, line 33: I was always thinking about that - but what if `null` in a valid value for this optional thing? What if `null` means "there is a value, it is obtained from external source/whatever, it is initialized, and it is `null` right now"?

Arkatufus: No, by definition Option<T> will return None or Some, what is the point of using Option if you still allow null to leak through? you wouldn't fix the original NRE problem.

Arkatufus: Even with this, you'll still have NRE problem if you forgot to check for None.

Arkatufus: In essence, Option is a half assed attempt to fix NRE problem, its just moving the problem somewhere else.

Arkatufus: If you still need a null, then what you need is a tri-state object, where its state can be null, none, and some

IgorFedchenko: And this will mean that `new Option(default(int)).HasValue == true` but `new Option(default(object)).HasValue == false`, which may make `Option` class less obvious... 
I realize the intention to void `new Option(null).HasValue` been `true`, but just thinking - what if this is a feature, not a bug? Besides, this may become a pretty tricky breaking change...

Just need to think twice before merging this change I think.

Arkatufus: BTW, the scala implementation actually throws a `NoSuchElementException` if you try to access a null value, which i think the right implementation... but changing that would probably break binary compatibility

Arkatufus: Our implementation is flawed anyway, Option<T> is supposed to be an IOption<T> with Some<T> and None<T> concrete classes

IgorFedchenko: Yeah, I agree with all of that, just thinking about how many things we can break because people are passing null (and how often they are doing this by design)

Arkatufus: No idea, its broken from the start

Arkatufus: Yeah, it even broke our own code.
This is stupid, I feel like I'm chasing my own tail.

IgorFedchenko: This is a good fix actually, I would love to see exception thrown here too - but maybe this should be done on some breaking change release, i.e. Akka.NET 1.5, or even 2.0, because chances are that many users will need to debug their app and fix things.

Arkatufus: I think the proper thing to do right now is to implement the right `IOption<T>` interface but make it internal and keep the old `Option<T>` struct for public consumption.
But this will require so so many work to convert every `Option<T>` in our code x__x

Aaronontheweb: Why do we need the `IOption<T>` interface exactly? I get the `Some<T>` and `None<T>`

Aaronontheweb (commented):
Left some comments

Arkatufus:
We needed `IOption<T>` to simulate the scala implementation to the T, it would make Option unambiguous, a None is always a None and a Some will always contain something, it makes it save to do a typed switch on the type of the returned value.

We don't need it per se, but by implementing it we can separate the internal implementation of Option with the one that is being used by user, just in case changing Option will break someone elses code. Pretty much a bandaid with backward compatibility.

Arkatufus:
But it will be a very big bandaid, akin to using a nuclear warhead to kill an ant.

Arkatufus:
The big fear right now is that, even changing that line in `Option` might break someone elses code.
It already break ours, we're also (ab)using `Option` as a tri-state return value inside the `Collect` stream dsl.

Aaronontheweb:
As long as:

1. It doesn't take long to implement;
2. Doesn't introduce additional complexities / issues in terms of API compatibility; and
3. Doesn't introduce lots of new allocations.

Then I think we should do it "the right way"

Arkatufus:
1. Yes, it will take awhile to implement.
2. We can change it as carefully as possible so that it doesn't change the public API
3. It will probably have some memory impact.

Arkatufus:
Memory impact will most probably came from stack allocation, I think struct was chosen exactly because of this.

Aaronontheweb:
> Memory impact will most probably came from stack allocation, I think struct was chosen exactly because of this.

that's fine then

Aaronontheweb:
as long as we don't allocate on every `Option.[name of read method]`

PR #4616, opened 2020-11-11
Docs: Programmatic Akka.NET Configuration

Aaronontheweb:
* [x] Programmatic serialization configuration
* [x] General explanation of `ActorSystemSetup` and `BootstrapSetup`

PR #4617, opened 2020-11-11
removed InternalApi attribute from ActorSystemSetup

Aaronontheweb:
ActorSystemSetup is part of the public API and is used in the creation of `ActorSystem`s. It should not be marked as `InternalApi`.


Aaronontheweb:
This is a non-breaking change. But it does affect how DocFx API documentation is generated - noticed this when I was working on https://github.com/akkadotnet/akka.net/pull/4616 earlier

PR #4618, opened 2020-11-11
if not specified in HOCON, set AppVersion to executing assembly version

Aaronontheweb:
just so users don't have to manage application versions in two places -have Akka.Cluster.ClusterSettings automatically use the entry assembly's 'Major.Minor.Revision' version - that way if the developers are automatically bumping their entry executable's version, that is what will be reflected in HOCON.

If the user takes the time to manually input their akka.cluster.app-version - that value will be used instead.

cc @zbynek001


Aaronontheweb:
Looks like we have a compilation failure here after 35 minutes? That's weird...

Aaronontheweb:
Looks like we timed out while compiling the multi-node test runner:

https://dev.azure.com/dotnet/Akka.NET/_build/results?buildId=41558&view=logs&j=a29a2371-0793-5e14-39b9-20eb0beef0db&t=3641e6b2-7849-5b4b-3026-a175bcdbe12a&l=2722

Aaronontheweb:
looks like the hung build was a fluke - ran fine the second time

Aaronontheweb:
Looks like there's an error with how the revision number gets used in this PR - but so far this feature is working very nicely.

zbynek001:
Not sure, but it might be better to hide this functionality behind some config switch.
In some scenarios, e.g. when using different assemblies joining the same cluster, it might have some unexpected behavior

Aaronontheweb:
> ter to hide this functionality behind some config switch.
> In some scenarios, e.g. when using different assemblies joining the same cluster, it mig

good point - I'll change the configuration around and to the following:

1. Have a `app-version = assembly` config by default, which executes the app-version-from-assembly behavior;
2. If that value is overridden with anything else, the explicit config value gets used instead.

Does that seem like a reasonable compromise?

src/core/Akka.Cluster/Configuration/Cluster.conf
```diff
@@ -92,7 +92,16 @@ akka {
...
+    #
+    # DEFAULT: by default the app-version will default to the entry assembly's version,
+    # i.e. the assembly of the executable running `Program.cs`
+    # 
+    # Values can be "assembly-version" or a version string as defined above, i.e.
+    # app-version = "1.0.0"
+    # app-version = "1.1-beta1"
+    # app-version = "1"
+    # app-version = "1.1"
+    app-version = assembly-version
```
Aaronontheweb, line 104: Changed the default value here to `assembly-version`, which activates this assembly-scanning default behavior

src/core/Akka/Util/AppVersion.cs
```diff
@@ -32,20 +34,50 @@ namespace Akka.Util
...
         {
-            this.Version = version;
+            Version = version;
         }
 
         public static AppVersion Create(string version)
         {
-            var v = new AppVersion(version);
+            // check to see if we're going to use the assembly-version
+            if (version.Equals(AssemblyVersionMarker, StringComparison.InvariantCultureIgnoreCase))
```
Aaronontheweb, line 54: Check to see if we're using `assembly-version` - if so, get the assembly data. If not, parse the version string as usual.

src/core/Akka/Util/AppVersion.cs
```diff
@@ -32,20 +34,50 @@ namespace Akka.Util
...
+        /// </remarks>
+        internal static AppVersion AppVersionFromAssemblyVersion()
+        {
+            // user hasn't specified AppVersion in HOCON
+            // try looking it up via assembly
+            var entryAssembly = Assembly.GetEntryAssembly();
+
+            // if the entryAssembly is null (which can happen when we're called from unmanaged code)
+            // then fall back to the executing assembly for the version number.
+            var targetAssembly = entryAssembly ?? Assembly.GetExecutingAssembly();
```
Aaronontheweb, line 77: Grab the `EntryAssembly` preferably. If this value is `null`, which can happen when called from unmanaged code (this happens when the XUnit test runner is executing from within visual studio, for instance) then fallback to the `ExecutingAssembly` version. Shouldn't happen in most Akka.NET applications, which tend to be launched as 100% managed processes.

Aaronontheweb (commented):
Left some comments describing the most recent round of changes

Arkatufus (approved):
Looks good

PR #4633, opened 2020-11-16
fixed documentation errors with Akka.Docts.Tests ReceiveTimeoutSpecs

Aaronontheweb:


src/core/Akka.Docs.Tests/Actors/ReceiveTimeoutSpecs.cs
```diff
@@ -51,7 +51,7 @@ protected override void PreStart()
                 Context.SetReceiveTimeout(TimeSpan.FromMilliseconds(100));
             }
         }
-        // <ReceiveTimeoutActor>
+        // </ReceiveTimeoutActor>
```
Aaronontheweb, line 54: Didn't close the `ReceiveTimeoutActor` section correctly, so this caused the documentation to render incorrectly here

PR #4638, opened 2020-11-19
added .YML and .YAML mime types to IIS

Aaronontheweb:
Need this so other applications that build on top of Akka.NET can reference Akka.NET's `xrefmap` via DocFx remote references: https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html#advanced-more-options-for-cross-reference

### DIFF
--- a/docs/web.config
+++ b/docs/web.config
@@ -3,6 +3,8 @@
     <system.webServer>
         <staticContent>
             <mimeMap fileExtension=".json" mimeType="application/json" />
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Testing if this works on another DocFx website first, before deploying to Akka.NET's website

Aaronontheweb:
Works great - excellent. This change will make it possible to reference Akka.NET types and content via DocFx in other websites, such as the Alpakka website can refer to Akka.NET's types and cross link automatically.

PR #4642, opened 2020-11-28
Add Akka.Persistence.Linq2Db and corresponding unit tests to /contrib/

to11mtm:
This commit would put Akka.Persistence.Linq2Db into the main repo under /contrib/

I'm honestly not sure whether this should live here or elsewhere, since on one hand this is where Sql.Common lives (which this could replace long term) but most of the implementations live in other repositories.

# Advantages over SQL.Common:

  - One library to rule them all. Linq2Db generates agnostic SQL and lets us even handle the type conversions when DBs have specific behavior (for example, SQLite using int64s as times in snapshot).

  - Based on akka-persistence-jdbc, using Akka.Streams for Batched Writes
    - By being based on and close to the JVM version, we may stand on the shoulders of giants.
    - Main differences:
       - We handle SerializerId and Manifest differently, they are just now dealing with some of these challenges on their end.
       - Pipeline is optimized for .NET Environment based on profiling and benchmarks.

# Disadvantages:
 
 - Main one that comes to mind is that some of the persistence queries are not implemented. This can of course be changed in future PRs.

# Performance

Tests based on i-7 8750H, 32GB Ram, 2TB SSD, Windows 10 version Version 10.0.19041.630.
Databases running on Docker WSL2.

All numbers are in msg/sec. 

|Test |SqlServer (normal)       |  SqlServer Batching     | Linq2Db     |vs Normal| vs Batching| 
|:-------------  |:------------- | :----------: | -----------: | -----------: | -----------: |
|Persist |  164|427| 235|143.29%|55.04%|
|PersistAll |  782|875| 5609|717.26%|641.03%|
|PersistAsync |  630|846| 16099|2555.40%|1902.96%|
|PersistAllAsync |  2095|902| 15681|748.50%|1738.47%|
|PersistGroup10 |  590|680| 1069|181.19%|157.21%|
|PersistGroup100 |  607|965| 5537|912.19%|573.78%|
|PersistGroup200 |  628|1356| 7966|1268.47%|587.46%|
|PersistGroup25 |  629|675| 2189|348.01%|324.30%|
|PersistGroup400 |  612|1011| 7237|1182.52%|715.83%|
|PersistGroup50 |  612|654| 3867|631.86%|591.28%|
|Recovering |  41903|38766| 42592|101.64%|109.87%|
|Recovering8 |  75466|65515| 63960|84.75%|97.63%|
|RecoveringFour |  59259|51355| 58437|98.61%|113.79%|
|RecoveringTwo |  41745|35512| 41108|98.47%|115.76%|


src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs
```diff
@@ -0,0 +1,189 @@
...
+        {
+            try
+            {
+                _journalConfig = new JournalConfig(config);
+                _mat = Materializer.CreateSystemMaterializer((ExtendedActorSystem)Context.System,
+                    ActorMaterializerSettings.Create(Context.System)
+                        .WithDispatcher(_journalConfig.MaterializerDispatcher)
+                    ,
+                    "l2dbWriteJournal"
+                );
```
to11mtm, line 64: Since persistence plugins run under /system/ , we should run Materializer there as well; otherwise shutdowns can get messy (shutting /user/ down first will kill the stream.)

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -0,0 +1,421 @@
...
+        BaseJournalDaoWithReadMessages,
+        IJournalDaoWithUpdates
+    {
+
+        public ISourceQueueWithComplete<WriteQueueEntry> WriteQueue;
+        protected JournalConfig _journalConfig;
+        protected FlowPersistentReprSerializer<JournalRow> Serializer;
+
+        private Lazy<object> logWarnAboutLogicalDeletionDeprecation =
+            new Lazy<object>(() => { return new object(); },
```
to11mtm, line 33: TODO: This is supposed to log and warn that logical delete is deprecated... but only do so once

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -0,0 +1,421 @@
...
+                            await transaction.RollbackAsync();
+                        }
+                        catch (Exception )
+                        {
+                            //If rollback fails, Don't throw as it will mask
+                            //original exception
+                        }
+                        
+                        throw;
+                    }
```
to11mtm, line 286: Should we really be async here? and/or make async optional?

Concern is that (with SQLServer, SQLite, anything else with lots of locking) we don't know how quickly we will get back to this task chain under load, leaving the table(s) locked even longer.

I can say from experience running the driver live, this is a thing; turning off delete compatibility (delete compatibility increases the db ops here) solved a bunch of stability issues on a project I've been working on.

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/DAO/BaseByteArrayJournalDao.cs
```diff
@@ -0,0 +1,421 @@
...
+                {
+                    query = query.Take((int) max);
+                }
+
+                
+                //TODO: Is there a better way to do this async?
+                //return Source
+                //    .FromObservable(query.AsAsyncEnumerable().ToObservable())
+                
+                return Source.From(await query.ToListAsync())
```
to11mtm, line 397: FWIW, there is potential to use -real- IAsyncEnumerable here in future, Linq2Db supports it, I just couldn't find an abstraction in Akka.Streams to utilize with actual benefit

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs
```diff
@@ -0,0 +1,189 @@
...
+                    {
+                        Context.GetLogger().Warning(e,
+                            "Unable to Initialize Persistence Journal Table!");
+                    }
+
+                }
+            }
+            catch (Exception ex)
+            {
+                Context.GetLogger().Warning(ex,"Unexpected error initializing journal!");
```
to11mtm, line 96: Probably need to `throw;`

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs
```diff
@@ -0,0 +1,189 @@
...
+                catch (Exception)
+                {
+                    //We don't have 'Recover' in C# so this is intentionally empty
+                    //Basically we just wanted to wait for write to succeed OR fail
+                }
+                var hsn =await _journal.HighestSequenceNr(persistenceId,
+                    fromSequenceNr);
+                return hsn;
+            }
+            return await _journal.HighestSequenceNr(persistenceId, fromSequenceNr);
```
to11mtm, line 157: Honestly, not sure if this was the right/best way to port intent, original [here](https://github.com/akka/akka-persistence-jdbc/blob/master/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala#L104-L113)

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs
```diff
@@ -0,0 +1,189 @@
...
+            
+            //When we are done, we want to send a 'WriteFinished' so that
+            //Sequence Number reads won't block/await/etc.
+            future.ContinueWith((p) =>
+                    self.Tell(new WriteFinished(persistenceId, future)),
+                CancellationToken.None,
+                TaskContinuationOptions.ExecuteSynchronously,
+                TaskScheduler.Default);
+            //But we still want to return the future from `AsyncWriteMessages`
+            return await future;
```
to11mtm, line 180: May want to Double-Check Porting from [here](https://github.com/akka/akka-persistence-jdbc/blob/master/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala#L94-L98)

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/Types/JournalRow.cs
```diff
@@ -0,0 +1,27 @@
...
+        
+        public long Timestamp { get; set; } = 0;
+
+        public bool deleted { get; set; }
+        [Column(Configuration = ProviderName.SQLite, IsPrimaryKey = false)]
+        [Column(IsPrimaryKey = true, CanBeNull = false)]
+        public string persistenceId { get; set; }
+        [Column(Configuration = ProviderName.SQLite, IsPrimaryKey = false)]
+        [Column(IsPrimaryKey = true)]
+        public long sequenceNumber { get; set; }
```
to11mtm, line 20: in SQLite, if you have an auto-increment sequence, it MUST be the primary key, so this is a tiny bit hacky since linq2db doesn't let us specify a unique constraint

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Streams/ExtSeq.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Threading.Tasks;
+using Akka.Streams.Dsl;
+using LanguageExt;
+
+namespace Akka.Persistence.Sql.Linq2Db.Streams
+{
+    public static class ExtSeq
+    {
+        public static Sink<TIn, Task<Seq<TIn>>> Seq<TIn>() => Sink.FromGraph(new ExtSeqStage<TIn>());
+    }
```
to11mtm, line 10: At some point benchmarks indicated Language.Ext's `Seq<A>` was far better for this use case than other immutable structures (mostly around concatenating/etc,) so we are using it here in a custom graph/flow.

Aaronontheweb:
@ismaelhamed are you interested in reviewing this? If not I'll see about making some time to do it

ismaelhamed:
@Aaronontheweb I'd love to, but I'm not sure when I'm going to have the time and I don't want to hold this PR.

One thing though: like @to11mtm  mentioned, there's a big change in the current schemas (both Journal and Snapshot) coming up in the next version (see [#318](https://github.com/akka/akka-persistence-jdbc/issues/318)). But they're still figuring out how to best migrate current users (I think they have something that leverages [Flyway](https://flywaydb.org/)) so, if we release this before having those changes integrated too, IMO we should do it as a beta and with a big warning of not using it on production just yet.

to11mtm:
> One thing though: like @to11mtm mentioned, there are some changes in the current schemas (both Journal and Snapshot) coming up in the next version (see #318). But they're still figuring out how to best migrate current users (I think they have something that leverages Flyway) so, if we release this before having those changes integrated too, IMO we should do it as a beta and with a big warning of not using it on production just yet.

Based on discussions, I believe the intent here would be to provide an 'opt-in' path over a longer time period. This way we can ensure that all DB providers are properly supported (i.e. no tests for Oracle or MySql compatibility or Migration exist yet, I'm sure they would each be simple PRs though), ensure we have feature parity for Persistence Queries, and give people time to migrate/upgrade.

At this point I can say it works in a production environment quite well :), with the main caveat that `delete-compatibility-mode` wasn't the best under high throughput/concurrency scenarios on SQL Server due to how it does page/table locks (hence the notes I put on the PR.)

With regard to their schema changes, we were already in an 'in-between' state with our Persistence layout; The schema changes are to add:

  - ~Serializer ID/Manifests~ (We already did this in .NET side, so the port already has this functionality)

  - Event Manifest (at first glance I'm not entirely certain whether this is the event adapter used, or the `.Manifest` from our `IPersistentRepresentation`)

  - Writer UUID (Just a GUID, I believe general intent is to know what actor/incarnation wrote to the journal; I know this sort of thing is helpful for spotting racy cluster handoffs).

So the only 'migration' we will likely have to worry about will be for Event Manifest. My gut says that we could write a migration tool easily enough there.

to11mtm:
@Aaronontheweb I've gone ahead and added Specs that show Snapshot Compatibility with:

 - SQLite
 - SQL Server
 - PostgreSQL

I've also added Journal Compatibility tests for SQLite, giving us Journal compatibility with:

 - SQLite
 - SQL Server
 - PostgreSQL

The intent of each series of tests is to ensure that Journals/Snapshots are cross-compatible at this stage, and users can go 'both ways' without worrying about backout plans.

to11mtm:
Looks like the tests may be failing on the build server because of the docker usage.

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs
to11mtm: done.

src/contrib/persistence/Akka.Persistence.Sql.Linq2Db/Journal/Types/JournalRow.cs
to11mtm: This has been changed now to be in the mapping

Aaronontheweb:
> Looks like the tests may be failing on the build server because of the docker usage.

@IgorFedchenko could you take a look at that?

IgorFedchenko:
I will take a closer look at this tomorrow, but seems like it is `net-core-tests-windows` failing when trying to use docker images.
 Like `postgres:latest` or `mcr.microsoft.com/mssql/server:latest`.

For example, the mssql server container seems to be linux only (from [it's page](https://hub.docker.com/_/microsoft-mssql-server)). It feels like `postgres` is a linux-based container too. But when Azure CI is running on windows, it is using "windows containers" mode - so most likely that is why it fails to restore containers.

@to11mtm Could this be the case? Can you run this tests locally with "windows containers" mode selected in Docker Desktop? Because if so, we will need to either select another image names depending on OS which is running the tests, or use some attribute like this to make selected tests skipped under windows:
```csharp
public class UnixFact : FactAttribute
{
	private string _skip;

	/// <inheritdoc />
	public override string Skip
	{
		get
		{
			var isUnix = Environment.OSVersion.Platform == PlatformID.Unix;
			return isUnix ? _skip : SkipWindowsReason ?? "Skipped under Unix platforms";
		}
		set => _skip = value;
	}
	
	public string SkipWindowsReason { get; set; }
}
```

Or move this tests to separate assembly and make it running only under unix with `isWindows` flag usage at `build.fsx` script's `RunTests` target.

to11mtm:
> 
> 
> I will take a closer look at this tomorrow, but seems like it is `net-core-tests-windows` failing when trying to use docker images.
> Like `postgres:latest` or `mcr.microsoft.com/mssql/server:latest`.
> 
> For example, the mssql server container seems to be linux only (from [it's page](https://hub.docker.com/_/microsoft-mssql-server)). It feels like `postgres` is a linux-based container too. But when Azure CI is running on windows, it is using "windows containers" mode - so most likely that is why it fails to restore containers.
> 
> @to11mtm Could this be the case? Can you run this tests locally with "windows containers" mode selected in Docker Desktop? Because if this is the reason, we will need to either select another image names depending on OS which is running the tests, or use some attribute like this to make selected tests skipped under windows:
> 
> ```cs
> public class UnixFact : FactAttribute
> {
> 	private string _skip;
> 
> 	/// <inheritdoc />
> 	public override string Skip
> 	{
> 		get
> 		{
> 			var isUnix = Environment.OSVersion.Platform == PlatformID.Unix;
> 			return isUnix ? _skip : SkipWindowsReason ?? "Skipped under Unix platforms";
> 		}
> 		set => _skip = value;
> 	}
> 	
> 	public string SkipWindowsReason { get; set; }
> }
> ```
> 
> Or move this tests to separate assembly and make it running only under unix with `isWindows` flag usage at `build.fsx` script's `RunTests` target.

This is all sounding -very- familiar to when I was trying to get the docker tests running locally (As they are based on the existing persistence modules.) If this is the case I'll work on determining the best way to have these only run under unix OS, I think it will have to be the assembly route since they are same base tests, just different harnesses.

to11mtm:
I've separated out the Docker dependent tests and updated build.fsx appropriately. 

I also kept SQLite separate since that doesn't require docker and should give us a good enough confidence in the other providers working under windows/framework.

For some reason it looks like the SQLite tests didn't run at all on Framework though, I'll have to check that (unless someone wants to just do a magic poke ;)

I did also notice in this that I somehow failed to bring the Persistence.Query tests over from my repo. I'll do that by EOW, but the rest of the code will not change (Tests were already passing, I just have to copy them back over).

build.fsx
```diff
@@ -233,6 +233,8 @@ Target "RunTests" (fun _ ->
         let rawProjects = match (isWindows) with
                             | true -> !! "./src/**/*.Tests.*sproj"
                                       ++ "./src/**/Akka.Streams.Tests.TCK.csproj"
+                                      -- "src/**/*.Tests.Docker.*sproj"
+                                      -- "src/**/*.*Tests.Docker.*sproj"
```
IgorFedchenko, line 237: Yep - exactly what I was talking about 👍 
@to11mtm One question - to some people, `Tests.Docker` may not meen that tests will be running under Unix-only systems by default. Should we use this as a convention for Akka.NET that `Tests.Docker` is always something running under Unix only? Or renaming it to `Tests.Unix` or something like that will be more intuitive for other contributors?

IgorFedchenko:
Congradulations @to11mtm , seems like Docker issue is resolved! 🍾 

Left you a minor note about the new tests assembly naming convention.

to11mtm:
> 
> 
> Congradulations @to11mtm , seems like Docker issue is resolved! 🍾
> 
> Left you a minor note about the new tests assembly naming convention.

Thanks! Do you think something like `DockerLinux` be descriptive enough? My worry is calling the tests `.Unix` may cause people to think they are only meaningful if you want to run on *nix.


Also, Persistence.Query SQLite tests seem angry on the build server. I'm guessing it's related to the lifecycle of the inmem journal, will try to look at sometime today after work.

Aaronontheweb:
Might be a better idea to stick this in the Akka.Persistence.Sql repository since that's already equipped for spinning up SQL Server during CI. Any objection to that?

IgorFedchenko:
> Do you think something like DockerLinux be descriptive enough? 

I guess so. Unless you will move all of this to Akka.Persistence.Sql  repo as @Aaronontheweb suggested.

to11mtm:
@IgorFedchenko I think based on last nights discussions on gitter chat this will be moving to it's own repo; moving to Akka.Persistence.SqlServer would still be a little suboptimal/confusing in that we also have postgres tests in this codebase

Aaronontheweb:
Resolved via https://github.com/akkadotnet/Akka.Persistence.Linq2Db/pull/1

Issue #4641, opened 2020-11-27
Akka.IO.UdpExt.Manager: OverflowException when sending UDP packets to terminated clients


motmot80:
**Akka.Net Version**: v1.4.8
**Platform**: .NET core 3.1 (windows)

**Summary**:
The UDP actor is sometimes crashing when sending data to UDP clients which already have been terminated.
Sometimes this is followed by an `OverflowException`. Sometimes by an `ArgumentException`. Sometimes I receive empty messages from the UDP actor.

**Analysis**:
While the error occured, I recorded the udp traffic with wire shark. 
The exception is occuring when the server tries to send a UDP packet to a non existing client and then receives an ICMP response **"Destination unreachable (Port unreachable)"** (Type:3, Code:3). And according to the wire shark trace there was no inbound traffic except the ICMP port unreachable packet. The empty messages I received from the UDP actor are also matching timestamps from the ICMP port unreachable packets.

_Hint: The **original** UDP message itself is **included** within the ICMP response to the server._

```
127.0.0.1 127.0.0.1 UDP 44 6666 -> 55023
127.0.0.1 127.0.0.1 ICMP 72 Destination unreachable (Port unreachable) [Afterwards the original UDP message is included]
```

I am not sure if this is really causing the issue and I know this is far from obvious, but according to the call stack (see exceptions below) and the empty messages received - for whatever reason the listener on the port seems to react to the ICMP unreachable packets including the original UDP packets. Interestingly wire shark does something similar: It's showing the associated ICMP unreachable packets even though I filtered "_udp and udp.port == 6666_".

**Reproduction**:
1. Bind the UDP manager to a local port f.e. 6666
2. Start a client listening to a local high port fe. 52023
3. Send a UDP packet to the client 6666 => 52023
4. Kill the client 
5. Send another UDP packet to the 6666 => 52023
5.1 (Wait for an ICMP port unreachable packet)

**Expected**:
UDP actor is not crashing.

**Actual**:
UDP actor is crashing sporadically.

**ArgumentException:**
```
[ERROR][27.11.2020 21:57:52][Thread 0036][akka://SYSTEMNAME/system/IO-UDP-FF/$a] Provided length [1585734688] of array to copy doesn't fit array length [256000] within given offset [254976] (Parameter 'count')
Cause: System.ArgumentException: Provided length [1585734688] of array to copy doesn't fit array length [256000] within given offset [254976] (Parameter 'count')
[...]

PR #4643, opened 2020-11-28
Fix: UdpListener does not react properly on SocketError != Success (Issue 4641)

motmot80:
Fix for https://github.com/akkadotnet/akka.net/issues/4641

### DIFF
--- a/src/core/Akka/IO/UdpListener.cs
+++ b/src/core/Akka/IO/UdpListener.cs
@@ -118,7 +118,15 @@ namespace Akka.IO
         {
             try
             {
-                handler.Tell(new Received(ByteString.CopyFrom(e.Buffer, e.Offset, e.BytesTransferred), e.RemoteEndPoint));
+                if (!IsICMPError(e))
+                {
+                    if (e.SocketError != SocketError.Success)
+                        throw new SocketException((int)e.SocketError);
+
+                    handler.Tell(new Received(ByteString.CopyFrom(e.Buffer, e.Offset, e.BytesTransferred),
+                        e.RemoteEndPoint));
+                }
+
                 ReceiveAsync();
             }
             finally
@@ -127,6 +135,21 @@ namespace Akka.IO
                 Udp.SocketEventArgsPool.Release(e);
                 Udp.BufferPool.Release(buffer);
             }
+        }
+
+        /// <summary>
+        /// Checks if the socket event is an ICMP error message.
+        /// </summary>
+        /// <seealso href="https://tools.ietf.org/html/rfc1122#page-78"/>
+        private bool IsICMPError(SocketAsyncEventArgs e)
+        {
+            if (e.SocketError == SocketError.ConnectionReset)
+            {
+                Log.Debug("Ignoring client connection reset.");
+                return true;
+            }
+
+            return false;
         }
 
         /// <summary>
@@ -151,6 +174,7 @@ namespace Akka.IO
         private void ReceiveAsync()
         {
             var e = Udp.SocketEventArgsPool.Acquire(Self);
+            
             var buffer = Udp.BufferPool.Rent();
             e.SetBuffer(buffer.Array, buffer.Offset, buffer.Count);
             e.RemoteEndPoint = Socket.LocalEndPoint;
--- a/src/core/Akka.Tests/IO/TcpListenerSpec.cs
+++ b/src/core/Akka.Tests/IO/TcpListenerSpec.cs
@@ -28,7 +28,7 @@ namespace Akka.Tests.IO
         { }
 
         [Fact]
-        public void A_TCP_Listner_must_let_the_bind_commander_know_when_binding_is_complete()
+        public void A_TCP_Listener_must_let_the_bind_commander_know_when_binding_is_complete()
         {
             new TestSetup(this, pullMode: false).Run(x =>
             {
@@ -37,7 +37,7 @@ namespace Akka.Tests.IO
         }
 
         [Fact]
-        public void A_TCP_Listner_must_continue_to_accept_connections_after_a_previous_accept()
+        public void A_TCP_Listener_must_continue_to_accept_connections_after_a_previous_accept()
         {
             new TestSetup(this, pullMode: false).Run(x =>
             {
@@ -49,17 +49,17 @@ namespace Akka.Tests.IO
         }
 
         [Fact]
-        public void A_TCP_Listner_must_react_to_unbind_commands_by_replying_with_unbound_and_stopping_itself()
+        public void A_TCP_Listener_must_react_to_unbind_commands_by_replying_with_unbound_and_stopping_itself()
         {
             new TestSetup(this, pullMode:false).Run(x =>
             {
                 x.BindListener();
 
                 var unbindCommander = CreateTestProbe();
-                unbindCommander.Send(x.Listner, Tcp.Unbind.Instance);
+                unbindCommander.Send(x.Listener, Tcp.Unbind.Instance);
 
                 unbindCommander.ExpectMsg(Tcp.Unbound.Instance);
-                x.Parent.ExpectTerminated(x.Listner);
+                x.Parent.ExpectTerminated(x.Listener);
             });    
         }
 
@@ -107,7 +107,7 @@ namespace Akka.Tests.IO
                 new Socket(_endpoint.AddressFamily, SocketType.Stream, ProtocolType.Tcp).Connect(_endpoint);
             }
 
-            public IActorRef Listner { get { return _parentRef.UnderlyingActor.Listner; } }
+            public IActorRef Listener { get { return _parentRef.UnderlyingActor.Listener; } }
 
             public TestProbe SelectorRouter
             {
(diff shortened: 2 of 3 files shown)


Arkatufus:
I think the original code is suspect to begin with, there are some execution path that are not accounted for and a suspect try...finally block

src/core/Akka/IO/UdpListener.cs
```diff
@@ -118,6 +118,12 @@ private void DoReceive(SocketAsyncEventArgs e, IActorRef handler)
         {
             try
             {
+                if (e.SocketError != SocketError.Success)
```
Arkatufus, line 121: Can you move this closer to the source of the error?
In this case, this should probably moved to the `if (!Socket.ReceiveFromAsync(e))` line inside the `ReceiveAsync()` method.

Arkatufus (changes requested):
You need to also change the `CoreAPISpec.ApproveCore.approved.txt` file inside the `src\core\Akka.API.Tests` folder. Details can be seen in this [documentation](https://getakka.net/community/public-api-changes.html).

src/core/Akka/IO/UdpListener.cs
Arkatufus: Should also note that the original Scala code simply throws when any socket error occured. We should do the same and let the socket actor parent/guardian decide what to do with the failing socket actor.

motmot80: By closing the udp socket actor it will be vulnerable to different denial of service attacks. In my case the issue was caused by the underlying network just telling the **receiver** socket that his **sender** socket had a problem reaching a udp client. Underneath this was an ICMP packet sent from the client to the server _"Dear sender listening on port x the port you are trying to reach isn't available - stop sending data"_. This is an external hint and no unrecoverable internal error.

Would you mind providing a link to the original akka scala source? I didn't find it quickly within the original akka udp implementation.

Thanks in advance and best regards

motmot80:
I moved the validation a bit deeper to the source of the error until we find a more suitable solution fitting the original akka code but not causing any vulnerability by stopping the actor in case of client side ICMP error message.

> UDP MUST pass to the application layer all ICMP error
> messages that it receives from the IP layer.  Conceptually
> at least, this may be accomplished with an upcall to the
> ERROR_REPORT routine

https://tools.ietf.org/html/rfc1122#page-78

I am not sure how this `ERROR_REPORT routine` is working in scala. In .NET this is accomplished by using the `System.Net.Sockets.SocketError`.

Arkatufus:
> I am not sure how this ERROR_REPORT routine is working in scala. In .NET this is accomplished by using the System.Net.Sockets.SocketError

In Scala, they're throwing either `SocketException` or `IOException`, depending on where the actual error occured.

System.Net.Sockets.SocketError is basically just a bunch of long constants converted into an enum. We could create a new exception that encapsulate this or just use `SocketException` and pass the long constant as a parameter, but it is unintuitive.

src/core/Akka/IO/Udp.cs
```diff
@@ -76,6 +76,16 @@ public SocketConnected(SocketAsyncEventArgs eventArgs)
             }
         }
 
+        internal sealed class SocketFailed : SocketCompleted
+        {
+            public readonly SocketAsyncEventArgs EventArgs;
+
+            public SocketFailed(SocketAsyncEventArgs eventArgs)
```
Arkatufus, line 83: It just occured to me that it is bad to pass `SocketAsyncEventArgs` around like this because it actually contains the buffer being used by `Socket.Receive()`. Doing this can cause memory leak later on.

Arkatufus (changes requested):

src/core/Akka/IO/Udp.cs
motmot80: This was just a copy from the similar implementations above. I didn't question the existing command implementations.

motmot80:
Unfortunatly I lost track a bit. 

Where I came from: I have a running solution with an UDP server actor crashing often when clients are not reachable any more. This is because the UDP implementation isn't handling ICMP error messages in a meaningful way. Right now it is randomly allocating memory in this cases, crashing and leaving the socket allocated.

If the solution is to stop the actor (unbind) everytime such an ICMP error reaches the socket this solution is of no use for me. It may be comprehensible from an actor point of view but it is not a resilient internet ready UDP solution restarting the server socket every time one of many clients have a little network complaint.

I just wanted to help you out with a quick fix. Refactoring some of the UDP networking code would currently just overwhelm me.

motmot80:
Unfortunatly I didn't find any code within the original akka UDP implementation throwing `SocketException `or `IOExceptions`. The `UdpListener` is just reading buffers and I wouldn't expect java to throw a `SocketException` or `IOExceptions` in this scenario.
SO_ERROR is used for asynchronous errors not related directly to active operations.

src/core/Akka/IO/UdpListener.cs
Arkatufus: The problem is that `System.Net.Sockets.SocketError` does not just report ICMP error, it also reports a host of other error as outlined [here](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socketerror?view=net-5.0), trying to address all that in a robust way will be quite a challange.

The related Scala code is [here](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/io/UdpListener.scala#L107-L111), as you can see, it is not trying to handle any errors, it just does a simple try...finally and release the buffer it borrowed from a buffer pool.

motmot80:
Looks like someone has had the same issue with Akka-Scala :)

https://github.com/akka/akka/issues/26903

Maybe we should take a closer look at the fix.

Arkatufus:
> Unfortunatly I didn't find any code within the original akka UDP implementation throwing SocketException or IOExceptions
It does, it just doesn't explicitly say that it will throw them. `channel.receive(buffer)` is a call that potentially throws if the socket underneath it fails.

motmot80:
They fixed it by handling this specific `java.net.PortUnreachableException` error individually:

https://github.com/akka/akka/blob/a4acf23d0541471ecf81b9ba5dd97c2e97670814/akka-actor/src/main/scala/akka/io/UdpConnection.scala#L135

motmot80:
Maybe we should explicitly identify this "port unreachable" SocketError and throwing an exception in every other case as you recommended. This is more or less the same as the akka scala solution and is fitting my need.

Arkatufus (approved):

motmot80:
I will add some basic unit tests for the UdpListener and one which is reproducing the issue.

Aaronontheweb:
@Arkatufus are you good with these most recent changes?

Arkatufus:
@motmot80 
It looks like the new test is failing under Linux due to socket port collision.

```
[xUnit.net 00:01:24.66]     Akka.Tests.IO.UdpListenerSpec.A_UDP_Listener_must_be_able_to_send_and_receive_when_server_goes_away [FAIL]
[xUnit.net 00:01:24.66]       System.Net.Sockets.SocketException : Address already in use
[xUnit.net 00:01:24.66]       Stack Trace:
[xUnit.net 00:01:24.66]            at System.Net.Sockets.Socket.UpdateStatusAfterSocketErrorAndThrowException(SocketError error, String callerName)
[xUnit.net 00:01:24.66]            at System.Net.Sockets.Socket.DoBind(EndPoint endPointSnapshot, SocketAddress socketAddress)
[xUnit.net 00:01:24.67]            at System.Net.Sockets.Socket.Bind(EndPoint localEP)
[xUnit.net 00:01:24.67]            at System.Net.Sockets.UdpClient..ctor(Int32 port, AddressFamily family)
[xUnit.net 00:01:24.67]            at System.Net.Sockets.UdpClient..ctor(Int32 port)
[xUnit.net 00:01:24.67]         /home/vsts/work/1/s/src/core/Akka.Tests/IO/UdpListenerSpec.cs(129,0): at Akka.Tests.IO.UdpListenerSpec.TestSetup.SendData(Byte[] buffer, IPEndPoint receiverEndpoint, IPEndPoint senderEndpoint)
[xUnit.net 00:01:24.67]         /home/vsts/work/1/s/src/core/Akka.Tests/IO/UdpListenerSpec.cs(71,0): at Akka.Tests.IO.UdpListenerSpec.<>c.<A_UDP_Listener_must_be_able_to_send_and_receive_when_server_goes_away>b__3_0(TestSetup x)
[xUnit.net 00:01:24.67]         /home/vsts/work/1/s/src/core/Akka.Tests/IO/UdpListenerSpec.cs(109,0): at Akka.Tests.IO.UdpListenerSpec.TestSetup.Run(Action`1 test)
[xUnit.net 00:01:24.67]         /home/vsts/work/1/s/src/core/Akka.Tests/IO/UdpListenerSpec.cs(77,0): at Akka.Tests.IO.UdpListenerSpec.A_UDP_Listener_must_be_able_to_send_and_receive_when_server_goes_away()
```

Aaronontheweb:
Yeah we need to fix that before we can merge it

motmot80:
Will take care. Sorry for the trouble.

motmot80:
I changed the test to use the existing actor to send the udp packet to the non existing client. So it is not necessary to use the existing listening port as sender port. Hopefully this fixes the issue on the linux platform.

Arkatufus:
The linux unit tests are passing, we just need to wait for some flaky tests to pass.

Aaronontheweb:
thanks for your contribution @motmot80

Issue #4215, opened 2020-02-13
MultiNodeSpecs: turn off debug logging for all of them


Aaronontheweb:
Version: 1.4.0-beta4

We generate MASSIVE amounts of logs each time we run the MNTR. Most of the tests in the suite are pretty stable, especially since the most recent changes we introduced since Akka.NET v1.4.0-beta4 was released. Let's cut down on the noise by setting all of the logging levels back to INFO or lower.

cc @IgorFedchenko

razvangoga:
@Aaronontheweb from what I gather from [here](https://getakka.net/community/building-akka-net.html), the logs this issue refers are the ones generated from **build.cmd MultiNodeTestsNetCore** or **build.cmd MultiNodeTestsNet**. Correct?

If yes how long does a run usually take? It's been running for 30min+ on my quite powerfull machine (Razer Blade 2018 / i7 6core / 32GB / fast ssd). Or am I doing something wrong?

Aaronontheweb:
@razvangoga I think each run on our build server usually takes 1 hour and 15 minutes. The tests are extremely long-running in some cases.

razvangoga:
Ok, so 1h11 locally is in line.

Can you confirm that the logs i mentioned above are the ones the issue is
about?

razvangoga:
@Aaronontheweb looking into the 18MB log dump i got from last night's run, i've got the following (ball-park) breakdown:

total lines  : 59.5k

- build output :  1.3k
  -   compile warnings :  0.8k
- tests output : 58.0k 
  -  **DEBUG    : 20.0k** 
  -  **INFO     : 13.0k**
  -  WARNING  : 10.0k
  -  ERROR    :  0.5k
  -  others   : 14.0k (stacktrace lines from errors / warnings, empty lines, logs without level)

Basically if the logging would be configured with the level
-  >=WARNING, you would shave 50% of it : -33k lines
- >=ERROR, you would  maybe about 75% : -43k + some of the stacktrace lines (as DotNetty connection errors like this one are logged as WARNING) - so maybe 50k?
![image](https://user-images.githubusercontent.com/1067494/100345617-dc4bc680-2fe2-11eb-8068-ac5e8963e407.png)

IgorFedchenko:
[...]

PR #4644, opened 2020-11-30
MultiNodeSpecs: turn off debug logging

razvangoga:
PR for #4215

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
@@ -43,7 +43,7 @@ namespace Akka.Cluster.Sharding.Tests
                             ""System.Object"" = hyperion
                         }}
                     }}
-                    akka.loglevel = INFO
+                    akka.loglevel = WARNING
                     akka.actor.provider = cluster
                     akka.remote.log-remote-lifecycle-events = off
                     akka.cluster.auto-down-unreachable-after = 0s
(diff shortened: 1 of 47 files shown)


src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
```diff
@@ -43,7 +43,7 @@ protected ClusterShardingFailureSpecConfig(string mode)
                             ""System.Object"" = hyperion
                         }}
                     }}
-                    akka.loglevel = INFO
+                    akka.loglevel = WARNING
```
IgorFedchenko, line 46: So, instead of changing logging level per spec, could you try to configure logging level on the MultiNodeTestRunner side? So that all existing specs could still write logs as they want, and we did not have any restrictions on that for new specs - but MNTR would write only given log level to it's outputs.

This way we could easily configure that in future. For example, let it output DEBUG level when we are running / debugging new tests on local machine.

IgorFedchenko: That would also save you a lot of efforts from modifying other specs we have. Does it make sense?

razvangoga: I guess that will also work. 
I'll take a look in the MNTR to see how it works and get back to you with specific questions if any.

IgorFedchenko: Sure, I'll be glad to help! Right now I do not have all the details from my head, but I remember that there are some actors that are handling arriving log events and producing output. So it should be a good place to inject filtering.

eaba: [WIP PR](https://github.com/akkadotnet/akka.net/pull/4651)

PR #4648, opened 2020-12-01
Fixes #3723

### LINKED ISSUE #3723: FSM: exception in LogTermination changes stopEvent.Reason to Shutdown
When stopping an FSM with `return Stop(new Failure(null))`, at first sight it seems to work. But this actually causes an exception in `LogTermination(Reason reason)` at https://github.com/akkadotnet/akka.net/blob/41c116d2a3d280481e584401608ec848dc147632/src/core/Akka/Actor/FSM.cs#L1353

Later on in `OnTermination`, the type of `stopEvent.Reason` will be `Shutdown` instead of `Failure`.

I'd expect one of these to happen:
* an `ArgumentNullException` is thrown when calling `new Failure(null)`
* exceptions in `LogTermination(...)` are caught so the original stop reason is preserved

eaba:


eaba:
https://github.com/akkadotnet/akka.net/issues/3723#issuecomment-737037116

src/core/Akka/Actor/FSM.cs
```diff
@@ -1336,7 +1337,7 @@ protected virtual void LogTermination(Reason reason)
                 {
                     _log.Error(failure.Cause.AsInstanceOf<Exception>(), "terminating due to Failure");
                 }
-                else
+                else if(!(failure.Cause is null))
```
Arkatufus, line 1340: This is a bad way of doing this. A failure is a failure, you should not block the error log.
A better way to handle this is to use the string `"null"` if `failure.Cause` is `null`.

Arkatufus (changes requested):

src/core/Akka/Actor/FSM.cs
```diff
@@ -1325,6 +1325,7 @@ protected override void PostStop()
         /// <summary>
         /// By default, <see cref="Failure"/> is logged at error level and other
         /// reason types are not logged. It is possible to override this behavior.
+        /// Should handle the case when Cause is null and don't log. If that is not the desired behavior, LogTermination can always be overridden
```
Arkatufus, line 1328: The summary tag is a documentation tag, it will be emmited into the API documentation. You should use a more affirmative language and tell the user what the method is actually doing.

Arkatufus (changes requested):

src/core/Akka/Actor/FSM.cs
eaba: Removed that now and reverted back to the original

src/core/Akka/Actor/FSM.cs
eaba: Done in the latest commit

src/core/Akka/Actor/FSM.cs
```diff
@@ -1338,7 +1338,7 @@ protected virtual void LogTermination(Reason reason)
                 }
                 else
                 {
-                    _log.Error(failure.Cause.ToString());
+                    _log.Error(failure.Cause is null? "null": failure.Cause.ToString());
```
Arkatufus, line 1341: Good change 👍

Arkatufus (approved):

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Performance/SqliteJournalPerfSpec.cs
```diff
@@ -10,7 +10,7 @@
 using Akka.Util.Internal;
 using Xunit.Abstractions;
 
-namespace Akka.Persistence.Redis.Tests.Performance
+namespace Akka.Persistence.Sqlite.Tests.Performance
```
Aaronontheweb, line 13: LGTM

Aaronontheweb:
Just a note - it's probably best to do your work on a feature branch rather than the `dev` branch of your personal fork of Akka.NET. Otherwise it will be really difficult to sync your commits up with Akka.NET when you start working on multiple things at once.

PR #4651, opened 2020-12-03
[WIP ] MultiNodeSpecs: turn off debug logging for all of them

eaba:


src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
```diff
@@ -28,29 +28,29 @@ public ClusterShardingCustomShardAllocationSpecConfig()
...
-                        serialization-bindings {
+                        }}
+                        serialization-bindings {{
                             ""System.Object"" = hyperion
-                        }
-                    }
-
+                        }}
+                    }}
+                    akka.loglevel = {LogLevel}
```
IgorFedchenko, line 40: So, the idea is that, let's try to not modify any specs - let them write the logs as there were logging before. Let's just make MultiNodeTestRunner to not output logs of particular level to the console / test output provider.

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingFailureSpec.cs
```diff
@@ -43,7 +43,7 @@ protected ClusterShardingFailureSpecConfig(string mode)
                             ""System.Object"" = hyperion
                         }}
                     }}
-                    akka.loglevel = INFO
+                    akka.loglevel = {LogLevel}
```
IgorFedchenko, line 46: Same as above, and in all other places like that

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs
IgorFedchenko: I will not leave a comment under each modification like this, but you got the idea

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -138,7 +138,8 @@ static void Main(string[] args)
             var specName = CommandLine.GetPropertyOrDefault("multinode.spec", "");
             var platform = CommandLine.GetPropertyOrDefault("multinode.platform", "net");
             var reporter = CommandLine.GetPropertyOrDefault("multinode.reporter", "console");
-            
+            var logLevel = CommandLine.GetPropertyOrDefault("multinode.log.level", "INFO"); //-Dmultinode.log.level=DEBUG
```
IgorFedchenko, line 141: To keep the behavior backward compatible, shouldn't we use DEBUG level here?

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -36,6 +36,9 @@ public abstract class MultiNodeConfig
         // it will remain safe by defaut
         Config _commonConf = ConfigurationFactory.Empty;
 
+        //makes log level configurable
+        public string LogLevel = "DEBUG";
```
IgorFedchenko, line 40: Do you think we need to modify the spec definition/configuration? Or probably we can keep specs untouched, and just apply logs filtering on MultiNodeTestRunner side?

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -264,7 +265,8 @@ static void Main(string[] args)
                                     .Append($@"-Dmultinode.index={nodeTest.Node - 1} ")
                                     .Append($@"-Dmultinode.role=""{nodeTest.Role}"" ")
                                     .Append($@"-Dmultinode.listen-address={listenAddress} ")
-                                    .Append($@"-Dmultinode.listen-port={listenPort} ");
+                                    .Append($@"-Dmultinode.listen-port={listenPort} ")
+                                    .Append($@"-Dmultinode.log.level={logLevel}");
```
IgorFedchenko, line 269: So, probably, instead of passing log level to `NoteTestRunner` process, we could use this `logLevel` value to apply logs filtering on the output? In this program, there is console / test output adapter actors configuration, and we could use this log level to ask those adapters to just skip the logs with level below the one we need.

Do you think we can try that?

IgorFedchenko (commented):
Left you few comments, to share my thoughts about how we could resolve the issue with less efforts / changes.

src/core/Akka.MultiNodeTestRunner/Program.cs
eaba: @IgorFedchenko am a bit confused.

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
eaba: When LogLevel is supplied vai command line, it is stored here.
````csharp
LogLevel = CommandLine.GetPropertyOrDefault("multinode.log.level", "DEBUG");
            
````

File: src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs (hunk 1 above)
Comment (eaba) [reply src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs:40]: > 
> 
> So, the idea is that, let's try to not modify any specs - let them write the logs as there were logging before. Let's just make MultiNodeTestRunner to not output logs of particular level to the console / test output provider.

I think we are not really modifying any Specs - `MultiNodeConfig` is what is being modified.

Comment (IgorFedchenko) [reply src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingCustomShardAllocationSpec.cs:40]: I mean, we should try to not modify the specs configuration - just the way how MNTR process is printing this logs to the output.

File: src/core/Akka.MultiNodeTestRunner/Program.cs (hunk 5 above)
Comment (IgorFedchenko) [reply src/core/Akka.MultiNodeTestRunner/Program.cs:269]: See discussion here https://github.com/akkadotnet/akka.net/issues/4215#issuecomment-738071227

I think this PR is going to do what we need, @razvangoga was working on that already for some time (still needs to be updated according to the comment above): https://github.com/akkadotnet/akka.net/pull/4644

Comment (IgorFedchenko) [reply src/core/Akka.MultiNodeTestRunner/Program.cs:269]: It's just I do not think we need to pass log level to the child processes - it can be used in MNTR process itself for output filtering.

So the idea is not to change how the logs are logged (under which level) - but how they are redirected to the console/output from MNTR process. Because we do not case about what logs are logged by NodeTestRunner processes (they are not getting to the output), it's MNTR who is collecting logs into single place and writing them to the output.
That's why the idea was to just apply filtering on the MNTR process level. This would also allow us to not touch specs configurations or how they run in NTR processes.

Comment (eaba) [reply src/core/Akka.MultiNodeTestRunner/Program.cs:269]: It is clearer now: the filtering should happen at the `MultiNodeTestRunner`.

File: src/core/Akka.MultiNodeTestRunner/Program.cs [hunk 6]
```diff
@@ -138,7 +138,7 @@ static void Main(string[] args)
             var specName = CommandLine.GetPropertyOrDefault("multinode.spec", "");
             var platform = CommandLine.GetPropertyOrDefault("multinode.platform", "net");
             var reporter = CommandLine.GetPropertyOrDefault("multinode.reporter", "console");
-            var logLevel = CommandLine.GetPropertyOrDefault("multinode.log.level", "DEBUG"); //-Dmultinode.log.level=DEBUG
+            var logLevel = CommandLine.GetPropertyOrDefault("multinode.log.level", "INFO"); //-Dmultinode.log.level=INFO
```
Aaronontheweb, line 141: please use real commit messages, as a matter of standards, going forward

Aaronontheweb:
Superceded via https://github.com/akkadotnet/akka.net/pull/4652

PR #4652, opened 2020-12-03
MultiNodeSpecs: turn off debug logging for all of them #4215

razvangoga:
PR for #4215

### DIFF
--- a/src/common.props
+++ b/src/common.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Copyright>Copyright © 2013-2020 Akka.NET Team</Copyright>
     <Authors>Akka.NET Team</Authors>
-    <VersionPrefix>1.4.13</VersionPrefix>
+    <VersionPrefix>1.4.12</VersionPrefix>
     <PackageIconUrl>https://getakka.net/images/akkalogo.png</PackageIconUrl>
     <PackageProjectUrl>https://github.com/akkadotnet/akka.net</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/akkadotnet/akka.net/blob/master/LICENSE</PackageLicenseUrl>
@@ -28,7 +28,7 @@
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
   </PropertyGroup>
   <PropertyGroup>
-    <PackageReleaseNotes>Placeholder for Nightlies**</PackageReleaseNotes>
+    <PackageReleaseNotes>Placeholder for nightlies**</PackageReleaseNotes>
   </PropertyGroup>
   <!-- SourceLink support for all Akka.NET projects -->
   <ItemGroup>
--- a/src/common.props
+++ b/src/common.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Copyright>Copyright © 2013-2020 Akka.NET Team</Copyright>
     <Authors>Akka.NET Team</Authors>
-    <VersionPrefix>1.4.12</VersionPrefix>
+    <VersionPrefix>1.4.13</VersionPrefix>
     <PackageIconUrl>https://getakka.net/images/akkalogo.png</PackageIconUrl>
     <PackageProjectUrl>https://github.com/akkadotnet/akka.net</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/akkadotnet/akka.net/blob/master/LICENSE</PackageLicenseUrl>
@@ -28,7 +28,7 @@
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
   </PropertyGroup>
   <PropertyGroup>
-    <PackageReleaseNotes>Placeholder for nightlies**</PackageReleaseNotes>
+    <PackageReleaseNotes>Placeholder for Nightlies**</PackageReleaseNotes>
   </PropertyGroup>
   <!-- SourceLink support for all Akka.NET projects -->
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
cc @IgorFedchenko

src/core/Akka.MultiNodeTestRunner/Program.cs
```diff
@@ -125,7 +122,9 @@ static void Main(string[] args)
         {
             OutputDirectory = CommandLine.GetPropertyOrDefault("multinode.output-directory", string.Empty);
             FailedSpecsDirectory = CommandLine.GetPropertyOrDefault("multinode.failed-specs-directory", "FAILED_SPECS_LOGS");
-            TestRunSystem = ActorSystem.Create("TestRunnerLogging");
+            
+            string logLevel = CommandLine.GetPropertyOrDefault("multinode.loglevel", "WARNING");
+            TestRunSystem = ActorSystem.Create("TestRunnerLogging", $"akka.loglevel={logLevel}");
```
IgorFedchenko, line 127: LGTM

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/TimelineLogCollectorActor.cs
```diff
@@ -14,25 +14,28 @@
...
     public class TimelineLogCollectorActor : ReceiveActor
     {
         private readonly SortedList<DateTime, HashSet<LogMessageInfo>> _timeline = new SortedList<DateTime, HashSet<LogMessageInfo>>();
-        
-        public TimelineLogCollectorActor()
+        private readonly LogLevel _minimumLogLevel;
+
+        public TimelineLogCollectorActor(string minimumLogLevel)
         {
+            LogMessageInfo.TryParseLogLevel(minimumLogLevel, out _minimumLogLevel);
```
IgorFedchenko, line 27: Yep, this should do the trick! Nice work!

Now, when I am looking at it, it seems like log level setting is specific to `PrintToConsole` command handling. That is, when receiving `DumpToFile` message we do not apply any filtering based on the log level provided. And that's great by the way - we want to reduce amount of logs printed in the console, but for debugging perposes (and for some MNTR reports) it is nice to have all logs saved without filtering.

So, this might be a good idea to pass required log level as an optional parameter of `PrintToConsole` command - do you agree? That would make it clear that we are just filtering console output.

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/TimelineLogCollectorActor.cs
```diff
@@ -87,26 +91,26 @@ public class LogMessageInfo
...
+            public LogLevel? LogLevel { get; }
             public string Message { get; }
 
             public LogMessageInfo(LogMessage msg)
             {
                 OriginalMessage = msg.Message;
                 Node = msg.Node;
                 When = DateTime.UtcNow;
-                LogLevel = LogLevel.InfoLevel; // In case if we could not find log level, assume that it is Info
+                LogLevel = null; //some log lines coming from nodes have no level (ex: stack traces) => live the level null and output always
```
IgorFedchenko, line 102: Yep, that's a good thing, making this more explicit :+1:

IgorFedchenko (changes requested):
Requested some minor changes. Nice work @razvangoga , this will make MNTR test output much more easy to read in case of failures!

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/TimelineLogCollectorActor.cs
razvangoga: Makes sense - modified accordingly

src/common.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Copyright>Copyright © 2013-2020 Akka.NET Team</Copyright>
     <Authors>Akka.NET Team</Authors>
-    <VersionPrefix>1.4.12</VersionPrefix>
+    <VersionPrefix>1.4.13</VersionPrefix>
```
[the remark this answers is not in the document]
razvangoga: done

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/TimelineLogCollectorActor.cs
ismaelhamed: *leave?

IgorFedchenko: ah, good catch :+1: "live" --> "leave"

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/TimelineLogCollectorActor.cs
```diff
@@ -46,23 +46,25 @@ public TimelineLogCollectorActor()
...
-                
+
                 File.AppendAllLines(dump.FilePath, _timeline.Select(pairs => pairs.Value).SelectMany(msg => msg).Select(m => m.ToString()));
                 Sender.Tell(Done.Instance);
             });
-            
+
             Receive<PrintToConsole>(_ =>
             {
+                LogMessageInfo.TryParseLogLevel(_.MinimumLogLevel, out var minimumLogLevel);
```
IgorFedchenko, line 66: `_` is usually used as not used argument, so we should probably rename it to something more meaningful - just to avoid people thinking that there are no parameters used

IgorFedchenko (changes requested):
One very minor change, and I will be happy to approve this 👍

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/TimelineLogCollectorActor.cs
razvangoga: i've already fixed the typo - please look at the files (comments seem stuck to the original version)

src/core/Akka.MultiNodeTestRunner.Shared/Sinks/TimelineLogCollectorActor.cs
razvangoga: done

IgorFedchenko (approved):
LGTM - would love to see this in our development branch :+1:

Aaronontheweb:
nice work @razvangoga

PR #4657, opened 2020-12-08
Set `PublishRepositoryUrl` to false

eaba:
Prevents .NET 5 tools from adding "RepositoryUrl" AssemblyMetadataAttribute

### DIFF
--- a/src/common.props
+++ b/src/common.props
@@ -35,7 +35,7 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
   <PropertyGroup>
-    <PublishRepositoryUrl>true</PublishRepositoryUrl>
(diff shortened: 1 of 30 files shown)


src/common.props
```diff
@@ -35,7 +35,7 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
   <PropertyGroup>
-    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <PublishRepositoryUrl>false</PublishRepositoryUrl>
```
Aaronontheweb, line 38: Hmmmm... I wonder if this will break SourceLink then. Maybe we need to hard code the RepositoryUrl instead so it can't be overridden by a personal fork?

Aaronontheweb (changes requested):

PR #4661, opened 2020-12-10
Deprecate PatternMatch

### LINKED ISSUE #4658: Deprecate PatternMatch
I guess this wouldn't hurt, and it will also serve as a reminder to refactor any file we touch in Akka.NET that still uses it. 

I will send a PR if this is fine:

```csharp
/// <summary>
/// Class PatternMatch.
/// </summary>
[Obsolete("Use instead the pattern matching feature introduced in C# 7.0")]
public static class PatternMatch
```

ismaelhamed:
Closes #4658


Aaronontheweb (approved):
LGTM

PR #4666, opened 2020-12-11
Provide sensible default values for Akka.Persistence circuit breaker configuration values

Arkatufus:


src/core/Akka.Persistence/Snapshot/SnapshotStore.cs
```diff
@@ -39,9 +39,9 @@ protected SnapshotStore()
             var config = extension.ConfigFor(Self);
             _breaker = CircuitBreaker.Create(
                 Context.System.Scheduler,
-                config.GetInt("circuit-breaker.max-failures", 0),
```
Arkatufus, line 42: These default values does not make sense. I'm replacing it with a more sensible default values from the HOCON configuration file.

Arkatufus: These default value assignments, in most cases, never get applied to the breaker because it usually filled with a sensible fallback value by any well behaving plugin. It would create a bug if a plugin does not provide any fallback configurations.

Aaronontheweb: Ah got it, that makes sense - so this change is probably ineffective in most cases.

Aaronontheweb: Ineffective = "values are overridden by the implementation usually so these settings often get ignored"

Aaronontheweb (approved):
LGTM

PR #4669, opened 2020-12-15
Hardcode RepositoryUrl

eaba:
Hard coded RepositoryUrl(https://github.com/akkadotnet/akka.net) to prevent Build from adding the repositoryurl of forked repository

### DIFF
--- a/src/common.props
+++ b/src/common.props
@@ -35,7 +35,9 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
   <PropertyGroup>
-    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <!--https://github.com/dotnet/sourcelink/issues/659-->  
+    <!--<PublishRepositoryUrl>true</PublishRepositoryUrl>-->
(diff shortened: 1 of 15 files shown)


Aaronontheweb:
Got the tests to pass locally on my machine with these settings - let's see what the build server says.

Aaronontheweb:
looks like the issue is that the build server never generates the `﻿[assembly: System.Reflection.AssemblyMetadataAttribute("RepositoryUrl", "https://github.com/akkadotnet/akka.net")]` metadata in the assembly files - that's probably because this issue was introduced, IIRC, by the .NET 5 SDK

Resolve https://github.com/akkadotnet/akka.net/pull/4671 first and then we'll re-check this. Wish we could get a better message out of the ApiApprover to validate this.

eaba:
> 
> 
> looks like the issue is that the build server never generates the `﻿[assembly: System.Reflection.AssemblyMetadataAttribute("RepositoryUrl", "https://github.com/akkadotnet/akka.net")]` metadata in the assembly files - that's probably because this issue was introduced, IIRC, by the .NET 5 SDK
> 
> Resolve #4671 first and then we'll re-check this. Wish we could get a better message out of the ApiApprover to validate this.

It was a pain.....

eaba:
Superceded by https://github.com/akkadotnet/akka.net/pull/4683

PR #4675, opened 2020-12-16
API - added overload for IScheduler that accepts IRunnable

Aaronontheweb:
Purpose of this change is to make it possible to take Akka.Dispatch.IRunnable, the abstraction we use inside Dispatchers for processing mailboxes among other things, and pass it directly to the IScheduler for execution without having to wrap it.

This is largely a performance optimization and I don't expect this API to be used outside of Akka.NET's dispatching system.


IgorFedchenko (approved):
LGTM - at least from the refactoring standpoint

Issue #4636, opened 2020-11-17
High Idle CPU in DotNetty


Zetanova:
I still have the issue with idle nodes more or less like in https://github.com/akkadotnet/akka.net/issues/4434
docker akka.net 1.4.11 dotnet 3.1.404  debug and release builds

All 7 nodes are idling and consume 100%  (docker is limited to 3 cores)
The main hotpatch is still in dotnetty

![image](https://user-images.githubusercontent.com/1385312/99338205-0800db00-2884-11eb-8a8b-891ccfe04a45.png)

messages/traffic is low the node is idling.

Zetanova:
The CuonccurencyQueue in the class `Helios.Concurrency.DedicatedThreadPool.ThreadPoolWorkQueue` could be replaced with the new [`System.Threading.Channels`](https://www.nuget.org/packages/System.Threading.Channels) API of the dotnet/runtime

This would get rid of the `UnfairSemaphore` implementation for good or bad.

https://github.com/akkadotnet/akka.net/blob/a80ddd79240382fee172e41fa48060034c95754f/src/core/Akka/Helios.Concurrency.DedicatedThreadPool.cs#L391-L399

I don't have the setup/knowledge to measure the perf effects of this change, can test it?

Zetanova:
I made a branch https://github.com/Zetanova/akka.net/tree/helios-idle-cpu with a commit that changes the `ThreadPoolWorkQueue` to `System.Threading.Channels.Channel`

Please, can somebody run a test and benchmark

Or explain me how to run get the Akka.MultiNodeTestRunner.exe started.

Aaronontheweb:
Cc @to11mtm - guess I need to move up the time table on doing that review

Aaronontheweb:
@Zetanova  I’ll give your branch a try - OOF for a couple of days but I’ll get on it

to11mtm:
@Zetanova I'll try to run this through the paces as well in the next few days. :)

Zetanova:
I made a helios-io/DedicatedThreadPool fork https://github.com/Zetanova/DedicatedThreadPool/tree/try-channels

The problem is that the **benchmark does not count the spin waits / idle CPU**

## CURRENT
--------------- RESULTS: Helios.Concurrency.Tests.Performance.DedicatedThreadPoolBenchmark+ThreadpoolBenchmark ---------------
--------------- DATA ---------------
[...]

Issue #4563, opened 2020-09-08
Akka.Remote - exhaustion of TCP buffer after updating from 1.3.8 to 1.4.6


voltcode:
**Scenario:**

We have a Windows service that pings other windows services and iis app pools (ca.10 services) every 10 seconds using Akka.Remote. Ping is implemented as Ask.  

Some of them may be dead for a while, etc. - it is fairly normal (for example one of the services is an IIS app pool, which will be activated only after user action. All services are on the same machine.

With Akka 1.3.8 it was working correctly.

On akka 1.4.6  we started observing the following problem on our test virtual machines:
1) start VM
2) services wake up, ping is working correctly
3) after at least 10 hours, the ping stops working. 

**Error**
We get the following exception when one of our services tries to talk to another using for example WCF using http:

```
Unable to connect to the remote server ---> System.Net.Sockets.SocketException: An operation on a socket could not be performed because the system lacked sufficient buffer space or because a queue was full 127.0.0.1:80
   at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
   at System.Net.ServicePoint.ConnectSocketInternal(Boolean connectFailure, Socket s4, Socket s6, Socket& socket, IPAddress& address, ConnectSocketState state, IAsyncResult asyncResult, Exception& exception)
```

basically, the OS prevents the services to talk to each other at all. 

We've never had such problems before Akka upgrade.

Another symptom is that the services stop being responsive at all.  Our test VM is resource constrained (1  CPU cpu @4 GHz, 4GB RAM).

debugging snapshot looks like this:
```
mscorlib_ni!System.Threading.Tasks.Task.InnerInvoke() 
mscorlib_ni!System.Threading.Tasks.Task.Execute() 
mscorlib_ni!System.Threading.Tasks.Task.ExecutionContextCallback(System.Object) 
mscorlib_ni!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.
Threading.ContextCallback, System.Object, Boolean) 
mscorlib_ni!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threadin
g.ContextCallback, System.Object, Boolean) 
mscorlib_ni!System.Threading.Tasks.Task.ExecuteWithThreadLocal(System.Threading.Tasks.Task ByRef) 
mscorlib_ni!System.Threading.Tasks.Task.ExecuteEntry(Boolean) 
[...]

PR #4678, opened 2020-12-17
Move DotNetty batching scheduling off of DotNetty STEE and onto HashedWheelTimer

Aaronontheweb:
work for https://github.com/akkadotnet/akka.net/issues/4563 and https://github.com/akkadotnet/akka.net/issues/4636

### DIFF
--- a/src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
+++ b/src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Threading.Tasks;
+using Akka.Actor;
 using DotNetty.Buffers;
 using DotNetty.Common.Concurrency;
 using DotNetty.Transport.Channels;
@@ -83,12 +84,14 @@ namespace Akka.Remote.Transport.DotNetty
     internal class BatchWriter : ChannelHandlerAdapter
     {
         public readonly BatchWriterSettings Settings;
+        public readonly IScheduler Scheduler;
 
         internal bool CanSchedule { get; private set; } = true;
 
-        public BatchWriter(BatchWriterSettings settings)
+        public BatchWriter(BatchWriterSettings settings, IScheduler scheduler)
         {
             Settings = settings;
+            Scheduler = scheduler;
         }
 
         private int _currentPendingWrites = 0;
@@ -146,7 +149,7 @@ namespace Akka.Remote.Transport.DotNetty
         {
             // Schedule a recurring flush - only fires when there's writable data
             var task = new FlushTask(context, Settings.FlushInterval, this);
-            context.Executor.Schedule(task, Settings.FlushInterval);
+            task.Start();
         }
 
         public void Reset()
@@ -160,12 +163,24 @@ namespace Akka.Remote.Transport.DotNetty
             private readonly IChannelHandlerContext _context;
             private readonly TimeSpan _interval;
             private readonly BatchWriter _writer;
+            private ICancelable _cancellation;
 
             public FlushTask(IChannelHandlerContext context, TimeSpan interval, BatchWriter writer)
             {
                 _context = context;
                 _interval = interval;
                 _writer = writer;
+            }
+
+            public void Start()
+            {
+                _cancellation =
+                    _writer.Scheduler.Advanced.ScheduleRepeatedlyCancelable(_interval, _interval, () => Run());
+            }
+
+            public void Cancel()
+            {
+                _cancellation?.Cancel();
             }
 
             public void Run()
@@ -177,8 +192,8 @@ namespace Akka.Remote.Transport.DotNetty
                     _writer.Reset();
                 }
 
-                if(_writer.CanSchedule)
-                    _context.Executor.Schedule(this, _interval); // reschedule
+                if(!_writer.CanSchedule)
+                   Cancel(); // unschedule
             }
         }
     }
--- a/src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
+++ b/src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
@@ -34,7 +34,7 @@ namespace Akka.Remote.Transport.DotNetty
             FlushInterval = hocon.GetTimeSpan("flush-interval", DefaultFlushInterval, false);
         }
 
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Raw throughput numbers were the same as `dev` on this branch - so no real impact to end to end performance.

However, significant change in idle CPU numbers!

Some important notes:

- `clusterwebcrawler_webcrawler.lighthouse2_1` - isn't in the cluster although it is running Akka.Remote
- `clusterwebcrawler_webcrawler.lighthouse_1 ` IS in the cluster but IS NOT using batching
- all other services inside the cluster are using batching

## Before

```
CONTAINER ID        NAME                                            CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O   
 PIDS                                                                                                                                                             
84ccf601e906        clusterwebcrawler_webcrawler.lighthouse2_1      0.55%               35.52MiB / 50.17GiB   0.07%               1.2kB / 0B          0B / 0B     
 21                                                                                                                                                               
37fd1e313cdb        clusterwebcrawler_webcrawler.web_1              9.98%               79.53MiB / 50.17GiB   0.15%               98.8kB / 97kB       0B / 0B     
 56                                                                                                                                                               
57d530c02a47        clusterwebcrawler_webcrawler.trackerservice_1   10.96%              48.97MiB / 50.17GiB   0.10%               99.9kB / 99kB       0B / 0B     
 39                                                                                                                                                               
5737cbaed752        clusterwebcrawler_webcrawler.crawlservice_1     11.31%              44.49MiB / 50.17GiB   0.09%               97.3kB / 97.4kB     0B / 0B     
 39                                                                                                                                                               
66dd2a7fe096        clusterwebcrawler_webcrawler.lighthouse_1       4.36%               45.39MiB / 50.17GiB   0.09%               106kB / 109kB       0B / 0B     
 27                                                                                                                                                               
```

## After

```
 CONTAINER ID        NAME                                            CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O
 PIDS
b332f021fbe1        clusterwebcrawler_webcrawler.lighthouse2_1      0.52%               35.91MiB / 50.17GiB   0.07%               1.27kB / 0B         0B / 0B
 21
0093f3510df3        clusterwebcrawler_webcrawler.crawlservice_1     5.64%               52.1MiB / 50.17GiB    0.10%               201kB / 202kB       0B / 0B
 40
d738c3391a89        clusterwebcrawler_webcrawler.trackerservice_1   4.96%               48.66MiB / 50.17GiB   0.09%               201kB / 201kB       0B / 0B
 39
0392e8f7835d        clusterwebcrawler_webcrawler.web_1              5.29%               85.66MiB / 50.17GiB   0.17%               211kB / 204kB       0B / 0B
 57
fd492a3b6ff5        clusterwebcrawler_webcrawler.lighthouse_1       4.42%               46.12MiB / 50.17GiB   0.09%               216kB / 221kB       0B / 0B
 23
```

Looks like a ~50% drop in idle CPU consumption to me. The nodes in the cluster using batching are now in the same ballpark is the ones that aren't.

src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
```diff
@@ -82,13 +83,22 @@ public BatchWriterSettings(Config hocon)
...
+        {
+            public static readonly FlushCmd Instance = new FlushCmd();
+            private FlushCmd() { }
+        }
+
         public readonly BatchWriterSettings Settings;
+        public readonly IScheduler Scheduler;
+        private ICancelable _flushSchedule;
 
         internal bool CanSchedule { get; private set; } = true;
```
to11mtm, line 96: It looks like we don't use `CanSchedule` anywhere, we are only setting it. Should it be removed?

src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
```diff
@@ -134,52 +144,50 @@ public override Task WriteAsync(IChannelHandlerContext context, object message)
...
         {
             // Schedule a recurring flush - only fires when there's writable data
-            var task = new FlushTask(context, Settings.FlushInterval, this);
-            context.Executor.Schedule(task, Settings.FlushInterval);
+            _flushSchedule = Scheduler.Advanced.ScheduleRepeatedlyCancelable(Settings.FlushInterval,
+                Settings.FlushInterval,
+                () =>
+                {
+                    // want to fire this event through the top of the pipeline
+                    context.Channel.Pipeline.FireUserEventTriggered(FlushCmd.Instance);
```
Aaronontheweb, line 183: I wanted to use an event for this as this will allow me to do some sampling around how often the `FlushCmd` has to be invoked separately from other flushes. But that may end up being a distinction without difference depending on how I write the sampler.

src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
```diff
@@ -134,52 +144,50 @@ public override Task WriteAsync(IChannelHandlerContext context, object message)
...
+                });
```
to11mtm, line 184: So this is one heck of an edge case.... But do we want to have any sort of warning if the flush interval is for some unknown reason less than `tick-duration` on HWT?

to11mtm (approved):
One question and one nit but looks good otherwise

src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
Aaronontheweb: I'll remove it

src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
Aaronontheweb: No we don't - but it shouldn't be an issue. The scheduler will just put it in the next available bucket if that's the case.

src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
```diff
@@ -114,72 +121,66 @@ public override Task WriteAsync(IChannelHandlerContext context, object message)
              * across the network.
              */
             var write = base.WriteAsync(context, message);
-            if (Settings.EnableBatching)
```
Aaronontheweb, line 117: Removed the `Settings.EnableBatching` checks - this stage never runs if this is false.

PR #4684, opened 2020-12-22
fixed bug with IPersistentRepresentation.Update and timestamp propagation

Aaronontheweb:
cc @ismaelhamed 

Tried to see if I could make the `Persistent` constructor backwards compatible per my comment on #4680 - turns out that's not feasible. Not without making an even nastier breaking change or adding some type of adapter. All plugins are going to need to be recompiled with Akka.NET v1.4.14.


ismaelhamed:
@Aaronontheweb Yep, you can't have multiple overloads where all the params are optional, that's why I only introduced a contructor overload in `EventEnvelope`. I understand we should try not to make breaking changes as much as possible, but sometimes it is what it is. I should've left a comment though, my bad.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -684,7 +684,7 @@ public virtual async Task InsertBatchAsync(DbConnection connection, Cancellation
                     var evt = entry.Key;
                     var tags = entry.Value;
 
-                    WriteEvent(command, evt.WithTimestamp(DateTime.Now.Ticks), tags);
+                    WriteEvent(command, evt.WithTimestamp(DateTime.UtcNow.Ticks), tags);
```
ismaelhamed, line 687: Good catch

src/core/Akka.Persistence/Persistent.cs
```diff
@@ -369,7 +369,7 @@ public IPersistentRepresentation WithManifest(string manifest)
         /// <inheritdoc />
         public IPersistentRepresentation Update(long sequenceNr, string persistenceId, bool isDeleted, IActorRef sender, string writerGuid)
         {
-            return new Persistent(payload: Payload, sequenceNr: sequenceNr, persistenceId: persistenceId, manifest: Manifest, isDeleted: isDeleted, sender: sender, writerGuid: writerGuid);
+            return new Persistent(payload: Payload, sequenceNr: sequenceNr, persistenceId: persistenceId, manifest: Manifest, isDeleted: isDeleted, sender: sender, writerGuid: writerGuid, Timestamp);
```
ismaelhamed, line 372: I didn't change this because it hasn't in the JVM and no test seems broken

Aaronontheweb: I'll double check that

Aaronontheweb: Reason I added it - seems like we could set the timestamp and then forget it with this method.

Aaronontheweb: Fixed

PR #4685, opened 2020-12-22
Auto-tuning DotNetty batching + removing scheduler from batching system

Aaronontheweb:
Ported the `FlushConsolidationHandler` from Netty and used it to auto-tune batching inside Akka.Remote without having to set explicit thresholds and without having to rely on the `IScheduler`.

This accomplishes:

1. Lower Idle CPU consumption than even https://github.com/akkadotnet/akka.net/pull/4678
2. Significantly lower latencies on systems that don't write heavily
3. No need for users to manually tune their threshold settings in `akka.remote.dot-netty.tcp.batching` - this is now handled automatically by the `FlushConsolidationHandler`.

close https://github.com/akkadotnet/akka.net/issues/4636
close https://github.com/akkadotnet/akka.net/issues/4563

We still have some more work to do optimizing the `DedicatedThreadPool` to scale down automatically, but this eliminates most of the CPU noise coming from DotNetty.

### DIFF
--- a/src/common.props
+++ b/src/common.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Copyright>Copyright © 2013-2020 Akka.NET Team</Copyright>
     <Authors>Akka.NET Team</Authors>
-    <VersionPrefix>1.4.13</VersionPrefix>
+    <VersionPrefix>1.4.14</VersionPrefix>
     <PackageIconUrl>https://getakka.net/images/akkalogo.png</PackageIconUrl>
     <PackageProjectUrl>https://github.com/akkadotnet/akka.net</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/akkadotnet/akka.net/blob/master/LICENSE</PackageLicenseUrl>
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
## How It Works
The `FlushConsolidationHandler` works by batching flushes together, rather than writes, and in this patch we've changed the `TcpAssociationHandle` to always call `IChannel.WriteAndFlushAsync` - which creates a 1:1 correlation between writes and flushes.

The algorithm is designed to capture flushes that occur in rapid succession and group them together in order to lower the total number of system calls to the socket, which improves average throughput and CPU utilization. 

The flushes are batched together when:

1. The socket is currently performing a read or
2. The total number of flushes is less than `DefaultExplicitFlushAfterFlushes` or whatever the configured value is - defaults to 30 in Akka.NET.

The flushes are allowed to pass and write out to the socket when:

1. The socket is writing but not reading (immediate write);
2. The total number of flushes is __equal to__ `DefaultExplicitFlushAfterFlushes`; or 
3. A flush has been scheduled onto the `EventLoop` without being cancelled.

In the third case, we don't use a time-based delay to flush the socket - instead the "flush" call is simply added to the same event queue where all of the read and write events are. It works identically to an actor's mailbox. All of the writes queued up prior to that event get flushed at the same time.

src/common.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <Copyright>Copyright © 2013-2020 Akka.NET Team</Copyright>
     <Authors>Akka.NET Team</Authors>
-    <VersionPrefix>1.4.13</VersionPrefix>
+    <VersionPrefix>1.4.14</VersionPrefix>
```
IgorFedchenko, line 5: I guess this should not be changed in this PR, it's just `build,cmd` updated this file when was running locally?

src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
```diff
@@ -6,181 +6,242 @@
...
-        private ICancelable _flushSchedule;
+        public override void Flush(IChannelHandlerContext context)
+        {
+            if (_readInProgress)
+            {
+                // If there is still a read in progress we are sure we will see a channelReadComplete(...) call. Thus
+                // we only need to flush if we reach the explicitFlushAfterFlushes limit.
+                if (++_flushPendingCount == ExplicitFlushAfterFlushes)
+                {
+                    FlushNow(context);
```
IgorFedchenko, line 95: So when reading is complete we will flush in `ChannelReadComplete` anyway. But we are trying to flush right here if there are too many flushes are pending? Are we able to flush while reading is in process? Or this is safe?

IgorFedchenko (approved):
LGTM - just left minor note and a question about flushing while reading is in process

src/common.props
Aaronontheweb: That's correct

src/core/Akka.Remote/Transport/DotNetty/BatchWriter.cs
Aaronontheweb: So if a read is sitting in the channel pipeline right now we assume it means that some data is being read by the current application and that might, presumably, be used to produce a response right away - we're trying to batch those writes together into as few flushes as possible and when the read "completes" (we've finished reading all of the data currently inside the buffer) it's safe to flush any writes that are currently pending. This helps reduce latency in lower traffic system AND increases throughput in higher traffic systems.

PR #4686, opened 2020-12-23
fix NRE inside RemotingTerminator

### LINKED ISSUE #4677: NullReferenceException when updating to 1.4.13 version
Hi guys,
We were using distributed data in a 3 machine cluster with version 1.4.10. We had an sporadic issue that was fixed in version 1.4.13. 
We have updated the following to version 1.4.13 to check that the issue is resolved and we receive an exception, without changing the code.

I've made a test project using Akka.Cluster nuget package with different versions (it's the only nuget the project has) and it starts to throw the exception in version 1.4.11.

The exception is thrown when we create the actor system. After reading the docs, we still can not figure it out why this exception started to be thrown. Any clue would be much appreciated. We believe that it has to be with the issue: https://github.com/akkadotnet/akka.net/issues/4530

This is the exception:

System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Akka.Remote
  StackTrace:
   at Akka.Remote.RemoteActorRefProvider.get_Transport() in D:\a\1\s\src\core\Akka.Remote\RemoteActorRefProvider.cs:line 156
   
This is the only file in the test project:

namespace AkkaTest.Poc.Working
{
    using System;
    using Akka.Actor;
    using Akka.Configuration;

    class Program
    {
        {
            var akkaConfigurationContent = @"
                akka 
                {
                    actor.provider = ""cluster""
                    remote 
                    {
                        dot-netty
                        {
                            tcp
                            {
                                port = 8091
                                hostname = ""localhost""
                            }
                        }
                    }
	                cluster
                    {
                        seed-nodes = 
                        [
                            ""akka.tcp://the-cluster@localhost:8091""
                        ]
                    }
                }";

            var config = ConfigurationFactory.ParseString(akkaConfigurationContent);

            var actorSystem = ActorSystem.Create("the-cluster", config);

            Console.ReadLine();

        }
    }
}

This is the stack trace of the thread where the exception is thrown:
image

Aaronontheweb:
close #4677 

This patch also fixes all NRE's originating from the logging system for any `FSM<TState,TData` inside Akka.NET


src/core/Akka.Remote/RemoteActorRefProvider.cs
```diff
@@ -713,8 +713,7 @@ private void InitFSM()
             {
                 When(TerminatorState.Uninitialized, @event =>
                 {
-                    var internals = @event.FsmEvent as Internals;
-                    if (internals != null)
+                    if (@event.FsmEvent is Internals internals)
```
IgorFedchenko, line 716: It's always lovely to see how new language features make old code pretty :+1:

src/core/Akka/Actor/FSM.cs
```diff
@@ -1100,20 +1100,20 @@ private static StateFunction OrElse(StateFunction original, StateFunction fallba
...
-                return true;
-            }
-
-            if (message is Timer timer)
-            {
-                if (ReferenceEquals(timer.Owner, this) && _timers.TryGetValue(timer.Name, out var oldTimer) && oldTimer.Generation == timer.Generation)
+                case Timer timer:
                 {
+                    if (!ReferenceEquals(timer.Owner, this) || !_timers.TryGetValue(timer.Name, out var oldTimer) ||
+                        oldTimer.Generation != timer.Generation) return true;
```
IgorFedchenko, line 1116: Can we move this `return true` statement to the new line, or reformat thing somehow to see it more clearly here?

src/core/Akka/Actor/FSM.cs
```diff
@@ -671,7 +671,7 @@ public override string ToString()
     /// <typeparam name="TData">The state data type</typeparam>
     public abstract class FSM<TState, TData> : FSMBase, IListeners, IInternalSupportsTestFSMRef<TState,TData>
     {
-        private readonly ILoggingAdapter _log = Context.GetLogger();
+        private readonly ILoggingAdapter _log = Context.GetLoggerStartup();
```
IgorFedchenko, line 674: So - does this mean that if logger did not get the remote context on start, then this actor instance will never log with remoting context?
I am not familiar with how this "remoting context" thing impacts the logger output, but just thinking if there is some way to give it a "retry" mechanism so that once remoting is up and running, the loggers could catch it up somehow. 

Well, maybe `GetLoggerStartup` could receive `ILoggingAdapter currentLogger` parameter and see if this logger has remoting context skipped and has to be replaced, or `currentLogger` is good to go... Too complicated?

Of course my thoughts make sense only if having remoting context in logger is important - otherwise, current fix should be everything we need 👌

IgorFedchenko (commented):
That's a great change - definitely need to make loggers creation to be a safe operation.

Left some comments, not sure about ones which are non-refactoring though

src/core/Akka/Actor/FSM.cs
Aaronontheweb: Yeah I'll take a look at cleaning that up - the original code was pretty messy too

src/core/Akka/Actor/FSM.cs
Aaronontheweb: > So - does this mean that if logger did not get the remote context on start, then this actor instance will never log with remoting context?

That's correct - should only happen for a small number of `/system` actors that are launched before the transport is bound (i.e. the ones responsible for managing that transport)

Aaronontheweb: > I am not familiar with how this "remoting context" thing impacts the logger output, but just thinking if there is some way to give it a "retry" mechanism so that once remoting is up and running, the loggers could catch it up somehow.

What we do currently is basically this - crash the actor and re-attempt to acquire the transport after it's been bound. Unfortunately when Visual Studio is attached users intercept the error and report it, which has happened in at least 3 separate issues now. This is my attempt to put it to bed for good.

IgorFedchenko: Ah, I see - then there is nothing to worry about

src/core/Akka/Actor/FSM.cs
IgorFedchenko: So now this is going to be the only change I would like to suggest for this PR

Aaronontheweb: Resolved this, I believe

IgorFedchenko (approved):
LGTM

PR #4692, opened 2020-12-24
Add new test to AkkaSerializationSpec. Failed for HyperionSerializer.

seriousbox:
An error occurred while serializing the `ActorPath` collection with duplicate paths. For the last `ActorPath`, `Hyperion.ValueSerializers.ObjectSerializer` is used instead of `Hyperion.ValueSerializers.FromSurrogateSerializer`. This causes the error `System.InvalidCastException: Unable to cast object of type 'Surrogate' to type 'Akka.Actor.ActorPath'`. For the `Poco` (`ISurrogated`) collection of objects no error occurs regardless of the stored values.

In the test I gave an example with a simple array of elements, but in a real task the following class `State` is used.

```c#
public class Scheduled
{
    public Scheduled(string scheduleId, ActorPath executor)
    {
        ScheduleId = scheduleId;
        Executor = executor;
    }

    public string ScheduleId { get; }
    public ActorPath Executor { get; }
}

public sealed class State
{
    private readonly ImmutableDictionary<string, Scheduled> _entries;

    public State()
    {
        _entries = ImmutableDictionary<string, ScheduledTest>.Empty;
    }

    public State(IEnumerable<Scheduled> entries)
    {
        _entries = ImmutableDictionary.CreateRange(entries.Select(e => new KeyValuePair<string, Scheduled>(e.ScheduleId, e)));
    }

    public IEnumerable<Scheduled> Entries => _entries.Values;
}
```

### DIFF
--- a/src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
+++ b/src/contrib/serializers/Akka.Serialization.TestKit/AkkaSerializationSpec.cs
@@ -384,6 +384,25 @@ my-settings{
         }
 
         [Fact]
+        public void CanSerializeActorPathCollection()
+        {
+            var collection = new []
+            {
+                ActorPath.Parse("akka.tcp://sys@localhost:9100/user1/actor"),
+                ActorPath.Parse("akka.tcp://sys@localhost:9100/user2/actor"),
+                ActorPath.Parse("akka.tcp://sys@localhost:9100/user1/actor")
+            };
+            var serializer = Sys.Serialization.FindSerializerFor(collection);
+            var serialized = serializer.ToBinary(collection);
+            var deserialized = (ActorPath[])serializer.FromBinary(serialized, typeof(ActorPath[]));
+
+            Assert.Equal(collection.Length, deserialized.Length);
+            Assert.Equal(collection[0], deserialized[0]);
+            Assert.Equal(collection[1], deserialized[1]);
+            Assert.Equal(collection[2], deserialized[2]);
+        }
+
+        [Fact]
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
hi @seriousbox - @eaba is looking into this now

eaba:
The root of this issue can be found in Hyperion:
````csharp
if (session.Serializer.Options.PreserveObjectReferences && session.TryGetObjectId(value, out existingId))
                {
                    //write the serializer manifest
                    ObjectReferenceSerializer.Instance.WriteManifest(stream, session);
                    //write the object reference id
                    ObjectReferenceSerializer.Instance.WriteValue(stream, existingId, session);
                }
````
It just return the object as it was written:
````csharp
public override object ReadValue(Stream stream, DeserializerSession session)
        {
            var id = stream.ReadInt32(session);
            var obj = session.GetDeserializedObject(id);
            return obj;
        }
````
@Aaronontheweb can't `ObjectReferenceSerializer` hold the `Manifest` of the actual Serializer?
One way to avoid this issue is to turn off `PreserveObjectReferences`

eaba:
How can `ObjectReferenceSerializer` be improved to hold the Manifest of the actual `Serializer` for the object without any breaking changes? Suggestion are welcomed

Arkatufus:
After reviewing the code, I've found several bugs that complicates this issue, each one caused by problems in the `Hyperion` code base. I'm moving this to the `Hyperion` repo instead.

Arkatufus:
@eaba please look at https://github.com/akkadotnet/Hyperion/issues/200

eaba:
@Arkatufus okay

PR #4693, opened 2020-12-25
Update ActorTaskSchedulerMessage with defensive coding to prevent possible NRE problems in the future.

### LINKED ISSUE #4645: Both _scheduler and _task can be null inside ActorTaskScheduler
https://github.com/akkadotnet/akka.net/blob/ec01db19f1c49aecc6e3e828d264156a628ebadf/src/core/Akka/Dispatch/SysMsg/ISystemMessage.cs#L635

And the way this code is called inside `ActorCell`:

https://github.com/akkadotnet/akka.net/blob/ec01db19f1c49aecc6e3e828d264156a628ebadf/src/core/Akka/Actor/ActorCell.DefaultMessages.cs#L337-L351

Can potentially throw an NRE depending on the content included inside the message that is sent back to the actor. I haven't seen any user reports of that occurring in the ~5-6 years we've had this code so if it can happen it's probably rare, but still - this looks like a bit of a footgun.

Arkatufus:
Closes #4645

### DIFF
--- a/src/core/Akka/Dispatch/SysMsg/ISystemMessage.cs
+++ b/src/core/Akka/Dispatch/SysMsg/ISystemMessage.cs
@@ -602,8 +602,8 @@ namespace Akka.Dispatch.SysMsg
         /// <param name="message">TBD</param>
         public ActorTaskSchedulerMessage(ActorTaskScheduler scheduler, Task task, object message)
         {
-            _scheduler = scheduler;
-            _task = task;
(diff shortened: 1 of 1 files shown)


Aaronontheweb (approved):
LGTM

Aaronontheweb:
Linux build timed out - going to see if I can recreate it. Not sure why that happened - looks like it's related to the Reactive.Streams.TCK.

PR #4696, opened 2020-12-28
FSharp Quotations Evaluator to 2.1.0

eaba:
When FSharp.Quotations.Evaluator is Bumped to 2.1.0 `can spawn actor from expression` fails

image

image

image

image


eaba:
FSharp Quotations Evaluator to 1.1.3
image

eaba:
FSharp Quotations Evaluator to 2.1.0
image

src/core/Akka.FSharp/FsApi.fs
```diff
@@ -374,9 +379,12 @@ module Linq =
...
     let toExpression<'Actor>(f : System.Linq.Expressions.Expression) = 
             match f with
-            | Lambda(_, (Call(null, Method "ToFSharpFunc", Ar [| Lambda(_, p) |]))) 
-            | Call(null, Method "ToFSharpFunc", Ar [| Lambda(_, p) |]) -> 
-                Expression.Lambda(p, [||]) :?> System.Linq.Expressions.Expression<System.Func<'Actor>>
+            | Lambda(_, body) -> 
+                match body with  
+                | :? ConstantExpression as cst -> Some(cst.Value, cst.Type)
+                | _ -> failwith "Doesn't match"
+            //| Call(null, Method "ToFSharpFunc", Ar [| Lambda(_, p) |]) ->
```
eaba, line 386: @Aaronontheweb am still pondering if this will have some side effects. Right now are tests are passing!

src/core/Akka.FSharp/FsApi.fs
```diff
@@ -374,9 +379,12 @@ module Linq =
 
     let toExpression<'Actor>(f : System.Linq.Expressions.Expression) = 
             match f with
-            | Lambda(_, (Call(null, Method "ToFSharpFunc", Ar [| Lambda(_, p) |]))) 
-            | Call(null, Method "ToFSharpFunc", Ar [| Lambda(_, p) |]) -> 
-                Expression.Lambda(p, [||]) :?> System.Linq.Expressions.Expression<System.Func<'Actor>>
+            | Lambda(_, body) -> 
+                match body with  
+                | :? ConstantExpression as cst -> Some(cst.Value, cst.Type)
+                | _ -> failwith "Doesn't match"
```
eaba, line 385: Added `Call` here but got error - maybe there is a way to add it to the switch

Aaronontheweb:
My own F# experience is too limited to properly review this, so I asked for help here: https://twitter.com/Aaronontheweb/status/1343973365778968576

src/core/Akka.FSharp/FsApi.fs
voronoipotato: This might just be bike shedding or nitpicking but I wonder if the result type might be a good fit here. That way you can get your fail but not have to worry about consumers forgetting to catch.

Aaronontheweb: What would that look like? My F# experience is limited to managing FAKE scripts

voronoipotato: Result type is a lot like Option, but with an Ok<T> | Error<U> instead of a Some<T> | None. You can pass along a message about what went wrong in the err, or what specifically didn't match and the consumer is forced to cover the possible failure. It communicates to the consumer that "Failure is something that can happen, and may happen and you need to have a response prepared to it". Now some people will match on the Error and just failwith as you did here, but they must make that decision.

voronoipotato: For an example... 
```fsharp
    match body with  
    | :? ConstantExpression as cst -> Ok(cst.Value, cst.Type)
    | _ -> Error (sprintf  "Unable to identify the expression %s" typeof<body>)
```

you could also just put 
```fsharp
    match body with  
    | :? ConstantExpression as cst -> Ok(cst.Value, cst.Type)
    | _ -> Error "Doesn't match"`
```

The consumer basically consumes it the same way they would Some | None, but they use 
```fsharp
match result with
| Ok(val,typename) -> failwith "not implemented"
| Error message -> failwith "not implemented"
```

Appologies if there's any slop there, I'm in a bit of a hurry lol. I'll be back later.

Aaronontheweb: @eaba think you can do this? https://twitter.com/7sharp9_/status/1344010659021922306

baronfel:
@eaba I have some details that may help you out.

The change in Expression shape you're seeing is because from the prior version of FSharp.Quotations.Evaluator to this version, the internal representation of the evaluated expression seems to have change, so the match statements in `toExpression` in this code base no longer align.

If you look [here](https://github.com/fsprojects/FSharp.Quotations.Evaluator/blob/a9a25de1e8c6fa1a20da175349c52ef836bb7bc8/src/FSharp.Quotations.Evaluator/Tools.fs#L57-L196), you can see a series of optimized closure functions that FQE uses now to reduce the overhead of calling F# functions. These types are then consumed/created [here](https://github.com/fsprojects/FSharp.Quotations.Evaluator/blob/a9a25de1e8c6fa1a20da175349c52ef836bb7bc8/src/FSharp.Quotations.Evaluator/QuotationsEvaluator.fs#L454-L567) for F# Lambda expressions to make System.Linq.Expressions.Expressions.

This all goes to say that if we look at the lambda returned (".Lambda #Lambda1<System.Func`2[Microsoft.FSharp.Core.Unit,FSharp.Quotations.Evaluator.Tools+FuncFSharp`3[Microsoft.FSharp.Core.Unit,Microsoft.FSharp.Core.Unit,Akka.FSharp.Tests.ApiTests+TestActor]]>(Microsoft.FSharp.Core.Unit $var1)\r\n{\r\n    .Constant<FSharp.Quotations.Evaluator.Tools+FuncFSharp`3[Microsoft.FSharp.Core.Unit,Microsoft.FSharp.Core.Unit,Akka.FSharp.Tests.ApiTests+TestActor]>(FSharp.Quotations.Evaluator.Tools+FuncFSharp`3[Microsoft.FSharp.Core.Unit,Microsoft.FSharp.Core.Unit,Akka.FSharp.Tests.ApiTests+TestActor])}"

This is to say that now, we should expand this set of matches to cover all of the different 'exit points' of the lambda match provided by FQE:
* calling `FuncConvert.ToFSharpFunc` for lambdas with more than 19 args: https://github.com/fsprojects/FSharp.Quotations.Evaluator/blob/a9a25de1e8c6fa1a20da175349c52ef836bb7bc8/src/FSharp.Quotations.Evaluator/QuotationsEvaluator.fs#L452
* Lambdas that are calls to this underlying function helper with no captured parameters (eg unit): https://github.com/fsprojects/FSharp.Quotations.Evaluator/blob/a9a25de1e8c6fa1a20da175349c52ef836bb7bc8/src/FSharp.Quotations.Evaluator/QuotationsEvaluator.fs#L528
* lambdas that capture parameters from the executing environment: https://github.com/fsprojects/FSharp.Quotations.Evaluator/blob/a9a25de1e8c6fa1a20da175349c52ef836bb7bc8/src/FSharp.Quotations.Evaluator/QuotationsEvaluator.fs#L561-L567

baronfel:
A bit more detail: Here's what I consider to be a more full-featured fix at [my fork](https://github.com/akkadotnet/akka.net/commit/c46be4cee26beb7bea08542b1c22c07a7dc6f05e).

I've extracted the constant value and invoked the `Invoke` member, passing along the parameters, for the unit-case mentioned in my comment above, and as a result the tests fail at a _new and exciting_ location:

```
[xUnit.net 00:00:09.46]     Akka.FSharp.Tests.ApiTests.can spawn actor from expression [FAIL]
  Failed Akka.FSharp.Tests.ApiTests.can spawn actor from expression [8 s]
  Error Message:
   System.InvalidCastException : Unable to cast object of type 'System.Linq.Expressions.InstanceMethodCallExpression1' to type 'System.Linq.Expressions.NewExpression'.
  Stack Trace:
     at Akka.Util.Internal.Extensions.AsInstanceOf[T](Object self) in E:\Code\akka.net\src\core\Akka\Util\Internal\Extensions.cs:line 27
   at Akka.Actor.Props.Create[TActor](Expression`1 factory, SupervisorStrategy supervisorStrategy) in E:\Code\akka.net\src\core\Akka\Actor\Props.cs:line 397
   at Akka.FSharp.Spawn.spawnObjOpt[a](IActorRefFactory actorFactory, String name, FSharpExpr`1 f, FSharpList`1 options) in E:\Code\akka.net\src\core\Akka.FSharp\FsApi.fs:line 574
   at Akka.FSharp.Spawn.spawnObj[a](IActorRefFactory actorFactory, String name, FSharpExpr`1 f) in E:\Code\akka.net\src\core\Akka.FSharp\FsApi.fs:line 585
   at Akka.FSharp.Tests.ApiTests.can spawn actor from expression() in E:\Code\akka.net\src\core\Akka.FSharp.Tests\ApiTests.fs:line 136
```

Props.Create expects specifically a `NewExpression` only to be the result, but this change in FQE seems to pre-compile the expression inside the quotation, and it's not easy to get at that thing to investigate it.

baronfel:
So even if you use reflection to get at the already-compiled inner-function, I'm not sure this is going to work, because the inner expression has been compiled down by FQE [here](https://github.com/fsprojects/FSharp.Quotations.Evaluator/blob/a9a25de1e8c6fa1a20da175349c52ef836bb7bc8/src/FSharp.Quotations.Evaluator/QuotationsEvaluator.fs#L509) and we can't reverse that out in any way I'm aware of. We may have to sidestep using FQE for these expressions and use the quotation traversals in FSharp.Core directly, unless FQE learns a flag for optionally skipping this optimization on their part.

baronfel:
It might be easier at this point to sidestep using FQE entirely, if our only goal is to parse simple actor expressions.  At my fork [here](https://github.com/akkadotnet/akka.net/commit/b4833fe30c9dff8d2239edd1fb2f0b194a70214d) is the minimal work necessary to understand the example in the tests, and it works with quotations that are a simple lambda and that pass in constant values. This has a few benefits:

* it's very easy to add on to what you as a project want to support
* you're not tied to an external dependency
*

baronfel:
Thinking on this a bit more, I've added tests [here](https://github.com/akkadotnet/akka.net/commit/176d98c9789192342cb8f53ed60178e212349b9b) and a bit more quotation traversal so that spawning actors that look like this work:

```fsharp
[<Fact>]
let ``can spawn actor with constant args from expression`` () =
    let system = Configuration.load() |> System.create "test"
    let actor = spawnObj system "test-actor" <@ fun () -> TestActorWithArgs(box 1, box true, box "yo") @>
    ()

[<Fact>]
let ``can spawn actor with captured args from expression`` () =
    let system = Configuration.load() |> System.create "test"
    let arg1 = 1
    let arg2 = true
    let arg3 = "yo"
    let actor = spawnObj system "test-actor" <@ fun () -> TestActorWithArgs(box arg1, box arg2, box arg3) @>
    ()
```

These are all constant values, but the mechanism is general enough to handle parameters of non-constant values as well.  I added these tests to verify things, and it all seems to be working quite well.  I think this could be a good direction for this project to go, as it decreases the blast radius of dependency changes (as shown by this MR), and also makes it quite clear what exactly is supported/could be extended in the future.

eaba:
Thank you @baronfel for the BIG GENEROUS HAND!!!

Aaronontheweb (approved):
LGTM

PR #4700, opened 2020-12-29
Move RouterActor routing logic controller actor instantiation from OnReceive to constructor

### LINKED ISSUE #4679: Cluster.Metrics Adaptive Load Balancer not collecting metrics
When using the Adaptive Load Balancer with a Cluster Group Router the metrics are not collected by the AdaptiveLoadBalancingMetricsListener unless a new router management message that is not handled by the ClusterRouterActor and the RouterActor is sent to the router,

`
       // RouterActor
       
        {
            switch (message)
            {
                case GetRoutees getRoutees:
                    Sender.Tell(new Routees(Cell.Router.Routees));
                    break;
                case AddRoutee addRoutee:
                    Cell.AddRoutee(addRoutee.Routee);
                    break;
                case RemoveRoutee removeRoutee:
                    Cell.RemoveRoutee(removeRoutee.Routee, stopChild: true);
                    StopIfAllRouteesRemoved();
                    break;
                case Terminated terminated:
                    Cell.RemoveRoutee(new ActorRefRoutee(terminated.ActorRef), stopChild: false);
                    StopIfAllRouteesRemoved();
                    break;
                default:
                    // Any other unhandled management message wil be forwarded to the controller and start collecting metrics
                    RoutingLogicController?.Forward(message);
                    break;
            }
        }
`

- Akka 1.14.12
- dotnet core 3.1

Arkatufus:
Closes #4679

`AdaptiveLoadBalancingMetricsListener` will never get created until `RouterActor` receives at least one routing message because the routing logic actor is instantiated inside `OnReceive` instead of the class constructor. This can be problematic for `AdaptiveLoadBalancingGroup` router.


Aaronontheweb (approved):
LGTM

PR #4708, opened 2021-01-05
[Akka.DependencyInjection] Microsoft.Extensions.DependencyInjection support

### LINKED ISSUE #4590: Proper Microsoft.Extensions.DependencyInjection Support, Akka.DI.Common Rewrite, and Actor Integration
The short version of this issue:

Now that `ActorSystemSetup` is available, it allows us to easily pass in external code dependencies into an `ActorSystem` in an immutable way before it gets instantiated. It should now be possible to accomplish the following:

1. Rewrite Akka.DI.Common to be immutable, deprecating the need for those mutable methods that will let someone swap out kernels at runtime;
2. Allow us to properly support Microsoft.Extensions.DependencyInjection in the core APIs for making DI more transparent inside Akka.NET;
3. Allow us to accomplish item 2 without explicitly taking a dependency on `Microsoft.Extenions.*`, which resolves one of my bugaboos for years; and
4. Should make it really easy for actors running inside the `ActorSystem` to grab ahold of the current DI container without end-users needing to manually pass it around anyway.

I want to revisit the trail that @Horusiath blazed with https://github.com/akkadotnet/akka.net/pull/3862 - never felt comfortable changing the default pipeline to use Microsoft.Extensions.DependencyIjnection directly (QA surface area is enormous) but making it possible to opt-into using the pipeline using an updated `Setup` introduced by Akka.DI.Common should take the risk factor down significantly.

Going to open this issue to cover the architectural bits, user requests, and comments from contributors.

Should also help resolve https://github.com/akkadotnet/akka.net/issues/1675, https://github.com/akkadotnet/akka.net/issues/3050, and https://github.com/akkadotnet/akka.net/issues/1267

Aaronontheweb:
close #4590 

I took a lot of inspiration from the code samples other contributors submitted - and ultimately decided to go with a very, very simple approach for integrating Microsoft.Extensions.DependencyInjection into Akka.NET.

The short version of what this pull request does:

1. **Introduces a new module, Akka.DependencyInjection, which will replace Akka.DI.Core and all of the libraries which depend on it**. This seemed like the easiest way to preserve backwards compatibility - simple deprecate the old package and replace it with a new one.
2. **Allows users to pass in an [`IServiceProvider`](https://docs.microsoft.com/en-us/dotnet/api/system.iserviceprovider) into the `ActorSystem` before the latter is created, via [a new kind of programmatic configuration `Setup` that was introduced in Akka.NET v1.4](https://getakka.net/articles/concepts/configuration.html#programmatic-configuration-with-setup)**. This ensures that the `IServiceProvider` is immutable within an `ActorSystem`, a problem that we have currently with Akka.DI.Core.
3. **Makes it easy to access the `IServiceProvider` via the `Akka.DependencyInjection.ServiceProvider.For(ActorSystem)` method**, which is what we can use to get access to the service provider in order to consume required services inside our actors.
4. **Makes it possible to dependency inject some parameters will dynamically specifying some others**, something that users have been asking for since 2015.

I kept the code extremely simple - Akka.DependencyInjection does not try to perform magic of any kind. It simply makes the process of creating a `Props` class that consumes `IServiceProvider` extremely trivial. See my comments on the PR code itself to get an idea for how it will work in practice.


src/contrib/dependencyinjection/Akka.DI.Core/Akka.DI.Core.csproj
```diff
@@ -3,7 +3,7 @@
 
   <PropertyGroup>
     <AssemblyTitle>Akka.DI.Core</AssemblyTitle>
-    <Description>Dependency injection support for Akka.NET</Description>
+    <Description>Dependency injection support for Akka.NET [DEPRECATED - replaced by Akka.DependencyInjection]</Description>
```
Aaronontheweb, line 6: Deprecation warning in the description for the old package. I'll also mark the package as deprecated on NuGet and in a subsequent PR, mark the old code as obsolete.

src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProvider.cs
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /// <summary>
+    /// This class represents a specialized <see cref="Akka.Actor.Props"/> that uses delegate invocation
+    /// to create new actor instances, rather than a traditional <see cref="System.Activator"/>.
+    ///
+    /// Relies on having an active <see cref="IServiceProvider"/> implementation available
+    /// </summary>
+    /// <typeparam name="TActor">The type of the actor to create.</typeparam>
+    internal class ServiceProviderProps<TActor> : Props where TActor : ActorBase
```
Aaronontheweb, line 93: So this class, the `ServiceProviderProps<TActor>`, is where the real magic happens. This class takes a delegate of type `Func<IServiceProvider, TActor>` and invokes that in order to create a new actor instance.

That function will be re-run each time the actor is restarted or each time a new actor is created using the same `Props` - so it's very easy to do things like:

1. Create a new `IServiceScope` and pass that into the actor, so the actor can manage its own lifecycle dependencies upon start / restart;
2. Manually resolve a dependency using `IServiceProvider.GetRequiredService<T>` and pass that directly into the actors' constructor - nice to have in the event that your actor doesn't need any scoped dependencies.

src/contrib/dependencyinjection/Akka.DependencyInjection.Tests/ActorServiceProviderPropsWithScopesSpecs.cs
```diff
@@ -0,0 +1,199 @@
...
+            var deps2 = ExpectMsg<CurrentDependencies>();
+            deps2.Dependencies.All(x => x.Disposed).Should().BeFalse();
+        }
+
+        [Fact(DisplayName =
+            "DI: actors who receive a mix of dependencies via IServiceScope should dispose ONLY of their scoped dependencies and recreate upon restart")]
+        public void ActorsWithMixedDependenciesShouldDisposeAndRecreateScopedUponRestart()
+        {
+            var spExtension = ServiceProvider.For(Sys);
+            var props = spExtension.Props(sp => new MixedActor(sp.GetRequiredService<AkkaDiFixture.ISingletonDependency>(), sp.CreateScope()));
```
Aaronontheweb, line 91: An example of mitching and maxing DI scopes here - note that the container _will_ do this for you since the dependencies are scoped properly inside our class fixture, but I wanted an example inside the test suite.

Note: you could also pass in an closed-over argument here that would be included inside the `Props` for the actor when it's created. This is how you can mix and match DI easily.

Theoretically, using this method, you could also create the actor instance _itself_ with DI:

```
spExtension.Props(sp => sp.GetService<MixedActor>());
```

That would also work.

Aaronontheweb (commented):
Added some comments explaining the key features of my changes

Aaronontheweb:
I tried to keep this as simple as possible - no cute magic trying to guess how to scope users' dependencies. I wanted to make it as simple as possible for them to do it themselves and I think this succeeds in doing that.

src/contrib/dependencyinjection/Akka.DependencyInjection.Tests/ActorServiceProviderPropsWithScopesSpecs.cs
Zetanova: ```spExtension.Props(sp => sp.GetService<MixedActor>());``` will not work, because MixedActor is not registered as a transient service into the ServiceCollection

This is the why to go und should have some kind of shorter support:
```spExtension.Props(sp => ActivatorUtilities.CreateInstance<MixedActor>(sp, "extraParam1", "extraParam2"));```

Props have a Arguments array that can me used for the extra params to make this like the already overload ```spExtension.Props<MixedActor>("extraParam1", "extraParam2");```
Props.Arguments would then hold the extra params

The Props should then call following method to create the actor instance internal:
```csharp
public override ActorBase NewActor() { 
   //...
   var instance = (ActorBase)ActivatorUtilities.CreateInstance(sp, Type, Arguments);
   //...
   return instance;
}
```

In the simplest form the ```Props.Arguments``` are empty.

Maybe the SP-Props could replace Normal-Props as default actor factory (configurable).

An other long standing question for me is the lifetime of the Props instance itself.
There is a small logical Problem. A Props-Instance hold/reference all Parameters  
for the lifetime of the actor. It can be accessed by ```Context.Props.Arguments```
But all example show ````Props.Create(() => new MyActor(param1, param2))``` 
that would create an capture-scope and delegate loaded for the whole life of the actor,
even with a Stopping Strategy (do not restart)

Aaronontheweb: I like it - I'll add it to this PR.

The `MixedActor` sample will work, however, as the service we're injecting via a manual call is a singleton so having a registered service isn't necessary in this instance, but your point is taken.

Aaronontheweb: So I just updated this in my most recent commit - problem: actors created using `ActivatorUtilities` and DI container registration still have to dispose of their own dependencies, by the looks of things. Had to implement a manual `IDisposable` method to get the test to pass.

src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProvider.cs
```diff
@@ -0,0 +1,145 @@
...
+        ///// Uses a delegate to dynamically instantiate an actor where some of the constructor arguments are populated via dependency injection
+        ///// and others are not.
+        ///// </summary>
+        ///// <remarks>
+        ///// YOU ARE RESPONSIBLE FOR MANAGING THE LIFECYCLE OF YOUR OWN DEPENDENCIES. AKKA.NET WILL NOT ATTEMPT TO DO IT FOR YOU.
+        ///// </remarks>
+        ///// <typeparam name="T">The type of actor to instantiate.</typeparam>
+        ///// <param name="producer">The delegate used to create a new instance of your actor type.</param>
+        ///// <returns>A new <see cref="Props"/> instance which uses DI internally.</returns>
+        //public Props Props<T>(Func<IServiceProvider, T> producer) where T : ActorBase
```
Aaronontheweb, line 60: if I get approval on the API stuff so far, I'll remove this before merging it in

src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProvider.cs
```diff
@@ -0,0 +1,145 @@
...
+            _provider = provider;
+        }
+
+        /// <summary>
+        /// Creates a new actor using the configured factory method.
+        /// </summary>
+        /// <returns>The actor created using the factory method.</returns>
+        public override ActorBase NewActor()
+        {
+            return ActivatorUtilities.CreateInstance<TActor>(_provider, Arguments);
```
Aaronontheweb, line 129: Removed the `Func<IServiceProvider, TActor>` delegate here due to the memory leak point that @Zetanova raised. This is much cleaner.

src/contrib/dependencyinjection/Akka.DependencyInjection.Tests/ActorServiceProviderPropsWithScopesSpecs.cs
```diff
@@ -0,0 +1,283 @@
...
+        public class NonDiArgsActor : ReceiveActor
+        {
+            private readonly AkkaDiFixture.ISingletonDependency _singleton;
+            private readonly IServiceScope _scope;
+            private AkkaDiFixture.ITransientDependency _transient;
+            private AkkaDiFixture.IScopedDependency _scoped;
+            private string _arg1;
+            private string _arg2;
+
+            public NonDiArgsActor(AkkaDiFixture.ISingletonDependency singleton, IServiceProvider sp, string arg1, string arg2)
```
Aaronontheweb, line 248: Using this to test out whether or not we can mix DI and constructor arguments provided directly via the `ServiceProvider.Props`.

src/contrib/dependencyinjection/Akka.DependencyInjection.Tests/ActorServiceProviderPropsWithScopesSpecs.cs
```diff
@@ -0,0 +1,283 @@
...
+        }
+
+        [Fact(DisplayName =
+            "DI: actors who receive a non-DI'd dependencies should be started correctly")]
+        public void ActorsWithNonDiDependenciesShouldStart()
+        {
+            var spExtension = ServiceProvider.For(Sys);
+            var arg1 = "foo";
+            var arg2 = "bar";
+            var props = spExtension.Props<NonDiArgsActor>(arg1, arg2);
```
Aaronontheweb, line 128: Passing in the non-DI'd arguments here

Aaronontheweb (commented):
Updated code and comments based on @Zetanova 's feedback

Arkatufus (commented):
Looks good to me, I just have reservation with the .Net 5 dependency.

src/contrib/dependencyinjection/Akka.DependencyInjection/Akka.DependencyInjection.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <PropertyGroup>
+        <AssemblyTitle>Akka.DependencyInjection</AssemblyTitle>
+        <Description>Dependency injection support for Akka.NET</Description>
+        <TargetFrameworks>$(NetStandardLibVersion)</TargetFrameworks>
+        <PackageTags>$(AkkaPackageTags);dependency injection</PackageTags>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="5.0.0" />
```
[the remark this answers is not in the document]
Aaronontheweb: I think so. I can always try rolling it back - but it looks to me like it shouldn't be a massive change.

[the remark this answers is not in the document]
Aaronontheweb: I reviewed this and it is fine - .NET Standard 2.0 compatible

PR #4713, opened 2021-01-06
Upgraded to FluentAssertions 4.14.0 -> v5.10.3

Aaronontheweb:

PR #4715, opened 2021-01-07
upgraded test target for .NET framework to .NET 4.7.2

Aaronontheweb:
Added this to solve build problems related to https://github.com/akkadotnet/akka.net/pull/4713 - which comes from the .NET 4.6.1 target referencing the .NET 4.5 version of FluentAssertions 5.10, which in turn depends on the System.ValueTuple NuGet package - which causes a build conflict on Linux.

Moving off of .NET 4.6.1 and onto 4.7.1 forces the system to resolve that dependency as .NET Standard 2.0 instead.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -53,7 +53,7 @@ let runIncrementally = hasBuildParam "incremental"
 let incrementalistReport = output @@ "incrementalist.txt"
 
 // Configuration values for tests
-let testNetFrameworkVersion = "net461"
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Incorporated this change into https://github.com/akkadotnet/akka.net/pull/4713/commits/d0e83bfa15a99cdc1b94f6db9af19e80aac5a08e just to fully validate the fix. Worked on my machine.

PR #4739, opened 2021-01-22
Allow different versions of MS Abstractions nuget package for Akka.DependencyInjection

IgorFedchenko:
Instead of multi-targeting, we can just allow lower versions of MS nuget package to be used, since all we need is `ActivatorUtilities` which exists in a wide version range of that package.


Aaronontheweb (approved):
I'm at odds about what to do here to some extent largely because I'm not sure whether this is going to introduce a new set of problems.

The other alternative is just multi-targeting, which we can certainly do... But we have been trying to avoid that. 

I think we should do a quick release with this included and see how things go, since .NET Core 3.1 seems to be the dominant one and the only real stable release. .NET 5 users should be able to roll-forward on top of this without any issues either.

PR #4742, opened 2021-01-23
Allow Streams to be consumed as IAsyncEnumerable

to11mtm:
This pull request adds two methods on top of `Source<TOut,TMat>`:

 - `RunAsAsyncEnumerable`
 - `RunAsAsyncEnumerableBuffer` <-- allows overriding default buffer settings

This is a semi-naive but very workable implementation, utilizing `ISinkQueue<TOut>` alongside a Killswitch for disposal semantics.

This approach has the advantage of being concise and safe for users; multiple iterations re-run the stream which is probably what most users would expect.


src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -0,0 +1,120 @@
...
+using Akka.Streams.TestKit;
+using Akka.TestKit;
+using FluentAssertions;
+using Nito.AsyncEx.Synchronous;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Streams.Tests.Dsl
+{
+#if NETCOREAPP
```
to11mtm, line 23: You can't `await foreach` in the versions of Framework we are targeting with our test runners.

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -0,0 +1,120 @@
...
+            var task = Source.FromPublisher(probe).RunAsAsyncEnumerable(materializer);
+            materializer.Shutdown();
+            Func<Task> a = async () =>
+            {
+                await foreach (var a in task)
+                {
+                    
+                }
+            };
+            a.ShouldThrow<IllegalStateException>();
```
to11mtm, line 114: Wasn't sure if we should throw something different here.

src/core/Akka.Streams/Akka.Streams.csproj
```diff
@@ -63,6 +63,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Google.Protobuf" Version="$(ProtobufVersion)" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="5.0.0" />
```
to11mtm, line 66: @Aaronontheweb I'm so so sorry ;_;

src/core/Akka.Streams/Dsl/Source.cs
```diff
@@ -362,6 +362,41 @@ public Task<TOut> RunSum(Func<TOut, TOut, TOut> reduce, IMaterializer materializ
...
+        /// <param name="maxBuffer">The Max input buffer size.</param>
+        /// <returns>A lazy <see cref="IAsyncEnumerable{T}"/> that will run each time it is enumerated.</returns>
+        public IAsyncEnumerable<TOut> RunAsAsyncEnumerableBuffer(
+            IMaterializer materializer, int minBuffer = 4,
+            int maxBuffer = 16) =>
+            new StreamsAsyncEnumerableRerunnable<TOut>(
+                ViaMaterialized(KillSwitches.Single<TOut>(), Keep.Right)
+                    .ToMaterialized(
+                        Sink.Queue<TOut>().WithAttributes(
+                            Attributes.CreateInputBuffer(minBuffer, maxBuffer)),
```
to11mtm, line 396: Is this the right way to set this up?

src/core/Akka.Streams/Implementation/AsyncEnumerable.cs
```diff
@@ -0,0 +1,77 @@
...
+    public sealed class SinkQueueAsyncEnumerator<T> : IAsyncEnumerator<T>
+    {
+        private ISinkQueue<T> _sinkQueue;
+        private IKillSwitch _killSwitch;
+        private CancellationToken _token;
+        public SinkQueueAsyncEnumerator((UniqueKillSwitch killSwitch,ISinkQueue<T> sinkQueue) queueAndSwitch, CancellationToken token)
+        {
+            _sinkQueue = queueAndSwitch.sinkQueue;
+            _killSwitch = queueAndSwitch.killSwitch;
+            _token = token;
```
to11mtm, line 50: So, I'm honestly not sure if this is the right way to use the `CancellationToken` here or not.

src/core/Akka.Streams/Implementation/AsyncEnumerable.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+        public async ValueTask DisposeAsync()
+        {
+            //If we are disposing, let's shut down the stream
+            //so that we don't have data hanging around.
+            _killSwitch.Shutdown();
```
to11mtm, line 56: Since we are materializing every time we get enumerator, we can dispose neatly here by just hitting the killswitch.

to11mtm:
Hmm, it looks like behavior of one of the unit tests is Racy; it looks like shutting down the stream could cause -either- an `InvalidStateException` or an `AbruptTerminationException` depending on some timing.

Do we want to try to normalize to AbruptTermination? Or wrap these altogether?

src/core/Akka.Streams/Implementation/AsyncEnumerable.cs
to11mtm: Added tests to confirm this is the right way.

Horusiath:
Sorry, but why like this? AFAIK I already implemented `IAsyncEnumerable` sink in the past (together with ChannelReader/ChannelWriter stages). The thing is that in akka streams junction points with external world (outside of akka streams graph) are provided by materialized values, so `IAsyncEnumerable` stage has shape `Sink<T, IAsyncEnumerable<T>>`. Why?

Akka.Streams is build to provide composability - sometimes you may want to return more then one materialized value (eg. input channel to a graph and async enumerable output from a graph). This is possible via combinator operators and `Keep.Both` option. With the approach you proposed it's no longer possible.

to11mtm:
> 
> 
> Sorry, but why like this? AFAIK I already implemented `IAsyncEnumerable` sink in the past, together with ChannelReader/ChannelWriter stages (I think it was in AkkaContrib repo, which is no longer present). The thing is that in akka streams junction points with external world (outside of akka streams graph) are provided by materialized values, so `IAsyncEnumerable` stage has shape `Sink<T, IAsyncEnumerable<T>>`. Why?

Is that approach still somewhere? I came up with this because I could not find such an implementation, because you are right about below, this is definitely not 'composable' in the way Akka Streams normally provides (thus the `RunAs` naming.)

> 
> Akka.Streams is build to provide composability - sometimes you may want to return more then one materialized value (eg. input channel to a graph and async enumerable output from a graph). This is possible via combinator operators and `Keep.Both` option. With the approach you proposed it's no longer possible.

Horusiath:
I think, this could be implemented in more general terms of [Akka.Streams.Channels](https://github.com/akkadotnet/Alpakka/blob/dev/src/Channels/Akka.Streams.Channels/ChannelSink.cs#L52) I once did in Alpakka project. From there after materializing stream as channel reader, all that's left to represent it as `IAsyncEnumerable` is to use natively supported extension method [ReadAllAsync](https://docs.microsoft.com/en-us/dotnet/api/system.threading.channels.channelreader-1.readallasync?view=net-5.0). Example:

```csharp
using System.Threading.Channels;

var events =
    source
    .ToMaterialized(ChannelSink.AsReader<int>(bufferSize: 10), Keep.Right)
    .Run(materializer)
    .ReadAllAsync(cancellationToken);
    
await foreach (var e in events) {
  await Handle(e)
}
```

This is more generalized implementation: it supports channels, which AFAIK are part of std lib in .NET 5.0. I can make a PR if we're ready to include that into core akka.

Aaronontheweb:
> This is more generalized implementation: it supports channels, which AFAIK are part of std lib in .NET 5.0. I can make a PR if we're ready to include that into core akka.

One thing we're going to explore in Akka.NET v1.5, per some of @Zetanova's legwork, is replacing our current dispatching system that uses separate threadpools with one that uses separate channels on top of the shared .NET threadpool, since we want the benefits (exclusive scheduling) without the overhead (auto-scaling and managing threadpools on-demand.) We're hopeful that System.Threading.Channels can accomplish that.

So we're cool to add that as a dependency to Akka.Streams now, IMHO. I don't see the harm.

Aaronontheweb:
I want to take another look at this again....

Aaronontheweb:
Looks like we have a compilation error - I suspect it's related to a FluentAssertions upgrade we merged in after this PR was made.

Arkatufus (commented):
LGTM

PR #4751, opened 2021-02-02
Update finite-state-machine.md

brah-mcdude:
There is no "arrow operator" in .NET.


Aaronontheweb:
Good call. Thanks for the PRs!

PR #4756, opened 2021-02-02
harden Akka.Remote.Tests.Transport.DotNettySslSupportSpec.Secure_transport_should_be_possible_between_systems_using_thumbprint

Aaronontheweb:


src/core/Akka.Remote/Transport/DotNetty/AkkaLoggingHandler.cs
```diff
@@ -103,7 +103,9 @@ public override void ChannelRead(IChannelHandlerContext ctx, object message)
         {
             if (_log.IsDebugEnabled)
             {
-                _log.Debug("Channel {0} received a message ({1}) of type [{2}]", ctx.Channel, message, message == null ? "NULL" : message.GetType().TypeQualifiedName());
+                
+                // have to force a .ToString() here otherwise the reference count on the buffer might be illegal
+                _log.Debug("Channel {0} received a message ({1}) of type [{2}]", ctx.Channel, message?.ToString(), message == null ? "NULL" : message.GetType().TypeQualifiedName());
```
Aaronontheweb, line 108: So this may be responsible for all sorts of racy failures in the MNTR suite, Akka.Cluster, Akka.Remote, and more.

When the DotNetty `IByteBuffer` is passed to the Akka.NET `LoggingBus` for debugging, the `IByteBuffer` is reference counted - and the reference count drops to zero when the message is passed to the `EndpointReader` actor, which causes the memory to be garbage collected or returned to the pool. If the `LoggingBus` processes the message after this occurs, it can blow up the logger and possibly do some damage elsewhere.

Aaronontheweb: Copying the contents via a `.ToString()` method will do the trick - this logging isn't enabled by default and is definitely not recommended in production.

Aaronontheweb:
@akkadotnet/core merged this in to stop the racy DotNetty specs from barfing as a result of the logging handler triggering reference count exceptions in DotNetty

PR #4764, opened 2021-02-07
Add Serialization.WithTransport overload that takes TState

to11mtm:
This PR adds the following method to Akka.Serialization.Serialization:

```
public static T WithTransport<TState, T>(ExtendedActorSystem system, TState state, Func<TState, T> action)
```

The purpose of this overload is to allow a TState value to be passed into the `action` when executed, allowing consumers to avoid allocations due to delegate captures if present.

It looks like remoting already does something similar to this internally since it has access to the right internals, but there are other aspects of Akka (specifically persistence) that could benefit from this overload.

This PR also contains a benchmark of both versions of `WithTransport`. It looks like we are a little bit faster:

|                              Method |     Mean |     Error |    StdDev |
|------------------------------------ |---------:|----------:|----------:|
| Serialization_WithTransport_NoState | 2.991 us | 0.0201 us | 0.0178 us |
|   Serialization_WithTransport_State | 2.864 us | 0.0482 us | 0.0427 us |

I'm not sure why it comes up slower with `[MemoryDiagnoser]`, but important thing is we save ~90 bytes here:

|                              Method |     Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------------------------------ |---------:|----------:|----------:|-------:|-------:|------:|----------:|
| Serialization_WithTransport_NoState | 2.849 us | 0.0366 us | 0.0324 us | 0.8087 | 0.0038 |     - |   3.72 KB |
|   Serialization_WithTransport_State | 2.893 us | 0.0518 us | 0.0485 us | 0.7858 | 0.0038 |     - |   3.63 KB |

That may not look like much but since this is called deep in pipelines I'm expecting at least some gains in Persistence benches. Working on looking at that in Sql.Common

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
@@ -778,27 +778,30 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// <returns>TBD</returns>
         protected virtual void WriteEvent(DbCommand command, IPersistentRepresentation e, IImmutableSet<string> tags)
         {
-            var payloadType = e.Payload.GetType();
-            var serializer = Serialization.FindSerializerForType(payloadType, Configuration.DefaultSerializer);
+            
+            var serializer = Serialization.FindSerializerForType(e.Payload.GetType(), Configuration.DefaultSerializer);
 
             // TODO: hack. Replace when https://github.com/akkadotnet/akka.net/issues/3811
-            string manifest = "";
-            var binary = Akka.Serialization.Serialization.WithTransport(Serialization.System, () =>
+            var (binary,manifest) = Akka.Serialization.Serialization.WithTransport(Serialization.System,(e.Payload,serializer) ,(state) =>
             {
                 
-                if (serializer is SerializerWithStringManifest stringManifest)
+                var (thePayload, theSerializer) = state;
+                string thisManifest = "";
+                if (theSerializer is SerializerWithStringManifest stringManifest)
                 {
-                    manifest = stringManifest.Manifest(e.Payload);
+                    thisManifest = stringManifest.Manifest(thePayload);
                 }
                 else
                 {
-                    if (serializer.IncludeManifest)
+                    if (theSerializer.IncludeManifest)
                     {
-                        manifest = e.Payload.GetType().TypeQualifiedName();
+                        thisManifest = thePayload.GetType().TypeQualifiedName();
                     }
                 }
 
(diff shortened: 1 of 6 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -778,27 +778,30 @@ protected DbCommand GetCommand(DbConnection connection, string sql)
         /// <returns>TBD</returns>
         protected virtual void WriteEvent(DbCommand command, IPersistentRepresentation e, IImmutableSet<string> tags)
         {
-            var payloadType = e.Payload.GetType();
-            var serializer = Serialization.FindSerializerForType(payloadType, Configuration.DefaultSerializer);
+            
+            var serializer = Serialization.FindSerializerForType(e.Payload.GetType(), Configuration.DefaultSerializer);
```
to11mtm, line 782: Idea here is to avoid a local we don't use anywhere else.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -778,27 +778,30 @@ protected DbCommand GetCommand(DbConnection connection, string sql)
...
 
             // TODO: hack. Replace when https://github.com/akkadotnet/akka.net/issues/3811
-            string manifest = "";
-            var binary = Akka.Serialization.Serialization.WithTransport(Serialization.System, () =>
+            var (binary,manifest) = Akka.Serialization.Serialization.WithTransport(Serialization.System,(e.Payload,serializer) ,(state) =>
```
to11mtm, line 785: by deconstructing here, we avoid requiring a capture on `manifest`.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -778,27 +778,30 @@ protected DbCommand GetCommand(DbConnection connection, string sql)
...
             {
                 
-                if (serializer is SerializerWithStringManifest stringManifest)
+                var (thePayload, theSerializer) = state;
+                string thisManifest = "";
```
to11mtm, line 789: We're keeping a local here because changing this into a Ternary at end was both:

 - Harder to read
 - Slower because the complex ternary doesn't optimize well =/

to11mtm:
It looks like in Sql.Common `Persist` and others appear to be both more consistent and a liiiitle bit faster.

Aaronontheweb (approved):
LGTM - great idea

PR #4768, opened 2021-02-08
Fix TestJournal exception aggregation

### LINKED ISSUE #4762: TestJournal PersistAll never invokes callback handler if 2 or more events are persisted.
Akka.Net version: 1.4.16
```
        <PackageReference Include="Akka.Persistence.TestKit.Xunit2" Version="1.4.16" />
        <PackageReference Include="Akka.TestKit" Version="1.4.16" />
        <PackageReference Include="Akka.TestKit.Xunit2" Version="1.4.16" />
```

Platform: macOS Catalina, TargetFramework=netcoreapp3.1

Example to reproduce:

```
public class TestActor : ReceivePersistentActor
{
    public void TestActor()
    {
        Command<TestCommand>(cmd =>
        {
            var event1 = ...
            var event2 = ...
            var events = new List<TestEvent>{ event1, event2 };
            PersistAll(events, _ =>
            {
                Sender.Tell(Done.Instance);
            });
        });
    }
}

public class Tests : PersistenceTestKit
{
    [Fact]
    public async Task ExampleTest()
    {
        await WithJournalWrite(write => write.Pass(), () =>
        {
            var actor = ActorOf(() => new TestActor());
            Watch(actor);

            var command = new TestCommand("1");
            actor.Tell(command, TestActor);
            ExpectMsg<Done>();
        });
    }
}
```

The WriteMessagesAsync method of TestJournal will emit 1 null exception entry for each message included in the atomic write from this line of code here:

https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/core/Akka.Persistence.TestKit/Journal/TestJournal.cs#L56

This will cause t.Result.Count to be 2 (List<Exception>{null, null}) which does not match the atomicWriteCount of 1 and so errors and does not send the WriteMessageSuccessful message back to the actor resulting in the callback for PeristAll to never get fired.

https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs#L393

Arkatufus:
Closes #4762

### DIFF
--- a/src/core/Akka.Persistence.TestKit.Tests/Bug4762FixSpec.cs
+++ b/src/core/Akka.Persistence.TestKit.Tests/Bug4762FixSpec.cs
@@ -1,0 +1,79 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Event;
+using Xunit;
+
+namespace Akka.Persistence.TestKit.Tests
+{
+    /// <summary>
+    /// Fix spec for bug https://github.com/akkadotnet/akka.net/issues/4762
+    /// </summary>
+    public class Bug4762FixSpec : PersistenceTestKit
+    {
+        class WriteMessage
+        { }
+
+        class TestEvent
+        { }
+
+        class TestActor2 : UntypedPersistentActor
+        {
+            private readonly IActorRef _probe;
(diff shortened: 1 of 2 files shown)


Aaronontheweb (approved):
LGTM

PR #4769, opened 2021-02-09
Allow cluster singleton zero buffer size to disable buffering

### LINKED ISSUE #4763: on singleton-proxy setting buffer-size to 0 via Hocon fail on exception
on Akka.Cluster.Tools 1.4.8.0
Windows OS
set config with 

akka.cluster{
  singleton{
            singleton-name= "CoordinatorActor",
            role="Coordinator"
            hand-over-retry-interval = @@HandOverRetryInterval@@
            min-number-of-hand-over-retries = @@MinNumberOfHandOverRetries@@
          }

          singleton-proxy {
              # The actor name of the singleton actor that is started by the ClusterSingletonManager
              singleton-name = "CoordinatorActor",

              # The role of the cluster nodes where the singleton can be deployed. 
              # If the role is not specified then any node will do.
              role = "Coordinator"

              # Interval at which the proxy will try to resolve the singleton instance.
              singleton-identification-interval = 1s

              # If the location of the singleton is unknown the proxy will buffer this
              # number of messages and deliver them when the singleton is identified. 
              # When the buffer is full old messages will be dropped when new messages are
              # sent via the proxy.
              # Use 0 to disable buffering, i.e. messages will be dropped immediately if
              # the location of the singleton is unknown.
              # Maximum allowed buffer size is 10000.
              buffer-size = 1 	#0 is not working - AKKA bug
            }
}

IgorFedchenko:
Closes #4763

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxySettings.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxySettings.cs
@@ -99,8 +99,8 @@ namespace Akka.Cluster.Tools.Singleton
                 throw new ArgumentNullException(nameof(singletonName));
             if (singletonIdentificationInterval == TimeSpan.Zero)
                 throw new ArgumentException("singletonIdentificationInterval must be positive", nameof(singletonIdentificationInterval));
-            if (bufferSize <= 0)
-                throw new ArgumentException("bufferSize must be positive", nameof(bufferSize));
+            if (bufferSize < 0)
+                throw new ArgumentException("bufferSize must not be negative", nameof(bufferSize));
 
             SingletonName = singletonName;
             Role = role;
--- a/src/core/Akka.Docs.Tests/Actors/ReceiveTimeoutSpecs.cs
+++ b/src/core/Akka.Docs.Tests/Actors/ReceiveTimeoutSpecs.cs
@@ -30,25 +30,22 @@ namespace DocsExamples.Actors
         public class CheckTimeout : INotInfluenceReceiveTimeout { }
         public class ReceiveTimeoutActor : ReceiveActor
         {
-            private bool _timedOut = false;
+            private readonly TimeSpan _inactivityTimeout;
 
-            public ReceiveTimeoutActor()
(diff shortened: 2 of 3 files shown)


src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxySettings.cs
```diff
@@ -99,8 +99,8 @@ public ClusterSingletonProxySettings(string singletonName, string role, TimeSpan
                 throw new ArgumentNullException(nameof(singletonName));
             if (singletonIdentificationInterval == TimeSpan.Zero)
                 throw new ArgumentException("singletonIdentificationInterval must be positive", nameof(singletonIdentificationInterval));
-            if (bufferSize <= 0)
-                throw new ArgumentException("bufferSize must be positive", nameof(bufferSize));
+            if (bufferSize < 0)
+                throw new ArgumentException("bufferSize must not be negative", nameof(bufferSize));
```
IgorFedchenko, line 103: The `0` value is actually supported in the code: https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs#L252

So this is just a minor validation issue.

src/core/Akka.Docs.Tests/Actors/ReceiveTimeoutSpecs.cs
```diff
@@ -30,40 +30,38 @@ public class ReceiveTimeoutSpecs : TestKit
         public class CheckTimeout : INotInfluenceReceiveTimeout { }
         public class ReceiveTimeoutActor : ReceiveActor
         {
-            private bool _timedOut = false;
+            private readonly TimeSpan _inactivityTimeout;
```
IgorFedchenko, line 33: Nevermind, I will fix this racy spec in separate PR: https://github.com/akkadotnet/akka.net/pull/4775

IgorFedchenko:
PR is sometimes passing, sometimes fails on some racy queries - but it is ready for review and merge I think.

PR #4774, opened 2021-02-09
Allow persistence providers to customize SQL query parameters generation

IgorFedchenko:
Work for https://github.com/akkadotnet/Akka.Persistence.SqlServer/issues/159

In this PR, we are adding support for loading column sizes - optional to implement in Sql journals providers. This column sizes are used to set static parameter sizes and reduce possible overhead.

Some results for SqlServer provider are shown here (we will merge that PR once this update is released in `Akka.Persistence` package): https://github.com/akkadotnet/Akka.Persistence.SqlServer/pull/197

UPDATE: Made update more generic and simple in this base repository. Just need to allow plugins to customize SQL query parameters generation, and that's it.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
@@ -892,7 +892,7 @@ namespace Akka.Persistence.Sql.Common.Journal
                 await connection.OpenAsync();
                 
                 // pre-load column sizes to use in queries
-                LoadColumnSizes(connection, Setup.NamingConventions);
+                await LoadColumnSizes(connection);
 
                 using (var tx = connection.BeginTransaction(Setup.IsolationLevel))
                 using (var command = (TCommand)connection.CreateCommand())
@@ -1414,14 +1414,13 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// </para>
         /// 
         /// </summary>
-        /// <param name="connection"></param>
-        /// <param name="conventions"></param>
-        protected virtual ColumnSizesInfo LoadColumnSizesInternal(DbConnection connection, QueryConfiguration conventions)
+        /// <param name="connection">Already opened connection to the database</param>
+        protected virtual Task<ColumnSizesInfo> LoadColumnSizesInternal(DbConnection connection)
         {
-            return ColumnSizesInfo.Default;
+            return Task.FromResult(ColumnSizesInfo.Default);
         }
 
-        private void LoadColumnSizes(DbConnection connection, QueryConfiguration conventions)
+        private async Task LoadColumnSizes(DbConnection connection)
         {
             // load schema/initialize column sizes only once
             if (ColumnSizes != null)
@@ -1429,7 +1428,7 @@ namespace Akka.Persistence.Sql.Common.Journal
 
             try
             {
-                ColumnSizes = LoadColumnSizesInternal(connection, conventions);
+                ColumnSizes = await LoadColumnSizesInternal(connection);
             }
             catch (Exception exception)
             {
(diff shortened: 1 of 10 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -994,7 +994,7 @@ protected virtual async Task<long> ReadHighestSequenceNr(string persistenceId, T
             command.CommandText = HighestSequenceNrSql;
 
             command.Parameters.Clear();
-            AddParameter(command, "PersistenceId", DbType.String, persistenceId);
+            AddParameter(command, "@PersistenceId", DbType.String, persistenceId);
```
IgorFedchenko, line 997: See https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlparameter.parametername?view=dotnet-plat-ext-5.0#remarks

IgorFedchenko:
Now need to test if this will produce queries we need

IgorFedchenko:
Nope, `GetSchema` returns wrong thing. Will check for other provider-agnosting way of doing this (have some ideas).

IgorFedchenko:
Ok, now I see that this feature has to be provider-specific. Will finish this PR first, and then optionally implement this in plugins.

IgorFedchenko:
All right, added a virtual method that allows different plugins to add this "parameter size" optimization.

Will check if this really works on SqlServer plugin as a test, and if it works - will mark as "ready for review".

IgorFedchenko:
Approvals are failing - will fix this as well if queries look like what we need.

IgorFedchenko:
Ok, this is ready for review now.

Some results for SqlServer provider are shown here (we will merge that PR once this update is released in `Akka.Persistence` package): https://github.com/akkadotnet/Akka.Persistence.SqlServer/pull/197

IgorFedchenko:
API approvals were failing - fixed that

IgorFedchenko:
I realized that there are multiple journals, and all of them should have this supported... Updating my PR for that

IgorFedchenko:
All right, this PR was simplified pretty much, moved all logic to SQLServer plugin since that's going to be provider-specific feature.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1375,9 +1375,19 @@ protected void AddParameter(TCommand command, string paramName, DbType dbType, o
...
+            
             command.Parameters.Add(param);
         }
 
+        /// <summary>
+        /// Override this to customize <see cref="DbParameter"/> creation used for building database queries
+        /// </summary>
+        /// <param name="command"><see cref="DbCommand"/> used to define a parameter in.</param>
+        /// <param name="param">Parameter to customize</param>
+        protected virtual void CustomDbParameterSetup(TCommand command, DbParameter param) { }
```
to11mtm, line 1389: Would it make more sense to call it something like 'OnBeforeParameterAddToCommand`? (ok maybe that one is a bit much). 

Thinking about how we have other things named around 'hookpoints' (i.e. `AroundPreRestart`) here

to11mtm (approved):

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
IgorFedchenko: Yeah, this makes sense... Let me update it to `PreAddParameterToCommand`?

IgorFedchenko: Does it seem better now @to11mtm ?

PR #4779, opened 2021-02-11
Updated specs that might be used in documentation to not use internal properties

### LINKED ISSUE #4772: Remove internal properties like ReadView from public code samples
We have some public-facing specs, that are used by Akka.NET users - like https://getakka.net/articles/testing/multi-node-testing.html .

But in this spec `ReadView` internal property is used (maybe other internal fields as well), which cause people to not understand how can they use TestKit for writing specs in their applications.

Need to refactor this specs to use only public-facing API.

IgorFedchenko:
Close #4772

Actually, initially I was going to remove `InternalsVisibleTo` attribute from `Akka.Cluster` assembly for test assemblies, so that we were using only public API in our own tests - but turned out that sometimes we really need this internals to be visible for assertions, and replacing them with public-only properties would be complicated (and would lead to code duplication). Users still do not need those internals.

So I have just did what I can - basically, replaced only those internals which were referenced in documentation. The specs where other internal properties were used are unlikely to be referenced in documentation since they are assertion our "internal stuff".

### DIFF
--- a/src/core/Akka.Cluster.Tests.MultiNode/RestartFirstSeedNodeSpec.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/RestartFirstSeedNodeSpec.cs
@@ -108,14 +108,14 @@ namespace Akka.Cluster.Tests.MultiNode
                     AwaitAssert(() =>
                     {
                         Cluster.Get(seed1System.Value)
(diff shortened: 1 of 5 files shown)


IgorFedchenko:
Set this to be merged automatically, since at the end changes are very minor.

PR #4780, opened 2021-02-11
Fix EventFilter output issues with NUnit

### LINKED ISSUE #4733: How to use Akka.DependencyInjection with Akka.TestKit?
- Akka.Net 1.4.15
- .Net 5

Excuse, me.
I can't found sample code. How to use TestKit with BootstrapSetup?
- https://github.com/seungyongshim/Mirero.Akka.Extensions.DependencyInjection

Best regards

IgorFedchenko:
Close #4733

### DIFF
--- a/src/Akka.sln
+++ b/src/Akka.sln
@@ -242,6 +242,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "AspNetCore", "AspNetCore", 
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Samples.Akka.AspNetCore", "examples\AspNetCore\Samples.Akka.AspNetCore\Samples.Akka.AspNetCore.csproj", "{D62F4AD6-318F-4ECC-B875-83FA9933A81B}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Akka.TestKit.NUnit.Tests", "core\Akka.TestKit.NUnit.Tests\Akka.TestKit.NUnit.Tests.csproj", "{87DDD75B-2F1D-470E-9BDE-1707194AEE57}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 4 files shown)


IgorFedchenko:
Added a reproduction project with NUnit support - not going to merge this, but need for testing.

Funny thing is that this code logs only `error` level, but suppress info:
```
[SetUp]
public void Setup()
{
	Console.SetOut(TestContext.Out);
}

[Test]
public void Test1()
{
	EventFilter.Info().Expect(3, () =>
	{
		Log.Error("error1");
		Log.Error("error2");
		Log.Info("info");
	});
}
```

IgorFedchenko:
All right, so the reason is following:

In `EventFilterApplier` we are muting events while making assertion. Not sure why:
https://github.com/akkadotnet/akka.net/blob/076d8de5b260a75c82b2a0151e56fce410d24cd4/src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs#L282

Internally, this ends up with `TestEventListener` not printing events to the `Console`:
https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/core/Akka.TestKit/EventFilter/TestEventListener.cs#L63

Now, for XUnit there is a separate logger that subscribes to `LogEvent` events from `System.EventStream` and prints them to the `ITestOutputHelper`:
https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/contrib/testkits/Akka.TestKit.Xunit/TestKit.cs#L126

For `NUnit` there is nothing like that - and we have to use `Console.SetOut(TestContext.Out);` to redirect console output to NUnit output sink. But we could do the same thing that XUnit does and have dedicated logger which will use `TestContext.WriteLine` for writing events to the output without interaction with console.

Based on that, here is a workaround everyone can use:
```csharp
public class LoggingActor : ReceiveActor
{
	public LoggingActor(TextWriter writer) => Receive<LogEvent>(e => writer.WriteLine(e.ToString()));
}

// this code should be inside [SetUp], no Console output redirect needed anymore
var loggingActor = Sys.ActorOf(Props.Create(() => new LoggingActor(TestContext.Out)));
Sys.EventStream.Subscribe<LogEvent>(loggingActor);
```

This way, all logging events are getting written to the test adapter output without interaction with console.

PR #4788, opened 2021-02-17
Fixed actor cancellation message namespace for SourceRef

### LINKED ISSUE #4744: InputStreamPublisher cannot be cancelled and StreamRef will be blocked
Akka.Net version: 1.4.14
Akka.Streams version: 1.4.14
platform: win 10, .net core 3.1
gist: https://gist.github.com/ClockGet/a3c114c053caba681eddbec4b3b3c75b

Hello everyone,
Thank you for this great project and I am trying to use this amazing project recently. I found that when an [ArgumentException](https://gist.github.com/ClockGet/a3c114c053caba681eddbec4b3b3c75b#file-teststream-cs-L162) threw by a task in ReadStreamActor, the [task](https://gist.github.com/ClockGet/a3c114c053caba681eddbec4b3b3c75b#file-teststream-cs-L64) in SendStreamActor will be blocked.
What's more, when I referenced the source code of Akka.Stream and found out after debugging, when the ArgumentException is thrown, the InputStreamPublisher can receive an [Akka.Streams.Actors.Cancel](https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/core/Akka.Streams/Implementation/IO/InputStreamPublisher.cs#L79) message but not handle. Then I replace Akka.Streams.Implementation.Cancel with Akka.Streams.Actors.Cancel (`With<Cancel>()` -> `With<Actors.Cancel>()`), the task in SendStreamActor can be completed normally. I am not sure if this is a bug or there is a problem with my usage, I look forward to your reply.
Thank you very much.

IgorFedchenko:
Close #4744

Making our code to match scala implementation:
```scala
def receive = {
    case ActorPublisherMessage.Request(elements) ⇒ readAndSignal()
    case Continue                                ⇒ readAndSignal()
    case ActorPublisherMessage.Cancel            ⇒ context.stop(self)
  }
```

### DIFF
--- a/src/core/Akka.Streams/Implementation/IO/FilePublisher.cs
+++ b/src/core/Akka.Streams/Implementation/IO/FilePublisher.cs
@@ -130,7 +130,7 @@ namespace Akka.Streams.Implementation.IO
             => message.Match()
                 .With<Request>(() => ReadAndSignal(_maxBuffer))
                 .With<Continue>(() => ReadAndSignal(_maxBuffer))
-                .With<Cancel>(() => Context.Stop(Self))
+                .With<Actors.Cancel>(() => Context.Stop(Self))
                 .WasHandled;
 
         private void ReadAndSignal(int maxReadAhead)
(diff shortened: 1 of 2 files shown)


Aaronontheweb (approved):
LGTM

PR #4789, opened 2021-02-17
Fix finding of temprorary actor for Ask pattern when having multiple ActorSystems

### LINKED ISSUE #4384: Ask can't find temporary actor inside async context
Version: 1.4.4

The following code produces an `AskTimeoutException`:

```csharp
public async Task ConsistentHashingPoolRoutersShouldWorkAsExpectedWithHashMapping()
        {
            var poolRouter =
                Sys.ActorOf(Props.Create(() => new ReporterActor(TestActor)).WithRouter(new ConsistentHashingPool(5,
                        msg =>
                        {
                            if (msg is IConsistentHashable c)
                                return c.ConsistentHashKey;
                            return msg;
                        })),
                    "router1");

            // use some auto-received messages to ensure that those still work
            var numRoutees = (await poolRouter.Ask<Routees>(new GetRoutees(), TimeSpan.FromSeconds(2))).Members.Count();

            // establish association between ActorSystems
            var sys2Probe = CreateTestProbe(Sys2);
            var secondActor = Sys.ActorOf(act => act.ReceiveAny((o, ctx) => ctx.Sender.Tell(o)), "foo");

            Sys2.ActorSelection(new RootActorPath(Sys1Address) / "user" / secondActor.Path.Name).Tell("foo", sys2Probe);
            sys2Probe.ExpectMsg("foo");

            // have ActorSystem2 message it via tell
            var sel = Sys2.ActorSelection(new RootActorPath(Sys1Address) / "user" / "router1");
            sel.Tell(new HashableString("foo"));
            ExpectMsg<HashableString>(str => str.Str.Equals("foo"));

            // have ActorSystem2 message it via Ask
            sel.Ask(new Identify("bar2"), TimeSpan.FromSeconds(3)).PipeTo(sys2Probe);
            var remoteRouter = sys2Probe.ExpectMsg<ActorIdentity>(x => x.MessageId.Equals("bar2"), TimeSpan.FromSeconds(5)).Subject;

            var s2Actor = Sys2.ActorOf(act =>
            {
                act.ReceiveAny((o, ctx) =>
                    sel.Ask<ActorIdentity>(new Identify(o), TimeSpan.FromSeconds(3)).PipeTo(sys2Probe));
            });
            s2Actor.Tell("hit");
            sys2Probe.ExpectMsg<ActorIdentity>(x => x.MessageId.Equals("hit"), TimeSpan.FromSeconds(5));
        }
```

Specifically, this line:

```csharp
// have ActorSystem2 message it via Ask
sel.Ask(new Identify("bar2"), TimeSpan.FromSeconds(3)).PipeTo(sys2Probe);
var remoteRouter = sys2Probe.ExpectMsg<ActorIdentity>(x => x.MessageId.Equals("bar2"), TimeSpan.FromSeconds(5)).Subject;
```

From the logs:

```
DEBUG][4/22/2020 9:11:36 PM][Thread 0042][LocalActorRefProvider(akka://test2)] Resolve of path sequence [/temp/f] failed
[INFO][4/22/2020 9:11:36 PM][Thread 0009][akka://test2/temp/f] Message [ActorIdentity] from akka.tcp://test@localhost:55844/user/router1 to akka://test2/temp/f was not delivered. [1] dead letters encountered .This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
```

The issue might be with running two `ActorSystem` instances in the same context, but this is 100% reproducible.

IgorFedchenko:
Close #4384

### DIFF
--- a/src/core/Akka/Actor/Futures.cs
+++ b/src/core/Akka/Actor/Futures.cs
@@ -129,20 +129,20 @@ namespace Akka.Actor
         }
 
         /// <summary>
-        /// TBD
+        /// Resolves <see cref="IActorRefProvider"/> for Ask pattern
         /// </summary>
-        /// <param name="self">TBD</param>
-        /// <returns>TBD</returns>
+        /// <param name="self">Reference to someone we are sending Ask request to</param>
+        /// <returns>Provider used for Ask pattern implementation</returns>
         internal static IActorRefProvider ResolveProvider(ICanTell self)
         {
-            if (ActorCell.Current != null)
-                return InternalCurrentActorCellKeeper.Current.SystemImpl.Provider;
+            if (self is ActorSelection)
+                return ResolveProvider(self.AsInstanceOf<ActorSelection>().Anchor);
 
             if (self is IInternalActorRef)
                 return self.AsInstanceOf<IInternalActorRef>().Provider;
-
-            if (self is ActorSelection)
-                return ResolveProvider(self.AsInstanceOf<ActorSelection>().Anchor);
+            
+            if (ActorCell.Current != null)
+                return InternalCurrentActorCellKeeper.Current.SystemImpl.Provider;
 
             return null;
         }
--- a/src/core/Akka.Remote.Tests/BugFixes/BugFix4384Spec.cs
+++ b/src/core/Akka.Remote.Tests/BugFixes/BugFix4384Spec.cs
@@ -14,6 +14,7 @@ using Akka.Actor.Dsl;
 using Akka.Configuration;
 using Akka.Routing;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Akka.Tests.Actor
 {
@@ -24,7 +25,7 @@ namespace Akka.Tests.Actor
         public ActorSystem Sys2 { get; }
         public Address Sys2Address { get; }
         
-        public BugFix4384Spec() : base()
+        public BugFix4384Spec(ITestOutputHelper outputHelper) : base(nameof(BugFix4384Spec), outputHelper)
         {
             var sys1Port = GetFreeTcpPort();
             var sys2Port = GetFreeTcpPort();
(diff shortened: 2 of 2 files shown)


src/core/Akka/Actor/Futures.cs
```diff
@@ -129,20 +129,20 @@ public static async Task<T> Ask<T>(this ICanTell self, Func<IActorRef,object> me
...
+                return ResolveProvider(self.AsInstanceOf<ActorSelection>().Anchor);
 
             if (self is IInternalActorRef)
                 return self.AsInstanceOf<IInternalActorRef>().Provider;
-
-            if (self is ActorSelection)
-                return ResolveProvider(self.AsInstanceOf<ActorSelection>().Anchor);
+            
+            if (ActorCell.Current != null)
+                return InternalCurrentActorCellKeeper.Current.SystemImpl.Provider;
```
IgorFedchenko, line 145: So the reason of the issue was the following:

Usually, using `ActorCell.Current` works when since it is some actor from Sys2 requesting `Ask` to actor of Sys1. But here, `ActorCell.Current` is `akka://test/temp/XXX`, the test actor of test system. This test actor's path is not resolvable for remote actors of Sys1 or Sys2 (test `Sys` is not configured to use remoting and does not have association with any of this systems).

Checks ordering change here allows to use correct provider (RemoteRefProvider in this case) with `Sys2` system for hosting temporary actor.

Aaronontheweb: WOW. Nice catch. I think this error might also have a lot to do with some of the "unable to resolve X" errors that occur in remote Ask operations too. I'm going to do some more testing around this locally, but this is an amazing find.

Aaronontheweb (approved):
LGTM

src/core/Akka/Actor/Futures.cs
IgorFedchenko: Yeah, that could lead to such "unable to resolve" errors I think.

IgorFedchenko:
Noticed that PR name is set to the first commit message, which is not representative. Updated it.

PR #4804, opened 2021-03-02
clarified Akka.Remote.EndpointWriter ArgumentException error message

Aaronontheweb:
cleaning up the error message logged by the `EndpointWriter` that occurs when the underlying serializer throws an `ArgumentException`.


Aaronontheweb:
This error message made it sound like the definition for the message type could not be found, but truthfully it is more generic than that.

Arkatufus (approved):
LGTM

PR #4807, opened 2021-03-02
Fix broken SerializationSetup, bindings insertion was reversed between HOCON and SerializationSetup.

### LINKED ISSUE #4761: Overriding Default Serializer for typeof(object) programmatically
I'm trying to override the default `NewtonsoftJsonSerializer` with my own implementation of `Akka.Serialization.Serializer` and I want this to be done programmatically as it's in a base library and I really don't want all of my apps to need this extra bit of a config. I'm initializing my `ActorSystem` by creating a `SerializationSetup` that creates my custom serializer and instructs it via `SerializerDetails` to register under `typeof(object)` -- expected behavior is that this would *overwrite* the default `NewtonsoftJsonSerializer` but alas, it doesn't.

If running the attached sample in debug mode, you can break after the `ActorSystem` is created and inspect `system.Serialization._serializerMap` and observe the `typeof(object)` mapping still pointing to `NewtonsoftJsonSerializer`. There is a mapping for my custom `World` type but my serializer isn't even being invoked for that type. If I manually invoke `system.Serializer.AddSerializerMap(typeof(object), new CustomSerializer(...))` then it does start working.

> 1.4.16

> dotnetcore v3.1 running on win10

- A list of steps to reproduce the issue. Or an gist or github repo which can be easily used to reproduce your case.
Please see this repository for a working example and comments/discussion on the observed behavior:
> https://github.com/mchandschuh/default-serializer-programmatic-override

Arkatufus:
Closes #4761

### DIFF
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -186,15 +186,10 @@ namespace Akka.Serialization
             }
 
             // Add any serializers that are registered via the SerializationSetup
+            // This has to be done here because SerializationSetup ALWAYS win.
             foreach (var details in _serializerDetails)
             {
                 AddSerializer(details.Alias, details.Serializer);
-
-                // populate the serialization map
-                foreach (var t in details.UseFor)
-                {
-                    AddSerializationMap(t, details.Serializer);
-                }
             }
 
             foreach (var kvp in serializerBindingConfig)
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
And this is done because the fallback serialization bindings in HOCON will always override the user-defined ones in the `SerializationSetup`?

Arkatufus:
The way it is set up before, both `SerializationSetup` alias mapping and type binding mapping were done just before we process the HOCON bindings, hence why the HOCON bindings always win (the `System.Object` binding in the original failed case).

I just split these mappings to after each of the respective HOCON mapping so that anything that came from `SerializationSetup` overrides HOCON.

Aaronontheweb (approved):
LGTM

mchandschuh:
Awesome work guys, thanks for proving I wasn't just crazy 😛

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -210,17 +205,24 @@ public Serialization(ExtendedActorSystem system)
...
                     continue;
                 }
-
                 AddSerializationMap(messageType, serializer);
             }
 
-            
+            // Add any serializer bindings that are registered via the SerializationSetup
+            // This has to be done here because SerializationSetup ALWAYS win.
+            foreach (var details in _serializerDetails)
```
mchandschuh, line 218: yep -- I suspected it was an ordering issue -- applying the defaults after apply the user defined settings. easy peasy!

mchandschuh:
@Aaronontheweb -- do we know what version this fix will be released in? I'll be able to remove my work around once the new packages are live. Thank you!

Aaronontheweb:
We just released Akka.NET v1.4.17 yesterday - this should be included in that release. @mchandschuh

mchandschuh:
yep -- just saw the twitter announcement -- will follow for future updates, thanks @Aaronontheweb

Issue #4806, opened 2021-03-02
Strange exception using distributed data


tesgu:
Hi guys, we are using Akka.DistributedData 1.4.16

When load is high, the system starts to throw the following exception and the distributed data stops finding items.
We are using an ORDictionary<string, POC>

Any clue would be much appreciated.

System.InvalidCastException: Unable to cast object of type 'NoDelta' to type 'DeltaGroup[System.String,Payvision.ThreeDS.Processor.DistributedCache.Domain.TransactionState]'.
   at Akka.DistributedData.ORDictionary`2.DeltaGroup.Merge(IReplicatedData other)
   at Akka.DistributedData.DeltaPropagationSelector.CollectPropagations()
   at Akka.DistributedData.Replicator.ReceiveDeltaPropagationTick()
   at lambda_method(Closure , Object , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Action`1 , Object[] )
   at Akka.Tools.MatchHandler.PartialHandlerArgumentsCapture`16.Handle(T value)
   at Akka.Actor.ReceiveActor.ExecutePartialMessageHandler(Object message, PartialAction`1 partialAction)
   at Akka.Actor.UntypedActor.Receive(Object message)
   at Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message)
   at Akka.Actor.ActorCell.ReceiveMessage(Object message)
   at Akka.Actor.ActorCell.Invoke(Envelope envelope)

Aaronontheweb:
@Arkatufus could you take a look at this?

tesgu:
If it helps (we are researching by our own, because it's affecting our production environment) We insert the item with consistency WriteLocal. After that we retrieve the item successfully also with ReadLocal. But after a minute the transaction doesn't exists in the local ORDictionary nor in the remote ORDictionaries.... The item has vanished...
There are 3 machines involved an none of the 3 has the transaction... I said that this was happening under heavy loads, but it's also happening under what I consider it's not a heavy load (24 trx/sec)

Arkatufus:
@tesgu I'm looking into the problem right now

Arkatufus:
[...]

PR #4809, opened 2021-03-04
Re-port DeltaPropagationSelector.CollectPropagations(), trying to avoid null leak

Arkatufus:
Related to #4806

### DIFF
--- a/src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
+++ b/src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
@@ -10,8 +10,10 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using Akka.Actor;
+using Akka.Cluster;
 using Akka.DistributedData.Internal;
 using Akka.Event;
+using Akka.Util.Internal;
 
 namespace Akka.DistributedData
 {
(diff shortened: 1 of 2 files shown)


src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
```diff
@@ -55,89 +57,83 @@ public void Delete(string key)
...
+                        {
+                            deltaGroup = deltaEntriesAfterJ.Values.Aggregate((d1, d2) =>
                             {
-                                using (var e = deltaEntriesAfterJ.Values.GetEnumerator())
-                                {
-                                    e.MoveNext();
-                                    deltaGroup = e.Current;
-                                    while (e.MoveNext())
-                                    {
-                                        deltaGroup = deltaGroup.Merge(e.Current);
```
Arkatufus, line 114: The bug happened here, a `NoDelta` null delta marker got passed into this line and `ORDictionary` could not cast it to `DeltaGroup`

src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
```diff
@@ -55,89 +57,83 @@ public void Delete(string key)
...
-                            // in most cases the delta group merging will be the same for each node,
-                            // so we cache the merged results
-                            var cacheKey = (key, fromSeqNr, toSeqNr);
-                            if (!cache.TryGetValue(cacheKey, out var deltaGroup))
+                        // in most cases the delta group merging will be the same for each node,
+                        // so we cache the merged results
+                        var cacheKey = (key, fromSeqNr, toSeqNr);
+                        if (!cache.TryGetValue(cacheKey, out var deltaGroup))
+                        {
+                            deltaGroup = deltaEntriesAfterJ.Values.Aggregate((d1, d2) =>
```
Arkatufus, line 107: Replaced the suspicious looking code block with one that makes a lot more sense

src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
```diff
@@ -55,89 +57,83 @@ public void Delete(string key)
...
-                                }
-
-                                cache[cacheKey] = deltaGroup;
-                            }
-
-                            deltas[key] = (deltaGroup, fromSeqNr, toSeqNr);
-                            _deltaSentToNode = _deltaSentToNode.SetItem(key, deltaSentToNodeForKey.SetItem(node, toSeqNr));
+                                var merged = ReferenceEquals(d2, DeltaPropagation.NoDeltaPlaceholder) 
+                                    ? DeltaPropagation.NoDeltaPlaceholder 
+                                    : d1.Merge(d2);
```
Arkatufus, line 111: This is the same line on the new code that the bug might happen again, but now it is guarded with a reference check

src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
Arkatufus: `reduceLeft` works exactly the same as LINQ `Aggregate`

src/contrib/cluster/Akka.DistributedData/Replicator.cs
```diff
@@ -1422,14 +1435,14 @@ protected override ImmutableArray<Address> AllNodes
...
-                
+
 
             protected override DeltaPropagation CreateDeltaPropagation(ImmutableDictionary<string, (IReplicatedData data, long from, long to)> deltas)
             {
                 // Important to include the pruning state in the deltas. For example if the delta is based
                 // on an entry that has been pruned but that has not yet been performed on the target node.
                 var newDeltas = deltas
-                    .Where(x => !Equals(x.Value.data, DeltaPropagation.NoDeltaPlaceholder))
+                    .Where(x => !ReferenceEquals(x.Value.data, DeltaPropagation.NoDeltaPlaceholder))
```
Aaronontheweb, line 1445: Why the `ReferenceEquals` change here?

Arkatufus: because it makes a lot more sense, `NoDeltaPlaceholder` is a reference to a singleton instance of `NoDelta`, it makes more sense to compare it by reference

Arkatufus:
Closing this in favor of #4811

PR #4811, opened 2021-03-04
Fix potential NoDelta leak into IReplicatedData.Merge()

Arkatufus:
Related to #4806
Minimize the number of changes to only those that directly relates to the bug.

### DIFF
--- a/src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
+++ b/src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
@@ -12,6 +12,7 @@ using System.Linq;
 using Akka.Actor;
 using Akka.DistributedData.Internal;
 using Akka.Event;
+using Akka.Util.Internal;
 
 namespace Akka.DistributedData
 {
@@ -68,13 +69,11 @@ namespace Akka.DistributedData
                 if (all.Length <= sliceSize) slice = all;
                 else
                 {
-                    var start = (int)(_deltaNodeRoundRobinCounter % all.Length);
-                    var buffer = new Address[sliceSize];
-                    for (var i = 0; i < sliceSize; i++)
-                    {
-                        buffer[i] = all[(start + i) % all.Length];
-                    }
-                    slice = ImmutableArray.CreateRange(buffer);
+                    var i = (int)(_deltaNodeRoundRobinCounter % all.Length);
+                    slice = all.Slice(i, sliceSize).ToImmutableArray();
+                
+                    if (slice.Length != sliceSize)
+                        slice = slice.AddRange(all.Take(sliceSize - slice.Length));
                 }
 
                 _deltaNodeRoundRobinCounter += sliceSize;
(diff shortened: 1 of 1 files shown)


src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
```diff
@@ -105,19 +104,17 @@ public void Delete(string key)
                             var cacheKey = (key, fromSeqNr, toSeqNr);
                             if (!cache.TryGetValue(cacheKey, out var deltaGroup))
                             {
-                                using (var e = deltaEntriesAfterJ.Values.GetEnumerator())
+                                deltaGroup = deltaEntriesAfterJ.Values.Aggregate((d1, d2) =>
```
Arkatufus, line 107: Replaced the suspicious looking code block with one that makes a lot more sense.
This is the scala code for reference:
```scala
val group = deltaEntriesAfterJ.valuesIterator.reduceLeft { (d1, d2) =>
  val merged = d2 match {
    case NoDeltaPlaceholder => NoDeltaPlaceholder
    case _                  =>
      // this is fine also if d1 is a NoDeltaPlaceholder
      d1.merge(d2.asInstanceOf[d1.T])
  }
  merged match {
    case s: ReplicatedDeltaSize if s.deltaSize >= maxDeltaSize =>
      // discard too large deltas
      NoDeltaPlaceholder
    case _ => merged
  }
}
```
`reduceLeft` works exactly the same as LINQ `Aggregate`

src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
```diff
@@ -105,19 +104,17 @@ public void Delete(string key)
...
                                 {
-                                    e.MoveNext();
-                                    deltaGroup = e.Current;
-                                    while (e.MoveNext())
-                                    {
-                                        deltaGroup = deltaGroup.Merge(e.Current);
```
Arkatufus, line 114: The bug happened here, a `NoDelta` null delta marker got passed into this line and `ORDictionary` could not cast it to `DeltaGroup`

src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
```diff
@@ -68,13 +69,11 @@ public void Delete(string key)
                 if (all.Length <= sliceSize) slice = all;
                 else
                 {
-                    var start = (int)(_deltaNodeRoundRobinCounter % all.Length);
```
Arkatufus, line 71: This part of the code also looks suspicious, so I reported it with a cleaner one.
Reference scala code:
```scala
val i = (deltaNodeRoundRobinCounter % all.size).toInt
val first = all.slice(i, i + sliceSize)
if (first.size == sliceSize) first
else first ++ all.take(sliceSize - first.size)
```

src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs
```diff
@@ -68,13 +69,11 @@ public void Delete(string key)
...
-                    var buffer = new Address[sliceSize];
-                    for (var i = 0; i < sliceSize; i++)
-                    {
-                        buffer[i] = all[(start + i) % all.Length];
-                    }
-                    slice = ImmutableArray.CreateRange(buffer);
+                    var i = (int)(_deltaNodeRoundRobinCounter % all.Length);
+                    slice = all.Slice(i, sliceSize).ToImmutableArray();
```
Arkatufus, line 73: This code looks different than the reference scala because in scala, slice parameters are `from` and `to`, while in C#, slice is `from` and `count`

Aaronontheweb: Thanks - I was just about to ask

Aaronontheweb (approved):
LGTM

PR #4814, opened 2021-03-06
Abstraction of ServiceProvider, Improving Akka.DependencyInjection

SamEmber:
Hi Guys,

I started using the new Akka.DependencyInjection library for MS-DI recently and I have run across a few issues that I would like to address.

The initial problem I faced was the tight coupling of the actor system DI to IServiceProvider. My team have a testing library where we substitute child actors for test probes, previously we did this via AddDependencyResolver however this has now been removed. Currently this coupling is making it very hard to achieve the previous functionality.

I propose a change to the ServiceProvider where instead of passing in an IServiceProvider we instead pass in a IDependencyResolver interface, by default this IDependencyResolver will be a SeriveProviderDependencyResolver that uses the IServiceProvider to resolve any DI dependencies. This will allow others to create their own implementations of the IDependencyResolver for testing.

I want to make it clear that I don't want to change the functionality of the current DI service resolution, its still extremely important that users manage the lifetime of their dependencies and this change will not effect that.

Secondly I have added the ability to resolve actor props from a Type param, this should make it easier to create actors dynamically.

I am very open to any feedback. I have not properly commented the code yet as I want to make sure this is a reasonable proposal from the community before I spent more time to complete this.

Thanks,

Sam


Aaronontheweb:
Thanks @SamEmber - we'll take a look at this soon

src/contrib/dependencyinjection/Akka.DependencyInjection/DependencyResolver.cs
```diff
@@ -0,0 +1,97 @@
...
+using Akka.Event;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Akka.DependencyInjection
+{
+    /// <summary>
+    /// Provides users with immediate access to the <see cref="IDependencyResolver"/> bound to
+    /// this <see cref="ActorSystem"/>, if any.
+    /// </summary>
+    public sealed class DependencyResolver : IExtension
```
Aaronontheweb, line 20: We probably shouldn't rename the extension here, for API compatibility reasons

src/contrib/dependencyinjection/Akka.DependencyInjection/IDependencyResolver.cs
```diff
@@ -0,0 +1,21 @@
...
+// //     Copyright (C) 2013-2021 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using System;
+using Akka.Actor;
+
+namespace Akka.DependencyInjection
+{
+    public interface IDependencyResolver
```
Aaronontheweb, line 13: So I'm supportive of adding this abstraction, after discussing it with @to11mtm and reviewing some of Mark Seeman's advice here: https://blog.ploeh.dk/2014/05/19/di-friendly-framework/

We've already run into "conforming container" problems with Microsoft.Extensions.DependencyInjection 3.0 vs. 5.0 - so an abstraction like yours would be helpful for users who need to roll-forward before the library itself can.

src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProviderProps.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Akka.DependencyInjection
+{
+    /// <summary>
+    /// This class represents a specialized <see cref="Akka.Actor.Props"/> that uses delegate invocation
+    /// to create new actor instances, rather than a traditional <see cref="System.Activator"/>.
+    ///
+    /// Relies on having an active <see cref="IServiceProvider"/> implementation available
+    /// </summary>
+    internal class ServiceProviderProps : Props
```
Aaronontheweb, line 20: We're going to remove this concept altogether in https://github.com/akkadotnet/akka.net/issues/4853 and replace it with an IIndirectActorProducer, similar to what we did in the original Akka.DI.Core library. 

We should probably merge that PR in first once it gets submitted, then rebase this on top of it.

src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProviderScope.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Akka.DependencyInjection
+{
+    public interface IResolverScope : IDisposable
+    {
+        IDependencyResolver Resolver { get; }
+    }
+    
+    public class ServiceProviderScope : IResolverScope
```
Aaronontheweb, line 18: Fine with this abstraction too

Aaronontheweb (changes requested):
So I approve of the concept and the changes, but we need to introduce these in a more backwards-compatible way - along with the changes we're making to the `ServiceProviderProps` in a separate PR.

I'll help make this happen as part of the 1.4.18 release.

SamEmber:
Thanks for taking the time to look at this Aaron, these all sound like sensible changes to me. Pleased that my suggestion will be taken into account, please let me know if there is anything I can do to help out.

Aaronontheweb:
Thanks @SamEmber - we'll be discussing these changes at our community standup today: https://github.com/akkadotnet/akka.net/discussions/4856 - my hope is to get them rolled up and out soon

Aaronontheweb:
Need to get this one merged and approved into `dev` first, then we can integrate this PR: https://github.com/akkadotnet/akka.net/pull/4858

SamEmber:
Hey @Aaronontheweb, was this ever integrated?

Aaronontheweb:
@SamEmber not yet - was up to my neck in Akka.Cluster issues over the past month. We will probably bump this to 1.4.20

chris-sung:
> Secondly I have added the ability to resolve actor props from a Type param, this should make it easier to create actors dynamically.
I thought we need this too. there's no easy way to dynamically create an actor with ServiceProvider and that is my pain point writing test with ServiceProvider

Also, I've been writing `using ServiceProvider = Akka.DependencyInjection.ServiceProvider;`  everywhere I use ServiceProvider due to conflicting class name and is quite a pain, could we mark the old class obsolete and let it sit backward compatible way and moving on?

Aaronontheweb:
>Also, I've been writing using ServiceProvider = Akka.DependencyInjection.ServiceProvider; everywhere I use ServiceProvider due to conflicting class name and is quite a pain, could we mark the old class obsolete and let it sit backward compatible way and moving on?

We'll give it our best shot.

Aaronontheweb:
Ok, just introduced my non-breaking API implementation version of this - but someone else (@chris-sung @SamEmber included) will have to review it before we can merge it.

* [x] Old APIs should be marked as obsolete
* [x] Should offer type-checking (i.e. generics) all the way down the stack up until we reach `Props` if wanted
* [x] New APIs should be painless to switch to
* [x] Documentation should feature new APIs
* [x] No breaking API changes if we can absolutely help it

Pretty sure I did all of that, but need the review to verify.

src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProviderDependencyResolver.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+        
+        public Props Props(Type type, params object[] args)
+        {
+            return Akka.Actor.Props.CreateBy(new ServiceProviderActorProducer(ServiceProvider, type, args));
+        }
+        
+        public Props Props(Type type)
+        {
+            return Akka.Actor.Props.CreateBy(new ServiceProviderActorProducer(ServiceProvider, type, Array.Empty<object>()));
```
Arkatufus, line 50: Its probably better if this calls Props(Type, params object[]) and have a type check using `Type.IsAssignableTo` to make sure that the type passed in actually inherits `ActorBase` (type safety)

Arkatufus (changes requested):
Everything looks fine, I just have a little suggestion for one of the methods.

src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProviderDependencyResolver.cs
Aaronontheweb: Good idea

Aaronontheweb (approved):
LGTM

src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProviderDependencyResolver.cs
Aaronontheweb: Resolved

SamEmber:
Awesome, this looks great @Aaronontheweb. Thanks for taking the time to make the required changes.

Aaronontheweb:
Crap, looks like I broke the build - that's what I get for attempting to edit a PR using the Github editor :p

Aaronontheweb:
No problem @SamEmber - thank you for your contribution and sorry it took me so long to get around to this. Didn't take me long to implement the changes we discussed at all, just had a full plate up until this week.

christallire:
Fantastic. Thanks for the quick resolution

PR #4816, opened 2021-03-09
Add check for null message in LogDeadLetter.

### LINKED ISSUE #4815: Null message causes crash of DeadLetterListener
Akka doesn't support sending NULL messages between actors, however DeadLetterListener should not crash because of unhandled NULL message. If such message appears in the system, it causes failure in LogDeadLetter method which tries to evaluate its type. Because of the failure it's not possible to find out the message origin and recipient.

I am sending a PR for this issue shortly.

object:
This PR fixes #4815 by checking if a message is null prior evaluating its type, so the DeadLetterListener does not crash.


Aaronontheweb (approved):
LGTM

PR #4844, opened 2021-03-11
Update API Approver reporting

Arkatufus:
The reporting of `Akka.API.Tests` really needs an overhaul.
1. It relies on a very flaky `DiffEngine` to try and guess the best diff viewer installed in the system based on the file extension.
2. It assumes that the person running the test can access both the approved.txt and received.txt file to view the diff
3. `DiffEngine` often fails, resulting in an exception being thrown.
4. There is no diff viewer in the CI pipeline, making debugging impossible.

The better route is to actually embed the diff into the unit test report itself, so that it appears in the failing unit test report.
Diff message format:
```
<<<<<<<<< [ Approved file name ]
[ Line number ] + This line is newly added
[ Line number ] - This line is removed
[ Line number ] ? This line is modified
=========
[ Line number ] + This line is newly added
[ Line number ] - This line is removed
[ Line number ] ? This line is modified
>>>>>>>>> [ Received file name ]
```

Issue #4765, opened 2021-02-08
Approval specs do not report failure properly


IgorFedchenko:
When we have public API modified, approval specs are failing. But instead of proper failure report we get NRE:
```
System.NullReferenceException
Object reference not set to an instance of an object.
   at ApprovalTests.Reporters.MultiReporter.Report(String approved, String received)
   at ApprovalTests.Approvers.FileApprover.ReportFailure(IApprovalFailureReporter reporter)
   at ApprovalTests.Core.Approver.Verify(IApprovalApprover approver, IApprovalFailureReporter reporter)
   at ApprovalTests.Approvals.Verify(IApprovalApprover approver, IApprovalFailureReporter reporter)
   at ApprovalTests.Approvals.Verify(IApprovalWriter writer, IApprovalNamer namer, IApprovalFailureReporter reporter)
   at ApprovalTests.Approvals.Verify(IApprovalWriter writer)
   at ApprovalTests.Approvals.Verify(String text, Func`2 scrubber)
   at Akka.API.Tests.CoreAPISpec.ApproveCore()
```

Contributor is able to check the reason locally, but still getting NRE in approval specs is frustrating.

IgorFedchenko:
So far debugging shows be that when reporting failure, `ApprovalTests.Reporters.ApprovalReporter` class is used, and is trying to resolve `Type.GetType("Xunit.Assert, xunit")`, and fails.

And this is happening because `XUnitReporter` is used instead of `XUnit2Reporter` internally during approval report generation.
The following line returns `false` in `XUnit2Reporter`:
```csharp
AppDomain.CurrentDomain.GetAssemblies().Any<Assembly>(a => a.FullName.Contains("xunit.assert"))
```

`xunit.assert` is not included in the currently loaded assemblies.

IgorFedchenko:
Seems like the problem is due to the fact that this assembly is lazily loaded into the context. Not sure if/how ApprovalTests library should fix that, but I will push PR to fix this for our CI.

PR #4848, opened 2021-03-12
Fix Akka.API.Tests spec always throws NRE when failed in CI/CD pipeline

Arkatufus:
The built-in FrameworkAssertReporter that is being called inside the DiffReporter class is buggy in a CI/CD environment because it is trying to be clever, could not distinguish between XUnit and XUnit2, and will throw Null Reference Exception every time it ran.

This is probably fixed in latest version of ApiApprover but we couldn't switch to the latest version because the latest ApiGenerator returns a different API report format.

FIX: This hack removes FrameworkAssertReporter from the possible reporter list and retains all of the other reporters in DiffReporter

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.cs
+++ b/src/core/Akka.API.Tests/CoreAPISpec.cs
@@ -26,18 +26,21 @@ using Akka.Persistence.Query.Sql;
 using Akka.Persistence.Sql.Common.Journal;
 using ApprovalTests.Core;
 using ApprovalTests.Reporters;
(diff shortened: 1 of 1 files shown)


IgorFedchenko (commented):
I think you can just get this line back:
https://github.com/akkadotnet/akka.net/pull/4844/files#diff-bac245bb155e2a86b4f2d26e6059df95a7e7161cc1ba4fc740fd3cc1afa9153bL37

There was an issue for that with fix that we accidently removed in previous PR: https://github.com/akkadotnet/akka.net/issues/4765

PR #4858, opened 2021-03-17
ServiceProviderProps and Props Changes

### LINKED ISSUE #3599: Props methods for propagating and setting IActorProducer are pretty weird
Akka.NET Version: v1.3.9

This is how we set the `ActorProducer` inside `Props` today when we're using dependency injection:

https://github.com/akkadotnet/akka.net/blob/3c6115eb99960cf496a4ec56e565a0d34f0c991b/src/core/Akka/Actor/Props.cs#L292-L298

It's indirect, and we repeat this process of using reflection on the input type to figure out if we should use a custom `ActorProducer` whenever someone calls any of the fluent interface methods on `Props`, all of which end up calling the `Props.Copy()` method:

https://github.com/akkadotnet/akka.net/blob/3c6115eb99960cf496a4ec56e565a0d34f0c991b/src/core/Akka/Actor/Props.cs#L583-L586

I'm not sure why we designed it this way, but it's gross, relies on magic, and not all that extensible. What irks me most about it is that the `Type` we feed into `Props` can mean different things other than the actor's implementation. I'd rather change the signatures on `Props` to make it 100% explicit "I AM USING THIS PRODUCER" and not let the arguments of various `Props` constructors take on different meanings for the same input.

How we do this without breaking 100% of existing DI plugins for Akka.NET is an interesting question. I guess a reasonable way of handling that is just "upgrade all of them" since this change doesn't have a meaningful impact on end-user code today, but it would break binary compatibility since fixing this issue might end up changing either the signatures of existing `Props` methods or changing their meanings. That doesn't seem like a big deal.

### LINKED ISSUE #4853: Akka.DependencyInjection: ServiceProviderProps is broken - needs to be rewritten
Version: 1.4.17

Ran into this while investigating https://github.com/petabridge/phobos-issues/issues/16 - the problem is that any of the methods you might typically use on `Props`, i.e. `Props.WithSupervisorStrategy` et al don't work with the `ServiceProviderProps`.

Instead, we need to refactor that mechanism to use the `IIndirectActorProducer` much like how we did with the original Akka.DI.Core implementation here:

https://github.com/akkadotnet/akka.net/blob/444432a9ef6cf8c82140996843be0c59b877ec55/src/contrib/dependencyinjection/Akka.DI.Core/DIActorProducer.cs#L16-L66

So this is going to be a breaking change to Akka.DependencyInjection, although one that shouldn't have much real-world impact as the underlying types are all abstracted away from the users of the library and the `ServiceProvider` interface shouldn't have any problems with it.

Aaronontheweb:
close #4853 
close #3599

API approvals and `ServiceProviderProps` changes aren't done yet - just wanted to kick off the test suite to make sure I haven't massively blown it on introducing these changes safely.

### DIFF
--- a/src/core/Akka/Actor/Props.cs
+++ b/src/core/Akka/Actor/Props.cs
@@ -177,13 +177,12 @@ namespace Akka.Actor
         /// <param name="deploy">The configuration used to deploy the actor.</param>
         /// <param name="strategy">The supervisor strategy to use.</param>
         /// <param name="args">The arguments needed to create the actor.</param>
-        internal Props(IIndirectActorProducer producer, Deploy deploy, SupervisorStrategy strategy, params object[] args)
+        internal Props(IIndirectActorProducer producer, Deploy deploy, params object[] args)
         {
             Deploy = deploy;
             _inputType = producer.ActorType;
             Arguments = args ?? NoArgs;
             _producer = producer;
-            SupervisorStrategy = strategy;
         }
 
         /// <summary>
@@ -361,7 +360,7 @@ namespace Akka.Actor
 
             var args = newExpression.GetArguments();
 
-            return new Props(new ActivatorProducer(typeof(TActor), args), DefaultDeploy, supervisorStrategy, args);
+            return new Props(new ActivatorProducer(typeof(TActor), args), DefaultDeploy, args){ SupervisorStrategy = supervisorStrategy };
         }
 
         /// <summary>
@@ -372,7 +371,7 @@ namespace Akka.Actor
         /// <returns>The newly created <see cref="Akka.Actor.Props" />.</returns>
         public static Props Create<TActor>(params object[] args) where TActor : ActorBase
         {
-            return new Props(new ActivatorProducer(typeof(TActor), args), DefaultDeploy, SupervisorStrategy.DefaultStrategy, args);
+            return new Props(new ActivatorProducer(typeof(TActor), args), DefaultDeploy, args);
         }
 
         /// <summary>
@@ -395,7 +394,7 @@ namespace Akka.Actor
         /// <returns>The newly created <see cref="Akka.Actor.Props" />.</returns>
         public static Props CreateBy(IIndirectActorProducer producer, params object[] args)
         {
-            return new Props(producer, DefaultDeploy, SupervisorStrategy.DefaultStrategy, args);
+            return new Props(producer, DefaultDeploy, args);
         }
 
         /// <summary>
@@ -406,7 +405,7 @@ namespace Akka.Actor
         /// <returns>The newly created <see cref="Akka.Actor.Props" />.</returns>
         public static Props Create<TActor>(SupervisorStrategy supervisorStrategy) where TActor : ActorBase, new()
         {
-            return new Props(typeof(TActor), supervisorStrategy);
+            return new Props(new ActivatorProducer(typeof(TActor), NoArgs), DefaultDeploy, NoArgs){ SupervisorStrategy = supervisorStrategy };
         }
 
         /// <summary>
@@ -556,7 +555,7 @@ namespace Akka.Actor
         /// <returns>The newly created <see cref="Akka.Actor.Props" /></returns>
         protected virtual Props Copy()
         {
-            return new Props(Deploy, _inputType, Arguments) { SupervisorStrategy = SupervisorStrategy };
+            return new Props(_producer, Deploy, Arguments) { SupervisorStrategy = SupervisorStrategy };
         }
 
         [Obsolete("we should not be calling this method. Pass in an explicit IIndirectActorProducer reference isntead.")]
--- a/src/core/Akka/Actor/Props.cs
+++ b/src/core/Akka/Actor/Props.cs
@@ -162,7 +162,7 @@ namespace Akka.Actor
         /// <param name="args">The arguments needed to create the actor.</param>
         /// <exception cref="ArgumentException">This exception is thrown if <paramref name="type" /> is an unknown actor producer.</exception>
         public Props(Deploy deploy, Type type, params object[] args) 
-            : this(CreateProducer(type, args), deploy, args) // have to preserve the "CreateProducer" call here to preserve backwards compat with Akka.DI.Core
+            : this(CreateProducer(type, args), deploy, SupervisorStrategy.DefaultStrategy, args) // have to preserve the "CreateProducer" call here to preserve backwards compat with Akka.DI.Core
         {
 
         }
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
Should be good to go now.

src/core/Akka/Actor/Props.cs
```diff
@@ -285,37 +155,51 @@ public Props(Deploy deploy, Type type, IEnumerable<object> args)
...
+        ///     Initializes a new instance of the <see cref="Props" /> class.
         /// </summary>
         /// <param name="deploy">The configuration used to deploy the actor.</param>
         /// <param name="type">The type of the actor to create.</param>
         /// <param name="args">The arguments needed to create the actor.</param>
-        /// <exception cref="ArgumentException">This exception is thrown if <paramref name="type"/> is an unknown actor producer.</exception>
-        public Props(Deploy deploy, Type type, params object[] args)
+        /// <exception cref="ArgumentException">This exception is thrown if <paramref name="type" /> is an unknown actor producer.</exception>
+        public Props(Deploy deploy, Type type, params object[] args) 
+            : this(CreateProducer(type, args), deploy, args) // have to preserve the "CreateProducer" call here to preserve backwards compat with Akka.DI.Core
```
Aaronontheweb, line 165: First big change - this used to be the "bottom" constructor that all other constructors call. I've added one more `internal` one below it. 

We needed to keep the mysterious / bad `CreateProducer` call here in order not to completely blow up Akka.DI.Core support, which many users are probably still using.

src/core/Akka/Actor/Props.cs
```diff
@@ -285,37 +155,51 @@ public Props(Deploy deploy, Type type, IEnumerable<object> args)
...
+        /// <param name="args">The arguments needed to create the actor.</param>
+        internal Props(IIndirectActorProducer producer, Deploy deploy, params object[] args)
         {
             Deploy = deploy;
-            inputType = type;
-            Arguments = args ?? noArgs;
-            producer = CreateProducer(inputType, Arguments);
+            _inputType = producer.ActorType;
+            Arguments = args ?? NoArgs;
+           _producer = producer;
```
Aaronontheweb, line 184: `IIndirectActorProducer` is now passed in directly, rather than instantiated via reflection magic.

src/core/Akka/Actor/Props.cs
```diff
@@ -398,35 +357,47 @@ public static Props Empty
...
-        /// Creates an actor using the given arguments.
+        ///     Creates an actor using the given arguments.
         /// </summary>
         /// <typeparam name="TActor">The type of the actor to create.</typeparam>
         /// <param name="args">The arguments needed to create the actor.</param>
         /// <returns>The newly created <see cref="Akka.Actor.Props" />.</returns>
         public static Props Create<TActor>(params object[] args) where TActor : ActorBase
         {
-            return new Props(typeof(TActor), args);
+            return new Props(new ActivatorProducer(typeof(TActor), args), DefaultDeploy, args);
```
Aaronontheweb, line 373: Use the `ActivatorProducer` directly, rather than use reflection to create it.

src/core/Akka/Actor/Props.cs
```diff
@@ -398,35 +357,47 @@ public static Props Empty
...
 
         /// <summary>
-        /// Creates an actor using a specified actor producer.
+        ///     Creates an actor using a specified actor producer.
         /// </summary>
         /// <typeparam name="TProducer">The type of producer used to create the actor.</typeparam>
         /// <param name="args">The arguments needed to create the actor.</param>
         /// <returns>The newly created <see cref="Akka.Actor.Props" />.</returns>
+        [Obsolete("Do not use this method. Call CreateBy(IIndirectActorProducer, params object[] args) instead")]
         public static Props CreateBy<TProducer>(params object[] args) where TProducer : class, IIndirectActorProducer
```
Aaronontheweb, line 383: This method sucks - marked as `Obsolete`.

src/core/Akka/Actor/Props.cs
```diff
@@ -398,35 +357,47 @@ public static Props Empty
...
         }
 
         /// <summary>
-        /// Creates an actor using a specified supervisor strategy.
+        ///     Creates an actor using a specified actor producer.
+        /// </summary>
+        /// <param name="producer">The actor producer that will be used to create the underlying actor..</param>
+        /// <param name="args">The arguments needed to create the actor.</param>
+        /// <returns>The newly created <see cref="Akka.Actor.Props" />.</returns>
+        public static Props CreateBy(IIndirectActorProducer producer, params object[] args)
```
Aaronontheweb, line 394: Take an explicit `IIndirectActorProducer` - the `ServiceProvider` extension now calls this method directly.

src/core/Akka/Actor/Props.cs
```diff
@@ -398,35 +357,47 @@ public static Props Empty
...
+        {
+            return new Props(producer, DefaultDeploy, args);
```
Aaronontheweb, line 396: even though the args presumably get passed intro the `IIndirectActorProducer`, we still have to pass them into the `Props` constructor because they're used for things like remote deployments.

src/core/Akka/Actor/Props.cs
```diff
@@ -533,88 +500,152 @@ public Props WithDeploy(Deploy deploy)
...
-        /// <returns>The newly created <see cref="Akka.Actor.Props"/></returns>
+        /// <returns>The newly created <see cref="Akka.Actor.Props" /></returns>
         protected virtual Props Copy()
         {
-            return new Props(Deploy, inputType, Arguments) { SupervisorStrategy = SupervisorStrategy };
+            return new Props(Deploy, _inputType, Arguments) { SupervisorStrategy = SupervisorStrategy };
+        }
+
+        [Obsolete("we should not be calling this method. Pass in an explicit IIndirectActorProducer reference isntead.")]
+        private static IIndirectActorProducer CreateProducer(Type type, object[] args)
```
Aaronontheweb, line 561: Source of most of the weirdness I identified in https://github.com/akkadotnet/akka.net/issues/3599 stem from this method, which conflates accepting a `Type` for an actor with a `Type` for an `IIndirectActorProducer`, which has the effect of making it really difficult to reason about what is happening internally and performing reflection twice for all actor invocations.

Aaronontheweb (commented):
Described my changes in this PR

Aaronontheweb:
I think I have the test suite under control now - going to run some benchmarks next.

Aaronontheweb:
With this PR:

|      Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------ |---------:|---------:|---------:|-------:|-------:|------:|----------:|
| Actor_spawn | 23.75 us | 1.253 us | 0.829 us | 1.5700 | 0.3400 |     - |   9.31 KB |

Dev:

|      Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------ |---------:|---------:|---------:|-------:|-------:|------:|----------:|
| Actor_spawn | 24.10 us | 1.209 us | 0.799 us | 1.5600 | 0.3400 |     - |   9.31 KB |

Looks like we got a small improvement by stripping out some of the extra reflection.

Aaronontheweb:
Eh, performance difference looks about the same actually - probably no negative or positive impact from this change.

Arkatufus (approved):
LGTM

PR #4861, opened 2021-03-18
Fix actor TestActorRef<> loosing Self reference in async context

### LINKED ISSUE #4823: Actor loses context in the context of test
When creating a new issue, please make sure the following information is part of your issue description. (if applicable). Thank You!

Pacakages:
- Akka 1.3.17
- Akka.TestKit.NUnit: 1.3.8

OS:
- Windows 10

Issue:

When running in a actor test and using ReceiveAsync the actor loses context.

```
var props = Props.Create(() => new MyActor());
var identity = ActorOfAsTestActorRef<MyActor>(props, TestActor);
var msg=new MyMessage();
identity.Tell(msg);

```
image

This works when the Self.Tell is moved to before the async message, but the code in the msg being send in the image is never received.

I've had similar issues in other tests as well.

IgorFedchenko:
Closes #4823

### DIFF
--- a/src/core/Akka.Tests/Actor/BugFix4823Spec.cs
+++ b/src/core/Akka.Tests/Actor/BugFix4823Spec.cs
@@ -1,0 +1,58 @@
+// //-----------------------------------------------------------------------
+// // <copyright file="BugFix4823Spec.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2021 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2021 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Event;
+using Akka.TestKit;
+using Xunit.Abstractions;
+using FluentAssertions;
+using Xunit;
+
+namespace Akka.Tests.Actor
+{
+    public class BugFix4823Spec : AkkaSpec
+    {
+        public BugFix4823Spec(ITestOutputHelper outputHelper) : base(outputHelper)
+        {
+        }
+
+        [Fact]
+        public void Actor_should_not_loose_self_context_after_async_call()
+        {
+            var props = Props.Create(() => new MyActor(TestActor));
+            var identity = ActorOfAsTestActorRef<MyActor>(props, TestActor);
+            identity.Tell(new MyMessage());
+            var selfBefore = ExpectMsg<IActorRef>();
+            var selfAfter = ExpectMsg<IActorRef>();
+            selfAfter.Should().Be(selfBefore);
+        }
+
+        class MyActor : ReceiveActor
+        {
+            public MyActor(IActorRef testActor)
+            {
+                ReceiveAsync<MyMessage>(async msg =>
+                {
+                    testActor.Tell(Self);
+                    Become(Tracking);
+                    await SomeAsyncProcessing(msg);
+                    testActor.Tell(Self);
+                });
+            }
+
+            private void Tracking()
+            {
+                ReceiveAny(m => Context.GetLogger().Info($"Received {m} in tracking state"));
+            }
+
+            private Task SomeAsyncProcessing(MyMessage msg) => Task.Delay(100); // changing to Task.CompletedTask does not reproduce bug
+        }
+        
+        class MyMessage { }
+    }
+}
--- a/src/core/Akka.Tests/Actor/BugFix4823Spec.cs
+++ b/src/core/Akka.Tests/Actor/BugFix4823Spec.cs
@@ -24,8 +24,8 @@ namespace Akka.Tests.Actor
         [Fact]
         public void Actor_should_not_loose_self_context_after_async_call()
         {
-            var props = Props.Create(() => new MyActor(TestActor));
-            var identity = ActorOfAsTestActorRef<MyActor>(props, TestActor);
+            var identity = ActorOfAsTestActorRef<MyActor>(Props.Create(() => new MyActor(TestActor)), TestActor);
+            // var identity = Sys.ActorOf(Props.Create(() => new MyActor(TestActor)));
             identity.Tell(NotUsed.Instance);
             var selfBefore = ExpectMsg<IActorRef>();
             var selfAfter = ExpectMsg<IActorRef>();
--- a/src/core/Akka.Tests/Actor/BugFix4823Spec.cs
+++ b/src/core/Akka.Tests/Actor/BugFix4823Spec.cs
@@ -25,7 +25,6 @@ namespace Akka.Tests.Actor
         public void Actor_should_not_loose_self_context_after_async_call()
         {
             var identity = ActorOfAsTestActorRef<MyActor>(Props.Create(() => new MyActor(TestActor)), TestActor);
-            // var identity = Sys.ActorOf(Props.Create(() => new MyActor(TestActor)));
             identity.Tell(NotUsed.Instance);
             var selfBefore = ExpectMsg<IActorRef>();
             var selfAfter = ExpectMsg<IActorRef>();
--- a/src/core/Akka.Tests/Actor/BugFix4823Spec.cs
+++ b/src/core/Akka.Tests/Actor/BugFix4823Spec.cs
@@ -26,7 +26,7 @@ namespace Akka.Tests.Actor
         {
             var props = Props.Create(() => new MyActor(TestActor));
             var identity = ActorOfAsTestActorRef<MyActor>(props, TestActor);
-            identity.Tell(new MyMessage());
+            identity.Tell(NotUsed.Instance);
             var selfBefore = ExpectMsg<IActorRef>();
             var selfAfter = ExpectMsg<IActorRef>();
             selfAfter.Should().Be(selfBefore);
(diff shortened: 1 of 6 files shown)


IgorFedchenko:
Worth to mention that changing `TestActorRef` usage to using `MyActor` directly stops spec from failing - which means that the issue is most likely related to test suite / `TestActorRef` implementation rather then to actor's message handling itself.

IgorFedchenko:
So, the problem is following.

Here we are setting current actor execution context, while running it's `Receive` handler:
https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/core/Akka/Actor/ActorCell.cs#L373

We are setting current actor context, executing action in that context, then switching context back. So far so good.

Now, for `async` receive handlers, this code is used:
https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/core/Akka/Dispatch/ActorTaskScheduler.cs#L145

Basically, we start the task in background, but do not process any more message from actor's mailbox until task is completed.

But! Right when we call `await ProcessMessage(msg)` inside actor receive handler, the synchronous part of handling is finished, receive action is completed and context is set back to the previous one in this code:
https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/core/Akka/Actor/ActorCell.cs#L370-L379

That's why, after calling async operation, context is almost immediately lost.

Note that this happens only with `InternalTestActorRef`, because it's `Tell` call ends up with synchronous receiver handler execution. I have a feeling that this could be somehow reproduced without `InternalTestActorRef` (because the code above will set old context before async handled executed for all types of actor cells), but I guess that would require caller and receiver actor running in the same thread, or something like that.

IgorFedchenko:
Looks like it is `ActorCellKeepingSynchronizationContext` who does not set valid current context for async task continuation:
https://github.com/akkadotnet/akka.net/blob/b76b98af84232b45327cc8d0a336d39434d9018c/src/core/Akka.TestKit/ActorCellKeepingSynchronizationContext.cs#L47

It sets default actor cell to test actor cell, which is then used in receiver handler after `await` call.

IgorFedchenko:
All right, test suite is passing to nothing important should be broken by this fix. Technically, this should only influent only those test cases where TestActor underlying actor executes any async handling, but still need someone to review this.

Arkatufus (commented):
I'm not sure if this is actually addressing the real problem.

What needs to be fixed is the real underlying problem of passing context around in an async environment, the test kit should never be touched and should instead be used as a tool to prove that the underlying implementation is correct.

IgorFedchenko:
Well, yeah, that's a kind of a hack to make test suite working...

Context is stored in `[ThreadStatic]` static properly, which is resetted after `await` call inside message handler. Ideally this had to be fully rewritten, but 1) such change is very dangerous and 2) everything works fine when you are using usual actor instead of `TestActorRef`, because it is using different synchronization context I believe. 

So here I am trying to make a hack that would fix out test suite hack which is used to support `async` in test environment (I mean `ActorCellKeepingSynchronizationContext` context from test suite).

This PR should help to the user with original issue - on the other hand, probably we need to provide some easy workaround for him and keep the issue open. What do you think @Arkatufus ?

Aaronontheweb:
> What needs to be fixed is the real underlying problem of passing context around in an async environment, the test kit should never be touched and should instead be used as a tool to prove that the underlying implementation is correct.

Both the bug and the fix look specific to the `TestKit` to me. Based on the screenshot provided by the original user, it looks like the problem is that the `TestActorOf<>` gets excluded upon the continuation and it's only the underlying "wrapped" actor that is included in the context. The way we pass around async context right now is parking / unparking a thread-static when an actor gets activated and that mechanism is working in this reproduction - the problem is that we hack on top of that mechanism in the `TestActorRef<T>` in order to make it possible to inspect what's happening inside the concrete actor implementation and that's where the bug is.

src/core/Akka/Dispatch/ActorTaskScheduler.cs
```diff
@@ -21,11 +21,26 @@ namespace Akka.Dispatch
...
         /// TBD
         /// </summary>
         public object CurrentMessage { get; private set; }
 
+        /// <summary>
+        /// Wraps async task scheduling with pre/post actions
+        /// </summary>
+        /// <param name="beforeStartStarted">Executed before scheduling the task call/></param>
+        /// <param name="afterTaskCompleted">Executed in task continuation after task completion</param>
+        internal void WrapTaskScheduling(Action beforeStartStarted = null, Action afterTaskCompleted = null)
```
Aaronontheweb, line 38: Better to have the `TestActorRef` subclass the `ActorTaskScheduler` with a custom implementation that does this rather than introduce mutable state inside the `ActorTaskScheduler` implementation shared / used by all actors.

src/core/Akka.Tests/Actor/BugFix4823Spec.cs
```diff
@@ -0,0 +1,47 @@
...
+namespace Akka.Tests.Actor
+{
+    public class BugFix4823Spec : AkkaSpec
+    {
+        public BugFix4823Spec(ITestOutputHelper outputHelper) : base(outputHelper)
+        {
+        }
+
+        [Fact]
+        public void Actor_should_not_loose_self_context_after_async_call()
```
Aaronontheweb, line 25: Reproduction spec looks good

Aaronontheweb (changes requested):
LGTM, but need to not touch the `ActorTaskScheduler` - subclass instead

Aaronontheweb:
@IgorFedchenko looks like we're missing API approvals

IgorFedchenko:
It looks like only racy specs are failing now @Aaronontheweb

src/core/Akka/Dispatch/ActorTaskScheduler.cs
```diff
@@ -142,6 +152,8 @@ public static void RunTask(Func<Task> asyncAction)
             ActorTaskScheduler actorScheduler = context.TaskScheduler;
             actorScheduler.CurrentMessage = context.CurrentMessage;
```
Aaronontheweb, line 153: I'm thinking, after having just looked at this code for a bunch of Phobos issues (i.e. https://github.com/petabridge/phobos-issues/issues/19) - I wonder if we should just preserve the entire `Context` here inside the default `ActorTaskScheduler`, since it seems like this issue could possibly come up inside an end-user actor and not just during unit tests.

IgorFedchenko: Well, to be completely honest, I am not sure about the reason why this was failing only in test actors so far. All that I can tell is that if I change TestActor to real actor in verification spec, it is passing the test.

So yeah, there is a chance that the issue is more serious. But this is a core part of the framework, and I am not sure how to fix it properly - that's why I found kind of "easy fix" for the test actor case. We can merge this as is to make it working better then it was before, and keep an eye on other user requests that complain to `async` support. But I do not mind dropping this PR and creating a new one with more fundamental fix... It's just I will unlikely be able to propose something better then this :)

Aaronontheweb:
Ultimately this is less risky than https://github.com/akkadotnet/akka.net/pull/5126 - going to merge this in instead.

I don't have time to validate a big `ActorTaskScheduler` redesign right now, but this should suffice.

Aaronontheweb (approved):

Aaronontheweb:
> I'm not sure if this is actually addressing the real problem.
> 
> What needs to be fixed is the real underlying problem of passing context around in an async environment, the test kit should never be touched and should instead be used as a tool to prove that the underlying implementation is correct.

I agree with this, but it's a giant project that can't be safely introduced without a major version bump https://github.com/akkadotnet/akka.net/issues/5128

pirovorster:
Hi @Aaronontheweb / @IgorFedchenko 

Just a quick question on this fix vs #5128: What are the caveats we need to know about? ie what does this merge not cover from that #5128 will?

pirovorster:
@Aaronontheweb  / @IgorFedchenko ?

Aaronontheweb:
sorry for the delay @pirovorster 

There are no caveats to this change really - I looked under the covers pretty deeply here in November to see about introducing `await` tracing inside Phobos (which we shipped with [Phobos 1.3.0](https://phobos.petabridge.com/articles/releases/RELEASE_NOTES.html#130httpssdkbincompublisherpetabridgeproductphobospackagesphobosactorversions130-november-18th-2021)) and I didn't see any need to override the `ActorTaskScheduler` or any of those `virtual` methods that @IgorFedchenko added. We only really override those inside the TestKit so `TestActorRef<T>`, which doesn't behave 100% like a normal actor due to how it's exposed for testability, doesn't lose track of its own context during `ReceiveAsync` calls.

https://github.com/akkadotnet/akka.net/issues/5128 - this on the other hand involves rewriting how all dispatchers work. Right now an actor's internal `Receive` is a blocking synchronous method and support its `await`s using some internal messaging and suspending the actor's mailbox during each `await` in order to preserve message order guarantees. This change would effectively turn the actor's mailbox processing routine into something resembling an `IAsyncEnumerable` (not literally) - which has a lot of benefits in terms of simplifying how dispatching works especially when it comes to supporting `await` inside actors.

It's also a change that could have massive side effects, hence why I think me should probably hold off on doing it until a major version bump.

pirovorster:
Hi @Aaronontheweb 

Thanks. I'll remove some of the work arounds we built for these and give it a go.

PR #4868, opened 2021-03-23
made Deploy methods virtual

Aaronontheweb:
Made the `Deploy` methods virtual so `Deploy` can be overridden in subsequent systems to carry additional actor information.

### DIFF
--- a/src/core/Akka/Actor/Props.cs
+++ b/src/core/Akka/Actor/Props.cs
@@ -63,7 +63,7 @@ namespace Akka.Actor
         /// </summary>
         /// <param name="copy">The object that is being cloned.</param>
         protected Props(Props copy)
-            : this(copy.Deploy, copy._inputType, copy.Arguments)
+            : this(copy._producer, copy.Deploy, copy.Arguments)
         {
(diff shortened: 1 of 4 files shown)


src/core/Akka/Actor/Props.cs
```diff
@@ -63,8 +63,9 @@ protected Props()
         /// </summary>
         /// <param name="copy">The object that is being cloned.</param>
         protected Props(Props copy)
-            : this(copy.Deploy, copy._inputType, copy.SupervisorStrategy, copy.Arguments)
+            : this(copy.Deploy, copy._inputType, copy.Arguments)
         {
+            SupervisorStrategy = copy.SupervisorStrategy;
```
Arkatufus, line 68: Should we also copy over the producer at this step?

Arkatufus (changes requested):

src/core/Akka/Actor/Props.cs
Aaronontheweb: Yes, we should.

Arkatufus (approved):
LGTM

PR #4871, opened 2021-03-24
Add a way to signal a manual flush to stream FileSubscriber

### LINKED ISSUE #4867: Akka.Streams.IO: need to add "Flush" command for FileIO that can be invoked from outside
Note - need to add an issue to make this programmable from the outside - i.e. a flow stage that can send a `Flush` signal to the `FileSubscriberActor`.

_Originally posted by @Aaronontheweb in https://github.com/akkadotnet/akka.net/pull/4859#r599545880_

Arkatufus:
Closes  #4867

@Aaronontheweb What do you think of this implementation


Aaronontheweb (changes requested):
LGTM - looks like a reasonable way to side-car an Akka.Streams

Aaronontheweb (approved):
LGTM

Issue #4537, opened 2020-08-05
DedicatedThreadPool Performance Optimization


Aaronontheweb:
Version: 1.4.9

Prior to merging in https://github.com/akkadotnet/akka.net/pull/4511 , which moves all `/system` actors on a `ForkJoinDispatcher`, we need to go about improving the structure of the `DedicatedThreadPool` class (used by the `ForkJoinDispatcher`), which is sourced originally to this repository:

https://github.com/helios-io/DedicatedThreadPool

But that source code is referenced as a file directly into our project here:

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Helios.Concurrency.DedicatedThreadPool.cs

The two code bases don't diverge much and that can easily be remedied. The last real change we made to the `DedicatedThreadPool` class was in 2017: https://github.com/akkadotnet/akka.net/pull/2714

## Problems that #4511 May Introduce: Rise in Idle CPU Consumption
Running all of the `/system` actors on the `ForkJoinDispatcher` will help improve performance in really busy systems - it will allow workloads for Akka.Persistence, Akka.Cluster, etc independently from the primary .NET `ThreadPool` since the `/system` actors have to operate under real-time constraints. The only point of contention will be the OS thread scheduler, rather than the length of the work queue inside the shared `ThreadPool`. We've used this strategy successfully for years with Akka.Remote.

The problem is: what happens with `ActorSytem` instances that make virtually zero use of `/system` actors, such as the `ActorSystem`s that are used inside stand-alone plugins like https://github.com/petabridge/Petabridge.Tracing.Zipkin ? The answer is we may start seeing increases in background CPU consumption as worker threads spin / wait / sleep for work, such as what we can observe with the `HashedWheelTimerScheduler` here: https://github.com/akkadotnet/akka.net/issues/4031

We need to avoid that to the extent that it's possible by freeing threads that don't have enough work do to and by adding new threads gradually as total demand increases. Spinning threads up and down is expensive, but having idle threads continuously sleeping and checking for work is also very expensive. We need to reach a happy medium.

## Existing Issues with `DedicatedThreadPool`
Other issues that the `DedicatedThreadPool` has currently:

1. End-to-end throughput - it's at least 20% slower than the `ThreadPool` and possibly moreso now since the DTP hasn't benefited from any of the performance improvements introduced during subsequent versions of .NET Core. Finding ways to improve this through the queuing mechanism, thread activation, and work distribution would be a huge plus.
[...]

PR #4882, opened 2021-03-30
Introduce `ChannelExecutor`

Aaronontheweb:
Working on deprecating the ole' `ForkJoinDispatcher` and thought I'd try using System.Threading.Channels - results are pretty disappointing so far. Current built-in implementations of the `ThreadPool` dispatchers absolutely stomp these ones in terms of performance.

### DIFF
--- a/docs/articles/actors/dispatchers.md
+++ b/docs/articles/actors/dispatchers.md
@@ -171,7 +171,7 @@ In Akka.NET v1.4.19 we will be introducing an opt-in feature, the `ChannelExecut
 
 During its initial development and benchmarks, we observed the following:
 
-1. The `ChannelExecutor` tremendously reduced idle CPU and max busy CPU even during peak message throughput, primarily as a result of dynamically shrinking the total `ThreadPool` to only the necessary size. This resolves one of the largest complaints large users of Akka.NET have today. However, **in order for this setting to be effective `ThreadPool.SetMin(0,0)` must also be set**. We are considering doing this inside the `ActorSystem.Create` method, but if you those settings don't work for you you can easily override them by simply calling `ThreadPool.SetMin(yourValue, yourValue)` again after `ActorSystem.Create` has exited.
+1. The `ChannelExecutor` tremendously reduced idle CPU and max busy CPU even during peak message throughput, primarily as a result of dynamically shrinking the total `ThreadPool` to only the necessary size. This resolves one of the largest complaints large users of Akka.NET have today. However, **in order for this setting to be effective `ThreadPool.SetMin(0,0)` must also be set**. We are considering doing this inside the `ActorSystem.Create` method, those settings don't work for you you can easily override them by simply calling `ThreadPool.SetMin(yourValue, yourValue)` again after `ActorSystem.Create` has exited.
 2. The `ChannelExecutor` actually beat the `ForkJoinDispatcher` and others on performance even in environments like Docker and bare metal on Windows.
 
 > [!NOTE]
@@ -201,7 +201,7 @@ akka.actor.internal-dispatcher = {
 
 akka.remote.default-remote-dispatcher {
     type = Dispatcher
-	  executor = channel-executor
+    executor = channel-executor
     fork-join-executor {
       parallelism-min = 2
       parallelism-factor = 0.5
--- a/docs/articles/actors/dispatchers.md
+++ b/docs/articles/actors/dispatchers.md
@@ -74,6 +74,7 @@ Some dispatcher configurations are available out-of-the-box for convenience. You
 * **task-dispatcher** - A configuration that uses the [TaskDispatcher](#taskdispatcher).
 * **default-fork-join-dispatcher** - A configuration that uses the [ForkJoinDispatcher](#forkjoindispatcher).
 * **synchronized-dispatcher** - A configuration that uses the [SynchronizedDispatcher](#synchronizeddispatcher).
+* **channel-executor** - new as of v1.4.19, the [`ChannelExecutor`](#channelexecutor) is used to run on top of the .NET `ThreadPool` and allow Akka.NET to dynamically scale thread usage up and down with demand in exchange for better CPU and throughput performance.
 
 ## Built-in Dispatchers
 
@@ -164,6 +165,63 @@ private void Form1_Load(object sender, System.EventArgs e)
   system.ActorOf(Props.Create<UIWorker>().WithDispatcher("synchronized-dispatcher"), "ui-worker");
 }
 ```
+
+### `ChannelExecutor`
+In Akka.NET v1.4.19 we will be introducing an opt-in feature, the `ChannelExecutor` - a new dispatcher type that re-uses the same configuration as a `ForkJoinDispatcher` but runs entirely on top of the .NET `ThreadPool` and is able to take advantage of dynamic thread pool scaling to size / resize workloads on the fly.
+
+During its initial development and benchmarks, we observed the following:
+
+1. The `ChannelExecutor` tremendously reduced idle CPU and max busy CPU even during peak message throughput, primarily as a result of dynamically shrinking the total `ThreadPool` to only the necessary size. This resolves one of the largest complaints large users of Akka.NET have today. However, **in order for this setting to be effective `ThreadPool.SetMin(0,0)` must also be set**. We are considering doing this inside the `ActorSystem.Create` method, but if you those settings don't work for you you can easily override them by simply calling `ThreadPool.SetMin(yourValue, yourValue)` again after `ActorSystem.Create` has exited.
+2. The `ChannelExecutor` actually beat the `ForkJoinDispatcher` and others on performance even in environments like Docker and bare metal on Windows.
+
+> [!NOTE]
+> We are in the process of gathering data from users on how well `ChannelExecutor` performs in the real world. If you are interested in trying out the `ChannelExecutor`, please read the directions in this document and then comment on [the "Akka.NET v1.4.19: ChannelExecutor performance data" discussion thread](https://github.com/akkadotnet/akka.net/discussions/4983).
+
+The `ChannelExectuor` re-uses the same threading settings as the `ForkJoinExecutor` to determine its effective upper and lower parallelism limits, and you can configure the `ChannelExecutor` to run inside your `ActorSystem` via the following HOCON configuration:
+
+```
+akka.actor.default-dispatcher = {
+    executor = channel-executor
+    fork-join-executor { #channelexecutor will re-use these settings
+      parallelism-min = 2
+      parallelism-factor = 1
+      parallelism-max = 64
+    }
+}
+
+akka.actor.internal-dispatcher = {
+    executor = channel-executor
+    throughput = 5
+    fork-join-executor {
+      parallelism-min = 4
+      parallelism-factor = 1.0
+      parallelism-max = 64
+    }
+}
+
+akka.remote.default-remote-dispatcher {
+    type = Dispatcher
+	  executor = channel-executor
+    fork-join-executor {
+      parallelism-min = 2
+      parallelism-factor = 0.5
+      parallelism-max = 16
+    }
+}
+
+akka.remote.backoff-remote-dispatcher {
+  executor = channel-executor
+  fork-join-executor {
+    parallelism-min = 2
+    parallelism-max = 2
+  }
+}
+```
+
+This will enable the `ChannelExecutor` to run everywhere and all Akka.NET loads, with the exception of anything you manually allocate onto a `ForkJoinDispatcher` or `PinnedDispatcher`, will be managed by the `ThreadPool`.
+
+> [!IMPORTANT]
+> As of Akka.NET v1.4.19, we call `ThreadPool.SetMinThreads(0,0)` inside the `ActorSystem.Create` method as we've found that the default `ThreadPool` minimum values have a negative impact on performance. However, if this causes undesireable side effects for you inside your application you can always override those settings by calling `ThreadPool.SetMinThreads(yourValue, yourValue)` again after you've created your `ActorSystem`.
 
 #### Common Dispatcher Configuration
 
--- a/docs/articles/clustering/cluster-singleton.md
+++ b/docs/articles/clustering/cluster-singleton.md
@@ -34,9 +34,6 @@ This pattern may seem to be very tempting to use at first, but it has several dr
 
 Especially the last point is something you should be aware of — in general when using the Cluster Singleton pattern you should take care of downing nodes yourself and not rely on the timing based auto-down feature.
 
-> [!WARNING]
-> Be very careful when using Cluster Singleton together with Automatic Downing, since it allows the cluster to split up into two separate clusters, which in turn will result in `multiple Singletons` being started, one in each separate cluster!
-
 ## An Example
 Assume that we need one single entry point to an external system. An actor that receives messages from a JMS queue with the strict requirement that only one JMS consumer must exist to be make sure that the messages are processed in order. That is perhaps not how one would like to design things, but a typical real-world scenario when integrating with external systems.
 
--- a/src/benchmark/RemotePingPong/Program.cs
+++ b/src/benchmark/RemotePingPong/Program.cs
@@ -44,7 +44,7 @@ namespace RemotePingPong
         public static Config CreateActorSystemConfig(string actorSystemName, string ipOrHostname, int port)
         {
             var baseConfig = ConfigurationFactory.ParseString(@"
-                akka {
+            akka {
               actor.provider = remote
               loglevel = ERROR
               suppress-json-serializer-warning = on
@@ -57,6 +57,7 @@ namespace RemotePingPong
                     port = 0
                     hostname = ""localhost""
                 }
+                
               }
             }");
 
--- a/src/core/Akka.Remote/Configuration/Remote.conf
+++ b/src/core/Akka.Remote/Configuration/Remote.conf
@@ -578,23 +578,20 @@ akka {
 
     ### Default dispatcher for the remoting subsystem
 
-    ### Default dispatcher for the remoting subsystem
-
     default-remote-dispatcher {
-      type = ForkJoinDispatcher
-	  executor = fork-join-executor
-      dedicated-thread-pool {
-        # Fixed number of threads to have in this threadpool
-        thread-count = 4
+	    executor = fork-join-executor
+      fork-join-executor {
+        parallelism-min = 2
+        parallelism-factor = 0.5
+        parallelism-max = 16
       }
     }
     
     backoff-remote-dispatcher {
-      type = ForkJoinDispatcher
-	  executor = fork-join-executor
-      dedicated-thread-pool {
-        # Fixed number of threads to have in this threadpool
-        thread-count = 4
+	    executor = fork-join-executor
+      fork-join-executor {
+        parallelism-min = 2
+        parallelism-max = 2
       }
     }
   }
--- a/src/core/Akka/Actor/ActorSystem.cs
+++ b/src/core/Akka/Actor/ActorSystem.cs
@@ -278,6 +278,10 @@ namespace Akka.Actor
 
         private static ActorSystem CreateAndStartSystem(string name, Config withFallback, ActorSystemSetup setup)
         {
+            // allows the ThreadPool to scale up / down dynamically
+            // by removing minimum thread count, which in our benchmarks
+            // appears to negatively impact performance
+            ThreadPool.SetMinThreads(0, 0); 
             var system = new ActorSystemImpl(name, withFallback, setup, Option<Props>.None);
             system.Start();
             return system;
--- a/src/core/Akka/Actor/ActorSystem.cs
+++ b/src/core/Akka/Actor/ActorSystem.cs
@@ -93,7 +93,7 @@ namespace Akka.Actor
     }
 
     /// <summary>
-    /// Core boostrap settings for the <see cref="ActorSystem"/>, which can be created using one of the static factory methods
+    /// Core bootstrap settings for the <see cref="ActorSystem"/>, which can be created using one of the static factory methods
     /// on this class.
     /// </summary>
     public sealed class BootstrapSetup : Setup.Setup
--- a/src/core/Akka/Dispatch/AbstractDispatcher.cs
+++ b/src/core/Akka/Dispatch/AbstractDispatcher.cs
@@ -116,6 +116,26 @@ namespace Akka.Dispatch
         public IDispatcherPrerequisites Prerequisites { get; private set; }
     }
 
+    internal sealed class ChannelExecutorConfigurator : ExecutorServiceConfigurator
+    {
+        public ChannelExecutorConfigurator(Config config, IDispatcherPrerequisites prerequisites) : base(config, prerequisites)
+        {
+            var fje = config.GetConfig("fork-join-executor");
+            MaxParallelism = ThreadPoolConfig.ScaledPoolSize(
+                        fje.GetInt("parallelism-min"), 
+                        fje.GetDouble("parallelism-factor", 1.0D), // the scalar-based factor to scale the threadpool size to 
+                        fje.GetInt("parallelism-max"));
+        }
+
+        public int MaxParallelism {get;}
+
+        public override ExecutorService Produce(string id)
+        {
+            Prerequisites.EventStream.Publish(new Debug($"ChannelExecutor-[id]", typeof(FixedConcurrencyTaskScheduler), $"Launched Dispatcher [{id}] with MaxParallelism=[{MaxParallelism}]"));
+            return new TaskSchedulerExecutor(id, new FixedConcurrencyTaskScheduler(MaxParallelism));
+        }
+    }
+
     /// <summary>
     /// INTERNAL API
     /// 
@@ -306,6 +326,8 @@ namespace Akka.Dispatch
                     return new CurrentSynchronizationContextExecutorServiceFactory(Config, Prerequisites);
                 case "task-executor":
                     return new DefaultTaskSchedulerExecutorConfigurator(Config, Prerequisites);
+                case "channel-executor":
+                    return new ChannelExecutorConfigurator(Config, Prerequisites);
                 default:
                     Type executorConfiguratorType = Type.GetType(executor);
                     if (executorConfiguratorType == null)
--- a/src/core/Akka/Dispatch/Dispatchers.cs
+++ b/src/core/Akka/Dispatch/Dispatchers.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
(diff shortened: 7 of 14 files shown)


src/benchmark/Akka.Benchmarks/BenchmarkDotNet.Artifacts/results/Akka.Benchmarks.Actor.PingPongBenchmarks-report.html
```diff
@@ -0,0 +1,32 @@
...
+  Job-VZNDHW : .NET Core 3.1.13 (CoreCLR 4.700.21.11102, CoreFX 4.700.21.11602), X64 RyuJIT
+</code></pre>
+<pre><code>IterationCount=3  LaunchCount=3  RunStrategy=Monitoring  
+WarmupCount=3  
+</code></pre>
+
+<table>
+<thead><tr><th>                         Method</th><th>Mean</th><th>Error</th><th>StdDev</th>
+</tr>
+</thead><tbody><tr><td>Actor_ping_pong_single_pair_in_memory</td><td>382.4 ns</td><td>9.51 ns</td><td>5.66 ns</td>
```
Aaronontheweb, line 29: This number looked suspiciously good, so I decided to check it against several other benchmarks - for whatever reason Channels did well here but got absolutely destroyed in the original PingPong benchmark as well as in the NBench suite.

src/benchmark/RemotePingPong/Program.cs
```diff
@@ -57,6 +61,11 @@ public static Config CreateActorSystemConfig(string actorSystemName, string ipOr
                     port = 0
                     hostname = ""localhost""
                 }
+
+                default-remote-dispatcher = {
```
Aaronontheweb, line 65: RemotePingPong numbers running two different `Channel<T>` dispatchers were about half of what they are in `dev`

src/core/Akka/Dispatch/Dispatchers.cs
```diff
@@ -137,6 +138,54 @@ public override void Shutdown()
...
+
+        public ChannelExecutor(string id, TaskScheduler scheduler, int degreeOfParallelism) : base(id)
+        {
+            _channel = Channel.CreateUnbounded<IRunnable>(new UnboundedChannelOptions()
+            {
+                AllowSynchronousContinuations = true, SingleReader = false, SingleWriter = true
+            });
+
+            for (var i = 0; i < degreeOfParallelism; i++)
+                Task.Run(() => ReadChannel(_channel.Reader));
```
Aaronontheweb, line 157: I suspect this is the biggest performance problem, how the `ChannelReader<IRunnable>` actually performs consumption via the `ThreadPool` (or in this case, the default `TaskScheduler`) but this is how a lot of other projects and implementations do it. I suspect their performance requirements aren't nearly as high as ours.

Horusiath: Take a look at how [`UnboundedChannel` is implemented](https://github.com/dotnet/runtime/blob/79ae74f5ca5c8a6fe3a48935e85bd7374959c570/src/libraries/System.Threading.Channels/src/System/Threading/Channels/UnboundedChannel.cs#L15). The basic approach is that it's a `ConcurrentQueue<>` + a deque with lock for awaiting readers subscription. 

If I understand correctly major incentive over just dispatching tasks to threadpool is control over `degreeOfParallelism`.

You could probably just replace channel with concurrent queue with asynchronous version of AutoResetEvent (it can have very simple construction as it's never awaited more than once).

Aaronontheweb: Yeah, the basic goal is to bound the degree of parallelism for multiple workloads in order to prevent starvation of `/system` and `/remote` actors - either the queue has to exist at the level of the OS' thread scheduler (which is what happens today) or it has to exist above it in the form of a virtual work queue. Our working theory is that we're probably better off letting the .NET `ThreadPool` manage the former and for us to apply an abstraction above the latter.

Aaronontheweb:
Made some changes to how we execute the read-side of the `ChannelDispatcher` and it is now significantly faster. Have some more tweaks I want to make, but this is progress.

Aaronontheweb:
So latest results from the benchmarks:

1. PingPong benchmark - need to re-run these, but replacing the default dispatcher and the ForkJoinDispatcher with a `ChannelExecutor` resulted in a performance increase. Looks promising.
2. RemotePingPong benchmark - a clean `dev` branch ran showed about 80-81 active threads running on my machine (16 cores) and an upper limit of 192,000 msg/s. Running with the `ChannelExecutor` showed 50 active threads (bear in mind this is the sum of two `ActorSystem`s running inside the same process for both cases) but peak performance dropped to 164,000 msg/s.

This is a big performance increase from my first attempt at using `System.Threading.Channels` and my goal was to cut down on the number of threads waiting for work in Akka.NET - that appears to be a success so far but it's come at the cost of 20% throughput. Not sure that's worth it. Going to keep tweaking this to see if I can improve performance on it.

Aaronontheweb:
Related: https://github.com/akkadotnet/akka.net/issues/4537

src/core/Akka/Dispatch/Dispatchers.cs
```diff
@@ -143,40 +143,60 @@ internal sealed class ChannelExecutor : ExecutorService
...
                 // suppress exceptions
             }
+            finally
+            {
+                int initialValue, newValue;
+                do
+                {
+                    initialValue = _readers;
+                    newValue = _readers - 1;
+                } while (initialValue != Interlocked.CompareExchange(ref _readers, newValue, initialValue));
```
Horusiath, line 181: `Interlocked.Decement(ref _readers)`?

Aaronontheweb: yeah that should probably be fine for this use case

Aaronontheweb:
Using https://github.com/akkadotnet/akka.net-integration-tests/tree/master/src/ClusterPingPong to measure the efficacy of these changes, running as separate .NET 5 processes on the same host machine. 2 node cluster.

I tried running these benchmarks inside `docker-compose` initially, but the overhead of WSL / docker Vnet / docker itself made it impossible to get a clear signal on the throughput (numbers were pretty similar for both.)

Running on bare metal itself tells a different story though.

## With `TaskScheduler(MaxConcurrency = VirtualCpuCount)`

```
OSVersion:                         Microsoft Windows NT 10.0.19041.0
Node Count:                        2
Active Connections:                1
Msgs sent/received per connection: 200000  (2e5)
Thread count per node:             45

Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms]
       1,        5,       10,  1,200,000,    179,543,    6683.62
       1,       10,       20,  2,200,000,    170,565,   12898.27
       1,       15,       30,  3,200,000,    158,513,   20187.56
       1,       20,       40,  4,200,000,    156,176,   26892.67
       1,       25,       50,  5,200,000,    156,961,   33129.11
       1,       30,       60,  6,200,000,    153,478,   40396.54
```

## With v1.4.18 (`ForkJoinExecutor` and `ThreadPoolExecutor`)

```
OSVersion:                         Microsoft Windows NT 10.0.19041.0
Node Count:                        2
Active Connections:                1
Msgs sent/received per connection: 200000  (2e5)
Thread count per node:             62

Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms]
       1,        5,       10,  1,200,000,    214,524,    5593.78
       1,       10,       20,  2,200,000,    198,502,   11082.99
       1,       15,       30,  3,200,000,    193,198,   16563.26
       1,       20,       40,  4,200,000,    188,137,   22324.13
       1,       25,       50,  5,200,000,    186,125,   27938.12
       1,       30,       60,  6,200,000,    182,807,   33915.48
```

Looks like the `TaskSchedulerExecutor` got dwarfed by our control - need to try using the `ChannelExecutor` next.

Aaronontheweb:
Made one small change:

`ThreadPool.SetMinThreads(0,0)` - this tells .NET to not set a minimum number of I/O and worker threads and instead allows the pool to scale up and down dynamically. 

```
OSVersion:                         Microsoft Windows NT 10.0.19041.0
Node Count:                        2
Active Connections:                1
Msgs sent/received per connection: 200000  (2e5)
Thread count per node:             45

Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms]
       1,        5,       10,  1,200,000,    190,670,    6293.58
       1,       10,       20,  2,200,000,    176,469,   12466.75
       1,       15,       30,  3,200,000,    185,703,   17231.78
       1,       20,       40,  4,200,000,    199,370,   21066.26
       1,       25,       50,  5,200,000,    203,247,   25584.53
       1,       30,       60,  6,200,000,    200,973,   30849.79
```

What a difference that makes! And these numbers were all generated at well below 100% CPU, which means that the bottleneck is now somewhere closer to the I/O pipeline:

image

Aaronontheweb:
Performance numbers using our control for `ThreadPool.SetMinThreads(0,0)` were pretty much the same - including the gradual performance degradation as load increased.

Aaronontheweb:
Trying this out using the `System.Threading.Channels` implementation instead:

```
OSVersion:                         Microsoft Windows NT 10.0.19041.0
Node Count:                        2
Active Connections:                1
Msgs sent/received per connection: 200000  (2e5)
Thread count per node:             45

Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms]
       1,        5,       10,  1,200,000,    178,031,    6740.37
       1,       10,       20,  2,200,000,    170,268,   12920.74
       1,       15,       30,  3,200,000,    167,911,   19057.66
       1,       20,       40,  4,200,000,    177,717,   23632.95
       1,       25,       50,  5,200,000,    175,869,   29567.32
       1,       30,       60,  6,200,000,    193,511,   32039.38
```

Going to try re-running that one more time since the last figure has me scratching my head...

Aaronontheweb:
Looked better the second time around:

```
OSVersion:                         Microsoft Windows NT 10.0.19041.0
Node Count:                        2
Active Connections:                1
Msgs sent/received per connection: 200000  (2e5)
Thread count per node:             45

Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms]
       1,        5,       10,  1,200,000,    194,142,    6181.02
       1,       10,       20,  2,200,000,    176,987,   12430.24
       1,       15,       30,  3,200,000,    179,937,   17783.94
       1,       20,       40,  4,200,000,    192,664,   21799.51
       1,       25,       50,  5,200,000,    189,140,   27492.84
       1,       30,       60,  6,200,000,    189,949,   32640.24
```

Aaronontheweb:
Numbers using the latest https://github.com/akkadotnet/akka.net-integration-tests/tree/master/src/ClusterPingPong

## Docker

### Using v1.4.18

```
seed_1          | OSVersion:                         Unix 5.4.72.2
seed_1          | ProcessorCount:                    16
seed_1          | Node Count:                        2
seed_1          | Active Connections:                1
seed_1          | Msgs sent/received per actor:      200000  (2e5)
seed_1          | Is Server GC:                      True
seed_1          | Thread count per node:             41
seed_1          |
seed_1          | Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms], Threads
seed_1          |        1,        5,             10,    1,000,000,    118,759,    8420.36,      3 -> 57
seed_1          |        1,       10,             20,    2,000,000,    133,184,   15016.80,      1 -> 51
seed_1          |        1,       15,             30,    3,000,000,    135,418,   22153.53,      1 -> 50
seed_1          |        1,       20,             40,    4,000,000,    135,450,   29530.99,      1 -> 50
seed_1          |        1,       25,             50,    5,000,000,    140,668,   35544.44,      2 -> 50
seed_1          |        1,       30,             60,    6,000,000,    133,377,   44985.23,      2 -> 50
```

### Using `ChannelExecutor` in v1.4.19

```
seed_1          | OSVersion:                         Unix 5.4.72.2
seed_1          | ProcessorCount:                    16
seed_1          | Node Count:                        2
seed_1          | Active Connections:                1
seed_1          | Msgs sent/received per actor:      200000  (2e5)
seed_1          | Is Server GC:                      True
seed_1          | Thread count per node:             41
seed_1          |
seed_1          | Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms], Threads
seed_1          |        1,        5,             10,    1,000,000,    111,815,    8943.31,    3 -> 57
seed_1          |        1,       10,             20,    2,000,000,    136,361,   14666.87,    2 -> 51
seed_1          |        1,       15,             30,    3,000,000,    142,152,   21104.12,    2 -> 50
seed_1          |        1,       20,             40,    4,000,000,    141,230,   28322.51,    1 -> 50
seed_1          |        1,       25,             50,    5,000,000,    138,405,   36125.68,    2 -> 50
seed_1          |        1,       30,             60,    6,000,000,    135,683,   44220.70,    2 -> 50
```

Numbers inside Docker look very similar, including the thread counts. But I was able to verify that the `ChannelExecutor` is indeed running.

## Bare Metal

### Using v1.4.18

```
OSVersion:                         Microsoft Windows NT 10.0.19041.0                  
Node Count:                        2                                                  
Active Connections:                1                                                  
Msgs sent/received per actor:      200000  (2e5)                                      
Thread count per node:             61                                                 
                                                                                      
Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms], Threads   
       1,        5,             10,    1,000,000,    174,766,    5721.91,      5 -> 84
       1,       10,             20,    2,000,000,    179,417,   11147.18,      7 -> 83
       1,       15,             30,    3,000,000,    177,977,   16856.02,      3 -> 79
       1,       20,             40,    4,000,000,    181,384,   22052.61,      3 -> 71
       1,       25,             50,    5,000,000,    184,619,   27082.80,      2 -> 70
       1,       30,             60,    6,000,000,    176,577,   33979.39,      2 -> 68
```

### Using `ChannelExecutor` in v1.4.19

```
OSVersion:                         Microsoft Windows NT 10.0.19041.0
Node Count:                        2
Active Connections:                1
Msgs sent/received per actor:      200000  (2e5)
Thread count per node:             44

Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms], Threads
       1,        5,             10,    1,000,000,    141,820,    7051.18,    9 -> 78
       1,       10,             20,    2,000,000,    154,477,   12946.90,    9 -> 78
       1,       15,             30,    3,000,000,    177,678,   16884.40,    6 -> 65
       1,       20,             40,    4,000,000,    193,058,   20719.06,    8 -> 57
       1,       25,             50,    5,000,000,    193,859,   25791.82,    5 -> 54
       1,       30,             60,    6,000,000,    195,050,   30761.33,    5 -> 51
```

Much more visible results here - the `ChannelExecutor` wins in terms of overall performance once it has a chance to tune its `ThreadPool` by round 4 - but it also is producing these numbers using ~40% CPU.

Aaronontheweb:
Going to try this one more time using the `TaskSchedulerExecutor` instead, since that will save us the trouble of having to install another dependency and is largely backwards-compatible with what we have today.

Aaronontheweb:
# `ChannelExecutor` Performance

## Docker

```
seed_1          | OSVersion:                         Unix 5.4.72.2
seed_1          | ProcessorCount:                    16
seed_1          | Node Count:                        2
seed_1          | Active Connections:                1
seed_1          | Msgs sent/received per actor:      200000  (2e5)
seed_1          | Is Server GC:                      True
seed_1          | Thread count per node:             41
seed_1          |
seed_1          | Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms], Threads
seed_1          |        1,        5,             10,    1,000,000,    114,908,    8702.59,    1 -> 57
seed_1          |        1,       10,             20,    2,000,000,    134,613,   14857.36,    1 -> 50
seed_1          |        1,       15,             30,    3,000,000,    140,762,   21312.48,    2 -> 50
seed_1          |        1,       20,             40,    4,000,000,    138,127,   28958.82,    2 -> 50
seed_1          |        1,       25,             50,    5,000,000,    138,427,   36119.92,    2 -> 50
seed_1          |        1,       30,             60,    6,000,000,    138,373,   43360.90,    1 -> 50
```

## Bare Metal

```
OSVersion:                         Microsoft Windows NT 10.0.19041.0
Node Count:                        2
Active Connections:                1
Msgs sent/received per actor:      200000  (2e5)
Thread count per node:             45

Connections, Actors/node, Total [actor], Total [msg], Msgs/sec, Total [ms], Threads
       1,        5,             10,    1,000,000,    154,427,    6475.54,    9 -> 79
       1,       10,             20,    2,000,000,    157,232,   12719.98,   13 -> 79
       1,       15,             30,    3,000,000,    165,067,   18174.39,    6 -> 61
       1,       20,             40,    4,000,000,    188,741,   21192.97,    6 -> 54
       1,       25,             50,    5,000,000,    194,141,   25754.36,    6 -> 54
       1,       30,             60,    6,000,000,    187,075,   32072.54,    8 -> 54
```

Looks about even with the `TaskScheduler` but I think the former's numbers are slightly better...

Aaronontheweb:
Full `ChannelExecutor` source code, even though I'm removing it from this PR in favor of the `FixedConcurrencyTaskScheduler` implementation:

```csharp
internal sealed class ChannelExecutor : ExecutorService
    {
        private Channel<IRunnable> _channel;
        private CancellationTokenSource _cts = new CancellationTokenSource();

        private int _readers = 0;

        public int DegreeOfParallelism { get; }

        public ChannelExecutor(string id, int degreeOfParallelism) : base(id)
        {
            _channel = Channel.CreateUnbounded<IRunnable>(new UnboundedChannelOptions()
            {
                AllowSynchronousContinuations = true,
                SingleReader = false,
                SingleWriter = true
            });

            DegreeOfParallelism = degreeOfParallelism;
        }

        private static WaitCallback Executor = o => ReadChannel((ChannelExecutor)o);

        private static void ReadChannel(ChannelExecutor executor)
        {
            try
            {
                while ( executor._channel.Reader.TryRead(out var runnable))
                {
                    runnable.Run();
                }
            }
            catch
            {
                // suppress exceptions
            }
            finally
            {
                Interlocked.Decrement(ref  executor._readers);
            }
        }

        public override void Execute(IRunnable run)
        {
            if (_channel.Writer.TryWrite(run))
            {                
                if (_readers < DegreeOfParallelism)
                {
                    var initial = _readers;
                    var newVale = _readers + 1;
                    if (initial == Interlocked.CompareExchange(ref _readers, newVale, initial))
                    {
                        // try to start a new worker
                        ThreadPool.UnsafeQueueUserWorkItem(Executor, this);
                    }
                }
            }
        }

        public override void Shutdown()
        {
            _channel.Writer.Complete();
            _cts.Cancel();
        }
    }
```

Aaronontheweb:
Writing up some documentation for this now, but once that's done this should be good to go. Going to make this an opt-in feature as part of Akka.NET v1.4.19.

src/core/Akka.Remote/Configuration/Remote.conf
```diff
@@ -578,23 +578,20 @@ akka {
...
-    ### Default dispatcher for the remoting subsystem
-
     default-remote-dispatcher {
-      type = ForkJoinDispatcher
-	  executor = fork-join-executor
-      dedicated-thread-pool {
-        # Fixed number of threads to have in this threadpool
-        thread-count = 4
+	    executor = fork-join-executor
+      fork-join-executor {
```
Aaronontheweb, line 583: This is really the right configuration, rather than a hard thread count of 4.

src/core/Akka/Actor/ActorSystem.cs
```diff
@@ -278,6 +278,10 @@ public static ActorSystem Create(string name)
 
         private static ActorSystem CreateAndStartSystem(string name, Config withFallback, ActorSystemSetup setup)
         {
+            // allows the ThreadPool to scale up / down dynamically
+            // by removing minimum thread count, which in our benchmarks
+            // appears to negatively impact performance
+            ThreadPool.SetMinThreads(0, 0);
```
Aaronontheweb, line 284: The differences in the benchmarks with this and without it are massive. However, if this causes problems for users they can easily reset it by calling `ThreadPool.SetMinThreads(yourValue, yourValue)` after the `ActorSystem` is created. I doubt many users will need to do that and changing this default likely works to the benefit of most.

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -116,6 +116,26 @@ protected ExecutorServiceConfigurator(Config config, IDispatcherPrerequisites pr
         public IDispatcherPrerequisites Prerequisites { get; private set; }
     }
 
+    internal sealed class ChannelExecutorConfigurator : ExecutorServiceConfigurator
```
Aaronontheweb, line 119: Used to configure the `TaskSchedulerExecutor(id, new FixedConcurrencyTaskScheduler(MaxParallelism));`

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -306,6 +326,8 @@ protected ExecutorServiceConfigurator ConfigureExecutor()
                     return new CurrentSynchronizationContextExecutorServiceFactory(Config, Prerequisites);
                 case "task-executor":
                     return new DefaultTaskSchedulerExecutorConfigurator(Config, Prerequisites);
+                case "channel-executor":
+                    return new ChannelExecutorConfigurator(Config, Prerequisites);
```
Aaronontheweb, line 330: This is how we actually pass the new executor into dispatcher configurations.

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -116,6 +116,26 @@ protected ExecutorServiceConfigurator(Config config, IDispatcherPrerequisites pr
...
+                        fje.GetInt("parallelism-min"), 
+                        fje.GetDouble("parallelism-factor", 1.0D), // the scalar-based factor to scale the threadpool size to 
+                        fje.GetInt("parallelism-max"));
+        }
+
+        public int MaxParallelism {get;}
+
+        public override ExecutorService Produce(string id)
+        {
+            Prerequisites.EventStream.Publish(new Debug($"ChannelExecutor-[id]", typeof(FixedConcurrencyTaskScheduler), $"Launched Dispatcher [{id}] with MaxParallelism=[{MaxParallelism}]"));
```
Aaronontheweb, line 134: `Debug` event you can listen for if you want to be certain that the `ChannelExecutor` is loaded.

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -116,6 +116,26 @@ protected ExecutorServiceConfigurator(Config config, IDispatcherPrerequisites pr
         public IDispatcherPrerequisites Prerequisites { get; private set; }
     }
 
+    internal sealed class ChannelExecutorConfigurator : ExecutorServiceConfigurator
+    {
+        public ChannelExecutorConfigurator(Config config, IDispatcherPrerequisites prerequisites) : base(config, prerequisites)
+        {
+            var fje = config.GetConfig("fork-join-executor");
+            MaxParallelism = ThreadPoolConfig.ScaledPoolSize(
```
Aaronontheweb, line 124: Re-use `ForkJoinDispatcher` configuration block, since it expresses what we need.

docs/articles/clustering/cluster-singleton.md
```diff
@@ -34,9 +34,6 @@ This pattern may seem to be very tempting to use at first, but it has several dr
 
 Especially the last point is something you should be aware of — in general when using the Cluster Singleton pattern you should take care of downing nodes yourself and not rely on the timing based auto-down feature.
 
-> [!WARNING]
```
Aaronontheweb, line 37: Whoops - this was actually from a previous PR.

Aaronontheweb (commented):
Described my changes.

src/core/Akka/Dispatch/Dispatchers.cs
Aaronontheweb: Made this change

Arkatufus (changes requested):
Code looks good, just a few changes needed in the documentation

docs/articles/actors/dispatchers.md
```diff
@@ -165,6 +166,63 @@ private void Form1_Load(object sender, System.EventArgs e)
...
+    fork-join-executor {
+      parallelism-min = 4
+      parallelism-factor = 1.0
+      parallelism-max = 64
+    }
+}
+
+akka.remote.default-remote-dispatcher {
+    type = Dispatcher
+	  executor = channel-executor
```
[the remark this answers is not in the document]
Aaronontheweb: Looks like VS Code mixed tabs and spaces there.

Aaronontheweb:
> We might actually need to copy HOCON lines over and make them part of the executor config in the future, because having one part of the system depending on a config of another part can be a point of confusion for the end user

Agree, but not going to do that today since this is meant to be a drop-in replacement for `ForkJoinDispatcher`. Goal is to eventually purge `DedicatedThreadPool` out of the code base for good.

docs/articles/actors/dispatchers.md
```diff
@@ -165,6 +166,63 @@ private void Form1_Load(object sender, System.EventArgs e)
 }
 ```
 
+### `ChannelExecutor`
+In Akka.NET v1.4.19 we will be introducing an opt-in feature, the `ChannelExecutor` - a new dispatcher type that re-uses the same configuration as a `ForkJoinDispatcher` but runs entirely on top of the .NET `ThreadPool` and is able to take advantage of dynamic thread pool scaling to size / resize workloads on the fly.
+
+During its initial development and benchmarks, we observed the following:
+
+1. The `ChannelExecutor` tremendously reduced idle CPU and max busy CPU even during peak message throughput, primarily as a result of dynamically shrinking the total `ThreadPool` to only the necessary size. This resolves one of the largest complaints large users of Akka.NET have today. However, **in order for this setting to be effective `ThreadPool.SetMin(0,0)` must also be set**. We are considering doing this inside the `ActorSystem.Create` method, but if you those settings don't work for you you can easily override them by simply calling `ThreadPool.SetMin(yourValue, yourValue)` again after `ActorSystem.Create` has exited.
```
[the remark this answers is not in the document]
Aaronontheweb: Done

docs/articles/actors/dispatchers.md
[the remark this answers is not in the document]
Aaronontheweb: Fixed

Aaronontheweb:
> > We might actually need to copy HOCON lines over and make them part of the executor config in the future, because having one part of the system depending on a config of another part can be a point of confusion for the end user
> 
> Agree, but not going to do that today since this is meant to be a drop-in replacement for `ForkJoinDispatcher`. Goal is to eventually purge `DedicatedThreadPool` out of the code base for good.

Should also point out, this is a sub-config of the current HOCON node anyway

Arkatufus (approved):

Aaronontheweb:
`ThreadPool.SetMinThreads(0,0)` definitely skewed some of the tests inside the `TestKit` - going to need to take a look at that.

Zetanova:
I would not recommend to set `ThreadPool.SetMinThreads(0,0)` or do any global ThreadPool settings.
Why set both limits work and IO threads?

I don't realy understand the effect that would have other then the ThreadPool schedule algo would be used instantly
and the first messages/IRunnable would be a bit delayed.

An idea would be first to add a totalReaders counter to count how may readers got called over the test-run. 
Less would be better.  I believe only to set `ThreadPool.SetMinThreads(0,0)`  would have the effect to reduce the totalReaders count.
And other counters  to count worker executions that got 0, 1 or many work-items.
Counters: `totalReaders, zeroReader, singleReaders`

To have the same effect as `ThreadPool.SetMinThreads(0,0)` a wrapped IRunnable could be used to start the next worker from one of the currently running wrokers.

Aaronontheweb:
> I would not recommend to set ThreadPool.SetMinThreads(0,0) or do any global ThreadPool settings.
> Why set both limits work and IO threads?

This setting removes the lower bound on the number of minimum threads - now there are no limits, as the lower bound is zero. The previous value was `Environment.ProcessorCount`, which as our benchmarks showed was probably more than even most busy systems need.

In the event of an extremely busy system, this change probably won't have much impact as those threads will continue to be used. The one downside might be that a bunch of threads might be killed and recreated right away at startup, and Ben Adams pointed out some potential memory overhead from work-stealing that might be incurred by this change too: https://twitter.com/ben_a_adams/status/1387236911018819596

If the setting has adverse affects for a lot of our users, we'll roll it back and make it toggleable - but changing that setting improved all of the Akka.NET v1.4.18 defaults too.

> An idea would be first to add a totalReaders counter to count how may readers got called over the test-run.
Less would be better. I believe only to set ThreadPool.SetMinThreads(0,0) would have the effect to reduce the totalReaders count.
> And other counters to count worker executions that got 0, 1 or many work-items.
> Counters: totalReaders, zeroReader, singleReaders

We'd need to keep track of how busy all of the current readers are and factor that in too, the time component. But I'm open to auto-tuning our use of workers on the thread pool as well. Definitely interested in doing more work in that area.

Zetanova:
My proposed coutners would only for perf messurement.
I believe that the new work item will get often "stolen" by an other worker 
and the queued up worker would run empty. `zeroReaders++`

Most likely the effect of `ThreadPool.SetMinThreads(0,0)` would be that 
worker-1, worker-2, worker-3 would be queued in the ThreadPool queue
and worker-0 would process all work-items more or less by himself.

Aaronontheweb:
@Zetanova I'd be very interested in seeing some experiments with this construct from contributors like you - your original `Channel<T>` implementation was one of the inspirations for this!

Zetanova:
@Aaronontheweb 
I will do one and test the current one at the weekend.

The refactor of the `HashWheelTaskScheduler` is on my todo list too.

Aaronontheweb:
> The refactor of the HashWheelTaskScheduler is on my todo list too. 

That's another good one - as @to11mtm often says, adding support for "Schedule with fixed delay" is something we should aim for by the time Akka.NET v1.5 is released.

Zetanova:
@Aaronontheweb  I didn't used the new dotnet counters for now, but would it be not an idea to use implement them here 
https://docs.microsoft.com/en-gb/dotnet/core/diagnostics/dotnet-counters
Of course about the overhead we would need to be very careful.  

The benefit would be that somebody could exec the dotnet counter output on production / kubernetes cluster 
without a spezial build

PR #4887, opened 2021-03-30
Fix ClusterMetricsExtensionSpec racy spec

### LINKED ISSUE #4267: Akka.Cluster.Metrics: ClusterMetricsExtensionSpec is racy
Version: Akka.NET v1.4.0

The following spec fails from time to time:

https://github.com/akkadotnet/akka.net/blob/c97e5b6820d33970c2ce06d15fee77d8273553eb/src/contrib/cluster/Akka.Cluster.Metrics.Tests/ClusterMetricsExtensionSpec.cs#L67-L104

Issue is that waiting for the collector to run within a specified window of time isn't always reliable within the configured timeframes. This test was disabled as part of https://github.com/akkadotnet/akka.net/pull/4257

IgorFedchenko:
Close #4267


IgorFedchenko:
Made this spec much more simple and less sensible to timing.

PR #4893, opened 2021-04-01
Handle CoordinatedShutdown exiting-completed when not joined

zbynek001:
Handle CoordinatedShutdown exiting-completed when not joined
* assertion failed: Nodes not part of cluster have marked the Gossip as seen
* trying to mark the Gossip as seen before it has joined, which may happen
  if CoordinatedShutdown is running before the node has joined

migrated from https://github.com/akka/akka/pull/26835


Aaronontheweb:
No idea what the .NET Core Tests on Linux failed - looked like a build failure but there wasn't any error message at all. Re-running it.

Aaronontheweb (approved):
LGTM

PR #4895, opened 2021-04-02
Add retry pattern with delay calculation support

ismaelhamed:
Ported `Retry` pattern that will retry a `Task` some number of times with a delay between each attempt.


Danthar:
Do we really need something like this in the core codebase when there are libraries like Polly ?
What problem are we solving here ?

ismaelhamed:
You don't have to take a dependency on yet another library just to simply retrying a Task. I've been asked at work why is there a `RetryFlow` but not a `Retry` pattern to use in plain actors. But [there is](https://doc.akka.io/docs/akka/current/futures.html#retry).

https://github.com/akka/akka/issues/22276

Aaronontheweb (approved):
LGTM

PR #4896, opened 2021-04-02
Added DoNotInherit annotation

### LINKED ISSUE #4883: Missing DoNotInherit annotation
Is there a reason why `DoNotInherit` is missing from the annotations? I can see it all over the place in the JVM. Would a PR be fine?

```csharp
/// <summary>
/// <para>
/// Marks APIs that are designed under a closed-world assumption for and are NOT meant to be
/// extended by user-code. It is fine to extend these classes within Akka itself however.
/// </para>
/// <para>
/// This is most useful for binary compatibility purposes when a set of classes and interfaces
/// assume a "closed world" between them, and gain the ability to add methods to the interfaces
/// without breaking binary compatibility for users of this code. Specifically this assumption may be
/// understood intuitively: as all classes that implement this interface are in this compilation unit
/// artifact, it is impossible to obtain a "old" class with a "new" interface, as they are part of
/// the same dependency.
/// </para>
/// </summary>
[AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface)]
public sealed class DoNotInheritAttribute: Attribute
{
}
```

ismaelhamed:
Closes #4883

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -2099,6 +2099,11 @@ namespace Akka.Annotations
     {
         public ApiMayChangeAttribute() { }
     }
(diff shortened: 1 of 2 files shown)


Aaronontheweb (approved):
LGTM

PR #4902, opened 2021-04-06
replace reflection magic in MNTR with reading of `MultiNodeConfig` properties

### LINKED ISSUE #4901: MNTR: needs to discover roles from Config, not reflection
Looks like I found a small bug in the MNTR here:

```
private IEnumerable<RoleName> RoleNames(Type specType)
        {
            var ctorWithConfig = FindConfigConstructor(specType);
            var configType = ctorWithConfig.GetParameters().First().ParameterType;
            var args = ConfigConstructorParamValues(configType);
            var configInstance = Activator.CreateInstance(configType, args);
            var roleType = typeof(RoleName);
            var configProps = configType.GetProperties(BindingFlags.Instance | BindingFlags.Public);
            var roleProps = configProps.Where(p => p.PropertyType == roleType && p.Name != "Myself").Select(p => (RoleName)p.GetValue(configInstance));
            var configFields = configType.GetFields(BindingFlags.Instance | BindingFlags.Public);
            var roleFields = configFields.Where(f => f.FieldType == roleType && f.Name != "Myself").Select(f => (RoleName)f.GetValue(configInstance));
            var roles = roleProps.Concat(roleFields).Distinct();
            return roles;
        }
```

This can only find `RoleName`s that have been declared as fields on an MNTR config - any code that dynamically returns a number of Roles, such as the `StressSpec`, will not run. I'll have to patch this also.

_Originally posted by @Aaronontheweb in https://github.com/akkadotnet/akka.net/issues/4899#issuecomment-813778565_

Aaronontheweb:
close #4901

### DIFF
--- a/src/core/Akka.MultiNodeTestRunner/Discovery.cs
+++ b/src/core/Akka.MultiNodeTestRunner/Discovery.cs
@@ -108,14 +108,8 @@ namespace Akka.MultiNodeTestRunner
             var ctorWithConfig = FindConfigConstructor(specType);
             var configType = ctorWithConfig.GetParameters().First().ParameterType;
             var args = ConfigConstructorParamValues(configType);
-            var configInstance = Activator.CreateInstance(configType, args);
-            var roleType = typeof(RoleName);
-            var configProps = configType.GetProperties(BindingFlags.Instance | BindingFlags.Public);
-            var roleProps = configProps.Where(p => p.PropertyType == roleType && p.Name != "Myself").Select(p => (RoleName)p.GetValue(configInstance));
-            var configFields = configType.GetFields(BindingFlags.Instance | BindingFlags.Public);
-            var roleFields = configFields.Where(f => f.FieldType == roleType && f.Name != "Myself").Select(f => (RoleName)f.GetValue(configInstance));
-            var roles = roleProps.Concat(roleFields).Distinct();
-            return roles;
+            var configInstance = (MultiNodeConfig)Activator.CreateInstance(configType, args);
+            return configInstance.Roles;
         }
 
         internal static ConstructorInfo FindConfigConstructor(Type configUser)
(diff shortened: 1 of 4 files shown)


src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -168,7 +168,7 @@ internal ImmutableList<string> Deployments(RoleName node)
             return deployments == null ? _allDeploy : deployments.AddRange(_allDeploy);
         }
 
-        internal ImmutableList<RoleName> Roles
+        public ImmutableList<RoleName> Roles
```
Aaronontheweb, line 171: Made this property public so the data is just available to anyone who needs it - preferred this over friend assembly magic.

src/core/Akka.MultiNodeTestRunner/Discovery.cs
```diff
@@ -108,14 +108,8 @@ private IEnumerable<RoleName> RoleNames(Type specType)
             var ctorWithConfig = FindConfigConstructor(specType);
             var configType = ctorWithConfig.GetParameters().First().ParameterType;
             var args = ConfigConstructorParamValues(configType);
-            var configInstance = Activator.CreateInstance(configType, args);
-            var roleType = typeof(RoleName);
-            var configProps = configType.GetProperties(BindingFlags.Instance | BindingFlags.Public);
```
Aaronontheweb, line 113: Don't believe we need the property checking anymore using reflection - a full MNTR run (which this PR will trigger) will show up with many fewer tests and much lower execution time if this change made several tests suddenly stop showing up.

Aaronontheweb:
Looks like this had no negative impact on MNTR test discoverability - I didn't expect that it would.

PR #4908, opened 2021-04-06
Fix/standardout logger format error

Aaronontheweb:
This `System.FormatException` can be thrown when a bad format string is passed into the `StandardOutLogger.PrintLog` method - ironically enough, it was our `catch` block for intercepting this same error that re-threw the exception.

**Issue is still not solved in this PR**

Pretty sure I found the bag log message that triggers this here:

https://github.com/akkadotnet/akka.net/blob/e7b3fb2956eafa867cfb9e2eb7ac5cc6a87e94e8/src/core/Akka.Cluster/SBR/SplitBrainResolver.cs#L179-L184

Missed the number 2 index on the format string.

### DIFF
--- a/src/core/Akka/Event/StandardOutLogger.cs
+++ b/src/core/Akka/Event/StandardOutLogger.cs
@@ -138,15 +138,15 @@ namespace Akka.Event
             catch (FormatException)
             {
                 /*
-                 * If we've reached this point, the `logEvent` itself is informatted incorrectly. 
+                 * If we've reached this point, the `logEvent` itself is formatted incorrectly. 
                  * Therefore we have to treat the data inside the `logEvent` as suspicious and avoid throwing
                  * a second FormatException.
                  */
                 var sb = new StringBuilder();
                 sb.AppendFormat("[ERROR][{0}][Thread {1}][StandardOutLogger] ", logEvent.Timestamp, 0);
-                sb.AppendFormat("Encoutered System.FormatException while recording log: [" +
+                sb.AppendFormat("Encountered System.FormatException while recording log: [" +
                                 logEvent.LogLevel().PrettyNameFor() + "]")
-                    .AppendFormat("[" + logEvent.LogSource + "][" + logEvent.Message + "]");
+                    .Append("[" + logEvent.LogSource + "][" + logEvent.Message + "]");
 
                 string msg;
                 switch (logEvent.Message)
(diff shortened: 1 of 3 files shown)


src/core/Akka/Event/StandardOutLogger.cs
```diff
@@ -138,15 +138,15 @@ public static void PrintLogEvent(LogEvent logEvent)
...
                  * Therefore we have to treat the data inside the `logEvent` as suspicious and avoid throwing
                  * a second FormatException.
                  */
                 var sb = new StringBuilder();
                 sb.AppendFormat("[ERROR][{0}][Thread {1}][StandardOutLogger] ", logEvent.Timestamp, 0);
-                sb.AppendFormat("Encoutered System.FormatException while recording log: [" +
+                sb.AppendFormat("Encountered System.FormatException while recording log: [" +
                                 logEvent.LogLevel().PrettyNameFor() + "]")
-                    .AppendFormat("[" + logEvent.LogSource + "][" + logEvent.Message + "]");
+                    .Append("[" + logEvent.LogSource + "][" + logEvent.Message + "]");
```
Aaronontheweb, line 149: I thought this was the fix, but sadly no

src/core/Akka/Event/StandardOutLogger.cs
```diff
@@ -138,15 +138,15 @@ public static void PrintLogEvent(LogEvent logEvent)
...
                 /*
-                 * If we've reached this point, the `logEvent` itself is informatted incorrectly. 
+                 * If we've reached this point, the `logEvent` itself is formatted incorrectly. 
                  * Therefore we have to treat the data inside the `logEvent` as suspicious and avoid throwing
                  * a second FormatException.
                  */
                 var sb = new StringBuilder();
                 sb.AppendFormat("[ERROR][{0}][Thread {1}][StandardOutLogger] ", logEvent.Timestamp, 0);
-                sb.AppendFormat("Encoutered System.FormatException while recording log: [" +
+                sb.AppendFormat("Encountered System.FormatException while recording log: [" +
```
Aaronontheweb, line 147: Wonder if we need to make that change here too?

src/core/Akka/Event/StandardOutLogger.cs
Arkatufus: `logEvent.Message` is the culprit

Arkatufus (changes requested):

Issue #4843, opened 2021-03-11
ORDictionary<string, POC> missing items


tesgu:
Hi guys

We are experiencing a weird behavior with ORDictionary. Before adding the new item to the dictionary we log the contents of it and, obviously the item doesn't exist. We add the item to the dictionary with WriteLocal consistency and the AddOrUpdate method. Before exiting the Merge function in the update message to the replicator, we log the items in the dictionary and the item exists. Sometimes, when we try to get the item from the dictionary with ReadLocal in the same machine or in the other machines, the item doesn't exist. Nobody has actively remove it, but the item doesn't exist. Item keys are completely unique (guids) and there is only one AddOrUpdate for all the machines... we trust gossip to propagate the value to the other machines.

This happens in a cluster with three machines that are adding and removing items from the dictionary. Not sure what is happening.

As I told we are adding new items with AddOrUpdate, removing them with Remove and getting the items with TryGetValue.

The POC is also derived from IReplicated data. It has a creation date and the Merge method always returns the newest item. If it merges with null, the non null value is returned.

Not sure what is happening... I would think that when trying to get from other machines maybe the gossip didn't work and can not find the value, but when trying to get it in the local machine where the AddOrUpdate was executed... I think it should exist if nobody has deleted it (and nobody did at least actively)

This is our code for the POC merge:
```
        public TransactionState Merge(TransactionState other) =>
            other != null && other.Created > this.Created ? other : this;

        /// <inheritdoc />
        public IReplicatedData Merge(IReplicatedData other) =>
            other is TransactionState state ? this.Merge(state) : this;
```

Any clue about what could be happening?

Arkatufus:
@tesgu I'll be looking into this shortly

Arkatufus:
@tesgu 
Can try to subscribe to the local `Replicator` and check that you never receive any `DataDeleted` message from your `Replicator`?
Can you also turn Debug logging on and check that the number of "Received update for deleted key {0}" messages matches the number of received `DataDeleted` message count?

tesgu:
Sorry @Arkatufus . Maybe we are using distributed data in a way it's not intended to because we are not receiving any DataDeleted message from the replicator as we are not deleting any key. I explain:

- We create an ORDictionaryKey<string, TransactionState> that contains an ORDictionary<string, TransactionState>.
- When we wat to add a TransactionState to the ORDictionary we use AddOrUpdate method in the Merge method passed in the Update message for the replicator
```
[...]

PR #4910, opened 2021-04-06
ORDictionary with POCO value missing items, add MultiNode spec

Arkatufus:
Related to #4843
- Add a multinode spec for ORDictionary<string, PocoObject>
- Fix problems with DData cluster killing itself because of excessive deadlettering of unprocessable messages


Aaronontheweb:
@Arkatufus can you do a self-review on this and point out where the significant changes are and why they are there?

src/contrib/cluster/Akka.DistributedData/Replicator.cs
```diff
@@ -479,35 +486,91 @@ private void Load()
...
+            {
+                if(!NormalReceive(msg))
+                    Unhandled(msg);
+            });
+
+            ReceiveAny(_ => Stash.Stash());
         }
 
-        private void NormalReceive()
+        private bool NormalReceive(object message)
```
Arkatufus, line 499: This method is practically the same, needed to change it to untyped receive so that we can call it from somewhere else and checks if it handled the message or not (partial function in scala)

src/contrib/cluster/Akka.DistributedData/Replicator.cs
```diff
@@ -479,35 +486,91 @@ private void Load()
...
-            Receive<ClusterEvent.UnreachableMember>(u => ReceiveUnreachable(u.Member));
-            Receive<ClusterEvent.ReachableMember>(r => ReceiveReachable(r.Member));
-            Receive<GetKeyIds>(_ => ReceiveGetKeyIds());
-            Receive<Delete>(d => ReceiveDelete(d.Key, d.Consistency, d.Request));
-            Receive<RemovedNodePruningTick>(r => ReceiveRemovedNodePruningTick());
-            Receive<GetReplicaCount>(_ => ReceiveGetReplicaCount());
+            switch (message)
+            {
+                case IDestinationSystemUid msg:
+                    if (msg.ToSystemUid.HasValue && msg.ToSystemUid != _selfUniqueAddress.Uid)
```
Arkatufus, line 504: Special case for `Status` and `Gossip` messages, we filter out messages that are not intended for this specific node. Processing these messages can poison the whole network with unprocessable messages that ends up being dead lettered.

src/contrib/cluster/Akka.DistributedData/Replicator.cs
```diff
@@ -479,35 +486,91 @@ private void Load()
...
+                            case Status s:
+                                ReceiveStatus(s.Digests, s.Chunk, s.TotalChunks); return true;
+                            case Gossip g : 
+                                ReceiveGossip(g.UpdatedData, g.SendBack); return true;
+                        }
+                    }
+                    return true;
+
+                case ISendingSystemUid msg:
+                    if (msg.FromNode != null && !IsKnownNode(msg.FromNode.Address))
```
Arkatufus, line 529: Special case for `Read`, `Write`, and `DeltaPropagation`, we filter out messages from all unknown nodes. Reason is the same as above.

src/contrib/cluster/Akka.DistributedData/Replicator.cs
```diff
@@ -441,8 +449,6 @@ private void Load()
             var startTime = DateTime.UtcNow;
             var count = 0;
 
-            NormalReceive();
```
Arkatufus, line 444: Instead of letting the node to process everything while loading, we will stash messages instead. The only things that are processed normally during load are cluster domain events.

src/contrib/cluster/Akka.DistributedData/Replicator.cs
```diff
@@ -461,6 +467,7 @@ private void Load()
                 _log.Debug("Loading {0} entries from durable store took {1} ms", count,
                     (DateTime.UtcNow - startTime).TotalMilliseconds);
                 Become(NormalReceive);
+                Stash.UnstashAll();
```
Arkatufus, line 470: We unstash everything when data has been loaded.

src/contrib/cluster/Akka.DistributedData/ReplicatorSettings.cs
```diff
@@ -52,35 +52,32 @@ public static ReplicatorSettings Create(Config config)
...
             }
 
-            // TODO: This constructor call fails when these fields are not populated inside the Config object:
-            // TODO: `pruning-marker-time-to-live` key depends on Config.GetTimeSpan() to return a TimeSpan.Zero default.
             return new ReplicatorSettings(
-                role: config.GetString("role"),
-                gossipInterval: config.GetTimeSpan("gossip-interval"),
-                notifySubscribersInterval: config.GetTimeSpan("notify-subscribers-interval"),
-                maxDeltaElements: config.GetInt("max-delta-elements"),
+                role: config.GetString("role", string.Empty),
```
Arkatufus, line 69: These are just code clean-up, while providing default values for most of the settings.

src/contrib/cluster/Akka.DistributedData/WriteAggregator.cs
```diff
@@ -87,50 +89,58 @@ protected override void PreStart()
...
-                if (IsDone) Reply(isTimeout: false);
-            })
-            .With<StoreFailure>(x =>
-            {
-                _gotLocalStoreReply = true;
-                _gotNackFrom = _gotNackFrom.Remove(_selfUniqueAddress.Address);
-                if (IsDone) Reply(isTimeout: false);
-            })
-            .With<SendToSecondary>(x =>
+        protected override bool Receive(object message)
```
Arkatufus, line 92: Practically the same, modernizing Match()...With() with a switch instead.

src/contrib/cluster/Akka.DistributedData/Replicator.cs
Aaronontheweb: Nice catch - good to know.

src/contrib/cluster/Akka.DistributedData/Replicator.cs
Aaronontheweb: And that works even with a `ReceiveActor` ? Never tried that before.

src/contrib/cluster/Akka.DistributedData/Replicator.cs
```diff
@@ -479,35 +486,91 @@ private void Load()
...
             Receive(new Action<Status>(IgnoreDebug));
             Receive(new Action<Gossip>(IgnoreDebug));
+
+            Receive<ClusterEvent.IClusterDomainEvent>(msg =>
+            {
+                if(!NormalReceive(msg))
+                    Unhandled(msg);
+            });
+
+            ReceiveAny(_ => Stash.Stash());
```
Aaronontheweb, line 496: LGTM

src/contrib/cluster/Akka.DistributedData/WriteAggregator.cs
Aaronontheweb: LGTM

Aaronontheweb (approved):
Overall LGTM

src/contrib/cluster/Akka.DistributedData/Replicator.cs
Arkatufus: Good point, I should probably change the base class to `UntypedActor` instead.

Aaronontheweb: Sounds good to me

Arkatufus:
With all the modifications, it seemed like POCO are serialized properly now, but I really have no actual reproduction spec for the original issue.

src/contrib/cluster/Akka.DistributedData/Internal/Internal.cs
```diff
@@ -878,7 +888,7 @@ private sealed class NoDelta : IDeltaReplicatedData<IReplicatedData, IReplicated
         /// </summary>
         public static readonly IReplicatedDelta NoDeltaPlaceholder = NoDelta.Instance;
 
-        public readonly UniqueAddress FromNode;
+        public UniqueAddress FromNode { get; }
```
Aaronontheweb, line 891: Why is this a property but everything else is a field?

Arkatufus: Because of the new interface `ISendingSystemUid`

Aaronontheweb: Why not make everything properties then? Just for consistency.

Arkatufus: Mainly to keep the change to the API and possible impact from code change as minimum as possible, should I change them all to properties?

Aaronontheweb: Yeah I would - if we're going to break binary compatibility (actually this may not be binary-incompatible since the return types, member names, are the same) we might as well clean it up. We use properties everywhere inside most message types.

Arkatufus: Alright, I'll do a code clean-up then

Arkatufus: Done

Aaronontheweb (approved):
LGTM

PR #4912, opened 2021-04-06
cleaned up the SpecPass / SpecFail messages

Aaronontheweb:
more cleanup on the `SpecPass` and `SpecFail` message types - removed unnecessary hiding and some of the message procssing inside `Sink`

### DIFF
--- a/src/core/Akka.MultiNodeTestRunner.Shared/Sinks/Spec.cs
+++ b/src/core/Akka.MultiNodeTestRunner.Shared/Sinks/Spec.cs
@@ -59,14 +59,11 @@ namespace Akka.MultiNodeTestRunner.Shared.Sinks
             FailureMessages = new List<string>();
             FailureStackTraces = new List<string>();
             FailureExceptionTypes = new List<string>();
-            Timestamp = DateTime.UtcNow;
         }
 
         public IList<string> FailureMessages { get; private set; }
         public IList<string> FailureStackTraces { get; private set; }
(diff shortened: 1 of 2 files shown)


src/core/Akka.MultiNodeTestRunner.Shared/Sinks/Spec.cs
```diff
@@ -59,15 +59,12 @@ public SpecFail(int nodeIndex, string nodeRole, string testDisplayName) : base(n
             FailureMessages = new List<string>();
             FailureStackTraces = new List<string>();
             FailureExceptionTypes = new List<string>();
-            Timestamp = DateTime.UtcNow;
```
Aaronontheweb, line 62: This was hiding the `Timestamp` property I later added to `SpecPass`, which is the base class of `SpecFail` - I added them the other way around the first time, so I wanted to clean that up here.

src/core/Akka.NodeTestRunner/Sink.cs
```diff
@@ -40,47 +40,47 @@ public Sink(int nodeIndex, string nodeRole, IActorRef logger)
 
         public bool OnMessage(IMessageSinkMessage message)
         {
-            var resultMessage = message as ITestResultMessage;
-            if (resultMessage != null)
+            if (message is ITestResultMessage resultMessage)
```
Aaronontheweb, line 43: Resharper'd all of this code to take advantage of newer C# language features

PR #4913, opened 2021-04-07
reduce allocations inside PhiAccrualFailureDetector

Aaronontheweb:
made `HeartbeatHistory` into a `readonly struct` and cleaned up some other old LINQ calls inside the data structure

### DIFF
--- a/src/core/Akka.Remote/PhiAccrualFailureDetector.cs
+++ b/src/core/Akka.Remote/PhiAccrualFailureDetector.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using Akka.Actor;
 using Akka.Configuration;
@@ -35,12 +36,12 @@ namespace Akka.Remote
     /// </summary>
     public class PhiAccrualFailureDetector : FailureDetector
     {
-        private double _threshold;
-        private int _maxSampleSize;
+        private readonly double _threshold;
+        private readonly int _maxSampleSize;
         private TimeSpan _minStdDeviation;
         private TimeSpan _acceptableHeartbeatPause;
         private TimeSpan _firstHeartbeatEstimate;
-        private Clock _clock;
+        private readonly Clock _clock;
 
         /// <summary>
         /// Procedural constructor for PhiAccrualDetector
@@ -182,7 +183,7 @@ namespace Akka.Remote
         {
             var timestamp = _clock();
             var oldState = state;
-            HeartbeatHistory newHistory = null;
+            HeartbeatHistory newHistory;
 
             if (!oldState.TimeStamp.HasValue)
             {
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
This isn't even necessarily a super high-performance area of Akka.NET, but since these history objects are frequently updated for each connection in Akka.Cluster I wanted to move them off of the heap if possible.

src/core/Akka.Remote/PhiAccrualFailureDetector.cs
```diff
@@ -371,7 +354,7 @@ public double StdDeviation
         {
             if (history._intervals.Count < history._maxSampleSize)
             {
-                return new HeartbeatHistory(history._maxSampleSize, history._intervals.Concat(new[] { interval }).ToList(),
+                return new HeartbeatHistory(history._maxSampleSize, history._intervals.Add(interval),
```
Aaronontheweb, line 357: Use Systems.Collections.Immutable here instead of classic collections and LINQ (this is 7-8 year old code)

src/core/Akka.Remote/PhiAccrualFailureDetector.cs
```diff
@@ -275,14 +276,14 @@ internal double Phi(long timeDiff, double mean, double stdDeviation)
...
 
-        private long MinStdDeviationMillis
+        private double MinStdDeviationMillis
         {
-            get { return (long)_minStdDeviation.TotalMilliseconds; }
+            get { return _minStdDeviation.TotalMilliseconds; }
         }
 
-        private long AcceptableHeartbeatPauseMillis
+        private double AcceptableHeartbeatPauseMillis
```
Aaronontheweb, line 284: Remove boxing back to `long` in order to prevent loss of precision.

Aaronontheweb: (may not have been an issue anyway)

src/core/Akka.Remote/PhiAccrualFailureDetector.cs
```diff
@@ -382,7 +365,8 @@ public double StdDeviation
 
         private static HeartbeatHistory DropOldest(HeartbeatHistory history)
         {
-            return new HeartbeatHistory(history._maxSampleSize, history._intervals.Skip(1).ToList(), history._intervalSum - history._intervals.First(), history._squaredIntervalSum - Pow2(history._intervals.First()));
+            return new HeartbeatHistory(history._maxSampleSize, history._intervals.RemoveAt(0), history._intervalSum - history._intervals.First(),
```
Arkatufus, line 368: This looks a bit suspect, assuming that code inside the parameters are executed in order from left to right, you removed the first item on the list, and then subtract the sum with the first item on the list. This is, of course, an assumption, but we can't really be sure of the code execution order, that really depends on the compiler and code optimizer.

The sum should be subtracted by the item removed from the list, not the first item on the list, because we don't really know which got executed first.

Arkatufus (changes requested):

Arkatufus (approved):
LGTM

PR #4917, opened 2021-04-07
cleaned up RemoteWatcher

Aaronontheweb:
This PR doesn't make any substantive changes to the `RemoteWatcher`, but it cleans up the code using newer C# language constructs and eliminates some very awkward conditional branching that could be simplified.

### DIFF
--- a/src/core/Akka.Remote/RemoteWatcher.cs
+++ b/src/core/Akka.Remote/RemoteWatcher.cs
@@ -128,18 +128,10 @@ namespace Akka.Remote
             {
             }
 
-            private static readonly Heartbeat _instance = new Heartbeat();
-
             /// <summary>
             /// TBD
             /// </summary>
-            public static Heartbeat Instance
-            {
-                get
-                {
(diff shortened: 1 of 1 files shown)


src/core/Akka.Remote/RemoteWatcher.cs
```diff
@@ -485,12 +475,7 @@ private void ReceiveHeartbeatRsp(int uid)
 
             if (WatcheeByNodes.ContainsKey(from) && !Unreachable.Contains(from))
             {
-                if (_addressUids.TryGetValue(from, out int addressUid))
```
Aaronontheweb, line 488: Simplified this to 

```
 if (!_addressUids.TryGetValue(from, out int addressUid) || addressUid != uid)
```

Aaronontheweb:
Another mysterious build failure on Linux - I don't see any compilation errors in the logs though.

Arkatufus (approved):
LGTM

PR #4922, opened 2021-04-08
Add spec for handling delegates in DI

Arkatufus:


Aaronontheweb:
Looks like one of these specs failed

Arkatufus:
It was a racy spec

Aaronontheweb:
@Arkatufus yes, but it looks like a new one we just introduced - we should try to harden that before it gets merged into the suite.

Aaronontheweb:
This is the spec that failed - it's only the tests that you are adding in this PR that are racy in this instance; incrementalist won't run any of the tests from outside this test suite since the only library that is affected is Akka.DependencyInjection:

**Akka.DependencyInjection.Tests.ActorServiceProviderPropsWithScopesSpecs.ActorsWithNonDiDependenciesShouldStart**

```
Expected deps1.Dependencies.Where(x => (x is AkkaDiFixture.ISingletonDependency)).All(x => x.Disposed) to be true, but found False.`
```

Arkatufus:
I didn't write that test.
I'll look into it.

Aaronontheweb:
> I didn't write that test.
> I'll look into it.

Whoops - my bad

Issue #4873, opened 2021-03-24
Need to integration test Akka.DistributedData w/ Durability and Akka.Cluster.Sharding


Aaronontheweb:
Version: 1.4.18

We're getting some bug reports that `remember-entities` doesn't work with Akka.DistributedData when an Akka.NET cluster undergoes a full reboot - naturally, that makes sense because all of the in-memory CRDTs are dumped when the entire cluster is shut down and spun back up again.

However, we've _also_ received reports that when enabling "durable" DData, which should be using LightningDb to save CRDT data to disk, this still doesn't work.

So we need to create a reproduction spec in the MNTR test suite to verify this:

1. Create a 3-node sharded cluster using DData w/ remember-entities=on and durable DData enabled;
2. Create at least 1 shard per node;
3. Shut down the entire cluster;
4. Reboot the cluster;
5. Validate that all entities are started back up.

If that test can pass, we need to document how to enable these settings inside the Akka.Cluster.Sharding documentation and also close out https://github.com/akkadotnet/akka.net/issues/4810 while we're at it - if it can't pass, we need to fix it.

Aaronontheweb:
Relevant Gitter thread: https://gitter.im/akkadotnet/akka.net?at=6059d56a9ebdfd1640921ba8

Arkatufus:
We actually turned this off...
https://github.com/akkadotnet/akka.net/issues/4262

Arkatufus:
I just compared our code and the JVM code, it turns out that we're missing a large chunk of DData sharding code, it might take awhile to port those into our code base.

Aaronontheweb:
@Arkatufus I think we turned it off at the time simply because we didn't have Durable DData support completed yet when we released it. That was fixed later but we never went back and revisited it.

Arkatufus:
Yes, it was the `Akka.DistributedData.LightningDB` port that I did, but it turns out, we're missing more code than just that.

Arkatufus:
These are the files that needs to be ported:
![image](https://user-images.githubusercontent.com/8293237/112875795-d8de5e00-90ee-11eb-958b-f21f95b7a8dd.png)

Aaronontheweb:
@Arkatufus go ahead and get started on it

Aaronontheweb:
```
[...]

PR #4926, opened 2021-04-09
Turn on DData related ClusterSharding MNTR specs

### LINKED ISSUE #4262: Akka.Cluster.Sharding: need to enable remember-entities to work in durable DData mode
As of https://github.com/akkadotnet/akka.net/pull/4255 remember-entities will work inside Akka.Cluster.Sharding with `storage-mode = ddata` - _however_, in the event of a full cluster restart the sharding system won't recover any of its previously remembered entities. This is a bit of a problem.

In order to fix that, we will first need to implement https://github.com/akkadotnet/akka.net/issues/4261 followed by getting all of the DDataDurable_Specs to pass.

Arkatufus:
Related to #4873
Accidentally also fixes  #4262 (no idea why)


ismaelhamed:
> Accidentally also fixes #4262 (no idea why)

Could it be because `durable.keys` option was commented and that would default to an empty list, which is how you [disable](https://github.com/akka/akka/blob/bff26488a40789ba05f242ec0a4431370549c20b/akka-docs/src/main/paradox/typed/cluster-sharding.md#remember-entities-distributed-data-mode) persisting the remeber entities store to disk in DData?

Arkatufus:
Could be, it also fixes inconsistency in the hocon config, instead of having 2 separate configs, it actually references the DistributedData config, so we don't have to maintain 2 separate configurations for DData and sharding

Arkatufus:
and of course it immediately failed in CI/CD while it worked locally...

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -491,6 +491,10 @@ private Props CoordinatorProps(string typeName, bool rebalanceEntities, bool rem
                 .WithFallback(Sys.Settings.Config.GetConfig("akka.cluster.sharding"));
             var settings = ClusterShardingSettings.Create(config, Sys.Settings.Config.GetConfig("akka.cluster.singleton"))
                 .WithRememberEntities(rememberEntities);
+            var majorityMinCap = Sys.Settings.Config.GetInt("akka.cluster.sharding.distributed-data.majority-min-cap");
+            if (IsDDataMode)
+                return DDataShardCoordinator.Props(typeName, settings, allocationStrategy, ReplicatorRef,
```
Arkatufus, line 496: Fix for the spec, need to return the correct coordinator for each test

src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingSpec.cs
```diff
@@ -540,14 +544,11 @@ public void ClusterSharding_specs()
 
             ClusterSharding_should_be_easy_API_for_starting();
 
-            if (!IsDDataMode)
```
Arkatufus, line 543: Turn DData test on

src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
```diff
@@ -277,7 +277,8 @@ public ClusterSharding(ExtendedActorSystem system)
         /// <returns>TBD</returns>
         public static Config DefaultConfig()
         {
-            return ConfigurationFactory.FromResource<ClusterSharding>("Akka.Cluster.Sharding.reference.conf");
+            return ConfigurationFactory.FromResource<ClusterSharding>("Akka.Cluster.Sharding.reference.conf")
+                .WithFallback(DistributedData.DistributedData.DefaultConfig());
```
Arkatufus, line 281: Changed how configuration works. Instead of copying the structure of `akka.harding.distributed-data` into sharding config,`akka.cluster.sharding.distributed-data` works as an override for `akka.cluster.istributed-data` instead.
This is done to minimize problems where we have 2 different distributed data settings in the system, potentially bugging the system.

src/contrib/cluster/Akka.Cluster.Sharding/ClusterShardingGuardian.cs
```diff
@@ -251,7 +251,9 @@ public ClusterShardingGuardian()
 
         private ReplicatorSettings GetReplicatorSettings(ClusterShardingSettings shardingSettings)
         {
-            var configuredSettings = ReplicatorSettings.Create(Context.System.Settings.Config.GetConfig("akka.cluster.sharding.distributed-data"));
+            var config = Context.System.Settings.Config.GetConfig("akka.cluster.sharding.distributed-data")
+                .WithFallback(Context.System.Settings.Config.GetConfig("akka.cluster.distributed-data"));
```
Arkatufus, line 255: Same reason as above.

src/contrib/cluster/Akka.Cluster.Sharding/DDataShard.cs
```diff
@@ -127,7 +127,37 @@ internal sealed class DDataShard : ActorBase, IShard, IWithUnboundedStash, IWith
             }
         }
 
-        public void EntityTerminated(IActorRef tref) => this.BaseEntityTerminated(tref);
+        public void EntityTerminated(IActorRef tref) 
+        {
+            if (!IdByRef.TryGetValue(tref, out var id)) return;
```
Arkatufus, line 132: In the original implementation, `DDataShard` inherits from `RememberingShard` class, not the base `Shard` class through a mixin.
In our implementation, since C# doesn't support mixins, it's implemented with an override. The bug here is that `DDataShard` didn't override the default `EntityTerminated` implementation.

src/contrib/cluster/Akka.Cluster.Sharding/DDataShardCoordinator.cs
```diff
@@ -345,7 +345,7 @@ private void SendCoordinatorStateUpdate(PersistentShardCoordinator.IDomainEvent
         {
             var s = CurrentState.Updated(e);
             _replicator.Tell(Dsl.Update(_coordinatorStateKey,
-                new LWWRegister<PersistentShardCoordinator.State>(Cluster.SelfUniqueAddress, PersistentShardCoordinator.State.Empty),
+                new LWWRegister<PersistentShardCoordinator.State>(Cluster.SelfUniqueAddress, PersistentShardCoordinator.State.Empty.WithRememberEntities(Settings.RememberEntities)),
```
Arkatufus, line 348: Bug, need to pass `RememberEntities` settings into the state.

Arkatufus:
Test is somehow failing in .NET Framework, LightningDB is somehow not disposed.

Arkatufus:
Changed the LightningDB actor code.
The LightningDB implementation that we used does not like it when its dispose method are called from the finalizer, which somehow kept happening in .NET Framework.

Aaronontheweb (approved):
LGTM

PR #4930, opened 2021-04-10
cleaned up protobuf CLI and definitions

Aaronontheweb:
- Remove all `optional` fields (not allowed in Protobuf3, as all fields are optional by default unless specifically defined as `required`)
- Removed `--experimental_allow_proto3_optional` call from `protoc` compiler as it's no longer supported / needed
- Doesn't have any impact on existing wire formats, especially for `ClusterMessages.proto` which is where I removed all of the `optional` commands


src/protobuf/ClusterMessages.proto
```diff
@@ -30,7 +30,7 @@ import "ContainerFormats.proto";
 message Join {
   UniqueAddress node = 1;
   repeated string roles = 2;
-  optional string appVersion = 3;
```
Aaronontheweb, line 33: Not a meaningful wire format change

src/protobuf/ClusterMessages.proto
```diff
@@ -138,7 +138,7 @@ message Member {
   int32 upNumber = 2;
   MemberStatus status = 3;
   repeated int32 rolesIndexes = 4 [packed = true];
-  optional int32 appVersionIndex = 5;
```
Aaronontheweb, line 141: Not a meaningful wire format change

src/core/Akka.Cluster/Serialization/ClusterMessageSerializer.cs
```diff
@@ -106,7 +106,7 @@ private static byte[] JoinToByteArray(InternalClusterAction.Join join)
         private static InternalClusterAction.Join JoinFrom(byte[] bytes)
         {
             var join = Proto.Msg.Join.Parser.ParseFrom(bytes);
-            AppVersion ver = join.HasAppVersion ? AppVersion.Create(join.AppVersion) : AppVersion.Zero;
+            var ver = !string.IsNullOrEmpty(join.AppVersion) ? AppVersion.Create(join.AppVersion) : AppVersion.Zero;
```
Aaronontheweb, line 109: Wire type didn't change, but the `HasAppVersion` property disappeared.

Issue #4849, opened 2021-03-15
Akka.Cluster: quarantining / reachability changes appear to be extremely sensitive


Aaronontheweb:
Akka.NET version: 1.4.17

As the description states, in a cluster with the following composition:

~20 unreachable nodes
~40 weakly up nodes
~30 up nodes

Any attempts to `Down` the unreachable nodes successfully change their status from `Up` / `WeaklyUp` -> `Down`, but they are never correctly exited from the cluster nor are any of the `WeaklyUp` nodes promoted to `Up`. Using a split brain resolver and a `pbm cluster down-unreachable` command have both demonstrated this.

This looks like a gossip convergence issue to me - going to investigate.

ismaelhamed:
Is this related to #4757 ? Every single time a DOWN does not seem to propagate is because there's at least one node in TERMINATED state.

Aaronontheweb:
Changed the title to: "Akka.Cluster: quarantining / reachability changes appear to be extremely sensitive"

Looks to me like there's some issues, which can be easily reproduced during a large Akka.Cluster deployment (i.e. scaling from 3 to 25 nodes via `kubectl`), that really shouldn't be occurring. Message traffic isn't all that high when this occurs - I think it has to do with the `failure-detector` being extremely overtuned in Akka.Cluster.

Aaronontheweb:
https://gist.github.com/Aaronontheweb/66095c9340437c0576cf55876d65c1f7 - some related theories I'm probing related to this

Aaronontheweb:
Worked on this as part of a "heartbeats not setting sent fast enough" hypothesis https://github.com/akkadotnet/akka.net/pull/4882

Aaronontheweb:
Working theories as to what is causing this:

- [ ] Addition of `AppVersion` data skews reachability detection when a "rolling upgrade" is detected
- [ ] https://github.com/akkadotnet/akka.net/pull/4511 - introduction of the `InternalDispatcher` actually made heartbeat scheduling less reliable, especially on larger node rings
- [ ] https://github.com/akkadotnet/akka.net/pull/4097 - that this change, which should have made it more difficult for the cluster to arrive at REACHABLE vs. UNREACHABLE decisions, doesn't really have any effect and any single node in the observability ring can effectively mark any of the others as unreachable.
- [ ] Bug in the cluster heartbeat system that causes it to be overly sensitive to early heartbeat data at startup. This would be an issue with the Phi Accrual system if that were the case.
[...]

PR #4946, opened 2021-04-18
Configure duration for applying `MemberStatus.WeaklyUp`  to joining nodes

Aaronontheweb:
port of https://github.com/akka/akka/pull/29665

close https://github.com/akkadotnet/akka.net/issues/4849

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCluster.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCluster.approved.txt
@@ -220,6 +220,7 @@ namespace Akka.Cluster
         public string UseDispatcher { get; }
         public bool VerboseGossipReceivedLogging { get; }
         public bool VerboseHeartbeatLogging { get; }
+        public System.TimeSpan WeaklyUpAfter { get; }
     }
     [Akka.Annotations.InternalApiAttribute()]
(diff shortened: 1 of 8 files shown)


Aaronontheweb:
Looks like this did help some:

**Before**

```
[INFO][4/18/2021 7:41:06 PM][Thread 0069][akka.trttl.gremlin.tcp://StressSpec@localhost:58399/user/result6] [join 2 to seed nodes, in 9 nodes cluster] completed in [8095] ms
```

**After**

```
[INFO][4/18/2021 7:14:12 PM][Thread 0066][akka.trttl.gremlin.tcp://StressSpec@localhost:57599/user/result6] [join 2 to seed nodes, in 9 nodes cluster] completed in [7318] ms
```

Aaronontheweb:
Looks like some issues with the `WeaklyUp` MNTR specs - I'll take a look at those.

Aaronontheweb:
> Looks like some issues with the `WeaklyUp` MNTR specs - I'll take a look at those.

This was a bug in the `ClusterSettings` code - I was validating that the `allow-weakly-up-members` had to be less than 7 seconds, when the check really should be that it's greater than zero. I've fixed this since.

Arkatufus (approved):
LGTM

Issue #4948, opened 2021-04-19
Bug: `ClusterEvent.MemberRemoved` can jump from `MemberLeaving` without being `MemberExited`


Aaronontheweb:
Version: Akka.NET v1.4.18

Based on some build log failures detected in PR https://github.com/akkadotnet/akka.net/pull/4946/

```
Akka.Cluster.Tests.ClusterSpec.A_cluster_must_leave_via_CoordinatedShutdownRun_when_member_status_is_Joining
Expected removed.PreviousStatus to equal MemberStatus.Exiting(3) by value, but found MemberStatus.Leaving(2).
```

This isn't legal according to the member transition spec - issue occurs here:

https://github.com/akkadotnet/akka.net/blob/65e5a2228061b562a628889c6a812f0e7c306736/src/core/Akka.Cluster/ClusterEvent.cs#L880-L903

Going to need to see if we can reproduce this with a simple use case, as this is definitely a gossip propagation / merging / diffing bug.

Aaronontheweb:
The unit test that was able to reproduce this bug (and this unit test has a very low flip rate, historically)

https://github.com/akkadotnet/akka.net/blob/65e5a2228061b562a628889c6a812f0e7c306736/src/core/Akka.Cluster.Tests/ClusterSpec.cs#L468-L500

Aaronontheweb:
Pretty sure this has been resolved.

PR #4949, opened 2021-04-19
restored Akka.Cluster model-based FsCheck specs

Aaronontheweb:
Going to attempt to use these, and specs similar to these, to track down https://github.com/akkadotnet/akka.net/issues/4948 and any other bugs that might be possibly related to it.


Aaronontheweb:
Bleh, this is going to do a full run of the test suite since I removed some precompilation directives from Akka.Tests.Shared.Internals.

PR #4953, opened 2021-04-19
Fix BatchingSqlJournal emitting WriteMessageSuccess before transaction was complete.

### LINKED ISSUE #4265: Some persistent actors are stuck with RecoveryTimedOutException after circuit breaker opens
This issue looks similar to #3870, however it happens when using latest version of Akka.NET.

OS: Windows Server 2016
Platform: .NET Core 3.1
Akka.NET packages: 1.4.0-beta14 (used in a cluster)

Scenario:

1. Akka.Persistence.SqlServer.Journal.BatchingSqlServerJournal raises exception with message "Circuit Breaker is open; calls are failing fast", most likely due to a temporary db outage

2. Attempt to recover state of some persistent actors fail with RecoveryTimedOutException. Here's a typical sequence of events, taken from our log:

```
Started (Akka.Pattern.BackoffOnRestartSupervisor)
now supervising akka://Oddjob/system/sharding/upload/0/ps~msui30002111/msc:ps~msui30002111
now watched by [akka://Oddjob/system/sharding/upload/0/ps~msui30002111#1585193596]
now watched by [akka://Oddjob/system/recoveryPermitter#1099929798
Spawned MediaSetController actor
now watched by [akka://Oddjob/system/sharding/upload/0#1224240942]
Started (Akkling.Persistence.FunPersistentActor`1[System.Object])
Restoring state from snapshot
(after 1 minute)
["", null, "Akka.Persistence.RecoveryTimedOutException: Recovery timed out, didn't get event within 60s, highest sequence number seen 312."] {AckInfo} {Exception}
Passivating started on entity "ps~msui30002111"
received AutoReceiveMessage <Terminated>: [akka://Oddjob/system/sharding/upload/0/ps~msui30002111#1585193596] - ExistenceConfirmed=True
Entity stopped after passivation ["ps~msui30002111"]

```

3. Once a persistent actor fails with such exception, it is stuck until the system is restarted. Other actors may be successfully recovered.

Arkatufus:
Should also fix #4265, but would need further field testing.


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1388,17 +1404,59 @@ protected void AddParameter(TCommand command, string paramName, DbType dbType, o
...
+        /// no guarantee that no concrete implementation is using them.
+        /// -- Greg
+        /// </summary>
         private RequestChunk DequeueChunk(int chunkId)
         {
-            var operationsCount = Math.Min(Buffer.Count, Setup.MaxBatchSize);
-            var array = new IJournalRequest[operationsCount];
-            for (int i = 0; i < operationsCount; i++)
+            // Need a lock here to ensure that buffer doesn't change during this operation
+            lock (_lock)
```
Arkatufus, line 1421: This is the very janky code to split the queued requests into either a read request chunk or write request chunk.
Explanation on why it is coded in this peculiar way is written in the method comment.

src/core/Akka.Persistence/JournalProtocol.cs
```diff
@@ -22,6 +22,16 @@ public interface IJournalMessage : IPersistenceMessage { }
     /// </summary>
     public interface IJournalRequest : IJournalMessage { }
 
+    /// <summary>
+    /// Marker for batched write operations
+    /// </summary>
+    public interface IJournalWrite : IJournalRequest { }
```
Aaronontheweb, line 28: Let's remove this and try to avoid making the `BatchingJournal`'s problem all of Akka.Persistence's problem.

Aaronontheweb (changes requested):
Let's not modify Akka.Persistence API surface area to save the `BatchingSqlJournal`.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -733,31 +748,65 @@ protected sealed override bool Receive(object message)
...
+                        {
+                            if (!(envelope is AtomicWrite write)) 
+                                continue;
+
+                            var writes = (IImmutableList<IPersistentRepresentation>)write.Payload;
+                            foreach (var unadapted in writes)
+                            {
+                                if (cause is DbException)
+                                {
+                                    // database-related exceptions should result in failure
```
Aaronontheweb, line 776: LGTM

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -829,10 +878,20 @@ private void NotifyPersistenceIdChanged(string persistenceId)
...
+            if (WriteBuffer.Count + ReadBuffer.Count > Setup.MaxBufferSize)
+            {
                 OnBufferOverflow(message);
+            }
             else
-                Buffer.Enqueue(message);
+            {
+                var id = _bufferIdCounter.GetAndIncrement();
+                // Enqueue writes and delete operation requests into the write queue,
+                // else if they are query operations, enqueue them into the read queue
```
Aaronontheweb, line 889: LGTM

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1388,29 +1389,111 @@ protected void AddParameter(TCommand command, string paramName, DbType dbType, o
...
+        /// items from it, and return it as a chunk that needs to be batched
+        /// </summary>
         private RequestChunk DequeueChunk(int chunkId)
         {
-            var operationsCount = Math.Min(Buffer.Count, Setup.MaxBatchSize);
-            var array = new IJournalRequest[operationsCount];
-            for (int i = 0; i < operationsCount; i++)
+            var currentBuffer = _buffers
+                .Where(q => q.Count > 0)
+                .OrderBy(q => q.Peek().id).First();
```
Aaronontheweb, line 1400: Is the buffer already sorted this way?

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1388,29 +1389,111 @@ protected void AddParameter(TCommand command, string paramName, DbType dbType, o
         /// <param name="param">Parameter to customize</param>
         protected virtual void PreAddParameterToCommand(TCommand command, DbParameter param) { }
         
+        /// <summary>
+        /// Select the buffer that has the smallest id on its first item, retrieve a maximum Setup.MaxBatchSize
```
Aaronontheweb, line 1393: Because this indicates the "oldest" outstanding operation?

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1388,29 +1389,111 @@ protected void AddParameter(TCommand command, string paramName, DbType dbType, o
...
+                {
+                    operations.Add(currentBuffer.Dequeue().request);
+                    if (operations.Count == Setup.MaxBatchSize)
+                        break;
+                }
             }
-
-            return new RequestChunk(chunkId, array);
+            
+            return new RequestChunk(chunkId, operations.ToArray());
```
Aaronontheweb, line 1426: Why not just pass the `List<>` instead?

Aaronontheweb (approved):
LGTM overall but left some questions for @Arkatufus

Aaronontheweb:
cc @ismaelhamed

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
Arkatufus: This is just a way to pick the buffer with the lowest id in its first item. We're trying to make sure that request order are roughly executed in order, if it was a choice between a read and a write.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
Arkatufus: Yes, it is "roughly first in, first out". Question is, how sensitive is journal request ordering between read and write operations, are they strict or can operations happen in asynchrony?
If they're tightly coupled, I'd have to change this logic so that only consecutive requests of the same operations are allowed to be batched, ie. all queries, all writes, and all deletes.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
Arkatufus: RequestChunk is a struct, I'm not sure if there is a specific reason why it is using an array and not a list, I assumed that it is for optimization reason, so I didn't change it.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
Arkatufus: This is actually a big question that needs to be addressed.

to11mtm: FWIW, I know that persistence-jdbc makes a point of blocking `ReadHighestSequenceNrAsync` for a given PersistenceId ([example in ported code](https://github.com/akkadotnet/Akka.Persistence.Linq2Db/blob/dev/src/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs#L141-L159).

I think this is done, because the first step of Recovery in `AsyncWriteJournal` is to read the max sequence number from which to replay. This way, if somehow the PersistentActor crashes before a `PersistAsync` write completes, Recovery sequence will still read up to the correct sequence number.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -926,21 +990,12 @@ private async Task<BatchComplete> ExecuteChunk(RequestChunk chunk, IActorContext
...
-                            {
-                                context.System.EventStream.Publish(chunk.Requests[i]);
-                            }
-                        }
                     }
-                    catch (Exception e)
+                    catch (Exception)
                     {
-                        cause = e;
                         tx.Rollback();
```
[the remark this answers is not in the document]
Aaronontheweb: It would probably be a good idea to include it

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
Arkatufus: So, to make things as safe as possible, i guess i should only batch consecutive similar commands, unless it is a ReadHighestSequenceNr, in which i should stop. This is turning into an `UnBatchedSqlJournal` real fast...

Arkatufus: Well, all query types should be safe to be batched together as long s they're consecutive, right. The danger comes when it is interleaved with writes, and for write operations, consecutive writes and consecutive deletes are fine.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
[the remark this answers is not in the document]
Arkatufus: Added a try...catch around the rollback and throw an aggregated exception of it and the original exception

ismaelhamed:
LGTM. This was trickier than I thought, nice job! 

Would a `-beta` of the `Akka.Persistence.Sql.Common` package with this branch be appropriate until we're sure this won't backfire in production? 

I'm not concerned with @Arkatufus coding at all ;) but, given that this has deviated from the original implementation and that the original author was not involved to pick his brain in some of the issues that came up, it may be more sensible to do so.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -733,31 +748,65 @@ protected sealed override bool Receive(object message)
 
         private void FailChunkExecution(ChunkExecutionFailure message)
         {
+            _remainingOperations++;
             var cause = message.Cause;
-            Log.Error(cause, "Failed to execute chunk for {0} requests", message.Requests.Length);
-
-            foreach (var req in message.Requests)
+            Log.Error(cause, "An error occurred during event batch processing (chunkId: {0})", message.ChunkId);
+
```
ismaelhamed, line 754: I'd keep the number of requests (i.e., ` (chunkId: {0}) of {1} requests`)

Aaronontheweb:
Yep, we can do that @ismaelhamed - I can make those into beta releases

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
to11mtm: After pondering it, I think we're probably Safe since BatchingSqlJournal is not from AsyncJournalBase and therefore doing recovery differently.

At -worst-, I think we would see a scenario where after recovery, the actor crashes b/c it didn't pick up the right sequence number. In that at worst, I can't imagine it would be much more than restarting the actor (or at worst the whole system.) 

That's still a better worst case than with current state of BatchingSqlJournal (where writes are being confirmed when they may still fail.) Under heavy loads I would see torn write behavior (i.e. missing sequence numbers in DB) basically corrupting the journal state.

Aaronontheweb: @to11mtm do you approve this set of changes then?

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
Arkatufus: Done

Aaronontheweb (approved):
LGTM

PR #4961, opened 2021-04-20
Fix dependabot Hyperion issue

Arkatufus:


Aaronontheweb:
LOL I never thought to send a PR to the dependabot branch - great idea! I'll definitely do this next time we run into a compat / SemVer issue in the future.

Arkatufus:
It makes it a lot easier to just fix the problem in the dependabot branch, and we can always tell it to rebase/recreate if it doesn't work

PR #4963, opened 2021-04-20
Clean up bad outbound ACKs in Akka.Remote

Aaronontheweb:
port of https://github.com/akka/akka/pull/20093

Might be responsible for some quarantines in Akka.Cluster / Akka.Remote when nodes are restarting on identical addresses.

### DIFF
--- a/src/core/Akka.Remote/Endpoint.cs
+++ b/src/core/Akka.Remote/Endpoint.cs
@@ -1139,7 +1139,7 @@ namespace Akka.Remote
             }
             _buffer.Clear();
 
-            if (_handle != null) _handle.Disassociate(_stopReason);
+            _handle?.Disassociate(_stopReason);
             EventPublisher.NotifyListeners(new DisassociatedEvent(LocalAddress, RemoteAddress, Inbound));
         }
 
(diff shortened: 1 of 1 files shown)


src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1227,56 +1233,50 @@ private void Handoff()
         /// <param name="message">TBD</param>
         protected override void Unhandled(object message)
         {
-            if (message is Terminated)
+            switch (message)
```
Aaronontheweb, line 1236: Just resharper'd this.

src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1219,6 +1219,12 @@ private void Handoff()
                 BecomeWritingOrSendBufferedMessages();
             });
             Receive<EndpointManager.Send>(send => EnqueueInBuffer(send));
+
+            // Ignore outgoing acks during take over, since we might have
+            // replaced the handle with a connection to a new, restarted, system
+            // and the ack might be targeted to the old incarnation.
+            // Relates to https://github.com/akka/akka/pull/20093
+            Receive<OutboundAck>(_ => { });
```
Aaronontheweb, line 1227: Bug fix.

src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1227,56 +1233,50 @@ private void Handoff()
...
-                {
+                    break;
+                case StopReading stop:
                     // initializing, buffer and take care of it later when buffer is sent
-                    EnqueueInBuffer(message);
+                    EnqueueInBuffer(stop);
+                    break;
+                case TakeOver takeover:
+                    // Shutdown old reader
+                    _handle.Disassociate("the association was replaced by a new one", _log);
```
Aaronontheweb, line 1256: Cleaned up the disassociation message here too, so it's more clear when a remote connection is being taken over vs. when a node disassociates due to failure.

Aaronontheweb (commented):
Describes my changes

PR #4968, opened 2021-04-21
Refactored `Gossip` into `MembershipState`

Aaronontheweb:
port of https://github.com/akka/akka/pull/23291 - should make it much easier for me to write some `FsCheck` specs to test cluster membership transition changes, per my comment on https://github.com/akkadotnet/akka.net/issues/4849#issuecomment-823745088

### DIFF
--- a/src/core/Akka.Cluster/ClusterDaemon.cs
+++ b/src/core/Akka.Cluster/ClusterDaemon.cs
@@ -1742,14 +1742,8 @@ namespace Akka.Cluster
                 else
                     _cluster.LogInfo("Marking unreachable node [{0}] as [{1}]", member.Address, MemberStatus.Down);
 
-                // replace member (changed status)
-                var newMembers = localMembers.Remove(member).Add(member.Copy(MemberStatus.Down));
-                // remove nodes marked as DOWN from the 'seen' table
-                var newSeen = localSeen.Remove(member.UniqueAddress);
 
-                //update gossip overview
-                var newOverview = localOverview.Copy(seen: newSeen);
-                var newGossip = localGossip.Copy(members: newMembers, overview: newOverview); //update gossip
+                var newGossip = localGossip.MarkAsDown(member); //update gossip
                 UpdateLatestGossip(newGossip);
 
                 PublishMembershipState();
--- a/src/core/Akka.Cluster/ClusterDaemon.cs
+++ b/src/core/Akka.Cluster/ClusterDaemon.cs
@@ -17,6 +17,7 @@ using Akka.Remote;
 using Akka.Util;
 using Akka.Util.Internal;
 using Akka.Util.Internal.Collections;
+using static Akka.Cluster.MembershipState;
 
 namespace Akka.Cluster
 {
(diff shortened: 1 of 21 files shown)


src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -2566,7 +2568,7 @@ public void GossipStatusTo(UniqueAddress node, IActorRef destination)
         /// <returns>TBD</returns>
         public bool ValidNodeForGossip(UniqueAddress node)
         {
-            return !node.Equals(SelfUniqueAddress) && _latestGossip.Overview.Reachability.IsReachable(SelfUniqueAddress, node);
+            return !node.Equals(SelfUniqueAddress) && _membershipState.IsReachableExcludingDownedObservers(node);
```
Aaronontheweb, line 2571: This was definitely a bug with our gossip system

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -2629,19 +2637,21 @@ public void UpdateLatestGossip(Gossip newGossip)
...
 
         /// <summary>
         /// Publishes gossip to other nodes in the cluster.
         /// </summary>
-        /// <param name="newGossip">The new gossip to share.</param>
-        public void Publish(Gossip newGossip)
+        public void PublishMembershipState()
         {
-            _publisher.Tell(new InternalClusterAction.PublishChanges(newGossip));
+            if (_cluster.Settings.VerboseGossipReceivedLogging)
```
Aaronontheweb, line 2651: Added additional `Gossip` logging here

Aaronontheweb (commented):
Left some comments

Aaronontheweb:
Looks like I must not have ported everything correctly based on the test failures - I'll get that cleaned up

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1663,17 +1665,17 @@ public void Joining(UniqueAddress node, ImmutableHashSet<string> roles, AppVersi
...
+            if (!LatestGossip.Members.IsEmpty) throw new InvalidOperationException("Welcome can only be done from an empty state");
             if (!joinWith.Equals(from.Address))
             {
                 _cluster.LogInfo("Ignoring welcome from [{0}] when trying to join with [{1}]", from.Address, joinWith);
             }
             else
             {
                 _cluster.LogInfo("Welcome from [{0}]", from.Address);
-                _latestGossip = gossip.Seen(SelfUniqueAddress);
+                _membershipState = _membershipState.Copy(gossip).Seen();
```
Aaronontheweb, line 1676: Ah, this is probably why our tests were failing - forgot to copy the initial `Gossip` payload sent by the leader during a `Welcome` operation

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -2601,25 +2603,31 @@ public static IEnumerable<Member> GossipTargetsForExitingMembers(Gossip latestGo
...
+                    // Update the state with the new gossip
+                    return seenVersionedGossip;
+                }
             }
+
+            // Don't mark gossip state as seen while exiting is in progress, e.g.
+            // shutting down singleton actors. This delays removal of the member until
+            // the exiting tasks have been completed.
+            var newGossip = PickLatest();
+            _membershipState = _membershipState.Copy(newGossip);
```
Aaronontheweb, line 2630: Forgot to assign updated `MembershipState` here

Aaronontheweb:
Looks like the main Akka.Cluster / Tools / Sharding / DData test suites are all passing again. We'll see what the MNTR finds - and if that's all ok we'll get this reviewed and I can start work on some model-based tests to validate cluster and membership transitions.

Aaronontheweb:
`TransitionSpec` failed in two of the runs so far, so there's definitely some more kinks with this port that need to be worked out.

Aaronontheweb:
Have `TransitionSpec` passing locally again - was a goof on the `ValidNodeForGossip` check.

src/core/Akka.Cluster/ClusterDaemon.cs
Aaronontheweb: Not true actually - looks like we had it right.

Arkatufus (approved):
LGTM

PR #4970, opened 2021-04-21
Adding Akka.Cluster membership model-based spec

### LINKED ISSUE #4948: Bug: `ClusterEvent.MemberRemoved` can jump from `MemberLeaving` without being `MemberExited`

Aaronontheweb:
close #4948

per https://github.com/akkadotnet/akka.net/issues/4849#issuecomment-823745088

Using an FsCheck model-based spec to exhaustively validate some portions of our membership model .

This might be a bit of a fool's errand to do how much of the state for this work is done inside the `ClusterCoreDaemon`, and that might be very difficult to extract and emulate. Going to give it a stab and spend 5-6 hours on it though.

### DIFF
--- a/src/benchmark/Akka.Benchmarks/Actor/SpawnActorBenchmarks.cs
+++ b/src/benchmark/Akka.Benchmarks/Actor/SpawnActorBenchmarks.cs
@@ -15,38 +15,41 @@ using BenchmarkDotNet.Engines;
 namespace Akka.Benchmarks.Actor
 {
     [Config(typeof(MicroBenchmarkConfig))]
-    [SimpleJob(RunStrategy.Throughput, targetCount:10, warmupCount:5, invocationCount: ActorCount)]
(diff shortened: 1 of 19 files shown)


Aaronontheweb:
Looks like this is no longer necessary.

PR #4972, opened 2021-04-21
simplified the environment variable name for StressSpec

Aaronontheweb:
Did this to make it easier to set the maximum node count in PowerShell,

i.e. `$env:MNTR_STRESSSPEC_NODECOUNT="50"`


Aaronontheweb:
Was able to validate that this PowerShell syntax works great locally.

Issue #4967, opened 2021-04-21
Akka.Cluster FirstSeedNodeRestartSpec failure


Aaronontheweb:
Version: v1.4.18

Possibly related to https://github.com/akkadotnet/akka.net/issues/4948 - looks like we have a hard failure even with a full cohort of seed nodes where a full cluster failed to form correctly upon restarting the first seed node.

[first-seednode-restart-failurespec.txt](https://github.com/akkadotnet/akka.net/files/6347774/first-seednode-restart-failurespec.txt)

Noticed the failure on https://github.com/akkadotnet/akka.net/runs/2395764208

Aaronontheweb:
Rest of the related logs for this issue:

[node1__seed1__net5.txt](https://github.com/akkadotnet/akka.net/files/6353179/node1__seed1__net5.txt)
[node2__seed2__net5.txt](https://github.com/akkadotnet/akka.net/files/6353180/node2__seed2__net5.txt)
[node3__seed3__net5.txt](https://github.com/akkadotnet/akka.net/files/6353182/node3__seed3__net5.txt)

PR #4975, opened 2021-04-23
clean up seed node process

Aaronontheweb:
close https://github.com/akkadotnet/akka.net/issues/4967

Reviewed the code in the seed node process actors to try to address issues with the failing `FirstSeedNodeSpec` and everything looks kosher. Spec failed because of an issue with the non-first-seed nodes replying back on-time.


Arkatufus (approved):
LGTM

PR #4978, opened 2021-04-23
Update PluginSpec so that it can accept ActorSystem and ActorSystemSetup in its constructor

Arkatufus:


Aaronontheweb (approved):
LGTM - great change

src/core/Akka/Actor/ActorSystem.cs
```diff
@@ -93,7 +93,7 @@ internal static ProviderSelection GetProvider(string providerClass)
     }
 
     /// <summary>
-    /// Core boostrap settings for the <see cref="ActorSystem"/>, which can be created using one of the static factory methods
+    /// Core bootstrap settings for the <see cref="ActorSystem"/>, which can be created using one of the static factory methods
```
Aaronontheweb, line 96: Doh - nice catch

PR #4989, opened 2021-04-28
Change the base class of PrimitiveSerializers to SerializerWithStringManifest

### LINKED ISSUE #4986: Akka.Remote.Serialization.PrimitiveSerializers needs cross platform compatibility
Akka.Remote.Serialization.PrimitiveSerializers breaks when an actor system living inside a .Net core environment sends a remote message to ones that are living inside .Net Framework environment, and vice versa because of the namespace changes introduced between NetFx and NetCore.

Need to change the base class of the serializer to `SerializerWithStringManifest` to make sure that no future namespace changes will break the wire format again.

Arkatufus:
Closes  #4986

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+version: 2
+updates:
+- package-ecosystem: nuget
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+  ignore:
+  - dependency-name: NUnit
+    versions:
+    - 3.13.0
+    - 3.13.1
+  - dependency-name: Google.Protobuf
+    versions:
+    - 3.15.0
+    - 3.15.1
+    - 3.15.4
+  - dependency-name: FsCheck.Xunit
+    versions:
+    - 2.14.5
+    - 2.15.0
+  - dependency-name: FsCheck
+    versions:
+    - 2.14.5
+  - dependency-name: FluentAssertions
+    versions:
+    - 5.9.0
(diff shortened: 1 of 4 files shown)


Aaronontheweb (changes requested):
Need to make this backwards wire compatible - we have to accept the manifests that could be produced using reflection and the old type names too, and we'll have to add unit tests to verify that.

Arkatufus:
Alright, will do

Aaronontheweb (approved):
LGTM

Aaronontheweb:
@Arkatufus need to do API approvals as well here

Arkatufus:
Done

Martin-Molinero:
I'm getting `Serializer not defined for message with serializer id [6] and manifest []. Transient association error (association remains live). Cannot find manifest class [S] for serializer with id [17].` when using new akka remote `1.4.20` (or bigger) with an older version, think might be related to this change, breaks connection compatibility.

Aaronontheweb:
Hi @Martin-Molinero - this looks like a new version --> old version breaking wire format change that we caused, and the reason why is because we standardized the the primitive type manifests because they varied between .NET Core and .NET Framework due to the breaking changes Microsoft introduced at the inception of .NET Core.

We're really sorry about this - we probably should have made this an opt-in feature some months _after_ introducing this serializer so you wouldn't be hit with this issue during a live upgrade. The work around for this now is probably to either just bite the bullet and upgrade everything, or to override this serializer with the one from the v1.4.19 source code and drop it later.

Aaronontheweb:
@Martin-Molinero this should resolve this issue in v1.4.26 https://github.com/akkadotnet/akka.net/pull/5280

Martin-Molinero:
@Aaronontheweb awesome ty! is there an ETA for v1.4.26?

Aaronontheweb:
We can probably ship something simple soon

PR #4991, opened 2021-05-01
Fixed QueueSource PostStop and added a missing test case

ismaelhamed:
Ports [#23151](https://github.com/akka/akka/pull/23151)  and [#23509](https://github.com/akka/akka/pull/23509)

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/QueueSinkSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/QueueSinkSpec.cs
@@ -153,7 +153,7 @@ namespace Akka.Streams.Tests.Dsl
             }, _materializer);
         }
 
-        [Fact(Skip = "Racy, see https://github.com/akkadotnet/akka.net/pull/4424#issuecomment-632284459")]
+        [Fact]
         public void QueueSink_should_fail_pull_future_when_stream_is_completed()
         {
             this.AssertAllStagesStopped(() =>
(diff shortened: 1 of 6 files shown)


src/core/Akka.Streams.Tests/Dsl/QueueSinkSpec.cs
```diff
@@ -153,7 +153,7 @@ public void QueueSink_should_timeout_future_when_stream_cannot_provide_data()
             }, _materializer);
         }
 
-        [Fact(Skip = "Racy, see https://github.com/akkadotnet/akka.net/pull/4424#issuecomment-632284459")]
+        [Fact]
```
ismaelhamed, line 156: Let's see if this was fixed too

src/core/Akka.Streams.Tests/Dsl/QueueSinkSpec.cs
```diff
@@ -170,10 +170,8 @@ public void QueueSink_should_fail_pull_future_when_stream_is_completed()
                 var result = queue.PullAsync().Result;
                 result.Should().Be(Option<int>.None);
 
-                ((Task)queue.PullAsync()).ContinueWith(t =>
-                {
-                    t.Exception.InnerException.Should().BeOfType<IllegalStateException>();
-                }, TaskContinuationOptions.OnlyOnFaulted).Wait();
+                var exception = Record.ExceptionAsync(async () => await queue.PullAsync()).Result;
+                exception.Should().BeOfType<StreamDetachedException>();
```
ismaelhamed, line 174: Using XUnit API to catch the exception and unwrap. Also, this should always run, not just on `OnlyOnFaulted`, to make sure we catch regressions.

Aaronontheweb (approved):
LGTM

PR #4995, opened 2021-05-04
Fixed some internal-dispatcher usages

ismaelhamed:
These should run on the internal dispatcher too


Aaronontheweb (approved):
LGTM - these definitely need to run on the internal dispatcher.

PR #5002, opened 2021-05-10
Change PersistenceMessageSerializer base class to SerializerWithStringManifest

### LINKED ISSUE #5001: Need to make PersistenceMessageSerializer version tolerant
Version: v1.4.19

Right now our `PersistenceMessageSerializer` isn't using type manifests for its message names: 

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs

We need to perform the same type of transformation we just applied to the https://github.com/akkadotnet/akka.net/pull/4989 in order to preserve backwards compatibility with previously serialized message types while also decoupling any future changes to the internal type names from the manifests we use inside the persistence payloads.

Before we release v1.4.20 we will need to make sure this change is compatibility-tested against Akka.Persistence.SqlServer and its nightly builds.

Arkatufus:
Closes #5001

### DIFF
--- a/src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
+++ b/src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
@@ -19,22 +19,45 @@ using Google.Protobuf;
 
 namespace Akka.Persistence.Serialization
 {
-    public sealed class PersistenceMessageSerializer : Serializer
+    public sealed class PersistenceMessageSerializer : SerializerWithStringManifest
     {
+        public const string NullManifest = "null";
+
+        public const string IPersistentRepresentationManifest = "PR";
+        public const string AtomicWriteManifest = "AW";
+        public const string AtLeastOnceDeliverySnapshotManifest = "ALODS";
+        public const string StateChangeEventManifest = "FSM.SCE";
+        public const string PersistentFSMSnapshotManifest = "FSM.PS";
+
+        // Backward compatibility constants
+        public const string IPersistentRepresentationManifestNetCoreManifest = "Akka.Persistence.Persistent, Akka.Persistence";
+        public const string AtomicWriteNetCoreManifest = "Akka.Persistence.AtomicWrite, Akka.Persistence";
+        public const string AtLeastOnceDeliverySnapshotNetCoreManifest = "Akka.Persistence.AtLeastOnceDeliverySnapshot, Akka.Persistence";
+        public const string StateChangeEventNetCoreManifest = "Akka.Persistence.Fsm.PersistentFSM+StateChangeEvent, Akka.Persistence";
+        public const string PersistentFSMSnapshotNetCoreManifest = "Akka.Persistence.Fsm.PersistentFSM+PersistentFSMSnapshot";
+
+
         public PersistenceMessageSerializer(ExtendedActorSystem system) : base(system)
         {
         }
 
-        public override bool IncludeManifest { get; } = true;
-
         public override byte[] ToBinary(object obj)
         {
-            if (obj is IPersistentRepresentation repr) return GetPersistentMessage(repr).ToByteArray();
-            if (obj is AtomicWrite aw) return GetAtomicWrite(aw).ToByteArray();
-            if (obj is AtLeastOnceDeliverySnapshot snap) return GetAtLeastOnceDeliverySnapshot(snap).ToByteArray();
-            if (obj is PersistentFSM.StateChangeEvent stateEvent) return GetStateChangeEvent(stateEvent).ToByteArray();
+            switch (obj)
+            {
+                case IPersistentRepresentation repr:
+                    return GetPersistentMessage(repr).ToByteArray();
+                case AtomicWrite aw:
+                    return GetAtomicWrite(aw).ToByteArray();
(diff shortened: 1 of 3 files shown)


src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
```diff
@@ -236,10 +298,13 @@ private object GetPersistentFSMSnapshot(Type type, byte[] bytes)
                 timeout = TimeSpan.FromMilliseconds(message.TimeoutMillis);
             }
 
+            var payload = GetPayload(message.Data);
+
             // use reflection to create the generic type of PersistentFSM.PersistentFSMSnapshot
-            Type[] types = { typeof(string), type.GenericTypeArguments[0], typeof(TimeSpan?) };
-            object[] arguments = { message.StateIdentifier, GetPayload(message.Data), timeout };
+            Type[] types = { typeof(string), payload.GetType(), typeof(TimeSpan?) };
```
Arkatufus, line 304: Get the type of the payload from the payload type instead of relying on a generic passed into the method

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
```diff
@@ -148,18 +169,54 @@ private PersistentFSMSnapshot GetPersistentFSMSnapshot(object obj)
...
-            if (type == typeof(Persistent)) return GetPersistentRepresentation(PersistentMessage.Parser.ParseFrom(bytes));
-            if (type == typeof(IPersistentRepresentation)) return GetPersistentRepresentation(PersistentMessage.Parser.ParseFrom(bytes));
-            if (type == typeof(AtomicWrite)) return GetAtomicWrite(bytes);
-            if (type == typeof(AtLeastOnceDeliverySnapshot)) return GetAtLeastOnceDeliverySnapshot(bytes);
-            if (type == typeof(PersistentFSM.StateChangeEvent)) return GetStateChangeEvent(bytes);
-            if (type.GetTypeInfo().IsGenericType
-                && type.GetGenericTypeDefinition() == typeof(PersistentFSM.PersistentFSMSnapshot<>)) return GetPersistentFSMSnapshot(type, bytes);
+            switch (manifest)
+            {
+                case null:
```
Aaronontheweb, line 176: good catch on the `null`

src/core/Akka/Util/TypeExtensions.cs
```diff
@@ -58,13 +59,18 @@ public static bool Implements(this Type type, Type moreGeneralType)
...
         {
-            string shortened;
-            if (ShortenedTypeNames.TryGetValue(type, out shortened))
+            if (ShortenedTypeNames.TryGetValue(type, out var shortened))
             {
                 return shortened;
             }
 
-            shortened = cleanAssemblyVersionRegex.Replace(type.AssemblyQualifiedName, string.Empty);
+            // Defensive coding. `type.AssemblyQualifiedName` can return null if type is of generic type parameter type
```
Aaronontheweb, line 67: LGTM

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
```diff
@@ -148,18 +169,54 @@ private PersistentFSMSnapshot GetPersistentFSMSnapshot(object obj)
...
+                case PersistentFSMSnapshotManifest:
+                case var m when m.StartsWith(PersistentFSMSnapshotNetCoreManifest):
+                    return GetPersistentFSMSnapshot(bytes);
+            }
 
-            throw new SerializationException($"Unimplemented deserialization of message with type [{type}] in [{GetType()}]");
+            throw new SerializationException($"Unimplemented deserialization of message with manifest [{manifest}] in [{GetType()}]");
+        }
+
+        public override string Manifest(object obj)
```
Aaronontheweb, line 201: For the time being, we should return the legacy types - I believe there are some persistence plugins (i.e. SQL ones) that are still relying on manifest data using reflection and while I would like to eventually clean that up, probably best to change as little as possible here first.

Aaronontheweb (changes requested):
Minor backwards compat changes requested

src/core/Akka.Persistence/Serialization/PersistenceMessageSerializer.cs
```diff
@@ -203,18 +202,15 @@ public override string Manifest(object obj)
...
             {
                 case IPersistentRepresentation _:
-                    return IPersistentRepresentationManifest;
                 case AtomicWrite _:
-                    return AtomicWriteManifest;
                 case AtLeastOnceDeliverySnapshot _:
-                    return AtLeastOnceDeliverySnapshotManifest;
                 case PersistentFSM.StateChangeEvent _:
-                    return StateChangeEventManifest;
+                    return obj.GetType().TypeQualifiedName();
```
Arkatufus, line 208: Would something like this be OK?

Arkatufus: I'd have problem with `Akka.Persistence.Fsm.PersistentFSM.PersistentFSMSnapshot` because it is a generic class, hard coding would not work

Issue #5010, opened 2021-05-12
when using TestScheduler, tasks that are added during the same exact tick - might get ignored


brah-mcdude:
the following code is not thread safe:
```C#
        if (!_scheduledWork.TryGetValue(scheduledTime, out var tickItems))
        { 
            /// multiple threads *might* get in here
            /// multiple tickItems *might* be getting created (one per thread) 
            tickItems = new ConcurrentQueue<ScheduledItem>(); 

            /// the first tickItems for a certain scheduledTime will get added. 
            /// the rest of them will NOT get added. 
            _scheduledWork.TryAdd(scheduledTime, tickItems); 
        }

        var type = message == null ? ScheduledItem.ScheduledItemType.Action : ScheduledItem.ScheduledItemType.Message;

        /// tickItems might have just been created BUT NOT added to _scheduledWork
        /// so tickItems might be an orphan reference (about to get garbage collected after this line
        tickItems.Enqueue(new ScheduledItem(initialDelay ?? delay, delay, type, message, action,
            initialDelay.HasValue || deliveryCount > 0, receiver, sender, cancelable));
```
repro:
```C#
                 // call FSM<TState, TData>.SetTimer for a certain actor-system multiple times FROM DIFFERENT THREADS (simultaneously) aiming at the same tick
thread A: FSM<TState, TData>.SetTimer("timerName1", "msg1", TimeSpan.FromTicks(10)); // the first call TryAdd succeeds
thread B: FSM<TState, TData>.SetTimer("timerName1", "msg1", TimeSpan.FromTicks(10)); // TryAdd will fail and tickItems IS an orphan reference 
```

brah-mcdude:
pre-fix work around (this also serves as a proof of the issue and the fix):
```C#
        internal void SetTimer2(string timerName, MessageBase msg, TimeSpan timeout, bool repeat = false)
        {
            System.Diagnostics.Debug.Assert(!IsTimerActive(name: timerName));

            /// the following is a 100% work-around for tests that use TestScheduler
            lock (ActorSystem)
            {
                /// try to avoid multiple threads performing simultaneous _scheduledWork.TryGetValue 

                SetTimer(timerName, msg, timeout, repeat);
            }
        }
```

simple potential fix (using goto):
```C#
        try_again:
            if (!_scheduledWork.TryGetValue(scheduledTime, out var tickItems))
            {
                tickItems = new ConcurrentQueue<ScheduledItem>();
[...]

PR #5011, opened 2021-05-13
Update TestScheduler.cs

brah-mcdude:
fixed bug #5010

### DIFF
--- a/src/core/Akka.TestKit/TestScheduler.cs
+++ b/src/core/Akka.TestKit/TestScheduler.cs
@@ -88,10 +88,14 @@ namespace Akka.TestKit
         {
             var scheduledTime = _now.Add(initialDelay ?? delay).UtcTicks;
 
-            if (!_scheduledWork.TryGetValue(scheduledTime, out var tickItems))
+            ConcurrentQueue<ScheduledItem> tickItems;
+            while (!_scheduledWork.TryGetValue(scheduledTime, out tickItems))
             {
                 tickItems = new ConcurrentQueue<ScheduledItem>();
-                _scheduledWork.TryAdd(scheduledTime, tickItems);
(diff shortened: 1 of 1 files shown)


src/core/Akka.TestKit/TestScheduler.cs
```diff
@@ -88,10 +88,14 @@ public void AdvanceTo(DateTimeOffset when)
         {
             var scheduledTime = _now.Add(initialDelay ?? delay).UtcTicks;
 
-            if (!_scheduledWork.TryGetValue(scheduledTime, out var tickItems))
+            ConcurrentQueue<ScheduledItem> tickItems;
+            while (!_scheduledWork.TryGetValue(scheduledTime, out tickItems))
```
Aaronontheweb, line 92: so this is compare-and-swap basically @brah-mcdude ?

Aaronontheweb: Ah ok, I think I get this now after reading the original issue on https://github.com/akkadotnet/akka.net/issues/5010

brah-mcdude: tickItems is a queue of all the items that we want to schedule for a specific scheduledTime tick. 
We need to extract this queue from the dictionary of all _scheduledWork. 
If TryGetValue fails, then we need to create tickItems and add it to _scheduledWork. 
However, if by the time that we call TryAdd, someone else has already added a new tickItems at the scheduledTime tick entry, our call to TryAdd will fail.  In such a scenario - we need to start over.
In other words, we need to loop while TryGetValue fails.  
Actually, our thread's SECOND attempt to TryAdd MUST ALWAYS succeed. 
So, theoretically, we could re-write it as a 3 iterations for loop: (int i = 0; i < 3; i++) that will throw an exception if i reaches 3.

PR #5013, opened 2021-05-14
Fix SonarQube's "IEnumerable LINQs should be simplified" / Code Smell

marodev:
Hello,

This pull request is _very simple_ and aims to reduce the technical debt. It specifically targets the "IEnumerable LINQs should be simplified" [rule](https://rules.sonarsource.com/csharp/RSPEC-2971)  in SonarQube.

Let me know if you are interested in similar contributions.

Best,
Martin


Aaronontheweb:
Thanks for your contribution!

PR #5014, opened 2021-05-14
Make sure BackoffSupervisorSpec specs are not racy anymore, enable skipped tests.

Arkatufus:


src/core/Akka.Tests/Pattern/BackoffSupervisorSpec.cs
```diff
@@ -90,15 +91,30 @@ public static Props Props(IActorRef probe)
...
         #endregion
 
-        [Fact(Skip = "Racy on Azure DevOps")]
+        [Fact]
         public void BackoffSupervisor_must_start_child_again_when_it_stops_when_using_Backoff_OnStop()
         {
-            var supervisor = Create(OnStopOptions());
-            supervisor.Tell(BackoffSupervisor.GetCurrentChild.Instance);
-            var c1 = ExpectMsg<BackoffSupervisor.CurrentChild>().Ref;
+            var supervisor = Create(OnStopOptions().WithManualReset());
```
Arkatufus, line 97: Supervisor props need to be forced to use manual reset to make sure that it doesn't auto reset the internal restart count mid-spec and fails the spec.

src/core/Akka.Tests/Pattern/BackoffSupervisorSpec.cs
```diff
@@ -345,22 +361,22 @@ internal class DelayTable : IEnumerable<object[]>
             IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
         }
 
-        [Fact(Skip = "Racy on Azure DevOps")]
+        [Fact]
         public void BackoffSupervisor_must_stop_restarting_the_child_after_reaching_maxNrOfRetries_limit_using_BackOff_OnStop()
         {
-            var supervisor = Create(OnStopOptions(maxNrOfRetries: 2));
+            var supervisor = Create(OnStopOptions(maxNrOfRetries: 2).WithManualReset());
```
Arkatufus, line 367: Supervisor props need to be forced to use manual reset to make sure that it doesn't auto reset the internal restart count mid-spec and fails the spec.

src/core/Akka.Tests/Pattern/BackoffSupervisorSpec.cs
```diff
@@ -372,46 +388,52 @@ IActorRef WaitForChild()
...
         }
 
-        [Fact(Skip = "Racy on Azure DevOps")]
+        [Fact]
         public void BackoffSupervisor_must_stop_restarting_the_child_after_reaching_maxNrOfRetries_limit_using_BackOff_OnFailure()
         {
             EventFilter.Exception<TestException>().Expect(3, () =>
             {
-                var supervisor = Create(OnFailureOptions(maxNrOfRetries: 2));
+                var supervisor = Create(OnFailureOptions(maxNrOfRetries: 2).WithManualReset());
```
Arkatufus, line 424: Supervisor props need to be forced to use manual reset to make sure that it doesn't auto reset the internal restart count mid-spec and fails the spec.

src/core/Akka.Tests/Pattern/BackoffSupervisorSpec.cs
```diff
@@ -90,15 +91,30 @@ public static Props Props(IActorRef probe)
...
+        [Fact]
         public void BackoffSupervisor_must_start_child_again_when_it_stops_when_using_Backoff_OnStop()
         {
-            var supervisor = Create(OnStopOptions());
-            supervisor.Tell(BackoffSupervisor.GetCurrentChild.Instance);
-            var c1 = ExpectMsg<BackoffSupervisor.CurrentChild>().Ref;
+            var supervisor = Create(OnStopOptions().WithManualReset());
+
+            IActorRef c1 = null;
+            AwaitCondition(() =>
```
Arkatufus, line 100: Child might not be immediately available, have to wait until it actually is

src/core/Akka.Tests/Pattern/BackoffSupervisorSpec.cs
```diff
@@ -90,15 +91,30 @@ public static Props Props(IActorRef probe)
...
+                return !c1.IsNobody();
+            });
+            c1.Should().NotBeNull();
+
             Watch(c1);
             c1.Tell(PoisonPill.Instance);
-            ExpectTerminated(c1);
+            ExpectMsg<Terminated>().ActorRef.Should().Be(c1);
+
+            AwaitAssert(() =>
```
Arkatufus, line 112: There is no guarantee that the supervisor will observe the child termination immediately. We know that it should, sometime in the very near future, so we need to await this.

Aaronontheweb (approved):
LGTM

Arkatufus:
Still racy somehow, need to recheck the code

src/core/Akka.Tests/Pattern/BackoffSupervisorSpec.cs
```diff
@@ -372,46 +392,52 @@ IActorRef WaitForChild()
             var c1 = ExpectMsg<BackoffSupervisor.CurrentChild>().Ref;
             Watch(c1);
             c1.Tell(PoisonPill.Instance);
-            ExpectTerminated(c1);
```
Arkatufus, line 375: This can rarely fail, need to track down the real cause because its not supposed to happen.

Aaronontheweb: How can that fail? If `c1` processes the `PosionPill` it has to die? Is this taking longer than 3 seconds?

Arkatufus: No, the test received a Terminated message but from another actor, which is really weird.
It only happened once every 1000 test or so iteration, making it hard to debug.

Aaronontheweb: ah ok - so if we're death watching two actors the easiest way to solve that race condition is to create two different test probes, one to death watch each actor. No overlap that way. Although I expect in this case what makes it hard is that it's two different children from the same parent being watched, right?

Arkatufus: That might be a solution, will have to explore it a bit more

PR #5019, opened 2021-05-17
Turn on MiscMessageSerializer LocalScope message serialization.

Arkatufus:
`LocalScope` was supposed to be serialized using `MiscMessageSerializer`. The code is there but it is not turned on on the `Akka.Remote` configurations file.
Is there a reason why we're not using it?

### DIFF
--- a/src/core/Akka.Remote.Tests/Serialization/MiscMessageSerializerSpec.cs
+++ b/src/core/Akka.Remote.Tests/Serialization/MiscMessageSerializerSpec.cs
@@ -142,7 +142,7 @@ namespace Akka.Remote.Tests.Serialization
             AssertEqual(poisonPill);
         }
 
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
@Arkatufus is it because local scope is redundant? is there a case where we'd need to serialize local scope, since those actors are deployed locally by definition?

Arkatufus:
@Aaronontheweb I'm not sure, it is there in the MiscMessageSerializer code, spec, and configuration, but it is commented out in spec and configuration. It is using the default JSON serializer right now.

Aaronontheweb:
@Arkatufus but it is actually handled via the `misc` serializer now, right?

Arkatufus:
In this PR, yes. It is using NewtonSoft JSON in our current code.

PR #5020, opened 2021-05-17
Fix AkkaProtocolStressTest spec

Arkatufus:


src/core/Akka.Remote.Tests/Transport/AkkaProtocolStressTest.cs
```diff
@@ -190,7 +192,12 @@ public void AkkaProtocolTransport_must_guarantee_at_most_once_delivery_and_messa
                         new FailureInjectorTransportAdapter.Drop(0.1, 0.1)));
             AwaitCondition(() => mc.IsCompleted && mc.Result, TimeSpan.FromSeconds(3));
 
-            var here = Here;
+            IActorRef here = null;
+            AwaitCondition(() =>
```
Arkatufus, line 196: `Here` property uses `ActorSelection` that can return a null `Subject` that causes NRE down the line (caused by `EmptyLocalActorRef`). Using `AwaitCondition` here to force the spec to wait until `Here` returns a valid `IActorRef`

Aaronontheweb: LGTM - how many times did you run this one successively without failures?

Arkatufus: 1000 times minimum

Aaronontheweb: That works for me

PR #5021, opened 2021-05-18
Increase timeout value to avoid timeout failure

Arkatufus:
With a 2 second initial delay, the spec is squeazing it too close to the default default timeout of 3 seconds, increased the timeout to avoid sporadic spec fail.


Aaronontheweb (approved):
LGTM

PR #4990, opened 2021-04-29
adding ORSet benchmarks

Aaronontheweb:
Working on https://github.com/akkadotnet/akka.net/issues/4956

PR #5028, opened 2021-05-21
Improve MurmurHash string hash memory footprint

### LINKED ISSUE #5022: Akka.DistributedData: memory leak when recovering events from LMDB data store
Version: Akka.NET v1.4.19

Reproduction: https://github.com/andyfurnival/ddata 

Memory consumption grows steadily in a 2 node cluster, with both nodes recovering events from their own LMDB data stores:

image

The drivers of memory consumption appear to be non-stop gossip between the replicators:

image

I think there must be an equality check that is failing somewhere in this system, causing the same objects to be gossiped over and over again. We're not even producing any new events other than what's been deserialized from LMDB in this sample and yet the memory grows continuously.

Arkatufus:
Closes #5023

Edit by @Aaronontheweb : Removed the reference to #5022 since that has other issues still affecting it

### DIFF
--- a/src/contrib/cluster/Akka.DistributedData/VersionVector.cs
+++ b/src/contrib/cluster/Akka.DistributedData/VersionVector.cs
@@ -52,13 +52,13 @@ namespace Akka.DistributedData
         /// <summary>
         /// Marker to signal that we have reached the end of a version vector.
         /// </summary>
-        private static readonly KeyValuePair<UniqueAddress, long> EndMarker = new KeyValuePair<UniqueAddress, long>(null, long.MinValue);
+        private static readonly (UniqueAddress, long) EndMarker = (null, long.MinValue);
 
         public abstract bool IsEmpty { get; }
 
         public abstract int Count { get; }
 
-        public abstract IEnumerator<KeyValuePair<UniqueAddress, long>> VersionEnumerator { get; }
+        public abstract IEnumerator<(UniqueAddress, long)> VersionEnumerator { get; }
         public static readonly VersionVector Empty = new MultiVersionVector(ImmutableDictionary<UniqueAddress, long>.Empty);
 
         /// <summary>
@@ -163,11 +163,11 @@ namespace Akka.DistributedData
         private T NextOrElse<T>(IEnumerator<T> enumerator, T defaultValue) =>
             enumerator.MoveNext() ? enumerator.Current : defaultValue;
 
-        private Ordering Compare(IEnumerator<KeyValuePair<UniqueAddress, long>> i1,
-            IEnumerator<KeyValuePair<UniqueAddress, long>> i2, Ordering requestedOrder)
+        private Ordering Compare(IEnumerator<(UniqueAddress, long)> i1,
+            IEnumerator<(UniqueAddress, long)> i2, Ordering requestedOrder)
         {
-            var nt1 = NextOrElse(i1, EndMarker);
-            var nt2 = NextOrElse(i2, EndMarker);
+            var nt1 = NextOrElse<(UniqueAddress Key, long Value)>(i1, EndMarker);
+            var nt2 = NextOrElse<(UniqueAddress Key, long Value)>(i2, EndMarker);
             var currentOrder = Ordering.Same;
             while (true)
             {
@@ -214,13 +214,13 @@ namespace Akka.DistributedData
     [DebuggerDisplay("VersionVector({Node}->{Version})")]
     public sealed class SingleVersionVector : VersionVector
     {
-        private sealed class Enumerator : IEnumerator<KeyValuePair<UniqueAddress, long>>
+        private sealed class Enumerator : IEnumerator<(UniqueAddress, long)>
         {
             private bool _moved = false;
 
             public Enumerator(UniqueAddress node, long version)
             {
-                Current = new KeyValuePair<UniqueAddress, long>(node, version);
+                Current = (node, version);
             }
 
             /// <inheritdoc/>
@@ -241,7 +241,7 @@ namespace Akka.DistributedData
                 _moved = false;
             }
 
-            public KeyValuePair<UniqueAddress, long> Current { get; }
+            public (UniqueAddress, long) Current { get; }
 
             object IEnumerator.Current => Current;
         }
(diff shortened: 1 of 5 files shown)


src/core/Akka/Util/MurmurHash.cs
```diff
@@ -202,20 +202,20 @@ public static int StringHash(string s)
         {
             unchecked
             {
-                var sChar = s.ToCharArray();
```
Arkatufus, line 205: This is the biggest offending code. MurmurHash is called once for every DataEnvelope that are going to be sent as a gossip. For a system that are trying to gossip thousands of gossips, this can add up quite significantly.

src/contrib/cluster/Akka.DistributedData/VersionVector.cs
```diff
@@ -336,7 +378,7 @@ public MultiVersionVector(ImmutableDictionary<UniqueAddress, long> nodeVersions)
 
         public override bool IsEmpty => Versions.IsEmpty;
         public override int Count => Versions.Count;
-        public override IEnumerator<KeyValuePair<UniqueAddress, long>> VersionEnumerator => Versions.GetEnumerator();
```
Arkatufus, line 339: This is the second memory offender. Calling ImmutableDictionary GetEnumerator seemed to create a shallow copy of some of its internal data structure, consuming memory everytime it is called.

Fix by converting this to a Tuple IEnumerator and our own much simpler IEnumerator implementation.

Arkatufus:
There is a third memory offender, the serialization byte array allocation, but I have yet found a good way to fix this.

Aaronontheweb:
@Arkatufus can you do a before and after snapshot on this?

Arkatufus:
Would a dotMemory screenshot be enough?

src/core/Akka/Actor/Address.cs
```diff
@@ -40,7 +40,7 @@ public int Compare(Address x, Address y)
                 if (result != 0) return result;
                 result = string.CompareOrdinal(x.System, y.System);
                 if (result != 0) return result;
-                result = string.CompareOrdinal(x.Host ?? "", y.Host ?? "");
+                result = string.CompareOrdinal(x.Host ?? string.Empty, y.Host ?? string.Empty);
```
Aaronontheweb, line 43: LGTM

src/contrib/cluster/Akka.DistributedData/VersionVector.cs
```diff
@@ -257,7 +257,7 @@ public SingleVersionVector(UniqueAddress node, long version)
 
         public override bool IsEmpty => false;
         public override int Count => 1;
-        public override IEnumerator<KeyValuePair<UniqueAddress, long>> VersionEnumerator => new Enumerator(Node, Version);
```
Aaronontheweb, line 260: Is this a breaking API change or is this a subtype of an `internal` type above?

Aaronontheweb: Yeah, both of these are breaking changes.

I thought about doing this in https://github.com/akkadotnet/akka.net/pull/4990 - but I decided to avoid the breaking API change by exposing an internal API that should be used for this.

I don't think the risks are that high of this breaking change having a significant negative impact as DData is either primarily used directly or inside of cluster sharding and these APIs aren't called directly... but it's a judgement call.

Aaronontheweb (commented):
What do you think about the breaking API change trade-offs?

src/core/Akka/Util/MurmurHash.cs
Aaronontheweb: Before

|              Method |      Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------- |----------:|---------:|---------:|-------:|------:|------:|----------:|
|  Murmur_string_hash |  60.87 ns | 1.240 ns | 2.389 ns | 0.0229 |     - |     - |      96 B |
| Jenkins_string_hash | 221.14 ns | 0.767 ns | 0.680 ns | 0.0153 |     - |     - |      64 B |
|  Murmur_binary_hash | 166.72 ns | 2.177 ns | 2.036 ns |      - |     - |     - |         - |
| Jenkins_binary_hash | 370.40 ns | 0.709 ns | 0.554 ns |      - |     - |     - |         - |

Aaronontheweb:
``` ini

BenchmarkDotNet=v0.13.0, OS=Windows 10.0.19041.985 (2004/May2020Update/20H1)
.NET SDK=5.0.203
  [Host]     : .NET Core 3.1.15 (CoreCLR 4.700.21.21202, CoreFX 4.700.21.21402), X64 RyuJIT
  DefaultJob : .NET Core 3.1.15 (CoreCLR 4.700.21.21202, CoreFX 4.700.21.21402), X64 RyuJIT


```
|              Method |      Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------- |----------:|---------:|---------:|-------:|------:|------:|----------:|
|  Murmur_string_hash |  40.39 ns | 0.082 ns | 0.077 ns |      - |     - |     - |         - |
| Jenkins_string_hash | 218.54 ns | 0.467 ns | 0.390 ns | 0.0153 |     - |     - |      64 B |
|  Murmur_binary_hash | 166.06 ns | 2.224 ns | 2.080 ns |      - |     - |     - |         - |
| Jenkins_binary_hash | 372.00 ns | 2.132 ns | 1.664 ns |      - |     - |     - |         - |

Aaronontheweb: That is definitely a big improvement.

Arkatufus:
Well, this is highly annoying. I can't reproduce the bug using the original reproduction code anymore, memory consuption stays between 100-200 Mb no matter what I do.

Aaronontheweb:
Is the frequent gossip still occurring?

Arkatufus:
Yes, its still occuring, the console is scrolling endlessly, but the memory total didn't go above 200 Mb

Aaronontheweb:
@Arkatufus let's get this PR cleaned up for merge so we can benefit from the improved memory footprint. We can still work on tracking down the gossip issues in another PR.

Aaronontheweb (approved):
LGTM

PR #4874, opened 2021-03-24
Clean up Akka.Stream file stream

Arkatufus:

PR #5032, opened 2021-05-22
Fix a potential race condition in FileSubscriber

ismaelhamed:
Port [#19594](https://github.com/akka/akka/pull/19594)

This basically does the same thing @Arkatufus did in #4874 but uses the original code. Up to you guys.

### DIFF
--- a/src/core/Akka.Streams/Implementation/IO/FileSubscriber.cs
+++ b/src/core/Akka.Streams/Implementation/IO/FileSubscriber.cs
@@ -13,7 +13,6 @@ using Akka.Event;
 using Akka.IO;
 using Akka.Streams.Actors;
 using Akka.Streams.IO;
-using Akka.Util;
 
 namespace Akka.Streams.Implementation.IO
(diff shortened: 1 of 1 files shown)


src/core/Akka.Streams/Implementation/IO/FileSubscriber.cs
```diff
@@ -176,8 +166,8 @@ protected override bool Receive(object message)
                     }
                     catch (Exception ex)
                     {
-                        _log.Error(ex, $"Tearing down FileSink({_f.FullName}). File flush failed.");
-                        _completionPromise.TrySetResult(IOResult.Failed(_bytesWritten, ex));
+                        _log.Error(ex, "Tearing down FileSink({0}). File flush failed.", _f.FullName);
```
ismaelhamed, line 169: This allows loggers like Serilog to properly apply the output template.

Aaronontheweb (approved):
LGTM

PR #5033, opened 2021-05-24
Consolidate intercept

ismaelhamed:
Currently there are a few `Intercept` methods included in `AkkaSpec`, but they lack cohesion in their implementations. This PR aims at solving this by consolidating all of them under a single one that also overcome their biggest limitation: not unwrapping `AggregateExceptions`, which force users to deal with these inside the tests themselves. 

This PR also brings along the `AssertThrows` method (a proper implementation of the `Intercept` that returns `void` instead), which should be the preferred  between the two when no further inspection of the caught exception is needed.

Hopefully this will make it porting unit tests more straight forward and better aligned with the JVM.

### DIFF
--- a/src/core/Akka.Tests.Shared.Internals/AkkaSpec.cs
+++ b/src/core/Akka.Tests.Shared.Internals/AkkaSpec.cs
@@ -113,7 +113,7 @@ namespace Akka.TestKit
         }
 
         public static Config AkkaSpecConfig { get { return _akkaSpecConfig; } }
-        
+
         protected T ExpectMsgPf<T>(TimeSpan? timeout, string hint, Func<object, T> function)
         {
             MessageEnvelope envelope;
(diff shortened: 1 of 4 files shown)


src/core/Akka.Tests.Shared.Internals/AkkaSpec.cs
```diff
@@ -136,30 +136,88 @@ protected T ExpectMsgPf<T>(string hint, Func<object, T> pf)
...
             }
-            throw new ThrowsException(typeof(Exception));
+
+            throw new ThrowsException(typeof(T));
         }
-        
-        protected async Task InterceptAsync(Func<Task> asyncActionThatThrows)
+
+        [Obsolete("Use AssertThrows instead.")]
+        protected void Intercept(Action actionThatThrows)
```
ismaelhamed, line 216: Keeping this one since there are a bunch of tests using it, as a catch-all-exceptions, which is dubious at best.

ismaelhamed:
I meant to open this PR as a Draft, consider it as such.

Aaronontheweb (approved):
LGTM

PR #5044, opened 2021-05-27
[MNTR] fix test dll loading error

Arkatufus:


src/core/Akka.MultiNodeTestRunner.Shared/Sinks/MessageSink.cs
```diff
@@ -89,7 +89,7 @@ public enum MultiNodeTestRunnerMessageType
         };
 
         private const string NodePassStatusRegexString =
-            @"\[(\w){4}(?<node>[0-9]{1,2})(?<role>:\w+)?\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\]\[(?<status>(PASS|FAIL))\]{1}\s(?<test>.*)";
+            @"\[(\w){4}(?<node>[0-9]{1,2})(?<role>:[a-zA-Z0-9-]+)?\]\[(?<time>\d{1,4}[- /.]\d{1,4}[- /.]\d{1,4}\s\d{1,2}:\d{1,2}:\d{1,2}(\s(AM|PM)){0,1})\]\[(?<status>(PASS|FAIL))\]{1}\s(?<test>.*)";
```
Aaronontheweb, line 92: Why this change?

Arkatufus: it fails when role contains a number or a dash, some of the test have the role as "node-1"

Arkatufus: Fails as in the regex won't recognize the message type as a pass/fail message

Aaronontheweb: Looks like the MNTR internals suite is failing due to this now

Aaronontheweb:
LGTM

PR #5045, opened 2021-05-27
[MNTR] Add ActorSystemSetup support

### LINKED ISSUE #5027: No way to provide IServiceCollection to MultiNodeClusterSpec created ActorSystem
 1.4.20
 net5

Hey guys, I've been digging MNTR testing and having a hard time providing IServiceCollection to **MultiNodeClusterSpec** derived classes since **MultiNodeClusterSpec** constructor only accepts Config, no **ActorSystemSetup**.

It seems manual creating of ActorSystem is possible here:

https://github.com/akkadotnet/akka.net/blob/b5f6552a719cb75686686af69e8744f77bf47e3f/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs#L394-L410

But this is not:
https://github.com/akkadotnet/akka.net/blob/b5f6552a719cb75686686af69e8744f77bf47e3f/src/core/Akka.Cluster.TestKit/MultiNodeClusterSpec.cs#L151-L158

Am I missing something? is there any way other than copy-pasta MultiNodeClusterSpec.cs and reimplement it?

Arkatufus:
Closes #5027

### DIFF
--- a/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
+++ b/src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
@@ -17,6 +17,7 @@ using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading.Tasks;
 using Akka.Actor;
+using Akka.Actor.Setup;
 using Akka.Configuration;
 using Akka.Configuration.Hocon;
 using Akka.Event;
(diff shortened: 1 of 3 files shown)


src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -109,7 +109,7 @@ protected TestKitBase(ITestKitAssertions assertions, Config config, string actor
         {
         }
 
-        private TestKitBase(ITestKitAssertions assertions, ActorSystem system, ActorSystemSetup config, string actorSystemName, string testActorName)
+        protected TestKitBase(ITestKitAssertions assertions, ActorSystem system, ActorSystemSetup config, string actorSystemName, string testActorName)
```
Aaronontheweb, line 112: LGTM

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
```diff
@@ -401,7 +402,26 @@ public static Config BaseConfig
...
+            ImmutableList<RoleName> roles,
+            Func<RoleName, ImmutableList<string>> deployments)
+            : this(myself, null, setup, roles, deployments)
+        {
+        }
+
+        private MultiNodeSpec(
+            RoleName myself,
+            ActorSystem system,
+            ActorSystemSetup setup,
```
Aaronontheweb, line 421: Got it - so this is really an option - you can pass in either a full-blown `ActorSystem` or an `ActorSystemSetup` but only one of them gets used - the former if it's not `null`.-

Aaronontheweb (approved):
LGTM

src/core/Akka.Remote.TestKit/MultiNodeSpec.cs
Arkatufus: Yes, exactly.

PR #5051, opened 2021-05-28
Removed `StringBuilder` allocation from `Base64Encoding`

Aaronontheweb:
## Before (ActorSelection + Ask)

``` ini

BenchmarkDotNet=v0.13.0, OS=Windows 10.0.19041.985 (2004/May2020Update/20H1)
.NET SDK=5.0.203
  [Host]     : .NET Core 3.1.15 (CoreCLR 4.700.21.21202, CoreFX 4.700.21.21402), X64 RyuJIT
  DefaultJob : .NET Core 3.1.15 (CoreCLR 4.700.21.21202, CoreFX 4.700.21.21402), X64 RyuJIT


```
|                        Method | Iterations |       Mean |     Error |    StdDev |     Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------------ |----------- |-----------:|----------:|----------:|----------:|------:|------:|----------:|
| RequestResponseActorSelection |      10000 | 100.723 ms | 2.0080 ms | 5.6308 ms | 5000.0000 |     - |     - |     20 MB |
|          CreateActorSelection |      10000 |   7.564 ms | 0.2305 ms | 0.6795 ms |  953.1250 |     - |     - |      4 MB |

## After (ActorSelection + Ask)

``` ini

BenchmarkDotNet=v0.13.0, OS=Windows 10.0.19041.985 (2004/May2020Update/20H1)
.NET SDK=5.0.203
  [Host]     : .NET Core 3.1.15 (CoreCLR 4.700.21.21202, CoreFX 4.700.21.21402), X64 RyuJIT
  DefaultJob : .NET Core 3.1.15 (CoreCLR 4.700.21.21202, CoreFX 4.700.21.21402), X64 RyuJIT


```
|                        Method | Iterations |      Mean |     Error |    StdDev |     Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------------ |----------- |----------:|----------:|----------:|----------:|------:|------:|----------:|
| RequestResponseActorSelection |      10000 | 89.490 ms | 1.2584 ms | 0.9824 ms | 4666.6667 |     - |     - |     19 MB |
|          CreateActorSelection |      10000 |  5.445 ms | 0.0385 ms | 0.0342 ms |  953.1250 |     - |     - |      4 MB |

### DIFF
--- a/src/core/Akka/Util/Base64Encoding.cs
+++ b/src/core/Akka/Util/Base64Encoding.cs
@@ -32,6 +32,8 @@ namespace Akka.Util
 
         public static string Base64Encode(this long value, string prefix)
         {
+            // 11 is the number of characters it takes to represent long.MaxValue
+            // so we will never need a larger size for encoding longs
             Span<char> sb = stackalloc char[11 + prefix?.Length ?? 0];
             var spanIndex = 0;
             if (!string.IsNullOrEmpty(prefix) && prefix.Length > 0)
--- a/src/core/Akka/Util/Base64Encoding.cs
+++ b/src/core/Akka/Util/Base64Encoding.cs
@@ -5,6 +5,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System;
 using System.Text;
 
 namespace Akka.Util
@@ -24,7 +25,30 @@ namespace Akka.Util
         /// </summary>
         /// <param name="value">TBD</param>
         /// <returns>TBD</returns>
-        public static string Base64Encode(this long value) => Base64Encode(value, new StringBuilder()).ToString();
+        public static string Base64Encode(this long value)
+        {
+            return Base64Encode(value, "");
+        }
+
+        public static string Base64Encode(this long value, string prefix)
+        {
+            Span<char> sb = stackalloc char[11 + prefix?.Length ?? 0];
+            var spanIndex = 0;
+            if (!string.IsNullOrEmpty(prefix) && prefix.Length > 0)
+            {
+                prefix.AsSpan().CopyTo(sb);
+                spanIndex = prefix.Length;
+            }
+
+            var next = value;
+            do
+            {
+                var index = (int)(next & 63);
+                sb[spanIndex++] = Base64Chars[index];
+                next = next >> 6;
+            } while (next != 0);
+            return sb.Slice(0, spanIndex).ToString();
+        }
 
         /// <summary>
         /// TBD
@@ -32,6 +56,7 @@ namespace Akka.Util
         /// <param name="value">TBD</param>
         /// <param name="sb">TBD</param>
         /// <returns>TBD</returns>
+        [Obsolete("Do not use. Pass in prefix as a string instead.")]
         public static StringBuilder Base64Encode(this long value, StringBuilder sb)
         {
             var next = value;
--- a/src/core/Akka/Util/Base64Encoding.cs
+++ b/src/core/Akka/Util/Base64Encoding.cs
@@ -30,13 +30,13 @@ namespace Akka.Util
             return Base64Encode(value, "");
         }
 
-        public static string Base64Encode(this long value, string prefix)
+        internal static string Base64Encode(this long value, string prefix)
         {
             // 11 is the number of characters it takes to represent long.MaxValue
             // so we will never need a larger size for encoding longs
             Span<char> sb = stackalloc char[11 + prefix?.Length ?? 0];
             var spanIndex = 0;
-            if (!string.IsNullOrEmpty(prefix) && prefix.Length > 0)
+            if (!string.IsNullOrWhiteSpace(prefix) && prefix.Length > 0)
             {
                 prefix.AsSpan().CopyTo(sb);
                 spanIndex = prefix.Length;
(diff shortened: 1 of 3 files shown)


src/core/Akka/Util/Base64Encoding.cs
```diff
@@ -24,14 +25,38 @@ public static class Base64Encoding
...
         /// <returns>TBD</returns>
-        public static string Base64Encode(this long value) => Base64Encode(value, new StringBuilder()).ToString();
+        public static string Base64Encode(this long value)
+        {
+            return Base64Encode(value, "");
+        }
+
+        public static string Base64Encode(this long value, string prefix)
+        {
+            Span<char> sb = stackalloc char[11 + prefix?.Length ?? 0];
```
Aaronontheweb, line 35: Testing program is here:

https://gist.github.com/Aaronontheweb/449d565754d5f9bce475653842317277

`11` is the number of characters it takes to represent `long.MaxValue` in base64 encoded strings

src/core/Akka/Util/Base64Encoding.cs
```diff
@@ -24,14 +25,40 @@ public static class Base64Encoding
         /// </summary>
         /// <param name="value">TBD</param>
         /// <returns>TBD</returns>
-        public static string Base64Encode(this long value) => Base64Encode(value, new StringBuilder()).ToString();
+        public static string Base64Encode(this long value)
+        {
+            return Base64Encode(value, "");
+        }
+
+        public static string Base64Encode(this long value, string prefix)
```
to11mtm, line 33: After some sleep, I think we may want to mark this version as `internal`, since mis-use of long prefix strings will result in StackOverflowException, and negative `value` inputs will result in an `indexOutOfRange` exception.

I guess the old version has the second hazard, not sure if we should mark this all internal in that case.

src/core/Akka/Util/Base64Encoding.cs
```diff
@@ -24,14 +25,40 @@ public static class Base64Encoding
...
+        {
+            // 11 is the number of characters it takes to represent long.MaxValue
+            // so we will never need a larger size for encoding longs
+            Span<char> sb = stackalloc char[11 + prefix?.Length ?? 0];
+            var spanIndex = 0;
+            if (!string.IsNullOrEmpty(prefix) && prefix.Length > 0)
```
to11mtm, line 39: Is there a reason for checking both IsNullOrEmpty as well as length on the prefix? 

Also, would `.IsNullOrWhiteSpace` be a better choice? (idk?)

to11mtm (commented):
Just a couple little things.

src/core/Akka/Util/Base64Encoding.cs
Aaronontheweb: Yep, you're correct on both points - I'll go fix it.

src/core/Akka/Util/Base64Encoding.cs
Aaronontheweb: > Is there a reason for checking both IsNullOrEmpty as well as length on the prefix?

I thought it was easier to read than using the null-coalescing operator

Aaronontheweb: Changed to `IsNullOrWhiteSpace` since the latter is illegal and we already check for `string.Empty` with the length check.

src/core/Akka/Util/Base64Encoding.cs
```diff
@@ -24,14 +25,40 @@ public static class Base64Encoding
...
+            {
+                prefix.AsSpan().CopyTo(sb);
+                spanIndex = prefix.Length;
+            }
+
+            var next = value;
+            do
+            {
+                var index = (int)(next & 63);
+                sb[spanIndex++] = Base64Chars[index];
```
to11mtm, line 49: I wonder, is there any advantage at this point to changing Base64Chars to `ReadOnlyMemory`?

to11mtm (approved):

src/core/Akka/Util/Base64Encoding.cs
Aaronontheweb: I don't think so since it's basically just integer accessors

src/core/Akka/Util/Base64Encoding.cs
```diff
@@ -24,14 +25,40 @@ public static class Base64Encoding
...
+            return sb.Slice(0, spanIndex).ToString();
+        }
 
         /// <summary>
         /// TBD
         /// </summary>
         /// <param name="value">TBD</param>
         /// <param name="sb">TBD</param>
         /// <returns>TBD</returns>
+        [Obsolete("Do not use. Pass in prefix as a string instead.")]
```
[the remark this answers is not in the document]
Aaronontheweb: Still useful since this method really shouldn't have been exposed as `public` anyway, IMHO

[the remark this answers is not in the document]
Aaronontheweb: We had a bad habit in the very early days of Akka.NET making a lot of the internals public "in case someone wanted to play with them!" and we've been gradually undoing that in our minor version releases....

PR #5052, opened 2021-05-30
Bring UnfoldResourceSourceAsync up-to-date

ismaelhamed:


src/core/Akka.Streams/Dsl/Source.cs
```diff
@@ -930,7 +930,7 @@ public static class Source
         /// <param name="close">function that closes resource</param>
         /// <returns>TBD</returns>
         public static Source<T, NotUsed> UnfoldResourceAsync<T, TSource>(Func<Task<TSource>> create,
-            Func<TSource, Task<Option<T>>> read, Func<TSource, Task> close)
+            Func<TSource, Task<Option<T>>> read, Func<TSource, Task<Done>> close)
```
ismaelhamed, line 933: Breaking change, but the signature was wrong.

src/core/Akka/Util/Internal/TaskExtensions.cs
```diff
@@ -40,5 +40,55 @@ public static Task WithCancellation(this Task task, CancellationToken cancellati
...
                 .ContinueWith(t => { }, cancellationToken, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
         }
+
+        /// <summary>
+        /// When this Task is completed, either through an exception or a value, invoke the provided function.
+        /// If the Task has already been completed, this will either be applied immediately or be scheduled asynchronously.
+        /// </summary>
+        /// <param name="source">TBD</param>
+        /// <param name="f">The function to be executed when this Task completes</param>
+        public static Task OnComplete(this Task source, Action<Try<Done>> f)
```
ismaelhamed, line 50: I understand these extensions are not idiomatic in C#, but working with `ContinueWith` is cumbersome when you have to repeat a chunk of code like this one over and over again.

Maybe porting more of these extensions from the JVM wouldn't be such a bad idea after all.

src/core/Akka.Streams/Dsl/Source.cs
Aaronontheweb: Not a big deal, IMHO

src/core/Akka/Util/Internal/TaskExtensions.cs
Aaronontheweb: This stuff is all marked as part of our internal APIs anyways - so users shouldn't expect that this would be part of the public-facing API that we preserve. If these make things easier, we should include them.

Aaronontheweb (approved):
LGTM

Aaronontheweb:
Really appreciate you dealing with the racy specs in particular. We need to do a lot more of that on the Akka.Streams test suite still unfortunately.

PR #5053, opened 2021-05-31
Reduced allocations around `ActorCell` and `Task` handling

Aaronontheweb:
* Made `ActorCell.SendSystemMessage` `virtual` so it can be subclassed
* Made a number of `ActorCell` methods `static` to reduce object memory footprint
* Reduced number of possible `Task` allocations where they weren't needed

### DIFF
--- a/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
+++ b/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
@@ -401,7 +401,7 @@ namespace Akka.Actor
             }
         }
 
-        private void HandleSupervise(IActorRef child, bool async)
+        private static void HandleSupervise(IActorRef child, bool async)
         {
             if (async && child is RepointableActorRef @ref)
             {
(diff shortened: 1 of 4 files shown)


src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -506,11 +506,11 @@ public void Suspend()
...
 
         /// <summary>
-        /// TBD
+        /// Handles a <see cref="ISystemMessage"/>
         /// </summary>
         /// <remarks>➡➡➡ NEVER SEND THE SAME SYSTEM MESSAGE OBJECT TO TWO ACTORS ⬅⬅⬅</remarks>
-        /// <param name="systemMessage">TBD</param>
-        public void SendSystemMessage(ISystemMessage systemMessage)
+        /// <param name="systemMessage">The system message to process.</param>
+        public virtual void SendSystemMessage(ISystemMessage systemMessage)
```
Aaronontheweb, line 513: Working on adding the ability to properly trace `async` / `await` flows within an actor using [Phobos](http://phobos.petabridge.com/) but as it turns out, it's very difficult to instrument that right now since all internal `await` operations flow 100% through the `ActorCell` and don't touch any of the `SendSystemMessage` methods on the `IActorRef` implementations - hence why I've made this `virtual`.

System messages are infrequently sent so the overhead of this change (virtual method dispatch overhead) should be trivial.

Aaronontheweb (commented):
Described my changes - this PR is still in draft until I can make sure that this solves some of our correlation issues inside Phobos

Arkatufus (approved):
Code looks good

PR #5056, opened 2021-06-01
Fix DData infinite status/gossip round

### LINKED ISSUE #5022: Akka.DistributedData: memory leak when recovering events from LMDB data store

Arkatufus:
Closes #5022

### DIFF
--- a/src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
+++ b/src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
@@ -53,9 +53,11 @@ namespace Akka.DistributedData.LightningDB
         private readonly Akka.Serialization.Serialization _serialization;
         private readonly SerializerWithStringManifest _serializer;
         private readonly string _manifest;
+        private readonly long _mapSize;
 
         private readonly TimeSpan _writeBehindInterval;
         private readonly string _dir;
+        private bool _dirExists;
 
         private readonly Dictionary<string, DurableDataEnvelope> _pending = new Dictionary<string, DurableDataEnvelope>();
         private readonly ILoggingAdapter _log;
@@ -80,11 +82,16 @@ namespace Akka.DistributedData.LightningDB
                     TimeSpan.Zero :
                     _config.GetTimeSpan("write-behind-interval");
 
+            _mapSize = _config.GetByteSize("map-size") ?? 100 * 1024 * 1024;
+
             var path = _config.GetString("dir");
             _dir = path.EndsWith(DatabaseName)
                 ? Path.GetFullPath($"{path}-{Context.System.Name}-{Self.Path.Parent.Name}-{Cluster.Cluster.Get(Context.System).SelfAddress.Port}")
                 : Path.GetFullPath(path);
 
+            _dirExists = Directory.Exists(_dir);
+
+            _log.Info($"Using durable data in LMDB directory [{_dir}]");
             Init();
         }
 
@@ -103,19 +110,17 @@ namespace Akka.DistributedData.LightningDB
 
         private LightningEnvironment GetLightningEnvironment()
         {
-            _log.Info($"Using durable data in LMDB directory [{_dir}]");
-
-            var mapSize = _config.GetByteSize("map-size");
             LightningEnvironment env;
 
-            if (!Directory.Exists(_dir))
+            if (!_dirExists)
             {
                 var t0 = Stopwatch.StartNew();
                 Directory.CreateDirectory(_dir);
+                _dirExists = true;
 
                 env = new LightningEnvironment(_dir)
                 {
-                    MapSize = mapSize ?? 100 * 1024 * 1024,
+                    MapSize = _mapSize,
                     MaxDatabases = 1
                 };
                 env.Open(EnvironmentOpenFlags.NoLock);
@@ -134,7 +139,7 @@ namespace Akka.DistributedData.LightningDB
             {
                 env = new LightningEnvironment(_dir)
                 {
-                    MapSize = mapSize ?? 100 * 1024 * 1024,
+                    MapSize = _mapSize,
                     MaxDatabases = 1
                 };
                 env.Open(EnvironmentOpenFlags.NoLock);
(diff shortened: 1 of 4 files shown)


src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
```diff
@@ -80,11 +82,16 @@ public LmdbDurableStore(Config config)
...
+            _mapSize = _config.GetByteSize("map-size") ?? 100 * 1024 * 1024;
+
             var path = _config.GetString("dir");
             _dir = path.EndsWith(DatabaseName)
                 ? Path.GetFullPath($"{path}-{Context.System.Name}-{Self.Path.Parent.Name}-{Cluster.Cluster.Get(Context.System).SelfAddress.Port}")
                 : Path.GetFullPath(path);
 
+            _dirExists = Directory.Exists(_dir);
+
+            _log.Info($"Using durable data in LMDB directory [{_dir}]");
```
Arkatufus, line 94: Move logging here, should be called once at startup, not every time a new LMDB environment is created

src/contrib/cluster/Akka.DistributedData/Serialization/Proto/OtherMessage.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Akka.DistributedData.Serialization.Proto.Msg
+{
+    internal sealed partial class OtherMessage : IComparable<OtherMessage>
```
Arkatufus, line 7: Needed to allow `OtherMessage` can be sorted

src/contrib/cluster/Akka.DistributedData/Serialization/ReplicatedDataSerializer.cs
```diff
@@ -251,57 +251,77 @@ private static Type GetTypeFromDescriptor(TypeDescriptor t)
...
                 case ORSet<int> ints:
                     {
-                        var p = ORSetToProto(ints);
-                        p.TypeInfo.Type = ValType.Int;
-                        p.IntElements.Add(ints.Elements);
-                        return p;
+                        b.Vvector =  SerializationSupport.VersionVectorToProto(ints.VersionVector);
+                        b.TypeInfo.Type = ValType.Int;
+                        var intElements = new List<int>(ints.ElementsMap.Keys);
+                        intElements.Sort();
```
Arkatufus, line 273: ORSet elements need to be sorted during serialization so that it have consistent SHA-1 digest

src/contrib/cluster/Akka.DistributedData/Serialization/ReplicatedDataSerializer.cs
```diff
@@ -251,57 +251,77 @@ private static Type GetTypeFromDescriptor(TypeDescriptor t)
...
+                        foreach (var val in intElements)
+                        {
+                            b.IntElements.Add(val);
+                            b.Dots.Add(SerializationSupport.VersionVectorToProto(ints.ElementsMap[val]));
```
Arkatufus, line 277: ElementsMap isn't serialized as a dictionary, instead keys and version vector are serialized as separate lists. Due to this serialization scheme, key ordering and value ordering matters. VersionVector need to be inserted in the same order as the newly sorted keys.

src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
```diff
@@ -80,11 +82,16 @@ public LmdbDurableStore(Config config)
                     TimeSpan.Zero :
                     _config.GetTimeSpan("write-behind-interval");
 
+            _mapSize = _config.GetByteSize("map-size") ?? 100 * 1024 * 1024;
```
Aaronontheweb, line 85: This was always the default size, so no big deal.

src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
```diff
@@ -53,9 +53,11 @@ private sealed class WriteBehind : IDeadLetterSuppression
         private readonly Akka.Serialization.Serialization _serialization;
         private readonly SerializerWithStringManifest _serializer;
         private readonly string _manifest;
+        private readonly long _mapSize;
 
         private readonly TimeSpan _writeBehindInterval;
         private readonly string _dir;
+        private bool _dirExists;
```
Aaronontheweb, line 60: Good idea to cache this

src/contrib/cluster/Akka.DistributedData/Serialization/Proto/OtherMessage.cs
Aaronontheweb: Would it be easier to just write an [`IComparer`](https://docs.microsoft.com/en-us/troubleshoot/dotnet/csharp/use-icomparable-icomparer)? So we don't have to hack around a generated class?

Aaronontheweb (changes requested):
One small change

src/contrib/cluster/Akka.DistributedData/Serialization/OtherMessageComparer.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Akka.DistributedData.Serialization.Proto.Msg;
+
+namespace Akka.DistributedData.Serialization
+{
+    internal class OtherMessageComparer : IComparer<OtherMessage>
```
Aaronontheweb, line 8: We should make this `static` and have a `readonly` singleton here, but we can easily do that in an update

Aaronontheweb (approved):
LGTM

src/contrib/cluster/Akka.DistributedData/Serialization/OtherMessageComparer.cs
Aaronontheweb: Ah, you already did that inside the serializer

PR #5059, opened 2021-06-02
Revert `ThreadPool.SetMinThreads(0,0)`

Aaronontheweb:
Based on user feedback on https://github.com/akkadotnet/akka.net/discussions/4983


boris-rg (approved):

Aaronontheweb:
I have no idea why Linux craps out on every other build during compilation - has to be an issue with the VM.

christallire:
the document needs to be reverted too. https://getakka.net/articles/actors/dispatchers.html

Aaronontheweb:
@christallire good point

PR #5072, opened 2021-06-04
Fix Akka.Remote serialization exception bubbling and messages

### LINKED ISSUE #5062: Akka.Remote: `MessageContainerSerializer` throws misleading error message when can't deserialize payload
**Version Information**
Version of Akka.NET?  v1.4.20
Which Akka.NET Modules? Akka.Remote

**Describe the bug**
When deserializing an `ActorSelection` mesage that contains an unserializable payload, we throw an error message that looks like this:

```
[2021/06/02 16:56:59.792] [Error] [Akka.Actor.OneForOneStrategy] Exception has been thrown by the target of an invocation.
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
---> System.Runtime.Serialization.SerializationException: Cannot find serializer with id [13] (manifest [AA]). The most probable reason is that the configuration entry 'akka.actor.serializers' is not in sync between the two systems.
at Akka.Serialization.Serialization.Deser
```

This gets interpreted as "could not find serializer with Id=6", which is not true.

**To Reproduce**

1. Deserialize `ActorSelectionMessage` with unserializable payload (non-registered payload)
2. Log error message

**Expected behavior**
Should clearly state that it was the wrapped message for which deserialization failed, not the higher level `MessageConatinerSerializer`.

**Environment**
Windows and Linux

Arkatufus:
Closes #5062

### DIFF
--- a/src/core/Akka.Remote.Tests/Akka.Remote.Tests.csproj
+++ b/src/core/Akka.Remote.Tests/Akka.Remote.Tests.csproj
@@ -8,6 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\contrib\cluster\Akka.DistributedData\Akka.DistributedData.csproj" />
     <ProjectReference Include="..\Akka.Remote\Akka.Remote.csproj" />
     <ProjectReference Include="..\Akka.Tests.Shared.Internals\Akka.Tests.Shared.Internals.csproj" />
 
--- a/src/core/Akka.Remote.Tests/Akka.Remote.Tests.csproj
+++ b/src/core/Akka.Remote.Tests/Akka.Remote.Tests.csproj
@@ -8,7 +8,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\contrib\cluster\Akka.DistributedData\Akka.DistributedData.csproj" />
     <ProjectReference Include="..\Akka.Remote\Akka.Remote.csproj" />
     <ProjectReference Include="..\Akka.Tests.Shared.Internals\Akka.Tests.Shared.Internals.csproj" />
 
--- a/src/core/Akka.Remote.Tests/Serialization/BugFix5062Spec.cs
+++ b/src/core/Akka.Remote.Tests/Serialization/BugFix5062Spec.cs
@@ -39,22 +39,20 @@ namespace Akka.Remote.Tests.Serialization
         [Fact]
         public void Failed_serialization_should_give_proper_exception_message()
         {
+            var childName = "dummy";
             var message = new ActorSelectionMessage(
                 new SomeMessage(), 
-                new SelectionPathElement[] { new SelectChildName("dummy") }, 
+                new SelectionPathElement[] { new SelectChildName(childName) }, 
                 true);
 
             var node1 = new Address("akka.tcp", "Sys", "localhost", 2551);
             var serialized = MessageSerializer.Serialize((ExtendedActorSystem)Sys, node1, message);
 
             var o = new object();
-            var ex = o.Invoking(s => MessageSerializer.Deserialize((ExtendedActorSystem)Sys, serialized)).Should()
+            o.Invoking(s => MessageSerializer.Deserialize((ExtendedActorSystem)Sys, serialized)).Should()
                 .Throw<SerializationException>()
-                .WithMessage("Failed to deserialize object with serialization id [6] (manifest []).")
-                //.WithInnerExceptionExactly<SerializationException>()
-                //.WithMessage("Failed to deserialize object with serialization id [11] (manifest [E]).")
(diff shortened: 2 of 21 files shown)


src/core/Akka.Remote.Tests/Akka.Remote.Tests.csproj
```diff
@@ -8,6 +8,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\contrib\cluster\Akka.DistributedData\Akka.DistributedData.csproj" />
```
Aaronontheweb, line 11: Isn't this a circular reference?

Aaronontheweb: Can we reproduce this error without using that? Like via a custom serializer that doesn't exist?

Arkatufus: It shouldn't be

Aaronontheweb: It is - it depends on Akka.Cluster. Remove this.

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -429,6 +429,11 @@ public object Deserialize(byte[] bytes, int serializerId, string manifest)
 
                 return serializer.FromBinary(bytes, type);
             }
+            catch (Exception e)
+            {
+                throw new SerializationException(
```
Aaronontheweb, line 434: This isn't really the right spot for this. Should be in the Akka.Remote serializer for wrapped payload support. We already handle this exception in the inner `try` block

Aaronontheweb (changes requested):
Not fixed

src/core/Akka/Serialization/Serialization.cs
Arkatufus: Actually, yes, it is the right spot for it, because the inner try..catch catches the exception thrown from failed manifest lookup, the outer exception catches the exception from the actual serializer deserialization. If it is not placed here then we have to actually put a try..catch on every single `FromBinary` function on every serializer.

Declaring it here actually makes debugging a very pleasant experience because all of the exceptions will be cascaded up the stack call from the top most call to the inner most calls when it is iterated.

Arkatufus: Basically, this fixes all exception rethrows for all serializers, not just Akka.Remote

Aaronontheweb: Ok

Aaronontheweb (approved):

src/core/Akka.Remote.Tests/Serialization/BugFix5062Spec.cs
```diff
@@ -0,0 +1,90 @@
...
+                new SelectionPathElement[] { new SelectChildName("dummy") }, 
+                true);
+
+            var node1 = new Address("akka.tcp", "Sys", "localhost", 2551);
+            var serialized = MessageSerializer.Serialize((ExtendedActorSystem)Sys, node1, message);
+
+            var o = new object();
+            var ex = o.Invoking(s => MessageSerializer.Deserialize((ExtendedActorSystem)Sys, serialized)).Should()
+                .Throw<SerializationException>()
+                .WithMessage("Failed to deserialize object with serialization id [6] (manifest []).")
```
Aaronontheweb, line 53: This literally reproduces the error we want to get rid of in https://github.com/akkadotnet/akka.net/issues/5062

src/core/Akka.Remote.Tests/Serialization/BugFix5062Spec.cs
```diff
@@ -0,0 +1,90 @@
...
+                //.WithInnerExceptionExactly<SerializationException>()
+                //.WithMessage("Failed to deserialize object with serialization id [11] (manifest [E]).")
+                .WithInnerExceptionExactly<SerializationException>()
+                .WithMessage("Failed to deserialize object with serialization id [13] (manifest [SM]).");
```
Aaronontheweb, line 57: Your use case and reproduction here is wrong - need to have a scenario where `SomeMessage` isn't registered at all, not that it fails during deserialization.

The goal of this bug is to unpack the inner exception and throw _that_ - so the user gets a clear error message: that the payload carried by an `ActorSelection` couldn't be found because of a missing manifest, not that the `ActorSelection` _itself_ couldn't be deserialized.

Aaronontheweb (changes requested):
Addresses a totally different problem and introduces new ones.

src/core/Akka/Serialization/Serialization.cs
Aaronontheweb: Actually no - this is not a good fix.

PR #5077, opened 2021-06-07
Use TimestampProvider in QueryExecutor.InsertBatchAsync.

### LINKED ISSUE #5076: TimestampProvider not working or not being used
**Version Information**
Version of Akka.NET? 
1.4.19

Which Akka.NET Modules?
Akka.Persistence.Sql.Common.Journal

**Describe the bug**
I implemented a custom `ITimestampProvider` for SQL Persistence Journal.
I put a breakpoint in `GenerateTimestamp()` and let this method return 0. 
But every time I persist an entity I keep seeing DateTime.UtcNow.Ticks used as TimeStamp.

**To Reproduce**
1. Added customer `ITimestampProvider`. Let method return 0.
2. Add config:
` persistence : {
      journal : {
        plugin : akka.persistence.journal.sql-server
        sql-server : {
          timestamp-provider : "xx.xx.xx, xx.xx"
        }
      }`
3. Persist entity
4. Look in EventJournal table and see TimeStamp is a big long (not 0).

**Expected behavior**
Use long returned by custom `ITimestampProvider`.

**Actual behavior**
DateTime.UtcNow.Ticks is used as TimeStamp.

**Additional context**
I tried to debug this and found out that in `SqlServerJournal` line 62 is returning an instance of my own `TimetampProvider`. So that seems to work! 
https://github.com/akkadotnet/Akka.Persistence.SqlServer/blob/6f9eae1f0d17ab76adf60809e2a631d8acfa2e82/src/Akka.Persistence.SqlServer/Journal/SqlServerJournal.cs#L62

But another thing I saw in the source is that in `AbstractQueryExecutor` the TimestampProvider is passed as parameter, but only the property is set. It seems that the property is never used.
https://github.com/akkadotnet/akka.net/blob/55c117566bb7176351a4e2bc40e471e87f8eecfc/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs#L319

In InsertBatchAsync as timestamp is used 'DateTime.UtcNow.Ticks'. I should expect the TimestampProvider being used here?
https://github.com/akkadotnet/akka.net/blob/55c117566bb7176351a4e2bc40e471e87f8eecfc/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs#L687

**Screenshots**
Schermafbeelding 2021-06-07 om 13 33 48
Schermafbeelding 2021-06-07 om 13 30 55
Schermafbeelding 2021-06-07 om 13 30 23

MartijnSchoemaker:
Fixes #5076

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
@@ -684,7 +684,7 @@ namespace Akka.Persistence.Sql.Common.Journal
                     var evt = entry.Key;
                     var tags = entry.Value;
 
-                    WriteEvent(command, evt.WithTimestamp(DateTime.UtcNow.Ticks), tags);
+                    WriteEvent(command, evt.WithTimestamp(TimestampProvider.GenerateTimestamp(evt)), tags);
                     await command.ExecuteScalarAsync(cancellationToken);
                     command.Parameters.Clear();
                 }


Aaronontheweb (approved):
LGTM!

Aaronontheweb:
Thanks for your first Akka.NET contribution!

PR #5085, opened 2021-06-09
Add Blazor startup compat

Aaronontheweb:


Aaronontheweb:
Looks like this can't be solved without some major `ActorSystem` changes:

1. `Task.Result` is not supported, which we use in several places (i.e. logger startup)
2. `HashedWheelTimer` is not supported;
3. A number of built-in logging options, i.e. setting console colors, are not supported.

Are these issues that Unity3D or Xamarin users run into also?

to11mtm:
> 
> 
> Looks like this can't be solved without some major `ActorSystem` changes:
> 
>     1. `Task.Result` is not supported, which we use in several places (i.e. logger startup)
> 
>     2. `HashedWheelTimer` is not supported;
> 
>     3. A number of built-in logging options, i.e. setting console colors, are not supported.
> 
> 
> Are these issues that Unity3D or Xamarin users run into also?

I recall Xamarin users at least at one point had to work around some of the console color stuff.

There are two major hurdles for Blazor at this time, aside from the little bits like console colors:

 - Client side Blazor currently only runs a single thread. So `.Result` and `.Wait()` are not usable AFAIK.
 - Chrome has timer throttling, which results in timers being inaccurate https://github.com/dotnet/runtime/issues/51041 . This likely would have an impact on anything scheduled on a tight interval.

I believe Unity3D supports Threads and async/await (although I'm not certain what limitations, if any, exist around them).

Aaronontheweb:
Some of this will be addressed in https://github.com/akkadotnet/akka.net/issues/7246

PR #5087, opened 2021-06-10
return `T` from `IDependencyResolver.GetService<T>()`

### LINKED ISSUE #5086: `IDependencyResolver.GetService<T>()` is returning `object` rather than `<T>`
During upgrading to 1.4.21-beta1 we saw that `IDependencyResolver.GetService<T>()` is returning `object`. Is there a specific reason why it is not directly returning `<T>`?
Now we need to cast it to `T` before we can use the object.

_Originally posted by @wesselkranenborg in https://github.com/akkadotnet/akka.net/discussions/5070#discussioncomment-840243_

We should update the API to change this.

Aaronontheweb:
close #5086


Arkatufus (approved):
LGTM

PR #5091, opened 2021-06-14
cleaned up internal ActorCell and TestKit methods

Aaronontheweb:
Found some methods that could be made static on `ActorCell`, which should help with total actor memory footprint.


Arkatufus (approved):
LGTM

PR #5092, opened 2021-06-14
TestKit ReceiveWhile did not insert last inspected message properly.

Arkatufus:
While peeking through items inside the backing `BlockingCollection`, ReceiveWhile re-insert the last item it inspected in the last position instead of the first.

### DIFF
--- a/src/core/Akka.TestKit/Internal/BlockingQueue.cs
+++ b/src/core/Akka.TestKit/Internal/BlockingQueue.cs
@@ -22,7 +22,7 @@ namespace Akka.TestKit.Internal
     /// <typeparam name="T">The type of item to store.</typeparam>
     public class BlockingQueue<T>
     {
-        private readonly BlockingCollection<Positioned> _collection = new BlockingCollection<Positioned>();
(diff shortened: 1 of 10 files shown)


src/core/Akka.TestKit/Internal/BlockingQueue.cs
```diff
@@ -22,7 +22,7 @@ namespace Akka.TestKit.Internal
     /// <typeparam name="T">The type of item to store.</typeparam>
     public class BlockingQueue<T>
     {
-        private readonly BlockingCollection<Positioned> _collection = new BlockingCollection<Positioned>();
+        private readonly BlockingCollection<Positioned> _collection = new BlockingCollection<Positioned>(new QueueWithAddFirst());
```
Arkatufus, line 25: `QueueWithAddFirst` is the backing `IProducerConsumerCollection` for the internal `BlockingCollection` that allows proper item insertion in the head of the queue.

Aaronontheweb (approved):
LGTM - looks like we should have been using that positional queue this entire time.

PR #5093, opened 2021-06-15
GSet elements needs to be sorted during serialization

Arkatufus:
Just like ORSet, GSet elements need to be sorted during serialization so that the digest of any same GSets with the same elements with different element positioning will always be the same.


src/contrib/cluster/Akka.DistributedData/Serialization/OtherMessageComparer.cs
```diff
@@ -7,6 +7,11 @@ namespace Akka.DistributedData.Serialization
 {
     internal class OtherMessageComparer : IComparer<OtherMessage>
     {
+        public static OtherMessageComparer Instance { get; } = new OtherMessageComparer();
```
Aaronontheweb, line 10: Good idea

Aaronontheweb (approved):
LGTM

PR #5107, opened 2021-06-23
[FLAKY TESTS] Update and clean up FlowThrottleSpec

Arkatufus:


Aaronontheweb (approved):
LGTM - I agree that it's better to use `ReceiveWhile` than a custom abstraction.

PR #5108, opened 2021-06-23
Reduce AwaitAssert default interval to help tests that have tight time budgets

Arkatufus:


src/core/Akka.TestKit/TestKitBase_AwaitAssert.cs
```diff
@@ -33,7 +33,7 @@ public abstract partial class TestKitBase
         /// <param name="interval">The interval to wait between executing the assertion.</param>
         public void AwaitAssert(Action assertion, TimeSpan? duration=null, TimeSpan? interval=null)
         {
-            var intervalValue = interval.GetValueOrDefault(TimeSpan.FromMilliseconds(800));
```
Aaronontheweb, line 36: Wow, I had no idea the default value was this high....

Arkatufus: yeah, its a bit high

Aaronontheweb (approved):
LGTM

PR #5112, opened 2021-06-27
[ShardedDaemonProcess] Support for custom stop message

ismaelhamed:
These API changes allow to specify a custom `handOffMessage` that will be sent to all entities when they're supposed to stop themselves due to rebalance or cluster shutdown.


Aaronontheweb (approved):
LGTM

PR #5114, opened 2021-06-29
IsNobody for remote DeadLetters

Zetanova:
Remote Messages with a sender set to ActorRefs.NoSender should return true for IsNobody()


Aaronontheweb (approved):
LGTM

PR #5119, opened 2021-07-02
Make MessageContainerSerializer exception message more verbose.

Arkatufus:
Make MessageContainerSerializer exception message more verbose by concatenating the inner exception message


Aaronontheweb (approved):
Should merge this in with https://github.com/akkadotnet/akka.net/pull/5120

PR #5121, opened 2021-07-02
log full exception upon deserialization failure

Aaronontheweb:
We weren't capturing the exception thrown during deserialization anymore, just logging the error message. A throwback from before Akka.NET v1.4 when only the `ILoggingAdatper.Error` method accepted `Exception` arguments

### DIFF
--- a/src/core/Akka.Remote/Endpoint.cs
+++ b/src/core/Akka.Remote/Endpoint.cs
@@ -1500,7 +1500,7 @@ namespace Akka.Remote
             {
                 _log.Error(
                   ex,
-                  "Serializer not defined for message type [{0}]. Transient association error (association remains live)",
+                  "Serialization failed for message [{0}]. Transient association error (association remains live)",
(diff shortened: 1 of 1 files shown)


src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1500,7 +1500,7 @@ private bool WriteSend(EndpointManager.Send send)
             {
                 _log.Error(
                   ex,
-                  "Serializer not defined for message type [{0}]. Transient association error (association remains live)",
+                  "Serialization failed for message [{0}]. Transient association error (association remains live)",
```
Aaronontheweb, line 1503: Made this message less specific, since we don't necessarily know why an outbound write failed.

Aaronontheweb: (the inner exception should tell us)

Arkatufus (approved):
LGTM

PR #5123, opened 2021-07-04
Pipe of Canceled Tasks

Zetanova:
Canceled Task will have null as tresult.Exception.
The error handler need have a TaskCanceledException instance to indicate the cancellation.


Aaronontheweb (approved):
LGTM

Issue #5116, opened 2021-07-01
Cluster Sharding Serialiser occasionally failing when using Hyperion


andyfurnival:
**Version Information**
Version of Akka.NET? 
1.4.21
Which Akka.NET Modules? 
Akka.Cluster.Sharding
Akka.Persistence
Hyperion

**Describe the bug**
Occasionally, I'm getting exceptions on Coordinator and Shard replaying events (using Redis Persistence store), with the error message "System.Runtime.Serialization.SerializationException: Offset must be specified in whole minutes. (Parameter 'offset')" or "System.IndexOutOfRangeException: Index was outside the bounds of the array.". This is being thrown when the coordinator / shard is using Hyperion serialiser. I don't specify which serialiser to use for Cluster Sharding, its taken from the Default Config in Akka.Cluster.Sharding which looks to use Protobuf, but I do specify in the last fallback of my Akka configuration, my Journal and Snapshot serialisation which are all Protobuf

Full stack:
```
System.Runtime.Serialization.SerializationException: Offset must be specified in whole minutes. (Parameter 'offset')
 ---> System.ArgumentException: Offset must be specified in whole minutes. (Parameter 'offset')
   at System.DateTimeOffset.ValidateOffset(TimeSpan offset)
   at Hyperion.ValueSerializers.DateTimeOffsetSerializer.ReadDateTimeOffset(Stream stream, Byte[] bytes)
   at Hyperion.ValueSerializers.SessionAwareByteArrayRequiringValueSerializer`1.ReadValue(Stream stream, DeserializerSession session)
   at Hyperion.Serializer.Deserialize[T](Stream stream)
   at Akka.Serialization.HyperionSerializer.FromBinary(Byte[] bytes, Type type)
   --- End of inner exception stack trace ---
   at Akka.Serialization.HyperionSerializer.FromBinary(Byte[] bytes, Type type)
   at Akka.Serialization.Serializer.FromBinary[T](Byte[] bytes)
   at System.Linq.Enumerable.SelectArrayIterator`2.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Linq.OrderedEnumerable`1.TryGetFirst(Boolean& found)
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Boolean& found)
[...]

PR #5125, opened 2021-07-05
Fix persistence cluster sharding serialization failure

Arkatufus:
Related to #5116

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
@@ -295,7 +295,7 @@ namespace Akka.Cluster.Sharding
         /// TBD
         /// </summary>
         [Serializable]
-        public sealed class GetShardStats : IShardQuery
+        public sealed class GetShardStats : IShardQuery, IClusterShardingSerializable
         {
             /// <summary>
             /// TBD
@@ -311,7 +311,7 @@ namespace Akka.Cluster.Sharding
         /// TBD
         /// </summary>
         [Serializable]
-        public sealed class ShardStats
+        public sealed class ShardStats : IClusterShardingSerializable
         {
             /// <summary>
             /// TBD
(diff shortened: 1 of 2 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
```diff
@@ -295,7 +295,7 @@ public CurrentShardState(string shardId, IImmutableSet<string> entityIds)
         /// TBD
         /// </summary>
         [Serializable]
-        public sealed class GetShardStats : IShardQuery
+        public sealed class GetShardStats : IShardQuery, IClusterShardingSerializable
```
Aaronontheweb, line 298: Neither of these messages are persisted, so they couldn't cause the persistence failures included in https://github.com/akkadotnet/akka.net/issues/5116

Aaronontheweb (changes requested):
Does not fix the issue

src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
Arkatufus: They should be included anyway:
https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs#L155-L156

Aaronontheweb: Yep, that's a good fix but it doesn't close the issue.

Aaronontheweb: Do you want me to merge this issue separately?

Aaronontheweb: I thought the original header for this issue said "close #5116"

Arkatufus: no, its only related, found this problem while checking the serializer code

Aaronontheweb (approved):

PR #5126, opened 2021-07-05
working on #4823 redesign

Aaronontheweb:
Looking to replace PR https://github.com/akkadotnet/akka.net/pull/4861 with this one. Taking a closer look at some of our sync primitives around task scheduling.


src/core/Akka.TestKit/ActorCellKeepingSynchronizationContext.cs
```diff
@@ -58,27 +56,22 @@ public override void Post(SendOrPostCallback d, object state)
...
         {
-            var tcs = new TaskCompletionSource<int>();
-            Post(_ =>
+            var oldCell = InternalCurrentActorCellKeeper.Current;
+            var oldContext = Current;
+            SetSynchronizationContext(this);
+            InternalCurrentActorCellKeeper.Current = _cell;
+            try
             {
-                try
```
Aaronontheweb, line 71: Don't believe any of this code was ever necessary

Issue #5102, opened 2021-06-22
InvalidCastException in mixed-host cluster


evenbrenden:
**Version Information**
Version of Akka.NET? 

1.4.21

Which Akka.NET Modules?

Akka.Cluster

**Describe the bug**

As part of a seamless migration from a Windows VM cluster to a K8s cluster, we are running our Akka cluster on a mix of Windows and Linux hosts. The seed nodes are running on Windows hosts. The/a node running in the K8s cluster is able to establish a connection with the cluster, but crashes on a serialization error:

```2021-06-22 07:47:10 ERR Unable to cast object of type 'System.Int32' to type 'System.Int64'.
System.InvalidCastException: Unable to cast object of type 'System.Int32' to type 'System.Int64'.
   at lambda_method132(Closure , Stream , DeserializerSession )
   at Hyperion.ValueSerializers.ObjectSerializer.ReadValue(Stream stream, DeserializerSession session)
   at lambda_method130(Closure , Stream , DeserializerSession )
   at Hyperion.ValueSerializers.ObjectSerializer.ReadValue(Stream stream, DeserializerSession session)
   at Hyperion.Serializer.Deserialize[T](Stream stream)
   at Akka.Serialization.HyperionSerializer.FromBinary(Byte[] bytes, Type type)
   at Akka.Serialization.Serialization.Deserialize(Byte[] bytes, Int32 serializerId, String manifest)
   at Akka.Remote.MessageSerializer.Deserialize(ExtendedActorSystem system, Payload messageProtocol)
   at Akka.Remote.DefaultMessageDispatcher.Dispatch(IInternalActorRef recipient, Address recipientAddress, Payload message, IActorRef senderOption)
   at Akka.Remote.EndpointReader.<Reading>b__11_0(InboundPayload inbound)
   at lambda_method60(Closure , Object , Action`1 , Action`1 , Action`1 )
   at Akka.Actor.ReceiveActor.OnReceive(Object message)
   at Akka.Actor.UntypedActor.Receive(Object message)
   at Akka.Actor.ActorBase.AroundReceive(Receive receive, Object message)
   at Akka.Actor.ActorCell.ReceiveMessage(Object message)
   at Akka.Actor.ActorCell.Invoke(Envelope envelope)
--- End of stack trace from previous location ---
   at Akka.Actor.ActorCell.HandleFailed(Failed f)
[...]

PR #5134, opened 2021-07-07
Attempt to replicate Int64 to Int32 type mutation in mixed host cluster

Arkatufus:
Attempt to reproduce #5102 bug


object:
As @rogeralsing pointed oiut, the issue is caused by difference in ordering on different platforms.

I made a PR for Hyperion.
https://github.com/akkadotnet/Hyperion/pull/236

Arkatufus:
@object Noted and thank you for contributing :)

PR #5140, opened 2021-07-12
Akka.Cluster.Sharding should not use Hyperion serializer

Arkatufus:
Fishing for possible serializer problem with Akka.Cluster.Sharding. It should never use the default object serializer for any of its message types.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs
@@ -63,9 +63,11 @@ namespace Akka.Cluster.Sharding.Serialization
         private const string ShardStatsManifest = "DB";
         private const string GetShardRegionStatsManifest = "DC";
         private const string ShardRegionStatsManifest = "DD";
+        private const string GetCurrentRegionsManifest = "DG";
+        private const string CurrentRegionsManifest = "DH";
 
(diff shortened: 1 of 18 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/Serialization/ClusterShardingMessageSerializer.cs
```diff
@@ -224,6 +241,11 @@ public override string Manifest(object o)
                 case ShardRegionStats _: return ShardRegionStatsManifest;
                 case GetClusterShardingStats _: return GetClusterShardingStatsManifest;
                 case ClusterShardingStats _: return ClusterShardingStatsManifest;
+                case GetCurrentRegions _: return GetCurrentRegionsManifest;
+                case CurrentRegions _: return CurrentRegionsManifest;
+                case GetShardRegionState _: return GetShardRegionStateManifest;
+                case ShardState _: return ShardStateManifest;
```
Aaronontheweb, line 247: Oh boy, did we have a namespace overload here? I.e. `ShardState` and `Shard.ShardState`?

Arkatufus: Looks like it

Aaronontheweb (approved):
LGTM - but this will have some impact on users running Akka.Cluster.Sharding with Akka.Persistence. We need to offer a tool to purge all of the Cluster.Sharding data from the persistence store to make it easy to reset after a wire change like this.

Issue #5129, opened 2021-07-06
ddata replicator stops but doesn't look like it can be restarted easily


andyfurnival:
**Is your feature request related to a problem? Please describe.**
I  got this error raised in my system `Stopping distributed-data Replicator due to load or startup failure in durable store, caused by: "failed to load durable distributed-data"` which was caused by a sudden large spike in memory causing OOM. The issue was momentary, and whilst the rest of the actor system remained up and working (memory recovered), the ddata replicator never started back up.  This occurred on 2 of 7 nodes, so there was a trigger, but not cataclysmic event, which leads me to believe it may be possible to have a more fault tolerant supervisor strategy for the ddata replicator.

**Describe the solution you'd like**
looking at the code, I wonder if it can be improved to at least try again over a short period of time, maybe try 3 times over 30 seconds, to let the system get back into order?    We have this today for the supervisor
```
 protected override SupervisorStrategy SupervisorStrategy() => new OneForOneStrategy(e =>
        {
            var fromDurableStore = Equals(Sender, _durableStore) && !Equals(Sender, Context.System.DeadLetters);
            if ((e is LoadFailedException || e is ActorInitializationException) && fromDurableStore)
            {
                _log.Error(e,
                    "Stopping distributed-data Replicator due to load or startup failure in durable store, caused by: {0}",
                    e.Message);
                Context.Stop(Self);
                return Directive.Stop;
            }
            else return Actor.SupervisorStrategy.DefaultDecider.Decide(e);
        });
```

to something like this (may be something to make more configurable perhaps)
```
protected override SupervisorStrategy SupervisorStrategy() => new OneForOneStrategy(3, TimeSpan.FromSeconds(30), e =>
        {
            var fromDurableStore = Equals(Sender, _durableStore) && !Equals(Sender, Context.System.DeadLetters);
            if ((e is LoadFailedException || e is ActorInitializationException) && fromDurableStore)
            {
                _log.Error(e,
                    "Stopping distributed-data Replicator due to load or startup failure in durable store, caused by: {0}",
                    e.Message);
                Context.Stop(Self);
                return Directive.Stop;
            }
            else return Actor.SupervisorStrategy.DefaultDecider.Decide(e);
        });
```
[...]

PR #5143, opened 2021-07-14
ddata replicator stops but doesn't look like it can be restarted easily #5129

andyfurnival:
Wrapped DistributedData Replicator in a Backoff supervisor, so temporary durable store failures don't require a full system restart to recover.

I didn't add this directly into Replicator.Props as this changed the behaviour of many of the multi-node tests, and the trade-off of reworking tests for supporting a supervisor didn't make a whole lot of sense

I had considered putting the supervisor directly on the LmdbDurableStore, however we still run the risk of losing the Replicator with no means of recover (even when durable store isn't used)

Fixed a slightly annoying missing end bracket in ddata logs for GSet


src/contrib/cluster/Akka.DistributedData/DistributedData.cs
```diff
@@ -83,10 +85,23 @@ public DistributedData(ExtendedActorSystem system)
             else
             {
                 var name = config.GetString("name", null);
-                Replicator = system.ActorOf(Akka.DistributedData.Replicator.Props(_settings), name);
+                Replicator = system.ActorOf(GetSupervisedReplicator(_settings, name), name+"Supervisor");
             }
         }
 
+        private static Props GetSupervisedReplicator(ReplicatorSettings settings, string name) => BackoffSupervisor.Props(
```
Aaronontheweb, line 92: The impact this has on the actor path should not affect replication - HOWEVER, when introducing Akka.NET v1.4.22 into a cluster that is running v1.4.21, the replication system will break temporarily due to:

https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.DistributedData/Replicator.cs#L1104

Aaronontheweb: We should introduce a HOCON config setting:

```
akka.cluster.distributed-data.recreate-on-failure = off
```

When that setting is off, we don't use the `BackoffSupervisor` - and we will keep that setting to `off` by default. That way this change can be introduced in a non-breaking fashion. When you want to use the `BackoffSupervisor`, set this setting to `on`.
```

**Review by Aaronontheweb — APPROVED** [review]
See my comment about breaking changes during upgrades

File: src/contrib/cluster/Akka.DistributedData/DistributedData.cs [hunk 2]
```diff
@@ -83,10 +85,23 @@ public DistributedData(ExtendedActorSystem system)
             else
             {
                 var name = config.GetString("name", null);
-                Replicator = system.ActorOf(Akka.DistributedData.Replicator.Props(_settings), name);
+                Replicator = system.ActorOf(GetSupervisedReplicator(_settings, name), name+"Supervisor");
```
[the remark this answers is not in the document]
Arkatufus: Just to be clear, the new HOCON flag should switch between these 2 implementation, depending on its value, correct?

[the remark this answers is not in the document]
Aaronontheweb: Correct

Aaronontheweb:
I think the tests introduced here might be a little racy - probably need to allow more time for the first seed node to self-join

src/contrib/cluster/Akka.DistributedData/DistributedData.cs
andyfurnival: I've done that change, to default to the original implementation, with a setting to enable backoff.

PR #5144, opened 2021-07-14
Make ddata replicator actor configurable to have backoff supervisor

andyfurnival:
Make ddata replicator actor configurable to have backoff supervisor. Defaults to not (original implementation) 

The test maybe a little overkill, I just wanted to prove that the original implementation how users would have it today, behaved the same way, including the way it fails

### DIFF
--- a/src/contrib/cluster/Akka.DistributedData.Tests/ReplicatorResiliencySpec.cs
+++ b/src/contrib/cluster/Akka.DistributedData.Tests/ReplicatorResiliencySpec.cs
@@ -6,7 +6,6 @@
 // //-----------------------------------------------------------------------
 
 using System;
-using System.Collections.Immutable;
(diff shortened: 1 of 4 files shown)


Aaronontheweb:
Need to approve the changes to the public API https://getakka.net/community/public-api-changes.html

```
cd src/core/Akka.API.Tests
dotnet test -c Release --framework netcoreapp3.1`
```

andyfurnival:
I think @Arkatufus has done a similar change on #5145. I saw your comment to make it configurable, and assumed that was for me. If that other PR is good, i can decline this one

Arkatufus:
@andyfurnival Sure, I can merge your changes to mine.

Aaronontheweb:
@Arkatufus go ahead and do that

PR #5150, opened 2021-07-19
CircuitBreaker's open state now returns a faulted Task instead of throwing

### LINKED ISSUE #5117: CircuitBreaker's Open state should return a faulted Task instead of throwing
**Version Information**  
Akka 1.4.21

**Describe the bug**  
CircuitBreaker's `Open` and `HalfOpen` states should return a faulted `Task` with the `OpenCircuitException` exception instead of throwing.

**To Reproduce**

```csharp
circuitBreaker.WithCircuitBreaker(dangerousAsyncOperation)
    .PipeTo(Self, failure: exception => new Success.Failure(exception));
```

**Expected behavior**  
Be able to handle the `OpenCircuitException` with just the `PipeTo`, without needing to add another `ContinueWith` to the mix.

**Actual behavior**  
`OpenCircuitException` cannot be handled in the failure callback, and the actor restarts due to the unhadled exception.

**Proposed fix**  
Return a faulted `Task`:

```csharp
public override Task Invoke(Func<Task> body) => Task.FromException(
    new OpenCircuitException(_breaker.LastCaughtException, RemainingDuration()));

```

...instead of throwing

```csharp
public override Task Invoke(Func<Task> body) =>
    throw new OpenCircuitException(_breaker.LastCaughtException, RemainingDuration());
```

ismaelhamed:
Fixes #5117

### DIFF
--- a/src/core/Akka/Pattern/CircuitBreakerState.cs
+++ b/src/core/Akka/Pattern/CircuitBreakerState.cs
@@ -43,27 +43,21 @@ namespace Akka.Pattern
         }
 
         /// <summary>
-        /// N/A
+        /// Fail-fast on any invocation
         /// </summary>
         /// <typeparam name="T">N/A</typeparam>
         /// <param name="body">Implementation of the call that needs protected</param>
-        /// <exception cref="OpenCircuitException">This exception is thrown automatically since the circuit is open.</exception>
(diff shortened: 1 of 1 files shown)


Aaronontheweb (approved):
LGTM

PR #5152, opened 2021-07-24
extend boostrap-setup to register extensions on prestart

Zetanova:
Currently there is only the possibility to add extensions prestart only over the config `akka.extensions`
and external initialized extensions are only possible to add after startup.

This PR fixes it with an `BootstrapSetup.Extensions` array.

### DIFF
--- a/src/core/Akka/Actor/ActorSystem.cs
+++ b/src/core/Akka/Actor/ActorSystem.cs
@@ -16,6 +16,9 @@ using Akka.Event;
 using Akka.Util;
 using ConfigurationFactory = Akka.Configuration.ConfigurationFactory;
 using Config = Akka.Configuration.Config;
+using System.Collections.Immutable;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Akka.Actor
 {
@@ -101,16 +104,20 @@ namespace Akka.Actor
         internal BootstrapSetup()
             : this(
                 Option<Config>.None,
-                Option<ProviderSelection>.None)
+                Option<ProviderSelection>.None,
+                ImmutableArray<IExtensionId>.Empty)
         {
         }
 
         internal BootstrapSetup(
             Option<Config> config,
-            Option<ProviderSelection> actorRefProvider)
+            Option<ProviderSelection> actorRefProvider,
+            ImmutableArray<IExtensionId> extensions)
         {
             Config = config;
             ActorRefProvider = actorRefProvider;
+            Extensions = !extensions.IsDefault
+                ? extensions : ImmutableArray<IExtensionId>.Empty;
         }
 
         /// <summary>
@@ -126,6 +133,8 @@ namespace Akka.Actor
         /// </summary>
         public Option<ProviderSelection> ActorRefProvider { get; }
 
+        public ImmutableArray<IExtensionId> Extensions { get; }
+
         /// <summary>
         /// Create a new <see cref="BootstrapSetup"/> instance.
         /// </summary>
@@ -136,18 +145,35 @@ namespace Akka.Actor
 
         public BootstrapSetup WithActorRefProvider(ProviderSelection name)
         {
-            return new BootstrapSetup(Config, name);
+            return new BootstrapSetup(Config, name, Extensions);
         }
 
         public BootstrapSetup WithConfig(Config config)
         {
-            return new BootstrapSetup(config, ActorRefProvider);
+            return new BootstrapSetup(config, ActorRefProvider, Extensions);
         }
 
         public BootstrapSetup WithConfigFallback(Config config)
             => Config.HasValue
-                ? new BootstrapSetup(Config.Value.SafeWithFallback(config), ActorRefProvider)
+                ? new BootstrapSetup(Config.Value.SafeWithFallback(config), ActorRefProvider, Extensions)
                 : WithConfig(config);
+
+        public BootstrapSetup WithExtensions(IExtensionId extension)
+        {
+            return Extensions.Contains(extension) ? this
+                : new BootstrapSetup(Config, ActorRefProvider, Extensions.Add(extension));
+        }
+
+        public BootstrapSetup WithExtensions(params IExtensionId[] extensions)
+        {
+            return WithExtensions((IEnumerable<IExtensionId>)extensions);
+        }
+
+        public BootstrapSetup WithExtensions(IEnumerable<IExtensionId> extensions)
+        {
+            return new BootstrapSetup(Config, ActorRefProvider, 
+                Extensions.AddRange(extensions.Where(n => !Extensions.Contains(n))));
+        }
     }
 
     /// <summary>
--- a/src/core/Akka/Actor/Internal/ActorSystemImpl.cs
+++ b/src/core/Akka/Actor/Internal/ActorSystemImpl.cs
@@ -8,15 +8,16 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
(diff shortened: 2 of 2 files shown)


src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -320,12 +322,21 @@ private void LoadExtensions()
...
             }
         }
 
+        private void ConfigureExtensions(ActorSystemSetup setup)
+        {
+            var extensions = setup.Get<BootstrapSetup>()
+                .Select(_ => _.Extensions)
+                .GetOrElse(ImmutableArray<IExtensionId>.Empty);
+
+            ConfigureExtensions(extensions);
```
Aaronontheweb, line 337: So `ConfigureExtensions` will be called twice if there are some extensions defined in HOCON _and_ some via the `BootstrapSetup`, correct? And the `RegisterExtension` method is idempotent, so even if the same extension was registered twice there shouldn't be any issues?

Aaronontheweb (commented):
Looks good to me - but I did leave some questions on behavior and compatibility with the old HOCON

Aaronontheweb:
I think we ought to add a unit test for this

Zetanova:
I added new BootstrapSetupSpec and switched the DI Extension to BootstrapSetup.Extensions.
The DependencyResolverSetup and it's ExtensionIdProvider are having the same function.

Zetanova:
I am getting in a local test run a successful result,
but with the same error output

image

Zetanova:
Why do this unit test builds fail?

Aaronontheweb:
> Why do this unit test builds fail?

Nothing to do with your PR - some of these specs are racy due to being designed to fit a bounded schedule, which is inherently brittle, and we haven't gone through and fixed all of them.

Aaronontheweb:
@Zetanova I will review this soon!

src/contrib/dependencyinjection/Akka.DependencyInjection.Tests/DelegateInjectionSpecs.cs
```diff
@@ -120,8 +120,9 @@ public AkkaService(IServiceProvider serviceProvider)
 
             public Task StartAsync(CancellationToken cancellationToken)
             {
-                var setup = DependencyResolverSetup.Create(_serviceProvider)
-                    .And(BootstrapSetup.Create().WithConfig(TestKitBase.DefaultConfig));
+                var setup = BootstrapSetup.Create()
+                        .WithDependencyInjection(_serviceProvider)
```
Aaronontheweb, line 124: Nice.

src/contrib/dependencyinjection/Akka.DependencyInjection/DependencyResolver.cs
```diff
@@ -111,15 +111,43 @@ public Props Props(Type type, params object[] args)
     /// </summary>
     public sealed class DependencyResolverExtension : ExtensionIdProvider<DependencyResolver>
     {
+        readonly IDependencyResolver _dependencyResolver;
+
+        public DependencyResolverExtension()
+        {
+        }
+
+        public DependencyResolverExtension(IDependencyResolver dependencyResolver)
```
Aaronontheweb, line 120: You don't want this code on the `ExtensionIdProvider` - you want it on the `DependencyResolver` extension itself. The  `ExtensionIdProvider` is just meant to be a unique identifier that the `ActorSystem` uses to uniquely identify each possible extension to ensure there's at most one instance of it per-`ActorSystem`.

src/contrib/dependencyinjection/Akka.DependencyInjection/DependencyResolverSetup.cs
```diff
@@ -8,6 +8,23 @@
...
 using Akka.Actor.Setup;
+using Akka.DependencyInjection;
+
+namespace Akka.Actor
+{
+    public static class DependencyInjectionSetupExtensions
+    {
+        public static BootstrapSetup WithDependencyInjection(this BootstrapSetup bootstrap, IServiceProvider provider)
+        {
+            return bootstrap.WithExtensions(new DependencyResolverExtension(provider));
```
Aaronontheweb, line 19: I'd just combine setups here and return an `ActorSystemSetup` - this is what we do with Phobos, for instance: https://phobos.petabridge.com/articles/setup/configuration.html#phobossetup-and-phobosconfigbuilder

```csharp
var phobosSetup = PhobosSetup.Create(phobosConfigBuilder)
    .And(BootstrapSetup.Create()
        .WithConfig(hocon)
        .WithActorRefProvider(PhobosProviderSelection.Cluster));
```

The `PhobosSetup.Create` method actually return an `ActorSystemSetup` because it combines both a `TracingSetup` and a separate `MonitoringSetup` type internally.

Aaronontheweb: Using that type of composition is preferable to changing the `DependencyResolverExtension` - IMHO.

src/core/Akka.Tests.Shared.Internals/AkkaSpec.cs
```diff
@@ -67,6 +67,12 @@ public AkkaSpec(ActorSystemSetup setup, ITestOutputHelper output = null)
             BeforeAll();
         }
 
+        public AkkaSpec(Setup setup, ITestOutputHelper output = null)
```
Aaronontheweb, line 70: Good idea

src/core/Akka/Actor/ActorSystem.cs
```diff
@@ -126,6 +133,8 @@ internal BootstrapSetup()
         /// </summary>
         public Option<ProviderSelection> ActorRefProvider { get; }
 
+        public ImmutableArray<IExtensionId> Extensions { get; }
```
Aaronontheweb, line 136: Need to add an XML-DOC comment here

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
Aaronontheweb: In the event that there's a HOCON collection for extensions and a `BootstrapSetup`, what happens?

Aaronontheweb (changes requested):
See my suggestions around modifying the `DependencyResolverExtension` - otherwise this looks good.

Also - should add some tests to see what happens when extensions are defined both via HOCON and `BootstrapSetup`

Zetanova (commented):
> In the event that there's a HOCON collection for extensions and a BootstrapSetup, what happens?

The same as trying to add the same extension twice, first wins and the rest gets ignored.
The point of the new `BoostrapSetup.Extensions` list setup and register custom instances of `Extension` factories.

Zetanova (commented):
i didn't press the review button

src/core/Akka/Actor/ActorSystem.cs
Zetanova: If the ExtensionId (the extension factory instance) should not hold any state values / fields to setup the extensions,
would it not be better, to use just the Type? 
```csharp
public ImmutableArray<Type> Extensions { get; }
```

Zetanova:
I looked now at the `ActorSystemImpl._extensions` its a mess, needs to be redone.

It registers Lazy Factory Functions one-by-one (foreach) and instantly initializing it. 

It is absolutely pointless.

The only thing what can happen is that the extension constructor can query itself
because of the safe-execution protection of the Lazy instance.

PR #5155, opened 2021-07-26
Prevent EndpointDisassociatedException from being thrown from EndpointReader

Arkatufus:
Hyperion can throw an InvalidCastException when deserializing a message that came from a system that uses a different sorting order compared to the receiving system. This exception should be treated like a SerializationException and should not cause the EndpointReader to fail.

### DIFF
--- a/src/core/Akka.Remote/Endpoint.cs
+++ b/src/core/Akka.Remote/Endpoint.cs
@@ -1961,6 +1961,10 @@ namespace Akka.Remote
                             {
                                 LogTransientSerializationError(ackAndMessage.MessageOption, e);
                             }
+                            catch (InvalidCastException e)
+                            {
+                                LogTransientSerializationError(ackAndMessage.MessageOption, e);
+                            }
                             catch (Exception e)
                             {
                                 throw;


src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1961,6 +1961,10 @@ private void Reading()
                             {
                                 LogTransientSerializationError(ackAndMessage.MessageOption, e);
                             }
+                            catch (InvalidCastException e)
+                            {
+                                LogTransientSerializationError(ackAndMessage.MessageOption, e);
```
Aaronontheweb, line 1966: LGTM

zbynek001:
just curious, isn't this already handled here?
https://github.com/akkadotnet/akka.net/blob/a4e26fb0e48edee3844f96775971e1ae329be78c/src/contrib/serializers/Akka.Serialization.Hyperion/HyperionSerializer.cs#L134

Aaronontheweb:
> just curious, isn't this already handled here?
> https://github.com/akkadotnet/akka.net/blob/a4e26fb0e48edee3844f96775971e1ae329be78c/src/contrib/serializers/Akka.Serialization.Hyperion/HyperionSerializer.cs#L134

Good question - cc @Arkatufus ?

Arkatufus:
No, that catches `SerializationException`, this is a specific failure that throws `InvalidCastException` inside Hyperion, most probably due to the incompatible sorting order between systems.
`InvalidCastException` isn't caught inside the `StreamReader`, causing it to fail and die, which is quite bad.

Aaronontheweb:
But if you click into that gist, that `SerializationException` gets raised when _any_ `Exception` is thrown via the `HyperionSerializer` - I think that was @zbynek001's observation

Arkatufus:
Yes, but the same problem will happen to people who doesn't use the latest Hyperion version, InvalidCastException will get bubbled up uncaught, so this is more of a backward compatibility patch

Aaronontheweb:
@Arkatufus that's a good point - we used to not catch those errors inside `HyperionSerializer` up until a month ago.

Aaronontheweb:
In fact, we probably haven't released a new version of Akka.NET since that bug fix was introduced. Meant to do that last week but ran out of time before I had to go on leave.

PR #5156, opened 2021-07-26
Update Azure Pipelines .NET SDK version to 5.0.302 (latest)

Arkatufus:


Aaronontheweb:
@Arkatufus might need to bump a `.cs` file in order to get Incrementalist to run things

Aaronontheweb:
Looks like Azure Pipelines doesn't like this change....

Arkatufus:
Ok, will look into this further.

PR #5161, opened 2021-08-02
exposed `CreateRemoteRef` method overload for creating remotely-deployed actors

Aaronontheweb:
Needed this primarily for Phobos support - in order to correctly trace remotely deployed actors.


Arkatufus (approved):
LGTM

PR #5162, opened 2021-08-02
fixed racy Akka.DistributedData.Tests.ReplicatorResiliencySpecs

Aaronontheweb:


src/contrib/cluster/Akka.DistributedData.Tests/ReplicatorResiliencySpec.cs
```diff
@@ -180,8 +180,9 @@ public async Task DistributedData_Replicator_Defaults_to_NoSupervisor()
             Watch(durableStore);
             durableStore.Tell(new InitFail());
 
-            ExpectTerminated(durableStore, TimeSpan.FromSeconds(10));
```
Aaronontheweb, line 183: Can't really do two `ExpectTerminated` here because these messages are sent by two different actors, therefore the delivery order is nondeterministic

src/contrib/cluster/Akka.DistributedData.Tests/WriteAggregatorSpec.cs
```diff
@@ -249,13 +249,13 @@ public void WriteAggregator_with_delta_must_timeout_when_less_than_required_ACKs
             probe.ExpectMsg<DeltaPropagation>();
             probe.LastSender.Tell(WriteAck.Instance);
             probe.ExpectMsg<DeltaPropagation>();
-            // nack
-            probe.LastSender.Tell(DeltaNack.Instance);
             probe.ExpectMsg<DeltaPropagation>();
-            // no reply
+            
+            // nack - will force a write to be sent right away
+            probe.LastSender.Tell(DeltaNack.Instance);
```
Aaronontheweb, line 255: The message ordering in this spec was wrong, hence why it was racy

Aaronontheweb: Honestly this spec could be simplified pretty significantly too - most of it is testing the receipt of timed messages, which I think is pretty low-calorie. It should be testing, simply, that the `WriteAggregator` times out when not enough ACKs are received.

Issue #5157, opened 2021-07-26
DData Max-Delta-Elements may not be fully honoured 


andyfurnival:
**Version Information**
Version of Akka.NET? 
1.4.21
Which Akka.NET Modules?
Akka.DistributedData

**Describe the bug**
I believe the max-delta-elements will not be fully respected due to the operator used. Since this is greater than or equal to max-delta-elements, when it hits max elements it will discard the delta update. I think it should honour the max elements, and discard for any amount over. This is how Distributed Pub Sub seems to behave

**To Reproduce**

https://github.com/akkadotnet/akka.net/blob/f2b99b48f3eb187ac4f8805396c8d267314850a3/src/contrib/cluster/Akka.DistributedData/DeltaPropagationSelector.cs#L113

_Links to working reproductions on Github / Gitlab are very much appreciated_

**Expected behavior**
It should be just using greater than &gt;

**Actual behavior**
Its using &gt;=

**Screenshots**

**Environment**
Are you running on Linux? Windows? Docker? Which version of .NET?
Linux

**Additional context**
I have a PR ready to go if you agree this needs amending (didn't want to just assume a change was required)

Aaronontheweb:
@Arkatufus care to take this one onto the triage list?

Arkatufus:
The original code _does_ use >= :
https://github.com/akka/akka/blob/master/akka-distributed-data/src/main/scala/akka/cluster/ddata/DeltaPropagationSelector.scala#L121

Aaronontheweb:
We think the original Scala code is probably wrong in this regard too

PR #5163, opened 2021-08-02
Make sure that DData propagation honor MaxDeltaSize settings

Arkatufus:
Change limit to inclusive by changing >= to >

close https://github.com/akkadotnet/akka.net/issues/5157


Aaronontheweb (approved):
LGTM

PR #5175, opened 2021-08-06
Remove LightningDB reference in cluster sharding, make ddata non-durable as default

Arkatufus:


src/contrib/cluster/Akka.Cluster.Sharding/reference.conf
```diff
@@ -184,8 +184,10 @@ akka.cluster.sharding {
   distributed-data {
     # minCap parameter to MajorityWrite and MajorityRead consistency level.
     majority-min-cap = 5
-
-    durable.keys = ["shard-*"]
```
Aaronontheweb, line 188: Do you recall why we added this in the first place?

Arkatufus: It was in the official JVM reference.conf, I didn't know that our default value is different than theirs

Aaronontheweb: Well it seems that our issue really comes down to the .NET LightningDb packages not having all of the native dependencies they need. We'll add this setting back when we move the default for Akka.Cluster.Sharding state store mode away from persistence and to DData.

Aaronontheweb (approved):
LGTM

src/contrib/cluster/Akka.Cluster.Sharding/reference.conf
```diff
@@ -184,8 +184,10 @@ akka.cluster.sharding {
...
+    
+    # Uncomment this line if you want to use the durable version of DData and make sure that
+    # you also reference the Akka.DistributedData.LightningDB nuget package.
+    # durable.keys = ["shard-*"]
```
Aaronontheweb, line 190: Make sure we tell users to add this setting back in the documentation (in the other PR you're working on)

PR #5177, opened 2021-08-06
fix MNTR local build for .NET 5

Aaronontheweb:
some .NET 5 SDKs are affected by https://github.com/dotnet/sdk/issues/17579

Added recommended work-around to ensure production of local NuGet packages still


Aaronontheweb:
Once the build server's NuGet stage passes with flying colors, I'll merge this in. Resolved the issue for my local builds on Windows. Pretty sure this was a regression introduced ~ .NET 5.0.300 and fixed subsequently.

Issue #5174, opened 2021-08-06
System.DllNotFoundException: Unable to load shared library 'lmdb' or one of its dependencies


fscavo:
**Version Information**
Version of Akka.NET?
**1.14.22**
Which Akka.NET Modules?
**Akka.Cluster
Akka.Cluster.Sharding**

**Describe the bug**
The service used to work fine with the version 1.14.21. After having updated to 1.14.22 the service stopped working and kept throwing **System.DllNotFoundException** with the following details:
**_Unable to load shared library 'lmdb' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: liblmdb: cannot open shared object file: No such file or directory_**

Aaronontheweb:
We just received another bug report via our paid support channel about this issue. This appears to be a regression in our HOCON configuration that didn't get caught because we have the https://www.nuget.org/packages/Akka.DistributedData.LightningDB/ package installed in our test suites.

You can work around this issue, in the short run, by just installing that package.

However, we're working on an emergency patch to roll this back as it represents a breaking change. We're also going to add a regression test to block for this in the future.

Aaronontheweb:
Actually, it looks like we are referencing it: https://www.nuget.org/packages/Akka.Cluster.Sharding/1.4.22

Aaronontheweb:
@fscavo is this on Linux or Windows?

fscavo:
> @fscavo is this on Linux or Windows?

This is on Linux and the docker image is mcr.microsoft.com/dotnet/aspnet:5.0.7-focal

Aaronontheweb:
Could be related to https://github.com/CoreyKaylor/Lightning.NET/issues/141

andyfurnival:
I’ve always installed lmdb libraries via apt-get since I started using ddata. I always assumed that was a prerequisite

Aaronontheweb:
Looks like this was even previously reported and fixed https://github.com/CoreyKaylor/Lightning.NET/issues/133

Not sure why this is happening now with v1.4.22.

Arkatufus:
@fscavo For a workaround, if you're not using durable ddata, you can set `akka.cluster.sharding.distributed-data.durable.keys=[]` to  make sure that LightningDB isn't loaded

Aaronontheweb:
[...]

PR #5180, opened 2021-08-06
move Akka.DistributedData.LightningDb onto vendored package

Aaronontheweb:
only until https://github.com/CoreyKaylor/Lightning.NET/issues/141 gets resolved

close https://github.com/akkadotnet/akka.net/issues/5174


Aaronontheweb:
So I'm able to get LightningDb.NET working just fine on Windows in this PR, so there must be a binary copy issue that's hitting the MNTR. Between LightningDb v0.13 and this vendored version they moved to using two different .DLLs for LightningDb on Windows. That might be part of it. I'll take a closer look at the binary output on a local development machine.

Aaronontheweb:
I was able to reproduce the issue locally - in does indeed appear to be an issue with the MNTR being unable to resolve the native assembly paths at runtime. I'm not sure why this wasn't an issue before - it could have to do with the .NET Standard 2.0 version of the package resolving different than the .NET Core 3.1 and .NET 5.0 versions since https://github.com/CoreyKaylor/Lightning.NET/pull/138

src/contrib/cluster/Akka.DistributedData.Tests.MultiNode/Akka.DistributedData.Tests.MultiNode.csproj
```diff
@@ -4,6 +4,7 @@
   <PropertyGroup>
     <AssemblyTitle>Akka.DistributedData.Tests.MultiNode</AssemblyTitle>
     <TargetFrameworks>$(NetFrameworkTestVersion);$(NetTestVersion);$(NetCoreTestVersion)</TargetFrameworks>
+    <Platforms>x64</Platforms> <!-- target x64 only due to LightningDb support -->
```
Aaronontheweb, line 7: I think `AnyCPU` might have something to do with the issues here. Going to see if I can get this all to work under x64.

Aaronontheweb: Doesn't appear to have helped, based on the data from my local machine

Aaronontheweb:
On this PR, the structure of my folders from the build output of `Akka.Cluster.Sharding.Tests.MultiNode.csproj` - all of the output targets have the `runtimes` folder, where the native .DLLs live, except for `net471`: https://gist.github.com/Aaronontheweb/99877df36004af89b8ceb8653f02b244

Aaronontheweb:
Alright, figured this one out - looks like the `lmdb.dll` is no longer being copied to the same directory as `LightningDb.dll` and that's why the `ImportDLL` call fails. Fun.

src/contrib/cluster/Akka.DistributedData.Tests.MultiNode/Akka.DistributedData.Tests.MultiNode.csproj
Aaronontheweb: LOL, applied this to the wrong project....

Aaronontheweb:
Looks like this is an MSBuild issue with the latest Lightning.NET sources: https://github.com/CoreyKaylor/Lightning.NET/issues/143

Definitely an edge case in how native .DLLs get resolved.

Aaronontheweb:
Might have to specify the [`DefaultDllImportSearchPaths`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.defaultdllimportsearchpathsattribute?view=netstandard-2.0) inside the `Akka.DistributedData.LightningDb` package as it seems that the MNTR is unable to search in the `runtimes` folder produced by the compilation: https://stackoverflow.com/questions/2864673/specify-the-search-path-for-dllimport-in-net/45384763#45384763

Other possibility I'm going to look at is that the MNTR's [Runtime Identifier (RID)](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog#rid-graph) is at issue.

https://github.com/akkadotnet/akka.net/blob/0bd23c981ead97c03363bdf7e785afd84f0a1f38/src/core/Akka.MultiNodeTestRunner/Akka.MultiNodeTestRunner.csproj#L9

Aaronontheweb:
I've tested this locally and changing the RIDs does not help at all - those directories simply aren't being searched for these DLLs.

Aaronontheweb:
Looks like this resolved the issue but I need to refine this a tad more....

src/contrib/cluster/lib-hack/LightningDB.targets
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
Aaronontheweb, line 2: Had to resolve this by re-using the `.targets` file from the original projects. 

Everything using the original `LightningDb` package runs just fine in a stand-alone project, but the MNTR isn't able to resolve these DLLs when it's running. Best bet is to just stick the DLLs in a folder adjacent to the .NET dlls, which is how things work in the current `dev` branch.

Aaronontheweb:
Bah, looks like I broke the .DLL copying again by trying to share it. I'll need to tweak that again.

src/contrib/cluster/lib-hack/LightningDB.targets
Aaronontheweb: Note: this targets file is only necessary because of the quirks of the MNTR, transitive dependencies, and multi-targeting in MSBuild. A normal software application using Akka.Cluster.Sharding will not have this issue.

Issue #5165, opened 2021-08-04
RestartFlow Akka Streams not working as expected


Danthar:
**Version Information**
Version of Akka.NET? 1.4.18
Which Akka.NET Modules? Akka Streams

**Describe the bug**
RestartFlow does not restart.

**To Reproduce**

```
        [Fact]
        public void RetryTestCases()
        {
            var flow = RestartFlow.OnFailuresWithBackoff(() =>
                {
                    return Flow.Create<int>() 
                        .Select(i =>
                        {
                            if(i == 6)
                                throw new ArgumentException($"BOOM");
                            return i;
                        }).Log("LOG ON ERROR", i => i, Sys.Log).WithAttributes(Attributes.CreateLogLevels(Attributes.LogLevels.Off, Attributes.LogLevels.Off, LogLevel.ErrorLevel));
                }, TimeSpan.FromMilliseconds(100), TimeSpan.FromSeconds(5), 0.3, 10);
            
            Source.From(new int[] { 1,2,3,4,5,6 })
                .Via(flow)
                .To(Sink.ForEach<int>(i =>
                {
                    Log.Info($"Received: {i}");
                })).Run(Sys.Materializer());
        }

```

**Expected behavior**
I expected the exception to occur 10 times. 

**Actual behavior**

```
[DEBUG][4-8-2021 10:12:10][Thread 0010][EventStream] subscribing [akka://testsystem/system/log-test#1666761903] to channel Akka.Event.LogEvent
[DEBUG][4-8-2021 10:12:11][Thread 0027][akka://testsystem/user] now supervising akka://testsystem/user/StreamSupervisor-0
[DEBUG][4-8-2021 10:12:11][Thread 0031][akka://testsystem/user/StreamSupervisor-0] Started (Akka.Streams.Implementation.StreamSupervisor)
[DEBUG][4-8-2021 10:12:11][Thread 0010][akka://testsystem/user/StreamSupervisor-0] now supervising akka://testsystem/user/StreamSupervisor-0/Flow-0-0-unknown-operation
[DEBUG][4-8-2021 10:12:11][Thread 0015][CoordinatedShutdown (akka://testsystem)] Performing phase [before-service-unbind] with [0] tasks.
[DEBUG][4-8-2021 10:12:11][Thread 0033][CoordinatedShutdown (akka://testsystem)] Performing phase [service-unbind] with [0] tasks.
[...]

PR #5181, opened 2021-08-07
Bugfix RestartFlow issue; #5165

Danthar:
For this PR i looked at what they did in the original bugfix on the JVM repo:
https://github.com/akka/akka/pull/24795/files#diff-ef6a29e6541e6af3e4452f69c790020e85116bad800388ce464dea7383eec9c0

And in whats currently in Master:
https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/scaladsl/RestartFlow.scala

Im not completely done yet. Still need to go through the Source and Sink implementations. And i want to see if i can add a UnitTest which verifies the behavior we want to fix.
But i wanted to get early feedback, if any.

I did ran into some things. It seems our Akka.Streams port could use some love. Lots of little changes that haven't been ported over the last year.
For example see https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/scaladsl/RestartFlow.scala#L425
The onDownstreamFinish has gotten a new Exception cause parameter. Which is absent in our codebase.


Danthar:
I looked at why they added a delay in cancellation. It seems there is a missing feature which up to this day they havent realised yet, namely the propagation of a cancellation cause or something. Which would enable the RestartFlow to prevent it from conditionally propagating the cancellation or something. So i went with the same Delay stage that they up to this day still have implemented.
If its good enough for them, its probably good enough for us. 

Please check my implementation, my scala-fu could be better, so i might have missed something subtle that im not aware of.

src/core/Akka.Streams/Dsl/RestartFlow.cs
```diff
@@ -481,31 +187,42 @@ internal sealed class RestartWithBackoffFlow<TIn, TOut, TMat> : GraphStage<FlowS
...
             }
 
             protected override void StartGraph()
             {
                 var sourceOut = CreateSubOutlet(_stage.In);
                 var sinkIn = CreateSubInlet(_stage.Out);
-                Source.FromGraph(sourceOut.Source).Via(_stage.FlowFactory()).RunWith(sinkIn.Sink, SubFusingMaterializer);
+                
+                var graph = Source.FromGraph(sourceOut.Source)
+                    //temp fix becaues the proper fix would be to have a concept of cause of cancellation. See https://github.com/akka/akka/pull/23909
```
Aaronontheweb, line 214: we should file a separate issue for this once it's merged, so we can revisit in the future.

src/core/Akka.Streams/Dsl/RestartFlow.cs
```diff
@@ -698,6 +418,92 @@ protected internal override void OnTimer(object timerKey)
...
+            {
+                _stage = stage;
+                
+                SetHandler(stage.Inlet, onPush: () => Push(stage.Outlet, Grab(stage.Inlet)));
+
+                SetHandler(stage.Outlet, onPull: () => Pull(stage.Inlet), onDownstreamFinish: OnDownStreamFinished );
+            }
+            
+            /// <summary>
+            /// We should really. port the Cause parameter functionality for the OnDownStreamFinished delegate
```
Aaronontheweb, line 482: Agree - we should file an issue

src/core/Akka.Streams/Dsl/RestartSink.cs
```diff
@@ -0,0 +1,138 @@
...
+// //-----------------------------------------------------------------------
+
+using System;
+using Akka.Pattern;
+using Akka.Streams.Stage;
+
+namespace Akka.Streams.Dsl
+{
+    /// <summary>
+    /// A RestartSink wraps a <see cref="Sink"/> that gets restarted when it completes or fails.
```
Aaronontheweb, line 15: When you say "completes" here - you don't mean that when the graph completes right? That wasn't clear from reading this.

Aaronontheweb (commented):
Left some comments - need to make sure you do API approval though: https://getakka.net/community/public-api-changes.html

Aaronontheweb:
Looks like the unit tests need to be updated too - lots of NREs

src/core/Akka.Streams/Dsl/RestartFlow.cs
```diff
@@ -481,31 +187,42 @@ internal sealed class RestartWithBackoffFlow<TIn, TOut, TMat> : GraphStage<FlowS
...
+            private readonly Attributes _inheritedAttributes;
             private Tuple<SubSourceOutlet<TIn>, SubSinkInlet<TOut>> _activeOutIn;
-
-            public Logic(RestartWithBackoffFlow<TIn, TOut, TMat> stage, string name)
+            private TimeSpan _delay;
+            
+            public Logic(RestartWithBackoffFlow<TIn, TOut, TMat> stage, Attributes inheritedAttributes, string name)
                 : base(name, stage.Shape, stage.In, stage.Out, stage.Settings, stage.OnlyOnFailures)
             {
+                _delay = _inheritedAttributes.GetAttribute<Delay>(new Delay(TimeSpan.FromMilliseconds(50))).Duration;
```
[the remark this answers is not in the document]
Danthar: Thats one thing im unsure about. 
In the scala source, the `Delay` class is a case class, wrapped in an object. (see: https://github.com/akka/akka/blob/master/akka-stream/src/main/scala/akka/stream/scaladsl/RestartFlow.scala#L402)  Im unsure if that means its a public api or not. But i figured undefined access modifiers means private/internal so i went with that.

The comments on the original scala bugfix suggest its overridable through a WithAttributes call on the RestartFlow, but im not sure if that works here (regardless of the access modifier). im pretty sure they are not applied to the flow the RestartFlow is wrapping.

However i think normally nobody will ever want to tweak stuff in their stream on this level. Im of the opinion that "nobody should ever need to tweak this, because it should just work". But considering what the fix is, they might have to ? Although i'd expect that the # of ppl that would need to do this would be vanishingly rare.

src/core/Akka.Streams/Dsl/RestartFlow.cs
Danthar: Agreed. Ill make a note for myself

src/core/Akka.Streams/Dsl/RestartSink.cs
Danthar: No idea. I only seperated the Sink and Source code to seperate files at this point. Still need to go through them. Ill take a closer look at the comments then as well.

Danthar:
Thx for the review. Answered some of your remarks. Yup the Units Tests needs work, ignored those for now. My next step is first going through the Source and Sink codebase to port their latest changes. And then im going to look at the UnitTests.
I expect that fixing those will take most of the time for this PR.

src/core/Akka.Streams/Dsl/RestartFlow.cs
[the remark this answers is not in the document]
Aaronontheweb: Yeah that's a public API - it's meant to be an attribute you can pass in at the time the flow is declared via the `inheritedAttributes` collection. So I guess, broadly speaking, this is meant to be configurable via the attributes approach.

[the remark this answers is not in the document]
Aaronontheweb: > Although i'd expect that the # of ppl that would need to do this would be vanishingly rare.

Agree - so the duration here just sets a 50ms threshold for restarting. That seems fine for most cases, where it's recovering from a crash.

src/core/Akka.Streams/Dsl/RestartSource.cs
```diff
@@ -0,0 +1,188 @@
...
+            {
+                _stage = stage;
+                _inheritedAttributes = inheritedAttributes;
+                Backoff();
+            }
+
+            protected override void StartGraph()
+            {
+                var sinkIn = CreateSubInlet(_stage.Out);
+                SubFusingMaterializer.Materialize(_stage.SourceFactory().To(sinkIn.Sink), _inheritedAttributes);
```
Danthar, line 176: Brought this in line with the scala sourcecode. IMO this used to contain a big ass bug. Before it never used the SourceFactory. How that ever worked i don't know. I can only guess the specs surrounding the RestartSource are bad.

src/core/Akka.Streams/Dsl/RestartSink.cs
Danthar: The docs are the same as on the scala side. I think what is being referenced here is the stage itself. And not the entire graph.

Danthar:
Ok, im looking at this bit of code in the Tests.
https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs#L691

While im sure its very handy to verify that a flow wrapped by the RestartFlow is actually restarted. I think it does nothing to verify that a message supplied to the graph of which the RestartFlow stage is a part, is actually Retried without having the entire graph being completed/failed.

I could not find the original unit tests in the scala repo btw, i wonder how they test this.

Aaronontheweb:
> Ok, im looking at this bit of code in the Tests.
> https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs#L691
> 
> While im sure its very handy to verify that a flow wrapped by the RestartFlow is actually restarted. I think it does nothing to verify that a message supplied to the graph of which the RestartFlow stage is a part, is actually Retried without having the entire graph being completed/failed.
> 
> I could not find the original unit tests in the scala repo btw, i wonder how they test this.

I guess the easiest way to do it would be to pass a numbered sequence 1-10, set number 5 to cause a crash inside the original flow, and see if [1,4] [6,10] still arrive before the stream completes.

src/core/Akka.Streams/Dsl/RestartSource.cs
Aaronontheweb: Given how racy those specs have been, I think that's a good assumption

Aaronontheweb:
@Danthar looks like the only specs that are failing right now are the API Approval specs https://getakka.net/community/public-api-changes.html

Danthar:
> 
> I guess the easiest way to do it would be to pass a numbered sequence 1-10, set number 5 to cause a crash inside the original flow, and see if [1,4] [6,10] still arrive before the stream completes.

Not quite. Since it should keep retrying the original message. So the error on number 5 should be transient, as in, only occur a few times. And then the flow should keep retrying untill it succeeds and end up emitting the entire sequence at the end.

I still want to add a test which verifies this. But that will be next week probably before i can get back to this.

Danthar:
Bad news. Playing around with a unit test shows it does not work. Now somehow it seems like the stage is not restarting.

src/core/Akka.Streams/Dsl/RestartFlow.cs
Danthar: Done

src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs
```diff
@@ -1010,8 +1075,8 @@ public void A_restart_with_backoff_flow_should_restart_on_failure_when_using_onl
             flowInProbe.RequestNext("c");
             flowOutProbe.SendNext("d");
             sink.RequestNext("d");
-
+            sink.Request(1);
```
Aaronontheweb, line 1078: Nice catch

Aaronontheweb (approved):
LGTM

PR #5182, opened 2021-08-07
MNTR: modernize RIDs, add Linux support

### LINKED ISSUE #3573: Possible issue with MultiNodeTestRunner on Unix environments
when running this in a net452 environment this does not happen, this only seems to happen on *nix (in my case OSX)

when running:

`dotnet run --project Akka.MultiNodeTestRunner.csproj -f netcoreapp1.1 "/Users/*/Workspace/Akka/akka.net/src/core/Akka.Cluster.Tests.MultiNode/bin/Debug/netcoreapp1.1/Akka.Cluster.Tests.MultiNode.dll" -Dmultinode.enable-filesink=on`

I get:
```
Unhandled Exception: System.ComponentModel.Win32Exception: No such file or directory
   at Interop.Sys.ForkAndExecProcess(String filename, String[] argv, String[] envp, String cwd, Boolean redirectStdin, Boolean redirectStdout, Boolean redirectStderr, Int32& lpChildPid, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start()
   at Akka.MultiNodeTestRunner.Program.Main(String[] args) in /Users/*/Workspace/Akka/akka.net/src/core/Akka.MultiNodeTestRunner/Program.cs:line 311
```
I am not particularly sure where the issue stems from but on gitter, Aaron said it might be a file globbing issue and its not using environment agnostic `Path.Join()`'s

Aaronontheweb:
close #3573 

Not 100% sure about the NuGet package output being germane, but we shall see.

Also: reduced build time significantly


Aaronontheweb:
Need to force this PR to try to run the entire MNTR

build.fsx
```diff
@@ -494,24 +494,24 @@ Target "PublishMntr" (fun _ ->
     if not skipBuild.Value then
         let executableProjects = !! "./src/**/Akka.MultiNodeTestRunner.csproj"
```
Aaronontheweb, line 495: I need to add a switch that does a Linux version of this, on-demand, and then update our build pipelines for it

src/core/Akka.MultiNodeTestRunner/Akka.MultiNodeTestRunner.nuspec
```diff
@@ -0,0 +1,22 @@
...
+    <owners>Akka.NET Team</owners>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <licenseUrl>https://github.com/akkadotnet/akka.net/blob/master/LICENSE</licenseUrl>
+    <projectUrl>https://github.com/akkadotnet/akka.net</projectUrl>
+    <iconUrl>http://getakka.net/images/AkkaNetLogo.Normal.png</iconUrl>
+    <description>Akka.NET Multi-node Test Runner; used for executing tests written with Akka.Remote.TestKit</description>
+    <copyright>Copyright � 2013-2021 .NET Foundation</copyright>
+    <tags>akka actors actor model Akka concurrency</tags>
+  </metadata>
+  <files>
```
Aaronontheweb, line 17: In order to publish a Linux version of our NuGet packages.... Maybe we should think about doing a .NET global tool

Aaronontheweb:
Having some issues on .NET Core / .NET 5 with "self contained app" bugs. This is due to our dependency on NodeTestRunner.exe from inside MultiNodeTestRunner.exe. We might need to copy over a config file to specify which platform we're running on.

PR #5183, opened 2021-08-07
adding NuGet logo and README support

Aaronontheweb:
Working on trying to implement https://devblogs.microsoft.com/nuget/add-a-readme-to-your-nuget-package/ but my MSBuild-foo is not yet strong enough....


Aaronontheweb:
NuGet package stage worked on the build server. I'll have to retry doing that locally.

Aaronontheweb:
Looks like the build system fails on my machine due to https://github.com/NuGet/Home/issues/10791 - can't include the `README` file with our NuGet package until `snupkg` adds support for it. Otherwise, we lose SourceLink support - which is more useful and valuable than a README in the NuGet package.

Aaronontheweb:
This doesn't fail on the build server because it uses an older version of the .NET 5 SDK than I do on my development machine. 

Server: 5.0.101
My machine: 5.0.300

Aaronontheweb:
In theory, I could split the `snupkg` and `nupkg` publication into two different discrete steps and manually pass in a flag to disable the `PackReadmeFile` variable but I think I'd rather just wait for a fix.

Aaronontheweb:
https://github.com/NuGet/Home/issues/10791#issuecomment-941247759 - need to upgrade to the .NET 6.0 runtime to resolve this issue.

Aaronontheweb:
Let's test this out now that we've merged in https://github.com/akkadotnet/akka.net/pull/5373

Aaronontheweb:
This now works properly on .NET 6! Going to update our `README.md` to be a little more NuGet-friendly in a subsequent issue.

PR #5187, opened 2021-08-09
use `Failure` to programmatically cancel `Ask`

### LINKED ISSUE #3019: Persistence plugins need to support SerializerWithStringManifest

### LINKED ISSUE #3091: Ask does not set exception on the underlying TaskCompletionSource when callee responds with Failure
Hello,

I am running Akka.net 1.3.0 on .NET Core, and came across the following issue:

I am calling `Ask<TResult>` from outside of the system and would like to have Ask throw an exception at the call site if the underlying actor responds with `new Failure() { Exception = ex }` instead of `TResult`.

This behaviour is described in the docs:

> To complete the Task with an exception you need send a Failure message to the sender. This is not done automatically when an actor throws an exception while processing a message.

- http://getakka.net/articles/actors/receive-actor-api.html#ask-send-and-receive-future
- http://getakka.net/articles/actors/untyped-actor-api.html#ask-send-and-receive-future

This feature is however not implemented, if you look at https://github.com/akkadotnet/akka.net/blob/a478c5e8c8d5de97dcf7bcea7e2a80fbaa5e6cc8/src/core/Akka/Actor/ActorRef.cs#L129 the `FutureActorRef` sets the result of the task without inspecting whether the message is a `Failure`.

This then breaks `Ask<TResult>` as the underlying extension method tries to cast `Failure` to `TResult`.

Unless I am mistaken the following modification could be used:
```
if (message is Failure failure) {
    _result.TrySetException(failure.Exception ?? new SomeDefaultException())
} else {
    _result.TrySetResult(message);
}
```

I am assuming this is still a behaviour you want to support. If not then the docs should be updated so that they don't put peoples hopes up :)

Many thanks,
Ondrej.

Aaronontheweb:
close #3091 - part of the `Ask` API that's never been fully implemented.

### DIFF
--- a/src/core/Akka/Actor/ActorRef.cs
+++ b/src/core/Akka/Actor/ActorRef.cs
@@ -146,6 +146,15 @@ namespace Akka.Actor
                 }
             }
         }
+        
+        /// <summary>
+        /// TBD
+        /// </summary>
+        /// <param name="message">TBD</param>
+        public override void SendSystemMessage(ISystemMessage message)
+        {
+            base.SendSystemMessage(message);
+        }
     }
 
 
--- a/src/core/Akka/Actor/ActorRef.cs
+++ b/src/core/Akka/Actor/ActorRef.cs
@@ -134,6 +134,10 @@ namespace Akka.Actor
                     {
                         _result.TrySetResult(t);
                     }
+                    else if (message is Failure f)
+                    {
+                        _result.TrySetException(f.Exception ?? new TaskCanceledException("Task cancelled by actor via Failure message."));
+                    }
                     else
                     {
                         _result.TrySetException(new ArgumentException(
@@ -141,15 +145,6 @@ namespace Akka.Actor
                     }
                 }
             }
-        }
-
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="message">TBD</param>
-        public override void SendSystemMessage(ISystemMessage message)
-        {
-            base.SendSystemMessage(message);
         }
     }
 


src/core/Akka/Actor/ActorRef.cs
```diff
@@ -142,15 +146,6 @@ protected override void TellInternal(object message, IActorRef sender)
                 }
             }
         }
-
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="message">TBD</param>
-        public override void SendSystemMessage(ISystemMessage message)
```
Aaronontheweb, line 150: Bah, getting rid of this is a "public API change"

Aaronontheweb: Reverted that

Arkatufus (approved):
LGTM

PR #5192, opened 2021-08-11
Use TimestampProvider in BatchingSqlJournal

MartijnSchoemaker:
PR #5077 fixed using timestamp provider for SqlJournal.
With this new PR the timestamp provider will be used in BatchingSqlJournal as well.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
@@ -236,6 +236,9 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// </summary>
         public string DefaultSerializer { get; }
 
+        /// <summary>
+        /// The fully qualified name of the type that should be used as timestamp provider.
+        /// </summary>
         public string TimestampProviderTypeName { get; }
 
         /// <summary>
@@ -532,6 +535,9 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// </summary>
         protected readonly bool CanPublish;
 
+        /// <summary>
+        /// The timestamp provider that will be used for the timestamp column when writing messages to the database.
+        /// </summary>
         protected ITimestampProvider TimestampProvider { get; }
 
         /// <summary>
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
@@ -22,6 +22,7 @@ using Akka.Configuration;
 using Akka.Event;
 using Akka.Pattern;
 using Akka.Persistence.Journal;
+using Akka.Persistence.Sql.Common.Journal;
 using Akka.Serialization;
 using Akka.Util;
 using Akka.Util.Internal;
@@ -233,7 +234,9 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// <summary>
         /// The default serializer used when not type override matching is found
         /// </summary>
-        public string DefaultSerializer { get; } 
+        public string DefaultSerializer { get; }
+
+        public string TimestampProviderTypeName { get; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="BatchingSqlJournalSetup" /> class.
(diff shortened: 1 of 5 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -528,6 +535,11 @@ public RequestChunk(int chunkId, IJournalRequest[] requests)
         /// </summary>
         protected readonly bool CanPublish;
 
+        /// <summary>
+        /// The timestamp provider that will be used for the timestamp column when writing messages to the database.
+        /// </summary>
+        protected ITimestampProvider TimestampProvider { get; }
```
Aaronontheweb, line 541: Need to add API approval for this https://getakka.net/community/public-api-changes.html

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1286,7 +1299,7 @@ private async Task<WriteMessagesResult> HandleWriteMessages(WriteMessages req, T
                         persistent = persistent.WithPayload(tagged.Payload);
                     }
 
-                    WriteEvent(command, persistent.WithTimestamp(DateTime.UtcNow.Ticks), tagBuilder.ToString());
+                    WriteEvent(command, persistent.WithTimestamp(TimestampProvider.GenerateTimestamp(persistent)), tagBuilder.ToString());
```
Aaronontheweb, line 1302: LGTM

Aaronontheweb (changes requested):
LGTM, but need to approve the public API changes via https://getakka.net/community/public-api-changes.html

Aaronontheweb (approved):

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
Aaronontheweb: I took care of this

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/ITimestampProvider.cs
```diff
@@ -34,4 +35,16 @@ public sealed class DefaultTimestampProvider : ITimestampProvider
...
+    {
+        public static ITimestampProvider GetTimestampProvider(string typeName, IActorContext context)
+        {
+            var type = Type.GetType(typeName, true);
+            var withSystem = type.GetConstructor(new[] { context.System.GetType() }) != null;
+            return withSystem ?
+                (ITimestampProvider)Activator.CreateInstance(type, context.System) :
+                (ITimestampProvider)Activator.CreateInstance(type);
+        }
+    }
```
to11mtm, line 49: Unit tests are currently failing:

```
  ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
  ---> System.ArgumentNullException: Value cannot be null. (Parameter 'typeName')
    at System.RuntimeType.GetType(String typeName, Boolean throwOnError, Boolean ignoreCase, StackCrawlMark& stackMark)
    at System.Type.GetType(String typeName, Boolean throwOnError)
    at Akka.Persistence.Sql.Common.Journal.TimestampProviderProvider.GetTimestampProvider(String typeName, IActorContext context) in D:\a\1\s\src\contrib\persistence\Akka.Persistence.Sql.Common\Journal\ITimestampProvider.cs:line 45
```

I think you need to return a default provider here that works the way things do now, in the case the string is null or empty.

Aaronontheweb: Man, I totally missed that. Good catch.

MartijnSchoemaker: Theoretically the `timestamp-provider` config value always has the default value of `Akka.Persistence.Sql.Common.Journal.DefaultTimestampProvider`. So there should always be a timestamp provider. 

But still I think it would indeed better to check `type` here as well and return an instance of Akka.Persistence.Sql.Common.Journal.DefaultTimestampProvider if null or empty.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
MartijnSchoemaker: Thanks

PR #5196, opened 2021-08-12
added reproduction for DU serialization

### LINKED ISSUE #5194: Akka.FSharp: need to re-enable discriminated union (DU) serialization tests
**Version Information**
Version of Akka.NET? v1.4.23
Which Akka.NET Modules? Akka.FSharp

**Describe the bug**
https://stackoverflow.com/questions/68741561/akka-unable-to-send-discriminated-unions-as-messages-in-f#comment121522181_68741561 - looks like default DU serialization isn't working with Newtonsoft.Json per this user's report. 

As it so happens, our DU serialization tests are out of date and commented out:

https://github.com/akkadotnet/akka.net/blob/d1463872dc399020b73831943e2c0c976c6504a8/src/core/Akka.FSharp.Tests/ApiTests.fs#L41-L76

I think a lot of this stems from the Akka.NET v1.3 push and the fact that F# lagged behind a bit on .NET Core support at the time. Now that everything is pretty reliable there for the most part we should reintroduce these tests as part of our suite.

**Actual behavior**
Not sure, but it looks like the DU content is lost in translation.

**Expected behavior**
DUs should be serializable via:

* Newtonsoft.Json
* Hyperion

With all of its contents correctly preserved.

Aaronontheweb:
close #5194

### DIFF
--- a/src/core/Akka.FSharp.Tests/RemoteSpecs.fs
+++ b/src/core/Akka.FSharp.Tests/RemoteSpecs.fs
@@ -1,5 +1,6 @@
 ﻿module Akka.FSharp.Tests.RemoteSpecs
 
+open System
 open Akka.Actor
 open Xunit
 open Xunit.Abstractions
@@ -36,6 +37,10 @@ type TestUnion =
 type TestUnion2 = 
     | C of string * TestUnion
     | D of int
+type Record1 = {Name:string;Age:int;}
+
+type Msg =
+    | R of Record1
 
 type RemoteSpecs(output:ITestOutputHelper) as this =
     inherit AkkaSpec((remoteConfig 0), output)
@@ -60,12 +65,47 @@ type RemoteSpecs(output:ITestOutputHelper) as this =
                    | C("a-11", B(11, "a-12")) -> mailbox.Sender() <! msg
                    | _ -> mailbox.Unhandled msg) @>
                 [SpawnOption.Deploy (Akka.Actor.Deploy(RemoteScope(this.GetAddress)))]
-                
+        
+        // test record serialization too
+        let testFire = { Name = "aaron"; Age = 30 }
+        aref.Tell(testFire, this.TestActor)
+        
         let msg = C("a-11", B(11, "a-12"))
         aref.Tell(msg, this.TestActor)
          
         // assert
-        this.ExpectMsg(msg)
+        this.ExpectMsg(testFire) |> ignore
+        this.ExpectMsg(msg) |> ignore
+        
+    [<Fact>]
+    member _.``can serialize and deserialize discriminated F# records over remote nodes using default serializer`` () =
+        
+        // arrange
+        use clientSys = System.create "clientSys" (remoteConfig 0)
+        let addr2 = getAddress clientSys
+        
+        // act
+        let aref = 
+            spawne clientSys "a-1" <@ actorOf2 (fun mailbox (msg:obj) -> 
+                   match msg with
+                   | :? Record1 as r -> mailbox.Log.Value.Info("Received message {0}", r)
+                                        mailbox.Sender() <! r
+                   | _ ->  mailbox.Unhandled msg) @>
+                [SpawnOption.Deploy(Deploy.None)]
+        
+        let msg = { Name = "aaron"; Age = 30 }
+        
+        let selection = this.Sys.ActorSelection(new RootActorPath(addr2) / "user" / "a-1")
+        let remoteRef =
+            async {
+                let! rRef = selection.ResolveOne this.RemainingOrDefault |> Async.AwaitTask
+                return rRef
+             } |> Async.RunSynchronously       
+        
+        remoteRef.Tell(msg, this.TestActor)
+        
+        // assert
+        this.ExpectMsg(msg) |> ignore
         
         
 
--- a/src/core/Akka.FSharp.Tests/RemoteSpecs.fs
+++ b/src/core/Akka.FSharp.Tests/RemoteSpecs.fs
@@ -18,6 +18,12 @@ let remoteConfig port =
             actor {{
                 ask-timeout = 5s
                 provider = remote
+                serializers {{
+                  hyperion = "Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion"
+                }}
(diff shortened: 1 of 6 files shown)


src/core/Akka.FSharp.Tests/RemoteSpecs.fs
```diff
@@ -0,0 +1,67 @@
...
+type TestUnion = 
+    | A of string
+    | B of int * string
+
+type TestUnion2 = 
+    | C of string * TestUnion
+    | D of int
+
+type RemoteSpecs(output:ITestOutputHelper) as this =
+    inherit AkkaSpec((remoteConfig 0), output)
```
Aaronontheweb, line 39: A little convoluted, but this is how you can pass through the XUnit `ITestOutputHelper` and the `AkkaSpec` / `TestKit` base members into an F# test

src/core/Akka.FSharp.Tests/RemoteSpecs.fs
```diff
@@ -0,0 +1,71 @@
...
+        // arrange
+        use clientSys = System.create "clientSys" (remoteConfig 0)
+        let addr2 = getAddress clientSys
+        
+        // act
+        let aref = 
+            spawne clientSys "a-1" <@ actorOf2 (fun mailbox msg -> 
+                   match msg with
+                   | C("a-11", B(11, "a-12")) -> mailbox.Sender() <! msg
+                   | _ -> mailbox.Unhandled msg) @>
```
Aaronontheweb, line 61: We have a clear reproduction here - we receive a `JObject` instead of a DU here.

Aaronontheweb: Updated this to use Hyperion - even a basic Newtonsoft.Json serialization setup can't handle DUs

src/core/Akka.FSharp.Tests/SerializationSpecs.fs
```diff
@@ -0,0 +1,73 @@
...
+            | :? SerializerWithStringManifest as str -> this.Sys.Serialization.Deserialize(bytes, serializer.Identifier, str.Manifest msg) 
+            | _ -> this.Sys.Serialization.Deserialize(bytes, serializer.Identifier, t)
+
+        Assert.Equal(msg, deserialized)
+        
+    [<Theory>]
+    [<MemberData(nameof(SerializationSpecs.FSharpTypes))>]
+    member _.``Must verify serialization of F#-specific types`` (t:obj) =
+        this.VerifySerialization t
+        // uncomment the section below when experimenting with IncludeManifest = true on the Newtonsoft.Json serializer
```
Aaronontheweb, line 56: If we enable `IncludeManifest = true` in the `NewtonsoftJsonSerializer` all of these tests will pass, including the ones currently using Hyperion right now. However, I'm not sure how significant of a change that might be and I'm loathe to touch it given that it powers so many Akka.Persistence serialization schemes.

src/core/Akka.FSharp/FsApi.fs
```diff
@@ -35,7 +35,7 @@ module Serialization =
         override __.FromBinary(bytes, _) = deserializeFromBinary fsp bytes
         
                         
-    let internal exprSerializationSupport (system: ActorSystem) =
+    let exprSerializationSupport (system: ActorSystem) =
```
Aaronontheweb, line 38: made this public

src/core/Akka.FSharp.Tests/SerializationSpecs.fs
```diff
@@ -0,0 +1,73 @@
...
+//        let s = this.Sys.Serialization.FindSerializerFor t
+//        let manifest = Akka.Serialization.Serialization.ManifestFor(s, t)
+//        Assert.True(manifest.Length > 0)
+        
+        
+    [<Fact(Skip="JSON.NET really does not support even basic DU serialization")>]
+    member _.``JSON.NET must serialize DUs`` () =
```
Aaronontheweb, line 63: Smoking gun - `JsonConvert` can't deserialize a DU even with `DiscriminatedUnionConverter` and none of our other customizations to Newtonsoft.Json. Will need to file an issue over there.

Aaronontheweb (commented):
Described my work on adding better F# DU testing support going forward - looks like Newtonsoft.Json really can't support DUs even with `DiscriminatedUnionConverter`

Arkatufus (approved):
LGTM

src/core/Akka.FSharp.Tests/RemoteSpecs.fs
Horusiath: If I remember correctly - but I might be wrong - the issue might be caused by the fact that you didn't configured JSON.NET to include type name as part of the payload. Since at the deserialization time payload type is still unknown and JSON.NET will use JObject as representation of `System.Object` instance.

Aaronontheweb: Yes, with type hints Newtonsoft.Json correctly deserializes DUs: 

https://github.com/akkadotnet/akka.net/blob/d2014528380ac7c00ea97084ba59495ba8ae76f2/src/core/Akka.FSharp.Tests/SerializationSpecs.fs#L41-L50

We disabled exposing type hints during the advent of Akka.NET v1.3 because of `mscorlib` namespace issues introduced by the creation of .NET Core: https://twitter.com/Danthar/status/1426204694310014981 cc @Danthar 

We could fix the issue by turning type hints on, but it would blow up anyone doing .NET Framework --> .NET Core interop with Newtonsoft.Json (population of people doing that is great than zero) - or we could make it configurable to turn that setting on or off.

Horusiath: I think you can potentially handle that as special case inside of typed actor itself: https://github.com/akkadotnet/akka.net/blob/d2014528380ac7c00ea97084ba59495ba8ae76f2/src/core/Akka.FSharp/FsApi.fs#L291-L297

Before down casting message, check if `obj` is `JObject` and if so use its [ToObject](https://www.newtonsoft.com/json/help/html/M_Newtonsoft_Json_Linq_JToken_ToObject__1_1.htm) method.

PR #5205, opened 2021-08-16
allow `null` response to `Ask<T>` operation

### LINKED ISSUE #5204: Sender.Tell(null) in response to Ask<T> now throws exception
**Version Information**
Version of Akka.NET?  1.4.23
Which Akka.NET Modules?
Akka (main module)

**Describe the bug**
When I do .Ask<T>(...) and in the actor I try to reply with null - Sender.Tell(null) an exception is thrown:

```
   at Akka.Actor.FutureActorRef`1.TellInternal(Object message, IActorRef sender)
   at Akka.Actor.ActorRefBase.Tell(Object message, IActorRef sender)
```

In version 1.4.22 you changed FutureActorRef to FutureActorRef&lt;T>  and in `protected override void TellInternal` you are checking type of the message using the `is` operator. This operator returns false with null and subsequent `message.GetType()` throws an exception.

**To Reproduce**
1. Spawn an actor, obtain IActorRef
2. await .Ask
3. In the actor reply with Sender.Tell(null)
4. An exception is thrown

**Expected behavior**
The null response is passed to the awaiter

**Actual behavior**
An exception occurs

**Additional context**
We use null response to indicate that an operation has completed when there is nothing to reply. Usually from an mvc controller that then returns no-content.

Aaronontheweb:
close #5204

### DIFF
--- a/src/core/Akka.Tests/Actor/AskSpec.cs
+++ b/src/core/Akka.Tests/Actor/AskSpec.cs
@@ -8,10 +8,12 @@
 using Akka.TestKit;
 using Xunit;
 using Akka.Actor;
+using Akka.Actor.Dsl;
 using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Util.Internal;
+using FluentAssertions;
 using Nito.AsyncEx;
 
 namespace Akka.Tests.Actor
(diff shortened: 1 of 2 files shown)


Arkatufus (approved):
LGTM

PR #5209, opened 2021-08-17
Akka.Cluster.Sharding performance benchmarks

Aaronontheweb:
close #3083

Initial numbers:

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19041.1165 (2004/May2020Update/20H1)
AMD Ryzen 7 1700, 1 CPU, 16 logical and 8 physical cores
.NET SDK=5.0.302
  [Host]     : .NET Core 3.1.17 (CoreCLR 4.700.21.31506, CoreFX 4.700.21.31502), X64 RyuJIT
  Job-HXSFLM : .NET Core 3.1.17 (CoreCLR 4.700.21.31506, CoreFX 4.700.21.31502), X64 RyuJIT

InvocationCount=1  UnrollFactor=1  

```
|                              Method |   StateMode | MsgCount |         Mean |      Error |      StdDev |       Median |
|------------------------------------ |------------ |--------- |-------------:|-----------:|------------:|-------------:|
|  **SingleRequestResponseToLocalEntity** | **Persistence** |    **10000** |   **110.158 ms** |  **2.1994 ms** |   **3.0832 ms** |   **109.456 ms** |
|              StreamingToLocalEntity | Persistence |    10000 |     5.156 ms |  0.3821 ms |   1.0964 ms |     4.934 ms |
| SingleRequestResponseToRemoteEntity | Persistence |    10000 | 4,157.097 ms | 82.5627 ms | 199.3982 ms | 4,069.618 ms |
|             StreamingToRemoteEntity | Persistence |    10000 |           NA |         NA |          NA |           NA |
|  **SingleRequestResponseToLocalEntity** |       **DData** |    **10000** |   **111.700 ms** |  **2.2282 ms** |   **6.0620 ms** |   **110.184 ms** |
|              StreamingToLocalEntity |       DData |    10000 |     4.897 ms |  0.3378 ms |   0.9360 ms |     4.858 ms |
| SingleRequestResponseToRemoteEntity |       DData |    10000 | 4,157.913 ms | 82.7139 ms | 165.1886 ms | 4,152.249 ms |
|             StreamingToRemoteEntity |       DData |    10000 |   506.119 ms | 10.1116 ms |  10.3839 ms |   504.516 ms |

Benchmarks with issues:
  ShardMessageRoutingBenchmarks.StreamingToRemoteEntity: Job-HXSFLM(InvocationCount=1, UnrollFactor=1) [StateMode=Persistence, MsgCount=10000]

[...]

PR #5210, opened 2021-08-17
fixed NRE inside ClusterSingletonManager

Aaronontheweb:
saw this error come up while running benchmarks in https://github.com/akkadotnet/akka.net/pull/5209


Aaronontheweb:
FYI, this usually happened when:

1. Current singleton manager was shutting down
2. Next node the singleton could be hosted on was _also_ shutting down

Issue #5213, opened 2021-08-19
IsNobody throws NRE


BaterPL:
**Version Information**
Version of Akka.NET? v1.4.24
Which Akka.NET Modules? Akka

**Describe the bug**
After changes in `IsNobody` implementation, our tests started failing because of NRE. It seems that the new implementation assumes that `Path` is not null which might not be true. In our case, it's caused by the mock of `IActorRef`.

**To Reproduce**
1. Prepare a test where you mock IActorRef using some mocking library (e.g. NSubstitute)
2. Call IsNobody() on mock
3. NRE is thrown

Code sample: https://github.com/BaterPL/akka-isNobody-example/blob/main/src/IsNobodyTests.cs

**Expected behavior**
IsNobody returns true instead of throwing an exception

**Actual behavior**
NullReferenceException

**Environment**
Windows, .NET 5

Aaronontheweb:
Thanks - we should patch this

jlaur:
See #5114

Aaronontheweb:
closed via https://github.com/akkadotnet/akka.net/pull/5220

PR #5220, opened 2021-08-24
fix IsNobody for actorref-mock and add unit test for it.

Zetanova:
fix and unit test for https://github.com/akkadotnet/akka.net/issues/5213


Aaronontheweb (approved):
LGTM - thanks @Zetanova

PR #5221, opened 2021-08-25
Ask should push unhandled answers into deadletter

Zetanova:
Currently invalid or timed out answers of Ask getting invisibly ignored by FutureActorRef<T>
This PR will push all invalid answers into deadletter.
Invalid answers are timeout-answer, too many answers, answer in a bad state of FutureActorRef<T>.

The only exception is a canceled Ask, this answer will still be ignored.

### DIFF
--- a/src/core/Akka/Actor/ActorRef.cs
+++ b/src/core/Akka/Actor/ActorRef.cs
@@ -97,11 +97,6 @@ namespace Akka.Actor
         /// </summary>
         public override IActorRefProvider Provider => _provider;
 
-
-        private const int INITIATED = 0;
-        private const int COMPLETED = 1;
-        private int status = INITIATED;
-
         /// <summary>
         /// TBD
         /// </summary>
@@ -113,7 +108,7 @@ namespace Akka.Actor
             {
                 SendSystemMessage(sysM);
             }
-            else if (Interlocked.Exchange(ref status, COMPLETED) == INITIATED)
+            else 
             {
                 var handled = false;
 
@@ -138,11 +133,6 @@ namespace Akka.Actor
                 //ignore canceled ask and put unhandled answers into deadletter
                 if (!handled && !_result.Task.IsCanceled)
                     _provider.DeadLetters.Tell(message ?? default(T), this);
-            }
-            else
-            {
-                //put too many answers into deadletter
-                _provider.DeadLetters.Tell(message ?? default(T), this);
             }
         }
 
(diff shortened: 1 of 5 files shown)


Aaronontheweb:
>but maybe for a future type some help to decide to use Interfaces or sealed classes

that takes me back to 2014, when I argued for using an `ActorRef` base class rather than an `IActorRef` interface - but I lost that argument unfortunately :p

Zetanova:
Your have been 100% on point, virt calls have a coast.
This has most of the time absolutely no impack.

But for IActorRef.Tel(msg) it has ... for sure.

src/core/Akka/Actor/ActorRef.cs
```diff
@@ -121,43 +104,37 @@ public override IActorRefProvider Provider
...
+            var handled = false;
 
-            if (message is ISystemMessage sysM) //we have special handling for system messages
-            {
-                SendSystemMessage(sysM);
-            }
-            else
+            switch (message)
             {
-                if (Interlocked.Exchange(ref status, COMPLETED) == INITIATED)
```
Aaronontheweb, line 131: Do we lose thread-safety without this? What happens if multiple threads try to complete the same Ask at the same time?

src/core/Akka/Actor/ActorRef.cs
```diff
@@ -121,43 +104,37 @@ public override IActorRefProvider Provider
         /// <param name="sender">TBD</param>
         protected override void TellInternal(object message, IActorRef sender)
         {
+            var handled = false;
```
Aaronontheweb, line 107: handled is a local variable, so if we attempt to send multiple messages to this `FutureActorRef<T>` this will evaluate to `true` each time, will it not?

Aaronontheweb (changes requested):
Not sure about thread-safety and re-entrancy ehere

Zetanova (commented):
@Aaronontheweb What is missing and what question is unanswered?

src/core/Akka/Actor/ActorRef.cs
```diff
@@ -121,43 +104,37 @@ public override IActorRefProvider Provider
...
-                        _result.TrySetException(f.Exception ?? new TaskCanceledException("Task cancelled by actor via Failure message."));
-                    }
-                    else
-                    {
-                        _result.TrySetException(new ArgumentException(
-                            $"Received message of type [{message.GetType()}] - Ask expected message of type [{typeof(T)}]"));
-                    }
-                }
+                case ISystemMessage sysM:
+                    SendSystemMessage(sysM); //we have special handling for system messages
```
Aaronontheweb, line 112: I would throw an exception here - not supposed to send any system messages to temporary actors

Aaronontheweb:
@Zetanova thanks! That answers both of those questions - only one last issue and then I think we're good to go:

https://github.com/akkadotnet/akka.net/pull/5221/files#r703000593

Would you mind submitting a new PR that implements that? Sorry for the screw-up here, again

PR #5226, opened 2021-08-26
Improve status messages

Zetanova:
`Akka.Actor.Failure` and `Akka.Actor.Failures` are unused and got obsoleted.

`Akka.Actor.Status.Success` and `Akka.Actor.Status.Failure` got sealed to improve pattern matching perf.
Failure got a new Field 'State' that the receiver can send the command or a correlation-id to the sender back.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
@@ -123,23 +123,27 @@ namespace Akka.Persistence.Sql.Common.Snapshot
             }
             catch (Exception e)
             {
-                return new Failure {Exception = e};
+                return new Status.Failure(e);
             }
         }
 
-        private bool WaitingForInitialization(object message) => message.Match()
-            .With<Initialized>(_ =>
+        private bool WaitingForInitialization(object message)
+        {
+            switch(message)
             {
-                UnbecomeStacked();
-                Stash.UnstashAll();
-            })
-            .With<Failure>(failure =>
-            {
-                Log.Error(failure.Exception, "Error during snapshot store initialization");
-                Context.Stop(Self);
-            })
-            .Default(_ => Stash.Stash())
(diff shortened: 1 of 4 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
```diff
@@ -123,23 +123,27 @@ private async Task<object> Initialize()
             }
             catch (Exception e)
             {
-                return new Failure {Exception = e};
+                return new Status.Failure(e);
```
Aaronontheweb, line 126: Is this going to break downstream Akka.Persistence.Sql plugin implementations @Zetanova ?

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1720,16 +1722,20 @@ namespace Akka.Actor
     public abstract class Status
     {
         protected Status() { }
-        public class Failure : Akka.Actor.Status
+        public sealed class Failure : Akka.Actor.Status
```
Aaronontheweb, line 1725: LGTM

Aaronontheweb (commented):
Had some questions on whether or not this will impact our Akka.Persistence.Sql plugins

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
Zetanova: It's only internally used with `Initialize().PipeTo(Self);`
The whole akka libery didn't use the `Akka.Actor.Failure` type

Aaronontheweb: Ok, sounds good.

Aaronontheweb (approved):
LGTM

PR #5229, opened 2021-08-26
singleton fix cleanup of overdue _removed members

zbynek001:
current logic seems reverted


Aaronontheweb (approved):
LGTM - definitely looks like a bug

Issue #5203, opened 2021-08-16
[PERF] Cluster Sharding perf issue


carl-camilleri-uom:
**Version Information**
Version of Akka.NET?  1.4.23
Which Akka.NET Modules? Akka.Cluster.Sharding

**Describe the performance issue**

 A minimum viable repo that reproduces the issue is at https://github.com/carlcamilleri/benchmark-akka-cluster

Running two n2-standard-8 nodes in GCP (8 CPU @ 2.80 GHz and 32GB RAM) with Windows Server 2019 in GCP ("instance-1" and "instance-2"), and a third machine to run the benchmarks from

First check:
`curl http://instance-1:5000/5`
Response: `akka://ping-pong-cluster-system/system/sharding/PingPongActor/13/5(pid=2372,hostname=instance-2)`

Therefore entity id 5 actor is hosted on instance-2 server

`wrk -t48 -c400 -d30s http://instance-2:5000/5`

```
Running 30s test @ http://instance-2:5000/5
48 threads and 400 connections
Thread Stats Avg Stdev Max +/- Stdev
Latency 19.54ms 70.10ms 1.09s 94.22%
Req/Sec 2.41k 429.13 12.12k 86.40%
3360471 requests in 30.10s, 666.60MB read
Socket errors: connect 0, read 0, write 202, timeout 0
Requests/sec: 111645.29
Transfer/sec: 22.15MB
```

`wrk -t48 -c400 -d30s http://instance-1:5000/5`
```

Running 30s test @ http://instance-1:5000/5
48 threads and 400 connections
Thread Stats Avg Stdev Max +/- Stdev
Latency 254.21ms 286.27ms 1.02s 78.54%
Req/Sec 119.09 127.71 0.94k 85.22%
73325 requests in 30.03s, 14.55MB read
Socket errors: connect 0, read 0, write 216, timeout 0
Requests/sec: 2441.41
Transfer/sec: 495.91KB
```

For interest I've also repeated test (1) (i.e. workload on the endpoint which requests the local actor) but with `serialize-messages = on`, and the result is:

```
Running 30s test @ http://instance-2:5000/5
48 threads and 400 connections
Thread Stats Avg Stdev Max +/- Stdev
Latency 24.77ms 90.16ms 1.15s 94.33%
Req/Sec 1.89k 339.47 4.94k 90.42%
2579946 requests in 30.08s, 511.77MB read
Socket errors: connect 0, read 0, write 246, timeout 0
[...]

PR #5232, opened 2021-08-29
Benchmark to test SingleRequestResponseToRemoteEntity with a local proxy

carlcamilleri:
Adding benchmark to test performance of sending messages to a remote entity without relying on remote `ASK` as per the approach discussed at https://github.com/akkadotnet/akka.net/issues/5203#issuecomment-907587322

### DIFF
--- a/src/benchmark/Akka.Cluster.Benchmarks/Sharding/ShardMessageRoutingBenchmarks.cs
+++ b/src/benchmark/Akka.Cluster.Benchmarks/Sharding/ShardMessageRoutingBenchmarks.cs
@@ -12,6 +12,7 @@ using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Benchmarks.Configurations;
 using Akka.Cluster.Sharding;
+using Akka.Routing;
 using BenchmarkDotNet.Attributes;
 using static Akka.Cluster.Benchmarks.Sharding.ShardingHelper;
 
@@ -33,6 +34,7 @@ namespace Akka.Cluster.Benchmarks.Sharding
 
         private IActorRef _shardRegion1;
         private IActorRef _shardRegion2;
+        private IActorRef _localRouter;
 
         private string _entityOnSys1;
         private string _entityOnSys2;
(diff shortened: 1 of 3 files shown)


src/benchmark/Akka.Cluster.Benchmarks/Sharding/ShardMessageRoutingBenchmarks.cs
```diff
@@ -43,6 +45,50 @@ public class ShardMessageRoutingBenchmarks
         private IActorRef _batchActor;
         private Task _batchComplete;
 
+#if (DEBUG)
```
Aaronontheweb, line 48: What's this for?

carlcamilleri: I was using this to debug the functionality of the benchmark (be able to attach breakpoints etc.) as per https://benchmarkdotnet.org/articles/guides/troubleshooting.html - if another approach is usually taken, happy to remove or change as needed

Thanks

Aaronontheweb: Anything different about code in this block vs. the `#else` block other than `async Task` vs. `void`? Was trying to spot some but none jumped out at me. The other statement, with the debugging config added, makes sense to me.

carlcamilleri: Correct I was running into an issue with async code in debug mode - the benchmark was executing before the task completion. I can refactor for a cleaner approach if it’s better (still would need two methods I believe due to async signature)

Aaronontheweb: Got it. Thank you!

Nope, that's ok - just adding a comment to explain why it's there would be sufficient. We can do that in a subsequent PR.

Thank you for your contribution!

PR #5236, opened 2021-08-31
Take Akka.Persistence.Sql.Common out of beta

Aaronontheweb:
If there were a major issue with the batching journals since v1.4.19, it would have been reported by now.


Arkatufus (approved):
LGTM

PR #5237, opened 2021-08-31
Refactor AdaptiveLoadBalancingRoutingLogic to use IClusterMetricsRoutingLogic

Arkatufus:


Aaronontheweb:
Even though we may not need this, I think it's a good change.

src/contrib/cluster/Akka.Cluster.Metrics/Routing/ClusterMetricsRouting.cs
```diff
@@ -328,7 +333,7 @@ public override Router CreateRouter(ActorSystem system)
         /// <inheritdoc />
         public override Props RoutingLogicController(RoutingLogic routingLogic)
         {
-            return Actor.Props.Create(() => new AdaptiveLoadBalancingMetricsListener(routingLogic as AdaptiveLoadBalancingRoutingLogic));
+            return Actor.Props.Create(() => new AdaptiveLoadBalancingMetricsListener((IClusterMetricsRoutingLogic)routingLogic));
```
Aaronontheweb, line 336: Blowing up here is definitely better than doing it as an NRE inside the `AdaptiveLoadBalancingMetricsListener`, IMHO.

PR #5238, opened 2021-08-31
Make `Router.Route` virtual

Aaronontheweb:


src/core/Akka/Routing/Router.cs
```diff
@@ -294,17 +292,14 @@ public Router(RoutingLogic logic, IActorRef routee, params IActorRef[] routees)
...
+            Routees = routees ?? Array.Empty<Routee>();
             RoutingLogic = logic;
         }
         
         /// <summary>
         /// The set of <see cref="Routees"/> that this router is currently targeting.
         /// </summary>
-        public IEnumerable<Routee> Routees
-        {
-            get { return _routees; }
```
Aaronontheweb, line 306: Potentially a breaking change, although in practice this code isn't ever used outside of Akka.NET's internals.

Changed this because there's zero benefit to returning an `IEnumable<Routee>` here and this actually results in more allocations from `.ToList()` calls throughout the downstream code within Akka.NET and we're better off just returning the underlying array directly - which is also what the `RoutingLogic` parameters demand.

Arkatufus: I guess the main argument against this is immutability? But of course the original code doesn't observe immutability either.

Aaronontheweb: Yeah you're right that the original design was probably to preserve immutability - or at least to make it difficult to get access to the underlying mutable array, even though `Router` itself is immutable by and large.

None of the code that adds or removes routees ever tries to modify the contents of this array directly so I think it's probably safe. The alternative being to keep this API as-is and have the subclassed routers in external projects (i.e. Phobos) maintain their own copy of the array or to change this type and the type that `RoutingLogic` uses to be an interface, which may introduce interface dispatch inside all routers... Which I'd prefer to avoid even though it's a very small performance penalty.

Arkatufus (approved):
LGTM

src/core/Akka/Routing/Router.cs
Aaronontheweb: Reverted this - as you said in chat, "too much work for a micro-optimization that probably doesn't improve performance much"

PR #5239, opened 2021-08-31
adding `GetChild` benchmark

Aaronontheweb:
Working on cleaning up the read-side resolve pipeline inside Akka.Cluster.Sharding and Akka.Remote. This seemed like a good place to start.

### DIFF
--- a/src/benchmark/Akka.Benchmarks/Actor/GetChildBenchmark.cs
+++ b/src/benchmark/Akka.Benchmarks/Actor/GetChildBenchmark.cs
@@ -1,0 +1,113 @@
+﻿// //-----------------------------------------------------------------------
+// // <copyright file="GetChildBenchmark.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2021 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2021 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Benchmarks.Configurations;
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Engines;
+using FluentAssertions;
+
+namespace Akka.Benchmarks.Actor
+{
+    /// <summary>
+    /// Used to measure how quickly an <see cref="IActorContext.Child"/> call can be executed in the wild.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


to11mtm:
How many layers deep do the children get here? Is there a way we can somehow measure difference in perf based on how deep the path is?

Aaronontheweb:
Data so far:

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19041.1165 (2004/May2020Update/20H1)
.NET SDK=5.0.302
  [Host]     : .NET Core 3.1.17 (CoreCLR 4.700.21.31506, CoreFX 4.700.21.31502), X64 RyuJIT
  DefaultJob : .NET Core 3.1.17 (CoreCLR 4.700.21.31506, CoreFX 4.700.21.31502), X64 RyuJIT


```
|       Method |     Mean |    Error |   StdDev | Allocated |
|------------- |---------:|---------:|---------:|----------:|
| ResolveChild | 62.07 ns | 0.971 ns | 0.811 ns |         - |

Aaronontheweb:
> How many layers deep do the children get here? Is there a way we can somehow measure difference in perf based on how deep the path is?

I can add a method to traverse the graph a few layers deeper, yes

Aaronontheweb:
Ideally, this is the path we want to optimize read-side in Akka.Remote:

https://github.com/akkadotnet/akka.net/blob/426111f48944a4c594a50b18c7f1afda019b8ee1/src/core/Akka/Actor/ActorRefProvider.cs#L465-L479

It gets called twice per-read and doesn't appear to be cached at all.

Aaronontheweb:
Ah, I take that back - looks like we do cache inside RARP here:

```csharp
/// <summary>
        /// Resolves a deserialized path into an <see cref="IActorRef"/>
        /// </summary>
        /// <param name="path">The path of the actor we are attempting to resolve.</param>
        /// <returns>A local <see cref="IActorRef"/> if it exists, <see cref="ActorRefs.Nobody"/> otherwise.</returns>
        public IActorRef ResolveActorRef(string path)
        {
            if (IgnoreActorRef.IsIgnoreRefPath(path))
                return IgnoreRef;

            // using thread local LRU cache, which will call InternalResolveActorRef
            // if the value is not cached
            if (_actorRefResolveThreadLocalCache == null)
            {
                return InternalResolveActorRef(path); // cache not initialized yet
            }
            return _actorRefResolveThreadLocalCache.Cache.GetOrCompute(path);
        }
```

Aaronontheweb:
Actually, funnily enough no - that doesn't get used at all inside the `AkkaPduCodec`

PR #5247, opened 2021-09-03
moved `IInboundMessageDispatcher` execution into separate sub-actor

Aaronontheweb:
This PR has the effect of temporally decoupling the `EndpointReader` from the `IInboundMessageDispatcher`, which means the work of parsing the Protobuf message and deserializing any user-defined messages are now asynchronous processes.

Message order is still preserved as the `MsgDispatcherActor` still runs serially and it runs on the same dispatcher as the rest of the remoting actors in order to ensure that  it similarly benefits from `ThreadLocal` cache re-use.

### DIFF
--- a/src/core/Akka.Remote/Endpoint.cs
+++ b/src/core/Akka.Remote/Endpoint.cs
@@ -1842,10 +1842,57 @@ namespace Akka.Remote
 
     }
 
+    internal sealed class MsgDispatcherActor : ReceiveActor
+    {
+        private readonly IInboundMessageDispatcher _msgDispatch;
+        private readonly ILoggingAdapter _log = Context.GetLogger();
+
+        public MsgDispatcherActor(IInboundMessageDispatcher dispatcher)
+        {
+            _msgDispatch = dispatcher;
+            
+            Receive<AckAndMessage>(ackAndMessage =>
+            {
+                try
+                {
+                    _msgDispatch.Dispatch(ackAndMessage.MessageOption.Recipient,
+                        ackAndMessage.MessageOption.RecipientAddress,
+                        ackAndMessage.MessageOption.SerializedMessage,
+                        ackAndMessage.MessageOption.SenderOptional);
+                }
+                catch (SerializationException e)
+                {
+                    LogTransientSerializationError(ackAndMessage.MessageOption, e);
+                }
+                catch (ArgumentException e)
+                {
+                    LogTransientSerializationError(ackAndMessage.MessageOption, e);
+                }
+                catch (InvalidCastException e)
+                {
+                    LogTransientSerializationError(ackAndMessage.MessageOption, e);
+                }
+            });
+        }
+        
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Performance numbers:

## Dev

```
OSVersion:                         Microsoft Windows NT 6.2.9200.0 
Messages sent/received per client: 200000  (2e5)                   
Thread count:                      111                             
                                                                   
         1,  200000,    122175,    1637.86                         
         5, 1000000,    241197,    4146.44                         
        10, 2000000,    244469,    8181.71                         
        15, 3000000,    242994,   12346.94                         
        20, 4000000,    241488,   16564.95                         
        25, 5000000,    240686,   20774.23                         
        30, 6000000,    240520,   24946.24                         
```

## This PR

```
Messages sent/received per client: 200000  (2e5)
Thread count:                      111          
                                                
         1,  200000,    139373,    1435.28      
         5, 1000000,    276015,    3623.40      
        10, 2000000,    274613,    7283.22      
        15, 3000000,    270393,   11095.59      
        20, 4000000,    268727,   14885.98      
        25, 5000000,    266909,   18733.29      
        30, 6000000,    266726,   22495.50      
```

Aaronontheweb:
Akka.Remote test suite not liking this change so far - not clear why. Smells like an issue with remotely deployed actors.

Aaronontheweb:
> Akka.Remote test suite not liking this change so far - not clear why. Smells like an issue with remotely deployed actors.

Figured that one out - missed a stop inside the `DeliverAndAck` method where the dispatcher gets invoked.

src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1842,10 +1842,57 @@ public OutboundAck(Ack ack)
 
     }
 
+    internal sealed class MsgDispatcherActor : ReceiveActor
```
to11mtm, line 1845: Since this is a hot path and we don't need `ReceiveAsync` behavior here, should this be a more base actor to save some delegate calls?

Aaronontheweb: Done - measuring this change now...

Aaronontheweb: Performance appears to be noticeably worse - 20k msg/s worse with `UntypedActor` and an `if... is` statement than a `ReceiveActor`.

Been able to reproduce it in both directions several times - going to revert probably.

Aaronontheweb: Done - reverted it. I think it was @ismaelhamed who pointed out that the `UntypedActor` itself may be faster, but the way `Receive<T>` works in the receive actor is inherently faster than some of what happens when powering `switch` and `if .. is _` statements. For whatever reason, that appears to be what's at work here.

Aaronontheweb:
Still having some issues here in the test suite - this change probably won't make it into v1.4.25.

PR #5251, opened 2021-09-03
Make sure stdout-loglevel setting is honored through the whole actor system lifecycle

### LINKED ISSUE #5246: Akka.Actor: need to enforce `stdout-loglevel = off` all the way through `ActorSystem` lifecycle
> The StandardOutLogger gets started at the very end to capture any output from shutting down the other parts of the logging system as it shuts down, so it happens at the very, very end. There shouldn't be anything else important happening by this point as the rest of the application has already been terminated by the time the ActorSystem shutdown process reaches this point.

@Aaronontheweb In my case it's very unfortunate - I have created simple console app which gets invoked with commands then do it's job and quits - like Git, for example _git status_ shows current status then quits app.

I designed it with Akka.Net - I know downsides of this decision, I know that every time I invoke app, whole actor system must be created and in my case it's acceptable performance, so this is not the subject to this comment (in the future I can move logic to separate long living process, like Docker does, invoking _docker ps_ doesn't start up whole _Docker_, it just connects to deamon).

Of course I want logs, so I have configured _NLog_ to save logs to file. But I never want them in console output, as this is my _user interface_, so I have explicitly configured `stdout-loglevel = off`

But with akka's design quoted in the beginning of this comment, every time job is done I see in my console output:

```
[DEBUG][2021-09-02 07:50:46][Thread 0009][EventStream] Shutting down: StandardOutLogger started
[DEBUG][2021-09-02 07:50:46][Thread 0009][EventStream] unsubscribing [akka://Q/system/log1-NLogLogger#918463714] from all channels
```

So to understand why I don't want this, imagine I that I now want to create _new Git version_ rewritten with Akka.Net and every time user wants to see repo status sees something like this:
```
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
[DEBUG][2021-09-02 07:50:46][Thread 0009][EventStream] Shutting down: StandardOutLogger started
[DEBUG][2021-09-02 07:50:46][Thread 0009][EventStream] unsubscribing [akka://Q/system/log1-NLogLogger#918463714] from all channels
```

In my view (and this is what I asks for) this behavior should be changed and starts to respect `stdout-loglevel` log level (which can be set to _off_) or maybe some new configuration should be added, something like `start-up-stdout-during-shutdown`.

_Originally posted by @tometchy in https://github.com/akkadotnet/akka.net/issues/4045#issuecomment-911355093_

Arkatufus:
Closes #5246

### DIFF
--- a/docs/articles/utilities/logging.md
+++ b/docs/articles/utilities/logging.md
@@ -25,6 +25,28 @@ Akka.NET comes with two built in loggers.
 * __StandardOutLogger__
 * __BusLogging__
 
+### StandardOutLogger
+`StandardOutLogger` is considered as a minimal logger and implements the `MinimalLogger` abstract
+class. Its job is simply to output all `LogEvent`s emitted by the `EventBus` onto the console. 
+Since it is not an actual actor, ie. it doesn't need the `ActorSystem` to operate, it is also 
+used to log other loggers activity at the very start and very end of the `ActorSystem` life cycle. 
+You can change the minimal logger start and end life cycle behaviour by changing the 
+`akka.stdout-loglevel` HOCON settings to `OFF` if you do not need these feature in your application.
+
+### Advanced MinimalLogger Setup
+You can also replace `StandardOutLogger` by making your own logger class with an empty constructor 
+that inherits/implements the `MinimalLogger` abstract class and passing the fully qualified class 
+name into the `akka.stdout-logger-class` HOCON settings. 
+
+> [!WARNING]
+> Be aware that `MinimalLogger` implementations are __NOT__ real actors and will __NOT__ have any 
+> access to the `ActorSystem` and all of its extensions. All logging done inside a `MinimalLogger` 
+> have to be done in as simple as possible manner since it is used to log how other loggers are 
+> behaving at the very start and very end of the `ActorSystem` life cycle.
+> 
+> Note that `MinimalLogger` are __NOT__ interchangeable with other Akka.NET loggers and there can 
+> only be one `MinimalLogger` registered with the `ActorSystem` in the HOCON settings.
+
 ## Contrib Loggers
 These loggers are also available as separate nuget packages
 
--- a/src/contrib/testkits/Akka.TestKit.Xunit/Internals/Loggers.cs
+++ b/src/contrib/testkits/Akka.TestKit.Xunit/Internals/Loggers.cs
@@ -5,6 +5,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System;
 using Akka.Actor;
 using Akka.Event;
 using Xunit.Abstractions;
@@ -16,20 +17,44 @@ namespace Akka.TestKit.Xunit.Internals
     /// </summary>
     public class TestOutputLogger : ReceiveActor
     {
+        private readonly ITestOutputHelper _output;
+        
         /// <summary>
         /// Initializes a new instance of the <see cref="TestOutputLogger"/> class.
         /// </summary>
         /// <param name="output">The provider used to write test output.</param>
         public TestOutputLogger(ITestOutputHelper output)
         {
-            Receive<Debug>(e => output.WriteLine(e.ToString()));
-            Receive<Info>(e => output.WriteLine(e.ToString()));
-            Receive<Warning>(e => output.WriteLine(e.ToString()));
-            Receive<Error>(e => output.WriteLine(e.ToString()));
+            _output = output;
+            
+            Receive<Debug>(Write);
+            Receive<Info>(Write);
+            Receive<Warning>(Write);
+            Receive<Error>(Write);
             Receive<InitializeLogger>(e =>
             {
                 e.LoggingBus.Subscribe(Self, typeof (LogEvent));
             });
+        }
+
+        private void Write(LogEvent e)
+        {
+            try
+            {
+                _output.WriteLine(e.ToString());
+            }
+            catch (FormatException ex)
+            {
+                if (e.Message is LogMessage msg)
+                {
+                    var message =
+                        $"Received a malformed formatted message. Log level: [{e.LogLevel()}], Template: [{msg.Format}], args: [{string.Join(",", msg.Args)}]";
+                    if(e.Cause != null)
+                        throw new AggregateException(message, ex, e.Cause);
+                    throw new FormatException(message, ex);
+                }
+                throw;
+            }
         }
     }
 }
--- a/src/core/Akka.TestKit/EventFilter/TestEventListener.cs
+++ b/src/core/Akka.TestKit/EventFilter/TestEventListener.cs
@@ -30,57 +30,65 @@ namespace Akka.TestKit
         /// <returns>TBD</returns>
         protected override bool Receive(object message)
         {
-            if(message is InitializeLogger)
+            switch (message)
             {
-                var initLogger = (InitializeLogger)message;
-                var bus = initLogger.LoggingBus;
-                var self = Context.Self;
-                bus.Subscribe(self, typeof(Mute));
-                bus.Subscribe(self, typeof(Unmute));
-                bus.Subscribe(self, typeof(DeadLetter));
-                bus.Subscribe(self, typeof(UnhandledMessage));
-                Sender.Tell(new LoggerInitialized());
-            }
-            else if(message is Mute)
-            {
-                var mute = (Mute)message;
-                foreach(var filter in mute.Filters)
+                case InitializeLogger initLogger:
                 {
-                    AddFilter(filter);
+                    base.Receive(message);
+                    var bus = initLogger.LoggingBus;
+                    var self = Context.Self;
+                    bus.Subscribe(self, typeof(Mute));
+                    bus.Subscribe(self, typeof(Unmute));
+                    bus.Subscribe(self, typeof(DeadLetter));
+                    bus.Subscribe(self, typeof(UnhandledMessage));
+                    Sender.Tell(new LoggerInitialized());
+                    break;
                 }
-            }
-            else if(message is Unmute)
-            {
-                var unmute = (Unmute)message;
-                foreach(var filter in unmute.Filters)
+                case Mute mute:
                 {
-                    RemoveFilter(filter);
+                    foreach(var filter in mute.Filters)
+                    {
+                        AddFilter(filter);
+                    }
+
+                    break;
                 }
-            }
-            else if(message is LogEvent)
-            {
-                var logEvent = (LogEvent)message;
-                if(!ShouldFilter(logEvent))
+                case Unmute unmute:
                 {
-                    Print(logEvent);
+                    foreach(var filter in unmute.Filters)
+                    {
+                        RemoveFilter(filter);
+                    }
+
+                    break;
                 }
+                case LogEvent logEvent:
+                {
+                    if(!ShouldFilter(logEvent))
+                    {
+                        Print(logEvent);
+                    }
+
+                    break;
+                }
+                case DeadLetter letter:
+                    HandleDeadLetter(letter);
+                    break;
+                
+                case UnhandledMessage un:
+                {
+                    var rcp = un.Recipient;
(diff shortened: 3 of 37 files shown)


src/contrib/testkits/Akka.TestKit.Xunit/Internals/Loggers.cs
```diff
@@ -16,20 +17,44 @@ namespace Akka.TestKit.Xunit.Internals
...
             {
                 e.LoggingBus.Subscribe(Self, typeof (LogEvent));
             });
         }
+
+        private void Write(LogEvent e)
+        {
+            try
+            {
+                _output.WriteLine(e.ToString());
```
Arkatufus, line 44: Add missing format exception catcher to Akka.TestKit.Xunit that was added to Akka.TestKit.Xunit2

src/core/Akka.TestKit/EventFilter/TestEventListener.cs
```diff
@@ -30,57 +30,65 @@ public class TestEventListener : DefaultLogger
         /// <returns>TBD</returns>
         protected override bool Receive(object message)
         {
-            if(message is InitializeLogger)
```
Arkatufus, line 33: Change if...then...else structure to switch

src/core/Akka/Actor/Settings.cs
```diff
@@ -109,6 +110,31 @@ public Settings(ActorSystem system, Config config, ActorSystemSetup setup)
 
             LogLevel = Config.GetString("akka.loglevel", null);
             StdoutLogLevel = Config.GetString("akka.stdout-loglevel", null);
+
+            var stdoutClassName = Config.GetString("akka.stdout-logger-class", null);
```
Arkatufus, line 114: Moved the hard-wired StandardOutLogger instance from the Logging class to the Settings class.
Default StandardOutLogger is now customizable through the HOCON configuration

src/core/Akka/Configuration/Pigeon.conf
```diff
@@ -42,6 +42,10 @@ akka {
   # Log level for the very basic logger activated during AkkaApplication startup
   # Options: OFF, ERROR, WARNING, INFO, DEBUG
   stdout-loglevel = "WARNING"
+  
+  # Fully qualified class name (FQCN) of the very basic logger used on startup and shutdown.
+  # You can substitute the logger by supplying an FQCN to a custom class that implements Akka.Event.MinimumLogger
+  stdout-logger-class = "Akka.Event.StandardOutLogger"
```
Arkatufus, line 48: New HOCON setting to customize the standard out logger

src/core/Akka/Event/LogLevel.cs
```diff
@@ -31,6 +31,11 @@ public enum LogLevel
         /// The error log level.
         /// </summary>
         ErrorLevel,
+        
+        /// <summary>
+        /// The off log level.
+        /// </summary>
+        OffLevel = int.MaxValue
```
Arkatufus, line 38: Added OffLevel to LogLevel enum, this would act as a marker for the off state

src/core/Akka/Event/StandardOutLogger.cs
```diff
@@ -13,6 +13,38 @@
 
 namespace Akka.Event
 {
+    public abstract class MinimalLogger : MinimalActorRef
```
Arkatufus, line 16: MinimalLogger is the abstract base class for implementing minimal loggers.

src/contrib/testkits/Akka.TestKit.Xunit/Internals/Loggers.cs
Aaronontheweb: LGTM

src/core/Akka.Tests/Loggers/ShutdownLoggerSpec.cs
```diff
@@ -0,0 +1,84 @@
...
+akka.loglevel = OFF
+akka.stdout-loglevel = OFF
+akka.stdout-logger-class = ""Akka.Tests.Loggers.ThrowingLogger, Akka.Tests""");
+
+        public ShutdownLoggerSpec(ITestOutputHelper output) : base(Config, output)
+        {
+        }
+
+        [Fact(DisplayName = "StandardOutLogger should not be called on shutdown when stdout-loglevel is set to OFF")]
+        public async Task StandardOutLoggerShouldNotBeCalled()
```
Aaronontheweb, line 35: LGTM

src/core/Akka/Actor/Settings.cs
Aaronontheweb: I believe this was @zbynek001's suggestion in one of those threads - we should document this inside https://getakka.net/articles/utilities/logging.html#standard-loggers

src/core/Akka/Actor/Settings.cs
```diff
@@ -109,6 +110,31 @@ public Settings(ActorSystem system, Config config, ActorSystemSetup setup)
...
+                if(!typeof(MinimalLogger).IsAssignableFrom(stdoutLoggerType))
+                    throw new ArgumentException("Standard out logger type must inherit from the MinimalLogger abstract class.");
+
+                try
+                {
+                    StdoutLogger = (MinimalLogger)Activator.CreateInstance(stdoutLoggerType);
+                }
+                catch (MissingMethodException)
+                {
+                    throw new MissingMethodException(
```
Aaronontheweb, line 133: LGTM

src/core/Akka/Event/LoggerInitialized.cs
```diff
@@ -12,7 +12,7 @@ namespace Akka.Event
     /// <summary>
     /// This class represents a message used to notify subscribers that a logger has been initialized.
     /// </summary>
-    public class LoggerInitialized : INoSerializationVerificationNeeded
+    public class LoggerInitialized : INoSerializationVerificationNeeded, IDeadLetterSuppression
```
Aaronontheweb, line 15: LGTM

src/core/Akka/Event/LogLevel.cs
Aaronontheweb: I'd revert this change - it's a major part of the public API and will require an update to all downstream loggers probably.

Aaronontheweb (changes requested):
We should probably leave the `LogLevel` enum alone - as that's a pretty viral change without much benefit, but other than that this PR looks good. Need to add some documentation as well.

src/core/Akka/Event/Logging.cs
```diff
@@ -193,8 +187,6 @@ public static string StringFor(this LogLevel logLevel)
                     return Warning;
                 case LogLevel.ErrorLevel:
                     return Error;
-                case OffLogLevel:
-                    return Off;
```
Aaronontheweb, line 197: Need to add this back

src/core/Akka/Event/Logging.cs
```diff
@@ -145,12 +145,6 @@ public static string SimpleName(Type t)
         private const string Warning = "WARNING";
         private const string Error = "ERROR";
         private const string Off = "OFF";
-        private const LogLevel OffLogLevel = (LogLevel) int.MaxValue;
```
Aaronontheweb, line 148: Need to add this back

src/core/Akka/Event/Logging.cs
```diff
@@ -284,7 +276,7 @@ public static LogLevel LogLevelFor(string logLevel)
                 case Error:
                     return LogLevel.ErrorLevel;
                 case Off:
-                    return OffLogLevel;
+                    return (LogLevel) int.MaxValue;
```
Aaronontheweb, line 279: Should add back previous call

Aaronontheweb (changes requested):
Some more cleanup needed to undo the `LogLevel.Off` change

src/core/Akka/Event/StandardOutLogger.cs
```diff
@@ -107,7 +119,7 @@ protected override void TellInternal(object message, IActorRef sender)
         /// Prints a specified event to the console.
         /// </summary>
         /// <param name="logEvent">The event to print</param>
-        public static void PrintLogEvent(LogEvent logEvent)
+        internal static void PrintLogEvent(LogEvent logEvent)
```
Aaronontheweb, line 122: Good call

src/core/Akka/Event/StandardOutLogger.cs
Aaronontheweb: In a separate PR, we should add some documentation for this

Aaronontheweb (approved):
LGTM

Arkatufus:
Don't merge just yet, I just spotted a few problems that needs to be changed, plus the documentation.

Arkatufus:
Done, documentation added.

Arkatufus:
Should be ready for review now

docs/articles/utilities/logging.md
```diff
@@ -25,6 +25,28 @@ Akka.NET comes with two built in loggers.
...
+### StandardOutLogger
+`StandardOutLogger` is considered as a minimal logger and implements the `MinimalLogger` abstract
+class. Its job is simply to output all `LogEvent`s emitted by the `EventBus` onto the console. 
+Since it is not an actual actor, ie. it doesn't need the `ActorSystem` to operate, it is also 
+used to log other loggers activity at the very start and very end of the `ActorSystem` life cycle. 
+You can change the minimal logger start and end life cycle behaviour by changing the 
+`akka.stdout-loglevel` HOCON settings to `OFF` if you do not need these feature in your application.
+
+### Advanced MinimalLogger Setup
+You can also replace `StandardOutLogger` by making your own logger class with an empty constructor
```
Aaronontheweb, line 37: Probably need to do more showing rather than telling here, but this is good for now

Aaronontheweb (approved):
LGTM

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -226,7 +226,7 @@ public void StartStdoutLogger(Settings config)
         private void SetUpStdoutLogger(Settings config)
         {
             var logLevel = Logging.LogLevelFor(config.StdoutLogLevel);
-            SubscribeLogLevelAndAbove(logLevel, Logging.StandardOutLogger);
+            SubscribeLogLevelAndAbove(logLevel, config.StdoutLogger);
```
tometchy, line 229: This line change is what PR title says, right?
Before change in this line (229) `akka.stdout-loglevel` wasn't honored and now, after this line has changed, it is honored, so _off_ means off even during shutdown, right?

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -214,9 +214,9 @@ private string CreateLoggerName(Type actorClass)
         }
 
         /// <summary>
-        /// Starts the <see cref="StandardOutLogger"/> logger.
+        /// Starts the <see cref="MinimalLogger"/> logger.
         /// </summary>
-        /// <param name="config">The configuration used to configure the <see cref="StandardOutLogger"/>.</param>
+        /// <param name="config">The configuration used to configure the <see cref="MinimalLogger"/>.</param>
         public void StartStdoutLogger(Settings config)
```
tometchy, line 220: If I get it right, it's a bit misleading, that if you set your own _MinimalLogger_ then you need to configure it's logging level through `akka.stdout-loglevel` not for example `akka.minimallogger-loglevel`.

src/core/Akka/Event/LoggingBus.cs
Arkatufus: Now, when you set `stdout-loglevel`, nothing will be outputted by the StandardOutLogger, correct.

src/core/Akka/Event/LoggingBus.cs
Arkatufus: StandardOutLogger derives from MinimalLogger, it is the barest minimum logger used to log how the system behave without actual loggers running, ie. a logger to log the loggers themselves, it runs at the very start and the very end of the ActorSystem life cycle. The only thing you can make by inheriting MinimalLogger is a replacement for the StandardOutLogger, thus it will use the same settings as StandardOutLogger.

PR #5252, opened 2021-09-04
Added TakeUntil stage

ptjhuang:
Small addition, but seems to me a common requirement.  Differs from TakeWhile in that it does not pull the next element from upstream after the terminating condition is met.

### DIFF
--- a/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
+++ b/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
@@ -288,6 +288,99 @@ namespace Akka.Streams.Implementation.Fusing
         public override string ToString() => "TakeWhile";
     }
 
+
+    /// <summary>
+    /// Take until
+    /// </summary>
+    /// <typeparam name="T">TBD</typeparam>
+    [InternalApi]
+    public sealed class TakeUntil<T> : SimpleLinearGraphStage<T>
+    {
+        #region Logic
+
+        private sealed class Logic : InAndOutGraphStageLogic
+        {
+            private readonly TakeUntil<T> _stage;
+            private readonly Decider _decider;
+
+            public Logic(TakeUntil<T> stage, Attributes inheritedAttributes) : base(stage.Shape)
+            {
+                _stage = stage;
+                var attr = inheritedAttributes.GetAttribute<ActorAttributes.SupervisionStrategy>(null);
+                _decider = attr != null ? attr.Decider : Deciders.StoppingDecider;
+
+                SetHandler(stage.Outlet, this);
+                SetHandler(stage.Inlet, this);
+            }
+
+            public override void OnPush()
+            {
+                try
+                {
+                    var element = Grab(_stage.Inlet);
+                    if (!_stage._predicate(element))
+                        Push(_stage.Outlet, element);
+                    else
+                    {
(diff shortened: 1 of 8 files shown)


src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -288,6 +288,99 @@ protected override GraphStageLogic CreateLogic(Attributes inheritedAttributes)
...
+                SetHandler(stage.Outlet, this);
+                SetHandler(stage.Inlet, this);
+            }
+
+            public override void OnPush()
+            {
+                try
+                {
+                    var element = Grab(_stage.Inlet);
+                    if (!_stage._predicate(element))
```
ismaelhamed, line 321: All this code just to end up inverting this statement, are we missing something?

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -288,6 +288,99 @@ protected override GraphStageLogic CreateLogic(Attributes inheritedAttributes)
...
+                        Push(_stage.Outlet, element);
+                    else
+                    {
+                        if (_stage._inclusive)
```
ismaelhamed, line 325: Shouldn't `TakeUntil` always be inclusive?

ptjhuang: Agreed - TakeUntil exclusive is basically TakeWhile exclusive with an inverted predicate

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
ptjhuang: Agreed - extracted the Logic out of TakeWhile and shared it.

ismaelhamed:
I did a little bit of research, and the `TakeUntil` operator either was part of Akka at some point or they were about to include it but eventually settled with adding the `inclusive` param to `TakeWhile`. 

But that doesn't seem to cover the use case for this PR.

I know the Lightbend team is hesitant of adding new operators (to avoid an explosion, since one may argue that there are already too many operators as it is) and favor combining the existing ones instead. But in this case, IMO, it won't harm having this operator as it is very common in other libs too (for instance, Rx).

cc: @Aaronontheweb

Aaronontheweb:
> I did a little bit of research, and the `TakeUntil` operator either was part of Akka at some point or they were about to include it but eventually settled with adding the `inclusive` param to `TakeWhile`.
> 
> But that doesn't seem to cover the use case for this PR.
> 
> I know the Lightbend team is hesitant of adding new operators (to avoid an explosion, since one may argue that there are already too many operators as it is) and favor combining the existing ones instead. But in this case, IMO, it won't harm having this operator as it is very common in other libs too (for instance, Rx).
> 
> cc: @Aaronontheweb

So, should we just refactor this operator to do that then? Agree with the reasoning

ismaelhamed:
> So, should we just refactor this operator to do that then? Agree with the reasoning
@Aaronontheweb , @ptjhuang already extracted the `TakeWhile` StageLogic and is currently shared between both operators. The question is if this something to we want in Akka Streams core, or should we have something like an [Akka.Streams.Contrib](https://github.com/akka/akka-stream-contrib)

PR #5256, opened 2021-09-06
harden Bugfix4360Spec

Aaronontheweb:
The `AwaitAssert` is actually what made this spec racy

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Bugfix4360Spec.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Bugfix4360Spec.cs
@@ -136,14 +136,10 @@ auto-initialize = on
             ExpectTerminated(recoveryActor);
 
             // recreate the actor and recover
+            var recoveryActor2 = Sys.ActorOf(Props.Create(() => new RecoverActor(TestActor)), recoveryActor.Path.Name);
 
-            AwaitAssert(() =>
-            {
-                var recoveryActor2 = Sys.ActorOf(Props.Create(() => new RecoverActor(TestActor)), recoveryActor.Path.Name);
-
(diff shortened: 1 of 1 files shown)


src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Bugfix4360Spec.cs
```diff
@@ -136,14 +136,10 @@ public void Should_recover_persistentactor_sqlite()
             ExpectTerminated(recoveryActor);
 
             // recreate the actor and recover
+            var recoveryActor2 = Sys.ActorOf(Props.Create(() => new RecoverActor(TestActor)), recoveryActor.Path.Name);
```
Aaronontheweb, line 139: Once the `Terminated` is received, this actor should be safe to recreate again right away.

If it's not - we should do a "double-tap" and get a second `ExpectTerminated` instead of an `AwaitAssert` in this case.

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Bugfix4360Spec.cs
```diff
@@ -136,14 +136,10 @@ public void Should_recover_persistentactor_sqlite()
...
 
-            AwaitAssert(() =>
-            {
-                var recoveryActor2 = Sys.ActorOf(Props.Create(() => new RecoverActor(TestActor)), recoveryActor.Path.Name);
-
-                var r2 = ExpectMsg<IEnumerable<string>>(TimeSpan.FromMilliseconds(100));
-                r2.Should().Contain(new[] { "foo", "bar" });
-            }, interval:TimeSpan.FromMilliseconds(150));
+            var r2 = ExpectMsg<IEnumerable<string>>();
```
Aaronontheweb, line 141: Use a longer timeout instead of `AwaitAssert`

PR #5280, opened 2021-09-15
Add backward compatibility to PrimitiveSerializers

### LINKED ISSUE #5279: #4989 causes backward compatibility problem
#4989 causes new version to old version serialization problem, need to add backward compatibility.

Reported in: https://github.com/akkadotnet/akka.net/pull/4989#issuecomment-920358590

Arkatufus:
Closes #5279

### DIFF
--- a/src/core/Akka.Remote/Serialization/PrimitiveSerializers.cs
+++ b/src/core/Akka.Remote/Serialization/PrimitiveSerializers.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Text;
 using Akka.Actor;
+using Akka.Configuration;
 using Akka.Serialization;
 using Akka.Util;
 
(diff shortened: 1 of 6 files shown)


src/core/Akka.Remote/Serialization/PrimitiveSerializers.cs
```diff
@@ -78,6 +85,9 @@ public override object FromBinary(byte[] bytes, string manifest)
         /// <inheritdoc />
         public override string Manifest(object obj)
         {
+            if (_useNeutralPrimitives)
+                return obj.GetType().TypeQualifiedName();
```
Arkatufus, line 89: Need to emulate the old serializer behaviour, which is to embed the type name as manifest.
Related code:
https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote/MessageSerializer.cs#L62-L74

src/core/Akka.Remote/Serialization/PrimitiveSerializers.cs
```diff
@@ -29,12 +30,18 @@ public sealed class PrimitiveSerializers : SerializerWithStringManifest
...
         internal const string Int64ManifestNetFx = "System.Int64, mscorlib";
 
+        private readonly bool _useNeutralPrimitives;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="PrimitiveSerializers" /> class.
         /// </summary>
         /// <param name="system">The actor system to associate with this serializer. </param>
-        public PrimitiveSerializers(ExtendedActorSystem system) : base(system)
+        public PrimitiveSerializers(ExtendedActorSystem system, Config config) : base(system)
```
Aaronontheweb, line 39: Do we need the config parameter here or can that be extracted from `System.Settings`?

Arkatufus: Have to do this here because of our own convention.
Serializer settings are injected into the serializer from `akka.actor.serialization-settings`, and if the configuration for specific serializer is defined in this configuration setting, the serializer is then instantiated with the configuration injected in the constructor.
The code will fail if that config parameter is ommited.

Arkatufus: Do you prefer to have this setting separate from our convention on providing settings to serializers?

Aaronontheweb: This is fine then

Aaronontheweb (approved):
LGTM - please send in a second PR with reference documentation in HOCON for this

PR #5281, opened 2021-09-16
Add documentation for PrimitiveSerializers backward compatibility feature

Arkatufus:


docs/articles/remoting/transports.md
```diff
@@ -155,5 +161,27 @@ Why is this distinction important? Why do we care about registering an publicly
...
+
+In order for a v1.4.26 and above system to communicate properly with v1.4.19 and below systems, 
+you will have to  add this into the HOCON settings:
+```
+akka.actor.serialization-settings.primitive.use-neutral-primitives = true
+```
+This will revert the new serializer to the old behaviour and allows them to be completely compatible
+with older versions of Akka.NET.
+
+### Rolling an Update From v1.4.19 to v1.4.26
```
Aaronontheweb, line 180: This oversells the problem, IMHO - the issue only occurs when string, int, etc are sent directly over the wire without a wrapper message. It's been reported once in the several months since v1.4.20 has been out.

docs/articles/remoting/transports.md
```diff
@@ -155,5 +161,27 @@ Why is this distinction important? Why do we care about registering an publicly
...
+akka.actor.serialization-settings.primitive.use-neutral-primitives = true
+```
+This will revert the new serializer to the old behaviour and allows them to be completely compatible
+with older versions of Akka.NET.
+
+### Rolling an Update From v1.4.19 to v1.4.26 
+To do a rolling update between v1.4.19 and below systems to v1.4.26 and above, you will have to
+do 2 rolling updates:
+- Roll an update of __all__ v1.4.19 and below systems to v1.4.26 and above using the setting described above.
+- Roll another update with the setting removed.
```
Aaronontheweb, line 184: You don't need to do this actually - it's just a single upgrade.

Aaronontheweb: We will turn this setting off by default in v1.5

docs/articles/remoting/transports.md
```diff
@@ -13,6 +13,12 @@ In this section we'll expand a bit more on what transports are and how Akka.Remo
 ## What are Transports?
 Transports in Akka.Remote are abstractions on top of actual network transports, such as TCP and UDP sockets, and in truth transports have pretty simple requirements.
 
+> [!WARNING]
+> Due to a bug in the remoting serializer, you __can not__ mix v1.4.20 - v1.4.25 systems with __older__
```
Aaronontheweb, line 17: Again, this oversells the issue - it makes it sound like the entire remoting system is torched, not one particular serializer.

docs/articles/remoting/transports.md
```diff
@@ -155,5 +161,27 @@ Why is this distinction important? Why do we care about registering an publicly
...
+serialized message size. This change introduces a bug that causes remote communication problem 
+between systems with v1.4.20 and above and systems with v1.4.19 and below. However, this bug was not 
+found until v1.4.25, which means that __v1.4.20 - v1.4.25 can not be mixed with any older 
+versions of Akka.NET__. In order for a new system to communicate with systems with v1.4.19 or below, 
+you will have to use v1.4.26 or above.
+
+In order for a v1.4.26 and above system to communicate properly with v1.4.19 and below systems, 
+you will have to  add this into the HOCON settings:
+```
+akka.actor.serialization-settings.primitive.use-neutral-primitives = true
```
Aaronontheweb, line 175: Isn't this automatically added via HOCON when you upgrade to v1.4.26?

Aaronontheweb (changes requested):
Feedback

docs/articles/remoting/transports.md
Arkatufus: the default is off, turning it on will make v1.4.26 compatible with v1.4.19 and below

Aaronontheweb:
Closed in favor of https://github.com/akkadotnet/akka.net/pull/5289

PR #5284, opened 2021-09-16
Reverse the MinimalLogger feature PR

Arkatufus:
The stdout-loglevel changes are still intact, just removing the MinimalLogger changes.

### DIFF
--- a/src/core/Akka/Event/StandardOutLogger.cs
+++ b/src/core/Akka/Event/StandardOutLogger.cs
@@ -13,12 +13,20 @@ using System.Threading;
 
 namespace Akka.Event
 {
-    public abstract class MinimalLogger : MinimalActorRef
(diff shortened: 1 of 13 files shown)


src/core/Akka/Event/StandardOutLogger.cs
```diff
@@ -115,6 +101,11 @@ protected override void Log(object message)
         /// </summary>
         public static bool UseColors { get; set; }
 
+        /// <summary>
+        /// FOR TESTING PURPOSES ONLY, DO NOT USE
+        /// </summary>
+        internal event Action<object> Invoked;
```
Arkatufus, line 107: This is the only addition to the code, I have to add this in for testing because there is no other way to check if StandardOutLogger was called or not.

PR #5288, opened 2021-09-28
Modify MNTR to work with the new MNTR package

Arkatufus:


src/core/Akka.Remote.TestKit/CommandLine.cs
```diff
@@ -26,17 +26,19 @@ namespace Akka.Remote.TestKit
...
-    public class CommandLine
+    public static class CommandLine
     {
-        private static readonly StringDictionary Values;
+        private static readonly StringDictionary Values = new StringDictionary();
 
         static CommandLine()
         {
-            Values = new StringDictionary();
+            Initialize(Environment.GetCommandLineArgs());
```
Arkatufus, line 35: Have to do this, else net472 tests would fail. It seems like static scope in net472 is different than in netcore, static somehow didn't get propagated from the node runner to the xunit test runner.

src/core/Akka.Remote.TestKit/CommandLine.cs
```diff
@@ -26,17 +26,19 @@ namespace Akka.Remote.TestKit
...
+        }
 
+        public static void Initialize(string[] args)
```
Arkatufus, line 38: Have to be able to inject arguments from other sources than `Environment.GetCommandLineArgs()` since we're not relying on CLI args anymore.

src/core/Akka.MultiNodeTestRunner.Shared.Tests/MultiNodeTestRunnerDiscovery/DiscoverySpec.cs
```diff
@@ -62,6 +63,7 @@ public void Only_MultiNodeConfig_role_count_used()
 
         private static Dictionary<string, List<NodeTest>> DiscoverSpecs()
         {
+            Environment.SetEnvironmentVariable(MultiNodeFactAttribute.MultiNodeTestEnvironmentName, "1");
```
Aaronontheweb, line 66: What's this for?

src/core/Akka.Remote.TestKit/CommandLine.cs
Aaronontheweb, line 35: Ahhh got it - need to defer the `Environment.GetCommandLineArgs` call until we're ready...

src/core/Akka.MultiNodeTestRunner.Shared.Tests/MultiNodeTestRunnerDiscovery/DiscoverySpec.cs
Aaronontheweb: "Flag to mark that the current test is being run inside the MNTR" according to @Arkatufus in chat - got it

Aaronontheweb (approved):
LGTM

Aaronontheweb:
This completed a full run of the MNTR and everything ran successfully by the looks of it. LGTM

PR #5289, opened 2021-09-28
Added Akka.NET v1.4 upgrade advisories area to website

Aaronontheweb:
In instances where we need to document a regression or an edge-case that has to be factored in during upgrades, we will put them on this area of the website.


Arkatufus (approved):
LGTM

PR #5290, opened 2021-09-28
Change PrimitiveSerializer compatibility switch setting name from `use-neutral-primitives` to `use-legacy-behavior` for less ambiguity

Arkatufus:


Aaronontheweb (approved):
LGTM

src/core/Akka.Remote/Configuration/Remote.conf
```diff
@@ -12,7 +12,7 @@ akka {
   actor {
 	serialization-settings {
       primitive {
-        use-neutral-primitives = off
+        use-legacy-behavior = on
```
Aaronontheweb, line 15: Yep, good idea - need to have this set to `on` by default so users don't hit the issue while upgrading.

PR #5309, opened 2021-10-12
Add spellcheck for docs

SeanKilleen:
We've had this over at the NUnit docs for a bit and I just did it for Marten, so I figured I wouldn't leave Akka .NET out of the fun. :) 

* [x] Add GitHub Action
* [x] Run spellcheck
* [x] cSpell: Add known words (that we want suggested) to `words`
* [x] cSpell: Add words that we want ignored to `ignore`
* [x] Update ignore rules / files


SeanKilleen:
FYI @Aaronontheweb or others, this proposes a GH Actions workflow change and I think that means there's a button on the PR somewhere that you need to hit "Approve" on in order to run it.

Issue #5325, opened 2021-10-18
DirectBufferPool: The buffer is not properly freed in the Release function


jjmSE:
**Version Information**
Version of Akka.NET? 1.4.25
Which Akka.NET Modules? Akka.IO

**Describe the bug**
The DirectBufferPool class throws an exception because the buffer is not properly freed in the Release function. 
(https://github.com/akkadotnet/akka.net/blob/master/src/core/Akka/IO/Buffers/DirectBufferPool.cs)

**Expected behavior**
The buffer should be released.

**Actual behavior**
The buffer is not released and new data is added using the Push function (see screenshots). 

**Screenshots**
![image](https://user-images.githubusercontent.com/92517977/137758385-ec591667-c5ca-4dc1-bc10-b31deb7c13eb.png)

![image](https://user-images.githubusercontent.com/92517977/137758475-269cbb7f-fce8-4d48-9159-95c8a8018ba5.png)

**Environment**
Windows
.NET Framework 4.8

Arkatufus:
Maybe I should create a new class without the memory optimization, lets call it `SimpleBufferPool`, for backward binary compatibility?

Aaronontheweb:
I think this already does that @Arkatufus https://github.com/akkadotnet/akka.net/blob/master/src/core/Akka/IO/Buffers/DisabledBufferPool.cs

Arkatufus:
The `DisabledBufferPool` does not do any instance pooling, each ArraySegment is created on the fly on rent and left for the GC to collect on release.

Arkatufus:
I'm reading up on it, that would need some code changes though. I'll try to make another PR for it.

PR #5326, opened 2021-10-18
Add SimpleBufferPool implementation

Arkatufus:
SimpleBufferPool is an Akka.IO IBufferPool implementation that is closer to the JVM implemention. It does not concern itself with memory management and memory allocation optimization, all it does is ArraySegment caching to reduce the amount of memory allocation and garbage collection in a chatty network environment. 

This is a bandaid patch for buffer pool under/overrun that happened in #5325


Aaronontheweb:
Superseded by https://github.com/akkadotnet/akka.net/pull/5404

Issue #4852, opened 2021-03-15
Akka.Streams: add monitoring stages to measure flow throughput / graph execution time


Aaronontheweb:
Without getting overly prescriptive on the specs for this:

1. Want one flow stage which measures the rate at which stream items are flowing through it and periodically emits that value to an external `Sink` of some kind. That `Sink` can be a `Sink.ActorRef` or whatever else, but the idea being that this stream can observe the quantity of messages / second passing through it and periodically emit that counter value out to the `Sink`. This can be used explicitly for monitoring / observability purposes or it can be used for flow-control, i.e. observing how close a system is performing compared to some internal or external resource limit and pre-emptively throttling.
2. This one is more complicated - need a way of observing how long it takes a specific "tracer round" or a user-defined message to pass from one point of the stream to another. This is designed to measure per-message latency as a gauge or histogram - so in the event that a stream begins rapidly backpressuring, that value can be observed in an external monitoring system or perhaps an alert can be fired.

We are planning on building some Phobos-specific implementations of these, but given the value they can bring outside of observability (flow control, adaptive backoff, etc) we thought they merited inclusion into the Akka.Streams core library itself.

Aaronontheweb:
closed via #5357 - that should help quite a bit.

PR #5357, opened 2021-10-31
Ported WireTap combinator

ismaelhamed:
This should also help with #4852

### DIFF
--- a/src/core/Akka.Streams/Dsl/Internal/InternalFlowOperations.cs
+++ b/src/core/Akka.Streams/Dsl/Internal/InternalFlowOperations.cs
@@ -196,6 +196,37 @@ namespace Akka.Streams.Dsl.Internal
         }
 
         /// <summary>
+        /// This is a simplified version of <seealso cref="WireTap{T}"/> that takes only a simple procedure.
+        /// Elements will be passed into this "side channel" delegate, and any of its results will be ignored.
+        /// <para>
+        /// If the wire-tap operation is slow (it backpressures), elements that would've been sent to it will be dropped instead.
+        /// </para>
+        /// <para>
+        /// This operation is useful for inspecting the passed through element, usually by means of side-effecting
+        /// operations (such as `Log`, or emitting metrics), for each element without having to modify it.
+        /// </para>
+        /// <para>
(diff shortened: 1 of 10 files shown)


src/core/Akka.Streams/Dsl/Internal/InternalFlowOperations.cs
```diff
@@ -195,6 +195,37 @@ internal static class InternalFlowOperations
...
+        /// <para>Cancels when downstream cancels</para>
+        /// </summary>
+        /// <typeparam name="TOut">TBD</typeparam>
+        /// <typeparam name="TMat">TBD</typeparam>
+        /// <param name="flow">TBD</param>
+        /// <param name="action">TBD</param>
+        /// <returns>TBD</returns>
+        public static IFlow<TOut, TMat> WireTap<TOut, TMat>(this IFlow<TOut, TMat> flow, Action<TOut> action) =>
+            //flow.WireTap(Sink.ForEach(action));
+            flow.WireTapMaterialized(Sink.ForEach(action), Keep.Left);
```
ismaelhamed, line 227: This should be `flow.WireTap(Sink.ForEach(action))`, but I can't seem to reconcile TMat that way (I'm missing something). I made it work with `WireTapMaterialized` in the meantime.

Aaronontheweb: I'll clone the source and take a look at it.

Aaronontheweb (commented):
Looks good overall - I'll clone the source and see what's up with that materializer call

src/core/Akka.Streams/Dsl/Internal/InternalFlowOperations.cs
Aaronontheweb: Problem here is how `Sink.ForEach` is hard-coded to return a `Task` materialization value, which is appropriate given how the stage is normally used. Passing in the `Keep.Left` call is the only real way to discard that.

This might be a translation issue due to .NET's "real generics" vs. the JVM's "erasable" generics - I'll see if there's a better way around this but that might be the bottom line in this instance.

Aaronontheweb (approved):
LGTM

PR #5360, opened 2021-11-02
made `GraphStageLogic.LogSource` virtual and change default `StageLogic` `LogSource`

Aaronontheweb:
Also, changed the default `LogSource` to be the output of the `Stage.ToString()` method. Right now all `GraphStageLogic` log sources default to just "Logic (akka://{ActorSystemName})"

### DIFF
--- a/src/core/Akka.Streams/Stage/GraphStage.cs
+++ b/src/core/Akka.Streams/Stage/GraphStage.cs
@@ -839,6 +839,7 @@ namespace Akka.Streams.Stage
         /// <param name="shape">TBD</param>
         protected GraphStageLogic(Shape shape) : this(shape.Inlets.Count(), shape.Outlets.Count())
         {
+            LogSource = Akka.Event.LogSource.Create(shape.ToString());
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Aaronontheweb:
Here's the `Shape.ToString()` method implementation:

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Streams/Shape.cs#L232

Is this too noisy for the logs? cc @ismaelhamed

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -839,6 +839,7 @@ protected GraphStageLogic(int inCount, int outCount)
         /// <param name="shape">TBD</param>
         protected GraphStageLogic(Shape shape) : this(shape.Inlets.Count(), shape.Outlets.Count())
         {
+            LogSource = Akka.Event.LogSource.Create(shape);
```
Aaronontheweb, line 842: Decided to just log the `Shape.GetType` here instead - less verbose this way.

Arkatufus (approved):
LGTM

PR #5361, opened 2021-11-02
rewrite `PipeTo` to use `async` / `await` instead of a callback

Aaronontheweb:
Doing this for two reasons:

1. Allows the `AsyncLocal` context to flow into the `PipeTo` and its resultant `IActorRef.Tell` call, which is helpful for telemetry purposes
2. Eliminates a `ContinueWith` delegate everywhere it's used.

### DIFF
--- a/src/core/Akka/Actor/PipeToSupport.cs
+++ b/src/core/Akka/Actor/PipeToSupport.cs
@@ -27,28 +27,24 @@ namespace Akka.Actor
         /// <param name="sender">TBD</param>
         /// <param name="success">TBD</param>
         /// <param name="failure">TBD</param>
-        /// <returns>TBD</returns>
-        public static Task PipeTo<T>(this Task<T> taskToPipe, ICanTell recipient, IActorRef sender = null, Func<T, object> success = null, Func<Exception, object> failure = null)
+        /// <returns>A detached task</returns>
+        public static async Task PipeTo<T>(this Task<T> taskToPipe, ICanTell recipient, IActorRef sender = null, Func<T, object> success = null, Func<Exception, object> failure = null)
         {
             sender = sender ?? ActorRefs.NoSender;
-            return taskToPipe.ContinueWith(tresult =>
+            
+            try
             {
-                if (tresult.IsFaulted)
-                    recipient.Tell(failure != null
-                        ? failure(tresult.Exception)
-                        : new Status.Failure(tresult.Exception), sender);
-                else if (tresult.IsCanceled)
-                {
-                    var ex = tresult.Exception ?? new AggregateException(new TaskCanceledException());
-                    recipient.Tell(failure != null
-                        ? failure(ex)
-                        : new Status.Failure(ex), sender);
-                }
-                else if (tresult.IsCompleted)
-                    recipient.Tell(success != null
-                        ? success(tresult.Result)
-                        : tresult.Result, sender);
-            }, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
+                var result = await taskToPipe.ConfigureAwait(false);
+                recipient.Tell(success != null
+                    ? success(result)
+                    : result, sender);
+            }
+            catch (Exception ex)
+            {
+                recipient.Tell(failure != null
+                    ? failure(ex)
+                    : new Status.Failure(ex), sender);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/core/Akka/Actor/PipeToSupport.cs
```diff
@@ -27,28 +27,24 @@ public static class PipeToSupport
         /// <param name="sender">TBD</param>
         /// <param name="success">TBD</param>
         /// <param name="failure">TBD</param>
-        /// <returns>TBD</returns>
-        public static Task PipeTo<T>(this Task<T> taskToPipe, ICanTell recipient, IActorRef sender = null, Func<T, object> success = null, Func<Exception, object> failure = null)
+        /// <returns>A detached task</returns>
+        public static async Task PipeTo<T>(this Task<T> taskToPipe, ICanTell recipient, IActorRef sender = null, Func<T, object> success = null, Func<Exception, object> failure = null)
```
Aaronontheweb, line 31: So long as `PipeTo` is not itself `await`ed then the behavior will be identical to what it is in Akka.NET v1.4.27.

Going to let the unit tests demonstrate if I'm incorrect on this.

Aaronontheweb:
Looks like this definitely had an impact on the type of `Exception`s propagated back via `PipeTo`

ismaelhamed:
> Looks like this definitely had an impact on the type of `Exception`s propagated back via `PipeTo`

AFAIK, `await` will not return an `AggregateException`, like `ContinueWith`, which is what many tests are expecting. IMO, any code invoking an async method should be expected to deal with `AggregateException`s too.

Aaronontheweb:
> AFAIK, `await` will not return an `AggregateException`, like `ContinueWith`, which is what many tests are expecting. IMO, any code invoking an async method should be expected to deal with `AggregateException`s too.

The `await` is going to flatten any `AggregateException`s into whatever the underlying `Exception` was.

Well since this is a bit of a behavioral change, I can do two things:

* Delay this change until Akka.NET v1.5;
* Wrap the `Exception` in an `AggregateException` and re-throw it for v1.4 backwards compat - and subsequently remove that in v1.5.

Or we can leave it as-is. What do you think?

ismaelhamed:
I see no harm in this PR, except that you'll need to fix them tests ;)

Aaronontheweb:
I'll also verify that this fixes some of our telemetry issues I've been working on as well

Aaronontheweb:
Fixed the specs - going to work on the telemetry issues.

Aaronontheweb:
Have another round of tests I need to fix first.

Aaronontheweb:
I think that should be all of the specs now. Going to test this out locally and make sure this actually solved some of our telemetry issues.

Aaronontheweb:
Looks like this didn't solve my correlation problem on Phobos, but that's because our OpenTracing `ActorScopeManager` doesn't use `AsyncLocal` when correlating spans that originate from an actor - this will be resolved when we move forward with OpenTelemetry, which does use `AsyncLocal`. I'll keep working on this, but I think this is a good change in that it solves a very common "close over `Sender`" UX / bug source for all Akka.NET users.

Aaronontheweb:
Had some of our tests timeout after 60 minutes, which strikes me as odd. Re-running it.

src/core/Akka/Actor/PipeToSupport.cs
Arkatufus: so if we await a PipeTo, it will act as if we're doing an await to an Ask?

Aaronontheweb: So long as the `PipeTo` itself is not `await`ed, it will continue to run as a detached `Task` - `PipeTo` could have also been `await`-ed in current versions of Akka.NET too.

Aaronontheweb: But yes, if you `await` the `PipeTo` the `PipeTo` will also away on the antecedent `Task`, which is structurally identical to what it does now from a flow control perspective but it's semantically different because now `AsyncLocal` context propagates with it using local methods and context. 

Basically this means we don't have to close over the `Sender` anymore outside the `PipeTo` call as that's now evaluated in real time inside the `PipeTo` method prior to its `await`.

Aaronontheweb (commented):
Responding to comments

src/core/Akka/Actor/PipeToSupport.cs
Aaronontheweb: > Will the exception result be different? As I remember, await will automatically unroll the AggregatedException thrown, so this can be a breaking change

Yeah this is what we discussed with @ismaelhamed - it definitely eliminates the `AggregateException` but still produces the correct `Status.Failure` message with the underlying `Exception` tucked inside the `Status.Failure.Cause` field so I think that should be fine. That property was never hard-coded to be an `AggregateException` in the first place, but it almost always was due to upstream `Task` handling.

So if that's too big of a change to fit into Akka.NET v1.4.28, I'd be fine moving it to v1.5

Arkatufus (approved):
LGTM

Aaronontheweb:
@Arkatufus @ismaelhamed I'm going to hold off on including this in v1.4.28 and include it in 1.4.29, so it has some more time to get used in nightly builds et al prior to having end-users work with it.

PR #5373, opened 2021-11-09
add .NET 6 SDK support

Aaronontheweb:
Adding .NET 6 SDK support. Replaces .NET 5 support. Might also drop .NET Core 3.1 as part of this.


Aaronontheweb:
Need to upgrade Incrementalist first: https://github.com/petabridge/Incrementalist

Aaronontheweb:
@Arkatufus do you know if we still need a separate MNTR task template in Azure DevOps now that we can just `dotnet test` the MNTR?

Arkatufus:
If we're dropping the old MNTR then no, we don't need them, since it will be released from a different repo

PR #5374, opened 2021-11-10
Return OverflowStrategy applied even when QueueOfferResult.Enqueued

vladud:


Aaronontheweb:
@vladud did you mean to close this?

Issue #5136, opened 2021-07-08
Port akka.io.dns and akka.discovery.dns


Arkatufus:
Porting akka.discovery.dns over from the JVM implementation would also require a port of the akka.io.dns namespace.
We could not use the built-in DNS client in dotnet because it is seriouly lacking in features, as it only handles the ANAME DNS query, so we would also need to port the JVM DNS client implementation or roll our own DNS client implementation.

- Port DNS client 
  - Able to handle DNS protocol, especially the A, AAAA, and SRV records and be able to extract TTL information
  - Have a cache for all resolved DNS record lookup and be able to do periodic check for record expiry
  - Test suite
- akka.discovery.dns.DnsServiceDiscovery
  - Test suite

ismaelhamed:
@Arkatufus I'd say [DnsClient](https://www.nuget.org/packages/DnsClient/) checks all the boxes.

Aaronontheweb:
I'd be fine taking a dependency on that inside Akka.Discovery. Probably wouldn't want that as part of Akka.IO. Do you think that'd be a reasonable compromise?

ismaelhamed:
Sounds good. Actually, what would be a use case for `Akka.IO.Dns`? I don't think I've seen it being used anywhere.

Aaronontheweb:
> Sounds good. Actually, what would be a use case for `Akka.IO.Dns`? I don't think I've seen it being used anywhere.

Other than Akka.Discovery - I have no idea. Doesn't make a lot of sense to have it IMHO.

anpin:
DNS support was merged into Akka.Management repo, this is issue can be closed

Aaronontheweb:
> DNS support was merged into Akka.Management repo, this is issue can be closed

Thank you for that, BTW. send me an email - aaron AT petabridge DOT com and I'll send you some Akka.NET swag.

PR #5380, opened 2021-11-10
make Akka.Discovery RTM, rather than beta status

Aaronontheweb:
Given how things are going with [Akka.Management](https://github.com/akkadotnet/Akka.Management), I think it would be worthwhile to take Akka.Discovery out of beta so we can also begin releasing Akka.Management RTM packages in the near future.


Aaronontheweb:
cc @Arkatufus @ismaelhamed

Aaronontheweb:
@ismaelhamed I just weighed in on that here: https://github.com/akkadotnet/akka.net/issues/5136#issuecomment-966345551

PR #5384, opened 2021-11-15
added thread count tracking to RemotePingPong

Aaronontheweb:
Added this to help diagnose throughput drops on .NET 6, which we suspect are related to https://github.com/dotnet/runtime/pull/53471

Now displays a data table that looks like this - with the Start Threads and End Threads columns attached.

image

### DIFF
--- a/src/benchmark/RemotePingPong/Program.cs
+++ b/src/benchmark/RemotePingPong/Program.cs
@@ -101,7 +101,7 @@ namespace RemotePingPong
             Console.WriteLine();
 
             //Print tables
-            Console.WriteLine("Num clients, Total [msg], Msgs/sec, Total [ms]");
+            Console.WriteLine("Num clients, Total [msg], Msgs/sec, Total [ms], Start Threads, End Threads");
 
             _firstRun = false;
         }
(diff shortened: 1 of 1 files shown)


src/benchmark/RemotePingPong/Program.cs
```diff
@@ -188,6 +188,8 @@ private static async Task<(bool, long, int)> Benchmark(int numberOfClients, long
                 PrintSysInfo();
             }
 
+            var startThreads = Process.GetCurrentProcess().Threads.Count;
```
Aaronontheweb, line 191: Recorded after Akka.Remote association is established, but before pressure is applied.

src/benchmark/RemotePingPong/Program.cs
```diff
@@ -197,6 +199,8 @@ private static async Task<(bool, long, int)> Benchmark(int numberOfClients, long
             var waiting = Task.WhenAll(tasks);
             await Task.WhenAll(waiting);
             sw.Stop();
+            
+            var endThreads = Process.GetCurrentProcess().Threads.Count;
```
Aaronontheweb, line 203: Collected at the end of the benchmark but before the `ActorSystem`s are terminated.

PR #5387, opened 2021-11-16
Fix Akka.Streams InitialAttributes is injected to the end of the Module.Attributes

Arkatufus:


Aaronontheweb:
Definitely a bug.

PR #5390, opened 2021-11-18
DRAFT: Add SemaphoreStateV2, Eliminate Delegate allocations in ForkJoinExecutor

to11mtm:


src/core/Akka/Dispatch/Dispatchers.cs
```diff
@@ -262,7 +262,7 @@ public override void Execute(IRunnable run)
         {
             if (Volatile.Read(ref _shuttingDown) == 1)
                 throw new RejectedExecutionException("ForkJoinExecutor is shutting down");
-            _dedicatedThreadPool.QueueUserWorkItem(run.Run);
+            _dedicatedThreadPool.QueueUserWorkItem(r=>r.Run(), run);
```
to11mtm, line 265: Frankly, we would be better off here if we just took `IRunnable` in here and called .Run() directly, which would be a good next step. That would eliminate a bit more indirection as well as further improve allocations.

Aaronontheweb:
Ran `RemotePingPong` on my Intel laptop and the stddev gets pretty bad for both the baseline and this PR as a result of the cooling system kicking in. Going to try some of the Benchmark.NET ones instead and see if I can get a more normalized figure.

PR #5399, opened 2021-11-24
Fix: Akka.Cluster Components

Zetanova:
Multiple components in the cluster startup like the IDowningProvider retrieved the cluster instance over `Cluster.Get(Context.System)` in the constructor. 
this is racy because the completion of the Cluster constructor extension is not guaranteed.


src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -816,6 +816,8 @@ public ClusterEvent.IClusterDomainEvent Event
     /// </summary>
     internal sealed class ClusterDaemon : ReceiveActor, IRequiresMessageQueue<IUnboundedMessageQueueSemantics>
     {
+        private Cluster _cluster;
```
Aaronontheweb, line 819: Where does this value get assigned?

Aaronontheweb (changes requested):
as far as I can tell, the `ClusterDaemon` never assigns its internal `_cluster` property.

src/core/Akka.Cluster/ClusterDaemon.cs
Zetanova: in `ClusterDaemon .CreateChildren` method
The message `InternalClusterAction.GetClusterCoreRef` acts as the init message from `Cluster` constructor

Zetanova (commented):
resolved

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -888,11 +890,12 @@ private void AddCoordinatedLeave()
             _coordShutdown.AddTask(CoordinatedShutdown.PhaseClusterShutdown, "wait-shutdown", () => _clusterPromise.Task);
         }
 
-        private void CreateChildren()
+        private void CreateChildren(Cluster cluster)
         {
+            _cluster = cluster;
```
Aaronontheweb, line 895: Ok, got it

Aaronontheweb (approved):
LGTM

PR #5403, opened 2021-11-30
Add new ChannelTaskScheduler Extension 

Zetanova:
As requested inlining https://github.com/Zetanova/Akka.Experimental.ChannelTaskScheduler into akka and using it as the default 
`channel-executor`

usage like before, the `channel-executor.priority` config value is already set in the default config.

to activate use:
``` 
akka.actor.default-dispatcher = {
    executor = "channel-executor"
    #channel-executor.priority = "normal"
}

akka.actor.internal-dispatcher = {
    executor = "channel-executor"
    #channel-executor.priority = "high"
}

akka.remote.default-remote-dispatcher {
    executor = "channel-executor"
    #channel-executor.priority = "high"
}

akka.remote.backoff-remote-dispatcher {
    executor = "channel-executor"
    #channel-executor.priority = "low"
}
```

All akka components can thread-safe access the priority TaskSchedulers over:
```
ChannelTaskScheduler.Get(system).High
ChannelTaskScheduler.Get(system).Normal
ChannelTaskScheduler.Get(system).Low
ChannelTaskScheduler.Get(system).Idle
```

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -3260,8 +3260,17 @@ namespace Akka.IO.Buffers
         public BufferPoolAllocationException(string message) { }
         protected BufferPoolAllocationException(System.Runtime.Serialization.SerializationInfo info, System.Runtime.Serialization.StreamingContext context) { }
     }
+    public class BufferPoolInfo
+    {
+        public BufferPoolInfo(System.Type type, long totalSize, long free, long used) { }
+        public long Free { get; }
+        public long TotalSize { get; }
+        public System.Type Type { get; }
+        public long Used { get; }
+    }
     public interface IBufferPool
     {
+        Akka.IO.Buffers.BufferPoolInfo Diagnostics();
         void Release(System.ArraySegment<byte> buf);
         void Release(System.Collections.Generic.IEnumerable<System.ArraySegment<byte>> buf);
         System.ArraySegment<byte> Rent();
@@ -3382,6 +3391,7 @@ namespace Akka.IO
     }
     public interface ISocketEventArgsPool
     {
+        Akka.IO.Buffers.BufferPoolInfo BufferPoolInfo { get; }
         System.Net.Sockets.SocketAsyncEventArgs Acquire(Akka.Actor.IActorRef actor);
         void Release(System.Net.Sockets.SocketAsyncEventArgs e);
     }
@@ -3904,14 +3914,12 @@ namespace Akka.IO
     {
         public UdpConnectedExt(Akka.Actor.ExtendedActorSystem system) { }
         public UdpConnectedExt(Akka.Actor.ExtendedActorSystem system, Akka.IO.UdpSettings settings) { }
-        public Akka.IO.Buffers.IBufferPool BufferPool { get; }
         public override Akka.Actor.IActorRef Manager { get; }
     }
     public class UdpExt : Akka.IO.IOExtension
     {
         public UdpExt(Akka.Actor.ExtendedActorSystem system) { }
         public UdpExt(Akka.Actor.ExtendedActorSystem system, Akka.IO.UdpSettings settings) { }
-        public Akka.IO.Buffers.IBufferPool BufferPool { get; }
         public override Akka.Actor.IActorRef Manager { get; }
     }
     public class static UdpExtensions
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
@@ -2509,6 +2509,22 @@ namespace Akka.Dispatch
         public System.TimeSpan PushTimeout { get; }
         public override Akka.Dispatch.MessageQueues.IMessageQueue Create(Akka.Actor.IActorRef owner, Akka.Actor.ActorSystem system) { }
     }
+    public sealed class ChannelTaskScheduler : Akka.Actor.IExtension, System.IDisposable
+    {
+        public ChannelTaskScheduler(Akka.Actor.ExtendedActorSystem system) { }
+        public System.Threading.Tasks.TaskScheduler High { get; }
+        public System.Threading.Tasks.TaskScheduler Idle { get; }
+        public System.Threading.Tasks.TaskScheduler Low { get; }
+        public System.Threading.Tasks.TaskScheduler Normal { get; }
+        public void Dispose() { }
+        public static Akka.Dispatch.ChannelTaskScheduler Get(Akka.Actor.ActorSystem system) { }
+        public System.Threading.Tasks.TaskScheduler GetScheduler(Akka.Dispatch.TaskSchedulerPriority priority) { }
+    }
+    public sealed class ChannelTaskSchedulerProvider : Akka.Actor.ExtensionIdProvider<Akka.Dispatch.ChannelTaskScheduler>
+    {
+        public ChannelTaskSchedulerProvider() { }
+        public override Akka.Dispatch.ChannelTaskScheduler CreateExtension(Akka.Actor.ExtendedActorSystem system) { }
+    }
     public sealed class CurrentSynchronizationContextDispatcher : Akka.Dispatch.Dispatcher
     {
         public CurrentSynchronizationContextDispatcher(Akka.Dispatch.MessageDispatcherConfigurator configurator, string id, int throughput, System.Nullable<long> throughputDeadlineTime, Akka.Dispatch.ExecutorServiceFactory executorServiceFactory, System.TimeSpan shutdownTimeout) { }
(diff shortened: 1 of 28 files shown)


src/core/Akka/Akka.csproj
```diff
@@ -19,6 +19,7 @@
     <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
     <PackageReference Include="System.Reflection.Emit" Version="4.7.0" />
       <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
+      <PackageReference Include="System.Threading.Channels" Version="6.0.0" />
```
Aaronontheweb, line 22: We probably need to align all of our `System.*` packages on a common version range - see https://github.com/akkadotnet/akka.net/pull/5365#issuecomment-984763805

Separate issue from this PR, but I'm seeing a trend here...

src/core/Akka/Configuration/Pigeon.conf
```diff
@@ -269,6 +269,15 @@ akka {
       }
     }
 
+    channel-scheduler {
```
Aaronontheweb, line 272: How do `work-max` et al relate to other items such as `dispatcher.throughput` for individual actor mailboxes?

src/core/Akka/Configuration/Pigeon.conf
```diff
@@ -377,6 +386,7 @@ akka {
         parallelism-factor = 1.0
         parallelism-max = 64
       }
+      channel-executor.priority = "high"
```
Aaronontheweb, line 389: Is this meant to be `channel-scheduler` or `channel-executor`? The section above where all of the defaults are defined is `channel-scheduler`

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -2677,6 +2693,18 @@ namespace Akka.Dispatch
         public RejectedExecutionException(string message = null, System.Exception inner = null) { }
         protected RejectedExecutionException(System.Runtime.Serialization.SerializationInfo info, System.Runtime.Serialization.StreamingContext context) { }
     }
+    public enum TaskSchedulerPriority
```
Aaronontheweb, line 2696: Is there a built-in enum we can leverage instead of adding our own? Or is that all "thread priority" stuff and not "task priority" ?

src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
```diff
@@ -0,0 +1,389 @@
...
+{
+    public sealed class ChannelTaskSchedulerProvider : ExtensionIdProvider<ChannelTaskScheduler>
+    {
+        public override ChannelTaskScheduler CreateExtension(ExtendedActorSystem system)
+        {
+            return new ChannelTaskScheduler(system);
+        }
+    }
+
+    public sealed class ChannelTaskScheduler : IExtension, IDisposable
```
Aaronontheweb, line 20: Need XML-DOC comments here

Aaronontheweb: Most importantly what we need, for future bug fixing and contributors, a summation of how your algorithm works and what role each component plays.

src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
```diff
@@ -0,0 +1,389 @@
...
+    {
+        [ThreadStatic]
+        private static TaskSchedulerPriority _threadPriority = TaskSchedulerPriority.None;
+
+        private readonly Task _controlTask;
```
Aaronontheweb, line 25: For instance, what is the `_controlTask` doing? How does it relate to the `Timer` and the channels themselves?

src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
```diff
@@ -0,0 +1,389 @@
...
+                    readTasks[2] = lowReader.WaitToReadAsync().AsTask();
+                if (readTasks[3].IsCompleted)
+                    readTasks[3] = idleReader.WaitToReadAsync().AsTask();
+
+                readTask = await Task.WhenAny(readTasks).ConfigureAwait(false);
+            }
+            while (readTask.Result && !_cts.IsCancellationRequested);
+        }
+
+        private void ScheduleCoWorkers(object state)
```
Aaronontheweb, line 154: Please provide a comment explaining, generally, how this method is supposed to work and what its configurable limitations are.

Not thinking about this for users so much as other contributors / maintainers.

Aaronontheweb (changes requested):
The code looks fine - I think we also need to add:

* Some configuration specs, validating that default values can be successfully changed via HOCON
* Some documentation additions to https://getakka.net/articles/actors/dispatchers.html explaining how to configure this and when you might want to use it
* Address some of my inline comments, which ask for more clarification on the internals so other contributors / maintainers can reason about the algorithm.

src/core/Akka/Configuration/Pigeon.conf
Zetanova: It is not related, says only that a worker is processing max count of work items (normal priority) before it loops
The loop forces the worker to check high-priority items

src/core/Akka/Configuration/Pigeon.conf
Zetanova: This are to separate components.

The `channel-scheduler` is a singleton extension (can be even standalone)
It hosts all priority task schedulers and runs the workers

The `channel-executor` is a normal akka executor component like the `ForkJoinExecutor`
It has currently only one config field: the priority. 
With the priority field it can pick the TaskScheduler from the ChannelScheduler

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Zetanova: Yes, the values are the same like on windows for thread and process priority
I did not find any good inside .net, in worst case they can be casted, when ever needed.

src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
Zetanova: Can I then include todo and ideas comments too?

Aaronontheweb: Of course!

src/core/Akka/Akka.csproj
Zetanova: done

src/core/Akka/Configuration/Pigeon.conf
Zetanova: In my field test there was only the main worker running. 
This means that only if normal-priority work with `max-work`  takes longer 
as `work-internal / work-step` (250ms) then a coworker gets started and executes the high priority work.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
Zetanova: I added a comment

src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
Zetanova: done

Zetanova (commented):
I added the comments and rollback dependency to 5.0.0

Zetanova:
About the https://getakka.net/articles/actors/dispatchers.html where to edit it, can u do it?
The only thing that changed between the last `channel-executor` and this,
is the `channel-executor.priority` config instead of the `max-concurrency` calculated from `fork-join-executor` settings

Aaronontheweb:
I’ll reply with a link to the .MD file once I’m back at my desk. On mobile at the moment.

Aaronontheweb:
> About the https://getakka.net/articles/actors/dispatchers.html where to edit it, can u do it?

Yep, I can do it.

Aaronontheweb (approved):
LGTM

src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
```diff
@@ -0,0 +1,389 @@
...
+            _normalScheduler.Dispose();
+            _highScheduler.Dispose();
+
+            _cts.Cancel();
+            _timer.Dispose();
+        }
+
+        sealed class PriorityTaskScheduler : TaskScheduler, IDisposable
+        {
+            readonly Channel<Task> _channel;
```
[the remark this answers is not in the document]
Zetanova: Good point, I will take a look the the Channel source code too.

In the end the queue should lock-free queued and dequeued, 
where adding side should have the heavier processing (splitting in priorities)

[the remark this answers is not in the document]
Zetanova: I looked up and yes there is a brief lock. 

The lock will not have much perf. issue on the executing thread itself,
but of course on locking threads. Don't know why it is not implemented in a free-lock style. 
Maybe the lock collation just don't happen that often to take it into effect?
I trust ["stephentoub" ](https://github.com/stephentoub) in his decision to use a lock here.

The `lock` statement is on itself is not "bad" but misused it can be very fast.

The main benefit of the Channel class is following line, it is reducing the idle cpu to near zero:
https://github.com/dotnet/runtime/blob/386f87139c7fe48f6b733e3cae6d09128c827192/src/libraries/System.Threading.Channels/src/System/Threading/Channels/UnboundedChannel.cs#L295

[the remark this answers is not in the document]
Zetanova: I rechecked now the `UnboundedChannel.TryWrite` code
1) it would be absolutely hard not to use a sync lock
2) Collisions are very unlikely, even under heavy queue load

PR #5404, opened 2021-12-01
Fix UDP memory leak

Arkatufus:
Fix UDP memory leak problem because the rented byte buffer memory never got released when SocketAsyncEventArgs got lost.

Closes https://github.com/akkadotnet/akka.net/issues/5325

- Moved ByteBuffer from Udp and UdpConnected into PreallocatedSocketEventAgrsPool. The socket event pool is resposible for renting and releasing buffers now, so we only need to care about renting and releasing socket events.
- SocketAsyncEventArgs never leaves `Udp`, `UdpListener`, `UdpConnect`, and `UdpConnection` class. Required data are copied into the response messages instead of embedding `SocketAsyncEventArgs` inside the `Received` message. `SocketAsyncEventArgs` will either be released immediately if `Socket.ReceiveAsync` returned false or released when the `OnComplete` event fires.
- Removed `SocketAsyncEventArgs` pooling functionality, caching `SocketAsyncEventArgs` is causing a lot more grief than its worth, we would have to catch all the edge cases where it can fail. `SocketAsyncEventArgs` is not a simple struct that can be cleaned and reused, it actually have internal states that can cause a lot of problems when reused.

### DIFF
--- a/src/core/Akka.Tests/IO/UdpConnectedIntegrationSpec.cs
+++ b/src/core/Akka.Tests/IO/UdpConnectedIntegrationSpec.cs
@@ -8,8 +8,10 @@
 using System;
 using System.Linq;
 using System.Net;
+using System.Threading;
 using Akka.Actor;
 using Akka.IO;
+using Akka.IO.Buffers;
 using Akka.TestKit;
 using FluentAssertions;
 using FluentAssertions.Extensions;
@@ -26,12 +28,20 @@ namespace Akka.Tests.IO
             : base(@"
                     akka.actor.serialize-creators = on
                     akka.actor.serialize-messages = on
+
+                    akka.io.udp-connected.buffer-pool = ""akka.io.udp-connected.direct-buffer-pool""
                     akka.io.udp-connected.nr-of-selectors = 1
-                    akka.io.udp-connected.direct-buffer-pool-limit = 100
-                    akka.io.udp-connected.direct-buffer-size = 1024
+                    # This comes out to be about 1.6 Mib maximum total buffer size
+                    akka.io.udp-connected.direct-buffer-pool.buffer-size = 512
+                    akka.io.udp-connected.direct-buffer-pool.buffers-per-segment = 32
+                    akka.io.udp-connected.direct-buffer-pool.buffer-pool-limit = 100
+
+                    akka.io.udp.buffer-pool = ""akka.io.udp.direct-buffer-pool""
                     akka.io.udp.nr-of-selectors = 1
-                    akka.io.udp.direct-buffer-pool-limit = 100
-                    akka.io.udp.direct-buffer-size = 1024
+                    # This comes out to be about 1.6 Mib maximum total buffer size
+                    akka.io.udp.direct-buffer-pool.buffer-size = 512
+                    akka.io.udp.direct-buffer-pool.buffers-per-segment = 32
+                    akka.io.udp.direct-buffer-pool.buffer-pool-limit = 100
                     akka.io.udp.trace-logging = true
                     akka.loglevel = DEBUG", output)
         {
@@ -140,6 +150,72 @@ namespace Akka.Tests.IO
             raw = client.ReceiveN(3);
             msgs = raw.Cast<UdpConnected.Received>();
             msgs.Sum(x => x.Data.Count).Should().Be(data.Count * 3);
+        }
+        
+        [Fact]
+        public void The_UDP_connection_oriented_implementation_must_not_leak_memory()
+        {
+            const int batchSize = 5000;
+            
+            var serverAddress = _addresses[0];
+            var clientAddress = _addresses[1];
+            var udpConnection = UdpConnected.Instance.Apply(Sys);
+            
+            var poolInfo = udpConnection.SocketEventArgsPool.BufferPoolInfo;
+            poolInfo.Type.Should().Be(typeof(DirectBufferPool));
+            poolInfo.Free.Should().Be(poolInfo.TotalSize);
+            poolInfo.Used.Should().Be(0);
+            
+            var server = CreateTestProbe();
+            udpConnection.Manager.Tell(new UdpConnected.Connect(server, clientAddress, serverAddress), server);
+            server.ExpectMsg<UdpConnected.Connected>();
+            var serverEp = server.LastSender;
+
+            var client = CreateTestProbe();
+            udpConnection.Manager.Tell(new UdpConnected.Connect(client, serverAddress, clientAddress), client);
+            client.ExpectMsg<UdpConnected.Connected>();
+            var clientEp = client.LastSender;
+            
+            var data = ByteString.FromString("Fly little packet!");
+
+            // send a lot of packets through, the byte buffer pool should not leak anything
+            for (var j = 0; j < batchSize; ++j)
+            {
+                serverEp.Tell(UdpConnected.Send.Create(data));
+            }
+
+            var msgs = client.ReceiveN(batchSize, TimeSpan.FromSeconds(10));
+            var cast = msgs.Cast<UdpConnected.Received>();
+            cast.Sum(m => m.Data.Count).Should().Be(data.Count * batchSize);
+
+            // stop all connections so all receives are stopped and all pending SocketAsyncEventArgs are collected
+            serverEp.Tell(UdpConnected.Disconnect.Instance, server);
+            server.ExpectMsg<UdpConnected.Disconnected>();
+            clientEp.Tell(UdpConnected.Disconnect.Instance, client);
+            client.ExpectMsg<UdpConnected.Disconnected>();
+            
+            // wait for all SocketAsyncEventArgs to be released
+            Thread.Sleep(1000);
+            
+            poolInfo = udpConnection.SocketEventArgsPool.BufferPoolInfo;
+            poolInfo.Type.Should().Be(typeof(DirectBufferPool));
+            poolInfo.Free.Should().Be(poolInfo.TotalSize);
+            poolInfo.Used.Should().Be(0);
+        }
+        
+        internal class UdpClientHandler: ReceiveActor
+        {
+            public int ReceivedMessages;
+            public int TotalData;
+            public UdpClientHandler()
+            {
+                Receive<UdpConnected.Received>(rcv =>
+                {
+                    ReceivedMessages++;
+                    TotalData += rcv.Data.Count;
+                });
+                ReceiveAny(obj => throw new Exception("Invalid message"));
+            }
         }
     }
 }
--- a/src/core/Akka.Tests/IO/UdpConnectedIntegrationSpec.cs
+++ b/src/core/Akka.Tests/IO/UdpConnectedIntegrationSpec.cs
@@ -155,7 +155,8 @@ namespace Akka.Tests.IO
         [Fact]
         public void The_UDP_connection_oriented_implementation_must_not_leak_memory()
         {
-            const int batchSize = 5000;
+            const int batchCount = 2000;
+            const int batchSize = 100;
             
             var serverAddress = _addresses[0];
             var clientAddress = _addresses[1];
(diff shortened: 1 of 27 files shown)


src/core/Akka/IO/Buffers/DirectBufferPool.cs
```diff
@@ -33,6 +33,22 @@ protected BufferPoolAllocationException(SerializationInfo info, StreamingContext
         }
     }
 
+    public class BufferPoolInfo
```
Arkatufus, line 36: Diagnostic class to get the state of the buffer pool without exposing the byte buffer class

src/core/Akka/IO/Buffers/DirectBufferPool.cs
```diff
@@ -195,8 +224,10 @@ public IEnumerable<ByteBuffer> Rent(int minimumSize)
         public void Release(ByteBuffer buf)
         {
             // only release buffers that have actually been taken from one of the segments
-            if (buf.Count == _bufferSize && _segments.Contains(buf.Array))
-                _buffers.Push(buf);
+            if (buf.Count != _bufferSize || !_segments.Contains(buf.Array))
+                throw new Exception("Wrong ArraySegment<byte> was released to DirectBufferPool");
```
Arkatufus, line 228: Sanity check to make sure that any `ArraySegment<byte>` returned to the buffer actually belongs to the buffer, should only happen during development.

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -21,48 +22,81 @@ public interface ISocketEventArgsPool
     {
         SocketAsyncEventArgs Acquire(IActorRef actor);
         void Release(SocketAsyncEventArgs e);
+        
+        BufferPoolInfo BufferPoolInfo { get; }
     }
 
     internal class PreallocatedSocketEventAgrsPool : ISocketEventArgsPool
     {
+        private readonly IBufferPool _bufferPool;
```
Arkatufus, line 31: Buffer pool belongs to the socket event pool now, no one should touch this. ever.

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -21,48 +22,81 @@ public interface ISocketEventArgsPool
...
         private readonly EventHandler<SocketAsyncEventArgs> _onComplete;
-        private readonly ConcurrentStack<SocketAsyncEventArgs> _pool = new ConcurrentStack<SocketAsyncEventArgs>();
+        private readonly ConcurrentQueue<SocketAsyncEventArgs> _pool = new ConcurrentQueue<SocketAsyncEventArgs>();
```
Arkatufus, line 33: Changed to make sure that the content of the pool are rotated evenly, instead of only using the top ones. Minimizes the problem where SocketAsyncEventArgs could not be reused because it is locked by the previous socket.
To be honest, this whole class is a bad idea because SocketAsyncEventArgs are only supposed to be recycled if they're used by a single socket.

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -21,48 +22,81 @@ public interface ISocketEventArgsPool
...
+                e.Completed += _onComplete;
+                _pool.Enqueue(e);
             }
         }
 
         public SocketAsyncEventArgs Acquire(IActorRef actor)
         {
-            if (!_pool.TryPop(out var e))
-                e = CreateSocketAsyncEventArgs();
+            var buffer = _bufferPool.Rent();
```
Arkatufus, line 50: Byte buffer pool rent can only happen here and no where else.

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -21,48 +22,81 @@ public interface ISocketEventArgsPool
...
 
-            e.UserToken = actor;
+                    e.SetBuffer(buffer.Array, buffer.Offset, buffer.Count);
+                    e.UserToken = actor;
+                    e.Completed += _onComplete;
+                    acquired = true;
+                }
+                catch (InvalidOperationException)
+                {
+                    // it can be that for some reason socket is in use and haven't closed yet. Dispose anyway to avoid leaks.
```
Arkatufus, line 67: This will always be the case if this pool are being used by multiple sockets...

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -21,48 +22,81 @@ public interface ISocketEventArgsPool
...
+            SocketAsyncEventArgs e = null;
+            while (!acquired)
+            {
+                try
+                {
+                    if (!_pool.TryDequeue(out e))
+                        e = new SocketAsyncEventArgs();
 
-            e.UserToken = actor;
+                    e.SetBuffer(buffer.Array, buffer.Offset, buffer.Count);
```
Arkatufus, line 60: This is fine because we only rent once and if we fail to acquire an instance, we immediately dispose it.

src/core/Akka/IO/Udp.cs
```diff
@@ -34,45 +34,51 @@ public class Udp : ExtensionIdProvider<UdpExt>
...
         #region internal connection messages
 
-        internal abstract class SocketCompleted : INoSerializationVerificationNeeded { }
+        internal abstract class SocketCompleted : INoSerializationVerificationNeeded
+        {
+            public ByteString Data { get; }
+
+            protected SocketCompleted(SocketAsyncEventArgs eventArgs)
+            {
+                Data = ByteString.CopyFrom(eventArgs.Buffer, eventArgs.Offset, eventArgs.BytesTransferred);
```
Arkatufus, line 43: Instead of embedding the SocketAsyncEventArgs, we extract the data and store that instead.

src/core/Akka/IO/Udp.cs
```diff
@@ -556,23 +559,17 @@ private void OnComplete(object sender, SocketAsyncEventArgs e)
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private Udp.SocketCompleted ResolveMessage(SocketAsyncEventArgs e)
+        private static Udp.SocketCompleted ResolveMessage(SocketAsyncEventArgs e)
         {
             switch (e.LastOperation)
             {
                 case SocketAsyncOperation.Receive:
                 case SocketAsyncOperation.ReceiveFrom:
                     return new Udp.SocketReceived(e);
-                case SocketAsyncOperation.Send:
```
Arkatufus, line 569: These are never used. Removed to remove red herrings and possible problem in the future.

src/core/Akka/IO/UdpConnected.cs
```diff
@@ -433,23 +433,17 @@ private void OnComplete(object sender, SocketAsyncEventArgs e)
...
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private UdpConnected.SocketCompleted ResolveMessage(SocketAsyncEventArgs e)
+        private static UdpConnected.SocketCompleted ResolveMessage(SocketAsyncEventArgs e)
         {
             switch (e.LastOperation)
             {
                 case SocketAsyncOperation.Receive:
                 case SocketAsyncOperation.ReceiveFrom:
                     return new UdpConnected.SocketReceived(e);
-                case SocketAsyncOperation.Send:
```
Arkatufus, line 446: Same reason as above, this isn't being used.

src/core/Akka/IO/UdpConnectedManager.cs
```diff
@@ -42,22 +40,7 @@ protected override bool Receive(object message)
                         Context.ActorOf(Props.Create(() => new UdpConnection(_udpConn, commander, connect)));
                         return true;
                     }
-                case DeadLetter dl when dl.Message is UdpConnected.SocketCompleted completed:
```
Arkatufus, line 45: No need for this anymore, because SocketAsyncEventArgs will never be embedded in messages.

src/core/Akka/IO/UdpConnection.cs
```diff
@@ -151,26 +156,25 @@ private bool Connected(object message)
...
+                    return true;
+                }
                 default: return false;
             }
         }
 
         private void DoRead(SocketReceived received, IActorRef handler)
         {
-            var e = received.EventArgs;
-            var buffer = new ByteBuffer(e.Buffer, e.Offset, e.BytesTransferred);
```
Arkatufus, line 170: This should never happen anymore, all handled in the SocketAsyncEventArgs pool

src/core/Akka/IO/UdpConnection.cs
```diff
@@ -233,19 +236,20 @@ private void ReportConnectFailure(Action thunk)
...
 
+        private SocketAsyncEventArgs _pendingReceive;
+        
         private void ReceiveAsync()
         {
+            if (_pendingReceive != null)
+                return;
+            
             var e = Udp.SocketEventArgsPool.Acquire(Self);
-            var buffer = Udp.BufferPool.Rent();
```
Arkatufus, line 239: Same here, we don't have to worry about this anymore.

Aaronontheweb:
@Arkatufus 

* Have a number of UDP test failures
* API approval failures

src/core/Akka/IO/Buffers/DirectBufferPool.cs
Aaronontheweb: Is this still necessary if we're disabling the pool for both TCP and UDP?

src/core/Akka.Tests/IO/UdpConnectedIntegrationSpec.cs
```diff
@@ -141,5 +151,58 @@ public void The_UDP_connection_oriented_implementation_must_to_send_batch_writes
...
+            clientEp.Tell(UdpConnected.Disconnect.Instance, client);
+            client.ExpectMsg<UdpConnected.Disconnected>();
+            
+            // wait for all SocketAsyncEventArgs to be released
+            Thread.Sleep(1000);
+            
+            poolInfo = udpConnection.SocketEventArgsPool.BufferPoolInfo;
+            poolInfo.Type.Should().Be(typeof(DirectBufferPool));
+            poolInfo.Free.Should().Be(poolInfo.TotalSize);
+            poolInfo.Used.Should().Be(0);
```
Aaronontheweb, line 205: LGTM

src/core/Akka.Tests/IO/UdpIntegrationSpec.cs
```diff
@@ -194,6 +204,97 @@ void CheckSendingToServer(ByteString expected)
...
+            client.Tell(Udp.Unbind.Instance, clientProbe);
+            clientProbe.ExpectMsg<Udp.Unbound>();
+            
+            // wait for all SocketAsyncEventArgs to be released
+            Thread.Sleep(1000);
+            
+            poolInfo = udp.SocketEventArgsPool.BufferPoolInfo;
+            poolInfo.Type.Should().Be(typeof(DirectBufferPool));
+            poolInfo.Free.Should().Be(poolInfo.TotalSize);
+            poolInfo.Used.Should().Be(0);
```
Aaronontheweb, line 252: LGTM

src/core/Akka/IO/Buffers/DirectBufferPool.cs
```diff
@@ -131,6 +149,17 @@ public DirectBufferPool(int bufferSize, int buffersPerSegment, int initialSegmen
             }
         }
 
+        public BufferPoolInfo Diagnostics()
```
Aaronontheweb, line 152: Good idea

src/core/Akka/IO/Buffers/DirectBufferPool.cs
Aaronontheweb: Why would it only happen during development?

src/core/Akka/IO/SocketEventArgsPool.cs
Aaronontheweb: Add a comment saying this and explaining why

src/core/Akka/IO/Udp.cs
Aaronontheweb: And we make a copy of the data, so there's no problems if the `SocketAsyncEventArgs` get processed and the buffer pool is released.

src/core/Akka/IO/UdpConnected.cs
Aaronontheweb: Got it

Aaronontheweb (approved):
Left some questions, nitpicks, and minor suggestions but this looks like a great structural improvement for Akka.IO.Udp. Nice work.

src/core/Akka/IO/Buffers/DirectBufferPool.cs
Aaronontheweb: Ah nevermind, I got it. We still pool bytes but not SAEA.

Arkatufus:
I mean its there so we can catch it while we're developing Akka, I think its better if I change that to an error log instead.

Aaronontheweb:
You can use 'Debug.Assert' for that

Arkatufus:
Done with the changes

PR #5412, opened 2021-12-03
Mark socket connection messages with IDeadLetterSuppression

### LINKED ISSUE #5318: UDP Connected: Unhandled message SocketReceived
**Version Information**
Version of Akka.NET? 1.4.25

**Describe the bug**
When communication with the device is complete, we send a UdpConnected.Disconnect message to close the socket. 
After that, we receive UdpConnected.Disconnected but we see an unhandled message in UdpConnection actor: 

  Message [SocketReceived] to [akka://dlmshe/system/IO-UDP-CONN/$a#267206278] was not delivered. [1] dead letters 
  encountered. If this is not an expected behavior then [akka://dlmshe/system/IO-UDP-CONN/$a#267206278] may have 
  terminated unexpectedly. This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 
  'akka.log-dead-letters-during-shutdown'.

**Expected behavior**
UDP socket is closed successfully. 

**Actual behavior**
UDP socket is closed successfully but we see an unhandled message. 

**Environment**
Windows
.NET Framework 4.8

DeadLetter
[akka_seq_logs.csv](https://github.com/akkadotnet/akka.net/files/7345357/akka_seq_logs.csv)

Arkatufus:
Closes #5318
Socket connection messages like SocketReceived can arrive late to a dead socket listener after the socket actor is closed.
The log message from the deadletter actor made it look like it is a bigger problem than it is, when we know that this is a benign behaviour.
Marking these messages as IDeadLetterSuppression so it will not confuse user when it happens.

### DIFF
--- a/src/core/Akka/IO/Tcp.cs
+++ b/src/core/Akka/IO/Tcp.cs
@@ -6,13 +6,10 @@
 //-----------------------------------------------------------------------
 
 using System;
-using System.Reflection;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
-using System.Net.Sockets;
-using System.Runtime.CompilerServices;
 using Akka.Actor;
 using Akka.Configuration;
 using Akka.Dispatch;
@@ -53,7 +50,8 @@ namespace Akka.IO
 
         #region internal connection messages
         
-        internal abstract class SocketCompleted : INoSerializationVerificationNeeded { }
+        internal abstract class SocketCompleted : INoSerializationVerificationNeeded, IDeadLetterSuppression 
+        { }
 
         internal sealed class SocketSent : SocketCompleted
         {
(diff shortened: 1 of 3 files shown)


src/core/Akka/IO/Tcp.cs
```diff
@@ -53,7 +50,8 @@ public override TcpExt CreateExtension(ExtendedActorSystem system)
 
         #region internal connection messages
         
-        internal abstract class SocketCompleted : INoSerializationVerificationNeeded { }
+        internal abstract class SocketCompleted : INoSerializationVerificationNeeded, IDeadLetterSuppression
```
Aaronontheweb, line 53: LGTM

Aaronontheweb (approved):
LGTM - no need for API approval as these are all `internal` classes.

Aaronontheweb:
Clearing milestone since the issue this resolves is part of it.

PR #5416, opened 2021-12-04
Refactor Props

Zetanova:
refactor of class Props to reduce view allocations, seal the class, 
and adding the possibility to use the same `IIndirectActorProducer` for multiple Props instances.

There is new an new obsolete `IIndirectActorProducerWithActorType` for legacy Akka.DI.Core,
but I don't know the legacy DI well to know if there will be some binary combability issues.

Improvements for **each props instance**:
1x less object instance with min. 1 reference field (the producer instance) 
1x less reference field (outtype field) 
View less vcalls


src/contrib/dependencyinjection/Akka.DI.Core/DIExt.cs
```diff
@@ -38,7 +38,7 @@ public void Initialize(IDependencyResolver dependencyResolver)
         /// <returns>A <see cref="Akka.Actor.Props"/> configuration object for the given actor type.</returns>
         public Props Props(Type actorType)
         {
-            return new Props(typeof(DIActorProducer), new object[] { dependencyResolver, actorType });
```
Zetanova, line 41: If this was the only place where the old DIActorProducer got with `new object[] { dependencyResolver, actorType }` created, then we could remove the obsolete `Props.CreateProducer` method.

src/core/Akka/Actor/Props.cs
```diff
@@ -554,20 +538,29 @@ public virtual ActorBase NewActor()
...
+                return (InvalidProducer.Default, typeof(EmptyActor), NoArgs);
 
             if (typeof(IIndirectActorProducer).IsAssignableFrom(type))
-                return Activator.CreateInstance(type, args).AsInstanceOf<IIndirectActorProducer>();
+            {
+                var producer = (IIndirectActorProducer)Activator.CreateInstance(type, args);
+                if (producer is IIndirectActorProducerWithActorType p)
+                    return (producer, p.ActorType, NoArgs);
+
+                //maybe throw new ArgumentException($"Unsupported legacy actor producer [{type.FullName}]", nameof(type));
```
Zetanova, line 558: here I am not sure, maybe to throw is better

src/core/Akka/Actor/Props.cs
```diff
@@ -689,92 +669,20 @@ public FactoryConsumer(Func<TActor> factory)
                 _factory = factory;
             }
 
-            public ActorBase Produce()
+            public ActorBase Produce(Props props)
             {
                 return _factory.Invoke();
```
Zetanova, line 674: maybe insert here a safety type check with `props.Type` ?

src/core/Akka/Actor/Props.cs
```diff
@@ -244,8 +244,11 @@ public bool Equals(Props other)
         {
             if (ReferenceEquals(null, other)) return false;
             if (ReferenceEquals(this, other)) return true;
-            return CompareDeploy(other) && CompareSupervisorStrategy(other) && CompareArguments(other) &&
-                   CompareInputType(other);
+            return Type == other.Type
+                && _producer.GetType() == other._producer.GetType()
```
Zetanova, line 248: I added here a simple producer type compare, there was  no comparison on the producer before

Zetanova:
Some F# test are failing with an `ArgumentException` instead of an `InvalidCastException` 
I dont think that the InvalidCastException was right before.

I can not fix F# test.

Zetanova:
All tests running successfully on my local machine.
The `probe.ExpectMsg(...)` got somehow very racy, didn't make any changes related to this method. 

Is the build server just over committed or throttled and just slow?

src/core/Akka/Actor/Props.cs
Zetanova: One other idea would be to let test if the producer implements `IEqualityComparer<Akka.Props>` 
or maybe add a field for it into the producer interface.

Aaronontheweb:
@Zetanova the test suite has become much more flakey since the .NET 6 upgrade due to the new "deadlock detection" algos that the managed thread pool uses. We're going to need to refactor all of it to be `async` probably, which we'd already started gradually doing.

Zetanova:
What exactly happens with the new "deadlock detection" ?

The message is not pushed into the Queue of TestProp.
This can only mean 3 things:
1) the adding thread of to the queue and the waiting thread are the same.
2) the actor-system had internal errors and try-catch-ignore them
3) the actor-system add log-errors over AkkaLogger that itself uses a Mailbox. 
The effect is that Log messages getting ignored at startup or do not write out after a failed test

Aaronontheweb:
> What exactly happens with the new "deadlock detection" ?

I'm not entirely sure, but the behavior of the `ThreadPool` now means that new threads will be spawned when a thread is blocking for longer than it "should" based on some heuristics. We need to basically start awaiting on things inside our test suite to avoid triggering this, rather than blocking.

Zetanova:
Some tests failed only in "debug" test run, but not with a simple "run" test
Both in debug builds

I found In the `CircuitBreaker` a sync method that waited only 50ms for execution.
The task just didn't get activated/run in the period by the threadpool/task-scheduler
I could refactor the method to fix it. 

But the underlining problem is most likely that scheduled task just do not activated in time,
specially with a very low timeout (under 50ms)

Zetanova:
This failed test in streams is strange
The actor system begins to shutdown near instantly at test-start
even before the timeout in the test itself

This indicates some startup-error and lost-log messages,
like mention in a comments above

image

Zetanova:
@Aaronontheweb  two failing tests are left, most likely it's just the racy akka-system startup or?
The test do not fail locally in debug-test-run and normal test-run

Aaronontheweb:
@Zetanova the DotNetty spec I think is just designed poorly. As for the ClusterSingletonProxy spec that failed - not entirely sure. All of our flip rate data has changed a lot since we added .NET 6 to the pipeline.

Zetanova:
I added a hack into akka-system.start to wait for the provider ready, this helps a lot.
We removed view PR's ago some locks in the startup process and because of it the tests got racy.

@Aaronontheweb 
Something strange here, where I would need second eye pair:
The test `Akka.Remote.Tests.Transport.DotNettySslSupportSpec.Secure_transport_should_be_possible_between_systems_using_thumbprint` is failing only if its run single/alone, but is always successful if its run together with the other test.

It produces this exception:
 ---> System.ComponentModel.Win32Exception (0x8009030D): The credentials supplied to the package were not recognized

image

the screenshot with two tests:
image

Aaronontheweb:
> I added a hack into akka-system.start to wait for the provider ready, this helps a lot.,

ah, is this because we no longer block while waiting for the loggers to startup?

Zetanova:
The start hack in the last commit fixed the startup

But it looks like that a new Thread().Start() does not get created instantly anymore,
maybe when the system is under CPU/MEM pressure?

This test is a good example, has nothing to do with System startup more or less:
Akka.Tests.Actor.Scheduler.DefaultScheduler_TellScheduler_Cancellation_Tests.When_canceling_existing_running_repeaters_by_scheduling_the_cancellation_ahead_of_time_Then_their_future_actions_should_not_be_invoked

the test code: 
image

It failed at `ExpectNoMsg(200);`


The hashwheelscheduler thread gets only started with the first message,
But 2 got send out, this mean that _thread.Start() took longer and/or _startTime was 1 
or got cached for long a period  >150ms

Zetanova:
some component in the startup still is racy, hard to find

We would need to refactor the startup completely, best insert bootstraps/phases queue like on shutdown.

Zetanova:
I cant reproduce failed test on my dev machine i7-920 (running since 2009)
Must be something with L2/L3 Cache, 
I think, one or two `Volatile.Write()` are missing somewhere on startup

Zetanova:
@Aaronontheweb  I added now a hack into the TestPrope.Create method
https://github.com/akkadotnet/akka.net/pull/5416/commits/fcf18c24996d8df2b836386e7155a0ce002398ef

This produces now errors local, but a lot on the build-server, strange only in `Akka.Cluster.Scharding` and `Akka.Persistence`
and all other failed test are gone.

Aaronontheweb:
You have a lot of test failures now (77) - looks like one of your changes broke something in the logging system that the TestKit uses.

Zetanova:
I put a hack inside the commit to instantly await a log message and it is failing now

Need to analyze why it is not support one some tests

Zetanova:
I removed now the testprope startup await hack,
will try to find the racy component locally and create a new PR for it, 
but it is not from this PR

@Aaronontheweb  if its ok for can be merged

Aaronontheweb:
Finally looking at this now

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -225,6 +225,17 @@ public void Start()
                 {
                     _log.Info(Settings.ToString());
                 }
+
+                //HACK: ensure provider started
+                {
```
Aaronontheweb, line 230: Why is this necessary? Just to stop NREs from being raised by the remoting system? I don't like this - the idea behind `_provider.DefaultAddress` possibly returning `null` or throwing is that those are handled exceptions that occur as a result of the underlying transport not yet being bound, which is going to happen. Blocking and forcing the transport before we start creating actors is going to create problems. I'd like to see this removed unless there's a really good reason to keep it.

src/core/Akka/Pattern/CircuitBreaker.cs
```diff
@@ -226,15 +226,12 @@ public Task WithCircuitBreaker(Func<Task> body)
...
-            }
-            if (cbTask.Exception != null)
-            {
-                ExceptionDispatchInfo.Capture(cbTask.Exception).Throw();
-            }
+            var cts = new CancellationTokenSource(CallTimeout);
+            var task = new Task(body, cts.Token);
+            var cbTask = WithCircuitBreaker(() => task);
+            task.RunSynchronously();
+            if (!task.IsCanceled)
```
Aaronontheweb, line 233: Yeah this is a much cleaner way of writing what we had before

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
```diff
@@ -82,7 +81,7 @@ public HashedWheelTimerScheduler(Config scheduler, ILoggingAdapter log) : base(s
         /// <summary>
         /// 0 - init, 1 - started, 2 - shutdown
         /// </summary>
-        private volatile int _workerState = WORKER_STATE_INIT;
+        private int _workerState = WORKER_STATE_INIT;
```
Aaronontheweb, line 84: Guess it really did not need to be `volatile` if we're using `Interlocked` to access it primarily.

src/core/Akka/Event/Logging.cs
```diff
@@ -105,14 +105,12 @@ public static string FromActor(IActorContext actor, ActorSystem system)
...
-            catch // can fail if the ActorSystem (remoting) is not completely started yet
-            {
-                return a.Path.ToString();
-            }
+            Address address = null;
+            if (system is ExtendedActorSystem extendedSystem && !(extendedSystem.Provider is null))
+                address = extendedSystem.Provider.DefaultAddress;
+
+            // maybe the ActorSystem (remoting) is not completely started yet
+            return address is null ? a.Path.ToString() : a.Path.ToStringWithAddress(address);
```
Aaronontheweb, line 113: This is a good change too

src/core/Akka/Actor/Props.cs
```diff
@@ -689,92 +661,20 @@ public FactoryConsumer(Func<TActor> factory)
...
-
-    /// <summary>
-    ///     This class represents a specialized <see cref="Akka.Actor.Props" /> that uses dynamic invocation
-    ///     to create new actor instances, rather than a traditional <see cref="System.Activator" />.
-    ///     <note>
-    ///         This is intended to be used in conjunction with Dependency Injection.
-    ///     </note>
-    /// </summary>
-    /// <typeparam name="TActor">The type of the actor to create.</typeparam>
-    internal class DynamicProps<TActor> : Props where TActor : ActorBase
```
Aaronontheweb, line 733: I take it this wasn't used anywhere?

src/core/Akka/Actor/Props.cs
```diff
@@ -162,10 +141,12 @@ public Props(Deploy deploy, Type type, IEnumerable<object> args)
...
         /// <param name="args">The arguments needed to create the actor.</param>
         /// <exception cref="ArgumentException">This exception is thrown if <paramref name="type" /> is an unknown actor producer.</exception>
-        public Props(Deploy deploy, Type type, params object[] args) 
-            : this(CreateProducer(type, args), deploy, args) // have to preserve the "CreateProducer" call here to preserve backwards compat with Akka.DI.Core
+        public Props(Deploy deploy, Type type, params object[] args)
         {
-
+            Deploy = deploy;
+            // have to preserve the "CreateProducer" call here to preserve backwards compat with Akka.DI.Core
+            (_producer, Type, Arguments) = CreateProducer(type, args);
```
Aaronontheweb, line 148: Ok, good

src/core/Akka/Actor/Props.cs
```diff
@@ -32,12 +30,12 @@ namespace Akka.Actor
     /// </code>
     ///     </example>
     /// </summary>
-    public class Props : IEquatable<Props>, ISurrogated
+    public sealed class Props : IEquatable<Props>, ISurrogated
```
Aaronontheweb, line 33: I'm not sure how I feel about this, in terms of keeping this class open for extension. I checked the Phobos source to see if we subclass `Props` anywhere there and we do not (we do subclass `Deploy`, one of the sub-components of `Props`) so if that doesn't need to extend `Props` I can't imagine many other plugins that do...

Aaronontheweb (changes requested):
Mostly looks good, but have some changes that either need to be explained in more detail or removed

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1409,31 +1413,29 @@ namespace Akka.Actor
...
         [Newtonsoft.Json.JsonIgnoreAttribute()]
         public string Dispatcher { get; }
         public static Akka.Actor.Props Empty { get; }
         [Newtonsoft.Json.JsonIgnoreAttribute()]
         public string Mailbox { get; }
         [Newtonsoft.Json.JsonIgnoreAttribute()]
         public Akka.Routing.RouterConfig RouterConfig { get; }
-        public Akka.Actor.SupervisorStrategy SupervisorStrategy { get; set; }
+        public Akka.Actor.SupervisorStrategy SupervisorStrategy { get; }
+        public static Akka.Actor.Props Terminated { get; }
```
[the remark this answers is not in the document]
Zetanova: Yes, the error message.
`Props.Terminated` is like `Props.Empty` but with the new `ÌnvalidProviderActorProducer` to throw the error. 
It is currently only used by ActorCell and its only a refactor

src/core/Akka/Actor/Props.cs
Zetanova: Yes, or with the old Akka.DI.Core once

src/core/Akka/Actor/Props.cs
Aaronontheweb: If it's _actually_ unsupported now, we should throw here.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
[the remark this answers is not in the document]
Aaronontheweb: Ok, got it

src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProvider.cs
```diff
@@ -93,43 +93,23 @@ public override ServiceProvider CreateExtension(ExtendedActorSystem system)
...
     /// </summary>
-    internal class ServiceProviderActorProducer : IIndirectActorProducer
+    internal sealed class ServiceProviderActorProducer : IIndirectActorProducer
     {
         private readonly IServiceProvider _provider;
-        private readonly object[] _args;
-
-        public ServiceProviderActorProducer(IServiceProvider provider, Type actorType, object[] args)
+        
+        public ServiceProviderActorProducer(IServiceProvider provider)
```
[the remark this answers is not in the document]
Zetanova: Yes, the hole point of this PR 1 less object for each Prop (and view references)

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
Zetanova: The akka-system startup is a complete mess, we removed 2-3 locks with our last commits 
and akka.remote got completely racy, so many components rely on `Provider.DefaultAddress` 
and just NRE in the unit tests, 

this hack comes more to the behavior of version 1.4.21

src/core/Akka/Pattern/CircuitBreaker.cs
Zetanova: Not only cleaner, before it was racy

src/core/Akka/Actor/Scheduler/HashedWheelTimerScheduler.cs
Zetanova: yes, absolutely `Interlocked` has a memory barrier 
and `volatile` fields used often are a perf killer.

Best is to use `Violete.Write()`  for fields that are written only once or view times.

src/core/Akka/Event/Logging.cs
Zetanova: There should never be a expected throw (if possible)
This is why `int.TryParse("aa", out var i)` got added,
`try { i = int.Parse("aa"); } catch(FormatException ex) { ... }`  was a perf killer

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
Aaronontheweb: We should probably rewrite, maybe for Akka.NET 2.0 or 1.5, `ActorSystem.Create` to return a `Task` and just `await` all of the startup functions that occur inside the `ActorRefProvider` - so once an `ActorSystem` is started it's either really started or the startup failed and an exception was thrown right away.

Aaronontheweb: But as far as this PR goes - I guess we can live with this

src/contrib/dependencyinjection/Akka.DI.Core/DIExt.cs
Zetanova: how do we mark deletion in version?
simply put some text in the obsolete attribute message?

src/core/Akka/Actor/Props.cs
Zetanova: `FactoryConsumer` was not used, I reactivated it for the Producer by Custom-Functon

src/contrib/dependencyinjection/Akka.DI.Core/DIExt.cs
Aaronontheweb: Correct - just mention that in the `ObsoleteAttribute`'s text

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.approved.txt
```diff
@@ -1072,10 +1072,14 @@ namespace Akka.Actor
     }
     public interface IIndirectActorProducer
     {
-        System.Type ActorType { get; }
-        Akka.Actor.ActorBase Produce();
```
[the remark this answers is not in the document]
Aaronontheweb: I think we're good

Aaronontheweb (approved):
LGTM - we're doing the 1.4.29 release today and I don't want to incorporate this change right away without having some further testing on subsequent PRs (I did the same thing with my `PipeTo` changes between 1.4.28 and 1.4.29), but this will make it into 1.4.30

src/core/Akka/Actor/Props.cs
```diff
@@ -353,15 +323,15 @@ public override int GetHashCode()
...
-            if (newExpression == null)
-                throw new ArgumentException("The create function must be a 'new T (args)' expression");
+            var newExpression = factory.Body as NewExpression
+                ?? throw new ArgumentException("The create function must be a 'new T (args)' expression");
 
             var args = newExpression.GetArguments();
+            args = args.Length > 0 ? args : NoArgs;
 
-            return new Props(new ActivatorProducer(typeof(TActor), args), DefaultDeploy, args){ SupervisorStrategy = supervisorStrategy };
+            return new Props(ActivatorProducer.Instance, DefaultDeploy, typeof(TActor), args) { SupervisorStrategy = supervisorStrategy };
```
to11mtm, line 334: If we're already getting the args, would there be any benefit in also sniffing out the Ctor in `.GetArguments()`? (Can we even do that in a way that works OK with expectations of Props?)

That method was originally CopyPasta'd from OddJob, where we weren't newing up objects but rather were getting arguments for method calls. If you can actually hang on to what constructor is called (would be easy) you there are opportunities to elide penalties involved with `Activator.CreateInstance`

src/core/Akka/Actor/Props.cs
```diff
@@ -353,15 +323,15 @@ public override int GetHashCode()
...
+                return new Props(new FactoryConsumer<TActor>(factory.Compile()), DefaultDeploy, typeof(TActor), NoArgs);
 
-            var newExpression = factory.Body.AsInstanceOf<NewExpression>();
-            if (newExpression == null)
-                throw new ArgumentException("The create function must be a 'new T (args)' expression");
+            var newExpression = factory.Body as NewExpression
+                ?? throw new ArgumentException("The create function must be a 'new T (args)' expression");
 
             var args = newExpression.GetArguments();
+            args = args.Length > 0 ? args : NoArgs;
```
to11mtm, line 332: Consider, if you don't do the deed commented below about pulling ConstructorInfo:

```
var args = newExpression.Arguments > 0 ? newExpression.GetArguments() : NoArgs;
```

src/core/Akka/Actor/Props.cs
```diff
@@ -162,10 +141,12 @@ public Props(Deploy deploy, Type type, IEnumerable<object> args)
         /// <param name="type">The type of the actor to create.</param>
         /// <param name="args">The arguments needed to create the actor.</param>
         /// <exception cref="ArgumentException">This exception is thrown if <paramref name="type" /> is an unknown actor producer.</exception>
-        public Props(Deploy deploy, Type type, params object[] args) 
-            : this(CreateProducer(type, args), deploy, args) // have to preserve the "CreateProducer" call here to preserve backwards compat with Akka.DI.Core
+        public Props(Deploy deploy, Type type, params object[] args)
         {
-
+            Deploy = deploy;
+            // have to preserve the "CreateProducer" call here to preserve backwards compat with Akka.DI.Core
```
to11mtm, line 147: Thank you. :)

src/core/Akka/Actor/Props.cs
Aaronontheweb: That's a good idea

src/core/Akka.TestKit/ActorCellKeepingSynchronizationContext.cs
```diff
@@ -41,23 +41,25 @@ public ActorCellKeepingSynchronizationContext(ActorCell cell)
         /// <param name="state">TBD</param>
         public override void Post(SendOrPostCallback d, object state)
         {
-            ThreadPool.QueueUserWorkItem(_ =>
+            ThreadPool.QueueUserWorkItem(s =>
             {
+                var t = ((SendOrPostCallback, object, ActorCellKeepingSynchronizationContext, ActorCell))s;
```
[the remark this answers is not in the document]
Aaronontheweb: This only affects the `TestKit` so I think we should be ok

src/contrib/dependencyinjection/Akka.DI.Core/DIActorProducer.cs
```diff
@@ -33,26 +34,20 @@ public DIActorProducer(IDependencyResolver dependencyResolver, Type actorType)
...
+        
         /// <summary>
         /// Creates an actor based on the container's implementation specific actor factory.
         /// </summary>
         /// <returns>An actor created by the container.</returns>
-        public ActorBase Produce()
+        public ActorBase Produce(Props props)
         {
-            return actorFactory();
+            if (props.Type != _actorType)
```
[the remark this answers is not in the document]
Aaronontheweb: This is part of the deprecated DI library so I'm not as worried about it - but did we end up changing something here that we shouldn't have?

src/core/Akka/Actor/Props.cs
Aaronontheweb: Can you file an issue for this so we remember to do this as part of 1.5?

src/contrib/dependencyinjection/Akka.DI.Core/DIActorProducer.cs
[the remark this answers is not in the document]
Zetanova: I insert here only a safty check the `props.Type` and `_actorType` should have the same type,
because it's only copy

src/core/Akka.TestKit/ActorCellKeepingSynchronizationContext.cs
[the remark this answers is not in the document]
Zetanova: I don't know what's faster here `ValueTuple` or `System.Tuple`, wanted only remove the scope. 

It does not matter here, but when a scope is used with ThreadPool, 
the context will not be GC until the action runs, it can lead to a self-reference in some scenarios.

src/core/Akka/Actor/Props.cs
Zetanova: @to11mtm We could move the logic into a custom-actor-producer and use the instance for mass creation of actors with the same props-instance, everything else will always require runtime binding or only dynamic-compile once.

PR #5418, opened 2021-12-07
[Documentation] Add extra help information on serializer id errors

Arkatufus:


Arkatufus:
Written documentation is not complete yet, code is complete.

docs/articles/toc.yml
```diff
@@ -202,3 +202,5 @@
     href: utilities/circuit-breaker.md
 - name: Configurations
   href: configuration/toc.yml
+- name: Serialization
+  href: serialization/toc.yml
```
Aaronontheweb, line 206: We should probably migrate some stuff to this section

Aaronontheweb (approved):
LGTM

Aaronontheweb:
This should help quite a bit when debugging serialization configuration errors.

docs/articles/toc.yml
Arkatufus: There's a plan on moving Hyperion to its own page, so I think this section would host it.

PR #5424, opened 2021-12-10
MDB throws a MDB_NOTFOUND exception when the database directory already exists

### LINKED ISSUE #5423: [DData] LMDB throws a MDB_NOTFOUND exception when the database directory already exists
**Version Information**
Version of Akka.NET? 
1.4.28
Which Akka.NET Modules?
Akka.DistributedData.LightningDB

**Describe the bug**
When using a user defined directory name to store LMDB database, if the directory already exist, LmdbDurableStore will throw a MDB_NOTFOUND exception and fail during the `LoadAll` state.

**To Reproduce**

**Expected behavior**
The actor should reuse the database directory

**Actual behavior**
The actor crashed

**Screenshots**

Arkatufus:
Closes #5423

- [x] Create a bug reproduction spec
- [x] Create fix


src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
```diff
@@ -110,41 +112,24 @@ protected override void PostStop()
...
-
-                t0.Stop();
-                if (_log.IsDebugEnabled)
-                    _log.Debug($"Init of LMDB in directory [{_dir}] took [{t0.ElapsedMilliseconds} ms]");
-            }
-            else
+            using (var tx = env.BeginTransaction())
+            using (tx.OpenDatabase(DatabaseName, new DatabaseConfiguration { Flags = DatabaseOpenFlags.Create }))
             {
-                env = new LightningEnvironment(_dir)
```
Arkatufus, line 140: The problem was here, the code assumes that the database is already created if the directory existed.
Just have to force LMDB to recreate the database if it doesn't exist.

src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
```diff
@@ -207,12 +192,13 @@ private void Init()
...
                 var t0 = Stopwatch.StartNew();
-                using (var env = GetLightningEnvironment())
-                using (var tx = env.BeginTransaction(TransactionBeginFlags.ReadOnly))
-                using(var db = tx.OpenDatabase(DatabaseName))
-                using(var cursor = tx.CreateCursor(db))
+                
+                try
                 {
-                    try
+                    using (var env = GetLightningEnvironment())
```
Arkatufus, line 198: Might as well re-arrange the try...catch block to include the LMDB environment initialization code so it can catch exceptions thrown there

src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
```diff
@@ -251,23 +237,25 @@ private void DoWriteBehind()
...
                 var t0 = Stopwatch.StartNew();
                 using (var env = GetLightningEnvironment())
-                using(var tx = env.BeginTransaction())
-                using (var db = tx.OpenDatabase(DatabaseName))
+                using (var tx = env.BeginTransaction())
                 {
                     try
                     {
-                        foreach (var entry in _pending)
+                        using (var db = tx.OpenDatabase(DatabaseName))
```
Arkatufus, line 244: Same as above, rearranged the try...catch block

src/contrib/cluster/Akka.DistributedData.LightningDB/LmdbDurableStore.cs
Aaronontheweb: Makes sense

Arkatufus:
I can't replicate that failure locally, i think it has something to do with the test file system

src/contrib/cluster/Akka.DistributedData.Tests/Akka.DistributedData.Tests.csproj
```diff
@@ -5,6 +5,7 @@
   <PropertyGroup>
     <AssemblyTitle>Akka.DistributedData.Tests</AssemblyTitle>
     <TargetFrameworks>$(NetFrameworkTestVersion);$(NetTestVersion);$(NetCoreTestVersion)</TargetFrameworks>
+    <PlatformTarget>x64</PlatformTarget>
```
Aaronontheweb, line 8: Might want to add a comment explaining why this is necessary - feel free to add it in a new PR @Arkatufus

Aaronontheweb (approved):
LGTM

PR #5430, opened 2021-12-14
Added InverseFishForMessage and nuget package fluent-assertions

brah-mcdude:
This pull request is based on a thread I had with @Aaronontheweb on Gitter: https://gitter.im/akkadotnet/akka.net?at=61ae5b0acd3f06175df32eda.

### DIFF
--- a/src/core/Akka.TestKit/Akka.TestKit.csproj
+++ b/src/core/Akka.TestKit/Akka.TestKit.csproj
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <EmbeddedResource Include="Configs\TestScheduler.conf;Internal\Reference.conf" />
+    <PackageReference Include="FluentAssertions" Version="5.10.3" />
     <ProjectReference Include="..\Akka\Akka.csproj" />
   </ItemGroup>
 
--- a/src/core/Akka.TestKit/TestKitBase_Receive.cs
+++ b/src/core/Akka.TestKit/TestKitBase_Receive.cs
@@ -9,7 +9,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using Akka.TestKit.Internal;
+using FluentAssertions;
 
 namespace Akka.TestKit
 {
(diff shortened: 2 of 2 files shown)


src/core/Akka.TestKit/Akka.TestKit.csproj
```diff
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <EmbeddedResource Include="Configs\TestScheduler.conf;Internal\Reference.conf" />
+    <PackageReference Include="FluentAssertions" Version="5.10.3" />
```
Aaronontheweb, line 14: We should probably just take a dependency on FluentAssertions in the TestKit. I don't know why we haven't yet. 

I'll make a bigger issue for this and we'll revisit that prior to releasing v1.4.30.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -56,6 +58,23 @@ public T FishForMessage<T>(Predicate<T> isMessage, TimeSpan? max = null, string
             }
         }
 
+        /// <summary>
+        /// Receives messages until <paramref name="max"/>. Ignores all messages except for a message of type <typeparamref name="T"/>. Asserts that all messages are not of the of type <typeparamref name="T"/>. Note that when comparing types, inheritance is ignored, in other words, only perfectly matching types are asserted.
```
Aaronontheweb, line 62: Need to format this into a few separate lines.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -56,6 +58,23 @@ public T FishForMessage<T>(Predicate<T> isMessage, TimeSpan? max = null, string
...
+        /// </summary>
+        /// <typeparam name="T">The type that the message is not supposed to be.</typeparam>
+        /// <param name="probe"></param>
+        /// <param name="max"></param>
+        public async static Task InverseFishForMessage<T>(TestProbe probe, TimeSpan? max = null)
```
Aaronontheweb, line 67: Should probably name this `ReceiveUntil`, `FishUntil` or something like that.

Aaronontheweb (approved):
LGTM - left behind some nitpicks but we can deal with those in separate PRs

PR #5431, opened 2021-12-15
Remove TestUtils.TemporaryServerAddress, suspected racy test source

Arkatufus:


src/core/Akka.Tests.Shared.Internals/AkkaSpec.cs
```diff
@@ -115,9 +115,12 @@ private static string GetCallerName()
         public static Config AkkaSpecConfig { get { return _akkaSpecConfig; } }
 
         protected T ExpectMsgPf<T>(TimeSpan? timeout, string hint, Func<object, T> function)
+            => ExpectMsgPf(timeout, hint, this, function);
+        
+        protected T ExpectMsgPf<T>(TimeSpan? timeout, string hint, TestKitBase probe, Func<object, T> function)
```
Arkatufus, line 120: Added ExpectMsgPf overload that can take a test probe as parameter

src/core/Akka.Tests/IO/TestUtils.cs
```diff
@@ -5,37 +5,15 @@
...
-using System.Linq;
 using System.Net;
-using System.Net.Sockets;
-using static Akka.IO.IpExtensions;
 
 namespace Akka.Tests.IO
 {
     public static class TestUtils
     {
-        public static IPEndPoint TemporaryServerAddress(string hostName = "127.0.0.1", bool udp = false)
```
Arkatufus, line 18: Removed, suspected port binding race

src/core/Akka.Tests.Shared.Internals/AkkaSpec.cs
Aaronontheweb: Nice

src/core/Akka/IO/UdpConnection.cs
```diff
@@ -73,7 +73,7 @@ private void DoConnect(EndPoint address)
         {
             ReportConnectFailure(() =>
             {
-                _socket = new Socket(SocketType.Dgram, ProtocolType.Udp) { Blocking = false };
+                _socket = new Socket(AddressFamily.InterNetwork, SocketType.Dgram, ProtocolType.Udp) { Blocking = false };
```
Aaronontheweb, line 76: Do we need to do IPV4 / IPV6 handling here?

Aaronontheweb (commented):
Left one question on the `UdpConnection` changes

src/core/Akka/IO/UdpConnection.cs
Arkatufus: Maybe it should match the connect message address type?

Arkatufus: Matched the behavior to UdpListener and I caught a bug in the code

src/core/Akka/IO/UdpConnection.cs
```diff
@@ -73,15 +73,17 @@ private void DoConnect(EndPoint address)
...
+                           new Inet.DatagramChannelCreator()).Create(address.AddressFamily);
+                _socket.Blocking = false;
 
                 foreach (var option in _connect.Options)
                 {
                     option.BeforeDatagramBind(_socket);
                 }
 
                 if (_connect.LocalAddress != null)
-                    _socket.Bind(_connect.LocalAddress);
```
Arkatufus, line 84: Should not use the local address provided in the `Connect` message, because it could be changed by DNS lookup

Aaronontheweb (approved):
LGTM

Aaronontheweb:
@Arkatufus `Akka.Tests.IO.UdpConnectedIntegrationSpec.The_UDP_connection_oriented_implementation_must_to_send_batch_writes_and_reads` failed in a couple of places in this PR - could you investigate?

src/core/Akka.Tests/IO/UdpConnectedIntegrationSpec.cs
```diff
@@ -158,43 +171,36 @@ public void The_UDP_connection_oriented_implementation_must_not_leak_memory()
...
-
-            var client = CreateTestProbe();
-            udpConnection.Manager.Tell(new UdpConnected.Connect(client, serverAddress, clientAddress), client);
-            client.ExpectMsg<UdpConnected.Connected>();
-            var clientEp = client.LastSender;
+            var serverProbe = CreateTestProbe();
+            var (server, serverEndPoint) = BindUdp(serverProbe);
+
+            var clientProbe = CreateTestProbe();
+            var (client, clientEndPoint) = ConnectUdp(serverEndPoint, clientProbe);
```
Arkatufus, line 184: Changed this from using UdpConnected both ways to just one way because using connected both ways would require us to know both ports ahead of time, which is racy.

PR #5433, opened 2021-12-15
Make updates to `FishUntil`

Aaronontheweb:
Implement nitpicks from https://github.com/akkadotnet/akka.net/pull/5430

cc @brah-mcdude


brah-mcdude:
Extra nitpick.
About the following line: 
`                    return false; // we are not returning anything
`
It should actually be:
`                    return true; // we are not returning anything
`
Reason: 
It does not change the logic of the method. But, if we want to be more efficient and be congruent with the comment: "we are not returning anything".

Aaronontheweb:
Done

Aaronontheweb:
Going to have a compilation failure until I take this down and get a chance to update the test suite

PR #5436, opened 2021-12-16
add async startup

Zetanova:


Aaronontheweb:
Breaking changes in here for sure break compat with Phobos but we'll need to test to see if this resolves the regression in 1.4.29

Zetanova:
If there are conflicts with phobos, we can resolve them later on

Aaronontheweb:
Yep, it's not a deal breaker. Fixing clustering is the bigger priority.

Zetanova:
@Aaronontheweb Why did the test failed?
The log its a bid confusing, no unit tests failed

Aaronontheweb:
Part of the test suite locked up and timed out after 30 minutes, which means we don't get the report indicating which test failed. And wherever it failed, it did it consistently. Need to click through to the build log to see it:

https://dev.azure.com/dotnet/Akka.NET/_build/results?buildId=62800&view=logs&j=d3d8bb3a-a87f-5af3-6a16-90b99d49172b&t=1d19db56-c24e-5fbb-3f13-e91c53ee1789&l=3886

Looks like it's the Akka.FSharp specs

Zetanova:
Yes, i saw this too but don't understand why it failed and how to resolve it.
the FSharpSpecs running local successful

Aaronontheweb:
Looks like the remoting system deadlocked - see all of the `Uhandled` warnings that go on for a while before the test times out?

Zetanova:
I think i found it, its something in RemoteActorRefProvider.CreateInternals()
The dispatcher of the components in Internals get faster executed then the CreateInternals() returns

Zetanova:
one other major problem is the logging.
if something throws at startup, there is no output. 

One other suspicious thing is the terminator routine, 
I think it can deathlock with a startup exception.

Aaronontheweb:
I think this is great progress - just need to keep whittling down the issues the test suite raises

Aaronontheweb:
> one other major problem is the logging.
> if something throws at startup, there is no output.

Is your issue related to this? https://github.com/akkadotnet/akka.net/pull/4424

Aaronontheweb:
It's still an issue with remoting.

If I had to guess, the problem here is that the `/system` actors that run remoting _have to start_ before the `Init` call on the `RemoteActorRefProvider` exits. Something about the way this method has been updated prevents them from doing that consistently.

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -198,21 +199,63 @@ public override IActorRef SystemActorOf<TActor>(string name = null)
...
+                    _provider.Init(this);
+                    if (_provider is IInitializable init)
+                        await init.InitializeAsync(cancellationToken);
+                }
+
+                //load and init extensions
+                {
+                    var extensions = LoadExtensions();
+                    foreach (var init in extensions.OfType<IInitializable>())
+                        await init.InitializeAsync(cancellationToken);
```
Aaronontheweb, line 234: I think this is part of your problem - the old system had ordering in-place created as an effect of lazy loading. If `Cluster` depends on the `RARP` extension, `Cluster` would start `RARP` if it wasn't created already and would load the cached value.

Scanning the assembly and arbitrarily front-loading the extensions is non-deterministic and will result in the types of deadlocks you're seeing now.

I wouldn't front-load any of the extensions at all - lazy loading until they're needed on-demand is still the better approach.

src/core/Akka.Remote/Remoting.cs
```diff
@@ -114,18 +114,18 @@ internal interface IPriorityMessage { }
...
     {
         private readonly ILoggingAdapter _log;
-        private volatile IDictionary<string, HashSet<ProtocolTransportAddressPair>> _transportMapping;
-        private volatile IActorRef _endpointManager;
+        private IDictionary<string, HashSet<ProtocolTransportAddressPair>> _transportMapping;
+        private IActorRef _endpointManager;
 
         // This is effectively a write-once variable similar to a lazy val. The reason for not using a lazy val is exception
         // handling.
-        private volatile HashSet<Address> _addresses;
```
Aaronontheweb, line 122: Why replace this with a `Volatile.Write` down below?

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -198,21 +199,63 @@ public override IActorRef SystemActorOf<TActor>(string name = null)
...
                 RegisterOnTermination(StopScheduler);
-                _provider.Init(this);
-                LoadExtensions();
+
+                //init provider
+                {
+                    _provider.Init(this);
+                    if (_provider is IInitializable init)
+                        await init.InitializeAsync(cancellationToken);
+                }
```
Aaronontheweb, line 228: Why would we call `Init` _and_ `InitializeAsync` ?

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
Aaronontheweb: BTW, one simple fix here to see if it resolves the deadlock - don't `await` each individual plugin starting up. Compose them into a `WhenAll` and `await` on the group.

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -302,13 +337,13 @@ private void StopScheduler()
             sched?.Dispose();
         }
 
-        private void LoadExtensions()
+        private List<object> LoadExtensions()
```
Aaronontheweb, line 340: Can we type this usefully in order to avoid boxing?

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
Zetanova: Only Cluster and the RemoteActorRefProvider are supporting it,
yes the issue is WHO and WHEN the Cluster Extensions get cold and exit

src/core/Akka.Remote/Remoting.cs
Zetanova: Its a multi read and only single write field.
volatile as perf issues if something is accessing it multiple times

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
Zetanova: `IInitializable` is a feature faced, it does not replace the old API

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
Zetanova: This are only the extensions loaded from config, excluding lazy loaded extensions

Zetanova:
It was simple the requirement that the `Cluster` extension need be initialized inside/right after the `ActorRefProvider`
It does not support "lazy" loading but it is used like it.

The whole extensions system is none-deterministic.
Because many system actors are lazy loading extensions
and there execution depends on Dispatcher.

Changing the Dispatcher can change the extension loading order.

Zetanova:
I can not reproduce the lock localy.

Something with the UnitTest dispatcher and most likely the Extesions-Lazy<>.Value is locking somewhere/somehow.

I mark this one as WIP

Aaronontheweb:
I think it'd be worth writing up a spec on this first and reworking the problem from the top down that way

Zetanova:
If I could reproduce the behavior then we could write a spec for it.
Sry, that I abuse the build-server, all unit-tests are sucessful on my local machine

I think that I found now the issue the `addressPromise` TCS of `Remoting.StartAsync() ` leaked the dispatcher thread.

Zetanova:
@Aaronontheweb Would need help,
1) I don't know how to activate more log output 
2) Have no idea, to reproduce the deadlock locally with debug to find the death-lock

I removed basically all from the `Cluster.cotr`  and moved it into `InitializeAsync()` 
and the same with the `Transport.StartAsync()`

The call sequence didn't change and still there is some death lock, 
what I cannot find.

The secquence is `Transport.StartAsync()` and after right away `Cluster.Get(sys).InitializeAsync()` 
and nothing is accessing the Cluster extension before

I could even remove the `Cluster.ClusterCore` late init,
it is not required anymore and if it is not set just throws.

The only idea what i have left is inside `Cluster.Get(sys).InitializeAsync()`
``` 
var = await _clusterDaemons.Ask<IActorRef>(new InternalClusterAction.GetClusterCoreRef(this), 
                    System.Settings.CreationTimeout, cancellationToken).ConfigureAwait(false);
```
that is just simply death-locking instantly are the answer-mailbox is not processed

Zetanova:
@Aaronontheweb I moved the fixes out to a new PR and open a discussion for the startup here:
https://github.com/akkadotnet/akka.net/discussions/5447

Issue #5435, opened 2021-12-16
Possible race condition in Akka.Cluster


ismaelhamed:
**Version Information**
Akka 1.4.29
Akka.Cluster

**Describe the bug**
After upgrading to Akka v1.4.29, I've been hit with this a few times already during testing:

```
[akka://XXXXXX/system/cluster/$a#415417592]: Akka.Actor.ActorInitializationException: Exception during creation ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Akka.Cluster.OnMemberStatusChangedListener.PreStart()
   at Akka.Actor.ActorCell.UseThreadContext(Action action)
   at Akka.Actor.ActorCell.Create(Exception failure)
   --- End of inner exception stack trace ---
   at Akka.Actor.ActorCell.Create(Exception failure)
   at Akka.Actor.ActorCell.SysMsgInvokeAll(EarliestFirstSystemMessageList messages, Int32 currentState)
```
Looks like this only happens in nodes with a proxy to a sharding, or that's my observation so far. 

And then I can also see this in some of those nodes:

```
System.NotSupportedException: Nobody does not provide
   at Akka.Actor.Nobody.get_Provider()
   at Akka.Actor.Futures.Ask[T](ICanTell self, Func`2 messageFactory, Nullable`1 timeout, CancellationToken cancellationToken)
   at Akka.Actor.Futures.Ask[T](ICanTell self, Object message, Nullable`1 timeout, CancellationToken cancellationToken)
   at Akka.Actor.Futures.Ask[T](ICanTell self, Object message, Nullable`1 timeout)
```
Affected nodes won't recover from this on their own (although the cluster is fully formed and healthy).

I haven't found a way to reproduce it: just restarting nodes or the whole cluster during testing will eventually hit this.

Zetanova:
A view PR's since akka 1.4.21 removed view thread-locks from components and startup 
This made already existing race condition in the startup visible

I fixed already view of them in PR https://github.com/akkadotnet/akka.net/pull/5416
but it did not made it to 1.4.29 

@ismaelhamed Pls try your test with the nightly build.

Yes its very hard to reproduce them.
we already planning to rework the akka startup process

Aaronontheweb:
Looks like this is what introduced those issues https://github.com/akkadotnet/akka.net/pull/5398

Aaronontheweb:
[...]

PR #5437, opened 2021-12-17
Resolve Akka.Cluster startup

Aaronontheweb:
I think some of the race conditions that began to appear as a result of .NET 6's threadpool changes hid some of the faults exposed by this PR and some of the other Akka.Cluster ones. We need to carefully and quickly restore this functionality since https://github.com/akkadotnet/akka.net/issues/5435 is a critical issue.

### DIFF
--- a/src/core/Akka.Cluster.Tests.MultiNode/TransitionSpec.cs
+++ b/src/core/Akka.Cluster.Tests.MultiNode/TransitionSpec.cs
@@ -122,12 +122,12 @@ namespace Akka.Cluster.Tests.MultiNode
 
         private void LeaderActions()
         {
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
I've tested this on my machine using @Arkatufus' suggestion: https://github.com/akkadotnet/akka.net/issues/5435#issuecomment-996030510

@Arkatufus - are you able to reproduce the error from v1.4.29 on this PR?

Arkatufus:
I'll give it a go

Arkatufus:
Works just fine under a single core environment, all Akka.Cluster.Tools.Tests and Akka.Cluster.Tests specs passed.

Zetanova:
I am not finished, view things i want to improve

PR #5442, opened 2021-12-17
racy spec fix: `BackoffOnRestartSupervisor_must_respect_maxNrOfRetries_property_of_OneForOneStrategy`

Aaronontheweb:
issue was that using `AutoReset` could cause a race where if the scheduler could reset the count in-flight and throw off the estimates used by the tests. Using a `ManualReset` avoids this issue.


Aaronontheweb:
`Akka.Tests.Actor.ReceiveTimeoutSpec.An_actor_with_receive_timeout_must_get_timeout_while_receiving_NotInfluenceReceiveTimeout_messages` - this is the test I'm working on fixing next.

PR #5444, opened 2021-12-18
Custom frame size computation support in Framing

ismaelhamed:
Original issue: [#22129](https://github.com/akka/akka/issues/22129)

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/FramingSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/FramingSpec.cs
@@ -240,12 +240,20 @@ namespace Akka.Streams.Tests.Dsl
 
         private static readonly List<int> FieldOffsets = new List<int> {0, 1, 2, 3, 15, 16, 31, 32, 44, 107};
 
-        private static ByteString Encode(ByteString payload, int fieldOffset, int fieldLength, ByteOrder byteOrder)
+        private static ByteString Encode(ByteString payload, int fieldOffset, int fieldLength, ByteOrder byteOrder) =>
+            EncodeComplexFrame(payload, fieldLength, byteOrder, ByteString.FromBytes(new byte[fieldOffset]), ByteString.Empty);
+
+        private static ByteString EncodeComplexFrame(
+            ByteString payload, 
+            int fieldLength, 
+            ByteOrder byteOrder, 
+            ByteString offset, 
+            ByteString tail)
         {
             var h = ByteString.FromBytes(new byte[4].PutInt(payload.Count, order: byteOrder));
             var header = byteOrder == ByteOrder.LittleEndian ? h.Slice(0, fieldLength) : h.Slice(4 - fieldLength);
 
-            return ByteString.FromBytes(new byte[fieldOffset]) + header + payload;
+            return offset + header + payload + tail;
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


src/core/Akka.Streams.Tests/Dsl/FramingSpec.cs
```diff
@@ -240,43 +241,98 @@ private static string RandomString(int length)
...
+                        .Via(Rechunk)
+                        .Via(Framing.LengthField(fieldLength, int.MaxValue, fieldOffset, byteOrder))
+                        .Grouped(10000)
+                        .RunWith(Sink.First<IEnumerable<ByteString>>(), Materializer)
+                        .ContinueWith(t => (t.Result, encodedFrames, (byteOrder, fieldOffset, fieldLength)));
+                }
+            }
+
+            Parallel.ForEach(GetFutureResults(), async futureResult => 
+            {
```
ismaelhamed, line 285: Check this out, because introducing parallelism in tests that do not require to execute their asserts in order could save a lot of time. 

This commit brought the execution time of each of both tests from `1.2 min` to less than `2s`

ismaelhamed:
An unrelated test in `SqliteEventsByTagSpec` is making one of the validations to fail.

src/core/Akka.Streams.Tests/Dsl/FramingSpec.cs
```diff
@@ -384,6 +440,39 @@ public void Length_field_based_framing_must_fail_the_stage_on_negative_length_fi
...
         }
+        
+        [Fact]
+        public void Length_field_based_framing_must_ignore_length_field_value_when_provided_computeFrameSize()
+        {
+            int ComputeFrameSize(IReadOnlyList<byte> offset, int length) => 8;
+
+            var tempArray = new byte[4].PutInt(unchecked((int)0xFF010203), order: ByteOrder.LittleEndian);
+            Array.Resize(ref tempArray, 8);
+            var bs = ByteString.FromBytes(tempArray.PutInt(checked(0x04050607), order: ByteOrder.LittleEndian));
```
ismaelhamed, line 451: We're missing the [ByteStringBuilder](https://github.com/akka/akka/blob/d703a2afe0e0c6dc8f1c0321b0f4652600e2ae99/akka-actor/src/main/scala-2.12/akka/util/ByteString.scala#L1010) which would've made this chunk of code more readable

Aaronontheweb: I believe that was removed in 1.3 or 1.4 and I don't recall why

src/core/Akka.Streams.Tests/Dsl/FramingSpec.cs
Aaronontheweb: Oh wow, I wonder where else we could take advantage of this.

src/core/Akka.Streams.Tests/Dsl/FramingSpec.cs
```diff
@@ -240,43 +241,98 @@ private static string RandomString(int length)
...
+            EncodeComplexFrame(payload, fieldLength, byteOrder, ByteString.FromBytes(new byte[fieldOffset]), ByteString.Empty);
+
+        private static ByteString EncodeComplexFrame(
+            ByteString payload, 
+            int fieldLength, 
+            ByteOrder byteOrder, 
+            ByteString offset, 
+            ByteString tail)
         {
             var h = ByteString.FromBytes(new byte[4].PutInt(payload.Count, order: byteOrder));
```
Aaronontheweb, line 254: With the `ByteStringBuilder` I bet we could also save on allocations here too...

Aaronontheweb (approved):
LGTM

PR #5445, opened 2021-12-18
Unlock cluster startup

Zetanova:
View improvements and fixes that improves the akka remote and cluster startup 
found and copied from https://github.com/akkadotnet/akka.net/pull/5436

### DIFF
--- a/src/core/Akka.Remote/Remoting.cs
+++ b/src/core/Akka.Remote/Remoting.cs
@@ -196,17 +196,12 @@ namespace Akka.Remote
                     var akkaProtocolTransports = addressPromise.Task.Result;
                     if (akkaProtocolTransports.Count == 0)
                         throw new ConfigurationException(@"No transports enabled under ""akka.remote.enabled-transports""");
-                    _addresses = new HashSet<Address>(akkaProtocolTransports.Select(a => a.Address));
 
-                    IEnumerable<IGrouping<string, ProtocolTransportAddressPair>> tmp =
-                        akkaProtocolTransports.GroupBy(t => t.ProtocolTransport.SchemeIdentifier);
-                    _transportMapping = new Dictionary<string, HashSet<ProtocolTransportAddressPair>>();
-                    foreach (var g in tmp)
-                    {
-                        var set = new HashSet<ProtocolTransportAddressPair>(g);
-                        _transportMapping.Add(g.Key, set);
-                    }
+                    var transportMapping = new Dictionary<string, HashSet<ProtocolTransportAddressPair>>();
+                    foreach (var g in akkaProtocolTransports.GroupBy(t => t.ProtocolTransport.SchemeIdentifier))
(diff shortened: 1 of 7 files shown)


src/core/Akka.Remote/Remoting.cs
```diff
@@ -186,35 +186,29 @@ public override void Start()
...
                     // of this promise to include them.
-                    _endpointManager.Tell(new EndpointManager.Listen(addressPromise)); 
+                    _endpointManager.Tell(new EndpointManager.Listen(addressPromise));
 
                     addressPromise.Task.Wait(Provider.RemoteSettings.StartupTimeout);
                     var akkaProtocolTransports = addressPromise.Task.Result;
-                    if(akkaProtocolTransports.Count==0)
+                    if (akkaProtocolTransports.Count == 0)
                         throw new ConfigurationException(@"No transports enabled under ""akka.remote.enabled-transports""");
-                    _addresses = new HashSet<Address>(akkaProtocolTransports.Select(a => a.Address));
```
to11mtm, line 199: Pardon my ignorance, but where does this get set if not here?

src/core/Akka.Remote/Remoting.cs
```diff
@@ -186,35 +186,29 @@ public override void Start()
...
-                        akkaProtocolTransports.GroupBy(t => t.ProtocolTransport.SchemeIdentifier);
-                    _transportMapping = new Dictionary<string, HashSet<ProtocolTransportAddressPair>>();
-                    foreach (var g in tmp)
-                    {
-                        var set = new HashSet<ProtocolTransportAddressPair>(g);
-                        _transportMapping.Add(g.Key, set);
-                    }
+                    var transportMapping = new Dictionary<string, HashSet<ProtocolTransportAddressPair>>();
+                    foreach (var g in akkaProtocolTransports.GroupBy(t => t.ProtocolTransport.SchemeIdentifier))
+                        transportMapping.Add(g.Key, new HashSet<ProtocolTransportAddressPair>(g));
```
to11mtm, line 202: Nice simplification. Should we .As/ToDictionary() to go further?

to11mtm (commented):
Left some comments.

BTW, Stupid question; shoudld we have TaskCompletionOptions.RunContinuationsAsync on TCS created by Join?

src/core/Akka.Remote/Remoting.cs
Zetanova: 3 lines below

src/core/Akka.Remote/Remoting.cs
Zetanova: will try the key point was that it set the dictionary and filled it up and not set the filled directory to the public field

Zetanova:
I didn't got this function and/or the Cluster startup to run in a async method (see other PR)
will now remove the volatile fields from Remoting Transport.

The `TaskCompletionOptions.RunContinuationsAsync` i sent in the endpoint manager, 
because it's necessary when `Transport.Start()` would be converted to async 
or else the dispatcher thread will leak

PR #5449, opened 2021-12-20
Async file sink

Zetanova:
fix some akka io specs and make file sink use async IO

### DIFF
--- a/src/core/Akka.Streams/Implementation/IO/FileSubscriber.cs
+++ b/src/core/Akka.Streams/Implementation/IO/FileSubscriber.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Event;
@@ -54,13 +55,17 @@ namespace Akka.Streams.Implementation.IO
 
         private readonly FileInfo _f;
         private readonly TaskCompletionSource<IOResult> _completionPromise;
+        private readonly CancellationTokenSource _cts;
         private readonly long _startPosition;
         private readonly FileMode _fileMode;
         private readonly ILoggingAdapter _log;
-        private readonly WatermarkRequestStrategy _requestStrategy;
+        private readonly FileStreamRequestStrategy _requestStrategy;
         private readonly bool _autoFlush;
         private FileStream _chan;
         private long _bytesWritten;
+        private readonly int _fileBufferSize;
+
+        private const int DefaultFileBufferSize = 4 * 1024; //regular file buffer size
 
         /// <summary>
         /// TBD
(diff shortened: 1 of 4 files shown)


Zetanova:
this PR is finished

Will open a new one to hunt racy specs

src/core/Akka.Streams/Implementation/IO/FileSubscriber.cs
```diff
@@ -105,113 +117,158 @@ protected override void PreStart()
...
+        sealed class FileStreamRequestStrategy : MaxInFlightRequestStrategy
+        {
+            public FileStreamRequestStrategy(int max) : base(max)
+            {
+            }
+
+            /// <summary>
+            /// How many complete file stream buffers are queued
+            /// </summary>
+            public int FileBufferCount { get; set; }
```
Arkatufus, line 268: Should this be made immutable?

src/core/Akka.Streams/Implementation/IO/FileSubscriber.cs
```diff
@@ -105,113 +117,158 @@ protected override void PreStart()
...
-                            _chan.Flush(true);
+                        var bytes = (ByteString)next.Element;
+                        if (_bufferedByteString.Count > 0)
+                        {
+                            bytes = _bufferedByteString.Concat(bytes);
+                            _bufferedByteString = ByteString.Empty;
+                            _requestStrategy.FileBufferCount = 0;
+                        }
+                        _streamTask = WriteAsync(Self, bytes, _autoFlush);
+                        _fileRequests++;
```
Arkatufus, line 148: Is it possible that `_fileRequest` be greater than 1? If not, is it better to change this into a `bool` instead, or just check and see if `_streamTask` is null, since `_streamTask` has to be assigned if a write operation is currently in progress?

Arkatufus (changes requested):
Have a few question regarding the code, mainly the use of `_fileRequests` and `_streamTask`

src/core/Akka.Streams/Implementation/IO/FileSubscriber.cs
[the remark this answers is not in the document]
Zetanova: `FileBufferCount` is used for the backpressure feature, it need to be mutable
default are 16 max, 5 min and file-buffer default 4KB 
Meaning that it will request until buffer is 64KB big

src/core/Akka.Streams/Implementation/IO/FileSubscriber.cs
Zetanova: `_streamTask` is not used (maybe for debug), because `_streamTask.IsCompleted` is not writable by the actor itself,
_fileRequests can only be 0 or 1 for this actor, but this pattern can be used for other things
its very good for fan-out / fad-in to async operations

Arkatufus: But does fan-out make any sense for a file sink? This should be a sink that does a write to a single file, and not multiple files.

Zetanova: Should i convert `int _fileRequests` to `boolean _fileRequested` ?

Arkatufus: I think so, making it a bool would make the intention more clear. Or we can use the _streamTask as a flag to check if a write is progress or not.

Zetanova: `_streamTask.IsCompleted` can not be used, because it would pre-indicate inflight completion messages.
I will change the state field to a bool.

Arkatufus: Well, you can always check to see if `_streamTask` is null or not, set it to null whenever a write operation is completed.

Issue #5453, opened 2021-12-20
Build system: need to automatically check API changes against plugins


Aaronontheweb:
# Problem
With the release of Akka.NET v1.4.30, we've inadvertently broken existing Akka.DI plugin support as a result of merging in https://github.com/akkadotnet/akka.net/pull/5416/files#r767775286 - and specifically I reviewed this section and thought that level of API change was fine, attaining to our "[Practical SemVer](https://aaronstannard.com/oss-semver/)" standard. Any actively used Akka.DI plugins can be updated to 1.4.30 and this issue will be fixed on their end, but we shouldn't tolerate surprises like this. It's beneath our quality standard.

Akka.NET project members are more interested in making changes to older APIs in order to create improvements, take advantage of newer .NET features like `Span<T>` and so forth. Those are all well and good, but we don't support the ability to do that sufficiently safely right now without applying a ridiculous standard like strict SemVer.

> A separate but related issue, [explicitly documenting Akka.NET's standards and processes for contributors](https://github.com/akkadotnet/akka.net/projects/7), has been on my to-do list for a while and I'm scheduled to complete it by December 31s.

## Solution
What I think we need is some more rigorous compatibility testing against a set of downstream components to ensure that when APIs do get changed they are:

1. Wire compatible
2. Binary compatible
3. Behaviorally compatible

How I propose we do this:

![image](https://user-images.githubusercontent.com/326939/146843961-f07ee241-d4b7-4768-a75d-be5aff251328.png)

First and foremost, we should never do this automatically on every pull request as it'll be prohibitively time-expensive. But when a member of @akkadotnet/core sees a significant change to an API approval or a Protobuf definition, they should always manually kick off one of these jobs via a command on the PR thread.

We'll have to design some specifications to actually implement this. Conceptually, they're easy.

### API Compat Specs
1. Code against current minor version of Akka.NET;
2. Take explicit dependencies on last stable version of some plugin (Akka.Logger.Serilog, Alpakka, Akka.Management) - this step is mandatory. Without taking a downstream dependency we can't test for upstream binary-compatibility breakages.
3. Upgrade Akka.NET to PR'd NuGet packages;
4. Run a set of integration tests orchestrated using XUnit, Docker.NET et al.
5. Transform the pass / fail / skip output into a Github comment for the full set of API compat specs.

### Wire Format Compat Specs
1. Build a set of stock docker images against each stable release of Akka.NET, labeled with that version;
[...]

PR #5454, opened 2021-12-20
Revert #5416, broke DI backward compatibility

Arkatufus:


Aaronontheweb (approved):
LGTM

Aaronontheweb:
I liked #5416 on balance, but due to the issues I identified in https://github.com/akkadotnet/akka.net/issues/5453 we can't let this hang out there in the open. Broke every Akka.DI plugin. Even though we're deprecating those we haven't communicated that to end users or set an explicit deadline. Can't break things by accident in a revision release like that.

Zetanova:
What exactly break and why do we revert racy spec and bug fixes too?

If its only the IActorProducer interface we could change it back and use a better binary compatible version/pattern

Aaronontheweb:
We're reverting the entire PR because it was the most expedient way to fix the error. We can't tolerate a surprise breaking change like that affecting user's applications with no warning.

I would love to reintroduce your changes but they need to be done in smaller chunks; going forward I'm going to be more aggressive about rejecting PRs that get too big and touch too many things. Shouldn't be racy spec fixes and other bug fixes on that same PR that touches core API changes _if those changes are unrelated_. Could you please reintroduce those in smaller PRs?

As I mentioned here https://github.com/akkadotnet/akka.net/issues/5453 a few moments ago, I'm going to document our standards explicitly going forward. We have to.

Aaronontheweb:
For the record, this breaking change and the subsequent revert is my fault - I haven't helped us build an API change management system that is more resilient, I haven't documented the standards we need to make it easy to contribute to the project, I approved the PR. I'm going to bump that up to the top of my list for Jan 2022.

Aaronontheweb:
I'm sorry for making the process of contributing to Akka.NET more frustrating that necessary for you @Zetanova - that's my fault too. I'll put a dent in https://github.com/akkadotnet/akka.net/projects/7 and update `CONTRIBUTORS.md` and our pull request templates. I need to do that before a new full time engineer for Petabridge starts in January anyway.

Zetanova:
We expected that it will not binary compatible with Akka.DI, you wanted to test it and then it was merged.
I suspected that nobody is using Akka.DI 

I will copy the fixes out, but not this week. 

What we need is also a good system to mark things obsolete and to know when they can be deleted.
There still so many functions marked "obsolete" from version 1.0.0 (2014-2015)

Aaronontheweb:
> We expected that it will not binary compatible with Akka.DI, you wanted to test it and then it was merged.
I suspected that nobody is using Akka.DI

You're absolutely right and that's on me.

Aaronontheweb:
> What we need is also a good system to mark things obsolete and to know when they can be deleted.
There still so many functions marked "obsolete" from version 1.0.0 (2014-2015)

Agreed - we usually do this between minor versions, but again - I've not documented that.

to11mtm:
> We expected that it will not binary compatible with Akka.DI, you wanted to test it and then it was merged. I suspected that nobody is using Akka.DI
> 
> I will copy the fixes out, but not this week.
> 
> What we need is also a good system to mark things obsolete and to know when they can be deleted. There still so many functions marked "obsolete" from version 1.0.0 (2014-2015)

Ironically I -do- still use Akka.DI, although this hasn't yet impacted us yet as we are a little more conservative on non-critical package upgrades.

PR #5455, opened 2021-12-20
Update RELEASE_NOTES.md for 1.4.31 release

Arkatufus:


RELEASE_NOTES.md
```diff
@@ -1,9 +1,16 @@
-#### 1.4.30 December 13 2021 ####
+#### 1.4.31 December 20 2021 ####
+Akka.NET v1.4.30 is a minor release that contains some bug fixes.
+
+Bug fixes:
+* [Akka: Reverted Props code refactor](https://github.com/akkadotnet/akka.net/pull/5454)
```
Aaronontheweb, line 5: Need to add some notes on here as to why this was reverted.

"Akka.NET v1.4.30 contained a breaking change that broke binary compatibility with all Akka.DI plugins. Even though those plugins are deprecated that change is not compatible with our SemVer standards and needed to be reverted. We regret the error."

Arkatufus: Done

PR #5458, opened 2021-12-24
Update Akka.MultiNode.TestAdapter to 1.1.0-beta2

Arkatufus:


Arkatufus:
This PR looks bad, but its really is not. There are a lot of one liner change needed because we now have 2 `MultiNodeFactAttribute`, one in the old MNTR implementation and one in the standalone version.

src/contrib/cluster/Akka.Cluster.Metrics.Tests.MultiNode/ClusterMetricsExtensionSpec.cs
```diff
@@ -17,6 +17,7 @@
 using Akka.Configuration;
 using FluentAssertions.Extensions;
 using ConfigurationFactory = Akka.Configuration.ConfigurationFactory;
+using MultiNodeFactAttribute = Akka.MultiNode.TestAdapter.MultiNodeFactAttribute;
```
Arkatufus, line 20: Need to add this using on all of our spec file because we now have 2 `MultiNodeFactAttribute` class, one in the old implementation and one in the new standalone implementation.

src/contrib/cluster/Akka.Cluster.Metrics.Tests.MultiNode/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,10 @@
+﻿// //-----------------------------------------------------------------------
+// // <copyright file="AssemblyInfo.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2021 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2021 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using Xunit;
+
+[assembly: TestFramework("Akka.MultiNode.TestAdapter.MultiNodeTestFramework", "Akka.MultiNode.TestAdapter")]
```
Arkatufus, line 10: This is the one liner that needs to be added to make sure that all classes in the assembly is grouped in a single collection and parallelization is turned off/ignored.

src/core/Akka.Cluster.Tests.MultiNode/StressSpec.cs
```diff
@@ -46,71 +47,74 @@ public StressSpecConfig()
...
-      # scale the *-duration settings with this factor
-      duration-factor = 1
-      join-remove-duration = 90s
-      idle-gossip-duration = 10s
-      expected-test-duration = 600s
-      # scale convergence within timeouts with this factor
-      convergence-within-factor = 1.0
-    }
-    akka.actor.provider = cluster
+akka.test.cluster-stress-spec {
```
Arkatufus, line 50: Reformat the config so that it is a lot easier to read

src/core/Akka.Cluster.Tests.MultiNode/StressSpec.cs
```diff
@@ -46,71 +47,74 @@ public StressSpecConfig()
...
-              }
-            }
+    publish-stats-interval = 1s
+}
+
+akka.loggers = [""Akka.TestKit.TestEventListener, Akka.TestKit""]
+akka.loglevel = INFO
+akka.remote.log-remote-lifecycle-events = off
+akka.actor.default-dispatcher = {
+    executor = fork-join-executor
```
Arkatufus, line 93: Changed the executor from `channel-executor` to `fork-join-executor`, this spec is not configured properly for the channel executor.

Aaronontheweb:
> This PR looks bad, but its really is not. There are a lot of one liner change needed because we now have 2 `MultiNodeFactAttribute`, one in the old MNTR implementation and one in the standalone version.

Remind me again - what did we decide we were going to do about the old attribute? Mark it as Obsolete with a warning telling them to change it over to a new namespace + a link to an article about upgrading?

Arkatufus:
Marking them obsolete would be an option, i think we said that we'll drop the old MNTR and just remove it.

Aaronontheweb:
> Marking them obsolete would be an option, I think we said that we'll drop the old MNTR and just remove it.

I think we should mark them as obsolete for 1.4 and remove in 1.5.

Arkatufus:
Done. Marked it as obsolete.

src/contrib/cluster/Akka.Cluster.Metrics.Tests.MultiNode/Properties/AssemblyInfo.cs
Aaronontheweb: Does this need to be added to all projects? I.e. is this part of the installation instructions for the new MNTR?

src/core/Akka.Remote.TestKit/MultiNodeFact.cs
```diff
@@ -10,6 +10,10 @@
 
 namespace Akka.Remote.TestKit
 {
+    [Obsolete(
+        "This attribute and the Akka.MultiNodeTestRunner NuGet package is being deprecated in favor of the new " +
```
Aaronontheweb, line 14: LGTM

Aaronontheweb (commented):
Generally LGTM but left one comment for you

src/contrib/cluster/Akka.Cluster.Metrics.Tests.MultiNode/Properties/AssemblyInfo.cs
Arkatufus: Ah you're right, this needs to be added to the migration/usage guide.

Aaronontheweb: No problem - we _might_ be able to automate the inclusion of this through a `.props` file that we can include inside the MNTR packages but for the time being let's add this to the instructions and explain why it's helpful.

PR #5497, opened 2022-01-10
code cleanup on Akka.Persistence

Aaronontheweb:
## Changes

Cleaning up some of the `Eventsourced` and `AsyncWriteJournal` code to use newer C# features - should help improve with compiler optimization in some small areas, but mostly serves to help improve readability.

### DIFF
--- a/src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
+++ b/src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
@@ -309,8 +309,7 @@ namespace Akka.Persistence.Journal
         /// <returns>TBD</returns>
         protected Exception TryUnwrapException(Exception e)
         {
-            var aggregateException = e as AggregateException;
-            if (aggregateException != null)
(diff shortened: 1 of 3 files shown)


src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -354,21 +353,21 @@ private void HandleWriteMessages(WriteMessages message)
                 writeResult = Task.FromResult((IImmutableList<Exception>)Enumerable.Repeat(e, atomicWriteCount).ToImmutableList());
             }
 
-            Action<Func<IPersistentRepresentation, Exception, object>, IImmutableList<Exception>> resequence = (mapper, results) =>
+            void Resequence(Func<IPersistentRepresentation, Exception, object> mapper, IImmutableList<Exception> results)
```
Aaronontheweb, line 356: Using a local function instead of explicit delegate allocation

Arkatufus (approved):
LGTM

PR #5505, opened 2022-01-13
cleaning up `async Task` internals inside `AsyncWriteJournal`

Aaronontheweb:
Fixes https://github.com/petabridge/phobos-issues/issues/31 and https://github.com/petabridge/phobos-issues/issues/30

## Changes
This PR is a work in progress. I'm attempting to clean up the internals of the Akka.Persistence `Journal` and `SnapshotStore` for three reasons:

1. To make Akka.Persistence traceable on both recovery and persistence, per the two Phobos issues linked above.
2. To modernize the way we work with asynchronous programming in C#, similar to what we've done with `Ask<T>` and `PipeTo`. See https://github.com/akkadotnet/akka.net/pull/5361 for a clear example of this.
3. To improve per-operation performance by reducing delegate allocations, collection allocations, and simplifying some of the async workflows.

**This is a work in progress**

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).

### Latest `dev` Benchmarks 

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19041.1415 (2004/May2020Update/20H1)
AMD Ryzen 7 1700, 1 CPU, 16 logical and 8 physical cores
.NET SDK=6.0.100
  [Host]     : .NET 6.0.0 (6.0.21.52210), X64 RyuJIT
  Job-WKWXNT : .NET 6.0.0 (6.0.21.52210), X64 RyuJIT

InvocationCount=1  UnrollFactor=1  

```
|             Method | PersistentActors | WriteMsgCount |       Mean |     Error |    StdDev |     Median |      Gen 0 |     Gen 1 | Allocated |
|------------------- |----------------- |-------------- |-----------:|----------:|----------:|-----------:|-----------:|----------:|----------:|
| **WriteToPersistence** |                **1** |           **100** |   **2.380 ms** | **0.1841 ms** | **0.5399 ms** |   **2.058 ms** |          **-** |         **-** |    **631 KB** |
[...]

PR #5511, opened 2022-01-13
[WIP][PERF] `SqlJournal` perf improvements

Aaronontheweb:
Fixes #

## Changes

Pairs with https://github.com/akkadotnet/akka.net/pull/5505 - working on improving some inefficiencies inside the `SqlJournal`, namely an absolute ton of unnecessary allocations.

## Checklist

### Latest `dev` Benchmarks 

### This PR's Benchmarks


src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -315,9 +349,40 @@ protected Exception TryUnwrapException(Exception e)
...
             return e;
         }
 
+        // ValueDelegate implementation for Func<IPersistentRepresentation, Exception, object> mapper
+        private interface IMapperDelegate
+        {
+            object Invoke(WriteMessages message, IPersistentRepresentation rep, Exception ex);
+        }
+
+        private sealed class SuccessDelegate : IMapperDelegate
```
to11mtm, line 362: To Fully ValueDelegate these, change to a struct and pass in `deault(SuccessDelegate)` instead of `SuccessDelegate.Instance`.

May want to test both ways, since that's where we'll be doing the Virt call cost vs code size tradeoff

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
```diff
@@ -135,8 +135,7 @@ protected override bool ReceivePluginInternal(object message)
         /// <returns>TBD</returns>
         protected override async Task<IImmutableList<Exception>> WriteMessagesAsync(IEnumerable<AtomicWrite> messages)
```
to11mtm, line 136: so, this is out of left field, but should we consider using `ToImmutableArray` instead of [`ToImmutableList` in PreparePersistentBatch](https://github.com/akkadotnet/akka.net/blob/6ee8b4fd86fc299445dddfd7a2daba3db69b35df/src/core/Akka.Persistence/Journal/WriteJournal.cs#L37-L43) and consider doing so in other places of persistence where we are creating Read-only immutable lists?

Reason being, `ImmutableArray<T>` does implement `IImmutableList<T>`, but uses an array rather than AVL Trees. Might help on locality for PersistAll (and a tiny bit on other things)

to11mtm (commented):
Couple comments. May have more later.

PR #5514, opened 2022-01-14
[Cluster] Switch Persistence benchmark from using SQLite to MemoryJournal

Arkatufus:
Switch Akka.Cluster.Benchmark persistence benchmark from using SQLite to MemoryJournal to minimize external variable


Aaronontheweb (approved):
LGTM

Issue #5498, opened 2022-01-11
Error when using ChannelTaskScheduler for internal-dispatcher in akka cluster node


SchiessMax:
**Version Information**
Version of Akka.NET? 1.4.31
Which Akka.NET Modules? Akka.Actor/Akka.Cluster

**Describe the bug**
When trying to use the ChannelTaskScheduler as akka.actor.internal-dispatcher an exception occurs (see stack trace below). I made the configuration according to https://github.com/akkadotnet/akka.net/pull/5403.
The exception does not occur when the ChannelTaskScheduler is used for the other dispatchers.
Please let me know if you need more information.

Stack trace:

```
Exception during creation
 ---> System.TypeLoadException: Error while creating actor instance of type Akka.Cluster.ClusterHeartbeatSender with 0 args: ()
 ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.InvalidOperationException: ValueFactory attempted to access the Value property of this instance.
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Akka.Actor.Internal.ActorSystemImpl.TryGetExtension[T](T& extension)
   at Akka.Actor.Internal.ActorSystemImpl.GetExtension[T]()
   at Akka.Actor.ActorSystemWithExtensions.WithExtension[T,TI](ActorSystem system)
   at Akka.Cluster.Cluster.Get(ActorSystem system)
   at Akka.Cluster.ClusterHeartbeatSender..ctor()
   --- End of inner exception stack trace ---
   at System.RuntimeTypeHandle.CreateInstance(RuntimeType type, Boolean publicOnly, Boolean wrapExceptions, Boolean& canBeCached, RuntimeMethodHandleInternal& ctor, Boolean& hasNoDefaultCtor)
   at System.RuntimeType.CreateInstanceDefaultCtorSlow(Boolean publicOnly, Boolean wrapExceptions, Boolean fillCache)
   at System.RuntimeType.CreateInstanceDefaultCtor(Boolean publicOnly, Boolean skipCheckThis, Boolean fillCache, Boolean wrapExceptions)
   at System.Activator.CreateInstance(Type type, Boolean nonPublic, Boolean wrapExceptions)
[...]

PR #5515, opened 2022-01-15
Cluster startup hard block

Zetanova:
Fixes https://github.com/akkadotnet/akka.net/issues/5498
Fix startup with missing config for channel-executor

## Changes

The constructor thread of cluster startup got reused by the default TaskScheduler used by the ChannelExecutor.

### DIFF
--- a/src/core/Akka.Cluster/Cluster.cs
+++ b/src/core/Akka.Cluster/Cluster.cs
@@ -140,7 +140,11 @@ namespace Akka.Cluster
             _readView = new ClusterReadView(this);
 
             // force the underlying system to start
-            _clusterCore = GetClusterCoreRef().Result;
+            // and hard block the current thread
+            var clusterCoreTask = Task.Run(GetClusterCoreRef);
+            if (!clusterCoreTask.Wait(System.Settings.CreationTimeout))
+                throw new TimeoutException("cluster startup");
+            _clusterCore = clusterCoreTask.Result;
 
             system.RegisterOnTermination(Shutdown);
 
--- a/src/core/Akka/Dispatch/AbstractDispatcher.cs
+++ b/src/core/Akka/Dispatch/AbstractDispatcher.cs
@@ -6,7 +6,6 @@
 //-----------------------------------------------------------------------
 
 using System;
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
@@ -59,9 +58,9 @@ namespace Akka.Dispatch
         /// <param name="settings">TBD</param>
         /// <param name="mailboxes">TBD</param>
         public DefaultDispatcherPrerequisites(
-            EventStream eventStream, 
-            IScheduler scheduler, 
-            Settings settings, 
+            EventStream eventStream,
+            IScheduler scheduler,
+            Settings settings,
             Mailboxes mailboxes)
         {
             Mailboxes = mailboxes;
@@ -120,8 +119,26 @@ namespace Akka.Dispatch
     {
         public ChannelExecutorConfigurator(Config config, IDispatcherPrerequisites prerequisites) : base(config, prerequisites)
         {
-            var cfg = config.GetConfig("channel-executor");
-            Priority = (TaskSchedulerPriority)Enum.Parse(typeof(TaskSchedulerPriority), cfg.GetString("priority", "normal"), true);
+            var priorityName = config.GetString("channel-executor.priority", "None") ?? "None";
+            Priority = (TaskSchedulerPriority)Enum.Parse(typeof(TaskSchedulerPriority), priorityName, true);
+            if (Priority == TaskSchedulerPriority.None)
+            {
+                var dispatcherName = config.Root.GetObject().GetKey("name").ToString();
+                switch (dispatcherName)
+                {
+                    case "internal-dispatcher":
+                    case "default-remote-dispatcher":
+                        Priority = TaskSchedulerPriority.High;
+                        break;
+                    case "backoff-remote-dispatcher":
+                        Priority = TaskSchedulerPriority.Low;
+                        break;
+                    default:
+                        Priority = TaskSchedulerPriority.Normal;
+                        break;
+                };
+            }
+                
         }
 
         public TaskSchedulerPriority Priority { get; }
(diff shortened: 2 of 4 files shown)


src/core/Akka.Cluster/Cluster.cs
```diff
@@ -140,7 +140,11 @@ public Cluster(ActorSystemImpl system)
             _readView = new ClusterReadView(this);
 
             // force the underlying system to start
-            _clusterCore = GetClusterCoreRef().Result;
+            // and hard block the current thread
+            var clusterCoreTask = Task.Run(GetClusterCoreRef);
```
to11mtm, line 144: Are there any flags we should consider adding here? i.e. would it be better to do `Task.Factory.StartNew` with `LongRunning` and `DenyChildAttach`?

Zetanova: `_clusterCore = GetClusterCoreRef().Result` was always locking and I think its still is.

I think what happend is that after the Ask got `improved` the current thread gets used for the actorcell dispatcher.
I don't know how/where exactly, but the child-actors of cluster-daemon are using it.

´Task.Run(GetClusterCoreRef)´ should force the ask on a different thread, 
this is the only thing that matters, that it the current thread does not leak.

to11mtm: > I think what happend is that after the Ask got improved the current thread gets used for the actorcell dispatcher.

Interesting. Would there be other implications if this is true?

cc: @Aaronontheweb

Zetanova: The actor system extensions implementation and the locking cluster constructor are both just anti-patterns 
and we need to rework it in the future.
https://github.com/akkadotnet/akka.net/discussions/5447

Zetanova: Maybe its just the "ConfigureAwait(false)" of 
https://github.com/akkadotnet/akka.net/blob/44c51f6786df0f97914381bd1c8b7f84013f7c0e/src/core/Akka.Cluster/Cluster.cs#L155

The Cluster Extension is called by the ClusterActorRefProvider and it is created in the ActorSystem Startup.
The thread is the same as of the ActorSystem creator and with the normal ForkJoinExecutor they will never mix.
But with the ChannelExecutor that uses the normal ThreadPool of dotnet, the awaiting thread can be used for an ActorCell. 

``` 
var task = Task.Run(...);
task.Wait();
``` 
I hope that simply resolve the problem.

`task.Result` blocked at .net 4.5 for sure, 
I don't know why/how the thread gets now reused.

Aaronontheweb:
I have not had time to review this - hopefully I'll be able to come up for air next week.

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -120,8 +119,26 @@ internal sealed class ChannelExecutorConfigurator : ExecutorServiceConfigurator
...
+            var priorityName = config.GetString("channel-executor.priority", "None") ?? "None";
+            Priority = (TaskSchedulerPriority)Enum.Parse(typeof(TaskSchedulerPriority), priorityName, true);
+            if (Priority == TaskSchedulerPriority.None)
+            {
+                var dispatcherName = config.Root.GetObject().GetKey("name").ToString();
+                switch (dispatcherName)
+                {
+                    case "internal-dispatcher":
+                    case "default-remote-dispatcher":
+                        Priority = TaskSchedulerPriority.High;
```
Arkatufus, line 131: These default values should be baked in the default HOCON config, not in code; i.e. they have to be transparent to the user, user should only read the config file to figure out the default value of a setting, not dig through the source code.

src/core/Akka/Dispatch/AbstractDispatcher.cs
```diff
@@ -120,8 +119,26 @@ internal sealed class ChannelExecutorConfigurator : ExecutorServiceConfigurator
     {
         public ChannelExecutorConfigurator(Config config, IDispatcherPrerequisites prerequisites) : base(config, prerequisites)
         {
-            var cfg = config.GetConfig("channel-executor");
-            Priority = (TaskSchedulerPriority)Enum.Parse(typeof(TaskSchedulerPriority), cfg.GetString("priority", "normal"), true);
+            var priorityName = config.GetString("channel-executor.priority", "None") ?? "None";
```
Arkatufus, line 122: Unless you're accessing `ActorSystem.Settings.Config` directly, always assume that a Config object passed into a method can be null. Need to check for null config instance here.

Arkatufus, line 122: These changes does not have anything to do with making cluster startup to block, is it? Can you remove these and move it to a new PR?

src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
```diff
@@ -62,14 +62,17 @@ public ChannelTaskScheduler(ExtendedActorSystem system)
             //config channel-scheduler
             var config = system.Settings.Config.GetConfig("akka.channel-scheduler");
             _maximumConcurrencyLevel = ThreadPoolConfig.ScaledPoolSize(
-                        config.GetInt("parallelism-min"),
-                        config.GetDouble("parallelism-factor", 1.0D), // the scalar-based factor to scale the threadpool size to 
-                        config.GetInt("parallelism-max"));
+                        config?.GetInt("parallelism-min", 4) ?? 4,
```
Arkatufus, line 65: Same thing here, these should be placed in a different PR

Arkatufus (changes requested):

src/core/Akka/Dispatch/AbstractDispatcher.cs
Zetanova: they are the default in the config.
But the testkit does not load the default config

src/core/Akka/Dispatch/AbstractDispatcher.cs
Zetanova: this is required to use the channel-executor in the spec.

I added one spec that simple starts the cluster with the channel-executor and discovered this problems.

The spec itself will not trigger the original failure, because the testkit is using a SynchronisationContext
for the akka system startup

Arkatufus: Yes, but this PR can easily be broken into 2 PR, one depending to the other.

Arkatufus: This should not be a problem anymore, fixed in #5568

PR #5516, opened 2022-01-17
Add custom persistence provider documentation

### LINKED ISSUE #5493: Add documentation on making custom Akka.Persistence provider
Need to add documentation on how to create a custom Akka.Persistence provider for other database storage system such as RavenDB. This should document how to create custom snapshot, journal, and query.

Arkatufus:
Fixes #5493

## Changes

- [x] Adds custom persistence provider page to the documentation
- [x] Add sample project for code linking in documentation page


docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
+# Writing A Custom Akka.Persistence Provider
```
eaba, line 1: Missing uid:
```
---
uid: custom-persistent-provider
title: Custom Persistent Provider
---
````

File: docs/articles/persistence/custom-persistent-provider.md [hunk 2]
```diff
@@ -0,0 +1,511 @@
...
+## Implementing Akka.Persistence Journal
+
+- Discuss design goals
+- Discuss Journal responsibility
+
+Akka.Persistence journal is responsible for storing all events for playback in the event of a recovery.
+
+All of the code examples in this documentation will assume a Sqlite database and the `Journal` table schema we will be using are:
+
+```sqlite
```
eaba, line 12: `[!code-sql[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?range=29-38,41-44)]`?

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+    persistence_id VARCHAR(255) NOT NULL,
+    sequence_nr INTEGER(8) NOT NULL,
+    PRIMARY KEY (persistence_id, sequence_nr));
+```
+
+### Required Method Implementations
+
+#### ReplayMessagesAsync
+
+```c#
```
eaba, line 34: `[!code-csharp[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L169-175)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+- The returned `Task` must be completed when all messages (matching the sequence number bounds) have been replayed.
+- The `Task` must be completed with a failure if any of the persistent messages could not be replayed.
+
+The `toSequenceNr` is the lowest of what was returned by `ReadHighestSequenceNrAsync` and what the user specified as the recovery `Recovery` parameter. This does imply that this call is always preceded by reading the highest sequence number for the given `persistenceId`.
+
+This call is NOT protected with a circuit-breaker because it may take a long time to replay all events. The plugin implementation itself must protect against an unresponsive backend store and make sure that the returned `Task` is completed with success or failure within reasonable time. It is not allowed to ignore completing the `Task`.
+
+`ByPersistenceIdSql` in this example refers to this SQL query statement:
+
+```sqlite
```
eaba, line 63: `[!code-sql[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L47-57)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+        e.payload as Payload,
+        e.serializer_id as SerializerId
+    FROM event_journal e
+    WHERE e.persistence_id = @PersistenceId
+    AND e.sequence_nr BETWEEN @FromSequenceNr AND @ToSequenceNr
+    ORDER BY sequence_nr ASC;
+```
+
+Semi pseudo-code example:
+```c#
```
eaba, line 79: `[!code-csharp[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L169-222)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+                command.Cancel();
+            }
+        }
+    }
+}
+```
+
+#### ReadHighestSequenceNrAsync
+
+```
```
eaba: `[!code-csharp[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L224)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+
+This method should asynchronously reads the highest stored sequence number for provided `persistenceId`. The persistent actor will use the highest sequence number after recovery as the starting point when persisting new events. This sequence number is also used as `toSequenceNr` in subsequent calls to `ReplayMessagesAsync` unless the user has specified a lower `toSequenceNr`. Journal must maintain the highest sequence number and never decrease it.
+
+This call is protected with a circuit-breaker.
+
+Please also note that requests for the highest sequence number may be made concurrently to writes executing for the same `persistenceId`, in particular it is possible that a restarting actor tries to recover before its outstanding writes have completed.
+
+`HighestSequenceNrSql` in this example refers to this SQL query statement:
+
+```sqlite
```
eaba, line 156: Did not find this in the source. `sql` instead of `sqlite`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+SELECT MAX(e.ordering) as Ordering
+FROM event_journal e;
+```
+
+```c#
```
eaba, line 161: `[!code-csharp[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L224-244)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+            // Return the result if one is returned by the query result, else return zero
+            return result is long ? Convert.ToInt64(result) : 0L;
+        }
+    }
+}
+```
+
+#### WriteMessagesAsync
+
+```
```
eaba: `[!code-csharp[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L246)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+
+Please note that the `IPersistentRepresentation.Sender` of the contained `Persistent` objects has been nulled out (i.e. set to `ActorRefs.NoSender` in order to not use space in the journal for a sender reference that will likely be obsolete during replay.
+
+Please also note that requests for the highest sequence number may be made concurrently to this call executing for the same `persistenceId`, in particular it is possible that a restarting actor tries to recover before its outstanding writes have completed. In the latter case it is highly desirable to defer reading the highest sequence number until all outstanding writes have completed, otherwise the `PersistentActor` may reuse sequence numbers.
+
+This call is protected with a circuit-breaker.
+
+`InsertEventSql` in this example refers to this SQL query statement:
+
+```sqlite
```
eaba, line 219: `[!code-sql[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L67-82)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+    @Timestamp,
+    @IsDeleted,
+    @Manifest,
+    @Payload,
+    @SerializerId
+);
+```
+
+Semi pseudo-code example:
+```c#
```
eaba, line 241: `[!code-csharp[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L246-326)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+        .ContinueWhenAll(writeTasks,
+            tasks => tasks.Select(t => t.IsFaulted ? TryUnwrapException(t.Exception) : null).ToImmutableList());
+
+    return result;
+}
+```
+
+#### DeleteMessagesToAsync
+
+```
```
eaba: `[!code-csharp[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L329)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+    long toSequenceNr)
+```
+
+Asynchronously deletes all persistent messages up to inclusive `toSequenceNr` bound. This operation has to be done atomically.
+
+This call is protected with a circuit-breaker.
+
+`DeleteBatchSql` in this example refers to this SQL query statement:
+
+```sqlite
```
eaba, line 343: `[!code-sql[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L85-88)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+DELETE FROM event_journal
+    WHERE persistence_id = @PersistenceId AND sequence_nr <= @ToSequenceNr;
+DELETE FROM journal_metadata
+    WHERE persistence_id = @PersistenceId AND sequence_nr <= @ToSequenceNr;
+```
+
+`HighestSequenceNrSql` in this example refers to this SQL query statement:
+
+```sqlite
```
eaba, line 352: Did not find this in the source. `sql` instead of `sqlite`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+SELECT MAX(e.ordering) as Ordering
+FROM event_journal e;
+```
+
+`UpdateSequenceNrSql` in this example refers to this SQL query statement:
+
+```sqlite
```
eaba, line 359: `[!code-sql[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L91-92)]`

docs/articles/persistence/custom-persistent-provider.md
```diff
@@ -0,0 +1,511 @@
...
+INSERT INTO journal_metadata (persistence_id, sequence_nr)
+VALUES (@PersistenceId, @SequenceNr);
+```
+
+```c#
```
eaba, line 364: `[!code-csharp[Main](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs#L329-387)]`

eaba (commented):
@Arkatufus `src/example` is the right folder for the custom persistence provider code!

src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs
```diff
@@ -0,0 +1,411 @@
...
+                    // Return the result if one is returned by the query result, else return zero
+                    return result is long ? Convert.ToInt64(result) : 0L;
+                }
+            }
+        }
+
+        protected sealed override async Task<IImmutableList<Exception>> WriteMessagesAsync(IEnumerable<AtomicWrite> messages)
+        {
+            // For each of the atomic write request, create an async Task 
+            var writeTasks = messages.Select(async message =>
```
to11mtm, line 249: Questions:

 - What is the risk that this task spawning pattern writes messages out of order? (In theory this may not be a big deal, but could lead to interesting edge cases)
 
 - What is the risk of this pattern hammering the datasource with open connections and causing issues on scale?

Aaronontheweb: @Arkatufus when we were discussing this documentation, we had raised the issue of having a "caveats and pitfalls" section - I think these two questions definitely belong in there: thinking about operation order (how does Akka.Persistence manage this for you vs. what do you need to worry about yourself) and connection management.

The built-in Akka.Persistence fundamentals should take care of the former (`AtomicWrite` is ordered and all operations in the `AtomicWrite` should be written in the same batch) but that semantic needs to be explained so an implementer can do this correctly.

Aaronontheweb:
@eaba are those sections of the Sqlite journal code that @Arkatufus could reference directly? We could decorate that code using `// <tagname> // </tagname>` blocks to target it.

src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs
Arkatufus: That is an interesting question, in theory, there is no guarantee that each `AtomicWrite` Task would be written in order, especially if one of them in the middle failed to be written to the datasource, the ordering would be screwed. @Aaronontheweb, do you have any input on this?

For the second question, I really have no answer. I would need to analyze the call chain and see if there are any optimization such as batching upstream of this method, I have a feeling that you're correct, that this method is being called for each produced message, but I would have to confirm it first.

Aaronontheweb: So when a `PersistentActor` writes multiple events via `PersistAll` - all of those events end up inside the same `AtomicWrite`; they're considered to be an atomic transaction from the point of view of the actor who wrote them.

The journal is supposed to process all of the `IPersistentRepresentation` (I think) objects that are included in the `AtomicWrite` in a single transaction to preserve atomicity. If there are multiple `AtomicWrite`s to process at once the order doesn't matter since they all came from different actors or the same actor using `PersistAsync` - if writes complete in different orders coming from the same actor I believe the `Resequencer` actor, built into the journal, will unroll the order in which the `WriteMessageSuccess` messages are delivered back to the original actor based on the sequence numbers that were generated prior to the write.

So the amount of things the implementor has to worry about is relatively small - the biggest one is making sure they don't group multiple `AtomicWrite`s into the same transaction. There are some implementations which explicitly break these atomicity guarantees, such as the `BatchingSqlJournal`, but that should be done only while explicitly making a trade off in the name of performance or throughput - not simply because the implementor didn't know better.

Aaronontheweb: > For the second question, I really have no answer. I would need to analyze the call chain and see if there are any optimization such as batching upstream of this method, I have a feeling that you're correct, that this method is being called for each produced message, but I would have to confirm it first.

Don't worry about this specific implementation. Worry about the general advice you're to give all implementors about connection management.

to11mtm: > That is an interesting question, in theory, there is no guarantee that each `AtomicWrite` Task would be written in order, especially if one of them in the middle failed to be written to the datasource, the ordering would be screwed. @Aaronontheweb, do you have any input on this?
> 
> For the second question, I really have no answer. I would need to analyze the call chain and see if there are any optimization such as batching upstream of this method, I have a feeling that you're correct, that this method is being called for each produced message, but I would have to confirm it first.

From the Docs on asyncwritejournal

Ordering:
>  Journals should aim to persist events in-order for a given `persistenceId`  as otherwise in case of a failure, the persistent state may be end up being inconsistent.

It's not a hard requirement, but it saves a lot of pain down the road, particularly if `persistAsync` is involved.

Atomicity:

> All `PersistentRepr` of the `AtomicWrite` must be written to the data store atomically, i.e. all or none must be stored. If the journal (data store) cannot support atomic writes of multiple events it should reject such writes with a `Try` `Failure` 
> This limitation should also be documented by the journal plugin.

Blocking Sequence Reads on writes:

> Please also note that requests for the highest sequence number may be made concurrently to this call executing for the same `persistenceId`, in particular it is possible that a restarting actor tries to recover before its outstanding writes have completed. In the latter case it is highly desirable to defer reading the highest sequence number until all outstanding writes have completed, otherwise the <see cref="PersistentActor"/> may reuse sequence numbers.

Note there's some edge cases around the above, even if you get it right in the provider itself (e.x. too short a recovery timeout on cluster shards/singletons, and they somehow read before the previous system finished it's somehow-delayed write)

I do know that as far as ordering, [in persistince-jdbc's batching they used `mapAsync` rather than `mapAsyncUnordered`](https://github.com/akka/akka-persistence-jdbc/blob/c89d588108f20646976be10a5028a6dc22958a6e/core/src/main/scala/akka/persistence/jdbc/journal/dao/BaseDao.scala#L22-L33), even though the atomicWrites are atomically grouped (via batchWeighted). I'm not certain whether there's still some edge cases with aggressive PersistAsync usage that could cause issues with this pattern (I think EventSourced handles this with it's batching).

Also, they handle the case of `readHighestSequenceNr` blocking for writes via a Future [Well, here's the ported version since it may be more useful](https://github.com/akkadotnet/Akka.Persistence.Linq2Db/blob/dev/src/Akka.Persistence.Sql.Linq2Db/Journal/Linq2DbWriteJournal.cs#L167-L177).

> If there are multiple AtomicWrites to process at once the order doesn't matter since they all came from different actors or the same actor using PersistAsync

@Aaronontheweb So if you look at the ported code above, what's interesting is that it appears there is an implicit guarantee that for a given call to `WriteMessagesAsync`, all of the `AtomicWrite`s will be the same persistenceId. [here's the scala version for ref](https://github.com/akka/akka-persistence-jdbc/blob/c89d588108f20646976be10a5028a6dc22958a6e/core/src/main/scala/akka/persistence/jdbc/journal/JdbcAsyncWriteJournal.scala#L91-L94).

The implication of all this, is for a given call to `WriteMessagesAsync`, you're best off doing one of the following to keep things as 'Akka' as possible (i.e. fail fast in a recoverable way):

 - Treating all AtomicWrites in that call as a single transacted write to the store
   - very easy to implement for the example IMO
   - As noted above, basically what persistence-jdbc and Persistence.Linq2Db do.
 - Ensuring that all AtomicWrites are chunked into an `AtomicWriteGroup` such that:
   - A chunk contains 1 or more `AtomicWrite`s from the group
   - An `AtomicWrite` is only inside one chunk
   - Chunks are processed in-order
   - As soon as one chunk fails, all subsequent chunks in that Group ignored.

Obviously the second option is a good bit more complicated, but I threw it out there for others who may want to roll their own.

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
+---
+uid: custom-persistent-provider
+title: Writing A Custom Persistent Provider
+---
```
eaba, line 4: LGTM

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+The `Journal` table schemas we will be using are:
+
+[!code-csharp[Schema](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?name=schema "Journal table schemas")]
+
+### Required Method Implementations
+
+#### ReplayMessagesAsync
+
+```c#
```
eaba, line 26: Do we leave this as it is or make use of the range selection though the line number may change in the future?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+[!code-csharp[ByPersistenceIdSql](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?name=ByPersistenceIdSql "ByPersistenceIdSql SQL Statement")]
+
+SQLite code example:
+
+[!code-csharp[ReplayMessagesAsync](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?name=ReplayMessagesAsync "ReplayMessagesAsync method implementation")]
+
+#### ReadHighestSequenceNrAsync
+
+```c#
```
eaba, line 67: Do we leave this as it is or make use of the range selection though the line number may change in the future?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+[!code-csharp[HighestSequenceNrSql](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?name=HighestSequenceNrSql "HighestSequenceNrSql SQL Statement")]
+
+SQLite code example:
+
+[!code-csharp[ReadHighestSequenceNrAsync](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?name=ReadHighestSequenceNrAsync "ReadHighestSequenceNrAsync method implementation")]
+
+#### WriteMessagesAsync
+
+```c#
```
eaba, line 92: Do we leave this as it is or make use of the range selection though the line number may change in the future?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+[!code-csharp[InsertEventSql](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?name=InsertEventSql "InsertEventSql SQL Statement")]
+
+SQLite code example:
+
+[!code-csharp[WriteMessagesAsync](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?name=WriteMessagesAsync "WriteMessagesAsync method implementation")]
+
+#### DeleteMessagesToAsync
+
+```c#
```
eaba, line 155: Do we leave this as it is or make use of the range selection though the line number may change in the future?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+[!code-csharp[DeleteMessagesToAsync](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?name=DeleteMessagesToAsync "DeleteMessagesToAsync method implementation")]
+
+### Detail Implementation
+
+#### Reading HOCON Settings
+
+There are some HOCON settings that are by default loaded by the journal base class and these can be overriden in your HOCON settings. The minimum HOCON settings that need to be defined for your custom plugin are:
+
+```hocon
```
eaba, line 192: @Arkatufus what is the best option here?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+    custom-sqlite {
+        # qualified type name of the SQLite persistence journal actor
+        class = "Akka.Persistence.Custom.Journal.SqliteJournal, Akka.Persistence.Custom"
+    }
+}
+```
+
+The default HOCON settings are:
+
+```hocon
```
eaba, line 204: @Arkatufus @Aaronontheweb what is the best option here? Any chance that this won't be touched in the future?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+[!code-csharp[Startup](../../../src/examples/Akka.Persistence.Custom/Journal/SqliteJournal.cs?name=Startup "Journal startup sequence")]
+
+## Implementing Akka.Persistence SnapshotStore
+
+### Required Method Implementations
+
+#### LoadAsync
+
+```c#
```
eaba, line 282: Do we leave this as it is or make use of the range selection though the line number may change in the future?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+[!code-csharp[SelectSnapshotSql](../../../src/examples/Akka.Persistence.Custom/Snapshot/SqliteSnapshotStore.cs?name=SelectSnapshotSql "SelectSnapshotSql SQL Statement")]
+
+SQLite code example:
+
+[!code-csharp[LoadAsync](../../../src/examples/Akka.Persistence.Custom/Snapshot/SqliteSnapshotStore.cs?name=LoadAsync "LoadAsync method implementation")]
+
+#### SaveAsync
+
+```c#
```
eaba, line 305: Do we leave this as it is or make use of the range selection though the line number may change in the future?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+[!code-csharp[InsertSnapshotSql](../../../src/examples/Akka.Persistence.Custom/Snapshot/SqliteSnapshotStore.cs?name=InsertSnapshotSql "InsertSnapshotSql SQL Statement")]
+
+SQLite code example:
+
+[!code-csharp[SaveAsync](../../../src/examples/Akka.Persistence.Custom/Snapshot/SqliteSnapshotStore.cs?name=SaveAsync "SaveAsync method implementation")]
+
+#### DeleteAsync
+
+```c#
```
eaba, line 328: Do we leave this as it is or make use of the range selection though the line number may change in the future?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+[!code-csharp[DeleteSnapshotRangeSql](../../../src/examples/Akka.Persistence.Custom/Snapshot/SqliteSnapshotStore.cs?name=DeleteSnapshotRangeSql "DeleteSnapshotRangeSql SQL Statement")]
+
+SQLite code example:
+
+[!code-csharp[DeleteAsync](../../../src/examples/Akka.Persistence.Custom/Snapshot/SqliteSnapshotStore.cs?name=DeleteAsync "DeleteAsync method implementation")]
+
+#### DeleteAsync
+
+```c#
```
eaba, line 354: Do we leave this as it is or make use of the range selection though the line number may change in the future?

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,467 @@
...
+
+[!code-csharp[DeleteAsync2](../../../src/examples/Akka.Persistence.Custom/Snapshot/SqliteSnapshotStore.cs?name=DeleteAsync2 "DeleteAsync method implementation")]
+
+### Detail Implementation
+
+#### Reading HOCON Settings
+
+There are some HOCON settings that are by default loaded by the snapshot store base class and these can be overriden in your HOCON settings. The minimum HOCON settings that need to be defined for your custom plugin are:
+
+```hocon
```
eaba, line 381: What is the best option here?

docs/articles/toc.yml
```diff
@@ -64,6 +64,8 @@
     href: persistence/storage-plugins.md
   - name: Custom serialization
     href: persistence/custom-serialization.md
+  - name: Writing A Custom Persistent Provider
+    href: persistence/custom-persistence-provider.md
   - name: Persistence Query
     href: persistence/persistence-query.md
   - name: Persistence Testing
```
eaba, line 71: While we are here, can we create `toc.yml` for the persistence section and use that here? For example: `https://github.com/Arkatufus/akka.net/tree/master/docs/articles/intro/getting-started/toc.yml`

docs/articles/persistence/custom-persistence-provider.md
Arkatufus: That is the name of the function that needs to be implemented, the code block is used here just for formatting

docs/articles/persistence/custom-persistence-provider.md
Arkatufus: see above

docs/articles/persistence/custom-persistence-provider.md
Arkatufus: see above

docs/articles/persistence/custom-persistence-provider.md
Arkatufus: Nothing, HOCON is a proprietary format, there are no CSS equivalent

docs/articles/persistence/custom-persistence-provider.md
Arkatufus: see above

docs/articles/persistence/custom-persistence-provider.md
Arkatufus: see above

docs/articles/persistence/custom-persistence-provider.md
Arkatufus: see above

docs/articles/persistence/custom-persistence-provider.md
Arkatufus: Nothing, HOCON is a proprietary format, there are no CSS equivalent, no support in DocFX

docs/articles/persistence/custom-persistence-provider.md
Arkatufus: Nothing, HOCON is a proprietary format, there are no CSS equivalent, no support in DocFX

docs/articles/toc.yml
Arkatufus: That would be out of scope of this PR, you're mixing 2 kinds of issues

docs/articles/persistence/custom-persistence-provider.md
Aaronontheweb: You can still reference a HOCON file use a doc include - we do this in all of the default config docs

Arkatufus: I think the support is limited to code line range? and since HOCON is a nested structure in nature, linking to a line of code inside a HOCON is a bit meaningless because you would not get all of the parent property names before and can actually lead to more confusion by the reader

Aaronontheweb: you'd have to either include an entire stand-alone HOCON file (which works) or wrap around a C# format string using line numbers. Separate file is probably best.

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,466 @@
...
+```c#
+protected abstract Task<IImmutableList<Exception>> WriteMessagesAsync(
+    IEnumerable<AtomicWrite> messages)
+```
+
+Asynchronously writes a batch of persistent messages to the journal.
+
+This call is protected with a circuit-breaker.
+
+Calls to this method are serialized by the enclosing journal actor. If you spawn work in asynchronous tasks it is alright that they complete the futures in any order, but the actual writes for a specific persistenceId should be serialized to avoid issues such as events of a later write are visible to consumers (query side, or replay) before the events of an earlier write are visible. A `PersistentActor` will not send a new `WriteMessages` request before the previous one has been completed.
```
Aaronontheweb, line 101: Not 100% true - `PersistAsync` explicitly does this. The journal may save events out of order, but they're ACKed back to the caller in the original order in which they were sent. Since the caller sets the `SeqNo`, which determines the ordering, this operation is safe.

docs/articles/persistence/custom-persistence-provider.md
```diff
@@ -0,0 +1,466 @@
...
+
+Please note that the `IPersistentRepresentation.Sender` of the contained `IPersistentRepresentation` objects has been set to null (i.e. set to `ActorRefs.NoSender`) in order to save space in the journal and to prevent saving a sender reference that will likely be obsolete during replay.
+
+Please also note that requests for the highest sequence number may be made concurrently to this call executing for the same `persistenceId`, in particular it is possible that a restarting actor tries to recover before its outstanding writes have completed. In the latter case it is highly desirable to defer reading the highest sequence number until all outstanding writes have completed, otherwise the `PersistentActor` may reuse sequence numbers.
+
+##### Batching
+
+The batch is only for performance reasons, i.e. all messages don't have to be written atomically. Higher throughput can typically be achieved by using batch inserts of many records compared to inserting records one-by-one, but this aspect depends on the underlying data store and a journal implementation can implement it as efficiently as possible.
```
Aaronontheweb, line 109: It's not clear what you're referring to here by "the Batch"

PR #5523, opened 2022-01-19
Removed comment that no longer applies.

brah-mcdude:
When I wrote this comment originally, it applied. Since then, the implementation of the method has changed. Now, the comment no longer applies. Therefore, I removed the comment.


Aaronontheweb (approved):
LGTM

PR #5529, opened 2022-01-23
added 2 tests - a true and a false positive

brah-mcdude:
Hi @Aaronontheweb. 
[DO NOT MERGE - This is not designed to be an actual pull request - this is just a question - but a very important one - please read]
I have been sitting on this for a couple of days. So I am turning to you as a last resort. 
I am unable to understand an issue that I have been having. 
I added 2 tests to a sample test. 
1. ReplyingToProbeMessagesFunc
2. ReplyingToProbeMessagesAction

Notice that 1 passes and 2 fails. 
But if you debug 1, you will see that the test should fail. 
As the condition asserted is false. 
So while debugging, an assertion pops up. 
But if you continue debugging, the TEST PASSES. 
What is going on?
This is something I find difficult to live with, 
as this seems to only occur with EventFilter's ExpectAsync actionAsync
which I depend on in all my tests.
Which means that I cannot use the await keyword inside my test code:(  
Is it a bug in Visual Studio?
In xUnit?
In Akka?
Or is it my bad?
Either way, we need to at the very least be aware of whatever it is. 
Thank you.

### DIFF
--- a/src/core/Akka.Docs.Tests/Testkit/ProbeSampleTest.cs
+++ b/src/core/Akka.Docs.Tests/Testkit/ProbeSampleTest.cs
@@ -5,6 +5,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.TestKit;
 using Akka.TestKit.Xunit2;
@@ -14,7 +15,7 @@ namespace DocsExamples.Testkit
 {
     public class ProbeSampleTest : TestKit
     {
-#region ProbeSample_0
+        #region ProbeSample_0
         public class Forwarder : ReceiveActor
         {
             private IActorRef target;
@@ -32,7 +33,7 @@ namespace DocsExamples.Testkit
         {
             //create a test probe
             var probe = CreateTestProbe();
-            
+
             //create a forwarder, injecting the probo's testActor
             var props = Props.Create(() => new Forwarder(probe));
             var forwarder = Sys.ActorOf(props, "forwarder");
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
I'll try to get you an answer on this today or tomorrow

Aaronontheweb:
So I see that `DocsExamples.Testkit.ProbeSampleTest.ReplyingToProbeMessagesAction` fails consistently with:

```
Akka.TestKit.Xunit2.Internals.AkkaEqualException : Assert.Equal() Failure
Expected: world2
Actual:   world
```

This fails because

```csharp
await EventFilter.Error().ExpectAsync(0, action: () =>
  {
      probe.Tell("hello");
      probe.ExpectMsg("hello");
      probe.Reply("world");
      ExpectMsg("world2");
      Assert.Equal(probe.Ref, LastSender);
  });
```
We reply with `world` and not `world2` - you probably figured that one out already.

The biggest mystery is why does this pass:

```csharp
/// <summary>
/// this test is a false positive
/// </summary>
/// <returns></returns>
[Fact]
public async Task ReplyingToProbeMessagesFunc()
{
    await EventFilter.Error().ExpectAsync(0, actionAsync: async () =>
    {
        probe.Tell("hello");
        probe.ExpectMsg("hello");
        probe.Reply("world");
        await Task.Run(() => { ExpectMsg("world2"); });
        Assert.Equal(probe.Ref, LastSender);
    });
}
```

Looks to me like we forgot to.... you know, `await` the `Task` returned inside here: 
https://github.com/akkadotnet/akka.net/blob/38b7443fd03f0025ac0e7a7a1ed5e1f636bfb6ba/src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs#L454-L460

So I suspect the ` await Task.Run(() => { ExpectMsg("world2"); });` is running as a detatched task, which is incredibly annoying. We just need to update the TestKit to await those tasks internally and then this should pass.

Aaronontheweb:
This is definitely a bug, we should patch it, and we can use your spec as a reproduction. Akka.TestKit.Tests is a good place to add this.

brah-mcdude:
Would you like me to do it?

Aaronontheweb:
> Would you like me to do it?

That would be great

brah-mcdude:
On it:)

PR #5531, opened 2022-01-23
Renamed FishUntilMessage to FishUntilMessageAsync

brah-mcdude:
This PR renaming breaks backwards compatibility. 
But since the original fixed code is very young - I think it is worth it.


Aaronontheweb:
Approved this because I'm pretty sure @brah-mcdude is the only person using this method so far, given how recently it was introduced.

PR #5536, opened 2022-01-25
Add CPU benchmark project

Arkatufus:
## Changes

Add a standalone console CPU benchmark project to measure Akka.Cluster CPU usage

```
Usage: dotnet run -c Release
Usage: dotnet run -c Release -- [OPTIONS]
Usage: Akka.Cluster.Cpu.Benchmark [OPTIONS]
Options:
  -d, --sample-duration=VALUE  
      Sets the sample point duration in seconds. Default: 5 seconds
  --delay=VALUE 
      Sets the initial delay to wait for cluster to stabilize before benchmark starts in seconds. Default: 5 seconds
  -s, --samples=VALUE 
      Sets how many samples are taken during the benchmark. Default: 60 samples
  -c, --cluster-size=VALUE 
      Sets how many nodes to be added to the cluster in addition to the tested node. Default: 9 nodes
  -w, --warm-up-count=VALUE      
      Sets how many blank samples to be performed as benchmark warm-up before benchmark starts. Default: 5 samples
  -h, -?, --help       
      Shows help
```


Aaronontheweb (approved):
LGTM

Issue #5537, opened 2022-01-25
InternalExpectAsync does not await actionAsync() - causing actionAsync to run as a detached task


brah-mcdude:
why does this pass:

```csharp
/// <summary>
/// this test is a false positive
/// </summary>
/// <returns></returns>
[Fact]
public async Task ReplyingToProbeMessagesFunc()
{
    await EventFilter.Error().ExpectAsync(0, actionAsync: async () =>
    {
        var probe = CreateTestProbe();
        probe.Tell("hello");
        probe.ExpectMsg("hello");
        probe.Reply("world");
        await Task.Run(() => { ExpectMsg("world2"); });
        Assert.Equal(probe.Ref, LastSender);
    });
}
```

Looks to me like we forgot to.... you know, `await` the `Task` returned inside here: 
https://github.com/akkadotnet/akka.net/blob/38b7443fd03f0025ac0e7a7a1ed5e1f636bfb6ba/src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs#L454-L460

So I suspect the ` await Task.Run(() => { ExpectMsg("world2"); });` is running as a detatched task, which is incredibly annoying. We just need to update the TestKit to await those tasks internally and then this should pass.

_Originally posted by @Aaronontheweb in https://github.com/akkadotnet/akka.net/issues/5529#issuecomment-1020502375_

Aaronontheweb:
Just need to `await actionAsync()` and delete the `Task.FromResult` call there I think.

brah-mcdude:
```c#
Akka.TestKit.Xunit2.Internals.AkkaEqualException
  HResult=0x80131500
  Message=Assert.Equal() Failure
Expected: world2
Actual:   world
  Source=Akka.TestKit.Xunit2
  StackTrace:
   at Akka.TestKit.Xunit2.XunitAssertions.AssertEqual[T](T expected, T actual, String format, Object[] args) in C:\Users\brahm\source\repos\akka.net\src\contrib\testkits\Akka.TestKit.Xunit2\XunitAssertions.cs:line 68
   at Akka.TestKit.TestKitBase.<>c__DisplayClass108_0`1.<ExpectMsg>b__0(T m) in C:\Users\brahm\source\repos\akka.net\src\core\Akka.TestKit\TestKitBase_Expect.cs:line 54
[...]

PR #5538, opened 2022-01-25
fixed - InternalExpectAsync does not await actionAsync() - causing ac…

brah-mcdude:
…tionAsync to run as a detached task #5537

Fixes #
InternalExpectAsync does not await actionAsync() - causing actionAsync to run as a detached task #5537

## Changes
1. added a test ExpectAsync_should_await_actionAsync that reproduced the issue by failing and stopped failing when fix was applied. 
2. the actual fix: added async and await. 

1. the issue was that a test that should have failed, was not failing. 
2. so I created test that reproduces this issue - a test that fails - when a failed assertion was not failing (a false positive).
3. once I created the test (RED), I fixed the issue (GREEN). 
4. done

Include data from before / after this change here.

Before: 
```c#
﻿ Akka.TestKit.Tests.Xunit2.TestEventListenerTests.CustomEventFilterWarningTests.ExpectAsync_should_await_actionAsync
   Source: AllTestForEventFilterBase.cs line 180
   Duration: 3.6 sec

  Message: 
Assert.Throws() Failure
Expected: typeof(Akka.TestKit.Xunit2.Internals.AkkaEqualException)
Actual:   (No exception was thrown)

  Stack Trace: 
AllTestForEventFilterBase`1.ExpectAsync_should_await_actionAsync() line 183

```

After: 

```c#
﻿ Akka.TestKit.Tests.Xunit2.TestEventListenerTests.CustomEventFilterWarningTests.ExpectAsync_should_await_actionAsync
   Source: AllTestForEventFilterBase.cs line 180
   Duration: 450 ms

```

### DIFF
--- a/src/core/Akka.TestKit.Tests/TestEventListenerTests/AllTestForEventFilterBase.cs
+++ b/src/core/Akka.TestKit.Tests/TestEventListenerTests/AllTestForEventFilterBase.cs
@@ -7,7 +7,9 @@
 
 using System;
 using System.Threading.Tasks;
+using Akka.Actor;
 using Akka.Event;
+using Akka.TestKit.Xunit2.Internals;
 using FluentAssertions;
 using Xunit;
 using Xunit.Sdk;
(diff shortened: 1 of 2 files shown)


Aaronontheweb (approved):
LGTM

brah-mcdude:
Just for the record, here is the original pull request where I described and reproduced the issue: 
https://github.com/akkadotnet/akka.net/pull/5529/files

PR #5545, opened 2022-01-27
Added note that Test Receive method violates the actor model

brah-mcdude:
It is just a comment. But a rather important one.


src/core/Akka.TestKit/TestActorRefBase.cs
```diff
@@ -39,6 +39,10 @@ protected TestActorRefBase(ActorSystem system, Props actorProps, IActorRef super
         /// Directly inject messages into actor receive behavior. Any exceptions
         /// thrown will be available to you, while still being able to use
         /// become/unbecome.
+        /// Note: This method violates the actor model and could cause unpredictable
```
Aaronontheweb, line 42: Yep and I guess TestKit users should be made aware of that.

PR #5547, opened 2022-01-28
Catch `StreamDetachedException`

### LINKED ISSUE #5546: [RACY TESTS] RunAsAsyncEnumerable Throws on Abrupt Stream termination
https://github.com/akkadotnet/akka.net/blob/4edb257d96c82f569a22e93e778c3a62ab9f483c/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs#L97-L125

When the test fails, it fails with the error message: `Akka.Streams.StreamDetachedException : Stream is terminated. Materialized value is detached.` 

I found two area where this is likely been thrown or set:

https://github.com/akkadotnet/akka.net/blob/db28aa32a6a6f5c98bc6fcd9ba275372cbf71229/src/core/Akka.Streams/Implementation/Sinks.cs#L808-L810

https://github.com/akkadotnet/akka.net/blob/db28aa32a6a6f5c98bc6fcd9ba275372cbf71229/src/core/Akka.Streams/Implementation/Sources.cs#L168-L178

Seeing that it is Set when `Logic` stops, two questions:
- Is lopgic not designed to be stopped?
- What could be the cause of `Logic` stopping?

eaba:
Fixes #5546

## Changes

Handle `StreamDetachedException` @to11mtm

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
@@ -117,6 +117,10 @@ namespace Akka.Streams.Tests.Dsl
             {
                 await a;
             }
+            catch (StreamDetachedException e) 
+            { 
+                thrown = true; 
+            } 
             catch (AbruptTerminationException e)
             {
                 thrown = true;


Aaronontheweb (approved):
LGTM

PR #5552, opened 2022-01-30
Added one more message to the test

### LINKED ISSUE #5551: FishForMessage should allow an allMessages optional parameter that can return all the messages until isMessage returns true
Is your feature request related to a problem? Please describe: 
I need this feature in my tests. 

Describe the solution you'd like:
1. add a param: List<object> allMessages = null
2. add this line in the loop: allMessages?.Add(msg);

Describe alternatives you've considered: 
Write a separate method.

brah-mcdude:
close #5551


brah-mcdude:
The issue for this PR: 
FishForMessage should allow an allMessages optional parameter that can return all the messages until isMessage returns true #5551

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -40,8 +40,9 @@ public object FishForMessage(Predicate<object> isMessage, TimeSpan? max = null,
         /// <param name="isMessage">The is message.</param>
         /// <param name="max">The maximum.</param>
         /// <param name="hint">The hint.</param>
+        /// <param name="allMessages">If null then will be ignored. If not null then will be filled with all the messages until <paramref name="isMessage"/> returns <c>true</c></param>
         /// <returns>Returns the message that <paramref name="isMessage"/> matched</returns>
-        public T FishForMessage<T>(Predicate<T> isMessage, TimeSpan? max = null, string hint = "")
+        public T FishForMessage<T>(Predicate<T> isMessage, TimeSpan? max = null, string hint = "", List<object> allMessages = null)
```
Aaronontheweb, line 45: In order to make this change binary compatible - you need to introduce a new overload here, rather than an optional parameter.

Aaronontheweb (changes requested):
Need to support our API compatibility guidelines - see https://getakka.net/community/contributing/api-changes-compatibility.html#how-to-safely-introduce-public-api-changes-extend-only-design for an explanation on how to introduce these types of changes in a binary-compatible way.

src/core/Akka.TestKit/TestKitBase_Receive.cs
brah-mcdude: As can be seen in https://github.com/akkadotnet/akka.net/pull/5552/commits/4677480c592eb5ad10f1341ca60252dd7286831e, I got a build error: "Error	CS0121	The call is ambiguous between the following methods or properties: 'TestKitBase.FishForMessage<T>(Predicate<T>, TimeSpan?, string)' and 'TestKitBase.FishForMessage<T>(Predicate<T>, TimeSpan?, string, List<object>)'	Akka.TestKit.Tests (net471), Akka.TestKit.Tests (net6.0), Akka.TestKit.Tests (netcoreapp3.1)	C:\Users\brahm\source\repos\akka.net\src\core\Akka.TestKit.Tests\TestKitBaseTests\ReceiveTests.cs	54"

brah-mcdude:
```csharp
        [Fact]
        public void FishForMessage_should_return_matched_message()
        {
            TestActor.Tell(1);
            TestActor.Tell(2);
            TestActor.Tell(10);
            TestActor.Tell(20);
            FishForMessage<int>(i => i >= 10).ShouldBe(10); /// <<<<< ""Error CS0121 The call is ambiguous..."
        }


```

Aaronontheweb: Might need to rename the method then. Either way, can't break binary compat.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -56,7 +56,7 @@ public T FishForMessage<T>(Predicate<T> isMessage, TimeSpan? max = null, string
         /// <param name="hint">The hint.</param>
         /// <param name="allMessages">If null then will be ignored. If not null then will be filled with all the messages until <paramref name="isMessage"/> returns <c>true</c></param>
         /// <returns>Returns the message that <paramref name="isMessage"/> matched</returns>
-        public T FishForMessage<T>(Predicate<T> isMessage, TimeSpan? max = null, string hint = "", List<object> allMessages = null)
+        public T FishForMessage2<T>(Predicate<T> isMessage, TimeSpan? max = null, string hint = "", List<object> allMessages = null)
```
Aaronontheweb, line 59: Maybe a better way to do this, since `FishForMessage2` isn't all that intuitive, is to just make the `allMessages` parameter not optional on the overload - place it before the optional `TimeSpan`. That would be better I think.

Aaronontheweb: Did you mean to push a new commit to resolve this message?

brah-mcdude: No. I clicked "resolve" by mistake.

brah-mcdude: I just pushed a new commit.

Aaronontheweb (approved):
LGTM

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -42,9 +43,25 @@ public object FishForMessage(Predicate<object> isMessage, TimeSpan? max = null,
...
+        /// Receives messages until <paramref name="isMessage"/> returns <c>true</c>.
+        /// Use it to ignore certain messages while waiting for a specific message.
+        /// </summary>
+        /// <typeparam name="T">The type of the expected message. Messages of other types are ignored.</typeparam>
+        /// <param name="isMessage">The is message.</param>
+        /// <param name="max">The maximum.</param>
+        /// <param name="hint">The hint.</param>
+        /// <param name="allMessages">If null then will be ignored. If not null then will be initially cleared, then filled with all the messages until <paramref name="isMessage"/> returns <c>true</c></param>
+        /// <returns>Returns the message that <paramref name="isMessage"/> matched</returns>
+        public T FishForMessage<T>(Predicate<T> isMessage, ArrayList allMessages, TimeSpan? max = null, string hint = "")
```
Aaronontheweb, line 60: Shouldn't this be just a `List<T>`?

Aaronontheweb (changes requested):
API suggestion before merge

src/core/Akka.TestKit/TestKitBase_Receive.cs
brah-mcdude: In other words: We are fishing for type T. Whereas all-messages might NOT be of type T.

Arkatufus (approved):
LGTM

Aaronontheweb (approved):

Aaronontheweb:
Looks we have several test failures here too

Aaronontheweb (changes requested):
Akka.Testkit.Tests.TestKitBaseTests.ReceiveTests.WaitForRadioSilenceAsync_should_reset_timer_twice_only
Expected messages to be a collection with 2 item(s), but found an empty collection.

brah-mcdude:
> Akka.Testkit.Tests.TestKitBaseTests.ReceiveTests.WaitForRadioSilenceAsync_should_reset_timer_twice_only Expected messages to be a collection with 2 item(s), but found an empty collection.

I created a separate issue: 
test WaitForRadioSilenceAsync_should_reset_timer_twice_only fails with empty collection #5640

And I submitted a PR to hopefully fix this issue: 
slow down the test - WaitForRadioSilenceAsync_should_reset_timer_twice_only #5641

Aaronontheweb:
thanks @brah-mcdude - going to see if your other PR helped

brah-mcdude:
@Aaronontheweb - please let me know if everything is ok or if I need to modify / fix anything.

PR #5556, opened 2022-01-31
Remove direct reference to Props in LocalActorRef

### LINKED ISSUE #5553: Memory leak in DistributedData on cross-node replication
**Version Information**
Version: Akka.NET 1.4.32
Akka.NET Modules: Akka.DistributedData
Environment: Windows 11, .Net 6.0, no containers

Akka.DistributedData module leaks memory while replicating DData updates between nodes.

1. Create new project using _Petabridge Akka.Cluster Starter App_ template (pb-akka-cluster, v.1.2.0).
2. Add Akka.DistributedData module.
3. Modify existing _AkkaService_ class to perform periodic updates of the _LWWRegister_ items through the DData replicator with WriteAll consistency. 
4. Run 2 or more instances of the test app to form a cluster with data replication among nodes.
5. Each DData update operation would leave GC non-collectable LWWRegister instance locked on ActorRefResolveCache's ThreadLocal instance which eventually would flood Gen2.

Reproduction: [Akka.DistributedData-MemLeak](https://github.com/Tavriets/Akka.DistributedData-MemLeak)

**Expected behavior**
DData replicator should not leave zombie instances of LocalActorRef type after DData replication is done.

**Actual behavior**
scr1

The number of retained ActorRefResolveCache will grow proportionally to the number of DData updates:
scr2

A typical retention tree of the objects survived through GCs:
scr3

Some additional observations:
- Problem seems to be replicable only within multi-node cluster. Single-node cluster without cross-node DData replication behaves well.
- Problem is not only limited to _LWWRegister_. Other CRDT types would stuck in Gen2 in the same way.
- The retrieval of values stored in the DData does not cause memleak, so it seems to be related to the update replication only.

Arkatufus:
Fixes #5553

## Changes

Remove reference to Props in LocalActorRef, this have a huge memory impact when LocalActorRef is being cached.

### DIFF
--- a/src/core/Akka/Actor/LocalActorRef.cs
+++ b/src/core/Akka/Actor/LocalActorRef.cs
@@ -22,7 +22,6 @@ namespace Akka.Actor
     public class LocalActorRef : ActorRefWithCell, ILocalRef
     {
         private readonly ActorSystemImpl _system;
-        private readonly Props _props;
         private readonly MessageDispatcher _dispatcher;
         private readonly IInternalActorRef _supervisor;
         private readonly ActorPath _path;
@@ -57,7 +56,6 @@ namespace Akka.Actor
             IInternalActorRef supervisor, ActorPath path)
         {
             _system = system;
-            _props = props;
             _dispatcher = dispatcher;
             MailboxType = mailboxType;
             _supervisor = supervisor;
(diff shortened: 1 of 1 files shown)


Arkatufus:
Before, snapshot taken at 2 minutes 52 seconds:
image

image

After, snapshot taken at 5 minutes 25 seconds:
image

image

Aaronontheweb (approved):
LGTM - very nice work

PR #5560, opened 2022-02-02
Created a dedicated `prune-interval`

### LINKED ISSUE #5352: DistributedPubSub: publishing to topics that have no subscribers doesn't produce any dead letters
**Version Information**
Version of Akka.NET? 1.4.27

**Describe the bug**
Once the final subscriber to a topic unsubscribes, no one is available to receive these messages published to this topic. Yet, when messages are published no `DeadLetter`s are produced.

**To Reproduce**
1. Subscribe to topic "foo" with one `IActorRef`
2. Unsubscribe `IActorRef` from "foo"
3. Wait for `UnsubscribeAck`
4. Query `CurrentTopics` from Mediator - "foo" still appears.

This appears to be because:

https://github.com/akkadotnet/akka.net/blob/838b6a3405b43632f5fee4e09af6651cecc7d533/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs#L248-L253

And the `Prune` operation never un-register any empty buckets once they've been created.

**Expected behavior**
Topic registrations should eventually be removed when not in use - or should publish to dead letters when there are no subscribers.

**Actual behavior**
If no one can receive a message published to a topic, whatever the reason, it should be logged as such - usually a `DeadLetter`.

eaba:
Fixes #5352

## Changes

Added `prune-interval` property to `DistributedPubSubSettings`

## Checklist
* [ ] Design discussion issue #5352 

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
@@ -12,6 +12,9 @@ using Akka.Actor.Dsl;
 using Akka.Cluster.Tools.PublishSubscribe;
 using Akka.TestKit;
 using Xunit;
+using Akka.Event;
+using FluentAssertions;
+using System.Threading.Tasks;
 
 namespace Akka.Cluster.Tools.Tests.PublishSubscribe
 {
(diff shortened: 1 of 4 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
```diff
@@ -54,5 +57,94 @@ public void DistributedPubSubMediator_can_be_activated_in_child_actor()
                 a.ShouldNotBe(ActorRefs.Nobody);
             });
         }
+
+        [Fact]
+        public async Task DistributedPubSubMediator_should_publish_to_deadletters_if_no_active_subscriber()
```
eaba, line 62: This test will need improvement. Did this for quick testing!

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -159,7 +160,7 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
 
             //Start periodic gossip to random nodes in cluster
             _gossipCancelable = Context.System.Scheduler.ScheduleTellRepeatedlyCancelable(_settings.GossipInterval, _settings.GossipInterval, Self, GossipTick.Instance, Self);
-            _pruneInterval = new TimeSpan(_settings.RemovedTimeToLive.Ticks / 2);
+            _pruneInterval = _settings.PruneInterval;
```
eaba, line 163: The prune interval is now set with the PruneInterval setting

eaba (commented):
When the Topic actor is Terminate, it is still considered alive here:
https://github.com/akkadotnet/akka.net/blob/dbe49992f3537485363ecf26aff2348e08b78b81/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs#L503-L529

Aaronontheweb (changes requested):
Isn't this just duplicating the functionality of the TTL system?

PR #5561, opened 2022-02-02
Fix: no `DeadLetter`s when publishing to a `Topic` with no subscribers

eaba:
The summary of the PubSub design is this, as it relates to #5352, if there are no subscribers, wait on the Deadline and if no new subscriber within that window terminate the Topic actor.

It is the `waiting on the Deadline` that is the actual issue.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
@@ -21,11 +21,14 @@ namespace Akka.Cluster.Tools.Tests.PublishSubscribe
     [Collection(nameof(DistributedPubSubMediatorSpec))]
     public class DistributedPubSubMediatorSpec : AkkaSpec
     {
-        public DistributedPubSubMediatorSpec() : base(GetConfig()) { }
+        public DistributedPubSubMediatorSpec() : base(GetConfig())
+        {
+        }
 
         public static Config GetConfig()
         {
-            return ConfigurationFactory.ParseString("akka.actor.provider = \"Akka.Cluster.ClusterActorRefProvider, Akka.Cluster\"");
+            return ConfigurationFactory.ParseString(
+                @"akka.actor.provider = cluster");
         }
 
         /// <summary>
@@ -44,11 +47,8 @@ namespace Akka.Cluster.Tools.Tests.PublishSubscribe
                         mediator = DistributedPubSub.Get(actorContext.System).Mediator;
                     };
 
-                    dsl.Receive<string>(s => s.Equals("check"), (s, actorContext) =>
-                    {
-                        actorContext.Sender.Tell(mediator);
-                    });
-
+                    dsl.Receive<string>(s => s.Equals("check"),
+                        (s, actorContext) => { actorContext.Sender.Tell(mediator); });
                 }, "childActor");
 
                 actor.Tell("check");
@@ -58,30 +58,35 @@ namespace Akka.Cluster.Tools.Tests.PublishSubscribe
             });
         }
 
+        /// <summary>
+        /// Reproduction for https://github.com/akkadotnet/akka.net/issues/5352
+        /// </summary>
         [Fact]
         public async Task DistributedPubSubMediator_should_send_messages_to_dead_letter()
         {
-            await EventFilter.DeadLetter<object>().ExpectAsync(10, TimeSpan.FromMinutes(3), () =>
-            {
-                var mediator = DistributedPubSub.Get(Sys).Mediator;
-                var actor = Sys.ActorOf((dsl, context) =>
-                {
-                }, "childActor");
-                mediator.Tell(new Subscribe("pub-sub", actor));
-                _ = ExpectMsg<SubscribeAck>();
-                mediator.Tell(new Unsubscribe("pub-sub", actor));
-                _ = ExpectMsg<UnsubscribeAck>();
-                Sys.Scheduler.Advanced.ScheduleOnce(TimeSpan.FromMinutes(2.50), () => 
-                { 
-                    for(var i = 0; i < 10; i++)
-                    {
-                        mediator.Tell(new Publish("pub-sub", $"Good {i}"));
-                    }
-                });
+            // arrange
+            var mediator = DistributedPubSub.Get(Sys).Mediator;
+            var actor = Sys.ActorOf((dsl, context) => { }, "childActor");
 
-            });            
+            // act
+            // create a topic
+            mediator.Tell(new Subscribe("pub-sub", actor));
+            _ = ExpectMsg<SubscribeAck>();
+
+            // all subscribers should be removed from this topic
+            // topic actor will still be alive for default value of 120s
+            mediator.Tell(new Unsubscribe("pub-sub", actor));
+            _ = ExpectMsg<UnsubscribeAck>();
+
+            // assert
+
+            // slightly racey, since the 0-subscriber topics are removed on a scheduled task that runs once per second
+            // hence why we use AwaitAssert to re-run the assertion
+            await EventFilter.DeadLetter<object>().ExpectAsync(1,
+                () => { mediator.Tell(new Publish("pub-sub", $"hit")); });
         }
     }
+
     public sealed class QueryTopics
     {
         public static QueryTopics Instance = new QueryTopics();
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
@@ -14,6 +14,7 @@ using Akka.TestKit;
 using Xunit;
 using Akka.Event;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace Akka.Cluster.Tools.Tests.PublishSubscribe
 {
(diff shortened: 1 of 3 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
```diff
@@ -54,5 +56,58 @@ public void DistributedPubSubMediator_can_be_activated_in_child_actor()
...
+                        actorContext.System.Log.Info($"Received deadletter {s.Message}");
+                    }); 
+                    dsl.Receive<UnsubscribeAck>((s, actorContext) =>
+                    {
+                        actorContext.System.Log.Info($"{s.Unsubscribe.Topic}");
+                        cancelable = context.System.Scheduler.ScheduleTellOnceCancelable(TimeSpan.FromMinutes(3), context.Self, PublishTopic.Instance, context.Self);
+                    });
+
+                }, "childActor");
+                Thread.Sleep(TimeSpan.FromMinutes(5));
```
Aaronontheweb, line 100: No need to sleep the thread for 5 minutes

src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
```diff
@@ -54,5 +56,58 @@ public void DistributedPubSubMediator_can_be_activated_in_child_actor()
...
+            EventFilter.Exception<NullReferenceException>().Expect(0, () =>
+            {
+                var actor = Sys.ActorOf((dsl, context) =>
+                {
+                    IActorRef mediator = null;
+                    ICancelable cancelable = null;
+                    dsl.OnPreStart = actorContext =>
+                    {
+                        Sys.EventStream.Subscribe(context.Self, typeof(DeadLetter));
+                        mediator = DistributedPubSub.Get(actorContext.System).Mediator;
```
Aaronontheweb, line 72: move the "arrange" steps to outside of the actor, to make things more comprehensible

src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
```diff
@@ -54,5 +56,58 @@ public void DistributedPubSubMediator_can_be_activated_in_child_actor()
...
+                    };
+
+                    dsl.Receive<string>(s => s.Equals("UnSub"), (s, actorContext) =>
+                    {
+                        mediator.Tell(new Unsubscribe("pub-sub", context.Self));
+                    }); 
+                    dsl.Receive<SubscribeAck>((s, actorContext) =>
+                    {
+                        actorContext.System.Log.Info($"{s.Subscribe.Topic}");
+                        cancelable = context.System.Scheduler.ScheduleTellOnceCancelable(TimeSpan.FromSeconds(5), context.Self, "UnSub", context.Self);
```
Aaronontheweb, line 83: No need to use the scheduler here - use `ExpectMsg<T>` outside the actor for the test

src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
```diff
@@ -54,5 +56,58 @@ public void DistributedPubSubMediator_can_be_activated_in_child_actor()
...
+                    }); 
+                    dsl.Receive<PublishTopic>((s, actorContext) =>
+                    {
+                        mediator.Tell(new Publish("pub-sub", "Good"));
+                    }); 
+                    dsl.Receive<DeadLetter>((s, actorContext) =>
+                    {
+                        actorContext.System.Log.Info($"Received deadletter {s.Message}");
```
Aaronontheweb, line 91: Assert that a `DeadLetter` was raised via the `EventFilter`.

Aaronontheweb (changes requested):
Left some comments on the structure of the test

src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
```diff
@@ -58,47 +59,27 @@ public void DistributedPubSubMediator_can_be_activated_in_child_actor()
         }
 
         [Fact]
-        public void DistributedPubSubMediator_should_get_dead_letter()
+        public async Task DistributedPubSubMediator_should_send_messages_to_dead_letter()
         {
-            EventFilter.Exception<NullReferenceException>().Expect(0, () =>
+            await EventFilter.DeadLetter<object>().ExpectAsync(10, TimeSpan.FromMinutes(3), () =>
```
eaba, line 64: The default `removed-time-to-live` is 2 minutes - that is when pruning takes effect if there are no subscribers to a Topic. Hence, the reason for the `TimeSpan.FromMinutes(3)`

src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
```diff
@@ -58,47 +59,27 @@ public void DistributedPubSubMediator_can_be_activated_in_child_actor()
...
-                    {
-                        actorContext.System.Log.Info($"Received deadletter {s.Message}");
-                    }); 
-                    dsl.Receive<UnsubscribeAck>((s, actorContext) =>
+                }, "childActor");
+                mediator.Tell(new Subscribe("pub-sub", actor));
+                _ = ExpectMsg<SubscribeAck>();
+                mediator.Tell(new Unsubscribe("pub-sub", actor));
+                _ = ExpectMsg<UnsubscribeAck>();
+                Sys.Scheduler.Advanced.ScheduleOnce(TimeSpan.FromMinutes(2.50), () =>
```
eaba, line 74: The default `removed-time-to-live` is 2 minutes - that is when pruning takes effect if there are no subscribers to a Topic. Hence, the reason for the `TimeSpan.FromMinutes(2,50)`

eaba (commented):
@Aaronontheweb improved the test with the latest commit

src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubMediatorSpec.cs
```diff
@@ -54,5 +57,40 @@ public void DistributedPubSubMediator_can_be_activated_in_child_actor()
...
+            mediator.Tell(new Subscribe("pub-sub", actor));
+            _ = ExpectMsg<SubscribeAck>();
+
+            // all subscribers should be removed from this topic
+            // topic actor will still be alive for default value of 120s
+            mediator.Tell(new Unsubscribe("pub-sub", actor));
+            _ = ExpectMsg<UnsubscribeAck>();
+
+            // assert
+            await EventFilter.DeadLetter<object>().ExpectAsync(1,
```
Aaronontheweb, line 82: Reproduction - as @eaba pointed out this spec would fail before because the TTL value kept the topic actor alive for 120s. The real bug was that the topic actor never published any `DeadLetter` instances if it was alive but had no subscribers.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -643,7 +643,7 @@ private long NextVersion()
 
         private IActorRef NewTopicActor(string encodedTopic)
         {
-            var t = Context.ActorOf(Actor.Props.Create(() => new Topic(_settings.RemovedTimeToLive, _settings.RoutingLogic)), encodedTopic);
+            var t = Context.ActorOf(Actor.Props.Create(() => new Topic(_settings.RemovedTimeToLive, _settings.RoutingLogic, _settings.SendToDeadLettersWhenNoSubscribers)), encodedTopic);
```
Aaronontheweb, line 646: Propagate `akka.cluster.pub-sub.send-to-dead-letters-when-no-subscribers` to the topic actors, who are the ones who actually need to implement this.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
```diff
@@ -11,28 +11,53 @@
...
 using Akka.Routing;
 
 namespace Akka.Cluster.Tools.PublishSubscribe.Internal
 {
     /// <summary>
-    /// TBD
+    /// A <see cref="DeadLetter"/> published when there are no subscribers
+    /// for a topic that has received a <see cref="Publish"/> event.
+    /// </summary>
+    internal readonly struct NoSubscribersDeadLetter
```
Aaronontheweb, line 24: Created an envelope `NoSubscribersDeadLetter` in order to help make it clear inside the user's logging system why this `DeadLetter` was published. Created it as a `readonly struct` because it will immediately be captured inside a `DeadLetter` and therefore we can save on an allocation here in the event of a large number of deadletters being published all at once.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
```diff
@@ -44,31 +69,39 @@ internal abstract class TopicLike : ActorBase
...
+
+        /// <summary>
+        /// Creates a new instance of a topic or group actor.
         /// </summary>
-        /// <param name="emptyTimeToLive">TBD</param>
-        protected TopicLike(TimeSpan emptyTimeToLive)
+        /// <param name="emptyTimeToLive">The TTL for how often this actor will be removed.</param>
+        /// <param name="sendToDeadLettersWhenNone">When set to <c>true</c>, this actor will
+        /// publish a <see cref="DeadLetter"/> for each message if the total number of subscribers == 0.</param>
+        protected TopicLike(TimeSpan emptyTimeToLive, bool sendToDeadLettersWhenNone)
```
Aaronontheweb, line 82: Updated constructor to receive new value. These are all `internal` classes so no breaking changes here.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
```diff
@@ -116,22 +151,23 @@ protected bool DefaultReceive(object message)
                 default:
                     foreach (var subscriber in Subscribers)
                         subscriber.Forward(message);
+
+                    // no subscribers
+                    if (Subscribers.Count == 0 && SendToDeadLettersWhenNoSubscribers)
```
Aaronontheweb, line 156: The bugfix: @eaba pointed out that this actor never was programmed to publish to deadletters when the subscriber count is zero, so we do this after the `foreach` loop (arbitrary) if we're configured for it. This causes the reproduction spec to now pass.

Aaronontheweb (commented):
Left some comments on the changes included in here

PR #5564, opened 2022-02-03
Docs: Update "Distributed Publish Subscribe in Cluster"

### LINKED ISSUE #5557: DistributedPubSub: add config toggle for publishing topics with no subscribers to deadletters
Related: https://github.com/akkadotnet/akka.net/issues/5352

It's not clear right now what happens when you publish to a topic that _never had any subscribers_ - whereas https://github.com/akkadotnet/akka.net/issues/5352 deals with what happens when a topic that previously had subscribers now no longer has any.

Both scenarios - no more subscribers and never had any subscribers need to have consistent behavior.

That behavior should also be toggleable by end users - I've just watched a user record tens of thousands of dead letters in a load test for a topic that never had any subscriber. I'm sure they would love to turn that off to free their test from noise.

i.e.

```
akka.cluster.akka.cluster.pub-sub.deadletters-for-empty-topics = on / off
```

eaba:
close #5557

Import HOCON properties with docfx code reference

### DIFF
--- a/docs/articles/clustering/distributed-publish-subscribe.md
+++ b/docs/articles/clustering/distributed-publish-subscribe.md
@@ -211,36 +211,7 @@ In the example above the mediator is started and accessed with the `Akka.Cluster
 
 The `DistributedPubSub` extension can be configured with the following properties:
 
-```hocon
-# Settings for the DistributedPubSub extension
-akka.cluster.pub-sub {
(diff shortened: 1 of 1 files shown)


Aaronontheweb (approved):
Thanks for bringing this page up to code

PR #5567, opened 2022-02-03
Updated configuration section with sample code

### LINKED ISSUE #5413: Docs: need to show how to pass custom configuration to Akka.TestKit
https://getakka.net/articles/actors/testing-actor-systems.html#configuration - it literally says

> TODO describe how to pass custom config

We should have filled this in by now.

eaba:
Fixes #5413
## Changes

Replaced TODO with sample code

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonConfigSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonConfigSpec.cs
@@ -5,6 +5,8 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+#region ClusterSingletonConfigSpec
+
 using System;
 using Akka.Cluster.Tools.Singleton;
 using Akka.Configuration;
@@ -54,3 +56,4 @@ namespace Akka.Cluster.Tools.Tests.Singleton
         }
     }
 }
+#endregion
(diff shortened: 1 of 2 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonConfigSpec.cs
```diff
@@ -5,6 +5,8 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+#region ClusterSingletonConfigSpec
```
Aaronontheweb, line 8: Don't want to include the _entire_ test here - only need the constructor.

Use the `//<{name}>` syntax to target a small area of code, rather than a `#region` 

See: https://getakka.net/community/contributing/documentation-guidelines.html#code-samples-must-use-code-references

Aaronontheweb (changes requested):
Need to fix the DocFx reference

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonConfigSpec.cs
```diff
@@ -15,6 +15,7 @@
 
 namespace Akka.Cluster.Tools.Tests.Singleton
 {
+    // <ClusterSingletonConfigSpec>
```
eaba, line 18: I am showing how the Test Class is derived from the `TeskKit`, and the custom config, `GetConfig()`, itself so that there are confusion about the setup!

Aaronontheweb: LGTM

Aaronontheweb (approved):
LGTM

PR #5568, opened 2022-02-04
Fix ChannelExecutor configuration backward compatibility problem

### LINKED ISSUE #5541: Non-documented breaking change of channel-executor configuration
**Version Information**
Version of Akka.NET? 1.4.32
Which Akka.NET Modules? Akka

**Describe the bug**
After update from 1.4.20 to 1.4.32 channel-executor configuration (see [doc](https://getakka.net/articles/actors/dispatchers.html#channelexecutor)) is not valid anymore

**To Reproduce**
1. Create a console app with Akka 1.4.29+
2. Execute:
```
using Akka.Actor;

namespace ChannelExecutorIssue
{
    internal class Program
    {
        {
            // Following: https://getakka.net/articles/actors/dispatchers.html#channelexecutor
            var config = @"
                          akka.actor.default-dispatcher = { 
                                 executor = channel-executor
                          }";
            
            // Throws NRE in 1.4.29+
            var actorSystem = ActorSystem.Create("test", config);
        }
    }
}
```
3. Observe

**Expected behavior**
ActorSystem created successfully

**Actual behavior**
NRE is thrown:
```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Akka.Dispatch.ChannelExecutorConfigurator..ctor(Config config, IDispatcherPrerequisites prerequisites)
   at Akka.Dispatch.MessageDispatcherConfigurator.ConfigureExecutor()
   at Akka.Dispatch.DispatcherConfigurator..ctor(Config config, IDispatcherPrerequisites prerequisites)
   at Akka.Dispatch.Dispatchers.ConfiguratorFrom(Config cfg)
   at Akka.Dispatch.Dispatchers.LookupConfigurator(String id)
   at Akka.Dispatch.Dispatchers.Lookup(String dispatcherName)
   at Akka.Dispatch.Dispatchers..ctor(ActorSystem system, IDispatcherPrerequisites prerequisites, ILoggingAdapter logger)
   at Akka.Actor.Internal.ActorSystemImpl.ConfigureDispatchers()
   at Akka.Actor.Internal.ActorSystemImpl..ctor(String name, Config config, ActorSystemSetup setup, Nullable`1 guardianProps)
   at Akka.Actor.ActorSystem.CreateAndStartSystem(String name, Config withFallback, ActorSystemSetup setup)
   at Akka.Actor.ActorSystem.Create(String name, Config config)
   at ChannelExecutorIssue.Program.Main(String[] args)
```
   

The issue is in the default config for `default-dispatcher` that has no `channel-executor.priority` set. Introduced in #5403    

Sidenote: default `channel-scheduler` configuration is expected to be at: `akka.channel-scheduler` https://github.com/akkadotnet/akka.net/blob/ccb467076f061239aecc976983c6990c372811dd/src/core/Akka/Dispatch/ChannelSchedulerExtension.cs#L63
But it is actually located under `akka.actor.channel-scheduler`     
Current workaround:
```
using Akka.Actor;

namespace ChannelExecutorIssue
{
    internal class Program
    {
        {
            //Copy akka.actor.channel-scheduler from Akka/Configuration/Pigeon.conf
            var config = @"
                          akka.channel-scheduler {
                              parallelism-min = 4    #same as for ForkJoinDispatcher
                              parallelism-factor = 1 #same as for ForkJoinDispatcher
                              parallelism-max = 64   #same as for ForkJoinDispatcher
                              work-max = 10          #max executed work items in sequence until priority loop
                              work-interval = 500    #time target of executed work items in ms
                              work-step = 2          #target work item count in interval / burst
                          }
                          akka.actor.default-dispatcher = { 
                              executor = channel-executor
                              channel-executor.priority = normal
                          }";
            
            var actorSystem = ActorSystem.Create("test", config);
        }
    }
}
```

Arkatufus:
Fixes #5541

## Changes

Make sure that `channel-executor` configuration is injected properly when it is missing

### DIFF
--- a/src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
+++ b/src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
@@ -60,7 +60,7 @@ namespace Akka.Dispatch
         public ChannelTaskScheduler(ExtendedActorSystem system)
         {
             //config channel-scheduler
-            var config = system.Settings.Config.GetConfig("akka.channel-scheduler");
+            var config = system.Settings.Config.GetConfig("akka.actor.channel-scheduler");
             _maximumConcurrencyLevel = ThreadPoolConfig.ScaledPoolSize(
                         config.GetInt("parallelism-min"),
                         config.GetDouble("parallelism-factor", 1.0D), // the scalar-based factor to scale the threadpool size to 
@@ -349,7 +349,7 @@ namespace Akka.Dispatch
         /// and to help execute queued works internaly. 
         /// It supports task-inlining only for task equal or above the own priority
         /// </summary>
-        sealed class PriorityTaskScheduler : TaskScheduler, IDisposable
+        internal sealed class PriorityTaskScheduler : TaskScheduler, IDisposable
         {
             readonly Channel<Task> _channel;
 
--- a/src/core/Akka.Tests/Dispatch/ChannelExecutorConfigurationSpec.cs
+++ b/src/core/Akka.Tests/Dispatch/ChannelExecutorConfigurationSpec.cs
@@ -1,0 +1,67 @@
+﻿// //-----------------------------------------------------------------------
+// // <copyright file="ChannelExecutorConfigurationSpec.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Dispatch;
+using Akka.TestKit;
+using Xunit;
+using FluentAssertions;
+
+namespace Akka.Tests.Dispatch
+{
+    public class ChannelExecutorConfigurationSpec : AkkaSpec
+    {
+        [Fact]
+        public void ChannelExecutor_config_should_be_injected_when_it_doesnt_exist()
+        {
+            var config = ConfigurationFactory.ParseString(@"executor = channel-executor");
+            var configurator = new ChannelExecutorConfigurator(config, Sys.Dispatchers.Prerequisites);
+            configurator.Priority.Should().Be(TaskSchedulerPriority.Normal);
+        }
+
+        [Fact]
+        public void ChannelExecutor_default_should_be_overriden_by_config()
+        {
+            var config = ConfigurationFactory.ParseString(@"
+executor = channel-executor
+channel-executor.priority = high");
+            var configurator = new ChannelExecutorConfigurator(config, Sys.Dispatchers.Prerequisites);
+            configurator.Priority.Should().Be(TaskSchedulerPriority.High);
+        }
+
+        [Fact]
+        public void ChannelExecutorConfigurator_should_use_default_when_config_is_null()
+        {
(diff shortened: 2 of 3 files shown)


src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
```diff
@@ -60,7 +60,7 @@ public static ChannelTaskScheduler Get(ActorSystem system)
         public ChannelTaskScheduler(ExtendedActorSystem system)
         {
             //config channel-scheduler
-            var config = system.Settings.Config.GetConfig("akka.channel-scheduler");
+            var config = system.Settings.Config.GetConfig("akka.actor.channel-scheduler");
```
Aaronontheweb, line 63: LGTM

Aaronontheweb (approved):
LGTM

src/core/Akka/Dispatch/ChannelSchedulerExtension.cs
eaba: Is this correct? @Zetanova said the `channel-scheduler` is a standalone feature, does not belong to the actor config

eaba: He said `channel-scheduler` inside the actor config was a mistake.

Zetanova: The `channel-scheduler` is not directly related with actor or dispatchers
the `channel-executor` is using the `channel-scheduler` to queue work-items

Aaronontheweb: If we move it into `akka` - is that going to break current users on 1.4.32? Or since the config reading is already broken, should that not be an issue @Arkatufus ?

Arkatufus: It shouldn't be an issue since it doesn't work in the first place.

Arkatufus: If it doesn't have anything to do with actors and dispatchers, why is it in the Dispatch namespace?

Zetanova: > * If it doesn't have anything to do with actors and dispatchers, why is it in the Dispatch namespace?
The config was always planed to be under `akka.channel-scheduler`, I made simply a copy-past mistake 

> * Is there anything else in Akka that is using this beside actors?
The `ChannelTaskScheduler` is a singleton and currently only used by the new `channel-executor` dispatcher
It is thread safe and could be used in any place as a regular `TaskScheduler`

> * Maybe it shouldn't be a separate component?
It need to be a singleton

> * Is using the component pattern an anti-pattern? Is it over-engineering it a bit?
It need to be a singleton and it can be lazy initialized

> * Will putting it outside the `akka.actor` sub config makes it harder to explain this to users?
The `ChannelTaskScheduler` is a standalone component. 

>   * Because the name is ambiguous, what is "channel scheduler"?
Because the `channel-executor` was already named, I decided to skip on the `task` and use `channel`
the correct naming would be `channel-task-scheduler` or `priority-task-scheduler`

>   * Does it have anything to do with distributed pub sub? cluster?
the system and remote dispatcher has priority "high" as default to always priorities cluster heartbeats and the rest over regular tasks.

>   * What is a channel?
The word `channel` comes from the namespace used `System.Threading.Channels`

>   * Why do I have to configure something I want to use inside an actor so far away from the actor configuration itself?
The default config is just working fine, but of course should be explicitly configurable. 
I just copy-pasted the default block into the wrong section on my PR

PR #5571, opened 2022-02-04
Fix support for `serialization-identifier` identifier override

### LINKED ISSUE #5569: Setting `akka.actor.serialization-identifiers` did not override previous settings
**Version Information**
Version of Akka.NET? 1.4.32
Which Akka.NET Modules? Akka

**Describe the bug**
Reported in Gitter, setting `akka.actor.serialization-identifiers` did not override previous settings.
For example, this setting __SHOULD__ replace `Akka.Serialization.NewtonSoftJsonSerializer` serializer, but it does not.
```
akka.actor.serialization-identifiers {
    "My.ThrowingSerializer, MyAssembly" = 1
}
```

Arkatufus:
Fixes #5569

## Changes

Change the serialization initializer from using `Dictionary.Add()` to the index property, `Dictionary.Add()` does not like it when a key is inserted multiple times.

### DIFF
--- a/src/core/Akka/Serialization/Serialization.cs
+++ b/src/core/Akka/Serialization/Serialization.cs
@@ -337,8 +337,8 @@ namespace Akka.Serialization
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void AddSerializer(string name, Serializer serializer)
         {
-            _serializersById.Add(serializer.Identifier, serializer);
-            _serializersByName.Add(name, serializer);
+            _serializersById[serializer.Identifier] = serializer;
+            _serializersByName[name] = serializer;
         }
 
         /// <summary>
--- a/src/core/Akka.Tests/Serialization/CustomSerializerSpec.cs
+++ b/src/core/Akka.Tests/Serialization/CustomSerializerSpec.cs
@@ -87,6 +87,38 @@ namespace Akka.Tests.Serialization
             }
         }
         
+        // Fix for issue #5569, could not declare multiple serializer identifier
+        [Fact]
+        public void Configuration_should_be_able_to_override_serialization_identifiers()
+        {
+            var config = ConfigurationFactory.ParseString(@"
+                akka.actor {
+                    serializers {
+                        custom = ""Akka.Tests.Serialization.CustomThrowingSerializer, Akka.Tests""
+                    }
+                    serialization-bindings {
+                        ""Akka.Tests.Serialization.MessageBase, Akka.Tests"" = custom
+                    }
+                    serialization-identifiers {
+	                    ""Akka.Tests.Serialization.CustomThrowingSerializer, Akka.Tests"" = 1
+                    }
+                }
+            ");
+            
+            using (var system = ActorSystem.Create(nameof(CustomSerializerSpec), config))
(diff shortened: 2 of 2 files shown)


src/core/Akka/Serialization/Serialization.cs
```diff
@@ -337,8 +337,8 @@ public void AddSerializer(Serializer serializer)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void AddSerializer(string name, Serializer serializer)
         {
-            _serializersById.Add(serializer.Identifier, serializer);
-            _serializersByName.Add(name, serializer);
+            _serializersById[serializer.Identifier] = serializer;
```
eaba, line 340: It is likely going to introduce some breaking changes? If there are multiple serializer with same name, the last inserted will override the previous one - the effect is that the last one will win if there are two different serializer with same name.

```csharp
__serializersById = new Dictionary<string, List<Type>>();
``` 
I don't know much about `Serialization.cs`, but if  there are no other way to solve that problem, is that doable, is it "legal"?

eaba (commented):
Left you a comment @Arkatufus

src/core/Akka/Serialization/Serialization.cs
Arkatufus: That is the point, to enable people to override serializers

eaba: Okay. Will the custom serializer be registered last after the default serializers have been registered?

Aaronontheweb: I would like to add a warning when a serializer is being overridden, to help protect against accidental footguns. We should check here to see if this key is already contained in the collection or not and then log "you are replacing X with Y - did you mean to do this?" as a WARNING loglevel

Aaronontheweb (changes requested):
Would like to add a warning so people know they're doing this, if it happens by accident.

We should also add a test case where the serializer id I set programmatically (defined on the `Serializer` class as an overrideable property) and see if that also works.

Arkatufus:
Done, had to use stdout-logger because Serialization is set up before logger.

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -350,9 +375,21 @@ public void AddSerializer(string name, Serializer serializer)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void AddSerializationMap(Type type, Serializer serializer)
         {
+            if(_serializerMap.ContainsKey(type))
```
Aaronontheweb, line 378: This is the wrong place to log this, is it not? Isn't this our cache for which message type goes with which serializer?

Arkatufus: The configuration were parsed directly to these mapping dictionaries inside the constructor [here](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Serialization/Serialization.cs#L185) and [here](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka/Serialization/Serialization.cs#L192)

Aaronontheweb (approved):

PR #5542, opened 2022-01-26
Check for possible broken documentation by failing on DocFX warning

Arkatufus:
## Changes
Add --warningsAsErrors flag to DocFX argument

close https://github.com/akkadotnet/akka.net/issues/5528

### Warnings Log
1-26-2022: 754 Warning(s)
2-4-2022: 702 Warning(s)
2-7-2022: 590 Warning(s)
2-8-2022: 411 Warning(s)
2-9-2022: 312 Warning(s)
**2-11-2022: 41 Warning(s)**
2-15-2022: 25 Warning(s)

## NOTE
This PR __WILL__ fail until we fix all of the DocFX warnings in our documentation.

PR #5573, opened 2022-02-04
Deleted `inheritdoc` on external assemble's methods

eaba:
## Changes

Deleted `inheritdoc` for external assembly in `Akka.Streams.Shape.cs`


Aaronontheweb:
part of https://github.com/akkadotnet/akka.net/pull/5542

PR #5575, opened 2022-02-04
Docs: Fixed `inheritdoc` issue in `Akka.Streams.Implementation.StreamLayout.cs`

eaba:
## Changes

Deleted `inheritdoc` for external assembly in `Akka.Streams.Implementation.StreamLayout.cs`


Aaronontheweb:
part of https://github.com/akkadotnet/akka.net/pull/5542

PR #5584, opened 2022-02-07
Lock markdownlint-cli version to 0.31.0 and markdownlint-rule-titlecase to 0.1.0

### LINKED ISSUE #5583: CI: need to lock down version of markdown linter package
@eaba noticed on https://github.com/akkadotnet/akka.net/pull/5579 that we had all sorts of new markdown linting errors appear out of nowhere:

```
MD049/emphasis-style Emphasis style should be consistent [Expected: asterisk; Actual: underscore]
```

While that might be helpful, I prefer it if we didn't have a dependency that can be updated on its own and blow up our builds across ALL PRs. The version of the markdown linter package we install needs to be locked.

https://github.com/akkadotnet/akka.net/blob/9144d5445d88e4cd97518c00ba402ffc76a06e61/build-system/pr-validation.yaml#L51 - I think we just need to pass in a version number on the `npm install` command here.

Arkatufus:
Fixes #5583

## Changes

Lock build script dependency versions, markdownlint-cli version to 0.31.0 and markdownlint-rule-titlecase to 0.1.0

### DIFF
--- a/build-system/pr-validation.yaml
+++ b/build-system/pr-validation.yaml
@@ -48,7 +48,7 @@ jobs:
       - task: Npm@1
         inputs:
           command: "custom"
-          customCommand: "install -g markdownlint-cli markdownlint-rule-titlecase"
+          customCommand: "install -g markdownlint-cli@0.31.0 markdownlint-rule-titlecase@0.1.0"
       - task: CmdLine@2
         inputs:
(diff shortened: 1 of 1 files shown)


Aaronontheweb (approved):
LGTM

Issue #5586, opened 2022-02-07
InterceptAsync seems to run func() as a detached task


brah-mcdude:
This issue seems to be very similar to this issue: 
InternalExpectAsync does not await actionAsync() - causing actionAsync to run as a detached task #5537

Repro: 
This is my code: 

```c#
        protected async Task ExpectLogNoWarningsNorErrorsAsync(Func<Task> action)
        {
            await CreateEventFilter(Sys)
                .Custom(CustomErrorOrWarningFilter)
                .ExpectAsync(0, TimeSpan.FromSeconds(3),
                async () =>
                {
                    await action();
                });
        }
```

Invokes this code: 
```c#
        /// <summary>
        /// Async version of <see cref="Intercept{T}"/>
        /// </summary>
        protected Task<T> InterceptAsync<T>(Func<T> func, ActorSystem system, TimeSpan? timeout, int? expectedOccurrences, MatchedEventHandler matchedEventHandler = null)
        {
            return InterceptAsync(() => Task.FromResult(func()), system, timeout, expectedOccurrences, matchedEventHandler);
        }

```
Which seems to invoke: 
func()
without awaiting it, hence causing func() to run as a detached task. 

_Originally posted by @brah-mcdude in https://github.com/akkadotnet/akka.net/issues/5537#issuecomment-1031799197_

PR #5588, opened 2022-02-08
Intercept async detached func 5586

brah-mcdude:
Fixes issue #5586

A brief description of the changes:
The issue is similar to a previous issue (#5537). Here too, the callback ("func" in this case) is not getting awaited. 
I modified an interface by adding an optional timeout parameter. 
This way, I am forcing  the compiler to call the "correct" ExpectAsync version that does await func.

I also modified the interface implementation to reflect the change in the interface. 

I left notes for users to see that it is still possible to call ExpectAsync and have func not get awaited. 

@Aaronontheweb  - you tell me what you think. Obviously, this it not optimal. 

I wrote a test to reproduce the issue. 
This test failed as expected. 
In other words: The issue was getting reproduced as expected. 
Then, I fixed the issue. 
And then I saw that the test passes as expected.

Compatibility: 

[I think so] This change follows the [Akka.NET API Compatibility Guidelines]

[Yes] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).

[Change in Akka.TestKit Interface change - not reviewed] Changes in public API reviewed, if any.

Included data from after this change here:

Output from visual studio tests: 

﻿Akka.TestKit.Tests (net6.0)
  Tests in group: 255
   Total Duration: 57.2 sec

Outcomes
   254 Passed
   1 Skipped


src/core/Akka.TestKit/EventFilter/IEventFilterApplier.cs
```diff
@@ -97,7 +97,8 @@ public interface IEventFilterApplier
         /// </summary>
         /// <param name="expectedCount">The expected number of events</param>
         /// <param name="actionAsync">The async action.</param>
-        Task ExpectAsync(int expectedCount, Func<Task> actionAsync);
+        /// <param name="timeout"></param>
+        Task ExpectAsync(int expectedCount, Func<Task> actionAsync, TimeSpan? timeout = null);
```
Aaronontheweb, line 101: Breaking change here - need to add a method overload instead.

Aaronontheweb (changes requested):
Need to add a method overload rather than introducing a new optional parameter, per https://getakka.net/community/contributing/api-changes-compatibility.html cc @Arkatufus

src/core/Akka.TestKit/EventFilter/IEventFilterApplier.cs
Arkatufus: Fixed

PR #5589, opened 2022-02-08
simplified test. i checked that it properly reproduces the issue.

brah-mcdude:
No Fixes. I only simplified a test that I wrote. 

A brief description of the change:
The previous test was very complicated. 
I realized that I can simplify it. 

This is not a significant change.

I ran all the tests. Here is the output: 

﻿Akka.TestKit.Tests (net6.0)
  Tests in group: 247
   Total Duration: 51 sec

Outcomes
   246 Passed
   1 Skipped


brah-mcdude:
This is the original issue for which this test was created.

PR #5593, opened 2022-02-08
Docs: Fix `Akka.Util` unresolved base documentaion

eaba:
Part fix for #5542

## Changes


Aaronontheweb (approved):
LGTM

PR #5606, opened 2022-02-08
Docs: Fix unresolved base documentaion 

eaba:
Part fix for #5542

## Changes


src/core/Akka/Actor/LocalActorRef.cs
```diff
@@ -110,7 +110,7 @@ public override IActorRefProvider Provider
             get { return _cell.SystemImpl.Provider; }
         }
 
-        /// <inheritdoc cref="IInternalActorRef"/>
```
Aaronontheweb, line 113: Shouldn't this, in theory, be correct? This is an internal CREF - it's external ones that have been failing.

Aaronontheweb (changes requested):
Are we sure this is right?

Aaronontheweb (approved):

PR #5610, opened 2022-02-08
Docs: Fix unresolved base documentaion WARNINGS in `Remote*.cs`

eaba:
Part fix for #5542

## Changes

### DIFF
--- a/src/core/Akka.Remote/RemoteActorRef.cs
+++ b/src/core/Akka.Remote/RemoteActorRef.cs
@@ -154,7 +154,7 @@ namespace Akka.Remote
             SendSystemMessage(new Akka.Dispatch.SysMsg.Suspend());
         }
 
-        /// <inheritdoc cref="IInternalActorRef"/>
+        
         public override bool IsLocal => false;
 
         /// <inheritdoc cref="IActorRef"/>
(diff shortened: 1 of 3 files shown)


src/core/Akka.Remote/RemoteActorRef.cs
```diff
@@ -154,7 +154,7 @@ public override void Suspend()
             SendSystemMessage(new Akka.Dispatch.SysMsg.Suspend());
         }
 
-        /// <inheritdoc cref="IInternalActorRef"/>
```
Aaronontheweb, line 157: This is an internal CREF, which should work fine. If we're resolving to the wrong base class we just need to change that, not delete it.

Aaronontheweb (changes requested):
We should try to fix internal CREFs.

src/core/Akka.Remote/RemoteActorRef.cs
eaba: `/// <inheritdoc/>` could suffice since the base `IsLocal` has `/// <inheritdoc cref="IInternalActorRef"/>` tag?

Aaronontheweb: Yeah that's fine I think

PR #5647, opened 2022-02-12
Change WatchTermination signature to Task<Done>

ismaelhamed:
Fixes `WatchTermination` signature.

## Changes

Change ` WatchTermination` signature from `Task` to `Task<Done>`. I don't think there's a way to introduce this change without breaking binary compatibility.

## Checklist


Aaronontheweb:
Why is this change needed exactly? Is there a bug that this resolves?

ismaelhamed:
> Why is this change needed exactly? Is there a bug that this resolves?

There's an old PR in JVM where they went from `Future` to `Future[Done]` everywhere. I tried to find it to link it here, but couldn't find it. I'll give it another shot.

So now it is impossible to "correctly" write some streams from the JVM in Akka.NET (actually, the other PR I sent goes in the same direction). Also, an empty `Task` is not good enough to signal completion of streams, for the same reason we return `Option<>` instead of `null`.

Aaronontheweb:
> So now it is impossible to "correctly" write some streams from the JVM in Akka.NET (actually, the other PR I sent goes in the same direction). Also, an empty Task is not good enough to signal completion of streams, for the same reason we return Option<> instead of null.

Got it. Ok. Even though this is a "breaking change" it's minor enough and in an area of the API that is sufficiently low-traffic to a point where I think this is fine.

PR #5648, opened 2022-02-12
Reorder Source/FlowWithContext type parameters

ismaelhamed:
Fixes `FlowWithContext` parameters order to follow actual shape of elements

## Changes

Reorder Source/FlowWithContext type parameters

## Checklist

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/FlowWithContextSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/FlowWithContextSpec.cs
@@ -1,0 +1,80 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="FlowWithContextSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2021 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2021 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Linq;
+using Akka.Streams.Dsl;
+using Akka.Streams.TestKit;
+using Akka.TestKit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Streams.Tests.Dsl
+{
+    public class FlowWithContextSpec : AkkaSpec
+    {
+        private ActorMaterializer Materializer { get; }
+
+        public FlowWithContextSpec(ITestOutputHelper helper) : base(helper)
+        {
+            var settings = ActorMaterializerSettings.Create(Sys);
+            Materializer = ActorMaterializer.Create(Sys, settings);
+        }
+
+        [Fact]
+        public void A_FlowWithContext_must_get_created_from_FlowAsFlowWithContext()
+        {
+            var flow = Flow.Create<Message>().Select(m => m.Copy(data: m.Data + "z"));
+            var flowWithContext = flow.AsFlowWithContext<Message, long, Message, long, NotUsed, Message>((m, o) => new Message(m.Data, o), m => m.Offset);
+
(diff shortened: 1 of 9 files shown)


src/core/Akka.Streams.Tests/Dsl/FlowWithContextSpec.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            var settings = ActorMaterializerSettings.Create(Sys);
+            Materializer = ActorMaterializer.Create(Sys, settings);
+        }
+
+        [Fact]
+        public void A_FlowWithContext_must_get_created_from_FlowAsFlowWithContext()
+        {
+            var flow = Flow.Create<Message>().Select(m => m.Copy(data: m.Data + "z"));
+            var flowWithContext = flow.AsFlowWithContext<Message, long, Message, long, NotUsed, Message>((m, o) => new Message(m.Data, o), m => m.Offset);
```
ismaelhamed, line 32: Why can't the compiler infer the type arguments? This should be just: `flow.AsFlowWithContext((m, o) => new Message(m.Data, o), m => m.Offset);`

ismaelhamed:
Unrelated test failing

Arkatufus:
Though a good idea, this is a giant public API change with no possible backward compatibility with the old API. We just can't have this kind of change without having other users to potentially have their code suddenly broken when they update their Akka.NET version.

ismaelhamed:
Agree, but this is the actual API in the JVM. So, we need to figure out a way to be able to correct things. Akka.NET is still missing at least 3 more PRs related to `Source/FlowWithContext`, and they need this API in place first.

src/core/Akka.Streams/Dsl/FlowWithContext.cs
```diff
@@ -17,28 +18,27 @@ namespace Akka.Streams.Dsl
...
+
         ///<summary>
         /// Transform this flow by the regular flow. The given flow must support manual context propagation by
         /// taking and producing tuples of (data, context).
         /// 
         /// This can be used as an escape hatch for operations that are not (yet) provided with automatic
         /// context propagation here.
         ///</summary>
-        public FlowWithContext<TCtxIn, TIn, TCtx2, TOut2, TMat> Via<TCtx2, TOut2, TMat2>(
+        public FlowWithContext<TIn, TCtxIn, TOut2, TCtx2, TMat> Via<TOut2, TCtx2, TMat2>(
```
eaba, line 38: Instead of changing the existing APIs, can we make it call the new API? We can adapt the existing API @ismaelhamed ? 

This applies to all other API changes. We can deprecate the old ones and have it call the new ones, that is, old over new. Is that possible?

eaba (commented):
@ismaelhamed We can have the old APIs call the new APIs? To avoid introducing any compatibility issues

ismaelhamed:
@eaba I don't think so as this PR touches the public APIs, but feel free to prove me wrong. 

All this code should've been annotated with `ApiMayChange` -since they were not stable APIs at the time of the original PR- and now it wouldn't be such an issue.

Aaronontheweb:
So in terms of this API change, help me understand - what issue is it fixing exactly? Does the `FlowWithContext` stage not function correctly in its current implementation because the parameters are wrong?

ismaelhamed:
@Aaronontheweb per [#26345](https://github.com/akka/akka/issues/26345), this PR fixes the type parameters order in `FlowWithContext`, and adds the missing `FlowWithContextSpec`.

With this API in place, we can keep adding the rest of the Source/Flow WithContext features (MapMaterializedValue and Throttle).

Aaronontheweb:
Got it - thanks!

Aaronontheweb (approved):
LGTM

PR #5649, opened 2022-02-12
Expect log no warnings nor errors async

brah-mcdude:
Fixes nothing. This is a new feature.

Changes:
1. I added a new public API ExpectLogNoWarningsNorErrorsAsync. 
2. I added 3 tests: 
   1. [Test 1] logging info and debug should not cause ExpectLogNoWarningsNorErrorsAsync to fail.
   2. ExpectLogNoWarningsNorErrorsAsync should fail when logging:
      1. [Test 2] a warning
      2. [Test 3] an error

A brief description of the changes:

1. ExpectLogNoWarningsNorErrorsAsync as its name suggests - expects the log to contain no warnings nor errors and it is an async call. 
2. I have been using ExpectLogNoWarningsNorErrorsAsync for a few months now. I think that this is a critical feature. I include it in literally all my tests. I honestly don't understand how one can run a test without asserting that the log has no warnings nor errors.

Checklist:

This is not a significant change. 

* [yes] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [yes] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [yes] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [no] Design discussion
* [none] Changes in public API reviewed, if any.
* [what are these?] I have added website documentation for this feature.

Data from after this change:

﻿Akka.TestKit.Tests (net6.0)
  Tests in group: 265
   Total Duration: 1.2 min

Outcomes
   264 Passed
   1 Skipped

Note: 
I placed the tests in the file: src/core/Akka.TestKit.Tests/TestEventListenerTests/ExceptionEventFilterTests.cs. 
However, these tests have nothing to do with ExceptionEventFilterTests. But, I found this file to be the most appropriate. 
I recommend that we add at least one more file for my new tests: either a MiscellaneousTests.cs file or ExpectLogNoWarningsNorErrors.cs file.  And then move these tests there.


brah-mcdude:
@Aaronontheweb - what do you think?

Aaronontheweb:
> @Aaronontheweb - what do you think?

I'll take a look!

Arkatufus:
@brah-mcdude 
There are 2 ways that this function can be achieved right now, one is through the event filter as you have shown and the other is by using a TestProbe and subscribing to the ActorSystem event bus for Warning and Error event log messages and then making sure that no such logs were emmited using ExpectNoMsg.

Though useful, I think we don't quite need this addition to the public API. It is better to keep the public API as concise and simple as possible.

Aaronontheweb:
> @brah-mcdude There are 2 ways that this function can be achieved right now, one is through the event filter as you have shown and the other is by using a TestProbe and subscribing to the ActorSystem event bus for Warning and Error event log messages and then making sure that no such logs were emmited using ExpectNoMsg.
> 
> Though useful, I think we don't quite need this addition to the public API. It is better to keep the public API as concise and simple as possible.

Isn't there a fluent way to do this through the `EventFilter` API?

brah-mcdude:
> > @brah-mcdude There are 2 ways that this function can be achieved right now, one is through the event filter as you have shown and the other is by using a TestProbe and subscribing to the ActorSystem event bus for Warning and Error event log messages and then making sure that no such logs were emmited using ExpectNoMsg.
> > Though useful, I think we don't quite need this addition to the public API. It is better to keep the public API as concise and simple as possible.
> 
> Isn't there a fluent way to do this through the `EventFilter` API?

I am not familiar with such a way.

Arkatufus:
@brah-mcdude You're right, the fluent builder pattern is broken, you're supposed to be able to chain filters together like this:
```
var filter = CreateEventFilter(Sys)
    .ForLogLevel(logLevel: LogLevel.Warning, source: "MyWarningActor") // warning source was an actor class named "MyWarningActor"
    .ForLogLevel(logLevel: LogLevel.Error, start: "MyErrorTag"); // error message starts with "MyErrorTag"

await filter.ExpectAsync(0, actionAsync, timeout);
```
apparently, its not designed correctly.

brah-mcdude:
> @brah-mcdude There are 2 ways that this function can be achieved right now, one is through the event filter as you have shown and the other is by using a TestProbe and subscribing to the ActorSystem event bus for Warning and Error event log messages and then making sure that no such logs were emmited using ExpectNoMsg.
> 
> Though useful, I think we don't quite need this addition to the public API. It is better to keep the public API as concise and simple as possible.

@Arkatufus - Shouldn't ALL of akka.net's tests include an assert such as: ExpectLogNoWarningsNorErrorsAsync?

Arkatufus:
Ah, I was looking at the wrong code, sorry. This is how you create a message filter using the build pattern:

```
var filter = EventFilter // this is the built in event filter factory property inside TestKitBase
    .Warning(source: "MyWarningActor") // warning source was an actor class named "MyWarningActor"
    .And.Error(start: "MyErrorTag"); // error message starts with "MyErrorTag"

await filter.ExpectAsync(0, actionAsync, timeout);
```

Arkatufus:
Yes, it is how we use it inside our unit tests:
https://github.com/akkadotnet/akka.net/blob/700f33d5d3b47ba75c21f85ccbba0b5e78bd1f84/src/core/Akka.Tests/Actor/SupervisorHierarchySpec.cs#L270-L302

Aaronontheweb:
Ok, so does that syntax work for you @brah-mcdude ?

brah-mcdude:
> Ok, so does that syntax work for you @brah-mcdude ?

Actually, no. I am looking for a simple one-liner that I can add to all my tests like this:

```c#
        [Fact]
        public async Task Some_class_should_do_something()
        {
            await ExpectLogNoWarningsNorErrorsAsync(async delegate
            {
                /// my test that some class should do something
            });
        }
```

Also, currently, all of akka.net tests do not check the log for warnings, errors and exceptions. There might be tests that are "passing" while they are actually writing failure messages to the log. 

I think that adding ExpectLogNoWarningsNorErrorsAsync to all akka.net tests would be a good idea.  This would add much value to each test, at the cost of one line of code. 

And if we do this, we should add ExpectLogNoWarningsNorErrorsAsync to the public API.

Aaronontheweb:
I feel that the compositions we allow in the `EventFilter` today are sufficient for this purpose and are flexible enough to allow for things like muting expected errors / warnings. Your API doesn't allow that as it hides the `EventFilter` so it can't be extended through additional calls.

It's possible for errors and warnings to be raised in our tests especially when we're networking. That's why you'll see a number of `AwaitAssert` calls where things like the second `ActorSystem`'s socket being open and fully associated is non-deterministic at the point of the assertion. That'll log a `DeadLetter` warning if that message doesn't get delivered, but it's ultimately not harmful to the test.

I'm still mulling over whether or not I would want to add this to the TestKit's public API - I'm leaning towards "no" simply because I feel like our composable API is a better tool for doing this for the reasons I stated above.

PR #5650, opened 2022-02-13
Add Stateful methods for circuitbreaker

### LINKED ISSUE #5512: Perf: add stateful `CircuitBreaker`
Should we consider adding a `WithCircuitBreaker<TState,TResult>` overload to the circuit breakers to get rid of an allocation here? (Pass `this` and `message` in, get vaue of `readHighestSequenceNrFrom` inside the function:)

```
var highSequenceNr = await _breaker.WithCircuitBreaker((this,message),(state) =>
                        ReadHighestSequenceNrAsync(message.PersistenceId, Math.Max(0L, message.FromSequenceNr - 1)));
```

That should? actually also get rid of one local on top of the closure allocation in question.

OTOH, this is the read side, which isn't as hot as the write side.

_Originally posted by @to11mtm in https://github.com/akkadotnet/akka.net/pull/5505#discussion_r783996956_

to11mtm:
Fixes #5512 

## Changes

Added overloads to CircuitBreaker that take a `TState` object in, allowing lower allocations in calls.

Note: I'm not quite sure yet what the best way to handle `WithSyncCircuitBreaker` will be. This -does- actually let us get rid of an allocation on them, but I'm still trying to figure out a good way to optimize that.


src/core/Akka/Util/Internal/AtomicState.cs
```diff
@@ -113,6 +113,38 @@ public async Task<T> CallThrough<T>(Func<Task<T>> task)
             }
             return result;
         }
+        
+        public async Task<T> CallThrough<T,TState>(TState state, Func<TState,Task<T>> task)
```
Aaronontheweb, line 117: Safe to use `ValueTask` here? I don't know of any instances where multiple `await`ers await on the same `CircuitBreaker` at the same time off the top of my head. Maybe @ismaelhamed does?

ismaelhamed: You can call `WithSyncCircuitBreaker` multiple times without waiting for the previous ones to finish, if that was the question. I'm thinking CircuitBreaker + PipeTo.

Aaronontheweb: The `ValueTask` constraint I'm thinking of here is that you can't have multiple awaiters on the same `ValueTask` - if `WithCircuitBreaker` is called multiple times those are all different tasks right? Or is that method returning the same one each time possibly?

to11mtm: @Aaronontheweb I think there are spots right now where we `return` the task all the way up the callchain to `WithCircuitBreaker`, if so it would propagate up there into public API space.

If we want to further reduce allocations we may want to consider adapting something like PooledAwait (which uses special async method builders to avoid some of the state machine costs, also they have a fancy thing that replaces TCS, lowering allocation on most platforms with safe fallback if not.)

Aaronontheweb: Ok, so `ValueTask` is out then under those conditions. Got it.

Aaronontheweb:
@to11mtm is this ready to be taken out of draft state yet?

to11mtm:
@Aaronontheweb Lets start with this, I will likely do another PR shortly though.

to11mtm:
I should have some time Sunday to get API docs etc done.

Aaronontheweb:
Thanks @to11mtm - I look forward to reviewing it.

Aaronontheweb (changes requested):
API Approval specs are still failing

Aaronontheweb (approved):
LGTM

PR #5656, opened 2022-02-14
Channels implementation

eaba:
This PR makes use of System.Threading.Channels

### DIFF
--- a/src/core/Akka.TestKit/Internal/ChannelQueue.cs
+++ b/src/core/Akka.TestKit/Internal/ChannelQueue.cs
@@ -10,7 +10,9 @@ using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Channels;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace Akka.TestKit.Internal
 {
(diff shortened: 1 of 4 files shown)


src/core/Akka.TestKit/Internal/ChannelQueue.cs
```diff
@@ -20,33 +22,41 @@ namespace Akka.TestKit.Internal
...
         /// </summary>
-        public int Count { get { return _collection.Count; } }
+        public int Count { get { return _channel.Reader.Count; } }
 
         /// <summary>
         /// Adds the specified item to the end of the queue.
         /// </summary>
         /// <param name="item">The item to add to the queue.</param>
-        public void Enqueue(T item)
+        public async ValueTask Enqueue(T item)
```
Arkatufus, line 39: Incomplete implementation, should create a new method that accept a CancellationToken and figure out a way to guard this method with a time out mechanism

src/core/Akka.TestKit/Internal/ChannelQueue.cs
```diff
@@ -57,25 +67,25 @@ public void AddFirst(T item)
...
         /// </summary>
-        /// <param name="item">The item to remove from the queue.</param>
-        /// <returns><c>true</c> if the item was removed; otherwise, <c>false</c>.</returns>
-        public bool TryTake(out T item)
+        /// <returns><c>(true, T)</c> if the item was removed; otherwise, <c>(false, default)</c>.</returns>
+        public async ValueTask<(bool Success, T Item)> TryTake()
         {
-            if(_collection.TryTake(out var p))
+            var reader = _channel.Reader;   
+            if(reader.TryRead(out var p))
```
Arkatufus, line 84: This is also blocking

Arkatufus (changes requested):
This would not work, will need a major revision.

PR #5655, opened 2022-02-14
BufferBlock Implementation

eaba:
This PR makes use of BufferBlock

https://github.com/akkadotnet/akka.net/pull/5656

PR #5657, opened 2022-02-15
AsyncCollection Implementation

eaba:
As @Arkatufus  pointed out: "What we need is a collection that behaves like a queue on read but can write at the tail or the head of the queue". 

We were able to achieve this with `BlockingCollection<T>` by passing in a custom queue implementation via the constructor. AsyncEx's `AsyncCollection`, as the name suggests, can be considered a `BlockingCollection<T>` with `async/await` support and it is actively being maintained!

[`Channel<T>`](https://github.com/akkadotnet/akka.net/pull/5655) and [`BufferBlock<T>`](https://github.com/akkadotnet/akka.net/pull/5656) will require some customization and we cannot be sure of what edge cases we may end up dealing with. 

## AsyncQueue demo
image

### DIFF
--- a/src/core/Akka.TestKit/Akka.TestKit.csproj
+++ b/src/core/Akka.TestKit/Akka.TestKit.csproj
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <EmbeddedResource Include="Configs\TestScheduler.conf;Internal\Reference.conf" />
+    <PackageReference Include="Nito.AsyncEx" Version="5.1.2" />
     <ProjectReference Include="..\Akka\Akka.csproj" />
   </ItemGroup>
 
--- a/src/core/Akka.TestKit/Internal/AsyncQueue.cs
+++ b/src/core/Akka.TestKit/Internal/AsyncQueue.cs
@@ -11,6 +11,8 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
+using Nito.AsyncEx;
 
 namespace Akka.TestKit.Internal
 {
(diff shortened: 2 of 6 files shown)


src/core/Akka.TestKit/Internal/AsyncQueue.cs
```diff
@@ -86,15 +91,14 @@ public bool TryTake(out T item)
         /// <param name="millisecondsTimeout">The number of milliseconds to wait for the remove to complete.</param>
         /// <param name="cancellationToken">The cancellation token that can be used to cancel the operation.</param>
         /// <returns><c>true</c> if the remove completed within the specified timeout; otherwise, <c>false</c>.</returns>
-        public bool TryTake(out T item, int millisecondsTimeout, CancellationToken cancellationToken)
+        public async ValueTask<(bool Success, T Item)> TryTake(int millisecondsTimeout, CancellationToken cancellationToken)
         {
-            if(_collection.TryTake(out var p, millisecondsTimeout, cancellationToken))
+            if (await _collection.OutputAvailableAsync(cancellationToken))
```
eaba, line 97: Missing waiting for ` millisecondsTimeout` in case not message yet!

src/core/Akka.TestKit/Internal/AsyncQueue.cs
```diff
@@ -122,11 +122,8 @@ public async ValueTask<T> Take(CancellationToken cancellationToken)
         /// <returns>A <see cref="List{T}"/> containing copies of the elements of the collection</returns>
         public IEnumerable<T> ToList()
         {
-            while(_collection.OutputAvailable())
-            {
-                var p = _collection.Take(); 
-                yield return p.Value;   
-            }
+            foreach (var p in _queue.ToList())
```
eaba, line 125: Finally understood @Arkatufus's `will empty the queue.`

src/core/Akka.TestKit/Internal/AsyncQueue.cs
eaba: All the timeouts in my libraries are expressed as CancellationTokens https://github.com/StephenCleary/AsyncEx/issues/235#issuecomment-796238375

src/core/Akka.TestKit/Internal/AsyncQueue.cs
Aaronontheweb: This method really should return a `List<T>`

src/core/Akka.TestKit/Akka.TestKit.csproj
```diff
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <EmbeddedResource Include="Configs\TestScheduler.conf;Internal\Reference.conf" />
+    <PackageReference Include="Nito.AsyncEx" Version="5.1.2" />
```
Aaronontheweb, line 14: Not opposed to taking a dependency here

Issue #5609, opened 2022-02-08
[TEST] Update Akka.TestKit to run asynchronous internally.


Arkatufus:
Right now, `Akka.TestKit` relies on a `BlockingQueue` internally inside `TestState` making all test to run async over sync, In order to support full asynchronous test operations, we will need to replace this with something that supports async operation better. The main goal of this spec is to invert all operations into sync over async instead.

The biggest problem right now is the `InternalTryReceiveOne()` method in the `TestKitBase` class, it blocks as it waits for a message to get queued into the `BlockingQueue<T>`. This method gets called by virtually all akka test function that waits for one message or another. This method needs to be converted into an async method.

Options:
* Wrap `InternalTryReceiveOne()` inside an async function that uses `Task.Run()` to spin it in another thread.
```
private async Task<(bool, MessageEnvelope)> InternalTryReceiveOneAsync(TimeSpan? max, CancellationToken cancellationToken, bool shouldLog)
{
    return await Task.Run(() =>
    {
        var result = InternalTryReceiveOne(out var envelope, max, cancellationToken, shouldLog);
        return (result, envelope);
    }, cancellationToken);
}
```
* Replace the internal `BlockingQueue<T>` implementation with a custom `System.Threading.Channels.Channel<T>` that supports async read and write
* Replace the internal `BlockingQueue<T>` implementation with [AsyncEx](https://github.com/StephenCleary/AsyncEx) `AsyncCollection` or `AsyncProducerConsumerQueue`

brah-mcdude:
Please let me know if you want me to help. I need this feature more than anyone.

Arkatufus:
Will do, thanks for the offer.

Arkatufus:
All the code are done but can't really be tested until all of the unit tests are converted to the new async code

PR #5661, opened 2022-02-16
Add BlockingQueue.TakeAll to replace BlockingQueue.ToList

Arkatufus:
Related to #5609


Arkatufus:
Could not be done, having an emptying deadletter logging broke some MNTR tests.

PR #5675, opened 2022-02-21
[Experimental] Switch from using `Queue<T>` to `ConcurrentQueue<T>`

eaba:
## Changes

Replaced `Queue<T>` with `ConcurrentQueue<T>`

### DIFF
--- a/src/core/Akka.TestKit/Internal/AsyncQueue.cs
+++ b/src/core/Akka.TestKit/Internal/AsyncQueue.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


src/core/Akka.TestKit/Internal/AsyncQueue.cs
```diff
@@ -175,100 +176,62 @@ public List<T> ToList()
...
             IEnumerator IEnumerable.GetEnumerator()
             {
                 return GetEnumerator();
             }
 
-            public object SyncRoot { get; } = new object();
 
             public bool IsSynchronized => true;
+
+            public object SyncRoot => new object();
```
Aaronontheweb, line 234: should return a singleton, otherwise this method allocates

Aaronontheweb:
cc @Arkatufus

PR #5676, opened 2022-02-21
Switch internal implementation to async

eaba:
## Changes

- Created `InternalTryReceiveOneAsync` and `InternalTryPeekOneAsync`
- Changed the old `InternalTryReceiveOne` and `InternalTryPeekOne` methods to use their async counterpart by wrapping them in sync-over-async pattern.

### DIFF
--- a/src/core/Akka.TestKit/TestKitBase_Receive.cs
+++ b/src/core/Akka.TestKit/TestKitBase_Receive.cs
@@ -139,9 +139,15 @@ namespace Akka.TestKit
         /// <returns>The message if one was received; <c>null</c> otherwise</returns>
         public object ReceiveOne(TimeSpan? max = null)
         {
-            MessageEnvelope envelope;
-            if (TryReceiveOne(out envelope, max, CancellationToken.None))
-                return envelope.Message;
+            return ReceiveOneAsync(max).GetAwaiter().GetResult();
+        }
+
+        /// <inheritdoc cref="ReceiveOne(TimeSpan?)"/>
+        public async ValueTask<object> ReceiveOneAsync(TimeSpan? max = null)
+        {
+            var received = await TryReceiveOneAsync(max, CancellationToken.None).ConfigureAwait(false);
+            if (received.success)
+                return received.envelope.Message;
             return null;
         }
 
@@ -152,10 +158,16 @@ namespace Akka.TestKit
         /// <param name="cancellationToken">A token used to cancel the operation</param>
         /// <returns>The message if one was received; <c>null</c> otherwise</returns>
         public object ReceiveOne(CancellationToken cancellationToken)
+        {            
+            return ReceiveOneAsync(cancellationToken).GetAwaiter().GetResult();
+        }
+
+        /// <inheritdoc cref="ReceiveOne(CancellationToken)"/>
+        public async ValueTask<object> ReceiveOneAsync(CancellationToken cancellationToken)
         {
-            MessageEnvelope envelope;
-            if (TryReceiveOne(out envelope, Timeout.InfiniteTimeSpan, cancellationToken))
-                return envelope.Message;
+            var received = await TryReceiveOneAsync(Timeout.InfiniteTimeSpan, cancellationToken).ConfigureAwait(false);
+            if (received.success)
+                return received.envelope.Message;
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -139,9 +139,15 @@ public async Task<ArrayList> WaitForRadioSilenceAsync(TimeSpan? max = null, uint
         /// <returns>The message if one was received; <c>null</c> otherwise</returns>
         public object ReceiveOne(TimeSpan? max = null)
         {
-            MessageEnvelope envelope;
-            if (TryReceiveOne(out envelope, max, CancellationToken.None))
-                return envelope.Message;
+            return ReceiveOneAsync(max).GetAwaiter().GetResult();
```
Arkatufus: You should not do this, the spec only calls for changes on the `InternalTryReceiveOneAsync` and `InternalTryReceiveOne` methods. It was intentionally done that way to isolate changes in incremental way so we can catch problems as new things are introduced into the new system.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -315,50 +365,64 @@ public bool TryPeekOne(out MessageEnvelope envelope, TimeSpan? max = null)
...
+        /// <inheritdoc cref="TryPeekOne(out MessageEnvelope, TimeSpan?, CancellationToken)"/>
+        public async ValueTask<(bool success, MessageEnvelope envelope)> TryPeekOneAsync(TimeSpan? max, CancellationToken cancellationToken)
+        {
+            return await InternalTryPeekOneAsync(max, cancellationToken, true).ConfigureAwait(false);
+        }
 
         private bool InternalTryPeekOne(out MessageEnvelope envelope, TimeSpan? max, CancellationToken cancellationToken, bool shouldLog)
         {
-            bool didPeek;
+            var received = InternalTryPeekOneAsync(max, cancellationToken, shouldLog).GetAwaiter().GetResult();
```
Arkatufus, line 329:
```
var task = InternalTryPeekOneAsync(max, cancellationToken, shouldLog).AsTask().ConfigureAwait(false);
task.Wait();
var received = task.Result;
```

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -201,47 +220,61 @@ public bool TryReceiveOne(out MessageEnvelope envelope, TimeSpan? max, Cancellat
...
+        /// <inheritdoc cref="TryReceiveOne(out MessageEnvelope, TimeSpan?, CancellationToken)"/>
+        public async ValueTask<(bool success, MessageEnvelope envelope)> TryReceiveOneAsync(TimeSpan? max, CancellationToken cancellationToken)
+        {
+            return await InternalTryReceiveOneAsync(max, cancellationToken, true).ConfigureAwait(false);
+        }
+
         private bool InternalTryReceiveOne(out MessageEnvelope envelope, TimeSpan? max, CancellationToken cancellationToken, bool shouldLog)
         {
-            bool didTake;
+            var received = InternalTryReceiveOneAsync(max, cancellationToken, shouldLog).GetAwaiter().GetResult();
```
Arkatufus, line 205:
```
var task = InternalTryReceiveOneAsync(max, cancellationToken, shouldLog).AsTask().ConfigureAwait(false);
task.Wait();
var received = task.Result;
```

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -201,47 +220,61 @@ public bool TryReceiveOne(out MessageEnvelope envelope, TimeSpan? max, Cancellat
...
+        {
+            (bool didTake, MessageEnvelope env) take;
             var maxDuration = GetTimeoutOrDefault(max);
-            var start = Now;
+            var start = Now;    
             if (maxDuration.IsZero())
             {
                 ConditionalLog(shouldLog, "Trying to receive message from TestActor queue. Will not wait.");
-                didTake = _testState.Queue.TryTake(out envelope);
+                take = await _testState.Queue.TryTakeAsync(cancellationToken).ConfigureAwait(false);
```
Arkatufus, line 218: This case requires the test kit to return immediately with a value, regardless of the queue state. Don't use the async version of this method, use `TryTake` instead.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -315,50 +365,64 @@ public bool TryPeekOne(out MessageEnvelope envelope, TimeSpan? max = null)
...
+            (bool didPeek, MessageEnvelope env) peek;
             var maxDuration = GetTimeoutOrDefault(max);
             var start = Now;
-            
+
             if (maxDuration.IsZero())
             {
                 ConditionalLog(shouldLog, "Trying to peek message from TestActor queue. Will not wait.");
-                didPeek = _testState.Queue.TryPeek(out envelope);
+                peek = await _testState.Queue.TryPeekAsync(cancellationToken).ConfigureAwait(false);
```
Arkatufus, line 342: This case requires the test kit to return immediately with a value, regardless of the queue state. Don't use the async version of this method, use `TryPeek` instead.

Arkatufus (changes requested):
Need some changes

Arkatufus (approved):

PR #5684, opened 2022-02-23
Make PipeTo ConfigureAwait() optional

Arkatufus:
## Changes
Removed `ConfigureAwait(false)` inside the PipeTo extension

This change is needed for telemetry reasons, we need to preserve async context so that Actor messages inside an Ask can be correlated correctly inside a PipeTo.


Aaronontheweb (approved):
LGTM

src/core/Akka/Actor/PipeToSupport.cs
```diff
@@ -21,20 +21,30 @@ public static class PipeToSupport
...
+        /// <param name="useConfigureAwait">If set to true, <c>taskToPipe</c> will be awaited using <c>ConfigureAwait(false)</c></param>
         /// <returns>A detached task</returns>
-        public static async Task PipeTo<T>(this Task<T> taskToPipe, ICanTell recipient, IActorRef sender = null, Func<T, object> success = null, Func<Exception, object> failure = null)
+        public static async Task PipeTo<T>(
+            this Task<T> taskToPipe,
+            ICanTell recipient,
+            IActorRef sender = null, 
+            Func<T, object> success = null,
+            Func<Exception, object> failure = null,
+            bool useConfigureAwait = true)
```
Aaronontheweb, line 38: Have to add an overload. Can't do a new optional parameter.

Aaronontheweb (changes requested):

src/core/Akka/Actor/PipeToSupport.cs
Arkatufus: Adding an overload will actually break everything because it will create an abiguous method fingerprint

Aaronontheweb: The re-arrange the overload as a non-optional parameter that appears before the optional ones - that will resolve that.

Aaronontheweb (changes requested):
Still need to make this binary compatible.

src/core/Akka/Actor/PipeToSupport.cs
Aaronontheweb: We need to keep the current callsite unchanged - adding a new callsite with a different signature (new explicit boolean parameter) is no problem.

Arkatufus:
Done

Aaronontheweb (approved):
LGTM

PR #5688, opened 2022-02-25
Optimize PipeTo ConfigureAwait call

Arkatufus:


src/core/Akka/Actor/PipeToSupport.cs
```diff
@@ -60,9 +60,7 @@ public static class PipeToSupport
             
             try
             {
-                var result = useConfigureAwait
-                    ? await taskToPipe.ConfigureAwait(false)
-                    : await taskToPipe;
+                var result = await taskToPipe.ConfigureAwait(!useConfigureAwait);
```
Aaronontheweb, line 63: just to simplify this - change the parameter name to `configureAwait` and have it default to `false` - get rid of the double negative.

Aaronontheweb (approved):

PR #5702, opened 2022-03-04
[BUILD] Add targetBranch parameter for Incrementalist

Arkatufus:


Arkatufus:
Not doable at this time, there is no built-in variables for target git branch in Azure Pipelines

PR #5703, opened 2022-03-04
[TEST] Implement ExpectMsgFrom as async

Arkatufus:


src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -383,24 +371,13 @@ private void AssertPredicateIsTrueForMessage<T>(Predicate<T> isMessage, T m, str
...
-        }
-        
         private async ValueTask<T> InternalExpectMsgAsync<T>(
             TimeSpan? timeout,
             Action<T, IActorRef> assert,
             string hint,
             CancellationToken cancellationToken)
         {
-            var envelope = await InternalExpectMsgEnvelopeAsync(timeout, assert, hint, cancellationToken: cancellationToken);
+            var envelope = await InternalExpectMsgEnvelopeAsync(timeout, assert, hint, cancellationToken);
```
Aaronontheweb, line 380: Need `ConfigureAwait` here?

Aaronontheweb: Or not really, because this is just a thin wrapper `ValueTask`

Aaronontheweb (commented):
Just one small comment

PR #5704, opened 2022-03-04
[TEST] Fix and cleanup TestKit codes

Arkatufus:
* Convert all async methods that returns an `IReadOnlyCollection` to `IAsyncEnumerable`
* Clean sync over async codes to always call `.ConfigureAwait(false)`
* Clean async await codes to always call  `.ConfigureAwait(false)`
* Fix erroneous HOCON strings
* Add missing `base.AfterAll()` in `AbstractSerializationTransportInformationSpec`, leaving a leaked ActorSystem living in the background.

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Bugfix4360Spec.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Bugfix4360Spec.cs
@@ -22,30 +22,30 @@ namespace Akka.Persistence.Sqlite.Tests
     {
         public static Config TestConf = @"
 akka.persistence {
-journal {
-plugin = ""akka.persistence.journal.sqlite""
-sqlite {
-class = ""Akka.Persistence.Sqlite.Journal.SqliteJournal, Akka.Persistence.Sqlite""
-plugin-dispatcher = ""akka.actor.default-dispatcher""
-connection-string = ""DataSource=AkkaJournalfxR16.db""
-connection-timeout = 25s
-table-name = event_journal
-auto-initialize = on
-}
-}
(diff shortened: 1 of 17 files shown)


src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Bugfix4360Spec.cs
```diff
@@ -129,7 +129,7 @@ public void Should_recover_persistentactor_sqlite()
             recoveryActor.Tell("foo");
             recoveryActor.Tell("bar");
             recoveryActor.Tell(RecoverActor.DoSnapshot.Instance);
-            ExpectMsgAllOf("foo", "bar");
+            ExpectMsgAllOf(new []{ "foo", "bar" });
```
Aaronontheweb, line 132: Why is this necessary exactly? I thought the `params` handled this?

Aaronontheweb (commented):
See a lot of the same change here, so I need to understand why it was made before I can keep reviewing this PR

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Bugfix4360Spec.cs
Arkatufus: `params` need to be replaced with `IEnumerable<T>` because we need to pass in a `CancellationToken` with a default value

Aaronontheweb: Got it, thank you. Are we going to warn users about that somehow? Roslyn rule?

PR #5705, opened 2022-03-04
Fix AbstractSerializationTransportInformationSpec

Arkatufus:
## Changes

Call `base.AfterAll()` to kill TestKit ActorSystem. The unit test file overrides the `AfterAll()` method but failed to call the base class method. This causes the TestKit ActorSystem to leak and can freeze the test runner.
image
As can be seen from the screenshot, the "no currently active test" error got repeated over and over until the test runner can kill the process.


Aaronontheweb:
Nice find @Arkatufus

PR #5706, opened 2022-03-05
Fix confusing logging when receiving gossip from unknown

ismaelhamed:
Fixes logging when receiving gossip from unknown node.
Port: [#23883](https://github.com/akka/akka/pull/23883)


Aaronontheweb (approved):
LGTM

PR #5718, opened 2022-03-14
Convert Akka.Persistence.TestKit.Tests to async

Arkatufus:


Aaronontheweb:
Still getting deadlocks in here.

Arkatufus:
--blame-crash only available on .NET 5.0, so for .NET FX and .NET core 3.1 we can only use --blame

Arkatufus:
Deadlock fixed

Arkatufus:
Deadlock fixed

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -189,6 +188,30 @@ protected void InitializeTest(ActorSystem system, ActorSystemSetup config, strin
             _testState.TestActor = testActor;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static void WaitUntilTestActorIsReady(IActorRef testActor)
```
Aaronontheweb, line 192: Great catch - should we move this to its own PR so we can get a clean signal on this test library?

Arkatufus: Will do

Aaronontheweb:
Ok, let's see if this gives us a clean signal now - might have to do a rebase if not

Aaronontheweb:
Looks good to me @Arkatufus - onto the next suite

PR #5719, opened 2022-03-14
fix video embeds on mobile

Aaronontheweb:
Fixes https://twitter.com/Philco78/status/1501911099184238597 - video embeds on the new documentation website break the layout on mobile.


Aaronontheweb:
Validated that this works on my machine

PR #5720, opened 2022-03-14
Fixed `IActorRef` leak inside `EventStream`

### LINKED ISSUE #5717: EventStream subscription does not get automatically removed on actor death causing memory leak
Akka.Net v1.4.34 (was discovered in v1.4.17 as that's what we were using but verified to still exist in latest)

Running on .Net Core 3.1

If an Actor subscribes to a message Type on the EventStream and doesn't unsubscribe before dying the actor is garbage collected but the subscription remains and it keeps the ActorCell alive forever along with the Props used to create it.

In our app we have a set of Job actors that are loaded on demand and stay in memory until idle for a time at which point they kill themselves. This works great except the Props (ours are unique per instance due to a GUID in the parameters) and the ActorCell stay in memory forever due to the EventStream subscription never getting freed up. The EventStream subscription dictionaries continue to grow in size as well. We have 1000s of Job actors churning 24/7 so in a 12 hour period, I measured on average 2000 live Job actors at any given moment and 68,000 dead and gone at the end of the period. However 70,000 ActorCells and Props were still alive resulting in about 30MB/24 hours memory leakage at the current churn rate. The total memory leakage depends on the ActorCell overhead and the size of the Props as the contents of the Props are kept alive too - we had 7000 extra GUIDs alive after 12 hours too due to this as they are in the Props.

To repro this I wrote a test program that slowly creates and then slowly kills 100 actors (slow so I can see what's going on and snapshot it easily) and can be configured to either not subscribe at all, subscribe in PreStart and unsubscribe in PostStop or to only subscribe and not unsubscribe. I monitored the memory with SciTech Memory Profiler snapshotting before, during and after for each case.

If set to either not subscribe at all or sub and unsub all is well. All Props and ActorCells are gone except for a few ones used by Akka internally and my Manager Actor that creates my 100 instances.

If instances subscribe and are killed without unsubscribing (like our app), all ActorCells and Props stay around forever although the Actors are gone. It's the EventStream subscriptions holding the reference to the IActorRef (ActorCell) that's doing it. As you can see below I've got 107 ActorCells left. There's just the expected 7 when I unsub manually or don't sub at all.

I ran my test app linked directly with the Akka.Net source code so I could debug into the Akka code and the message to Register to unsubscribe is never received by the EventBusUnsubscriber so it never Context.Watch the actors therefore it doesn't care if they die and are still subscribed. I verified the EventBusUnsubscriber is running it just never does anything. I am unable to figure out why the Register message is never received but I think it never gets sent due to some issue with the code in EventStream.cs RegisterWithUnsubscriber. The IF never seems to evaluate to TRUE that I've seen. I can't figure out what this code is trying to avoid doing but it seems to avoid it all the time...it's too clever for me :-)

Explicitly unsubscribing Self in the PostStop of the Actor 100% fixes the issue.

Here is the code I believe the problem lies in.
Thanks
Tom

*edit - added Repro

```
        private void RegisterWithUnsubscriber(IActorRef subscriber)
        {
            _initiallySubscribedOrUnsubscriber.Match().With<Left<IImmutableSet<IActorRef>>>(v =>
            {
                if (_initiallySubscribedOrUnsubscriber.CompareAndSet(v,
                    Either.Left<IImmutableSet<IActorRef>>(v.Value.Add(subscriber))))
                {
                    RegisterWithUnsubscriber(subscriber);  // <---- this line never gets called
                }

            }).With<Right<IActorRef>>(unsubscriber =>
            {
                unsubscriber.Value.Tell( new EventStreamUnsubscriber.Register(subscriber)); // <---- neither does this
            });
        }
```

akkaeventstreams

[AkkaMemoryLeakRepro.zip](https://github.com/akkadotnet/akka.net/files/8232887/AkkaMemoryLeakRepro.zip)

Aaronontheweb:
Reproduced `IActorRef` leak inside the `EventStream`

Fixes #5717

## Changes

`IActorRef`s leaked because:

* `EventStreamUnsubscriber` actor subscribed to the wrong `Terminated` event and
* `EventStream`'s unsubscriber management code was immensely complicated and very likely wrong.

## Checklist

### DIFF
--- a/src/core/Akka.Tests/Event/Bugfix5717Specs.cs
+++ b/src/core/Akka.Tests/Event/Bugfix5717Specs.cs
@@ -1,0 +1,49 @@
+﻿// //-----------------------------------------------------------------------
+// // <copyright file="Bugfix5717Specs.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using Akka.TestKit;
+using Xunit;
+
+namespace Akka.Tests.Event
+{
+    public class Bugfix5717Specs : AkkaSpec
+    {
+        /// <summary>
+        /// Reproduction for https://github.com/akkadotnet/akka.net/issues/5717
+        /// </summary>
+        [Fact]
+        public void Should_unsubscribe_from_all_topics_on_Terminate()
+        {
+            var es = Sys.EventStream;
+            var tm1 = 1;
+            var tm2 = "FOO";
+            var a1 = CreateTestProbe();
+            var a2 = CreateTestProbe();
+
+            es.Subscribe(a1.Ref, typeof(int));
+            es.Subscribe(a2.Ref, typeof(int));
+            es.Subscribe(a2.Ref, typeof(string));
+            es.Publish(tm1);
+            es.Publish(tm2);
+            a1.ExpectMsg(tm1);
+            a2.ExpectMsg(tm1);
+            a2.ExpectMsg(tm2);
+
+            // kill second test probe
+            Watch(a2);
+            Sys.Stop(a2);
+            ExpectTerminated(a2);
+
+            EventFilter.DeadLetter().Expect(0, () =>
+            {
+                es.Publish(tm1);
+                es.Publish(tm2);
+                a1.ExpectMsg(tm1);
+            });
+        }       
+    }
+}
--- a/src/core/Akka.Tests/Event/Bugfix5717Specs.cs
+++ b/src/core/Akka.Tests/Event/Bugfix5717Specs.cs
@@ -5,6 +5,8 @@
 // // </copyright>
 // //-----------------------------------------------------------------------
 
+using System;
+using System.Threading.Tasks;
 using Akka.Configuration;
 using Akka.TestKit;
 using Xunit;
@@ -20,7 +22,7 @@ namespace Akka.Tests.Event
         /// Reproduction for https://github.com/akkadotnet/akka.net/issues/5717
         /// </summary>
         [Fact]
-        public void Should_unsubscribe_from_all_topics_on_Terminate()
+        public async Task Should_unsubscribe_from_all_topics_on_Terminate()
         {
             var es = Sys.EventStream;
             var tm1 = 1;
@@ -42,12 +44,23 @@ namespace Akka.Tests.Event
             Sys.Stop(a2);
             ExpectTerminated(a2);
 
-            EventFilter.DeadLetter().Expect(0, () =>
+            /*
+             * It's possible that the `Terminate` message may not have been processed by the
+             * Unsubscriber yet, so we want to try this operation more than once to see if it
+             * eventually executes the unsubscribe on the EventStream.
+             *
+             * If it still fails after multiple attempts, the issue is that the unsub was never
+             * executed in the first place.
+             */
+            await AwaitAssertAsync(async () =>
             {
-                es.Publish(tm1);
-                es.Publish(tm2);
-                a1.ExpectMsg(tm1);
-            });
+                await EventFilter.DeadLetter().ExpectAsync(0, () =>
+                {
+                    es.Publish(tm1);
+                    es.Publish(tm2);
+                    a1.ExpectMsg(tm1);
+                });
+            }, interval:TimeSpan.FromSeconds(250));
         }       
     }
 }
--- a/src/core/Akka.Tests/Event/Bugfix5717Specs.cs
+++ b/src/core/Akka.Tests/Event/Bugfix5717Specs.cs
@@ -5,13 +5,17 @@
 // // </copyright>
 // //-----------------------------------------------------------------------
 
+using Akka.Configuration;
 using Akka.TestKit;
 using Xunit;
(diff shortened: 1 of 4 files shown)


src/core/Akka.Tests/Event/Bugfix5717Specs.cs
```diff
@@ -0,0 +1,49 @@
...
+// //     Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using Akka.TestKit;
+using Xunit;
+
+namespace Akka.Tests.Event
+{
+    public class Bugfix5717Specs : AkkaSpec
```
Aaronontheweb, line 13: Reproduction for #5717

src/core/Akka/Event/EventBusUnsubscriber.cs
```diff
@@ -45,140 +42,83 @@ public EventStreamUnsubscriber(EventStream eventStream, ActorSystem system, bool
...
-        /// </summary>
-        /// <param name="message">TBD</param>
-        /// <returns>TBD</returns>
+        
         protected override bool Receive(object message)
         {
-            return message.Match().With<Register>(register =>
+            switch (message)
             {
-                if (_debug)
```
Aaronontheweb, line 58: The code before was very old and still used `PatternMatch`. Refactored to a `switch` statement.

src/core/Akka/Event/EventStream.cs
```diff
@@ -29,8 +30,13 @@ public class EventStream : LoggingBus
     {
         private readonly bool _debug;
 
-        private readonly AtomicReference<Either<IImmutableSet<IActorRef>, IActorRef>> _initiallySubscribedOrUnsubscriber =
```
Aaronontheweb, line 32: This is complicated and it's not clear that it works very well.

src/core/Akka/Event/EventStream.cs
```diff
@@ -29,8 +30,13 @@ public class EventStream : LoggingBus
...
-            new AtomicReference<Either<IImmutableSet<IActorRef>, IActorRef>>();
+        // used to uniquely name unsubscribers instances, should there be more than one ActorSystem / EventStream
+        private static readonly AtomicCounter UnsubscribersCounter = new AtomicCounter(0);
+        private readonly AtomicReference<IActorRef> _unsubscriber = new AtomicReference<IActorRef>(ActorRefs.NoSender);
```
Aaronontheweb, line 35: Replaced with 1 unsubscriber `IActorRef` and a `ConcurrentSet<IActorRef>`, the latter of which records subscribers that have been started prior to the `ActorSystem` itself starting.

src/core/Akka/Event/EventStream.cs
```diff
@@ -108,87 +114,69 @@ public override bool Unsubscribe(IActorRef subscriber)
...
-        {
-            if (system == null)
-            {
-                return false;
-            }
-            return _initiallySubscribedOrUnsubscriber.Match().With<Left<IImmutableSet<IActorRef>>>(v =>
+            if (_unsubscriber.Value.IsNobody())
             {
-                if (_initiallySubscribedOrUnsubscriber.CompareAndSet(v, Either.Right(unsubscriber)))
+                lock (this)
```
Aaronontheweb, line 125: No need for complicated CAS behavior here. Just lock it.

Aaronontheweb (commented):
Review my own PR

src/core/Akka/Event/EventBusUnsubscriber.cs
```diff
@@ -45,140 +42,83 @@ public EventStreamUnsubscriber(EventStream eventStream, ActorSystem system, bool
...
-            /// TBD
-            /// </summary>
-            public IActorRef Actor { get; private set; }
-        }
-
-
-        /// <summary>
-        /// TBD
-        /// </summary>
-        internal class Terminated
```
Aaronontheweb, line 113: Most probable cause of #5717 - we were handling our own `Terminated` message rather than the built-in one. This looks like a porting error that never was caught years ago.

Aaronontheweb:
> Akka.Tests.Event.Bugfix5717Specs.Should_unsubscribe_from_all_topics_on_Terminate
> Received 2 messages too many. Expected 0 messages but received 2 that matched filter [DeadLetter]
> Expected: True
> Actual:   False

Issue with my test. I'll get that fixed.

Aaronontheweb:
> 

Seems to indicate that the unsubscribe wasn't processed yet.

src/core/Akka/Event/EventStream.cs
```diff
@@ -108,87 +114,69 @@ public override bool Unsubscribe(IActorRef subscriber)
...
-                {
-                    UnregisterIfNoMoreSubscribedChannels(subscriber);
-                }
-
-            }).With<Right<IActorRef>>(unsubscriber =>
-            {
-                unsubscriber.Value.Tell(new EventStreamUnsubscriber.UnregisterIfNoMoreSubscribedChannels(subscriber));
-            });
+                _unsubscriber.Value.Tell(new EventStreamUnsubscriber.UnregisterIfNoMoreSubscribedChannels(subscriber));
+            }
```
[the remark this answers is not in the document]
Aaronontheweb: No, because this event only gets fired when the actor is already unsubscribed to all topics. If the `_unsubscriber` hasn't been created yet then it can't have the deathwatch subscription that this message removes in the first place.

Arkatufus (approved):
LGTM, just had a question

PR #5721, opened 2022-03-16
Convert AwaitCondition in TestKitBase initializer, can not use `.Wait()` in ctor

Arkatufus:
Using sync over async in constructors can cause deadlocks


Aaronontheweb (approved):
LGTM

PR #5722, opened 2022-03-16
supress `InvalidOperationException` in xUnit OutputLogger

Aaronontheweb:
Whenever I have a test run long and the `ActorSystem` has to shutdown behind the scenes, my logs are filled with thousands of `InvalidOperationException: no active test" as a result of calling `_output.WriteLine` once the XUnit test has been terminated. Each time this actor crashes and restarts it generates another one of those logs.

## Changes

Added `InvalidOperationException` to the `catch` list and we just write out to `Console` now when this occurs, as xUnit test `IOutputHelper` is no longer available.


Arkatufus (approved):
LGTM

PR #5729, opened 2022-03-18
fix `ActorSystemSetup.And`

### LINKED ISSUE #5728: `ActorSystemSetup.And` forgets types
**Version Information**
Version of Akka.NET? v1.4.34
Which Akka.NET Modules? Akka

**Describe the bug**

Discovered this while using Akka.Hosting.

Suppose you create multiple `Setup` classes of discrete types:

1. `Akka.DependencyInjection.DependencyResolverSetup` and
2. `Akka.Serialization.Hyperion.HyperionSerializerSetup`

Right now Akka.NET only allows one setup of each type to be passed into the `ActorSystem`, by way of an `ActorSystemSetup` - which you can think of as an "aggregate" setup.

```csharp
// earlier in the class
private readonly HashSet<Setup> _setups = new HashSet<Setup>();

  var diSetup = DependencyResolverSetup.Create(sp);
  var bootstrapSetup = BootstrapSetup.Create().WithConfig(Configuration.GetOrElse(Config.Empty));
  if (ActorRefProvider.HasValue) // only set the provider when explicitly required
  {
      bootstrapSetup = bootstrapSetup.WithActorRefProvider(ActorRefProvider.Value);
  }


  var actorSystemSetup = bootstrapSetup.And(diSetup);
  foreach (var setup in _setups)
  {
      actorSystemSetup = actorSystemSetup.And(setup);
  }
```

> `WithSetup` is called by `And`

Because each individual `Setup` is downcast to its base class (`Setup`), the key that `ActorSystemSetup.And` is going to use to store each of the `Setup`s inside `_setups` is.... `Setup` - therefore, the contents of `_setups` all override each other and only the last `Setup` in the set actually gets used inside the `ActorSystemSetup`.

**Expected behavior**

What should happen instead is that the type key of the actual concrete type should be included, rather than the generic parameter argument:

https://github.com/akkadotnet/akka.net/blob/21762b90ac1d7db670487dda1368d84c61d693e5/src/core/Akka/Actor/Setup/ActorSystemSetup.cs#L73-L76

Rather than do `typeof(T)` we should probably do `setup.GetType()` in order to ensure that this type collision does not occur.

**Actual behavior**
What actually happened and how did it differ from your expectations?

All `Setup`s included in a collection of `Setup` types overrode each other.

Aaronontheweb:
close #5728 

## Changes

Modified the `ActorSystemSetup.WithSetup` method to evaluate the type of the `Setup` object being passed in rather than the generic method argument type.

## Checklist

### DIFF
--- a/src/core/Akka.Tests/Actor/Setup/ActorSystemSetupSpec.cs
+++ b/src/core/Akka.Tests/Actor/Setup/ActorSystemSetupSpec.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using Akka.Actor;
 using Akka.Actor.Setup;
 using Akka.TestKit;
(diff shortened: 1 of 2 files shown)


src/core/Akka.Tests/Actor/Setup/ActorSystemSetupSpec.cs
```diff
@@ -107,5 +108,36 @@ public void ActorSystemSettingsShouldBeAvailableFromExtendedActorSystem()
...
+
+            // act
+            var dummySetup = actorSystemSetup.Get<DummySetup>();
+            var dummySetup2 = actorSystemSetup.Get<DummySetup2>();
+            
+            // shouldn't exist
+            var dummySetup3 = actorSystemSetup.Get<DummySetup3>();
+
+            // assert
+            dummySetup.HasValue.Should().BeTrue();
```
Aaronontheweb, line 138: This assertion failed prior to these changes.

src/core/Akka/Actor/Setup/ActorSystemSetup.cs
```diff
@@ -72,7 +72,8 @@ internal ActorSystemSetup(ImmutableDictionary<Type, Setup> setups)
         /// <returns>A new, immutable <see cref="ActorSystemSetup"/> instance.</returns>
         public ActorSystemSetup WithSetup<T>(T setup) where T : Setup
         {
-            return new ActorSystemSetup(_setups.SetItem(typeof(T), setup));
+            var typeT = setup.GetType();
```
Aaronontheweb, line 75: Resolve type from the object, rather than the method signature.

Arkatufus (approved):
LGTM

PR #5743, opened 2022-03-22
marked `EventBusUnsubscriber` messages as `INoSerializationVerificationNeeded`

Aaronontheweb:
Eliminates issues with some `MinimalActorRef` actors subscribing to the `EventStream` when `akka.actor.serialize-messages = on`.

### DIFF
--- a/src/core/Akka/Event/EventBusUnsubscriber.cs
+++ b/src/core/Akka/Event/EventBusUnsubscriber.cs
@@ -90,7 +90,7 @@ namespace Akka.Event
         ///
         /// Registers a new subscriber to be death-watched and automatically unsubscribed.
         /// </summary>
-        internal class Register
+        internal class Register : INoSerializationVerificationNeeded
         {
             public Register(IActorRef actor)
             {
@@ -108,7 +108,7 @@ namespace Akka.Event
         ///
         /// Unsubscribes an actor that is no longer subscribed and does not need to be death-watched any longer.
         /// </summary>
-        internal class UnregisterIfNoMoreSubscribedChannels
+        internal class UnregisterIfNoMoreSubscribedChannels : INoSerializationVerificationNeeded
         {
             public UnregisterIfNoMoreSubscribedChannels(IActorRef actor)
             {


src/core/Akka/Event/EventBusUnsubscriber.cs
```diff
@@ -90,7 +90,7 @@ protected override void PreStart()
         ///
         /// Registers a new subscriber to be death-watched and automatically unsubscribed.
         /// </summary>
-        internal class Register
+        internal class Register : INoSerializationVerificationNeeded
```
eaba, line 93: Is this the intended design for this class? Classes that implement `INoSerializationVerificationNeeded` won't be serialized if configuration setting 'akka.actor.serialize-messages' is "on"

src/core/Akka/Event/EventBusUnsubscriber.cs
```diff
@@ -108,7 +108,7 @@ public Register(IActorRef actor)
         ///
         /// Unsubscribes an actor that is no longer subscribed and does not need to be death-watched any longer.
         /// </summary>
-        internal class UnregisterIfNoMoreSubscribedChannels
+        internal class UnregisterIfNoMoreSubscribedChannels : INoSerializationVerificationNeeded
```
eaba, line 111: Same as above

eaba (commented):
@Aaronontheweb left you some comments

src/core/Akka/Event/EventBusUnsubscriber.cs
Aaronontheweb: Yes, this is entirely internal and is `IActorRef` is immutable,

src/core/Akka/Event/EventBusUnsubscriber.cs
Aaronontheweb: Same reply as above.

src/core/Akka/Event/EventBusUnsubscriber.cs
eaba: Got it!

PR #5747, opened 2022-03-23
Fix MSBuild does not copy xunit.runner dlls correctly

Arkatufus:
## Changes

Added xunit.runner.visualstudio nuget package, MSBuild fails to copy the dlls from the MNTR package


Aaronontheweb (approved):
LGTM

PR #5797, opened 2022-03-30
Port `Akka.Tests.IO` tests to `async/await` - `TcpListenerSpec`

eaba:


src/core/Akka.Tests/IO/TcpListenerSpec.cs
```diff
@@ -32,36 +33,36 @@ public TcpListenerSpec()
         [Fact]
         public void A_TCP_Listener_must_let_the_bind_commander_know_when_binding_is_complete()
         {
-            new TestSetup(this, pullMode: false).Run(x =>
+            new TestSetup(this, pullMode: false).Run(async x =>
```
Arkatufus, line 36: This doesn't work, change the `Run` method to async

Arkatufus (changes requested):

src/core/Akka.Tests/IO/TcpListenerSpec.cs
eaba: How do I do that? See below:
```csharp
public void Run(Action<TestSetup> test)
{
    test(this);
 }
```

Arkatufus (approved):
LGTM

PR #5816, opened 2022-04-01
Fix leaky coordinated shutdown

### LINKED ISSUE #5802: ActorSystem leaks memory
**Version Information**
1.4.6 - 1.4.35
Did not test every single version but at least 1.4.6 and 1.4.35 leaks.
No leak on 1.4.5

.NET 6

Using Akka NuGet package

**Describe the bug**
Creating ActorSystem, terminating it and disposing it in a loop, leaves memory behind.

**To Reproduce**

Complete console app code example:
```
using Akka.Actor;

while (true)
{
    var actorSystem = ActorSystem.Create("test");

    await actorSystem.Terminate();
    actorSystem.Dispose();
    actorSystem = null;

    GC.Collect();
    GC.WaitForPendingFinalizers();
    GC.Collect();
}
```

Look at Visual Studio diagnostics memory usage or publish the app and look at memory usage in task manager. It will rise continously.

**Expected behavior**
Memory usage should flatline and not keep rising.

**Actual behavior**
Memory usage kept rising and did not stop like in 1.4.5.

**Screenshots**
1.4.35 memory usage graph: https://imgur.com/qJgGHFw
1.4.5 memory usage graph: https://imgur.com/ggyhBrc
1.4.35 memory usage snapshot diff: https://imgur.com/Mhdlwpc

**Environment**
Windows 10
.NET 6

**Additional context**

Arkatufus:
Fixes #5802

## Changes

* Fix possible infinite loop 
* Fix circular reference memory leak

### DIFF
--- a/src/core/Akka/Actor/CoordinatedShutdown.cs
+++ b/src/core/Akka/Actor/CoordinatedShutdown.cs
@@ -653,8 +653,7 @@ namespace Akka.Actor
 
                     if (terminateActorSystem)
                     {
-                        system.FinalTerminate();
-                        return system.Terminate().ContinueWith(tr =>
+                        return system.FinalTerminate().ContinueWith(tr =>
                         {
                             if (exitClr && !coord._runningClrHook)
                             {
--- a/src/core/Akka/Actor/CoordinatedShutdown.cs
+++ b/src/core/Akka/Actor/CoordinatedShutdown.cs
@@ -270,7 +270,7 @@ namespace Akka.Actor
         /// </summary>
         internal readonly List<string> OrderedPhases;
 
-        private readonly ConcurrentBag<Func<Task<Done>>> _clrShutdownTasks = new ConcurrentBag<Func<Task<Done>>>();
+        private readonly ConcurrentSet<Func<Task<Done>>> _clrShutdownTasks = new ConcurrentSet<Func<Task<Done>>>();
         private readonly ConcurrentDictionary<string, ImmutableList<(string, Func<Task<Done>>)>> _tasks = new ConcurrentDictionary<string, ImmutableList<(string, Func<Task<Done>>)>>();
         private readonly AtomicReference<Reason> _runStarted = new AtomicReference<Reason>(null);
         private readonly AtomicBoolean _clrHooksStarted = new AtomicBoolean(false);
@@ -335,7 +335,7 @@ namespace Akka.Actor
         {
             if (!_clrHooksStarted)
             {
-                _clrShutdownTasks.Add(hook);
+                _clrShutdownTasks.TryAdd(hook);
             }
         }
 
@@ -662,6 +662,7 @@ namespace Akka.Actor
 
                             coord.System = null;
                             coord.Log = null;
+                            coord._tasks.Clear(); // Clear the dictionary, just in case it is retained in memory
                             return Done.Instance;
                         });
                     }
@@ -693,7 +694,11 @@ namespace Akka.Actor
                 var exitTask = TerminateOnClrExit(coord);
                 // run all hooks during termination sequence
                 AppDomain.CurrentDomain.ProcessExit += exitTask;
-                system.WhenTerminated.ContinueWith(tr => { AppDomain.CurrentDomain.ProcessExit -= exitTask; });
+                system.WhenTerminated.ContinueWith(tr =>
+                {
+                    AppDomain.CurrentDomain.ProcessExit -= exitTask;
+                    coord._clrShutdownTasks.Clear(); // Clear the tasks, just in case it is retained in memory
+                });
 
                 coord.AddClrShutdownHook(() =>
                 {
--- a/src/core/Akka/Actor/CoordinatedShutdown.cs
+++ b/src/core/Akka/Actor/CoordinatedShutdown.cs
@@ -251,7 +251,7 @@ namespace Akka.Actor
         /// <summary>
         /// The <see cref="ActorSystem"/>
         /// </summary>
-        public ExtendedActorSystem System { get; }
+        public ExtendedActorSystem System { get; private set; }
 
         /// <summary>
         /// The set of named <see cref="Phase"/>s that will be executed during coordinated shutdown.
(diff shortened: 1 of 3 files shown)


src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -653,13 +653,15 @@ internal static void InitPhaseActorSystemTerminate(ActorSystem system, Config co
 
                     if (terminateActorSystem)
                     {
-                        system.FinalTerminate();
-                        return system.Terminate().ContinueWith(tr =>
```
Arkatufus, line 657: This potentially an infinite loop call, if the call came from `ActorSystem.Terminate()` in the first place.

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -653,13 +653,15 @@ internal static void InitPhaseActorSystemTerminate(ActorSystem system, Config co
...
+                        return system.FinalTerminate().ContinueWith(tr =>
                         {
                             if (exitClr && !coord._runningClrHook)
                             {
                                 Environment.Exit(0);
                             }
+
+                            coord.System = null;
```
Arkatufus, line 663: Clear the reference to the `ActorSystem` to prevent a circular reference, preventing the `ActorSystem` from being garbage collected afterward.

Aaronontheweb (approved):
LGTM - need to see if CI digs up anything harmful here with the changing of the `CoordinatedShutdown` tasks though

Arkatufus:
dotMemory profiling of 500 consecutive creation and termination of a single `ActorSystem`.

### Before

image

### After

image

Aaronontheweb:
Nice work, definitely looks like it solved the memory issue.

isurakka:
Hey, I'm still able to reproduce the leak with this code:

```
using Akka.Actor;

while (true)
{
    var actorSystem = ActorSystem.Create("test");

    await actorSystem.Terminate();
    actorSystem.Dispose();
    actorSystem = null;

    GC.Collect();
    GC.WaitForPendingFinalizers();
    GC.Collect();
}
```

Arkatufus:
@isurakka 
Are you using a custom build of Akka using this new PR?
This fix is still not in effect, we still need to have this merged in and included in a release.

Arkatufus:
OK, this should really fix it.
[](https://user-images.githubusercontent.com/8293237/161600051-b46d110c-10af-4a09-8f32-989e44de5a6a.png)Found the real culprit, CoordinatedShutdown._clrShutdownTasks is a ConcurrentBag, ConcurrentBag uses ThreadLocal, ThreadLocal acts like a static inside the thread and never get GC-ed.

CoordinatedShutdown+<>c_DisplayClass55_0 is a compiler generated EventHandler function for the CoordinatedShutdown.InitClrHook method.
image

Memory allocation after the fix:
image

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -270,7 +270,7 @@ public class ClusterJoinUnsuccessfulReason : Reason
         /// </summary>
         internal readonly List<string> OrderedPhases;
 
-        private readonly ConcurrentBag<Func<Task<Done>>> _clrShutdownTasks = new ConcurrentBag<Func<Task<Done>>>();
```
Arkatufus, line 273: This is the real cause of the memory leak. `ConcurrentBag` uses `ThreadLocal` that pinned the lambdas in memory, preventing them from being GC-ed.

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -691,7 +694,11 @@ internal static void InitClrHook(ActorSystem system, Config conf, CoordinatedShu
                 var exitTask = TerminateOnClrExit(coord);
                 // run all hooks during termination sequence
                 AppDomain.CurrentDomain.ProcessExit += exitTask;
-                system.WhenTerminated.ContinueWith(tr => { AppDomain.CurrentDomain.ProcessExit -= exitTask; });
+                system.WhenTerminated.ContinueWith(tr =>
+                {
+                    AppDomain.CurrentDomain.ProcessExit -= exitTask;
+                    coord._clrShutdownTasks.Clear(); // Clear the tasks, just in case it is retained in memory
```
Arkatufus, line 700: I'm pretty sure that `ConcurrentSet` does not use `ThreadLocal`, but lets add this just in case that changes in the future.

Arkatufus:
This is the code I use to test for the memory leak, you will need JetBrains dotMemory and add the `JetBrains.Profiler.SelfApi` nuget package:
```c#
using Akka.Actor;
using Akka.Configuration;
using JetBrains.Profiler.SelfApi;

Console.WriteLine("Downloading DotMemory...");
await DotMemory.EnsurePrerequisiteAsync();
Console.WriteLine("Download complete.");

var profileConfig = new DotMemory.Config()
    .SaveToDir("./TestResults/")
    .OpenDotMemory();
DotMemory.Attach(profileConfig);

var config = ConfigurationFactory.ParseString(@"
akka.coordinated-shutdown.run-by-actor-system-terminate = on
akka.coordinated-shutdown.run-by-clr-shutdown-hook = on
");

Console.WriteLine("Grabbing 1st snapshot");
DotMemory.GetSnapshot();
Console.WriteLine("Starting loop");
try
{
    for (var i = 0; i < 100; ++i)
    {
        Console.WriteLine(i);
        var actorSystem = ActorSystem.Create("test", config);
        await actorSystem.Terminate();
    }
    await Task.Delay(1000);
    GC.Collect();
    GC.WaitForPendingFinalizers();
    GC.Collect();
    await Task.Delay(1000);
    Console.WriteLine("Grabbing 2nd snapshot");
    DotMemory.GetSnapshot();
}
finally
{
    DotMemory.Detach();
}
Console.WriteLine("Profiling done.");
```

src/core/Akka/Actor/CoordinatedShutdown.cs
Aaronontheweb: yikes

Aaronontheweb (approved):
LGTM

PR #5821, opened 2022-04-05
Be more explicit that a node is shutting down as it was marked as DOWN

ismaelhamed:
Port: [#26861](https://github.com/akka/akka/pull/26861)


Aaronontheweb (approved):
LGTM

PR #5822, opened 2022-04-05
Revert changes that are unrelated to the CoordinatedShutdown memory leak

Arkatufus:


Arkatufus:
Memory profiler result after the changes:
image

PR #5835, opened 2022-04-08
Change the failure log inside `AllEventPublisher` from Debug to Error

### LINKED ISSUE #5834: [Persistence] Query does not report failure correctly
Related to #5824 

Failure in `AllEvents` query is reported on debug level, which is incorrect. It seems like error and failure handling in Akka.Persistence.Query is not standardized. We need to standardize error and failure reporting in Akka.Persistence.Query.

Arkatufus:
Closes #5834

This doesn't really fix a bug. We're just making sure that an error is logged even when nothing is catching the stream failure exception.


Aaronontheweb (approved):
LGTM

PR #5838, opened 2022-04-12
Fix LeaseProvider instance Activator exception handling

Arkatufus:
Since the first Activator call was wrapped using a universal try..catch that swallows any exception thrown from inside activated instance constructor, making it impossible to inspect the actual exception since the second Actovator call exception will obscure the exception thrown by the first Activator.


Aaronontheweb (approved):
LGTM

PR #5849, opened 2022-04-18
Fix Persistence.TCK specs

Arkatufus:
- DeleteMessagesFailure message stack trace were suppressed in log
- Make JournalSpec facts overridable


src/contrib/cluster/Akka.Cluster.Sharding/PersistentShard.cs
```diff
@@ -125,7 +125,7 @@ protected override bool ReceiveCommand(object message)
                     DeleteSnapshots(new SnapshotSelectionCriteria(deleteTo, DateTime.MaxValue, deleteFrom));
                     break;
                 case DeleteMessagesFailure m:
-                    Log.Warning("PersistentShard messages to [{0}] deletion failure: [{1}]", m.ToSequenceNr, m.Cause.Message);
+                    Log.Warning(m.Cause, "PersistentShard messages to [{0}] deletion failure: [{1}]", m.ToSequenceNr, m.Cause.Message);
```
Aaronontheweb, line 128: Nice catch

Aaronontheweb (approved):
LGTM

PR #4629, opened 2020-11-12
akka.cluster.sharding - storage api update

zbynek001:
Another big one ^^
This MR is updating sharding to current scala version, including unit tests and multinode tests.

What's still missing compared to scala:
- ~~`ExternalShardAllocation`~~
- `RemoveInternalClusterShardingData` app

There were quite many changes related to the storage api change, so basically some parts were completely rewritten. Especially `Shard` and `ShardCoordinator`.

Also includes some pre-requirements to make some part easier, especially `MessageBuffer` and `Dropped` deadletter. Not sure if you'll like it.

~~Tests are now using `sqlite` persistence provider~~
Tests are now using shared inmem persistence providers for both journal & snapshots to avoid issues with files

In theory, everything "should" be backwards compatible 

Additionally included in this pr:
* Race condition in DeprecatedSupervisionSpec fixed (migrated from https://github.com/akka/akka/pull/29914)
* Small clarification of recovery strategy in config (migrated from https://github.com/akka/akka/pull/30167)
* Resolve snapshot check skipped for some events (migrated from https://github.com/akka/akka/pull/30226)
* Defer coordinator stop until region graceful stop has completed (migrated from https://github.com/akka/akka/pull/30338)
* sharding: actively signal 'region stopped' to the coordinator (migrated from https://github.com/akka/akka/pull/30402)

PR #5857, opened 2022-04-20
Zbynek001 sharding update2

Aaronontheweb:
https://github.com/akkadotnet/akka.net/pull/4629 but with merge conflicts resolved.


Aaronontheweb:
Test worked fine; retrying with `remember-entities=on`.

Then I'll retest with `state-store-mode=ddata`.

Aaronontheweb:
Upgrade ran without any problems. Re-trying on DData.

Aaronontheweb:
DData with remember-entities has some major problems in 1.4 by the looks of it; upgrading to this PR helped, but there are still some fundamental issues with the `DDataShardCoordinator` that need to be fixed. Will start working on that now once this is merged.

PR #5863, opened 2022-04-21
fix sharding recovery

### LINKED ISSUE #5862: Incorrect invocation of the 'IEnumerable<T>.Union' method
Hi,
I found the suspicious code via the PVS-Studio analyzer.

Here is the code:
```cs
...
case ShardId shardId:
  _shards.Add(shardId); // <=
  return true;
case SnapshotOffer offer when (offer.Snapshot is ShardCoordinator.CoordinatorState state):
  _shards.UnionWith(state.Shards.Keys.Union(state.UnallocatedShards)); // <=
  return true;
case SnapshotOffer offer when (offer.Snapshot is State state):
  _shards.Union(state.Shards); // <=
...

```
Here is [the link](https://github.com/akkadotnet/akka.net/blob/8e7f02149451856c9160f9cd28d687ff8a166a11/src/contrib/cluster/Akka.Cluster.Sharding/Internal/EventSourcedRememberEntitiesCoordinatorStore.cs#L123) to the sources.

Several methods were called for the `_shards` variable. The last invocation looks suspicious. 

`Add` and `UnionWith` are methods of the `HashSet<T>` type. They change the `_shards` object state. But `Union` is the extension method which does not change the `_shards` state. The result of this invocation is not used as well.

zbynek001:
Fixes #5862

## Changes

Fixes incorrect use of Union in sharding recovery

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/Internal/EventSourcedRememberEntitiesCoordinatorStore.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/Internal/EventSourcedRememberEntitiesCoordinatorStore.cs
@@ -120,7 +120,7 @@ namespace Akka.Cluster.Sharding.Internal
                     _shards.UnionWith(state.Shards.Keys.Union(state.UnallocatedShards));
                     return true;
                 case SnapshotOffer offer when (offer.Snapshot is State state):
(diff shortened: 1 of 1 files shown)


Aaronontheweb (approved):
LGTM - thanks @zbynek001

PR #5864, opened 2022-04-21
Fix Akka.DistributedData.Tests.MultiNode ReplicatorPruningSpec

Arkatufus:


src/contrib/cluster/Akka.DistributedData.Tests.MultiNode/ReplicatorPruningSpec.cs
```diff
@@ -189,8 +189,11 @@ public void Pruning_of_CRDT_should_move_data_from_removed_node()
             void UpdateAfterPruning(ulong expectedValue)
             {
                 // inject data from removed node to simulate bad data
-                _replicator.Tell(Dsl.Update(_keyA, GCounter.Empty, new WriteAll(_timeout), x => x.Merge(oldCounter).Increment(_cluster, 1)));
-                ExpectMsg<UpdateSuccess>(msg =>
```
Arkatufus, line 193: This is wrong, the test need to separate the `ExpectMsg<UpdateSuccess>` and the `ExpectMsg<GetSuccess>` test because they're independent events, `UpdateSuccess` need to happen first, then we need to wait until `GetSuccess` happened.

The way it is set up in the old test is that the second test only get executed once because `UpdateSuccess` only happens once, but we wanted to repeatedly try to get until it succeed or failed within the timeout period.

Aaronontheweb (approved):
LGTM

PR #5880, opened 2022-04-25
[Async TestKit] Convert Akka.Remote.Tests to async - Transport.AkkaProtocolSpec

Arkatufus:


Arkatufus:
Was racy before the conversion

Aaronontheweb:
> Was racy before the conversion

yep, this is one I want to keep an eye on once it's merged in

Aaronontheweb (approved):
LGTM

PR #5887, opened 2022-04-26
[Async TestKit] Convert Akka.Remote.Tests to async - RemoteRouterSpec

Arkatufus:


src/core/Akka.Remote.Tests/Akka.Remote.Tests.csproj
```diff
@@ -4,8 +4,8 @@
 
   <PropertyGroup>
     <AssemblyTitle>Akka.Remote.Tests</AssemblyTitle>
-    <TargetFrameworks>$(NetFrameworkTestVersion);$(NetTestVersion);$(NetCoreTestVersion)</TargetFrameworks>
```
Aaronontheweb, line 7: Are we dropping .NET Framework support here?

Aaronontheweb:
Rest of the tests look fine, save for that one comment above.

src/core/Akka.Remote.Tests/Akka.Remote.Tests.csproj
Aaronontheweb: Ah, _that_ again. Ok, fine with me.

PR #5894, opened 2022-04-26
[Async TestKit] Convert Akka.Remote.Tests to async - DotNettySslSupportSpec

Arkatufus:


Aaronontheweb:
@Arkatufus looks like one of the tests we refactored failed:

```
Akka.Remote.Tests.RemoteDeathWatchSpec.Must_quarantine_systems_after_unsuccessful_system_message_delivery_if_have_not_communicated_before
Failed: Expected no messages during 00:00:05, instead we received <<Terminated>: [akka.tcp://extinct-system@localhost:37980/user/noone] - ExistenceConfirmed=False> from [akka.tcp://extinct-system@localhost:37980/user/noone] after 00:00:00.0000214
Expected: True
Actual:   False
```

src/core/Akka.Remote.Tests/RemoteDeathWatchSpec.cs
```diff
@@ -141,6 +141,7 @@ public async Task Must_quarantine_systems_after_unsuccessful_system_message_deli
 
             var probe = CreateTestProbe();
             probe.Watch(extinctRef);
+            (await probe.ExpectMsgAsync<Terminated>()).ActorRef.ShouldBe(extinctRef);
             probe.Unwatch(extinctRef);
```
Arkatufus, line 145: I don't know what the purpose of this rapid `Watch` `Unwatch` calls is, but the `extinctRef` actor ref does get terminated and if the terminated message arrive just between these 2 method calls, this test will fail. So we can either remove the rapid `Watch`-`Unwatch` or explicitly wait until the `Teminated` message arrives.

Aaronontheweb: I thought we were already waiting for the Terminated message?

Arkatufus: Nope, the terminated message was never awaited

Aaronontheweb:
```
 (await probe.ExpectMsgAsync<Terminated>()).ActorRef.ShouldBe(extinctRef);
```

Ah, so this is a fix you _just_ introduced. Got it.

PR #5896, opened 2022-04-27
[Async TestKit] Convert Akka.Remote.Tests to async - Transport.DotNettyTransportShutdownSpec

Arkatufus:


src/core/Akka.Remote.Tests/RemotingTerminatorSpecs.cs
```diff
@@ -58,6 +58,7 @@ public async Task RemotingTerminator_should_shutdown_promptly_with_no_associatio
         public async Task RemotingTerminator_should_shutdown_promptly_with_some_associations()
         {
             _sys2 = ActorSystem.Create("System2", RemoteConfig);
+            InitializeLogger(_sys2);
```
Arkatufus, line 61: Couldn't find what's wrong with the failing test, adding xunit logging support on the second ActorSystem so I can debug it better if it happens again. Need to see if the second system successfully shuts down gracefully.

PR #5900, opened 2022-04-28
Convert Akka.Remote.Tests to async - Remove AkkaSpec dependency

Arkatufus:
All these tests does not need to inherit AkkaSpec, removing it shaves about 400ms per test

### DIFF
--- a/src/core/Akka.Remote.Tests/AccrualFailureDetectorSpec.cs
+++ b/src/core/Akka.Remote.Tests/AccrualFailureDetectorSpec.cs
@@ -8,14 +8,13 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using Akka.TestKit;
 using Akka.Util.Internal;
 using Xunit;
 
 namespace Akka.Remote.Tests
 {
 
-    public class AccrualFailureDetectorSpec : AkkaSpec
+    public class AccrualFailureDetectorSpec
     {
         public static IEnumerable<(T, T)> Slide<T>(IEnumerable<T> values)
         {
--- a/src/core/Akka.Remote.Tests/AckedDeliverySpec.cs
+++ b/src/core/Akka.Remote.Tests/AckedDeliverySpec.cs
@@ -16,7 +16,7 @@ using Xunit;
 namespace Akka.Remote.Tests
 {
     
-    public class AckedDeliverySpec : AkkaSpec
+    public class AckedDeliverySpec
     {
         sealed class Sequenced : IHasSequenceNumber
         {
--- a/src/core/Akka.Remote.Tests/DeadlineFailureDetectorSpec.cs
+++ b/src/core/Akka.Remote.Tests/DeadlineFailureDetectorSpec.cs
@@ -8,14 +8,13 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using Akka.TestKit;
 using FluentAssertions;
 using FluentAssertions.Extensions;
 using Xunit;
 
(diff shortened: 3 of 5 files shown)


Aaronontheweb:
```
Akka.Remote.Tests.Transport.AkkaProtocolSpec.ProtocolStateActor_must_in_outbound_mode_delay_readiness_until_handshake_finished
```

this spec failed - has this already been ported to the async test methods @Arkatufus ?

Arkatufus:
It has, let me recheck it

src/core/Akka.Remote.Tests/Transport/AkkaProtocolSpec.cs
```diff
@@ -117,19 +117,18 @@ private Collaborators GetCollaborators()
...
+            public override bool IsAvailable => Volatile.Read(ref _isAvailable);
+            public void SetAvailable(bool available) => Volatile.Write(ref _isAvailable, available);
+            
+            private volatile bool _called;
+            public override bool IsMonitoring => Volatile.Read(ref _called);
 
             public override void HeartBeat()
             {
-                called = true;
+                Volatile.Write(ref _called, true);
```
Arkatufus, line 129: Volatile is being accessed inconsistently, make sure that all volatile are accessed properly.

src/core/Akka.Remote.Tests/Transport/AkkaProtocolSpec.cs
```diff
@@ -167,7 +166,7 @@ public async Task ProtocolStateActor_must_in_inbound_mode_accept_payload_after_A
 
             reader.Tell(TestAssociate(33), TestActor);
 
-            await AwaitConditionAsync(() => collaborators.FailureDetector.called, DefaultTimeout);
+            await AwaitConditionAsync(() => collaborators.FailureDetector.IsMonitoring, DefaultTimeout);
```
Arkatufus, line 169: On second thought, its safer to await the condition instead

src/core/Akka.Remote.Tests/Transport/AkkaProtocolSpec.cs
```diff
@@ -236,7 +235,8 @@ public async Task ProtocolStateActor_must_in_outbound_mode_delay_readiness_until
                 failureDetector: collaborators.FailureDetector));
 
             await AwaitConditionAsync(() => LastActivityIsAssociate(collaborators.Registry, 42), DefaultTimeout);
-            Assert.True(collaborators.FailureDetector.called);
+            
+            await AwaitConditionAsync(() => collaborators.FailureDetector.IsMonitoring, DefaultTimeout);
```
Arkatufus, line 239: On second thought, its safer to await the condition instead

Arkatufus, line 239: On second thought, its safer to await the condition instead

src/core/Akka.Remote.Tests/Transport/AkkaProtocolSpec.cs
```diff
@@ -181,7 +180,7 @@ public async Task ProtocolStateActor_must_in_inbound_mode_accept_payload_after_A
 
             wrappedHandle.ReadHandlerSource.SetResult(new ActorHandleEventListener(TestActor));
 
-            Assert.True(collaborators.FailureDetector.called);
+            Assert.True(collaborators.FailureDetector.IsMonitoring);
```
Arkatufus, line 183: Need to change this to the same assertion if the last change proves to fix the racy condition

PR #5901, opened 2022-04-28
[Async TestKit] Convert Akka.Remote.Tests to async - Transport.ThrottlerTransportAdapterSpec

### LINKED ISSUE #5432: Akka.TestKit: take dependency on FluentAssertions?
We should probably just take a dependency on FluentAssertions in the TestKit. I don't know why we haven't yet. 

I'll make a bigger issue for this and we'll revisit that prior to releasing v1.4.30.

_Originally posted by @Aaronontheweb in https://github.com/akkadotnet/akka.net/pull/5430#discussion_r769172219_

Arkatufus:
Closes #5432

Guard all async calls with timeouts to make sure that the test will never hung


src/contrib/testkits/Akka.TestKit.Xunit2/Akka.TestKit.Xunit2.csproj
```diff
@@ -4,14 +4,15 @@
...
-    <TargetFrameworks>$(NetStandardLibVersion)</TargetFrameworks>
+    <TargetFramework>$(NetStandardLibVersion)</TargetFramework>
     <PackageTags>$(AkkaPackageTags);testkit;xunit</PackageTags>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\..\core\Akka.TestKit\Akka.TestKit.csproj" />
     <PackageReference Include="xunit" Version="$(XunitVersion)" />
+    <PackageReference Include="FluentAssertions" Version="$(FluentAssertionsVersion)" />
```
Aaronontheweb, line 15: Do we need the dependency on FluentAssertions for the `TaskExtensions` class?

src/core/Akka.Remote.Tests/Transport/ThrottlerTransportAdapterSpec.cs
```diff
@@ -183,23 +186,21 @@ public ThrottlerTransportAdapterSpec(ITestOutputHelper output)
         [Fact]
```
Aaronontheweb, line 186: The tests themselves look fine

Aaronontheweb (commented):
Left some comments about the testkit changes

src/contrib/testkits/Akka.TestKit.Xunit2/Akka.TestKit.Xunit2.csproj
Arkatufus: I guess not, I wanted to introduce the `ShouldCompleteWithin` method in tests; I guess we can use our own `WithTimeout` method.

Aaronontheweb: I'm fine bringing it in since we seem to use it everywhere, but maybe we should make it part of the base testkit if we're going to take a dependency on it. We can punt on it for now and just do our own thing, or we can try to close out https://github.com/akkadotnet/akka.net/issues/5432

Aaronontheweb: How would you like to proceed on that?

Arkatufus: well, we can't because FluentAssertion depends on XUnit, that's why I'm putting it here instead in the base TestKit

Arkatufus: Oops, didn't know that FluentAssertion is unit test framework agnostic... I'm recaftoring this to TestKit instead.

Aaronontheweb:
@Arkatufus merged it

PR #5906, opened 2022-04-30
[Async TestKit] Convert Akka.Stream.TestKit to async - Refactor TestKit.Tests

Arkatufus:
Depends on #5901 to be merged first

- Refactor classes out of Akka.Streams.TestKit.Tests into Akka.Streams.TestKit, these classes are not related to the TestKit tests but are used in other stream unit tests. Decouples other tests from Akka.Streams.TestKit.Tests package.
- Skip all racy tests that relates to `Akka.Streams.TestKit` so we would not waste time waiting for them to clear for each PR
- Convert `Akka.Streams.TestKit.Utils` to async

### DIFF
--- a/src/core/Akka.Persistence.TCK/Performance/JournalPerfSpec.cs
+++ b/src/core/Akka.Persistence.TCK/Performance/JournalPerfSpec.cs
@@ -201,7 +201,7 @@ namespace Akka.Persistence.TestKit.Performance
             });
         }
 
-        [Fact]
+        [Fact(Skip = "Skipped for async_testkit conversion build")]
         public void PersistenceActor_performance_must_measure_PersistAsync()
         {
             var p1 = BenchActor("PersistAsyncPid", EventsCount);
--- a/src/core/Akka.Persistence.TCK/Performance/JournalPerfSpec.cs
+++ b/src/core/Akka.Persistence.TCK/Performance/JournalPerfSpec.cs
@@ -212,7 +212,7 @@ namespace Akka.Persistence.TestKit.Performance
             });
         }
 
-        [Fact]
+        [Fact(Skip = "Skipped for async_testkit conversion build")]
         public void PersistenceActor_performance_must_measure_PersistAllAsync()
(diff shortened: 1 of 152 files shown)


src/core/Akka.Persistence.TCK/Performance/JournalPerfSpec.cs
```diff
@@ -212,7 +212,7 @@ public void PersistenceActor_performance_must_measure_PersistAsync()
             });
         }
 
-        [Fact]
+        [Fact(Skip = "Skipped for async_testkit conversion build")]
```
Arkatufus, line 215: I'm skipping these while we're converting the rest of Streams.Tests over so it wouldn't block every PRs

src/core/Akka.Streams.TestKit.Tests/Akka.Streams.TestKit.Tests.csproj
```diff
@@ -8,7 +8,6 @@
   </PropertyGroup>
   
   <ItemGroup>
-    <EmbeddedResource Include="reference.conf" />
```
Arkatufus, line 11: reference.conf is moved to Akka.Streams.TestKit

src/core/Akka.Streams.Tests.Performance/FlowSelectBenchmark.cs
```diff
@@ -54,7 +54,7 @@ public class FlowSelectBenchmark
         public void Setup(BenchmarkContext context)
         {
             _actorSystem = ActorSystem.Create("FlowSelectBenchmark", Config.WithFallback(
-                ConfigurationFactory.FromResource<ScriptedTest>("Akka.Streams.TestKit.Tests.reference.conf")));
```
Arkatufus, line 57: The magic text for loading the config is moved to a static property in the `StreamTestDefaultMailbox` class

Aaronontheweb (approved):
lgtm

PR #5861, opened 2022-04-21
enable `ChannelTaskScheduler` to work inside Akka.Cluster without causing errors inside `/system` actors

Aaronontheweb:
enable `ChannelTaskScheduler` to work inside Akka.Cluster without causing errors inside `/system` actors

Fixes #5498

## Changes

Modifying the constructors of some `/system` actors to avoid errors when running with the `ChannelTaskScheduler`.

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [ ] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).

Issue #5385, opened 2021-11-15
[PERF] .NET6 Performance Regression


Aaronontheweb:
**Version Information**
Version of Akka.NET? v1.4.28
Which Akka.NET Modules? Akka, Akka.Remote

**Describe the performance issue**
Our [`RemotePingPong` benchmark](https://github.com/akkadotnet/akka.net/blob/dev/src/benchmark/RemotePingPong/Program.cs) has been the standard used for 7 years or so to measure throughput passing over a single Akka.Remote connection between two `ActorSystem` instances. It's a crucial benchmark because it measures the biggest bottleneck in Akka.Remote networks: the end to end response time over a single connection.

Over the lifespan of .NET Core since 2017 we've seen steady improvements in the benchmark numbers each time a new version of the .NET runtime is released usually as an improvement of underlying threading / concurrency / IO primitives introduced into the runtime itself.

With the release of .NET 6, however, we've noticed that while the overall throughput in some measures remains higher than on .NET 5 for these same reasons - there are steady, reproducible, **long-lasting** drops in total throughput that occur only on .NET 6.

**Data and Specs**
Here are the RemotePingPong numbers from my local development machine, a Gen 1 8-core Ryzen, on .NET Core 3.1:

![image](https://user-images.githubusercontent.com/326939/141868542-1f7dec73-8891-4420-af8f-bbe956e86763.png)

**Edit**: update the .NET Core 3.1 benchmark numbers to include the settings from https://github.com/akkadotnet/akka.net/pull/5386

And here are the equivalent numbers for this same benchmark on .NET 6:

![image](https://user-images.githubusercontent.com/326939/141866868-9e0cd0d6-0a68-4c13-9704-b73416d6254d.png)

I've been able to reproduce this consistently - a sustained drop in throughput that lasts for roughly 30s. We've also noticed this in the Akka.NET test suite since merging in https://github.com/akkadotnet/akka.net/pull/5373 - the number of failures in the test suite has grown and has started to include tests that historically have not been racy. We've also observed this separately in the [Phobos](https://phobos.petabridge.com/) repository which we also upgraded to use the .NET 6 SDK.

There is definitely something amiss here with how Akka.NET runs on top of .NET 6.

**Expected behavior**
A consistent level of performance across all benchmarks.

**Actual behavior**
Intermittent lag, declines in throughput, and unexplained novel race conditions.

**Environment**
.NET 6, Windows

**Additional context**
[...]

PR #5908, opened 2022-05-03
[DO NOT MERGE] Test PR for ChannelExecutor as the default Dispatcher executor

Arkatufus:


Aaronontheweb:
testing to see if there are any other issues we need to fix in https://github.com/akkadotnet/akka.net/pull/5861 before we make this the default.

Aaronontheweb:
Now that #5861 has been merged into `dev` this should be a much more interesting test

Aaronontheweb:
https://github.com/akkadotnet/akka.net/issues/5385#issuecomment-1119566953 - ran some benchmarks for this change. Looks like a ~50% drop in Akka.Remote throughput under this configuration... That's a bitter pill to swallow.

PR #5927, opened 2022-05-07
Create Akka_Actor_ILogReceive.md

brah-mcdude:
Added a minimal explanation.


Aaronontheweb:
@Arkatufus this should go in the XML-DOC comments for the `ILogReceive` interface probably, right?

PR #5949, opened 2022-05-18
[PORT] Akka/Akka#27266 - Propagate stream cancellation causes

Arkatufus:
Port of https://github.com/akka/akka/pull/27266 and https://github.com/akka/akka/pull/27547

From the original PR:
>This PR is focussed on adding APIs and machinery to provide cancellation causes across stages and reactive streams interfaces. The new APIs are (so far) completely optional:
> * A reactive streams subscription can implement SubscriptionWithCancelException if it wants to support that feature
> * There's a new GraphStage.cancel(cause) method that can be used to provide a cause (otherwise, SubscriptionWithCancelException.NoMoreElementsNeeded is used)
> * There's a new OutHandler.onDownstreamFinished(cause) method to override if you want access to (or to propagate) the cancellation cause. If not overridden it calls the old method.
> * Similar APIs (SubSink etc.) have already also been adapted.
>
> These changes should be completely transparent and only noticeable to those parties that implement the APIs.

### DIFF
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveStreams.Core.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveStreams.Core.verified.txt
@@ -723,6 +723,10 @@ namespace Akka.Streams
     {
         Akka.Streams.Dsl.Source<TOut, Akka.NotUsed> Source { get; }
     }
+    public interface ISubscriptionWithCancelException : Reactive.Streams.ISubscription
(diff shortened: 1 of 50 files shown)


Aaronontheweb (approved):
LGTM

ismaelhamed:
LGTM. 

This felt like a punishment to review. Now I get why @Aaronontheweb sometimes complains about PRs too big ;)

ismaelhamed (approved):

PR #5958, opened 2022-05-24
Fix broken images in NuGet.org readme

Aaronontheweb:
Changed all file paths to absolute Uris in the `README.md` so they could be referenced from `github.com` directly, a "trusted domain" on NuGet.org.


Arkatufus (approved):
LGTM

Issue #5962, opened 2022-05-25
Error in SplitBrainResolver.PreStart when using ChannelTaskScheduler for internal-dispatcher


SchiessMax:
**Version Information**
Version of Akka.NET? 1.4.38
Which Akka.NET Modules? Akka.Cluster

**Describe the bug**
When using the Akka.Cluster.SBR.SplitBrainResolver while having configured the ChannelTaskScheduler as internal dispatcher an error is thrown in the PreStart method. The error is thrown when `Cluster.Get(Context.System)` is called.
I think this might be related to [#5861](https://github.com/akkadotnet/akka.net/pull/5861) 

The error that is logged:

```
ValueFactory attempted to access the Value property of this instance.
      [akka://CasActorSystem/system/cluster/core/daemon/downingProvider#1885816536]: Akka.Actor.ActorInitializationException: Exception during creation
       ---> System.InvalidOperationException: ValueFactory attempted to access the Value property of this instance.
         at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
         at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
         at System.Lazy`1.CreateValue()
         at System.Lazy`1.get_Value()
         at Akka.Actor.Internal.ActorSystemImpl.TryGetExtension[T](T& extension)
         at Akka.Actor.Internal.ActorSystemImpl.GetExtension[T]()
         at Akka.Actor.ActorSystemWithExtensions.WithExtension[T,TI](ActorSystem system)
         at Akka.Cluster.Cluster.Get(ActorSystem system)
         at Akka.Cluster.SBR.SplitBrainResolver.PreStart()
         at Akka.Actor.ActorBase.AroundPreStart()
         at Akka.Actor.ActorCell.<>c__DisplayClass176_0.<Create>b__0()
         at Akka.Actor.ActorCell.UseThreadContext(Action action)
         at Akka.Actor.ActorCell.Create(Exception failure)
         --- End of inner exception stack trace ---
         at Akka.Actor.ActorCell.Create(Exception failure)
         at Akka.Actor.ActorCell.SysMsgInvokeAll(EarliestFirstSystemMessageList messages, Int32 currentState) 
```

**To Reproduce**
1. Create an ActorSystem with the following configuration: 
```
akka {
    actor {
        provider = cluster
        default-dispatcher = {
            executor = "channel-executor"
            channel-executor.priority = "normal"
[...]

PR #5966, opened 2022-05-26
Spec for #5962, SplitBrainResolver DowningProvider never got restarted after a PreStart failure.

Arkatufus:
Spec for #5962


Aaronontheweb:
Your repro definitely works @Arkatufus

PR #5967, opened 2022-05-26
Allow `PersistentShardCoordinator` to tolerate duplicate `ShardHomeAllocated` messages

### LINKED ISSUE #5604: Akka.Cluster.Sharding: make `PersistentShardCoordinator` a tolerant reader
https://stackoverflow.com/questions/70651266/exception-of-shard-0-already-allocated/70668206#70668206 - this happens often with sharded `ActorSystem`s not shutting down properly and it's a massive footgun.

I feel like the current persistence design is inherently brittle and we've just accepted that as the status quo for years, but I think it could be better.

The problem occurs here 

https://github.com/akkadotnet/akka.net/blob/618776918b5c894f56ddee4cca3838c3e75cdf09/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs#L117

If the previous state hadn't been properly disposed of when the previous `PersistentShardCoordinator` singleton was terminated, the new incarnation of this actor fails upon recovery - this is done in order to guarantee the consistency of shard allocation data, since that's prioritized by the sharding system. If duplicate records are encountered during recovery, for any given `ShardRegion` or `ShardHome`, the recovery process fails and the underlying journal basically has to be repaired.

I think we can modify the recovery process to make this more robust - if we're recovering old records that are going to be overwritten by newer ones that are in the journal, we can make the evaluation at the very end of the recovery process as to whether the recovered state is healthy or not, rather than blowing up the entire recovery process and making the cluster essentially non-operable.

Essentially what I'd like to do is model how to make the `PersistentShardCoordinator` recovery process idempotent and less brittle.

Some examples:

* Duplicate `ShardRegionTerminated` messages don't matter - it's already dead.
* Duplicate `ShardRegionRegistered` messages don't matter - overwrite the old one with the new.
* Duplicate `ShardRegionProxyRegistered` messages don't matter - it's a no-op.

Where things get tricky are the `ShardHomeAllocation` messages - that needs to be either evaluated at the end of the recovery and checked for data consistency or kept as-is. I'd prefer the former.

Using `state-store-mode=ddata` is a viable workaround for this, but the problem is that DurableData does not play nicely at all with K8s due to the complexity of volume mounts and doesn't scale super well for `remember-entities=on` where entity counts are quite large.

https://github.com/petabridge/Akka.Cluster.Sharding.RepairTool - but that's a heavy duty solution to this problem that I'd rather make obsolete through updates to Akka.NET itself.

**Additional context**

Aaronontheweb:
Fixes #5604

## Changes

Allows the `PersistentShardCoordinator` to tolerate multiple `ShardHomeAllocated` messages in the journal for the same shard - this is done in order to allow the sharding system to try its best to restart itself without bringing the system offline in the event of duplicate data appearing in the journal.

## Checklist

* [x] Design discussion issue #5604

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
@@ -109,12 +109,15 @@ namespace Akka.Cluster.Sharding
             }
 
             /// <summary>
-            /// TBD
+            /// Feed an event into the ShardCoordinator state.
             /// </summary>
-            /// <param name="e">TBD</param>
-            /// <exception cref="ArgumentException">TBD</exception>
-            /// <returns>TBD</returns>
-            public State Updated(IDomainEvent e)
+            /// <param name="e">The event to process.</param>
+            /// <param name="isRecovering">A flag to indicate if we're trying to recover state previously stored in the database.
+            /// We need to be more tolerant when this happens in the name of trying to accelerate recovery, so the system doesn't compromise
+            /// itself and go offline.</param>
+            /// <exception cref="ArgumentException">Thrown if an event is illegal in the current state.</exception>
+            /// <returns>An update copy of this state.</returns>
+            public State Updated(IDomainEvent e, bool isRecovering = false)
             {
                 switch (e)
                 {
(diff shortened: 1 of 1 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
```diff
@@ -109,12 +109,15 @@ public IImmutableSet<ShardId> AllShards
...
-            /// <exception cref="ArgumentException">TBD</exception>
-            /// <returns>TBD</returns>
-            public State Updated(IDomainEvent e)
+            /// <param name="e">The event to process.</param>
+            /// <param name="isRecovering">A flag to indicate if we're trying to recover state previously stored in the database.
+            /// We need to be more tolerant when this happens in the name of trying to accelerate recovery, so the system doesn't compromise
+            /// itself and go offline.</param>
+            /// <exception cref="ArgumentException">Thrown if an event is illegal in the current state.</exception>
+            /// <returns>An update copy of this state.</returns>
+            public State Updated(IDomainEvent e, bool isRecovering = false)
```
Aaronontheweb, line 120: We now accept a new parameter here, `isRecovering` - when this value is set to `true` then we enable "reader tolerance" inside the `PersistentShardCoordinator`.

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
```diff
@@ -153,7 +156,27 @@ public State Updated(IDomainEvent e)
                             if (!Regions.TryGetValue(message.Region, out var shardRegions))
                                 throw new ArgumentException($"Region {message.Region} not registered", nameof(e));
                             if (Shards.ContainsKey(message.Shard))
-                                throw new ArgumentException($"Shard {message.Shard} is already allocated", nameof(e));
+                            {
+                                if (!isRecovering)
```
Aaronontheweb, line 160: I've decided to implement reader tolerance in only one location for now the `ShardHomeAllocated` message, since this is where the issue has come up most often and it's also probably the safest place to do it.

src/contrib/cluster/Akka.Cluster.Sharding/PersistentShardCoordinator.cs
```diff
@@ -1346,26 +1369,26 @@ protected override bool ReceiveRecover(object message)
                     switch (evt)
                     {
                         case ShardRegionRegistered _:
```
Aaronontheweb, line 1371: Make sure all of the `Recovery` methods are updated to support the new `isRecovering` flag.

Arkatufus (approved):
LGTM

Aaronontheweb:
Conversation on how safe these changes are https://github.com/akkadotnet/akka.net/pull/5970#discussion_r883747361

PR #5965, opened 2022-05-26
pass `Akka.Cluster.Cluster` into `IDowningProvider` directly

Aaronontheweb:
close #5962

Fixes #5962

## Changes

Passes a `Cluster` instance directly into the SBR constructor in order to avoid same-thread issues with `Lazy<Cluster>`.

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [x] Design discussion issue https://github.com/akkadotnet/akka.net/pull/5861
* [x] Changes in public API reviewed, if any.
* [x] I have added website documentation for this feature.

PR #5968, opened 2022-05-26
Backport of #5965 - pass Akka.Cluster.Cluster into IDowningProvider directly

Arkatufus:
Backport of #5965 

> Passes a Cluster instance directly into the SBR constructor in order to avoid same-thread issues with Lazy<Cluster>.


Aaronontheweb (approved):
LGTM

PR #5970, opened 2022-05-26
Backport of #5967 - Allow PersistentShardCoordinator to tolerate duplicate ShardHomeAllocated messages

Arkatufus:
Backport of #5967

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ShardCoordinator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ShardCoordinator.cs
@@ -1144,12 +1144,15 @@ namespace Akka.Cluster.Sharding
             }
 
             /// <summary>
-            /// TBD
+            /// Feed an event into the ShardCoordinator state.
             /// </summary>
-            /// <param name="e">TBD</param>
-            /// <exception cref="ArgumentException">TBD</exception>
-            /// <returns>TBD</returns>
-            public CoordinatorState Updated(IDomainEvent e)
+            /// <param name="e">The event to process.</param>
+            /// <param name="isRecovering">A flag to indicate if we're trying to recover state previously stored in the database.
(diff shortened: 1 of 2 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/ShardCoordinator.cs
```diff
@@ -1188,7 +1191,26 @@ public CoordinatorState Updated(IDomainEvent e)
...
                             if (Shards.ContainsKey(message.Shard))
-                                throw new ArgumentException($"Shard {message.Shard} already allocated: {this}", nameof(e));
+                            {
+                                if(!isRecovering)
+                                    throw new ArgumentException($"Shard {message.Shard} already allocated: {this}",
+                                        nameof(e));
+                                
+                                // per https://github.com/akkadotnet/akka.net/issues/5604
+                                // we're going to allow new value to overwrite previous
+                                var newRegions = Regions;
```
ismaelhamed, line 1201: A couple questions: 1) Is this safe all the times?; 2) Will there be a way to disable this new behavior?

Aaronontheweb: 1. Not sure - the alternative is defaulting the current behavior, which is that the `ShardCoordinator` blows up and the cluster needs to be restarted and all of the `ShardCoordinator` data deleted. Given that, I think this probably is safer than the current defaults. The worst case scenario I can imagine here is that the `Shard` was _actually_ allocated to two different nodes, but that would if that were the case then the sharding system would already be in very bad shape (i.e. violating its consistency rules) and this would allow the newest home for the shard to supersede the old one, which would stop receiving message traffic. In the logs where I saw this occuring it was duplicate records for the same `Shard` in the same `ShardRegion IActorRef` - so this change would just make the recovery idempotent.

2. We could add it, but I don't think it should be necessary.

Arkatufus:
Could be superseeded by #5976

PR #5971, opened 2022-05-27
Trap all exception thrown while trying to dispatch messages in Akka.Remote.EndpointReader

Arkatufus:
## Changes

Trap all exceptions when we try to dispatch/deserialize an InboundPayload in Akka.Remote.EndpointReader receiver.

### DIFF
--- a/src/core/Akka.Remote/Endpoint.cs
+++ b/src/core/Akka.Remote/Endpoint.cs
@@ -1953,21 +1953,9 @@ namespace Akka.Remote
                                     ackAndMessage.MessageOption.SerializedMessage,
                                     ackAndMessage.MessageOption.SenderOptional);
                             }
-                            catch (SerializationException e)
-                            {
-                                LogTransientSerializationError(ackAndMessage.MessageOption, e);
-                            }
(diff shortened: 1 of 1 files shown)


src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1953,21 +1953,9 @@ private void Reading()
...
-                                LogTransientSerializationError(ackAndMessage.MessageOption, e);
-                            }
-                            catch (InvalidCastException e)
-                            {
-                                LogTransientSerializationError(ackAndMessage.MessageOption, e);
-                            }
                             catch (Exception e)
                             {
-                                throw;
+                                LogTransientSerializationError(ackAndMessage.MessageOption, e);
```
Aaronontheweb, line 1958: Have to just make this a catch all because libraries like Newtonsoft.Json don't throw either of the handled cases above. If you have a serialization problem, it's your fault - we'll log it and not let it blow up the connection. On the user to fix it.

Aaronontheweb (approved):
LGTM

PR #5974, opened 2022-05-31
Model-based tests for `PersistentShardCoordinator.State`

### LINKED ISSUE #5604: Akka.Cluster.Sharding: make `PersistentShardCoordinator` a tolerant reader

Aaronontheweb:
Fixes #5604

## Changes

Added model-based test with FsCheck to validate that `PersistentShardCoordinator.State` is designed correctly.

## Checklist

### DIFF
--- a/src/core/Akka.Tests.Shared.Internals/Helpers/FSharpDelegateHelper.cs
+++ b/src/core/Akka.Tests.Shared.Internals/Helpers/FSharpDelegateHelper.cs
@@ -35,5 +35,13 @@ namespace Akka.Tests.Shared.Internals.Helpers
                 value1 => { return Create<T2, T3, TResult>((value2, value3) => func(value1, value2, value3)); };
             return FSharpFunc<T1, FSharpFunc<T2, FSharpFunc<T3, TResult>>>.FromConverter(conv);
         }
+        
+        public static FSharpFunc<T1, FSharpFunc<T2, FSharpFunc<T3, FSharpFunc<T4, TResult>>>> Create<T1, T2, T3, T4, TResult>(
+            Func<T1, T2, T3, T4, TResult> func)
+        {
+            Converter<T1, FSharpFunc<T2, FSharpFunc<T3, FSharpFunc<T4, TResult>>>> conv =
+                value1 => { return Create<T2, T3, T4, TResult>((value2, value3, value4) => func(value1, value2, value3, value4)); };
+            return FSharpFunc<T1, FSharpFunc<T2, FSharpFunc<T3, FSharpFunc<T4, TResult>>>>.FromConverter(conv);
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


Aaronontheweb:
Tried this with a complex configuration and the tests passed after several minutes.

> Ok, passed 10000 tests (100% long sequences (>6 commands)).

No issues with the `State` class, but I believe I found the error...

Aaronontheweb:
https://github.com/akkadotnet/akka.net/pull/5976 - must be merged first.

src/core/Akka.Tests.Shared.Internals/Helpers/FSharpDelegateHelper.cs
```diff
@@ -35,5 +35,13 @@ public static class FsharpDelegateHelper
                 value1 => { return Create<T2, T3, TResult>((value2, value3) => func(value1, value2, value3)); };
             return FSharpFunc<T1, FSharpFunc<T2, FSharpFunc<T3, TResult>>>.FromConverter(conv);
         }
+        
+        public static FSharpFunc<T1, FSharpFunc<T2, FSharpFunc<T3, FSharpFunc<T4, TResult>>>> Create<T1, T2, T3, T4, TResult>(
```
Aaronontheweb, line 39: Needed to support 4 argument FSharp functions

PR #5984, opened 2022-06-07
[Async TestKit] Convert Akka.Cluster.Tests to async - ClusterSpec

Arkatufus:
## Changes
Change how `Cluster.JoinAsync()` and `Cluster.JoinSeedNodesAsync()` work, does not rely on `MemberRemoved` anymore, work on a fixed timer instead, configurable via the `akka.cluster.seed-node-timeout` HOCON settings.

### DIFF
--- a/src/core/Akka.Cluster/Cluster.cs
+++ b/src/core/Akka.Cluster/Cluster.cs
@@ -9,7 +9,6 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
-using System.Reflection;
 using System.Runtime.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/core/Akka.Cluster/Cluster.cs
```diff
@@ -263,12 +262,24 @@ public void Join(Address address)
...
+        public Task JoinAsync(Address address, CancellationToken token = default)
         {
-            var completion = new TaskCompletionSource<NotUsed>();
-            this.RegisterOnMemberUp(() => completion.TrySetResult(NotUsed.Instance));
-            this.RegisterOnMemberRemoved(() => completion.TrySetException(
-                new ClusterJoinFailedException($"Node has not managed to join the cluster using provided address: {address}")));
+            var completion = new TaskCompletionSource<NotUsed>(TaskCreationOptions.RunContinuationsAsynchronously);
+            
+            var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(token);
+            timeoutCts.CancelAfter(Settings.SeedNodeTimeout);
```
Aaronontheweb, line 270: Should be join-retry timeout

Aaronontheweb (changes requested):
One minor change needed - just need to use the right timeout value

src/core/Akka.Cluster/Cluster.cs
Arkatufus: `akka.cluster.retry-unsuccessful-join-after`?

Arkatufus: Done.

Aaronontheweb (approved):

Aaronontheweb:
LGTM

PR #5989, opened 2022-06-10
[Async TestKit] Modernize TcpStage and TcpListener

Arkatufus:
Changes:
Make TcpListener binding and unbinding code non-blocking by running them on another thread.

### DIFF
--- a/src/core/Akka/IO/Tcp.cs
+++ b/src/core/Akka/IO/Tcp.cs
@@ -667,6 +667,10 @@ namespace Akka.IO
             /// TBD
             /// </summary>
             public static readonly ResumeWriting Instance = new ResumeWriting();
+
+            private ResumeWriting()
+            {
+            }
         }
 
         /// <summary>
@@ -679,7 +683,7 @@ namespace Akka.IO
             /// <summary>
             /// TBD
             /// </summary>
-            public static SuspendReading Instance = new SuspendReading();
+            public static readonly SuspendReading Instance = new SuspendReading();
 
             private SuspendReading()
             {
@@ -695,7 +699,7 @@ namespace Akka.IO
             /// <summary>
             /// TBD
             /// </summary>
-            public static ResumeReading Instance = new ResumeReading();
+            public static readonly ResumeReading Instance = new ResumeReading();
 
             private ResumeReading()
             {
@@ -830,7 +834,11 @@ namespace Akka.IO
             /// <summary>
             /// TBD
             /// </summary>
-            public static WritingResumed Instance = new WritingResumed();
+            public static readonly WritingResumed Instance = new WritingResumed();
+
+            private WritingResumed()
+            {
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


Aaronontheweb (changes requested):
Missing API approval

src/core/Akka/IO/Tcp.cs
```diff
@@ -830,7 +834,11 @@ public class WritingResumed : Event
             /// <summary>
             /// TBD
             /// </summary>
-            public static WritingResumed Instance = new WritingResumed();
+            public static readonly WritingResumed Instance = new WritingResumed();
+
+            private WritingResumed()
```
Aaronontheweb, line 839: These are breaking API changes, but no big deal. Need to be able to do stuff like this in Akka.NET v1.5.

src/core/Akka/IO/TcpListener.cs
```diff
@@ -40,79 +40,136 @@ partial class TcpListener : ActorBase, IRequiresMessageQueue<IUnboundedMessageQu
...
+                yield return saea;
+            }
+        }
+
+        private async Task<Tcp.Bound> BindAsync()
+        {
+            _socket = new Socket(_bind.LocalAddress.AddressFamily, SocketType.Stream, ProtocolType.Tcp) { Blocking = false };
+                
+            _bind.Options.ForEach(x => x.BeforeServerSocketBind(_socket));
+            _socket.Bind(_bind.LocalAddress);
```
Aaronontheweb, line 151: No `await`-able methods in .NET Standard 2.0?

src/core/Akka/IO/TcpListener.cs
```diff
@@ -40,79 +40,136 @@ partial class TcpListener : ActorBase, IRequiresMessageQueue<IUnboundedMessageQu
...
+                    
                     saea.AcceptSocket = null;
-
                     if (!_socket.AcceptAsync(saea))
                         Self.Tell(new SocketEvent(saea));
                     return true;
 
                 case Tcp.ResumeAccepting resumeAccepting:
                     _acceptLimit = resumeAccepting.BatchSize;
+                    // TODO: this is dangerous, previous async args are not disposed and there's no guarantee that they're not still receiving data
```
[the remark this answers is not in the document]
Arkatufus: `_saeas` is an array of `SocketAsyncEventArgs`, assigned by the `Accept` method.
In this case, the old code actually discarded all of the old args while most of them still active accepting connections, we would lose all reference to those event args.
Best case they cause a memory leak, worse case the got finalized and disposed and something breaks.

src/core/Akka/IO/TcpListener.cs
Arkatufus: No, `Bind` and `Listen` are blocking

src/core/Akka/IO/TcpListener.cs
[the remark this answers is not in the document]
Aaronontheweb: aren't these SAEA only handling bind events though, which have no buffers?

Aaronontheweb (approved):

PR #5999, opened 2022-06-14
Add Cluster.Sharding ActorInitializationException spec

Arkatufus:
Add spec to make sure that Cluster.Sharding can survive any transient exceptions thrown from entity constructor and PreStart method

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/PersistentShardSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/PersistentShardSpec.cs
@@ -6,14 +6,17 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Cluster.Tools.Singleton;
 using Akka.Configuration;
+using Akka.Event;
 using Akka.TestKit;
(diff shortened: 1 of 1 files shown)


src/contrib/cluster/Akka.Cluster.Sharding.Tests/PersistentShardSpec.cs
```diff
@@ -87,5 +136,96 @@ public void Persistent_Shard_must_remember_entities_started_with_StartEntity()
...
+                {
+                    case ShardSpec.EntityEnvelope msg:
+                        return msg.Id.ToString();
+                }
+                return null;
+            };            
+            
+            var settings = ClusterShardingSettings.Create(Sys);
+            var tuning = settings.TuningParameters;
+            settings = settings.WithTuningParameters(new TuningParameters
```
Arkatufus, line 166: This is a very bad coding experience. We need to at least add a `Copy()` method to `TuningParameter` class so no one should be forced to write this. Or use the record feature in .NET 5.0

Aaronontheweb (approved):
LGTM - port this to 1.5 also.

Aaronontheweb:
Looks like this test fails under some conditions.

Aaronontheweb (changes requested):
Have an intermittent test failure.

Aaronontheweb (approved):

PR #6001, opened 2022-06-15
Port of #5999: Add Cluster.Sharding ActorInitializationException spec

Arkatufus:
Ports #5999


Aaronontheweb:
@Arkatufus need to re-run API approval specs

Arkatufus:
Done

PR #6009, opened 2022-06-20
Add environment variable controlled skipable LocalFactAttribute

Arkatufus:
This custom Fact attribute will be used for timing sensitive and performace related unit tests so that they can be skipped when running in the CI/CD environment.


src/contrib/testkits/Akka.TestKit.Xunit2/Attributes/LocalFactAttribute.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System;
+using Xunit;
+
+namespace Akka.TestKit.Xunit2.Attributes
+{
+    /// <summary>
+    /// <para>
+    /// This custom XUnit Fact attribute will skip unit tests if the environment variable
+    /// "XUNIT_SKIP_LOCAL_FACTS" exists and is set to the string "true"
```
Aaronontheweb, line 15: Do we want to distribute this as part of the testkit for XUnit generally? Or just internally for Akka.NET use? I'm open to either - just want to make sure I know what you want to do @Arkatufus

Arkatufus: I figured that its a fairly harmless addition, nothing will happen if you use it unless you specifically set the environment variable. We might as well make it a publicly available attribute.

Aaronontheweb: ok, sounds good to me.

PR #6010, opened 2022-06-20
[Async TestKit] Fix WithinAsync and Remaining logic

Arkatufus:
## Changes
- Make sure that `Remaining` always return a positive `TimeSpan`
- Put a limit on how long `WithinAsync` will wait for the function block to complete. Prevents bad unit tests from deadlocking the entire unit test suite.
- Make `WithinAsync` use the `cancellationToken` passed in the function parameter
- Fix some assertion and error message errors
- Warn user if unit test is waiting or peeking with infinite timeout
- Guard against negative duration values in `Dilated`

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/InactiveEntityPassivationSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/InactiveEntityPassivationSpec.cs
@@ -151,7 +151,7 @@ namespace Akka.Cluster.Sharding.Tests
 
             var timeSinceOneSawAMessage = DateTime.Now.Ticks - timeOneSawMessage;
             var time = settings.PassivateIdleEntityAfter - TimeSpan.FromTicks(timeSinceOneSawAMessage) + smallTolerance;
-            return time < TimeSpan.Zero ? TimeSpan.Zero : time;
+            return time < smallTolerance ? smallTolerance : time;
         }
     }
 
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/InactiveEntityPassivationSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/InactiveEntityPassivationSpec.cs
@@ -150,7 +150,8 @@ namespace Akka.Cluster.Sharding.Tests
             probe.ExpectMsg<Entity.GotIt>().Id.ShouldBe("2");
 
             var timeSinceOneSawAMessage = DateTime.Now.Ticks - timeOneSawMessage;
-            return settings.PassivateIdleEntityAfter - TimeSpan.FromTicks(timeSinceOneSawAMessage) + smallTolerance;
+            var time = settings.PassivateIdleEntityAfter - TimeSpan.FromTicks(timeSinceOneSawAMessage) + smallTolerance;
+            return time < TimeSpan.Zero ? TimeSpan.Zero : time;
         }
     }
 
@@ -197,7 +198,7 @@ namespace Akka.Cluster.Sharding.Tests
             var probe = CreateTestProbe();
             var region = Start(probe);
             var time = await TimeUntilPassivate(region, probe);
-            probe.ExpectNoMsg(time);
+            await probe.ExpectNoMsgAsync(time);
         }
     }
 }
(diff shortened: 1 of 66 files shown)


Arkatufus:
Looks like it creeped past the last merge, it should be fixed in #6012

src/contrib/cluster/Akka.Cluster.Sharding.Tests/InactiveEntityPassivationSpec.cs
```diff
@@ -150,7 +150,8 @@ protected async Task<TimeSpan> TimeUntilPassivate(IActorRef region, TestProbe pr
             probe.ExpectMsg<Entity.GotIt>().Id.ShouldBe("2");
 
             var timeSinceOneSawAMessage = DateTime.Now.Ticks - timeOneSawMessage;
-            return settings.PassivateIdleEntityAfter - TimeSpan.FromTicks(timeSinceOneSawAMessage) + smallTolerance;
+            var time = settings.PassivateIdleEntityAfter - TimeSpan.FromTicks(timeSinceOneSawAMessage) + smallTolerance;
+            return time < smallTolerance ? smallTolerance : time;
```
Aaronontheweb, line 154: LGTM - prevents negative timespans from being returned. Nice catch.

src/core/Akka.TestKit/Internal/TimeSpanExtensions.cs
```diff
@@ -130,7 +130,7 @@ public static bool IsInfiniteTimeout(this TimeSpan? timeSpan)
         public static void EnsureIsPositiveFinite(this TimeSpan timeSpan, string parameterName)
         {
             if(!IsPositiveFinite(timeSpan))
-                throw new ArgumentException($"The timespan must be greater than zero. Actual value: {timeSpan}", nameof(parameterName));
```
Aaronontheweb, line 133: Yeah, this was a bug. Nice catch.

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -413,9 +413,15 @@ public TimeSpan Remaining
...
             get
             {
-                // ReSharper disable once PossibleInvalidOperationException
-                if (_testState.End.IsPositiveFinite()) return _testState.End.Value - Now;
-                throw new InvalidOperationException(@"Remaining may not be called outside of ""within""");
+                if(_testState.End is null)
+                    throw new InvalidOperationException(@"Remaining may not be called outside of ""within""");
+                
+                if (_testState.End < TimeSpan.Zero)
+                    throw new InvalidOperationException($"End can not be negative, was: {_testState.End}");
```
Aaronontheweb, line 420: Sanity check

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -448,7 +456,7 @@ protected TimeSpan RemainingOr(TimeSpan duration)
         public TimeSpan RemainingOrDilated(TimeSpan? duration)
         {
             if(!duration.HasValue) return RemainingOrDefault;
-            if(duration.IsInfinite()) throw new ArgumentException("max duration cannot be infinite");
```
Aaronontheweb, line 451: This only considered `-1`, whereas now we check for all possible negative timespan values.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -267,7 +267,7 @@ public async ValueTask<object> ReceiveOneAsync(TimeSpan? max = null, Cancellatio
             }
             else if (maxDuration == Timeout.InfiniteTimeSpan)
             {
-                ConditionalLog(shouldLog, "Trying to receive message from TestActor queue. Will wait indefinitely.");
+                Log.Warning("Trying to receive message from TestActor queue with infinite timeout! Will wait indefinitely!");
```
Aaronontheweb, line 270: Need to explicitly warn here.

src/core/Akka.TestKit/TestKitBase_Within.cs
```diff
@@ -261,14 +271,32 @@ public abstract partial class TestKitBase
...
+            T ret = default;
+            using (var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
             {
-                _testState.End = prevEnd;
+                try
+                {
+                    var executionTask = function();
+                    // Limit the execution time block to the maximum allowed execution time.
+                    // 200 milliseconds is added because Task.Delay() timer is not precise and can return prematurely.
+                    var resultTask = await Task.WhenAny(executionTask, Task.Delay(max + 200.Milliseconds(), cts.Token));
```
Aaronontheweb, line 282: Got it, so this ensures that even if the CTS isn't fired we can still cap the execution time for any spec to `max + 200.milliseconds`

Aaronontheweb (approved):
LGTM - caught a lot of small errors in the testkit itself. Should help a lot with test hygiene going forward. Nice work @Arkatufus

PR #6026, opened 2022-06-27
Add support to `ByteString` for copying to/from `Memory` and `Span`

Arkatufus:
## Changes
Add `CopyTo` and `CopyFrom` `Memory` and `Span` support to `ByteString`


to11mtm:
Should we consider going away from `ArraySegment` for internal storage and use `Memory` instead? On one hand it would be a bit of a yak shave, on the other hand it should greatly simplify the internal logic (since with `Memory` we can slice and not have to consider offset/length in other methods)

Arkatufus:
I don't think we will get any improvement switching over to Memory<T>, even microsoft ReadOnlySequence<T> still uses ArraySegment<T> underneath.

Aaronontheweb:
> I don't think we will get any improvement switching over to Memory, even microsoft ReadOnlySequence still uses ArraySegment underneath.

being able to work with new native APIs in .NET 6 without copying is a big improvement: https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.receiveasync?view=net-6.0#system-net-sockets-socket-receiveasync(system-memory((system-byte))-system-net-sockets-socketflags-system-threading-cancellationtoken)

Arkatufus:
Hmmm... true, as long as we ditch dotnetty

Aaronontheweb (approved):
LGTM

Aaronontheweb:
This is small PR and doesn't do much to change the internals of `ByteString` - we should address the internals of this later, once we are in a better position to support System.Memory throughout the serialization and transport system.

PR #6027, opened 2022-06-27
[Async TestKit] Fix racy specs

Arkatufus:


src/core/Akka.Streams.Tests/Akka.Streams.Tests.csproj
```diff
@@ -35,6 +34,12 @@
     <PackageReference Include="System.Runtime.Extensions" Version="4.3.1" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="xunit.runner.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
Arkatufus, line 39: Embedding `xunit.runner.json` directly inside the project folder to see if test raciness was caused by a missing setting file

src/core/Akka.Streams.Tests/Dsl/FlowAskSpec.cs
```diff
@@ -243,14 +243,14 @@ public void Flow_with_ask_must_produce_future_elements_in_order()
...
 
-        [LocalFact(SkipLocal = "Racy on Azure DevOps")]
+        [Fact]
         public void Flow_with_ask_must_signal_ask_failure() => this.AssertAllStagesStopped(() =>
         {
             var failsOn = ReplierFailOn(1);
             var c = this.CreateManualSubscriberProbe<Reply>();
 
             var p = Source.From(Enumerable.Range(1, 5))
-                .Ask<Reply>(failsOn, _timeout, 4)
```
Arkatufus, line 253: Faulty test. Ask flow was using 4 parallellism but the test expects that the first message will always be an error while running in an async environment.

src/core/Akka.Streams.Tests/Dsl/RestartSpec.cs
```diff
@@ -144,7 +144,7 @@ await probe.AsyncBuilder()
                     .ExecuteAsync();
 
                 // There should be a delay of at least _minBackoff before we receive the element after restart
-                var deadline = (_minBackoff - TimeSpan.FromMilliseconds(1)).FromNow();
+                var deadline = (_minBackoff - TimeSpan.FromMilliseconds(100)).FromNow();
```
Arkatufus, line 147: Added a bit more leeway to the timing.

PR #6028, opened 2022-07-01
Fix faulty AddLogger in LoggingBus

Arkatufus:
Exception handling inside AddLogger is wrong

### DIFF
--- a/src/core/Akka/Event/LoggingBus.cs
+++ b/src/core/Akka/Event/LoggingBus.cs
@@ -9,13 +9,11 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
-using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Actor.Internal;
 using Akka.Configuration;
-using Akka.Dispatch.SysMsg;
 
 namespace Akka.Event
 {
@@ -24,10 +22,29 @@ namespace Akka.Event
     /// </summary>
     public class LoggingBus : ActorEventBus<object, Type>
     {
+        private sealed class LoggerStartInfo
+        {
+            private readonly string _name;
+            
+            public LoggerStartInfo(string loggerName, Type loggerType, IActorRef actorRef)
+            {
+                _name = $"{loggerName} [{loggerType.FullName}]";
+                ActorRef = actorRef;
+            }
+
+            public IActorRef ActorRef { get; }
+            
+            public override string ToString() => _name;
+        }
+        
         private static readonly LogLevel[] AllLogLevels = Enum.GetValues(typeof(LogLevel)).Cast<LogLevel>().ToArray();
 
         private static int _loggerId;
         private readonly List<IActorRef> _loggers = new List<IActorRef>();
+        
+        // Async load support
+        private readonly Dictionary<Task, LoggerStartInfo> _startupState = new Dictionary<Task, LoggerStartInfo>();
+        private readonly CancellationTokenSource _shutdownCts = new CancellationTokenSource();
 
         /// <summary>
         /// The minimum log level that this bus will subscribe to, any <see cref="LogEvent">LogEvents</see> with a log level below will not be subscribed to.
@@ -89,17 +106,19 @@ namespace Akka.Event
         internal void StartDefaultLoggers(ActorSystemImpl system)
         {
             var logName = SimpleName(this) + "(" + system.Name + ")";
-            var logLevel = Logging.LogLevelFor(system.Settings.LogLevel);
             var loggerTypes = system.Settings.Loggers;
             var timeout = system.Settings.LoggerStartTimeout;
             var shouldRemoveStandardOutLogger = true;
+
+            LogLevel = Logging.LogLevelFor(system.Settings.LogLevel);
 
             foreach (var strLoggerType in loggerTypes)
             {
                 var loggerType = Type.GetType(strLoggerType);
                 if (loggerType == null)
                 {
-                    throw new ConfigurationException($@"Logger specified in config cannot be found: ""{strLoggerType}""");
+                    Publish(new Error(null, logName, GetType(), $@"Logger specified in config cannot be found: ""{strLoggerType}"""));
+                    continue;
                 }
 
                 if (typeof(MinimalLogger).IsAssignableFrom(loggerType))
@@ -108,17 +127,22 @@ namespace Akka.Event
                     continue;
                 }
 
-                try
-                {
-                    AddLogger(system, loggerType, logLevel, logName, timeout);
-                }
-                catch (Exception ex)
-                {
-                    throw new ConfigurationException($"Logger [{strLoggerType}] specified in config cannot be loaded: {ex}", ex);
-                }
+                AddLogger(system, loggerType, logName);
             }
 
-            LogLevel = logLevel;
+            if (!Task.WaitAll(_startupState.Keys.ToArray(), timeout))
+            {
+                foreach (var kvp in _startupState)
+                {
+                    if (!kvp.Key.IsCompleted)
+                    {
+                        Publish(new Warning(logName, GetType(),
+                            $"Logger {kvp.Value} did not respond within {timeout} to InitializeLogger(bus), " +
+                            $"{nameof(LoggingBus)} will try and wait until it is ready. Since it start up is delayed, " +
+                            "this logger may not capture all startup events correctly."));
+                    }
+                }
+            }
 
             if (system.Settings.DebugUnhandledMessage)
             {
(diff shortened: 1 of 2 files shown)


src/core/Akka/Event/LoggingBus.cs
```diff
@@ -188,12 +188,14 @@ private void AddLogger(ActorSystemImpl system, Type loggerType, LogLevel logLeve
             object response = null;
             try
             {
-                response = askTask.Result;
```
Arkatufus, line 191: Calling `.Result` throws `AggregateException` instead of the expected `AskTimeoutException` exception.

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -108,28 +107,13 @@ internal void StartDefaultLoggers(ActorSystemImpl system)
                     continue;
                 }
 
-                if (asyncStart)
```
Arkatufus, line 111: This is not relevant anymore, ActorSystem startup will not throw on logger initialization timeout.

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -188,12 +188,14 @@ private void AddLogger(ActorSystemImpl system, Type loggerType, LogLevel logLeve
...
             {
-                response = askTask.Result;
+                response = askTask.ConfigureAwait(false).GetAwaiter().GetResult();
             }
             catch (Exception ex) when (ex is TaskCanceledException || ex is AskTimeoutException)
             {
                 Publish(new Warning(loggingBusName, GetType(),
                      string.Format("Logger {0} [{2}] did not respond within {1} to InitializeLogger(bus)", loggerName, timeout, loggerType.FullName)));
+                logger.Tell(PoisonPill.Instance);
+                return;
```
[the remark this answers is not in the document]
Aaronontheweb: Send it a system message via `ActorSystem.Stop` instead - if the logger initialization is `await`-ed internally that will guarantee that the actor is still shutdown, whereas `PoisonPill.Instance` will wait inside the user messages queue.

src/core/Akka/Event/LoggingBus.cs
Aaronontheweb: When did we make that change? Is there a PR you can link to here?

Aaronontheweb (changes requested):
Have some questions and one minor requested change

Aaronontheweb:
@Arkatufus this will need to be sent to v1.5 as well?

Arkatufus:
This is the original fix: 
https://github.com/akkadotnet/akka.net/commit/a155dcc6787b074c7d4250900632a8a0a069647e#diff-d271e591ce53138713aa0d7b825006f40bc2ed5591cd8cb00d7247d7f4ea900fR110-R132

It was made because `AddLogger` `Ask` timed out, throwing an `AggregateException` instead of an `AskTimeoutException` which was the exception that was expected.

Arkatufus:
@Aaronontheweb
I've decided to change how loggers are loaded, all loggers are loaded asynchronously now. A warning will be emitted for each late responders if they didn't respond within the alloted timeout time.
The con with this approach is that the `Ask` task will be awaited forever for throwing logger actors, since we do not have a way to trap this inside `LoggingBus`, but this is basically what happens with the old implementation, a detached task with no timeout.

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -88,18 +106,19 @@ protected override Type GetClassifier(object @event)
...
 
+            LogLevel = Logging.LogLevelFor(system.Settings.LogLevel);
+
             foreach (var strLoggerType in loggerTypes)
             {
                 var loggerType = Type.GetType(strLoggerType);
                 if (loggerType == null)
                 {
-                    throw new ConfigurationException($@"Logger specified in config cannot be found: ""{strLoggerType}""");
+                    Publish(new Error(null, logName, GetType(), $@"Logger specified in config cannot be found: ""{strLoggerType}"""));
```
Arkatufus, line 120: We log an error instead of stopping the ActorSystem when a logger type can not be found

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -108,33 +127,23 @@ internal void StartDefaultLoggers(ActorSystemImpl system)
...
-                            {
-                                Console.WriteLine($"Logger [{strLoggerType}] specified in config cannot be loaded: {t.Exception}");
-                            }
-                        });
-                }
-                else
+                AddLogger(system, loggerType, logName);
+            }
+
+            if (!Task.WaitAll(_startupState.Keys.ToArray(), timeout))
```
Arkatufus, line 133: We will wait for all loggers to be ready here, limiting the time to the timeout period

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -108,33 +127,23 @@ internal void StartDefaultLoggers(ActorSystemImpl system)
...
                 {
-                    try
-                    {
-                        AddLogger(system, loggerType, logLevel, logName, timeout);
-                    }
-                    catch (Exception ex)
+                    if (!kvp.Key.IsCompleted)
                     {
-                        throw new ConfigurationException($"Logger [{strLoggerType}] specified in config cannot be loaded: {ex}", ex);
+                        Publish(new Warning(logName, GetType(),
```
Arkatufus, line 139: We log a warning if any of the `Ask`-ed loggers hasn't replied after the `Wait`

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -164,46 +173,68 @@ internal void StopDefaultLoggers(ActorSystem system)
                 Publish(new Debug(SimpleName(this), GetType(), $"Shutting down: {Logging.SimpleName(system.Settings.StdoutLogger)} started"));
             }
 
+            // Cancel all pending logger initialization tasks
+            _shutdownCts.Cancel(false);
```
Arkatufus, line 177: This is the mechanism to shutdown all pending `Ask` if they're still in progress even when the ActorSystem is shutting down.

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -164,46 +173,68 @@ internal void StopDefaultLoggers(ActorSystem system)
...
-            SubscribeLogLevelAndAbove(logLevel, logger);
-            Publish(new Debug(loggingBusName, GetType(), $"Logger {loggerName} [{loggerType.Name}] started"));
+        private void AddLogger(ActorSystemImpl system, Type loggerType, string loggingBusName)
+        {
+            var loggerName = CreateLoggerName(loggerType);
+            var logger = system.SystemActorOf(Props.Create(loggerType).WithDispatcher(system.Settings.LoggersDispatcher), loggerName);
+            var askTask = logger.Ask(new InitializeLogger(this), Timeout.InfiniteTimeSpan, _shutdownCts.Token);
             
+            _startupState[askTask] = new LoggerStartInfo(loggerName, loggerType, logger);
+            askTask.ContinueWith(t =>
```
Arkatufus, line 208: We process all `Ask` result with a continuation here

src/core/Akka/Event/LoggingBus.cs
Aaronontheweb: If you consider a logger to be critical infrastructure, isn't this a bad idea? i.e. if I don't have visibility into how my service runs at a basic level, how can I be sure it's working correctly? 

I'd probably leave the old fail-fast mechanic in-place, which should be even rarer now with the changes in Akka.Hosting.

src/core/Akka/Event/LoggingBus.cs
```diff
@@ -164,46 +173,68 @@ internal void StopDefaultLoggers(ActorSystem system)
...
+                {
+                    var info = _startupState[t];
```
Aaronontheweb, line 210: Not safe to do this in a continuation, since this is now shared mutable state.

Aaronontheweb: I'd keep the rest of the continuation but basically just purge the `_startupState` after the `Task.WaitAll`.

src/core/Akka/Event/LoggingBus.cs
Aaronontheweb: This is different than the logger taking too long to startup - nothing is necessarily wrong with the users' configuration in that case. But in this case, the user's config is 100% wrong and they must fix it. That's why throwing an error here is a good idea.

Aaronontheweb (changes requested):
Has some issues

src/core/Akka/Event/LoggingBus.cs
Arkatufus: Alright, I'll bring the exception code back

src/core/Akka/Event/LoggingBus.cs
Arkatufus: It should be safe since this is the only class where it is modified and there is no cross-pollination, each task handles its own continuation and have their own Task as their key.
We had to held on to `_startupState` because of the possibility that some loggers are not ready after the `Task.WaitAll` and they needed to retrieve their info. Would it be safe to use closure instead?

Arkatufus: Removed the dictionary, changed it to closure. Just needed the logger name for logging purposes outside of the Task continuation

Aaronontheweb (approved):
LGTM

Aaronontheweb:
@Arkatufus needs a backport to v1.5

PR #6030, opened 2022-07-04
harden `HubSpec` duration

Aaronontheweb:
## Changes

Needed to boost time to process 20,000 elements with buffer size =1 from 3 cumulative seconds to 300. On low quality AzDo machines there's no telling what delays we'll hit.


Arkatufus (approved):
LGTM

PR #6032, opened 2022-07-05
Improve Akka.Actor.Cancelable

Arkatufus:
## Changes
- Add new Create methods that takes CancellationToken params
- Clean up the Dispose method
- Expose the _isDisposed field as a property

### DIFF
--- a/src/core/Akka/Actor/Cancellation/Cancelable.cs
+++ b/src/core/Akka/Actor/Cancellation/Cancelable.cs
@@ -218,11 +218,9 @@ namespace Akka.Actor
                     }
                     finally
                     {
-                        // base.dispose(disposing);
-                    }
-                    IsDisposed = true;
+                        IsDisposed = true;
+                    }                    
                 }
-                //Clean up unmanaged resources
             }
         }
     }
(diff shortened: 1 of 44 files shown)


src/core/Akka/Actor/Cancellation/Cancelable.cs
```diff
@@ -207,27 +206,17 @@ protected virtual void Dispose(bool disposing)
...
+                    _source?.Dispose();
                 }
-                _isDisposed = true;
-            }
-            finally
-            {
-                // base.dispose(disposing);
+                //Clean up unmanaged resources
             }
+            IsDisposed = true;
```
Aaronontheweb, line 219: Why get rid of the `try...finally` here,  in case `source.Dispose()` throws?

Aaronontheweb (changes requested):
API changes look fine, but have questions on the underlying behavior.

src/core/Akka/Actor/Cancellation/Cancelable.cs
Arkatufus: Hmmm... I can't see a CancellationTokenSource to throw inside its Dispose method, but i'll add the try...catch block back

Arkatufus: To my defense, the original code does not catch any exception thrown from Dispose, it only have an empty finally block

Aaronontheweb (approved):
LGTM

PR #6033, opened 2022-07-05
Make JoinAsync and JoinSeedNodesAsync more robust by using an async state

Arkatufus:
## Changes
- JoinAsync and JoinSeedNodesAsync is now idempotent, always returning the same Task instance when called multiple times.
- More robust memory claen-up

## Still a problem
- There is still no way to remove the cluster message listener actors, these listeners will persist until the Cluster is shut down.
- It is possible that variables inside these listener delegates went out of scope or are nullified, causing NRE to be thrown, filling the log with confusing error messages.

### DIFF
--- a/src/core/Akka.Cluster/Cluster.cs
+++ b/src/core/Akka.Cluster/Cluster.cs
@@ -243,6 +243,9 @@ namespace Akka.Cluster
             ClusterCore.Tell(new ClusterUserAction.JoinTo(FillLocal(address)));
         }
 
+        // This object holds the current asynchronous join state
+        private AsyncJoinState _asyncJoinState;
+        
         /// <summary>
         /// Try to asynchronously join this cluster node specified by <paramref name="address"/>.
         /// A <see cref="Join"/> command is sent to the node to join. Returned task will be completed
@@ -264,27 +267,24 @@ namespace Akka.Cluster
         /// <returns>Task which completes, once current cluster node reaches <see cref="MemberStatus.Up"/> state.</returns>
         public Task JoinAsync(Address address, CancellationToken token = default)
         {
-            var completion = new TaskCompletionSource<NotUsed>(TaskCreationOptions.RunContinuationsAsynchronously);
-            
-            var timeout = Settings.RetryUnsuccessfulJoinAfter ?? TimeSpan.FromSeconds(10);
-            var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(token);
-            timeoutCts.CancelAfter(timeout);
-            timeoutCts.Token.Register(() =>
+            if (_asyncJoinState != null)
             {
-                timeoutCts.Dispose();
-                completion.TrySetException(new ClusterJoinFailedException(
-                    $"Node has not managed to join the cluster using provided address: {address}"));
-            });
+                _log.Error("Another async cluster join is already in progress");
+                return _asyncJoinState.Task;
+            }
             
-            RegisterOnMemberUp(() =>
-            {
-                timeoutCts.Dispose();
-                completion.TrySetResult(NotUsed.Instance);
-            });
+            _asyncJoinState = new AsyncJoinState(
+                cluster: this, 
+                failException: new ClusterJoinFailedException(
+                    $"Node has not managed to join the cluster using provided address: {address}"), 
+                onComplete: () => _asyncJoinState = null,
+                token: token);
 
+            // Guard against possibility that _asyncJoinState got nullified immediately
+            var task = _asyncJoinState.Task;
             Join(address);
 
-            return completion.Task.WithCancellation(token);
+            return task;
         }
 
         private Address FillLocal(Address address)
@@ -333,28 +333,27 @@ namespace Akka.Cluster
         /// </summary>
         /// <param name="seedNodes">TBD</param>
         /// <param name="token">TBD</param>
-        public Task JoinSeedNodesAsync(IEnumerable<Address> seedNodes, CancellationToken token = default(CancellationToken))
+        public Task JoinSeedNodesAsync(IEnumerable<Address> seedNodes, CancellationToken token = default)
         {
-            var completion = new TaskCompletionSource<NotUsed>(TaskCreationOptions.RunContinuationsAsynchronously);
-            
-            var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(token);
-            timeoutCts.CancelAfter(Settings.SeedNodeTimeout);
-            timeoutCts.Token.Register(() =>
+            if (_asyncJoinState != null)
             {
-                timeoutCts.Dispose();
-                completion.TrySetException(new ClusterJoinFailedException(
-                    $"Node has not managed to join the cluster using provided seed node addresses: {string.Join(", ", seedNodes)}."));
-            });
+                _log.Error("Another async cluster join is already in progress");
+                return _asyncJoinState.Task;
+            }
             
-            RegisterOnMemberUp(() =>
-            {
-                timeoutCts.Dispose();
-                completion.TrySetResult(NotUsed.Instance);
-            });
+            var nodes = seedNodes.ToList();
+            _asyncJoinState = new AsyncJoinState(
+                cluster: this, 
+                failException: new ClusterJoinFailedException(
+                    $"Node has not managed to join the cluster using provided seed node addresses: {string.Join(", ", nodes)}."), 
+                onComplete: () => _asyncJoinState = null,
+                token: token);
 
-            JoinSeedNodes(seedNodes);
+            // Guard against possibility that _asyncJoinState got nullified immediately
+            var task = _asyncJoinState.Task;
+            JoinSeedNodes(nodes);
 
-            return completion.Task.WithCancellation(token);
+            return task;
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
> It is possible that variables inside these listener delegates went out of scope or are nullified, causing NRE to be thrown, filling the log with confusing error messages, and if not guarded properly, propagates to the ClusterDaemon and kills it.

Unlikely with closures, but theoretically possible. I wouldn't worry about it - end-user's responsibility.

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -264,27 +267,24 @@ public void Join(Address address)
...
-            var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(token);
-            timeoutCts.CancelAfter(timeout);
-            timeoutCts.Token.Register(() =>
+            if (_asyncJoinState != null)
             {
-                timeoutCts.Dispose();
-                completion.TrySetException(new ClusterJoinFailedException(
-                    $"Node has not managed to join the cluster using provided address: {address}"));
-            });
+                _log.Error("Another async cluster join is already in progress");
```
Aaronontheweb, line 272: I'd just log a `DEBUG` here and return the task without making a fuss, in case this came up.

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -264,27 +267,24 @@ public void Join(Address address)
...
+                return _asyncJoinState.Task;
```
Aaronontheweb, line 273: But if `_asyncJoinState` gets nulled out upon completion, won't this fail?

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -333,28 +333,27 @@ public void JoinSeedNodes(IEnumerable<Address> seedNodes)
         /// </summary>
         /// <param name="seedNodes">TBD</param>
         /// <param name="token">TBD</param>
-        public Task JoinSeedNodesAsync(IEnumerable<Address> seedNodes, CancellationToken token = default(CancellationToken))
+        public Task JoinSeedNodesAsync(IEnumerable<Address> seedNodes, CancellationToken token = default)
         {
-            var completion = new TaskCompletionSource<NotUsed>(TaskCreationOptions.RunContinuationsAsynchronously);
```
Aaronontheweb, line 338: Something we don't have today that we should add - a bailout condition in the event that we're already a part of the cluster. Either we should just return an immediate `Task.Completed` or an exception. I'd probably prefer to do the former and keep chugging along.

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -333,28 +333,27 @@ public void JoinSeedNodes(IEnumerable<Address> seedNodes)
...
-
-            JoinSeedNodes(seedNodes);
-
-            return completion.Task.WithCancellation(token);
+            var nodes = seedNodes.ToList();
+            _asyncJoinState = new AsyncJoinState(
+                cluster: this, 
+                failException: new ClusterJoinFailedException(
+                    $"Node has not managed to join the cluster using provided seed node addresses: {string.Join(", ", nodes)}."), 
+                onComplete: () => _asyncJoinState = null,
```
Aaronontheweb, line 349: I wouldn't null the state out if we're going to return this value downstream.

Aaronontheweb (changes requested):
These changes don't look 100% safe to me

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -264,27 +267,24 @@ public void Join(Address address)
         /// <returns>Task which completes, once current cluster node reaches <see cref="MemberStatus.Up"/> state.</returns>
         public Task JoinAsync(Address address, CancellationToken token = default)
         {
-            var completion = new TaskCompletionSource<NotUsed>(TaskCreationOptions.RunContinuationsAsynchronously);
-            
-            var timeout = Settings.RetryUnsuccessfulJoinAfter ?? TimeSpan.FromSeconds(10);
-            var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(token);
-            timeoutCts.CancelAfter(timeout);
-            timeoutCts.Token.Register(() =>
+            if (_asyncJoinState != null)
```
to11mtm, line 270: This could still fail under a number of scenarios including two different threads calling JoinAsync 'close enough'

Given this is not a hot path, we probably could get away with a quick `lock` here; An uncontended lock will be fast (Which is the happy path) and is only needed for the time it takes to check the state and new up the Task/AsyncState

src/core/Akka.Cluster/Cluster.cs
to11mtm: I think it would be more polite to return Task.Completed, IIRC that is more or less what existing behavior is.

src/core/Akka.Cluster/Cluster.cs
Arkatufus: I'll add a lock to make sure that it is more thread safe

src/core/Akka.Cluster/Cluster.cs
Arkatufus: The state object is there as a marker that a join operation is already in progress, the most important part is the `Task` object from the `TaskCompletionSource` inside it.
Its totally fine that it is nulled.

src/core/Akka.Cluster/Cluster.cs
Arkatufus: That sounds like a better idea, I'll try to revamp this PR to fit around that idea instead.

Arkatufus:
Changed how the state is handled, no longer using null as a marker.
Note that the code assumes that you can only join a cluster successfully once, just like leaving.

@Aaronontheweb @to11mtm, Would be great if you can re-review this code again.

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -143,6 +143,17 @@ public Cluster(ActorSystemImpl system)
...
             system.RegisterOnTermination(Shutdown);
 
+            // Async join support
+            RegisterOnMemberUp(() =>
+            {
+                lock (_asyncJoinLock)
+                {
+                    _isUp.GetAndSet(true);
+                    // If there is an async join operation in progress, complete it.
+                    _asyncJoinTaskSource?.Complete();
```
to11mtm, line 153: This looks a -little- closer...

IMO Ideally, we should only lock long enough to see if there is an existing continuation set up, and if so re-use that.

Does `_asyncJoinTaskSource` have a `TryComplete` to use instead?

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -264,27 +275,27 @@ public void Join(Address address)
...
         public Task JoinAsync(Address address, CancellationToken token = default)
         {
-            var completion = new TaskCompletionSource<NotUsed>(TaskCreationOptions.RunContinuationsAsynchronously);
-            
-            var timeout = Settings.RetryUnsuccessfulJoinAfter ?? TimeSpan.FromSeconds(10);
-            var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(token);
-            timeoutCts.CancelAfter(timeout);
-            timeoutCts.Token.Register(() =>
+            if (_isUp || _isTerminated)
+                return Task.CompletedTask;
```
to11mtm, line 279: Should we return a Task with error if the system is terminated here?

src/core/Akka.Cluster/Cluster.cs
Arkatufus: It uses `TryComplete` internally, the lock is to guard against possible null value

Arkatufus: `_asyncJoinTaskSource` is actually a custom wrapper class around `TaskCompletionSource` to provide timeout with custom exception support.

Aaronontheweb:
Here is how I would simplify this:

1. Update `MemberUp` listeners for callbacks to terminate themselves once fired - no need to have them around forever once they've processed their event.
2. Start a new one for each `JoinAsync` attempt - no need for shared state concurrency that way. They'll all get the same event regardless of who started the join.
3. If we're already part of a cluster, just return a completed task (we're not going to handle edge cases where a user wants us to join a node different than the ones we're already in a cluster with - they can read the logs.) You can determine this by checking [`Cluster.SelfMember`](https://getakka.net/api/Akka.Cluster.Cluster.html#Akka_Cluster_Cluster_SelfMember).

Arkatufus:
I've checked the cluster status change listener actor, its appropriately designed, no problem there.

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -421,7 +433,10 @@ private Task LeaveSelf()
                 return leaveTask;
 
             // Subscribe to MemberRemoved events
-            _clusterDaemons.Tell(new InternalClusterAction.AddOnMemberRemovedListener(() => tcs.TrySetResult(null)));
+            _clusterDaemons.Tell(new InternalClusterAction.AddOnMemberRemovedListener(() =>
+            {
+                tcs.TrySetResult(null);
```
Aaronontheweb, line 438: LGTM

Aaronontheweb (approved):
LGTM

PR #6037, opened 2022-07-08
IAsyncEnumerable

eaba:
## Changes

IAsyncEnumerable<T> as a streams source

### DIFF
--- a/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
+++ b/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
@@ -3789,7 +3789,7 @@ namespace Akka.Streams.Implementation.Fusing
                 try
                 {
                     _currentIterator = new IterableAdapter<TOut>(_plainConcat(Grab(_stage._in)).GetAsyncEnumerator());
-                    PushPull();
+                    PushPull().Wait();
                 }
                 catch (Exception ex)
                 {
@@ -3816,11 +3816,11 @@ namespace Akka.Streams.Implementation.Fusing
 
             public override void OnUpstreamFinish()
             {
-                if (!HasNext)
+                if (!HasNext().Result)
                     CompleteStage();
             }
 
-            public override void OnPull() => PushPull();
+            public override void OnPull() => PushPull().Wait();
 
             private void RestartState()
             {
@@ -3828,14 +3828,21 @@ namespace Akka.Streams.Implementation.Fusing
                 _currentIterator = null;
             }
 
-            private bool HasNext => _currentIterator != null && _currentIterator.HasNext();
+            private async Task<bool> HasNext()
+            { 
+                if(_currentIterator == null)
+                    return false;   
+                
+               return await _currentIterator.HasNextAsync();
+            }
 
-            private void PushPull()
+
+            private async Task PushPull()
             {
-                if (HasNext)
+                if (await HasNext())
                 {
                     Push(_stage._out, _currentIterator.Next());
-                    if (!HasNext && IsClosed(_stage._in))
+                    if (!await HasNext() && IsClosed(_stage._in))
                         CompleteStage();
                 }
                 else if (!IsClosed(_stage._in))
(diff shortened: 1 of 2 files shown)


src/core/Akka.Streams/Util/IteratorAdapter.cs
```diff
@@ -91,4 +100,66 @@ public T Next()
             return _enumerator.Current;
         }
     }
+    internal class IterableAdapter<T> : IIterator<T>
```
Aaronontheweb, line 103: What is this class for?

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3762,6 +3762,140 @@ public RecoverWith(Func<Exception, IGraph<SourceShape<TOut>, TMat>> partialFunct
...
+                SetHandler(stage._in, this);
+                SetHandler(stage._out, this);
+            }
+
+            public override void OnPush()
+            {
+                try
+                {
+                    _currentIterator = new IterableAdapter<TOut>(_plainConcat(Grab(_stage._in)).GetAsyncEnumerator());
+                    PushPull().Wait();
```
Aaronontheweb, line 3792: No, can't block here. Use the `GetAsyncCallBack` and create a delegate.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3762,6 +3762,140 @@ public RecoverWith(Func<Exception, IGraph<SourceShape<TOut>, TMat>> partialFunct
...
+                            break;
+                        default:
+                            throw new ArgumentOutOfRangeException();
+                    }
+                }
+            }
+
+            public override void OnUpstreamFinish()
+            {
+                if (!HasNext().Result)
```
Aaronontheweb, line 3819: Again, don't block here. This can cause a deadlock when terminating the stream graph if the upstream `IAsyncEnumerable` ahs not been disposed. 

But the business logic is wrong anyway - there can't be an upstream if we're the `Source`.

Aaronontheweb: Actually, it looks like this stage is a Flow and not a Source - why is that?

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3762,6 +3762,140 @@ public RecoverWith(Func<Exception, IGraph<SourceShape<TOut>, TMat>> partialFunct
...
+                    CompleteStage();
+            }
+
+            public override void OnPull() => PushPull().Wait();
```
Aaronontheweb, line 3823: Can't make this blocking. Has to be async.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3762,6 +3762,140 @@ public RecoverWith(Func<Exception, IGraph<SourceShape<TOut>, TMat>> partialFunct
     /// </summary>
     /// <typeparam name="TIn">TBD</typeparam>
     /// <typeparam name="TOut">TBD</typeparam>
+    [InternalApi]
+    public sealed class StatefulSelectManyAsync<TIn, TOut> : GraphStage<FlowShape<TIn, TOut>>
```
Aaronontheweb, line 3766: This is designed wrong. Must be a source, not a flow.

Aaronontheweb (changes requested):
Needs to be rewritten.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
to11mtm: @Aaronontheweb  FWIW there are situations where it -may- make sense to unfold an async enumerator from an earlier stage... i.e. I wouldn't mind a `SelectMany` that takes `IAsyncEnumerable` from a previous stage and behaves the way the existing one does.

Agreed this probably needs a redesign though. I'd assume that handling an `IAsyncEnumerable` would be almost like a special case of `UnfoldResourceAsync`, at least in naieve implementation.

PR #6040, opened 2022-07-12
Fix InterceptAsync and TestKit.RemainingOrDilated

Arkatufus:
Superceeds #6039

## Changes
- Change `RemainingOrDilated` behavior so it is consistent with `RemainingOrDefault`.
- Fix WithinAsync to use `GetAwaiter().GetResult()` instead of `Result` to automatically unroll `AggregatedException`.
- Fix `EventFilterApplier.InterceptAsync` to consider `WithinAsync` `Remaining` value as its timeout value.
- Add unit tests

### DIFF
--- a/src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs
+++ b/src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs
@@ -442,7 +442,7 @@ namespace Akka.TestKit.Internal
                 ? TestKitExtension.For(system).TestEventFilterLeeway
                 : _testkit.TestKitSettings.TestEventFilterLeeway;
 
-            var timeoutValue = timeout.HasValue ? _testkit.Dilated(timeout.Value) : leeway;
+            var timeoutValue = timeout.HasValue ? _testkit.Dilated(timeout.Value) : _testkit.RemainingOrDilated(leeway);
             matchedEventHandler ??= new MatchedEventHandler();
             system.EventStream.Publish(new Mute(_filters));
             try
--- a/src/core/Akka.TestKit/TestKitBase_Within.cs
+++ b/src/core/Akka.TestKit/TestKitBase_Within.cs
@@ -283,7 +283,7 @@ namespace Akka.TestKit
 
                     if (resultTask == executionTask)
                     {
-                        ret = executionTask.Result;
+                        ret = executionTask.GetAwaiter().GetResult();
                     }
                     else
                     {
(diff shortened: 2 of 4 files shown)


src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs
```diff
@@ -442,7 +442,7 @@ public EventFilterFactory And
                 ? TestKitExtension.For(system).TestEventFilterLeeway
                 : _testkit.TestKitSettings.TestEventFilterLeeway;
 
-            var timeoutValue = timeout.HasValue ? _testkit.Dilated(timeout.Value) : leeway;
+            var timeoutValue = timeout.HasValue ? _testkit.Dilated(timeout.Value) : _testkit.RemainingOrDilated(leeway);
```
Arkatufus, line 445: Check to see if we're inside a `Within` block by using `RemainingOrDilated`

src/core/Akka.TestKit/TestKitBase_Within.cs
```diff
@@ -283,7 +283,7 @@ public abstract partial class TestKitBase
 
                     if (resultTask == executionTask)
                     {
-                        ret = executionTask.Result;
+                        ret = executionTask.GetAwaiter().GetResult();
```
Arkatufus, line 286: Use `GetAwaiter().GetResult()` to unroll any `AggregateException` that might be thrown from inside the async Task

src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs
```diff
@@ -442,7 +442,23 @@ public EventFilterFactory And
...
+            TimeSpan timeoutValue;
+            if (timeout.HasValue)
+            {
+                timeoutValue = _testkit.Dilated(timeout.Value);
+            }
+            else
+            {
+                timeoutValue = _testkit.RemainingOrDilated(leeway);
+                // If we expected 0 occurence and we're inside a Within block, timeout need to be slightly less else Within will throw
+                if (_testkit.IsInsideWithin && expectedOccurrences == 0)
```
Arkatufus, line 454: Special check when occurance is 0 and we're within a Within block. timeout need to be slightly less or Within WILL throw

src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs
```diff
@@ -442,7 +442,23 @@ public EventFilterFactory And
...
 
-            var timeoutValue = timeout.HasValue ? _testkit.Dilated(timeout.Value) : leeway;
+            TimeSpan timeoutValue;
+            if (timeout.HasValue)
+            {
+                timeoutValue = _testkit.Dilated(timeout.Value);
+            }
+            else
+            {
+                timeoutValue = _testkit.RemainingOrDilated(leeway);
```
Arkatufus, line 452: Check if we're inside a Within block, if we do, use Remaining, else use a dilated leeway

Aaronontheweb:
@Arkatufus 

this looks suspicious

```
DocsExamples.Actors.ReceiveTimeoutSpecs.ShouldReceiveTimeoutActors
Failed: Expected no messages during 00:00:00.0500000, instead we received <timeout> from [akka://test/user/receive-timeout#219375173] after 00:00:00.9111398
Expected: True
Actual:   False
```

PR #6043, opened 2022-07-13
[BACKPORT #6038] SSL Configuration fails even EnbleSsl property is set to false

eaba:
## Changes

SSL Configuration Fails even EnbleSsl property is set to false (https://github.com/akkadotnet/akka.net/commit/b21f5fae69a1c2ad89d346bf1f0fd81d62e5d921)


Arkatufus (changes requested):
The actual fix port themselves looks fine, but there are some spec failure.

Aaronontheweb:
Need to rewrite the tests for v1.4 to not use any of the async methods in the TestKit, as those are v1.5-specific

Aaronontheweb (approved):
LGTM

PR #6044, opened 2022-07-13
AsyncEnumerable

eaba:
## Changes

IAsyncEnumerable as a streams source


src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3756,7 +3757,126 @@ public RecoverWith(Func<Exception, IGraph<SourceShape<TOut>, TMat>> partialFunct
...
 
+        /// <summary>
+        /// TBD
+        /// </summary>
+        /// <param name="asyncEnumerable">TBD</param>
+        public AsyncEnumerable(IAsyncEnumerable<T> asyncEnumerable)
+        {
+            //TODO: when to dispose async enumerable? Should this be a part of ownership of current stage, or should it
+            // be a responsibility of the caller?
+            _asyncEnumerable = asyncEnumerable;
```
to11mtm, line 3860: @Aaronontheweb :

Not entirely related but Should we be taking a `Func<IAsyncEnumerable<T>>` here? (In my opinion yes, in ideal-state, a Source itself should be idempotent and able to be materialized multiple times.)

As to this question, for whatever my opinion is worth I'd *think* that since we are consuming the stream till the end, we should be disposing when we are done whether it is provided by a func producer or live instance.

Aaronontheweb:
the code in this PR is copied from @Horusiath [Akka.Persistence.Pulsar](https://github.com/Horusiath/Akka.Persistence.Pulsar) - and while the copyright headers and licenses are identical to Akka.NET's we'll need to obtain his permission before proceeding.

Horusiath:
@Aaronontheweb @eaba I don't mind coping the code. Feel free to use it.

Aaronontheweb:
thanks @Horusiath - we'll move forward with these changes then.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
Aaronontheweb: > Not entirely related but Should we be taking a Func<IAsyncEnumerable<T>> here? (In my opinion yes, in ideal-state, a Source itself should be idempotent and able to be materialized multiple times.)

Now that is a good question. We should absolutely program this component to accept this `Func` so we can restart the stream in the event of errors. That'll also make disposal of the `IAsyncEnumerable` less ambigious too: we created it, we own it, so we clean it up.

Aaronontheweb: @eaba can you implement this change?

Aaronontheweb:
Weird that we'd have that problem given that we already have `IAsyncEnumerable` support elsewhere in Akka.Streams

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -26,19 +26,20 @@
 
 namespace Akka.Streams.Tests.Dsl
 {
-#if NETCOREAPP
```
Aaronontheweb, line 24: You need to add this back @eaba  - your new `#if` is what's causing the compilation error.

Aaronontheweb:
Current test failures are unrelated to your PR I think - looks like mostly Akka.Persistence issues.

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveStreams.Core.verified.txt
```diff
@@ -3983,6 +3984,15 @@ namespace Akka.Streams.Implementation.Fusing
         public override string ToString() { }
     }
     [Akka.Annotations.InternalApiAttribute()]
+    public sealed class AsyncEnumerable<T> : Akka.Streams.Stage.GraphStage<Akka.Streams.SourceShape<T>>
+    {
+        public AsyncEnumerable(System.Func<System.Collections.Generic.IAsyncEnumerable<T>> factory) { }
```
Aaronontheweb, line 3989: LGTM

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3756,7 +3757,127 @@ public RecoverWith(Func<Exception, IGraph<SourceShape<TOut>, TMat>> partialFunct
...
+                    }
+                    else
+                    {
+                        // if result is false, it means enumerator was closed. Complete stage in that case.
+                        CompleteStage();
+                    }
+                }
+                else
+                {
+                    vtask.AsTask().ContinueWith(_handleContinuation);
```
Aaronontheweb, line 3828: This would probably be better with a local function that does `await` but we can refactor this later.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3756,7 +3757,127 @@ public RecoverWith(Func<Exception, IGraph<SourceShape<TOut>, TMat>> partialFunct
...
+        private readonly Outlet<T> _outlet = new Outlet<T>("EnumerableSource.out");
+        private readonly Func<IAsyncEnumerable<T>> _factory;
 
+        /// <summary>
+        /// TBD
+        /// </summary>
+        /// <param name="factory">TBD</param>
+        public AsyncEnumerable(Func<IAsyncEnumerable<T>>  factory)
+        {
+            //TODO: when to dispose async enumerable? Should this be a part of ownership of current stage, or should it
```
Aaronontheweb, line 3858: No need to have this here anymore. We're now responsible for disposing the `IAsyncEnumerable` - hence the factory.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3756,7 +3757,127 @@ public RecoverWith(Func<Exception, IGraph<SourceShape<TOut>, TMat>> partialFunct
...
+            //_asyncEnumerable = asyncEnumerable;
+            _factory = factory;
+            Shape = new SourceShape<T>(_outlet);
+        }
+
+        public override SourceShape<T> Shape { get; }
+        protected override GraphStageLogic CreateLogic(Attributes inheritedAttributes) => new Logic(Shape, _factory().GetAsyncEnumerator());
+
+        /// <summary>
+        /// TBD
```
Aaronontheweb, line 3869: Delete - this is less useful than having no comments.

Aaronontheweb (approved):
LGTM

PR #6045, opened 2022-07-18
cleaned up `IAsyncEnumerable` Source to use local functions

Aaronontheweb:
## Changes

Better to use `await` around `ValueTask` rather than converting to `Task` and doing `ContinueWith`

Cleanup from #6044

## Checklist

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
@@ -10,19 +10,16 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Pattern;
-using Akka.Routing;
 using Akka.Streams.Dsl;
 using Akka.Streams.TestKit;
 using Akka.TestKit;
 using FluentAssertions;
-using Nito.AsyncEx.Synchronous;
 using Xunit;
 using Xunit.Abstractions;
 using System.Collections.Generic;
-using Akka.Actor;
-using Akka.Streams.Actors;
-using Akka.Streams.Tests.Actor;
-using Reactive.Streams;
+using System.Runtime.CompilerServices;
+using Akka.Util;
+using FluentAssertions.Extensions;
 
 namespace Akka.Streams.Tests.Dsl
 {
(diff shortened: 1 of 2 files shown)


src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3775,7 +3775,6 @@ private sealed class Logic : OutGraphStageLogic
             private readonly Action<T> _onSuccess;
             private readonly Action<Exception> _onFailure;
             private readonly Action _onComplete;
-            private readonly Action<Task<bool>> _handleContinuation;
 
             public Logic(SourceShape<T> shape, IAsyncEnumerator<T> enumerator) : base(shape)
```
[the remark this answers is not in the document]
Aaronontheweb: Absolutely we need to do that. I made a mistake looking over @eaba's PR and assumed it did that already.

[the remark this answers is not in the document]
Aaronontheweb: Looks like we already do this in the stream stage - it's where we create the `Logic` that we should do this. So this is a smaller refactor than I thought.

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -150,47 +152,128 @@ async Task ShouldThrow()
...
 
-            var subscription = subscriber.ExpectSubscription();
+            var subscription = await subscriber.ExpectSubscriptionAsync();
             subscription.Request(100);
-            for (int i = 1; i <= 20; i++)
+            await subscriber.ExpectCompleteAsync();
+        }
+
+        [Fact]
+        public async Task AsyncEnumerableSource_Must_Process_All_Elements()
```
Aaronontheweb, line 177: LGTM

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -150,47 +152,128 @@ async Task ShouldThrow()
...
+
+            var subscription = await subscriber.ExpectSubscriptionAsync();
+            subscription.Request(101);
+
+            await subscriber.ExpectNextNAsync(Enumerable.Range(0, 50));
+
+            var exception = await subscriber.ExpectErrorAsync();
+            
+            // Exception should be automatically unrolled, this SHOULD NOT be AggregateException
+            exception.Should().BeOfType<TestException>();
```
Aaronontheweb, line 210: LGTM

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -150,47 +152,128 @@ async Task ShouldThrow()
...
+            exception.Message.Should().Be("BOOM!");
+        }
+
+        [Fact]
+        public async Task AsyncEnumerableSource_Must_Cancel_Running_Source_If_Downstream_Completes()
```
Aaronontheweb, line 215: LGTM

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3822,24 +3815,55 @@ public override void OnPull()
                         // if result is false, it means enumerator was closed. Complete stage in that case.
                         CompleteStage();
                     }
+                } 
+                else if (vtask.IsCompleted) // IsCompleted covers Faulted, Cancelled, and RanToCompletion async state
```
Aaronontheweb, line 3819: Nice catch - I totally missed this case.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3822,24 +3815,55 @@ public override void OnPull()
...
+                {
+                    // vtask will always contains an exception because we know we're not successful and always throws
+                    try
+                    {
+                        // This does not block because we know that the task already completed
```
Aaronontheweb, line 3824: LGTM

Aaronontheweb (commented):
LGTM

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3775,24 +3776,16 @@ private sealed class Logic : OutGraphStageLogic
...
             private readonly Action _onComplete;
-            private readonly Action<Task<bool>> _handleContinuation;
+            private readonly CancellationTokenSource _completionCts;
 
-            public Logic(SourceShape<T> shape, IAsyncEnumerator<T> enumerator) : base(shape)
+            public Logic(SourceShape<T> shape, IAsyncEnumerable<T> enumerable) : base(shape)
             {
-                _enumerator = enumerator;
+                _completionCts = new CancellationTokenSource();
+                _enumerator = enumerable.GetAsyncEnumerator(_completionCts.Token);
```
to11mtm, line 3784: Interesting question, should this be in prestart vs the constructor?

Aaronontheweb: The only thing that would throw, I guess, would be if there was something wrong with the `IAsyncEnumerable` (i.e. `null`)?

In that case I guess it'd be better to have this run in `PreStart`

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3822,24 +3815,55 @@ public override void OnPull()
...
-                {
-                    CompleteStage(); // if dispose completed immediately, complete stage directly
-                }
-                else
-                {
-                    // for async disposals use async callback
-                    vtask.GetAwaiter().OnCompleted(_onComplete);
-                }
+                _completionCts.Cancel();
+                _completionCts.Dispose();
```
to11mtm, line 3865: Do we have guarantees that cancelling will dispose the enumerator if needed?

Arkatufus: Oh, you're totally correct, I forgot that we need to dispose the enumerator... I guess the previous code is still needed here.

Arkatufus:
```
            public override void OnDownstreamFinish(Exception cause)
            {
                _completionCts.Cancel();
                _completionCts.Dispose();
                CompleteStage();
                _enumerator.DisposeAsync().AsTask().ContinueWith(t => _baseOnDownstreamFinishCallback(cause));
            }
```
or
```
            public override void OnDownstreamFinish(Exception cause)
            {
                _completionCts.Cancel();
                _completionCts.Dispose();
                CompleteStage();

                _enumerator.DisposeAsync().GetAwaiter().GetResult();
                base.OnDownstreamFinish(cause);
            }
```

Arkatufus: I will add the try...catch later, just wanted to get the point across first

Aaronontheweb: I think we should go with the latter - I doubt there will be `IAsyncEnumerator.DisposeAsync` implementations that can block for prolonged periods of time.

Aaronontheweb (commented):
LGTM

PR #6048, opened 2022-07-22
[BACKPORT  #6045] cleaned up `IAsyncEnumerable` Source to use local functions

Arkatufus:
Cherry-picked from https://github.com/akkadotnet/akka.net/commit/0c92aac98c9421a4112692155ec33b28dbc4940d


Aaronontheweb (approved):
LGTM - provided that we're using the correct TestKit methods.

Arkatufus:
We should be, had to downgrade all the async testkit methods

PR #6050, opened 2022-07-27
Singleton class and settings based on current Akka Typed implementation

### LINKED ISSUE #6036: Akka.Cluster.Singleton: should singleton creation match what we do with `ShardRegion`s?

In Akka.Cluster.Sharding, if you are creating a `ShardRegion` that can only be hosted on nodes with role `foo` from a node that __does not__ have a role of `foo`, you get a `ShardRegionProxy` back in exchange.

Akka.Cluster.Tools could work the same way with `SingletonManager`s and `SingletonProxy`s, potentially. I think it may not be 100% feasible without creating some type of `ClusterSingleton` extension for the `ActorSystem` to handle this, similarly to how `ShardRegion`s are handled. But would this make the creation of singletons easier and less error-prone for end users?

Worth noting, this is exactly how we do it inside Akka.Hosting right now - although we always start a singleton + proxy at the same time on each node anyway:

https://github.com/akkadotnet/Akka.Hosting/blob/db3d82780fc29445cbfbf826e029012bfb314057/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs#L319-L356

Maybe just sticking with Akka.Cluster.Hosting for this.

ismaelhamed:
Fixes #6036

## Changes

Since this already exists in Akka Typed and it isn't built on top of anything Akka-Typed specific, I put together a quick port with this functionality:

- If there already is a manager running for the given `singletonName` on this node, no additional manager is started.
- If there already is a proxy running for the given `singletonName` on this node, an IActorRef to that is returned.

### Usage

```
var singleton = ClusterSingleton.Get(system);
var settings = ClusterSingletonSettings.Create(system).WithRole("singleton");

var proxy = singleton.Init(SingletonActor.Create(Props.Create<GlobalCounter>(), "GlobalCounter").WithSettings(settings));
proxy.Tell(Counter.Increment);
```

This will required a complete cluster restart though, since the names for both the `ClusterSingletonManager` and `ClusterSingletonManagerProxy` are now automatically constructed (i.e., based on the code above, the new manager will be named `singletonManagerGlobalCounter` and the proxy `singletonProxyGlobalCounter`).

## Checklist

* [X] Design discussion issue #6036

### DIFF
--- a/docs/articles/clustering/cluster-singleton.md
+++ b/docs/articles/clustering/cluster-singleton.md
@@ -4,6 +4,8 @@ title: Cluster Singleton
 ---
 # Cluster Singleton
 
+## Introduction
+
 For some use cases it is convenient and sometimes also mandatory to ensure that you have exactly one actor of a certain type running somewhere in the cluster.
 
 Some examples:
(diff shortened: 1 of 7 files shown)


Aaronontheweb:
This looks great @ismaelhamed - I'll review this shortly.

ismaelhamed:
The only thing I've found with this implementation is that if the user already have a `ClusterSingletonManager` that is persistent, due to how the manager's name is now automatically [constructed](https://github.com/akkadotnet/akka.net/pull/6050/files#diff-1a8b145d0fbf473a56b9f08d5b8fe561d7304960619f712d0582080f5c79d98fR81), it probably won't match with the existing `PersistentId` in the journal/snapshot-store tables.

ismaelhamed:
Added some test and documentation.

Aaronontheweb:
> The only thing I've found with this implementation is that if the user already have a `ClusterSingletonManager` that is persistent, due to how the manager's name is now automatically [constructed](https://github.com/akkadotnet/akka.net/pull/6050/files#diff-1a8b145d0fbf473a56b9f08d5b8fe561d7304960619f712d0582080f5c79d98fR81), it probably won't match the existing `PersistentId` in the journal/snapshot-store tables.

What would an end-user need to do to work around that? Is there something we'll have to document or help them migrate with?

ismaelhamed:
> > The only thing I've found with this implementation is that if the user already have a `ClusterSingletonManager` that is persistent, due to how the manager's name is now automatically [constructed](https://github.com/akkadotnet/akka.net/pull/6050/files#diff-1a8b145d0fbf473a56b9f08d5b8fe561d7304960619f712d0582080f5c79d98fR81), it probably won't match the existing `PersistentId` in the journal/snapshot-store tables.
> 
> What would an end-user need to do to work around that? Is there something we'll have to document or help them migrate with?

I would probably only recommend using this newer API for brand new applications since, at the very least, a cluster restart is needed. IMO, if you already have your singletons figured out, this will only save you a few lines of code which may not be worth the effort.

docs/articles/clustering/cluster-singleton.md
```diff
@@ -35,6 +43,65 @@ This pattern may seem to be very tempting to use at first, but it has several dr
 
 Especially the last point is something you should be aware of — in general when using the Cluster Singleton pattern you should take care of downing nodes yourself and not rely on the timing based auto-down feature.
 
+## An Example (Simplified API)
+
+Any `Actor` can be run as a singleton. E.g. a basic counter:
+
+```csharp
```
Aaronontheweb, line 50: Generally, we prefer to use DocFx's `[code]` blocks that reference a runnable source file instead of traditional markdown blocks - see https://getakka.net/community/contributing/documentation-guidelines.html#code-samples-must-use-code-references for examples on how to do this.

Reason why we care: if the sample is updated, the documentation is also automatically updated too.

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonApiSpec.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            _system2 = ActorSystem.Create(
+                Sys.Name,
+                ConfigurationFactory.ParseString("akka.cluster.roles = [\"singleton\"]").WithFallback(Sys.Settings.Config));
+
+            _clusterNode2 = Cluster.Get(_system2);
+        }
+
+        [Fact]
+        public void A_cluster_singleton_must_be_accessible_from_two_nodes_in_a_cluster()
```
Aaronontheweb, line 89: LGTM

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingleton.cs
```diff
@@ -0,0 +1,131 @@
...
+                {
+                    _system.ActorOf(ClusterSingletonManager.Props(
+                        singletonProps: singleton.Props,
+                        terminationMessage: singleton.StopMessage.GetOrElse(PoisonPill.Instance),
+                        settings: settings.ToManagerSettings(singleton.Name)),
+                        managerName);
+                }
+                catch (InvalidActorNameException ex) when (ex.Message.EndsWith("is not unique!"))
+                {
+                    // This is fine. We just wanted to make sure it is running and it already is
```
Aaronontheweb, line 59: LGTM

Aaronontheweb (approved):
LGTM - greatly simplifies the API for managing `ClusterSingleton`s. This is a major usability improvement.

docs/articles/clustering/cluster-singleton.md
Aaronontheweb: I'm going to merge this in now and we can update those docs to reference your unit tests' code in a separate PR.

PR #6057, opened 2022-08-03
Tutorial improvements

Nicolaidavies:
## Changes

This PR improves the tutorial by:

- adding an assertion to tutorial 3. Without this assertion, there is no check of the newly implemented functionality.
- removing unused arguments that would otherwise stop the code from compiling.
- including code that is relevant for step 4, that would stop the code from compiling at that point in the tutorial.

I made all the changes while following the tutorial.


Aaronontheweb:
Thanks @Nicolaidavies - looks like there's something DocFx isn't happy about in the build logs. I'll check it out today - it's usually something small. We appreciate your contribution!

Aaronontheweb:
Actually, looks like it wasn't even a DocFx validation issue - NuGet 503'd during a package restore. I'll just re-run those steps.

Nicolaidavies:
Thank you @Aaronontheweb !

PR #6058, opened 2022-08-08
Fixes log and uses cluster sharding `LeaseSettings`, if configured, i…

ismaelhamed:


Aaronontheweb (approved):
LGTM - do we need to backport this to Akka.NET v1.4 @ismaelhamed ?

ismaelhamed:
@Aaronontheweb no, I don't think so.

PR #6063, opened 2022-08-11
[Cluster] Enable HeartbeatResponse message type

### LINKED ISSUE #6061: Akka.Cluster: need to enable `HeartbeatResponse` message type

First introduced in Akka.NET v1.4.19 via https://github.com/akkadotnet/akka.net/pull/4934, we need to enable the new `HeartbeatRsp` message to be sent by default. All clients going back to v1.4.19 and earlier have the ability to deserialize this message, but we never introduced the response payload since it could result in a break in wire compatibility.

Thus, we followed our best practice and allowed the read-side code to "age into" existing installations over a long-ish period of time. Now with Akka.NET v1.5 come up it's the right time to enable this particular feature as it will ultimately assist with debugging heartbeat issues inside Akka.Cluster.

Need to update the `ClusterMessageSerializer` to produce the typed protobuf `HeartbeatResponse` message instead of the old `UniqueAddress` response message we currently write here:

https://github.com/akkadotnet/akka.net/blob/03de287884074de1d3a13ef93f77df4de1df99bb/src/core/Akka.Cluster/Serialization/ClusterMessageSerializer.cs#L72

You can see what the typed `HeartbeatResponse` message looks like here:

https://github.com/akkadotnet/akka.net/blob/03de287884074de1d3a13ef93f77df4de1df99bb/src/protobuf/ClusterMessages.proto#L73-L83

Leaving this as-is and making it opt-in, but we decided against that in Akka.NET v1.4.19. It should be safe for most clusters to just upgrade straight v1.5.0.

Arkatufus:
Fixes #6061

## Changes
- Enable `HeartbeatResponse` and `Heartbeat` protobuf wire format for cluster hearbeats.
- Add `akka.cluster.use-legacy-heartbeat-message` HOCON settings, defaults to false. When set to true, the serializer will use the old legacy `Heartbeat` messages.
- Add `HeartbeatResponse` and `Heartbeat` proto serializer

### DIFF
--- a/src/core/Akka.Cluster.Tests/ClusterSpecBase.cs
+++ b/src/core/Akka.Cluster.Tests/ClusterSpecBase.cs
@@ -7,6 +7,7 @@
 
 using Akka.Configuration;
 using Akka.TestKit;
+using Xunit.Abstractions;
 
 namespace Akka.Cluster.Tests
 {
@@ -15,20 +16,23 @@ namespace Akka.Cluster.Tests
     /// </summary>
     public abstract class ClusterSpecBase : AkkaSpec
     {
-        protected ClusterSpecBase(Config config) : base(config.WithFallback(BaseConfig))
+        protected ClusterSpecBase(Config config, ITestOutputHelper output, bool useLegacyHeartbeat) 
+            : base(config.WithFallback(BaseConfig(useLegacyHeartbeat)), output)
         {
             
         }
 
-        protected ClusterSpecBase()
-            : base(BaseConfig)
+        protected ClusterSpecBase(ITestOutputHelper output, bool useLegacyHeartbeat)
+            : base(BaseConfig(useLegacyHeartbeat), output)
         {
 
         }
 
-        protected static readonly Config BaseConfig = ConfigurationFactory.ParseString(@"
-                            akka.actor.serialize-messages = on
-                            akka.actor.serialize-creators = on");
+        private static Config BaseConfig(bool useLegacyHeartbeat) => 
+            ConfigurationFactory.ParseString($@"
+                akka.actor.serialize-messages = on
+                akka.actor.serialize-creators = on
+                akka.cluster.use-legacy-heartbeat-message = {(useLegacyHeartbeat ? "true" : "false")}");
     }
 }
 
--- a/src/core/Akka.Cluster.Tests/Serialization/ClusterMessageSerializerSpec.cs
+++ b/src/core/Akka.Cluster.Tests/Serialization/ClusterMessageSerializerSpec.cs
@@ -15,6 +15,7 @@ using Akka.TestKit;
 using Xunit;
 using FluentAssertions;
 using Akka.Util;
+using Akka.Util.Internal;
 using Google.Protobuf;
 using Xunit.Abstractions;
 
@@ -209,10 +210,12 @@ akka.cluster.use-legacy-heartbeat-message = {(useLegacyHeartbeat ? "true" : "fal
 
         private T AssertAndReturn<T>(T message)
         {
-            var serializer = Sys.Serialization.FindSerializerFor(message);
-            var serialized = serializer.ToBinary(message);
+            var serializer = (SerializerWithStringManifest) Sys.Serialization.FindSerializerFor(message);
             serializer.Should().BeOfType<ClusterMessageSerializer>();
-            return serializer.FromBinary<T>(serialized);
+            
+            var serialized = serializer.ToBinary(message);
+            var manifest = serializer.Manifest(message);
+            return (T) serializer.FromBinary(serialized, manifest);
         }
 
         private void AssertEqual<T>(T message)
--- a/src/core/Akka.Cluster.Tests/Serialization/ClusterMessageSerializerSpec.cs
+++ b/src/core/Akka.Cluster.Tests/Serialization/ClusterMessageSerializerSpec.cs
@@ -37,11 +37,13 @@ namespace Akka.Cluster.Tests.Serialization
     
     public abstract class ClusterMessageSerializerBase : AkkaSpec
     {
+        private readonly bool _useLegacyHeartbeat;
         public ClusterMessageSerializerBase(ITestOutputHelper output, bool useLegacyHeartbeat)
             : base($@"
 akka.actor.provider = cluster
 akka.cluster.use-legacy-heartbeat-message = {(useLegacyHeartbeat ? "true" : "false")}", output)
         {
+            _useLegacyHeartbeat = useLegacyHeartbeat;
         }
 
         private static readonly Member a1 = TestMember.Create(new Address("akka.tcp", "sys", "a", 2552), MemberStatus.Joining, appVersion: AppVersion.Create("1.0.0"));
@@ -54,8 +56,11 @@ akka.cluster.use-legacy-heartbeat-message = {(useLegacyHeartbeat ? "true" : "fal
         public void Can_serialize_Heartbeat()
         {
             var address = new Address("akka.tcp", "system", "some.host.org", 4711);
-            var message = new ClusterHeartbeatSender.Heartbeat(address, -1, -1);
-            AssertEqual(message);
+            var legacyMessage = new ClusterHeartbeatSender.Heartbeat(address, -1, -1);
+            var message =  new ClusterHeartbeatSender.Heartbeat(address, 10, 3);
+            
+            // Legacy heartbeat serializer will replace the sequence number and creation date with -1 and -1 respectively
+            AssertEqual(message, _useLegacyHeartbeat ? legacyMessage : message);
         }
 
         [Fact]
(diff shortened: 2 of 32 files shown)


src/core/Akka.Cluster.Tests/ClusterSpecBase.cs
```diff
@@ -15,20 +16,23 @@ namespace Akka.Cluster.Tests
...
         }
 
-        protected static readonly Config BaseConfig = ConfigurationFactory.ParseString(@"
-                            akka.actor.serialize-messages = on
-                            akka.actor.serialize-creators = on");
+        private static Config BaseConfig(bool useLegacyHeartbeat) => 
+            ConfigurationFactory.ParseString($@"
+                akka.actor.serialize-messages = on
+                akka.actor.serialize-creators = on
+                akka.cluster.use-legacy-heartbeat-message = {(useLegacyHeartbeat ? "true" : "false")}");
```
Aaronontheweb, line 35: I take it this is for cases where a user needs to upgrade a much older version of Akka.NET v1.3 / v1.4 to v1.5.

src/core/Akka.Cluster/ClusterSettings.cs
```diff
@@ -118,6 +118,7 @@ TimeSpan GetWeaklyUpDuration()
 
             WeaklyUpAfter = GetWeaklyUpDuration();
 
+            UseLegacyHeartbeatMessage = clusterConfig.GetBoolean("use-legacy-heartbeat-message", true);
```
Aaronontheweb, line 121: needs to default to `false` - we want users to have to use the new message starting with v1.5. It should be opt-out, not opt-in.

src/core/Akka.Cluster/Configuration/Cluster.conf
```diff
@@ -169,6 +169,8 @@ akka {
     # greater than this value.
     reduce-gossip-different-view-probability = 400
 
+    use-legacy-heartbeat-message = true
```
Aaronontheweb, line 172: Needs to default to `false`.

Aaronontheweb (changes requested):
Need to change our defaults, but otherwise LGTM

src/core/Akka.Cluster/ClusterSettings.cs
Arkatufus: Got it

src/core/Akka.Cluster/Configuration/Cluster.conf
Arkatufus: Got it

src/core/Akka.Cluster.Tests/ClusterSpecBase.cs
Arkatufus: Basically I've abstracted the unit test so that we can test that both settings would work with all of our current unit tests that touches the heartbeat messages

Arkatufus:
Done, changed the setting to opt-out, not opt-in

Aaronontheweb (approved):

src/core/Akka/Serialization/Serializer.cs
```diff
@@ -137,7 +137,7 @@ protected SerializerWithStringManifest(ExtendedActorSystem system) : base(system
         /// <returns>The object contained in the array</returns>
         public sealed override object FromBinary(byte[] bytes, Type type)
         {
-            var manifest = type.TypeQualifiedName();
```
Arkatufus, line 140: There's a bug here, `SerializerWithStringManifest` uses different manifest in this code path

src/core/Akka/Serialization/Serializer.cs
```diff
@@ -168,6 +168,15 @@ public sealed override object FromBinary(byte[] bytes, Type type)
         /// <param name="o">The object for which the manifest is needed.</param>
         /// <returns>The manifest needed for the deserialization of the specified <paramref name="o"/>.</returns>
         public abstract string Manifest(object o);
+
+        /// <summary>
+        /// Returns the manifest (type hint) that will be provided in the <see cref="FromBinary(byte[],System.Type)"/> method.
+        /// </summary>
+        /// <param name="type">The <see cref="Type"/> for which the manifest is needed.</param>
+        /// <returns>The manifest needed for the deserialization of the specified <paramref name="type"/>.</returns>
+        public virtual string Manifest(Type type)
```
Arkatufus, line 177: This might not be an ideal solution, would love to hear any input on this one

src/core/Akka.Cluster/Serialization/ClusterMessageSerializer.cs
```diff
@@ -141,38 +155,37 @@ public override object FromBinary(byte[] bytes, string manifest)
...
-                case GossipEnvelope _:
-                    return GossipEnvelopeManifest;
-                case ClusterRouterPool _:
-                    return ClusterRouterPoolManifest;
-                default:
-                    throw new ArgumentException($"Can't serialize object of type [{o.GetType()}] in [{GetType()}]");
-            }
+            => Manifest(o.GetType());
+        
+        public override string Manifest(Type type)
```
Aaronontheweb, line 160: I'd prefer it if we didn't add this

src/core/Akka/Serialization/Serializer.cs
Aaronontheweb: This method should not be called on `SerializerWithStringManifest` unless an end-user directly casts it into a `Serializer`, so I don't think this needs to be changed for this test.

src/core/Akka/Serialization/Serializer.cs
Aaronontheweb: I'd prefer if we didn't do this

Aaronontheweb (changes requested):
I think we can skip the serializer changes and just modify 

```csharp
 private T AssertAndReturn<T>(T message)
        {
            var serializer = Sys.Serialization.FindSerializerFor(message);
            var serialized = serializer.ToBinary(message);
            serializer.Should().BeOfType<ClusterMessageSerializer>();
            return serializer.FromBinary<T>(serialized);
        }
```

to cast the returned serializer into `SerializerWithStringManifest` and then manually call `.Manifest` on that. If the serialization binding is wrong then the cast will fail.

src/core/Akka.Cluster/Serialization/ClusterMessageSerializer.cs
Arkatufus: Done, reverted, changes moved to the spec instead

src/core/Akka.Cluster/Configuration/Cluster.conf
```diff
@@ -169,6 +169,8 @@ akka {
     # greater than this value.
     reduce-gossip-different-view-probability = 400
 
+    use-legacy-heartbeat-message = false
```
Aaronontheweb, line 172: Probably should add a comment here, but we can do that in a future PR

src/core/Akka.Cluster/ClusterSettings.cs
```diff
@@ -300,6 +301,8 @@ TimeSpan GetWeaklyUpDuration()
         /// The leader will move <see cref="MemberStatus.WeaklyUp"/> members to <see cref="MemberStatus.Up"/> status once convergence has been reached.
         /// </summary>
         public TimeSpan WeaklyUpAfter { get; }
+        
+        public bool UseLegacyHeartbeatMessage { get; }
```
Aaronontheweb, line 305: Should also add an XML-DOC comment here. Can do that in a future PR.

src/core/Akka.Cluster.Tests/Serialization/ClusterMessageSerializerSpec.cs
```diff
@@ -63,20 +67,6 @@ public void Can_serialize_HeartbeatRsp()
             AssertEqual(message);
```
Aaronontheweb, line 67: This test only succeeds because `-1,`, -`1` are the hard-coded values when legacy==on. Need to harden this to make sure that the correct, non-`-1` values are supported when `legacy==off`.

Aaronontheweb: The `AssertEqual` regardless of config is what's not going to work here. Change 

```csharp
 var message = new ClusterHeartbeatSender.HeartbeatRsp(uniqueAddress, -1, -1);
```

to 

```csharp
 var message = new ClusterHeartbeatSender.HeartbeatRsp(uniqueAddress, 10, 3);
```

When legacy off, 10,3

When legacy on, -1, -1

Aaronontheweb (changes requested):
Need to harden tests

src/core/Akka.Cluster/Configuration/Cluster.conf
Arkatufus: Done

src/core/Akka.Cluster/ClusterSettings.cs
Arkatufus: Done

src/core/Akka.Cluster.Tests/Serialization/ClusterMessageSerializerSpec.cs
Arkatufus: Done

Aaronontheweb (approved):
LGTM - nice work @Arkatufus

Aaronontheweb:
TODO: need to add an issue for this letting users know about this setting as part of the Akka.NET 1.5 upgrade advisories

PR #6065, opened 2022-08-15
Cluster singleton should consider Member AppVersion during hand over.

### LINKED ISSUE #6035: Akka.Cluster.Tools: `ClusterSingleton` should factor in `AppVersion` when choosing next likely home for messages during deployments

Akka.Cluster.Sharding does this currently - shards are homed onto nodes with the highest `AppVersion` when a deployment is detected inside the cluster. This helps us guarantee that we aren't allocating shards, again, on locations that are going to be removed later during a deployment.

We _do not_ do this for Akka.Cluster singletons, which have the same issue as sharded entity actors: they prefer to be moved _only once_ if possible.

Consider `AppVersion` in the re-homing calculation for cluster singletons.

Arkatufus:
Fixes #6035

## Changes

Changes the way oldest member sorting comparator works, taking Member.AppVersion into consideration. Members with higher AppVersion will always be bumped to the front of the list, regardless of Member age.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestart3Spec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestart3Spec.cs
@@ -31,6 +31,7 @@ namespace Akka.Cluster.Tools.Tests.Singleton
               akka.cluster.app-version = ""1.0.0""
               akka.cluster.auto-down-unreachable-after = 2s
               akka.cluster.singleton.min-number-of-hand-over-retries = 5
+              akka.cluster.singleton.consider-app-version = true
               akka.remote {
                 dot-netty.tcp {
                   hostname = ""127.0.0.1""
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestart3Spec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestart3Spec.cs
@@ -1,0 +1,134 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ClusterSingletonRestart2Spec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2021 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2021 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Cluster.Tools.Singleton;
+using Akka.Configuration;
+using Akka.TestKit;
+using FluentAssertions;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Cluster.Tools.Tests.Singleton
+{
+    public class ClusterSingletonRestart3Spec : AkkaSpec
+    {
+        private readonly ActorSystem _sys1;
+        private readonly ActorSystem _sys2;
+        private readonly ActorSystem _sys3;
+
+        public ClusterSingletonRestart3Spec(ITestOutputHelper output) : base(@"
+              akka.loglevel = DEBUG
+              akka.actor.provider = ""cluster""
+              akka.cluster.app-version = ""1.0.0""
+              akka.cluster.auto-down-unreachable-after = 2s
+              akka.cluster.singleton.min-number-of-hand-over-retries = 5
+              akka.remote {
+                dot-netty.tcp {
+                  hostname = ""127.0.0.1""
+                  port = 0
+                }
+              }", output)
+        {
+            _sys1 = Sys;
+            _sys2 = ActorSystem.Create(Sys.Name, Sys.Settings.Config);
+            InitializeLogger(_sys2);
+            _sys3 = ActorSystem.Create(Sys.Name, ConfigurationFactory.ParseString("akka.cluster.app-version = \"1.0.2\"")
+                .WithFallback(Sys.Settings.Config));
+            InitializeLogger(_sys3);
+        }
+
+        public void Join(ActorSystem from, ActorSystem to)
+        {
+            from.ActorOf(ClusterSingletonManager.Props(Props.Create(() => new Singleton()),
+                PoisonPill.Instance,
+                ClusterSingletonManagerSettings.Create(from)), "echo");
+
+
+            Within(TimeSpan.FromSeconds(45), () =>
+            {
+                AwaitAssert(() =>
+                {
+                    Cluster.Get(from).Join(Cluster.Get(to).SelfAddress);
+                    Cluster.Get(from).State.Members.Select(x => x.UniqueAddress).Should().Contain(Cluster.Get(from).SelfUniqueAddress);
+                    Cluster.Get(from)
+                        .State.Members.Select(x => x.Status)
+                        .ToImmutableHashSet()
+                        .Should()
+                        .Equal(ImmutableHashSet<MemberStatus>.Empty.Add(MemberStatus.Up));
+                });
+            });
+        }
+
+        [Fact]
+        public void Singleton_should_consider_AppVersion_when_handing_over()
+        {
+            Join(_sys1, _sys1);
+            Join(_sys2, _sys1);
+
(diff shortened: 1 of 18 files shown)


src/core/Akka/Util/AppVersion.cs
```diff
@@ -243,12 +243,12 @@ public int CompareTo(AppVersion other)
 
         public bool Equals(AppVersion other)
         {
-            return other != null && Version == other.Version;
+            return other != null && string.Equals(Version, other.Version, StringComparison.Ordinal);
         }
 
         public override bool Equals(object obj)
         {
-            return base.Equals(obj as AppVersion);
```
Arkatufus, line 251: Since AppVersion extends Object directly, base.Equals() does a direct reference equality, which is wrong.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/MemberAgeOrdering.cs
```diff
@@ -24,6 +24,20 @@ private MemberAgeOrdering(bool ascending)
...
+            if (y is null)
+                return _ascending ? 1 : -1;
+
+            if (x is null)
+                return _ascending ? -1 : 1;
+
+            // prefer nodes with the highest app version, even if they're younger
+            var appVersionDiff = x.AppVersion.CompareTo(y.AppVersion);
+            if (appVersionDiff != 0)
+                return _ascending ? appVersionDiff : appVersionDiff * -1;
```
Arkatufus, line 39: AppVersion difference always wins compared to Member age difference

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestart3Spec.cs
```diff
@@ -0,0 +1,135 @@
...
+                dot-netty.tcp {
+                  hostname = ""127.0.0.1""
+                  port = 0
+                }
+              }", output)
+        {
+            _sys1 = Sys;
+            _sys2 = ActorSystem.Create(Sys.Name, Sys.Settings.Config);
+            InitializeLogger(_sys2);
+            _sys3 = ActorSystem.Create(Sys.Name, ConfigurationFactory.ParseString("akka.cluster.app-version = \"1.0.2\"")
```
Aaronontheweb, line 45: LGTM

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestart3Spec.cs
```diff
@@ -0,0 +1,135 @@
...
+            {
+                AwaitAssert(() =>
+                {
+                    var probe = CreateTestProbe(_sys2);
+                    proxy2.Tell("poke", probe.Ref);
+
+                    // note that _sys3 has a higher app-version, so the singleton should start there
+                    var singleton = probe.ExpectMsg<Member>(TimeSpan.FromSeconds(1));
+                    singleton.Should().Be(Cluster.Get(_sys3).SelfMember);
+                    singleton.AppVersion.Version.Should().Be("1.0.2");
```
Aaronontheweb, line 112: LGTM

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -880,7 +880,9 @@ private void InitializeFSM()
                 {
                     case StartOldestChangedBuffer _:
                         {
-                            _oldestChangedBuffer = Context.ActorOf(Actor.Props.Create<OldestChangedBuffer>(_settings.Role).WithDispatcher(Context.Props.Dispatcher));
+                            _oldestChangedBuffer = Context.ActorOf(
+                                Actor.Props.Create(() => new OldestChangedBuffer(_settings.Role, _settings.ConsiderAppVersion))
```
Aaronontheweb, line 884: LGTM

src/contrib/cluster/Akka.Cluster.Tools/Singleton/MemberAgeOrdering.cs
```diff
@@ -15,15 +15,25 @@ namespace Akka.Cluster.Tools.Singleton
...
         }
 
         /// <inheritdoc/>
         public int Compare(Member x, Member y)
         {
+            if (_considerAppVersion)
+            {
+                // prefer nodes with the highest app version, even if they're younger
+                var appVersionDiff = x.AppVersion.CompareTo(y.AppVersion);
+                if (appVersionDiff != 0)
```
Aaronontheweb, line 33: So we have to run the sort in two phases here:

1. AppVersion
2. Age within the AppVersion cohort.

Looks to me like the sorter disregards age when `AppVersion` is considered - is that correct?

Aaronontheweb (commented):
Important design question

src/contrib/cluster/Akka.Cluster.Tools/Singleton/MemberAgeOrdering.cs
Aaronontheweb: This design is wrong then - without the member age requirement there's no guarantee of stable singleton placement, which defeats the purpose of adding this feature (i.e. being able to predict the next location of where a singleton will be placed during a deployment)

We need to do the sort in multiple phases:

1. Members by role
2. Members within a role that have the highest app version
3. Oldest member within the results of phase 2.

Let's design it to work that way.

Aaronontheweb (changes requested):
Need to fix member ordering guarantee

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/MemberAgeOrderingSpec.cs
```diff
@@ -0,0 +1,89 @@
...
+            };
+
+            var seq = members.ToList();
+            seq.Count.Should().Be(3);
+            seq[0].Should().Be(Create(Address.Parse("akka://sys@darkstar:1113"), upNumber: 1));
+            seq[1].Should().Be(Create(Address.Parse("akka://sys@darkstar:1112"), upNumber: 3));
+            seq[2].Should().Be(Create(Address.Parse("akka://sys@darkstar:1111"), upNumber: 9));
+        }
+        
+        [Fact(DisplayName = "MemberAgeOrdering should sort based on Address if UpNumber is the same")]
```
Aaronontheweb, line 38: Not possible for UpNumber to be the same - that'd be the result of data corruption.

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/MemberAgeOrderingSpec.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+        
+        [Fact(DisplayName = "MemberAgeOrdering should prefer AppVersion over UpNumber")]
+        public void SortByAppVersionTest()
+        {
+            var members = new SortedSet<Member>(MemberAgeOrdering.DescendingWithAppVersion)
+            {
+                Create(Address.Parse("akka://sys@darkstar:1112"), upNumber: 3, appVersion: AppVersion.Create("1.0.0")),
+                Create(Address.Parse("akka://sys@darkstar:1113"), upNumber: 1, appVersion: AppVersion.Create("1.0.0")),
+                Create(Address.Parse("akka://sys@darkstar:1116"), upNumber: 6, appVersion: AppVersion.Create("1.0.2")),
```
Aaronontheweb, line 62: Could you humor me and add this item last on this list? That'd prove that the upnumber is canon, not insertion order.

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/MemberAgeOrderingSpec.cs
Arkatufus: Its part of the `MemberAgeOrdering` code, it defaults to member `Address` ordering if the `UpNumber` is the same.
Since this is a spec for the comparator, I might as well make sure that it works.

Aaronontheweb: ok, that's fine.

Aaronontheweb (approved):
LGTM

Aaronontheweb:
TODO: need to add changelog and website documentation for this feature

Aaronontheweb:
@Arkatufus API approvals are out of date

PR #6085, opened 2022-08-30
Akka.TestKit.Hosting Proof of Concept

Arkatufus:
A proof of concept implementation of Akka.Hosting test kit.
Comments and reviews are welcome.

## Configuring the spec

The biggest difference is that, since the test is started asynchronously, the TestKit and ActorSystem would not be available in the unit test class constructor anymore. Since the spec depends on Microsoft HostBuilder, configuration has to be broken down into steps. There are overridable methods that user can use to override virtually all of the setup process. 

The steps of what overridable methods gets called is:
1. `ConfigureLogging(ILoggingBuilder)` -- Add custom logger and filtering rules here.
2. `ConfigureHostConfiguration(IConfigurationBuilder)` -- Inject any additional hosting environment configuration here, such as faking environment variables.
3. `ConfigureAppConfiguration(HostBuilderContext, IConfigurationBuilder)` -- Inject the application configuration here.
4. `ConfigureServices(HostBuilderContext, IServiceCollection)` -- Add additional services needed by the test here, such as mocked up services used inside the dependency injection.
5. `ConfigureAkka(AkkaConfigurationBuilder, IServiceProvider)` -- This is called inside `AddAkka`, use this to configure the `ActorSystem`

All of the original `Akka.TestKit` methods will be wrapped by the `HostingSpec` class (only partially for this POC), All of the properties, such as `Sys` and `TestActor` will be available once the unit test class is invoked.


Arkatufus:
As per request, moving this to `Akka.Hosting` repo.

PR #6092, opened 2022-09-08
[Akka.TestKit] TestKitBase should take arbitrary ActorSystem without throwing

Arkatufus:
## Fixes 
Problem when passing arbitrary `ActorSystem` into a TestKit, it would throw a `ConfigurationException` because the TestKit HOCON configuration was not injected into the `ActorSystem`

## Changes
- Injects `TestKitBase._defaultConfig` if the TestKit did not create the `ActorSystem` being tested
- Add unit tests


Aaronontheweb:
@Arkatufus does this also need to be ported to v1.4?

Arkatufus:
It will be, yes

PR #6093, opened 2022-09-09
Change TestKitBase.InitializeTest to virtual so it can be deferred

Arkatufus:
## Changes
- Change TestKitBase.InitializeTest to virtual so it can be deferred
  This is needed for Akka.Hosting.TestKit


Aaronontheweb (approved):
LGTM - need to backport to 1.5 also.

PR #6099, opened 2022-09-14
Remove the PatternMatch class

Arkatufus:
Remove the PatternMatch class, updating all the codes that uses it with .NET language version 7.0 switch...case pattern matching


Aaronontheweb:
Thanks @Arkatufus - I'll give this a detailed review, but the test results look good.

Aaronontheweb (approved):
LGTM - nice work @Arkatufus - this will probably help improve performance somewhat in some key areas of Akka.Persistence.Query, DData, and other areas that were still using `PatternMatch` heavily.

PR #6101, opened 2022-09-16
Akka.Cluster.Sharding `GetEntityLocation` Query

Aaronontheweb:
## Changes

Designed to make it easier for testing and telemetry, the new `GetEntityLocation` is an `IShardRegionQuery` that checks for the presence of an entity in a given `ShardRegion` and reports back if this entity is able to be located.

This query is not designed to be supported over the network and is meant to be local-only.

I've added this primary for _testing and telemetry_ purposes - as we're going to look at using this inside the [`pbm cluster-sharding` palette for Petabridge.Cmd](https://cmd.petabridge.com/articles/commands/cluster-sharding-commands.html).

## Checklist

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
@@ -936,56 +936,71 @@ namespace Akka.Cluster.Sharding
             {
                 return shardOrRegionRef.Path.Address.HasGlobalScope ? shardOrRegionRef.Path.Address : Cluster.SelfAddress;
             }
-            
-            var shardId = this.ExtractShardId(getEntityLocation.EntityId);
-            if (string.IsNullOrEmpty(shardId))
-            {
-                // unsupported entityId - could only happen in highly customized extractors
-                sender.Tell(new EntityLocation(getEntityLocation.EntityId, shardId, Address.AllSystems, Option<IActorRef>.None));
-                return;
-            }
-            
-            if(!Shards.TryGetValue(shardId, out var shardActorRef))
-            {
-                // shard is not homed yet, so try looking up the ShardRegion
-                if (!RegionByShard.TryGetValue(shardId, out var shardRegionRef))
-                {
-                    // shardRegion isn't allocated either
-                    sender.Tell(new EntityLocation(getEntityLocation.EntityId, shardId, Address.AllSystems, Option<IActorRef>.None));
-                }
-                else
-                {
-                    // ShardRegion exists, but shard is not homed
-                    sender.Tell(new EntityLocation(getEntityLocation.EntityId, shardId, GetNodeAddress(shardRegionRef), Option<IActorRef>.None));
-                }
-                
-                return;
-            }
-            
-            var destinationAddress = GetNodeAddress(shardActorRef);
 
-            async Task ResolveEntityRef()
+            try
             {
-                // we have a ShardRef - now we just need to check to see if an entity ref exists
-                // we are going to
-
-                var entityPath = shardActorRef.Path / shardId;
-                
-                try
+                var shardId = ExtractShardId(new StartEntity(getEntityLocation.EntityId));
+                if (string.IsNullOrEmpty(shardId))
                 {
-                    var entityRef = await Context.ActorSelection(entityPath).ResolveOne(getEntityLocation.Timeout);
-                    sender.Tell(new EntityLocation(getEntityLocation.EntityId, shardId, destinationAddress, new Option<IActorRef>(entityRef)));
+                    // unsupported entityId - could only happen in highly customized extractors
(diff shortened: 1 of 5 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -917,12 +983,90 @@ private void HandleShardRegionQuery(IShardRegionQuery query)
...
+                case GetEntityLocation g:
+                    ReplyToGetEntityLocationQuery(g, Sender);
+                    break;
                 default:
                     Unhandled(query);
                     break;
             }
         }
 
+        private void ReplyToGetEntityLocationQuery(GetEntityLocation getEntityLocation, IActorRef sender)
```
Aaronontheweb, line 995: This is where the meat of these changes are

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -917,12 +983,90 @@ private void HandleShardRegionQuery(IShardRegionQuery query)
...
+            Address GetNodeAddress(IActorRef shardOrRegionRef)
+            {
+                return shardOrRegionRef.Path.Address.HasGlobalScope
+                    ? shardOrRegionRef.Path.Address
+                    : Cluster.SelfAddress;
+            }
+
+            try
+            {
+                var shardId = ExtractShardId(new StartEntity(getEntityLocation.EntityId));
```
Aaronontheweb, line 1008: One thing that is awkward about this query is that the `ShardExtractor` has to support the `ShardRegion.StartEntity` type in order for this to work - the reason being is that even though we already know what the entity id is, the extractors that resolve the `ShardId` are looking for an _input message_ to generate the `ShardId`, not the `EntityId`. Therefore I have to smuggle the `EnttityId` back through a `ShardRegion.StartEntity` message. That makes this a bit hacky but it's the best compromise I could make.

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -917,12 +983,90 @@ private void HandleShardRegionQuery(IShardRegionQuery query)
...
+                    if (!RegionByShard.TryGetValue(shardId, out var shardRegionRef))
+                    {
+                        // shardRegion isn't allocated either
+                        sender.Tell(new EntityLocation(getEntityLocation.EntityId, shardId, Address.AllSystems,
+                            Option<IActorRef>.None));
+                    }
+                    else
+                    {
+                        // ShardRegion exists, but shard is not homed
+                        // NOTE: in the event that we're querying a shard's location from a ShardRegionProxy
```
Aaronontheweb, line 1046: Validated this during our unit tests.

src/contrib/cluster/Akka.Cluster.Sharding/ShardingMessages.cs
```diff
@@ -153,6 +158,77 @@ private GetCurrentRegions()
...
+    ///
+    /// Creates a <see cref="EntityLocation"/> message in response.
+    /// </summary>
+    /// <remarks>
+    /// This is used primarily for testing and telemetry purposes.
+    ///
+    /// In order for this query to work, the <see cref="MessageExtractor"/> must support <see cref="ShardRegion.StartEntity"/>,
+    /// which is also used when remember-entities=on.
+    /// </remarks>
+    public sealed class GetEntityLocation : IShardRegionQuery
```
Aaronontheweb, line 173: This message never gets sent over the wire, so it doesn't require any changes to the wire type nor does it need to support `IClusterShardingSerializable`.

Aaronontheweb (commented):
Reviewed my own PR to make things easier.

Arkatufus (approved):
LGTM, yeah, the `StartEntity` requirement is a bit of a bummer. But I guess this will make the solution work "out-of-the-box"-ish if the remember entities is turned on?

Aaronontheweb:
> LGTM, yeah, the `StartEntity` requirement is a bit of a bummer. But I guess this will make the solution work "out-of-the-box"-ish if the remember entities is turned on?

That's correct - I went for the "least bad" solution here. I don't expect many users to actually use this query much in production, but the one problematic issue here is what we're going to do inside Petabridge.Cmd....

Aaronontheweb:
Going to do what we discussed in discord and see if I can eliminate the `ShardRegion.StartEntity` requirement on this pull request - maybe I'll use that as a fallback instead.

Aaronontheweb:
Sigh - looks like there's no easy way to do this without `ShardRegion.StartEntity`. Even built-in extractors like the `HashCodeMessageExtractor` are all designed to extract `ShardId` from a message - there's nothing exposed to compute the `ShardId` directly from the `EntityId` and there's not a lot of need for that inside the `ShardRegion`. Going to add some documentation for this query and then merge it in.

Issue #6106, opened 2022-09-21
Exceeding max-concurrent-recoveries triggers circuit breaker


lucavice:
**Version Information**
Version of Akka.NET? 1.4.40
Which Akka.NET Modules? Akka.Cluster.Sharding 1.4.40, Akka.Persistence.SqlServer 1.4.35

**Describe the bug**
In certain situations, exceedeing temporarily the max-concurrenct-recoveries parameter triggers a circuit breaker that prevents Akka Persistence to persist any further events for the duration of the circuit breaker.

See sequence of logged events here:
![image](https://user-images.githubusercontent.com/14258744/191468001-584c557c-1dd5-40fc-b38d-0a800de711e2.png)

I have been unable to reproduce reliably this problem, as it seems to happen fairly randomly on our production instance (a few times per day).
Setting locally a max-concurrenct-recoveries equal to 1 and force recover of multiple actor at once does not seem to create the issue, so it must be triggered by a combination of factors.

We can't find the root of the error that triggers the circuit breaker. There is no information in the logged OpenCircuitException, and that's the only error that appears in the log (hundreds of times for the duration of the open circuit breaker).

**To Reproduce**
I don't have reliable steps to trigger the problem.
I would appreciate hints on what I could try to understand better the underlying problem and come up with a strategy to reproduce reliably. It may be possible that this is entirely caused by some bad programming on my side, but I'm a bit lost in what to look for.

**Environment**
Windows on .NET 6

Aaronontheweb:
Looks like @ismaelhamed has a fix for this targeting the v1.5 branch - we'll see if we can also backport that to a future v1.4 release as well. Going to review Ismael's work along with @Arkatufus right now.

lucavice:
@ismaelhamed @Aaronontheweb I was able to retrieve some errors today which include a more detailed stack trace.

I'm posting it here, in case it helps identifying if it's actually related to https://github.com/akkadotnet/akka.net/pull/6109

```
Akka.Pattern.OpenCircuitException: Circuit Breaker is open; calls are failing fast
 ---> System.ArgumentException: The tasks argument contains no tasks. (Parameter 'tasks')
   at System.Threading.Tasks.TaskFactory.CheckMultiContinuationTasksAndCopy(Task[] tasks)
[...]

PR #6108, opened 2022-09-23
Ported CircuitBreakerStressSpec

ismaelhamed:
Might be related to #6106

This PR is to confirm that the `CircuitBreaker`, in its current implementation, could give false positives (signal tasks as done when in reality they failed). 

I ported the `CircuitBreakerStressSpec` and slightly modified it to make all tasks take longer than the CB's `CallTimeout`, so that all of them fail. Instead, by running the test you can see that they are all "marked" as succeeded (`DoneCount`).

Upon further inspection:

- The tasks indeed all failed with a `TimeException`, but the exception is swallowed by the `CallFail` method. This makes impossible to capture the `TimeException` outside the CB --as demonstrated by the `StressActor`.

- Because we are awaiting the task first, and only after it finishes we check whether it took longer than the `CallTimeout`, we could potentially be awaiting indefinitely for the task to complete.

I'll send next another PR with an alternative implementation leveraging the new `Task.WaitAsync` in .NET6

**TEST OUTPUT**
```
FailCount:0, DoneCount:1000, CircCount:0, TimeoutCount:0
FailCount:0, DoneCount:1000, CircCount:0, TimeoutCount:0
FailCount:0, DoneCount:1000, CircCount:0, TimeoutCount:0
```

### DIFF
--- a/src/core/Akka.Tests/Pattern/CircuitBreakerStressSpec.cs
+++ b/src/core/Akka.Tests/Pattern/CircuitBreakerStressSpec.cs
@@ -1,0 +1,134 @@
+//-----------------------------------------------------------------------
+// <copyright file="CircuitBreakerStressSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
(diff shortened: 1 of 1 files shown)


ismaelhamed:
Superseded by #6109

PR #6109, opened 2022-09-23
Alternative implementation of `AtomicState` leveraging WaitAsync

ismaelhamed:
Might be related to #6106

The circuit breaker, in its current implementation, could give false positives (signal tasks as done when in reality they failed).

I ported the `CircuitBreakerStressSpec` and slightly modified it to make all tasks take longer than the CB's `CallTimeout`, so that all of them fail. Instead, by running the test you can see that they are all "marked" as succeeded (`DoneCount`).

Upon further inspection:

- The tasks indeed all failed with a `TimeException`, but the exception is swallowed by the `CallFail` method. This makes impossible to capture the `TimeException` outside the CB --as demonstrated by the `StressActor`.

- Because we are awaiting the task first, and only after it finishes we check whether it took longer than the `CallTimeout`, we could potentially be awaiting indefinitely for the task to complete. This PR leverages the new `Task.WaitAsync` in .NET6 instead.

When compared with the results of the test in the [previous](https://github.com/akkadotnet/akka.net/pull/6108) PR, we now get the correct behavior:

**BEFORE**
```
FailCount:0, DoneCount:1000, CircCount:0, TimeoutCount:0
FailCount:0, DoneCount:1000, CircCount:0, TimeoutCount:0
FailCount:0, DoneCount:1000, CircCount:0, TimeoutCount:0
```

**AFTER**
```
FailCount:0, DoneCount:0, CircCount:106753, TimeoutCount:1001
FailCount:0, DoneCount:0, CircCount:110008, TimeoutCount:1001
FailCount:0, DoneCount:0, CircCount:110216, TimeoutCount:1001
```

### DIFF
--- a/src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
+++ b/src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
@@ -7,174 +7,137 @@
 
 using System;
 using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Pattern;
 using Akka.TestKit;
-using Akka.TestKit.Extensions;
-using Akka.Tests.Util;
+using Akka.Util.Internal;
 using Xunit;
+using Xunit.Sdk;
 
 namespace Akka.Tests.Pattern
 {
     public class ASynchronousCircuitBreakerThatIsClosed : CircuitBreakerSpecBase
     {
-        [Fact(DisplayName = "A synchronous circuit breaker that is closed should allow call through")]
-        public void Should_Allow_Call_Through( )
+        [Fact(DisplayName = "A synchronous circuit breaker that is closed must allow calls through")]
+        public void Must_allow_calls_through()
         {
-            var breaker = LongCallTimeoutCb( );
-            var result = breaker.Instance.WithSyncCircuitBreaker( ( ) => "Test" );
-
-            Assert.Equal( "Test", result );
+            var breaker = LongCallTimeoutCb();
+            breaker.Instance.WithSyncCircuitBreaker(SayHi).ShouldBe("hi");
         }
 
-        [Fact( DisplayName = "A synchronous circuit breaker that is closed should increment failure count when call fails" )]
-        public void Should_Increment_FailureCount_When_Call_Fails( )
+        [Fact(DisplayName = "A synchronous circuit breaker that is closed must increment failure count on failure")]
+        public void Must_increment_failure_count_on_failure()
         {
-            var breaker = LongCallTimeoutCb( );
-
-            Assert.Equal(0, breaker.Instance.CurrentFailureCount);
-            Assert.True( InterceptExceptionType<TestException>( ( ) => breaker.Instance.WithSyncCircuitBreaker( ThrowException ) ) );
-            Assert.True( CheckLatch( breaker.OpenLatch ) );
-            Assert.Equal( 1, breaker.Instance.CurrentFailureCount );
-        }
-
-        [Fact( DisplayName = "A synchronous circuit breaker that is closed should reset failure count when call succeeds" )]
-        public void Should_Reset_FailureCount_When_Call_Succeeds( )
-        {
-            var breaker = MultiFailureCb( );
-
-            Assert.Equal(0, breaker.Instance.CurrentFailureCount);
-            Assert.True( InterceptExceptionType<TestException>( ( ) => breaker.Instance.WithSyncCircuitBreaker( ThrowException ) ) );
-            Assert.Equal(1, breaker.Instance.CurrentFailureCount);
-
-            breaker.Instance.WithSyncCircuitBreaker( ( ) => "Test" );
-
-            Assert.Equal( 0, breaker.Instance.CurrentFailureCount );
-        }
-
-        [Fact(DisplayName = "A synchronous circuit breaker that is closed should increment failure count when call times out")]
-        public void Should_Increment_FailureCount_When_Call_Times_Out( )
-        {
-            var breaker = ShortCallTimeoutCb( );
-
-            breaker.Instance.WithSyncCircuitBreaker( ( ) => Thread.Sleep( 500 ) );
-
-            Assert.True( CheckLatch( breaker.OpenLatch ) );
-            Assert.Equal( 1, breaker.Instance.CurrentFailureCount );
+            var breaker = LongCallTimeoutCb();
+            breaker.Instance.CurrentFailureCount.ShouldBe(0);
+            InterceptException<TestException>(() => breaker.Instance.WithSyncCircuitBreaker(ThrowException));
+            Assert.True(CheckLatch(breaker.OpenLatch));
+            breaker.Instance.CurrentFailureCount.ShouldBe(1);
         }
 
         [Fact(DisplayName = "A synchronous circuit breaker that is closed must increment failure count on fail method")]
         public void Must_increment_failure_count_on_fail_method()
         {
             var breaker = LongCallTimeoutCb();
-            Assert.True(breaker.Instance.CurrentFailureCount == 0);
+            breaker.Instance.CurrentFailureCount.ShouldBe(0);
             breaker.Instance.Fail();
             Assert.True(CheckLatch(breaker.OpenLatch));
-            Assert.True(breaker.Instance.CurrentFailureCount == 1);
+            breaker.Instance.CurrentFailureCount.ShouldBe(1);
         }
 
-        [Fact(DisplayName = "A synchronous circuit breaker that is closed must reset failure count and clears cached last exception after success method")]
+        [Fact(DisplayName = "A synchronous circuit breaker that is closed must invoke onOpen if call fails and breaker transits to open state")]
+        public void Must_invoke_onOpen_if_call_fails_and_breaker_transits_to_open_state()
+        {
+            var breaker = ShortCallTimeoutCb();
+            InterceptException<TestException>(() => breaker.Instance.WithSyncCircuitBreaker(ThrowException));
+            Assert.True(CheckLatch(breaker.OpenLatch));
+        }
+
+        [Fact(DisplayName = "A synchronous circuit breaker that is closed must reset failure count after success")]
+        public void Must_reset_failure_count_after_success()
+        {
+            var breaker = MultiFailureCb();
+            breaker.Instance.CurrentFailureCount.ShouldBe(0);
+            InterceptException<TestException>(() => breaker.Instance.WithSyncCircuitBreaker(ThrowException));
+            breaker.Instance.CurrentFailureCount.ShouldBe(1);
+            breaker.Instance.WithSyncCircuitBreaker(SayHi);
+            breaker.Instance.CurrentFailureCount.ShouldBe(0);
+        }
+
+        [Fact(DisplayName = "A synchronous circuit breaker that is closed must reset failure count after success method")]
         public void Must_reset_failure_count_after_success_method()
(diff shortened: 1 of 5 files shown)


ismaelhamed:
NOTE: for whatever reason, `Task.WaitAsync` in .NET6 also fails sometimes.

ismaelhamed:
BTW, some of the CB's tests are failing now because they seem tweaked to work with the current CB implementation. If this PR goes ahead, I'll make sure to fix them all.

Aaronontheweb:
> BTW, some of the CB's tests are failing now because they seem tweaked to work with the current CB implementation. If this PR goes ahead, I'll make sure to fix them all.

got it - thanks for letting us know!

src/core/Akka.Tests/Pattern/CircuitBreakerStressSpec.cs
```diff
@@ -0,0 +1,134 @@
...
+            var stressActors = Enumerable.Range(0, 3).Select(i => Sys.ActorOf(Props.Create<StressActor>(breaker))).ToList();
+
+            for (var i = 0; i < 1000; i++)
+                foreach (var stressActor in stressActors)
+                {
+                    stressActor.Tell(RequestJob.Instance);
+                }
+
+            // let them work for a while
+            Thread.Sleep(3000);
```
Aaronontheweb, line 121: Nitpick - better to call `await Task.Delay` here. Has no semantic impact on the test but avoids taking a thread out of commission.

src/core/Akka/Util/Internal/AtomicState.cs
```diff
@@ -73,149 +74,78 @@ await Task
...
         public async Task<T> CallThrough<T>(Func<Task<T>> task)
         {
-            var deadline = DateTime.UtcNow.Add(_callTimeout);
-            ExceptionDispatchInfo capturedException = null;
-            T result = default(T);
+            var result = default(T);
             try
             {
-                result = await task().ConfigureAwait(false);
+                result = await task().WaitAsync(_callTimeout).ConfigureAwait(false);
```
Aaronontheweb, line 85: Got it - this fix makes sense to me. Can't trust the `task()` to necessarily cancel itself. Have to control whether or not it completes on-time via our own abstractions.

src/core/Akka/Util/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Akka.Util.Extensions
+{
+    internal static class TaskExtensions
+    {
+#if NETSTANDARD
+        public static async Task WaitAsync(this Task task, TimeSpan timeout)
```
Aaronontheweb, line 17: Separate issue that can be resolved at a later time: wonder if it's possible to reduce allocations around this via `ValueTask` or some of the other concepts + value types that were introduced in the .NET 6 implementation: https://andrewlock.net/a-deep-dive-into-the-new-task-waitasync-api-in-dotnet-6/

Not huge issue worth spending a lot of cycles on right now unless they're easy wins.

Aaronontheweb (approved):
LGTM - please proceed

ismaelhamed:
> Does #6108 need to be merged first or does this?

No, I've included the `CircuitBreakerStressSpec` in this one too.

src/core/Akka/Util/Extensions/TaskExtensions.cs
ismaelhamed: I'll include a improved version, based on the .NET6 implementation and with `CancellationToken` support to properly cancel the task.

src/core/Akka/Util/Internal/AtomicState.cs
ismaelhamed: Yep, this is where we were potentially awaiting a task indefinitely (depending on how long the task would take) despite the `CallTimeout`

ismaelhamed:
For `WithSyncCircuitBreaker` [#L259](https://github.com/akkadotnet/akka.net/blob/1b96294db2de1a85e670e3829c7c93a1f7a93687/src/core/Akka/Pattern/CircuitBreaker.cs#L259), I wonder if we should just do:

```csharp
public void WithSyncCircuitBreaker(Action body) => 
    WithCircuitBreaker(body, b => Task.Run(b)).GetAwaiter().GetResult();
```

Aaronontheweb:
LMK when this is ready for review

Aaronontheweb:
> For `WithSyncCircuitBreaker` [#L259](https://github.com/akkadotnet/akka.net/blob/1b96294db2de1a85e670e3829c7c93a1f7a93687/src/core/Akka/Pattern/CircuitBreaker.cs#L259), I wonder if we should just do:
> 
> ```cs
> public void WithSyncCircuitBreaker(Action body) => 
>     WithCircuitBreaker(body, b => Task.Run(b)).GetAwaiter().GetResult();
> ```

I think that's fine IMHO

ismaelhamed:
@Aaronontheweb this is ready for review. I have an improved version of the `WaitAsync` with some optimizations and support token cancellation (taken from the .NET6 implementation), but I'd prefer an initial review first.

Aaronontheweb:
Looks like you have a test suite issue here:

```
Akka.Tests.Pattern.ASynchronousCircuitBreakerThatIsClosed.A synchronous circuit breaker that is closed must increment failure count on callTimeout before call finishes
System.AggregateException : One or more errors occurred. (Timeout 00:00:00.1000000 expired while waiting for condition.
Expected: True
Actual:   False)
---- Timeout 00:00:00.1000000 expired while waiting for condition.
Expected: True
Actual:   False
```

Aaronontheweb:
Going to re-review this this week

src/core/Akka.TestKit/TestKitBase_AwaitConditions.cs
```diff
@@ -293,10 +293,59 @@ protected static async Task<bool> InternalAwaitConditionAsync(Func<Task<bool>> c
             return true;
         }
 
-        private static void ConditionalLog(ILoggingAdapter logger, string format, params object[] args)
+        protected void AwaitCond(Func<bool> p, TimeSpan? max = null, TimeSpan? interval = null, string message = "")
```
Aaronontheweb, line 296: Don't we already have an `AwaitCondition` method or does this do something different?

ismaelhamed:
I haven't had the time to work on this some more, and I'm still not sure why that test keeps failing. I even went ahead and reimplemented `Within` and `AwaitCond` (this one specially doesn't work like the JVM, so it might be skewing tests), but no luck.

src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
```diff
@@ -7,174 +7,137 @@
...
-            Assert.True(breaker.Instance.LastCaughtException is null);
+            breaker.Instance.CurrentFailureCount.ShouldBe(0);
         }
-    }
 
-    public class ASynchronousCircuitBreakerThatIsHalfOpen : CircuitBreakerSpecBase
-    {
-        [Fact(DisplayName = "A synchronous circuit breaker that is half open should pass call and transition to close on success")]
-        public void Should_Pass_Call_And_Transition_To_Close_On_Success( )
+        [Fact(DisplayName = "A synchronous circuit breaker that is closed must increment failure count on callTimeout before call finishes")]
```
Aaronontheweb, line 81: So this spec is currently racy

src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
```diff
@@ -7,174 +7,137 @@
...
-            Assert.True( CheckLatch( breaker.HalfOpenLatch ) );
-
-            var result = breaker.Instance.WithSyncCircuitBreaker( ( ) => SayTest( ) );
-
-            Assert.True( CheckLatch( breaker.ClosedLatch ) );
-            Assert.Equal( SayTest( ), result );
+            var breaker = ShortCallTimeoutCb();
+            Task.Run(() => breaker.Instance.WithSyncCircuitBreaker(() => Thread.Sleep(Dilated(TimeSpan.FromSeconds(1)))));
+            Within(TimeSpan.FromMilliseconds(900),
+                () => AwaitCond(() => breaker.Instance.CurrentFailureCount == 1, Dilated(TimeSpan.FromMilliseconds(100))));
```
Aaronontheweb, line 87: I think your problem here is your parameters on `AwaitCond` - you meant to set `Dilated(TimeSpan.FromMilliseconds(100))` as the interval but it's being used as the max value here. @ismaelhamed

src/core/Akka/Util/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+//     Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+namespace System.Threading.Tasks
+{
+    internal static class TaskExtensions
+    {
+//#if NETSTANDARD
+        public static async Task WaitAsync1(this Task task, TimeSpan timeout)
```
Aaronontheweb, line 13: Need help with the `#if` targeting here or is this just commented out for debugging purposes?

Aaronontheweb (commented):
I think I have your failing test figured out

src/core/Akka.TestKit/TestKitBase_AwaitConditions.cs
ismaelhamed: The behavior of `AwaitCondition` is different from the JVM, that's why I tried a straight port instead. No luck, keeps failing.

src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
ismaelhamed: https://github.com/akka/akka/blob/13892f7d01e2b0b0a4fbb7e20b3bed3bafee3dde/akka-actor-tests/src/test/scala/akka/pattern/CircuitBreakerSpec.scala#L413

src/core/Akka/Util/Extensions/TaskExtensions.cs
ismaelhamed: Reverted back

Aaronontheweb:
@ismaelhamed is this ready for review?

ismaelhamed:
> @ismaelhamed is this ready for review?

The implementation yes, but I couldn't figure out where the problem with the specs was. I'll give it another shot soon.

src/core/Akka.TestKit/TestKitBase_AwaitConditions.cs
ismaelhamed: `AwaitCondition` arbitrarily calculates the `interval`(when not specified) [as a 10th](https://github.com/akkadotnet/akka.net/blob/dc232c6c38dd346631a5d15605be40fca8633244/src/core/Akka.TestKit/TestKitBase_AwaitConditions.cs#L45) of `max`. It should be a fixed `100 ms`, otherwise tests ported from the JVM might not behave as expected.

src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
ismaelhamed: @Aaronontheweb , see my comment above. In the meantime, I've fixed it by passing both `max` and `interval`.

ismaelhamed:
Unrelated test failing now.

Aaronontheweb:
@ismaelhamed is this good for me to review again?

Aaronontheweb:
ah yes, it is - you just requested one from me! I'll get right on it.

Aaronontheweb:
Queued for auto-merge - nice work @ismaelhamed

PR #6120, opened 2022-09-27
Fix GetEntityLocation use wrong actor path

Arkatufus:
Fixes:
`GetEntityLocation` does not work with `HashCodeMessageExtractor`

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
@@ -1005,6 +1005,7 @@ namespace Akka.Cluster.Sharding
 
             try
             {
+                var entityId = getEntityLocation.EntityId;
                 var shardId = ExtractShardId(new StartEntity(getEntityLocation.EntityId));
                 if (string.IsNullOrEmpty(shardId))
                 {
@@ -1046,7 +1047,7 @@ namespace Akka.Cluster.Sharding
                         // NOTE: in the event that we're querying a shard's location from a ShardRegionProxy
                         // the shard may not be technically "homed" inside the proxy, but it does exist.
 #pragma warning disable CS4014
-                        ResolveEntityRef(GetNodeAddress(shardRegionRef), shardRegionRef.Path / shardId / shardId); // needs to run as a detached task
+                        ResolveEntityRef(GetNodeAddress(shardRegionRef), shardRegionRef.Path / shardId / entityId); // needs to run as a detached task
 #pragma warning restore CS4014
                     }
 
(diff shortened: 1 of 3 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -1046,15 +1047,15 @@ async Task ResolveEntityRef(Address destinationAddress, ActorPath entityPath)
                         // NOTE: in the event that we're querying a shard's location from a ShardRegionProxy
                         // the shard may not be technically "homed" inside the proxy, but it does exist.
 #pragma warning disable CS4014
-                        ResolveEntityRef(GetNodeAddress(shardRegionRef), shardRegionRef.Path / shardId / shardId); // needs to run as a detached task
+                        ResolveEntityRef(GetNodeAddress(shardRegionRef), shardRegionRef.Path / shardId / entityId); // needs to run as a detached task
```
Arkatufus, line 1050: The actor path should be `shardRegionRef.Path / shardId / entityId` not `shardRegionRef.Path / shardId / shardId`

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
```diff
@@ -1046,15 +1047,15 @@ async Task ResolveEntityRef(Address destinationAddress, ActorPath entityPath)
...
 #pragma warning restore CS4014
                     }
 
                     return;
                 }
                 
 #pragma warning disable CS4014
-                ResolveEntityRef(GetNodeAddress(shardActorRef), shardActorRef.Path / shardId); // needs to run as a detached task
+                ResolveEntityRef(GetNodeAddress(shardActorRef), shardActorRef.Path / entityId); // needs to run as a detached task
```
Arkatufus, line 1058: Same here, actor path should be `shardActorRef.Path / entityId`, not `shardActorRef.Path / shardId`

src/contrib/cluster/Akka.Cluster.Sharding.Tests/ShardRegionQueriesSpecs.cs
```diff
@@ -70,11 +70,11 @@ private string ExtractShardId(object message)
             switch (message)
             {
                 case int i:
-                    return (i % 10).ToString();
+                    return (i % 10 + 1).ToString();
                 // must support ShardRegion.StartEntity in order for
                 // GetEntityLocation to work properly
                 case ShardRegion.StartEntity se:
-                    return se.EntityId;
+                    return (int.Parse(se.EntityId) % 10 + 1).ToString();
```
Arkatufus, line 77: Change the test `IMessageExtractor` so that the `ShardId` does not equal to `EntityId` to catch actor path bugs

Aaronontheweb: doh - that's a good catch

src/contrib/cluster/Akka.Cluster.Sharding/ShardRegion.cs
Aaronontheweb: Yep, my mistake

Aaronontheweb (approved):
LGTM - probably need to push a new release

PR #6152, opened 2022-10-06
Block Xunit2 TestKit until unit test logger is ready

### LINKED ISSUE #6125: `TestkitBase.Log` doesn't log
**Version Information**
1.5.0

**Describe the bug**
If I call `Log.*()` from inside a test, I expect the log to show in the test output window (as long as the `akka.loglevel` allows it).

**To Reproduce**
In this instance of a `CircuitBreaker` test, directly logging to the `OutputHelper` works but via the `Log` property doesn't:

```csharp
public class ASynchronousCircuitBreakerThatIsClosed : CircuitBreakerSpecBase
{
    public ASynchronousCircuitBreakerThatIsClosed(ITestOutputHelper outputHelper)
        : base(outputHelper)
    { }

    [Fact(DisplayName = "A synchronous circuit breaker that is closed should allow call through")]
    public void Should_Allow_Call_Through( )
    {
        Output.WriteLine("LongCallTimeoutCb1"); // works
        Log.Info("LongCallTimeoutCb2"); // this one doesn't work

        var breaker = LongCallTimeoutCb( );
        var result = breaker.Instance.WithSyncCircuitBreaker( ( ) => "Test" );

        Assert.Equal( "Test", result );
    }
}

public class CircuitBreakerSpecBase : AkkaSpec
{
    private static Config Config => ConfigurationFactory.ParseString("akka.loglevel = INFO");

    public CircuitBreakerSpecBase(ITestOutputHelper outputHelper)
        : base(outputHelper, Config)
    { }
    ...
}
```

Also, even thought the `akka.loglevel` has been set to `INFO`, you can always see these three DEBUG traces at the end:
```
LongCallTimeoutCb1
[DEBUG][29/09/2022 8:16:31][Thread 0012][ActorSystem(CircuitBreakerSpecBase-1)] System shutdown initiated
[DEBUG][29/09/2022 8:16:31][Thread 0022][EventStream] Shutting down: StandardOutLogger started
[DEBUG][29/09/2022 8:16:31][Thread 0022][EventStream] All default loggers stopped
```

Arkatufus:
Fixes #6125

Note that since we're blocking during test start, there are a chance that some tests that timed to system startup might start to racily fail.

### DIFF
--- a/src/contrib/testkits/Akka.TestKit.Xunit2/Internals/Loggers.cs
+++ b/src/contrib/testkits/Akka.TestKit.Xunit2/Internals/Loggers.cs
@@ -35,6 +35,7 @@ namespace Akka.TestKit.Xunit2.Internals
             Receive<InitializeLogger>(e =>
             {
                 e.LoggingBus.Subscribe(Self, typeof (LogEvent));
+                Sender.Tell(new LoggerInitialized());
             });
         }
 
--- a/src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
+++ b/src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
@@ -145,7 +145,8 @@ namespace Akka.TestKit.Xunit2
             {
                 var extSystem = (ExtendedActorSystem)system;
                 var logger = extSystem.SystemActorOf(Props.Create(() => new TestOutputLogger(Output)), "log-test");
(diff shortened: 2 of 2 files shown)


Arkatufus:
Need to see if this breaks anything in the unit test or not.

Aaronontheweb (approved):
LGTM

PR #6175, opened 2022-10-13
Fix DbCommand.CommandTimeout

### LINKED ISSUE #6174: Connection timeout is not respected for event journal commands when using BatchingSqlJournal
**Version Information**
Akka - v1.4.43
Akka.Persistence - v1.4.43
Akka.Persistence.SqlServer - v1.4.35

**Describe the bug**
The connection timeout is not respected when using the `BatchingSqlJournal`.

**To Reproduce**
1. Configure `BatchingSqlServerJournal` as the persistence journal
2. Set a connection timeout for the journal (e.g. with HOCON: `akka.persistence.journal.sql-server.connection-timeout = 5s`)
3. Lock the `EventJournalTable` (e.g. running `SELECT * FROM [EventJournal] WITH (TABLOCKX, HOLDLOCK)` in a separate transaction)
4. Try to recover a persistent actor or try to persist an event from a persistent actor

**Expected behavior**
The event journal command should time out after the configured connection timeout duration.

**Actual behavior**
The event journal command does not time out after the configured connection timeout duration.

**Environment**
Windows 10
.NET 6

**Additional context**
The issue appears to be caused by [this](https://github.com/akkadotnet/akka.net/blob/eba7bf50d4766ca7160edad6af18bdaf4970fbc7/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs#L977) line in the `BatchingSqlJournal` base class. The `CommandTimeout` property on the command [expects](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlcommand.commandtimeout?view=sqlclient-dotnet-core-5.0) the timeout duration to be specified in seconds but it is specified in milliseconds. I.e. a connection timeout of 5 seconds in the journal configuration results in a timeout of 5000 seconds on the command.

Arkatufus:
Fixes #6174

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
@@ -974,7 +974,7 @@ namespace Akka.Persistence.Sql.Common.Journal
                 using (var tx = connection.BeginTransaction(Setup.IsolationLevel))
                 using (var command = (TCommand)connection.CreateCommand())
                 {
-                    command.CommandTimeout = (int)Setup.ConnectionTimeout.TotalMilliseconds;
+                    command.CommandTimeout = (int)Setup.ConnectionTimeout.TotalSeconds;
                     command.Transaction = tx;
                     try
                     {


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -974,7 +974,7 @@ private async Task<BatchComplete> ExecuteChunk(RequestChunk chunk, IActorContext
                 using (var tx = connection.BeginTransaction(Setup.IsolationLevel))
                 using (var command = (TCommand)connection.CreateCommand())
                 {
-                    command.CommandTimeout = (int)Setup.ConnectionTimeout.TotalMilliseconds;
+                    command.CommandTimeout = (int)Setup.ConnectionTimeout.TotalSeconds;
```
Arkatufus, line 977: `DbCommand.CommandTimeout` is expressed in seconds, not milliseconds.

Aaronontheweb:
Needs to be backported to v1.4

PR #6176, opened 2022-10-13
Fix OldCoordinatorStateMigrationEventAdapter.Manifest

### LINKED ISSUE #6159: Akka.Persistence: returning `string.Empty` in `IWriteEventAdapter` can overwrite valid manifest
**Version Information**
Version of Akka.NET? v1.4.44, 1.5.0-alpha1
Which Akka.NET Modules? Akka.Persistence, Akka.Cluster.Sharding

**Describe the bug**

While working with the `OldCoordinatorStateMigrationEventAdapter`

https://github.com/akkadotnet/akka.net/blob/1b96294db2de1a85e670e3829c7c93a1f7a93687/src/contrib/cluster/Akka.Cluster.Sharding/OldCoordinatorStateMigrationEventAdapter.cs#L16-L36

I noticed that the rows in my SQL Server database for testing purposes had the following in their "manifest" column:

image

Those messages should belong to `serializerId=13`, the sharding serializer. Instead they belong to the `StringSerializer`.

**To Reproduce**

Create any `IWriteEventAdapter` and implement the `Manifest` method to return `string.Empty` in combination with a `StringWithManifest` serializer.

**Expected behavior**

Original manifest is preserved unless `string.IsNullOrEmpty` returns `false` for the `IWriteEventAdapter.Manifest` method.

**Actual behavior**
What actually happened and how did it differ from your expectations?

Original manifest is overwritten with a `System.String` type.

**Additional context**

It seems like this would be a major issue for anyone using tags et al unless everyone is using Newtonsoft.Json / Hyperion, which embed type metadata directly into the payloads. Either that or it's a bug with the new Akka.Cluster.Sharding code in v1.5.

Arkatufus:
Fixes #6159

## Changes

- Refactor `ClusterShardingMessageSerializer.Manifest()` logic to static method
- Make `OldCoordinatorStateMigrationEventAdapter.Manifest()` forward the proper manifest to the journal

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/OldCoordinatorStateMigrationEventAdapter.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/OldCoordinatorStateMigrationEventAdapter.cs
@@ -5,6 +5,8 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using Akka.Cluster.Sharding.Serialization;
+using Akka.Persistence;
 using Akka.Persistence.Journal;
 
 namespace Akka.Cluster.Sharding
@@ -17,7 +19,10 @@ namespace Akka.Cluster.Sharding
     {
         public string Manifest(object evt)
         {
-            return "";
+            if (evt is IPersistentRepresentation p)
+                return p.Manifest;
+            
+            return ClusterShardingMessageSerializer.GetManifest(evt);
         }
 
         public object ToJournal(object evt)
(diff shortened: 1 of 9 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/OldCoordinatorStateMigrationEventAdapter.cs
```diff
@@ -17,7 +19,10 @@ internal sealed class OldCoordinatorStateMigrationEventAdapter : IEventAdapter
     {
         public string Manifest(object evt)
         {
-            return "";
+            if (evt is IPersistentRepresentation p)
+                return p.Manifest;
+            
+            return ClusterShardingMessageSerializer.GetManifest(evt);
```
Aaronontheweb, line 25: Make a comment here - "this should and will throw if `evt` is not a type supported by the ClusterShardingSerializer"

This actually does need to throw.

Aaronontheweb (changes requested):
Make `internal` rather than `public`

Aaronontheweb:
@Arkatufus I think you need to re-run API approvals

Aaronontheweb (approved):

PR #6181, opened 2022-10-14
remove `Deadline` allocations from Akka.Remote

Aaronontheweb:
## Changes

Made `Deadline` into a `readonly struct` - it's a minor source of allocations but relatively easy to remove.

## Checklist

### Latest `dev` Benchmarks 

Benchmark throughput data emerging from this change is negligible - showed up in Rider's dynamic profiling report as a source of 200mb of allocation over the course of running RemotePingPong.

### DIFF
--- a/src/benchmark/RemotePingPong/Program.cs
+++ b/src/benchmark/RemotePingPong/Program.cs
@@ -116,9 +116,11 @@ namespace RemotePingPong
                 var bestThroughput = 0L;
                 foreach (var throughput in GetClientSettings())
                 {
+                    GC.Collect();
                     var result1 = await Benchmark(throughput, repeat, bestThroughput, redCount);
                     bestThroughput = result1.Item2;
                     redCount = result1.Item3;
+                    GC.Collect(); // for good measure
                 }
             }
 
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
@@ -66,7 +66,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Internal
         /// <summary>
         /// TBD
         /// </summary>
-        protected Deadline PruneDeadline = null;
+        protected Deadline PruneDeadline = Deadline.Never;
 
         /// <summary>
         /// Used to toggle what we do during publication when there are no subscribers
@@ -109,7 +109,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Internal
                 case Subscribe subscribe:
                     Context.Watch(subscribe.Ref);
                     Subscribers.Add(subscribe.Ref);
-                    PruneDeadline = null;
+                    PruneDeadline = Deadline.Never;
                     Context.Parent.Tell(new Subscribed(new SubscribeAck(subscribe), Sender));
                     return true;
 
@@ -124,9 +124,9 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Internal
                     return true;
 
                 case Prune _:
-                    if (PruneDeadline != null && PruneDeadline.IsOverdue)
+                    if (PruneDeadline != Deadline.Never && PruneDeadline.IsOverdue)
                     {
-                        PruneDeadline = null;
+                        PruneDeadline = Deadline.Never;
                         Context.Parent.Tell(NoMoreSubscribers.Instance);
                     }
 
@@ -224,7 +224,7 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Internal
                             NewGroupActor(encodedGroup).Forward(message);
                         }
                     });
-                    PruneDeadline = null;
+                    PruneDeadline = Deadline.Never;
                     return true;
 
                 case Unsubscribe unsubscribe when unsubscribe.Group != null:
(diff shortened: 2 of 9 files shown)


src/core/Akka.Remote/Deadline.cs
```diff
@@ -115,6 +120,16 @@ public static Deadline Now
                 return deadline;
         }
 
+        public static bool operator ==(Deadline left, Deadline right)
+        {
+            return left.Equals(right);
```
Arkatufus, line 125: I just checked the `Equals()` method (line 62) and it is doing a dangerous explicit cast, which will throw if anything else is passed into it. We might as well fix that in this PR.
A ReferenceEquals() shortcut in there would also be good.

src/benchmark/RemotePingPong/Program.cs
```diff
@@ -116,9 +116,11 @@ private static async Task Start(uint timesToRun)
                 var bestThroughput = 0L;
                 foreach (var throughput in GetClientSettings())
                 {
+                    GC.Collect();
```
Aaronontheweb, line 119: Doh, I'll need to roll this back  - didn't mean to commit this.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
```diff
@@ -66,7 +66,7 @@ internal abstract class TopicLike : ActorBase
         /// <summary>
         /// TBD
         /// </summary>
-        protected Deadline PruneDeadline = null;
+        protected Deadline PruneDeadline = Deadline.Never;
```
Aaronontheweb, line 69: `Deadline.Never` is the new default. A non-expiring `Deadline`.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
```diff
@@ -124,9 +124,9 @@ protected bool DefaultReceive(object message)
                     return true;
 
                 case Prune _:
-                    if (PruneDeadline != null && PruneDeadline.IsOverdue)
+                    if (PruneDeadline != Deadline.Never && PruneDeadline.IsOverdue)
```
Aaronontheweb, line 127: Can probably get rid of the `PruneDeadline != Deadline.Never` here.

src/core/Akka.Cluster/Gossip.cs
```diff
@@ -478,7 +478,7 @@ public GossipEnvelope(UniqueAddress from, UniqueAddress to, Gossip gossip, Deadl
         /// <summary>
         /// TBD
         /// </summary>
-        public Deadline Deadline { get; set; }
+        public Deadline? Deadline { get; set; }
```
Aaronontheweb, line 481: Still going to use nullable Deadlines in Akka.Cluster gossip.

src/core/Akka.Remote/Endpoint.cs
```diff
@@ -528,7 +528,7 @@ protected override SupervisorStrategy SupervisorStrategy()
                 _log.Warning("Association with remote system {0} has failed; address is now gated for {1} ms. Reason is: [{2}]", _remoteAddress, _settings.RetryGateClosedFor.TotalMilliseconds, ex);
                 UidConfirmed = false; // Need confirmation of UID again
 
-                if ((_resendBuffer.Nacked.Any() || _resendBuffer.NonAcked.Any()) && _bailoutAt == null)
+                if ((_resendBuffer.Nacked.Any() || _resendBuffer.NonAcked.Any()) && _bailoutAt == Deadline.Never)
```
Aaronontheweb, line 531: Check for `_bailloutAt == Deadline.Never` instead of `null`; always reset to `Deadline.Never` instead of `null`.

Aaronontheweb (commented):
Left some comments on my own PR.

Arkatufus (changes requested):
Needed a small fix, would be better if we modernize the `Equals()` override too.

src/core/Akka.Remote/Deadline.cs
Aaronontheweb: I don't see an explicit cast, only this:

```csharp
public override bool Equals(object obj) =>
                obj is Deadline deadline && Equals(deadline);
```

Arkatufus: Hmmm... weird... I'm talking about these code:
https://github.com/akkadotnet/akka.net/blob/aca6850eb68bbdb64f029bad019d3e58e4bb290f/src/core/Akka.Remote/Deadline.cs#L60-L69

Aaronontheweb: Found it - very odd

Arkatufus (approved):
LGTM

Aaronontheweb:
Looks like we definitely broke this test:

```
Akka.Cluster.Metrics.Tests.ClusterMetricsExtensionSpec.Metrics_extension_Should_collect_metrics_after_start_command
Expected MetricsHistorySize to be greater or equal to 10, but found 4.
```

I can't reproduce locally yet, but I'll look into it.

Aaronontheweb:
Still dealing with some bugs emerging from making `Deadline` a `struct`

Aaronontheweb:
There's still something wrong with the `Cluster.Join` code and how these `Deadline` changes affect it.

PR #6201, opened 2022-10-19
Revert ConfigurationException due to binary incompatibility

### LINKED ISSUE #6200: #6167 is not backward compatible
**Version Information**
Version of Akka.NET? 1.4.44
Which Akka.NET Modules? Core

**Describe the bug**
A user reported that #6167 causes Lighthouse to crash because it could not find `public static ConfigurationException NullOrEmptyConfig<T>(string path = null)` method fingerprint in `ConfigurationException` class

Arkatufus:
Fixes #6200

## Changes
Revert changes to the static method

### DIFF
--- a/src/core/Akka.Cluster/ClusterSettings.cs
+++ b/src/core/Akka.Cluster/ClusterSettings.cs
@@ -35,7 +35,7 @@ namespace Akka.Cluster
             //TODO: Requiring!
             var clusterConfig = config.GetConfig("akka.cluster");
             if (clusterConfig?.GetConfig("failure-detector") == null)
-                throw ConfigurationException.NullOrEmptyConfig<ClusterSettings>("akka.cluster", "Did you forgot to set the 'akka.cluster.provider' HOCON property to 'cluster'?");
+                throw new ConfigurationException($"Failed to instantiate {nameof(ClusterSettings)}: Configuration does not contain `akka.cluster` node. Did you forgot to set the 'akka.cluster.provider' HOCON property to 'cluster'?");
 
             LogInfoVerbose = clusterConfig.GetBoolean("log-info-verbose", false);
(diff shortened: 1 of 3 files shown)


src/core/Akka/Configuration/ConfigurationException.cs
```diff
@@ -16,11 +16,11 @@ namespace Akka.Configuration
     /// </summary>
     public class ConfigurationException : AkkaException
     {
-        public static ConfigurationException NullOrEmptyConfig<T>(string path = null, string reason = null)
+        public static ConfigurationException NullOrEmptyConfig<T>(string path = null)
```
Arkatufus, line 19: Reverting this back to the old method fingerprint

src/core/Akka.Cluster/ClusterSettings.cs
```diff
@@ -35,7 +35,7 @@ public ClusterSettings(Config config, string systemName)
             //TODO: Requiring!
             var clusterConfig = config.GetConfig("akka.cluster");
             if (clusterConfig?.GetConfig("failure-detector") == null)
-                throw ConfigurationException.NullOrEmptyConfig<ClusterSettings>("akka.cluster", "Did you forgot to set the 'akka.cluster.provider' HOCON property to 'cluster'?");
+                throw new ConfigurationException($"Failed to instantiate {nameof(ClusterSettings)}: Configuration does not contain `akka.cluster` node. Did you forgot to set the 'akka.cluster.provider' HOCON property to 'cluster'?");
```
Arkatufus, line 38: Just bake the exception message instead of using the helper method

Aaronontheweb (changes requested):
Needs API approvals

Arkatufus:
Done

Aaronontheweb (approved):

Aaronontheweb:
@Arkatufus going to give tests a chance to run, then we can push a new release

PR #6203, opened 2022-10-19
Improve Akka.Cluster.Metrics collected values

### LINKED ISSUE #4142: Improve CPU/Memory metrics collection at Akka.Cluster.Metrics
## Introduction ##

Once #4126 will be merged, we will need to improve metrics loading, that is implemented in `DefaultCollector`.

The basic idea is to collect:

- Current CPU load on the node, in %
- Current Memory usage, in %

Using that information, `AdaptiveLoadBalancingRoutingLogic` will calculate availability of each node and will perform "smart" routing.

Ideally, we should collect more complete list:

- CPU usage by current process
- CPU total usage on the node machine (this will be used for routing)
- Memory usage by current process (in bytes)
- Memory total usage on the node machine
- Memory available on the node machine (this with previous one will give total utilization in %)

### CPU ###

There is no API available in `netstandard2.1` to collect CPU metrics out-of-the-box, like `PerformanceCounters`. So what we are doing now is using `Process.TotalProcessorTime` property, to get time, dedicated to current process. Having total time elapsed, we can give some estimation of CPU usage by current process.

But talking about CPU **total** usage, this approach would require to get all processes info with `Process.GetProcesses()` - which is very time consuming (especially when we have to deal with access violation exceptions here), when there are lots of processes.
So total CPU usage is just **the same** as current process CPU usage now. This is more or less fine for routing based on .NET process load, but not ideal if there are some other heavy processes running on machine.

## Memory ##

 Candidate list includes:

- `GC.GetTotalMemory` to get currently allocated managed memory size. There is also [`GC.GetGCMemoryInfo`](https://docs.microsoft.com/ru-ru/dotnet/api/system.gc.gettotalmemory?view=netframework-4.8) - that will provide struct with `TotalAvailableMemoryBytes` property, but this method is only available at `.netstandard3.0`, and we are targeting `2.1`

- PerformanceCouters, which are working under Windows, and there is Mono implementation. There are some other Windows-only ways to get metrics.

- `Process` class, which provides multiple memory-related properties

- Using P/Invoke and working with native API

- Getting some shell commands output, specific for OS

Currently, we are using the cross-platform sources available for `netstandard2.1` - the `Process` class. 

#### First issue ####
Same as for CPU: this is quite heavy to get all processes information. So current implementation treats `MemoryUsage` as current process usage, which is useful, but not ideal for nodes routing.

#### Second issue ####
Another issue is understanding the term of "used" memory, and getting "available" memory info. 

To track unmanaged memory as well as managed, [`Process.PrivateMemorySize64`](https://docs.microsoft.com/ru-ru/dotnet/api/system.diagnostics.process.privatememorysize64?view=netframework-4.8) is used instead of `GC.GetTotalMemory`. It works well by itself. But it is hard to know the upper limit for this value, because it is not the *allocated physical memory from RAM* (see documentation). 
Getting "available" memory is much more tricky, and I did not find anything available under .NET Core sdk to get this value. Ideally would be getting available size of installed physical memory (or available part of it in cloud environment). So far, the `Process.VirtualMemorySize64` is used - but is is just a number of bytes in **virtual** address space, and does not correlate much with really available memory. But still it is one of the upper bounds for available memory, and can be used to get % of memory load (relative to other node).

In my understanding, ideally would be loading `Available MBytes` PerformanceCounter (but on all platforms) to get available memory, and get some way to load installed total available memory. This two would allow to get `% Used Memory` on the node, and perform routing. And provide all different `Process` properties in addition, like `WorkingSet`, `PrivateMemorySize64`, and others.

Maybe there is some other convenient approach. The main idea here is that while current `used / available` relation is `Process.PrivateMemorySize64 / Process.VirtualMemorySize64` - it is always is range of `[0, 1]` and reflects the memory load. So we can compare nodes based on this. But value of `0.5` does not guarantee that there is available memory on the node at all, so need some more accurate values for node's  memory capacity calculation.

Arkatufus:
close #4142

This improves the values collected by the metrics plugin, it does not fix the heap selector since it is still missing informations.
Heap metric based routing can only be implemented after we implement the new `EventCounters` API.


Aaronontheweb (approved):
LGTM but needs API approvals

Aaronontheweb:
We will also need to backport this to v1.5 @Arkatufus

Aaronontheweb:
@Arkatufus looks like some of the MNTR specs for Akka.Cluster.Metrics are not passing. Could you investigate?

Arkatufus:
Changed the code, available memory is the virtual size, used memory is the working set

src/contrib/cluster/Akka.Cluster.Metrics/Collectors/DefaultCollector.cs
```diff
@@ -55,12 +55,11 @@ public NodeMetrics Sample()
...
             {
+                process.Refresh();
                 var metrics = new List<NodeMetrics.Types.Metric>()
                 {
                     // Memory
-                    // Forcing garbage collection to keep metrics more resilent to occasional allocations
-                    NodeMetrics.Types.Metric.Create(StandardMetrics.MemoryUsed, GC.GetTotalMemory(true)).Value,
-                    // VirtualMemorySize64 is not best idea here...
+                    NodeMetrics.Types.Metric.Create(StandardMetrics.MemoryUsed, process.WorkingSet64).Value,
                     NodeMetrics.Types.Metric.Create(StandardMetrics.MemoryAvailable, process.VirtualMemorySize64).Value,
```
Aaronontheweb, line 63: this should be working set

src/contrib/cluster/Akka.Cluster.Metrics/Collectors/DefaultCollector.cs
```diff
@@ -55,12 +55,11 @@ public NodeMetrics Sample()
         {
             using (var process = Process.GetCurrentProcess())
             {
+                process.Refresh();
                 var metrics = new List<NodeMetrics.Types.Metric>()
                 {
                     // Memory
-                    // Forcing garbage collection to keep metrics more resilent to occasional allocations
-                    NodeMetrics.Types.Metric.Create(StandardMetrics.MemoryUsed, GC.GetTotalMemory(true)).Value,
```
Aaronontheweb, line 62: This should be GC

src/contrib/cluster/Akka.Cluster.Metrics/Collectors/DefaultCollector.cs
```diff
@@ -55,12 +55,11 @@ public NodeMetrics Sample()
         {
             using (var process = Process.GetCurrentProcess())
```
Aaronontheweb, line 56: I wonder if this is wrong - shouldn't we just cache this value somewhere?

Aaronontheweb:
The right algorithm, which we can't implement on .NET Standard 2.0 without a lot of hacks:

used memory = gc.totalsize OR process.workingset64
available memory = GC.GetGCMemoryInfo().TotalAvailableMemoryBytes 

We should be measuring how close we're getting to hitting the total amount of available memory on the system as a percentage. In the meantime, we can do the following on .NET Standard 2.0:

used memory = gc.totalsize 
available memory = process.workingset64 + process.pagedmemorysize64

This will tell us how much memory is being consumed by the .NET application itself - it won't be able to factor in issues like system memory and some unmanaged memory without the user manually calling `GC.AddMemoryPressure` but that's probably the best compromise we can make.

Aaronontheweb:
Testing out the latter now - we'll go back to the drawing board if we keep running into problems getting the MNTR to pass.

Aaronontheweb:
@Arkatufus needs a port to v1.5

PR #6207, opened 2022-10-21
Fix syntax error in Testing actor systems documentation

tjaart:
Fixes #
The call to `Within()` in the documents contained invalid c# syntax.


Aaronontheweb (approved):
LGTM

Issue #6214, opened 2022-10-25
PingPongBenchmarks unexpected results warning 


brah-mcdude:
When running the PingPongBenchmarks I get a warning: 

// * Warnings *
Configuration
  Summary -> The exporter BenchmarkDotNet.Exporters.MarkdownExporter is already present in configuration. There may be unexpected results.

Aaronontheweb:
I think this is a function of us using some Benchmark.NET configuration elements that are due to be deprecated. Would happily accept a PR fixing them in any of our BDN projects.

brah-mcdude:
Already on it.

PR #6215, opened 2022-10-25
PingPongBenchmarks unexpected results warning

brah-mcdude:
Removed [Config(typeof(MonitoringConfig))].
Issue seems to be resolved.

Fixes PingPongBenchmarks unexpected results warning #6214


src/benchmark/Akka.Benchmarks/Actor/PingPongBenchmarks.cs
```diff
@@ -15,7 +15,6 @@
 
 namespace Akka.Benchmarks.Actor
 {
-    [Config(typeof(MonitoringConfig))]
```
Aaronontheweb, line 18: Wasn't the issue with how `MonitoringConfig` is defined?

Aaronontheweb (changes requested):
Isn't the issue in this case how we're using BDN's configuration? We don't really want to fix the issue by removing that config altogether.

Aaronontheweb:
Did you see my comments @brah-mcdude

PR #6221, opened 2022-11-01
Report cause for Akka/IO TCP CommandFailed events

ismaelhamed:
Port [#22954](https://github.com/akka/akka/pull/22954)

## Changes

> I think we should finally fix this once and forever. We have so many failed test jobs where we cannot be sure why "Connect failed" and we have no way of finding out without enabling DEBUG log level everywhere. That same pain is felt by every user.

## Checklist


src/core/Akka/IO/Tcp.cs
```diff
@@ -808,18 +810,32 @@ public sealed class CommandFailed : Event
...
+            public Option<Exception> Cause { get; private set; } = Option<Exception>.None;
+
+            /// <summary>
+            /// Creates a copy of this object with a new cause set.
+            /// </summary>
+            [InternalApi]
+            public CommandFailed WithCause(Exception cause)
+            {
+                // Needs to be added with a mutable property for compatibility reasons
+                return new CommandFailed(Cmd) { Cause = cause };
```
ismaelhamed, line 832: Not sure if this holds true in C# or we could just add a second constructor instead.

Aaronontheweb: We could just add a constructor overload and that would work

Aaronontheweb: Although I think this is probably a fine way of doing it

Aaronontheweb (approved):
LGTM

Aaronontheweb:
@Arkatufus can we backport to v1.4?

PR #6223, opened 2022-11-01
Fix broken object type serializer in QueryExecutor

Arkatufus:
Fix for https://github.com/akkadotnet/Akka.Hosting/issues/127
Fix https://github.com/akkadotnet/akka.net/issues/6389

## Changes
* Removed persistence default deserializer feature

## Issue
Any non-serializer mapped objects being serialized by `Akka.Persistence.Sql.Common.Journal.QueryExecutor` will use the JSON serializer and not the object type serializer.

Problem is with this line:
https://github.com/akkadotnet/akka.net/blob/39a8f9e6822592b3c401fb55315f6f09a21ab260/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs#L783

which takes its value from the very obscure and very rarely used HOCON override here:
https://github.com/akkadotnet/akka.net/blob/39a8f9e6822592b3c401fb55315f6f09a21ab260/src/core/Akka.Persistence/persistence.conf#L119

## Workaround
Override the HOCON setting:

```
akka.persistence.journal-plugin-fallback.serializer = null
akka.persistence.snapshot-store-plugin-fallback.serializer = null
```

or declare the default non-mapped object serializer in the journal/snapshot settings:
```
akka.persistence.journal.sqlite.serializer = mySerializer
akka.persistence.snapshot-store.sqlite.serializer = mySerializer
```


src/contrib/persistence/Akka.Persistence.Sqlite.Tests/CustomObjectSerializerSpec.cs
```diff
@@ -0,0 +1,129 @@
...
+            actor.Tell("a", probe);
+            probe.ExpectMsg("a");
+
+            var conn = new SqliteConnection("DataSource=AkkaJournal.db");
+            conn.Open();
+            const string sql = "SELECT ej.serializer_id FROM event_journal ej WHERE ej.persistence_id = 'a'";
+            await using var cmd = new SqliteCommand(sql, conn);
+            var record = await cmd.ExecuteReaderAsync();
+            await record.ReadAsync();
+            record[0].Should().Be(9999);
```
Arkatufus, line 71: This fails because the payload is being serialized using the JSON serializer (1) and not MySerializer (9999)

Aaronontheweb:
Nice work @Arkatufus - yep, that will do it.

So what can be done to safely change this default, given that this behavior is somewhat destructive but has also been in-place for years?

Arkatufus:
To be honest, no idea... this is breaking change anyway we approach it.

Aaronontheweb:
https://github.com/akkadotnet/akka.net/blame/dev/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs#L815 <-- so it looks like for at least the past 5 years, we've been recording both the manifest and the serializer id in their own columns in all SQL persistence implementations. I think this means we can safely work around this default to some extent, but it will require making some backwards compatibility tests for the journal itself.

Basically, we need to try loading the explicit serializer by its ID first during reads - and on writes we should just use whatever the default `object` serializer is if there's not a specific match defined. If we can still read previous reads compatibly then this change is much less risky to implement.

cc @ismaelhamed

Aaronontheweb:
So what would a backwards compat spec look like? Using SQLite as an example:

1. Write events using the current config, commit SQLite file to repository (this is key - can't produce on the fly);
2. Change the configuration so there's no more default fallback serializer for persistence implementations - update the read event to use the serializer ID explicitly;
3. Verify that all of the expected events were read and recovered correctly.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -780,7 +780,7 @@ protected DbCommand GetCommand(DbConnection connection, string sql)
         protected virtual void WriteEvent(DbCommand command, IPersistentRepresentation e, IImmutableSet<string> tags)
         {
             
-            var serializer = Serialization.FindSerializerForType(e.Payload.GetType(), Configuration.DefaultSerializer);
```
Aaronontheweb, line 783: LGTM

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -846,7 +846,7 @@ protected virtual IPersistentRepresentation ReadEvent(DbDataReader reader)
             {
                 // Support old writes that did not set the serializer id
                 var type = Type.GetType(manifest, true);
-                var deserializer = Serialization.FindSerializerForType(type, Configuration.DefaultSerializer);
```
Aaronontheweb, line 849: so we'd need to keep this configuration setting here for the rare cases where the data is really old and never had a serialzier ID in-place - we should revert this change specifically.

Aaronontheweb: TL;DR; this is probably not the hotpath for reads on 99.999999% of events in Akka.Persistence.Sql

Aaronontheweb (changes requested):

ismaelhamed:
Don't forget this also applies to the `BatchingJournal`.

So, what happens now for users that have set a custom serializer for persistence that is different from the default serializer?

```
actor {
    serializers {
        hyperion = "Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion"
    }
    serialization-bindings {
        "System.Object" = hyperion
    }
}

akka.persistence.snapshot-store.sqlite.serializer = mySerializer
```

Aaronontheweb:
> So, what happens now for users that have set a custom serializer for persistence that is different from the default serializer?

New messages will be serialized using your `object` serializer, old messages that were previously serialized (incorrectly) with JSON are still deserialized with the JSON serializer. It's just a write-side change if we do it the way it's laid out in the review comments.

ismaelhamed:
I may be missing something, but with my example above, does it mean that messages that were being persisted with `msgpack` will now be serialized with `hyperion`?

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
to11mtm: This Do we have a backward compat spec around this specific condition? My concern is that if someone had a different default serializer defined, will we be using the correct one here?

Arkatufus:
No, this is a very specific bug where if there's any types that does not have any serializer mapping, the persistence plugin will use the serializer with the name "json", regardless of which serializer were set to handle the `System.Object` serialization.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
ismaelhamed: Exactly my question

Arkatufus:
This _is_ a breaking change for those people that sets their `akka.persistence.journal.{persistence_plugin}.serializer` setting in their HOCON if the serializer is not the same as the one being mapped to `System.Object`.

Arkatufus:
This is how very old legacy serialized data would look like. Notice that the `serializer_id` column is `DbNull`. In this very specific case, the `QueryExecutor.ReadEvent` will be forced to read the data using the serializer defined in `akka.persistence.{persistence_plugin}.journal.serializer` setting.

Arkatufus:
OK, looking at the history of why this setting is introduced in the first place, I'm in two minds about this.
This is the original reason: https://github.com/akkadotnet/akka.net/issues/2743#issuecomment-314790660

Seems like removing this feature would introduce a bigger problem than it would solve.

ismaelhamed:
> This _is_ a "breaking" change for those people that sets their `akka.persistence.journal.{persistence_plugin}.serializer` setting in their HOCON _if_ the serializer is not the same as the one being mapped to `System.Object` 

This is clear for me now, thanks @Arkatufus.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -234,6 +234,7 @@ public abstract class BatchingSqlJournalSetup
         /// <summary>
         /// The default serializer used when not type override matching is found
         /// </summary>
+        [Obsolete(message: "This property should never be used for writes, use the default `System.Object` serializer instead")]
```
Aaronontheweb, line 237: LGTM

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1325,7 +1326,7 @@ private async Task<WriteMessagesResult> HandleWriteMessages(WriteMessages req, T
         protected virtual void WriteEvent(TCommand command, IPersistentRepresentation persistent, string tags = "")
         {
             var payloadType = persistent.Payload.GetType();
-            var serializer = _serialization.FindSerializerForType(payloadType, Setup.DefaultSerializer);
```
Aaronontheweb, line 1328: LGTM - this is really the key fix here.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
Aaronontheweb: So we'll need to find which versions of Akka.Persistence didn't include a serializerId.


It looks like we first started adding serializerIds to Akka.Persistence.Sql.Common back in 1.3.1: https://github.com/akkadotnet/akka.net/releases/tag/v1.3.1

So, we should create a Sqlite database that runs on version 1.3.0 or earlier, writes some data, commit the binary to this repo, and then replay it using the latest development version of Akka.Persistence.Sql.Common - if we can still recover that data with this change, we're in good shape.

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
```diff
@@ -224,22 +224,5 @@ protected override async Task DeleteAsync(string persistenceId, SnapshotSelectio
                 await QueryExecutor.DeleteBatchAsync(connection, nestedCancellationTokenSource.Token, persistenceId, criteria.MaxSequenceNr, criteria.MaxTimeStamp);
             }
         }
-        
-        private SnapshotEntry ToSnapshotEntry(SnapshotMetadata metadata, object snapshot)
```
Aaronontheweb, line 228: Why was this removed?

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/CustomObjectSerializerSpec.cs
Aaronontheweb: Could you add a comment explaining what this test actually tests - like, what use case is it here to assert? Just need a snippet at the top explaining this.

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/CustomObjectSerializerSpec.cs
```diff
@@ -0,0 +1,194 @@
...
+            await using var cmd = new SqliteCommand(sql, conn);
+            var record = await cmd.ExecuteReaderAsync();
+            await record.ReadAsync();
+            
+            // In the bug this fails, the serializer id is JSON id instead of MySerializer id
+            record[0].Should().Be(9999);
+        }
+
+        [Fact(DisplayName = "Persistence.Sql should be able to read legacy data")]
+        public void LegacyDataTest()
```
Aaronontheweb, line 92: So to make this cover the issue referenced here: https://github.com/akkadotnet/akka.net/pull/6223/files#r1012961321 - we just need even older data, correct?

Aaronontheweb (changes requested):
Overall looks good, but we do need to harden the backwards compat test and add some additional comments for future committers to understand the significance of these tests - most of this backwards compatibility data is very "inside baseball."

Aaronontheweb:
Also, you'll need to add an entry to the upgrade advisories page for this change too @Arkatufus

Arkatufus:
Closing this issue, moving the changes to `dev` branch instead of `v1.4`.
PR moved to #6528

PR #6247, opened 2022-11-10
Upgrade to latest .NET 6 SDK on build agent

Aaronontheweb:
Upgrading to newest .NET 6 SDK for builds


Aaronontheweb:
Bleh, looks like DocFx doesn't like it - might need to upgrade that while we're at it.

Aaronontheweb:
Deprecated in favor of https://github.com/akkadotnet/akka.net/pull/6263

PR #6250, opened 2022-11-14
Make transport adapter messages public

Arkatufus:
Prerequisite for https://github.com/petabridge/petabridge.cmd/pull/587

## Changes
Make transport adapter messages public so it can be used in Petabridge.Cmd for testing network disruptions


src/core/Akka.API.Tests/CoreAPISpec.ApproveRemote.verified.txt
```diff
@@ -580,6 +580,58 @@ namespace Akka.Remote.Transport
         public FailureInjectorProvider() { }
         public Akka.Remote.Transport.Transport Create(Akka.Remote.Transport.Transport wrappedTransport, Akka.Actor.ExtendedActorSystem system) { }
     }
+    public class FailureInjectorTransportAdapter : Akka.Remote.Transport.AbstractTransportAdapter, Akka.Remote.Transport.IAssociationEventListener
```
Aaronontheweb, line 583: LGTM - this is sufficiently obscure that it's not going to regularly pop-up in end-user code, but still accessible to an extent that tooling like our planned "failure injection" feature for Petabridge.Cmd can still work.

Aaronontheweb (approved):
LGTM

PR #6254, opened 2022-11-15
Add CodeQL vulnerability scanning

Aaronontheweb:
## Changes

Akka.NET is a big project with a large installbase. It's time we add some security scanning tools to our suite. We'll see how CodeQL performs.


Aaronontheweb:
We're going to need to turn this off for PR validation and only run it as a chron job I think. 27 minutes to "autobuild" the project? Woof.

PR #6257, opened 2022-11-15
Update cluster SBR documentation on lease and how it works

Arkatufus:
## Changes
- Update SBR documentation

### DIFF
--- a/docs/articles/clustering/split-brain-resolver.md
+++ b/docs/articles/clustering/split-brain-resolver.md
@@ -285,7 +285,7 @@ As a precaution for that scenario, all nodes are downed if no decision can be ma
 
 This behavior is enabled by default for all new SBR strategies and by default the duration is derived as 3/4 of `stable-after` duration, but never less than 4 seconds.
 
-This duration can be overriden by changing the HOCON setting to a specific duration value or turned off by setting the HOCON setting to `off`:
+This duration can be overridden by changing the HOCON setting to a specific duration value or turned off by setting the HOCON setting to `off`:
 
 ```hocon
 akka.cluster.split-brain-resolver {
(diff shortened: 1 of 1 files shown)


docs/articles/clustering/split-brain-resolver.md
```diff
@@ -262,10 +262,60 @@ akka.cluster.split-brain-resolver.lease-majority {
 }
 ```
 
-A `Lease` is a type of distributed lock implementation.
+A `Lease` is a type of distributed lock implementation to decide which nodes are allowed to survive. Only one SBR instance can acquire the lease to make the decision to remain up. The other side will not be able to acquire the lease and will therefore down itself.
```
Aaronontheweb, line 265: `Lease` can also be used for other things like Akka.Cluster.Sharding and ClusterSingleton hand-over too - so I would add " in on the context of SBRs, the leader who acquires the `Lease` gets to make downing decisions for the entire cluster"

docs/articles/clustering/split-brain-resolver.md
```diff
@@ -262,10 +262,60 @@ akka.cluster.split-brain-resolver.lease-majority {
...
+    stable-after = 20s
+}
+```
+
+> [!WARNING]
+> It is recommended to keep `down-all-when-unstable` enabled and not to set it to be longer than `stable-after` because that can cause a delayed SBR decision on partition sides that should be downed, e.g. in the case of a clean network partition followed by continued instability on the side that should be downed. This can result in members being removed in one side but kept on running on the other side.
+
+### How Split Brain Resolver Works
+
+* The SBR actor subscribes itself to all cluster events from the event bus during actor `PreStart`.
```
Aaronontheweb, line 302: I'll see about turning this into a diagram in a separate PR

Aaronontheweb (approved):
LGTM but need to make the markdown linter happy

PR #6262, opened 2022-11-23
Akka.Streams: `ReuseLatest` stage to repeatedly emit the most recent value until a newer one is pushed

Aaronontheweb:
## Changes

Adds a new `RepeatPrevious` `Flow<TOut>` stage to Akka.Streams, designed for use with fan-in stages where one producer might only emit output intermittently - such as https://github.com/Aaronontheweb/AkkaStreams.Demo.HttpClient (with the `HttpClient` production stream.)

## Checklist


docs/articles/streams/buffersandworkingwithrate.md
```diff
@@ -189,3 +189,32 @@ var driftFlow = Flow.Create<int>()
 ```
 
 Note that all of the elements coming from upstream will go through expand at least once. This means that the output of this flow is going to report a drift of zero if producer is fast enough, or a larger drift otherwise.
+
+### Repeating Previous Values Downstream
```
Aaronontheweb, line 193: Wanted to add a lengthier article about `RepeatPrevious` in the context of different sources working at decoupled rates, which is why I added this stage in this first place.

src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.verified.txt
```diff
@@ -1383,6 +1383,8 @@ namespace Akka.Streams.Dsl
         [System.ObsoleteAttribute("Use RecoverWithRetries instead. [1.1.2]")]
         public static Akka.Streams.Dsl.Flow<TIn, TOut, TMat> RecoverWith<TIn, TOut, TMat>(this Akka.Streams.Dsl.Flow<TIn, TOut, TMat> flow, System.Func<System.Exception, Akka.Streams.IGraph<Akka.Streams.SourceShape<TOut>, TMat>> partialFunc) { }
         public static Akka.Streams.Dsl.Flow<TIn, TOut, TMat> RecoverWithRetries<TIn, TOut, TMat>(this Akka.Streams.Dsl.Flow<TIn, TOut, TMat> flow, System.Func<System.Exception, Akka.Streams.IGraph<Akka.Streams.SourceShape<TOut>, TMat>> partialFunc, int attempts) { }
+        public static Akka.Streams.Dsl.Flow<TIn, TOut, TMat> RepeatPrevious<TIn, TOut, TMat>(this Akka.Streams.Dsl.Flow<TIn, TOut, TMat> flow) { }
```
Aaronontheweb, line 1386: Flow DSL changes

src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.verified.txt
```diff
@@ -1791,6 +1793,14 @@ namespace Akka.Streams.Dsl
         public Pulse(System.TimeSpan interval, bool initiallyOpen = False) { }
         protected override Akka.Streams.Stage.GraphStageLogic CreateLogic(Akka.Streams.Attributes inheritedAttributes) { }
     }
+    public sealed class RepeatPrevious<T> : Akka.Streams.Stage.GraphStage<Akka.Streams.FlowShape<T, T>>
```
Aaronontheweb, line 1796: `RepeatPrevious<T>` API implementation

src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.verified.txt
```diff
@@ -2059,6 +2069,8 @@ namespace Akka.Streams.Dsl
         [System.ObsoleteAttribute("Use RecoverWithRetries instead. [1.1.2]")]
         public static Akka.Streams.Dsl.Source<TOut, TMat> RecoverWith<TOut, TMat>(this Akka.Streams.Dsl.Source<TOut, TMat> flow, System.Func<System.Exception, Akka.Streams.IGraph<Akka.Streams.SourceShape<TOut>, TMat>> partialFunc) { }
         public static Akka.Streams.Dsl.Source<TOut, TMat> RecoverWithRetries<TOut, TMat>(this Akka.Streams.Dsl.Source<TOut, TMat> flow, System.Func<System.Exception, Akka.Streams.IGraph<Akka.Streams.SourceShape<TOut>, TMat>> partialFunc, int attempts) { }
+        public static Akka.Streams.Dsl.Source<T, TMat> RepeatPrevious<T, TMat>(this Akka.Streams.Dsl.Source<T, TMat> source) { }
```
Aaronontheweb, line 2072: Source DSL changes

src/core/Akka.API.Tests/CoreAPISpec.ApproveStreams.verified.txt
```diff
@@ -2280,6 +2292,7 @@ namespace Akka.Streams.Dsl
         public abstract Akka.Streams.Dsl.IFlow<T2, TMat> Via<T2, TMat2>(Akka.Streams.IGraph<Akka.Streams.FlowShape<TOut, T2>, TMat2> flow);
         public abstract Akka.Streams.Dsl.IFlow<T2, TMat3> ViaMaterialized<T2, TMat2, TMat3>(Akka.Streams.IGraph<Akka.Streams.FlowShape<TOut, T2>, TMat2> flow, System.Func<TMat, TMat2, TMat3> combine);
     }
+    public delegate void SwapPrevious<T>(T previousValue, T newValue);
```
Aaronontheweb, line 2295: I created an explicit delegate type here in order to make it clear which value was the previous and which one was the new, since I personally had some confusion with it while I was working on the sample.

src/core/Akka.Streams.Tests/Dsl/RepeatPreviousSpec.cs
```diff
@@ -0,0 +1,84 @@
...
+            var (queue, source) = Source.Queue<int>(10, OverflowStrategy.Backpressure).PreMaterialize(Materializer);
+            
+            // populate 1 into queue
+            await queue.OfferAsync(1);
+
+            // take 4 items from the queue
+            var result = await source.RepeatPrevious().Take(4).RunWith(Sink.Seq<int>(), Materializer);
+            
+            // the most recent queue item will be repeated 3 times, plus the original element
+            result.Should().BeEquivalentTo(1,1,1,1);
```
Aaronontheweb, line 80: The downstream ` Take` stage will grab 4 elements from the queue almost immediately. Even if I tried to `queue.OfferAsync` a second value the amount of time it takes for those values to propagate would be greater than how quickly this stage could execute. This is by design - `RepeatPrevious` totally decouples the rate of production from the rate of consumption.

src/core/Akka.Streams.Tests/Dsl/RepeatPreviousSpec.cs
```diff
@@ -0,0 +1,84 @@
...
+        private ActorMaterializer Materializer { get; }
+
+        public RepeatPreviousSpec(ITestOutputHelper testOutputHelper) : base(Config.Empty, output: testOutputHelper)
+        {
+            var settings = ActorMaterializerSettings.Create(Sys);
+            Materializer = ActorMaterializer.Create(Sys, settings);
+        }
+        
+        [Fact]
+        public async Task RepeatPrevious_should_immediately_terminate_with_Empty_source()
```
Aaronontheweb, line 30: Instant completion test case.

src/core/Akka.Streams.Tests/Dsl/RepeatPreviousSpec.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+        
+        [Fact]
+        public async Task RepeatPrevious_should_complete_when_upstream_completes()
+        {
+            var source = Source.Single(1).RepeatPrevious();
+            var result = await source.RunWith(Sink.Seq<int>(), Materializer);
+
+            // as a side-effect of RepeatPrevious' buffering process, there's going to be an extra element in the result
+            result.Should().BeEquivalentTo(1, 1);
```
Aaronontheweb, line 44: Unfortunately, and I spent a couple of hours trying to fix this - we're going to get an extra event due to the work involved in detecting that the upstream has completed. Comes with the territory if you're separating upstream from downstream in the manner that this stage does.

src/core/Akka.Streams.Tests/Dsl/RepeatPreviousSpec.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        [Fact]
+        public async Task RepeatPrevious_should_fail_when_upstream_fails()
```
Aaronontheweb, line 48: Tests to validate that failure was propagated.

src/core/Akka.Streams/Attributes.cs
```diff
@@ -301,7 +301,7 @@ public Attributes(params IAttribute[] attributes)
         /// INTERNAL API
         /// </summary>
         internal bool IsAsync
-            => _attributes.Count() > 0 &&
+            => _attributes.Length > 0 &&
```
Aaronontheweb, line 304: Performance optimization / syntax clean up

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
```diff
@@ -0,0 +1,105 @@
...
+
+                SetHandler(_stage._in, this);
+                SetHandler(_stage._out, this);
+            }
+
+            public override void OnPush()
+            {
+                var next = Grab(_stage._in);
+                if (_last.HasValue)
+                    _swapPrevious(_last.Value, next);
```
Aaronontheweb, line 73: Swap function is only needed for scenarios where removing the previous value requires some cleanup - i.e. in my original usecase I needed to call `Dispose` on an `HttpClient`

Aaronontheweb (commented):
Described all my changes

Aaronontheweb:
In order to get this to pass I'm going to need to upgrade the build system to use .NET 7 - otherwise the extra .DLL metadata is going to cause conflicts with the build system.

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
to11mtm: My concern here is that there may or may not be some implicit buffering concerns here.

Is there a risk that either:

 1. There are 'extra' instances of the item getting disposed in a buffer in-between stages
 2. This stage is processing input while the downstream stage is trying to use the last value?

My knowledge of akka streams stage runs is not the best, but I know there are cases where internal buffers can lead to surprising results.

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        public RepeatPrevious(SwapPrevious<T> swapPrevious)
+        {
+            _swapPrevious = swapPrevious;
+        }
+
+        protected override GraphStageLogic CreateLogic(Attributes inheritedAttributes) =>
+            new Logic(this, _swapPrevious);
+
+        private sealed class Logic : InAndOutGraphStageLogic
```
to11mtm, line 54: I will say that overall this reminds me of a version of `Expand` with a bit more sane of an API. I'm not sure if that means we can de-dupe any logic between the two however (Expand is... weird to me.)

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    Push(_stage._out, _last.Value);
+                }
+            }
+
+            public override void OnPull()
+            {
+                if (_last.HasValue)
+                {
+                    if (!HasBeenPulled(_stage._in))
```
to11mtm, line 86: To your comments about repeat elements in the test, I wonder if guard code in pull would help. Expand does something like this:

```
if (_iterator.HasNext())
                {
                    if (!_expanded)
                    {
                        _expanded = true;
                        if (IsClosed(_stage.In))
                        {
                            Push(_stage.Out, _iterator.Next());
                            CompleteStage();
                        }
                        else
                        {
                            // expand needs to pull first to be "fair" when upstream is not actually slow
                            Pull(_stage.In);
                            Push(_stage.Out, _iterator.Next());
                        }
                    }
                    else
                        Push(_stage.Out, _iterator.Next());
                }
```

And I wonder if that IsClosed guard would help?

src/core/Akka.Streams.Tests/Dsl/RepeatPreviousSpec.cs
to11mtm: This comment may help: https://github.com/akkadotnet/akka.net/pull/6262#discussion_r1031828989

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace Akka.Streams.Dsl
+{
+    /// <summary>
+    /// Delegate used to customize replacement functions.
+    ///
+    /// This is used for things like calling <see cref="IDisposable.Dispose"/> on the previous element.
+    /// when a new value is emitted.
+    /// </summary>
+    /// <typeparam name="T">The type of element handled by the <see cref="RepeatPrevious{T}"/></typeparam>
+    public delegate void SwapPrevious<T>(T previousValue, T newValue);
```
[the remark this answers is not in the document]
Aaronontheweb: That's a good idea - I'm not a fan of the custom delegate either but it was confusing as hell trying to make sense of which is which without naming the parameters.

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
Aaronontheweb: I'll take a look at this! I tried coming up with something similar to this but I just kept moving the problem somewhere else in the test suite. Let me give this a look though.

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
Aaronontheweb: I had no idea about `Expand` TBH - part of why I did what I did in the documentation with this stage is that I think our overview of built-in stages is pretty weak and not easy to grok.

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
Aaronontheweb: Yeah I tried this again and it's not going to work - the issue is that in order to detect that the upstream is closed it has to be pulled a second time and during that timeframe we're going to emit an extra item.

The `Expand` specs don't really check for the case where the upstream terminates - they're mostly just testing that expansion doesn't happen so long as downstream can keep up with demand.

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            public override void OnPush()
+            {
+                var next = Grab(_stage._in);
+                if (_last.HasValue)
+                    _onItemChanged(_last.Value, next);
+                _last = next;
+
+                if (IsAvailable(_stage._out))
```
Arkatufus, line 67: Since we're not buffering, if the `RepeatPrevious` stage is actually faster than its companion(s), would that mean that we'll lose events/data?

Arkatufus: We might want to document this, just in case.

Arkatufus (commented):
I think the stage name is a bit misleading since we're not buffering anything. Maybe `RepeatLatest` or `RetainLatest` be better?

src/core/Akka.Streams/Dsl/RepeatPrevious.cs
Aaronontheweb: We won't lose any events / data - we always emit the current element as it's being discovered and we'll repeatedly emit it downstream when a newer value hasn't been pushed.

Aaronontheweb:
> 

The name is `RepeatPrevious` - where is buffering implied there?

Arkatufus:
Its just a bit ambiguous, its not clear what "previous" is. 
Lets say we're zipping alphabets and numbers, alphabets being repeated.
1. 'a' pushed
2. '1' pushed, zipped as ('a', 1')
3. 'b' pushed
4. 'c' pushed
5. '2' pushed, would the user expect ('b', 2) or ('c', 2)

"Latest" is a bit more clear that the last received event got zipped

Aaronontheweb:
The user would receive (`c`,2) in that scenario

Arkatufus (approved):
LGTM, just need to remove the benchmark artifact files.

Aaronontheweb:
> LGTM, just need to remove the benchmark artifact files.

Done, sorry about that :(

Aaronontheweb:
This needs a port to v1.5 also, along with several other PRs that have been made to the v1.4 branch.

src/core/Akka.Streams/Dsl/FlowOperations.cs
```diff
@@ -2438,5 +2438,42 @@ public static class FlowOperations
...
+        /// <typeparam name="TIn">The input type.</typeparam>
+        /// <typeparam name="TOut">The output type.</typeparam>
+        /// <typeparam name="TMat">The materialization type.</typeparam>
+        /// <remarks>
+        /// This is designed to allow fan-in stages where output from one of the sources is intermittent / infrequent
+        /// and users just want the previous value to be reused.
+        /// </remarks>
+        public static Flow<TIn, TOut, TMat> RepeatPrevious<TIn, TOut, TMat>(this Flow<TIn, TOut, TMat> flow, Action<TOut, TOut> onItemChanged)
+        {
+            return flow.Via(new ReuseLatest<TOut>(onItemChanged));
```
He-Pin, line 2476: Why not named as `RepeatLatest` just as the stage?

src/core/Akka.Streams/Dsl/ReuseLatest.cs
```diff
@@ -0,0 +1,96 @@
...
+                else
+                {
+                    Pull(_stage._in);
+                }
+            }
+        }
+
+        public override string ToString()
+        {
+            return "RepeatPrevious";
```
He-Pin, line 93: `ReuseLatest`?

He-Pin:
@Aaronontheweb Hi, as the reuseLatest can be implemented with `expand` ,so this new operator is added for performance?

src/core/Akka.Streams/Dsl/FlowOperations.cs
Aaronontheweb: I got feedback on name changes and didn't propagate it everywhere I guess, although the stage implementation is really meant to be internal.

src/core/Akka.Streams/Dsl/ReuseLatest.cs
Aaronontheweb: Yes, also needs to be renamed - this got missed during the review

Aaronontheweb:
ReuseLatest is meant to be its own thing - Expand is funky and allocates lists each time

PR #6268, opened 2022-11-26
Move Channel Stages from Alpakka to main project.

to11mtm:
## Changes

 - Moves Alpakka Channels Extension into the main project.
 - Appends DSLs for main `Source` and `Sink` to have Channel methods
 - Adds capability for a Channel Source with a materialized `ChannelWriter<T>` result
   - This feels more symmetrical with the Sink API that allows both options, and allows for a better option-in-place-of `Source.Queue<T>`. 

The main motivations for this are as follows:

 - We already take `System.Threading.Channels` as a dep in core Akka
 - We can probably get rid of the ugly backing code behind `RunAsAsyncEnumerable`, even if it is a minorly-breaking API change it may be better long term.

## Checklist


src/core/Akka.Streams/Implementation/ChannelSources.cs
```diff
@@ -0,0 +1,155 @@
...
+                if (continuation.IsCompletedSuccessfully)
+                {
+                    var dataAvailable = continuation.GetAwaiter().GetResult();
+                    if (dataAvailable && _reader.TryRead(out element))
+                        Push(_outlet, element);
+                    else
+                        CompleteStage();
+                }
+                else
+                    continuation.AsTask().ContinueWith(_onReadReady);
```
to11mtm, line 81: This is a slight change from the Alpakka code:

 - The Alpakka code [looked a little racy with how it was checking flags](https://github.com/akkadotnet/Alpakka/blob/ca6ff62f664bee21f764efb18de42f459c490a77/src/Channels/Akka.Streams.Channels/Internal/ChannelReaderSource.cs#L74-L92).
 - The Failure/Completion stages are less perf sensitive than the main read loop, so the code on our happy path is now smaller

src/core/Akka.Streams/Implementation/ChannelSources.cs
```diff
@@ -0,0 +1,155 @@
...
+        public ChannelSourceLogic(SourceShape<T> source, Outlet<T> outlet,
+            ChannelReader<T> reader) : base(source)
+        {
+            _outlet = outlet;
+            _reader = reader;
+            _onValueRead = GetAsyncCallback<bool>(OnValueRead);
+            _onValueReadFailure =
+                GetAsyncCallback<Exception>(OnValueReadFailure);
+            _onReaderComplete = GetAsyncCallback<Exception>(OnReaderComplete);
+            _onReadReady = ContinueAsyncRead;
```
to11mtm, line 33: This change is here to avoid delegates potentially being made/initted repeatedly in OnPull.

src/core/Akka.Streams/Dsl/ChannelSink.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Creates a sink that upon materialization, returns a <see cref="ChannelReader{T}"/> connected with
+        /// this materialized graph. It can then be used to consume events incoming from the graph. It will
+        /// also be completed once upstream completes.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="bufferSize"></param>
+        /// <param name="singleReader"></param>
+        /// <param name="fullMode"></param>
+        /// <returns></returns>
+        public static Sink<T, ChannelReader<T>> AsReader<T>(int bufferSize, bool singleReader = false, BoundedChannelFullMode fullMode = BoundedChannelFullMode.Wait) =>
```
to11mtm, line 44: IDK if we want to keep these, I moved them over for the sake of migration.

src/core/Akka.Streams/Dsl/ChannelSource.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <summary>
+        /// Creates an Akka.Streams <see cref="Source{TOut,TMat}"/> from a given instance of
+        /// <see cref="ChannelReader{T}"/>. It will propagate backpressure from the downstream
+        /// to guarantee resource-safe communication as well as will react when <paramref name="reader"/>
+        /// will complete (successfully or with failure) and finish downstream accordingly.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="reader"></param>
+        /// <returns></returns>
+        public static Source<T, NotUsed> FromReader<T>(ChannelReader<T> reader)
```
to11mtm, line 30: IDK if we want to keep these, I moved them over for the sake of migration.

src/core/Akka.Streams/Dsl/ChannelSource.cs
```diff
@@ -0,0 +1,51 @@
...
+                ThrowArgumentNullException("reader");
+
+            return Source.FromGraph(new ChannelReaderSource<T>(reader));
+        }
+
+        public static Source<T, ChannelWriter<T>> Create<T>(int bufferSize,
+            bool singleWriter = false,
+            BoundedChannelFullMode fullMode = BoundedChannelFullMode.Wait)
+        {
+            return Source.FromGraph(
```
to11mtm, line 42: I added this for the sake of symmetry with the Sinks, as well as it's amazing utility in place of `Source.Queue<T>`

to11mtm (commented):
Left my own comments here.

Please let me know if files should be moved/deleted/etc.

src/core/Akka.Streams/Implementation/ChannelSinks.cs
```diff
@@ -0,0 +1,170 @@
...
+        public override void OnUpstreamFailure(Exception e)
+        {
+            base.OnUpstreamFailure(e);
+            if (_isOwner)
+                _writer.TryComplete(e);
+        }
+
+        public override void PreStart()
+        {
+            Pull(_inlet);
```
Aaronontheweb, line 77: Don't think this is necessary

src/core/Akka.Streams/Implementation/ChannelSinks.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (_writer.TryWrite(element))
+            {
+                Pull(_inlet);
+            }
+            else
+            {
+                var continuation = _writer.WaitToWriteAsync();
+                if (continuation.IsCompletedSuccessfully)
+                {
+                    var available = continuation.GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
Aaronontheweb: But I don't really see a way around it in this case either.

src/core/Akka.Streams/Implementation/ChannelSinks.cs
```diff
@@ -0,0 +1,170 @@
...
+    internal sealed class ChannelReaderSink<T> : GraphStageWithMaterializedValue<SinkShape<T>, ChannelReader<T>>
+    {
+        private readonly int _bufferSize;
+        private readonly bool _singleReader;
+        private readonly BoundedChannelFullMode _fullMode;
+
+        public ChannelReaderSink(int bufferSize, bool singleReader = false, BoundedChannelFullMode fullMode = BoundedChannelFullMode.Wait)
+        {
+            _bufferSize = bufferSize;
+            _singleReader = singleReader;
```
Aaronontheweb, line 131: Shouldn't `singleReader` be `true` by default?

src/core/Akka.Streams/Implementation/ChannelSources.cs
Aaronontheweb: LGTM

src/core/Akka.Streams/Implementation/ChannelSources.cs
```diff
@@ -0,0 +1,155 @@
...
+            if (_reader.TryRead(out var element))
+            {
+                Push(_outlet, element);
+            }
+            else
+            {
+                var continuation = _reader.WaitToReadAsync();
+                if (continuation.IsCompletedSuccessfully)
+                {
+                    var dataAvailable = continuation.GetAwaiter().GetResult();
```
Aaronontheweb, line 74: Any way we can use async callbacks here?

Aaronontheweb (commented):
Left some thoughts

src/core/Akka.Streams/Implementation/ChannelSinks.cs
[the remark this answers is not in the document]
to11mtm: Well, if the continuation is completed successfully, this is going to be non-blocking. It's confusing but legal AFAIK (FWIW this is what the original was more or less doing as well.)

src/core/Akka.Streams/Implementation/ChannelSinks.cs
```diff
@@ -0,0 +1,170 @@
...
+                        Pull(_inlet);
+                    }
+                    else
+                    {
+                        TryComplete();
+                    }
+                }
+                else
+                {
+                    var task = continuation.AsTask();
```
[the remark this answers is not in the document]
to11mtm: I thought about it, but getting it to have the same semantics was a bit too bold for my taste (I wanted to keep this as close as possible to the original code for reference purposes.)

src/core/Akka.Streams/Implementation/ChannelSinks.cs
to11mtm: The Pull on starting the sink was added as part of https://github.com/akkadotnet/Alpakka/pull/1068/ . Maybe there's a different way to fix that but without it you can't really use a sink properly.

src/core/Akka.Streams/Implementation/ChannelSinks.cs
to11mtm: This was 'as-found' but `singleReader = false` here is the safer default so I didn't think it was worth changing.

src/core/Akka.Streams/Implementation/ChannelSources.cs
to11mtm: I would need a bit more context to answer that?

AFAIK this code path is present to specifically avoid `GetAsyncCallBack` and being put on scheduler.

Aaronontheweb: Got it.

src/core/Akka.Streams/Implementation/ChannelSinks.cs
to11mtm: FWIW, `BoundedChannel` AFAIK doesn't even use `ChannelOptions.SingleReader`. it's only used in `UnboundedChannel`

src/core/Akka.Streams/Implementation/ChannelSinks.cs
Aaronontheweb: Ok, no worries - we'll keep it. Can always update it later.

src/core/Akka.Streams/Implementation/ChannelSinks.cs
Aaronontheweb: And we don't use `UnboundedChannel`, so I guess this setting doesn't really matter then?

Aaronontheweb:
We can add website documentation later, but I think this looks good - can you run API approvals @to11mtm ?

Aaronontheweb (approved):
LGTM but needs API approvals

src/core/Akka.Streams/Implementation/ChannelSinks.cs
to11mtm: > And we don't use `UnboundedChannel`, so I guess this setting doesn't really matter then?

Not _yet_ anyway. The challenge of course is that it may become relevant in the future.

Right now, `BoundedChannel<T>` uses a `Dequeue<T>` internally, so it uses a single lock for both readers/writers. In the future however, the internal structure may change (e.x. something more like `ConcurrentQueue<T>` with segments) and then it may become important.

Really it's a question of whether it is an oversight in `ChannelOptions` that it is exposed for both types.

Aaronontheweb:
I think I have the API approvals working here now

Aaronontheweb:
Looks like one of the new versions of Verify we started using might be causing problems with API approval tests.

PR #6281, opened 2022-11-30
Add ReceiveAsync feature to Akka.TestKit TestActorRef

### LINKED ISSUE #6265: `TestActorRef` can not catch exceptions on asynchronous methods
**Version Information**
Version of Akka.NET? `Akka 1.4.46`
Which Akka.NET Modules? `Akka.TestKit 1.4.46`, `xunit 2.4.1`

**Describe the bug**
I am attempting to test the behavior of Actors. I am not very good at it yet, so every once in a while there is an exception, and the test just swallows the exception, and I have to go search why `ExpectMsg` is timing out.  

I have found the best way to test the actor behavior is with `TestActorRef.Receive`, but it only works for synchronous methods.

**To Reproduce**
Test actor:
```C#
public class ExceptionActor : ReceiveActor
{
    public record GiveError();
    public record GiveErrorAsync();
    public ExceptionActor()
    {
        Receive<GiveError>((b) =>
        {
            throw new Exception("WAT");
        });

        ReceiveAsync<GiveErrorAsync>(async (b) =>
        {
            await Task.Delay(TimeSpan.FromSeconds(0.1));
            throw new Exception("WATASYNC");
        });
    }
}
```

This test fails (Expected behavior):
```C#
    [Fact]
    public void GetException()
    {
        var props = Props.Create<ExceptionActor>();
        var subject = new TestActorRef<ExceptionActor>(Sys, props, null, "testA");
        subject.Receive(new ExceptionActor.GiveError());
    }
```

This test does not fail (Not expected behavior): 
```C#
    [Fact]
    public void GetExceptionAsync()
    {
        var props = Props.Create<ExceptionActor>();
        var subject = new TestActorRef<ExceptionActor>(Sys, props, null, "testB");
        subject.Receive(new ExceptionActor.GiveErrorAsync());
    }
```

**Expected behavior**
To have a `ReceiveAsync` method for asynchronous method handling. 
OR 
Some mechanism that automatically watches for any unhandled exceptions in tests, and have the test failure display the information.

**Actual behavior**
Unhandled exceptions disappear.

**Environment**
.NET Core v7 on Windows

**Additional context**
I am fairly new to akka.net, but I am loving it. I am following the instructions on [Testing Actor Systems](https://getakka.net/articles/actors/testing-actor-systems.html), and it feels like I am missing something. Are there no better ways to watch that unhandled exceptions exist?

Arkatufus:
Fixes #6265

## Changes

- Add ReceiveAsync feature to Akka.TestKit TestActorRef
- Add IAsyncResultInterceptor interface to tap directly into ReceiveAsync execution result


src/core/Akka/Dispatch/ActorTaskScheduler.cs
```diff
@@ -158,18 +158,21 @@ public static void RunTask(Func<Task> asyncAction)
...
-
                                       context.CheckReceiveTimeout();
                                   }
                                   else
                                   {
                                       context.Self.AsInstanceOf<IInternalActorRef>().SendSystemMessage(new ActorTaskSchedulerMessage(exception, actorScheduler.CurrentMessage));
                                   }
+                                  
+                                  // Used by TestActorRef to intercept async execution result
+                                  if(actorScheduler is IAsyncResultInterceptor interceptor)
```
Aaronontheweb, line 173: I'd prefer not to do this on the `TaskScheduler` that all actors use, but I don't really see a way around it - this is probably the best call we can make in the short run, without redesigining the mailbox system to use `Task`.

src/core/Akka.TestKit/TestActorRefBase.cs
```diff
@@ -51,6 +52,22 @@ public void Receive(object message, IActorRef sender = null)
...
+        /// thrown will be available to you, while still being able to use
+        /// become/unbecome.
+        /// Note: This method violates the actor model and could cause unpredictable 
+        /// behavior. For example, a Receive call to an actor could run simultaneously 
+        /// (2 simultaneous threads running inside the actor) with the actor's handling 
+        /// of a previous Tell call. 
+        /// </summary>
+        /// <param name="message">The message.</param>
+        /// <param name="sender">The sender.</param>
+        public Task ReceiveAsync(object message, IActorRef sender = null)
```
Aaronontheweb, line 66: won't this require API approvals @Arkatufus ?

Aaronontheweb (approved):
LGTM but not sure if we need an API approval here

src/core/Akka.TestKit/TestActorRefBase.cs
Arkatufus: It doesn't look like it, at least not locally

src/core/Akka/Dispatch/ActorTaskScheduler.cs
Arkatufus: Yes, other alternatives would require a more radical change to implement, this is the shortest solution I can find without scrambling codes all over the place

Aaronontheweb:
Yep, did not require API approval

Issue #6279, opened 2022-11-29
Akka.Actor: `ReceiveAsync` loses track of `ActorContext` following an `await` on `WaitAsync`, `Sender` throws


Aaronontheweb:
**Version Information**
Version of Akka.NET? 1.4.46, 1.5.0-alpha3
Which Akka.NET Modules? Akka.Actor

**Describe the bug**

When using the following `ReceiveAsync` statement, the `Sender.Tell` crashes with an `System.NotSupportedException`:

```csharp
ReceiveAsync<SubscribeToEnvironment>(async cmd =>
{
    // NOTE: can't throw here - one subscription failing shouldn't blow up any of the others

    // create a ChannelReader, from which we'll create an IAsyncEnumerable in the service
    var (reader, sink) = ChannelSink.AsReader<IEnvironmentEvent>(100, true, BoundedChannelFullMode.DropOldest)
        .PreMaterialize(Context.Materializer());
    var (sinkRefTask, killSwitch) = StreamRefs.SinkRef<IEnvironmentEvent>()
        .ViaMaterialized(KillSwitches.Single<IEnvironmentEvent>(), Keep.Both)
        .Via(cmd.CancellationToken.AsFlow<IEnvironmentEvent>())
        .To(sink)
        .Run(Context.Materializer());

    try
    {
        using var cts = new CancellationTokenSource(_subscriptionTimeout);
        var sinkRef = await sinkRefTask.WaitAsync(cts.Token).ConfigureAwait(false);

        var resp = await _clusterStateActor.ExecuteCommand(
            new SubscriptionCommands.SubscribeToClusterEvents(_subscriberId, cmd.EnvironmentId, sinkRef),
            cts.Token).ConfigureAwait(false);
        if (resp.Status == ResponseStatus.Success)
        {
            Sender.Tell(new SubscribeSuccess(cmd.EnvironmentId, reader));
        }
        else // no-ops are failures here - no guarantee that the new stream is going to run
        {
            // deliver SubscribeFailed to the client
            Sender.Tell(new SubscribeFailed(cmd.EnvironmentId, resp.Message));
            // kill the stream
            killSwitch.Shutdown();
        }
    }
    catch (Exception ex)
    {
        _log.Error(ex, "Failed to create subscription for [EnvironmentId: {0}]", cmd.EnvironmentId);
        Sender.Tell(new SubscribeFailed(cmd.EnvironmentId, ex.Message));
        killSwitch.Abort(ex);
    }
});
[...]

PR #6287, opened 2022-12-01
Fix `ConfigureAwait` issues inside `Ask<T>` and `ReceiveAsync`

Aaronontheweb:
## Changes

Added reproduction for https://github.com/akkadotnet/akka.net/issues/6279

## Checklist

### DIFF
--- a/src/core/Akka.Tests/Actor/AskSpec.cs
+++ b/src/core/Akka.Tests/Actor/AskSpec.cs
@@ -10,19 +10,22 @@ using Xunit;
 using Akka.Actor;
 using Akka.Actor.Dsl;
 using System;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Util.Internal;
 using FluentAssertions;
 using Nito.AsyncEx;
 using Akka.Dispatch.SysMsg;
+using Akka.TestKit.TestActors;
+using Xunit.Abstractions;
 
 namespace Akka.Tests.Actor
 {
     public class AskSpec : AkkaSpec
     {
-        public AskSpec()
-            : base(@"akka.actor.ask-timeout = 3000ms")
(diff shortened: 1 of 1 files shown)


src/core/Akka/Actor/Futures.cs
```diff
@@ -117,8 +117,10 @@ public static Task<T> Ask<T>(this ICanTell self, object message, TimeSpan? timeo
         /// This exception is thrown if the system can't resolve the target provider.
         /// </exception>
         /// <returns>TBD</returns>
-        public static Task<T> Ask<T>(this ICanTell self, Func<IActorRef, object> messageFactory, TimeSpan? timeout, CancellationToken cancellationToken)
+        public static async Task<T> Ask<T>(this ICanTell self, Func<IActorRef, object> messageFactory, TimeSpan? timeout, CancellationToken cancellationToken)
         {
+            await SynchronizationContextManager.RemoveContext;
```
Aaronontheweb, line 122: Still fails even after I add this back - cc @to11mtm

Aaronontheweb:
Found the problem:

https://github.com/akkadotnet/akka.net/blob/5605d8303f3e901b297d9dface7efce51e90d27b/src/core/Akka/Dispatch/ActorTaskScheduler.cs#L157-L179

The continuation task never gets scheduled onto the actor's `DefaultTaskScheduler` - it still gets scheduled onto the default `TaskScheduler`. This might be a change in the way `TaskCompletionSource<T>(TaskCreationOptions.RunContinuationsAsynchronously)` behaves - because it previously used to not do that.

Aaronontheweb:
With `ConfigureAwait(true)` the `ActorTaskScheduler` is used - as expected.

Aaronontheweb:
```
return new TaskCompletionSource<T>(TaskCreationOptions.RunContinuationsAsynchronously); <-- removing Run Async made no difference
```

PR #6290, opened 2022-12-05
Akka:Streams Resolve `IAsyncEnumerator.DisposeAsync` bug

### LINKED ISSUE #6280: Akka.Streams: `System.NotSupportedException` when disposing stage with materialized `IAsyncEnumerable`
**Version Information**
Version of Akka.NET? v1.4.46, v1.5.0-alpha3

**Describe the bug**

When running a `Source<T>` built from an `IAsyncEnumerable<T>`, any attempt to dispose the `IAsyncEnumerable<T>` via its `DisposeAsync()` method results in the following exception being thrown:

```text
[WARNING][11/30/2022 4:36:06 PM][Thread 0008][LogSource (akka://OpsCenter)] Failed to dispose IAsyncEnumerator asynchronously
Cause: System.NotSupportedException: Specified method is not supported.
   at OpsCenter.ClusterState.Subscriptions.ClusterStreamManager.GetEnvironmentEventStream(String environmentId, ClusterStreamConfig config, CancellationToken cancellationToken)+System.IAsyncDisposable.DisposeAsync()
   at Akka.Streams.Implementation.Fusing.AsyncEnumerable`1.Logic.OnDownstreamFinish()
```

**To Reproduce**

I don't have a clean reproduction of this issue yet, but using the following in LinqPad with Akka.NET v1.4.46 installed:

```csharp
async Task Main()
{
		var actorSystem = ActorSystem.Create("MySys");

		using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(3));

		var (actorRef, src) = Source.ActorRef<int>(1000, OverflowStrategy.DropHead)
		.PreMaterialize(actorSystem.Materializer());
		
		Task.Run(() => {
			foreach(var i in Enumerable.Range(0, 100)){
				actorRef.Tell(i);
			}	
			//  shut down the stream
			//actorRef.Tell(PoisonPill.Instance);
		});


		await Source.From(() => src.RunAsAsyncEnumerable(actorSystem.Materializer())).RunForeach(i => Console.WriteLine($"{DateTime.UtcNow}: {i}"), actorSystem.Materializer());
}
```

I was able to trigger another bug with this stage:

```text
[ERROR][11/30/2022 9:02:21 PM][Thread 0082][[akka://MySys/user/StreamSupervisor-10/Flow-0-0-unknown-operation#943292100]] Error in stage [EnumerableSource]: Should never reach this code
Cause: System.InvalidOperationException: Should never reach this code
   at Akka.Streams.Implementation.Fusing.AsyncEnumerable`1.Logic.OnPull()
   at Akka.Streams.Implementation.Fusing.GraphInterpreter.Execute(Int32 eventLimit)
   at UserQuery.Main() in C:\Users\aaron\AppData
```

As for the error with this code - I'll try another reproduction but I think it happens when `IActorContext` used to materialize the `IAsyncEnumerable` source is terminated.

**Expected behavior**

`IAsyncEnumerable` stages should be able to shut down correctly.

**Actual behavior**

The `IAsyncEnumerable` stage errs in various ways, which aren't properly supported right now.

Aaronontheweb:
Fixes #6280 

## Changes

We no longer attempt to dispose `IAsyncEnumerator`s inside `Source` stages due to the reasons outlined here: https://github.com/akkadotnet/akka.net/issues/6280#issuecomment-1338156992

This prevents the log from being filled with `NotSupportedException` warnings from failed `DisposeAsync` operations.

## Checklist

* [x] Design discussion issue #6280

### DIFF
--- a/src/core/Akka.Streams.Tests/Akka.Streams.Tests.csproj
+++ b/src/core/Akka.Streams.Tests/Akka.Streams.Tests.csproj
@@ -5,6 +5,7 @@
   <PropertyGroup>
     <AssemblyName>Akka.Streams.Tests</AssemblyName>
     <TargetFrameworks>$(NetFrameworkTestVersion);$(NetTestVersion);$(NetCoreTestVersion)</TargetFrameworks>
+    <LangVersion>9</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
@@ -24,7 +24,7 @@ using FluentAssertions.Extensions;
 
 namespace Akka.Streams.Tests.Dsl
 {
-#if !NETFRAMEWORK
+#if !NETFRAMEWORK // disabling this causes .NET Framework 4.7.2 builds to fail on Linux
     public class AsyncEnumerableSpec : AkkaSpec
     {
         private ActorMaterializer Materializer { get; }
--- a/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
@@ -24,10 +24,12 @@ using FluentAssertions.Extensions;
 
 namespace Akka.Streams.Tests.Dsl
 {
+#if !NETFRAMEWORK
     public class AsyncEnumerableSpec : AkkaSpec
     {
         private ActorMaterializer Materializer { get; }
         private ITestOutputHelper _helper;
+
         public AsyncEnumerableSpec(ITestOutputHelper helper) : base(
             AkkaSpecConfig.WithFallback(StreamTestDefaultMailbox.DefaultConfig),
             helper)
@@ -45,7 +47,7 @@ namespace Akka.Streams.Tests.Dsl
 
             var cts = new CancellationTokenSource();
             var token = cts.Token;
-            
+
             var asyncEnumerable = Source.From(input).RunAsAsyncEnumerable(Materializer);
             var output = input.ToArray();
             bool caught = false;
@@ -60,7 +62,7 @@ namespace Akka.Streams.Tests.Dsl
             {
                 caught = true;
             }
-            
+
             caught.ShouldBeTrue();
         }
 
@@ -75,6 +77,7 @@ namespace Akka.Streams.Tests.Dsl
                 (output[0] == a).ShouldBeTrue("Did not get elements in order!");
                 output = output.Skip(1).ToArray();
             }
+
             output.Length.ShouldBe(0, "Did not receive all elements!");
         }
 
@@ -89,14 +92,16 @@ namespace Akka.Streams.Tests.Dsl
                 (output[0] == a).ShouldBeTrue("Did not get elements in order!");
                 output = output.Skip(1).ToArray();
             }
+
             output.Length.ShouldBe(0, "Did not receive all elements!");
-            
+
             output = input.ToArray();
             await foreach (var a in asyncEnumerable)
             {
                 (output[0] == a).ShouldBeTrue("Did not get elements in order!");
                 output = output.Skip(1).ToArray();
             }
+
             output.Length.ShouldBe(0, "Did not receive all elements in second enumeration!!");
         }
 
@@ -107,7 +112,7 @@ namespace Akka.Streams.Tests.Dsl
             var materializer = ActorMaterializer.Create(Sys);
             var probe = this.CreatePublisherProbe<int>();
             var task = Source.FromPublisher(probe).RunAsAsyncEnumerable(materializer);
-            
+
             var a = Task.Run(async () =>
             {
                 await foreach (var notused in task)
@@ -135,7 +140,7 @@ namespace Akka.Streams.Tests.Dsl
 
             thrown.ShouldBeTrue();
         }
-        
+
         [Fact]
         public async Task RunAsAsyncEnumerable_Throws_if_materializer_gone_before_Enumeration()
         {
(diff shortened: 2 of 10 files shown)


src/core/Akka.Streams.Tests/Akka.Streams.Tests.csproj
```diff
@@ -5,6 +5,7 @@
   <PropertyGroup>
     <AssemblyName>Akka.Streams.Tests</AssemblyName>
     <TargetFrameworks>$(NetFrameworkTestVersion);$(NetTestVersion);$(NetCoreTestVersion)</TargetFrameworks>
+    <LangVersion>9</LangVersion>
```
Aaronontheweb, line 8: Added this in order to natively support `IAsyncEnumerable` inside both Akka.Streams and its test suite. We can consider dropping the dependencies on the BCL NuGet package that pulls it in now.

Aaronontheweb:
```xml
  <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="5.0.0" />
```

Is referenced in Akka.Streams - let me see if I can remove it in an additional commit.

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -17,18 +17,13 @@
...
 using Akka.Streams.TestKit.Tests;
-using Akka.Streams.Tests.Actor;
-using Reactive.Streams;
 using System.Runtime.CompilerServices;
 using Akka.Util;
 using FluentAssertions.Extensions;
 
 namespace Akka.Streams.Tests.Dsl
 {
-#if NETCOREAPP
```
Aaronontheweb, line 31: `<LangVersion>9</LangVersion>` eliminates the need for these `#if` statements.

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -230,6 +225,47 @@ public async Task AsyncEnumerableSource_Must_Cancel_Running_Source_If_Downstream
...
+        /// </summary>
+        [Fact]
+        public async Task AsyncEnumerableSource_BugFix6280()
+        {
+            async IAsyncEnumerable<int> GenerateInts()
+            {
+                foreach (var i in Enumerable.Range(0, 100))
+                {
+                    if (i > 50)
+                        await Task.Delay(1000);
```
Aaronontheweb, line 240: This is designed to force the `ValueTask` returned from `IAsyncEnumerator.MoveNextAsync()` to be still running at the moment we attempt to dispose the `IAsyncEnumerator`.

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -230,6 +225,47 @@ public async Task AsyncEnumerableSource_Must_Cancel_Running_Source_If_Downstream
...
+                    yield return i;
+                }
+            }
+
+            var source = Source.From(GenerateInts);
+            var subscriber = this.CreateManualSubscriberProbe<int>();
+
+            await EventFilter.Warning().ExpectAsync(0, async () =>
```
Aaronontheweb, line 248: This assertion failed until I implemented my fix.

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -230,6 +225,47 @@ public async Task AsyncEnumerableSource_Must_Cancel_Running_Source_If_Downstream
...
+                    .ToMaterialized(Sink.FromSubscriber(subscriber), Keep.Left);
+            
+#pragma warning disable CS4014
+                var task = mat.Run(Materializer);
+#pragma warning restore CS4014
+
+                var subscription = subscriber.ExpectSubscription();
+                subscription.Request(50);
+                subscriber.ExpectNextN(Enumerable.Range(0, 50));
+                subscription.Request(10); // the iterator is going to start delaying 1000ms per item here
```
Aaronontheweb, line 261: Force the `IAsyncEnumerable` to emit the elements that start producing 1s delays.

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -230,6 +225,47 @@ public async Task AsyncEnumerableSource_Must_Cancel_Running_Source_If_Downstream
...
+                subscription.Cancel();
```
Aaronontheweb, line 262: And cancel immediately afterwards - this will trigger the exception before I made my changes.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3817,19 +3817,7 @@ public override void OnDownstreamFinish()
                 _completionCts.Cancel();
                 _completionCts.Dispose();
 
-                try
-                {
-                    _enumerator.DisposeAsync().GetAwaiter().GetResult();
```
Aaronontheweb, line 3822: The right solution here is to not dispose the iterator - iteration should stop automatically as a result of the `CancellationToken` and the `DisposeAsync` method won't do anything in scenarios where we're still waiting on the `Task` anyway. GC should still clean this up afterwards.

The alternative, adding a bunch of accounting code to safely call `DisposeAsync`, is really infeasible since we can't pass the `ValueTask` around (because it's a `struct`) and only supports a single `await`-er - unless we want to wrap it in a `Task` each time.

None of our other stages which retain an `IEnumerator<T>` call `Dispose` on it, so I don't think it's an issue.

Aaronontheweb (commented):
Detailed my changes

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
Aaronontheweb: Nope, had it add it back due to .NET Framework 4.7.1 compilation on Linux

Aaronontheweb:
No idea why AzDo isn't running....

Arkatufus (approved):
LGTM

PR #6294, opened 2022-12-06
add simple actor telemetry

### LINKED ISSUE #6293: Telemetry: publish events for each actor create / restart / terminate
### Discussed in https://github.com/akkadotnet/akka.net/discussions/5587

<div type='discussions-op-text'>

<sup>Originally posted by **Aaronontheweb** February  7, 2022</sup>

I've been thinking about this in a few areas:

1. [Phobos](https://phobos.petabridge.com/) - right now we use invocation counters to keep track of actor starts / stops to keep track the churn of actors. In platforms like Prometheus we compute the delta between start / stop events to compute the total number of alive actors. We can try using things like an OTel `UpDownCounter` (not yet implemented in .NET) when that gets added to reach a total number and maybe that will be more accurate.
2. [Petabridge.Cmd](https://cmd.petabridge.com/) - similar to how we report on shard counts right now, I'd love to update the `actor` commands to show a breakdown of how many actors are alive by hierarchy position and type.
3. Homespun telemetry solutions - right now most of these are done via the logging system or subscribing to various data channels inside Akka.NET (i.e. `Cluster.Subscribe`), but the actual structure of the actor hierarchy itself is difficult to capture here.

In particular, some of the work I've been doing around Phobos 2.0 has been leading me to wonder if it makes sense to emit some basic telemetry around actor creation inside Akka.NET itself.

I'll have to flesh this out in some more detail, but basically add three `EventStream` events that can be subscribed to:

```csharp
public readonly struct ActorStarted{
   public IActorRef Ref {get;} // can be used for topological sort / correlation
   public Type ActorType {get;}
}

public readonly struct ActorRestarted{
   public IActorRef Ref {get;} // can be used for topological sort / correlation
   public Type ActorType {get;}
   public Exception CrashReason {get;}
}

public readonly struct ActorTerminated{
   public IActorRef Ref {get;} // can be used for topological sort / correlation
   public Type ActorType {get;}
   public TerminatedReason Reason {get;} // stopped, parent decided, system shutting down
}
```

Having a `TerminatedReason` would be really helpful - implementing that alone would be a bit of work, but I think it's achievable.</div>

Aaronontheweb:
## Changes

Fixes #6293

This is an initial implementation designed to provide basic actor start / stop / restart telemetry for keeping track of the number of alive actor instances in a given Akka.NET process. We have plans to use this in both [Petabridge.Cmd](https://cmd.petabridge.com/) and [Phobos](https://phobos.petabridge.com/) to improve observability experiences for users.

This design relies on publishing simple, lightweight events via the `EventStream` to listeners that subscriber to those events in-process. These metrics are not enabled by default and must be explicitly enabled via the following HOCON setting:

```hocon
akka.actor.telemetry.enabled = true
```

## Checklist

* [x] Design discussion issue #6293 

### Latest `dev` Benchmarks 

__Need to run actor spawn benchmarks with this setting enabled and not enabled__

### This PR's Benchmarks

### DIFF
--- a/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.verified.txt
+++ b/src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.verified.txt
@@ -314,6 +314,13 @@ namespace Akka.Actor
         public static readonly Akka.Actor.IActorRef NoSender;
         public static readonly Akka.Actor.Nobody Nobody;
     }
+    public struct ActorRestarted : Akka.Actor.IActorTelemetryEvent, Akka.Actor.INoSerializationVerificationNeeded, Akka.Actor.INotInfluenceReceiveTimeout
+    {
+        public ActorRestarted(Akka.Actor.IActorRef subject, System.Type actorType, Akka.Actor.Reason reason) { }
+        public System.Type ActorType { get; }
+        public Akka.Actor.Reason Reason { get; }
+        public Akka.Actor.IActorRef Subject { get; }
+    }
     public class ActorSelection : Akka.Actor.ICanTell
     {
         public ActorSelection() { }
@@ -340,12 +347,24 @@ namespace Akka.Actor
         public Akka.Actor.ActorSelectionMessage Copy(object message = null, Akka.Actor.SelectionPathElement[] elements = null, System.Nullable<bool> wildCardFanOut = null) { }
         public override string ToString() { }
     }
+    public struct ActorStarted : Akka.Actor.IActorTelemetryEvent, Akka.Actor.INoSerializationVerificationNeeded, Akka.Actor.INotInfluenceReceiveTimeout
+    {
+        public ActorStarted(Akka.Actor.IActorRef subject, System.Type actorType) { }
+        public System.Type ActorType { get; }
+        public Akka.Actor.IActorRef Subject { get; }
+    }
     public class ActorStashPlugin : Akka.Actor.ActorProducerPluginBase
     {
         public ActorStashPlugin() { }
         public override void AfterIncarnated(Akka.Actor.ActorBase actor, Akka.Actor.IActorContext context) { }
         public override void BeforeIncarnated(Akka.Actor.ActorBase actor, Akka.Actor.IActorContext context) { }
         public override bool CanBeAppliedTo(System.Type actorType) { }
+    }
+    public struct ActorStopped : Akka.Actor.IActorTelemetryEvent, Akka.Actor.INoSerializationVerificationNeeded, Akka.Actor.INotInfluenceReceiveTimeout
+    {
+        public ActorStopped(Akka.Actor.IActorRef subject, System.Type actorType) { }
+        public System.Type ActorType { get; }
+        public Akka.Actor.IActorRef Subject { get; }
     }
     public abstract class ActorSystem : Akka.Actor.IActorRefFactory, System.IDisposable
     {
@@ -988,6 +1007,11 @@ namespace Akka.Actor
     {
         Akka.Actor.IStash Stash { get; set; }
     }
+    public interface IActorTelemetryEvent : Akka.Actor.INoSerializationVerificationNeeded, Akka.Actor.INotInfluenceReceiveTimeout
+    {
+        System.Type ActorType { get; }
+        Akka.Actor.IActorRef Subject { get; }
+    }
     public interface IAdvancedScheduler : Akka.Actor.IActionScheduler, Akka.Actor.IRunnableScheduler { }
     public interface IAutoReceivedMessage { }
     public interface ICanTell
@@ -1486,6 +1510,16 @@ namespace Akka.Actor
             public static readonly Akka.Actor.ProviderSelection.Remote Instance;
         }
     }
+    public sealed class Reason
+    {
+        public Reason(string type, string message = null) { }
+        public string Message { get; }
+        public string Type { get; }
+    }
+    public class static ReasonExtensions
+    {
+        public static Akka.Actor.Reason ToReason(this System.Exception ex) { }
+    }
     public delegate bool Receive(object message);
     public abstract class ReceiveActor : Akka.Actor.UntypedActor, Akka.Actor.Internal.IInitializableActor
     {
@@ -1682,6 +1716,7 @@ namespace Akka.Actor
         public bool DebugRouterMisconfiguration { get; }
         public bool DebugUnhandledMessage { get; }
         public int DefaultVirtualNodesFactor { get; }
+        public bool EmitActorTelemetry { get; }
         public bool FsmDebugEvent { get; }
         public bool HasCluster { get; }
         public string Home { get; }
--- a/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
+++ b/src/core/Akka/Actor/ActorCell.DefaultMessages.cs
@@ -449,6 +449,8 @@ namespace Akka.Actor
                 CheckReceiveTimeout();
                 if (System.Settings.DebugLifecycle)
                     Publish(new Debug(Self.Path.ToString(), created.GetType(), "Started (" + created + ")"));
+                if(System.Settings.EmitActorTelemetry)
+                    System.EventStream.Publish(new ActorStarted(Self, Props.Type));
             }
             catch (Exception e)
             {
--- a/src/core/Akka/Actor/ActorCell.FaultHandling.cs
+++ b/src/core/Akka/Actor/ActorCell.FaultHandling.cs
@@ -291,6 +291,9 @@ namespace Akka.Actor
                     var pipeline = _systemImpl.ActorPipelineResolver.ResolvePipeline(a.GetType());
                     pipeline.BeforeActorIncarnated(a, this);
                 }
+                
+                if(System.Settings.EmitActorTelemetry)
+                    System.EventStream.Publish(new ActorStopped(Self, Props.Type));
             }
             catch (Exception x)
             {
@@ -348,7 +351,9 @@ namespace Akka.Actor
 
                 if (System.Settings.DebugLifecycle)
                     Publish(new Debug(_self.Path.ToString(), freshActor.GetType(), "Restarted (" + freshActor + ")"));
-
+                if(System.Settings.EmitActorTelemetry)
+                    System.EventStream.Publish(new ActorRestarted(Self, Props.Type, cause.ToReason()));
+                
                 // only after parent is up and running again do restart the children which were not stopped
                 foreach (var survivingChild in survivors)
                 {
@@ -382,8 +387,7 @@ namespace Akka.Actor
             //the UID protects against reception of a Failed from a child which was
             //killed in preRestart and re-created in postRestart
 
-            ChildRestartStats childStats;
-            if (TryGetChildStatsByRef(failedChild, out childStats))
+            if (TryGetChildStatsByRef(failedChild, out var childStats))
             {
                 var statsUid = childStats.Child.Path.Uid;
                 if (statsUid == f.Uid)
@@ -429,8 +433,7 @@ namespace Akka.Actor
 
             // if the removal changed the state of the (terminating) children container,
             // then we are continuing the previously suspended recreate/create/terminate action
-            var recreation = status as SuspendReason.Recreation;
-            if (recreation != null)
+            if (status is SuspendReason.Recreation recreation)
             {
                 FinishRecreate(recreation.Cause, _actor);
             }
--- a/src/core/Akka/Actor/ActorTelemetry.cs
+++ b/src/core/Akka/Actor/ActorTelemetry.cs
@@ -1,0 +1,104 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="ActorTelemetry.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using Akka.Event;
+
+namespace Akka.Actor
+{
+    /// <summary>
+    /// A set of events designed to provide some basic telemetry functions for monitoring actor lifecycles.
+    ///
+    /// We want to track actor starts, stops, and restarts. More detailed metrics, such as mailbox size or message
+    /// processing rates will require something like Phobos [https://phobos.petabridge.com/].
+    /// </summary>
+    /// <remarks>
+    /// Not intended to be sent across network boundaries - should only be processed locally via the <see cref="EventStream"/>.
+    /// </remarks>
+    public interface IActorTelemetryEvent : INoSerializationVerificationNeeded, INotInfluenceReceiveTimeout
+    {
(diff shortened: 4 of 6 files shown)


Arkatufus:
Would it be better to add a timestamp to the event messages? Maybe a `DateTime.UtcNow.Ticks()`, so that it can be reconstructed later on by analytics

Aaronontheweb:
@Arkatufus I was think

> Would it be better to add a timestamp to the event messages? Maybe a `DateTime.UtcNow.Ticks()` or `DateTimeOffset.UtcNow.ToUnixTimeMilliseconds()`, so that it can be reconstructed later on by analytics.

Since this is just a metric emission I wanted to keep it utterly bare minimum - a downstream consumer can convert these events into timed log entries if they wish.

src/core/Akka/Actor/ActorTelemetry.cs
```diff
@@ -0,0 +1,104 @@
...
+        IActorRef Subject {get;}
+        
+        /// <summary>
+        /// The implementation type for this actor.
+        /// </summary>
+        Type ActorType { get; }
+    }
+    
+    // Create ActorTelemetryEvent messages for the following events: starting an actor, stopping an actor, restarting an actor
+    public readonly struct ActorStarted : IActorTelemetryEvent
```
to11mtm, line 36: Is it worth using Structs here if they are getting boxed anyway?

Aaronontheweb: I'll benchmark this and find it - what I'd _like_ to do is use structs and get rid of the boxing but that'll require a non-binary-compatible change to the `EventStream` class.

Aaronontheweb: So that change would probably be an Akka.NET v1.5 thing.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.verified.txt
```diff
@@ -1486,6 +1510,16 @@ namespace Akka.Actor
             public static readonly Akka.Actor.ProviderSelection.Remote Instance;
         }
     }
+    public sealed class Reason
+    {
+        public Reason(string type, string message = null) { }
+        public string Message { get; }
+        public string Type { get; }
+    }
```
Zetanova, line 1518: The `public guid Error` property is missing. 
It would make it possible to tag and serialize/log exceptions into elasticsearch, seq, opentelemetry, Redis or simply into a S3 bucket and only transmit this `Reason` to remote nodes. 
If needed the receiving remote system could try restore/deserialize the exception.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.verified.txt
```diff
@@ -1486,6 +1510,16 @@ namespace Akka.Actor
...
+    public class static ReasonExtensions
+    {
+        public static Akka.Actor.Reason ToReason(this System.Exception ex) { }
+    }
```
Zetanova, line 1522: This extensions method should be on the ActorContext or ActorSystem,
to make it possible to retrieve a new AkkaErrorService from Extensions list.
The Default implementation can simply convert the `Exception` to a `Reason` with `Error = Guid.Empty`

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.verified.txt
Zetanova: Maybe instead of `Guid Error` a property `string Url` would be better. 
It will allow to have many/mix of error service types.

Aaronontheweb: Why not use a log entry for this? These are just meant to be simple metric updates

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.verified.txt
Aaronontheweb: These metrics are only produced within the 'ActorCell' and it's performed automatically by Akka.NET

src/core/Akka/Actor/ActorTelemetry.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// <summary>
+    /// A set of events designed to provide some basic telemetry functions for monitoring actor lifecycles.
+    ///
+    /// We want to track actor starts, stops, and restarts. More detailed metrics, such as mailbox size or message
+    /// processing rates will require something like Phobos [https://phobos.petabridge.com/].
+    /// </summary>
+    /// <remarks>
+    /// Not intended to be sent across network boundaries - should only be processed locally via the <see cref="EventStream"/>.
+    /// </remarks>
+    public interface IActorTelemetryEvent : INoSerializationVerificationNeeded, INotInfluenceReceiveTimeout
```
Aaronontheweb, line 22: These are meant to be metric deltas / increments that can be aggregated by a subscriber on the `EventStream` - not used for tracing or logs. Specific to the `IActorRef` lifecycle.

Aaronontheweb: Handling matters like measuring message processing rates this way would be a disaster from a performance and memory management point of view - the overhead of measuring message processing rates would be greater than or equal to the overhead of processing the messages themselves. Use [Phobos](https://phobos.petabridge.com/) for that use case.

src/core/Akka/Actor/Settings.cs
```diff
@@ -242,6 +243,17 @@ public Settings(ActorSystem system, Config config, ActorSystemSetup setup)
...
+        /// <summary>
+        /// When set to <c>true</c>, all actors will emit <see cref="IActorTelemetryEvent"/>s when they are created, stopped, or restarted.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to <c>false</c>.
+        /// </remarks>
+        /// <code>
+        /// akka.actor.telemetry.enabled = on
+        /// </code>
+        public bool EmitActorTelemetry { get; }
```
Aaronontheweb, line 256: Setting that controls the emission of telemetry events. Off by default.

src/core/Akka/Configuration/Pigeon.conf
```diff
@@ -118,6 +118,12 @@ akka {
     # Default timeout for IActorRef.Ask.
     ask-timeout = infinite
 
+    # Controls telemetry settings built into Akka.NET
+    telemetry{
+      # Enables telemetry emission from actors - actors will emit events
+      # whenever they are started, stopped, or restarted.
+      enabled = false
```
Aaronontheweb, line 125: relevant HOCON setting.

src/core/Akka/Actor/ActorCell.FaultHandling.cs
```diff
@@ -348,7 +351,9 @@ private void FinishRecreate(Exception cause, ActorBase failedActor)
 
                 if (System.Settings.DebugLifecycle)
                     Publish(new Debug(_self.Path.ToString(), freshActor.GetType(), "Restarted (" + freshActor + ")"));
-
+                if(System.Settings.EmitActorTelemetry)
+                    System.EventStream.Publish(new ActorRestarted(Self, Props.Type, cause.ToReason()));
```
Aaronontheweb, line 355: Emit `ActorRestarted` events here automatically when an actor is being rebooted. Include details about the exception that caused it to restart (so that data can be aggregated downstream.)

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -449,6 +449,8 @@ private void Create(Exception failure)
                 CheckReceiveTimeout();
                 if (System.Settings.DebugLifecycle)
                     Publish(new Debug(Self.Path.ToString(), created.GetType(), "Started (" + created + ")"));
+                if(System.Settings.EmitActorTelemetry)
+                    System.EventStream.Publish(new ActorStarted(Self, Props.Type));
```
Aaronontheweb, line 453: Emit actor started events here - we validate via our specs that these aren't duplicated when an actor restarts.

src/core/Akka.Remote.Tests/RemoteActorTelemetrySpecs.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            protected override void PreStart()
+            {
+                Context.System.EventStream.Subscribe(Self, typeof(IActorTelemetryEvent));
+            }
+        }
+
+        // create a unit test where a second ActorSystem connects to Sys and receives an IActorRef from Sys and subscribes to Telemetry events
+        [Fact]
+        public async Task RemoteActorRefs_should_not_produce_telemetry()
```
Aaronontheweb, line 93: Any `IActorRef` without an `ActorCell` will not produce metrics (by design) - so `FutureActorRef`s from `Ask<T>` and `RemoteActorRef`s don't influence these totals.

src/core/Akka.Tests/Actor/ActorTelemetrySpecs.cs
```diff
@@ -0,0 +1,304 @@
...
+                // assert 100 stops recorded (only the child actors)
+                Assert.Equal(100, telemetry.ActorStopped);
+            }, RemainingOrDefault);
+        }
+        
+        /// <summary>
+        /// Pool routers should have their start / stop / restarts counted too
+        /// </summary>
+        [Fact]
+        public async Task ActorTelemetry_must_be_accurate_for_pool_router()
```
Aaronontheweb, line 255: Router actors have an `ActorCell` even though they don't have a typical `Mailbox`, therefore they will also emit the same events are other local actors do for starting / stopping / restarting.

src/core/Akka/Actor/ActorTelemetry.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            Subject = subject;
+            ActorType = actorType;
+            Reason = reason;
+        }
+
+        public IActorRef Subject { get; }
+        public Type ActorType { get; }
+        
+        public Reason Reason { get; }
```
Aaronontheweb, line 102: These messages are meant to be consumed in a local process - why not just include the `Exception` here so the consumer can decide how to report?

src/core/Akka/Actor/ActorTelemetry.cs
```diff
@@ -0,0 +1,104 @@
...
+        public IActorRef Subject { get; }
+        public Type ActorType { get; }
+    }
+    
+    /// <summary>
+    /// Emitted when an actor restarts.
+    /// </summary>
+    public readonly struct ActorRestarted : IActorTelemetryEvent
+    {
+        public ActorRestarted(IActorRef subject, Type actorType, Reason reason)
```
Aaronontheweb, line 92: Mark all constructors as `internal` so these events can be emitted from user code? If I'm building reporting tools that are supposed to accurately report on what current load in a large Akka.Cluster looks like, someone else emitting their own versions of these metrics into the `EventStream` is going to throw this off _and the other developers on their own team probably won't know that_.

Aaronontheweb (commented):
Detailed my own changes - still need to run benchmarks so we can get data on `struct` vs. `class` for these types.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.verified.txt
Zetanova: The `Reason` is need to transfer unknown Exception types between nodes transparently. 
Has not much to to with the `IActorTelemetryEvent` them self.

`Actor.Remote` creates the assumption that all remoting is transparent.  
But if a node transmit a unknown exception type to a node it fails horrible.

The general idea is to put the `Èxception` into memory cache or distributed storage/cache 
then can then be loaded when really needed,  
to get from Reason => Exception again

Aaronontheweb: By design, these events are not designed to be transmitted over the wire - local aggregation only. We don't want this data coupled to remoting, serialization, or persistence. "Simple metrics" is the goal.

Aaronontheweb:
# `SpawnActor` Benchmark Results

We ran the `Akka.Benchmarks.Actor.SpawnActorBenchmarks` on .NET 7.0.#6294 

## `readonly struct` Implementation

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19044.2251 (21H2)
.NET SDK=7.0.100
  [Host]     : .NET 7.0.0 (7.0.22.51805), X64 RyuJIT
  Job-ZEAKWQ : .NET 7.0.0 (7.0.22.51805), X64 RyuJIT

InvocationCount=1  IterationCount=10  RunStrategy=Throughput  
UnrollFactor=1  WarmupCount=5  

```
|      Method | ActorCount | EnableTelemetry |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 |     Gen 2 | Allocated |
|------------ |----------- |---------------- |--------:|---------:|---------:|------------:|-----------:|----------:|----------:|
| **Actor_spawn** |     **100000** |           **False** | **3.592 s** | **0.2078 s** | **0.1237 s** | **169000.0000** | **48000.0000** | **2000.0000** |    **986 MB** |
| **Actor_spawn** |     **100000** |            **True** | **3.620 s** | **0.1311 s** | **0.0867 s** | **172000.0000** | **50000.0000** | **4000.0000** |    **993 MB** |

## `sealed class` Implementation

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19044.2251 (21H2)
.NET SDK=7.0.100
  [Host]     : .NET 7.0.0 (7.0.22.51805), X64 RyuJIT
  Job-HEBSLY : .NET 7.0.0 (7.0.22.51805), X64 RyuJIT

InvocationCount=1  IterationCount=10  RunStrategy=Throughput  
UnrollFactor=1  WarmupCount=5  

```
|      Method | ActorCount | EnableTelemetry |    Mean |    Error |   StdDev |       Gen 0 |      Gen 1 |     Gen 2 | Allocated |
|------------ |----------- |---------------- |--------:|---------:|---------:|------------:|-----------:|----------:|----------:|
| **Actor_spawn** |     **100000** |           **False** | **3.543 s** | **0.2174 s** | **0.1438 s** | **170000.0000** | **49000.0000** | **3000.0000** |    **985 MB** |
| **Actor_spawn** |     **100000** |            **True** | **3.544 s** | **0.1591 s** | **0.1052 s** | **170000.0000** | **49000.0000** | **2000.0000** |    **993 MB** |


Looks to me like the impact of boxing is higher than the impact of having another class to GC. We should go with the `class` implementation.

src/core/Akka.API.Tests/CoreAPISpec.ApproveCore.verified.txt
Zetanova: I am asking only to make the `Reason` type multi purpose that it can be in general used instead of an exception.
The remote transparency has this intrinsic conflict with the `Exception` dotnet type.
The MS guys for `system.runtime.remoting` in framework 1.1 times already discovered it.

Arkatufus (approved):
LGTM 👍

Aaronontheweb:
In terms of website documentation - I think we're going to need to expand our real-world deployment area of the site to include some discussion points on telemetry.

Aaronontheweb:
@Arkatufus we're going to need to port this to v1.5 (the `dev` branch)

PR #6299, opened 2022-12-09
[Port #6294] add simple actor telemetry

Arkatufus:
Port of Port #6294

* added initial actor telemetry for https://github.com/akkadotnet/akka.net/issues/6293

* added basic telemetry tests for local actors

* added spec to validate that `RemoteActorRef` doesn't influence counters

* updated `SpawnActorBenchmarks` to include telemetry impact

* converted telemetry events into `sealed class`es with `internal` constructors

* removed `Reason`

(cheery-picked from 7f68c48d65ec9d1c7216cda309a81430f839d90c)


Aaronontheweb:
@Arkatufus looks like all of the API approvals barfed here

PR #6323, opened 2022-12-30
Read stash capacity from actor's mailbox or dispatcher configuration

ismaelhamed:
Fixes reading stash capacity from actor's mailbox or dispatcher configuration.

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [X] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [X] Changes in public API reviewed, if any.

PR #6325, opened 2023-01-04
Added support for `UnrestrictedStash`

ismaelhamed:
Currently, once you've marked your actor with any of the Stash interfaces (`IWithUnboundedStash` or `IWithBoundedStash`) there's no way to change it afterward via configuration (it is a design-time operation that requires recompiling).

But there are two traits that allow users to do that in the JVM:

- A default `Stash` trait (`IWithStash` in this PR) which allows using any Deque-based mailbox (i.e., bounded or unbounded) via configuration. By default, it uses an UnboundedDeque mailbox, so it is akin to the current `IWithUnboundedStash` if you don't specify a custom mailbox type.

- `UnrestrictedStash` trait, which does not enforce any mailbox type. The proper mailbox *must* be configured
manually.

## Checklist

### DIFF
--- a/docs/articles/actors/receive-actor-api.md
+++ b/docs/articles/actors/receive-actor-api.md
@@ -790,12 +790,15 @@ static void Main(string[] args)
 
 ## Stash
 
-The `IWithUnboundedStash` interface enables an actor to temporarily stash away messages that can not or should not be handled using the actor's current behavior. Upon changing the actor's message handler, i.e., right before invoking `Context.BecomeStacked()` or `Context.UnbecomeStacked()`;, all stashed messages can be "un-stashed", thereby prepending them to the actor's mailbox. This way, the stashed messages can be processed in the same order as they have been received originally. An actor that implements `IWithUnboundedStash` will automatically get a dequeue-based mailbox.
+The `IWithStash` interface enables an actor to temporarily stash away messages that can not or should not be handled using the actor's current behavior. Upon changing the actor's message handler, i.e., right before invoking `Context.Become()` or `Context.Unbecome()`, all stashed messages can be "un-stashed", thereby prepending them to the actor's mailbox. This way, the stashed messages can be processed in the same order as they have been received originally.
 
-Here is an example of the `IWithUnboundedStash` interface in action:
+> [!NOTE]
+> The interface `IWithStash` implements the marker interface `IRequiresMessageQueue<IDequeBasedMessageQueueSemantics>` which requests the system to automatically choose a deque-based mailbox implementation for the actor (defaults to an unbounded deque mailbox). If you want more control over the mailbox, see the documentation on mailboxes: [Mailboxes](xref:mailboxes).
+
+Here is an example of the `IWithStash` interface in action:
 
 ```csharp
-public class ActorWithProtocol : ReceiveActor, IWithUnboundedStash
+public class ActorWithProtocol : ReceiveActor, IWithStash
 {
     public IStash Stash { get; set; }
 
@@ -827,11 +830,18 @@ public class ActorWithProtocol : ReceiveActor, IWithUnboundedStash
 }
 ```
 
-Invoking `Stash()` adds the current message (the message that the actor received last) to the actor's stash. It is typically invoked when handling the default case in the actor's message handler to stash messages that aren't handled by the other cases. It is illegal to stash the same message twice; to do so results in an `IllegalStateException` being thrown. The stash may also be bounded in which case invoking `Stash()` may lead to a capacity violation, which results in a `StashOverflowException`. The capacity of the stash can be configured using the stash-capacity setting (an `Int`) of the mailbox's configuration.
+Invoking `Stash()` adds the current message (the message that the actor received last) to the actor's stash. It is typically invoked when handling the default case in the actor's message handler to stash messages that aren't handled by the other cases. It is illegal to stash the same message twice; to do so results in an `IllegalStateException` being thrown. The stash may also be bounded in which case invoking `Stash()` may lead to a capacity violation, which results in a `StashOverflowException`. The capacity of the stash can be configured using the stash-capacity setting (an `int`) of the mailbox's configuration.
 
 Invoking `UnstashAll()` enqueues messages from the stash to the actor's mailbox until the capacity of the mailbox (if any) has been reached (note that messages from the stash are prepended to the mailbox). In case a bounded mailbox overflows, a `MessageQueueAppendFailedException` is thrown. The stash is guaranteed to be empty after calling `UnstashAll()`.
 
-Note that the `stash` is part of the ephemeral actor state, unlike the mailbox. Therefore, it should be managed like other parts of the actor's state which have the same property. The `IWithUnboundedStash` interface implementation of `PreRestart` will call `UnstashAll()`, which is usually the desired behavior.
+Note that the stash is part of the ephemeral actor state, unlike the mailbox. Therefore, it should be managed like other parts of the actor's state which have the same property.
+
+However, the `IWithStash` interface implementation of `PreRestart` will call `UnstashAll()`. This means that before the actor restarts, it will transfer all stashed messages back to the actor’s mailbox.
+
+The result of this is that when an actor is restarted, any stashed messages will be delivered to the new incarnation of the actor. This is usually the desired behavior.
+
+> [!NOTE]
+> If you want to enforce that your actor can only work with an unbounded stash, then you should use the `IWithUnboundedStash` interface instead.
 
 ## Killing an Actor
 
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.Core.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.Core.verified.txt
@@ -1178,14 +1178,15 @@ namespace Akka.Actor
         void Become(Akka.Actor.UntypedReceive receive);
         void BecomeStacked(Akka.Actor.UntypedReceive receive);
     }
-    [System.ObsoleteAttribute("Bounded stashing is not yet implemented. Unbounded stashing will be used instead " +
-        "[0.7.0]")]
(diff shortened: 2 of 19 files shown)


src/core/Akka/Actor/Stash/IWithBoundedStash.cs
```diff
@@ -17,8 +17,8 @@ namespace Akka.Actor
     /// <code>public IStash Stash { get; set; }</code>
     /// </summary>
     // ReSharper disable once InconsistentNaming
-    [Obsolete("Bounded stashing is not yet implemented. Unbounded stashing will be used instead [0.7.0]")]
-    public interface IWithBoundedStash : IActorStash, IRequiresMessageQueue<IBoundedDequeBasedMessageQueueSemantics>
+    [Obsolete("Use `IWithStash` with a configured BoundedDeque-based mailbox instead.")]
+    public interface IWithBoundedStash : IWithUnrestrictedStash, IRequiresMessageQueue<IBoundedDequeBasedMessageQueueSemantics>
```
ismaelhamed, line 21: We can remove this type, since typically you'd want to change the mailbox or stash capacity with a bounded deque, which cannot be done programmatically (AFAIK)

src/core/Akka/Actor/Stash/StashFactory.cs
```diff
@@ -50,15 +48,14 @@ public static IStash CreateStash(this IActorContext context, IActorStash actorIn
...
-            {
+            if (actorType.Implements<IWithUnboundedStash>())
                 return new UnboundedStashImpl(context);
-            }
+
+            if (actorType.Implements<IWithUnrestrictedStash>())
+                return new UnrestrictedStashImpl(context);
 
             throw new ArgumentException($"Actor {actorType} implements an unrecognized subclass of {typeof(IActorStash)} - cannot instantiate", nameof(actorType));
         }
```
ismaelhamed, line 61: `BoundedStashImpl`, `UnboundedStashImpl` and `UnrestrictedStashImpl` are all empty implementations, so we could get away with a single type (i.e., `StashSupportImpl`) like in the JVM.

Aaronontheweb:
> UnrestrictedStash trait, which does not enforce any mailbox type. The proper mailbox must be configured
manually.

So in practice, how does this work? The user has to manually specify a mailbox type in addition to specifying the stash type?

ismaelhamed:
> > UnrestrictedStash trait, which does not enforce any mailbox type. The proper mailbox must be configured
> > manually.
> 
> So in practice, how does this work? The user has to manually specify a mailbox type in addition to specifying the stash type?

Correct. By using `IWithUnrestrictedStash` you're opting out from the framework enforcing any mailbox type, but you must specify one (there's no default mailbox in this case, obviously).

ismaelhamed:
Updated the documentation.

As an addendum to my previous comment, you now have two main stash interfaces to work with:

- `IWithStash`, which is "unrestricted" since it allows both bounded and unbounded deque-based mailboxes. Defaults to unbounded if not specified otherwise.
- `IWithUnboundedStash`, if you want to enforce that your actor can only work with an unbounded stash.

The `UnrestrictedStash` type is there to solve the single inheritance problem in JAVA, which I've respected for the sake of maintaining a compatible API. But since a Stash's mailbox *MUST* be deque-based anyway, using `IWithStash` or `IWithUnrestrictedStash` is basically the same (except that in the latter you must always specify a mailbox).

ismaelhamed:
@Aaronontheweb could we backport this and #6323 to `v1.4`?

ismaelhamed:
Unrelated tests failing

Aaronontheweb:
I put it on my planner to review this today - will do @ismaelhamed

Aaronontheweb:
> @Aaronontheweb could we backport this and #6323 to `v1.4`?

yep, I don't see why not provided that there's nothing high risk in the PR

docs/articles/actors/receive-actor-api.md
```diff
@@ -790,12 +790,15 @@ static void Main(string[] args)
...
+The `IWithStash` interface enables an actor to temporarily stash away messages that can not or should not be handled using the actor's current behavior. Upon changing the actor's message handler, i.e., right before invoking `Context.Become()` or `Context.Unbecome()`, all stashed messages can be "un-stashed", thereby prepending them to the actor's mailbox. This way, the stashed messages can be processed in the same order as they have been received originally.
 
-Here is an example of the `IWithUnboundedStash` interface in action:
+> [!NOTE]
+> The interface `IWithStash` implements the marker interface `IRequiresMessageQueue<IDequeBasedMessageQueueSemantics>` which requests the system to automatically choose a deque-based mailbox implementation for the actor (defaults to an unbounded deque mailbox). If you want more control over the mailbox, see the documentation on mailboxes: [Mailboxes](xref:mailboxes).
```
Aaronontheweb, line 796: LGTM

docs/articles/actors/receive-actor-api.md
```diff
@@ -827,11 +830,18 @@ public class ActorWithProtocol : ReceiveActor, IWithUnboundedStash
...
 
-Note that the `stash` is part of the ephemeral actor state, unlike the mailbox. Therefore, it should be managed like other parts of the actor's state which have the same property. The `IWithUnboundedStash` interface implementation of `PreRestart` will call `UnstashAll()`, which is usually the desired behavior.
+Note that the stash is part of the ephemeral actor state, unlike the mailbox. Therefore, it should be managed like other parts of the actor's state which have the same property.
+
+However, the `IWithStash` interface implementation of `PreRestart` will call `UnstashAll()`. This means that before the actor restarts, it will transfer all stashed messages back to the actor’s mailbox.
+
+The result of this is that when an actor is restarted, any stashed messages will be delivered to the new incarnation of the actor. This is usually the desired behavior.
+
+> [!NOTE]
+> If you want to enforce that your actor can only work with an unbounded stash, then you should use the `IWithUnboundedStash` interface instead.
```
Aaronontheweb, line 844: LGTM

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.Core.verified.txt
```diff
@@ -1178,14 +1178,15 @@ namespace Akka.Actor
...
-    public interface IWithBoundedStash : Akka.Actor.IActorStash, Akka.Dispatch.IRequiresMessageQueue<Akka.Dispatch.IBoundedDequeBasedMessageQueueSemantics> { }
+    [System.ObsoleteAttribute("Use `IWithStash` with a configured BoundedDeque-based mailbox instead.")]
+    public interface IWithBoundedStash : Akka.Actor.IActorStash, Akka.Actor.IWithUnrestrictedStash, Akka.Dispatch.IRequiresMessageQueue<Akka.Dispatch.IBoundedDequeBasedMessageQueueSemantics> { }
+    public interface IWithStash : Akka.Actor.IActorStash, Akka.Actor.IWithUnrestrictedStash, Akka.Dispatch.IRequiresMessageQueue<Akka.Dispatch.IDequeBasedMessageQueueSemantics> { }
     public interface IWithTimers
     {
         Akka.Actor.ITimerScheduler Timers { get; set; }
     }
-    public interface IWithUnboundedStash : Akka.Actor.IActorStash, Akka.Dispatch.IRequiresMessageQueue<Akka.Dispatch.IUnboundedDequeBasedMessageQueueSemantics> { }
+    public interface IWithUnboundedStash : Akka.Actor.IActorStash, Akka.Actor.IWithUnrestrictedStash, Akka.Dispatch.IRequiresMessageQueue<Akka.Dispatch.IUnboundedDequeBasedMessageQueueSemantics> { }
```
Aaronontheweb, line 1188: This is extend-only, so it's safe.

src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistenceSqlCommon.Net.verified.txt
```diff
@@ -272,7 +272,7 @@ namespace Akka.Persistence.Sql.Common.Journal
         public long Offset { get; }
         public Akka.Actor.IActorRef ReplyTo { get; }
     }
-    public abstract class SqlJournal : Akka.Persistence.Journal.AsyncWriteJournal, Akka.Actor.IActorStash, Akka.Actor.IWithUnboundedStash, Akka.Dispatch.IRequiresMessageQueue<Akka.Dispatch.IUnboundedDequeBasedMessageQueueSemantics>
+    public abstract class SqlJournal : Akka.Persistence.Journal.AsyncWriteJournal, Akka.Actor.IActorStash, Akka.Actor.IWithUnboundedStash, Akka.Actor.IWithUnrestrictedStash, Akka.Dispatch.IRequiresMessageQueue<Akka.Dispatch.IUnboundedDequeBasedMessageQueueSemantics>
```
Aaronontheweb, line 275: This makes me think that we might have to update all of the Akka.Persistence plugins, since the journal currently uses an unbounded stash. I don't think that should be important unless there's something on the normal `IWithUnrestrictedStash` interface that could cause a `MissingMethodException` but it looks to me like there isn't any. Just wanted to make note of this in case I'm wrong.

src/core/Akka/Actor/Stash/IWithBoundedStash.cs
Aaronontheweb: Let's remove this type in v1.5 in a separate PR - I'd like to backport this as-is to v1.4 first.

Aaronontheweb (approved):
LGTM

src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistenceSqlCommon.Net.verified.txt
ismaelhamed: Not really because the journal is still implementing `IWithUnboundedStash` which is more constrained. So, the mailbox is going to be selected based on it and not `IWithUnrestrictedStash`.

PR #6327, opened 2023-01-05
Add API for UntypedActorWithStash types

ismaelhamed:
Added utility abstract classes for `UntypedActor`s that support stashing: 

- `UntypedActorWithStash`
- `UntypedActorWithUnboundedStash` 
- `UntypedActorWithUnrestrictedStash`

Depends on #6325 

## Checklist

### DIFF
--- a/docs/articles/actors/untyped-actor-api.md
+++ b/docs/articles/actors/untyped-actor-api.md
@@ -708,18 +708,16 @@ static void Main(string[] args)
 
 ## Stash
 
-The `IWithStash` interface enables an actor to temporarily stash away messages that can not or should not be handled using the actor's current behavior. Upon changing the actor's message handler, i.e., right before invoking `Context.Become()` or `Context.Unbecome()`, all stashed messages can be "un-stashed", thereby prepending them to the actor's mailbox. This way, the stashed messages can be processed in the same order as they have been received originally.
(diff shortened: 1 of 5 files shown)


Aaronontheweb:
@ismaelhamed #6325 has been merged. I can start work on backporting it probably on Friday unless you want to do it (via `git cherry-pick`)

ismaelhamed:
> @ismaelhamed #6325 has been merged. I can start work on backporting it probably on Friday unless you want to do it (via `git cherry-pick`)

I think I can find the time to do it myself. I will be back porting #6323, #6325 and #6327 if you're fine with it.

Aaronontheweb:
> > @ismaelhamed #6325 has been merged. I can start work on backporting it probably on Friday unless you want to do it (via `git cherry-pick`)
> 
> I think I can find the time to do it myself. I will be backporting #6323, #6325 and #6327 if you're fine with it.

Yes, that would delight me

Aaronontheweb (commented):
Is the general goal behind this API change to move away from the actor creation pipeline, so the stash is provided automatically as part of the base class infrastructure itself rather than injected in from the outside?

Aaronontheweb (approved):
LGTM

ismaelhamed:
> Is the general goal behind this API change to move away from the actor creation pipeline, so the stash is provided automatically as part of the base class infrastructure itself rather than injected in from the outside?

@Aaronontheweb  If you think about it, using stashing is currently a little weird since users actually must implement an interface (including declaring the `Stash` property that, BTW, they are not instantiating themselves). That's because in scala traits can have behavior, which can't be done with interfaces in JAVA and CSharp (well, we have default interfaces now but you get the point).

So it is natural that these utility classes exist in the JAVA API, allowing the user to just derive from a class (like they would with normal actors) and just get everything needed for stashing already set up. They can still implement the interfaces themselves if they choose to, but this is a bit of an improvement on the usability side.

Aaronontheweb:
> They can still implement the interfaces themselves if they choose to, but this is a bit of an improvement on the usability side.

Time will tell I guess, but I see your point.

Issue #6362, opened 2023-01-25
BatchingSqlJournal's FailChunkExecution mishandles `DbExceptions` wrapped in an `AggregateException`


ismaelhamed:
**Version Information**
1.4.48
Akka.Persistence.Sql.Common

**Describe the bug**
Whenever a `DbException` is wrapped in an `AggregateException` (because of line [#1003](https://github.com/Arkatufus/akka.net/blob/4ad16ccfb4f044d164bd13bff2ba99e19abd3b29/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs#L1003)), the `FailChunkException` method will mistreat it as a "non-database related" failure, thus failing to stop the persistent actor and leaving it in an unstable state.

**Expected behavior**
Unwrap `AggregateException`s to inspect the actual cause of the exception.

**Actual behavior**
Mishandles `AggregateException`s altogether.

PR #6361, opened 2023-01-25
Fixes FailChunkExecution does not handle `DbExceptions` wrapped in an `AggregateException`

ismaelhamed:
Fixes #6362

* [X] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).

PR #6364, opened 2023-01-26
[BACKPORT] FailChunkExecution does not handle `DbExceptions` wrapped in an `AggregateException` (#6361)

ismaelhamed:
Backport of #6362


Aaronontheweb:
thanks @ismaelhamed - we'll do a release today after we ship 1-2 more additional minor bug fixes that we can get done shortly.

PR #6366, opened 2023-01-26
close #6295 - set default PoolRouter SupervisorStrategy to Restart

### LINKED ISSUE #6295: Pool routers use `Directive.Escalate` by default when supervising routees - this causes unintended consequences
**Version Information**
Version of Akka.NET? 1.4.46 and earlier, 1.5.0-alpha3 and earlier.
Which Akka.NET Modules? Akka

**Describe the bug**

Currently, the default `SupervisionStrategy` on all `Pool` routers uses `Directive.Escalate`, which causes the router to restart and kill all routees every time any of the routees restart. This can be problematic, as it means that all routees are restarted even if only one of them has failed.

Instead, I suggest changing the default `SupervisionStrategy` to return a `Directive.Restart `instead. This would cause only the failed routee to be restarted, rather than restarting all routees in the pool. This would allow for more granular control over how failures are handled, and can prevent unnecessary restarts of healthy routees.

**To Reproduce**

Reproduced with one of my unit tests on https://github.com/akkadotnet/akka.net/pull/6294

```csharp
// unit test to assert that a pool router actor's children generate telemetry events, but not the router actor itself
[Fact]
public async Task ActorTelemetry_must_be_accurate_for_pool_router()
{
    // create a TelemetrySubscriber actor
    var subscriber = Sys.ActorOf(Props.Create<TelemetrySubscriber>(), "subscriber");
    
    // request current telemetry values (ensure that the actor has started, so counter values will be accurate)
    await subscriber.Ask<TelemetrySubscriber.GetTelemetry>(TelemetrySubscriber.GetTelemetryRequest.Instance);
    
    // create a pool router
    var router = Sys.ActorOf(Props.Create<ChildActor>().WithRouter(new RoundRobinPool(10)), "router");

    // awaitassert collecting data from the telemetry subscriber until we can see that 10 actors have been created
    {
        var telemetry = await subscriber.Ask<TelemetrySubscriber.GetTelemetry>(TelemetrySubscriber.GetTelemetryRequest.Instance);
        Assert.Equal(12, telemetry.ActorCreated);
        // assert no restarts or stops recorded
        Assert.Equal(0, telemetry.ActorRestarted);
        Assert.Equal(0, telemetry.ActorStopped);
    }, RemainingOrDefault);
    
    // send a message to the router to restart all children
    router.Tell(new Broadcast(RestartChildren.Instance));

    // await assert collecting data from the telemetry subscriber until we can see that 10 actors have been restarted
    {
        var telemetry = await subscriber.Ask<TelemetrySubscriber.GetTelemetry>(TelemetrySubscriber.GetTelemetryRequest.Instance);
        // assert that actor start count is still 10
        Assert.Equal(12, telemetry.ActorCreated);
        Assert.Equal(10, telemetry.ActorRestarted);
        // assert no stops recorded
        Assert.Equal(0, telemetry.ActorStopped);
    }, RemainingOrDefault);
    
    // GracefulStop router actor and assert that 10 actors have been stopped
    await router.GracefulStop(RemainingOrDefault);
    {
        var telemetry = await subscriber.Ask<TelemetrySubscriber.GetTelemetry>(TelemetrySubscriber.GetTelemetryRequest.Instance);
        // assert that actor start count is still 10
        Assert.Equal(12, telemetry.ActorCreated);
        Assert.Equal(10, telemetry.ActorRestarted);
        Assert.Equal(10, telemetry.ActorStopped);
    }, RemainingOrDefault);
}
```

**Expected behavior**

Each child should be restarted individually and independently from each other.

**Actual behavior**

`Assert.Equal(10, telemetry.ActorRestarted);` should be true, but instead the `telemetry.ActorRestarted` value is 110!

That's because the following line:

```csharp
// send a message to the router to restart all children
  router.Tell(new Broadcast(RestartChildren.Instance));
```

Causes the router to restart all of its children 10 times apiece.

Aaronontheweb:
## Changes

Fixes #6295 - this fixes a _very_ old bug in Akka.NET where any time a routee on a `Pool` router restarted, the default supervision directive on the router was set to `Escalate` which would cause _the router itself_ to restart and _kill all of the routees in the process_. I believe this was done as a result of misunderstanding what Escalate does and how it works. We now know better and this sets the supervision directive to `Restart`, which is the safe default we use for everything else.

## Checklist

* [x] Design discussion issue #6295

### DIFF
--- a/src/core/Akka.Tests/Routing/RoutingSpec.cs
+++ b/src/core/Akka.Tests/Routing/RoutingSpec.cs
@@ -11,6 +11,7 @@ using System.Diagnostics;
 using System.Linq;
 using Akka.Configuration;
 using Akka.Actor;
+using Akka.Actor.Dsl;
 using Akka.Dispatch;
 using Akka.Routing;
 using Akka.TestKit;
@@ -293,27 +294,27 @@ namespace Akka.Tests.Routing
         }
 
         [Fact]
-        public void Routers_in_general_must_default_to_all_for_one_always_escalate_strategy()
+        public void Routers_in_general_must_default_to_all_for_one_restart_strategy()
         {
-            var restarter = new OneForOneStrategy(e =>
+            var props = Props.Create(() => new RestartActor(TestActor));
+            var routes = new[]
             {
-                TestActor.Tell(e);
-                return Directive.Restart;
-            });
+                Sys.ActorOf(props, "actor-0"),
+                Sys.ActorOf(props, "actor-1"),
+                Sys.ActorOf(props, "actor-2"),
+            };
 
-            var supervisor = Sys.ActorOf(Props.Create(() => new Supervisor(restarter)));
+            var router = Sys.ActorOf(new RoundRobinGroup(routes.Select(a => a.Path.ToString())).Props());
+            var restarted = new HashSet<string>();
 
-            supervisor.Tell(new RoundRobinPool(3).Props(Props.Create(() => new RestartActor(TestActor))));
-
-            var router = ExpectMsg<IActorRef>();
-            EventFilter.Exception<ArgumentException>("die").ExpectOne(() =>
+            for (var i = 0; i < 3; i++)
             {
                 router.Tell("die");
-            });
-            ExpectMsg<ArgumentException>().Message.Should().Be("die");
-            ExpectMsg("restarted");
-            ExpectMsg("restarted");
-            ExpectMsg("restarted");
+                ExpectMsg("restarted");
+                restarted.Add(LastSender.Path.Name);
+            }
+
+            restarted.Should().BeEquivalentTo("actor-0", "actor-1", "actor-2");
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


Arkatufus (approved):
LGTM

Aaronontheweb:
This will also need to be ported to v1.5 @Arkatufus

Aaronontheweb:
`Akka.Tests.Routing.RoutingSpec.Routers_in_general_must_default_to_all_for_one_always_escalate_strategy` - doh, missed that. I'm going to need to fix that test before this can be merged in.

Arkatufus:
OK

src/core/Akka.Tests/Routing/RoutingSpec.cs
```diff
@@ -293,27 +294,27 @@ public void Routers_in_general_must_set_supplied_supervisorStrategy_for_FromConf
...
-            });
-
-            var supervisor = Sys.ActorOf(Props.Create(() => new Supervisor(restarter)));
+                Sys.ActorOf(props, "actor-0"),
+                Sys.ActorOf(props, "actor-1"),
+                Sys.ActorOf(props, "actor-2"),
+            };
 
-            supervisor.Tell(new RoundRobinPool(3).Props(Props.Create(() => new RestartActor(TestActor))));
+            var router = Sys.ActorOf(new RoundRobinGroup(routes.Select(a => a.Path.ToString())).Props());
```
Aaronontheweb, line 307: @Arkatufus this needs to be a `RoundRobinPool` router

PR #6370, opened 2023-01-26
[PORT #6366] set default PoolRouter SupervisorStrategy to Restart

### LINKED ISSUE #6295: Pool routers use `Directive.Escalate` by default when supervising routees - this causes unintended consequences

Arkatufus:
Port of #6366

* close #6295 - set default PoolRouter SupervisorStrategy to Restart
* use the `SupervisionStrategy.DefaultStrategy`
* Fix unit test
* Change unit test to use RoundRobinPool router

(cherry-picked from ffd9a9e15e2841cabb7ac9e71db937c7ea320cbd)

### DIFF
--- a/src/core/Akka.Tests/Actor/ActorTelemetrySpecs.cs
+++ b/src/core/Akka.Tests/Actor/ActorTelemetrySpecs.cs
@@ -283,8 +283,7 @@ namespace Akka.Tests.Actor
                 // assert that actor start count is still 10
                 Assert.Equal(12, telemetry.ActorCreated);
                 
-                // bug due to https://github.com/akkadotnet/akka.net/issues/6295 - all routees and the router start each time
-                Assert.Equal(110, telemetry.ActorRestarted);
+                Assert.Equal(10, telemetry.ActorRestarted);
                 // assert no stops recorded
                 Assert.Equal(0, telemetry.ActorStopped);
             }, RemainingOrDefault);
@@ -296,7 +295,7 @@ namespace Akka.Tests.Actor
                 var telemetry = await subscriber.Ask<TelemetrySubscriber.GetTelemetry>(TelemetrySubscriber.GetTelemetryRequest.Instance);
                 // assert that actor start count is still 10
                 Assert.Equal(12, telemetry.ActorCreated);
-                Assert.Equal(110, telemetry.ActorRestarted);
+                Assert.Equal(10, telemetry.ActorRestarted);
                 Assert.Equal(11, telemetry.ActorStopped);
             }, RemainingOrDefault);
         }
--- a/src/core/Akka.Tests/Routing/RoutingSpec.cs
+++ b/src/core/Akka.Tests/Routing/RoutingSpec.cs
@@ -294,27 +294,20 @@ namespace Akka.Tests.Routing
         }
 
         [Fact]
-        public async Task Routers_in_general_must_default_to_all_for_one_always_escalate_strategy()
+        public void Routers_in_general_must_default_to_all_for_one_restart_strategy()
         {
-            var restarter = new OneForOneStrategy(e =>
-            {
-                TestActor.Tell(e);
-                return Directive.Restart;
-            });
+            var router = Sys.ActorOf(new RoundRobinPool(3).Props(Props.Create(() => new RestartActor(TestActor))));
+            var restarted = new HashSet<string>();
 
-            var supervisor = Sys.ActorOf(Props.Create(() => new Supervisor(restarter)));
-
-            supervisor.Tell(new RoundRobinPool(3).Props(Props.Create(() => new RestartActor(TestActor))));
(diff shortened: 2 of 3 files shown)


Aaronontheweb:
Ty!

src/core/Akka.Tests/Actor/ActorTelemetrySpecs.cs
```diff
@@ -296,7 +295,7 @@ public async Task ActorTelemetry_must_be_accurate_for_pool_router()
                 var telemetry = await subscriber.Ask<TelemetrySubscriber.GetTelemetry>(TelemetrySubscriber.GetTelemetryRequest.Instance);
                 // assert that actor start count is still 10
                 Assert.Equal(12, telemetry.ActorCreated);
-                Assert.Equal(110, telemetry.ActorRestarted);
+                Assert.Equal(10, telemetry.ActorRestarted);
```
Aaronontheweb, line 298: We will also need to backport this to v1.4 as well

Arkatufus: I wonder why we didn't caught this before... I guess we missed it amongst the racy unit test noise?

Arkatufus: added the unit test fix in #6376

PR #6371, opened 2023-01-27
Fix of EventFilter.And from Akka.testKit when one parameter is string (#6316).

### LINKED ISSUE #6316: Akka.TestKit: `EventFilter.And` does not correctly support typed parameters
**Version Information**
Version of Akka.NET? v1.4.47 (and probably most earlier versions)
Which Akka.NET Modules? Akka.TestKit

**Describe the bug**

Consider the following code sample:

```csharp
[Fact(DisplayName = "DeadLetterLoggingSpecs: Should log the underlying message, not the IWithTrace wrapper")]
public async Task ShouldLogTheUnderlyingMessage()
{
    // create a new SuspendableActor
    var actor = Sys.ActorOf(Props.Create<SuspendableActor>());
    Watch(actor);

    await EventFilter.DeadLetter<string>().And.DeadLetter<double>().ExpectAsync(2, () =>
    {
        // suspend the actor (via a blocking Task)
        actor.Tell(new SuspendableActor.Suspend());

        using (var newOp = Tracer.StartActiveSpan("startOp"))
        {
            // queue up dead letters (should be descendants of the current span)
            // this should force these messages to be wrapped inside a SpanEnvelope
            actor.Tell("hi");
            actor.Tell(1);
        }
        
        // sends a system message that will cause the actor to shutdown regardless of its mailbox state
        Sys.Stop(actor);
        
        // wait for the actor to shut down
        ExpectTerminated(actor);
    });

}
```

The `DeadLetter`s that should be received should be of type `string`, `int`, but the assertion checked for `string`, `bool`.

Worth noting: if I change the expected type on the first assertion away from `string` to something like `SuspendableActor` this assertion will fail.

**Expected behavior**

The `EventFilter` assertions should type check everything in a composed assertion.

**Actual behavior**

Only the first message in the `EventFilter` expression appears to be properly type-checked.

F0b0s:
Fixes #6316 

## Changes

The problem was in TestEventListenerClass because it calls `ShouldFilter` for deadLetter twice - first time with real parameter type and the second one with string param. So, for tests with `DeadLetter<string>` it might became false positive.

## Checklist

### DIFF
--- a/src/core/Akka.TestKit.Tests/TestEventListenerTests/DeadLettersEventFilterTests.cs
+++ b/src/core/Akka.TestKit.Tests/TestEventListenerTests/DeadLettersEventFilterTests.cs
@@ -5,6 +5,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Event;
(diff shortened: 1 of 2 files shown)


Aaronontheweb (approved):
LGTM - going to let CI run

PR #6378, opened 2023-01-31
Fix for issue #6377

### LINKED ISSUE #6377: Suppress warning for serializer configuration changes
**Problem description**
If the serializers default configuration is changed then a WARNING is written into the log to make sure if this is really meant to be changed. However if the change was indeed the intend, there is no possibility to switch off further notifications about this "problem" to keep focus on other possible unintended issues.

**Possible solution**
The desired solution would be to add a switch to disable this behavior. The default should be "on" (logging the possible issue at first sight).

**Considered alternatives**
The alternative of increasing the log level is not a reasonable solution (although you get rid of the serializer change messages) because this also hides notifications about other (unitended) issues, too.
I was not able to disable logging (through logger config) only for this class, but even if it would have worked... are there any other important notifications that could possibly swallowed without intention?

**Additional context**
The log messages to be suppressed are located within the  Serialization class at https://github.com/akkadotnet/akka.net/blob/8dfa8f85efd3c359a220d9b31499547aebb9fd11/src/core/Akka/Serialization/Serialization.cs#L333.

beachwalker:
* created a setting property LogSerializerOverrideOnStart to reflect if warning should be logged or not
* updated default config to enable logging for this issue by default
* added property to the configuration specification

Fixes #6377

## Changes

Allows to suppress the WARNING message about changes in the serializer settings.

Set the value in the config
`log-serializer-override-on-start = off`
to disable the logging of the message.
The default value is `on` to get notified at the first time and check if the change was by intention.

## Checklist

* [x] Design discussion issue #
It is worth to mention that the solution has the shortcomming of a global enable/disable behavior. It would be nice to just mark the intended change (for example by having a white list of checked changes configured in a mapping table (type <-> serializer)) but this seems to be an overkill for the rare cases to me.
I only added the change to the default config and documented the code (according the other configuration properties documentation). However, I wrote no Markdown files or so. Please ask, if required.

The changes do not affect the performance, so I added no benchmarks.

### DIFF
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.Core.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.Core.verified.txt
@@ -1721,6 +1721,7 @@ namespace Akka.Actor
         public bool LogDeadLettersDuringShutdown { get; }
         public System.TimeSpan LogDeadLettersSuspendDuration { get; }
         public string LogLevel { get; }
+        public bool LogSerializerOverrideOnStart { get; }
         public bool LoggerAsyncStart { get; }
         public System.TimeSpan LoggerStartTimeout { get; }
         public System.Collections.Generic.IList<string> Loggers { get; }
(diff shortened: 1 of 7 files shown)


beachwalker:
image

... no idea what is wrong here after trying to run the tests again. :-(

Aaronontheweb (approved):
LGTM

Issue #5522, opened 2022-01-19
[PERF] Reduce Akka.Persistence.Journal allocations + overhead


Aaronontheweb:
**Version Information**
Version of Akka.NET? 1.4.32
Which Akka.NET Modules? Akka.Persistence

**Describe the performance issue**
The current `AsyncWriteJournal` implementation needs some house cleaning in order to reduce excess allocations in the form of:

1. Boxing
2. Unnecessary array / collection copying
3. Delegate allocation
4. Possibly, inefficient LINQ usage (i.e. snapshotting with `ToList()` calls)

**Data and Specs**
Some examples pointed out by @to11mtm 

https://github.com/akkadotnet/akka.net/pull/5505#discussion_r783996956 
https://github.com/akkadotnet/akka.net/pull/5505#discussion_r783998326
https://github.com/akkadotnet/akka.net/pull/5505#discussion_r784008777

**Additional context**
Can use https://github.com/akkadotnet/akka.net/blob/dev/src/benchmark/Akka.Cluster.Benchmarks/Persistence/JournalWriteBenchmarks.cs to measure the improvements

PR #6412, opened 2023-02-17
Delete unnecessary / bad Sql.Common.Journal subscriptions

Aaronontheweb:
## Changes

All of the old SubscribeTo___ methods from the Akka.Persistence.Sql.Common journal implement some bad practices, i.e. not able to detect events published in other instances of the Akka.NET processes targeting the same journal. This method deletes them and marks the old queries in Sql.Common as `Obsolete`. Rewrote existing queries to poll the database rather than rely on in-memory events, which is the correct way to implement these.

Also: significantly reduces allocations and overheads for writes for all Akka.Persistence.Sql.Common implementations.

Related: https://github.com/akkadotnet/akka.net/issues/5522

## Checklist

* [x] Design discussion issue #5522

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
@@ -25,9 +25,6 @@ namespace Akka.Persistence.Sql.Common.Journal
     /// </summary>
     public abstract class SqlJournal : AsyncWriteJournal, IWithUnboundedStash
     {
-        private ImmutableDictionary<string, IImmutableSet<IActorRef>> _persistenceIdSubscribers = ImmutableDictionary.Create<string, IImmutableSet<IActorRef>>();
-        private ImmutableDictionary<string, IImmutableSet<IActorRef>> _tagSubscribers = ImmutableDictionary.Create<string, IImmutableSet<IActorRef>>();
-        private readonly HashSet<IActorRef> _newEventsSubscriber = new HashSet<IActorRef>();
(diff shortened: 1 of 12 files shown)


src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistenceSqlCommon.Core.verified.txt
```diff
@@ -91,9 +91,6 @@ namespace Akka.Persistence.Sql.Common.Journal
         protected virtual string ByPersistenceIdSql { get; }
         protected virtual string ByTagSql { get; }
         protected virtual string DeleteBatchSql { get; }
-        protected bool HasNewEventsSubscribers { get; }
```
Aaronontheweb, line 94: No longer need these defined on the journal implementations. Could add these back to be non-breaking, but they're obsolete now regardless and not used anywhere inside the journal.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
```diff
@@ -91,6 +74,10 @@ protected override bool ReceivePluginInternal(object message)
         {
             switch (message)
             {
+                case SelectCurrentPersistenceIds msg:
```
Aaronontheweb, line 77: Added support for proper database polling here.

Aaronontheweb (commented):
Reviewed my changes

Arkatufus (approved):
LGTM

Aaronontheweb:
FWIW, the Akka.Persistence.Sqlite.Tests project covers all of these changes.

Issue #6404, opened 2023-02-14
Akka.Persistence.Query: need backpressure control for queries


Aaronontheweb:
**Is your feature request related to a problem? Please describe.**

Here's a problem that affects users trying to run large-scale CQRS with Akka.NET: Akka.Persistence.Query does not currently constrain the number of parallel database connections being opened on any implementation of Akka.Persistence, therefore it's quite ease to instantly melt even a well-provisioned Postgres instance if each node is running thousands of `EventsByEntyId` queries, all of which poll at a rate of once per-second.

Akka.Persistence `Recovery` and writes both regulate the number of open parallel connections through the internals of the `AsyncWriteJournal` actor and in order to solve this very problem from occurring inside normal Akka.Persistence operations. We should incorporate an equivalent mechanism for adding backpressure support to Akka.Persistence.Query as well.

**Describe the solution you'd like**

I'd probably use Akka.Streams to do this - a pipeline with a fixed number of parallel queries (`SelectAsync(degreeOfParallelism`) which queues any pending query request until the ones ahead of it have completed. No need to worry about ordering here - a single requestor for a large, paginated query has to send its requests to the journal in chunks anyway. 

**Describe alternatives you've considered**

We might have to implement this on a per-journal basis, as there's no unified implementation for handling them currently: 

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Persistence.Query/PersistenceQuery.cs

We could try to develop some unified APIs around this but that might need to occur after the initial 1.5 release.

Aaronontheweb:
# Design

In order to rate-limit the number of open database connections at any given time, I'd propose we put a `ReadJournal` "front-end" in front of the `AsyncWriteJournal` implementation :

```mermaid
flowchart TD
    A[QuerySubmitted] -->|Offset,QueryType,Tag or EntityId| B(ReadJournal.ReceiveQuery)
    B --> |Buffer into Akka.Stream|C(SelectAsync w/ Max Parallelism = N)
    C --> D{SelectAsync}
    D --> |Capacity Avaiable| E(Query to AsyncWriteJournal)
    D --> |Capacity Unavailable - Backpressure| D
```

The `SelectAsync` is key - we're basically implementing a simple token bucket throttler that weights all queries the same. This likely means:

* All journal implementation have to have some sort of custom acknowledgement message or
* We bake a standard one into Akka.Persistence.Query that all journal implementations have to support in order to fulfill the Akka.Persistence.TCK.

[...]

PR #6417, opened 2023-02-20
Akka.Persistence.Query.Sql backpressure support

Aaronontheweb:
## Changes

Implementation of https://github.com/akkadotnet/akka.net/issues/6404 for Akka.Persistence.Query.Sql

## Checklist

* [x] Design discussion issue #6404


Aaronontheweb:
Doh! Meant to make this as a draft PR.

src/contrib/persistence/Akka.Persistence.Query.Sql/AllEventsPublisher.cs
```diff
@@ -25,11 +25,11 @@ public sealed class Continue
             private Continue() { }
         }
 
-        public static Props Props(long fromOffset, TimeSpan? refreshInterval, int maxBufferSize, string writeJournalPluginId)
+        public static Props Props(long fromOffset, TimeSpan? refreshInterval, int maxBufferSize, IActorRef writeJournal)
```
Aaronontheweb, line 28: I made this change on all of the `Query` implementations - for the simple reason that we're going to swap out the real underlying `JournalRef` with an `IActorRef` that points to the `SqlQueryThrottler` once this change is introduced. Injecting the `IActorRef` into the constructor of each query instead of having each query resolve the journal ref themselves makes it possible to throttle all queries launched by the same read journal plugin.

src/contrib/persistence/Akka.Persistence.Query.Sql/AllPersistenceIdsPublisher.cs
```diff
@@ -169,7 +170,7 @@ protected override bool Receive(object message)
                 case Request _:
                     Become(Waiting);
                     _journalRef
-                        .Ask<CurrentPersistenceIds>(new SelectCurrentPersistenceIds(_lastOrderingOffset, Self))
+                        .Ask<CurrentPersistenceIds>(new SelectCurrentPersistenceIds(_lastOrderingOffset, Self), DefaultQueryTimeout)
```
Aaronontheweb, line 173: Added default timeouts on all `Ask<T>` operations to the throttler / journal in order to make queries responsive in the event that the journal fails.

src/contrib/persistence/Akka.Persistence.Query.Sql/SqlQueryConstants.cs
```diff
@@ -0,0 +1,19 @@
...
+using System;
+
+namespace Akka.Persistence.Query.Sql
+{
+    /// <summary>
+    /// INTERNAL API
+    /// </summary>
+    internal static class SqlQueryConstants
+    {
+        public static readonly TimeSpan DefaultQueryTimeout = TimeSpan.FromSeconds(10);
```
Aaronontheweb, line 17: Can make this configurable - haven't yet.

src/contrib/persistence/Akka.Persistence.Query.Sql/EventsByPersistenceIdPublisher.cs
```diff
@@ -109,6 +106,7 @@ protected bool Idle(object message)
         protected void Replay()
         {
             var limit = MaxBufferSize - Buffer.Length;
+            // TODO: use Ask<T> here to limit time similar to other queries?
```
Aaronontheweb, line 109: Right now the time to complete a recovery is not bounded by the caller for `_byPersistentId` queries - should we change that?

src/contrib/persistence/Akka.Persistence.Query.Sql/SqlQueryThrottler.cs
```diff
@@ -0,0 +1,66 @@
...
+            Receive<IJournalRequest>(req =>
+            {
+                _throttler.Tell((req, Sender));
+            });
+        }
+        
+        protected override void PreStart()
+        {
+            var materializer = Context.Materializer();
+            var (actor, source) = Source.ActorRef<(IJournalRequest req, IActorRef sender)>(1000, OverflowStrategy.DropHead)
```
Aaronontheweb, line 41: Can store up to 1000 concurrent queries - oldest queries (most likely to be timed out) are dropped first.

Aaronontheweb: We can make this larger or configurable.

src/contrib/persistence/Akka.Persistence.Query.Sql/SqlQueryThrottler.cs
```diff
@@ -0,0 +1,66 @@
...
+                .PreMaterialize(materializer);
+            _throttler = actor;
+
+            source
+                .SelectAsyncUnordered(_maxConcurrentQueries, async request =>
```
Aaronontheweb, line 46: SelectAsyncUnordered for min latency here - query invocation order is determined by the caller, i.e. the same "EventsBy___" query can't kick off more than 1 request in a row - so each distinct query sitting inside this filter must come from a separate query instance. Therefore ordering does not matter - deliver results as fast as possible.

src/contrib/persistence/Akka.Persistence.Query.Sql/SqlReadJournal.cs
```diff
@@ -37,21 +38,28 @@ public static Config DefaultConfiguration()
         }
         
         private readonly TimeSpan _refreshInterval;
-        private readonly string _writeJournalPluginId;
         private readonly int _maxBufferSize;
+        private readonly int _maxConcurrentQueries = 30; // TODO: make configurable
```
Aaronontheweb, line 42: Need to make the max concurrent queries setting configurable.

Aaronontheweb (commented):
Detailed changes so far.

Aaronontheweb:
Added a "crush test" to apply maximum stress to Akka.Persistence.SqlServer: https://github.com/Aaronontheweb/AkkaSqlQueryCrushTest

1000 recoveries, 1000 * 10 writes, 1000 * 10 persistence queries doing 1 event per second

Using Akka.NET v1.4.49 and latest for Akka.Persistence.SqlServer:

```
[INFO][2/20/2023 11:24:09 PM][Thread 0054][akka://SqlSharding/user/recovery-tracker] Recovery complete - took 76855.5798ms
```

src/contrib/persistence/Akka.Persistence.Query.Sql/EventsByPersistenceIdPublisher.cs
Aaronontheweb: So it turns out @Arkatufus that this is a major vulnerability in Akka.Persistence.Query.Sql currently - anything that goes wrong between here and the journal causes queries to die. Relying on the journal to do the right thing is too weak a guarantee. We have to use `Ask<T>` here and explicitly time out the request.

Aaronontheweb: Was able to reproduce this in the crush test just now

Aaronontheweb: Fixed this for two of the queries -have two more that need fixing too....

Aaronontheweb:
Superseded via https://github.com/akkadotnet/akka.net/pull/6436

PR #6420, opened 2023-02-21
[TESTKIT] Reintroduce old code and mark them obsolete

Arkatufus:
Reintroduce `Dispose` and `AfterAll` method to `TestKit`

### DIFF
--- a/src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
+++ b/src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
@@ -21,7 +21,7 @@ namespace Akka.TestKit.Xunit2
     /// This class represents an Akka.NET TestKit that uses <a href="https://xunit.github.io/">xUnit</a>
     /// as its testing framework.
     /// </summary>
-    public class TestKit : TestKitBase, IAsyncLifetime
+    public class TestKit : TestKitBase, IAsyncLifetime, IDisposable
     {
         private class PrefixedOutput : ITestOutputHelper
         {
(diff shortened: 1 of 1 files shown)


src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
```diff
@@ -168,7 +179,24 @@ public virtual Task InitializeAsync()
...
+        /// Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
+        /// </summary>
+        /// <param name="disposing">
+        /// if set to <c>true</c> the method has been called directly or indirectly by a  user's code.
+        /// Managed and unmanaged resources will be disposed.<br /> if set to <c>false</c> the method
+        /// has been called by the runtime from inside the finalizer and only unmanaged resources can
+        ///  be disposed.
+        /// </param>
+        [Obsolete("Dispose(bool) is deprecated, please use DisposeAsync() instead")]
+        protected virtual void Dispose(bool disposing)
```
Aaronontheweb, line 198: lgtm

Aaronontheweb: Don't we still need to implement `IDisposable` on this class?

Aaronontheweb: Yes, we do https://github.com/akkadotnet/akka.net/blob/a45501991bd366a4aa7db60347760332b6cd30c4/src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs#L22

Aaronontheweb (changes requested):
Still need to add `IDisposable` back and do API approvals

src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
Arkatufus: Done

Aaronontheweb (approved):

Aaronontheweb (approved):
LGTM

PR #6422, opened 2023-02-21
Bump MultinodeTestAdapter to 1.5.0-beta1

Arkatufus:


Aaronontheweb (approved):
LGTM - let's see if we run into any issues 😬

PR #6426, opened 2023-02-21
Harden `Option<T>` by disallowing null value

### LINKED ISSUE #4604: Need to harden `Option<T>` to avoid gotcha bugs
The current implementation of `Option<T>` is lacking consistency and creates more problem than it actually solves.
1. It allows for a Some(null), which is, contractually broken (should be partially fixed in #4603).
2. It doesn't differentiate between Some and None, all we can go on with is to trust the `HasValue` property.
3. It allows access to null Value, which defeats the purpose of avoiding null reference propagation in the first place
    - Proper implementation is to stop null propagation immediately by throwing an exception

All in all, this allows for null to propagate inspite the use of `Option<T>` as a guard.

There are a few improvement options that can be taken:
1. (#4603) Assign `false` to `HasValue` if the passed value is null.
    This is the simplest "fix", it still doesn't differentiate between None, Some(null), and Some(T). You can still access the null value if you're not being careful and check for `Option<T>.HasValue()`. Might propagate into unexpected bug somewhere else. Can be slightly mitigated by:
2. Throw an exception if a null value is being accessed.
    This will stop any further null propagation. This guard will warn any possible code problem _during runtime_, making bug fixes faster. The problem is that, the more burried and rare a code is, the longer it will take to finally catch this exception, and it might happen in a critical timing during runtime.
3. Throw an exception if a null is passed into the contructor.
    This will guarantee that an `Option<T>` will always contain a value, unless it is reference equal to the `None` instance, the only `Option<T>` instance allowed to have a null value, which will throw if accessed. Slightly higher bug catch rate compared to option 2, but still can create problem down the road.
4. Replace `Option<T>` struct with `IOption<T>` interface and `Some<T>` and `None<T>` concrete class.
    - This is akin to open invasive radical surgery. All in all, it has the same benefit as option 3, with the added benefit that we can use switch...case based on the type returned by a function, exactly like coding in scala, making porting a lot easier.
    - This is how `Option<T>` is implemented in scala.
    - Actually works on primitive types since `Some` and `None` does not depend on a null value to define the state of `Option`
    - Some and None are clearly defined by the class type, and depending on how draconian we want to be, we can restrict certain methods to the concrete class, making sure that the user conciously aware of what they are using when coding.
    - **This is highly disruptive**, and **will cause** a major rewrite on all code paths that uses `Option<T>` as it is not a simple copy-paste and/or search-replace fix. On the flip side, this will make those code more robust and might uncover previously uncaught bugs in the system.
5. Improve upon option 4 and only allow creation of `IOption<T>` instances through a factory pattern.
    - This will slightly improve the experience of using `IOption<T>`, since the instance factory can automate `Some` or `None` instantiation decision based on the value passed into it. Can be a bit tricky to support primitive types.

Arkatufus:
Fixes #4604

Hardens Option<T>

## Changes

* Make `Option<T>` constructor throw an `ArgumentNullException` when null value is passed in
* Add a `Create` static method that returns a `None` when null is passed in

### DIFF
--- a/src/core/Akka/Util/Option.cs
+++ b/src/core/Akka/Util/Option.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 namespace Akka.Util
 {
@@ -18,6 +19,7 @@ namespace Akka.Util
     /// <typeparam name="T">TBD</typeparam>
     public readonly struct Option<T>
     {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static Option<T> Create(T value)
 #pragma warning disable CS0618
             => value is null ? None : new Option<T>(value);
--- a/src/core/Akka/Util/Option.cs
+++ b/src/core/Akka/Util/Option.cs
@@ -18,19 +18,24 @@ namespace Akka.Util
     /// <typeparam name="T">TBD</typeparam>
     public readonly struct Option<T>
     {
+        public static Option<T> Create(T value)
+#pragma warning disable CS0618
+            => value is null ? None : new Option<T>(value);
+#pragma warning restore CS0618
+
         /// <summary>
         /// None.
         /// </summary>
         public static readonly Option<T> None = new Option<T>();
 
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="value">TBD</param>
+        [Obsolete("Use Option<T>.Create() instead")] 
         public Option(T value)
         {
+            if (value is null)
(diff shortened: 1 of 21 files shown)


src/core/Akka/Util/Option.cs
```diff
@@ -50,7 +58,7 @@ public Option(T value)
         /// </summary>
         /// <param name="value">The object to convert</param>
         /// <returns>The result of the conversion.</returns>
-        public static implicit operator Option<T>(T value) => new Option<T>(value);
+        public static implicit operator Option<T>(T value) => Create(value);
```
Aaronontheweb, line 61: LGTM

Aaronontheweb: We should enable `nullable` in the future for this project - there's ways to do it in increments.

Aaronontheweb (approved):
LGTM

PR #6427, opened 2023-02-21
move `DateTime` / `TimeSpan` extension APIs out of Akka.Util and into Akka.Cluster.Metrics

Aaronontheweb:
## Changes

Moves a bunch of Akka.Cluster.Metrics-specific junk out of Akka.Util and into the projects where they're actually used. Also marked as `internal` so we don't hijack common types.

## Checklist


Arkatufus (approved):
LGTM

Aaronontheweb:
Looks like I broke some Akka.Cluster.Metrics specs. I'll take a look at that.

PR #6429, opened 2023-02-21
Remove unsafe `implicit` conversion operators in `AtomicBoolean` and `AtomicReference<T>`

### LINKED ISSUE #6392: Dangerous conversion operator in AtomicBoolean.cs
https://github.com/akkadotnet/akka.net/blob/192e9a886838ccb678be7287733a7cf5f6a983ad/src/core/Akka/Util/AtomicBoolean.cs#L91

I understand why it was added. It enables direct assignment to an AtomicBoolean variable.

e.g. 
```cs
AtomicBoolean myBool = true;
myBool = false;
// etc.
```

But I would consider conversion from bool to AtomicBool to be dangerous. this is basically a new allocation for the AtomicBool class itself, and this allocation isn't easily seen by other threads, as far as I understand. Although I do see it is generally allocated in the same place, and that we would be generally fine, but I don't see the C# standard guaranteeing it
Instead what should be promoted is to assign directly to the property of the original class.

```cs
AtomicBoolean myBool = new AtomicBoolean(true);
myBool.Value = false;
// etc.
```

By removing this dangerous implicit operator. This way you don't promote dangerous behaviour

Aaronontheweb:
## Changes

Fixes #6392 - we weren't using these conversion operators in a lot of places and it's better to be explicit about accessing / setting the values on these synchronized types.

## Checklist

* [x] Design discussion issue #6392


Aaronontheweb:
looks like this change introduced some hard failures

src/core/Akka/Util/AtomicBoolean.cs
```diff
@@ -70,30 +70,6 @@ public bool GetAndSet(bool newValue)
...
-        #region Conversion operators
-
-        /// <summary>
-        /// Performs an implicit conversion from <see cref="AtomicBoolean"/> to <see cref="System.Boolean"/>.
-        /// </summary>
-        /// <param name="atomicBoolean">The boolean to convert</param>
-        /// <returns>The result of the conversion.</returns>
-        public static implicit operator bool(AtomicBoolean atomicBoolean)
-        {
-            return atomicBoolean.Value;
```
Arkatufus, line 83: Shouldn't this be fine?

src/core/Akka/Util/AtomicReference.cs
```diff
@@ -74,30 +74,6 @@ public T GetAndSet(T newValue)
...
-        #region Conversion operators
-
-        /// <summary>
-        /// Performs an implicit conversion from <see cref="AtomicReference{T}"/> to <typeparamref name="T"/>.
-        /// </summary>
-        /// <param name="atomicReference">The reference to convert</param>
-        /// <returns>The result of the conversion.</returns>
-        public static implicit operator T(AtomicReference<T> atomicReference)
-        {
-            return atomicReference.Value;
```
Arkatufus, line 87: This should also be fine?

Arkatufus (changes requested):

src/core/Akka/Util/AtomicBoolean.cs
Aaronontheweb: Yeah I guess so - that's probably safe. I can add that one back.

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -264,7 +264,7 @@ public void Join(Address address)
         /// <returns>Task which completes, once current cluster node reaches <see cref="MemberStatus.Up"/> state.</returns>
         public Task JoinAsync(Address address, CancellationToken token = default)
         {
-            if (_isTerminated)
+            if (_isTerminated.Value)
```
Arkatufus, line 267: This can be avoided if implicit casting of AtomicBool -> bool is allowed. Would minimize the amount of work we need to do to fix this API breaking change

Arkatufus (changes requested):

src/core/Akka/Util/AtomicBoolean.cs
Aaronontheweb: Done

src/core/Akka.Cluster/Cluster.cs
Aaronontheweb: (the work is already done though)

Arkatufus (approved):
LGTM

PR #6439, opened 2023-02-24
Remove IWithBoundedStash

Arkatufus:
## Changes
* Remove IWithBoundedStash


Aaronontheweb:
cc @ismaelhamed on this one - not sure if his stash changes would be adversely impacted by this.

src/core/Akka/Actor/Stash/IWithStash.cs
```diff
@@ -30,7 +30,7 @@ namespace Akka.Actor
     /// that does not enforce any mailbox type see <see cref="IWithUnrestrictedStash"/>.
     /// </para>
     /// </summary>
-    public interface IWithStash : IWithUnrestrictedStash, IRequiresMessageQueue<IDequeBasedMessageQueueSemantics>
+    public interface IWithStash : IWithUnrestrictedStash, IRequiresMessageQueue<IBoundedDequeBasedMessageQueueSemantics>
     {
```
ismaelhamed, line 34: Default deque must be unbounded, this is change in behavior. 

Keep in mind that when using these marker interfaces, the mailbox type will be resolved via the `akka.actor.mailbox.requirements` configuration in the hocon, not via the `StashFactory` (maybe that's the confusion).

Arkatufus: `IWithStash` is the replacement for `IWithBoundedStash` interface, the unbounded behaviour is `IWithUnboundedStash`

Aaronontheweb: Yeah this is an unsafe change.

Aaronontheweb: Oh wait, are these the new APIs @ismaelhamed added?

Arkatufus: Ah, I get it now... IWithStash is supposed to replace both `IWithBoundedStash` and `IWithUnboundedStash` and the behaviour is dictated by the settings?

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.Core.verified.txt
```diff
@@ -1178,9 +1178,7 @@ namespace Akka.Actor
         void Become(Akka.Actor.UntypedReceive receive);
         void BecomeStacked(Akka.Actor.UntypedReceive receive);
     }
-    [System.ObsoleteAttribute("Use `IWithStash` with a configured BoundedDeque-based mailbox instead.")]
```
Arkatufus, line 1181: Obsolete since 1.5.0-alpha4

PR #6436, opened 2023-02-24
Akka.Persistence.Query Throttling implementation - "QueryPermitter"

Aaronontheweb:
## Changes

Implements https://github.com/akkadotnet/akka.net/issues/6404 using the same design as Akka.Persistence's base `RecoveryPermitter`. Gets used on every single iteration of each query.

Related: https://github.com/akkadotnet/akka.net/pull/6417, https://github.com/akkadotnet/akka.net/pull/6433

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [x] Design discussion issue #6404
* [ ] I have added website documentation for this feature.

### Latest `dev` Benchmarks 

Include data from the [relevant benchmark](https://getakka.net/community/contributing/index.html#improve-performance) prior to this change here.

### This PR's Benchmarks

Include data from after this change here.

PR #6440, opened 2023-02-24
Akka.Streams: memory-optimize `ActorMaterializer` HOCON injection

Aaronontheweb:
## Changes

creating lots of materializer instances (i.e. in actors that own their own streams) results in a tremendous amount of unreclaimable HOCON-related memory - on the order of 11Gb in https://github.com/Aaronontheweb/AkkaSqlQueryCrushTest

This tries to avoid injecting the top level Akka.Streams fallback every time a materializer is used. In addition to that, we try to cache the default materializer settings parsed from the `ActorSystem`s HOCON if none are provided.

## Checklist


Aaronontheweb (commented):
My inline comments are self-explanatory

Aaronontheweb:
This should probably be back-ported to v1.4, C#11 syntax not withstanding.

Aaronontheweb:
For reference 

image

Arkatufus (approved):
LGTM, do we have any before and after memory profile?

Aaronontheweb:
> LGTM, do we have any before and after memory profile?

I'll need to rebuild this and integrate it with https://github.com/akkadotnet/akka.net/pull/6436 to do that, and remove a materializer optimization I also made there.

Arkatufus:
Ok, the changes are benign so we can merge them anyway

Aaronontheweb:
According to test suite, this for sure broke something

Aaronontheweb:
Ah, the `static` is the issue - doesn't work when there's multiple `ActorSystem`s per process.

src/core/Akka.Streams/ActorMaterializer.cs
```diff
@@ -42,18 +44,18 @@ public static Config DefaultConfig()
...
         #region static
-        
+
         /// <summary>
         /// Injecting the top-level Materializer HOCON configuration over and over again is expensive, so we want to avoid
         /// doing it each time a materializer is instantiated. This flag will be set to true once the configuration has been
         /// injected the first time.
         /// </summary>
-        private static volatile bool _injectedConfig = false;
+        private static readonly ConcurrentDictionary<ActorSystem, bool> InjectedConfig = new();
```
Aaronontheweb, line 53: This would be unnecessary if the materializer were an `ActorSystemExtension`, but since it's not - have to do it this way.

src/core/Akka.Streams/ActorMaterializer.cs
```diff
@@ -87,16 +89,28 @@ public static ActorMaterializer Create(IActorRefFactory context, ActorMaterializ
...
 
-            if(!_injectedConfig)
+            if(!InjectedConfig.TryGetValue(system, out _) && InjectedConfig.TryAdd(system, true))
             {
                 // Inject the top-level fallback config for the Materializer once, and only once.
                 // This is a performance optimization to avoid having to do this on every materialization.
                 system.Settings.InjectTopLevelFallback(DefaultConfig());
-                _injectedConfig = true;
+
+                static async Task CleanUp(ActorSystem sys)
```
Aaronontheweb, line 98: Added a static delegate to cleanup `ActorSystem` references after shut down

Aaronontheweb (commented):
Reviewed changes

Aaronontheweb:
@Arkatufus re-ran my crush test with these changes - the materializer no longer even appears in the memory footprint now.

ismaelhamed:
@Aaronontheweb , in the JVM they have this [default system wide materializer](https://github.com/akka/akka/pull/27491) which is the recommended, since you typically only need one materializer per application.

Aaronontheweb:
> @Aaronontheweb , in the JVM they have this [default system wide materializer](https://github.com/akka/akka/pull/27491) which is the recommended, since you typically only need one materializer per application.

Yeah, I think we should go ahead and just do that too.

PR #6443, opened 2023-02-24
Remove obsolete methods from Akka.DistributedData

Arkatufus:


src/contrib/cluster/Akka.DistributedData/ReplicatorSettings.cs
```diff
@@ -166,105 +166,6 @@ public static ReplicatorSettings Create(Config config)
         /// </summary>
         public bool VerboseDebugLogging { get; }
 
-        [Obsolete]
-        public ReplicatorSettings(string role,
```
Arkatufus, line 170: Obsolete since v1.4.22 (2 years)

src/contrib/cluster/Akka.DistributedData/ReplicatorSettings.cs
```diff
@@ -166,105 +166,6 @@ public static ReplicatorSettings Create(Config config)
...
-                  durableStoreProps,
-                  pruningMarkerTimeToLive,
-                  durablePruningMarkerTimeToLive,
-                  maxDeltaSize,
-                  false
-                 )
-        {
-        }
-
-        [Obsolete]
```
Arkatufus, line 200: Obsolete since v1.4.22 (2 years)

Aaronontheweb:
This is probably fine - I doubt many people manually spin up `ReplicatorSettings` without parsing it from the `ActorSystem` first anyway.

PR #6446, opened 2023-02-24
Remove obsolete methods from Akka.Persistence.Query

Arkatufus:


src/core/Akka.Persistence.Query/EventEnvelope.cs
```diff
@@ -20,14 +20,6 @@ namespace Akka.Persistence.Query
     /// </summary>
     public sealed class EventEnvelope : IEquatable<EventEnvelope>
     {
-        /// <summary>
-        /// Initializes a new instance of the <see cref="EventEnvelope"/> class.
-        /// </summary>
-        [Obsolete("For binary compatibility with previous releases")]
-        public EventEnvelope(Offset offset, string persistenceId, long sequenceNr, object @event)
```
Arkatufus, line 27: Obsolete since v1.4.14

Aaronontheweb: I'm going to lean towards keeping this one.

PR #6448, opened 2023-02-24
Append message content to deadletter log message

Arkatufus:
## Changes

Quality of life improvement, append the message content to the end of deadletter log messages

### DIFF
--- a/src/core/Akka/Event/DeadLetterListener.cs
+++ b/src/core/Akka/Event/DeadLetterListener.cs
@@ -200,12 +200,19 @@ namespace Akka.Event
                     $"If this is not an expected behavior then {d.Recipient} may have terminated unexpectedly. ";
                     break;
             }
-            _eventStream.Publish(new Info(
-                d.Recipient.Path.ToString(),
-                d.Recipient.GetType(),
-                logMessage +
+
+            logMessage +=
                 "This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' " +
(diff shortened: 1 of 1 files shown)


src/core/Akka/Event/DeadLetterListener.cs
```diff
@@ -200,12 +200,19 @@ private void LogDeadLetter(AllDeadLetters d, string doneMsg)
...
+                "and 'akka.log-dead-letters-during-shutdown'.";
+
+            // Check that unwrapped object has an overriden ToString() method
+            var content = unwrapped?.ToString() ?? "null";
+            if (!content.Equals(messageStr))
+            {
+                logMessage += $" Message content: {content}";
+            }
+            
+            _eventStream.Publish(new Info(d.Recipient.Path.ToString(), d.Recipient.GetType(), logMessage));
```
Aaronontheweb, line 215: Add this to the original message we log with `DeadLetter`s, rather than as a separate log message.

Aaronontheweb (changes requested):

Aaronontheweb (approved):

src/core/Akka/Event/DeadLetterListener.cs
Aaronontheweb: Nevermind - I misread this.

Aaronontheweb:
We'll want to backport this to v1.4 @Arkatufus

PR #6462, opened 2023-02-27
Remove obsolete methods from Akka.Actor

Arkatufus:
Some "obsolete" methods, classes, and properties are not removed:
* `IInternalActorRef.IsTerminated`: Ingrained inside Akka.NET, requires extensive surgery to remove.
* `ActorCell.GetChildren()`: Removed the obsolete attribute, the replacement methods was never implemented (since v1.1.0).
* `Akka.Actor.Failure`: Ingrained inside Akka.NET, requires extensive surgery to remove.
* `ActorProducerPipeline`: Promised to be removed in v1.5, but replacement was never implemented


Aaronontheweb (dismissed):
LGTM

Aaronontheweb (changes requested):
Needs API approval, but I don't want anymore of these PRs today anyways. Closing.

PR #6465, opened 2023-02-27
copy `Obsolete` attributes to Akka.Persistence.TCK

Aaronontheweb:
## Changes

copy `Obsolete` attributes to Akka.Persistence.TCK - same changes as we made to the TestKit base classes for Xunit.

## Checklist

### DIFF
--- a/src/core/Akka.Persistence.TCK/PluginSpec.cs
+++ b/src/core/Akka.Persistence.TCK/PluginSpec.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Actor.Setup;
 using Akka.Configuration;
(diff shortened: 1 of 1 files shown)


Arkatufus (approved):
LGTM

src/core/Akka.Persistence.TCK/PluginSpec.cs
```diff
@@ -65,19 +66,21 @@ public void Subscribe<T>(IActorRef subscriber)
             Sys.EventStream.Subscribe(subscriber, typeof (T));
         }
 
-        
-        public void Dispose()
```
Aaronontheweb, line 69: Do we need to add this back?

Arkatufus:
Hold on, I think I smelled something fishy in the code

Arkatufus:
The old code was actually wrong, `Dispose()` and `Dispose(bool)` methods are already implemented in the base class here: https://github.com/akkadotnet/akka.net/blob/c7ac23b360c5bcc2826729536cc43ef1d5a475da/src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs#L197-L206

PR #6471, opened 2023-02-28
bug fix: timing regressions in `Cluster.JoinAsync` methods

Aaronontheweb:
## Changes

We used the `cluster.seed-node-timeout` property incorrectly in https://github.com/akkadotnet/akka.net/pull/6033 - that is a "how long did it take from me to hear back from a seed" setting, but we're using it like a "how much time do I have to join the cluster?" setting.

Added a function designed to give us at least 20s of leeway if the user does not provide a `CancellationToken` of their own.

If the user does provide their own `CancellationToken`, that should be the single source of truth for this operation - second-guessing it without own un-overridable timeout adds zero value.

## Checklist

### DIFF
--- a/src/core/Akka.Cluster/Cluster.cs
+++ b/src/core/Akka.Cluster/Cluster.cs
@@ -244,6 +244,16 @@ namespace Akka.Cluster
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Aaronontheweb (commented):
Detailed changes

PR #6475, opened 2023-03-01
Remove IAsyncLifetime from TestKit

Arkatufus:


src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
```diff
@@ -50,6 +50,9 @@ public void WriteLine(string format, params object[] args)
         /// </summary>
         protected readonly ITestOutputHelper Output;
 
+        private bool _disposed;
+        private bool _disposing;
```
Aaronontheweb, line 54: Don't need this

src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
```diff
@@ -122,26 +125,9 @@ public TestKit(string config, ITestOutputHelper output = null)
...
-            await ShutdownAsync();
-        }
-        
         /// <summary>
         /// This method is called when a test ends.
         /// </summary>
-        [Obsolete("AfterAll() is deprecated, please use AfterAllAsync() instead")]
         protected virtual void AfterAll()
         {
         }
```
Aaronontheweb, line 133: `Shutdown` needs to be called here

Aaronontheweb (changes requested):
Make the internals of the xUnit2 `TestKit` base class look exactly like how they did in v1.4, XML-DOC comments and all:

https://github.com/akkadotnet/akka.net/blob/a45501991bd366a4aa7db60347760332b6cd30c4/src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs#L22-L168

src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
Arkatufus: no, its not, it is the cause of multiple headaches, the actor system need to be called inside Dispose, not AfterAll

src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
Arkatufus: Dispose is a blocking long process now, can block for up to 5 seconds, disposing is needed just in case user called dispose multiple times

Aaronontheweb (approved):

PR #6489, opened 2023-03-06
[CS0809] pragma warning disable

eaba:
## Changes

- ActorRef  `public override bool IsTerminated`
- EmptyLocalActorRef  `public override bool IsTerminated`
- RemoteActorRef `public override bool IsTerminated`
- TestProbe `public override bool IsTerminated`


Aaronontheweb (approved):
Yep, this is the right call @eaba - this property, though it's obsolete, is buried too deep in the bowels of Akka.NET to remove without a lot of headache. Disabling the warning here is the best approach.

PR #6491, opened 2023-03-06
[disable] Warning CS0618

eaba:
## Changes

#pragma warning disable CS0618


src/benchmark/Akka.Benchmarks/Configurations/Configs.cs
```diff
@@ -19,8 +19,12 @@ public class MicroBenchmarkConfig : ManualConfig
     {
         public MicroBenchmarkConfig()
         {
+#pragma warning disable CS0618
```
Aaronontheweb, line 22: What's the full warning message here?

Aaronontheweb (changes requested):
What's the warning message here @eaba ?

PR #6494, opened 2023-03-07
[Obsolete][CS0618] `AddFirst`

eaba:
"This method will be removed from the public API in the future"


Aaronontheweb:
@Arkatufus is this still relevant in the era of the `async` TestKit?

Arkatufus:
Yes, it is still relevant

PR #6495, opened 2023-03-07
[Obsolete][CS0618]`ActorProducerPipeline*` 

eaba:
"Actor producer pipeline API will be removed in v1.5."

## Changes

- ActorProducerPipelineResolver
- ActorProducerPluginBase


Aaronontheweb:
I'm half tempted to just remove the `Obsolete` attribute off of these classes until we have something concrete to replace them with cc @akkadotnet/core

Aaronontheweb:
so @eaba , per our call this morning - plan is to remove the `Obsolete` attribute off of these types. You'll need to approve the API changes per https://getakka.net/community/contributing/api-changes-compatibility.html#api-change-procedures

Aaronontheweb (approved):
LGTM

PR #6497, opened 2023-03-07
Fix of PVS-Studio warnings.

### LINKED ISSUE #6073: Suspicious code fragments found by PVS-Studio
Hello,
PVS-Studio (static analyzer) issued a few warnings on the sources. I've described the most suspicious below. Hope it'll be useful. :)

**P.S.** Please, let me know if I should open separate issues for each code fragment.

<hr>

**Issue 1**
```
{
  if (!(obj is ActorMaterializerSettings s)) return false;
    return
                s.InitialInputBufferSize == InitialInputBufferSize &&
                s.MaxInputBufferSize == MaxInputBufferSize &&
                s.Dispatcher == Dispatcher &&
                s.SupervisionDecider == SupervisionDecider &&
                s.SubscriptionTimeoutSettings == SubscriptionTimeoutSettings &&   // <=
                s.IsDebugLogging == IsDebugLogging &&
                s.OutputBurstLimit == OutputBurstLimit &&
                s.SyncProcessingLimit == SyncProcessingLimit &&
                s.IsFuzzingMode == IsFuzzingMode &&
                s.IsAutoFusing == IsAutoFusing &&
                s.SubscriptionTimeoutSettings == SubscriptionTimeoutSettings &&   // <=
                s.StreamRefSettings == StreamRefSettings;
}
```

**Warning**: There are identical sub-expressions 's.SubscriptionTimeoutSettings == SubscriptionTimeoutSettings' to the left and to the right of the '&&' operator.

**Link to the sources**: [link](https://github.com/akkadotnet/akka.net/blob/0228d5d0c3af95b6d640e1b8ea4b249776cab2f0/src/core/Akka.Streams/ActorMaterializer.cs#L626)

Looks suspicious that the `SubscriptionTimeoutSettings` field is checked twice, whereas the `MaxFixedBufferSize` field is not checked.
<hr>

**Issue 2**

```
internal static string Base64Encode(this long value, string prefix)
{
  // 11 is the number of characters it takes to represent long.MaxValue
  // so we will never need a larger size for encoding longs
  Span<char> sb = stackalloc char[11 + prefix?.Length ?? 0];
  ....
}
```

**Warning**: Perhaps the '??' operator works in a different way than it was expected. Its priority is lower than priority of other operators in its left part. 

**Link to the sources**: [link](https://github.com/akkadotnet/akka.net/blob/0228d5d0c3af95b6d640e1b8ea4b249776cab2f0/src/core/Akka/Util/Base64Encoding.cs#L37).

If `prefix` is `null` then `sb` will be an empty array. Looks like it should contain 11 elements at least. If so, the correct expression should contain parentheses:
```
11 + (prefix?.Length ?? 0)
```
<hr>

**Issue 3**
```
protected override void PreStart() => Console.WriteLine("Good Morning, we are awake!", ConsoleColor.Green);
protected override void PostStop() => Console.WriteLine("Good Night, going to bed!", ConsoleColor.Red);
```

**Warning**: Incorrect format. A different number of format items is expected while calling 'WriteLine' function. Arguments not used: ConsoleColor.Green.

**Link to the sources**: [link](https://github.com/akkadotnet/akka.net/blob/0228d5d0c3af95b6d640e1b8ea4b249776cab2f0/src/examples/HelloAkka/HelloWorld/GreetingActor.cs#L24).

'WriteLine' method doesn't contain overloads that accept arguments of the 'ConsoleColor' type. So, they'll be ignored.
<hr>

**Issue 4**

```
 if (_actor == null)
 {
  ....
} 
else if (IsNormal)
{
  var failedActor = _actor;

    Publish(new Debug(_self.Path.ToString(), failedActor.GetType(), "Restarting"));

    if(!(failedActor is null))
    ....
}
....
```

**Warning**: Expression '!(failedActor is null)' is always true.

**Link to the sources**: [link](https://github.com/akkadotnet/akka.net/blob/0228d5d0c3af95b6d640e1b8ea4b249776cab2f0/src/core/Akka/Actor/ActorCell.FaultHandling.cs#L61).

The `failedActor` is always not `null` here. `_actor` is not `null` at the `else` branch, as a result `failedActor` too.

F0b0s:
Fixes #6073 

## Changes

Fix of Issue 2,3,4.

## Checklist

### DIFF
--- a/src/core/Akka.Tests/Util/Base64EncodingSpec.cs
+++ b/src/core/Akka.Tests/Util/Base64EncodingSpec.cs
@@ -1,0 +1,28 @@
+// //-----------------------------------------------------------------------
+// // <copyright file="Base64EncodingSpec.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2023 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2023 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using Akka.Util;
+using Xunit;
+
+namespace Akka.Tests.Util;
+
+public class Base64EncodingSpec
+{
+    [Fact]
+    public void When_prefix_is_null_it_should_work_correctly()
+    {
+        var actual = Base64Encoding.Base64Encode(12345, null);
+        Assert.Equal("5ad", actual);
+    }
+    
+    [Fact]
+    public void Should_calculate_base_64_correctly()
+    {
+        var actual = Base64Encoding.Base64Encode(12345, "");
+        Assert.Equal("5ad", actual);
+    }
+}
--- a/src/core/Akka/Actor/ActorCell.FaultHandling.cs
+++ b/src/core/Akka/Actor/ActorCell.FaultHandling.cs
@@ -58,30 +58,27 @@ namespace Akka.Actor
                 if (System.Settings.DebugLifecycle)
                     Publish(new Debug(_self.Path.ToString(), failedActor.GetType(), "Restarting"));
 
-                if(!(failedActor is null))
+                var optionalMessage = CurrentMessage;
+                try
                 {
-                    var optionalMessage = CurrentMessage;
-                    try
-                    {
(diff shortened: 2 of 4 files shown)


Aaronontheweb (approved):
LGTM

PR #6505, opened 2023-03-10
[obsolete] `ExpectNext(params T[] elems)` 

eaba:
## Changes

- `ExpectNext(CancellationToken cancellationToken, params T[] elems)`

## Checklist

### Latest `dev` Benchmarks 

### This PR's Benchmarks

### DIFF
--- a/src/core/Akka.Streams.TestKit/TestSubscriber_Fluent.cs
+++ b/src/core/Akka.Streams.TestKit/TestSubscriber_Fluent.cs
@@ -66,7 +66,6 @@ namespace Akka.Streams.TestKit
             /// <summary>
             /// Fluent DSL. Expect multiple stream elements.
             /// </summary>
-            [Obsolete("Use the method with CancellationToken support instead")]
             public ManualProbe<T> ExpectNext(params T[] elems)
             {
                 ExpectNextTask(this, null, default, elems)
--- a/src/core/Akka.Streams.Tests/Dsl/FlowThrottleSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/FlowThrottleSpec.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using Akka.IO;
 using Akka.Streams.Actors;
 using Akka.Streams.Dsl;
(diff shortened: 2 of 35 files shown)


src/core/Akka.Streams.Tests/Dsl/FlowThrottleSpec.cs
```diff
@@ -308,7 +309,7 @@ public void Throttle_for_single_cost_elements_must_properly_combine_shape_and_th
                     .Throttle(1, TimeSpan.FromMilliseconds(100), 5, ThrottleMode.Enforcing)
                     .RunWith(this.SinkProbe<int>(), Materializer)
                     .Request(5)
-                    .ExpectNext(1, 2, 3, 4, 5)
+                    .ExpectNext(CancellationToken.None, 1, 2, 3, 4, 5)
```
Aaronontheweb: @Arkatufus Maybe we need to add an overload to `ExpectNext` that implicitly sucks up a timeout. Passing in `CancellationToken.None` everywhere feels like code smell.

Aaronontheweb (changes requested):
Need feedback from @Arkatufus

src/core/Akka.Streams.Tests/Dsl/FlowThrottleSpec.cs
Arkatufus: The obsolete one is the one that does that. Maybe we should remove the Obsolete attribute instead.

Arkatufus: Note that this can be turned into an async call instead, it has a slightly different API.
```
await probe.AsyncBuilder()
    .ExpectNext("a-1", "a-2")
    .ExpectNoMsg(TimeSpan.FromMilliseconds(500))
    .ExecuteAsync();
```

Add the `.AsyncBuilder()` at front, then call `.ExecuteAsync()` at the end.

Aaronontheweb: > Maybe we should remove the Obsolete attribute instead.

I think that's a good idea - @eaba can you implement that on this PR instead?

eaba: done https://github.com/akkadotnet/akka.net/pull/6505/commits/dc0ebc86bf828cb9c273f16852fcb9d499804d60

src/core/Akka.Streams.TestKit/TestSubscriber_Fluent.cs
```diff
@@ -66,7 +66,6 @@ public ManualProbe<T> ExpectNext(TimeSpan? timeout, T element, CancellationToken
             /// <summary>
             /// Fluent DSL. Expect multiple stream elements.
             /// </summary>
-            [Obsolete("Use the method with CancellationToken support instead")]
```
Aaronontheweb, line 69: LGTM

Aaronontheweb (approved):
LGTM

Issue #4447, opened 2020-06-01
Akka.Remote cookies are not honored


petrikero:
I tried using the Akka.Remote cookies to avoid having two servers accidentally talk to each other, but the feature doesn't seem to work.

I tested using Akka.Net v1.4.7 and v1.4.4 which both produced the same results. I tested on Windows, but the issue seems platform-agnostic.

The relevant configs on the two nodes are:
`akka {
  remote {
    require-cookie = on
    secure-cookie = "first"
  }
}`
and
`akka {
  remote {
    require-cookie = on
    secure-cookie = "second"
  }
}`

Regardless of the differing cookies, the nodes connected just as without the cookies. Luckily, it's fairly easy for us to implement our own cookies on top of Akka.

Aaronontheweb:
We don't support this feature in Akka.NET and it was removed from JVM Akka because it never really worked in the first place.

PR #6515, opened 2023-03-14
Remove secure cookie

ismaelhamed:
As stated in #4447, secure cookies never worked plus they're deprecated in the JVM anyway.

### DIFF
--- a/src/protobuf/WireFormats.proto
+++ b/src/protobuf/WireFormats.proto
@@ -82,7 +82,7 @@ message AkkaControlMessage {
 message AkkaHandshakeInfo {
 	AddressData origin = 1;
 	fixed64 uid = 2;
-  optional string cookie = 3;
+  string cookie = 3; // not used
 }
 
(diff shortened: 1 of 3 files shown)


src/protobuf/WireFormats.proto
```diff
@@ -82,7 +82,7 @@ message AkkaControlMessage {
 message AkkaHandshakeInfo {
 	AddressData origin = 1;
 	fixed64 uid = 2;
-	string cookie = 3;
+  optional string cookie = 3;
```
Aaronontheweb: I don't think this is needed - these fields are already optional by default.

Aaronontheweb (approved):
LGTM

src/protobuf/WireFormats.proto
ismaelhamed: Everything in Proto3 is optional... I'll undo this.

Aaronontheweb: No worries - I took care of it

PR #6528, opened 2023-03-16
Fix broken object type serializer in QueryExecutor

### LINKED ISSUE #127: Custom Serializer setup not propagating to Persistence
**Version Information**
Akka - v1.4.43
Akka.Hosting - v0.4.3
Akka.Persistence.SqlServer.Hosting - v0.4.3
Akka.Serialization.Hyperion - v1.4.43

**Describe the bug**
When setting up Hyperion as the default serializer via Akka.Hosting configuration, persistence still uses Json as the default serializer and has to be configured separately to use Hyperion.

**To Reproduce**
I used a configuration similar to the following:

```csharp 
        services.AddAkka("actor-system", (configurationBuilder, _) =>
        {
            configurationBuilder
                .WithRemoting("localhost", 7910)
                .WithClustering(new ClusterOptions() { Roles = new[] { "Coordinator-Role" }, SeedNodes = Address[] { Address.Parse("akka.tcp://actor-system@localhost:7910") } })
                .WithSqlServerPersistence("Server=localhost,1433; Database=Akka; User Id=sa; Password=[password];")
                .WithCustomSerializer("hyperion", new[] { typeof(object) }, system => new HyperionSerializer(system));
        {
```

**Expected behavior**
When setting up Hyperion as the default serializer, this setting should be propagated through-out the Akka.net modules.

**Actual behavior**
Json was still the default serializer for persistence.

**Screenshots**
image

**Environment**
Windows 11
.Net 6.0

**Additional context**

### LINKED ISSUE #6389: Akka.Persistence: need to remove hard-coded Newtonsoft.Json `object` serializer
**Version Information**
Version of Akka.NET? v1.4.49
Which Akka.NET Modules? Akka.Persistence and all derivatives

Related: https://github.com/akkadotnet/akka.net/issues/3811

**Describe the bug**

Per https://github.com/akkadotnet/Akka.Hosting/issues/141

Persistence has its own default `System.Object` serializer settings here:

```hocon
akka.persistence.journal-plugin-fallback.serializer = json
akka.persistence.snapshot-store-plugin-fallback.serializer = json
```

https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Persistence/persistence.conf#L178

As far as I can tell, this cannot be overwritten.

I've attempted to override these in both of the following ways:

```csharp
 public static AkkaConfigurationBuilder AddLegacySerialization(this AkkaConfigurationBuilder builder)
    {
        return builder
            .AddHocon(@"
                akka.persistence.journal-plugin-fallback.serializer = json2
                akka.persistence.snapshot-store-plugin-fallback.serializer = json2
            ", HoconAddMode.Prepend)
        .AddHocon(@$"akka.actor {{
             serializers {{
                 json2 = ""{typeof(JsonSerializerWithManifest).AssemblyQualifiedName}""
             }}
            serialization-identifiers {{
                ""{typeof(JsonSerializerWithManifest).AssemblyQualifiedName}"" = 410
            }}
            serialization-bindings {{
                ""System.Object"" = json2
            }}
        }}", HoconAddMode.Prepend);
    }
```

I still get `Newtonsoft.Json` as my serializer, even though I have explicitly overridden all of the appropriate settings. Akka.Remote uses the correct serializer. Persistence does not.

I also tried overriding it programmatically:

```csharp
 public static AkkaConfigurationBuilder AddLegacySerialization(this AkkaConfigurationBuilder builder)
    {
        return builder.WithCustomSerializer("json2", new[] { typeof(object) },
                //system => new JsonSerializerWithManifest(system))
    }
```

Also did not work.

**Expected behavior**

Akka.Persistence should take its directions on serialization from Akka.NET, or at least be overridable using the appropriate settings.

**Actual behavior**

"Phasers had no effect, captain"

**Additional context**

I've discussed this before with @Arkatufus and we thought it might be unsafe to make this change - after looking more closely at it I don't think that's the case:

1. All old events serialized with a SerializerId of `1` will still be read using `Newtonsoft.Json`.
2. Any new events serialized going forward will use the `System.Object` serializer actually specified by the end-user.

In other words, this isn't a destructive change at all - old data read using old method, new data read using configured method.

Arkatufus:
Fixes https://github.com/akkadotnet/Akka.Hosting/issues/127
Fixes https://github.com/akkadotnet/akka.net/issues/6389

A Port of #6223 from v1.4 branch to dev branch

## Changes
* Removed persistence default deserializer feature

## Issue
Any non-serializer mapped objects being serialized by `Akka.Persistence.Sql.Common.Journal.QueryExecutor` will use the JSON serializer and not the object type serializer.

Problem is with this line:
https://github.com/akkadotnet/akka.net/blob/39a8f9e6822592b3c401fb55315f6f09a21ab260/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs#L783

which takes its value from the very obscure and very rarely used HOCON override here:
https://github.com/akkadotnet/akka.net/blob/39a8f9e6822592b3c401fb55315f6f09a21ab260/src/core/Akka.Persistence/persistence.conf#L119

## Workaround
Override the HOCON setting:

```
akka.persistence.journal-plugin-fallback.serializer = null
akka.persistence.snapshot-store-plugin-fallback.serializer = null
```

or declare the default non-mapped object serializer in the journal/snapshot settings:
```
akka.persistence.journal.sqlite.serializer = mySerializer
akka.persistence.snapshot-store.sqlite.serializer = mySerializer
```


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1228,7 +1229,7 @@ private async Task<WriteMessagesResult> HandleWriteMessages(WriteMessages req, T
         protected virtual void WriteEvent(TCommand command, IPersistentRepresentation persistent, string tags = "")
         {
             var payloadType = persistent.Payload.GetType();
-            var serializer = _serialization.FindSerializerForType(payloadType, Setup.DefaultSerializer);
+            var serializer = _serialization.FindSerializerForType(payloadType);
```
Arkatufus, line 1232: This is the actual fix, we do not use the default serializer set in the persistence plugin HOCON settings for writes anymore.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -780,7 +781,7 @@ protected DbCommand GetCommand(DbConnection connection, string sql)
         protected virtual void WriteEvent(DbCommand command, IPersistentRepresentation e, IImmutableSet<string> tags)
         {
             
-            var serializer = Serialization.FindSerializerForType(e.Payload.GetType(), Configuration.DefaultSerializer);
+            var serializer = Serialization.FindSerializerForType(e.Payload.GetType());
```
Arkatufus, line 784: This is the actual fix, we do not use the default serializer set in the persistence plugin HOCON settings for writes anymore.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -846,7 +847,10 @@ protected virtual IPersistentRepresentation ReadEvent(DbDataReader reader)
             {
                 // Support old writes that did not set the serializer id
                 var type = Type.GetType(manifest, true);
+#pragma warning disable CS0618
+                // Backward compatibility code, we still need to use the old default serializer on read to support legacy data
                 var deserializer = Serialization.FindSerializerForType(type, Configuration.DefaultSerializer);
```
Arkatufus, line 852: This is the only code in the SQL common journal that still uses the old settings. The purpose of this code is to make sure that really old (pre-v1.3.0) data are still readable by the journal. These pre-v1.3.0 event data does not have their `serializer_id` column populated.

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
```diff
@@ -224,22 +224,5 @@ protected override async Task DeleteAsync(string persistenceId, SnapshotSelectio
                 await QueryExecutor.DeleteBatchAsync(connection, nestedCancellationTokenSource.Token, persistenceId, criteria.MaxSequenceNr, criteria.MaxTimeStamp);
             }
         }
-        
-        private SnapshotEntry ToSnapshotEntry(SnapshotMetadata metadata, object snapshot)
```
Arkatufus, line 228: Byte-rot method, this method is not being used anymore.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
Arkatufus: Another piece of code that still uses this still uses this setting is the snapshot store reader, just like this one, it is there to make sure that legacy data are still readable.

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/Legacy/SqliteLegacyJournalSpec.cs
```diff
@@ -0,0 +1,127 @@
...
+        sqlite {{
+            auto-initialize = on
+            connection-string = ""{connectionString}""
+        }}
+    }}
+}}").WithFallback(SqlitePersistence.DefaultConfiguration());
+        }
+
+        [Fact(DisplayName = "Legacy v1.3.0 and below data (null serializer_id) should be read correctly")]
+        public void ReadLegacyDataTest()
```
Arkatufus, line 61: The purpose of this test is to make sure that data saved from v1.3.0 can still be recovered successfully. The Sqlite file were generated using v.1.3.0 code and is being recovered in this unit test.

src/contrib/persistence/Akka.Persistence.Sqlite.Tests/CustomObjectSerializerSpec.cs
```diff
@@ -0,0 +1,194 @@
...
+            // Read the database directly, make sure that we're using the correct object type serializer
+            var conn = new SqliteConnection(ConnectionString);
+            conn.Open();
+            const string sql = "SELECT ej.serializer_id FROM event_journal ej WHERE ej.persistence_id = 'a'";
+            await using var cmd = new SqliteCommand(sql, conn);
+            var record = await cmd.ExecuteReaderAsync();
+            await record.ReadAsync();
+            
+            // In the bug this fails, the serializer id is JSON id instead of MySerializer id
+            record[0].Should().Be(9999);
```
Arkatufus, line 88: This unit test makes sure that the old serializer HOCON setting are not being used in journal writes anymore.

Aaronontheweb (approved):
LGTM

PR #6534, opened 2023-03-20
[Obsolete] Fix all CS0618 for `IStage`

eaba:
## Checklist


Aaronontheweb:
@eaba looks like we have a merge conflict

Aaronontheweb:
@eaba also, you have some DocFx warnings to address as well

Aaronontheweb (changes requested):
This is a really bad idea - don't just disable `Obsolete` warnings across the board. There's a reason why these things are `Obsolete`. Figure out what it is and upgrade.

src/core/Akka.Streams.Tests/Dsl/FlowSelectSpec.cs
```diff
@@ -43,7 +43,9 @@ public void Select_should_select()
             var n = ThreadLocalRandom.Current.Next(10);
             for (int i = 0; i < n; i++)
             {
+#pragma warning disable CS0618 // Type or member is obsolete
```
Aaronontheweb, line 46: Whenever you disable an `Obsolete` warning, explain why this is appropriate.

PR #6538, opened 2023-03-22
`ForEachAsync` optimization, added missing `SinkForeachAsyncSpec`

### LINKED ISSUE #6391: [Feature - Akka.Streams] ForEachAsync
Imagine a scenario: consume input, process in a stream, sink to an external destination (write to database, send an api request etc). In this case, sink will be performing operation that is best run as a `Task<TResult>`. But currently there is no `Sink.ForEachAsync` to facilitate that, so available options are:

- `var mySink = Sink.ForEach<TIn>(async myVar => ...)`. This creates an async void delegate, which does not really behave as a loop where each interation does an async call. Also, exception propagation will not work, and IDEs/compiler will issue a warning for this kind of code.
- This workaround:
```
var mySink = Flow.Create<TIn>()
 .SelectAsync(...)
 .ToMaterialized(Sink.ForEach<TOut>(_ => { }), Keep.Right)
 .Named(...) 
```

Which works fine, but is not super elegant :)

OG Akka has this: https://doc.akka.io/docs/akka/current/stream/operators/Sink/foreachAsync.html

```
private Task<TResult> MySinkLogic(..)
{
  ...
}

var mySink = Sink.ForEachAsync(4, myVar => MySinkLogic(myVar))
```

ismaelhamed:
Resolves #6391 

## Checklist

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/SinkForeachAsyncSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/SinkForeachAsyncSpec.cs
@@ -1,0 +1,259 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="SinkForeachAsyncSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Streams.Dsl;
+using Akka.Streams.Supervision;
+using Akka.Streams.TestKit;
+using Akka.TestKit;
+using Akka.TestKit.Extensions;
+using Akka.Util.Internal;
+using FluentAssertions;
+using Nito.AsyncEx;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Streams.Tests.Dsl
+{
+    public class SinkForeachAsyncSpec : AkkaSpec
+    {
+        private ActorMaterializer Materializer { get; }
+
+        public SinkForeachAsyncSpec(ITestOutputHelper helper) : base(helper)
+        {
+            var settings = ActorMaterializerSettings.Create(Sys);
+            Materializer = ActorMaterializer.Create(Sys, settings);
+        }
+
+        [Fact]
+        public async Task A_ForeachAsync_must_handle_empty_source()
+        {
+            var p = Source.From(new List<int>()).RunWith(Sink.ForEachAsync<int>(3, _ => Task.CompletedTask), Materializer);
+            (await p.ShouldCompleteWithin(RemainingOrDefault)).Should().Be(Done.Instance);
+        }
+
+        [Fact]
+        public async Task A_ForeachAsync_must_be_able_to_run_elements_in_parallel()
+        {
+            var probe = CreateTestProbe();
+            var latch = Enumerable.Range(1, 4)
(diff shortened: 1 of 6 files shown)


src/core/Akka.Streams.Tests/Dsl/SinkForeachAsyncSpec.cs
```diff
@@ -0,0 +1,259 @@
...
+            threeCalled.ShouldBeTrue();
+            fourCalled.ShouldBeTrue();
+        }
+
+        [Fact]
+        public async Task A_ForeachAsync_must_produce_elements_in_the_order_they_are_ready()
+        {
+            var probe = CreateTestProbe();
+            var latch = Enumerable.Range(1, 4)
+                .Select(i => (i, new AsyncCountdownEvent(1)))
```
ismaelhamed, line 158: I had to used `AsyncCountdownEvent` directly in a bunch of tests because `TestLatch.Ready` currently blocks.

src/core/Akka.Streams/Dsl/Sink.cs
```diff
@@ -325,7 +325,7 @@ public static class Sink
         /// <param name="action">Async function delegate to be executed on all elements</param>
         /// <returns>TBD</returns>
         public static Sink<TIn, Task<Done>> ForEachAsync<TIn>(int parallelism, Func<TIn, Task> action) => Flow.Create<TIn>()
-            .SelectAsync(parallelism, async input =>
+            .SelectAsyncUnordered(parallelism, async input =>
```
Arkatufus, line 328: This will break the intent of `ForEachAsync`, we should preserve the behavior of our flow stream to be idiomatic to C#.
In this case a `foreach` should process the stream in order, which is why we need to use `SelectAsync`

src/core/Akka.Streams/Dsl/Sink.cs
```diff
@@ -375,6 +375,7 @@ public static class Sink
         /// <param name="parallelism">TBD</param>
         /// <param name="action">TBD</param>
         /// <returns>TBD</returns>
+        [Obsolete("Use `ForEachAsync` instead, it allows you to choose how to run the procedure, by calling some other API returning a Task or using Task.Run. Obsolete since 1.5.2")]
```
Arkatufus, line 378: Instead of making this obsolete, why not make an overloaed version that takes a `Func<TIn, Task>` instead of an `Action<TIn>` instead?

Arkatufus (changes requested):
I have some concern about the changes here

src/core/Akka.Streams/Dsl/Sink.cs
ismaelhamed: Actually, this replaces `ForEachParallel` which already utilizes `SelectAsyncUnordered`. The change in behavior was introduced by the current implementation of `ForEachAsync`.

Keep in mind that this operator is similar to `Parallel.ForEach` or `Parallel.ForEachAsync` in C#, which *also* does not guarantee the order of execution.

src/core/Akka.Streams/Dsl/Sink.cs
ismaelhamed: To eventually delete code instead of keep adding more: you can now do everything with `ForEachAsync`, hence why is deprecated. On top of that, we are keeping up date with Akka's public API.

Aaronontheweb: That makes sense - agree with @ismaelhamed

src/core/Akka.Streams/Dsl/Sink.cs
Arkatufus: OK, makes sense

src/core/Akka.Streams/Dsl/Sink.cs
Arkatufus: OK, make sense

Arkatufus (approved):

PR #6547, opened 2023-03-23
`ExpectNextUnordered` remove the `Obsolete` attribute

eaba:
## Checklist

### DIFF
--- a/src/core/Akka.Streams.TestKit/TestSubscriber_Fluent.cs
+++ b/src/core/Akka.Streams.TestKit/TestSubscriber_Fluent.cs
@@ -107,7 +107,6 @@ namespace Akka.Streams.TestKit
             /// <summary>
             /// Fluent DSL. Expect multiple stream elements in arbitrary order.
             /// </summary>
-            [Obsolete("Use the method with CancellationToken support instead")]
             public ManualProbe<T> ExpectNextUnordered(params T[] elems)
             {
                 ExpectNextUnorderedTask(this, null, default, elems)


src/core/Akka.Streams.TestKit/TestSubscriber_Fluent.cs
```diff
@@ -107,7 +107,6 @@ public ManualProbe<T> ExpectNext(TimeSpan? timeout, CancellationToken cancellati
             /// <summary>
             /// Fluent DSL. Expect multiple stream elements in arbitrary order.
             /// </summary>
-            [Obsolete("Use the method with CancellationToken support instead")]
```
Aaronontheweb, line 110: LGTM

Aaronontheweb (approved):
LGTM

src/core/Akka.Streams.TestKit/TestSubscriber_Fluent.cs
Aaronontheweb: Ah wait a minute - is there an `ExpectNextUnorderedAsync` cc @Arkatufus ?

Aaronontheweb (changes requested):
Actually, need to take a look - this is a blocking method.

src/core/Akka.Streams.TestKit/TestSubscriber_Fluent.cs
Arkatufus: Not for the non-async stream.
For the async version, you will need to use the `AsyncBuilder()` helper
```
await subscriber.AsyncBuilder()
  .ExpectNextUnordered()
  .ExecuteAsync();
```

Arkatufus: We agreed to remove this obsolete attribute before, so this change is good

Aaronontheweb (approved):

PR #6548, opened 2023-03-23
Log send time and sqNbr in verbose cluster heartbeat messages #6496.

### LINKED ISSUE #6496: Akka.Cluster: need to log send time in verbose heartbeat messages
https://github.com/akkadotnet/akka.net/blob/30e762c8c3abecc4b203827f6c30bda061ad13da/src/core/Akka.Cluster/ClusterHeartbeat.cs#L261

F0b0s:
Fixes #6496 

## Changes

Log send time and sqNbr in verbose cluster heartbeat messages

## Checklist

### DIFF
--- a/src/core/Akka.Cluster/ClusterHeartbeat.cs
+++ b/src/core/Akka.Cluster/ClusterHeartbeat.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Globalization;
 using System.Linq;
 using Akka.Actor;
 using Akka.Event;
@@ -43,8 +44,7 @@ namespace Akka.Cluster
             switch (message)
             {
                 case ClusterHeartbeatSender.Heartbeat hb:
-                    // TODO log the sequence nr once serializer is enabled
-                    if(VerboseHeartbeat) _cluster.CurrentInfoLogger.LogDebug("Heartbeat from [{0}]", hb.From);
+                    if (VerboseHeartbeat) _cluster.CurrentInfoLogger.LogDebug($"Heartbeat from [{hb.From}] - Sequence number [{hb.SequenceNr.ToString(CultureInfo.InvariantCulture)}]");
                     Sender.Tell(new ClusterHeartbeatSender.HeartbeatRsp(_cluster.SelfUniqueAddress,
                         hb.SequenceNr, hb.CreationTimeNanos));
                     break;
@@ -58,7 +58,6 @@ namespace Akka.Cluster
         {
             return Akka.Actor.Props.Create(() => new ClusterHeartbeatReceiver(getCluster));
         }
-
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/core/Akka.Cluster/ClusterHeartbeat.cs
```diff
@@ -258,8 +258,8 @@ private void DoHeartbeatRsp(HeartbeatRsp rsp)
         {
             if (_cluster.Settings.VerboseHeartbeatLogging)
             {
-                // TODO: log response time and validate sequence nrs once serialisation of sendTime is released
-                _log.Debug("Cluster Node [{0}] - Heartbeat response from [{1}]", _cluster.SelfAddress, rsp.From.Address);
+                _log.Debug("Cluster Node [{0}] - Heartbeat response from [{1}] - Sequence number [{2}] - Creation time [{3}]", _cluster.SelfAddress, rsp.From.Address,
+                    rsp.SequenceNr.ToString(CultureInfo.InvariantCulture), rsp.CreationTimeNanos.ToString(CultureInfo.InvariantCulture));
```
F0b0s, line 262: Why .ToString()
Because if we pass simple rsp.SequenceNr it will be boxed as its type is long.
Why .ToString(CultureInfo.InvariantCulture)
because ToString is culture dependent

src/core/Akka.Cluster/ClusterHeartbeat.cs
```diff
@@ -248,7 +247,8 @@ private void CheckTickInterval()
                     "Previous heartbeat was sent [{1}] ms ago, expected interval is [{2}] ms. This may cause failure detection " +
                     "to mark members as unreachable. The reason can be thread starvation, e.g. by running blocking tasks on the " +
                     "default dispatcher, CPU overload, or GC.",
-                    _cluster.SelfAddress, (now - _tickTimestamp).TotalMilliseconds, _cluster.Settings.HeartbeatInterval.TotalMilliseconds);
+                    _cluster.SelfAddress, (now - _tickTimestamp).TotalMilliseconds.ToString(CultureInfo.InvariantCulture),
```
F0b0s, line 250: Usage of CultureInfo.InvariantCulture here may be breaking changes, because without ToString logger calls object.ToString() which should use current culture.
Not sure do we need to use InvariantCulture here.

Aaronontheweb: `InvariantCulture` is the right call - won't be a breaking change for anyone (we don't consider updates to the content of log messages to be breaking changes.)

src/core/Akka.Cluster/ClusterHeartbeat.cs
Aaronontheweb: I'd convert the `CreationTimeNanos` back into a `DateTime` and log that, so the data is human-readable this way.

Aaronontheweb: https://github.com/akkadotnet/akka.net/blob/996cac701253749f63c55027db848c2a393dd660/src/core/Akka/Util/MonotonicClock.cs#L98-L101

That will get you back to ticks - from there you can convert back into a DateTime.

Aaronontheweb (changes requested):
One minor usability nitpick

src/core/Akka.Cluster/ClusterHeartbeat.cs
F0b0s: Sorry, can't get you.
`MonotonicClock` is just a `Stopwatch`, so it's something like `TimeSpan` since cluster start.
If i get ticks, how it can be converted to `DateTime`?
if i get ticks 5 min after cluster start it will be something like  `January 1, 0001 00:00:05`

Aaronontheweb: Doh, you're right - it's not absolute time. It's relative system time. You're correct. No need to convert to `DateTime` here.

Aaronontheweb (approved):

PR #6594, opened 2023-03-25
[47-74] `GraphUnzipSpec`

eaba:
## Changes


src/core/Akka.Streams.Tests/Dsl/GraphUnzipSpec.cs
```diff
@@ -62,24 +62,24 @@ public void A_Unzip_must_unzip_to_two_subscribers()
 
                 sub1.Request(1);
                 sub2.Request(2);
-                c1.ExpectNext(1*2);
```
Aaronontheweb, line 65: Needs to be converted to `async` @eaba - looks like this entire test fixture does.

Aaronontheweb (changes requested):
Needs to be converted to `async`

Aaronontheweb (approved):
LGTM

PR #6609, opened 2023-03-26
[61-74] `QueueSinkSpec`

eaba:
## Changes


src/core/Akka.Streams.Tests/Dsl/QueueSinkSpec.cs
```diff
@@ -57,14 +56,14 @@ public void QueueSink_should_send_the_elements_as_result_of_future()
                     queue.PullAsync().PipeTo(TestActor);
                     ExpectMsg(v);
```
Aaronontheweb: `ExpectMsgAsync`

Aaronontheweb (changes requested):
Look for async TestKit methods to use here instead of work-arounds.

Aaronontheweb (approved):
LGTM

PR #6610, opened 2023-03-26
[62-74] `QueueSourceSpec`

eaba:
## Changes


src/core/Akka.Streams.Tests/Dsl/QueueSourceSpec.cs
```diff
@@ -114,10 +114,9 @@ public void QueueSource_should_buffer_when_needed()
         }
 
         [Fact]
-        public void QueueSource_should_not_fail_when_0_buffer_space_and_demand_is_signalled()
```
Aaronontheweb, line 117: Use `async` methods instead - all tests.

Aaronontheweb (changes requested):

Aaronontheweb (approved):
LGTM

PR #6611, opened 2023-03-26
[63-74] `SinkForeachParallelSpec`

eaba:
## Changes


src/core/Akka.Streams.Tests/Dsl/SinkForeachParallelSpec.cs
```diff
@@ -19,6 +19,7 @@
 using System.Threading;
 using Akka.Streams.TestKit;
 using Akka.TestKit.Xunit2.Attributes;
+using System.Threading.Tasks;
```
Aaronontheweb, line 22: Use the new async TestKit methods and try to avoid `Task.Wait` et al

Aaronontheweb (changes requested):

Aaronontheweb (approved):
LGTM

PR #6613, opened 2023-03-26
[65-74] `SubscriberSinkSpec`

eaba:
## Changes


src/core/Akka.Streams.Tests/Dsl/SubscriberSinkSpec.cs
```diff
@@ -38,6 +38,7 @@ public void A_Flow_with_SubscriberSink_must_publish_elements_to_the_subscriber()
                 c.ExpectNext(2);
                 c.ExpectNext(3);
                 c.ExpectComplete();
```
Aaronontheweb: Use `async` methods here

Aaronontheweb (changes requested):
Use `async` methods instead of work-arounds

Aaronontheweb (approved):
LGTM

PR #6614, opened 2023-03-26
[66-74] `SubstreamSubscriptionTimeoutSpec`

eaba:
## Changes


src/core/Akka.Streams.Tests/Dsl/SubstreamSubscriptionTimeoutSpec.cs
```diff
@@ -91,14 +91,14 @@ public void GroupBy_and_SplitWhen_must_timeout_and_cancel_substream_publisher_wh
                 action.Should().Throw<SubscriptionTimeoutException>();
```
Aaronontheweb, line 96:
```csharp
 // sleep long enough for it to be cleaned up
                Thread.Sleep(1500);
```

Change to `Task.Delay`

src/core/Akka.Streams.Tests/Dsl/SubstreamSubscriptionTimeoutSpec.cs
```diff
@@ -116,6 +116,7 @@ public void GroupBy_and_SplitWhen_must_timeout_and_stop_groupBy_parent_actor_if_
 
                 subscriber.ExpectNext();
```
Aaronontheweb: Use `async` methods

Aaronontheweb (changes requested):
Use `async` methods instead of work-arounds

Aaronontheweb (approved):
LGTM

PR #6632, opened 2023-04-03
use correct capacity when creating DynamicQueue when FixedQueue is full

### LINKED ISSUE #6631: Queue with capacity = int.MaxValue and strategy DropNew still drops items after very small amount is added
**Version Information**
Version of Akka.NET? 1.4.28

**Describe the bug**
Hello, I am using Source.Queue<IdType>(int.MaxValue, OverflowStrategy.DropNew) to create a queue that always accepts new items, but after enqueuing around 64 of them OfferAsync returns Dropped.
I investigated the source code and found that it uses FixedQueue and in its Enqueue method it checks if there are 16 elements (constant) and allocates a dynamic queue like this:  var queue = new DynamicQueue(_head); The parameter _head has value of 64. I wonder if this is a bug and should be var queue = new DynamicQueue(Capacity); The problem is that queue IsFull returns _q.IsFull and since _q is now the DynamicQueue and its capacity is now 64 it returns true and the element is dropped.

**To Reproduce**
1. Create a queue with int.MaxValue limit: `Source.Queue<int>(int.MaxValue, OverflowStrategy.DropNew)`
2. Keep adding elements by calling `await q.OfferAsync ...`
3. Expect to get `QueueOfferResult.Enqueued` each time
4. Get `QueueOfferResult.Dropped` instead

**Expected behavior**
The queue should accept all items because of the high limit

**Actual behavior**
The queue still drops

**Environment**
GNU/Linux docker or local .net 6

ondravondra:
Fixes #6631

## Changes

When DynamicQueue is allocated, it's capacity is set to the original capacity of the buffer using it.

### DIFF
--- a/src/core/Akka.Streams/Implementation/Buffers.cs
+++ b/src/core/Akka.Streams/Implementation/Buffers.cs
@@ -375,7 +375,7 @@ namespace Akka.Streams.Implementation
             {
                 if (_tail - _head == Size)
                 {
-                    var queue = new DynamicQueue(_head);
+                    var queue = new DynamicQueue(Capacity);
                     while (NonEmpty)
                         queue.Enqueue(Dequeue());
                     queue.Enqueue(element);


src/core/Akka.Streams/Implementation/Buffers.cs
```diff
@@ -375,7 +375,7 @@ public void Enqueue(T element)
             {
                 if (_tail - _head == Size)
                 {
-                    var queue = new DynamicQueue(_head);
+                    var queue = new DynamicQueue(Capacity);
```
Aaronontheweb, line 378: LGTM

Aaronontheweb (approved):
LGTM

PR #6634, opened 2023-04-05
Add distributed pub-sub query feature

### LINKED ISSUE #3663: Enhancement: add ability to retrieve subscriber / topic stats from DistributedPubSubMediator
I'm thinking it'd be nice to add a metadata querying capability similar to what we have for Cluster.Sharding https://petabridge.com/blog/cluster-sharding-technical-overview-akkadotnet/#getting-shard-metadata for DistributedPubSub. 

I'd like the ability to query:

* All topics and the total number of subscribers on the current node for each - can currently get the topics by sending the mediator a message.
* List of all groups for a given topic on the current node and the number of subscribers in each group.
* The list of all nodes PubSub-ing.
* The ability to scatter-gather any of the above commands to the other mediators and gather responses back from all nodes.

My reasons for wanting these features:
1. Monitoring, whether via hand-rolled actors or perhaps via a [Petabridge.Cmd module](https://cmd.petabridge.com/articles/commands/index.html)
2. Run-time debugging - I can imagine some scenarios where I might to dump the state of the mediator to see if someone is / is not subscribed to something they should or should not be.

Bad (IMHO) reasons for wanting to have this feature:

1. Runtime flow-control - people should not be performing a request->response query against the subscriber list to check for subscribers every time they want to publish. That would be a no-no and would significantly limit the benefits of having a topic broker to begin with.

Are there any other bad reasons or objections to adding something like this?

Arkatufus:
Fixes #3663 

## Changes

## Checklist

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Query/Protocol.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Query/Protocol.cs
@@ -1,0 +1,118 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="Messages.cs" company="Akka.NET Project">
+//     Copyright (C) 2013-2023 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System.Collections.Immutable;
+using Akka.Actor;
+
+namespace Akka.Cluster.Tools.PublishSubscribe.Query;
+
+#region Request messages
+
+public sealed class GetLocalPubSubStats
+{
+    public static readonly GetLocalPubSubStats Instance = new();
+    private GetLocalPubSubStats() { }
+}
+
+public sealed class GetPubSubStats
+{
+    public static readonly GetPubSubStats Instance = new();
+    private GetPubSubStats() { }
+}
+
+public sealed class GetClusterPubSubStats
+{
+    public static readonly GetClusterPubSubStats Instance = new();
+    private GetClusterPubSubStats() { }
+}
+
+public sealed class GetLocalPubSubState
+{
+    public static readonly GetLocalPubSubState Instance = new();
+    private GetLocalPubSubState() { }
+}
+
+public sealed class GetPubSubState
+{
+    public static readonly GetPubSubState Instance = new();
+    private GetPubSubState() { }
+}
+
+public sealed class GetClusterPubSubState
+{
+    public static readonly GetClusterPubSubState Instance = new();
+    private GetClusterPubSubState() { }
+}
+
+#endregion
+
+#region Reply messages
+
+public sealed class LocalPubSubStats
+{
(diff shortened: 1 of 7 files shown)


src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Query/Protocol.cs
```diff
@@ -0,0 +1,118 @@
...
+    private GetPubSubStats() { }
+}
+
+public sealed class GetClusterPubSubStats
+{
+    public static readonly GetClusterPubSubStats Instance = new();
+    private GetClusterPubSubStats() { }
+}
+
+public sealed class GetLocalPubSubState
```
Aaronontheweb, line 32: What's the difference between state and stats?

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Query/Protocol.cs
```diff
@@ -0,0 +1,118 @@
...
+        Subscribers = subscribers;
+    }
+
+    public string Name { get; }
+    public ImmutableList<Address> Subscribers { get; }
+}
+
+public sealed class PubSubState
+{
+    public PubSubState(ImmutableDictionary<Address, LocalPubSubState> clusterStates)
```
Aaronontheweb, line 110: So this going to involve _a lot_ of duplicates on the receivers end - it's what everybody knows about everybody else. The "local state" should probably just be a list, literally, of node-local topic subscribers. In effect `TopicStats` already does this and does so more effectively by communicating the number of subscribers. Therefore, I don't think there's a lot of value in having the separate `state` commands at all. Merely showing "this node has N subscribers to T" is good enough and simpler.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Query/Protocol.cs
Arkatufus: Stats will return the subscriber count of each topic, state returns a list of actual subscriber addresses of each topic

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Query/Protocol.cs
Arkatufus: The problem with distributed pub sub local state cache is that its eventually consistent, do we need to expose this some how? A way to take a complete snapshot of the current local cluster state.

Aaronontheweb: I think keeping the protocol _really, really_ simple and providing a rough idea of where subscribers are for each topic at a given point in time is sufficient.

Aaronontheweb:
* Get list of available topics
* Get list of all topics and their subscriber counts locally
* Get list of all topics and their subscriber counts globally, with the node address as a relevant "column" in the results

PR #6639, opened 2023-04-07
Lock persistence journal table on write

Arkatufus:
## Changes

Change journal table write transaction isolation from undefined to serializable.
This is done to prevent a racy condition in a very write heavy environment, coupled with very eager read on the database.
The explanation for this is here: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property?view=sql-server-ver15#remarks
Basically, for an identity column to be guaranteed to be assigned sequential number during concurrent writes is to lock the table so that only one transaction can operate on the table at any given time.

## Performance Comparison

Performance was measured on MS SQL Server, 5 runs were done on each variant and only the highest result is taken, this is to assume that the most optimal condition was used on each test.
Numbers are measured in total messages per second.

| Spec | Old | New |
| ---- | --- | --- |
| Persist | 305 | 307 |
| PersistAll | 1318 | 1355 |
| PersistAllAsync | 3130 | 3158 |
| PersistAsync | 1026 | 1052 |
| PersistGroup10 | 962 | 1033 |
| PersistGroup100 | 920 | 1055 |
| PersistGroup200 | 986 | 1004 |
| PersistGroup25 | 982 | 1009 |
| PersistGroup50 | 783 | 777 |
| Recovering | 66507 | 65705 |
| RecoveringFour | 71158 | 69620 |
| RecoveringTwo | 55984 | 60593 |

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
@@ -676,8 +676,11 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// <returns>TBD</returns>
         public virtual async Task InsertBatchAsync(DbConnection connection, CancellationToken cancellationToken, WriteJournalBatch write)
         {
-            using (var command = GetCommand(connection, InsertEventSql))
-            using (var tx = connection.BeginTransaction())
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Performance results look acceptable to me - no major impact.

Aaronontheweb (approved):
LGTM

PR #6651, opened 2023-04-17
* Add a CombineLatest graph stage

### LINKED ISSUE #6650: Make a CombineLatest graph stage
As per conversation with @Aaronontheweb 

Create a custom graph stage that combines the latest elements from a set inputs. Emits first time when all inputs have an element available, afterwards each time a new value is available on any of the inputs.

arieldf:
Create a custom graph stage that combines the latest elements from a set inputs. Emits first time when all inputs have an element available, afterwards each time a new value is available on any of the inputs.

Closes #6650


arieldf:
Will file again with tests.

He-Pin:
would you like to contribute this to Akka and Pekko too？

Aaronontheweb:
> Will file again with tests.

sounds good to me

PR #6654, opened 2023-04-19
Add SQL query isolation level

### LINKED ISSUE #6640: Akka.Persistence.Sql.Common: make transaction isolation level configurable for reads and writes

From my Twitter thread yesterday https://twitter.com/Aaronontheweb/status/1645483986926329857

> Question for SQL Server pros: have a table written to frequently (1000/s), uses an IDENTITY column sorted in descending as our pagination index in that table, and our queries that run on fixed intervals BUT sometimes miss entries during SELECT. We think: concurrency problem

We've had reports from users of Akka.Persistence.Query that `EventsByTag` queries can periodically miss events in a high frequency journal.

We believe the issue, in this case, is that some of the rows being inserted by Akka.Persistence.SqlServer / Sql.Common are written but not yet committed and thus they are missed during the `SELECT TOP 100 FROM EventJournal WHERE Ordering > {PreviousOffset}` due to SqlServer's default configuration of using `ReadCommitted` isolation levels.

You can read more here: https://www.sqlservercentral.com/articles/isolation-levels-in-sql-server

It's a rare occurrence, but it does happen regularly ("once every 2-3 weeks", per the affected user.) We believe that lowering the isolation level to `ReadUncomitted` would resolve the issue in this scenario. However, making that change for all users in all use cases is a bit heavy-handed.

Therefore, I think exposing two new configuration settings to allow users to configure this is preferable - keeping the default as-is:

* `write-isolation-level` - configures the isolation level for deletes and inserts (doesn't affect performance much due to small transaction sizes, per the numbers in https://github.com/akkadotnet/akka.net/pull/6639)
* `read-isolation-level` - this would address the issue reported here.

**Additional context**

This issue was observed in SQL Server - not sure how well isolation levels carry over to other SQL platforms like Postgres / MySQL.

Arkatufus:
Fixes #6640

## Changes

* Add transaction to all database queries
* Add separate read and write isolation level to queries
* Add isolation level settings to sql journal and snapshot settings
* Add new HOCON key to control journal and snapshot query transaction level

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Extensions/DbConnectionExtensions.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Extensions/DbConnectionExtensions.cs
@@ -1,0 +1,74 @@
+﻿// //-----------------------------------------------------------------------
+// // <copyright file="DbConnectionExtensions.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2023 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2023 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using System;
+using System.Data;
+using System.Data.Common;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Akka.Persistence.Sql.Common.Extensions
+{
+    public static class DbConnectionExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static async Task ExecuteInTransaction(
+            this DbConnection connection,
+            IsolationLevel isolationLevel,
+            CancellationToken token,
+            Func<DbTransaction, CancellationToken, Task> task)
+        {
+            using var tx = connection.BeginTransaction(isolationLevel);
+            try
+            {
+                await task(tx, token);
+                tx.Commit();
+            }
+            catch (Exception ex1)
+            {
+                try
+                {
+                    tx.Rollback();
+                }
+                catch (Exception ex2)
+                {
+                    throw new AggregateException(ex2, ex1);
+                }
+                throw;
+            }
+        }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static async Task<T> ExecuteInTransaction<T>(
+            this DbConnection connection,
+            IsolationLevel isolationLevel,
+            CancellationToken token,
+            Func<DbTransaction, CancellationToken, Task<T>> task)
+        {
+            using var tx = connection.BeginTransaction(isolationLevel);
+            try
+            {
+                var result = await task(tx, token);
+                tx.Commit();
+                return result;
+            }
+            catch (Exception ex1)
+            {
+                try
+                {
+                    tx.Rollback();
+                }
+                catch (Exception ex2)
+                {
+                    throw new AggregateException(ex2, ex1);
+                }
+                throw;
+            }
+        }
+    }
+}
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Extensions/IsolationLevelExtensions.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Extensions/IsolationLevelExtensions.cs
@@ -22,7 +22,8 @@ namespace Akka.Persistence.Sql.Common.Extensions
             => level switch
             {
                 null => IsolationLevel.Unspecified,
-                "chaos" => IsolationLevel.Chaos,
+                "chaos" => // IsolationLevel.Chaos,
+                    throw new ConfigurationException($"{nameof(IsolationLevel)}.{IsolationLevel.Chaos} is not supported."),
                 "read-committed" => IsolationLevel.ReadCommitted,
                 "read-uncommitted" => IsolationLevel.ReadUncommitted,
                 "repeatable-read" => IsolationLevel.RepeatableRead,
@@ -30,7 +31,7 @@ namespace Akka.Persistence.Sql.Common.Extensions
                 "snapshot" => IsolationLevel.Snapshot,
                 "unspecified" => IsolationLevel.Unspecified,
                 _ => throw new ConfigurationException(
-                    "Unknown isolation-level value. Should be one of: chaos | read-committed | read-uncommitted | repeatable-read | serializable | snapshot | unspecified")
+                    "Unknown isolation-level value. Should be one of: read-committed | read-uncommitted | repeatable-read | serializable | snapshot | unspecified")
             };
         
     }
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Extensions/IsolationLevelExtensions.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Extensions/IsolationLevelExtensions.cs
@@ -1,0 +1,37 @@
+﻿// //-----------------------------------------------------------------------
+// // <copyright file="HoconExtensions.cs" company="Akka.NET Project">
+// //     Copyright (C) 2009-2023 Lightbend Inc. <http://www.lightbend.com>
+// //     Copyright (C) 2013-2023 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// // </copyright>
+// //-----------------------------------------------------------------------
+
+using System.Data;
+using System.Runtime.CompilerServices;
+using Akka.Configuration;
+
+namespace Akka.Persistence.Sql.Common.Extensions
+{
+    public static class IsolationLevelExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static IsolationLevel GetIsolationLevel(this Config config, string key)
+            => config.GetString(key).ToIsolationLevel();
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static IsolationLevel ToIsolationLevel(this string level)
+            => level switch
+            {
+                null => IsolationLevel.Unspecified,
+                "chaos" => IsolationLevel.Chaos,
+                "read-committed" => IsolationLevel.ReadCommitted,
+                "read-uncommitted" => IsolationLevel.ReadUncommitted,
+                "repeatable-read" => IsolationLevel.RepeatableRead,
+                "serializable" => IsolationLevel.Serializable,
+                "snapshot" => IsolationLevel.Snapshot,
+                "unspecified" => IsolationLevel.Unspecified,
+                _ => throw new ConfigurationException(
+                    "Unknown isolation-level value. Should be one of: chaos | read-committed | read-uncommitted | repeatable-read | serializable | snapshot | unspecified")
+            };
+        
+    }
+}
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
@@ -22,6 +22,7 @@ using Akka.Configuration;
 using Akka.Event;
 using Akka.Pattern;
 using Akka.Persistence.Journal;
+using Akka.Persistence.Sql.Common.Extensions;
 using Akka.Persistence.Sql.Common.Journal;
 using Akka.Serialization;
 using Akka.Util;
@@ -210,9 +211,10 @@ namespace Akka.Persistence.Sql.Common.Journal
         public TimeSpan ConnectionTimeout { get; }
 
         /// <summary>
-        /// Isolation level of transactions used during query execution.
+        /// Isolation level of transactions used during write query execution.
         /// </summary>
-        public IsolationLevel IsolationLevel { get; }
+        [Obsolete("Use WriteIsolationLevel property instead")]
+        public IsolationLevel IsolationLevel => WriteIsolationLevel;
 
         /// <summary>
         /// Settings specific to <see cref="CircuitBreaker"/>, which is used internally 
@@ -241,6 +243,16 @@ namespace Akka.Persistence.Sql.Common.Journal
         /// The fully qualified name of the type that should be used as timestamp provider.
         /// </summary>
         public string TimestampProviderTypeName { get; }
+        
+        /// <summary>
+        /// Isolation level of transactions used during read query execution.
+        /// </summary>
+        public IsolationLevel ReadIsolationLevel { get; }
+
+        /// <summary>
+        /// Isolation level of transactions used during write query execution.
+        /// </summary>
+        public IsolationLevel WriteIsolationLevel { get; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="BatchingSqlJournalSetup" /> class.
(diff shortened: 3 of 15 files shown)


src/contrib/persistence/Akka.Persistence.Sql.Common/Extensions/DbConnectionExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Akka.Persistence.Sql.Common.Extensions
+{
+    public static class DbConnectionExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static async Task ExecuteInTransaction(
```
Arkatufus, line 20: These extension methods wraps a query delegate inside a controlled transaction

src/contrib/persistence/Akka.Persistence.Sql.Common/Extensions/DbConnectionExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+                catch (Exception ex2)
+                {
+                    throw new AggregateException(ex2, ex1);
+                }
+                throw;
+            }
+        }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static async Task<T> ExecuteInTransaction<T>(
```
Arkatufus, line 47: These extension methods wraps a query delegate inside a controlled transaction.
This is the wrapper version that returns a value.

src/contrib/persistence/Akka.Persistence.Sql.Common/Extensions/IsolationLevelExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Akka.Persistence.Sql.Common.Extensions
+{
+    public static class IsolationLevelExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static IsolationLevel GetIsolationLevel(this Config config, string key)
+            => config.GetString(key).ToIsolationLevel();
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static IsolationLevel ToIsolationLevel(this string level)
```
Arkatufus, line 21: Convenience method to convert HOCON/string to `IsolationLevel`

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -276,30 +288,26 @@ protected BatchingSqlJournalSetup(Config config, QueryConfiguration namingConven
...
-                case "serializable": level = IsolationLevel.Serializable; break;
-                case "snapshot": level = IsolationLevel.Snapshot; break;
-                case "unspecified": level = IsolationLevel.Unspecified; break;
-                default: throw new ConfigurationException("Unknown isolation-level value. Should be one of: chaos | read-committed | read-uncommitted | repeatable-read | serializable | snapshot | unspecified");
-            }
+            ReadIsolationLevel = namingConventions.ReadIsolationLevel;
+            WriteIsolationLevel = namingConventions.WriteIsolationLevel;
+            
+            // backward compatibility
+            var level = config.GetString("isolation-level");
```
Arkatufus, line 295: Backward compatibility for older codes

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -862,25 +940,24 @@ private void TryProcess()
             {
                 _remainingOperations--;
 
-                var chunk = DequeueChunk(_remainingOperations);
+                var (chunk, isWrite) = DequeueChunk(_remainingOperations);
```
Arkatufus, line 943: `isWrite` is a flag to signal if the current batch chunk is a read or a write chunk.

`DequeueChunk` automatically separate the read and write requests in the queue, we need to pull the read and write information of the current batch chunk from there.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -862,25 +940,24 @@ private void TryProcess()
...
                 var context = Context;
-                _circuitBreaker.WithCircuitBreaker(() => ExecuteChunk(chunk, context))
+                _circuitBreaker.WithCircuitBreaker(() => ExecuteChunk(chunk, context, isWrite))
                     .PipeTo(Self, failure: ex => new ChunkExecutionFailure(ex, chunk.Requests, chunk.ChunkId));
             }
         }
 
-        private async Task<BatchComplete> ExecuteChunk(RequestChunk chunk, IActorContext context)
+        private async Task<BatchComplete> ExecuteChunk(RequestChunk chunk, IActorContext context, bool isWriteOperation)
```
Arkatufus, line 950: Need to tell `ExecuteBatch` the read/write context of the chunk so we can apply the appropriate read/write isolation level

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -1331,7 +1406,7 @@ protected void AddParameter(TCommand command, string paramName, DbType dbType, o
         /// Select the buffer that has the smallest id on its first item, retrieve a maximum Setup.MaxBatchSize
         /// items from it, and return it as a chunk that needs to be batched
         /// </summary>
-        private RequestChunk DequeueChunk(int chunkId)
+        private (RequestChunk chunk, bool isWrite) DequeueChunk(int chunkId)
```
Arkatufus, line 1409: `isWrite` is a flag to signal if the current batch chunk is a read or a write chunk.

`DequeueChunk` automatically separate the read and write requests in the queue, we need to pull the read and write information of the current batch chunk from there.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -323,19 +331,84 @@ protected BatchingSqlJournalSetup(Config config, QueryConfiguration namingConven
         /// <param name="replayFilterSettings">The settings used when replaying events from database back to the persistent actors.</param>
         /// <param name="namingConventions">The naming conventions used by the database to construct valid SQL statements.</param>
         /// <param name="defaultSerializer">The serializer used when no specific type matching can be found.</param>
-        protected BatchingSqlJournalSetup(string connectionString, int maxConcurrentOperations, int maxBatchSize, int maxBufferSize, bool autoInitialize, TimeSpan connectionTimeout, IsolationLevel isolationLevel, CircuitBreakerSettings circuitBreakerSettings, ReplayFilterSettings replayFilterSettings, QueryConfiguration namingConventions, string defaultSerializer)
+        [Obsolete("Use constructor with separate read and write isolation level instead. (since v1.5.2)")]
```
Arkatufus, line 334: Marking old constructors as obsolete

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -207,6 +217,7 @@ public class QueryConfiguration
         /// <param name="timeout">TBD</param>
         /// <param name="defaultSerializer">The default serializer used when not type override matching is found</param>
         /// <param name="useSequentialAccess">Uses the CommandBehavior.SequentialAccess when creating the command, providing a performance improvement for reading large BLOBS.</param>
+        [Obsolete("Use .ctor that accepts read and write IsolationLevel instead (since v1.5.2)")]
```
Arkatufus, line 220: Marking old constructors as obsolete

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -487,22 +577,25 @@ protected AbstractQueryExecutor(QueryConfiguration configuration, Akka.Serializa
...
         /// <param name="offset">TBD</param>
         /// <returns>TBD</returns>
-        public virtual async Task<ImmutableArray<string>> SelectAllPersistenceIdsAsync(DbConnection connection, CancellationToken cancellationToken, long offset)
+        public virtual async Task<ImmutableArray<string>> SelectAllPersistenceIdsAsync(
+            DbConnection connection, 
+            CancellationToken cancellationToken,
+            long offset)
         {
-            using (var command = GetCommand(connection, AllPersistenceIdsSql))
+            return await connection.ExecuteInTransaction(ReadIsolationLevel, cancellationToken, async (tx, token) =>
```
Arkatufus, line 585: Converts old code into delegate and tuck them inside the transaction wrapper. 

There should be zero impact from these wrap, the code is marked with aggresive inlining so that they are unrolled as soon as possible either during compile or inside JIT.

No logic were changed, we're just wrapping them inside explicit transaction.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -516,37 +609,35 @@ public virtual async Task<ImmutableArray<string>> SelectAllPersistenceIdsAsync(D
...
+            DbConnection connection,
+            CancellationToken cancellationToken,
+            string persistenceId,
+            long fromSequenceNr,
+            long toSequenceNr,
+            long max,
+            Action<IPersistentRepresentation> callback)
         {
-            using (var command = GetCommand(connection, ByPersistenceIdSql))
+            await connection.ExecuteInTransaction(ReadIsolationLevel, cancellationToken, async (tx, token) =>
```
Arkatufus, line 621: Same as above, wrap the old code inside transaction.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/SqlJournal.cs
```diff
@@ -27,7 +27,7 @@ public abstract class SqlJournal : AsyncWriteJournal, IWithUnboundedStash
         private IImmutableDictionary<string, long> _tagSequenceNr = ImmutableDictionary<string, long>.Empty;
 
         private readonly CancellationTokenSource _pendingRequestsCancellation;
-        private readonly JournalSettings _settings;
+        protected readonly JournalSettings Settings;
```
Arkatufus, line 30: Expose `JournalSettings` as protected so inheriting classes can use them.

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
```diff
@@ -132,6 +143,7 @@ public class QueryConfiguration
         /// <param name="timeout">TBD</param>
         /// <param name="defaultSerializer">The default serializer used when not type override matching is found</param>
         /// <param name="useSequentialAccess">Uses the CommandBehavior.SequentialAccess when creating the command, providing a performance improvement for reading large BLOBS.</param>
+        [Obsolete("Use the constructor that takes read and write isolation level argument (since v1.5.2)")]
```
Arkatufus, line 146: Mark old constructor as obsolete for backward compatibility

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/SqlSnapshotStore.cs
```diff
@@ -38,7 +38,7 @@ private sealed class Initialized
         /// </summary>
         private readonly CancellationTokenSource _pendingRequestsCancellation;
 
-        private readonly SnapshotStoreSettings _settings;
+        protected readonly SnapshotStoreSettings Settings;
```
Arkatufus, line 41: Expose `SnapshotStoreSettings` as protected so inheriting classes can use them

src/contrib/persistence/Akka.Persistence.Sqlite/sqlite.conf
```diff
@@ -31,6 +31,14 @@
 			# timestamp provider used for generation of journal entries timestamps
 			timestamp-provider = "Akka.Persistence.Sql.Common.Journal.DefaultTimestampProvider, Akka.Persistence.Sql.Common"
 			
+			# The isolation level of all database read query
+			# Valid values: "chaos", "read-committed", "read-uncommitted", "repeatable-read", "serializable", "snapshot", or "unspecified"
+			read-isolation-level = unspecified
```
Arkatufus, line 38: These new isolation level settings need to be propagated to all `Sql.Common` implementation when they are upgraded to the next `Akka.NET` release.

src/contrib/persistence/Akka.Persistence.Sqlite/sqlite.conf
```diff
@@ -31,6 +31,14 @@
...
+			
+			# The isolation level of all database write query
+			# Valid values: "chaos", "read-committed", "read-uncommitted", "repeatable-read", "serializable", "snapshot", or "unspecified"
+			write-isolation-level = unspecified
```
Arkatufus, line 44: These new isolation level settings need to be propagated to all `Sql.Common` implementation when they are upgraded to the next `Akka.NET` release.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -210,9 +211,10 @@ public abstract class BatchingSqlJournalSetup
         public TimeSpan ConnectionTimeout { get; }
 
         /// <summary>
-        /// Isolation level of transactions used during query execution.
+        /// Isolation level of transactions used during write query execution.
         /// </summary>
-        public IsolationLevel IsolationLevel { get; }
+        [Obsolete("Use WriteIsolationLevel property instead")]
```
Arkatufus, line 216: Make old property obsolete, old isolation level is used for both read and write, we now have fine grain control over both.

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/BatchingSqlJournal.cs
```diff
@@ -862,25 +940,24 @@ private void TryProcess()
...
-            var isWriteOperation = false;
             var stopwatch = new Stopwatch();
             using (var connection = CreateConnection(Setup.ConnectionString))
             {
                 await connection.OpenAsync();
 
                 // In the grand scheme of thing, using a transaction in an all read batch operation
                 // should not hurt performance by much, because it is done only once at the start.
-                using (var tx = connection.BeginTransaction(Setup.IsolationLevel))
+                using (var tx = connection.BeginTransaction(isWriteOperation ? _writeIsolationLevel : _readIsolationLevel))
```
Arkatufus, line 960: The kind of isolation level is controlled via the write flag argument

src/contrib/persistence/Akka.Persistence.Sqlite/sqlite.conf
Aaronontheweb: What's the difference between `chaos` and `unspecified`?

Arkatufus: The differences are covered here: https://learn.microsoft.com/en-us/dotnet/api/system.data.isolationlevel?#fields

Aaronontheweb (approved):
LGTM

PR #6662, opened 2023-04-23
EventStream deadLetters doc has been improved

### LINKED ISSUE #5334: Docs: need to improve documentation how to receive unhandled + dead letters
Original issue reported by user here at SO: https://stackoverflow.com/questions/69656719/akka-net-1-4-dead-letter-handler-eventstream-subscribe

> Just because it cost me a few hours, I'd like to share the following insight, may it be interesting for someone: If we want to handle dead letters by ourselves, the akka.net documentation (https://getakka.net/articles/utilities/event-bus.html) says we should use
> 
>     "system.EventStream.Subscribe(deadletterWatchActorRef, typeof(DeadLetter));" 
> 
> This is only partly true - yes, this catches those messages which are strictly "dead letters", those whose recipient is currently dead. But it does not catch the unhandled messages and the dropped ones. If you want that (and I'm pretty sure that 99% of the use cases don't want to finely differentiate between unprocessed, dropped and really dead letters), you have to use "AllDeadLetters" instead of "DeadLetters", so the proper call is
> 
>     system.EventStream.Subscribe(deadletterWatchActorRef, typeof(AllDeadLetters));

**Which pages have you looked at?**

1. https://getakka.net/articles/utilities/event-bus.html

**What did these pages not make clear?**
Doesn't distinguish between dead letters and unhandled messages, and doesn't explain how to capture both.

**How can we do it better?**
Go into detail on what the different types of unhandled messages mean and explain how to subscribe to all of them via the `EventStream`. Might want to make this information more easily discoverable in the navigation too.

F0b0s:
Fixes #5334 

## Changes

Add information about different EventBus message types to subscribe to.

## Checklist

### DIFF
--- a/docs/articles/utilities/event-bus.md
+++ b/docs/articles/utilities/event-bus.md
@@ -5,8 +5,9 @@ title: Event Bus
 # EventBus
 
 EventBus provides several types of custom messages to subscribe to:
+
 1. DeadLetter - messages that are not delivered to actor
-2. UnhandledMessage - messages which actor receives and doesn't understand 
+2. UnhandledMessage - messages which actor receives and doesn't understand
 3. SuppressedDeadLetter - similar to DeadLetter with the slight twist of NOT being logged by the default dead letters listener
 4. Dropped - messages dropped due to overfull queues or routers with no routees
 5. AllDeadLetters - shortcut for all types of beforementioned messages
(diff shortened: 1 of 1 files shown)


Aaronontheweb (changes requested):
LGTM

Aaronontheweb (approved):

Aaronontheweb:
DocFx timed out. Nothing wrong with your PR. Going to override the build system and merge this in.

Issue #6667, opened 2023-04-24
Akka.TestKit's "ExpectMsgAllOf" methods could be emulated to provide new Predicate-based testing methods.


malclear:
#### Problem
I’ve been running into issues with tests failing periodically in the Akkatecture library I maintain. Looking back, I see that the issue has existed for a while; the original Akkatecture author made note of it. Essentially, the tests are racy, “Heisenbug”-style problems. 
 
While looking for a solution, I see Akka.TestKit has a set of `ExpectMsgAllOf` methods that can be used for "expecting" a set of messages. It is described as a solution to racy tests here: [Debugging Akka.NET](https://getakka.net/community/contributing/debugging.html). I would like to use something like this to test for a set of events that should exist after a specific domain action. However, aspects of the `ExpectMsgAllOf` methods make them unsuitable for the purpose: 
1) They require whole objects to be submitted for an equality check, rather than allowing Predicates (as does the `ExpectMsg` functions). 
	1) If the contents of the events resulting from the test are not easily predicted, then using a Predicate to validate individual aspects of each message might suffice.
	2) I'm unable to supply our own value to the parameter `areEqual`, as it exists only in the private method `InternalExpectMsgAllOfAsync<T>(...)` (and that would be hackish anyway).
2) The `ExpectMsgAllOf` method expects a common message type, `T`. 
	1) In my use-case, events will be a variety of types, all derived from a common ancestor, and the details of the derived type would be required to satisfy the Predicate<T> associated with the particular message.

#### Proposed Solution
I propose a set of methods named as `ExpectMsgAllOfWithPredicates` that would allow setting the expectations for sets of messages using predicates. This would be similar to how the `ExpectMsg` methods accept `Predicate<T>` as a parameter.

To enable heterogeneous message types in the set,  the methods would accept a collection of a new type, `PredicateInfo`, that would contain the type of the expected input as an instance of `Type`. 

The wrapper class `PredicateInfo` would have only two properties of the types `Delegate` and `Type`. Static helper methods will be supplied as factories for safety and convenience.

Aaronontheweb:
closed via https://github.com/akkadotnet/akka.net/pull/6668

PR #6668, opened 2023-04-24
[Issue #6667] New methods and tests to enable "ExpectAll" with Predicates

malclear:
## Changes
Adds new methods to Akka.TestKit

A new set of methods named as ExpectMsgAllOfWithPredicates that would allow setting the expectations for sets of messages using predicates. This would be similar to how the ExpectMsg methods accept Predicate<T> as a parameter.

To enable heterogeneous message types in the set, the methods would accept a collection of a new type, PredicateInfo, that would contain the type of the expected input as an instance of Type.

The wrapper class PredicateInfo would have only two properties of the types Delegate and Type. Static helper methods will be supplied as factories for safety and convenience.

### DIFF
--- a/src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
+++ b/src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
@@ -1,0 +1,155 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.TestKit.Internal;
+
+namespace Akka.TestKit;
+
+public abstract partial class TestKitBase
+{
+    public IReadOnlyCollection<object> ExpectMsgAllOfMatchingPredicates(params PredicateInfo[] predicates)
+    {
+        using var cts = new CancellationTokenSource(RemainingOrDefault);
+        return ExpectMsgAllOfMatchingPredicates(predicates, cts.Token);
+    }
+
+    public IReadOnlyCollection<object> ExpectMsgAllOfMatchingPredicates(IReadOnlyCollection<PredicateInfo> predicates,
+        CancellationToken cancellationToken)
+    {
+        return ExpectMsgAllOfMatchingPredicatesAsync(predicates, cancellationToken)
+            .ToListAsync(cancellationToken)
+            .ConfigureAwait(false).GetAwaiter().GetResult();
+    }
+
+    public async IAsyncEnumerable<object> ExpectMsgAllOfMatchingPredicatesAsync(
+        IReadOnlyCollection<PredicateInfo> predicates,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        var enumerable = InternalExpectMsgAllOfMatchingPredicatesAsync(new TimeSpan(0, 0, 60), predicates,
+                cancellationToken: cancellationToken)
+            .ConfigureAwait(false).WithCancellation(cancellationToken);
+        await foreach (var item in enumerable)
+        {
+            yield return item;
+        }
+    }
+
+    public IReadOnlyCollection<object> ExpectMsgAllOfMatchingPredicates(TimeSpan max, params PredicateInfo[] predicates)
+    {
+        using var cts = new CancellationTokenSource(RemainingOrDefault);
+        return ExpectMsgAllOfMatchingPredicates(max, predicates, cts.Token);
+    }
+
+    public IReadOnlyCollection<object> ExpectMsgAllOfMatchingPredicates(TimeSpan max,
+        IReadOnlyCollection<PredicateInfo> predicates, CancellationToken cancellationToken)
+    {
+        return ExpectMsgAllOfMatchingPredicatesAsync(max, predicates, cancellationToken)
+            .ToListAsync(cancellationToken)
+            .ConfigureAwait(false).GetAwaiter().GetResult();
+    }
+
+    public async IAsyncEnumerable<object> ExpectMsgAllOfMatchingPredicatesAsync(
+        TimeSpan max,
+        IReadOnlyCollection<PredicateInfo> predicates,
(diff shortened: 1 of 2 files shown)


src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
```diff
@@ -0,0 +1,155 @@
...
+        if (parameters.Length != 1)
+        {
+            error += "Delegate does not take exactly one parameter.";
+        }
+
+        if (error.Length != 0)
+            throw new ArgumentException(error);
+
+        var predicate =
+            (Predicate<int>)Delegate.CreateDelegate(typeof(Predicate<int>), predicateT.Target, predicateT.Method);
```
Arkatufus: Shouldn't this be `Predicate<T>` instead of `Predicate<int>`?

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    public async IAsyncEnumerable<object> ExpectMsgAllOfMatchingPredicatesAsync(
+        IReadOnlyCollection<PredicateInfo> predicates,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        var enumerable = InternalExpectMsgAllOfMatchingPredicatesAsync(new TimeSpan(0, 0, 60), predicates,
+                cancellationToken: cancellationToken)
+            .ConfigureAwait(false).WithCancellation(cancellationToken);
+        await foreach (var item in enumerable)
```
Arkatufus, line 34: Can't we return the result of `InternalExpectMsgAllOfMatchingPredicateAsync` and avoid re-enumerating it?

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
```diff
@@ -0,0 +1,155 @@
...
+
+    public async IAsyncEnumerable<object> ExpectMsgAllOfMatchingPredicatesAsync(
+        TimeSpan max,
+        IReadOnlyCollection<PredicateInfo> predicates,
+        [EnumeratorCancellation] CancellationToken cancellationToken = default)
+    {
+        max.EnsureIsPositiveFinite("max");
+        var enumerable = InternalExpectMsgAllOfAsync(Dilated(max), predicates, cancellationToken: cancellationToken)
+            .ConfigureAwait(false).WithCancellation(cancellationToken);
+        await foreach (var item in enumerable)
```
Arkatufus, line 62: Same as above

Arkatufus (changes requested):
Left a few comments

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
malclear: You are correct. Copy/paste error.

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
malclear: That question came into (and out of) my mind as well. I was emulating the [existing code here](https://github.com/malclear/akka.net/blob/feature/addPredicateSetTesting/src/core/Akka.TestKit/TestKitBase_Expect.cs#L606) when I wrote that.
Should I remove it in both places?

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
malclear: More [cargo garbage](https://github.com/malclear/akka.net/blob/feature/addPredicateSetTesting/src/core/Akka.TestKit/TestKitBase_Expect.cs#L654)?

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
Arkatufus: I think it is fine to just return the result instead, as long as we keep the `.ConfigureAwait()` code.

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
Arkatufus: Yes, we should return the result instead.

malclear:
Responses added. I had questions before I send a new commit.

Arkatufus:
@malclear Responded to your questions, you can go ahead with the new commit.
Thank you for the contribution :)

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
Arkatufus: Actually, nevermind, this code is good as is, its not possible to return an `IAsyncEnumerable` with `.ConfigureAwait()`

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
Arkatufus: Actually, nevermind, this code is good as is, its not possible to return an IAsyncEnumerable with .ConfigureAwait()

malclear:
@Arkatufus , Would you like me to adjust the those two functions in the TestKitBase_Expect.cs file as well?

Arkatufus (approved):

Arkatufus:
@malclear PR approved, auto-merging it.
Thanks for the contribution :)

malclear:
I see the issue. The internal method implicitly returns a `ConfiguredCancelableAsyncEnumerable<object>` type which can't then be returned by the calling method. 

I see you've approved the PR and are auto-merging it. Did you make a change of the code where "<int>" should be "<T>"?

Arkatufus:
Yes, I did, the PR is good for merging.

Aaronontheweb (approved):
LGTM

PR #6672, opened 2023-04-25
[CS0618] AbstractStage Warning Disable

eaba:
## Checklist

### DIFF
--- a/src/core/Akka.Streams/Stage/AbstractStage.cs
+++ b/src/core/Akka.Streams/Stage/AbstractStage.cs
@@ -18,7 +18,9 @@ namespace Akka.Streams.Stage
     /// <typeparam name="TOut">TBD</typeparam>
     internal class PushPullGraphLogic<TIn, TOut> : GraphStageLogic, IDetachedContext<TOut>
     {
+#pragma warning disable CS0618 // Type or member is obsolete
         private AbstractStage<TIn, TOut> _currentStage;
+#pragma warning restore CS0618 // Type or member is obsolete
         private readonly FlowShape<TIn, TOut> _shape;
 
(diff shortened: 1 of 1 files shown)


src/core/Akka.Streams/Stage/AbstractStage.cs
```diff
@@ -59,7 +63,9 @@ internal class PushPullGraphLogic<TIn, TOut> : GraphStageLogic, IDetachedContext
         /// <summary>
         /// TBD
         /// </summary>
+#pragma warning disable CS0618 // Type or member is obsolete
         public AbstractStage<TIn, TOut> Stage { get; }
```
Aaronontheweb, line 67: What stage are we supposed to replace `AbstractStage` with @eaba ?

Aaronontheweb (commented):
Left a question

src/core/Akka.Streams/Stage/AbstractStage.cs
eaba: "Please use GraphStage instead. [1.1.2]" @Aaronontheweb

eaba: https://github.com/akkadotnet/akka.net/blob/b688e5cf2e786f582e7e440210c194b6161f8b98/src/core/Akka.Streams/Stage/AbstractStage.cs#L389-L393

eaba: https://github.com/akkadotnet/akka.net/blob/b688e5cf2e786f582e7e440210c194b6161f8b98/src/core/Akka.Streams/Stage/Stage.cs#L33-L36

PR #6676, opened 2023-04-25
[CS0618][Stage] Warning Disable

eaba:
## Checklist

### DIFF
--- a/src/core/Akka.Streams/Implementation/Stages/Stages.cs
+++ b/src/core/Akka.Streams/Implementation/Stages/Stages.cs
@@ -479,7 +479,9 @@ namespace Akka.Streams.Implementation.Stages
     /// </summary>
     /// <typeparam name="TIn">TBD</typeparam>
     /// <typeparam name="TOut">TBD</typeparam>
+#pragma warning disable CS0618 // Type or member is obsolete
     public interface ISymbolicStage<in TIn, out TOut> : IStage<TIn, TOut>
+#pragma warning restore CS0618 // Type or member is obsolete
     {
         /// <summary>
         /// TBD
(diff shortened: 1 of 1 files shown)


src/core/Akka.Streams/Implementation/Stages/Stages.cs
```diff
@@ -479,7 +479,9 @@ public SymbolicGraphStage(ISymbolicStage<TIn, TOut> symbolicStage) : base(symbol
     /// </summary>
     /// <typeparam name="TIn">TBD</typeparam>
     /// <typeparam name="TOut">TBD</typeparam>
+#pragma warning disable CS0618 // Type or member is obsolete
     public interface ISymbolicStage<in TIn, out TOut> : IStage<TIn, TOut>
```
Aaronontheweb, line 483: Is it `IStage` that is obsolete? Any reason why it was marked as Obsolete (in the `Reason` field?)

Aaronontheweb (commented):
Left a question

src/core/Akka.Streams/Implementation/Stages/Stages.cs
eaba: https://github.com/akkadotnet/akka.net/blob/b688e5cf2e786f582e7e440210c194b6161f8b98/src/core/Akka.Streams/Stage/Stage.cs#L33-L36

Aaronontheweb: @eaba so let's use `GraphStage` instead of `IStage`, rather than disable the warning.

Aaronontheweb (changes requested):
Need to follow the warning's advice rather than disable it.

src/core/Akka.Streams/Implementation/Stages/Stages.cs
Aaronontheweb: Ah, I think I see what the issue is here - this is an interface not a class - therefore it can't be fixed via inheriting from GraphStage, got it.

Aaronontheweb (approved):
LGTM

PR #6683, opened 2023-04-26
[CS0420] `ActorCell` Warning Disable

eaba:
## Checklist


src/core/Akka/Actor/ActorCell.cs
```diff
@@ -91,7 +91,9 @@ public ActorCell(ActorSystemImpl system, IInternalActorRef self, Props props, Me
         /// <summary>
         /// TBD
         /// </summary>
+#pragma warning disable CS0420 // A reference to a volatile field will not be treated as volatile
         public Mailbox Mailbox => Volatile.Read(ref _mailboxDoNotCallMeDirectly);
+#pragma warning restore CS0420 // A reference to a volatile field will not be treated as volatile
```
eaba: 1. `A volatile field should not normally be passed using a ref or out parameter, since it will not be treated as volatile within the scope of the function. There are exceptions to this, such as when calling an interlocked API. As with any warning, you may use the [#pragma warning](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives#pragma-warning) to disable this warning in those rare cases where you are intentionally using a volatile field as a reference parameter.` https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0420?f1url=%3FappId%3Droslyn%26k%3Dk(CS0420)

2. `public Mailbox Mailbox => _mailboxDoNotCallMeDirectly;`

eaba (commented):
@Aaronontheweb tell me the solution for this

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -91,7 +91,9 @@ public ActorCell(ActorSystemImpl system, IInternalActorRef self, Props props, Me
         /// <summary>
         /// TBD
         /// </summary>
+#pragma warning disable CS0420 // A reference to a volatile field will not be treated as volatile
         public Mailbox Mailbox => Volatile.Read(ref _mailboxDoNotCallMeDirectly);
```
Aaronontheweb: We should probably remove the `Volatile.Read` call - since the field is already `volatile`.

Aaronontheweb (approved):
LGTM

PR #6685, opened 2023-04-26
[CS0420][Fix] `Mailbox` 

eaba:
## Checklist


eaba:
We should probably remove the Volatile.Read call - since the field is already volatile. @Aaronontheweb

Aaronontheweb:
> We should probably remove the Volatile.Read call - since the field is already volatile. @Aaronontheweb

Complicated issue from a semantics point of view, but I wonder if we should do this the other way around in order to guarantee atomicity https://github.com/dotnet/runtime/issues/67330#issuecomment-1083106423

PR #6702, opened 2023-04-29
[CS0612] `QueryEventsSql` Warning Disable

eaba:
## Checklist


src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
```diff
@@ -502,10 +502,12 @@ public abstract class AbstractQueryExecutor : IJournalQueryExecutor
                     @SerializerId
                 )";
 
+#pragma warning disable CS0612 // Type or member is obsolete
```
Aaronontheweb: Why is this `Obsolete`?

Aaronontheweb (commented):
When disabling an `Obsolete` warning @eaba , we should reference the Obsoletion message and explain why disabling the warning is the right call, versus doing something else (like removing the method.)

src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs
eaba: maybe is for this I don't know @Aaronontheweb 

https://github.com/akkadotnet/akka.net/blob/9676cc18f29ce81730d43fc019c9413f0e690533/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs#L563-L564

https://github.com/akkadotnet/akka.net/blob/9676cc18f29ce81730d43fc019c9413f0e690533/src/contrib/persistence/Akka.Persistence.Sql.Common/Journal/QueryExecutor.cs#L472-L477

Aaronontheweb: @eaba is that field used at all?

eaba: Aah! I have checked, it is not used. I want to remove it!!

Aaronontheweb (approved):
LGTM

PR #6707, opened 2023-05-02
Fix FastLazy race condition and waiting thread hanging (#6336)

### LINKED ISSUE #6336: Utility class `FastLazy` has a race condition
**Version Information**
Version 1.4.48 and later (including 1.5.x)
Package `Akka`

**Describe the bug**
The utility class `FastLazy` has a race condition that can lead to `Value` being produced multiple times. See [FastAtomicLazy Issue #2](https://github.com/Aaronontheweb/FastAtomicLazy/issues/2).

This issue has been resolved in the meantime, so please use the latest version.

But there is another issue with this class: [FastAtomicLazy Issue #12](https://github.com/Aaronontheweb/FastAtomicLazy/issues/12). This issue is only relevant if the producer function throws an exception.

**To Reproduce**
Since the issue is a race condition, there is no simple way to reproduce it. But it's an obvious Schroedinbug described in the linked issue.

**Expected behavior**
Only one thread should produce a value.

**Actual behavior**
The code in the `producer` funcion can be executed multiple times in very rare situations since the block calling `producer` is not entered in an atomic fashion.

**Environment**
Windows, but the issue is not dependent on the environment.

**Additional context**
The `FastLazy` implementation uses a `SpinWait` which can lead to high CPU load if the code executed in the `producer` function is not extremely short (only some IL instructions).

Are you shure that this condition is holding? If not I would avoid a `SpinWait`.

F0b0s:
Fixes #6336 

## Changes

Fix of two issues:
1. race condition that leads to multiple producer calls
2. waiting thread hanging on producer thread exception

## Checklist

### DIFF
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
@@ -4995,14 +4995,8 @@ namespace Akka.Util
     public sealed class FastLazy<T>
     {
         public FastLazy(System.Func<T> producer) { }
-        public bool IsValueCreated { get; }
         public T Value { get; }
-    }
-    public sealed class FastLazy<S, T>
-    {
-        public FastLazy(System.Func<S, T> producer, S state) { }
-        public bool IsValueCreated { get; }
-        public T Value { get; }
+        public bool IsValueCreated() { }
     }
     public interface ISurrogate
     {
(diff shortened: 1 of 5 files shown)


Aaronontheweb:
@F0b0s looks like we need API approvals still: https://getakka.net/community/contributing/api-changes-compatibility.html#approving-a-new-change

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -4995,14 +4995,8 @@ namespace Akka.Util
...
         public FastLazy(System.Func<T> producer) { }
-        public bool IsValueCreated { get; }
-        public T Value { get; }
-    }
-    public sealed class FastLazy<S, T>
-    {
-        public FastLazy(System.Func<S, T> producer, S state) { }
-        public bool IsValueCreated { get; }
         public T Value { get; }
+        public bool IsValueCreated() { }
```
F0b0s, line 5000: Changed to method because with 
`[MethodImpl(MethodImplOptions.AggressiveInlining)]`
it shows slightly better performance

Aaronontheweb (approved):
LGTM

PR #6709, opened 2023-05-02
added real bounds checking to `ByteString.Slice`

Aaronontheweb:
## Changes

Fixed a very weird bit of code that could result in byte buffers constantly allocating despite being of illegal size. This caused parts of Petabridge.Cmd's Akka.IO message decoding to run out of memory during message decoding.

## Checklist

### DIFF
--- a/src/core/Akka.Tests/Util/ByteStringSpec.cs
+++ b/src/core/Akka.Tests/Util/ByteStringSpec.cs
@@ -8,6 +8,7 @@
 using System.Linq;
 using System.Text;
 using Akka.IO;
+using FluentAssertions;
 using FsCheck;
 using Xunit;
 
@@ -44,8 +45,10 @@ namespace Akka.Tests.Util
         [Fact]
         public void A_ByteString_must_have_correct_size_when_slicing_from_index()
         {
-            Prop.ForAll((ByteString a, ByteString b) => (a + b).Slice(b.Count).Count == a.Count)
-                .QuickCheckThrowOnFailure();
+            var a = ByteString.FromBytes(new byte[]{ 1, 2, 3, 4, 5, 6, 7, 8, 9} );
+            var b = ByteString.FromBytes(new byte[] { 10, 11, 12, 13, 14, 15, 16, 17, 18 });
+
+            (a + b).Slice(b.Count).Count.Should().Be(a.Count);
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


src/core/Akka.Tests/Util/ByteStringSpec.cs
```diff
@@ -57,8 +60,10 @@ public void A_ByteString_must_be_sequential_when_slicing_from_start()
         [Fact]
         public void A_ByteString_must_be_sequential_when_slicing_from_index()
         {
-            Prop.ForAll((ByteString a, ByteString b) => (a + b).Slice(a.Count).SequenceEqual(b))
```
Aaronontheweb, line 60: Some of these property-based tests no longer pass because, as should be no surprise, the `ByteString.Slice` method will no longer work with arbitrary index and count values - hence why certain random-generated values fail during testing now.

src/core/Akka/Util/ByteString.cs
```diff
@@ -332,21 +332,21 @@ public ByteString Compact()
         /// </summary>
         /// <param name="index">index inside current <see cref="ByteString"/>, from which slicing should start</param>
         /// <param name="count">Number of bytes to fit into new <see cref="ByteString"/>.</param>
-        /// <returns></returns>
+        /// <exception cref="ArgumentOutOfRangeException">If index or count result in an invalid <see cref="ByteString"/>.</exception>
         public ByteString Slice(int index, int count)
         {
-            //TODO: this is really stupid, but previous impl didn't throw if arguments
```
Aaronontheweb, line 338: @Horusiath 's comment here was the first sign that there was something weird going on inside this method, when I started investigating how our Petabridge.Cmd code could run out of memory. Sure enough - this method will produce arbitrary byte arrays regardless of what the `index` and `count` values are, which caused our looping mechanism for decoding messages to go haywire.

If someone is depending on this old behavior - sorry, but that's your problem now - we're going back to byte arrays meaning what they mean  starting in Akka.NET V1.5.

Aaronontheweb (commented):
Detailed my changes

Aaronontheweb:
My ideal solution to this issue is https://github.com/akkadotnet/akka.net/pull/6708 - but that's going to need to wait until Akka.NET v1.6 most likely. We'll need to get rid of `SocketAsyncEventArgs` and refactor Akka.IO to use System.Memory or System.Buffers constructs instead.

Arkatufus (approved):
LGTM

PR #6718, opened 2023-05-03
Fix ReceiveAsync resetting ReceiveTimeout

### LINKED ISSUE #6653: ReceiveAsync paired with Timers results in actors never stopping from idle timeout
**Version Information**
Version of Akka.NET? 
1.4.47
Which Akka.NET Modules? 
Akka Dependency Injection (also 1.4.47)

**Describe the bug**
When setting up a periodic timer for an actor with a message that inherits INotInfluenceReceiveTimeout, the actor from which the timers were set is never stopped from inactivity, even when only these timed messages have been arriving beyond the receive timeout. This only happens when the message the timer sends is handled by ReceiveAsync, instead of Receive.

This seems to be due to the fact ReceiveAsync resets any pending ReceiveTimeout messages as part of it's flow, and so if the timer is producing messages more frequently than the ReceiveTimeout for the actor, it will never stop. The relevant code is in the ActorTaskScheduler class in the RunTask method. This calls into the actor context CheckReceiveTimeout() method, which appears to cancel any pending ReceiveTimeout messages and schedules a new one to be sent after the actors Receive Timeout amount has elapsed

**To Reproduce**
1. Setup a Periodic Timer in a ReceiveActor (StartPeriodicTimer)
2. Ensure the ReceiveTimeout is set on the actor, but at a duration that is higher than the frequency of the periodic timer
3. Register an async method for the object being sent from the periodic timer in the actor (e.g. ReceiveAsync)
4. Add appropriate logging and note that the actor will never stop from inactivity

**Expected behavior**
I would expect that given we are inheriting from INotInfluenceReceiveTimeout, the timed messages should not reset the inactive duration for the actor, and that the actor should eventually be sent a ReceiveTimeout message.

**Actual behavior**
Actor continues running due to the timer indirectly causing the inactive duration to reset through the ReceiveAsync method

**Environment**
Running Windows and .NET 6.0

**Additional context**
This may be expected and understood behavior from the ReceiveAsync method - I just couldn't find any documentation explaining it as thus, and this setup also feels like a gotcha in terms of perceived behavior and actual behavior.

Would be interested in the thoughts around this, and if there's other ways around doing a timer in an actor who's message is handled asynchronously from within the Akka library, without resorting to a custom implementation.

Arkatufus:
Fixes #6653

## Changes
* Pass in boolean flag to stop `CheackReceiveTimeout` from resetting every time.

### DIFF
--- a/src/core/Akka/Dispatch/ActorTaskScheduler.cs
+++ b/src/core/Akka/Dispatch/ActorTaskScheduler.cs
@@ -162,7 +162,7 @@ namespace Akka.Dispatch
                                   if (exception == null)
                                   {
                                       dispatcher.Resume(context);
-                                      context.CheckReceiveTimeout();
+                                      context.CheckReceiveTimeout(context.CurrentMessage is not INotInfluenceReceiveTimeout);
                                   }
                                   else
                                   {
--- a/src/core/Akka.Tests/Actor/ReceiveTimeoutSpec.cs
+++ b/src/core/Akka.Tests/Actor/ReceiveTimeoutSpec.cs
@@ -16,16 +16,17 @@ using Akka.Util.Internal;
 using FluentAssertions;
 using FluentAssertions.Extensions;
 using Xunit;
+using Xunit.Abstractions;
 
 
 namespace Akka.Tests.Actor
 {
+    public class Tick { }
+
+    public class TransparentTick : INotInfluenceReceiveTimeout { }
+    
     public class ReceiveTimeoutSpec : AkkaSpec
     {
-        public class Tick { }
-
-        public class TransparentTick : INotInfluenceReceiveTimeout { }
-
         public class TimeoutActor : ActorBase
         {
             private TestLatch _timeoutLatch;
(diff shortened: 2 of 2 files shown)


src/core/Akka/Dispatch/ActorTaskScheduler.cs
```diff
@@ -162,7 +162,7 @@ public static void RunTask(Func<Task> asyncAction)
                                   if (exception == null)
                                   {
                                       dispatcher.Resume(context);
-                                      context.CheckReceiveTimeout();
+                                      context.CheckReceiveTimeout(context.CurrentMessage is not INotInfluenceReceiveTimeout);
```
Arkatufus, line 165: `reschedule` flag defaults to `true`, if we don't set this, `ReceiveAsync` will always resets receive timeout.

Aaronontheweb: LGTM

Aaronontheweb (approved):
LGTM

PR #6721, opened 2023-05-03
use char based overloads for StartsWith, EndsWith, and Contains

SimonCropp:
faster execution path than the string based overloads


Aaronontheweb:
@SimonCropp looks like there's a compilation error here someplace

SimonCropp:
weird. something was wrong with that branch. i will resubmit

PR #6724, opened 2023-05-03
use Directory.Build.props convention

SimonCropp:


Aaronontheweb (approved):
This is a great change and has been on my chore backlog for some time.

PR #6750, opened 2023-05-10
harden ReliableDeliveryShardingSpecs

Aaronontheweb:
## Changes

Fixes the following test race condition caused by forcing the test to wait until a probe gracefully stops, which can cause a re-delivery.

```
Akka.Cluster.Sharding.Tests.Delivery.ReliableDeliveryShardingSpec.ReliableDelivery_with_Sharding_must_deliver_unconfirmed_if_ShardingConsumerController_is_terminated
Expected delivery4b.Message to be Job(msg-4), but found Job(msg-3).

With configuration:
- Use declared types and members
- Compare enums by value
- Match member by name (or throw)
- Without automatic conversion.
- Be strict about the order of items in byte arrays
```

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Delivery/ReliableDeliveryShardingSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Delivery/ReliableDeliveryShardingSpec.cs
@@ -86,7 +86,7 @@ public class ReliableDeliveryShardingSpec : TestKit.Xunit2.TestKit
             $"producer-{_idCount}");
 
         // expecting 3 end messages, one for each entity: "entity-0", "entity-1", "entity-2"
-        await consumerEndProbe.ReceiveNAsync(3, TimeSpan.FromSeconds(25)).ToListAsync();
+        await consumerEndProbe.ReceiveNAsync(3, TimeSpan.FromSeconds(5)).ToListAsync();
     }
 
     [Fact]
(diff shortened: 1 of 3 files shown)


src/contrib/cluster/Akka.Cluster.Sharding.Tests/Delivery/ReliableDeliveryShardingSpec.cs
```diff
@@ -413,7 +413,7 @@ public async Task
         }
 
         // redeliver also when no more messages are sent to the entity
-        await consumerProbes[1].GracefulStop(RemainingOrDefault);
+        Sys.Stop(consumerProbes[1]); // don't wait for termination
```
Aaronontheweb, line 417: This might not be the only fix this tests needs, FYI. No way to know for certain other than gathering more data.

Aaronontheweb:
@Arkatufus so we have a hard bug in the async TestKit here:

https://github.com/akkadotnet/akka.net/blob/4f27bac819bba8e78b89ad0e59913114a84bbee3/src/contrib/cluster/Akka.Cluster.Sharding.Tests/Delivery/ReliableDeliveryShardingSpec.cs#L131

25 seconds is what we allot - it used to be 15 seconds prior to my last PR to harden this test (issue is: actors weren't being given enough time to complete the messaging routine.)

Looking at the logs:

First

```
[INFO][05/10/2023 01:41:12.789Z][Thread 0014][Cluster (akka://test)] Cluster Node [15.0.0] - Node [akka.tcp://test@0.0.0.0:64654] is JOINING itself (with roles [], version [15.0.0]) and forming a new cluster
```

Last

```
[INFO][05/10/2023 01:41:18.679Z][Thread 0079][akka.tcp://test@0.0.0.0:64654/system/sharding/TestConsumer-1Coordinator] Singleton manager stopping singleton actor [akka://test/system/sharding/TestConsumer-1Coordinator/singleton]
```

That's a grand total of 6 seconds - and there's several other asserts that happen between the 25 second one and the one prior.

src/core/Akka.Tests/Delivery/TestConsumer.cs
```diff
@@ -73,12 +73,18 @@ private void Active()
...
 
-            if (EndCondition(job))
+            if (EndCondition(job) && _messageCount > 0)
             {
                 _log.Debug("End at [{0}]", job.SeqNr);
                 EndReplyTo.Tell(new Collected(_processed.Select(c => c.Item1).ToImmutableHashSet(), _messageCount + 1));
                 Context.Stop(Self);
             }
+            else if (EndCondition(job))
+            {
```
Aaronontheweb, line 85: This is the figure - if the `EndCondition` is satisfied before we've processed any messages, we're probably being recreated via the `ProducerController<T>` + `ShardRegion`, therefore we want to ignore without sending anymore `Collected` responses.

PR #6768, opened 2023-05-17
Fix MNTR crashing because it is using PolyFill extension method

Arkatufus:


src/core/Akka.Remote.TestKit/CommandLine.cs
```diff
@@ -42,7 +42,7 @@ public static void Initialize(string[] args)
             var fixedArgs = new List<string>();
             for (var i = 1; i < args.Length - 1; ++i)
             {
-                if (args[i].Equals("-Dmultinode") && args[i + 1].StartsWith('.'))
+                if (args[i].Equals("-Dmultinode") && args[i + 1].StartsWith("."))
```
Aaronontheweb, line 45: Changing from a char to a string does the trick here?

Arkatufus: Yes, the `char` variant is a polyfill extension method, the string variant is the `String` class pure method

PR #6775, opened 2023-05-25
Add Cluster.Sharding DData backward compatibility wire format mode

### LINKED ISSUE #6704: Akka.Cluster.Sharding: potential wire format problem when upgrading from v1.4 to v1.5 with `state-store-mode=ddata` and `remember-entities=on`
**Version Information**
Version of Akka.NET? v1.4 --> v1.5

**Describe the bug**

When upgrading an existing Akka.NET cluster that has `remember-entities=on` and `state-store-mode=ddata`, the following error can appear on the v1.4 nodes:

image

This appears to be a wire format issue - the user who reported this issue hadn't materially changed any of their code at all.

**To Reproduce**

We need to reproduce this to see if this is an oversight in our management of the wire format for v1.5 for this specific group of users.

1. Clone https://github.com/petabridge/akkadotnet-code-samples/tree/master/src/clustering/sharding-sqlserver
2. Configure to run in DData mode using v1.4 binaries with remember-entities=on - populate some data (https://github.com/petabridge/akkadotnet-code-samples/blob/8ff65149ee4cbdbcadd6f5c39ff2716a343d9f61/src/clustering/sharding-sqlserver/SqlSharding.Host/Program.cs#L74-L76)
3. Run at least 3 nodes
4. Upgrade to v1.5 - only deploy a single v1.5 node with the other two remaining at v1.4
5. See if the error occurs.

**Expected behavior**

Upgrade should occur without any errors.

**Actual behavior**

To be determined.

Arkatufus:
Fixes #6704

## Changes

- Add `akka.cluster.sharding.distributed-data.backward-compatible-wire-format` to allow v1.5 to talk with v1.4
- Add DData hack to convert v1.4 to v1.5 data structure serialization

## NOTES

- When `backward-compatible-wire-format` is set to true, `Cluster.Sharding` will not be able to talk to any Cluster.Sharding v1.5.0-v1.5.7, so if `Cluster.Sharding` were to talk to v1.4.x, all v1.5 nodes have to be updated to the version with this fix.

### DIFF
--- a/src/contrib/cluster/Akka.DistributedData/Serialization/ReplicatedDataSerializer.cs
+++ b/src/contrib/cluster/Akka.DistributedData/Serialization/ReplicatedDataSerializer.cs
@@ -66,9 +66,13 @@ namespace Akka.DistributedData.Serialization
 
         private readonly byte[] _emptyArray = Array.Empty<byte>();
 
+        private readonly bool _backwardCompatWireFormat;
+        
         public ReplicatedDataSerializer(ExtendedActorSystem system) : base(system)
         {
             _ser = new SerializationSupport(system);
+            _backwardCompatWireFormat =
+                system.Settings.Config.GetBoolean("akka.cluster.sharding.distributed-data.backward-compatible-wire-format");
         }
 
 
@@ -735,6 +739,11 @@ namespace Akka.DistributedData.Serialization
             pLww.State = _ser.OtherMessageToProto(register.Value);
             pLww.Timestamp = register.Timestamp;
             pLww.TypeInfo = GetTypeDescriptor(r.RegisterType);
+            
+            // HACK: Really really ugly hack to make sure that v1.5 DData cluster sharding works with v1.4
+            if(_backwardCompatWireFormat && pLww.TypeInfo.TypeName == "Akka.Cluster.Sharding.ShardCoordinator+CoordinatorState, Akka.Cluster.Sharding")
+                pLww.TypeInfo.TypeName = "Akka.Cluster.Sharding.PersistentShardCoordinator+State, Akka.Cluster.Sharding";
+            
             return pLww;
         }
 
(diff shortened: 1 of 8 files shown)


Aaronontheweb (changes requested):
Do we need to add a simple regression test to the sharding message serializer to ensure that this fix works?

Arkatufus:
The fix was not in the sharding serializer, its in the DData `LWWRegister<T>` serializer

Aaronontheweb:
@Arkatufus  is it feasible to unit test this fix?

Arkatufus:
Well, the target class does not exist in v1.5 binary anymore, testing would require a v1.4 binary protobuf loaded from disk and see if the serializer can deserialize it?

Arkatufus:
Actually, we can probably test the `ToProto` methods to see if it generates the correct `TypeInfo`, ~~but that would require us to change the method from private to internal~~

Arkatufus:
Done

src/contrib/cluster/Akka.DistributedData/Serialization/ReplicatedDataSerializer.cs
```diff
@@ -735,6 +739,11 @@ private Proto.Msg.LWWRegister LWWToProto<T>(ILWWRegister r)
             pLww.State = _ser.OtherMessageToProto(register.Value);
             pLww.Timestamp = register.Timestamp;
             pLww.TypeInfo = GetTypeDescriptor(r.RegisterType);
+            
+            // HACK: Really really ugly hack to make sure that v1.5 DData cluster sharding works with v1.4
+            if(_backwardCompatWireFormat && pLww.TypeInfo.TypeName == "Akka.Cluster.Sharding.ShardCoordinator+CoordinatorState, Akka.Cluster.Sharding")
```
Aaronontheweb, line 744: LGTM

Aaronontheweb (approved):
LGTM

Issue #4698, opened 2020-12-29
Non-documented slow performance of Props.Create(Expression) overload


mralexes:
- Which Akka.Net version you are using: 1.4.13
- On which platform you are using Akka.Net: .Net Core 3.1 (both Windows and Linux)
- A list of steps to reproduce the issue: Run next benchmark code:
``` csharp
using Akka.Actor;
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Running;

namespace PropsCreateBenchmark
{
    class Program
    {
        static void Main(string[] args)
        {
            BenchmarkRunner.Run<PropsBenchmark>();
        }
    }

    [SimpleJob]
    public class PropsBenchmark
    {
        private static readonly string Arg = "actor arg";

        [Benchmark(Baseline = true)]
        public Props UsualProps()
        {
            var props = Props.Create<Actor>(Arg);
            return props;
        }

        [Benchmark]
        public Props ExpressionsProps()
        {
            var props = Props.Create(() => new Actor(Arg));
            return props;
        }
    }

    public class Actor : ReceiveActor
    {
        public string Arg { get; }

        public Actor(string arg)
        {
            Arg = arg;
        }
    }
}

```

And check results

``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.18363.592 (1909/November2018Update/19H2)
Intel Core i5-8600K CPU 3.60GHz (Coffee Lake), 1 CPU, 6 logical and 6 physical cores
.NET Core SDK=5.0.100
  [Host]     : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  DefaultJob : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
```
|           Method |         Mean |        Error |       StdDev |    Ratio | RatioSD |
|----------------- |-------------:|-------------:|-------------:|---------:|--------:|
|       UsualProps |     56.54 ns |     1.175 ns |     1.306 ns |     1.00 |    0.00 |
| ExpressionsProps | 56,668.76 ns | 1,096.394 ns | 1,262.609 ns | 1,002.71 |   31.31 |

As you can see from the benchmark report, creation of Props using expression is very slow and impacts performance in scenarios with a big amount of short-lived or dynamically-created actors. 
[...]

PR #6805, opened 2023-06-19
Do not use expression based options for backoff props

### LINKED ISSUE #6804: [PERF] Backoff Supervisor uses Expression based Props
Related to #4698 

**Version Information**
1.4.x

Backoff Supervisor Props use Expressions, which has a large perf impact for cases where an actor is created semi-transiently but still needs backoff.

**Data and Specs**
See #4698 for details. The benchmarks there illustrate the overall concern well enough.

**Expected behavior**
Don't use Expression based props where it's possible transient actors are being created.

**Actual behavior**
We use expression based props and it hurts perf by a significant margian

**Environment**
N/A

**Additional context**
This is not as big a problem for long lived actors.

We should still consider having an issue for other internals of akka where this may come up, even if it means more diligence from maintainers (although, FWIW, unit tests should cover when someone gets it wrong).

to11mtm:
Fixes #6804. See #4698 for background.

## Changes

Change `BackoffOptions` props to be non-expression based

### DIFF
--- a/src/core/Akka/Pattern/BackoffOnRestartSupervisor.cs
+++ b/src/core/Akka/Pattern/BackoffOnRestartSupervisor.cs
@@ -24,6 +24,10 @@ namespace Akka.Pattern
         private readonly OneForOneStrategy _strategy;
         private readonly ILoggingAdapter _log = Context.GetLogger();
 
+        /// <devremarks>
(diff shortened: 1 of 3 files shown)


Aaronontheweb (approved):
LGTM

Aaronontheweb:
We'll forward port this to v1.5 as well.

PR #6807, opened 2023-06-19
Do not use expression based props for long lived streams

### LINKED ISSUE #6806: [PERF] ActorGraphInterpreter and other Akka streams internals use Expression based props
Related to #4698 

**Version Information**
1.4.x

`ActorGraphInterpreter` Props static method uses Expressions, which has a large perf impact for cases where streams are created in a transient/semi-transient (i.e. stream per actor lifecycle) fashion.

Additionally some stages use this as well, and, per above, we shouldn't.

**Data and Specs**
See #4698 for details. The benchmarks there illustrate the overall concern well enough.

**Expected behavior**
Don't use Expression based props where it's possible transient actors are being created.

**Actual behavior**
We use expression based props and it hurts perf by a significant margin

**Environment**
N/A

**Additional context**
This is not as big a problem for single streams created.

This should almost certainly be covered by unit tests, as `ActorGraphInterpreter` uses a single arg at present and is not touched frequently.

to11mtm:
Fixes #6806 . See https://github.com/akkadotnet/akka.net/issues/4698 for background.

## Changes

Akka.Streams no longer uses Expression based Props (at least, in easily located scenarios.)

### DIFF
--- a/src/core/Akka.Streams/Implementation/ActorMaterializerImpl.cs
+++ b/src/core/Akka.Streams/Implementation/ActorMaterializerImpl.cs
@@ -648,7 +648,7 @@ namespace Akka.Streams.Implementation
         /// </summary>
         /// <param name="settings">TBD</param>
         /// <param name="haveShutdown">TBD</param>
-        /// If this changes you must also change <see cref="StreamSupervisor.Props"/> as well!
+        /// If this changes you must also change StreamSupervisor.Props as well!
         public StreamSupervisor(ActorMaterializerSettings settings, AtomicBoolean haveShutdown)
         {
             Settings = settings;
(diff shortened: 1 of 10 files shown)


src/core/Akka.Streams/Implementation/ActorMaterializerImpl.cs
```diff
@@ -648,6 +648,7 @@ public static Props Props(ActorMaterializerSettings settings, AtomicBoolean have
         /// </summary>
         /// <param name="settings">TBD</param>
         /// <param name="haveShutdown">TBD</param>
+        /// If this changes you must also change <see cref="StreamSupervisor.Props"/> as well!
```
SimonCropp: this is not a valid xml comment

Aaronontheweb: Fixed

PR #6818, opened 2023-06-28
Akka.Remote: log all layers of wrapped messages during errors

Aaronontheweb:
## Changes

Make sure we capture what the real underlying message types are when we encounter issues like serialization errors, oversized payload exceptions, and so on.

## Checklist

### DIFF
--- a/src/core/Akka.Remote.Tests/RemotingSpec.cs
+++ b/src/core/Akka.Remote.Tests/RemotingSpec.cs
@@ -11,6 +11,7 @@ using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Actor.Dsl;
 using Akka.Configuration;
+using Akka.Event;
 using Akka.Remote.Transport;
 using Akka.Routing;
 using Akka.TestKit;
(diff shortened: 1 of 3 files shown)


src/core/Akka.Remote.Tests/RemotingSpec.cs
```diff
@@ -651,6 +652,21 @@ await EventFilter.Exception<OversizedPayloadException>(start: "Discarding oversi
...
+        [Fact]
+        public void Log_Wrapped_messages_that_fail_to_Send()
+        {
+            // 2x wrapped message
+            var wrappedMessage =
+                new DeadLetter(new ActorSelectionMessage("hit", Array.Empty<SelectionPathElement>(), false), TestActor,
+                    TestActor);
+
+            var loggedType = EndpointWriter.LogPossiblyWrappedMessageType(wrappedMessage);
+            loggedType.Should().Contain("DeadLetter").And.Contain("ActorSelectionMessage").And.Contain("String");
```
Aaronontheweb, line 667: Simple unit test to assert that our log formatting helper captures all of the layers correctly.

This test produces the following output in the log:

```
Akka.Event.DeadLetter-->(Akka.Actor.ActorSelectionMessage-->System.String)
```

src/core/Akka.Remote/Endpoint.cs
```diff
@@ -1486,7 +1522,7 @@ private bool WriteSend(EndpointManager.Send send)
                         string.Format("Discarding oversized payload sent to {0}: max allowed size {1} bytes, actual size of encoded {2} was {3} bytes.",
                             send.Recipient,
                             Transport.MaximumPayloadBytes,
-                            send.Message.GetType(),
+                            LogPossiblyWrappedMessageType(send.Message),
```
Aaronontheweb, line 1525: Captured all of the relevant `EndpointWriter` callsites here - unfortunately, on the `EndpointReader` side we don't know what the underlying type is because it fails during deserialization, which is where we would extract that kind of information.

Aaronontheweb (commented):
This is primarily a debuggability improvement - it doesn't really impact the execution of Akka.Remote much. I'm looking to add this in order to track down some Akka.Cluster.DistributedData improvement areas related to large payload sizes.

Arkatufus (approved):
LGTM

PR #6828, opened 2023-07-10
Improve the code readability of client actor inherited method

jokoyoski:
In the refactored code:

Each if condition is kept separate for clarity and readability.
The code inside each condition is left unchanged.
The return true statement is placed after each block of code to indicate that the condition is satisfied and the execution should continue.
The final return false statement is kept outside of any condition, indicating that none of the conditions were satisfied and the default case is reached.
These changes improve the code's structure and maintain its original logic.

### DIFF
--- a/src/benchmark/PingPong/ClientActorBase.cs
+++ b/src/benchmark/PingPong/ClientActorBase.cs
@@ -28,36 +28,38 @@ namespace PingPong
 
         protected override bool Receive(object message)
         {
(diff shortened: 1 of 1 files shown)


Aaronontheweb (dismissed):
LGTM

Aaronontheweb:
Actually nah this doesn't make a lot of sense, and I can see this user has sent PRs to a bunch of other popular .NET repos with similarly low-effort stuff.

PR #6837, opened 2023-07-14
Harden LWWDictionary serialization null check

Arkatufus:
Harden LWWDictionary delta operation de/serialization with null checks to track down possible NRE bug.

## Changes

* Add null value checks for LWWDictionaryDelta to track down possible NRE bugs


Aaronontheweb (approved):
LGTM

PR #6841, opened 2023-07-18
Removed nesting levels in certain places through block-scoped `using` statement and inversion of `if` blocks

Lehonti:
## Changes

No behavior is  changed. Code is just leveraging more modern constructs to improve readability

### Benchmarks

There shouldn't be any performance changes

### DIFF
--- a/src/benchmark/SpawnBenchmark/Program.cs
+++ b/src/benchmark/SpawnBenchmark/Program.cs
@@ -17,14 +17,11 @@ namespace SpawnBenchmark
         static void Main(string[] args)
         {
             Console.WriteLine($"Is Server GC {GCSettings.IsServerGC}");
-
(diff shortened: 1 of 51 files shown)


Aaronontheweb:
Thanks! I'm going to need to take a look at the test suite - these look like racy tests.

Aaronontheweb:
Did you use an auto-formatter to make these changes?

Lehonti:
> Did you use an auto-formatter to make these changes?

I used Visual Studio to do most of the busy work, but I formatted some things manually, too.

decided to keep a formatting that was similar to the original, which means that several `using` statements span several lines. If it were my own code I would do it differently, though.

Aaronontheweb:
Yeah I don't think I'm going to merge this - touches way too many files with changes that I can't really review. If you want to help with formatting, please do it in smaller bite-sized PRs.

PR #6856, opened 2023-07-27
DistributedPubSub: make query to count local subscribers for topic public

### LINKED ISSUE #3663: Enhancement: add ability to retrieve subscriber / topic stats from DistributedPubSubMediator

Aaronontheweb:
## Changes

close #3663 - between this and `Topic`s all local data should be queryable for distributed pub-sub

## Checklist

* [x] Design discussion issue #3663

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
@@ -42,9 +42,9 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Internal
     }
 
     /// <summary>
-    /// TBD
+    /// Get all subscribers for a given topic.
     /// </summary>
-    internal sealed class CountSubscribers
+    public sealed class CountSubscribers
     {
         public string Topic { get; }
 
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
@@ -10,6 +10,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using Akka.Actor;
+using Akka.Annotations;
 using Akka.Event;
 using Akka.Routing;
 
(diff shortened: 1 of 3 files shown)


src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
```diff
@@ -42,9 +42,9 @@ internal sealed class Count
     }
 
     /// <summary>
-    /// TBD
+    /// Get all subscribers for a given topic.
     /// </summary>
-    internal sealed class CountSubscribers
+    public sealed class CountSubscribers
```
Aaronontheweb, line 49: Just made this `internal` class `public` - left it inside the `.Internal` namespace though (which I think this fine - it's public but we reserve the right to change it, per our API guidelines: https://getakka.net/community/contributing/api-changes-compatibility.html#akkanet-api-versioning-policy )

Arkatufus: Might want to decorate this class with `[ApiMayChange]` attribute for documentation purposes

Arkatufus (changes requested):
LGTM, just a nitpick

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
Aaronontheweb: Will do

Arkatufus (approved):

PR #6862, opened 2023-07-31
Expose `Tags` property in `EventEnvelope`

### LINKED ISSUE #6849: Add `Tags` property to `Akka.Persistence.Query` `EventEnvelope`
Currently, event tags are write-only construct that were written along side of the event in question. There is no way to inspect/retrieve these tags once they are written.
It would be trivial to return these tags as part of the event metadata inside the `EventEnvelope` class that are returned by any of the persistence queries.
This property can be made optional (nullable) so that existing plugins can opt in on implementing it.

Arkatufus:
## Changes

close #6849

* Expose a new property `Tags` in `EventEnvelope`

The `Tags` property is, by default, an empty array. It is up to the plugin implementation to populate it.

The built-in SQLite and in-mem journal is set to populate the property only for `EventsByTag` and `CurrentEventsByTag` queries, all SQL based plugin "should" automatically inherit this. It is not populated on any other persistence queries at the moment. 

Full support is planned for `Akka.Persistence.Sql` plugin.

Changes have been tested on `Akka.Persistence.SqlServer` and `Akka.Persistence.MongoDb` to see if it breaks any API or unit test, they worked just fine.

### DIFF
--- a/src/core/Akka.Persistence.Query/EventEnvelope.cs
+++ b/src/core/Akka.Persistence.Query/EventEnvelope.cs
@@ -26,7 +26,7 @@ namespace Akka.Persistence.Query
         /// </summary>
         [Obsolete("For binary compatibility with previous releases. Since 1.4.14")]
         public EventEnvelope(Offset offset, string persistenceId, long sequenceNr, object @event)
-            : this(offset, persistenceId, sequenceNr, @event, 0L, null)
+            : this(offset, persistenceId, sequenceNr, @event, 0L, Array.Empty<string>())
         { }
         
         /// <summary>
@@ -34,17 +34,17 @@ namespace Akka.Persistence.Query
         /// </summary>
         [Obsolete("For binary compatibility with previous releases. Since 1.5.11")]
         public EventEnvelope(Offset offset, string persistenceId, long sequenceNr, object @event, long timestamp)
-            : this(offset, persistenceId, sequenceNr, @event, timestamp, null)
+            : this(offset, persistenceId, sequenceNr, @event, timestamp, Array.Empty<string>())
         { }        
 
-        public EventEnvelope(Offset offset, string persistenceId, long sequenceNr, object @event, long timestamp, string[]? tags)
+        public EventEnvelope(Offset offset, string persistenceId, long sequenceNr, object @event, long timestamp, string[] tags)
         {
             Offset = offset;
             PersistenceId = persistenceId;
             SequenceNr = sequenceNr;
             Event = @event;
             Timestamp = timestamp;
-            Tags = tags;
+            Tags = tags ?? Array.Empty<string>();
         }        
 
         public Offset Offset { get; }
@@ -57,7 +57,7 @@ namespace Akka.Persistence.Query
         
         public long Timestamp { get; }
 
-        public string[]? Tags { get; }
+        public string[] Tags { get; }
 
         public bool Equals(EventEnvelope? other)
         {
--- a/src/core/Akka.Persistence.Query/EventEnvelope.cs
+++ b/src/core/Akka.Persistence.Query/EventEnvelope.cs
@@ -7,6 +7,7 @@
 
 using System;
 
+#nullable enable
 namespace Akka.Persistence.Query
 {
     /// <summary>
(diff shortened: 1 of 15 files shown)


src/core/Akka.Persistence.Query/EventEnvelope.cs
```diff
@@ -23,21 +24,27 @@ public sealed class EventEnvelope : IEquatable<EventEnvelope>
...
-            : this(offset, persistenceId, sequenceNr, @event, 0L)
+            : this(offset, persistenceId, sequenceNr, @event, 0L, null)
         { }
         
         /// <summary>
         /// Initializes a new instance of the <see cref="EventEnvelope"/> class.
         /// </summary>
+        [Obsolete("For binary compatibility with previous releases. Since 1.5.11")]
         public EventEnvelope(Offset offset, string persistenceId, long sequenceNr, object @event, long timestamp)
+            : this(offset, persistenceId, sequenceNr, @event, timestamp, null)
```
Aaronontheweb: Want to use `Array.Empty<string>` here instead of a nullable array?

src/core/Akka.Persistence.Query/EventEnvelope.cs
```diff
@@ -50,28 +57,30 @@ public EventEnvelope(Offset offset, string persistenceId, long sequenceNr, objec
         
         public long Timestamp { get; }
 
-        public bool Equals(EventEnvelope other)
+        public string[]? Tags { get; }
```
Aaronontheweb: See previous comment about non-nullable `Array.Empty<string>`

src/core/Akka.Persistence.TCK/Query/CurrentEventsByTagSpec.cs
```diff
@@ -59,30 +65,32 @@ public virtual void ReadJournal_query_CurrentEventsByTag_should_find_existing_ev
             var greenSrc = queries.CurrentEventsByTag("green", offset: NoOffset());
             var probe = greenSrc.RunWith(this.SinkProbe<EventEnvelope>(), Materializer);
             probe.Request(2);
-            probe.ExpectNext<EventEnvelope>(p => p.PersistenceId == "a" && p.SequenceNr == 2L && p.Event.Equals("a green apple"));
-            probe.ExpectNext<EventEnvelope>(p => p.PersistenceId == "a" && p.SequenceNr == 4L && p.Event.Equals("a green banana"));
+            ExpectEnvelope(probe, "a", 2L, "a green apple", "green");
```
Aaronontheweb, line 68: LGTM

src/core/Akka.Persistence.TCK/Query/CurrentEventsByTagSpec.cs
```diff
@@ -188,7 +214,9 @@ public virtual void ReadJournal_query_CurrentEventsByTag_should_see_all_150_even
         [Fact]
         public void ReadJournal_query_CurrentEventsByTag_should_include_timestamp_in_EventEnvelope()
         {
-            var queries = ReadJournal as ICurrentEventsByTagQuery;
+            if (ReadJournal is not ICurrentEventsByTagQuery queries)
+                throw IsTypeException.ForMismatchedType(nameof(ICurrentEventsByTagQuery), ReadJournal?.GetType().Name ?? "null");
```
Aaronontheweb, line 218: LGTM - shouldn't run these specs for journals that don't support it / should let the implementor know they've made a boo-boo.

src/core/Akka.Persistence.TCK/Query/CurrentEventsByTagSpec.cs
```diff
@@ -205,5 +233,19 @@ public void ReadJournal_query_CurrentEventsByTag_should_include_timestamp_in_Eve
...
+        private EventEnvelope ExpectEnvelope(TestSubscriber.Probe<EventEnvelope> probe, string persistenceId, long sequenceNr, string @event, string tag)
+        {
+            var envelope = probe.ExpectNext<EventEnvelope>(_ => true);
+            envelope.PersistenceId.Should().Be(persistenceId);
+            envelope.SequenceNr.Should().Be(sequenceNr);
+            envelope.Event.Should().Be(@event);
+            if (SupportsTagsInEventEnvelope)
+            {
+                envelope.Tags.Should().NotBeNull();
+                envelope.Tags.Should().Contain(tag);
```
Aaronontheweb, line 246: LGTM

src/core/Akka.Persistence.TCK/Query/EventsByTagSpec.cs
```diff
@@ -15,17 +15,21 @@
 using FluentAssertions;
 using Xunit;
 using Xunit.Abstractions;
+using Xunit.Sdk;
 using static Akka.Persistence.Query.Offset;
 
+#nullable enable
```
Aaronontheweb, line 21: :+1:

Aaronontheweb (changes requested):
One small comment, but otherwise LGTM

src/core/Akka.Persistence.Query/EventEnvelope.cs
Arkatufus: good idea

Aaronontheweb:
@Arkatufus I think you need to update API approval here

Arkatufus:
API Approval list updated

Aaronontheweb (approved):

CumpsD:
Nice one 👍🏻

PR #6865, opened 2023-08-01
cap total number of iterations for Sharding end to end BDN

Aaronontheweb:
## Changes

can't wait all day for these to finish


src/benchmark/Akka.Cluster.Benchmarks/Sharding/ShardMessageRoutingBenchmarks.cs
```diff
@@ -15,18 +15,20 @@
...
 using BenchmarkDotNet.Diagnostics.dotTrace;
+using BenchmarkDotNet.Engines;
 using static Akka.Cluster.Benchmarks.Sharding.ShardingHelper;
 
 namespace Akka.Cluster.Benchmarks.Sharding
 {
-    [DotTraceDiagnoser]
+    //[DotTraceDiagnoser]
     [Config(typeof(MonitoringConfig))]
+    [SimpleJob(RunStrategy.Monitoring, launchCount: 10, warmupCount: 10)]
```
Aaronontheweb, line 25: max 20 iterations of each benchmark variant (and they run 2x when `DotTraceDiagnoser` is enabled)

PR #6886, opened 2023-08-16
Improve Streams Throttle. Log errors, improve tests, and add supervisor strategy support

### LINKED ISSUE #6406: [Akka.Streams] Supervisor strategy is not applied when `calculateCost` throws an exception in `Throttle`
**Version Information**
Version of Akka.NET? 1.4.48

**Describe the bug**
When using a `.Throttle` stage in `Akka.Streams` with cost-based throttling arguments, if a `calculateCost` function throws an exception, and a stream has a `SupervisorStrategy` attached, it is ignored.

**To Reproduce**
Write two `xUnit` tests and compare results below. Test case is, make a `Source` that emits 5 elements and throttle cost throws an exception when the value is odd. In real life, this can be anything, from external API error or simple division by zero. When we attach a `ResumingDecider`, elements `1,3,5` should pass to the sink and `2,4` should be dropped. However:

```
[Fact]
public async Task FailingTest()
{
  var src = Source.From(new List<int> { 1, 2, 3, 4, 5 }).Throttle(
		  cost: 1,
		  per: TimeSpan.FromSeconds(1),
		  maximumBurst: 10,
		  calculateCost: e => e % 2 == 0 ? throw new Exception() : 1,
		  mode: ThrottleMode.Shaping)
	  .WithAttributes(new Attributes(new ActorAttributes.SupervisionStrategy(Deciders.ResumingDecider)));
			  
  var result = await src.RunWith(Sink.Seq<int>(), this.akkaFixture.Materializer);
  
  Assert.Equal(3, result.Count);
}
```

Running this test will fail with `System.Exception was thrown`. Also, when attaching a debugger, it doesn't stop at `ResumingDecider` code breakpoints - thus it is not called at all.
 
Now the second one:
```
[Fact]
public async Task WorkingTest()
{
  var src = Source.From(new List<int> { 1, 2, 3, 4, 5 })
                .Select(e => e % 2 == 0 ? throw new Exception() : (e, 1))
                .Throttle(
                    cost: 1,
                    per: TimeSpan.FromSeconds(1),
                    maximumBurst: 10,
                    calculateCost: e => e.Item2,
                    mode: ThrottleMode.Shaping)
                .Select(v => v.e)
                .WithAttributes(new Attributes(new ActorAttributes.SupervisionStrategy(Deciders.ResumingDecider)));
			  
  var result = await src.RunWith(Sink.Seq<int>(), this.akkaFixture.Materializer);
  
  Assert.Equal(3, result.Count);
}
```

This test will pass w/o issues.

**Expected behavior**
`SupervisionStrategy` and associated `Decider` is called when a `Throttle` stage fails to compute the cost via `calculateCost` function, throwing an exception from inside that function.

**Actual behavior**
`SupervisionStrategy`and associated `Decider` is **not** called when a `Throttle` stage fails to compute the cost via `calculateCost` function, throwing an exception from inside that function.

**Environment**
Reproduces on Linux, Windows, container/desktop, .NET6

Arkatufus:
Fixes #6406

## Changes

* Add supervisor strategy support for failing colculateCost delegate
* Add unit test
* Rewrite nonsensical tests to proper async tests

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/FlowThrottleSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/FlowThrottleSpec.cs
@@ -13,14 +13,17 @@ using System.Threading.Tasks;
 using Akka.IO;
 using Akka.Streams.Actors;
 using Akka.Streams.Dsl;
+using Akka.Streams.Supervision;
 using Akka.Streams.TestKit;
 using Akka.TestKit;
+using Akka.TestKit.Extensions;
 using Akka.TestKit.Xunit2.Attributes;
 using Akka.Util.Internal;
 using FluentAssertions;
+using FluentAssertions.Extensions;
 using Xunit;
 using Xunit.Abstractions;
-// ReSharper disable InvokeAsExtensionMethod
+using static FluentAssertions.FluentActions;
 
 namespace Akka.Streams.Tests.Dsl
 {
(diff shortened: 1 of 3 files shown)


src/core/Akka.Streams.Tests/Dsl/FlowThrottleSpec.cs
```diff
@@ -46,66 +49,101 @@ private static ByteString GenerateByteString(int length)
             return ByteString.FromBytes(bytes);
         }
 
+        [Fact(DisplayName = "Throttle with delegate calculateCost must resume when delegate throws")]
```
Aaronontheweb, line 52: LGTM

src/core/Akka.Streams/Dsl/Internal/InternalFlowOperations.cs
```diff
@@ -1893,7 +1893,7 @@ public SplitAfterMergeBack(IFlow<TOut, TMat> self, Func<TOut, bool> predicate, S
             int maximumBurst, ThrottleMode mode)
         {
             if (elements <= 0) throw new ArgumentException("Throttle elements must be > 0", nameof(elements));
-            if (per == TimeSpan.Zero) throw new ArgumentException("Throttle per timeout must not be zero", nameof(per));
+            if (per <= TimeSpan.Zero) throw new ArgumentException("Throttle per timeout must not be less than or equal to zero", nameof(per));
```
Aaronontheweb, line 1896: Nice catch

src/core/Akka.Streams/Implementation/Throttle.cs
```diff
@@ -29,35 +31,62 @@ private sealed class Logic : TimerGraphStageLogic, IInHandler, IOutHandler
...
-                    ScheduleOnce(TimerName, TimeSpan.FromTicks(delayTicks));
+                    var cost = _stage._costCalculation(element);
+                    var delayTicks = _tokenBucket.Offer(cost);
+
+                    if (delayTicks == 0)
+                        Push(_stage.Outlet, element);
+                    else
+                    {
+                        if (_enforcing)
+                            throw new OverflowException("Maximum throttle throughput exceeded.");
```
Aaronontheweb: Problem: if we're enforcing, do we need to fail regardless of what the supervision strategy says?

Aaronontheweb (changes requested):
Need to address issue of what happens when we're "enforcing" throttling?

src/core/Akka.Streams/Implementation/Throttle.cs
Arkatufus: I'm not sure, what do you think is best?

Arkatufus: I think it should fail regardless, I'll change the code

Aaronontheweb: I agree

Aaronontheweb (approved):
LGTM

PR #6887, opened 2023-08-17
remove redundant private set

SimonCropp:


Aaronontheweb (changes requested):
Looks like these changes broke a few of the tests in `Akka.Tests.Serialization.SerializationSpec`

SimonCropp:
hmm. if your serialization relies on private fields. probs best to kill this

PR #6888, opened 2023-08-17
fix broken editor config

SimonCropp:
when u have a single extension., eg `[*.{csproj}]`,  results the `{}` being treated as literals and not a group. so the settings for `{cs}` and `{csproj}` were being ignored. also having a second `[*.cs]` results in it overriding the first one, so i merged them


Aaronontheweb (approved):
LGTM

PR #6906, opened 2023-09-07
Fix possible NRE bug in Dispatchers

Arkatufus:


src/core/Akka/Dispatch/Dispatchers.cs
```diff
@@ -372,10 +372,10 @@ public static Config GetConfig(Config config, string id, int depth = 0)
         public Dispatchers(ActorSystem system, IDispatcherPrerequisites prerequisites, ILoggingAdapter logger)
         {
             _system = system;
+            _logger = logger;
```
Aaronontheweb, line 375: LGTM

Aaronontheweb (approved):
LGTM

PR #6932, opened 2023-09-28
Sealed uninherited `internal` classes in `Akka.Streams`

Lehonti:
## Changes
Adding `sealed` to `class`es that:
- Are in the `Akka.Streams` project
- Are `internal` or `private` (i.e. not part of the public API)
- Don't define any new `protected` or `virtual` members
- Aren't inherited by any other class

This helps us ensure more invariants (in this case, the certainty that the code of the class doesn't need to take inherited classes into account) at the language level. This could make the logic easier to understand

### DIFF
--- a/src/core/Akka.Streams.Tests/Implementation/FixedBufferSpec.cs
+++ b/src/core/Akka.Streams.Tests/Implementation/FixedBufferSpec.cs
@@ -164,8 +164,8 @@ namespace Akka.Streams.Tests.Implementation
                     buf.Dequeue().Should().Be(elem);
             }
         }
-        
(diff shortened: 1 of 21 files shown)


Aaronontheweb:
> This helps us ensure more invariants (in this case, the certainty that the code of the class doesn't need to take inherited classes into account) at the language level. This could make the logic easier to understand

This will also speed up certain types of operations with these classes too. Thanks @Lehonti - we'll review these changes.

to11mtm (approved):

Aaronontheweb (approved):
LGTM

PR #6935, opened 2023-09-29
Streams ensure cts cancel fix statecheck

### LINKED ISSUE #6903: Source from IAsyncEnumerable is not disposed on shutdown, while interrupted because of back-pressure
**Version Information**
1.5.12
Akka, Akka.Streams

**Describe the bug**
Assume stream composed of following:
  1. Source constructed from `IAsyncEnumerable<>`.
  2. Stream has a kill switch.
  3. Sink is an actor that sends back back-pressure signal (using acknowledge message).

When:
  1. Source is paused/interrupted because sink does not signal acknowledge message.
  2. Stream is shut down via kill switch.

Then:
  1. Enumerator provided by `IAsyncEnumerable<>` does not complete enumeration, leaving any resource allocated withing the enumerator not disposed. In my case this would be EF Core `DbContext` constructed via `IDbContextFactory<>`.

**To Reproduce**
```csharp
using Akka.Actor;
using Akka.Actor.Dsl;
using Akka.Streams;
using Akka.Streams.Dsl;
using System.Runtime.CompilerServices;

const bool interruptSinkActor = false;
using var system = ActorSystem.Create("Streams");

var sinkActorRef = system.ActorOf(act =>
{
    act.ReceiveAny((message, context) =>
    {
        switch (message)
        {
            case int value:
                Console.WriteLine(value);
                if (interruptSinkActor && value > 50)
                    Console.WriteLine("Interrupted. Press [Enter] to shutdown.");
                else
                    context.Sender.Tell(new Ack());
                break;

            case OnInit:
                context.Sender.Tell(new Ack());
                break;

            case Complete:
                break;
        }
    });
});

var stream = Source.From(() => Enumerate(default))
    .ViaMaterialized(KillSwitches.Single<int>(), Keep.Right)
    .ToMaterialized(
        Sink.ActorRefWithAck<int>(
            actorRef: sinkActorRef,
            onInitMessage: new OnInit(),
            ackMessage: new Ack(),
            onCompleteMessage: new Complete()),
        Keep.Left);

var killSwitch = stream.Run(system);

Console.ReadLine(); 
killSwitch.Shutdown();
await system.Terminate();
Console.WriteLine("System terminated.");

async static IAsyncEnumerable<int> Enumerate([EnumeratorCancellation] CancellationToken cancellationToken)
{
    await using var resource = new Resource(); 
    foreach (var i in Enumerable.Range(0, 100))
    {
        await Task.Delay(1, cancellationToken).ConfigureAwait(false);
        yield return i;
    }
}

public class Resource : IAsyncDisposable
{
    public ValueTask DisposeAsync()
    {
        Console.WriteLine("Enumerator completed and resource disposed");
        return ValueTask.CompletedTask;
    }
}

public record class OnInit();

public record class Ack();

public record class Complete();
```

_Links to working reproductions on Github / Gitlab are very much appreciated_

**Expected behavior**
Async enumerator should be either disposed, or maybe awaited next via `MoveNextAsync` with cancellation triggered, so that enumerator is correctly completed. Resources allocated within enumerator are thus correctly disposed.

**Actual behavior**
Async enumerator is left uncompleted. Resources allocated within enumerator are not disposed.

**Screenshots**
Interrupted enumerator:
image
Completed enumerator:
image

**Environment**
All environments are impacted.

**Additional context**
Disposal of the stream was removed previously by #6290, related to #6280.

to11mtm:
Fixes #6903

## Changes

Changed cleanup pattern to fix AsyncEnumerable stage, also fixed race condition in ValueTask state check.

## Checklist

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
@@ -269,6 +269,21 @@ namespace Akka.Streams.Tests.Dsl
             });
         }
 
+        [Fact]
+        public async Task AsyncEnumerableSource_Disposes_OnCancel()
+        {
+            var resource = new Resource();
+            var tcs = new System.Threading.Tasks.TaskCompletionSource<NotUsed>(TaskCreationOptions
+                .RunContinuationsAsynchronously);
+            var src = Source.From(() =>
+                CancelTestGenerator(tcs, resource, default));
+            src.To(Sink.Ignore<int>()).Run(Materializer);
+            await tcs.Task;
+            Materializer.Shutdown();
+            await Task.Delay(500);
+            Assert.False(resource.IsActive);
+        }
+
         private static async IAsyncEnumerable<int> RangeAsync(int start, int count, 
             [EnumeratorCancellation] CancellationToken token = default)
         {
@@ -306,6 +321,39 @@ namespace Akka.Streams.Tests.Dsl
                     yield break;
 
                 yield return i;
+            }
+        }
+
+        public static async IAsyncEnumerable<int> CancelTestGenerator(
+            TaskCompletionSource<NotUsed> tcs,
+            Resource resource,
+            [EnumeratorCancellation] CancellationToken token
+        )
+        {
+            await using var res = resource;
+            int i = 0;
+            bool isSet = false;
+            while (true)
+            {
+                await Task.Delay(1, token).ConfigureAwait(false);
+                yield return i++;
+                if (isSet == false)
+                {
+                    tcs.TrySetResult(NotUsed.Instance);
+                    isSet = true;
+                }
+            }
+            // ReSharper disable once IteratorNeverReturns
+        }
+
+        public class Resource : IAsyncDisposable
+        {
+            public bool IsActive = true;
+            public ValueTask DisposeAsync()
+            {
+                IsActive = false;
+                Console.WriteLine("Enumerator completed and resource disposed");
+                return new ValueTask();
             }
         }
     }
--- a/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
@@ -9,6 +9,7 @@ using System;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Akka.Actor;
 using Akka.Pattern;
 using Akka.Streams.Dsl;
 using Akka.Streams.TestKit;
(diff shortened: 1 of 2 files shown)


src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3812,18 +3812,19 @@ private sealed class Logic : OutGraphStageLogic
...
             public Logic(SourceShape<T> shape, IAsyncEnumerable<T> enumerable) : base(shape)
             {
+                
                 _enumerable = enumerable;
                 _outlet = shape.Outlet;
                 _onSuccess = GetAsyncCallback<T>(OnSuccess);
                 _onFailure = GetAsyncCallback<Exception>(OnFailure);
                 _onComplete = GetAsyncCallback(OnComplete);
-
+                _completionCts = new CancellationTokenSource();
```
to11mtm, line 3827: Controversial possibly, but I felt it safer to move this to ctor.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3859,26 +3880,12 @@ public override void OnPull()
                         // if result is false, it means enumerator was closed. Complete stage in that case.
                         CompleteStage();
                     }
-                } 
-                else if (vtask.IsCompleted) // IsCompleted covers Faulted, Cancelled, and RanToCompletion async state
```
to11mtm, line 3863: This is racy, so removed. Exception case is not as big a deal anyway.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3897,16 +3904,12 @@ async Task ProcessTask()
                         }
                     }
 
-#pragma warning disable CS4014
-                    ProcessTask();
-#pragma warning restore CS4014
+                    _ = ProcessTask();
```
to11mtm, line 3928: Discard = no pragma disable needed

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3897,16 +3904,12 @@ async Task ProcessTask()
...
                 }
             }
 
             public override void OnDownstreamFinish(Exception cause)
             {
-                _completionCts.Cancel();
-                _completionCts.Dispose();
```
to11mtm, line 3909: Moved to `PostStop()`. AFAIK we will call it after `CompleteStage()` in interpreter, plz correct if I'm wrong.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3838,9 +3839,29 @@ public Logic(SourceShape<T> shape, IAsyncEnumerable<T> enumerable) : base(shape)
...
-                _completionCts = new CancellationTokenSource();
                 _enumerator = _enumerable.GetAsyncEnumerator(_completionCts.Token);
             }
+            
+            public override void PostStop()
+            {
+                try
+                {
+                    _completionCts.Cancel();
+                    _completionCts.Dispose();
```
to11mtm, line 3850: We should ALWAYS be disposing the CTS.

What is more interesting to ask, is whether it's worth us having some sort of flag to know whether we should bother canceling or not. I went this route because in theory, under happy-completion path nobody cares about this CTS anyway. 

That said there -may- be value in some cases of not calling cancel, but can't think of any.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3838,9 +3839,29 @@ public Logic(SourceShape<T> shape, IAsyncEnumerable<T> enumerable) : base(shape)
...
+                }
+                catch
+                {
+                }
+                try
+                {
+                    _enumerator.DisposeAsync().ConfigureAwait(false);
```
to11mtm: Just feels like the right thing to do.

src/core/Akka.Streams.Tests/Dsl/AsyncEnumerableSpec.cs
```diff
@@ -269,6 +269,21 @@ async IAsyncEnumerable<int> GenerateInts()
...
+                .RunContinuationsAsynchronously);
+            var src = Source.From(() =>
+                CancelTestGenerator(tcs, resource, default));
+            src.To(Sink.Ignore<int>()).Run(Materializer);
+            await tcs.Task;
+            Materializer.Shutdown();
+            await Task.Delay(500);
+            Assert.False(resource.IsActive);
+        }
+
```
to11mtm, line 416: Adapted from #6903 report.
Calling `Materializer.Shutdown()` is, as far as I know, going to happen when the actorsystem shuts down. 

OTOH this test was added in a red->green refactoring, so I know without the changes in `AsyncEnumerable` it fails.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3838,9 +3839,33 @@ public Logic(SourceShape<T> shape, IAsyncEnumerable<T> enumerable) : base(shape)
...
+                    _completionCts.Dispose();
+                }
+                catch(Exception ex)
+                {
+                    // This should never happen
+                    Log.Error(ex, "AsyncEnumerable threw while cancelling CancellationTokenSource");
+                }
+                try
+                {
+                    var disposed = _enumerator.DisposeAsync().AsTask().Wait(TimeSpan.FromSeconds(10));
```
Arkatufus: Opted to wait for the dispose task to complete with a generous timeout value (10 seconds).

Arkatufus: Basically, since the ValueTask was never awaited, we would never catch any exception thrown, since it is ran detached.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
Aaronontheweb: Good call IMHO

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -3838,9 +3839,39 @@ public Logic(SourceShape<T> shape, IAsyncEnumerable<T> enumerable) : base(shape)
...
+                    _completionCts.Dispose();
+                }
+                catch(Exception ex)
+                {
+                    // This should never happen
+                    Log.Error(ex, "AsyncEnumerable threw while cancelling CancellationTokenSource");
+                }
+                try
+                {
+                    _enumerator.DisposeAsync().AsTask().ContinueWith(t =>
```
Aaronontheweb: Nitpick: why not just have a detatched `Task` running in a local function which `await`s on `DisposeAsync` ? You can place the entire `try...catch` inside of that and have cleaner error-handling semantics.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
Aaronontheweb: Handy!

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
Aaronontheweb: I believe that's the case.

Aaronontheweb (commented):
Overall LGTM, but had a suggestion around handling of `DisposeAsync`

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
Arkatufus: done

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
Arkatufus: I think that this is the actual fix, the enumerator needs to be disposed for the resources to be released.

Aaronontheweb: Yep, me too

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
Arkatufus: Reverted back to detached task

Aaronontheweb (approved):
LGTM

PR #6940, opened 2023-10-02
Sealed uninherited `internal` classes in `Akka.Streams` and `Akka.Remote`

Lehonti:
Also:
- In the files that were touched, fields that weren't being modified outside the constructors were made `readonly`
- In the classes that were `sealed`, new `protected` members were made `private`
- Exceptions were not `sealed` because they had a `protected` serialization constructors. It may well be the case that this constructor works well if it's made `private`, but I don't know the codebase well enough, so I will leave those alone for now


src/core/Akka.Remote/Endpoint.cs
Aaronontheweb, line 1: LGTM

src/core/Akka.Remote/EndpointManager.cs
Aaronontheweb, line 1: LGTM

src/core/Akka.Remote/PhiAccrualFailureDetector.cs
Aaronontheweb, line 1: LGTM

src/core/Akka.Remote/Remoting.cs
Aaronontheweb, line 1: LGTM

Aaronontheweb (approved):
LGTM

PR #6941, opened 2023-10-02
`struct`s in `Akka.Streams` with exclusively read-only members were made `readonly`

Lehonti:
## Changes

The title is self-explanatory, so I will just emphasize the fact that this change will allow spotting immutable `struct`s with just a glance


Aaronontheweb:
there's public API approvals missing on this PR

Aaronontheweb (approved):
LGTM

PR #6942, opened 2023-10-03
Fix broken illegal actor name unit test

Arkatufus:
## Changes

Illegal actor name exception string was changed, breaking the unit tests.


Aaronontheweb (approved):
LGTM

PR #6946, opened 2023-10-05
Fix PeriodicTimer scheduler deadlock during start

Arkatufus:
Scheduler PeriodicTimer async start can deadlock if disposed immediately

## Changes

* Change detached task execution to use `Task.Run()`


Aaronontheweb (approved):
LGTM

PR #6978, opened 2023-10-29
Akka.Persistence.Query: include more descriptive ToString() values for all Offset types #6927

### LINKED ISSUE #6927: Akka.Persistence.Query: include more descriptive `ToString()` values for all `Offset` types

I just had to write the following class in my own application that is making use of Akka.NET for EventSourcing and CQRS using Akka.Persistence.Query:

```csharp
public static long AsLong(this Offset offset)
{
    switch (offset)
    {
        case Sequence seq:
            return seq.Value;
        case NoOffset _:
            return 0;
        default:
            throw new NotSupportedException($"Offset type [{offset.GetType()}] is not supported.");
    }
}
```

This is because the default `ToString()` values for `Offset` aren't helpful - from one of my test logs:

```
PublisherCreated { PublisherId = ad07e579-69f5-4738-8b20-e6a0dd3a058d, CurrentStatus = Active, Timestamp = 9/20/2023 6:39:57 PM +00:00 }] - projection state updated to [Akka.Persistence.Query.Sequence]
```

Each `Offset` implementation should clearly log its value so users can plop that `Offset` directly into log statements without needing to filter through the underlying types. Only users with detailed knowledge about the internals of Akka.Persistence would know how to do this today.

It might be worth just exposing a `long` value type on each `Offset` anyway to help make it easier to serialize / persist too, but that gets into icky territory if someone wants to start using stuff like `TimeUuid`s for ordering records inside the Akka.Persistence event journal - hence why the `Offset` abstraction exists in the first place.

Cleaning up the `ToString()` method for logging purposes makes the most sense.

YariSPB:
Fixes #6927 

## Changes

Added mandatory override ToString() for Offset implementations, which displays Value. Added respective unit tests. Exposed new public API ToString().

## Checklist

### Latest `dev` Benchmarks 

### This PR's Benchmarks

### DIFF
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistenceQuery.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistenceQuery.DotNet.verified.txt
@@ -64,6 +64,7 @@ namespace Akka.Persistence.Query
     {
         public static Akka.Persistence.Query.NoOffset Instance { get; }
         public override int CompareTo(Akka.Persistence.Query.Offset other) { }
+        public override string ToString() { }
     }
     public abstract class Offset : System.IComparable<Akka.Persistence.Query.Offset>
     {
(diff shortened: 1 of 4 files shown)


Aaronontheweb (approved):
LGTM - the failing tests are racy (I need to look into that) and are unrelated to your changes.

PR #7007, opened 2023-11-22
hide task scheduler

Zetanova:
Fixes the accidentally leaking of the task scheduler inside from the actor cell,
by forcing all created async-tasks inside an actor to use the `TaskScheduler.Default`

## Changes

The ´ActorCell´ will use the custom TaskScheduler, but all created async Tasks will use `TaskScheduler.Default`

//before
```csharp
//inside actor message loop
Run(async () => UpdateAsync(_cts.Token)).PipeTo(Self, null);

async UpdateAsync(CancellationToken cancellationToken) {
    //this code will always run with `TaskScheduler.Default`
}
```

//after
```csharp
//inside actor message loop
UpdateAsync(_cts.Token).PipeTo(Self, null);

async UpdateAsync(CancellationToken cancellationToken) {
    //this code will always run with `TaskScheduler.Default`
}
```

## Checklist

### DIFF
--- a/src/core/Akka/Dispatch/Dispatchers.cs
+++ b/src/core/Akka/Dispatch/Dispatchers.cs
@@ -185,10 +185,9 @@ namespace Akka.Dispatch
     /// </summary>
     internal sealed class TaskSchedulerExecutor : ExecutorService
     {
-        /// <summary>
-        ///     The scheduler
-        /// </summary>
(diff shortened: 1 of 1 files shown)


Zetanova:
on hold, maybe the `TaskContinuationOptions.HideScheduler` continuation can have an effect on the nested async task 
like inside EF core.

PR #7008, opened 2023-11-22
migrate build and test system to .NET 8

Aaronontheweb:
## Changes

migrate build and test system to .NET 8 - has no impact on runtime targets.

Added legacy support for .NET 6 to important benchmarks for comparison purposes.


src/Directory.Build.props
```diff
@@ -26,7 +26,7 @@
     <NBenchVersion>2.0.1</NBenchVersion>
     <ProtobufVersion>3.25.0</ProtobufVersion>
     <BenchmarkDotNetVersion>0.13.10</BenchmarkDotNetVersion>
-    <NetTestVersion>net7.0</NetTestVersion>
+    <NetTestVersion>net8.0</NetTestVersion>
```
Arkatufus, line 29: You can probably change this to `<NetTestVersion>net8.0;net6.0</NetTestVersion>`

Arkatufus (changes requested):

src/Directory.Build.props
Aaronontheweb: I don't necessarily want to multi-target for tests, just a handful of benchmarks

Arkatufus (approved):

PR #7010, opened 2023-11-29
Akka.Serialization: `INoSerializationVerificationNeeded` does not handle `IWrappedMessage` correctly

Aaronontheweb:
## Changes

`INoSerializationVerificationNeeded` does not handle `IWrappedMessage` correctly. We handle `DeadLetter` correctly as a special case, but not any of the other `IWrappedMessage` implementations. We should handle this generically.

## Checklist

### DIFF
--- a/src/core/Akka.Tests/Serialization/SerializeAllMessagesSpec.cs
+++ b/src/core/Akka.Tests/Serialization/SerializeAllMessagesSpec.cs
@@ -1,0 +1,76 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="SerializeAllMessagesSpec.cs" company="Akka.NET Project">
+//     Copyright (C) 2009-2023 Lightbend Inc. <http://www.lightbend.com>
+//     Copyright (C) 2013-2023 .NET Foundation <https://github.com/akkadotnet/akka.net>
+// </copyright>
+//-----------------------------------------------------------------------
+
+using System;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Event;
+using Akka.TestKit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Tests.Serialization;
+
+public class SerializeAllMessagesSpec : AkkaSpec
+{
+    public SerializeAllMessagesSpec(ITestOutputHelper output) : base("akka.actor.serialize-messages = on", output)
+    {
+    }
+    
+    private class MyMessage : INoSerializationVerificationNeeded
+    {
+        public MyMessage(Action myAction)
+        {
+            MyAction = myAction;
+        }
+
+        // add an unserializable member, such as a delegate
+        public Action MyAction { get; }
+    }
+
+    private class MyWrappedMessage : IWrappedMessage
+    {
+        public MyWrappedMessage(object message)
+        {
+            Message = message;
+        }
+
+        public object Message { get; }
+    }
+    
+    // create an actor that will process a MyWrappedMessage and invoke the delegate
+    private class MyActor : ReceiveActor
+    {
+        public MyActor()
+        {
+            Receive<MyWrappedMessage>(msg =>
+            {
+                var myMessage = (MyMessage) msg.Message;
+                myMessage.MyAction();
+            });
+        }
+    }
+    
+    [Fact]
+    public async Task Should_not_serialize_WrappedMessage_with_INoSerializationVerificationNeeded()
+    {
+        // Arrange
+        var myProbe = CreateTestProbe();
+        var action = () => { myProbe.Tell("worked"); };
+        var message = new MyMessage(action);
+        var wrappedMessage = new MyWrappedMessage(message);
+        
+        var myActor = Sys.ActorOf(Props.Create(() => new MyActor()), "wrapped-message-actor");
+
+        await EventFilter.Error().ExpectAsync(0, async () =>
+        {
+            // Act
+            myActor.Tell(wrappedMessage);
+            await myProbe.ExpectMsgAsync("worked");
+        });
+    }
+}
--- a/src/core/Akka/Actor/ActorCell.cs
+++ b/src/core/Akka/Actor/ActorCell.cs
@@ -521,7 +521,7 @@ namespace Akka.Actor
             var unwrapped = envelope.Message;
             
             if(envelope.Message is IWrappedMessage wrapped)
-                unwrapped = wrapped.Message;
+                unwrapped = WrappedMessage.Unwrap(wrapped); // recursively unwraps message
 
             // don't do serialization verification if the underlying type doesn't require it
             if (unwrapped is INoSerializationVerificationNeeded)
--- a/src/core/Akka/Actor/ActorCell.cs
+++ b/src/core/Akka/Actor/ActorCell.cs
@@ -518,9 +518,12 @@ namespace Akka.Actor
 
         private Envelope SerializeAndDeserialize(Envelope envelope)
         {
-            DeadLetter deadLetter;
-            var unwrapped = (deadLetter = envelope.Message as DeadLetter) != null ? deadLetter.Message : envelope.Message;
+            var unwrapped = envelope.Message;
+            
+            if(envelope.Message is IWrappedMessage wrapped)
+                unwrapped = wrapped.Message;
 
+            // don't do serialization verification if the underlying type doesn't require it
             if (unwrapped is INoSerializationVerificationNeeded)
                 return envelope;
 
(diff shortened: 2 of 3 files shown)


Aaronontheweb:
Added a repro for this; working on a fix now.

src/core/Akka.Tests/Serialization/SerializeAllMessagesSpec.cs
```diff
@@ -0,0 +1,76 @@
...
+    
+    private class MyMessage : INoSerializationVerificationNeeded
+    {
+        public MyMessage(Action myAction)
+        {
+            MyAction = myAction;
+        }
+
+        // add an unserializable member, such as a delegate
+        public Action MyAction { get; }
```
Aaronontheweb, line 32: Newtonsoft.Json and Hyperion will both barf on delegate serialization, so this message will not be received when the parent `IWrappedMessage` is not handled correctly.

src/core/Akka.Tests/Serialization/SerializeAllMessagesSpec.cs
```diff
@@ -0,0 +1,76 @@
...
+using Akka.Event;
+using Akka.TestKit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Tests.Serialization;
+
+public class SerializeAllMessagesSpec : AkkaSpec
+{
+    public SerializeAllMessagesSpec(ITestOutputHelper output) : base("akka.actor.serialize-messages = on", output)
```
Aaronontheweb, line 20: Have to enable `akka.actor.serialize-messages=on` in order for this to even be an issue.

src/core/Akka.Tests/Serialization/SerializeAllMessagesSpec.cs
```diff
@@ -0,0 +1,76 @@
...
+            });
+        }
+    }
+    
+    [Fact]
+    public async Task Should_not_serialize_WrappedMessage_with_INoSerializationVerificationNeeded()
+    {
+        // Arrange
+        var myProbe = CreateTestProbe();
+        var action = () => { myProbe.Tell("worked"); };
```
Aaronontheweb, line 63: Invoking the delegate closes on the loop on this spec by messaging the closed-over `TestProbe`.

src/core/Akka.Tests/Serialization/SerializeAllMessagesSpec.cs
```diff
@@ -0,0 +1,76 @@
...
+        var message = new MyMessage(action);
+        var wrappedMessage = new MyWrappedMessage(message);
+        
+        var myActor = Sys.ActorOf(Props.Create(() => new MyActor()), "wrapped-message-actor");
+
+        await EventFilter.Error().ExpectAsync(0, async () =>
+        {
+            // Act
+            myActor.Tell(wrappedMessage);
+            await myProbe.ExpectMsgAsync("worked");
```
Aaronontheweb, line 73: Both the `EventFilter` will fail and the `ExpectMsgAsync` will fail if the `ActorCell` of the recipient attempts to serialize this message - which is exactly what happened before I applied my fix.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -518,9 +518,12 @@ public static IActorRef GetCurrentSenderOrNoSender()
 
         private Envelope SerializeAndDeserialize(Envelope envelope)
         {
-            DeadLetter deadLetter;
-            var unwrapped = (deadLetter = envelope.Message as DeadLetter) != null ? deadLetter.Message : envelope.Message;
+            var unwrapped = envelope.Message;
+            
+            if(envelope.Message is IWrappedMessage wrapped)
```
Aaronontheweb: for all `IWrappedMessage`s: if the underlying message doesn't want to be serialized (typically because it can't be), don't try it. Just return the message without performing any additional work.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -534,7 +537,8 @@ private Envelope SerializeAndDeserialize(Envelope envelope)
                 throw new SerializationException($"Failed to serialize and deserialize payload object [{unwrapped.GetType()}]. Envelope: [{envelope}], Actor type: [{Actor.GetType()}]", e);
             }
 
-            if (deadLetter != null)
+            // special case handling for DeadLetters
+            if (envelope.Message is DeadLetter deadLetter)
```
Aaronontheweb: Restore the special-case handling for `DeadLetter`s that we had before.

Aaronontheweb (commented):
Detailed my changes.

src/core/Akka/Actor/ActorCell.cs
Arkatufus: It seems like we're missing a round of unwrap here, this code will return an `Envelope` that wraps a `DeadLetter` that wraps a `DeadLetter`.
The `DeadLetter` message still needs to be deserialized like it was before and then re-wrapped inside a new `DeadLetter` message.

Arkatufus (changes requested):

src/core/Akka/Actor/ActorCell.cs
Aaronontheweb: Good point, I can use the `Unwrapped` utility from earlier to recursively unwrap it - let me look into that. Currently doing a full test suite run locally so I can fix some of the outstanding .NET 8 errors in the suite.

Arkatufus (approved):

src/core/Akka/Actor/BuiltInActors.cs
```diff
@@ -191,8 +191,10 @@ public static class WrappedMessage
     {
         public static object Unwrap(object message)
         {
-            if (message is IWrappedMessage wm)
-                return Unwrap(wm.Message);
+            while (message is IWrappedMessage wm)
+            {
+                message = wm.Message;
+            }
```
Arkatufus, line 197: Unwrap tail recursion into loop

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -534,11 +538,12 @@ private Envelope SerializeAndDeserialize(Envelope envelope)
...
             }
 
-            if (deadLetter != null)
-                return new Envelope(new DeadLetter(deserializedMsg, deadLetter.Sender, deadLetter.Recipient), envelope.Sender);
-            return new Envelope(deserializedMsg, envelope.Sender);
-
+            // special case handling for DeadLetters
+            return deadLetter is not null 
+                ? new Envelope(new DeadLetter(deserializedMsg, deadLetter.Sender, deadLetter.Recipient), envelope.Sender) 
+                : new Envelope(deserializedMsg, envelope.Sender);
```
Arkatufus: This code behaves exactly like the old code, only modernized

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -516,11 +516,15 @@ public static IActorRef GetCurrentSenderOrNoSender()
...
+        #nullable enable
         private Envelope SerializeAndDeserialize(Envelope envelope)
         {
-            DeadLetter deadLetter;
-            var unwrapped = (deadLetter = envelope.Message as DeadLetter) != null ? deadLetter.Message : envelope.Message;
-
+            var deadLetter = envelope.Message as DeadLetter;
+            
+            // recursively unwraps message
+            var unwrapped = WrappedMessage.Unwrap(deadLetter is not null ? deadLetter.Message : envelope.Message);
```
Arkatufus: If the envelope message is a `DeadLetter`, grab the dead letter message instead.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -516,11 +516,15 @@ public static IActorRef GetCurrentSenderOrNoSender()
             return current != null ? current.Sender : ActorRefs.NoSender;
         }
 
+        #nullable enable
         private Envelope SerializeAndDeserialize(Envelope envelope)
         {
-            DeadLetter deadLetter;
-            var unwrapped = (deadLetter = envelope.Message as DeadLetter) != null ? deadLetter.Message : envelope.Message;
-
+            var deadLetter = envelope.Message as DeadLetter;
```
Arkatufus: Check to see if the envelope message is a `DeadLetter`

src/core/Akka/Actor/ActorCell.cs
Arkatufus: Also unwraps the message recursively if it is an `IWrappedMessage`. There is no need to check if the message is `IWrappedMessage` or not, it is done inside the `Unwrap` static function.

Arkatufus (commented):
Self review

src/core/Akka/Actor/BuiltInActors.cs
Aaronontheweb: LGTM

src/core/Akka/Actor/ActorCell.cs
Aaronontheweb: Don't need to check for a `DeadLetter` here as it implements `IWrappedMessage`.

src/core/Akka/Actor/ActorCell.cs
Aaronontheweb: Looks fine to me

Aaronontheweb (commented):
Just one nitpick on the duplicate check for a `DeadLetter` @Arkatufus

PR #7026, opened 2023-12-13
Cleanup warnings

Aaronontheweb:
Just a bit of housekeeping around the code base


Aaronontheweb:
Gradually want to get `TreatWarningsAsErrors` turned on, but have a bit of technical debt that needs paying off between here and there

Aaronontheweb:
Normal racy unit test failure here - looks like I cleaned everything else up.

PR #7032, opened 2024-01-03
Make ClusterClient messages be serialized using ClusterClientMessageSerializer

### LINKED ISSUE #6803: Akka.Tools.ClusterClient: `ClusterClient.Send` and `ClusterClient.SendToAll` aren't serialized using `ClusterClientSerializer`
**Version Information**
Version of Akka.NET? 1.5.8 and previous
Which Akka.NET Modules?

**Describe the bug**

`ClusterClient.Send` and `ClusterClient.SendToAll` aren't serialized using `ClusterClientSerializer` - they are instead serialized using the `System.Object` serializer. This means that if a user defines a custom serializer for messages they wish to send over `ClusterClient`, none of the messages sent will use that serializer.

**Reproduction**

Will submit one via PR shortly.

**Expected behavior**

User-defined messages should be serialized using their own custom serializer, when specified, not overridden by a user-defined serializer

**Actual behavior**

Newtonsoft.Json serializer is being used instead.

**Additional context**

Since this involves making a change to the wire format (one of the [no-no's outlined in our developer guidelines](https://getakka.net/community/contributing/wire-compatibility.html)), we're going to need to add a feature flag to make this backwards compatible and enable that by default:

```hocon
akka.cluster.client.use-legacy-serialization = on
```

Users who want to opt-in to the new behavior will be able to do-so by turning off that setting.

Arkatufus:
Fixes #6803

## Changes

* Add ClusterClient specific message types to the `ClusterClientMessageSerializer` serializer
* Add new type mapping for ClusterClient specific message types to the serializer HOCON settings
* Add a backward compatibility flag to turn this new feature on and off (defaults to old behaviour)
* Add spec to make sure that new behavior is behaving correctly.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/ClusterClient/ClusterClientSerializerSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/ClusterClient/ClusterClientSerializerSpec.cs
@@ -1,0 +1,110 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="ClusterClientSerializerSpec.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2024 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2024 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Cluster.Tools.Client;
+using Akka.Cluster.Tools.Client.Serialization;
+using Akka.Configuration;
+using Akka.Serialization;
+using Akka.TestKit;
+using Xunit;
+using Xunit.Abstractions;
+using FluentAssertions;
+
+namespace Akka.Cluster.Tools.Tests.ClusterClient;
+
+public class ClusterClientSerializerLegacySpec : ClusterClientSerializerSpecBase
+{
+    public ClusterClientSerializerLegacySpec(ITestOutputHelper output) : base(output, true)
+    {
+    }
+}
+
+public class ClusterClientSerializerSpec : ClusterClientSerializerSpecBase
+{
+    public ClusterClientSerializerSpec(ITestOutputHelper output) : base(output, false)
+    {
+    }
+}
+
+public abstract class ClusterClientSerializerSpecBase : AkkaSpec
+{
+    private static Config Config(bool useLegacy)
+        => ConfigurationFactory.ParseString($"""
+akka.extensions = ["Akka.Cluster.Tools.Client.ClusterClientReceptionistExtensionProvider, Akka.Cluster.Tools"]
+akka.log-serializer-override-on-start = on
+akka.actor.provider = cluster
+akka.remote.dot-netty.tcp.port = 0
+akka.cluster.client.use-legacy-serialization = {(useLegacy ? "on" : "off")}
+""")
+            .WithFallback(ClusterClientReceptionist.DefaultConfig());
+
+    private readonly bool _useLegacy;
+    private readonly ActorPath _actorPath;
+    
+    protected ClusterClientSerializerSpecBase(ITestOutputHelper output, bool useLegacy) : base(Config(useLegacy), output)
+    {
+        _useLegacy = useLegacy;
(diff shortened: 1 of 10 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests/ClusterClient/ClusterClientSerializerSpec.cs
Aaronontheweb, line 1: LGTM

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClient.cs
Aaronontheweb, line 1: LGTM

src/contrib/cluster/Akka.Cluster.Tools/Client/Serialization/ClusterClientMessageSerializer.cs
```diff
@@ -99,13 +123,24 @@ public override object FromBinary(byte[] bytes, string manifest)
...
         public override string Manifest(object o)
         {
-            if (o is ClusterReceptionist.Contacts) return ContactsManifest;
-            if (o is ClusterReceptionist.GetContacts) return GetContactsManifest;
-            if (o is ClusterReceptionist.Heartbeat) return HeartbeatManifest;
-            if (o is ClusterReceptionist.HeartbeatRsp) return HeartbeatRspManifest;
-            if (o is ClusterReceptionist.ReceptionistShutdown) return ReceptionistShutdownManifest;
-
-            throw new ArgumentException($"Can't serialize object of type [{o.GetType()}] in [{nameof(ClusterClientMessageSerializer)}]");
+            return o switch
```
Aaronontheweb, line 126: LGTM

src/contrib/cluster/Akka.Cluster.Tools/Client/reference.conf
```diff
@@ -90,6 +90,10 @@ akka.cluster.client {
   # If this is not wanted it can be set to "off" to disable the timeout and retry
   # forever.
   reconnect-timeout = off
+  
+  # Turning this setting to off or false will cause the ClusterClient messages
+  # to be serialized using the ClusterClientMessageSerializer and not the default Object serializer. 
+  use-legacy-serialization = on
```
Aaronontheweb, line 96: LGTM

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -393,6 +393,41 @@ public void AddSerializationMap(Type type, Serializer serializer)
...
+                LogWarning($"Failed to remove serializer for type [{type}]");
+            }
+            
+            foreach (var serializerType in _serializerMap.Keys)
+            {
+                if (serializerType.IsAssignableFrom(type) && serializerType != _objectType)
+                {
+                    if (!_serializerMap.TryRemove(serializerType, out _))
+                    {
+
```
Aaronontheweb: Is something supposed to happen here if this can't be used?

src/core/Akka/Serialization/Serialization.cs
```diff
@@ -393,6 +393,41 @@ public void AddSerializationMap(Type type, Serializer serializer)
...
         }
 
+        /// <summary>
+        /// Remove the serializer mapping for <see cref="Type"/> <paramref name="type"/> and all other types
+        /// that were derived from it.
+        /// </summary>
+        /// <param name="type">The <see cref="Type"/> that will be removed from the serializer mapping</param>
+        /// <exception cref="InvalidOperationException">>Thrown when <paramref name="type"/> is of type <see cref="Object"/></exception>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void RemoveSerializationMap(Type type)
```
Aaronontheweb: Can we make this `internal` instead of `public` so this can't be abused by end-users?

Aaronontheweb (changes requested):
Mostly looks fine, but left one suggestion

src/core/Akka/Serialization/Serialization.cs
Arkatufus: Forgot to log a warning there

src/core/Akka/Serialization/Serialization.cs
Arkatufus: Good point

Aaronontheweb (approved):
LGTM

PR #7037, opened 2024-01-04
Akka.TestKit: remove `ConfigureAwait(false)` from internal synchronous TestKit methods and fix `Watch` / `Unwatch` bugs

### LINKED ISSUE #7038: Akka.TestKit: `ImplicitSender` (`TestActor`) is lost inside `async` tests whenever synchronous methods are called
**Version Information**
Version of Akka.NET? v1.5.14
Which Akka.NET Modules? Akka.TestKit

**Describe the bug**

When we originally introduced the async TestKit back in v1.5, we had to add a layer of backwards compatibility for all of the original synchronous methods in the test suite that are still heavily used through not just Akka.NET's test suite, but also many of our users.

This is what the `ExpectMsg<T>` implementation looks like in the current `dev` branch at the time of writing this:

https://github.com/akkadotnet/akka.net/blob/5620f02459d6b06fa89ac148a1317f4e67d8fdd6/src/core/Akka.TestKit/TestKitBase_Expect.cs#L38-L45

That method works fine and as-intended inside any un-updated synchronous test - but problems occur when we start blending `async Task` tests with older synchronous methods, such as this rather innocuous test I observed fail in an earlier pull request today:

https://github.com/akkadotnet/akka.net/blob/5620f02459d6b06fa89ac148a1317f4e67d8fdd6/src/core/Akka.Docs.Tests/Actors/SchedulerSpecs.cs#L151-L165

This test won't _consistently_ fail, but it does occasionally with the following error:

```
Failed: Timeout 00:00:03 while waiting for a message of type System.Int32
``` 

How could that happen?

As it turns out, the problem is the `ExpectMsg<T>` call implementation, specifically the `ConfigureAwait(false)`. This can cause, _not always but sometimes_, the `ImplicitSender` synchronization context to get lost during the `IActorRef.Tell` - specifically it's the _second_ `IActorRef.Tell` where the implicit sender is lost, because we've completed our `await` and are now scheduled on a different synchronization context than previously.

The reason why this is flaky: TaskScheduler non-determinism: sometimes it might just inline the `Task` execution, but other times it doesn't? I honestly don't know.

**To Reproduce**

I modified one of the TestKit.Tests to make this consistently fail:

```csharp
[Fact]
public async Task ImplicitSender_should_have_testActor_as_sender()
{
    var echoActor = Sys.ActorOf(c => c.ReceiveAny((_, ctx) => ctx.Sender.Tell(ctx.Sender)));


    // Churn the context a bit to make sure it's not just a fluke
    var attempts = 10;
    try
    {
        while (attempts-- > 0)
        {
            await Task.Delay(1).ConfigureAwait(false);
            echoActor.Tell("message");
            ExpectMsg<IActorRef>(actorRef => Equals(actorRef, TestActor));
            await Task.Delay(1);
        }
    }
    catch(Exception ex)
    {
        Assert.Fail($"Failed on attempt {attempts} with {ex}");
    }
}
```

The test, if written this way:

```csharp
[Fact]
public async Task ImplicitSender_should_have_testActor_as_sender()
{
    var echoActor = Sys.ActorOf(c => c.ReceiveAny((_, ctx) => ctx.Sender.Tell(ctx.Sender)));


    // Churn the context a bit to make sure it's not just a fluke
    var attempts = 10;
    try
    {
        while (attempts-- > 0)
        {
            await Task.Delay(1); // no ConfigureAwait(false) here
            echoActor.Tell("message");
            ExpectMsg<IActorRef>(actorRef => Equals(actorRef, TestActor));
            await Task.Delay(1);
        }
    }
    catch(Exception ex)
    {
        Assert.Fail($"Failed on attempt {attempts} with {ex}");
    }
}
```

May not consistently fail - it very much depends on some of the `async` contexts and scopes involved, plus how the `TaskScheduler` does its thing. 

**Expected behavior**

The simple test case from before should consistently pass.

**Actual behavior**

Occasionally fails - we believe this is due to the internal use of `ConfigureAwait(false)` inside the synchronous testkit's adapter layer.

### LINKED ISSUE #7064: Akka.TestKit: `Watch` and `Unwatch` are secretly asynchronous
Nasty, decade old bug in the TestKit - `Watch` and `Unwatch` have always been secretly asynchronous, which is why tests like this one failed:

```csharp
[Fact]
        public async Task Routers_in_general_must_evict_terminated_routees()
        {
            var router = Sys.ActorOf(new RoundRobinPool(2).Props(Props.Create<Echo>()));
            router.Tell("");
            router.Tell("");

            var c1 = await ExpectMsgAsync<IActorRef>();
            var c2 = await ExpectMsgAsync<IActorRef>();

            Watch(router);
            Watch(c2);
            Sys.Stop(c2);
            (await ExpectTerminatedAsync(c2)).ExistenceConfirmed.Should().BeTrue();

            // it might take a while until the Router has actually processed the Terminated message
            await AwaitConditionAsync(async () =>
            {
                router.Tell("");
                router.Tell("");

                var res = await ReceiveWhileAsync(100.Milliseconds(), x =>
                {
                    if (x is IActorRef actorRef)
                        return actorRef;

                    return null;
                }, msgs: 2).ToListAsync();

                return res.Count == 2 && res.All(c => c.Equals(c1));
            });

            Sys.Stop(c1);
            (await ExpectTerminatedAsync(router)).ExistenceConfirmed.Should().BeTrue();
        }
```

Failed on `(await ExpectTerminatedAsync(c2)).ExistenceConfirmed.Should().BeTrue();`, which can only mean the actor was dead before the `TestActor` had a chance to go through.

_Originally posted by @Aaronontheweb in https://github.com/akkadotnet/akka.net/pull/7037#discussion_r1454192313_

Aaronontheweb:
## Changes

close https://github.com/akkadotnet/akka.net/issues/7038 
close #7064 

Removed all internal usages of `ConfigureAwait(false)` from synchronous Akka.TestKit method implementations - this should help reduce the incidences of mysterious racy unit tests.

### DIFF
--- a/src/core/Akka.TestKit/Internal/InternalTestActor.cs
+++ b/src/core/Akka.TestKit/Internal/InternalTestActor.cs
@@ -16,7 +16,7 @@ namespace Akka.TestKit.Internal
     /// An actor that enqueues received messages to a <see cref="BlockingCollection{T}"/>.
     /// <remarks>Note! Part of internal API. Breaking changes may occur without notice. Use at own risk.</remarks>
     /// </summary>
-    public class InternalTestActor : ActorBase
+    public class InternalTestActor : UntypedActor
     {
         private readonly ITestActorQueue<MessageEnvelope> _queue;
         private TestKit.TestActor.Ignore _ignore;
@@ -37,7 +37,7 @@ namespace Akka.TestKit.Internal
         /// </summary>
         /// <param name="message">TBD</param>
         /// <returns>TBD</returns>
-        protected override bool Receive(object message)
+        protected override void OnReceive(object message)
         {
             try
             {
@@ -56,16 +56,26 @@ namespace Akka.TestKit.Internal
             {
                 case TestActor.SetIgnore setIgnore:
                     _ignore = setIgnore.Ignore;
-                    return true;
+                    return;
                 case TestActor.Watch watch:
                     Context.Watch(watch.Actor);
-                    return true;
+                    RunTask(async () =>
+                    {
+                        await watch.Actor.Ask<ActorIdentity>(new Identify("watch"));
+                        watch.WatchCompleted.TrySetResult(true);
+                    });
+                    return;
                 case TestActor.Unwatch unwatch:
                     Context.Unwatch(unwatch.Actor);
-                    return true;
+                    RunTask(async () =>
+                    {
+                        await unwatch.Actor.Ask<ActorIdentity>(new Identify("unwatch"));
+                        unwatch.UnwatchCompleted.TrySetResult(true);
+                    });
+                    return;
                 case TestActor.SetAutoPilot setAutoPilot:
                     _autoPilot = setAutoPilot.AutoPilot;
-                    return true;
+                    return;
                 case TestActor.Spawn spawn:
                 {
                     var actor = spawn.Apply(Context);
@@ -74,7 +84,7 @@ namespace Akka.TestKit.Internal
                         _supervisorStrategy.Update(actor, spawn._supervisorStrategy.Value);
                     }
                     _queue.Enqueue(new RealMessageEnvelope(actor, Self));
-                    return true;
+                    return;
                 }
             }
 
@@ -82,12 +92,11 @@ namespace Akka.TestKit.Internal
             if(_autoPilot != null)
             {
                 var newAutoPilot = _autoPilot.Run(actorRef, message);
-                if(!(newAutoPilot is KeepRunning))
+                if(newAutoPilot is not KeepRunning)
                     _autoPilot = newAutoPilot;
             }
             if(_ignore == null || !_ignore(message))
                 _queue.Enqueue(new RealMessageEnvelope(message, actorRef));
-            return true;
         }
     }
 }
--- a/src/core/Akka.TestKit/Internal/InternalTestActor.cs
+++ b/src/core/Akka.TestKit/Internal/InternalTestActor.cs
@@ -59,19 +59,11 @@ namespace Akka.TestKit.Internal
                     return;
                 case TestActor.Watch watch:
                     Context.Watch(watch.Actor);
-                    RunTask(async () =>
-                    {
-                        await watch.Actor.Ask<ActorIdentity>(new Identify("watch"));
-                        watch.WatchCompleted.TrySetResult(true);
-                    });
+                    watch.WatchCompleted.TrySetResult(true);
                     return;
                 case TestActor.Unwatch unwatch:
                     Context.Unwatch(unwatch.Actor);
-                    RunTask(async () =>
-                    {
-                        await unwatch.Actor.Ask<ActorIdentity>(new Identify("unwatch"));
-                        unwatch.UnwatchCompleted.TrySetResult(true);
-                    });
+                    unwatch.UnwatchCompleted.TrySetResult(true);
                     return;
                 case TestActor.SetAutoPilot setAutoPilot:
                     _autoPilot = setAutoPilot.AutoPilot;
--- a/src/core/Akka.TestKit/Internal/InternalTestActor.cs
+++ b/src/core/Akka.TestKit/Internal/InternalTestActor.cs
@@ -59,11 +59,11 @@ namespace Akka.TestKit.Internal
                     return;
                 case TestActor.Watch watch:
                     Context.Watch(watch.Actor);
-                    watch.WatchCompleted.TrySetResult(true);
+                    Sender.Tell(TestActor.WatchAck.Instance);
                     return;
                 case TestActor.Unwatch unwatch:
                     Context.Unwatch(unwatch.Actor);
-                    unwatch.UnwatchCompleted.TrySetResult(true);
+                    Sender.Tell(TestActor.UnwatchAck.Instance);
                     return;
                 case TestActor.SetAutoPilot setAutoPilot:
                     _autoPilot = setAutoPilot.AutoPilot;
--- a/src/core/Akka.TestKit/TestActor.cs
+++ b/src/core/Akka.TestKit/TestActor.cs
@@ -65,15 +65,19 @@ namespace Akka.TestKit
         /// </summary>
         public class Watch : INoSerializationVerificationNeeded
         {
-            public Watch(IActorRef actorToWatch, TaskCompletionSource<bool> watchCompleted)
+            public Watch(IActorRef actorToWatch)
             {
                 Actor = actorToWatch;
-                WatchCompleted = watchCompleted;
             }
             
             public IActorRef Actor { get; }
-
-            public TaskCompletionSource<bool> WatchCompleted { get; }
+        }
+        
+        internal sealed class WatchAck : INoSerializationVerificationNeeded
+        {
+            private WatchAck() { }
+            
+            public static readonly WatchAck Instance = new();
         }
 
         /// <summary>
@@ -82,15 +86,22 @@ namespace Akka.TestKit
         /// </summary>
         public class Unwatch : INoSerializationVerificationNeeded
         {
-            public Unwatch(IActorRef actorToUnwatch, TaskCompletionSource<bool> unwatchCompleted)
+            public Unwatch(IActorRef actorToUnwatch)
             {
                 Actor = actorToUnwatch;
-                UnwatchCompleted = unwatchCompleted;
             }
             
             public IActorRef Actor { get; }
+        }
+
+        /// <summary>
+        /// INTERNAL API
+        /// </summary>
+        internal sealed class UnwatchAck : INoSerializationVerificationNeeded
+        {
+            private UnwatchAck() { }
             
-            public TaskCompletionSource<bool> UnwatchCompleted { get; }
+            public static readonly UnwatchAck Instance = new();
         }
 
         /// <summary>
--- a/src/core/Akka.TestKit/TestActor.cs
+++ b/src/core/Akka.TestKit/TestActor.cs
@@ -5,6 +5,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Util;
 
@@ -64,18 +65,15 @@ namespace Akka.TestKit
         /// </summary>
         public class Watch : INoSerializationVerificationNeeded
         {
-            private readonly IActorRef _actorToWatch;
+            public Watch(IActorRef actorToWatch, TaskCompletionSource<bool> watchCompleted)
+            {
+                Actor = actorToWatch;
+                WatchCompleted = watchCompleted;
+            }
+            
+            public IActorRef Actor { get; }
 
-            /// <summary>
-            /// TBD
-            /// </summary>
-            /// <param name="actorToWatch">TBD</param>
-            public Watch(IActorRef actorToWatch) { _actorToWatch = actorToWatch; }
-
-            /// <summary>
-            /// TBD
-            /// </summary>
-            public IActorRef Actor { get { return _actorToWatch; } }
+            public TaskCompletionSource<bool> WatchCompleted { get; }
         }
 
         /// <summary>
@@ -84,18 +82,15 @@ namespace Akka.TestKit
         /// </summary>
         public class Unwatch : INoSerializationVerificationNeeded
         {
-            private readonly IActorRef _actorToUnwatch;
-
-            /// <summary>
-            /// TBD
-            /// </summary>
-            /// <param name="actorToUnwatch">TBD</param>
-            public Unwatch(IActorRef actorToUnwatch) { _actorToUnwatch = actorToUnwatch; }
-
-            /// <summary>
-            /// TBD
-            /// </summary>
-            public IActorRef Actor { get { return _actorToUnwatch; } }
+            public Unwatch(IActorRef actorToUnwatch, TaskCompletionSource<bool> unwatchCompleted)
+            {
+                Actor = actorToUnwatch;
+                UnwatchCompleted = unwatchCompleted;
+            }
+            
+            public IActorRef Actor { get; }
+            
+            public TaskCompletionSource<bool> UnwatchCompleted { get; }
         }
 
         /// <summary>
--- a/src/core/Akka.TestKit/TestKitBase.cs
+++ b/src/core/Akka.TestKit/TestKitBase.cs
@@ -357,7 +357,18 @@ namespace Akka.TestKit
         /// <returns>The actor to watch, i.e. the parameter <paramref name="actorToWatch"/></returns>
         public IActorRef Watch(IActorRef actorToWatch)
         {
-            _testState.TestActor.Tell(new TestActor.Watch(actorToWatch));
+             var taskCompletionSource = new TaskCompletionSource<bool>();
+            _testState.TestActor.Tell(new TestActor.Watch(actorToWatch, taskCompletionSource));
+            /*
+             * Look, I know what you're thinking: wow, what kind of idiot would add this line of code?
+             * Did you know that this method is secretly asynchronous? And has been responsible for possibly dozens of
+             * racy unit tests? Why yes, yes dear reader - in fact it has!
+             *
+             * We have to ensure that the Watch completes before this method exit, so unfortunately we have to block.
+             *
+             * "Nuke the site from orbit. It's the only way to be sure."
+             */
+            taskCompletionSource.Task.Wait(RemainingOrDefault);
             return actorToWatch;
         }
 
@@ -368,7 +379,10 @@ namespace Akka.TestKit
         /// <returns>The actor to unwatch, i.e. the parameter <paramref name="actorToUnwatch"/></returns>
         public IActorRef Unwatch(IActorRef actorToUnwatch)
         {
-            _testState.TestActor.Tell(new TestActor.Unwatch(actorToUnwatch));
+            var taskCompletionSource = new TaskCompletionSource<bool>();
+            _testState.TestActor.Tell(new TestActor.Unwatch(actorToUnwatch, taskCompletionSource));
+            // See previous comment in Watch method
+            taskCompletionSource.Task.Wait(RemainingOrDefault);
             return actorToUnwatch;
         }
 
--- a/src/core/Akka.TestKit/TestKitBase.cs
+++ b/src/core/Akka.TestKit/TestKitBase.cs
@@ -357,8 +357,6 @@ namespace Akka.TestKit
         /// <returns>The actor to watch, i.e. the parameter <paramref name="actorToWatch"/></returns>
         public IActorRef Watch(IActorRef actorToWatch)
         {
-             var taskCompletionSource = new TaskCompletionSource<bool>();
-            _testState.TestActor.Tell(new TestActor.Watch(actorToWatch, taskCompletionSource));
             /*
              * Look, I know what you're thinking: wow, what kind of idiot would add this line of code?
              * Did you know that this method is secretly asynchronous? And has been responsible for possibly dozens of
@@ -368,7 +366,24 @@ namespace Akka.TestKit
              *
              * "Nuke the site from orbit. It's the only way to be sure."
              */
-            taskCompletionSource.Task.Wait(RemainingOrDefault);
+            WatchAsync(actorToWatch).Wait(RemainingOrDefault);
+            return actorToWatch;
+        }
+        
+        /// <summary>
+        /// Have the <see cref="TestActor"/> watch an actor and receive 
+        /// <see cref="Terminated"/> messages when the actor terminates.
+        /// </summary>
+        /// <param name="actorToWatch">The actor to watch.</param>
+        /// <returns>The actor to watch, i.e. the parameter <paramref name="actorToWatch"/></returns>
+        /// <remarks>
+        /// This method exists in order to make the asynchronous nature of the Watch method explicit.
+        /// </remarks>
+        public async Task<IActorRef> WatchAsync(IActorRef actorToWatch)
+        {
+            var taskCompletionSource = new TaskCompletionSource<bool>();
+            _testState.TestActor.Tell(new TestActor.Watch(actorToWatch, taskCompletionSource));
+            await taskCompletionSource.Task;
             return actorToWatch;
         }
 
@@ -379,10 +394,24 @@ namespace Akka.TestKit
         /// <returns>The actor to unwatch, i.e. the parameter <paramref name="actorToUnwatch"/></returns>
         public IActorRef Unwatch(IActorRef actorToUnwatch)
         {
+            // See previous comment in Watch method
+            UnwatchAsync(actorToUnwatch).Wait(RemainingOrDefault);
+            return actorToUnwatch;
+        }
+        
+        /// <summary>
+        /// Have the <see cref="TestActor"/> stop watching an actor.
+        /// </summary>
+        /// <param name="actorToUnwatch">The actor to unwatch.</param>
+        /// <returns>The actor to unwatch, i.e. the parameter <paramref name="actorToUnwatch"/></returns>
+        /// <remarks>
+        /// This method exists in order to make the asynchronous nature of the Unwatch method explicit.
+        /// </remarks>
+        public async Task<IActorRef> UnwatchAsync(IActorRef actorToUnwatch)
+        {
             var taskCompletionSource = new TaskCompletionSource<bool>();
             _testState.TestActor.Tell(new TestActor.Unwatch(actorToUnwatch, taskCompletionSource));
-            // See previous comment in Watch method
-            taskCompletionSource.Task.Wait(RemainingOrDefault);
+            await taskCompletionSource.Task;
             return actorToUnwatch;
         }
 
--- a/src/core/Akka.TestKit/TestKitBase.cs
+++ b/src/core/Akka.TestKit/TestKitBase.cs
@@ -176,11 +176,11 @@ namespace Akka.TestKit
             // Wait for the testactor to start
             WaitUntilTestActorIsReady(testActor);
 
-            if (!(this is INoImplicitSender))
+            if (this is not INoImplicitSender)
             {
                 InternalCurrentActorCellKeeper.Current = (ActorCell)((ActorRefWithCell)testActor).Underlying;
             }
-            else if (!(this is TestProbe))
+            else if (this is not TestProbe)
             //HACK: we need to clear the current context when running a No Implicit Sender test as sender from an async test may leak
             //but we should not clear the current context when creating a testprobe from a test
             {
@@ -548,7 +548,7 @@ namespace Akka.TestKit
             SupervisorStrategy supervisorStrategy,
             CancellationToken cancellationToken = default)
             => ChildActorOfAsync(props, name, supervisorStrategy, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
 
         /// <summary>
         /// Spawns an actor as a child of this test actor, and returns the child's IActorRef
@@ -566,7 +566,7 @@ namespace Akka.TestKit
         {
             TestActor.Tell(new TestActor.Spawn(props, name, supervisorStrategy));
             return await ExpectMsgAsync<IActorRef>(cancellationToken: cancellationToken)
-                .ConfigureAwait(false);
+                ;
         }
 
         /// <summary>
@@ -579,7 +579,7 @@ namespace Akka.TestKit
         public IActorRef ChildActorOf(
             Props props, SupervisorStrategy supervisorStrategy, CancellationToken cancellationToken = default)
             => ChildActorOfAsync(props, supervisorStrategy, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
 
         /// <summary>
         /// Spawns an actor as a child of this test actor with an auto-generated name, and returns the child's ActorRef.
@@ -593,7 +593,7 @@ namespace Akka.TestKit
         {
             TestActor.Tell(new TestActor.Spawn(props, Option<string>.None, supervisorStrategy));
             return await ExpectMsgAsync<IActorRef>(cancellationToken: cancellationToken)
-                .ConfigureAwait(false);
+                ;
         }
 
         /// <summary>
@@ -605,7 +605,7 @@ namespace Akka.TestKit
         /// <returns></returns>
         public IActorRef ChildActorOf(Props props, string name, CancellationToken cancellationToken = default)
             => ChildActorOfAsync(props, name, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
         
         /// <summary>
         /// Spawns an actor as a child of this test actor with a stopping supervisor strategy, and returns the child's ActorRef.
@@ -619,7 +619,7 @@ namespace Akka.TestKit
         {
             TestActor.Tell(new TestActor.Spawn(props, name, Option<SupervisorStrategy>.None));
             return await ExpectMsgAsync<IActorRef>(cancellationToken: cancellationToken)
-                .ConfigureAwait(false);
+                ;
         }
         
         /// <summary>
@@ -630,7 +630,7 @@ namespace Akka.TestKit
         /// <returns></returns>
         public IActorRef ChildActorOf(Props props, CancellationToken cancellationToken = default)
             => ChildActorOfAsync(props, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
 
         /// <summary>
         /// Spawns an actor as a child of this test actor with an auto-generated name and stopping supervisor strategy, returning the child's ActorRef.
@@ -642,7 +642,7 @@ namespace Akka.TestKit
         {
             TestActor.Tell(new TestActor.Spawn(props, Option<string>.None, Option<SupervisorStrategy>.None));
             return await ExpectMsgAsync<IActorRef>(cancellationToken: cancellationToken)
-                .ConfigureAwait(false);
+                ;
         }
 
         /// <summary>
--- a/src/core/Akka.TestKit/TestKitBase.cs
+++ b/src/core/Akka.TestKit/TestKitBase.cs
@@ -381,9 +381,7 @@ namespace Akka.TestKit
         /// </remarks>
         public async Task<IActorRef> WatchAsync(IActorRef actorToWatch)
         {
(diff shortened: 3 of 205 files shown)


Aaronontheweb:
This innocuous test is still failing, even after these changes, so it makes me think there's more async funk going on inside the TestKit that isn't completely resolved yet:

https://github.com/akkadotnet/akka.net/blob/0486f97c68858b6725c846cce79c6d6012128646/src/core/Akka.Persistence.TestKit.Tests/Bug4762FixSpec.cs#L67-L85

Aaronontheweb:
@akkadotnet/core not going to include this in Akka.NET v1.5.15 - too much of a hot potato and it needs time for lots of subsequent CI/CD runs to validate it before we roll out the "mission accomplished!" banner to our users

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -357,7 +357,18 @@ public void IgnoreNoMessages()
...
+            /*
+             * Look, I know what you're thinking: wow, what kind of idiot would add this line of code?
+             * Did you know that this method is secretly asynchronous? And has been responsible for possibly dozens of
+             * racy unit tests? Why yes, yes dear reader - in fact it has!
+             *
+             * We have to ensure that the Watch completes before this method exit, so unfortunately we have to block.
+             *
+             * "Nuke the site from orbit. It's the only way to be sure."
+             */
+            taskCompletionSource.Task.Wait(RemainingOrDefault);
```
Aaronontheweb: Nasty, decade old bug in the TestKit - `Watch` and `Unwatch` have always been secretly asynchronous, which is why tests like this one failed:

```csharp
[Fact]
        public async Task Routers_in_general_must_evict_terminated_routees()
        {
            var router = Sys.ActorOf(new RoundRobinPool(2).Props(Props.Create<Echo>()));
            router.Tell("");
            router.Tell("");

            var c1 = await ExpectMsgAsync<IActorRef>();
            var c2 = await ExpectMsgAsync<IActorRef>();

            Watch(router);
            Watch(c2);
            Sys.Stop(c2);
            (await ExpectTerminatedAsync(c2)).ExistenceConfirmed.Should().BeTrue();

            // it might take a while until the Router has actually processed the Terminated message
            await AwaitConditionAsync(async () =>
            {
                router.Tell("");
                router.Tell("");

                var res = await ReceiveWhileAsync(100.Milliseconds(), x =>
                {
                    if (x is IActorRef actorRef)
                        return actorRef;

                    return null;
                }, msgs: 2).ToListAsync();

                return res.Count == 2 && res.All(c => c.Equals(c1));
            });

            Sys.Stop(c1);
            (await ExpectTerminatedAsync(router)).ExistenceConfirmed.Should().BeTrue();
        }
```

Failed on `(await ExpectTerminatedAsync(c2)).ExistenceConfirmed.Should().BeTrue();`, which can only mean the actor was dead before the `TestActor` had a chance to go through.

src/core/Akka.TestKit/TestActor.cs
```diff
@@ -64,18 +65,15 @@ public class SetIgnore : INoSerializationVerificationNeeded
...
         public class Watch : INoSerializationVerificationNeeded
         {
-            private readonly IActorRef _actorToWatch;
-
-            /// <summary>
-            /// TBD
-            /// </summary>
-            /// <param name="actorToWatch">TBD</param>
-            public Watch(IActorRef actorToWatch) { _actorToWatch = actorToWatch; }
+            public Watch(IActorRef actorToWatch, TaskCompletionSource<bool> watchCompleted)
```
Aaronontheweb: Messaging fixes to support the `Watch` and `Unwatch` raciness.

src/core/Akka.TestKit/Internal/InternalTestActor.cs
```diff
@@ -56,16 +56,26 @@ protected override bool Receive(object message)
             {
                 case TestActor.SetIgnore setIgnore:
                     _ignore = setIgnore.Ignore;
-                    return true;
+                    return;
                 case TestActor.Watch watch:
                     Context.Watch(watch.Actor);
-                    return true;
+                    RunTask(async () =>
```
Aaronontheweb: Asynchronously run a messaging loop back to the actor being watched in order to make sure that our `Watch` has, in fact, gone through.

Aaronontheweb: Loop may not even be necessary - just completing the `TCS` might be good enough.

Aaronontheweb (commented):
Left some comments about my `Watch` and `Unwatch` fixes.

src/core/Akka.TestKit/Internal/InternalTestActor.cs
Aaronontheweb: Fixed this and simplified

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -548,7 +591,7 @@ public TimeSpan GetTimeoutOrDefault(TimeSpan? timeout)
             SupervisorStrategy supervisorStrategy,
             CancellationToken cancellationToken = default)
             => ChildActorOfAsync(props, name, supervisorStrategy, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use `GetAwaiter().GetResult()` because this automatically unwraps all `AggregateException`, using `.Result` here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -357,7 +357,33 @@ public void IgnoreNoMessages()
...
+        /// </summary>
+        /// <param name="actorToWatch">The actor to watch.</param>
+        /// <returns>The actor to watch, i.e. the parameter <paramref name="actorToWatch"/></returns>
+        /// <remarks>
+        /// This method exists in order to make the asynchronous nature of the Watch method explicit.
+        /// </remarks>
+        public async Task<IActorRef> WatchAsync(IActorRef actorToWatch)
+        {
+            var taskCompletionSource = new TaskCompletionSource<bool>();
+            _testState.TestActor.Tell(new TestActor.Watch(actorToWatch, taskCompletionSource));
```
Arkatufus: Would it be less API breaking if we do an `Ask<T>()` with timeout here instead of passing a `TaskCompletionSource` to the test actor?

Arkatufus: We also need to make sure that `TestActor.Watch` and `TestActor.Unwatch` is future compatible, just in case someone uses it to message a remote actor.

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -579,7 +622,7 @@ public TimeSpan GetTimeoutOrDefault(TimeSpan? timeout)
         public IActorRef ChildActorOf(
             Props props, SupervisorStrategy supervisorStrategy, CancellationToken cancellationToken = default)
             => ChildActorOfAsync(props, supervisorStrategy, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -605,7 +648,7 @@ public TimeSpan GetTimeoutOrDefault(TimeSpan? timeout)
         /// <returns></returns>
         public IActorRef ChildActorOf(Props props, string name, CancellationToken cancellationToken = default)
             => ChildActorOfAsync(props, name, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -630,7 +673,7 @@ public IActorRef ChildActorOf(Props props, string name, CancellationToken cancel
         /// <returns></returns>
         public IActorRef ChildActorOf(Props props, CancellationToken cancellationToken = default)
             => ChildActorOfAsync(props, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -40,18 +40,16 @@ public abstract partial class TestKitBase
             string hint = null,
             CancellationToken cancellationToken = default)
         {
-            return ExpectMsgAsync<T>(duration, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+            return ExpectMsgAsync<T>(duration, hint, cancellationToken).Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -75,7 +73,7 @@ public abstract partial class TestKitBase
             CancellationToken cancellationToken = default)
         {
             return ExpectMsgAsync(message, timeout, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -115,7 +113,7 @@ public abstract partial class TestKitBase
             CancellationToken cancellationToken = default)
         {
             return ExpectMsgAsync(isMessage, timeout, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -159,7 +157,7 @@ public abstract partial class TestKitBase
             CancellationToken cancellationToken = default)
         {
             return ExpectMsgAsync(assert, timeout, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -195,7 +193,7 @@ public abstract partial class TestKitBase
             CancellationToken cancellationToken = default)
         {
             return ExpectMsgAsync(isMessageAndSender, timeout, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -242,7 +240,7 @@ public abstract partial class TestKitBase
             CancellationToken cancellationToken = default)
         {
             return ExpectMsgAsync(assertMessageAndSender, timeout, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -279,7 +277,7 @@ public abstract partial class TestKitBase
             CancellationToken cancellationToken = default)
         {
             return ExpectMsgAsync(expected, comparer, timeout, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -317,7 +315,7 @@ public abstract partial class TestKitBase
             CancellationToken cancellationToken = default)
         {
             return ExpectTerminatedAsync(target, timeout, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -368,7 +366,7 @@ private void AssertPredicateIsTrueForMessage<T>(Predicate<T> isMessage, T m, str
             CancellationToken cancellationToken)
         {
             return InternalExpectMsgAsync(timeout, msgAssert, senderAssert, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -537,7 +528,7 @@ private async ValueTask InternalExpectNoMsgAsync(TimeSpan duration, Cancellation
         public T ExpectMsgAnyOf<T>(IEnumerable<T> messages, CancellationToken cancellationToken = default)
         {
             return ExpectMsgAnyOfAsync(messages, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -590,15 +581,14 @@ public ValueTask<T> ExpectMsgAnyOfAsync<T>(IEnumerable<T> messages, Cancellation
         {
             return ExpectMsgAllOfAsync(messages, cancellationToken)
                 .ToListAsync(cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Expect.cs
```diff
@@ -638,7 +628,7 @@ await foreach (var item in enumerable)
         {
             return ExpectMsgAllOfAsync(max, messages, cancellationToken)
                 .ToListAsync(cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
```diff
@@ -21,7 +21,7 @@ public IReadOnlyCollection<object> ExpectMsgAllOfMatchingPredicates(params Predi
     {
         return ExpectMsgAllOfMatchingPredicatesAsync(predicates, cancellationToken)
             .ToListAsync(cancellationToken)
-            .ConfigureAwait(false).GetAwaiter().GetResult();
+            .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
```diff
@@ -48,7 +48,7 @@ public IReadOnlyCollection<object> ExpectMsgAllOfMatchingPredicates(TimeSpan max
     {
         return ExpectMsgAllOfMatchingPredicatesAsync(max, predicates, cancellationToken)
             .ToListAsync(cancellationToken)
-            .ConfigureAwait(false).GetAwaiter().GetResult();
+            .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
```diff
@@ -58,7 +58,7 @@ public IReadOnlyCollection<object> ExpectMsgAllOfMatchingPredicates(TimeSpan max
     {
         max.EnsureIsPositiveFinite("max");
         var enumerable = InternalExpectMsgAllOfAsync(Dilated(max), predicates, cancellationToken: cancellationToken)
-            .ConfigureAwait(false).WithCancellation(cancellationToken);
+            .WithCancellation(cancellationToken);
```
Arkatufus: We don't need `.WithCancellation()` here, token is already passed into the method.

src/core/Akka.TestKit/TestKitBase_ExpectAllWithPredicates.cs
```diff
@@ -78,7 +78,7 @@ await foreach (var item in enumerable)
 
         var unexpectedMessages = new List<object>();
         var receivedMessages = InternalReceiveNAsync(predicateInfos.Count, max, shouldLog, cancellationToken)
-            .ConfigureAwait(false).WithCancellation(cancellationToken);
+            .WithCancellation(cancellationToken);
```
Arkatufus: We don't need `.WithCancellation()` here, token is already passed into the method.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -38,7 +38,7 @@ public abstract partial class TestKitBase
             CancellationToken cancellationToken = default)
         {
             return FishForMessageAsync(isMessage, max, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -68,7 +68,7 @@ public abstract partial class TestKitBase
...
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -100,7 +100,7 @@ public abstract partial class TestKitBase
             CancellationToken cancellationToken = default)
         {
             return FishForMessageAsync(isMessage, allMessages, max, hint, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -198,15 +198,15 @@ public async Task<ArrayList> WaitForRadioSilenceAsync(TimeSpan? max = null, uint
         public object ReceiveOne(TimeSpan? max = null,CancellationToken cancellationToken = default)
         {
             return ReceiveOneAsync(max, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -149,7 +149,7 @@ public async Task FishUntilMessageAsync<T>(TimeSpan? max = null, CancellationTok
                     return true; // please continue receiving, don't stop
                 },
                 cancellationToken: cancellationToken)
-                .ConfigureAwait(false).WithCancellation(cancellationToken);
+                .WithCancellation(cancellationToken);
```
Arkatufus: We don't need `.WithCancellation()` here, token is already passed into the method.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -230,7 +230,7 @@ public async ValueTask<object> ReceiveOneAsync(TimeSpan? max = null, Cancellatio
             CancellationToken cancellationToken = default)
         {
             var (success, messageEnvelope) = TryReceiveOneAsync(max, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -309,14 +309,14 @@ public async ValueTask<object> ReceiveOneAsync(TimeSpan? max = null, Cancellatio
         public object PeekOne(TimeSpan? max = null, CancellationToken cancellationToken = default)
         {
             return PeekOneAsync(max, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -329,14 +329,14 @@ public async ValueTask<object> PeekOneAsync(TimeSpan? max = null, CancellationTo
         public object PeekOne(CancellationToken cancellationToken)
         {
             return PeekOneAsync(cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -361,7 +361,7 @@ public async ValueTask<object> PeekOneAsync(CancellationToken cancellationToken)
         public bool TryPeekOne(out MessageEnvelope envelope, TimeSpan? max, CancellationToken cancellationToken)
         {
             var (success, result) = InternalTryPeekOneAsync(max, true, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -445,7 +445,7 @@ public ValueTask<(bool success, MessageEnvelope envelope)> TryPeekOneAsync(TimeS
         {
             return ReceiveWhileAsync(max, filter, msgs, cancellationToken)
                 .ToListAsync(cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -488,7 +488,7 @@ await foreach (var item in enumerable)
         {
             return ReceiveWhileAsync(max, idle, filter, msgs, cancellationToken)
                 .ToListAsync(cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -532,7 +532,7 @@ await foreach (var item in enumerable)
         {
             return ReceiveWhileAsync(filter, max, idle, msgs, cancellationToken)
                 .ToListAsync(cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -628,7 +628,7 @@ await foreach (var item in enumerable)
         {
            return ReceiveWhileAsync(shouldContinue, max, idle, msgs, shouldIgnoreOtherMessageTypes, cancellationToken)
                 .ToListAsync(cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -721,7 +721,7 @@ private void PopPeekedEnvelope(MessageEnvelope envelope)
         {
             return ReceiveNAsync(numberOfMessages, cancellationToken)
                 .ToListAsync(cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -752,7 +752,7 @@ await foreach (var item in enumerable)
         {
             return ReceiveNAsync(numberOfMessages, max, cancellationToken)
                 .ToListAsync(cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Within.cs
```diff
@@ -161,7 +161,7 @@ public abstract partial class TestKitBase
                     hint: null,
                     epsilonValue: epsilonValue,
                     cancellationToken: cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestProbe.cs
```diff
@@ -184,7 +184,7 @@ public void SendSystemMessage(ISystemMessage message, IActorRef sender)
             CancellationToken cancellationToken = default)
             where T : ActorBase
             => ChildActorOfAsync(Props.Create<T>(), name, supervisorStrategy, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestProbe.cs
```diff
@@ -212,7 +212,7 @@ public void SendSystemMessage(ISystemMessage message, IActorRef sender)
             SupervisorStrategy supervisorStrategy, CancellationToken cancellationToken = default)
             where T : ActorBase
             => ChildActorOfAsync(Props.Create<T>(), supervisorStrategy, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestProbe.cs
```diff
@@ -235,7 +235,7 @@ public void SendSystemMessage(ISystemMessage message, IActorRef sender)
         public IActorRef ChildActorOf<T>(string name, CancellationToken cancellationToken = default)
             where T : ActorBase
             => ChildActorOfAsync(Props.Create<T>(), name, cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestProbe.cs
```diff
@@ -256,7 +256,7 @@ public async Task<IActorRef> ChildActorOfAsync<T>(string name, CancellationToken
         public IActorRef ChildActorOf<T>(CancellationToken cancellationToken = default)
             where T : ActorBase
             => ChildActorOfAsync(Props.Create<T>(), cancellationToken)
-                .ConfigureAwait(false).GetAwaiter().GetResult();
+                .Result;
```
Arkatufus: We still need to use GetAwaiter().GetResult() because this automatically unwraps all AggregateException, using .Result here breaks behavior consistency.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -456,7 +456,7 @@ public ValueTask<(bool success, MessageEnvelope envelope)> TryPeekOneAsync(TimeS
             [EnumeratorCancellation] CancellationToken cancellationToken = default)
         {
             var enumerable = ReceiveWhileAsync(filter, max, Timeout.InfiniteTimeSpan, msgs, cancellationToken)
-                .ConfigureAwait(false).WithCancellation(cancellationToken);
+                .WithCancellation(cancellationToken);
```
Arkatufus: We don't need `.WithCancellation()` here, token is already passed into the method.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -500,7 +500,7 @@ await foreach (var item in enumerable)
             [EnumeratorCancellation] CancellationToken cancellationToken = default)
         {
             var enumerable = ReceiveWhileAsync(filter, max, idle, msgs, cancellationToken)
-                .ConfigureAwait(false).WithCancellation(cancellationToken);
+                .WithCancellation(cancellationToken);
```
Arkatufus: We don't need `.WithCancellation()` here, token is already passed into the method.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -730,7 +730,7 @@ private void PopPeekedEnvelope(MessageEnvelope envelope)
             [EnumeratorCancellation] CancellationToken cancellationToken = default)
         {
             var enumerable = InternalReceiveNAsync(numberOfMessages, RemainingOrDefault, true, cancellationToken)
-                .ConfigureAwait(false).WithCancellation(cancellationToken);
+                .WithCancellation(cancellationToken);
```
Arkatufus: We don't need `.WithCancellation()` here, token is already passed into the method.

src/core/Akka.TestKit/TestKitBase_Receive.cs
```diff
@@ -763,7 +763,7 @@ await foreach (var item in enumerable)
         {
             max.EnsureIsPositiveFinite("max");
             var enumerable = InternalReceiveNAsync(numberOfMessages, Dilated(max), true, cancellationToken)
-                .ConfigureAwait(false).WithCancellation(cancellationToken);
+                .WithCancellation(cancellationToken);
```
Arkatufus: We don't need `.WithCancellation()` here, token is already passed into the method.

Arkatufus (changes requested):
I have some concern on the `Watch` and `Unwatch` implementation and also how sync over async is implemented

src/core/Akka.TestKit/TestKitBase.cs
Aaronontheweb: > We also need to make sure that TestActor.Watch and TestActor.Unwatch is future compatible, just in case someone uses it to message a remote actor.

Ah, for a remote `TestActor` ? Yeah good point - I don't think we have any use cases where that happens in the test suite currently but rewriting this method to just send an ACK back eliminates that as a possible issue altogether.

src/core/Akka.TestKit/TestKitBase_Receive.cs
Aaronontheweb: That call was there previously and I think I just kept it during search and replace, but yes I'll clean that up.

src/core/Akka.TestKit/TestKitBase.cs
Aaronontheweb: Resolved

Arkatufus (approved):
LGTM

PR #7039, opened 2024-01-04
Docs: defining Akka.Analyzer rules

Aaronontheweb:
These will be linked from the tooltips inside Visual Studio whenever an error is flagged.


Aaronontheweb:
This looks good to go, but to close out https://github.com/akkadotnet/akka.analyzers/issues/6 I'll need to make some updates to the Akka.Analyzers package itself and have it reference the URI scheme for Akka.NET Analyzer rule ids.

PR #7059, opened 2024-01-12
Fix sharding entity id extractor nullability

Arkatufus:
EntityId extractor should be able to return a null value.

## Changes

Make the `HashCodeMessageExtractor` `Create()` and `Implementation.ctor()` take function reference that returns a nullable string.


Aaronontheweb (approved):
LGTM - good catch

PR #7061, opened 2024-01-12
Fix cluster sharding benchmark

Arkatufus:
## Changes

Fix broken cluster sharding benchmark

### DIFF
--- a/src/benchmark/Akka.Benchmarks/Configurations/Configs.cs
+++ b/src/benchmark/Akka.Benchmarks/Configurations/Configs.cs
@@ -37,16 +37,19 @@ namespace Akka.Benchmarks.Configurations
             var benchmarkAttribute = benchmarkCase.Descriptor.WorkloadMethod.GetCustomAttribute<BenchmarkAttribute>();
             var totalOperations = benchmarkAttribute?.OperationsPerInvoke ?? 1;
 
-            if (summary.HasReport(benchmarkCase))
-            {
-                var report = summary[benchmarkCase];
-                var nsPerOperation = report.ResultStatistics.Mean;
-                var operationsPerSecond = 1 / (nsPerOperation / 1e9);
-
-                return operationsPerSecond.ToString("N2");  // or format as you like
-            }
+            if (!summary.HasReport(benchmarkCase)) 
+                return "<not found>";
             
-            return "<not found>";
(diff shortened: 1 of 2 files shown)


src/benchmark/Akka.Benchmarks/Configurations/Configs.cs
```diff
@@ -37,16 +37,19 @@ public string GetValue(Summary summary, BenchmarkCase benchmarkCase, SummaryStyl
...
-                var operationsPerSecond = 1 / (nsPerOperation / 1e9);
-
-                return operationsPerSecond.ToString("N2");  // or format as you like
-            }
+            if (!summary.HasReport(benchmarkCase)) 
+                return "<not found>";
+            
+            var report = summary[benchmarkCase];
+            var statistics = report?.ResultStatistics;
+            if(statistics is null)
```
Arkatufus, line 45: Both report and statistics can be null if a benchmark case failed to run.

src/benchmark/Akka.Cluster.Benchmarks/Sharding/ShardingInfrastructure.cs
```diff
@@ -33,11 +33,11 @@ public ResolveResp(string entityId, Address addr)
             public Address Addr { get; }
         }
         
-        public ShardedEntityActor()
+        public ShardedEntityActor(string entityId)
         {
-            Receive<ShardingEnvelope>(e =>
+            Receive<Resolve>(_ =>
```
Arkatufus, line 38: This is the cause for the bug.

`ShardingEnvelope` does not get passed to the shard entity anymore, it is automatically unwrapped by the new extractor.

Arkatufus (commented):
Self review

src/benchmark/Akka.Cluster.Benchmarks/Sharding/ShardingInfrastructure.cs
Aaronontheweb: BITTEN BY MY OWN BUG FIX

Aaronontheweb: Thank you @Arkatufus

Aaronontheweb (approved):
LGTM

PR #7069, opened 2024-01-22
Use NoOffset when running query from start

MattiasJakobsson:
## Changes

I have a issue with a persistence plugin where the first event has the offset of 0. This test made the assumption that Sequence(0) was the same as NoOffset(). I've changed the test to use NoOffset() instead.
I also made the test virtual to follow the pattern.


Aaronontheweb (approved):
LGTM

PR #7078, opened 2024-01-30
Tentative nullref fix for `LogSource`

### LINKED ISSUE #6966: ActorPath.ToStringWithAddress is given null address parameter
**Version Information**
1.5.13
Remote 1.5.13

**Describe the bug**
Configuring a simple remote server as in the ChatServer example crashes the program.

Same issue as described on StackOverflow https://stackoverflow.com/questions/69983136/akka-net-net-framework-akka-remote-config-always-errors-on-start-then-works

**To Reproduce**
1. Download Akka.net github repo
2. open solution in VS 2022 on Windows 11 22H2
3. start the **ChatServer** example

**Expected behavior**
should just run 

**Actual behavior**
throws `System.NullReferenceException`

**Screenshots**
image

**Additional context**
Adding a check in `ActorPath.ToStringWithAddress` if the given address is null works (e.g. the server listens), but I have no clue about the side effects and why a `null` `address` is given

```
        private string ToStringWithAddress(Address address, bool includeUid)
        {
            if (address == null ||  IgnoreActorRef.IsIgnoreRefPath(this))
            {
                // we never change address for IgnoreActorRef
                return ToString();
            }

            long? uid = null;
            if (includeUid && _uid != ActorCell.UndefinedUid)
                uid = _uid;

            if (_address.Host != null && _address.Port.HasValue)
                return Join(_address.ToString().AsSpan(), uid);

            return Join(address.ToString().AsSpan(), uid);
        }
```

ilmax:
Fixes #6966

## Changes

As mentioned [here](https://github.com/akkadotnet/akka.net/issues/6966#issuecomment-1902333991) I'm getting this error 100% of times when starting the application that's configure to use Akka.Remoting.
Looking at the source code it seems to me that there's a race between [Remoting](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Remote/Remoting.cs#L210) and the Logging configuration.

Apparently the logging always wins the race and consequently throws a nullref. This exception is swallowed but I was wondering if the suggested fix may be of any value since it's unlikely that the remoting system may successfully configure (this is an educated guess) after the introduced null check because the offending path has only:
- a method call
- a string compare
- an if statement

Before throwing the nullref, also since the fallback is exactly the same as what the null path takes in my PR, I think that removing the start-up exception may valuable, but I'm not an expert in Akka source and I may be missing out here.

@Arkatufus FYI since you responded in the aforementioned issue.

### DIFF
--- a/src/core/Akka/Event/Logging.cs
+++ b/src/core/Akka/Event/Logging.cs
@@ -105,19 +105,8 @@ namespace Akka.Event
 
         public static string FromActorRef(IActorRef a, ActorSystem system)
         {
-            try
-            {
-                var defaultAddress = system.AsInstanceOf<ExtendedActorSystem>().Provider.DefaultAddress;
-                if (defaultAddress is null)
-                {
-                    return a.Path.ToString();
-                }
-                return a.Path.ToStringWithAddress(defaultAddress);
-            }
-            catch // can fail if the ActorSystem (remoting) is not completely started yet
-            {
-                return a.Path.ToString();
-            }
+            var defaultAddress = system.AsInstanceOf<ExtendedActorSystem>().Provider.DefaultAddress;
+            return defaultAddress is null ? a.Path.ToString() : a.Path.ToStringWithAddress(defaultAddress);
         }
     }
 
--- a/src/core/Akka/Event/Logging.cs
+++ b/src/core/Akka/Event/Logging.cs
@@ -107,7 +107,12 @@ namespace Akka.Event
         {
             try
             {
-                return a.Path.ToStringWithAddress(system.AsInstanceOf<ExtendedActorSystem>().Provider.DefaultAddress);
+                var defaultAddress = system.AsInstanceOf<ExtendedActorSystem>().Provider.DefaultAddress;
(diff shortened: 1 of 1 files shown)


src/core/Akka/Event/Logging.cs
```diff
@@ -107,7 +107,12 @@ public static string FromActorRef(IActorRef a, ActorSystem system)
         {
             try
             {
-                return a.Path.ToStringWithAddress(system.AsInstanceOf<ExtendedActorSystem>().Provider.DefaultAddress);
+                var defaultAddress = system.AsInstanceOf<ExtendedActorSystem>().Provider.DefaultAddress;
+                if (defaultAddress is null)
+                {
+                    return a.Path.ToString();
+                }
+                return a.Path.ToStringWithAddress(defaultAddress);
```
Arkatufus: This can be simplified by removing the try...catch block altogether

Arkatufus (approved):
LGTM, thank you for your contribution :)

ilmax:
Oh, I'm glad I was helpful (and will be happy to not see this exception anymore).

PR #7093, opened 2024-02-09
Made all snapshot tests virtual

MattiasJakobsson:
## Changes
I have a need to disable a couple of tests that aren't supported for the EventStore persistence. This means I can't use the SnapshotStoreSpec right now, but have to write the tests myself. With this change I could just override the once I want to skip.


Aaronontheweb (approved):
LGTM

Aaronontheweb:
@MattiasJakobsson I still need to bring the EventStore Persistence repo in this organization and setup CI/CD, don't I?

MattiasJakobsson:
@Aaronontheweb Yeah, would be nice if we could move that over. No rush tough, we're not using it at the moment.

PR #7099, opened 2024-02-16
Fix `IMessageExtractor` technical debts in tests

Arkatufus:
## Changes

* Convert all usage of `ExtractEntityId` and `ExtractShardId` delegates in unit tests to `IMessageExtractor` interface implementation.
* Improve NRE message caused by `IMessageExtractor.EntityMessage()` returning a null.
  * We can convert this into an Akka.Analyzer error rule.

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
@@ -1819,7 +1819,9 @@ namespace Akka.Cluster.Sharding
 
         private void DeliverMessage(string entityId, object msg, IActorRef snd)
         {
-            var payload = _extractor.EntityMessage(msg)!; // payload can't be null unless dev really screwed up
+            var payload = _extractor.EntityMessage(msg); // payload can't be null unless dev really screwed up
+            if (payload is null)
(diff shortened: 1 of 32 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/Shard.cs
```diff
@@ -1819,7 +1819,9 @@ private void PassivateCompleted(EntityId entityId)
 
         private void DeliverMessage(string entityId, object msg, IActorRef snd)
         {
-            var payload = _extractor.EntityMessage(msg)!; // payload can't be null unless dev really screwed up
+            var payload = _extractor.EntityMessage(msg); // payload can't be null unless dev really screwed up
+            if (payload is null)
+                throw new ArgumentNullException($"IMessageExtractor.EntityMessage() should never return null. Message extractor: {_extractor.GetType()}");
```
Arkatufus, line 1824: "IMessageExtractor.EntityMessage() should never return null" message is much better than an obscure null reference exception being thrown later on in this method when we try to access the payload variable.

Arkatufus (commented):
self-review

Aaronontheweb (approved):
LGTM

PR #7101, opened 2024-02-16
[DOCS] Add custom sharding handoff message docs

### LINKED ISSUE #7090: Update Sharding Custom Handoff Message documentation
I'm trying to implement Akka.Cluster.Sharding in an application that creates thousands and sometimes tens of thousands of Persistent Actors at a time. During creation of said actors, they will need to be rebalanced and moved to a different shard region. I did not fully understand how rebalancing and handoff messages worked. We lost messages in the actors mailbox since the default handoff message PoisonPill was handled before our messages were handled. 

**Which pages have you looked at?**
For example:
1. https://getakka.net/articles/clustering/cluster-sharding.html#integrating-cluster-sharding-with-persistent-actors

**What did these pages not make clear?**
When a shard region decides to rebalance, it will send an actor a Handoff message. The default Handoff message is a PoisonPill, which can cause significant problems for persistent actors. Commands are turned into deadletters and are not automatically reprocessed, so messages are lost. 

**How can we do it better?**
I would add a section that talks about rebalancing and what messages are sent during this process. Either a dedicated section to shard rebalance, or a note that persistent implementations should override the default Handoff message. 

**Additional suggestions**
I would also add more documentation around PoisonPill and the potential dangers it possesses. The [documentation](https://getakka.net/articles/actors/receive-actor-api.html#poisonpill) does state that the PosionPill message is added to the mailbox like other messages, but it does not mention that it is not handled like other mailbox messages. There is a list of key messages that are always handled regardless of the FSM state of the actor. We did not implement PoisonPills in our systems, so I didn't account for their processing. The system has specific cases where it sends actors a PoisonPill, and an actor's FSM state set to different states on Startup depending on the type of the actor. So if the PoisonPill is added after a message (msgA) is sent to a mailbox, then it is implied from the documentation that msgA will be handled before the PoisonPill is handled. This is not always the case if the actor's FSM state is controlled by internal mechanisms, i.e. [Github Issue 6321](https://github.com/akkadotnet/akka.net/issues/6321)

Arkatufus:
Fixes #7090

## Changes

Add custom sharding rebalance handoff message documentation

### DIFF
--- a/docs/articles/clustering/cluster-sharding.md
+++ b/docs/articles/clustering/cluster-sharding.md
@@ -83,6 +83,35 @@ As you may have seen in the examples above shard resolution algorithm is one of 
 
 By default re-balancing process always happens from nodes with the highest number of shards, to the ones with the smallest one. This can be configured into by specifying custom implementation of the `IShardAllocationStrategy` interface in `ClusterSharding.Start` parameters.
 
+## Shard Re-Balancing
+
+In the shard re-balance process, the coordinator first notifies all `ShardRegion` actors that a handoff for a shard has started.
+That means they will start buffering incoming messages for that shard, in the same way as if the shard location is unknown.
+During the re-balance process, the coordinator will not answer any requests for the location of shards that are being rebalanced, i.e. local buffering will continue until the handoff is completed.
+The `ShardRegion` responsible for the rebalanced shard will stop all entities in that shard by sending the specified `handoffStopMessage` (default `PoisonPill`) to them.
+When all entities have been terminated the `ShardRegion` owning the entities will acknowledge the handoff as completed to the coordinator.
+Thereafter the coordinator will reply to requests for the location of the shard, thereby allocating a new home for the shard, and then buffered messages in the `ShardRegion` actors are delivered to the new location; this means that the state of the entities are not transferred or migrated.
+If the state of the entities are of importance it should be persistent (durable) with Persistence, so that it can be recovered at the new location.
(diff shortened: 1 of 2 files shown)


docs/articles/clustering/cluster-sharding.md
```diff
@@ -83,6 +83,35 @@ As you may have seen in the examples above shard resolution algorithm is one of
...
+The `ShardRegion` responsible for the rebalanced shard will stop all entities in that shard by sending the specified `handoffStopMessage` (default `PoisonPill`) to them.
+When all entities have been terminated the `ShardRegion` owning the entities will acknowledge the handoff as completed to the coordinator.
+Thereafter the coordinator will reply to requests for the location of the shard, thereby allocating a new home for the shard, and then buffered messages in the `ShardRegion` actors are delivered to the new location; this means that the state of the entities are not transferred or migrated.
+If the state of the entities are of importance it should be persistent (durable) with Persistence, so that it can be recovered at the new location.
+
+The logic that decides which shards to re-balance is defined in a pluggable shard allocation strategy. The default implementation `LeastShardAllocationStrategy` allocates new shards to the `ShardRegion` (node) with least number of previously allocated shards.
+
+## Intercepting Actor Shutdown During Shard Re-Balancing Handoff Using Custom Stop Message
```
Aaronontheweb, line 98: LGTM

Aaronontheweb (approved):
LGTM

PR #7102, opened 2024-02-26
Suppress extremely chatty TimerScheduler debug messages

Arkatufus:
## Changes

* Add new setting "akka.actor.debug.log-timers", when set to true, `TimerScheduler` will emit debug messages.
* Set "akka.actor.debug.log-timers" default to false'
* Suppress `TimerScheduler` debug messages
* Add unit tests


src/core/Akka.Tests/Actor/Scheduler/TimerSchedulerDebugSpec.cs
Aaronontheweb, line 1: LGTM

Aaronontheweb (approved):
LGTM

src/core/Akka/Actor/Scheduler/TimerScheduler.cs
```diff
@@ -78,10 +78,14 @@ public TimerMsgNotInfluenceReceiveTimeout(object key, int generation, TimerSched
...
         private readonly Dictionary<object, Timer> _timers = new();
         private readonly AtomicCounter _timerGen = new(0);
+        private readonly bool _logDebug;
+        private readonly ILoggingAdapter _log;
 
         public TimerScheduler(IActorContext ctx)
         {
             _ctx = ctx;
+            _log = _ctx.System.Log;
+            _logDebug = ctx.System.Settings.Config.GetBoolean("akka.actor.debug.log-timers");
```
Aaronontheweb: Eh, one wrinkle here - it's better to have this setting get pulled into the `Settings` class once rather than parsed by each and every actor that uses the scheduler. For processes with a large number of actors that overhead + allocations from the parser can become significant. It's not a big deal when running a small number of actors that use `IWithTimers`, but if you're running hundreds of thousands that can add up.

Aaronontheweb (changes requested):
Need to make some optimizations around config loading

src/core/Akka/Actor/Scheduler/TimerScheduler.cs
Arkatufus: good idea

Aaronontheweb (approved):
LGTM

PR #7118, opened 2024-03-12
Made `akka.cluster.sharding.verbose-debug-logging` accept a `LogLevel`

### LINKED ISSUE #7097: Enable debug logging for ShardCoordinator/ShardRegion/Shard actors
We are seeing potential issues with the shard coordinator when recovering millions of entities, where shard regions do not receive acknowledgement from the coordinator. We'd like to allow debug logs to be switched on for sharding system-level actors without having to enable debug logging for the whole system

A configuration switch to perhaps change the log level of all logs from the shard coordinator and shard region actors, so we can view more information, even when the rest of the system is set to `Information` or `Warning` level. 

We could switch debug logging on for the whole system but this would cause unnecessary stress on our logging infrastructure.

Aaronontheweb:
## Changes

close #7097

For users with a large number of sharded entities, the `akka.cluster.sharding.verbose-debug-logging` HOCON setting now accepts a `LogLevel` argument. It also still supports the same boolean values as before too.

However, if you were to pass in the following HOCON:

```hocon
akka.cluster.sharding.verbose-debug-logging = INFO
```

All of the internal `ShardCoordinator` logs could now be captured at a `LogLevel.Info` level, so there wouldn't be any need to turn `LogLevel.Debug` on globally to capture these logs.

## Checklist

* [x] Design discussion issue #7097

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingLeavingSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingLeavingSpec.cs
@@ -28,7 +28,7 @@ namespace Akka.Cluster.Sharding.Tests
 
         public ClusterShardingLeavingSpecConfig(StateStoreMode mode)
             : base(mode: mode, loglevel: "DEBUG", additionalConfig: @"
-            akka.cluster.sharding.verbose-debug-logging = on
+            akka.cluster.sharding.verbose-debug-logging = INFO
             akka.cluster.sharding.rebalance-interval = 1s # make rebalancing more likely to happen to test for https://github.com/akka/akka/issues/29093
             akka.cluster.sharding.distributed-data.majority-min-cap = 1
             akka.cluster.sharding.coordinator-state.write-majority-plus = 1
--- a/src/contrib/cluster/Akka.Cluster.Sharding/DDataShardCoordinator.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/DDataShardCoordinator.cs
@@ -15,6 +15,7 @@ using Akka.Cluster.Sharding.Internal;
 using Akka.DistributedData;
 using Akka.Event;
 using Akka.Pattern;
+using static Akka.Cluster.Sharding.InternalConfigUtilities;
 
 namespace Akka.Cluster.Sharding
 {
@@ -29,7 +30,7 @@ namespace Akka.Cluster.Sharding
 
         private sealed class RememberEntitiesStoreStopped
         {
-            public static RememberEntitiesStoreStopped Instance = new();
+            public static readonly RememberEntitiesStoreStopped Instance = new();
 
             private RememberEntitiesStoreStopped()
             {
@@ -77,7 +78,7 @@ namespace Akka.Cluster.Sharding
 
         private readonly IActorRef _replicator;
         private readonly ShardCoordinator _baseImpl;
-        private bool VerboseDebug => _baseImpl.VerboseDebug;
+        private LogLevel? VerboseDebug => _baseImpl.VerboseDebug;
 
         private readonly IReadConsistency _stateReadConsistency;
         private readonly IWriteConsistency _stateWriteConsistency;
@@ -108,7 +109,7 @@ namespace Akka.Cluster.Sharding
         {
             _replicator = replicator;
             var log = Context.GetLogger();
-            var verboseDebug = Context.System.Settings.Config.GetBoolean("akka.cluster.sharding.verbose-debug-logging");
+            var verboseDebug = ParseVerboseLogSettings(Context.System.Settings.Config);
 
             _baseImpl = new ShardCoordinator(typeName, settings, allocationStrategy,
                 Context, log, verboseDebug, Update, UnstashOneGetShardHomeRequest);
@@ -171,9 +172,9 @@ namespace Akka.Cluster.Sharding
                 {
                     case GetSuccess g when g.Key.Equals(_coordinatorStateKey):
                         var existingState = g.Get(_coordinatorStateKey).Value.WithRememberEntities(Settings.RememberEntities);
-                        if (VerboseDebug)
-                            Log.Debug("{0}: Received initial coordinator state [{1}]", TypeName, existingState);
-                        else
+                        if (VerboseDebug.HasValue)
+                            Log.Log(VerboseDebug.Value, null, "{0}: Received initial coordinator state [{1}]", TypeName, existingState);
+                        else if(Log.IsDebugEnabled)
                             Log.Debug(
                                 "{0}: Received initial coordinator state with [{1}] shards",
                                 TypeName,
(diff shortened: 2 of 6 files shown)


src/contrib/cluster/Akka.Cluster.Sharding.Tests.MultiNode/ClusterShardingLeavingSpec.cs
```diff
@@ -28,7 +28,7 @@ public class ClusterShardingLeavingSpecConfig : MultiNodeClusterShardingConfig
 
         public ClusterShardingLeavingSpecConfig(StateStoreMode mode)
             : base(mode: mode, loglevel: "DEBUG", additionalConfig: @"
-            akka.cluster.sharding.verbose-debug-logging = on
+            akka.cluster.sharding.verbose-debug-logging = INFO
```
Aaronontheweb, line 31: Added a test case to ensure that this actually works with INFO logging here. Going to inspect the output once this runs.

src/contrib/cluster/Akka.Cluster.Sharding/DDataShardCoordinator.cs
```diff
@@ -29,7 +30,7 @@ internal sealed class DDataShardCoordinator : ActorBase, IWithTimers, IWithUnbou
 
         private sealed class RememberEntitiesStoreStopped
         {
-            public static RememberEntitiesStoreStopped Instance = new();
+            public static readonly RememberEntitiesStoreStopped Instance = new();
```
Aaronontheweb, line 33: Minor fix unrelated to this PR, but noticed it while I was working.

src/contrib/cluster/Akka.Cluster.Sharding/DDataShardCoordinator.cs
```diff
@@ -77,7 +78,7 @@ private RememberEntitiesLoadTimeout()
 
         private readonly IActorRef _replicator;
         private readonly ShardCoordinator _baseImpl;
-        private bool VerboseDebug => _baseImpl.VerboseDebug;
+        private LogLevel? VerboseDebug => _baseImpl.VerboseDebug;
```
Aaronontheweb, line 81: Change from a `bool` to a `LogLevel?` - these are all private APIs so there's no user-facing impact here other than the logging output.

src/contrib/cluster/Akka.Cluster.Sharding/DDataShardCoordinator.cs
```diff
@@ -108,7 +109,7 @@ private RememberEntitiesLoadTimeout()
         {
             _replicator = replicator;
             var log = Context.GetLogger();
-            var verboseDebug = Context.System.Settings.Config.GetBoolean("akka.cluster.sharding.verbose-debug-logging");
+            var verboseDebug = ParseVerboseLogSettings(Context.System.Settings.Config);
```
Aaronontheweb, line 112: Added a utility method to parse these configuration settings into `LogLevel?` since we have to make this call from both the `DDataShardCoordinator` and the `PersistentShardCoordinator`.

src/contrib/cluster/Akka.Cluster.Sharding/DDataShardCoordinator.cs
```diff
@@ -171,9 +172,9 @@ bool ReceiveDelegate(object message)
                 {
                     case GetSuccess g when g.Key.Equals(_coordinatorStateKey):
                         var existingState = g.Get(_coordinatorStateKey).Value.WithRememberEntities(Settings.RememberEntities);
-                        if (VerboseDebug)
-                            Log.Debug("{0}: Received initial coordinator state [{1}]", TypeName, existingState);
-                        else
+                        if (VerboseDebug.HasValue)
```
Aaronontheweb, line 175: The way I handled this type of case, which came up in a few areas:

* In the past we use to have two different `Debug` calls - one with more data when `VerboseDebugging` was enabled and then a regular one that was used with `Log.IsDebugEnabled` was true.
* All of these calls have been converted to include the first call to use whatever `LogLevel` is specified by `VerboseDebugging` in the event that `VerboseDebugging` is non-`null`.
* If `VerboseDebugging` is `null` and `Log.IsDebugEnabled`, we record the "lighter" information.

src/contrib/cluster/Akka.Cluster.Sharding/InternalConfigUtilities.cs
Aaronontheweb, line 1: Internal utility class for parsing the `LogLevel?`

src/contrib/cluster/Akka.Cluster.Sharding/reference.conf
```diff
@@ -235,7 +235,9 @@ akka.cluster.sharding {
   lease-retry-interval = 5s
 
   # Provide a higher level of details in the debug logs, often per routed message. Be careful about enabling
-  # in production systems.
+  # in production systems. 
+  # Values can be `off`, `on`, or an `akka.loglevel` value such as `INFO`.
+  # When set to `on` these logs will be logged using `LogLevel.Debug`.
```
Aaronontheweb, line 240: New comments describing the underlying HOCON changes. The default value of the `verbose-debug-logging` setting has been left intact.

Aaronontheweb (commented):
Detailed my changes

src/contrib/cluster/Akka.Cluster.Sharding/InternalConfigUtilities.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// <param name="coordinatorHocon">The HOCON config used by a shard coordinator</param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentException"></exception>
+    public static LogLevel? ParseVerboseLogSettings(Config coordinatorHocon)
+    {
+        var logLevel = coordinatorHocon.GetString("akka.cluster.sharding.verbose-debug-logging").ToLowerInvariant();
+        return logLevel switch
+        {
+            "off" => null,
+            "on" => LogLevel.DebugLevel,
```
Arkatufus: To conform to HOCON standard, this will also need to check for "true" and "false" value.

Arkatufus (changes requested):
This is a very weird design decision, but I guess it is needed. I only have one change request.

Aaronontheweb:
> This is a very weird design decision, but I guess it is needed. I only have one change request.

I think this comes down to that our log sources aren't easily filterable on the back-end, so it has to happen on the front-end. This change is also pretty easy to reverse if there were an issue with it.

src/contrib/cluster/Akka.Cluster.Sharding/InternalConfigUtilities.cs
Aaronontheweb: Good point - will do.

Aaronontheweb:
> I think this comes down to that our log sources aren't easily filterable on the back-end

I actually have some ideas on how that can be addressed too, in the grand scheme of things, but that'd be a 1.6 or 2.0 feature.

Arkatufus (approved):

PR #7128, opened 2024-03-21
Add `ActorBase` lifecycle flow unit tests

Arkatufus:
## Changes

* Add new property `ActiveTimers` to `ITimerScheduler` to retrieve all currently active timer keys
* Adds `ActorBase` lifecycle flow unit tests in relation to `IWithTimers` lifecycle

## Checklist


src/core/Akka/Actor/Scheduler/TimerScheduler.cs
```diff
@@ -150,6 +151,11 @@ public bool IsTimerActive(object key)
             return _timers.ContainsKey(key);
         }
 
+        /// <summary>
+        /// Retrieves all current active timer keys
+        /// </summary>
+        public ImmutableArray<object> ActiveTimers => _timers.Keys.ToImmutableArray();
```
Aaronontheweb: I would just return an `ICollection<object>` or whatever here so we don't have to allocate an immutable array each time.

Aaronontheweb (changes requested):
need to make the read operation zero-allocation.

src/core/Akka/Actor/Scheduler/TimerScheduler.cs
Aaronontheweb: I understand that this is probably done for the purposes of making sure we don't get rugged by the collection changing out from under us, but it's better to take that risk and leave the API un-opionated

Arkatufus: Done, though accessing `_timers.Keys` actually allocates internally inside the `Dictionary`, but that's unavoidable.

Aaronontheweb (approved):
LGTM

PR #7136, opened 2024-03-28
[WIP] Akka.IO / Akka.Streams: `ByteString` rewrite

Aaronontheweb:
## Changes

A leftover milestone from v1.5 - we are dragging `ByteString` kicking and screaming into the .NET Core era with major refactoring, namely replacing all of its internals `ReadOnlyMemory<byte>` usage.

## Checklist

### Latest `dev` Benchmarks 

### This PR's Benchmarks


Aaronontheweb:
Tests are not currently compiling - working on that.

src/core/Akka.Streams/Dsl/Framing.cs
```diff
@@ -169,39 +169,45 @@ public FramingException(string message) : base(message)
             protected FramingException(SerializationInfo info, StreamingContext context) : base(info, context) { }
         }
 
-        private static readonly Func<IEnumerator<byte>, int, int> BigEndianDecoder = (enumerator, length) =>
+        private static int BigEndianDecoder(ref ReadOnlySpan<byte> enumerator, int length)
```
to11mtm: If usable, `BinaryPrimitives.ReverseEndianness` may be preferable here; you can use it after various primitives are read, and it will fix the endianness.

Version for `long`, exampled by MessagePack-CSharp (but int should be able to be same pattern!):

```
        public static bool TryReadBigEndian(ref this SequenceReader<byte> reader, out long value)
        {
            if (!BitConverter.IsLittleEndian)
            {
                return reader.TryRead(out value);
            }

            return  TryReadReverseEndianness(ref reader, out value);
        }

        private static bool TryReadReverseEndianness(ref SequenceReader<byte> reader, out long value)
        {
            if (reader.TryRead(out value))
            {
                value = BinaryPrimitives.ReverseEndianness(value);
                return true;
            }

            return false;
        }
```

src/core/Akka.Streams/Dsl/Framing.cs
```diff
@@ -440,9 +446,14 @@ private void TryPushFrame()
...
-                        var parsedLength = _stage._intDecoder(iterator, _stage._lengthFieldLength);
-
+                        var iterator = _buffer.Memory.Span.Slice(_stage._lengthFieldOffset);
+                        var parsedLength = _stage._byteOrder switch {
+                            ByteOrder.BigEndian => BigEndianDecoder(ref iterator, _stage._lengthFieldLength),
+                            ByteOrder.LittleEndian => LittleEndianDecoder(ref iterator, _stage._lengthFieldLength),
+                            _ => throw new NotSupportedException($"ByteOrder {_stage._byteOrder} is not supported")
+                        }; 
+
+                        // TODO: AVOID ARRAY COPYING AGAIN HERE
```
to11mtm: Is this going to be separate PR?

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -132,7 +101,8 @@ public static void SetBuffer(this SocketAsyncEventArgs args, IEnumerable<ByteStr
             else
             {
                 args.SetBuffer(null, 0, 0);
-                args.BufferList = dataCollection.SelectMany(d => d.Buffers).ToList();
+                // TODO: fix this before we ship
+                args.BufferList = new List<ArraySegment<byte>>(dataCollection.Select(bs => new ArraySegment<byte>(bs.ToArray())));
```
to11mtm, line 105: Thing that feels very relevant, 

Short term, This -screams- pooling scenario and/or `IMemoryOwner` even if we have to deal with semantics internally.

I remember hacking something together with one pool or another during my Covid lockdown 'Akka streams remote transport experiment' and it helped a LOT. I think it was somewhere during or just before the hand written protobuf stuff.

Where 'considerations' get important, is the pinning of the arrays underlying the bufferlist.

IDK the current pinning 'status' but last I dug into they all get pinned for a while, and that sucks for the GC.

Thought/suggestion would be in general to use a pooled buffer of some form, but also strongly to consider/measure a single final buffer per `SetBuffer()` call being set on SAEA.

Having a pooled buffer increases likelyhood that it will, long term, survive to an older segment and hopefully avoid pinning costs in Gen0/Gen1. (Also, longer term, we can hopefully `#IFDEF` said pooled buffers into POH and get more gains.)

src/core/Akka/IO/SocketEventArgsPool.cs
```diff
@@ -132,7 +101,8 @@ public static void SetBuffer(this SocketAsyncEventArgs args, IEnumerable<ByteStr
             else
             {
                 args.SetBuffer(null, 0, 0);
```
to11mtm, line 103: Food for thought:

 - _If_ the 'buffer' is going to be typically null and we instead normally use `.BufferList`, it is better to hide this behind a `if (args.Buffer != null)` since the underlying call has a try/catch.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -945,6 +945,7 @@ public override void DoWrite(ConnectionInfo info)
             {
                 try
                 {
+                    // TODO: avoid use of SocketAsyncEventArgs on newer platforms
```
to11mtm: Stupid question, are newer APIs to the point where SAEA isn't worth the squeeze? Again it's been a while but last I knew SAEA was still the 'best' way.

src/core/Akka/Util/ByteString.cs
```diff
@@ -459,53 +382,38 @@ public int IndexOf(byte b, int from)
         /// <returns></returns>
         public bool HasSubstring(ByteString other, int index)
         {
-            // quick check: if subsequence is longer than remaining size, return false
-            if (other.Count > _count - index) return false;
+            if (other.Count == 0) return true; // Empty spans are always "found".
```
to11mtm: per discord comments, if we can use the 'Contains' extensions that exist in `System.Memory`, namely `public static int IndexOf<T>(this Span<T> span, ReadOnlySpan<T> value) where T : IEquatable<T>`, it would be strongly encouraged.

src/core/Akka/Util/ByteString.cs
```diff
@@ -519,16 +427,14 @@ public ByteString Concat(ByteString other)
...
-            var count = _count + other._count;
-            var len1 = _buffers.Length;
-            var len2 = other._buffers.Length;
-            var array = new ByteBuffer[len1 + len2];
-            Array.Copy(this._buffers, 0, array, 0, len1);
-            Array.Copy(other._buffers, 0, array, len1, len2);
-
-            return new ByteString(array, count);
+            // combine the two ReadOnlyMemory<byte> instances
+            var array = new byte[_memory.Length + other._memory.Length];
```
to11mtm, line 449: possible over-opt:
not sure whether compiler will wind up reading length twice here?

src/core/Akka/Util/ByteString.cs
```diff
@@ -544,27 +450,17 @@ public int CopyTo(byte[] buffer, int index, int count)
...
-                Array.Copy(b.Array, b.Offset, buffer, position, toCopy);
-                position += toCopy;
-                remaining -= toCopy;
-
-                if (remaining == 0) return count;
-            }
-
-            return 0;
+            count = Math.Min(count, Count);
+            _memory.Span.CopyTo(buffer.AsSpan(index, count));
```
to11mtm, line 470: NIT: IMO at bare minimum this could/should call into `public int CopyTo(ref Memory<byte> buffer, int index, int count)`. But in general the `CopyTo` overloads can be consolidated.


Additional NIT: Torn as to whether the smaller IL is a benefit compared to lacking bounds checks. (OTOH, this is 'I'd have to sharplab and benchmark to reliably comment.)

src/core/Akka/Util/ByteString.cs
```diff
@@ -544,27 +450,17 @@ public int CopyTo(byte[] buffer, int index, int count)
...
+            return count;
         }
 
         /// <summary>
         /// Copies content of a current <see cref="ByteString"/> into a provided <see cref="Memory{T}"/>
         /// <paramref name="buffer"/>
         /// </summary>
         /// <returns>The number of bytes copied</returns>
+        [Obsolete("This method will be removed in future versions of Akka.NET.")]
```
to11mtm, line 479: butwhy.jpg?

(also same question for other more optimal overloads compared to the `byte[]` variant)

src/core/Akka/Util/ByteString.cs
```diff
@@ -646,91 +517,61 @@ public int CopyTo(ref Span<byte> buffer, int index, int count)
...
             if (stream == null) throw new ArgumentNullException(nameof(stream));
-
-            foreach (var buffer in _buffers)
-            {
-                stream.Write(buffer.Array, buffer.Offset, buffer.Count);
-            }
+            
+            // TODO: remove this method in future versions of Akka.NET and use System.Memory APIs
+            var array = _memory.ToArray();
+            stream.Write(array, 0, array.Length);
```
to11mtm, line 540: hmm..

need to sleep on whether chunking could be better.

Note: as mentioned in other ramblings, certain read/write operations on streams and the like with `Memory` may have an intermediate copy to a magical CLR-internal buffer.

to11mtm (commented):
Took a closer look at current state and wanted to add FTT based on some past hacking at the subject.

src/core/Akka.Streams/Dsl/Framing.cs
Aaronontheweb: That's great - I'll give that a try, although I'm not sure about .NET Standard 2.0 support for it.

src/core/Akka.Streams/Dsl/Framing.cs
Aaronontheweb: Going to try to do that here but haven't gotten around to it yet

src/core/Akka/IO/SocketEventArgsPool.cs
Aaronontheweb: Rather than go down that road, I'd prefer to rewrite Akka.IO to not depend on `SocketAsyncEventArgs` and all of the complicated junk inside the TCP and UDP Akka.IO actors. I think some major simplification is in-order there for .NET 6 at least.

src/core/Akka/Util/ByteString.cs
Aaronontheweb: yeah I guess I could use the local variables but I think this is probably fine

src/core/Akka/Util/ByteString.cs
Aaronontheweb: It's duplicative - just use the methods on the `ReadOnlyMemory<byte>` property directly that already do this.

Aaronontheweb: Essentially, I'm trying to lower TCO here. .NET Standard 2.0 will be going bye-bye soon enough. We can get by just using the native System.Memory methods instead.

Aaronontheweb:
Let's see how accurately ChatGPT was able to describe the current FSM behavior of Akka.IO's `TcpConnection` actor, one of the areas I'm looking to refactor:

```mermaid
stateDiagram-v2
    [*] --> WaitingForRegistration: Connection established
    WaitingForRegistration --> Connected: Register
    WaitingForRegistration --> Closing: CloseCommand
    WaitingForRegistration --> WaitingForRegistration: ResumeReading / SuspendReading
    Connected --> ClosingWithPendingWrite: CloseCommand with pending write
    Connected --> PeerSentEOF: PeerClosed (keep open)
    Connected --> Closing: CloseCommand
    Connected --> Connected: WriteCommand / SocketSent / SuspendReading / ResumeReading
    PeerSentEOF --> Closing: CloseCommand
    ClosingWithPendingWrite --> Closing: Write completed
    Closing --> Closed: Connection closed
    Closed --> [*]
```

Aaronontheweb:
Honestly, not a bad job ChatGPT ^

Aaronontheweb:
Looks like `SocketAsyncEventArgs` is still the way to go for high performance sockets in .NET: https://twitter.com/_MihaZupan/status/1774911865019113974

src/core/Akka.Streams/Dsl/Framing.cs
Aaronontheweb: The https://learn.microsoft.com/en-us/dotnet/api/system.buffers.binary.binaryprimitives.tryreadint32bigendian?view=net-8.0&viewFallbackFrom=netstandard-2.0 methods won't work for this, FYI, due to us supporting configurable-length fields. I'm inclined not to mess with this class at all given that.

Aaronontheweb:
Gonna need to redo this - stuff like the framing code is phenomenally messy and difficult to reason about. That needs to be refactored and simplified first.

PR #7179, opened 2024-04-29
Akka.Event: add log filtering system to prevent Akka.NET logs from being emitted in first place

### LINKED ISSUE #7097: Enable debug logging for ShardCoordinator/ShardRegion/Shard actors

Aaronontheweb:
## Changes

Fixes #7097 and probably lots of other usability issues.

The goal behind this system is to prevent Akka.NET `LogEvent`s from being emitted in the first place in the event that their data is not useful. This should allow developers to do things like `akka.LOGLEVEL=DEBUG` without getting absolutely flooded with noise from busy parts of the system like Akka.Remote, Akka.Streams, and so on.

The general idea is to pass in a new `Setup` class:

```csharp
/// <summary>
/// Used to specify filters that can be used to curtail noise from sources in the Akka.NET log stream.
/// </summary>
public sealed class LogFilterSetup : Setup
{
    public LogFilterBase[] Filters { get; }
    
    public LogFilterEvaluator CreateEvaluator() => new(Filters);

    public LogFilterSetup(LogFilterBase[] filters)
    {
        Filters = filters;
    }
}
```

These `LogFilterBase` types are intended to be simple functions, but while taking into account some easy performance considerations:

```csharp
public abstract class LogFilterBase : INoSerializationVerificationNeeded, IDeadLetterSuppression
{
    /// <summary>
    /// Which part of the log message this filter is evaluating?
    /// </summary>
    /// <remarks>
    /// This actually has a performance implication - if we're filtering on the source, which
    /// is already fully "expanded" into its final string representation, we can try to fail fast
    /// on that without any additional allocations.
    ///
    /// If we're filtering on the message, we have to fully expand the log message first which
    /// involves allocations. Users on really tight performance budgets should be aware of this.
    /// </remarks>
    public abstract LogFilterType FilterType { get; }

    /// <summary>
    /// Returns <c>true</c> if the message should be kept, <c>false</c> if it should be dropped.
    /// </summary>
    /// <param name="evt">The <see cref="LogEvent"/> being evaluated.</param>
    /// <remarks>
    /// If there are multiple filters, all of them must return <c>true</c> for the message to be kept.
    /// </remarks>
    public abstract bool ShouldKeepMessage(LogEvent evt);

    /// <summary>
    /// Fast path designed to avoid allocating strings if we're filtering on the message content.
    /// </summary>
    /// <param name="part">The part of the message to evaluate.</param>
    /// <param name="content">Usually the fully expanded message content.</param>
    public abstract LogFilterDecision ShouldKeepMessage(LogFilterType part, string content);
}
```

I'm even including a `LogFilterBuilder` that should make it easy to construct these using a simple builder pattern. For a log to be printed, all filters must evaluate to `true`. Will include some examples of this below.

## Checklist

* [x] Design discussion issue #7097


src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -1725,6 +1725,7 @@ namespace Akka.Actor
         public int LogDeadLetters { get; }
         public bool LogDeadLettersDuringShutdown { get; }
         public System.TimeSpan LogDeadLettersSuspendDuration { get; }
+        public Akka.Event.LogFilterEvaluator LogFilter { get; }
```
Aaronontheweb, line 1728: New setting - this is where the log filter actually lives. Any custom logging plugin can access it this way.

src/core/Akka.Tests/Loggers/LogFilterEvaluatorSpecs.cs
```diff
@@ -0,0 +1,224 @@
...
+
+namespace Akka.Tests.Loggers;
+
+/// <summary>
+/// Goal of these specs are mostly to make sure our default <see cref="Regex"/>s are working as expected
+/// </summary>
+// ReSharper disable once ClassNeverInstantiated.Global
+public class LogFilterEvaluatorSpecs
+{
+    public class LogFilterSetupSpecs : AkkaSpec
```
Aaronontheweb, line 28: End to end specs - I can't easily intercept the default output from the `StandardOutWriter` so I'm swapping in my own `MinimalLogger` implementation and testing the results using that at the end. Not 100% clean, but close enough.

src/core/Akka.Tests/Loggers/LogFilterEvaluatorSpecs.cs
```diff
@@ -0,0 +1,224 @@
...
+            loggingAdapter1.Warning("foo-bar");
+            
+            // should be filtered out based on message content
+            loggingAdapter2.Warning("foo-bar");
+            
+            // should be allowed through
+            loggingAdapter2.Warning("baz");
+            
+            // expect only the last message to be received
+            ReceiveN(3);
```
Aaronontheweb, line 97: Log events will still hit the `EventStream` - the filter runs at the layer beneath this. **AND THAT IS IMPORTANT**: if we have the log filter layer running at this layer, at the point where `LogEvent`s are EMITTED, this WILL KILL PERFORMANCE. 

Instead, we make the decision about whether or not to keep / include a log at the layer right before the log is rendered. **This is the correct place to do this**.

Aaronontheweb: Anyway, that's why we still receive all 3 log events.

src/core/Akka.Tests/Loggers/LogFilterEvaluatorSpecs.cs
```diff
@@ -0,0 +1,224 @@
...
+            
+            // check that the last message was the one that was allowed through
+            _logger.Events.Count.Should().Be(1);
```
Aaronontheweb: The real logger should only have recorded the 1 log that passes both filters.

src/core/Akka.Tests/Loggers/LogFilterEvaluatorSpecs.cs
```diff
@@ -0,0 +1,224 @@
...
+            var msg = _logger.Events[0];
+            msg.Message.Should().Be("baz");
+            msg.LogSource.Should().StartWith("Akka.Util.Test2");
+        }
+    }
+
+    public class LogSourceCases
```
Aaronontheweb, line 107: The rest of these test cases are really just testing the default `Regex` used by the `LogFilterBuilder`.

src/core/Akka/Actor/Settings.cs
```diff
@@ -111,6 +111,18 @@ public Settings(ActorSystem system, Config config, ActorSystemSetup setup)
 
             LogLevel = Config.GetString("akka.loglevel", null);
             StdoutLogLevel = Config.GetString("akka.stdout-loglevel", null);
+            
+            // FILTER MUST ALWAYS BE LOADED BEFORE STANDARD OUT LOGGER
+            // check to see if we have a LogFilterSetup in the ActorSystemSetup
+            var logFilterSetup = Setup.Get<LogFilterSetup>();
```
Aaronontheweb, line 117: Load the `LogFilterSetup` safely from the `ActorSystemSetup`. **There will be no HOCON equivalent for loading this**.

src/core/Akka/Actor/Settings.cs
```diff
@@ -136,6 +148,9 @@ public Settings(ActorSystem system, Config config, ActorSystemSetup setup)
                 }
             }
             
+            // set the filter
+            StdoutLogger!.Filter = LogFilter;
```
Aaronontheweb, line 152: New property added to the `MinimaLogger` - it's settable `internal`ly so we do that here. This is what ties the `LogFilterEvaluator` into the `StandardOutLogger` in default Akka.NET logging scenarios.

src/core/Akka/Configuration/akka.conf
```diff
@@ -49,7 +49,8 @@ akka {
   
   # Fully qualified class name (FQCN) of the very basic logger used on startup and shutdown.
   # You can substitute the logger by supplying an FQCN to a custom class that implements Akka.Event.MinimumLogger
-  stdout-logger-class = "Akka.Event.StandardOutLogger"
+  # Set to null by default so the default logger is used.
+  stdout-logger-class = ""
```
Aaronontheweb, line 53: Disabled the HOCON loading here, as we always `new` up this class by default if this property is left blank. Removes 1 AOT warning that every Akka.NET application 100% hits and improves type safety.

Aaronontheweb: You can still override this property if you want to  - I do this in my unit test, which passes.

src/core/Akka/Event/LogFilter.cs
```diff
@@ -0,0 +1,319 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Akka.Actor;
+using Akka.Actor.Setup;
+
+namespace Akka.Event;
+
+/*
+ * NOTE: We do not do caching here - the number of log sources can be very large and
```
Aaronontheweb, line 19: **Do not do caching here** - no real utility since we're out of band anyway, and the variety of logs is utterly massive. You want better performance? DON'T EMIT THE LOGS IN THE FIRST PLACE.

Aaronontheweb: LogLevels and granular log emission settings are the right tools to solve performance problems that stem from logging, not the `LogFilter` - the `LogFilter` is here to let people run at more verbose log without keeping all of it.

src/core/Akka.Tests/Loggers/LoggerSpec.cs
```diff
@@ -127,7 +127,7 @@ public async Task StandardOutLogger_WithBadFormattingMustNotThrow()
         public void StandardOutLogger_PrintLogEvent_WithBadLogFormattingMustNotThrow(LogEvent @event)
         {
             var obj = new object();
-            obj.Invoking(_ => StandardOutLogger.PrintLogEvent(@event)).Should().NotThrow();
+            obj.Invoking(_ => StandardOutLogger.PrintLogEvent(@event, LogFilterEvaluator.NoFilters)).Should().NotThrow();
```
Aaronontheweb, line 130: Used the default `NoFilters` placeholder to keep this safe.

src/core/Akka/Event/LogFilter.cs
```diff
@@ -0,0 +1,314 @@
...
+
+    public override LogFilterDecision ShouldKeepMessage(LogFilterType part, string content)
+    {
+        if (part == LogFilterType.Source)
+            return _sourceRegex.IsMatch(content) ? LogFilterDecision.Drop : LogFilterDecision.Keep;
+        return LogFilterDecision.NoDecision;
+    }
+}
+
+public sealed class ExactMatchLogSourceFilter : LogFilterBase
```
Aaronontheweb, line 113: I might get rid of this - because the name of the `ActorSystem` _usually_ gets appended to the `LogSource`, therefore I think this will never work properly for most users. But, you never know.

src/core/Akka/Event/LogFilter.cs
```diff
@@ -0,0 +1,314 @@
...
+        if (part == LogFilterType.Message)
+            return _messageRegex.IsMatch(content) ? LogFilterDecision.Drop : LogFilterDecision.Keep;
+        return LogFilterDecision.NoDecision;
+    }
+}
+
+/// <summary>
+/// Runs inside the logging actor and evaluates if a log message should be kept.
+/// </summary>
+public class LogFilterEvaluator
```
Aaronontheweb, line 159: Decision-making class on whether or not to include or exclude logs.

src/core/Akka/Event/LogFilter.cs
```diff
@@ -0,0 +1,325 @@
...
+    /// </summary>
+    public bool EvaluatesLogSourcesOnly { get; }
+
+    public LogFilterEvaluator(LogFilterBase[] filters)
+    {
+        _filters = filters;
+        EvaluatesLogSourcesOnly = filters.All(x => x.FilterType == LogFilterType.Source);
+    }
+
+    public virtual bool ShouldTryKeepMessage(LogEvent evt, out string expandedLogMessage)
```
Aaronontheweb, line 177: A bit of a perf optimization here - if we're asked to evaluate including / excluding a log based on the content of its message, we have to expand the log earlier than usual. That can actually be a somewhat expensive operation (in the context of millions / second).

Since we have to expand _anyway_ for filtration purposes, we might as well also make this actor responsible for the original log expansion and emit that back out to the `StandardOutLogger` so we only have to do it once. That's why we have the `out` parameter here.

src/core/Akka/Event/LogFilter.cs
```diff
@@ -0,0 +1,325 @@
...
+    {
+        expandedLogMessage = string.Empty;
+
+        // fast and slow paths available here
+        if (EvaluatesLogSourcesOnly)
+        {
+            foreach (var filter in _filters)
+            {
+                // saves on allocations in negative cases, where we can avoid expanding the message
+                if (filter.ShouldKeepMessage(LogFilterType.Source, evt.LogSource) == LogFilterDecision.Drop)
```
Aaronontheweb: Fast path - WE ARE ONLY LOOKING AT LOGSOURCES, which don't need to be expanded. This is great. This is good. We are hoping we hit a negative case and can bail out right away. No expansion of the log is needed at all when this happens.

src/core/Akka/Event/LogFilter.cs
```diff
@@ -0,0 +1,325 @@
...
+                    return false;
+            }
+        }
+        else
+        {
+            // allocate the message just once
+            var nullCheck = evt.Message.ToString();
```
Aaronontheweb, line 194: Scenario where we have to consider at least 1 rule pertaining to the body of the log message - unavoidable allocation here even in scenarios where we drop the log.

src/core/Akka/Event/LogFilter.cs
```diff
@@ -0,0 +1,325 @@
...
+            if (nullCheck == null)
+                return false; // no message to filter
+
+            expandedLogMessage = nullCheck;
+
+            foreach (var filter in _filters)
+            {
+                switch (filter.FilterType)
+                {
+                    case LogFilterType.Source:
```
Aaronontheweb: A single vote cast for `LogFilterDecision.Drop` is all it takes for us to bail out.

src/core/Akka/Event/LogFilter.cs
```diff
@@ -0,0 +1,325 @@
...
+                    default:
+                        throw new ArgumentOutOfRangeException(nameof(filter), "Unknown filter type: " + filter.FilterType);
+                }
+            }
+        }
+
+        // expand the message if we haven't already
+        // NOTE: might result in duplicate allocations in third party logging libraries. They'll have to adjust their
+        // code accordingly after this feature ships.
+        expandedLogMessage = (string.IsNullOrEmpty(expandedLogMessage) ? evt.Message.ToString() : expandedLogMessage)!;
```
Aaronontheweb, line 211: Have to expand the log down here, since we're keeping it.

src/core/Akka/Event/StandardOutLogger.cs
```diff
@@ -9,12 +9,13 @@
 using Akka.Actor;
 using Akka.Util;
 using System.Text;
-using System.Threading;
 
 namespace Akka.Event
 {
     public abstract class MinimalLogger : MinimalActorRef
     {
+        public LogFilterEvaluator Filter { get; internal set; } = LogFilterEvaluator.NoFilters;
```
Aaronontheweb, line 17: Default setting, which maintains previous behavior.

src/core/Akka/Event/StandardOutLogger.cs
```diff
@@ -118,10 +118,15 @@ protected override void Log(object message)
...
         /// <param name="logEvent">The event to print</param>
-        internal static void PrintLogEvent(LogEvent logEvent)
+        /// <param name="filter"></param>
+        internal static void PrintLogEvent(LogEvent logEvent, LogFilterEvaluator filter)
         {
             try
             {
+                // short circuit if we're not going to print this message
+                if (!filter.ShouldTryKeepMessage(logEvent, out var expandedLogMessage))
+                    return;
```
Aaronontheweb, line 128: Self-explanatory.

src/core/Akka/Event/StandardOutLogger.cs
```diff
@@ -144,7 +149,7 @@ internal static void PrintLogEvent(LogEvent logEvent)
                     }
                 }
 
-                StandardOutWriter.WriteLine(logEvent.ToString(), color);
+                StandardOutWriter.WriteLine(expandedLogMessage, color);
```
Aaronontheweb, line 152: Avoid expanding the log a second time.

Aaronontheweb (commented):
Detailed my changes

Aaronontheweb:
Going to leave this PR open for comments for a while before it gets merged in

Aaronontheweb:
Also: documentation. Will add it once I get review comments.

Arkatufus (approved):
LGTM

Aaronontheweb:
I still need to do API approvals here

Issue #7185, opened 2024-05-06
Some error messages are not logged unless info logging is turned on


oyzar:
**Describe the bug**

        public static void Error<T1, T2, T3, T4>(this ILoggingAdapter log, Exception cause, string format, T1 arg1,
            T2 arg2, T3 arg3, T4 arg4)
        {
            if (!log.IsInfoEnabled)
                return;

            log.Log(LogLevel.ErrorLevel, cause, format, arg1, arg2, arg3, arg4);
        }
        
        this should be IsErrorEnabled

Aaronontheweb:
I believe this has been fixed via #7185

PR #7186, opened 2024-05-06
Fix IsInfoEnabled on Error bug (#7185)

Hannott:
## Changes

Changed IsInfoEnabled to IsErrorEnabled on Error logging function


Aaronontheweb (approved):
LGTM

PR #7194, opened 2024-05-22
fix race conditions with `ReceivePersistentActorTests`

Aaronontheweb:
## Changes

Subscriptions to the `EventStream` need to happen _before_ the actor emitting events starts, otherwise we can end up in a scenario where everything works but the assertion fails.

## Checklist

### DIFF
--- a/src/core/Akka.Persistence.Tests/ReceivePersistentActorTests.cs
+++ b/src/core/Akka.Persistence.Tests/ReceivePersistentActorTests.cs
@@ -31,9 +31,9 @@ namespace Akka.Persistence.Tests
             //Given
             var pid = "p-1";
             WriteEvents(pid, 1, 2, 3);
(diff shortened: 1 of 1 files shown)


src/core/Akka.Persistence.Tests/ReceivePersistentActorTests.cs
```diff
@@ -31,9 +31,9 @@ public void Given_persistent_actor_with_no_receive_command_specified_When_receiv
             //Given
             var pid = "p-1";
             WriteEvents(pid, 1, 2, 3);
-            var actor = Sys.ActorOf(Props.Create(() => new NoCommandActor(pid)), "no-receive-specified");
             Sys.EventStream.Subscribe(TestActor, typeof(UnhandledMessage));
```
Aaronontheweb, line 34: Need to subscribe to the `EventStream` first, so we don't end up with a race condition behind the scenes.

Aaronontheweb (commented):
Detailed my changes

PR #7197, opened 2024-05-23
Fixed: Akka.Cluster.Singleton no longer moves when node with matching role and higher `AppVersion` joins cluster

### LINKED ISSUE #7196: Akka.Cluster.Tools.Singleton: singleton moves earlier than expected - as soon as new node joins
**Version Information**
Version of Akka.NET? v1.5.0

**Describe the bug**

**Expected behavior**

Aaronontheweb:
## Changes

Fixes #7196 - not done yet.

## Checklist

* [x] Design discussion issue #7196

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/OldChangedBufferSpecs.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/OldChangedBufferSpecs.cs
@@ -1,0 +1,57 @@
+// -----------------------------------------------------------------------
+//  <copyright file="OldChangedBufferSpecs.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2024 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2024 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+#nullable enable
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Cluster.Tools.Singleton;
+using Akka.Configuration;
+using Akka.TestKit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Cluster.Tools.Tests.Singleton;
+
+/// <summary>
+/// Reproduction for https://github.com/akkadotnet/akka.net/issues/7196 - clearly, what we did
+/// </summary>
+public class OldChangedBufferSpecs : AkkaSpec
+{
+    public IActorRef CreateOldestChangedBuffer(string role, bool considerAppVersion)
+    {
+        return Sys.ActorOf(Props.Create(() => new OldestChangedBuffer(role, considerAppVersion)));
+    }
+    
+    private readonly ActorSystem _otherNodeV1;
+    private readonly ActorSystem _nonHostingNode;
+    private ActorSystem? _otherNodeV2;
+    
+    public OldChangedBufferSpecs(ITestOutputHelper output) : base("""
+                                          
+                                                        akka.loglevel = INFO
+                                                        akka.actor.provider = "cluster"
+                                                        akka.cluster.roles = [singleton]
+                                                        akka.cluster.auto-down-unreachable-after = 2s
+                                                        akka.cluster.singleton.min-number-of-hand-over-retries = 5
+                                                        akka.remote {
+                                                          dot-netty.tcp {
+                                                            hostname = "127.0.0.1"
+                                                            port = 0
+                                                          }
+                                                        }
+                                          """, output)
+    {
+        _otherNodeV1 = ActorSystem.Create(Sys.Name, Sys.Settings.Config);
+        _nonHostingNode = ActorSystem.Create(Sys.Name, ConfigurationFactory.ParseString("akka.cluster.roles = [other]")
+            .WithFallback(Sys.Settings.Config));
+    }
+
+    [Fact(DisplayName = "Singletons should not move to higher AppVersion nodes until after older incarnation is downed")]
+    public async Task Bugfix7196Spec()
+    {
+        
+    }
+}
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/OldChangedBufferSpecs.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/OldChangedBufferSpecs.cs
@@ -4,217 +4,148 @@
 //      Copyright (C) 2013-2024 .NET Foundation <https://github.com/akkadotnet/akka.net>
 //  </copyright>
 // -----------------------------------------------------------------------
+
 #nullable enable
+using System;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Cluster.Tools.Singleton;
-using Akka.Util;
+using Akka.Configuration;
+using Akka.TestKit;
 using FluentAssertions;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Akka.Cluster.Tools.Tests.Singleton;
 
 /// <summary>
 /// Reproduction for https://github.com/akkadotnet/akka.net/issues/7196 - clearly, what we did
 /// </summary>
-// public class OldChangedBufferSpecs : AkkaSpec
-// {
-//     public IActorRef CreateOldestChangedBuffer(string role, bool considerAppVersion)
-//     {
-//         return Sys.ActorOf(Props.Create(() => new OldestChangedBuffer(role, considerAppVersion)));
-//     }
-//     
-//     private readonly ActorSystem _otherNodeV1;
-//     private readonly ActorSystem _nonHostingNode;
-//     private ActorSystem? _otherNodeV2;
-//     
-//     public OldChangedBufferSpecs(ITestOutputHelper output) : base("""
-//                                           
-//                                                         akka.loglevel = INFO
-//                                                         akka.actor.provider = "cluster"
-//                                                         akka.cluster.roles = [singleton]
-//                                                         akka.cluster.auto-down-unreachable-after = 2s
-//                                                         akka.cluster.singleton.min-number-of-hand-over-retries = 5
-//                                                         akka.remote {
-//                                                           dot-netty.tcp {
-//                                                             hostname = "127.0.0.1"
-//                                                             port = 0
-//                                                           }
-//                                                         }
-//                                           """, output)
-//     {
-//         _otherNodeV1 = ActorSystem.Create(Sys.Name, Sys.Settings.Config);
-//         _nonHostingNode = ActorSystem.Create(Sys.Name, ConfigurationFactory.ParseString("akka.cluster.roles = [other]")
-//             .WithFallback(Sys.Settings.Config));
-//     }
-//
-//     [Fact(DisplayName = "Singletons should not move to higher AppVersion nodes until after older incarnation is downed")]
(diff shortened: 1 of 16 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/OldChangedBufferSpecs.cs
```diff
@@ -0,0 +1,151 @@
...
+    {
+        _hostNodeV1 = ActorSystem.Create(Sys.Name,
+            ConfigurationFactory.ParseString("akka.cluster.roles = [singleton]").WithFallback(Sys.Settings.Config));
+        _otherNodeV2 = ActorSystem.Create(Sys.Name,
+            ConfigurationFactory.ParseString("akka.cluster.roles = [singleton]").WithFallback(V2NodeConfig(Sys)));
+    }
+
+    [Fact(DisplayName =
+        "Singletons should not move to higher AppVersion nodes until after older incarnation is downed")]
+    public async Task Bugfix7196Spec()
```
Aaronontheweb: End to End reproduction spec:

1. Forms 2 node cluster: 1 with the role enabled for singleton (` _hostNodeV1`), the other without (`Sys`). Both nodes are running with `AppVersion` == 1.0.0.
1. `Sys` will create the `ClusterSingletonProxy` - which will point to the `SingletonManger` and the resulting `Singleton` actor hosted on `_hostNodeV1`.
1. `_otherNodeV2` joins the cluster with `AppVersion` == 1.0.2. The bug from #7196 meant that this node would IMMEDIATELY become the `Oldest` and the singleton would move (and the old one would not die, creating a split brain.)
1. We validate that the singleton is still on `_hostNodeV1` - which means we've fixed the bug, but we still need to make sure the oldest gets replaced if the current one leaves.
1. We down `_hostNodeV1` and observe that the singleton migrates over to `_otherNodeV2` afterwards.

This validates, end to end, that the issue is resolved. The only way we could make this more robust is adding a second `_hostNodeV1` instance and observe that `_otherNodeV2` gets elected in favor over it, but we have other tests below that ensure this.

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/OldestChangedBufferStateSpecs.cs
Aaronontheweb, line 1: Tests for my `OldestChangedBuffersState` abstraction, which is now shared by both the `SingletonProxy` and the `OldestChangedBuffer`

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
```diff
@@ -5,6 +5,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+#nullable enable
```
Aaronontheweb, line 8: Enable nullability here - just trying to do that as I go back through older code.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
```diff
@@ -102,7 +99,8 @@ public ClusterSingletonProxy(string singletonManagerPath, ClusterSingletonProxyS
             _memberAgeComparer = settings.ConsiderAppVersion
                 ? MemberAgeOrdering.DescendingWithAppVersion
                 : MemberAgeOrdering.Descending;
-            _membersByAge = ImmutableSortedSet<Member>.Empty.WithComparer(_memberAgeComparer);
+            _state = new OldestChangedBufferState(ImmutableSortedSet<Member>.Empty.WithComparer(_memberAgeComparer),
```
Aaronontheweb, line 102: Use the new `OldestChangedBufferState` instead of separate fields

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
```diff
@@ -221,31 +223,27 @@ private void IdentifySingleton()
...
             if (MatchingRole(member))
-                TrackChanges(() =>
-                {
-                    _membersByAge = _membersByAge.Remove(member); //replace
-                    _membersByAge = _membersByAge.Add(member);
-                });
+            {
+                var (newState, oldestChanged) = _state.AddMember(member);
+                _state = newState;
+                if (oldestChanged)
```
Aaronontheweb, line 233: Check to see if the oldest has changed, if so retry identifying the singleton (consistent with the behavior we had before, just no `Action` delegates and the `TrackChanges` method to do it.)

src/contrib/cluster/Akka.Cluster.Tools/Singleton/MemberAgeOrdering.cs
Aaronontheweb, line 1: Enabled nullability here and cleaned up some comments - otherwise no changes.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBuffer.cs
```diff
@@ -4,14 +4,13 @@
 //     Copyright (C) 2013-2023 .NET Foundation <https://github.com/akkadotnet/akka.net>
 // </copyright>
 //-----------------------------------------------------------------------
-
+#nullable enable
```
Aaronontheweb, line 7: Enable nullability

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBuffer.cs
```diff
@@ -102,11 +92,12 @@ public OldestChanged(UniqueAddress oldest)
...
         public OldestChangedBuffer(string role, bool considerAppVersion)
         {
-            _role = role;
             _memberAgeComparer = considerAppVersion
                 ? MemberAgeOrdering.DescendingWithAppVersion
                 : MemberAgeOrdering.Descending;
-            _membersByAge = ImmutableSortedSet<Member>.Empty.WithComparer(_memberAgeComparer);
+            var membersByAge = ImmutableSortedSet<Member>.Empty.WithComparer(_memberAgeComparer);
+            
+            State = new OldestChangedBufferState(membersByAge, role);
```
Aaronontheweb, line 100: Use the new `OldestChangedBufferState` type.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBuffer.cs
```diff
@@ -135,65 +126,64 @@ private void SetupCoordinatedShutdown()
...
             // all members except Joining and WeaklyUp
-            _membersByAge = state.Members
-                .Where(m => m.UpNumber != int.MaxValue && MatchingRole(m))
+            var newMembersByAge = clusterState.Members
+                .Where(m => m.UpNumber != int.MaxValue && State.MatchingRole(m))
                 .ToImmutableSortedSet(_memberAgeComparer);
+            
+           State = State with { MembersByAge = newMembersByAge };
+           
+           // compute the initial oldest - doesn't matter if it's not "safe" or not. That's our parent's problem.
```
Aaronontheweb, line 143: See comment for details - behavior is the same as before, just computed by the `State`

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBuffer.cs
```diff
@@ -135,65 +126,64 @@ private void SetupCoordinatedShutdown()
...
+            {
+                var (newState, oldestChanged) = State.AddMember(member);
+                State = newState;
+                if (oldestChanged)
                 {
-                    // replace, it's possible that the upNumber is changed
-                    _membersByAge = _membersByAge.Remove(member);
-                    _membersByAge = _membersByAge.Add(member);
-                });
+                    _changes = _changes.Enqueue(new OldestChanged(State.CurrentOldest?.UniqueAddress));
```
Aaronontheweb, line 171: Same behavior as before, but this is just now internally handled by the `State`.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBufferState.cs
```diff
@@ -0,0 +1,74 @@
...
+#nullable enable
+using System.Collections.Immutable;
+using System.Linq;
+
+namespace Akka.Cluster.Tools.Singleton;
+
+/// <summary>
+/// Immutable data object that represents the state of the oldest changed buffer.
+/// </summary>
+internal sealed record OldestChangedBufferState
```
Aaronontheweb, line 17: New immutable data structure designed to do two things:

1. Predict who the next leader should be
2. Not actually allow the leader to be changed until the previous leader has exited the cluster <-- THIS IS THE KEY BUGFIX.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBufferState.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        return string.IsNullOrEmpty(state.Role) || member.HasRole(state.Role);
+    }
+
+    public static (OldestChangedBufferState newState, bool oldestChanged) AddMember(this OldestChangedBufferState state, Member member)
+    {
+        if (MatchingRole(state, member))
+        {
+            // remove then add node to replace it, as it's possible that the upNumber is changed
+            return ComputeNextOldest(state with { MembersByAge = state.MembersByAge.Remove(member).Add(member) });
```
Aaronontheweb, line 45: Same "replacement" behavior that we used inside the `OldestChanged` buffer before.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBufferState.cs
```diff
@@ -0,0 +1,74 @@
...
+            return ComputeNextOldest(state with { MembersByAge = state.MembersByAge.Remove(member) });
+        }
+
+        return (state, false);
+    }
+
+    public static (OldestChangedBufferState newState, bool oldestChanged) ComputeNextOldest(this OldestChangedBufferState state)
+    {
+        // if the current oldest has not been removed, then it remains the oldest
+        if(state.CurrentOldest is not null && state.MembersByAge.Contains(state.CurrentOldest))
```
Aaronontheweb, line 64: Key bugfix happens right here - we don't allow a new "oldest" node to be decided until AFTER the previous one is removed from the cluster. This prevents split brains from happening AND it stops the `ClusterSingletonManager` from moving immediately after a new with a high `AppVersion` and matching role joins the cluster.

Aaronontheweb (commented):
Detailed my changes

Aaronontheweb:
I have a racy failure / instability of some kind in my test. Working on ruling that out before I get anyone else to look at this PR. I think it might have been related to the auto-downing setting I copied and pasted from the `Restart2Spec`.

Arkatufus (approved):
LGTM

Aaronontheweb:
So this "fix" will actually create additional problems - I'm thinking that the right thing to do here is to remove the "consider app version" support from Akka.Cluster.Singleton altogether. That problem doesn't seem workable without adding another communication layer around singletons, which I am not inclined to do. Better that we have a single heuristic for deciding where singletons live so we can keep the design of the infrastructure as simple as possible and eliminate any possibility of split brains.

Aaronontheweb:
Superseded by https://github.com/akkadotnet/akka.net/pull/7297

PR #7200, opened 2024-05-23
Akka.Actor.Serialization: remember to serialize envelope AND underlying type when using `serialize-messages=on`

Aaronontheweb:
## Changes

Bug: need to serialize both the envelope and the underlying type when `serialize-messages=on`

We found an issue stemming from a change I made in by https://github.com/akkadotnet/akka.net/pull/7010 where the underlying `IWrappedMessage` envelope would get discarded when  `serialize-messages=on`. The solution to this problem is to force serialization of the envelope AND the underlying type and return the product of both of those. Right now we are discarding the envelope by accident, which causes the `DistributedPubSubMediator` to be unable to process any messages when `serialize-messages=on`

## Checklist


Aaronontheweb:
Now that https://github.com/akkadotnet/akka.net/pull/7202 has been merged in, we should be good to go

PR #7205, opened 2024-05-24
[DOC] Add Akka.Analyzers AK1006 documentation

Arkatufus:
---
uid: AK1006
title: Akka.Analyzers Rule AK1006 - "Should not call `Persist()` or `PersistAsync()` inside a loop"
---

# AK1006 - Warning

Calling `Persist()` or `PersistAsync()` inside a loop is an anti-pattern and is non-performant. Consider collecting all events in a collection and then call `PersistAll()` or `PersistAllAsync()` after the loop instead.

## Cause

Akka.NET persistence tries its best to batch consecutive calls to `Persist()` or `PersistAsync()` methods, but there is no guarantee that all of them will be batched properly. 

For example, lets assume that for each command, you generate 10 `Persist()` operations. This can potentially create 10 asynchronous database connection to complete:

```csharp
using Akka.Persistence;

public class MyActor: ReceivePersistentActor
{
    public override string PersistenceId { get; }
    public MyActor(string persistenceId)
    {
        PersistenceId = persistenceId;
        CommandAny(obj =>
        {
            for (var i=0; i<10; i++)
            {
                Persist(i, o => {});
            }
        });
    }
}
```

## Resolution

If you know that a group of events need to be batched, it is a lot more performant to batch all `Persist()` operations into a single `PersistAll()` operation after the loop

```csharp
using Akka.Persistence;

public class MyActor: ReceivePersistentActor
{
    public override string PersistenceId { get; }
    public MyActor(string persistenceId)
    {
        PersistenceId = persistenceId;
        CommandAny(obj =>
        {
            var events = new List<int>();
            for (var i=0; i<10; i++)
            {
                events.Add(i);
            }
            PersistAll(events, o => {});
        });
    }
}
```

The persist success callback will be called for each event that are successfully persisted, so you do not need to change your logic inside the callback.

### DIFF
--- a/docs/articles/debugging/akka-analyzers.md
+++ b/docs/articles/debugging/akka-analyzers.md
@@ -11,16 +11,17 @@ Akka.Analyzer is a [Roslyn Analysis and Code Fix](https://learn.microsoft.com/en
 
 ## Supported Rules
 
-| Id                    | Title                                                                                                                    | Severity | Category     |
-|-----------------------|--------------------------------------------------------------------------------------------------------------------------|----------|--------------|
-| [AK1000](xref:AK1000) | Do not use `new` to create actors.                                                                                       | Error    | Actor Design |
-| [AK1002](xref:AK1002) | Must not await `Self.GracefulStop()` inside `ReceiveAsync<T>()` or `ReceiveAnyAsync`.                                    | Error    | Actor Design |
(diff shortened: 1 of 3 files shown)


Aaronontheweb (approved):
LGTM

PR #7219, opened 2024-05-31
added docs for `Valve` Akka.Streams stage

### LINKED ISSUE #7217: Akka.Streams: weird behavior with `RepeatPrevious`
**Version Information**
Version of Akka.NET? v1.5.20

**Describe the bug**

Consider the following code:

```csharp
var actorSystem = ActorSystem.Create("TestSys");

var source1 = Source.From(new[] { 1, 2, 3, 4 });
var source2 = Source.From(new[] { "Celsius", "Fahrenheit" });
var source3 = Source.From(new[] { 10.0 });

var source4 = source1.Zip(source2.RepeatPrevious())
   .ZipWith(source3.RepeatPrevious(), (a, b) => (a.Item1, a.Item2, b));
   
await source4.RunForeach(c => Console.WriteLine(c), actorSystem);
```

We get the following output:

```text
(1, Celsius, 10)
(2, Celsius, 10)
```

**Expected behavior**

I would expect we would get at least 4 elements - 1 for each of the items in `source1`. Why don't we?

**Actual behavior**

We only get 2 elements - and we never have a single "Fahrenheit" element appear.

**Additional context**

This could be a misconfiguration with how the various sources are being zipped together, but it doesn't read that way to me - this seems legitimate and like it should produce 4 total elements.

Aaronontheweb:
## Changes

close #7217 - documented how the `Valve` stage works

## Checklist

* [x] Design discussion issue #7217

### DIFF
--- a/docs/articles/streams/builtinstages.md
+++ b/docs/articles/streams/builtinstages.md
@@ -787,6 +787,17 @@ a function has to be provided to calculate the individual cost of each element.
 
 **completes** when upstream completes
 
+### Valve
+
+Materializes into a Task with an [`IValveSwitch`](xref:Akka.Streams.Dsl.IValveSwitch) which provides a method that will pause or resume elements being emitted from the stream.
+
+As long as the valve is closed it will backpressure.
+
+> [!NOTE]
+> Closing the valve could result in one element being buffered inside the stage, and if the stream completes or fails while being closed, that element may be lost.
+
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Only real question I have is that we don't have a proper DSL stage for `Valve` currently - is it worth adding one?

PR #7221, opened 2024-06-02
Make InternalTestActor override its SupervisionStrategy

### LINKED ISSUE #7062: [TestKit] SupervisorStrategy for TestActor is unused
**Version Information**

v1.5

**Describe the bug**

https://github.com/akkadotnet/akka.net/pull/4148 added `public IActorRef ChildActorOf(Props props, string name, SupervisorStrategy supervisorStrategy)` and similar methods to TestKitBase. However, the `supervisorStrategy` does not take effect.

It looks like `InternalTestActor` doesn't override `SupervisorStrategy()` to actually use its private `_supervisorStrategy`. See https://github.com/akkadotnet/akka.net/blob/0782c7dc069618e20a3791aaab667076f487ad96/src/core/Akka.TestKit/Internal/InternalTestActor.cs#L24

yanpitangui:
Fixes #7062 

## Changes

## Checklist

### Latest `dev` Benchmarks 

### This PR's Benchmarks


Aaronontheweb (commented):
LGTM but you should add a very simple regression test to verify that this works, mostly to protect us if we make some changes to the test suite in the future. Does that make sense @yanpitangui ?

Aaronontheweb:
Aha, looks like we already have a test that touches this and it's now failing as a result of this change: `Akka.TestKit.Tests.TestActorRefTests.TestProbeSpec.TestProbe_restart_a_failing_child_if_the_given_supervisor_says_so`

my guess is that the test is also wrong - so maybe if we just fix that we should be in good shape @yanpitangui

PR #7229, opened 2024-06-05
`ShardedDaemonProcess` push mode

### LINKED ISSUE #7195: Akka.Cluster.Sharding: support "push" mode for communicating with `ShardedDaemonProcess`s

Related to https://github.com/akkadotnet/Akka.Hosting/issues/453, specifically this line I wrote:

> One thing that occurs to me now is that it might be the the `ShardedDaemonProcess` was always designed for work-pulling only, which is another problem we'd have to deal with in the main Akka.NET repo itself.

When I went looking to fix that issue, lo and behold this original implementation comment from @ismaelhamed 

https://github.com/akkadotnet/akka.net/blob/3f0be58a661150c3d14572cd4615b526ba5e037a/src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs#L79-L92

So the `ShardedDaemonProcess` was originally designed for work-pulling, which makes sense: using it for projections et al.

However, I think without too much headache we can retool the `ShardedDaemonProcess` to also support work-pushing. In a universe where remote deployment of actors no longer exists (something that has been discussed, from time to time) and you want to have a fixed number of processors distributed evenly throughout the cluster for running things like ETL jobs, `ShardedDaemonProcess`es are perfect for that.

I'd like to change the signature of `ShardedDaemonProcess.Init` to return an `IActorRef` that points to a lightweight router that sits on top of the `ShardRegion` `IActorRef` that is used, internally, to power the `ShardedDaemonProcess`. This router can use one of the existing Akka.NET routing strategies but by default we'll use RoundRobin and distribute work evenly across all of the sharded daemon process instances.

Alternatively we could just use a remote `Pool` router for this work, but that has two big disadvantages:

1. Requires remote deployment and all of the serialization baggage that comes with it
2. You can't control the number of instances "globally" for the pool without limiting yourself to a single router - each node that creates the router adds more capacity to the pool. 

**Additional context**

I think the `KeepAlivePinger`, part of the `ShardedDaemonProcess`, provides some basis for how we could go about adding a public-facing routing capability that feeds into the `HashCodeMessageExtractor` for communicating with the worker processes directly: 

https://github.com/akkadotnet/akka.net/blob/3f0be58a661150c3d14572cd4615b526ba5e037a/src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs#L17-L63

Aaronontheweb:
## Changes

close #7195 - adds "push" mode to communicating with sharded daemon processes.

## Checklist

* [x] Design discussion issue #7195

### DIFF
--- a/docs/articles/clustering/cluster-sharded-daemon-process.md
+++ b/docs/articles/clustering/cluster-sharded-daemon-process.md
@@ -32,5 +32,20 @@ when starting up:
 
 ## Scalability  
 
-This cluster tool is intended for small numbers of consumers and will not scale well to a large set. In large clusters
-it is recommended to limit the nodes the sharded daemon process will run on using a role.
+This cluster tool is intended for small numbers of consumers and will not scale well to a large set. In large clusters it is recommended to limit the nodes the sharded daemon process will run on using a role.
+
+## Push-Based Communication
+
+[`ShardedDaemonProcess`](xref:Akka.Cluster.Sharding.ShardedDaemonProcess) also supports push-based communication not too dissimilar from a round-robin `Router`:
+
+[!code-csharp[IActorRef returned by ShardedDaemonProcess](../../../src/contrib/cluster/Akka.Cluster.Sharding.Tests/ShardedDaemonProcessProxySpec.cs?name=PushDaemon)]
+
+The `ShardedDaemonProcess.Init` call returns an `IActorRef` - any messages you send to this `IActorRef` will be routed to one of the `n` worker instances you specified.
+
+## Daemon Proxies
+
+You can also interact with `ShardedDaemonProcess` on nodes that don't use the same [Akka.Cluster role](xref:member-roles) as the `ShardedDaemonProcess` host itself.
+
+[!code-csharp[IActorRef returned by ShardedDaemonProcess](../../../src/contrib/cluster/Akka.Cluster.Sharding.Tests/ShardedDaemonProcessProxySpec.cs?name=PushDaemonProxy)]
+
+Under the covers we use a [`ShardRegionProxy`](xref:Akka.Cluster.Sharding.ClusterSharding#Akka_Cluster_Sharding_ClusterSharding_StartProxy_System_String_System_String_Akka_Cluster_Sharding_IMessageExtractor_) to forward any messages you send to the `IActorRef` returned by the `ShardedDaemonProcess.InitProxy` method.
--- a/src/contrib/cluster/Akka.Cluster.Sharding.Tests/ShardedDaemonProcessSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding.Tests/ShardedDaemonProcessSpec.cs
@@ -12,6 +12,7 @@ using Akka.Cluster.Tools.Singleton;
 using Akka.Configuration;
 using Akka.Event;
 using Akka.TestKit;
+using FluentAssertions;
 using Xunit;
 
 namespace Akka.Cluster.Sharding.Tests
@@ -21,7 +22,10 @@ namespace Akka.Cluster.Sharding.Tests
         private sealed class Stop
         {
             public static Stop Instance { get; } = new();
-            private Stop() { }
+
+            private Stop()
+            {
+            }
         }
 
         private sealed class Started
@@ -81,14 +85,16 @@ namespace Akka.Cluster.Sharding.Tests
 
         public ShardedDaemonProcessSpec()
             : base(GetConfig())
-        { }
+        {
+        }
 
         [Fact]
         public void ShardedDaemonProcess_must_have_a_single_node_cluster_running_first()
         {
             var probe = CreateTestProbe();
             Cluster.Get(Sys).Join(Cluster.Get(Sys).SelfAddress);
-            probe.AwaitAssert(() => Cluster.Get(Sys).SelfMember.Status.ShouldBe(MemberStatus.Up), TimeSpan.FromSeconds(3));
+            probe.AwaitAssert(() => Cluster.Get(Sys).SelfMember.Status.ShouldBe(MemberStatus.Up),
+                TimeSpan.FromSeconds(3));
         }
 
         [Fact]
@@ -126,7 +132,9 @@ namespace Akka.Cluster.Sharding.Tests
 
             var probe = CreateTestProbe();
             var settings = ShardedDaemonProcessSettings.Create(Sys).WithRole("workers");
-            ShardedDaemonProcess.Get(Sys).Init("roles", 3, id => MyActor.Props(id, probe.Ref), settings, PoisonPill.Instance);
+            var actorRef = ShardedDaemonProcess.Get(Sys).Init("roles", 3, id => MyActor.Props(id, probe.Ref), settings,
+                PoisonPill.Instance);
+            actorRef.Should().BeNull();
 
             probe.ExpectNoMsg();
         }
@@ -152,8 +160,10 @@ namespace Akka.Cluster.Sharding.Tests
         private void DocExample()
         {
             #region tag-processing
+
             var tags = new[] { "tag-1", "tag-2", "tag-3" };
             ShardedDaemonProcess.Get(Sys).Init("TagProcessors", tags.Length, id => TagProcessor.Props(tags[id]));
+
             #endregion
         }
     }
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
@@ -211,7 +211,7 @@ namespace Akka.Cluster.Sharding
             {
                 var sharding = ClusterSharding.Get(_system);
                 var shardingRef = sharding.Start(
-                    typeName: $"sharded-daemon-process-{name}",
+                    typeName: FormatWorkerProcessName(name),
                     entityPropsFactory: entityId => propsFactory(int.Parse(entityId)),
                     settings: shardingSettings,
                     messageExtractor: new MessageExtractor(numberOfShards),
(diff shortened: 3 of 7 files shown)


Aaronontheweb:
Thanks for fixing my dumb mistake in the test suite @Arkatufus - much appreciated lol

docs/articles/clustering/cluster-sharded-daemon-process.md
```diff
@@ -32,5 +32,20 @@ when starting up:
 
 ## Scalability  
 
-This cluster tool is intended for small numbers of consumers and will not scale well to a large set. In large clusters
-it is recommended to limit the nodes the sharded daemon process will run on using a role.
+This cluster tool is intended for small numbers of consumers and will not scale well to a large set. In large clusters it is recommended to limit the nodes the sharded daemon process will run on using a role.
+
+## Push-Based Communication
```
Aaronontheweb, line 37: Documented the two new pieces of functionality introduced here, which I'll review further down in the code itself

src/contrib/cluster/Akka.Cluster.Sharding.Tests/ShardedDaemonProcessSpec.cs
```diff
@@ -126,7 +132,9 @@ public void ShardedDaemonProcess_must_not_run_if_the_role_does_not_match_node_ro
 
             var probe = CreateTestProbe();
             var settings = ShardedDaemonProcessSettings.Create(Sys).WithRole("workers");
-            ShardedDaemonProcess.Get(Sys).Init("roles", 3, id => MyActor.Props(id, probe.Ref), settings, PoisonPill.Instance);
+            var actorRef = ShardedDaemonProcess.Get(Sys).Init("roles", 3, id => MyActor.Props(id, probe.Ref), settings,
```
Aaronontheweb, line 135: Assert that we get a `null` `IActorRef` back here.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -5,34 +5,41 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+#nullable enable
```
Aaronontheweb, line 8: Every time I make API updates to an existing area of Akka.NET I always try to enable `nullable` so it'll be less jarring when we make it the default in the future.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -5,34 +5,41 @@
...
+            }
         }
 
         public string Name { get; }
         public string[] Identities { get; }
         public IActorRef ShardingRef { get; }
         public ShardedDaemonProcessSettings Settings { get; }
 
-        public ITimerScheduler Timers { get; set; }
+        public ITimerScheduler Timers { get; set; } = null!; // gets set by Akka.NET
```
Aaronontheweb, line 35: Made compiler warning go away.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -66,18 +74,55 @@ internal sealed class MessageExtractor : HashCodeMessageExtractor
...
-    /// <para>This extension runs a pre set number of actors in a cluster.</para>
+    /// Used to support push-based communication with the <see cref="ShardedDaemonProcess"/> - uses
+    /// round-robin messaging to distribute messages to the actors in the cluster.
+    /// </summary>
+    /// <remarks>
+    /// NOTE: does not use a traditional <see cref="Router"/> to distribute messages because we want to use the
+    /// <see cref="ShardRegion"/> mechanism in order to piggyback off of its reliability and message buffering
+    /// features, which <see cref="Router"/> and <see cref="ActorSelection"/> do not support.
+    /// </remarks>
+    internal sealed class DaemonMessageRouter : UntypedActor
```
Aaronontheweb, line 98: new class - `DaemonMessageRouter`. Performs round-robin message routing (according to the sending node - we don't care what other nodes who might be sending messages are doing) to all of the individual worker actors.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -66,18 +74,55 @@ internal sealed class MessageExtractor : HashCodeMessageExtractor
...
+            _entityIds = entityIds;
+            // validate that we have at least 1 entityId
+            if (_entityIds.Length == 0)
+                throw new ArgumentException("At least one entityId must be provided", nameof(entityIds));
+            _shardingRef = shardingRef;
+        }
+
+        protected override void OnReceive(object message)
+        {
+            var nextId = _entityIds[_index % _entityIds.Length];
```
Aaronontheweb, line 115: Use modulo to select the next available worker.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -66,18 +74,55 @@ internal sealed class MessageExtractor : HashCodeMessageExtractor
...
+            
+            // have to remember to always allow the sharding envelope to be forwarded
+            _shardingRef.Forward(new ShardingEnvelope(nextId, message));
```
Aaronontheweb, line 118: Forward the message and wrap it inside a `ShardingEnvelope` so the underlying Akka.Cluster.Sharding infrastructure can safely deliver it.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -66,18 +74,55 @@ internal sealed class MessageExtractor : HashCodeMessageExtractor
...
+            if (_index == int.MaxValue) _index = 0;
```
Aaronontheweb, line 119: Wrap-around if we're about to integer overflow.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -106,28 +151,35 @@ public class ShardedDaemonProcess : IExtension
         /// <param name="name">TBD</param>
         /// <param name="numberOfInstances">TBD</param>
         /// <param name="propsFactory">Given a unique id of `0` until `numberOfInstance` create an entity actor.</param>
-        public void Init(string name, int numberOfInstances, Func<int, Props> propsFactory) => 
+        /// <returns>A reference to a router actor that will distribute all messages evenly across the workers
+        /// using round-robin message routing. <c>null</c> if the ShardedDaemonProcess is misconfigured.</returns>
+        public IActorRef? Init(string name, int numberOfInstances, Func<int, Props> propsFactory) =>
```
Aaronontheweb, line 156: Modified all of these methods to return an `IActorRef?` - will return `null` if we're the wrong role type to be hosting a worker and we'll also now log a `WARNING` accordingly if that happens.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -150,29 +203,71 @@ public void Init(string name, int numberOfInstances, Func<int, Props> propsFacto
...
                 shardingBaseSettings.LeaseSettings);
 
-            if (string.IsNullOrEmpty(shardingSettings.Role) || Cluster.Get(_system).SelfRoles.Contains(shardingSettings.Role))
+            if (string.IsNullOrEmpty(shardingSettings.Role) ||
+                Cluster.Get(_system).SelfRoles.Contains(shardingSettings.Role))
             {
                 var sharding = ClusterSharding.Get(_system);
                 var shardingRef = sharding.Start(
-                    typeName: $"sharded-daemon-process-{name}",
+                    typeName: FormatWorkerProcessName(name),
```
Aaronontheweb, line 214: Use a shared function to name `ShardedDaemonProcess`es and proxies - uses the exact same lexical structure as before so it won't break any existing deployments.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -150,29 +203,71 @@ public void Init(string name, int numberOfInstances, Func<int, Props> propsFacto
...
+        /// Starts a proxy for a sharded daemon process running in a different role.
+        /// </summary>
+        /// <param name="name">The name of this daemon worker set</param>
+        /// <param name="numberOfInstances">The number of instances that belong on this set -
+        /// note that this value must match the value used in the `Init` call on the hosting-nodes
+        /// or messages may be lost.</param>
+        /// <param name="role">The role where the worker actors are hosted.</param>
+        /// <returns>A reference to a router actor that will distribute all messages evenly across the workers
+        /// using round-robin message routing.</returns>
+        public IActorRef InitProxy(string name, int numberOfInstances, string role)
```
Aaronontheweb, line 250: This is new - since we added push-based communication to interact with `ShardedDaemonProcess`es I thought it also made sense to leverage `ShardRegionProxy` to allow non-hosting nodes to communicate with the worker processes. The only issue here is that if the user gets the number of entities wrong, then our routing logic will be skewed or broken. Not too different than if the user gets the message extractor, role, or entity type wrong so it's not a big deal IMHO.

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcessSettings.cs
```diff
@@ -4,7 +4,7 @@
 //     Copyright (C) 2013-2023 .NET Foundation <https://github.com/akkadotnet/akka.net>
 // </copyright>
 //-----------------------------------------------------------------------
-
+#nullable enable
```
Aaronontheweb, line 7: Enable `nullable` on the settings class too

Aaronontheweb (commented):
Detailed my changes - none of these API changes are breaking from a binary or source-compatibility standpoint, since no one could use return values from a `void` method before anyway.

Arkatufus (approved):
LGTM

PR #7232, opened 2024-06-06
Tidied up the language on Fault Tolerance page

sctmike:
## Changes

Tidied up the language - no material changes to the code


Aaronontheweb (approved):
LGTM - much clearer than what we had before. Thank you!

PR #7251, opened 2024-06-12
Modernize Cluster.Client code

Arkatufus:
Cleaning up and modernizing Akka.Cluster.Tools.Client code, auto-refactored using Rider

## Changes

* Use file scoped namespace
* Change if..else message handling structure to switch..case
* Change switch..return pattern to use pattern matching


Aaronontheweb (approved):
LGTM

PR #7263, opened 2024-06-23
made examples, benchmarks, and internal libraries unpackable

Aaronontheweb:
Working on some future-proofing for the build system so we can migrate off of FAKE and eventually just do everything with `dotnet` plus some other external `dotnet tool` commands like `incrementalist` and `docfx`.

One thing we need in order to do this is to make it possible to run `dotnet publish` on the `src/Akka.sln` file itself is to lock down which projects can actually be published. This hasn't been an issue with production NuGet deploys yet but I ran into this as soon as I tried some local builds on Linux.


Aaronontheweb:
Tests failed because of `Akka.Tests.Shared.Internal` got touched, but these were some of the usual suspect racy specs - not anything affected by this change

PR #7270, opened 2024-06-27
Fix possible problems with ClusterClient Discovery

Arkatufus:
## Changes

* Limits the number of initial contacts being used in ClusterClient Discovery to improve lost connection detection
* Randomly choose the initial contacts to prevent thundering herd problem

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientDiscovery.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientDiscovery.cs
@@ -14,6 +14,7 @@ using Akka.Actor;
 using Akka.Discovery;
 using Akka.Event;
 using Akka.Util;
+using Akka.Util.Internal;
 
 #nullable enable
 namespace Akka.Cluster.Tools.Client;
@@ -248,28 +249,10 @@ public class ClusterClientDiscovery: UntypedActor, IWithUnboundedStash, IWithTim
         if (fullContact.Length <= count)
             return fullContact;
         
-        Shuffle(fullContact);
+        fullContact.Shuffle();
         return fullContact.Take(count).ToArray();
     }
 
-    /// <summary>
-    /// Fisher-Yates in-place array shuffle algorithm
-    /// </summary>
-    /// <param name="array"></param>
-    /// <typeparam name="T"></typeparam>
-    private static void Shuffle<T>(T[] array)
-    {
-        var rnd = ThreadLocalRandom.Current;
-        var n = array.Length;
-        for (var i = 0; i < (n - 1); i++)
-        {
-            var r = i + rnd.Next(n - i);
-            var t = array[r];
-            array[r] = array[i];
-            array[i] = t;
-        }
-    }
-    
     private Receive Active(Contact[] contacts)
     {
         if(_verboseLogging && _log.IsDebugEnabled)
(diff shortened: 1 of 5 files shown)


src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientDiscovery.cs
```diff
@@ -49,6 +50,7 @@ private sealed record ResolveResult(Contact Contact, IActorRef? Subject);
     private readonly string _targetActorSystemName;
     private readonly string _receptionistName;
     private readonly string _transportProtocol;
+    private readonly int _numberOfContacts;
```
Aaronontheweb, line 54: This is just the number of initial contacts we need, right?

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientDiscovery.cs
```diff
@@ -229,6 +236,40 @@ async Task<ResolveResult[]> VerifyContacts()
...
+        Shuffle(fullContact);
+        return fullContact.Take(count).ToArray();
+    }
+
+    /// <summary>
+    /// Fisher-Yates in-place array shuffle algorithm
+    /// </summary>
+    /// <param name="array"></param>
+    /// <typeparam name="T"></typeparam>
+    private static void Shuffle<T>(T[] array)
```
Aaronontheweb: Do we have another implementation of this lying around somewhere? I could have sworn I'd written one before

Aaronontheweb (commented):
Left some general comments

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientDiscovery.cs
Arkatufus: it limits the number of contacts we use when we instantiate the ClusterClient actor. Too many of these will set the initial contacts to a higher than optimum number. The thing is that ClusterClient initial contacts never gets trimmed when they can't be contacted, if we use the whole discovery resolve result as the ClusterClient initial contact list, there will be a big burst of network connection attempt when the ClusterClient actor starts up and whenever its internal contact list could not be contacted anymore.

Aaronontheweb: Got it, makes sense

src/contrib/cluster/Akka.Cluster.Tools/Client/ClusterClientDiscovery.cs
Arkatufus: no idea

Arkatufus: There is one in the Akka.Util.Internal.ArrayExtensions, but its internal

Aaronontheweb: I would add a link to that file and just use that (or you can add a friend assembly to Akka.Discovery; link is probably cleaner.)

Aaronontheweb: That's just a nitpick though

Arkatufus: Done, turns out the core Akka project is already a friend of Akka.Cluster.Tools

Aaronontheweb (approved):
LGTM

PR #7274, opened 2024-07-02
Improve ClusterClientDiscovery feature

Arkatufus:
## Changes

* Add exponential backoff to the discovery interval
* Add CoordinatedShutdown support
* Use the new Akka.Management "/cluster-client/receptionist" endpoint


Aaronontheweb (approved):
LGTM - did review on call with @Arkatufus

PR #7276, opened 2024-07-03
Fix ClusterClientDiscovery, missing port name in Lookup

Arkatufus:
## Changes

Fix ClusterClientDiscovery, missing port name argument in Akka.Discovery Lookup constructor


Aaronontheweb (approved):
LGTM

PR #7278, opened 2024-07-10
Akka.Cluster: harden ignored gossip messages

Aaronontheweb:
## Changes

Added our unique address just in case the issue here is gossip intended for previous incarnations of ourselves following a fast restart. This is just for improving diagnostic clarity.

## Checklist

### DIFF
--- a/src/core/Akka.Cluster/ClusterDaemon.cs
+++ b/src/core/Akka.Cluster/ClusterDaemon.cs
@@ -1801,14 +1801,17 @@ namespace Akka.Cluster
 
             if (remoteGossip.Equals(Gossip.Empty))
             {
-                _log.Debug("Cluster Node [{0}] - Ignoring received gossip from [{1}] to protect against overload",
+                _log.Debug("Cluster Node [{0}] - Ignoring empty received gossip from [{1}] to protect against overload",
                     _cluster.SelfAddress, from);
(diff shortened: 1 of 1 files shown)


src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1801,14 +1801,17 @@ public ReceiveGossipType ReceiveGossip(GossipEnvelope envelope)
...
                 return ReceiveGossipType.Ignored;
             }
             if (!envelope.To.Equals(SelfUniqueAddress))
             {
-                _cluster.LogInfo("Ignoring received gossip intended for someone else, from [{0}] to [{1}]",
-                    from.Address, envelope.To);
+                _cluster.LogInfo("Ignoring received gossip intended for someone else, from [{0}] to [{1}]. Our full address is [{2}]",
+                    from.Address, envelope.To, SelfUniqueAddress);
+                
+                // TODO: if the gossip is received for a version of ourselves with a different UID, do we issue a Down command?
```
Aaronontheweb, line 1813: Valid question - what we need to do here will be determined by what we see in the enhanced log messages, but I suspect in our cluster testing we're seeing some gossip for old incarnations of current nodes. Need to log the UID to be sure.

PR #7286, opened 2024-07-17
convert Failure to exception when using Ask<object>

### LINKED ISSUE #7254: Ask<object> will not throw exception when the response is Status.Failure
**Version Information**
Version of Akka.NET? 1.5.15

**Describe the bug**
Ask<object> will not throw exception when the response is Status.Failure

This is caused by the order of case statements in `FutureActorRef<T>.TellInternal`:

```
            switch (message)
            {
                case ISystemMessage msg:
                    handled = _result.TrySetException(new InvalidOperationException($"system message of type '{msg.GetType().Name}' is invalid for {nameof(FutureActorRef<T>)}"));
                    break;
                case T t:
                    handled = _result.TrySetResult(t);
                    break;
                case null:
                    handled = _result.TrySetResult(default);
                    break;
                case Status.Failure f:
                    handled = _result.TrySetException(f.Cause
                        ?? new TaskCanceledException("Task cancelled by actor via Failure message."));
                    break;
```

when T is object and message is Status.Failure the `case T t:` is executed instead of ` case Status.Failure f:`.

ondravondra:
close #7254


Aaronontheweb:
Thanks for the PR - looks like this broke some tests

ondravondra:
It seems that the framework relies on this behavior.
For example, test `Inbox_Receive_will_timeout_gracefully_if_timeout_is_already_expired` uses `Akka.Actor.Inbox.Receive` that calls `Ask<object>` and expects the result to be `Status.Failure`.
A bit of a pickle innit.

Aaronontheweb:
> It seems that the framework relies on this behavior. For example, test `Inbox_Receive_will_timeout_gracefully_if_timeout_is_already_expired` uses `Akka.Actor.Inbox.Receive` that calls `Ask<object>` and expects the result to be `Status.Failure`. A bit of a pickle innit.

Probably just poor test design - at least that was my gut feeling when I looked at this PR. I'll give it another look when I can

Arkatufus (approved):
LGTM

lucavice:
Hi here. 

I wanted to report that after updating to 1.5.45+, I noticed some parts of the project I am working on stopped working as intended. 
Turns out that some parts of our codebase relied on that behaviour that @ondravondra reported for some of those tests.
In my case, I found:
``` c#
 var orderQueryResponse = await DomainService<CheckoutDomainService>.Ask<object>(Mediator, getOrderQuery);
 if (orderQueryResponse is Status.Failure sf)
   {
      // does something if you got back a failure
```
This code was probably bad design in the first place, but I wanted to let you know just in case other people may have used the same pattern and they are experiencing a breaking change after updating. I actually like what the PR is introducing and I think it makes sense.

Aaronontheweb:
Thanks for weighing in @lucavice - I agree with all of this. Changing the behavior was probably the right call so the PR was a good idea, but yeah it's going to cause breaking behavioral changes for components that relied on the old sub-optimal behavior. Thanks for documenting this in case other users get bit by it too.

PR #7287, opened 2024-07-18
Revert singleton app version

Aaronontheweb:
## Changes

Correct solution to fixing https://github.com/akkadotnet/akka.net/issues/7196 and replaces https://github.com/akkadotnet/akka.net/issues/7197

Definitely a breaking change but the current functionality is both broken and dangerous.

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [ ] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [ ] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [ ] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [x] Design discussion issue #7196 and #7197
* [ ] Changes in public API reviewed, if any.

PR #7297, opened 2024-07-23
Akka.Cluster.Tools.Singleton / Akka.Cluster.Sharding: fix duplicate shards caused by incorrect `ClusterSingletonManager` `HandOver`

### LINKED ISSUE #6973: Akka.Cluster.Sharding: duplicate shards / entities
**Version Information**
Version of Akka.NET? v1.5 - all versions including v1.5.13

**Describe the bug**

We have very rough and loose data on this right now, but it's been reported by multiple users including on https://github.com/petabridge/Akka.Persistence.Azure/issues/350 - looks like there could be something wrong with Akka.Cluster.Sharding in v1.5 that allows a `Shard` to be allocated more than once. This is our thread to investigate.

## If you have run into this issue, provide a full dump of your config here please

### LINKED ISSUE #7196: Akka.Cluster.Tools.Singleton: singleton moves earlier than expected - as soon as new node joins

Aaronontheweb:
## Changes

Eliminates the source of #6793, which was caused by using the incorrect ordering methodology when it came to determining which `ClusterSingletonManager` to hand-over to during member state transitions.

close #6973
close #7196

Supersedes https://github.com/akkadotnet/akka.net/pull/7287 and https://github.com/akkadotnet/akka.net/pull/7197

## Checklist

* [x] Design discussion issue #6973 and #7197


src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
```diff
@@ -99,9 +99,7 @@ public ClusterSingletonProxy(string singletonManagerPath, ClusterSingletonProxyS
             _singletonPath = (singletonManagerPath + "/" + settings.SingletonName).Split('/');
             _identityId = CreateIdentifyId(_identityCounter);
 
-            _memberAgeComparer = settings.ConsiderAppVersion
-                ? MemberAgeOrdering.DescendingWithAppVersion
-                : MemberAgeOrdering.Descending;
+            _memberAgeComparer = Member.AgeOrdering;
```
Aaronontheweb, line 102: This is the bug fix - descending order of members by age means we always get THE YOUNGEST at the front of the `membersByAge` list. Members MUST BE IN ASCENDING ORDER, OLDEST FIRST. I'm not sure how far back this bug goes - at least to v1.5. I'll add another comment after I've dug that up.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/MemberAgeOrdering.cs
Aaronontheweb, line 1: No longer needed, sticking with DRY especially since `Member.AgeOrdering` is already thoroughly sanity checked via https://github.com/akkadotnet/akka.net/pull/7291

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBuffer.cs
```diff
@@ -103,9 +103,7 @@ public OldestChanged(UniqueAddress oldest)
         public OldestChangedBuffer(string role, bool considerAppVersion)
         {
             _role = role;
-            _memberAgeComparer = considerAppVersion
-                ? MemberAgeOrdering.DescendingWithAppVersion
-                : MemberAgeOrdering.Descending;
+            _memberAgeComparer = Member.AgeOrdering;
```
Aaronontheweb, line 106: Uses the same fix as the `SingletonProxy`

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBuffer.cs
```diff
@@ -221,21 +218,33 @@ protected override void PostStop()
...
             {
-                Remove(_cluster.ReadView.Self);
-                Sender.Tell(Done.Instance); // reply to ask
-            }
-            else if (message is GetNext && _changes.IsEmpty) Context.BecomeStacked(OnDeliverNext);
-            else if (message is GetNext) SendFirstChange();
-            else
-            {
-                Unhandled(message);
+                case ClusterEvent.CurrentClusterState state:
```
Aaronontheweb, line 223: ReSharper'd this to use a `switch` statement. No biggie.

src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonLeaseSpec.cs
```diff
@@ -71,11 +71,6 @@ public TestException(string message, Exception innerEx)
                 : base(message, innerEx)
             {
             }
-
-            protected TestException(SerializationInfo info, StreamingContext context)
```
Aaronontheweb, line 75: Obsolete, deleted it to remove a build warning.

Aaronontheweb (commented):
This resolves one of the nastiest bugs we've ever debugged - I knew it would be a stupid problem like this, but tracking it down was a herculean effort. Hundreds of hours spent between @Arkatufus and myself.

Aaronontheweb:
Going to put this into our test lab with a private build and verify the fix overnight.

Aaronontheweb:
Looked at the history - apparently this has ALWAYS been a bug with Akka.Cluster.Tools, going all the way back to when the Cluster Singeton was first introduced https://github.com/akkadotnet/akka.net/pull/1530

Aaronontheweb:
We'll see if tests and the MNTR need to be adjusted, but our tracing system found definitive proof that the hand off was being screwed up, which lead to two `DDataShardCoordinator`s running at the same time (hence #6973). Going to fire this up in the test lab and see if we can still reproduce that after this.

Arkatufus (approved):
LGTM

PR #7298, opened 2024-07-24
Akka.Cluster.Tools: fix mutability and oldest state bugs with `ClusterSingletonManager`

### LINKED ISSUE #6973: Akka.Cluster.Sharding: duplicate shards / entities

### LINKED ISSUE #7196: Akka.Cluster.Tools.Singleton: singleton moves earlier than expected - as soon as new node joins

Aaronontheweb:
## Changes

Looks like the list of available nodes was getting mangled - seems like a basic mutability issue inside the `ClusterSingletonManager`.

close https://github.com/akkadotnet/akka.net/issues/6973
close https://github.com/akkadotnet/akka.net/issues/7196

## Checklist

* [x] Design discussion issue #6973 and #7196

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestart2Spec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestart2Spec.cs
@@ -31,7 +31,7 @@ namespace Akka.Cluster.Tools.Tests.Singleton
                                                                                            akka.loglevel = INFO
                                                                                            akka.actor.provider = "cluster"
                                                                                            akka.cluster.roles = [singleton]
-                                                                                           #akka.cluster.auto-down-unreachable-after = 2s
+                                                                                           akka.cluster.split-brain-resolver.stable-after = 2s
                                                                                            akka.cluster.singleton.min-number-of-hand-over-retries = 5
                                                                                            akka.remote {
                                                                                              dot-netty.tcp {
@@ -120,7 +120,7 @@ namespace Akka.Cluster.Tools.Tests.Singleton
             await Join(_sys4, _sys3);
 
             // let it stabilize
-            //Task.Delay(TimeSpan.FromSeconds(5)).Wait();
+            await Task.Delay(TimeSpan.FromSeconds(5));
 
             await WithinAsync(TimeSpan.FromSeconds(10), () =>
             {
--- a/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
@@ -203,21 +203,22 @@ namespace Akka.Cluster.Tools.Singleton
     }
 
     /// <summary>
-    /// TBD
+    /// State for after we've successfully transitioned to oldest, so we're hosting
+    /// the singleton actor.
     /// </summary>
     [Serializable]
     internal sealed class OldestData : IClusterSingletonData
     {
         /// <summary>
-        /// TBD
+        /// The reference to the current singleton running on this node.
         /// </summary>
-        public IActorRef Singleton { get; }
-
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="singleton">TBD</param>
-        public OldestData(IActorRef singleton)
+        /// <remarks>
+        /// Cam be explicitly set to <c>null</c> when we are leaving the cluster
+        /// and the singleton has to be terminated.
+        /// </remarks>
+        public IActorRef? Singleton { get; }
+        
+        public OldestData(IActorRef? singleton)
         {
             Singleton = singleton;
         }
@@ -230,21 +231,24 @@ namespace Akka.Cluster.Tools.Singleton
     internal sealed class WasOldestData : IClusterSingletonData
     {
         /// <summary>
-        /// TBD
+        /// The reference to the singleton.
         /// </summary>
-        public IActorRef Singleton { get; }
+        /// <remarks>
+        /// Can be <c>null</c> in edge cases where the node became the oldest but was already shutting down.
+        /// Shouldn't happen very often, but it's not impossible.
+        /// </remarks>
+        public IActorRef? Singleton { get; }
 
         /// <summary>
-        /// TBD
+        /// The address of the new oldest node.
         /// </summary>
-        public UniqueAddress NewOldest { get; }
-
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="singleton">TBD</param>
-        /// <param name="newOldest">TBD</param>
-        public WasOldestData(IActorRef singleton, UniqueAddress newOldest)
+        /// <remarks>
+        /// Can be <c>null</c> if we don't know who the new oldest is - for instance, during a full cluster
+        /// shutdown (in which case, there won't be any hand-over.)
+        /// </remarks>
+        public UniqueAddress? NewOldest { get; }
+        
+        public WasOldestData(IActorRef? singleton, UniqueAddress? newOldest)
         {
             Singleton = singleton;
             NewOldest = newOldest;
@@ -281,7 +285,7 @@ namespace Akka.Cluster.Tools.Singleton
     }
 
     /// <summary>
-    /// TBD
+    /// For when we are transitioning to a stopping state.
     /// </summary>
     [Serializable]
     internal sealed class StoppingData : IClusterSingletonData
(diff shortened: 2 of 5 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests/Singleton/ClusterSingletonRestart2Spec.cs
```diff
@@ -31,7 +31,7 @@ public class ClusterSingletonRestart2Spec : AkkaSpec
                                                                                            akka.loglevel = INFO
                                                                                            akka.actor.provider = "cluster"
                                                                                            akka.cluster.roles = [singleton]
-                                                                                           #akka.cluster.auto-down-unreachable-after = 2s
+                                                                                           akka.cluster.split-brain-resolver.stable-after = 2s
```
Aaronontheweb, line 34: Bug fixes for this spec - after doing a thorough review of some the cases where we had a race condition, it all comes down to whether or not `_sys1` was able to fully get `MemberRemoved` or whether or not it became `Unreachable` first. Speeding up the downing provider's decision-making process both ensures that this member gets removed quickly AND ensures that the number of retries during the `HandOver` process get capped so the test can complete on-time.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
Aaronontheweb, line 1: I added nullability changes in here, cleaned up garbage XML-DOC comments, and did some minor reformatting - hence why this appears to be a "large" diff. However, I will comment on the real and true changes below.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -199,119 +173,117 @@ internal sealed class YoungerData : IClusterSingletonData
...
-        public List<UniqueAddress> PreviousOldest { get; }
-
-        /// <summary>
-        /// TBD
-        /// </summary>
-        /// <param name="previousOldest">TBD</param>
-        public BecomingOldestData(List<UniqueAddress> previousOldest)
+        public ImmutableList<UniqueAddress> PreviousOldest { get; }
+        
+        public BecomingOldestData(ImmutableList<UniqueAddress> previousOldest)
```
Aaronontheweb, line 199: Moved to using `Immutable` lists everywhere for all states and messages. These are only sent when movement is occurring in the cluster, which is inherently low-volume, so there's no performance concerns.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -769,15 +723,20 @@ private void GetNextOldestChanged()
...
+            _lease.Acquire(reason =>
             {
                 self.Tell(new LeaseLost(reason));
             }).ContinueWith(r =>
             {
                 if (r.IsFaulted || r.IsCanceled)
                     return (object)new AcquireLeaseFailure(r.Exception);
                 return new AcquireLeaseResult(r.Result);
-            }).PipeTo(Self);
+            }).PipeTo(self);
```
Aaronontheweb, line 739: Fixed a potential safety issue here with using the correct closure.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -769,15 +723,20 @@ private void GetNextOldestChanged()
         private State<ClusterSingletonState, IClusterSingletonData> TryAcquireLease()
         {
             var self = Self;
-            lease.Acquire(reason =>
+
+            if (_lease == null)
```
Aaronontheweb, line 727: Added a `null` check here to appease nullability checks

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -842,17 +799,16 @@ private void GetNextOldestChanged()
...
-            else if (handOverTo == null)
-            {
-                return GoTo(ClusterSingletonState.Younger).Using(new YoungerData(null));
-            }
-            else
+
+            if (handOverTo == null)
             {
-                return GoTo(ClusterSingletonState.End).Using(EndData.Instance);
+                return GoTo(ClusterSingletonState.Younger).Using(new YoungerData(ImmutableList<UniqueAddress>.Empty));
```
Aaronontheweb, line 805: No longer using null `List<T>` - using empty immutable lists instead.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -911,24 +867,38 @@ private void InitializeFSM()
                     case OldestChangedBuffer.OldestChanged oldestChanged when e.StateData is YoungerData youngerData:
                     {
                         _oldestChangedReceived = true;
-                        if (oldestChanged.Oldest != null && oldestChanged.Oldest.Equals(_selfUniqueAddress))
+                        if (oldestChanged.NewOldest != null && oldestChanged.NewOldest.Equals(_selfUniqueAddress))
```
Aaronontheweb, line 870: This and a corresponding change in the `OldestChangedBuffer` are the two big fixes in this PR, aside from dealing with some of the minor mutability issues.

## The Problem

When we receive a `MemberRemoved` for an older node that was previously hosting a cluster singleton, we intentionally delay processing of that `MemberRemoved` event via the `DelayedMemberRemoved` mechanism. This is designed to ensure that in the event that there is a split brain and the SBR was responsible for removing that old node, that node still gets a chance to shut itself down and die cleanly before we take over. This is especially important for singletons using Akka.Persistence, so we don't have two competing actors using the same persistence id and database. So, we still assume that node is the oldest up to ~20s after it's removal.

This is normally not an issue - when we attempt to send hand-overs to that node they'll time out eventually and we'll just recreate the singleton in a matter of 10-20 seconds typically.

However, where this design creates a problem is during rolling updates in a managed environment like Kubernetes, where restarts can happen really quickly.

Imagine this scenario:

* Cluster of `node1` (oldest), `node2`, `node3` - all have the right `role` to host the singleton.
* `node1` is replaced first by K8s - it's terminated via CLR process exit; uses `CoordinatedShutdown` to try to exit gracefully as best as it can.
* `node2` sees that `node1` is exiting and does the correct thing - messages it for a handover and becomes the new singleton host and is now the oldest node.
* Kubernetes restarts `node1`, which joins the cluster under a new `UniqueMemberAddress` but with the same `Address` as before.
* Kubernetes then takes down `node2` - per the update process.
* If all of the prior steps happen in under 20 seconds, `node3` will be notified that it is becoming the oldest node and it will start messaging `node1`, which is ACTUALLY YOUNGER THAN IT, to do the hand-over. `node1` will ACK that hand-over right away and we'll have two separate cluster singletons active at the same time, until `node2` completely exits. This issue is one of the two bugs that caused https://github.com/akkadotnet/akka.net/issues/6973 - the member ordering problem that we fixed on https://github.com/akkadotnet/akka.net/pull/7297 being the other.


So to fix this issue, we can either speed up member removals or do something even simpler - just include the address of the previous "oldest" node in the `OldestChanged` message and put that one at the front of the list in our `Younger` and `BecomingOldest` states. That way, we're always messaging the most recent home of the singleton using live data from the `OldestChangedBuffer`. I've run this experiment thousands of times in our test lab and can confirm that it eliminated the issue completely.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/OldestChangedBuffer.cs
```diff
@@ -62,31 +61,36 @@ public sealed class InitialOldestState
...
+        /// Message propagated once the previous oldest member is exiting / downed / removed.
         /// </summary>
         [Serializable]
         public sealed class OldestChanged
         {
             /// <summary>
-            /// TBD
+            /// The new "oldest" - this node will become the new singleton manager.
             /// </summary>
-            public UniqueAddress Oldest { get; }
```
Aaronontheweb, line 81: The big fix - make sure that the `OldestChanged` message includes BOTH the "new" oldest node and the previous oldest node. This ensures that we don't have any data consistency problems in the event of a rolling update.

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManager.cs
```diff
@@ -911,24 +867,38 @@ private void InitializeFSM()
...
                             if (youngerData.Oldest.All(m => _removed.ContainsKey(m)))
                             {
                                 return TryGotoOldest();
                             }
 
-                            Peer(youngerData.Oldest.Head().Address).Tell(HandOverToMe.Instance);
-                            return GoTo(ClusterSingletonState.BecomingOldest).Using(new BecomingOldestData(youngerData.Oldest));
+                            // explicitly re-order the list to make sure that the oldest, as indicated to us by the OldestChangedBuffer,
+                            //  is the first element - resolves bug https://github.com/akkadotnet/akka.net/issues/6973
+                            var newOldestState = oldestChanged.PreviousOldest switch
```
Aaronontheweb, line 880: Re-arrange the state we're going to retain so we make sure the `PreviousOldest` node, as reported by the `OldestChangedBuffer`, appears at the front of the list. This ensures that the `BecomingOldest` node is always performing the hand-off with the correct party in the event of a rolling update.

Aaronontheweb (commented):
Detailed my changes

Arkatufus (approved):
LGTM

PR #7300, opened 2024-07-24
Fix `ActorMaterializerImpl` `null` `LogSource`

Arkatufus:
## Changes

The ActorMaterializer.MakeLogger() method accesses a variable that was not initialized and most possibly be null


Aaronontheweb (approved):
LGTM

PR #7301, opened 2024-07-24
Add AlsoTo downstream failure propagation support

### LINKED ISSUE #7269: Akka Stream AlsoTo may not be failing graph when its sink throws exception
**Version Information**
Version of Akka.NET?  1.5.25
Which Akka.NET Modules? Akka.Streams 1.5.25

**Describe the bug**
A simple graph with source, sink and an intermediate AlsoTo stage with its own sink, if the sink referenced by the AlsoTo flow throws an exception the rest of the graph completes instead of failing the complete graph

**To Reproduce**

The following console app I think demonstrates the issue

```
// See https://aka.ms/new-console-template for more information
using Akka.Actor;
using Akka.Streams;
using Akka.Streams.Dsl;

Console.WriteLine("Starting Akka Stream");

var system = ActorSystem.Create("MySystem");
var materializer = system.Materializer();

var source = Source.From(Enumerable.Range(1, 100));

var sink = Sink.ForEach<int>(i => Console.WriteLine($"Main sink {i}"));

var alsoSink = Sink.ForEach<int>(i => throw new Exception("sink failed") );


var alsoFlow = Flow.Create<int>()
                .WireTap(i => Console.WriteLine($"also wiretap: {i}"))
                .To(alsoSink);

var task = source
           .AlsoTo(alsoFlow)           
           .ToMaterialized(sink, Keep.Right).Run(materializer);

await task;


```

_Links to working reproductions on Github / Gitlab are very much appreciated_

**Expected behavior**

When the AlsoTo flow fails the complete graph should fail

**Actual behavior**
No further output is seen from the AlsoTo stage but the other sink continues to receive elements

**Environment**

Windows 11, Dotnet 8

**Additional context**

I first noticed this when working with alpakka where the AlsoTo stream sink was an Azure Service bus queue, the main flow reading from an Azure Event Hub and it continued to emit elements even after the service bus sink threw an exception

Note no exception is returned by the failed AlsoTo stage in my Azure example the actor starting the stream has the task piped back to itself - and no fail message is received

Arkatufus:
Fixes #7269

## Changes

Add some new API for `AlsoTo` that allows users to opt-in to propagate downstream failures to its parent stream. If any flow/stage failed downstream of the `AlsoTo` flow, it will also cancel the parent stream with the same cause.

The default behavior will still ignore any cancellation/failures from downstream stages/flows

### DIFF
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveStreams.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveStreams.DotNet.verified.txt
@@ -2059,6 +2059,7 @@ namespace Akka.Streams.Dsl
         public static Akka.Streams.Dsl.Source<TOut2, TMat> Aggregate<TOut1, TOut2, TMat>(this Akka.Streams.Dsl.Source<TOut1, TMat> flow, TOut2 zero, System.Func<TOut2, TOut1, TOut2> fold) { }
         public static Akka.Streams.Dsl.Source<TOut2, TMat> AggregateAsync<TOut1, TOut2, TMat>(this Akka.Streams.Dsl.Source<TOut1, TMat> flow, TOut2 zero, System.Func<TOut2, TOut1, System.Threading.Tasks.Task<TOut2>> fold) { }
         public static Akka.Streams.Dsl.Source<TOut, TMat> AlsoTo<TOut, TMat>(this Akka.Streams.Dsl.Source<TOut, TMat> flow, Akka.Streams.IGraph<Akka.Streams.SinkShape<TOut>, TMat> that) { }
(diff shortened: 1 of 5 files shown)


Aaronontheweb (changes requested):
I would probably make this enabled by default - since that's what I'd expect if I were an end-user.

Arkatufus:
~~That would be a breaking change for existing implementation~~

~~Talked to Aaron about this offline, we decide that the correct behavior should be the default, even when it is a breaking change.~~

After some deliberation, the old behavior actually have a purpose, we're keeping the old behavior.

Arkatufus:
Done, failure propagation is the default behavior now

Aaronontheweb (approved):

PR #7302, opened 2024-07-25
Akka.Cluster.Tools: deprecate ClustersSingletonManagerSettings.ConsiderAppVersion

Aaronontheweb:
## Changes

Cleanup from PR https://github.com/akkadotnet/akka.net/pull/7298

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [x] Design discussion issue #7197
* [x] Changes in public API reviewed, if any.

PR #7292, opened 2024-07-22
DData: if `lmdb.dir` is null or empty, log a warning and set to default

Aaronontheweb:
## Changes

DData: if `lmdb.dir` is null or empty, log a warning and set to default. Should address some `NullReferenceException` issues that occured when the following HOCON was set:

```hocon
akka.cluster.distributed-data.durable.lmdb.dir = "shard-data"
```

Reported `Exception`:

```
[15:53:33 ERR] Stopping distributed-data Replicator due to load or startup failure in durable store, caused by: Exception during creation
[akka://cluster-system/system/sharding/replicator/durableStore#713307088]: Akka.Actor.ActorInitializationException: Exception during creation
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Akka.DistributedData.LightningDB.LmdbDurableStore.PreStart()
   at Akka.Actor.ActorBase.AroundPreStart()
   at Akka.Actor.ActorCell.<>cDisplayClass175_0.<Create>b0()
   at Akka.Actor.ActorCell.UseThreadContext(Action action)
   at Akka.Actor.ActorCell.Create(Exception failure)
   --- End of inner exception stack trace -
```

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).

PR #7281, opened 2024-07-10
Akka.Cluster: improve gossip serialization performance

Aaronontheweb:
## Changes

Performance golf to attempt to improve serialization performance on `Gossip` data structures

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).

### Latest `dev` Benchmarks 

```

BenchmarkDotNet v0.13.12, Windows 10 (10.0.19045.4529/22H2/2022Update)
AMD Ryzen 7 1700, 1 CPU, 16 logical and 8 physical cores
.NET SDK 8.0.303
  [Host]     : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.7 (8.0.724.31311), X64 RyuJIT AVX2


```
| Method                     | Mean        | Error       | StdDev      | Gen0   | Allocated |
|--------------------------- |------------:|------------:|------------:|-------:|----------:|
| Serialize_Heartbeat        |    392.8 ns |     7.83 ns |    21.84 ns | 0.0534 |     224 B |
| Deserialize_Heartbeat      |    914.4 ns |    18.04 ns |    40.34 ns | 0.1640 |     688 B |
| Serialize_HeartbeatRsp     |    492.4 ns |     9.44 ns |    24.69 ns | 0.0629 |     264 B |
| Deserialize_HeartbeatRsp   |  1,149.8 ns |    22.99 ns |    55.51 ns | 0.1907 |     800 B |
| Serialize_GossipEnvelope   | 27,185.0 ns |   675.36 ns | 1,980.72 ns | 1.7700 |    7496 B |
| Deserialize_GossipEnvelope | 51,976.9 ns | 1,186.86 ns | 3,499.47 ns | 3.9673 |   16624 B |
| Serialize_GossipStatus     |  3,124.2 ns |    71.97 ns |   211.07 ns | 0.3929 |    1648 B |
| Deserialize_GossipStatus   |  8,466.2 ns |   196.12 ns |   559.54 ns | 0.9766 |    4096 B |
[...]

PR #7271, opened 2024-06-28
Update ConfigServiceDiscovery to support multi-config

Arkatufus:
## Changes

* Make ConfigServiceDiscovery constructor compatible with the new multi-config feature

PR #7303, opened 2024-07-25
V1.5.27 release notes

Aaronontheweb:
#### 1.5.27 July 25th 2024 ####

Akka.NET v1.5.27 is a significant release that includes new features, mission-critical bug fixes, and some performance improvements.

**Major Akka.Cluster.Sharding and Akka.Cluster.Tools.Singleton Bug Fixes**

In _all prior versions_ of Akka.NET, there are two high impact distributed systems bugs:

1. [Akka.Cluster.Tools.Singleton: singleton moves earlier than expected - as soon as new node joins](https://github.com/akkadotnet/akka.net/issues/7196)
2. [Akka.Cluster.Sharding: duplicate shards / entities](https://github.com/akkadotnet/akka.net/issues/6973)

As we discovered during the course of our pains-taking bug investigation, these were, in fact, the same issue:

1. The `ClusterSingletonManager` is supposed to _always_ belong on the oldest node of a given role type, but an original design error from the time Akka.Cluster.Tools was first introduced to Akka.NET meant that nodes were always sorted in _descending_ order of `UpNumber`. This is backwards: nodes should always be sorted in _ascending_ order of `UpNumber` - this means that the oldest possible node is always at the front of the "who is oldest?" list held by the `ClusterSingletonManager`. This explains why the singleton could appear to move early during deployments and restarts.
2. The `ClusterSingletonManager` was suspectible to a race condition where if nodes were shutdown and restarted with the same address in under 20 seconds, the default "down removal margin" used by the `ClusterSingletonManager` to tolerate dirty exits, it would be possible after _multiple_ successive, fast, restarts for multiple instances of the singleton to be alive at the same time (for a short period.)

Both of these varieties of problem, duplicate singletons, is what lead to duplicate shards.

As a result we've made the following fixes:

* [Akka.Cluster.Tools: deprecate ClustersSingletonManagerSettings.ConsiderAppVersion](https://github.com/akkadotnet/akka.net/pull/7302) - `AppVersion` is no longer considered for singleton placement as it could easily result in split brains.
* [Akka.Cluster.Tools: fix mutability and oldest state bugs with `ClusterSingletonManager`](https://github.com/akkadotnet/akka.net/pull/7298) - resolves the issue with rapid rolling restarts creating duplicates. We've tested this fix in our test lab across thousands of coordinator restarts and haven't been able to reproduce the issue since (we could easily do it before.)
* [Akka.Cluster.Tools.Singleton / Akka.Cluster.Sharding: fix duplicate shards caused by incorrect `ClusterSingletonManager` `HandOver`](https://github.com/akkadotnet/akka.net/pull/7297) - we fixed the member age problem here, which could cause a second singleton to start at inappropriate times.

**Akka.Discovery and `ClusterClient` Discovery Support**

In Akka.NET v1.5.27 we've added support for using Akka.Cluster.Tools.ClusterClient alongside with [Akka.Discovery plugins](https://getakka.net/articles/discovery/index.html) to automatically discover the initial contacts you need for `ClusterClientReceptionist` instances in your environment.

You can read the documentation for how this works here: https://getakka.net/articles/clustering/cluster-client.html#contact-auto-discovery-using-akkadiscovery

Related PRs and issues:

* [Akka.Discovery: Add multi-config support to config-based discovery](https://github.com/akkadotnet/akka.net/issues/7271)
* [Cluster.Tools: Fix missing VerboseLogging in ClusterClientSettings.Copy method](https://github.com/akkadotnet/akka.net/issues/7272)
* [Cluster.Tools: Improve ClusterClientDiscovery to avoid thundering herd problem](https://github.com/akkadotnet/akka.net/issues/7270)
* [Cluster.Tools: Change ClusterClientDiscovery to use the new Akka.Management "/cluster-client/receptionist" endpoint](https://github.com/akkadotnet/akka.net/issues/7274)

**Other Bug Fixes and Improvements**

* [Akka.Cluster: improve gossip serialization performance](https://github.com/akkadotnet/akka.net/pull/7281)
* [Akka.Streams: Fix `ActorMaterializerImpl` `null` `LogSource`](https://github.com/akkadotnet/akka.net/pull/7300)
* [Akka.Streams: `AlsoTo` may not be failing graph when its sink throws exception](https://github.com/akkadotnet/akka.net/issues/7269)
* [Akka.DistributedData: if `lmdb.dir` is null or empty, log a warning and set to default](https://github.com/akkadotnet/akka.net/pull/7292)

To [see the full set of changes in Akka.NET v1.5.27, click here](https://github.com/akkadotnet/akka.net/milestone/109).

| --- | --- | --- | --- |
| 15 | 835 | 1001 | Aaron Stannard |
| 12 | 1123 | 207 | Gregorius Soedharmo |

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,50 @@
+#### 1.5.27 July 25th 2024 ####
+
+Akka.NET v1.5.27 is a significant release that includes new features, mission-critical bug fixes, and some performance improvements.
+
+**Major Akka.Cluster.Sharding and Akka.Cluster.Tools.Singleton Bug Fixes**
+
+In _all prior versions_ of Akka.NET, there are two high impact distributed systems bugs:
+
+1. [Akka.Cluster.Tools.Singleton: singleton moves earlier than expected - as soon as new node joins](https://github.com/akkadotnet/akka.net/issues/7196)
+2. [Akka.Cluster.Sharding: duplicate shards / entities](https://github.com/akkadotnet/akka.net/issues/6973)
+
+As we discovered during the course of our pains-taking bug investigation, these were, in fact, the same issue:
+
+1. The `ClusterSingletonManager` is supposed to _always_ belong on the oldest node of a given role type, but an original design error from the time Akka.Cluster.Tools was first introduced to Akka.NET meant that nodes were always sorted in _descending_ order of `UpNumber`. This is backwards: nodes should always be sorted in _ascending_ order of `UpNumber` - this means that the oldest possible node is always at the front of the "who is oldest?" list held by the `ClusterSingletonManager`. This explains why the singleton could appear to move early during deployments and restarts.
+2. The `ClusterSingletonManager` was suspectible to a race condition where if nodes were shutdown and restarted with the same address in under 20 seconds, the default "down removal margin" used by the `ClusterSingletonManager` to tolerate dirty exits, it would be possible after _multiple_ successive, fast, restarts for multiple instances of the singleton to be alive at the same time (for a short period.)
+
+Both of these varieties of problem, duplicate singletons, is what lead to duplicate shards.
+
+As a result we've made the following fixes:
+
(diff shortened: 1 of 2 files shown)


Arkatufus (approved):
LGTM

PR #7313, opened 2024-08-06
Akka.Persistence: Made `DateTime.UtcNow` the default timestamp for `SnapshotMetdata`

### LINKED ISSUE #7312: Akka.Persistence: `SnapshotMetadata` always uses `DateTime.MinValue`, which creates "duplicate key" exceptions when multiple snapshots are taken at same `seqNr`
**Version Information**
Version of Akka.NET? v1.5.27.1 and all prior versions
Which Akka.NET Modules? Akka.Persistence has the bug, but it's only the SQL-based implementations that have this problem

**Describe the bug**

A bug report from an end-user contained the following exception when using [Akka.Persistence.Sql](https://github.com/akkadotnet/Akka.Persistence.Sql):

```
      [WARNING][07/27/2024 07:11:47.184Z][Thread 0042][akka://Journey/user/journey] Failed to SaveSnapshot given metadata [SnapshotMetadata<pid: JourneyAggregateActorRouter, seqNr: 0, timestamp: 0001/01/01>] due to: [Akka.Pattern.OpenCircuitException: Circuit Breaker is open; calls are failing fast: Circuit Breaker is open; calls are failing fast]
fail: Akka.Actor.ActorSystem[0]
      [ERROR][07/27/2024 07:11:47.185Z][Thread 0037][akka://Journey/user/journey/10935440] Persistence failure when replaying events for persistenceId [10935440]. Last known sequence number [0]
```

The real kicker here is the timestamp value - `0001/01/01`. This is `DateTime.MinValue`, which is used as the default timestamp on all `SnapshotMetadata` instances used by Akka.Persistence's actor base classes:

https://github.com/akkadotnet/akka.net/blob/acc8bd3894bc9cf74778cf8051dee9e55e990e90/src/core/Akka.Persistence/Eventsourced.cs#L219-L222

And that CTOR always uses `DateTime.MinValue` for the snapshot timestamp:

https://github.com/akkadotnet/akka.net/blob/acc8bd3894bc9cf74778cf8051dee9e55e990e90/src/core/Akka.Persistence/SnapshotProtocol.cs#L62-L72

**Expected behavior**

It is perfectly valid and legal, according to the Akka.Persistence specification, to have multiple snapshots taken at the same `PersistentId` and same `seqNr` so long as they are separated by Timestamp. The combination of these three values is what is supposed to form the "uniqueness constraint" for a snapshot, hence why Timestamps are included in things like `SnapshotSelection` queries.

However, there is a secondary bug in the case of SQL-related plugins: the primary keys for Snapshots themselves are also wrong:

```sql
USE [Akka]
GO

SET ANSI_PADDING ON
GO

/****** Object:  Index [PK_SnapshotStore]    Script Date: 8/6/2024 10:01:04 AM ******/
ALTER TABLE [dbo].[SnapshotStore] ADD  CONSTRAINT [PK_SnapshotStore] PRIMARY KEY CLUSTERED 
(
	[PersistenceId] ASC,
	[SequenceNr] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, SORT_IN_TEMPDB = OFF, IGNORE_DUP_KEY = OFF, ONLINE = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
GO
```

I'll file a second issue for that.

**Actual behavior**

Saving two snapshots at the same `seqNr` fails.

**Environment**

Running on SQL Server, but I'm sure it's the same issue for the other SQL platforms we support as well.

**Additional context**

This bug has been around for ~10 years, so I'm sure it doesn't happen very often.

If you're impacted by this bug, I recommend the following workaround:

* Each time you want to save a snapshot, call `Persist` on a useless data type you don't care about (i.e. an integer)
* Add a `Recover<int>(i => { });` to just discard those values during recovery
* Inside the `Persist` callback function, THEN call `SaveSnapshot` - this will force the `seqNr` to implement even though you're not actually using it.

Aaronontheweb:
## Changes

Fixes #7312

## Checklist

* [x] Design discussion issue #7312

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
+++ b/src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
@@ -463,7 +463,7 @@ namespace Akka.Persistence.Sql.Common.Snapshot
             await connection.ExecuteInTransaction(WriteIsolationLevel, cancellationToken, async (tx, token) =>
             {
                 var sql = timestamp.HasValue
-                    ? DeleteSnapshotRangeSql + " AND { Configuration.TimestampColumnName} = @Timestamp"
+                    ? DeleteSnapshotSql + $" AND {Configuration.TimestampColumnName} <= @Timestamp"
                     : DeleteSnapshotSql;
 
                 using var command = GetCommand(connection, sql); 
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.DotNet.verified.txt
@@ -588,7 +588,7 @@ namespace Akka.Persistence
     public sealed class SnapshotMetadata : System.IEquatable<Akka.Persistence.SnapshotMetadata>
     {
         [System.ObsoleteAttribute("This constructor is deprecated and will be removed in v1.6. Use the constructor w" +
-            "ith the timestamp parameter instead.", true)]
+            "ith the timestamp parameter instead. Since v1.5.28", true)]
         public SnapshotMetadata(string persistenceId, long sequenceNr) { }
         [Newtonsoft.Json.JsonConstructorAttribute()]
         public SnapshotMetadata(string persistenceId, long sequenceNr, System.DateTime timestamp) { }
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.DotNet.verified.txt
@@ -587,7 +587,8 @@ namespace Akka.Persistence
     }
     public sealed class SnapshotMetadata : System.IEquatable<Akka.Persistence.SnapshotMetadata>
     {
-        public static System.DateTime TimestampNotSpecified;
+        [System.ObsoleteAttribute("This constructor is deprecated and will be removed in v1.6. Use the constructor w" +
+            "ith the timestamp parameter instead.", true)]
         public SnapshotMetadata(string persistenceId, long sequenceNr) { }
         [Newtonsoft.Json.JsonConstructorAttribute()]
         public SnapshotMetadata(string persistenceId, long sequenceNr, System.DateTime timestamp) { }
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.Net.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.Net.verified.txt
@@ -588,7 +588,7 @@ namespace Akka.Persistence
     public sealed class SnapshotMetadata : System.IEquatable<Akka.Persistence.SnapshotMetadata>
     {
         [System.ObsoleteAttribute("This constructor is deprecated and will be removed in v1.6. Use the constructor w" +
-            "ith the timestamp parameter instead.", true)]
+            "ith the timestamp parameter instead. Since v1.5.28", true)]
         public SnapshotMetadata(string persistenceId, long sequenceNr) { }
         [Newtonsoft.Json.JsonConstructorAttribute()]
         public SnapshotMetadata(string persistenceId, long sequenceNr, System.DateTime timestamp) { }
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.Net.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.Net.verified.txt
@@ -587,7 +587,8 @@ namespace Akka.Persistence
     }
     public sealed class SnapshotMetadata : System.IEquatable<Akka.Persistence.SnapshotMetadata>
     {
-        public static System.DateTime TimestampNotSpecified;
+        [System.ObsoleteAttribute("This constructor is deprecated and will be removed in v1.6. Use the constructor w" +
+            "ith the timestamp parameter instead.", true)]
         public SnapshotMetadata(string persistenceId, long sequenceNr) { }
         [Newtonsoft.Json.JsonConstructorAttribute()]
         public SnapshotMetadata(string persistenceId, long sequenceNr, System.DateTime timestamp) { }
--- a/src/core/Akka.Persistence.TCK/Query/PersistenceIdsSpec.cs
+++ b/src/core/Akka.Persistence.TCK/Query/PersistenceIdsSpec.cs
@@ -247,7 +247,7 @@ namespace Akka.Persistence.TCK.Query
                 ExpectMsg($"{persistenceId}-{i}-done");
             }
 
-            var metadata = new SnapshotMetadata(persistenceId, n + 10);
+            var metadata = new SnapshotMetadata(persistenceId, n + 10, Sys.Scheduler.Now.DateTime);
             SnapshotStore.Tell(new SaveSnapshot(metadata, $"s-{n}"), _senderProbe.Ref);
             _senderProbe.ExpectMsg<SaveSnapshotSuccess>();
 
--- a/src/core/Akka.Persistence.TCK/Snapshot/SnapshotStoreSpec.cs
+++ b/src/core/Akka.Persistence.TCK/Snapshot/SnapshotStoreSpec.cs
@@ -100,7 +100,7 @@ namespace Akka.Persistence.TCK.Snapshot
         {
             for (int i = 1; i <= 5; i++)
             {
-                var metadata = new SnapshotMetadata(Pid, i + 10);
+                var metadata = new SnapshotMetadata(Pid, i + 10, Sys.Scheduler.Now.DateTime);
                 SnapshotStore.Tell(new SaveSnapshot(metadata, $"s-{i}"), _senderProbe.Ref);
                 yield return _senderProbe.ExpectMsg<SaveSnapshotSuccess>().Metadata;
             }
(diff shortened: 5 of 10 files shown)


Aaronontheweb:
Not done with this PR by a long shot - looking at doing some potential fixes around the `ITimeProvider` implementation we have first in order to ensure that we ACTUALLY support dependency injection with it.

src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.DotNet.verified.txt
```diff
@@ -587,7 +587,8 @@ namespace Akka.Persistence
     }
     public sealed class SnapshotMetadata : System.IEquatable<Akka.Persistence.SnapshotMetadata>
     {
-        public static System.DateTime TimestampNotSpecified;
+        [System.ObsoleteAttribute("This constructor is deprecated and will be removed in v1.6. Use the constructor w" +
```
Aaronontheweb, line 590: Just marked the old CTOR as `Obsolete` - no other API changes.

src/core/Akka.Persistence.TCK/Query/PersistenceIdsSpec.cs
```diff
@@ -247,7 +247,7 @@ protected IActorRef WriteSnapshot(string persistenceId, int n)
                 ExpectMsg($"{persistenceId}-{i}-done");
             }
 
-            var metadata = new SnapshotMetadata(persistenceId, n + 10);
+            var metadata = new SnapshotMetadata(persistenceId, n + 10, Sys.Scheduler.Now.DateTime);
```
Aaronontheweb, line 250: Standard call we're going to use to get the current UTC time from the `ITimeProvider` built into the `IScheduler`.

src/core/Akka.Persistence/Eventsourced.cs
```diff
@@ -218,7 +218,7 @@ public void LoadSnapshot(string persistenceId, SnapshotSelectionCriteria criteri
         /// <param name="snapshot">TBD</param>
         public void SaveSnapshot(object snapshot)
         {
-            SnapshotStore.Tell(new SaveSnapshot(new SnapshotMetadata(SnapshotterId, SnapshotSequenceNr), snapshot));
+            SnapshotStore.Tell(new SaveSnapshot(new SnapshotMetadata(SnapshotterId, SnapshotSequenceNr, Context.System.Scheduler.Now.Date), snapshot));
```
Aaronontheweb, line 221: Using the same standard call now for both saving and deleting snapshots - big question I have though: does this break deleting snapshots? It SHOULDN'T, but we'll need to see how the specs perform to know for certain.

Aaronontheweb (commented):
Detailed my changes

src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.Net.verified.txt
```diff
@@ -587,7 +587,8 @@ namespace Akka.Persistence
     }
     public sealed class SnapshotMetadata : System.IEquatable<Akka.Persistence.SnapshotMetadata>
     {
-        public static System.DateTime TimestampNotSpecified;
+        [System.ObsoleteAttribute("This constructor is deprecated and will be removed in v1.6. Use the constructor w" +
```
Arkatufus, line 590: A little nitpick, Would be better if we follow english language sentence breaking point here, makes it easier to read.

src/core/Akka.Persistence/SnapshotProtocol.cs
```diff
@@ -52,32 +52,28 @@ public int Compare(SnapshotMetadata x, SnapshotMetadata y)
...
-
         /// <summary>
         /// Initializes a new instance of the <see cref="SnapshotMetadata"/> class.
         /// </summary>
         /// <param name="persistenceId">The id of the persistent actor fro which the snapshot was taken.</param>
         /// <param name="sequenceNr">The sequence number at which the snapshot was taken.</param>
+        [Obsolete("This constructor is deprecated and will be removed in v1.6. Use the constructor with the timestamp parameter instead.", true)]
         public SnapshotMetadata(string persistenceId, long sequenceNr)
-            : this(persistenceId, sequenceNr, TimestampNotSpecified)
+            : this(persistenceId, sequenceNr, DateTime.UtcNow)
```
Arkatufus: The only objection I have for this is that this breaks previous behaviour.
The way this works previously, if `DateTime.MinValue` were passed as the timestamp, it will be interpreted as a `null` value by the `SqlSnapshotStore`, and if its a null, it will not generate a timestamp SQL query statement inside the `QueryExecutor`

If we're passing `DateTime.UtcNow` here, we're basically changing the `QueryExecutor` SQL query from `DeleteSnapshotSql` to `DeleteSnapshotRangeSql + " AND { Configuration.TimestampColumnName} = @Timestamp"`

Arkatufus (changes requested):
After thinking over about it, I think the change is dangerous

src/core/Akka.Persistence/SnapshotProtocol.cs
Aaronontheweb: That's the point of this change though - we need to be passing a real timestamp value. I deprecated this CTOR and we aren't using it anymore - we're using the timestamp provided by the `ITimeProvider` in the 3-parameter CTOR.

src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.Net.verified.txt
Aaronontheweb: Good point - I can fix that.

Aaronontheweb: Ah, no I can't - that's an artifact from the rendering of the API diff. The real code looks like this:

```csharp
[Obsolete("This constructor is deprecated and will be removed in v1.6. Use the constructor with the timestamp parameter instead.", true)]
```

src/core/Akka.Persistence.TCK/Snapshot/SnapshotStoreSpec.cs
```diff
@@ -198,6 +201,54 @@ public virtual void SnapshotStore_should_delete_a_single_snapshot_identified_by_
...
+                result.ToSequenceNr == long.MaxValue
+                && result.Snapshot != null
+                && result.Snapshot.Metadata.Equals(Metadata[1])
+                && result.Snapshot.Snapshot.ToString() == "s-2");
+        }
+        
+        // Issue #7312, this is a side effect of the ctor signature changes
+        // DeleteSnapshot should not delete snapshot if timestamp value does not meet deletion criteria
+        [Fact]
+        public virtual void SnapshotStore_should_not_delete_snapshot_identified_by_SequenceNr_if_metadata_timestamp_is_less_than_stored_timestamp()
```
Arkatufus, line 231: This is a side effect of the changes made in this PR.
User need to either have a `SnapshotMetadata.Timestamp` that is greater than or equals to the stored value, or equals to `DateTime.MinValue` for the actual `Snapshot` to be deleted.

Arkatufus (commented):
Self review

Arkatufus (approved):

src/core/Akka.Persistence.TCK/Snapshot/SnapshotStoreSpec.cs
Arkatufus: We either have this as an actual spec, or change the code to disregard `SnapshotMetadata.Timestamp` altogether when deleting a `Snapshot`.

src/contrib/persistence/Akka.Persistence.Sql.Common/Snapshot/QueryExecutor.cs
```diff
@@ -463,7 +463,7 @@ protected virtual void SetManifestParameters(object snapshot, DbCommand command)
             await connection.ExecuteInTransaction(WriteIsolationLevel, cancellationToken, async (tx, token) =>
             {
                 var sql = timestamp.HasValue
-                    ? DeleteSnapshotRangeSql + " AND { Configuration.TimestampColumnName} = @Timestamp"
+                    ? DeleteSnapshotSql + $" AND {Configuration.TimestampColumnName} <= @Timestamp"
```
Aaronontheweb, line 466: LGTM - bugfix

src/core/Akka.Persistence.TCK/Snapshot/SnapshotStoreSpec.cs
Aaronontheweb: Should we add an overrideable `virtual` variable that allows this spec to be enable / disable on an implementation by implementation basis?

Aaronontheweb (commented):
Left some comments

src/core/Akka.Persistence.TCK/Snapshot/SnapshotStoreSpec.cs
Arkatufus: The unit test methods themselves are virtual, and this is part of the spec now.
I think it is reasonable to force the other SQL implementation to adhere to the new spec?

Arkatufus: Well, "force", someone can always override and skip the test in implementation

Aaronontheweb: Doh, I missed that. You're right. LGTM.

PR #7318, opened 2024-08-15
applied 2024 copyrightheader, reformatted

Aaronontheweb:
Running some testing locally to ensure that the reformat didn't break anything - going to let the build server scope it out in parallel.


Aaronontheweb:
This was also too ambitious, it seems.

PR #7322, opened 2024-08-16
fixed `ClusterClientDiscoverySpec`

Aaronontheweb:
## Changes

Fixed a race condition with the Akka.Cluster.Tools.MultiNode.Tests project

## Checklist

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/ClusterClient/ClusterClientDiscoverySpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/ClusterClient/ClusterClientDiscoverySpec.cs
@@ -107,8 +107,6 @@ akka {
             _config = config;
         }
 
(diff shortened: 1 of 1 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests.MultiNode/ClusterClient/ClusterClientDiscoverySpec.cs
```diff
@@ -107,8 +107,6 @@ protected ClusterClientDiscoverySpec(ClusterClientDiscoverySpecConfig config)
             _config = config;
         }
 
-        protected override int InitialParticipantsValueFactory => 3;
```
Aaronontheweb, line 110: This value should be set to `4`, which the test will do by default. With a value of `3` it's possible that the `RoleName("third")` will start up late and hit a bunch of barriers late, causing the coordinator node to crash.

PR #7324, opened 2024-08-20
Improve Akka.Persistence.TestKit

Arkatufus:
Improve Akka.Persistence.TestKit

## Changes

* Allow for external pre-instantiated ActorSystem to be passed in into the constructor
* Make all of the interceptors API public instead of internal to allow custom interceptor composition
* Add a new ConnectionInterceptor that intercepts mocked database connection, emulating connection failures
* Add a new interceptor `CancelableDelay` that is less prone to strict timing failure, a delay stops immediately when cancelled, lessening dependency on fragile timing issues.


Aaronontheweb (approved):
LGTM

PR #7333, opened 2024-09-04
Fix StreamRefSerializer NRE bug

Arkatufus:
`StreamRefSerializer` will throw an NRE if `ActorMaterializer.DefaultConfig()` were added to actor system configuration before `ActorSystem` is started.

## Changes

Remove `_serialization` field from .ctor and use `_system.Serialization` instead.

### DIFF
--- a/src/core/Akka.Streams/Serialization/StreamRefSerializer.cs
+++ b/src/core/Akka.Streams/Serialization/StreamRefSerializer.cs
@@ -10,7 +10,6 @@ using System.Text;
 using Akka.Actor;
 using Akka.Serialization;
 using Akka.Streams.Serialization.Proto.Msg;
-using Akka.Util;
 using Google.Protobuf;
 using Akka.Streams.Implementation.StreamRef;
 using CumulativeDemand = Akka.Streams.Implementation.StreamRef.CumulativeDemand;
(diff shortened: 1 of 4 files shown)


src/core/Akka.Streams/Serialization/StreamRefSerializer.cs
```diff
@@ -37,52 +36,51 @@ public sealed class StreamRefSerializer : SerializerWithStringManifest
         public StreamRefSerializer(ExtendedActorSystem system) : base(system)
         {
             _system = system;
-            _serialization = system.Serialization;
```
Arkatufus, line 40: This is the actual fix. `system.Serialization` is null if config were added before the ActorSystem is started because all serializer plugins were instantiated _before_ `ActorSystem.Serialization` is assigned.

src/core/Akka.Streams/Serialization/StreamRefSerializer.cs
```diff
@@ -129,7 +127,7 @@ private SequencedOnNext DeserializeSequenceOnNext(byte[] bytes)
         {
             var onNext = Proto.Msg.SequencedOnNext.Parser.ParseFrom(bytes);
             var p = onNext.Payload;
-            var payload = _serialization.Deserialize(
+            var payload = system.Serialization.Deserialize(
```
Arkatufus, line 130: Actual fix. Instead of using a field assigned `Serialization`, we just retrieve it from `_system.Serialization` when we need it.

src/core/Akka.Streams/Serialization/StreamRefSerializer.cs
```diff
@@ -169,7 +167,7 @@ private ByteString SerializeCumulativeDemand(CumulativeDemand demand) =>
         private ByteString SerializeSequencedOnNext(SequencedOnNext onNext)
         {
             var payload = onNext.Payload;
-            var serializer = _serialization.FindSerializerFor(payload);
+            var serializer = system.Serialization.FindSerializerFor(payload);
```
Arkatufus, line 170: Actual fix. Instead of using a field assigned `Serialization`, we just retrieve it from `_system.Serialization` when we need it.

Arkatufus (commented):
Self-review

src/core/Akka.Streams/Serialization/StreamRefSerializer.cs
Aaronontheweb: This could make for a decent analyzer rule

Aaronontheweb (approved):
lgtm

PR #7334, opened 2024-09-04
Fix SaveSnapshot timestamp metadata

Arkatufus:
Currently, when `SaveSnapshot()` is invoked, the `Metadata.Timestamp` is ignored and always replaced with `DateTime.UtcNow`

## Changes

SnapshotStore should only override the metadata timestamp if it is not set (`DateTime.MinValue`)

### DIFF
--- a/src/core/Akka.Persistence/Snapshot/SnapshotStore.cs
+++ b/src/core/Akka.Persistence/Snapshot/SnapshotStore.cs
@@ -75,7 +75,7 @@ namespace Akka.Persistence.Snapshot
             }
             else if (message is SaveSnapshot saveSnapshot)
             {
-                var metadata = new SnapshotMetadata(saveSnapshot.Metadata.PersistenceId, saveSnapshot.Metadata.SequenceNr, DateTime.UtcNow);
+                var metadata = new SnapshotMetadata(saveSnapshot.Metadata.PersistenceId, saveSnapshot.Metadata.SequenceNr, saveSnapshot.Metadata.Timestamp == DateTime.MinValue ? DateTime.UtcNow : saveSnapshot.Metadata.Timestamp);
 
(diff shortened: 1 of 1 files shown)


src/core/Akka.Persistence/Snapshot/SnapshotStore.cs
```diff
@@ -75,15 +75,15 @@ private bool ReceiveSnapshotStore(object message)
             }
             else if (message is SaveSnapshot saveSnapshot)
             {
-                var metadata = new SnapshotMetadata(saveSnapshot.Metadata.PersistenceId, saveSnapshot.Metadata.SequenceNr, DateTime.UtcNow);
+                var metadata = new SnapshotMetadata(saveSnapshot.Metadata.PersistenceId, saveSnapshot.Metadata.SequenceNr, saveSnapshot.Metadata.Timestamp == DateTime.MinValue ? DateTime.UtcNow : saveSnapshot.Metadata.Timestamp);
```
Arkatufus, line 78: Fix, instead of always replacing the metadata timestamp with `DateTime.UtcNow`, we only replace it if it is not set, i.e. has a `TimeStamp.MinValue` value

src/core/Akka.Persistence/Snapshot/SnapshotStore.cs
```diff
@@ -196,8 +196,7 @@ private bool ReceiveSnapshotStore(object message)
 
         private Exception TryUnwrapException(Exception e)
         {
-            var aggregateException = e as AggregateException;
-            if (aggregateException != null)
+            if (e is AggregateException aggregateException)
```
Arkatufus, line 199: Code modernization

Arkatufus (commented):
Self review

Aaronontheweb:
LGTM

PR #7335, opened 2024-09-04
Persistence.TCK: Add SnapshotStore SaveSnapshot spec

Arkatufus:
## Changes

Add new optional spec for Akka.Persistence.TCK
* Spec for saving snapshot with the same SequenceNr
* (Optional) Data integrity spec for concurrent saves
* Data integrity spec is an opt-in spec, to turn it on, override `SupportsConcurrentSaves` to return true.


Arkatufus:
~~This PR CI/CD will fail until #7334 is merged~~

Aaronontheweb (approved):
LGTM

PR #7337, opened 2024-09-06
Remove Hyperion from DData

Aaronontheweb:
## Changes

Apparently this dependency was added by accident in https://github.com/akkadotnet/akka.net/pull/3790

## Checklist


Aaronontheweb:
Tested this locally and it confirms that we never actually used it at all, so it's safe to remove.

Arkatufus (approved):
LGTM

Drakiss666:
Glad to hear all this.
I'm looking forward to the new version.

PR #7338, opened 2024-09-12
Streams: Fix SelectAsync race condition bug

### LINKED ISSUE #7330: Item can be handed downstream from a SelectAsync stage even though a earlier item throws a exception
**Version Information**
Version of Akka.NET? 1.5.27.1

**Describe the bug**
This might be intended behavior, I just want to file this here if it isn't.
I've run into a issue with the SelectAsync with high parallelism and failures where a later item could be returned downstream before the stream fails if a exception is thrown inside SelectAsync.
I think I've hunted the issue down to [this row](https://github.com/akkadotnet/akka.net/blob/dev/src/core/Akka.Streams/Implementation/Fusing/Ops.cs#L2647) where, in case of a failed task, it's trying to push the next item downstream and there is a race condition where the next item will either be run through the rest of the graph or the stream will fail before.
In my case this meant I could miss messages because I would restart the stream on failures but maybe from a later position then I would like. The workaround is pretty simple though, I just never throw a exception from the SelectAsync and handle that in a select stage after.

If this isn't a intended behavior I wouldn't mind having a look at fixing this, I just want to confirm it's a bug before doing anything.

**To Reproduce**
1. Create a stream with a SelectAsync stage with high parallelism (I used 100 in my case) where some item will throw a exception (not the first one).
2. Handle the result in a later stage.
3. Observe that a item that got to the SelectAsync after a item that throws a exception might arrive at the later stage even though the exception should fail the stream.

**Expected behavior**
I would expect that a item that got into a SelectAsync (with any parallelism) after a item that throws a exception would never be handed downstream before the stream is failed.

**Actual behavior**
In some cases a item that got into a SelectAsync after a item that throws a exception will be handed downstream.

**Environment**
Ubuntu 22.04.4 LTS

Arkatufus:
Fixes #7330

## Changes

* Add decider check inside the `PushOne()` loop

## Notes

Since this is a race condition bug, there is no reliable way to create a regression unit test for this bug.

This is the brute force unit test used to make sure that this fix works as intended:

```csharp
[Theory]
[Repeat(10000)]
public async Task A_Flow_with_SelectAsync_must_signal_error_at_appropriate_time_from_SelectAsync(int _)
{
    var failurePoint = ThreadLocalRandom.Current.Next(3, 50);
    var output = new List<int>();
    (await Awaiting(async () =>
        {
            await Source.From(Enumerable.Range(1, 500))
                .SelectAsync(100, n =>
                {
                    if (n % failurePoint == 0)
                        return Task.Run<int>(async () =>
                        {
                            await Task.Delay(ThreadLocalRandom.Current.Next(100, 200).Milliseconds());
                            throw new TestException("err2");
                        });

                    return Task.Run(async () =>
                    {
                        await Task.Delay(ThreadLocalRandom.Current.Next(20, 50).Milliseconds());
                        return n;
                    });
                })
                .Select(i =>
                {
                    output.Add(i);
                    return i;
                })
                .RunWith(Sink.Ignore<int>(), Materializer)
                .ShouldCompleteWithin(5.Seconds());
        }).Should().ThrowAsync<AggregateException>())
        .WithInnerExceptionExactly<TestException>()
        .WithMessage("err2");

    output.Count.Should().Be(failurePoint - 1);
}
```

### DIFF
--- a/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
+++ b/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
@@ -2529,13 +2529,15 @@ namespace Akka.Streams.Implementation.Fusing
             {
                 private readonly Action<Holder<T>> _callback;
 
-                public Holder(Result<T> element, Action<Holder<T>> callback)
+                public Holder(object message, Result<T> element, Action<Holder<T>> callback)
                 {
                     _callback = callback;
+                    Message = message;
                     Element = element;
                 }
 
                 public Result<T> Element { get; private set; }
+                public object Message { get; }
 
                 public void SetElement(Result<T> result)
                 {
@@ -2575,7 +2577,7 @@ namespace Akka.Streams.Implementation.Fusing
                 try
                 {
                     var task = _stage._mapFunc(message);
-                    var holder = new Holder<TOut>(NotYetThere, _taskCallback);
+                    var holder = new Holder<TOut>(message, NotYetThere, _taskCallback);
                     _buffer.Enqueue(holder);
 
                     // We dispatch the task if it's ready to optimize away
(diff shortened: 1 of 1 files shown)


Aaronontheweb (approved):
LGTM - nice catch @Arkatufus @MattiasJakobsson

PR #7341, opened 2024-09-23
Add IWithTimers API with sender override

Arkatufus:
`IWithTimers` always pass in `ActorRefs.Nobody` as the scheduler message sender for all timers/

## Changes

Add new API methods that allows user to specify the message sender for messages sent by the `IWithTimers.Timer`

### DIFF
--- a/src/core/Akka/Actor/Scheduler/TimerScheduler.cs
+++ b/src/core/Akka/Actor/Scheduler/TimerScheduler.cs
@@ -105,7 +105,25 @@ namespace Akka.Actor.Scheduler
         /// <param name="interval">Interval</param>
         public void StartPeriodicTimer(object key, object msg, TimeSpan interval)
         {
-            StartTimer(key, msg, interval, interval, true);
+            StartTimer(key, msg, interval, interval, true, ActorRefs.NoSender);
+        }
+
+        /// <summary>
+        /// Start a periodic timer that will send <paramref name="msg"/> to the "Self" actor at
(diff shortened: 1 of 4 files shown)


src/core/Akka/Actor/Scheduler/TimerScheduler.cs
```diff
@@ -210,9 +265,9 @@ private void StartTimer(object key, object msg, TimeSpan timeout, TimeSpan initi
 
             ICancelable task;
             if (repeat)
-                task = _ctx.System.Scheduler.ScheduleTellRepeatedlyCancelable(initialDelay, timeout, _ctx.Self, timerMsg, ActorRefs.NoSender);
+                task = _ctx.System.Scheduler.ScheduleTellRepeatedlyCancelable(initialDelay, timeout, _ctx.Self, timerMsg, sender);
             else
-                task = _ctx.System.Scheduler.ScheduleTellOnceCancelable(timeout, _ctx.Self, timerMsg, ActorRefs.NoSender);
+                task = _ctx.System.Scheduler.ScheduleTellOnceCancelable(timeout, _ctx.Self, timerMsg, sender);
```
Arkatufus, line 270: The fix, allow the API to specify who the sender is and only use `ActorRefs.Nobody` if none is specified

Arkatufus (commented):
Self-review

Aaronontheweb (approved):
LGTM

Issue #7342, opened 2024-09-26
Akka.Persistence: `RecoveryTick` dead letters


Aaronontheweb:
**Version Information**
Version of Akka.NET? v1.5.27.1
Which Akka.NET Modules? Akka.Persistence

**Describe the bug**

It looks like we have a race condition somewhere inside the recovery code for persistent actors that can cause them to continue sending themselves `RecoveryTick`s even after the actor has shut down. Not entirely sure how to reproduce it, but saw some data from a customer that demonstrated that this had clearly happened and continued at a steady interval for some time afterwards.

**To Reproduce**

N/A

**Expected behavior**

No more `RecoveryTick`s should be sent after a persistent actor finishes recovering.

**Actual behavior**

It appears that in some rare cases, the `RecoveryTick`s never stop.

Aaronontheweb:
I believe this issue has been fixed.

PR #7343, opened 2024-09-26
Fix Persistence RecoveryTick timer leak

Arkatufus:
Possible fix for #7342

## Changes

Make sure that scheduler timer is cancelled


Aaronontheweb (approved):
LGTM

PR #7350, opened 2024-09-30
Fix AskTimeoutException message formatting bug

Arkatufus:
## Changes

Fix AskTimeoutException message formatting, its outputting the whole TimeSpan, not just the seconds part.


Aaronontheweb (approved):
LGTM

PR #7349, opened 2024-09-30
Fix DateTimeOffset.DateTime UTC bug

Arkatufus:
Fixes https://github.com/akkadotnet/akka.net/blame/554086ccba07338194aca2a1eb8e0f9d801973b7/src/core/Akka.Persistence/Eventsourced.cs#L221

1. It uses `DateTimeOffset.Date` which strips the time portion of the `DateTime` struct
2. `DateTimeOffset.DateTime` returns a `DateTime` object with `DateTimeKind.Unspecified` and we need `DateTimeKind.Utc`

## Changes

Add a new property in `ITimeProvider` interface that returns a `DateTime.UtcNow`

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).

PR #7353, opened 2024-10-03
Revert #7349 and #7352

Arkatufus:
## Changes

API interface change led to catastrophic failure with all plugins that depends on the interface, reverting.


Aaronontheweb (approved):
LGTM

PR #7354, opened 2024-10-03
Fix SnapshotMetadata.Timestamp bug

Arkatufus:
Fixes https://github.com/akkadotnet/akka.net/blame/554086ccba07338194aca2a1eb8e0f9d801973b7/src/core/Akka.Persistence/Eventsourced.cs#L221

1. It uses DateTimeOffset.Date which strips the time portion of the DateTime struct
2. DateTimeOffset.DateTime returns a DateTime object with DateTimeKind.Unspecified and we need DateTimeKind.Utc

## Changes

Make sure that we use UTC DateTime everywhere

### DIFF
--- a/src/core/Akka.Persistence/Eventsourced.cs
+++ b/src/core/Akka.Persistence/Eventsourced.cs
@@ -218,7 +218,7 @@ namespace Akka.Persistence
         /// <param name="snapshot">TBD</param>
         public void SaveSnapshot(object snapshot)
         {
-            SnapshotStore.Tell(new SaveSnapshot(new SnapshotMetadata(SnapshotterId, SnapshotSequenceNr, Context.System.Scheduler.Now.Date), snapshot));
+            SnapshotStore.Tell(new SaveSnapshot(new SnapshotMetadata(SnapshotterId, SnapshotSequenceNr, Context.System.Scheduler.Now.UtcDateTime), snapshot));
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


Aaronontheweb (approved):
LGTM

Arkatufus:
I'm pretty sure I've covered the majority use cases here, at least for things that were covered by our unit tests

src/core/Akka.Persistence/Eventsourced.cs
```diff
@@ -218,7 +218,7 @@ public void LoadSnapshot(string persistenceId, SnapshotSelectionCriteria criteri
         /// <param name="snapshot">TBD</param>
         public void SaveSnapshot(object snapshot)
         {
-            SnapshotStore.Tell(new SaveSnapshot(new SnapshotMetadata(SnapshotterId, SnapshotSequenceNr, Context.System.Scheduler.Now.Date), snapshot));
+            SnapshotStore.Tell(new SaveSnapshot(new SnapshotMetadata(SnapshotterId, SnapshotSequenceNr, Context.System.Scheduler.Now.UtcDateTime), snapshot));
```
Arkatufus, line 221: This is the major fix for persistence

Arkatufus (commented):
Self review

src/core/Akka.Persistence/Eventsourced.cs
Aaronontheweb: Yes that's much better than extending the scheduler interface

PR #7360, opened 2024-10-18
Add DeleteAsync failure logging

### LINKED ISSUE #7359: A failure in snapshot save can cause unobserved task exception
**Version Information**
Version of Akka.NET? 1.5.28

**Describe the bug**
When snapshot save fails, Akka will then attempt to delete the specified snapshot, if the delete snapshot operation throws an exception, this exception will become unobserved.

**To Reproduce**
This can most likely be reproduced in multiple ways, but this is what happened in our scenario. 
1. Multiple load snapshot operations failed, causing an open circuit.
2. Save snapshot operation fails, because circuit is open.
3. Delete snapshot gets triggered, because snapshot save failed.
4. Delete snapshot operation returns a faulted task with an open circuit exception, this exception is never observed.

**Expected behavior**
The exception should get observed by Akka framework.

**Actual behavior**
Exception was not observed and caused our service to crash.

**Screenshots**
image

**Environment**
Windows Service and .NET 7

Arkatufus:
Fixes #7359

Add logging when `DeleteAsync()` was called in `SaveSnapshotFailure` handler and failed.

According to scala spec, persistence does not care if `DeleteAsync()` after failed snapshot save failed or not, `DeleteAsync()` operation is done in parallel with whatever is next on the persistence queue, which sped up the `SaveSnapshotFailure` handler execution. To fully fix this we would need to block longer and signal `SaveSnapshotFailure` after the `DeleteAsync()` operation is completed.

## Changes

* Add logging

### DIFF
--- a/src/core/Akka.Persistence/Snapshot/SnapshotStore.cs
+++ b/src/core/Akka.Persistence/Snapshot/SnapshotStore.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Threading.Tasks;
 using Akka.Actor;
+using Akka.Event;
 using Akka.Pattern;
 
 namespace Akka.Persistence.Snapshot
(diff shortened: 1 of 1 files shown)


Aaronontheweb (approved):
LGTM

PR #7362, opened 2024-10-21
Fix: swallowed exceptions in timed out `WithCircuitBreaker()`

### LINKED ISSUE #7358: Akka Persistence Throwing Unobserved Task Exception
**Version Information**
Version of Akka.NET? 1.5.28

**Describe the bug**
Bug appears to be in method **CallThrough** of class **AtomicState**. If **WaitAsync(Timeout timeout)** completes before  the task implemented by SnapshotStore/Journal, and the task in question throws an exception, then this exception will become unobserved.

**To Reproduce**
1. Snapshot loading takes longer than call timeout specified by Akka framework, and Snapshot call throws an exception.

**Expected behavior**
Would expect Akka framework to observe exception thrown by SnapshotStore/Journal.

**Actual behavior**
Causes unobserved task exception and crashes our service.

**Screenshots**
image

**Environment**
Windows Service and .NET 7

Arkatufus:
Fixes #7358

## Changes

* Modernize unit tests
* Add unit test

### DIFF
--- a/src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
+++ b/src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
@@ -246,8 +246,12 @@ namespace Akka.Tests.Pattern
         {
             var breaker = ShortCallTimeoutCb();
 
-            var innerFuture = SlowThrowing();
-            var future = breaker.Instance.WithCircuitBreaker(() => innerFuture);
+            Task innerFuture = null;
+            var future = breaker.Instance.WithCircuitBreaker(ct =>
+            {
+                innerFuture = SlowThrowing(ct);
+                return innerFuture;
+            });
 
             CheckLatch(breaker.OpenLatch).Should().BeTrue();
             breaker.Instance.CurrentFailureCount.Should().Be(1);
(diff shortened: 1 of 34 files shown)


src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
```diff
@@ -199,65 +200,86 @@ public class AnAsynchronousCircuitBreakerThatIsClosed : CircuitBreakerSpecBase
...
+            await InterceptException<TimeoutException>(() => future);
+            
+            // Issue https://github.com/akkadotnet/akka.net/issues/7358
+            // The actual exception is thrown out-of-band with no handler because inner Task is detached
+            // after a timeout and NOT protected
+            
+            // In the bug, the task is still running when it should've been stopped. 
+            innerFuture.IsCompleted.Should().BeTrue();
+            innerFuture.IsFaulted.Should().BeTrue();
+            innerFuture.Exception.Should().BeOfType<TestException>();
```
Arkatufus: The actual test, the inner Task should have stopped when the circuit breaker timed out, but it is still running as a detached Task.

Arkatufus (commented):
Self-review for unit test

Arkatufus (commented):
Self-review for minimum viable fix

src/core/Akka/Util/Internal/AtomicState.cs
```diff
@@ -94,16 +97,18 @@ public async Task<T> CallThrough<T>(Func<Task<T>> task)
             return result;
```
Arkatufus, line 97: Actual fix, this does not enforce that the underlying task be stopped, it is up to the task itself to use the cancellation token and stop itself. Will need to be propagated manually to anything that uses circuit breaker.

Arkatufus (commented):
Self-review minimum viable fix

Arkatufus:
Superseeded by https://github.com/akkadotnet/akka.net/pull/7624

PR #7363, opened 2024-10-23
Add missing cluster singleton detection

Arkatufus:
As per user request, add missing cluster singleton detection to ClusterSingletonProxy to help debug misconfigured singleton setup

Detection timer:
* starts when the proxy actor started
* starts when the singleton actor ref is lost
* continuously emits/logs timeout event/log until singleton is found

## Changes

* Add new HOCON setting to opt-out of this feature
* Add new HOCON setting to set the detection period (defaults to 5 minutes)
* Add detection code to ClusterSingletonProxy
* Add unit test

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
@@ -151,9 +151,7 @@ namespace Akka.Cluster.Tools.Singleton
                 if (m.Member.UniqueAddress.Equals(_cluster.SelfUniqueAddress))
                     Context.Stop(Self);
                 else
-                {
                     Remove(m.Member);
-                }
             });
             Receive<ClusterEvent.IMemberEvent>(_ =>
             {
--- a/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
@@ -37,7 +37,7 @@ namespace Akka.Cluster.Tools.Singleton
     /// Note that this is a best effort implementation: messages can always be lost due to the distributed nature of the actors involved.
     /// </remarks>
     /// </summary>
-    public sealed class ClusterSingletonProxy : ReceiveActor
+    public sealed class ClusterSingletonProxy : ReceiveActor, IWithTimers
     {
         /// <summary>
         /// TBD
@@ -63,21 +63,33 @@ namespace Akka.Cluster.Tools.Singleton
             private IdentifySingletonTimeOutTick() { }
         }
 
+        public enum IdentifyResult
+        {
+            Success,
+            Timeout,
+        }
+        
         /// <summary>
         /// Used by the proxy to signal that no singleton has been found after a period of time
         /// </summary>
-        public sealed class IdentifySingletonTimedOut : INoSerializationVerificationNeeded
+        public sealed class IdentifySingletonResult : INoSerializationVerificationNeeded
         {
-            public IdentifySingletonTimedOut(string singletonName, string role, TimeSpan duration)
+            public static IdentifySingletonResult Success(string singletonName, string role)
+                => new (singletonName, role, IdentifyResult.Success);
+            
+            public static IdentifySingletonResult Timeout(string singletonName, string role)
+                => new (singletonName, role, IdentifyResult.Timeout);
+            
+            public IdentifySingletonResult(string singletonName, string role, IdentifyResult result)
             {
                 SingletonName = singletonName;
                 Role = role;
-                Duration = duration;
+                Result = result;
             }
 
+            public IdentifyResult Result { get; }
             public string SingletonName { get; }
             public string Role { get; }
-            public TimeSpan Duration { get; }
         }
 
         /// <summary>
@@ -113,10 +125,9 @@ namespace Akka.Cluster.Tools.Singleton
         private int _identityCounter = 0;
         private string _identityId;
         private IActorRef _singleton = null;
-        private ICancelable _identityTimer = null;
-        private ICancelable _identityTimeoutTimer = null;
         private ImmutableSortedSet<Member> _membersByAge;
         private ILoggingAdapter _log;
+        private bool _isIdentifying;
 
         /// <summary>
         /// TBD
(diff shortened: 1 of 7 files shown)


src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
```diff
@@ -85,6 +114,7 @@ public static Props Props(string singletonManagerPath, ClusterSingletonProxySett
         private string _identityId;
         private IActorRef _singleton = null;
         private ICancelable _identityTimer = null;
+        private ICancelable _identityTimeoutTimer = null;
```
gabor-bocso: Simply out of curiosity, why not the IWithTimers used instead of Context.System.Scheduler?

src/contrib/cluster/Akka.Cluster.Tools/Singleton/reference.conf
```diff
@@ -61,6 +61,12 @@ akka.cluster.singleton-proxy {
 
   # Interval at which the proxy will try to resolve the singleton instance.
   singleton-identification-interval = 1s
+  
+  # Should the singleton proxy log a warning if no singleton actor were found after a period of time
+  log-singleton-identification-failure = true
+  
+  # The period the proxy will wait until it logs a missing singleton warning, defaults to 5 minutes 
+  singleton-identification-failure-period = 5m
```
Aaronontheweb: I'd shorten this to something more like 30 seconds and I'd keep raising warnings every time the period repeats without success - that way it's really, really clear that something is wrong in the logs

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
Arkatufus: Good point, I'll try and convert the class to use IWithTimers

src/contrib/cluster/Akka.Cluster.Tools/Singleton/reference.conf
Arkatufus: Ok, that would actually make the design a lot simpler, I just need to run a repeating timer whenever `_singleton` is null and fire a warning

Arkatufus: Done

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
Arkatufus: Done

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveClusterTools.Net.verified.txt
```diff
@@ -437,26 +437,48 @@ namespace Akka.Cluster.Tools.Singleton
...
         public static Akka.Configuration.Config DefaultConfig() { }
         protected override void PostStop() { }
         protected override void PreStart() { }
         public static Akka.Actor.Props Props(string singletonManagerPath, Akka.Cluster.Tools.Singleton.ClusterSingletonProxySettings settings) { }
+        public enum IdentifyResult
+        {
+            Success = 0,
+            Timeout = 1,
+        }
+        public sealed class IdentifySingletonResult : Akka.Actor.INoSerializationVerificationNeeded
```
Aaronontheweb, line 453: Should probably make this and the `enum` private so it doesn't appear in the public API surface

src/contrib/cluster/Akka.Cluster.Tools/Singleton/reference.conf
```diff
@@ -61,6 +61,12 @@ akka.cluster.singleton-proxy {
 
   # Interval at which the proxy will try to resolve the singleton instance.
   singleton-identification-interval = 1s
+
```
Aaronontheweb, line 64: Config changes look good

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
```diff
@@ -37,7 +37,7 @@ namespace Akka.Cluster.Tools.Singleton
     /// Note that this is a best effort implementation: messages can always be lost due to the distributed nature of the actors involved.
     /// </remarks>
     /// </summary>
-    public sealed class ClusterSingletonProxy : ReceiveActor
+    public sealed class ClusterSingletonProxy : ReceiveActor, IWithTimers
```
Aaronontheweb, line 40: LGTM

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
```diff
@@ -126,19 +167,40 @@ public ClusterSingletonProxy(string singletonManagerPath, ClusterSingletonProxyS
...
                          Context.ActorSelection(singletonAddress).Tell(new Identify(_identityId));
                      }
                  });
+            Receive<IdentifySingletonTimeOutTick>(_ =>
+            {
+                // We somehow missed a CancelTimer() and a singleton reference was found when we waited,
+                // ignoring the timeout tick message.
+                if (_singleton is not null)
+                {
+                    Timers.Cancel(IdentifySingletonTimeOutTick.Instance);
```
Aaronontheweb, line 192: I might consider adding a debug message here in case there's an actual bug that lead to this, but that's not a big deal. Good to check for this condition and just turn the timer off and return early though - LGTM

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
```diff
@@ -126,19 +167,40 @@ public ClusterSingletonProxy(string singletonManagerPath, ClusterSingletonProxyS
...
+                    return;
+                }
+                
+                Log.Warning(
+                    "ClusterSingletonProxy failed to find an associated singleton named [{0}] in role [{1}] after {2} seconds.",
+                    _settings.SingletonName, _settings.Role, _settings.SingletonIdentificationFailurePeriod.TotalSeconds);
+                
+                Context.System.EventStream.Publish(IdentifySingletonResult.Timeout(
```
Aaronontheweb, line 200: Ah, is this why `IdentifySingletonResult` is part of the public API? So apps can subscribe to it?

Aaronontheweb (commented):
Left some comments

src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonProxy.cs
Arkatufus: Yes, I think it is a good addition, people can use this to detect if something went wrong in their application by hooking to this event, instead of having to check their logs

PR #7364, opened 2024-10-25
Fix Hyperion serializer config bug

Arkatufus:
The `HyperionSerializer(ExtendedActorSystem)` ctor method uses the wrong HOCON config.
`HyperionSerializerConfig.Create()` expects the config from the "akka.actor.serialization-settings.hyperion" path, but the `HyperionSerializer` ctor passes the HOCON root config to it.

Behavior change after this PR:
No known behavioral change.

## Changes

* Make sure the config being passed to the constructor is the correct config.

### DIFF
--- a/src/contrib/serializers/Akka.Serialization.Hyperion/HyperionSerializer.cs
+++ b/src/contrib/serializers/Akka.Serialization.Hyperion/HyperionSerializer.cs
@@ -170,7 +170,8 @@ namespace Akka.Serialization
         /// <summary>
         /// Default settings used by <see cref="HyperionSerializer"/> when no config has been specified.
         /// </summary>
-        public static readonly HyperionSerializerSettings Default = Create(HyperionSerializer.DefaultConfiguration());
+        public static readonly HyperionSerializerSettings Default = Create(HyperionSerializer.DefaultConfiguration()
+            .GetConfig("akka.actor.serialization-settings.hyperion"));
 
         /// <summary>
         /// Creates a new instance of <see cref="HyperionSerializerSettings"/> using provided HOCON config.
--- a/src/contrib/serializers/Akka.Serialization.Hyperion/reference.conf
+++ b/src/contrib/serializers/Akka.Serialization.Hyperion/reference.conf
@@ -1,80 +1,80 @@
 ﻿akka.actor {
-	serializers {
-		hyperion = "Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion"
-	}
-	serialization-settings {
-		hyperion {
-			# If true, hyperion serializer will preserve references between objects. 
-			# This is necessary to support things such as cyclic dependencies between objects.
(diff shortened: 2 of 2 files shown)


src/contrib/serializers/Akka.Serialization.Hyperion/reference.conf
```diff
@@ -1,80 +1,80 @@
 ﻿akka.actor {
-	serializers {
-		hyperion = "Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion"
-	}
-	serialization-settings {
-		hyperion {
-			# If true, hyperion serializer will preserve references between objects. 
-			# This is necessary to support things such as cyclic dependencies between objects.
-			preserve-object-references = true
+  serializers {
```
Arkatufus, line 2: Nothing is changed on the default config file, all changes were white space changes from tab to spaces to make sure that text formatting is consistent

src/contrib/serializers/Akka.Serialization.Hyperion/HyperionSerializer.cs
```diff
@@ -170,7 +170,8 @@ public sealed class HyperionSerializerSettings
         /// <summary>
         /// Default settings used by <see cref="HyperionSerializer"/> when no config has been specified.
         /// </summary>
-        public static readonly HyperionSerializerSettings Default = Create(HyperionSerializer.DefaultConfiguration());
+        public static readonly HyperionSerializerSettings Default = Create(HyperionSerializer.DefaultConfiguration()
+            .GetConfig("akka.actor.serialization-settings.hyperion"));
```
Arkatufus, line 174: This is the actual config `HyperionSerializerSettings` was expecting, not the root config.

Arkatufus (commented):
Self-review

Aaronontheweb:
> No known behavioral change.

doesn't this mean that the defaults behaviors, if different than what the HOCON specifies, are now enforced?

Arkatufus:
Luckily, all of the default values are exactly the same to what we've assigned if the config were empty.
https://github.com/akkadotnet/akka.net/blob/07c2b51640e94bdc35dcff15155abb5bcbe4d8e2/src/contrib/serializers/Akka.Serialization.Hyperion/HyperionSerializer.cs#L251-L258
https://github.com/akkadotnet/akka.net/blob/07c2b51640e94bdc35dcff15155abb5bcbe4d8e2/src/contrib/serializers/Akka.Serialization.Hyperion/HyperionSerializer.cs#L394

So this PR have no behavior impact, unless we change the default values.

Arkatufus:
> > No known behavioral change.
> 
> doesn't this mean that the defaults behaviors, if different than what the HOCON specifies, are now enforced?

Yes and no, yes if the default were different, luckily it is not.

Aaronontheweb:
That's good to know - thanks @Arkatufus

PR #7371, opened 2024-10-30
fixed bugs with multiple `Member`s with same `Address` crashing `ClusterDaemon`

### LINKED ISSUE #7370: Akka.Cluster: unable to mark member as `Leaving` if another instance of member with same `Address` is being marked as `DOWN`
**Version Information**
Version of Akka.NET? v1.5.30
Which Akka.NET Modules? Akka.Cluster, Akka.Cluster.Sharding

**Describe the bug**

It's possible for multiple `Member`s in the cluster to have the same `Address` - usually after a node is rebooted and the old incarnation of the node hasn't been evicted from the cluster yet. This is why Akka.Cluster uses a separate `UniqueAddress` construct - to help us identify when these types of situations occur and to distinguish between two instances of nodes with the same `Address`.

We just recently fixed one error impacted by this non-uniqueness constraint in Akka.Cluster.Sharding with #7367 and it also looks like there are issues with Akka.Cluster's `ClusterDaemon` code itself where it's susceptible to these types of problems - for instance:

https://github.com/akkadotnet/akka.net/blob/5e2bd0e2232efaab529c4be1c4db9e6328fb266b/src/core/Akka.Cluster/ClusterDaemon.cs#L1635-L1655

This is a really subtle issue, but basically: we should be iterating through EACH of these members, THEN apply the condition, and THEN remove them. The way the loop is designed right now is guaranteed to produce a `["System.InvalidOperationException: Invalid member status transition Down -> Leaving` error if there are multiple instances of the node in the gossip at this time.

**Expected behavior**

Should be able to change status of `Member`s without error even if there are multiple instances of the same `Address` in-use inside Akka.Cluster.

**Actual behavior**

The daemons crash and Akka.Cluster destabilizes.

**Environment**

Environments with stable addresses (i.e. `StatefulSet`s in Kubernetes or bare metal) are susceptible to this problem - dynamically addressed environments are not.

Aaronontheweb:
## Changes

close #7370

## Checklist

* [x] Design discussion issue #7370

### DIFF
--- a/src/core/Akka.Cluster/ClusterDaemon.cs
+++ b/src/core/Akka.Cluster/ClusterDaemon.cs
@@ -1635,22 +1635,22 @@ namespace Akka.Cluster
         public void Leaving(Address address)
         {
             // only try to update if the node is available (in the member ring)
-            if (LatestGossip.Members.Any(m => m.Address.Equals(address) && m.Status is MemberStatus.Joining or MemberStatus.WeaklyUp or MemberStatus.Up))
+            foreach(var mem in LatestGossip.Members.Where(m => m.Address.Equals(address)))
             {
-                // mark node as LEAVING
-                var newMembers = LatestGossip.Members.Select(m =>
+                if (mem.Status is MemberStatus.Joining or MemberStatus.WeaklyUp or MemberStatus.Up)
                 {
-                    if (m.Address == address) return m.Copy(status: MemberStatus.Leaving);
-                    return m;
-                }).ToImmutableSortedSet(); // mark node as LEAVING
-                var newGossip = LatestGossip.Copy(members: newMembers);
+                    // mark node as LEAVING
+                    var newMembers = LatestGossip.Members
+                        .Remove(mem).Add(mem.Copy(status: MemberStatus.Leaving));
+                    var newGossip = LatestGossip.Copy(members: newMembers);
+                    
+                    UpdateLatestGossip(newGossip);
 
-                UpdateLatestGossip(newGossip);
-
-                _cluster.LogInfo("Marked address [{0}] as [{1}]", address, MemberStatus.Leaving);
-                PublishMembershipState();
-                // immediate gossip to speed up the leaving process
-                SendGossip();
+                    _cluster.LogInfo("Marked address [{0}] as [{1}]", address, MemberStatus.Leaving);
+                    PublishMembershipState();
+                    // immediate gossip to speed up the leaving process
+                    SendGossip();
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1635,22 +1635,22 @@ public void Welcome(Address joinWith, UniqueAddress from, Gossip gossip)
         public void Leaving(Address address)
         {
             // only try to update if the node is available (in the member ring)
-            if (LatestGossip.Members.Any(m => m.Address.Equals(address) && m.Status is MemberStatus.Joining or MemberStatus.WeaklyUp or MemberStatus.Up))
+            foreach(var mem in LatestGossip.Members.Where(m => m.Address.Equals(address)))
```
Aaronontheweb, line 1638: TLDR - iterate over each `Member` with the same `Address` and filter them one at a time - don't do what we did before, which is to check if there are members that match this condition and then try to state-transition them all at the same time. This is what creates the `IllegalOperation: invalid state transition` errors.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1674,40 +1674,43 @@ public void Downing(Address address)
...
             var localMembers = localGossip.Members;
             var localOverview = localGossip.Overview;
-            var localSeen = localOverview.Seen;
             var localReachability = _membershipState.DcReachability;
 
             // check if the node to DOWN is in the 'members' set
-            var member = localMembers.FirstOrDefault(m => m.Address == address);
-            if (member != null && member.Status != MemberStatus.Down)
+            var found = false;
+            foreach (var member in localMembers.Where(m => m.Address == address))
```
Aaronontheweb, line 1681: Found the same type of issue with the `Downing` function too and made similar fixes there.

Aaronontheweb (commented):
Detailed my changes - gathering MNTR data on the build server and locally on my machine now.

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1674,40 +1674,43 @@ public void Downing(Address address)
...
             {
-                if (localReachability.IsReachable(member.UniqueAddress))
-                    _cluster.LogInfo("Marking node [{0}] as [{1}]", member.Address, MemberStatus.Down);
-                else
-                    _cluster.LogInfo("Marking unreachable node [{0}] as [{1}]", member.Address, MemberStatus.Down);
+                found = true;
+                if (member.Status != MemberStatus.Down)
+                {
```
Arkatufus, line 1685: Shouldn't `found = true;` be inside the if block?

src/core/Akka.Cluster/ClusterDaemon.cs
```diff
@@ -1635,22 +1635,22 @@ public void Welcome(Address joinWith, UniqueAddress from, Gossip gossip)
...
+                    UpdateLatestGossip(newGossip);
 
-                _cluster.LogInfo("Marked address [{0}] as [{1}]", address, MemberStatus.Leaving);
-                PublishMembershipState();
-                // immediate gossip to speed up the leaving process
-                SendGossip();
+                    _cluster.LogInfo("Marked address [{0}] as [{1}]", address, MemberStatus.Leaving);
+                    PublishMembershipState();
+                    // immediate gossip to speed up the leaving process
+                    SendGossip();
```
Arkatufus, line 1652: If we're looping over the members, shouldn't the `PublishMembershipState` and `SendGossip` method be outside the loop so we don't waste bandwidth sending gossips multiple times?

Arkatufus (changes requested):
Had some questions

src/core/Akka.Cluster/ClusterDaemon.cs
Aaronontheweb: We don't need to log anything if the node is already being downed (which is what happens inside the `foreach` but outside the `if` block) - and we want to log an info message if someone tries to have us down a node that does not exist. So it's fine where it is.

src/core/Akka.Cluster/ClusterDaemon.cs
Aaronontheweb: I thought about that too but I think it's probably fine because:

1. If there are multiple nodes in an `Up` / `Joining` / `WeaklyUp` state with the same address then we are in serious trouble - Akka.Cluster has a fundamental problem that runs a lot deeper than this. So in other words, barring a major disaster with how this code surrounding this code is implemented, that should never happen.
2. Hitting the gossip twice in the event that there are multiple member transitions detected at once is probably ok.

Arkatufus (approved):

PR #7375, opened 2024-10-31
Fix EventSourced Stash filtering out identical envelopes

### LINKED ISSUE #7373: Akka.Persistence: `IStash` filters out messages with identical senders + references
**Version Information**
Version of Akka.NET? v1.5.30

**Describe the bug**

Two versions of the same test, against an actor that uses `Stash.Stash()` internally - the `CohortActor` in this case is an `UntypedPersistentActor` that implements `IWithStash`

```csharp
[Fact]
public async Task ShouldAllowDuplicatePauseCommands()
    {
        // arrange
        var cohortId = TestCohortId;
        var cohortActor = await ActorRegistry.GetAsync<SubscriptionsActorRegistryKeys.CohortActorKey>();
        var pauseCmd = new CohortSendingCommands.PauseCohortSends(cohortId, DateTime.UtcNow);
        
        // act
        cohortActor.Tell(pauseCmd, TestActor);
        cohortActor.Tell(pauseCmd, TestActor);
        
        // assert
        var resp1 = await ExpectMsgAsync<CommandResponse>();
        var resp2 = await ExpectMsgAsync<CommandResponse>();
        
        Assert.Equal(CommandResponseCode.Success, resp1.Code);
        Assert.Equal(CommandResponseCode.NoOp, resp2.Code);
    }
```

This test fails - the second `cohortActor.Tell(pauseCmd, TestActor);` does not get processed, so the second `ExpectMsgAsync` times out. That message simply disappears - no `DeadLetter` notification or anything. I even observed two messages both go into the stash `Stash.Count == 2` but when it comes time to `UnstashAll`, only a single instance makes it into the mailbox.

The second version of this test passes:

```csharp
[Fact]
public async Task ShouldAllowDuplicatePauseCommands()
{
    // arrange
    var cohortId = TestCohortId;
    var cohortActor = await ActorRegistry.GetAsync<SubscriptionsActorRegistryKeys.CohortActorKey>();
    var pauseCmd1 = new CohortSendingCommands.PauseCohortSends(cohortId, DateTime.UtcNow);
    var pauseCmd2 = new CohortSendingCommands.PauseCohortSends(cohortId, DateTime.UtcNow);
    
    // act
    cohortActor.Tell(pauseCmd1, TestActor);
    cohortActor.Tell(pauseCmd2, TestActor);
    
    // assert
    var resp1 = await ExpectMsgAsync<CommandResponse>();
    var resp2 = await ExpectMsgAsync<CommandResponse>();
    
    Assert.Equal(CommandResponseCode.Success, resp1.Code);
    Assert.Equal(CommandResponseCode.NoOp, resp2.Code);
}
```

The message content is effectively identical, but the `pauseCmd1` and `pauseCmd2` reference equality checks do not pass!

**To Reproduce**

Runnable reproduction in LinqPad here: https://share.linqpad.net/exn9a5du.linq - this one uses Akka.Persistence actors

**Expected behavior**

Both stashed messages should be successfully processed.

**Actual behavior**

In other tests, only one of the stashed messages gets processed.

**Additional context**

Worth noting - if I can't reproduce this with vanilla Akka even if I do weird stuff like change the behavior to non-default in the CTOR of the `UntypedActor`: https://share.linqpad.net/s7urq58p.linq

Arkatufus:
Fixes #7373

## Changes

Removed the `Distinct()` filter

### DIFF
--- a/src/core/Akka/Actor/Stash/Internal/AbstractStash.cs
+++ b/src/core/Akka/Actor/Stash/Internal/AbstractStash.cs
@@ -180,7 +180,7 @@ namespace Akka.Actor.Internal
         {
             // since we want to save the order of messages, but still prepending using AddFirst,
             // we must enumerate envelopes in reversed order
-            foreach (var envelope in envelopes.Distinct().Reverse())
+            foreach (var envelope in envelopes.Reverse())
             {
                 _theStash.AddFirst(envelope);
             }


Aaronontheweb (approved):
LGTM - I'll merge this in and then re-run https://github.com/akkadotnet/akka.net/pull/7374

to11mtm:
Ooops I did mean to leave one comment on this...

The behavior is semantically correct, should note it may or may not lead to 'vexing' behaviors for consumers  that are inadvertently dependent on the behavior (e.x. repeated externally fired messages and they don't have proper guarding.

tl;dr- Good mention in release notes would be nice, I know I've had to write code to get *around* this in the past. <3

Aaronontheweb:
> Ooops I did mean to leave one comment on this...
> 
> The behavior is semantically correct, should note it may or may not lead to 'vexing' behaviors for consumers that are inadvertently dependent on the behavior (e.x. repeated externally fired messages and they don't have proper guarding.
> 
> tl;dr- Good mention in release notes would be nice, I know I've had to write code to get _around_ this in the past. <3

We'll try to address this on the front-end with https://github.com/akkadotnet/akka.analyzers/issues/104 - but yes, we'll also add an upgrade advisory for this issue in v1.5.31

PR #7388, opened 2024-11-14
[DOCS] Add custom lease documentation

Arkatufus:


docs/articles/utilities/lease.md
```diff
@@ -0,0 +1,105 @@
+---
+uid: lease
+title: Lease
```
Aaronontheweb: "Distributed Locks with Akka.Coordination" would be better

docs/articles/utilities/lease.md
```diff
@@ -0,0 +1,105 @@
...
+---
+# Lease-Based Access Control System
+
+## General Definition
+
+A **lease-based access control system** is a distributed coordination mechanism in which access to a resource is granted by issuing a lease, or token, that represents temporary permission to access a resource. The lease must be acquired by any process intending to interact with the resource. This mechanism often involves the following elements:
```
Aaronontheweb: It'd simplify this to say Akka.Coordination provides a generalized "distributed lock" implementation called a `Lease` that uses a unique resource identifier inside a backing store (such as Azure Blob Storage or Kubernetes Custom Resource Definitions) to only allow one current "holder" of the lease to perform an action at any given time.

Akka.NET uses leases internally inside Split Brain Resolvers (link), Cluster.Sharding, and Cluster Singletons for this purpose - and in this document you can learn how to call and create leases in your own Akka.NET applications if needed.

docs/articles/utilities/lease.md
```diff
@@ -0,0 +1,105 @@
...
+## Akka.Coordination.Lease
+
+`Akka.Coordination.Lease` is an abstraction in the Akka.NET that provides a way to control access to resources in distributed environments. It enables a process to acquire exclusive access to a resource through a lease. The lease is represented by an abstract class `Lease`, allowing various lease implementations that define specific lease implementations.
+
+### Key Characteristics and Components
+
+* **Lease Name**: A unique identifier for the lease, which specifies the resource to be protected.
+* **Owner Name**: A unique identifier for the entity (usually an actor or node) that is attempting to acquire the lease.
+* **Lease Timeout**: A duration parameter that specifies how long the lease should last. Leases may be renewed or revoked depending on the implementation.
+
```
Aaronontheweb, line 27: Add a section here on "available lease implementations" and link to the appropriate docs in the Akka.Management repo

Aaronontheweb (changes requested):
Looks good overall but left some feedback

docs/articles/utilities/lease.md
Arkatufus: Done

docs/articles/utilities/lease.md
Arkatufus: Done

docs/articles/utilities/lease.md
Arkatufus: Done

Aaronontheweb (approved):

PR #7402, opened 2024-12-04
Added reproduction spec for missing custom mailbox when using DI

Aaronontheweb:
## Changes

Based on this Discord thread, https://discord.com/channels/974500337396375553/1313882787350188124 - it looks like there's a bug with using custom mailboxes with DI'd actors. This PR tests for the issue.

Ultimately we were not able to reproduce it after we corrected the test assertion for the custom mailbox.

### DIFF
--- a/src/contrib/dependencyinjection/Akka.DependencyInjection.Tests/DiPropsSpecs.cs
+++ b/src/contrib/dependencyinjection/Akka.DependencyInjection.Tests/DiPropsSpecs.cs
@@ -1,0 +1,74 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="DiPropsSpecs.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2024 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2024 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Dispatch;
+using Akka.Util.Internal;
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+using Xunit.Abstractions;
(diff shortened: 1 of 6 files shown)


src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -10,6 +10,7 @@
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("Akka.Cluster.Tests.MultiNode")]
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("Akka.Cluster.Tools")]
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("Akka.DependencyInjection")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("Akka.DependencyInjection.Tests")]
```
Aaronontheweb, line 13: Had to expose friend assembly bits here in order to make the test workable, and that's what triggered the API check-in.

src/contrib/dependencyinjection/Akka.DependencyInjection.Tests/DiPropsSpecs.cs
```diff
@@ -0,0 +1,73 @@
...
+        actor.Tell(1, probe);
+        actor.Tell("test", probe);
+        var result1 = await probe.ExpectMsgAsync<Message>();
+        result1.Value.Should().Be("I was injected" + thing);
+        var result2 = await probe.ExpectMsgAsync<Message>();
+        result2.Value.Should().Be("I was injected" + thing);
+
+        // Verify that the custom mailbox was used
+        var actorRef = (RepointableActorRef)actor;
+        actorRef.MailboxType.GetType().Should().Be(typeof(CustomMailbox));
```
Aaronontheweb, line 60: Fixed how we assert for the presence of the custom mailbox to match what we do in the core AkkaSpecs and this passes now - I was also able to verify with the debugger that the custom mailbox was used.

PR #7408, opened 2024-12-16
obsolete AddOrSet

SimonCropp:
ContainsKey in AddOrSet is redundant. so may as well just set directly and burn the extension

note this will also perform better


Aaronontheweb:
Added an issue to remove this from the API altogether in v1.6 https://github.com/akkadotnet/akka.net/issues/7463

PR #7409, opened 2024-12-16
better perf in HandleStashedInbound

SimonCropp:
* remove 1 dictionary lookup. since previously GetOrElse and Remove are a dictionary lookup
 * remove redundant list allocation
 * remove redundant enumeration over the above empty list


Aaronontheweb (approved):
LGTM

Aaronontheweb:
Test failure was not related to this spec - it's an Akka.Streams spec that's tightly bounded to timings, which makes it somewhat inherently racy.

PR #7418, opened 2024-12-19
harden `CircuitBreakerSpecs.cs`

Aaronontheweb:
## Changes

harden `CircuitBreakerSpecs.Must_increment_failure_count_on_callTimeout_before_call_finishes`

## Checklist

### DIFF
--- a/src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
+++ b/src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
@@ -90,8 +90,7 @@ namespace Akka.Tests.Pattern
             await AwaitConditionAsync(() => t.Status >= TaskStatus.Running); // need to kick off the task before we can check the latch
 #pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
             var epsilon = TimeSpan.FromMilliseconds(500); // need to pad timeouts due to non-determinism of OS scheduler
-            await WithinAsync(TimeSpan.FromMilliseconds(900) + epsilon,
-                () => AwaitConditionAsync(() => breaker.Instance.CurrentFailureCount == 1, Dilated(TimeSpan.FromMilliseconds(100)), TimeSpan.FromMilliseconds(100)));
+            await AwaitConditionAsync(() => breaker.Instance.CurrentFailureCount == 1, TimeSpan.FromMilliseconds(900) + epsilon, TimeSpan.FromMilliseconds(100));
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
```diff
@@ -85,7 +85,8 @@ public async Task Must_increment_failure_count_on_callTimeout_before_call_finish
             var breaker = ShortCallTimeoutCb();
 #pragma warning disable CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
             // meant to run as detached task
-            Task.Run(() => breaker.Instance.WithSyncCircuitBreaker(() => Thread.Sleep(Dilated(TimeSpan.FromSeconds(1)))));
+            var t = Task.Run(() => breaker.Instance.WithSyncCircuitBreaker(() => Thread.Sleep(Dilated(TimeSpan.FromSeconds(1)))));
+            await AwaitConditionAsync(() => t.Status >= TaskStatus.Running); // need to kick off the task before we can check the latch
```
Aaronontheweb, line 90: This spec is pretty racy due to non-determinism of the `TaskScheduler` - now we force our assertions to wait until the `Task.Run` actually gets scheduled for execution, because otherwise the `breaker` never gets a chance to be scheduled and this will throw the timing off.

src/core/Akka.Tests/Pattern/CircuitBreakerSpec.cs
```diff
@@ -226,10 +228,10 @@ public void Must_increment_failure_count_on_async_failure()
         public async Task Must_reset_failure_count_after_success()
         {
             var breaker = MultiFailureCb();
-            _ = breaker.Instance.WithCircuitBreaker(() => Task.Run(SayHi));
-            Enumerable.Range(1, 4).ForEach(_ => breaker.Instance.WithCircuitBreaker(() => Task.Run(ThrowException)));
+            await WaitForTaskToBeScheduled(breaker.Instance.WithCircuitBreaker(() => Task.Run(SayHi)));
+            await WaitForTaskToBeScheduled(Enumerable.Range(1, 4).Select(_ => breaker.Instance.WithCircuitBreaker(() => Task.Run(ThrowException))).ToList());
```
Aaronontheweb, line 231: Issues we've been running into this spec are due to scheduler lag again - when the CPU count is low these tasks may not get scheduled to run immediately.

Aaronontheweb:
These circuit breaker specs are kind of sketchy in the first place IMHO, as many of them rely on imprecise system clock behavior and the non-deterministic scheduling of the `TaskScheduler`. Doing my best to clean this up without just disabling them.

Aaronontheweb:
This has been impacted by https://github.com/akkadotnet/akka.net/issues/7419 a couple of times; we can debug that separately

PR #7423, opened 2024-12-19
Make TestKit startup timeouts configurable

### LINKED ISSUE #7259: Make timeouts used in TestKit(Base) class configurable
Unit tests often fail with timeout errors because of parallel execution of independent unit tests regarding the TestKit class. If the system is under heavy load, then it seems that the timeouts are to close...

To reproduce the problem you can try to start a cpu stress test and execute the "Failing_Test" until an Exception will be thrown. Usually it takes less than 10 repetitions until an error will occur (on my dev system):

[Fact]
public async Task Failing_Test() {
  var tasks = Enumerable.Range(0, 100).Select(_ => Task.Run(() => new TestKit()));
  
  var actors = await Task.WhenAll(tasks);
  foreach (var actor in actors) {
    actor.Dispose();
  }    
}

### Possible errors:
1. image
2. image

### It would be nice if it would be possible to configure the following timeouts:
1. image
3. image

Aaronontheweb:
## Changes

close #7259 

## Checklist

* [x] Design discussion issue #7259

### DIFF
--- a/src/core/Akka.TestKit.Tests/TestKit_Config_Tests.cs
+++ b/src/core/Akka.TestKit.Tests/TestKit_Config_Tests.cs
@@ -21,6 +21,7 @@ namespace Akka.TestKit.Tests.Xunit2
             TestKitSettings.SingleExpectDefault.ShouldBe(TimeSpan.FromSeconds(3));
             TestKitSettings.TestEventFilterLeeway.ShouldBe(TimeSpan.FromSeconds(3));
             TestKitSettings.TestTimeFactor.ShouldBe(1);
+            TestKitSettings.TestKitStartupTimeout.ShouldBe(TimeSpan.FromSeconds(5));
             var callingThreadDispatcherTypeName = typeof(CallingThreadDispatcherConfigurator).FullName + ", " + typeof(CallingThreadDispatcher).Assembly.GetName().Name;
             Assert.False(Sys.Settings.Config.IsEmpty);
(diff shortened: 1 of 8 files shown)


src/core/Akka.TestKit/Internal/Reference.conf
```diff
@@ -26,6 +26,10 @@ akka {
     # The timeout that is added as an implicit by DefaultTimeout trait 
     # This is used for Ask-pattern
     default-timeout = 5s
+    
+    # The amount of time it takes the testkit to startup
+    # Increase this value if you're running many tests in parallel
+    startup-timeout = 5s
```
Aaronontheweb, line 32: New setting here - you can increase this value to give tests more time to initialize, which might be useful if you're running many in parallel.

src/core/Akka.TestKit.Tests/TestKit_Config_Tests.cs
```diff
@@ -21,6 +21,7 @@ public void DefaultValues_should_be_correct()
             TestKitSettings.SingleExpectDefault.ShouldBe(TimeSpan.FromSeconds(3));
             TestKitSettings.TestEventFilterLeeway.ShouldBe(TimeSpan.FromSeconds(3));
             TestKitSettings.TestTimeFactor.ShouldBe(1);
+            TestKitSettings.TestKitStartupTimeout.ShouldBe(TimeSpan.FromSeconds(5));
```
Aaronontheweb, line 24: Validate that this value can be written / overwritten via HOCON.

Aaronontheweb (commented):
Detailed my changes

PR #7425, opened 2024-12-19
Throw clearer error message during `Stash()` in `null` message cases

### LINKED ISSUE #7398: Misleading error text when IllegalActorStateException is raised on attemp to stash a lifecycle message
**Version Information**
Version 1.5.31

**Describe the bug**
If an actor tries to stash a lifecycle message (e.g. **PreStart**), the **IllegalActorStateException** is raised with the text "Can't stash the same message  more than once". The text is supposed to include the actual message, but in case of PreStart current message is **null**, so no message information is included in the text. This is confusing because there is no attempt to stash a message twice, there is an attempt to stash a message that can't be stashed. 

**Expected behavior**
Either **InvalidOperationException** should be raised, or the text for **IllegalActorStateException** should state that such message can't be stashed.

**Actual behavior**
**IllegalActorStateException** is raised with the text "Can't stash the same message  more than once".

**Additional context**
In case a lifecycle message is being stashed, the **currMessage** is **null**, and both **_currentEnvelopeId** and **_actorCell.CurrentEnvelopeId** are **zero** (their equality triggers the exception). So there should be a check for nullable currMessage that should trigger either better exception text or **InvalidOperationException**.

Aaronontheweb:
## Changes

close #7398 

## Checklist

* [x] Design discussion issue #7398

### DIFF
--- a/src/core/Akka/Actor/Stash/Internal/AbstractStash.cs
+++ b/src/core/Akka/Actor/Stash/Internal/AbstractStash.cs
@@ -74,9 +74,11 @@ namespace Akka.Actor.Internal
         {
             var currMsg = _actorCell.CurrentMessage;
             var sender = _actorCell.Sender;
-
+            
             if (_actorCell.CurrentEnvelopeId == _currentEnvelopeId)
             {
+                if(currMsg is null)
+                    throw new InvalidOperationException("There is no message to stash right now. Stash() must be called inside an actor's Receive methods.");
                 throw new IllegalActorStateException($"Can't stash the same message {currMsg} more than once");
             }
             _currentEnvelopeId = _actorCell.CurrentEnvelopeId;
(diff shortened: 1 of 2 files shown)


src/core/Akka/Actor/Stash/Internal/AbstractStash.cs
```diff
@@ -74,9 +74,11 @@ public void Stash()
         {
             var currMsg = _actorCell.CurrentMessage;
             var sender = _actorCell.Sender;
-
+            
             if (_actorCell.CurrentEnvelopeId == _currentEnvelopeId)
             {
+                if(currMsg is null)
+                    throw new InvalidOperationException("There is no message to stash right now. Stash() must be called inside an actor's Receive methods.");
```
Aaronontheweb, line 81: Provide a clearer explanation of why we're throwing an exception

Aaronontheweb (commented):
Detailed my changes

PR #7433, opened 2024-12-23
Fix `Akka.Util.Result` edge case

### LINKED ISSUE #7432: Stream upstream error propagation can fail in async calls
**Version Information**
Version of Akka.NET? Latest dev branch
Which Akka.NET Modules? Akka.Stream

**Describe the bug**

Related code:
https://github.com/akkadotnet/akka.net/blob/61bdebf6b9d24f858a2aa42263fe08e4ed62784d/src/core/Akka.Streams/Implementation/Fusing/Ops.cs#L2591-L2592

A bug in `Result<T>.FromTask()` where it failed to check for an edge case where a `Task` can be in a cancelled state but have its `Task.Exception` property be null. In fact, `TaskCancelledException` would only be thrown if `Task.Result` was called.

* This causes an event where `Result<T>` is in a failed state with its `Result<T>.Exception` property null. 
* This in turn is picked up by the upstream error propagation and 
* the null exception is passed upstream and was caught by the `OutGraphStageLogic.OnDownstreamFinish()` method, causing it to throw an `ArgumentException`. 
* This causes the whole stream to shutdown immediately without completing the proper graceful shutdown logic

Any code listening to the stream completion will never receive the completion signal. 
This can be the root cause of why Akka.Streams.Kafka fails, because the parent stream could not pick up the death of its sub-stream

**To Reproduce**

```csharp
var tcs = new TaskCompletionSource<int>();
tcs.TrySetCanceled();

var result = Result.FromTask(tcs.Task);
result.IsSuccess.Should().BeFalse();
result.Exception.Should().NotBeNull(); // Fails
```

Arkatufus:
Fixes #7432

## Changes

* Make sure that `Akka.Util.Result<T>.FromTask()` checks for edge cases
* Change `Akka.Util.Result<T>.FromTask()` behavior, it will now throw if the `Task` argument is not completed.

## Checklist

* [x] Design discussion issue #7432

### DIFF
--- a/src/core/Akka/Util/Result.cs
+++ b/src/core/Akka/Util/Result.cs
@@ -137,7 +137,30 @@ namespace Akka.Util
         /// <returns>TBD</returns>
         public static Result<T> FromTask<T>(Task<T> task)
         {
-            return task.IsCanceled || task.IsFaulted ? new Result<T>(task.Exception) : new Result<T>(task.Result);
+            if(!task.IsCompleted)
+                throw new ArgumentException("Task is not completed. Result.FromTask only accepts completed tasks.", nameof(task));
+            
+            if(task.Exception is not null)
+                return new Result<T>(task.Exception);
+            
+            if (task.IsCanceled && task.Exception is null)
+            {
+                try
+                {
+                    _ = task.GetAwaiter().GetResult();
+                }
+                catch(Exception e)
+                {
+                    return new Result<T>(e);
+                }
+
+                throw new InvalidOperationException("Should never reach this line!");
+            }
+            
+            if(task.IsFaulted && task.Exception is null)
+                throw new InvalidOperationException("Should never happen! something is wrong with .NET Task code!");
+            
+            return new Result<T>(task.Result);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/core/Akka/Util/Result.cs
```diff
@@ -137,7 +137,30 @@ public static Result<T> Failure<T>(Exception exception)
...
+                {
+                    _ = task.GetAwaiter().GetResult();
+                }
+                catch(Exception e)
+                {
+                    return new Result<T>(e);
+                }
+
+                throw new InvalidOperationException("Should never reach this line!");
+            }
```
Arkatufus, line 158: Fix, detect edge case where a `Task` is cancelled and its `Exception` property is null.

Arkatufus (commented):
Self-review

src/core/Akka/Util/Result.cs
```diff
@@ -137,7 +137,30 @@ public static Result<T> Failure<T>(Exception exception)
         /// <returns>TBD</returns>
         public static Result<T> FromTask<T>(Task<T> task)
         {
-            return task.IsCanceled || task.IsFaulted ? new Result<T>(task.Exception) : new Result<T>(task.Result);
+            if(!task.IsCompleted)
+                throw new ArgumentException("Task is not completed. Result.FromTask only accepts completed tasks.", nameof(task));
```
Arkatufus, line 141: New behavior, `.FromTask()` did not check if the `Task` argument is completed or not, it should only accept completed tasks

Arkatufus (commented):
Self-review

src/core/Akka/Util/Result.cs
Aaronontheweb: LGTM

Aaronontheweb (approved):
LGTM

PR #7435, opened 2024-12-23
Harden Streams SelectAsync specs

Arkatufus:
## Changes

Harden Streams SelectAsync specs: 
* make sure that we're testing the right thing
* add missing tests


Arkatufus:
A few tests failed, we will need to look into it:
* SelectAsync should signal completion without requiring further demand. Failed if tasks did not complete immediately.
* Decider should be called only once when a task failed. Failed, decider was being called twice.

PR #7440, opened 2025-01-02
Update Akka.FSharp README Interop with Task Parallel Library Example

jasonmar:
## Changes

Update Akka.FSharp README Interop with Task Parallel Library Example

- `System.IO.StreamReader` method name has changed to `ReadToEndAsync`
- `|!>` requires `System.Threading.Tasks.Task` to be wrapped by `Async.AwaitTask`


Aaronontheweb (approved):
LGTM

PR #7448, opened 2025-01-08
Add per-plugin recovery permitter actors

### LINKED ISSUE #7447: [PERF] Akka.Cluster.Sharding remember-entity recovery timeouts for large entity counts
**Version Information**
Version of Akka.NET? v1.5.35 and earlier
Which Akka.NET Modules? Akka.Cluster.Sharding and Akka.Persistence 

So this is _really_ an Akka.Persistence issue, but Akka.Cluster.Sharding is impacted by it.

Imagine the following scenario: 1m remembered-entity actors, each one having 3-4 children that might also use Akka.Persistence (so we're looking at a total of 3-4m persistent actors total) spread out across 100 shards on 10 nodes.

1 node goes down, 10 shards with ~30-40k persistent actors apiece need to get rebalanced across the 9 remaining nodes.

The first shard on one of the newly allocated nodes begins recovering its remembered entities, but as soon as a second shard tries to recover it runs into https://github.com/akkadotnet/akka.net/blob/08dd5bd2217fa0b8e85325c3ed88d73eadc4ac70/src/core/Akka.Persistence/persistence.conf#L18

Uh oh, now there's a long queue with 10s of thousands of entity actors all trying to recover - and so the remember-entities infrastructure hits this timeout https://github.com/akkadotnet/akka.net/blob/08dd5bd2217fa0b8e85325c3ed88d73eadc4ac70/src/contrib/cluster/Akka.Cluster.Sharding/reference.conf#L148-L150

And over and over again, starts experiencing massive recovery delays that eventually result in the infrastructure failing and the sharding system not being able to recover.

The fix for this is to do what Akka.Cluster.Sharding has supported for years: have Sharding use its own Journal and SnapshotStore so it can have a separate recovery queue from everybody else. That was, in theory, how I would have solved this issue.

To my surprise, however, this is not possible due to a puzzling design decision in the internals of Akka.Persistence:

https://github.com/akkadotnet/akka.net/blob/5e2bd0e2232efaab529c4be1c4db9e6328fb266b/src/core/Akka.Persistence/Persistence.cs#L124-L128

This means that there is a single `RecoveryPermitter` instance _spread out across all journals_ - so there's no way of addressing this prioritization problem through giving the sharding system its own journal. There is therefore, no way of alleviating this recovery pressure on the sharding system. That's a problem

**Expected behavior**

Additional journals / snapshot stores should all have their own separate recovery permitter.

**Actual behavior**

They all share the same recovery permitter.

**Environment**
Are you running on Linux? Windows? Docker? Which version of .NET?

**Additional context**

I think the reason why it was designed this way is because a "recovery" might constitute separate journals / snapshot stores, so which piece of infrastructure should get to decide if the recovery permitter gets "scoped" to specific plugins?

My advice: the **journal** is the clear winner here as it has the heaviest burden in terms of total number of records that need to be retrieved, snapshots do not. So the recovery permitter should be scoped only to the journal - snapshot stores should have no say in the matter.

Arkatufus:
Fixes #7447

## Changes

All changes are applied to internal APIs.

* Move recovery permitter cache from global `Persistence` extension to each `EventSourced` actors.
* `RecoveryPermitter` actor is now lazily created and stored inside persistence `PluginHolder`
* Add option to declare "max-concurrent-recoveries" HOCON setting in each persistence plugin that, if declared, will override the global "akka.persistence.max-concurrent-recoveries"

## Checklist

* [x] Design discussion issue #7447

### DIFF
--- a/src/core/Akka.Persistence/Eventsourced.cs
+++ b/src/core/Akka.Persistence/Eventsourced.cs
@@ -81,6 +81,7 @@ namespace Akka.Persistence
         private readonly IStash _internalStash;
         private IActorRef _snapshotStore;
         private IActorRef _journal;
+        private IActorRef _recoveryPermitter;
         private List<IPersistentEnvelope> _journalBatch = new();
         private bool _isWriteInProgress;
         private long _sequenceNr;
@@ -166,6 +167,8 @@ namespace Akka.Persistence
         /// </summary>
         public IActorRef Journal => _journal ??= Extension.JournalFor(JournalPluginId);
 
+        internal IActorRef RecoveryPermitter => _recoveryPermitter ??= Extension.RecoveryPermitterFor(JournalPluginId);
+        
         /// <summary>
         /// TBD
         /// </summary>
--- a/src/core/Akka.Persistence/Persistence.cs
+++ b/src/core/Akka.Persistence/Persistence.cs
@@ -8,7 +8,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Linq;
-using System.Reflection;
 using System.Threading;
 using Akka.Actor;
 using Akka.Annotations;
@@ -21,11 +20,12 @@ namespace Akka.Persistence
 {
     internal struct PluginHolder
     {
-        public PluginHolder(IActorRef @ref, EventAdapters adapters, Config config)
+        public PluginHolder(IActorRef @ref, EventAdapters adapters, Config config, IActorRef recoveryPermitter)
         {
             Ref = @ref;
             Adapters = adapters;
             Config = config;
+            RecoveryPermitter = recoveryPermitter;
         }
 
         public IActorRef Ref { get; }
@@ -33,6 +33,8 @@ namespace Akka.Persistence
         public EventAdapters Adapters { get; }
 
         public Config Config { get; }
+        
+        public IActorRef RecoveryPermitter { get; }
     }
 
     /// <summary>
@@ -50,7 +52,6 @@ namespace Akka.Persistence
         private readonly Lazy<string> _defaultJournalPluginId;
         private readonly Lazy<string> _defaultSnapshotPluginId;
         private readonly Lazy<IStashOverflowStrategy> _defaultInternalStashOverflowStrategy;
-        private readonly Lazy<IActorRef> _recoveryPermitter;
 
         private readonly ConcurrentDictionary<string, Lazy<PluginHolder>> _pluginExtensionIds = new();
 
@@ -120,12 +121,6 @@ namespace Akka.Persistence
                     _log.Info("Auto-starting snapshot store `{0}`", id);
                 SnapshotStoreFor(id);
             });
-
-            _recoveryPermitter = new Lazy<IActorRef>(() =>
-            {
-                var maxPermits = _config.GetInt("max-concurrent-recoveries", 0);
-                return _system.SystemActorOf(Akka.Persistence.RecoveryPermitter.Props(maxPermits), "recoveryPermitter");
-            });
         }
 
         /// <summary>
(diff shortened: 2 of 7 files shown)


src/core/Akka.Persistence/Eventsourced.cs
```diff
@@ -166,6 +167,8 @@ public IStash Stash
         /// </summary>
         public IActorRef Journal => _journal ??= Extension.JournalFor(JournalPluginId);
 
+        internal IActorRef RecoveryPermitter => _recoveryPermitter ??= Extension.RecoveryPermitterFor(JournalPluginId);
```
Aaronontheweb, line 170: LGTM

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -21,18 +20,21 @@ namespace Akka.Persistence
...
+            RecoveryPermitter = recoveryPermitter;
         }
 
         public IActorRef Ref { get; }
 
         public EventAdapters Adapters { get; }
 
         public Config Config { get; }
+        
+        public IActorRef RecoveryPermitter { get; }
```
Aaronontheweb, line 37: LGTM

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -303,8 +310,9 @@ private static PluginHolder NewPluginHolder(ExtendedActorSystem system, string c
             var config = system.Settings.Config.GetConfig(configPath).WithFallback(system.Settings.Config.GetConfig(fallbackPath));
             var plugin = CreatePlugin(system, configPath, config);
             var adapters = CreateAdapters(system, configPath);
+            var recoveryPermitter = CreateRecoveryPermitter(system, configPath, config);
```
Aaronontheweb, line 313: LGTM

src/core/Akka.Persistence/Eventsourced.cs
```diff
@@ -81,6 +81,7 @@ public abstract partial class Eventsourced : ActorBase, IPersistentIdentity, IPe
         private readonly IStash _internalStash;
         private IActorRef _snapshotStore;
         private IActorRef _journal;
+        private IActorRef _recoveryPermitter;
```
Arkatufus, line 84: The recovery permitter is being cached here instead of in the `Persistence` extension. This is lazily created when the plugin is initialized for the first time.

Aaronontheweb (changes requested):
Design looks good - but we should add some specs:

1. Validate that we look for a plugin-specific `max-concurrent-recoveries` and that that can override the global default
2. Validate, with a quick test against the `Persistence` extension itself, that multiple recovery permitter actors are created if multiple journals are defined.

No need for a big End2End integration spec, I think that should be fine.

src/core/Akka.Persistence/Eventsourced.cs
Arkatufus, line 170: Lazy instantiation

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -21,18 +20,21 @@ namespace Akka.Persistence
 {
     internal struct PluginHolder
     {
-        public PluginHolder(IActorRef @ref, EventAdapters adapters, Config config)
+        public PluginHolder(IActorRef @ref, EventAdapters adapters, Config config, IActorRef recoveryPermitter)
         {
             Ref = @ref;
             Adapters = adapters;
             Config = config;
+            RecoveryPermitter = recoveryPermitter;
```
Arkatufus, line 28: Recovery permitter is also cached inside the `PluginHolder` so that it can be looked up.

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -152,9 +147,10 @@ public string PersistenceId(IActorRef actor)
         /// INTERNAL API: When starting many persistent actors at the same time the journal its data store is protected 
         /// from being overloaded by limiting number of recoveries that can be in progress at the same time.
         /// </summary>
-        internal IActorRef RecoveryPermitter()
+        internal IActorRef RecoveryPermitterFor(string journalPluginId)
```
Arkatufus, line 150: Same design as `AdaptersFor()` to look up recovery permitter actor per plugin basis

Arkatufus (commented):
Self-review

Arkatufus:
Done adding spec

Aaronontheweb (approved):
LGTM

PR #7452, opened 2025-01-09
Fix `Shard` failed to unwrap buffered messages correctly

Arkatufus:
## Changes

* Change Shard ctor to accept buffer message adapter
* Add reproduction test

## Checklist


Aaronontheweb:
Related: https://github.com/petabridge/phobos-issues/issues/84

src/contrib/cluster/Akka.Cluster.Sharding.Tests/WrappedShardBufferedMessageSpec.cs
```diff
@@ -0,0 +1,250 @@
...
+    {
+        public MessageExtractor() : base(10)
+        {
+        }
+
+        public override string EntityId(object message)
+        {
+            return message switch
+            {
+                Message m => m.Payload,
```
Aaronontheweb: This is kind of testing the wrong thing here - the `Unapply` method should be handling this.

Aaronontheweb (changes requested):
Need to re-arrange the reproduction - don't do anything custom inside the `HashCodeMessageExtractor`. Handle this in the `Unapply` method instead.

src/contrib/cluster/Akka.Cluster.Sharding.Tests/WrappedShardBufferedMessageSpec.cs
Arkatufus: no its not, we're testing `MyEnvelope`, that `Message` class is the actual message being sent, not a `string`.

Arkatufus: actually, you're right, there's something wrong with the repro code

Aaronontheweb (approved):

PR #7457, opened 2025-01-10
remove old ASP.NET Core sample

Aaronontheweb:


Aaronontheweb:
Yikes, this stuff is referenced in the docs. That'll make it harder to remove.

PR #7460, opened 2025-01-11
upgrade MSFT.EXT and BCL libraries to 8.0+

Aaronontheweb:
## Changes

time to upgrade to the next set of LTS - 8.0 is the baseline we should be using as 6.0 is all deprecated now.

## Checklist


Arkatufus (approved):
LGTM

PR #7474, opened 2025-01-17
Have `ShardingAdapter` recursively call the underlying `IMessageExtractor`

### LINKED ISSUE #7470: Akka.Cluster.Sharding: can't recursively unpack `ShardingEnvelope` contents in `IMessageExtractor`
**Version Information**
Version of Akka.NET? v1.5.15 to v1.5.35

**Describe the bug**

The bug stems from the `ExtractorAdapter` we introduced in v1.5.15 as part of https://github.com/akkadotnet/akka.net/pull/6863

https://github.com/akkadotnet/akka.net/blob/a1c0042b18e59aa3a4e3e6c23fa2713260239cca/src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs#L79

The `ShardingEnvelope` is meant to be invisible - a transparent way to route messages to the sharding system. We have a tiny bug here that I noticed while working on https://github.com/akkadotnet/akka.net/issues/7467 this morning: if you include a message type that you want handled by the `IMessageExtractor.EntityMessage` method inside your `ShardingEnvelope`, that does not get handled and you instead get whatever the value of the `object` is inside the `ShardingEnvelope`.

No one has reported this so I assume it's not a bug anyone has run into, but the correct behavior here is that the output of the `ShardingEnvelope` should PROBABLY be routed through the `_underlying.EntityMessage` call - that way the `ShardingEnvelope` can be side-effect free, as it was intended. In most scenarios this should arrive at the same result as today.

Aaronontheweb:
## Changes

close #7470

## Checklist

* [x] Design discussion issue #7470


src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
```diff
@@ -76,7 +76,7 @@ public ExtractorAdapter(IMessageExtractor underlying)
         {
             return message switch
             {
-                ShardingEnvelope se => se.Message,
+                ShardingEnvelope se => _underlying.EntityMessage(se.Message),
```
Aaronontheweb, line 79: Also call the underlying message extractor's `EntityMessage` method on the payload of the `ShardingEnvelope`, just in case that _also_ needs unwrapping. This is needed in order to make sure that messages still work properly when using tools like Akka.Cluster.Sharding.Delivery, which uses the `ShardingEnvelope` heavily.

Aaronontheweb (commented):
Detailed my changes

src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
Arkatufus: Should we do a recursive search here instead, just to make doubly sure?

something in the line of
```
private object UnwrapShardingEnvelope(object message)
{
  object unwrapped = message;
  while(unwrapped is IWrappedMessage)
  {
    if(unwrapped is ShardingEnvelope)
      return unwrapped;
    unwrapped = unwrapped.Message;
  }
  return message;
}
```

Arkatufus (commented):
Something to think about

src/contrib/cluster/Akka.Cluster.Sharding/ClusterSharding.cs
Aaronontheweb: No because that's introducing side-effects the user is ultimately responsible for - we only care about the `ShardingEnvelope` specifically

Arkatufus (approved):
LGTM

Aaronontheweb:
Looks like this introduced a real test failure

src/contrib/cluster/Akka.Cluster.Sharding/ShardedDaemonProcess.cs
```diff
@@ -78,7 +78,9 @@ public MessageExtractor(int maxNumberOfShards)
         }
 
         public override string? EntityId(object message) => (message as ShardingEnvelope)?.EntityId;
-        public override object? EntityMessage(object message) => (message as ShardingEnvelope)?.Message;
+        
+        // Due to https://github.com/akkadotnet/akka.net/issues/7470 we will want to return the underlying content too
+        public override object? EntityMessage(object message) => (message as ShardingEnvelope)?.Message ?? message;
```
Aaronontheweb, line 83: This PR would have broken this - which is making me wonder why https://getakka.net/articles/debugging/rules/AK2001.html didn't catch this. Any ideas @Arkatufus ?

PR #7475, opened 2025-01-20
made `ActorCell` use `nullable`

Aaronontheweb:
## Changes

This PR looks scarier than it is. I've implemented `nullability` on the `ActorCell`, the brains underneath each actor, in order to support some future changes we're going to make here. None of this should result in a behavior change or a breaking API change - practically speaking, given that most of these APIs are not called directly from user code.

## Checklist

### DIFF
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
@@ -141,7 +141,7 @@ namespace Akka.Actor
                 0,
                 1})] Akka.Util.Option<object> customMessage) { }
         public bool TryGetChildStatsByName(string name, out Akka.Actor.Internal.IChildStats child) { }
-        protected bool TryGetChildStatsByRef(Akka.Actor.IActorRef actor, out Akka.Actor.Internal.ChildRestartStats child) { }
+        protected bool TryGetChildStatsByRef(Akka.Actor.IActorRef actor, [System.Diagnostics.CodeAnalysis.NotNullWhenAttribute(true)] [System.Runtime.CompilerServices.NullableAttribute(2)] out Akka.Actor.Internal.ChildRestartStats child) { }
         public void UnbecomeStacked() { }
         protected void UnreserveChild(string name) { }
         public Akka.Actor.IActorRef Unwatch(Akka.Actor.IActorRef subject) { }
@@ -324,6 +324,7 @@ namespace Akka.Actor
     }
     public class static ActorRefs
     {
+        [System.Runtime.CompilerServices.NullableAttribute(2)]
         public static readonly Akka.Actor.IActorRef NoSender;
         public static readonly Akka.Actor.Nobody Nobody;
     }
@@ -2072,7 +2073,7 @@ namespace Akka.Actor.Internal
         public abstract Akka.Actor.Internal.IChildrenContainer Reserve(string name);
         public abstract Akka.Actor.Internal.IChildrenContainer ShallDie(Akka.Actor.IActorRef actor);
         public bool TryGetByName(string name, out Akka.Actor.Internal.IChildStats stats) { }
-        public bool TryGetByRef(Akka.Actor.IActorRef actor, out Akka.Actor.Internal.ChildRestartStats childRestartStats) { }
+        public bool TryGetByRef(Akka.Actor.IActorRef actor, [System.Diagnostics.CodeAnalysis.NotNullWhenAttribute(true)] [System.Runtime.CompilerServices.NullableAttribute(2)] out Akka.Actor.Internal.ChildRestartStats childRestartStats) { }
         public abstract Akka.Actor.Internal.IChildrenContainer Unreserve(string name);
     }
     public class EmptyChildrenContainer : Akka.Actor.Internal.IChildrenContainer
@@ -2106,7 +2107,7 @@ namespace Akka.Actor.Internal
         Akka.Actor.Internal.IChildrenContainer Reserve(string name);
         Akka.Actor.Internal.IChildrenContainer ShallDie(Akka.Actor.IActorRef actor);
         bool TryGetByName(string name, out Akka.Actor.Internal.IChildStats stats);
-        bool TryGetByRef(Akka.Actor.IActorRef actor, out Akka.Actor.Internal.ChildRestartStats stats);
+        bool TryGetByRef(Akka.Actor.IActorRef actor, [System.Diagnostics.CodeAnalysis.NotNullWhenAttribute(true)] [System.Runtime.CompilerServices.NullableAttribute(2)] out Akka.Actor.Internal.ChildRestartStats stats);
         Akka.Actor.Internal.IChildrenContainer Unreserve(string name);
     }
     public interface IInitializableActor
@@ -5653,7 +5654,9 @@ namespace Akka.Util.Internal
         public static void AddOrSet<TKey, TValue>(this System.Collections.Generic.IDictionary<TKey, TValue> hash, TKey key, TValue value) { }
         public static T AsInstanceOf<T>(this object self) { }
         public static string BetweenDoubleQuotes(this string self) { }
-        public static System.Collections.Generic.IEnumerable<T> Concat<T>(this System.Collections.Generic.IEnumerable<T> enumerable, T item) { }
+        public static System.Collections.Generic.IEnumerable<T> Concat<[System.Runtime.CompilerServices.NullableAttribute(2)]  T>([System.Runtime.CompilerServices.NullableAttribute(new byte[] {
+                2,
+                1})] this System.Collections.Generic.IEnumerable<T> enumerable, T item) { }
         public static System.Collections.Generic.IEnumerable<T> Drop<T>(this System.Collections.Generic.IEnumerable<T> self, int count) { }
         public static void ForEach<T>(this System.Collections.Generic.IEnumerable<T> source, System.Action<T> action) { }
         public static TValue GetOrElse<TKey, TValue>(this System.Collections.Generic.IDictionary<TKey, TValue> hash, TKey key, TValue elseValue) { }
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
@@ -59,12 +59,14 @@ namespace Akka.Actor
         protected virtual void Unhandled(object message) { }
     }
     [System.Diagnostics.DebuggerDisplayAttribute("{Self,nq}")]
+    [System.Runtime.CompilerServices.NullableAttribute(0)]
     public class ActorCell : Akka.Actor.IActorContext, Akka.Actor.IActorRefFactory, Akka.Actor.ICanWatch, Akka.Actor.ICell, Akka.Actor.IUntypedActorContext
     {
         public const int UndefinedUid = 0;
         public ActorCell(Akka.Actor.Internal.ActorSystemImpl system, Akka.Actor.IInternalActorRef self, Akka.Actor.Props props, Akka.Dispatch.MessageDispatcher dispatcher, Akka.Actor.IInternalActorRef parent) { }
         public Akka.Actor.Internal.IChildrenContainer ChildrenContainer { get; }
         public int CurrentEnvelopeId { get; }
+        [System.Runtime.CompilerServices.NullableAttribute(2)]
         public object CurrentMessage { get; }
         public Akka.Dispatch.MessageDispatcher Dispatcher { get; }
         public bool HasMessages { get; }
@@ -72,22 +74,24 @@ namespace Akka.Actor
         protected bool IsNormal { get; }
         public bool IsTerminated { get; }
         protected bool IsTerminating { get; }
+        [System.Runtime.CompilerServices.NullableAttribute(2)]
         public Akka.Dispatch.Mailbox Mailbox { get; }
         public int NumberOfMessages { get; }
         public Akka.Actor.IInternalActorRef Parent { get; }
         public Akka.Actor.Props Props { get; }
         public System.Nullable<System.TimeSpan> ReceiveTimeout { get; }
         public Akka.Actor.IActorRef Self { get; }
+        [System.Runtime.CompilerServices.NullableAttribute(2)]
         public Akka.Actor.IActorRef Sender { get; }
         public Akka.Actor.ActorSystem System { get; }
         public Akka.Actor.Internal.ActorSystemImpl SystemImpl { get; }
         public virtual Akka.Dispatch.ActorTaskScheduler TaskScheduler { get; }
-        public virtual Akka.Actor.IActorRef ActorOf(Akka.Actor.Props props, string name = null) { }
+        public virtual Akka.Actor.IActorRef ActorOf(Akka.Actor.Props props, [System.Runtime.CompilerServices.NullableAttribute(2)] string name = null) { }
         public Akka.Actor.ActorSelection ActorSelection(string path) { }
         public Akka.Actor.ActorSelection ActorSelection(Akka.Actor.ActorPath path) { }
         protected void AddWatcher(Akka.Actor.IActorRef watchee, Akka.Actor.IActorRef watcher) { }
         protected void AddressTerminated(Akka.Actor.Address address) { }
-        public virtual Akka.Actor.IActorRef AttachChild(Akka.Actor.Props props, bool isSystemService, string name = null) { }
+        public virtual Akka.Actor.IActorRef AttachChild(Akka.Actor.Props props, bool isSystemService, [System.Runtime.CompilerServices.NullableAttribute(2)] string name = null) { }
         protected virtual void AutoReceiveMessage(Akka.Actor.Envelope envelope) { }
         public void Become(Akka.Actor.Receive receive) { }
         public void BecomeStacked(Akka.Actor.Receive receive) { }
@@ -105,6 +109,7 @@ namespace Akka.Actor
         public static Akka.Actor.IActorRef GetCurrentSenderOrNoSender() { }
         public Akka.Actor.IInternalActorRef GetSingleChild(string name) { }
         public void Init(bool sendSupervise, Akka.Dispatch.MailboxType mailboxType) { }
+        [return: System.Runtime.CompilerServices.NullableAttribute(2)]
         public Akka.Actor.Internal.ChildRestartStats InitChild(Akka.Actor.IInternalActorRef actor) { }
         public void Invoke(Akka.Actor.Envelope envelope) { }
         protected virtual void PreStart() { }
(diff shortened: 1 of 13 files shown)


Aaronontheweb:
I've also cleaned up A TON of `TBD` stuff in the XML-DOCS while I was at it. Didn't get everything, but got a lot of it.

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
Aaronontheweb, line 1: API approval file - you should still look over it but there's nothing super scary here. Going to comment on the changes in the source files themselves.

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -36,10 +36,10 @@ private IReadOnlyCollection<IActorRef> Children
         }
 
         private ImmutableDictionary<string, FunctionRef> FunctionRefs => Volatile.Read(ref _functionRefsDoNotCallMeDirectly);
-        internal bool TryGetFunctionRef(string name, out FunctionRef functionRef) =>
+        internal bool TryGetFunctionRef(string name, out FunctionRef? functionRef) =>
```
Aaronontheweb: All of these `TryGetValue<TKey, TValue>` methods have a signature of `bool `TryGetValue<TKey, TValue>(TKey key, out TValue? val)` in the .NET standard libraries - so I've brought the `TryGetFunctionRef` up to speed.

src/core/Akka/Actor/ActorCell.Children.cs
Aaronontheweb, line 1: So a really useful over-arching point for all of these code changes in the review:

**We're not changing _what is actually returned_ - we're just _accurately annotating it_ going forward**.

Please bear that in mind - only _input_ behavior has changed (i.e. explicitly handling `null` inputs where we weren't before), not any _output_ behavior. Methods that could return `null` before could still return `null`, but it's just annotated now.

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -199,7 +199,7 @@ protected void UnreserveChild(string name)
         /// </summary>
         /// <param name="actor">TBD</param>
         /// <returns>TBD</returns>
-        public ChildRestartStats InitChild(IInternalActorRef actor)
+        public ChildRestartStats? InitChild(IInternalActorRef actor)
```
Aaronontheweb, line 203: `ChildRestartStats` will be `null` if we don't have a child by this name - should _never_ happen but it has always been theoretically possible

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -318,7 +318,7 @@ public bool TryGetChildStatsByName(string name, out IChildStats child)   //This
         /// <summary>
         /// Tries to get the stats for the child with the specified name. This ignores children for whom only names have been reserved.
         /// </summary>
-        private bool TryGetChildRestartStatsByName(string name, out ChildRestartStats child)
+        private bool TryGetChildRestartStatsByName(string name, out ChildRestartStats? child)
```
Aaronontheweb: Again, the `TryGetValue` pattern applied here

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -361,20 +361,20 @@ public IInternalActorRef GetSingleChild(string name)
 
                 if (TryGetFunctionRef(name, out var functionRef))
                 {
-                    return functionRef;
+                    return functionRef!;
```
Aaronontheweb: Can't be `null` if the function returned `true`

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -361,20 +361,20 @@ public IInternalActorRef GetSingleChild(string name)
...
                 }
             }
             else
             {
                 var (s, uid) = GetNameAndUid(name);
-                if (TryGetChildRestartStatsByName(s, out var stats) && (uid == ActorCell.UndefinedUid || uid == stats.Uid))
+                if (TryGetChildRestartStatsByName(s, out var stats) && (uid == UndefinedUid || uid == stats?.Uid))
                 {
-                    return stats.Child;
+                    return stats!.Child;
```
Aaronontheweb: Can't be `null` if the function returned `true`

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -361,20 +361,20 @@ public IInternalActorRef GetSingleChild(string name)
...
                 }
 
                 if (TryGetFunctionRef(s, uid, out var functionRef))
                 {
-                    return functionRef;
+                    return functionRef!;
```
Aaronontheweb: Can't be `null` if the function returned `true`

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -385,15 +385,14 @@ public IInternalActorRef GetSingleChild(string name)
         /// </summary>
         /// <param name="child">TBD</param>
         /// <returns>TBD</returns>
-        protected SuspendReason RemoveChildAndGetStateChange(IActorRef child)
+        protected SuspendReason? RemoveChildAndGetStateChange(IActorRef child)
```
Aaronontheweb, line 389: `SuspendReason` will be `null` if this child was not found - should never happen but it is theoretically possible.

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -471,35 +470,33 @@ private IInternalActorRef MakeChild(Props props, string name, bool async, bool s
             {
                 throw new InvalidOperationException("Cannot create child while terminating or terminated");
             }
-            else
```
Aaronontheweb, line 474: `else` statement was redundant - rest of the code was left unchanged.

src/core/Akka/Actor/ActorCell.DeathWatch.cs
```diff
@@ -238,7 +240,8 @@ protected void AddWatcher(IActorRef watchee, IActorRef watcher)
                 {
                     _state = _state.AddWatchedBy(watcher);
 
-                    if (System.Settings.DebugLifecycle) Publish(new Debug(Self.Path.ToString(), Actor.GetType(), string.Format("now watched by {0}", watcher)));
+                    if (System.Settings.DebugLifecycle) Publish(new Debug(Self.Path.ToString(), Actor?.GetType(),
```
Aaronontheweb, line 243: `Actor` can be `null` - should never happen, but possible. The logging system is built to handle that though.

Aaronontheweb: You'll see this pattern show up several times throughout the review - not going to comment on each one individually, but that is what's going on here.

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -207,7 +207,7 @@ private void ReceiveSelection(ActorSelectionMessage m)
         private int CalculateState()
         {
             if(IsWaitingForChildren) return SuspendedWaitForChildrenState;
-            if(Mailbox.IsSuspended()) return SuspendedState;
+            if(Mailbox!.IsSuspended()) return SuspendedState;
```
Aaronontheweb, line 216: `Mailbox` cannot be `null` at this point - otherwise we have had a catastrophic actor initialization failure and this actor will be killed before ever reaching this point.

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -286,7 +286,7 @@ private void SysMsgInvokeAll(EarliestFirstSystemMessageList messages, int curren
                             Supervise(s.Child, s.Async);
                             break;
                         default:
-                            throw new NotSupportedException($"Unknown message {message.GetType().Name}");
+                            throw new NotSupportedException($"Unknown message {message!.GetType().Name}");
```
Aaronontheweb, line 298: `message` cannot be `null` here.

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -445,14 +445,14 @@ protected virtual ActorStopped CreateActorStoppedEvent()
             return new ActorStopped(Self, Props.Type);
         }
 
-        private void Create(Exception failure)
+        private void Create(Exception? failure)
```
Aaronontheweb, line 457: The `Exception` was always optional on the `Create` method because during first initialization there won't be one.

src/core/Akka/Actor/ActorCell.DefaultMessages.cs
```diff
@@ -347,7 +347,7 @@ private void HandleActorTaskSchedulerMessage(ActorTaskSchedulerMessage m)
         /// </summary>
         /// <param name="mailbox">TBD</param>
         /// <returns>TBD</returns>
-        internal Mailbox SwapMailbox(Mailbox mailbox)
+        internal Mailbox? SwapMailbox(Mailbox mailbox)
```
Aaronontheweb, line 359: The `Mailbox` could be, in theory, `null` if the compare-and-swap operation fails - should never happen.

src/core/Akka/Actor/ActorCell.ReceiveTimeout.cs
```diff
@@ -19,10 +19,10 @@ public interface INotInfluenceReceiveTimeout
     public partial class ActorCell
     {
         private TimeSpan? _receiveTimeoutDuration;
-        private ICancelable _pendingReceiveTimeout;
+        private ICancelable? _pendingReceiveTimeout;
```
Aaronontheweb, line 22: This field is `null` until someone schedules a `ReceiveTimeout` the first time.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -28,23 +29,19 @@ namespace Akka.Actor
     public partial class ActorCell : IUntypedActorContext, ICell
     {
         /// <summary>NOTE! Only constructor and ClearActorFields is allowed to update this</summary>
-        private IInternalActorRef _self;
+        private readonly IInternalActorRef _self;
```
Aaronontheweb, line 33: Is only ever set inside the CTOR, so we can make it `readonly`

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -77,90 +74,113 @@ private LatestFirstSystemMessageList UnstashAll()
...
+        /// Will be set to <c>null</c> when not processing any message.
+        /// </remarks>
+        public object? CurrentMessage { get; internal set; }
+        
         /// <summary>
-        /// TBD
+        /// This actor's mailbox instance.
         /// </summary>
-        public Mailbox Mailbox => _mailboxDoNotCallMeDirectly;
+        public Mailbox? Mailbox => _mailboxDoNotCallMeDirectly;
```
Aaronontheweb, line 95: The `Mailbox` will be `null` until the `ActorCell.Init` method is invoked.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -77,90 +74,113 @@ private LatestFirstSystemMessageList UnstashAll()
...
+        /// is remotely deployed from elsewhere.
         /// </summary>
         public bool IsLocal { get { return true; } }
+        
         /// <summary>
-        /// TBD
+        /// A reference to the current actor instance
         /// </summary>
-        internal ActorBase Actor { get { return _actor; } }
+        internal ActorBase? Actor { get; private set; }
```
Aaronontheweb, line 111: Technically needs to be nullable because this field will briefly be `null` during restarts and before `ActorCell.Init` is called.  Has always been this way.

Also, refactored it to use an automatic backing field.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -77,90 +74,113 @@ private LatestFirstSystemMessageList UnstashAll()
...
         /// </summary>
-        public bool IsTerminated => Mailbox.IsClosed();
+        public bool IsTerminated => Mailbox!.IsClosed();
+        
         /// <summary>
-        /// TBD
+        /// A static reference to the current actor cell.
         /// </summary>
-        internal static ActorCell Current
+        internal static ActorCell? Current
```
Aaronontheweb, line 128: Could always be `null` if the `ActorCell.Current` was accessed outside of a valid actor context.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -77,90 +74,113 @@ private LatestFirstSystemMessageList UnstashAll()
...
+        
         /// <summary>
-        /// TBD
+        /// The actor who sent us <see cref="CurrentMessage"/>.
         /// </summary>
-        public IActorRef Sender { get; private set; }
+        /// <remarks>
+        /// Will be <c>null</c> when we are not processing messages.
+        /// </remarks>
+        public IActorRef? Sender { get; private set; }
```
Aaronontheweb, line 168: `Sender` will be `null` if we haven't received a message.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -77,90 +74,113 @@ private LatestFirstSystemMessageList UnstashAll()
...
+
         /// <summary>
-        /// TBD
+        /// The <see cref="Props"/> used to create this actor.
         /// </summary>
-        public Props Props { get { return _props; } }
+        /// <remarks>
+        /// Will be re-used in the event that the actor restarts.
+        /// </remarks>
+        public Props Props { get; private set; }
```
Aaronontheweb, line 149: Removed backing field and went with auto-implementation.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -77,90 +74,113 @@ private LatestFirstSystemMessageList UnstashAll()
...
+        /// Will be <c>null</c> when we are not processing messages.
+        /// </remarks>
+        public IActorRef? Sender { get; private set; }
+        
         /// <summary>
-        /// TBD
+        /// Will return <c>true</c> if <see cref="NumberOfMessages"/> is greater than zero.
         /// </summary>
-        public bool HasMessages { get { return Mailbox.HasMessages; } }
+        public bool HasMessages { get { return Mailbox!.HasMessages; } }
```
Aaronontheweb: `Mailbox` can't be `null` by the time anyone calls this method - or else there's been a catastrophic failure.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -77,90 +74,113 @@ private LatestFirstSystemMessageList UnstashAll()
...
+        
         /// <summary>
-        /// TBD
+        /// Current message count inside the mailbox.
         /// </summary>
-        public int NumberOfMessages { get { return Mailbox.NumberOfMessages; } }
+        public int NumberOfMessages { get { return Mailbox!.NumberOfMessages; } }
```
Aaronontheweb: `Mailbox` can't be `null` by the time anyone calls this method - or else there's been a catastrophic failure.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -216,7 +236,7 @@ public void Init(bool sendSupervise, MailboxType mailboxType)
             }
 
             SwapMailbox(mailbox);
-            Mailbox.SetActor(this);
+            Mailbox!.SetActor(this);
```
Aaronontheweb, line 260: `Mailbox` can't be `null` by the time anyone calls this method - or else there's been a catastrophic failure.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -331,30 +359,33 @@ private static long NewUid()
...
                 //We should investigate what we can do to handle this better
-                instance.SupervisorStrategyInternal = _props.SupervisorStrategy;
+                instance.SupervisorStrategyInternal = Props.SupervisorStrategy;
                 //defaults to null - won't affect lazy instantiation unless explicitly set in props
             });
-            return instance;
+
+
+            Assert.Assert(instance != null, (string)(nameof(instance) + " != null"));
+            return instance!;
```
Aaronontheweb, line 397: The assertion on the line above prevents this from ever actually returning `null`.

src/core/Akka/Actor/ActorCell.cs
```diff
@@ -507,13 +541,12 @@ public static IActorRef GetCurrentSelfOrNoSender()
...
         /// <summary>
-        /// TBD
+        /// Gets the current sender or <see cref="ActorRefs.NoSender"/>
         /// </summary>
-        /// <returns>TBD</returns>
         public static IActorRef GetCurrentSenderOrNoSender()
         {
             var current = Current;
-            return current != null ? current.Sender : ActorRefs.NoSender;
+            return current?.Sender ?? ActorRefs.NoSender;
```
Aaronontheweb, line 570: Since `Current` can be `null` and `Sender` can _separately_ be `null` (this has always been a possibility), we return `ActorRefs.NoSender` here - which, ironically, is `null` right now.

src/core/Akka/Actor/ChildrenContainer/Internal/ChildrenContainerBase.cs
```diff
@@ -188,19 +187,18 @@ public bool Contains(IActorRef actor)
         /// <param name="sb">TBD</param>
         /// <param name="kvp">TBD</param>
         /// <param name="index">TBD</param>
-        protected void ChildStatsAppender(StringBuilder sb, KeyValuePair<string, IChildStats> kvp, int index)
+        protected static void ChildStatsAppender(StringBuilder sb, KeyValuePair<string, IChildStats> kvp, int index)
```
Aaronontheweb, line 193: Not a breaking change - moving to `static` would only change inheritors, not callers.

Aaronontheweb: (and there could be no inheritors of this method before.)

Aaronontheweb (commented):
Detailed my changes

src/core/Akka/Util/Internal/Extensions.cs
```diff
@@ -177,13 +177,15 @@ public static TimeSpan Min(this TimeSpan @this, TimeSpan other)
         /// <param name="enumerable">TBD</param>
         /// <param name="item">TBD</param>
         /// <returns>TBD</returns>
-        public static IEnumerable<T> Concat<T>(this IEnumerable<T> enumerable, T item)
+        #nullable enable
+        public static IEnumerable<T> Concat<T>(this IEnumerable<T>? enumerable, T item)
```
Aaronontheweb, line 181: Good call

src/core/Akka/Actor/ActorRef.cs
```diff
@@ -226,7 +226,9 @@ public static class ActorRefs
         /// Use this value as an argument to <see cref="ICanTell.Tell"/> if there is not actor to
         /// reply to (e.g. when sending from non-actor code).
         /// </summary>
-        public static readonly IActorRef NoSender = null;
+        #nullable enable
+        public static readonly IActorRef? NoSender = null;
```
Aaronontheweb, line 230: LGTM

Arkatufus (approved):
LGTM

Issue #7480, opened 2025-01-22
Akka.Persistence: `Akka.Persistence.Journal.AsyncWriteJournal+Resequencer` is created as a top-level `/user` actor instead of a child of the journal


Aaronontheweb:
**Version Information**
Version of Akka.NET? v1.5.36 and earlier
Which Akka.NET Modules? Akka.Peristence and all implementations

**Describe the bug**

While reviewing some of the work we've been doing to help keep Phobos metrics clean, @object observed a massive number of `Akka.Persistence.Journal.AsyncWriteJournal+Resequencer` running around under the `/user` actor hierarchy according to Phobos metrics: https://github.com/petabridge/phobos-issues/issues/86#issuecomment-2608221426

As it turns out, this is indeed a bug - the `Resequencer` is being created in the wrong place: 

https://github.com/akkadotnet/akka.net/blob/759e93f55f66e79d8356f6ad1526144126d0b479/src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs#L84

Now as for why there's such a large number of them, two theories:

1. This actor is getting abused by plugin implementations that create `Resequencer`s on-demand and never terminate them
2. The journal actor is restarting frequently and due to this bug, the old `Resequencer`s are never terminated.

**Expected behavior**

The journal should only need a single `Resequencer` and it should be the journal's child.

**Actual behavior**

YOLO

PR #7486, opened 2025-01-23
added v1.5.37 release notes

Aaronontheweb:
#### 1.5.37 January 23rd 2025 ####

Akka.NET v1.5.36 is a maintenance release that rolls back earlier changes made in Akka.NET v1.5.35 that have caused problems in some downstream Akka.NET plugins.

* [Rollback to using 6.0 MSFT libraries](https://github.com/akkadotnet/akka.net/pull/7482) <- moving all of our BCL dependencies to 8.0 created issues for our .NET 6-9 users when adopting Akka.NET packages that only targeted .NET Standard, so for the time being we're normalizing everything back to 6.0
* [Akka.Persistence: `Akka.Persistence.Journal.AsyncWriteJournal+Resequencer` is created as a top-level `/user` actor instead of a child of the journal](https://github.com/akkadotnet/akka.net/issues/7480)

To [see the full set of changes in Akka.NET v1.5.37, click here](https://github.com/akkadotnet/akka.net/milestone/120?closed=1).

| --- | --- | --- | --- |
| 4 | 9 | 7 | Aaron Stannard |

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,15 @@
-#### 1.5.37 January 22nd 2025 ####
+#### 1.5.37 January 23rd 2025 ####
 
-*Placeholder for nightlies*
+Akka.NET v1.5.36 is a maintenance release that rolls back earlier changes made in Akka.NET v1.5.35 that have caused problems in some downstream Akka.NET plugins.
+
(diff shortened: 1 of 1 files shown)


Arkatufus (approved):
LGTM

PR #7491, opened 2025-01-30
Add ToReadOnlySequence() to ByteString

to11mtm:
## Changes

Adds `public ReadOnlySequence<byte> ToReadOnlySequence()` to `ByteString`.

## Checklist

### Latest `dev` Benchmarks 

N/A

### This PR's Benchmarks

N/A

#### Notes:

`ReadOnlySequence` is a bit of a tradeoff; On one hand, it is a larger struct (two objects and two ints) and does some fun typechecks against said objects which makes it not as fast as `ReadOnlySpan`.

On the other hand, for non-compacted `ByteString`, rather than requiring a full copy+alloc of new array, it's essentially a linked-list of the segments (each segment being a `long` of the total length of previous segments, a `Memory<T>` and the next `ReadOnlySequenceSegment`).

(For already 'compacted' `ByteString`s, no allocation is required since `ReadOnlySequence` has logic to to avoid allocation for a single `ReadOnlyMemory` being passed in.)

The other advantage however, is that `ReadOnlySequence` is *not` a `ref struct`. 

This makes it easier to do certain user code (i.e. not having to worry about `ref struct` rules as much, e.x. including in a class being passed as a message.

As far as ecosystem support, I know MessagePack-CSharp supports `ReadOnlySequence<byte>` for input, System.Text.Json should eventually include it based on https://github.com/dotnet/runtime/issues/95187 / https://github.com/dotnet/runtime/issues/67337, and Protobuf-Net appears to have support for it as well. Probably others too but those are the first that came to mind for serialization.


src/core/Akka/Util/ByteString.cs
```diff
@@ -28,6 +29,46 @@ namespace Akka.IO
...
+                var bArr = bs._buffers;
+                var first = new ByteStringReadOnlySequenceSegment(bArr[0],0);
+                ByteBuffer item = default;
+                ByteStringReadOnlySequenceSegment last = first;
+                if (bArr.Length == 2)
+                {
+                    item = bArr[1];
+                    last = new ByteStringReadOnlySequenceSegment(item, bs.Count-item.Count);
+                    first.Next = last;
+                }
```
to11mtm: TBH this might be premature optimization....

src/core/Akka/Util/ByteString.cs
```diff
@@ -28,6 +29,46 @@ namespace Akka.IO
...
+                    for (int i = 1; i < bArr.Length; i++)
+                    {
+                        item = bArr[i];
+                        var curr = new ByteStringReadOnlySequenceSegment(item, prior.RunningIndex+prior.Memory.Length);
+                        prior.Next = curr;
+                        prior = curr;
+                    }
+                    last = prior;
+                }
+                return new ReadOnlySequence<byte>(first,0,last,last.Memory.Length);
```
to11mtm: IDK if I like this less or more than just 'going backwards on the array' and just building the Segments from last to first.

Happy for feedback (I suppose main thing being the pattern here vs setting next explicitly in ctor could impact JIT optimizations... but probably overthinking things a bit.)

src/core/Akka/Util/ByteString.cs
```diff
@@ -28,6 +29,46 @@ namespace Akka.IO
     [DebuggerDisplay("(Count = {_count}, Buffers = {_buffers})")]
     public sealed class ByteString : IEquatable<ByteString>, IEnumerable<byte>
     {
+        public class ByteStringReadOnlySequenceSegment : ReadOnlySequenceSegment<byte>
```
to11mtm: Not sure where to put in class (or maybe even move to partial/etc?)

It's a nested class for the sake of being able to directly access `_buffers` in the create. OTOH if we 'flipped' the construction (see other comment) we wouldn't need to nest it and could just build in `.ToReadOnlySequence()`.

... Either way this should be `sealed` though

to11mtm:
Needs API review

to11mtm:
All observed failures in tests appear to be races since they wouldn't be using these new APIs.

src/core/Akka/Util/ByteString.cs
Aaronontheweb: Yes it is, please remove

Aaronontheweb (changes requested):
We are probably going to restructure `ByteString` to implement `System.Memory` directly, so anything that's more complicated than 

```csharp
public ReadOnlySequence<byte> ToReadOnlySequence()
        {
            if (_count == 0)
            {
                return ReadOnlySequence<byte>.Empty;
            }
            else if (_buffers.Length == 1)
            {
                //Happy path, we can just pass ArraySegment here and avoid alloc.
                return new ReadOnlySequence<byte>(_buffers[0]);
            }
            else
            {
                return ByteStringReadOnlySequenceSegment.CreateSequence(this);
            }
        }
```

needs to be deleted.

to11mtm:
> We are probably going to restructure `ByteString` to implement `System.Memory` directly, so anything that's more complicated than
> 
>  <snip>
> 
> needs to be deleted.

Did my best to simplify the creation path; If there's specific things to handle please LMK.

Aaronontheweb:
Thanks @to11mtm

PR #7494, opened 2025-02-01
Update ClusterSharding.Node example to use shared SQLite store

### LINKED ISSUE #3608: Sample Cleanup: have Akka.Cluster.Sharding sample run on multiple nodes
We should change 
https://github.com/akkadotnet/akka.net/tree/dev/src/examples/Cluster/ClusterSharding/ClusterSharding.Node to have it actually run on multiple nodes instead of just a single node, since that's more illustrative of real-world behavior than the single-node mode it uses right now.

Looks like that can be configured via the SQLite database, but it'd be easier to just discard that and have the code fire up multiple nodes like how it does for other samples.

dimabarbul:
Fixes #3608 

## Changes

ClusterSharding.Node example is updated with following:

- Correct Dockerfile to use net8.0 folder and net8.0 base image
- Use `least-shard-allocation-strategy.rebalance-absolute-limit` instead of `least-shard-allocation-strategy.rebalance-threshold` as the latter is deprecated
- Don't use host network in docker-compose.yaml to not mess with host machine ports
- Use shared SQLite database for all nodes (using docker volume) - this is crucial for rebalancing
- Use docker compose V2 as V1 is deprecated since mid of 2023

## Checklist

### DIFF
--- a/src/examples/Cluster/ClusterSharding/ClusterSharding.Node/Dockerfile
+++ b/src/examples/Cluster/ClusterSharding/ClusterSharding.Node/Dockerfile
@@ -1,3 +1,3 @@
-﻿FROM mcr.microsoft.com/dotnet/core/runtime:3.1
-COPY ["./bin/Release/netcoreapp3.1/publish", "."]
+﻿FROM mcr.microsoft.com/dotnet/runtime:8.0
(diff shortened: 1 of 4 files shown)


Aaronontheweb (approved):
LGTM - nice work

PR #7496, opened 2025-02-05
Implement nullability in streams Buffer

Arkatufus:
## Changes

* Implement nullability in Akka.Streams.Implementation.Buffer
* Propagate and implement nullability to stream classes that uses Buffer:
  * ActorRefSourceActor
  * Buffer
  * SelectAsync
  * SelectAsyncUnordered
  * Delay
  * FlattenMerge
  * SourceRefStageImpl
  * QueueSink
  * QueueSource

## Checklist


src/core/Akka.Streams/Implementation/Sources.cs
```diff
@@ -130,20 +132,22 @@ public void OnPull()
...
+                        }
+                    }
+                    if (_terminating && _buffer.IsEmpty)
+                    {
+                        _completion.SetResult(new object());
+                        CompleteStage();
                     }
-                }
-
-                if (_terminating && _buffer.IsEmpty)
```
Arkatufus, line 143: This is an actual bug, `_buffer` can be null if `_stage._maxBuffer` is less than 1. Moved the code to inside the previous if block

src/core/Akka.Streams/Implementation/Buffers.cs
```diff
@@ -54,7 +55,7 @@ internal interface IBuffer<T>
         /// TBD
         /// </summary>
         /// <returns>TBD</returns>
-        T Peek();
+        T? Peek();
```
Arkatufus, line 58: Original behavior is that `Peek()` can return null

Arkatufus (commented):
Self-review

src/core/Akka.Streams/Implementation/ActorRefSourceActor.cs
```diff
@@ -90,7 +90,7 @@ protected bool DefaultReceive(object message)
                 Context.Stop(Self);
             else if (message is Status.Success)
             {
-                if (BufferSize == 0 || Buffer.IsEmpty)
+                if (Buffer is null || Buffer.IsEmpty)
```
Aaronontheweb, line 93: LGTM

src/core/Akka.Streams/Implementation/ActorRefSourceActor.cs
```diff
@@ -63,13 +63,13 @@ public ActorRefSourceActor(int bufferSize, OverflowStrategy overflowStrategy, in
         {
             BufferSize = bufferSize;
             OverflowStrategy = overflowStrategy;
-            Buffer = bufferSize != 0 ? Implementation.Buffer.Create<T>(bufferSize, maxFixedBufferSize) : null;
+            Buffer = bufferSize > 0 ? Implementation.Buffer.Create<T>(bufferSize, maxFixedBufferSize) : null;
```
Aaronontheweb, line 66: LGTM

Aaronontheweb: Although, should we throw an exception here if the `bufferSize` is less than `0` ?

Aaronontheweb (commented):
Not done reviewing yet but left a comment

Aaronontheweb:
Another thing I'm wondering - is there a compelling reason to have a `null` buffer at all?

src/core/Akka.Streams/Implementation/ActorRefSourceActor.cs
Arkatufus: I think we should, I just tried to preserve the old behavior

Arkatufus:
In `QueueSource` case, maybe, though I don't know what the use case is and what the thought behind it, its a 1 to 1 behavior port from scala Akka.

Arkatufus:
Though for `QueueSource`, there is a difference in scala Akka, in their version, `_pendingOffer` is another buffer, so you have 2 buffers before the stage backpressure.

Aaronontheweb:
After thinking on this, yep, there are definitely scenarios where a user might want a `null` result inside the buffer. You were right @Arkatufus

PR #7508, opened 2025-02-18
[POC] Proof of concept `TypeLoader`, replacement for `Type.GetType()`

Arkatufus:
## `TypeLoader` String to Type Map

* TypeLoader should support the current configuration type naming pattern and be able to match:
  * Fully qualified assembly type name
  * Partial "full assembly name" with "{namespace}.{typeName}, {assemblyName}" string format
  * Partial "full assembly name" with "{typeName}, {assemblyName}" string format
  * Type full name with "{namespace}.{typeName}" format
  * Type name only
  * Arbitrary string to type mapping
* TypeLoader should be able to throw an exception when lookup failed, just like `Type.GetType()` does.
* TypeLoader should return `null` when lookup failed, just like `Type.GetType()` does.

Note that we discard version, culture, and public key metadata. Should we also match these?


Aaronontheweb (changes requested):
I'd prefer something simpler and less generic for this purpose.

```csharp
internal sealed class KnownTypes {
   // ideally, should not need to be concurrent - should all happen at ActorSystem launch
   private readonly Dictionary<string, Func<Mailbox>> typesOfMailboxes;
   public Func<... args for mailbox, Mailbox> GetMailbox (string typeName) {
         
   }
}
```

We're not going to support type lookup for arbitrary stuff - so no need for a generic abstraction here. It's my preference to make this all as strongly typed, fewest-moving-parts as possible.

PR #7515, opened 2025-03-06
Add SelectAsync regression test for Akka.Persistence.Sql

Arkatufus:


src/core/Akka.Streams.Tests/Dsl/FlowSelectAsyncSpec.cs
```diff
@@ -467,5 +469,36 @@ Task<int> Deferred()
...
+                .Queue<int>(bufferSize: 5000, overflowStrategy: OverflowStrategy.DropNew)
+                .BatchWeighted(
+                    max: 100,
+                    costFunction: i => i,
+                    seed: r => ImmutableList.Create([r]),
+                    aggregate: (oldRows, i) => oldRows.Add(i))
+                .SelectAsync(
+                    parallelism: 3,
+                    asyncMapper: mapFunc)
+                .AddAttributes(ActorAttributes.CreateSupervisionStrategy(Deciders.RestartingDecider))
```
Arkatufus, line 489: The stream would stop if this attribute isn't declared, `SelectAsync` default decider is `Stop`.

PR #7497, opened 2025-02-06
[Streams] Fix null exceptions being propagated on downstream completion

Arkatufus:
Fixes https://github.com/akkadotnet/Akka.Persistence.Sql/issues/498

## Changes

Replace null `Exception` with an empty exception to prevent stages from dying from improper task continuation handling.

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).

PR #7519, opened 2025-03-12
Revert #7497

Arkatufus:
> [!NOTE]
>
> Do not merge until we know that this is a good idea


Aaronontheweb:
We decided that this is not a good idea.

PR #7520, opened 2025-03-13
Akka.Util: improve `Result<T>`

Aaronontheweb:
## Changes

Designed to support our work on https://github.com/akkadotnet/akka.net/issues/7518

Makes the following changes to `Result<T>`:

- Enables nullability
- Converts fields to properties (should have been that way from the beginning IMHO) - this _might_ be a binary compatibility issue but I doubt it though as this is mostly used internally by Akka.Streams and other components.
- Redesigns the `struct` to be a `readonly record struct` so it an automatically benefit from equality by value and other compiler optimizations

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [x] Changes in public API reviewed, if any.

PR #7521, opened 2025-03-13
Fixed race conditions + unsafe struct assignment in `SelectAsync`

### LINKED ISSUE #7518: `SelectAsync` stopped working even when async code block is guarded by a try...catch block
**Version Information**
Version of Akka.NET? 1.5.38

**Describe the bug**

Related issues:
* https://github.com/akkadotnet/Akka.Persistence.Sql/issues/497
* https://github.com/akkadotnet/Akka.Persistence.Sql/issues/498
* https://github.com/akkadotnet/akka.net/issues/7432

This has been observed inside the `Akka.Persistence.Sql` plugin, where this `SelectAsync` stage suddenly stopped working due to `ArgumentException` "Cancellation cause must not be null (Parameter 'cause')":

https://github.com/akkadotnet/Akka.Persistence.Sql/blob/e034a57789a7a982fd18e385311298adc055df01/src/Akka.Persistence.Sql/Journal/Dao/BaseByteArrayJournalDao.cs#L72-L89

Note that the `SelectAsync` code body is running inside a try...catch block.

Image

We've fixed the symptom here: https://github.com/akkadotnet/akka.net/pull/7433 and here: https://github.com/akkadotnet/akka.net/pull/7497 , but the problem still persists, causing the `Akka.Persistence.Sql` write journal to stop working completely for some user with `StreamDetachedException` "Stream is terminated. Materialized value is detached.", the same symptom seen in https://github.com/akkadotnet/Akka.Persistence.Sql/issues/497 . So far, we tried to keep the write journal going by introducing a `Restart` decider, but logically, we shouldn't need it; the stream should've kept going until the process stops with no errors.

Something in `Akka.Streams` is stopping the stream with null exception as the cause of the stop. The only point where exceptions can start to bubble up the stream stages is here inside the `GraphInterpreter` class: https://github.com/akkadotnet/akka.net/blob/6a5c7c06ef45602eb6ce7ccbd8e0c4bc88da2572/src/core/Akka.Streams/Implementation/Fusing/GraphInterpreter.cs#L866 , so far this is the only place inside `Akka.Streams` that can cause a stage cancellation that can explain the original symptom. Code review so far hasn't found any probable site where and/or when a null can possibly be injected into this code.

Aaronontheweb:
## Changes

close #7518

`struct` assignments are not guaranteed to be atomic in .NET because each field, individually, has to be assigned - as explained by the brilliant Eric Lippert: https://ericlippert.com/2011/05/31/atomicity-volatility-and-immutability-are-different-part-two/

This PR fixes an unsafe `struct` assignment that caused #7518 

## Checklist

* [x] Design discussion issue #7518

### DIFF
--- a/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
+++ b/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
@@ -2664,7 +2664,7 @@ namespace Akka.Streams.Implementation.Fusing
                                     break;
                         
                                 default:
-                                    throw new ArgumentOutOfRangeException($"Unknown SupervisionStrategy directive: {strategy}");
+                                    throw new ArgumentOutOfRangeException($"Unknown SupervisionStrategy directive: {strategy}", result.Exception);
                             }
                             continue;
                         }
@@ -2705,7 +2705,7 @@ namespace Akka.Streams.Implementation.Fusing
                         break;
                     
                     default:
-                        throw new ArgumentOutOfRangeException($"Unknown SupervisionStrategy directive: {strategy}", exception ?? new Exception("Null exception"));
+                        throw new ArgumentOutOfRangeException($"Unknown SupervisionStrategy directive: {strategy}", exception);
                 }
             }
 
--- a/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
+++ b/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
@@ -22,6 +22,7 @@ using Akka.Streams.Supervision;
 using Akka.Streams.Util;
 using Akka.Util;
 using Akka.Util.Internal;
+using Debug = System.Diagnostics.Debug;
 using Decider = Akka.Streams.Supervision.Decider;
 using Directive = Akka.Streams.Supervision.Directive;
 
@@ -2512,12 +2513,12 @@ namespace Akka.Streams.Implementation.Fusing
         /// </returns>
         public override string ToString() => "Expand";
     }
+    
+    #nullable enable
 
     /// <summary>
     /// INTERNAL API
     /// </summary>
-    /// <typeparam name="TIn">TBD</typeparam>
-    /// <typeparam name="TOut">TBD</typeparam>
     [InternalApi]
     public sealed class SelectAsync<TIn, TOut> : GraphStage<FlowShape<TIn, TOut>>
     {
@@ -2525,31 +2526,17 @@ namespace Akka.Streams.Implementation.Fusing
 
         private sealed class Logic : InAndOutGraphStageLogic
         {
-            private class Holder<T>
+            private sealed class Holder<T>(object message, Result<T> element)
             {
-                private readonly Action<Holder<T>> _callback;
-
-                public Holder(object message, Result<T> element, Action<Holder<T>> callback)
-                {
-                    _callback = callback;
-                    Message = message;
-                    Element = element;
-                }
-
-                public Result<T> Element { get; private set; }
-                public object Message { get; }
+                public object Message { get; private set; } = message;
+                
+                public Result<T> Element { get; private set; } = element;
 
                 public void SetElement(Result<T> result)
                 {
                     Element = result.IsSuccess && result.Value == null
                         ? Result.Failure<T>(ReactiveStreamsCompliance.ElementMustNotBeNullException)
                         : result;
-                }
-
-                public void Invoke(Result<T> result)
-                {
-                    SetElement(result);
-                    _callback(this);
                 }
             }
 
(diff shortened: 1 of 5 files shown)


src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -2512,61 +2513,47 @@ public Expand(Func<TIn, IEnumerator<TOut>> extrapolate)
         /// </returns>
         public override string ToString() => "Expand";
     }
+    
+    #nullable enable
```
Aaronontheweb, line 2517: Enable `nullable` to take advantage of the changes made in https://github.com/akkadotnet/akka.net/pull/7520

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -2512,61 +2513,47 @@ public Expand(Func<TIn, IEnumerator<TOut>> extrapolate)
...
-    /// <typeparam name="TOut">TBD</typeparam>
     [InternalApi]
     public sealed class SelectAsync<TIn, TOut> : GraphStage<FlowShape<TIn, TOut>>
     {
         #region internal classes
 
         private sealed class Logic : InAndOutGraphStageLogic
         {
-            private class Holder<T>
+            private sealed class Holder<T>(object message, Result<T> element)
```
Aaronontheweb: used a primary CTOR because I felt like it 🤫

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -2577,19 +2564,34 @@ public override void OnPush()
...
+                    var holder = new Holder<TOut>(message!, NotYetThere);
                     _buffer.Enqueue(holder);
 
                     // We dispatch the task if it's ready to optimize away
                     // scheduling it to an execution context
                     if (task.IsCompleted)
-                    {
-                        holder.SetElement(Result.FromTask(task));
-                        HolderCompleted(holder);
+                    {;
```
mhbuck:
```suggestion
                    {
```

Looks like there is an extra `;`

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -2512,61 +2513,47 @@ public Expand(Func<TIn, IEnumerator<TOut>> extrapolate)
...
-                }
             }
 
             private static readonly Result<TOut> NotYetThere = Result.Failure<TOut>(new Exception());
 
             private readonly SelectAsync<TIn, TOut> _stage;
             private readonly Decider _decider;
             private IBuffer<Holder<TOut>> _buffer;
-            private readonly Action<Holder<TOut>> _taskCallback;
+            private readonly Action<(Holder<TOut>, Result<TOut>)> _taskCallback;
```
Aaronontheweb, line 2548: AsyncCallbacks in Akka.Streams can only accept a single input parameter, so we have to go with a `ValueTuple` here

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -2512,61 +2513,47 @@ public Expand(Func<TIn, IEnumerator<TOut>> extrapolate)
...
-                public Holder(object message, Result<T> element, Action<Holder<T>> callback)
-                {
-                    _callback = callback;
-                    Message = message;
-                    Element = element;
-                }
-
-                public Result<T> Element { get; private set; }
-                public object Message { get; }
+                public object? Message { get; private set; } = message;
```
Aaronontheweb, line 2531: `null` might be a legal value for the message for all we know

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -2577,19 +2564,33 @@ public override void OnPush()
...
                     {
-                        holder.SetElement(Result.FromTask(task));
-                        HolderCompleted(holder);
+                        HolderCompleted(holder, Result.FromTask(task));
                     }
                     else
-                        task.ContinueWith(t => holder.Invoke(Result.FromTask(t)),
-                            TaskContinuationOptions.ExecuteSynchronously);
+                    {
+                        async Task WaitForTask()
```
Aaronontheweb, line 2578: Used the latest .NET meta for relying on a local function + detached `Task` + `await` on parent task instead of using `ContinueWith` - this is designed to tremendously simplify error-handling mostly.

Aaronontheweb: This has the side effect of eliminating the `AggregateException`s previously emitted from `SelectAsync`, which is why I've had to update some of the unit tests.

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -2677,17 +2678,18 @@ private void PushOne()
                 }
             }
 
-            private void HolderCompleted(Holder<TOut> holder)
+            private void HolderCompleted(Holder<TOut> holder, Result<TOut> result)
```
Aaronontheweb, line 2681: This is the real fix - we now pass in both the `Holder<TOut>` and the `Result<TOut>` together into the StageActor's context and perform the assignment there, where it's guaranteed to be thread-safe, rather than doing it half-in / half-out like before.

Aaronontheweb (commented):
Detailed my changes

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
Aaronontheweb: Good catch - fixed

src/core/Akka.Streams/Implementation/Fusing/Ops.cs
```diff
@@ -2663,12 +2664,12 @@ private void PushOne()
                                     break;
                         
                                 default:
-                                    throw new AggregateException($"Unknown SupervisionStrategy directive: {strategy}", result.Exception);
+                                    throw new ArgumentOutOfRangeException($"Unknown SupervisionStrategy directive: {strategy}", result.Exception);
                             }
                             continue;
                         }
```
Arkatufus, line 2670: Since `SetElement()` always get invoked in the same thread now, does that mean that this bit of code isn't needed anymore?

Arkatufus: The whole error handling code inside `PushOne()` I mean

Aaronontheweb: I think we decided in Slack that we probably still need error-handling in both places since there can be a delay on `FailStage` taking effect in the `AsyncCallback` - we'd still want to check for errors whenever else we're calling `PushOne`

Arkatufus (approved):
LGTM

PR #7527, opened 2025-03-17
Akka.Analyzer: Add AK2002 documentation

Arkatufus:
# AK2002 - Warning

`Context.Materializer()` should not be invoked multiple times, use a cached value instead.

## Cause

Calling `Context.Materializer()` inside an actor will create a new instance of streams [`ActorMaterializer`](xref:Akka.Streams.ActorMaterializer). When you create multiple streams, most likely you will only need a single actor materializer; creating multiple `ActorMaterializer` would only create memory pressure and potential memory leak if they're not disposed properly.

> [!NOTE]
>
> Using `ActorSystem.Materializer()` is safe because it is cached internally and disposed of during shutdown.

An example:

```csharp
using Akka.Actor;
using Akka.Streams;
using Akka.Streams.Dsl;

{
    public MyActor()
    {
        ReceiveAsync<string>(
            s => s == "hello",
            async _ =>
            {
                await Source.Single(1)
                    .Select(x => x + 1)
                    .ToMaterialized(Sink.ForEach<int>(_ => { }), Keep.Right)
                    .Run(Context.Materializer());
            });
    }
}
```

## Resolution

There are two ways to resolve this problem:

```csharp
using Akka.Actor;
using Akka.Streams;
using Akka.Streams.Dsl;

{
    private readonly ActorMaterializer _materializer;
    
    public MyActor()
    {
        _materializer = Context.Materializer();
        
        ReceiveAsync<string>(
            s => s == "hello",
            async _ =>
            {
                await Source.Single(1)
                    .Select(x => x + 1)
                    .ToMaterialized(Sink.ForEach<int>(_ => { }), Keep.Right)
                    .Run(_materializer); // Use a cached ActorMaterializer
            });
    }

    protected override void PostStop()
    {
        _materializer.Dispose(); // Dispose the ActorMaterializer when the actor is stopped.
    }
}
```

```csharp
using Akka.Actor;
using Akka.Streams;
using Akka.Streams.Dsl;

{
    public MyActor()
    {
        ReceiveAsync<string>(
            s => s == "hello",
            async _ =>
            {
                await Source.Single(1)
                    .Select(x => x + 1)
                    .ToMaterialized(Sink.ForEach<int>(_ => { }), Keep.Right)
                    .Run(Context.System.Materializer()); // Use ActorSystem.Materializer(), this is cached internally
            });
    }
}
```


Aaronontheweb (approved):
LGTM

PR #7540, opened 2025-03-19
Akka.Persistence.TestKt.Tests: fix `JournalInterceptorSpec`

Aaronontheweb:
## Changes

Fix race condition in cancelable_delay test by using TaskCompletionSource


src/core/Akka.Persistence.TestKit.Tests/JournalInterceptorsSpecs.cs
```diff
@@ -67,20 +67,41 @@ public async Task cancelable_delay_must_call_next_interceptor_immediately_after_
...
+            // Ensure the probe hasn't been called yet (not using probe.WasCalled since it might have race conditions)
+            synchronizationTcs.Task.IsCompleted.Should().BeFalse();
+            
+            // Cancel the delay
             cts.Cancel();
-            await task;
-
+            
+            // Wait for the probe to be called
+            await synchronizationTcs.Task;
```
Aaronontheweb, line 97: Adds a more deterministic method for making sure the probe gets called before we assert

PR #7541, opened 2025-03-19
Akka.TestKit: ensure that `EventFilter` respects `WithinAsync` timeout blocks

Aaronontheweb:
## Changes

Discovered while debugging one of the `MergeHub` specs that a 10s timeout from a `WithinAsync` block was not being passed to the `EventFilter.ExpectOneAsync` method, which instead timed out to 3s.

## Checklist


src/core/Akka.TestKit/TestKitBase_Within.cs
```diff
@@ -260,6 +260,32 @@ public async Task<T> WithinAsync<T>(
...
+             *       });
+             *   });
+             * ```
+             *
+             * This test is designed just fine, but the EventFilter has to wait the full 10 seconds before it can
+             * complete. We don't want to fail the test because the Within block took 10.1 seconds to complete.
+             */
+            const double epsilonPercentage = 0.15; // 15% fudge factor
+            const double minimumEpsilonValueMs = 50.0d; // 50ms minimum epsilon value
+            epsilonValue ??= TimeSpan.FromMilliseconds(Math.Max(max.TotalMilliseconds * epsilonPercentage, minimumEpsilonValueMs));
```
Aaronontheweb, line 287: Calculate a default epsilon value

src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs
```diff
@@ -479,8 +386,21 @@ protected async Task<T> InterceptAsync<T>(
...
+            // Otherwise, if we're in a WithinAsync block use its remaining time
+            // Otherwise use the default leeway
+            TimeSpan timeoutValue;
+            if (timeout.HasValue)
+            {
+                timeoutValue = _testkit.Dilated(timeout.Value);
+            }
+            else
+            {
+                timeoutValue = _testkit.RemainingOrDefault;
```
Aaronontheweb, line 401: Make sure we use any time values set by `Within` / `WithinAsync`

Aaronontheweb (commented):
Detailed my changes

src/core/Akka.Cluster/Cluster.cs
```diff
@@ -282,7 +282,7 @@ public Task JoinAsync(Address address, CancellationToken token = default)
 
             var completion = new TaskCompletionSource<NotUsed>(TaskCreationOptions.RunContinuationsAsynchronously);
 
-            if (token != default)
+            if (token != CancellationToken.None)
```
Arkatufus: Does `CancellationToken.None` equals to `default`? Because the default method argument value is `default`.

Aaronontheweb:
I threw some TBD removals in here while I was poking around, since the test suite that was failing as a result of my first set of changes (have `EventFilter` respect `Within` settings) was Akka.Cluster.Tests

src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs
```diff
@@ -479,8 +386,21 @@ protected async Task<T> InterceptAsync<T>(
             var leeway = system.HasExtension<TestKitSettings>()
                 ? TestKitExtension.For(system).TestEventFilterLeeway
                 : _testkit.TestKitSettings.TestEventFilterLeeway;
-
-            var timeoutValue = timeout.HasValue ? _testkit.Dilated(timeout.Value) : leeway;
+            
+            // Calculate timeout - if an explicit timeout is provided, use that (after dilating)
+            // Otherwise, if we're in a WithinAsync block use its remaining time
+            // Otherwise use the default leeway
```
Arkatufus: I don't see `leeway` variable being used anywhere in this method?

Arkatufus (changes requested):
Left some questions

src/core/Akka.Cluster/Cluster.cs
Aaronontheweb: Yes, this is a C# linter suggestion - so I'd assume they're equal

src/core/Akka.TestKit/EventFilter/Internal/EventFilterApplier.cs
Aaronontheweb: Yeah I wondered about that too - basically `leeway` is going to conflict with the `WithinAsync` block. Leeway is really supposed to be the epsilon value, but we have the call chain backwards.

src/core/Akka.Cluster/Cluster.cs
Aaronontheweb: Reverted this

Arkatufus (approved):
LGTM

Aaronontheweb:
## Akka.TestKit.Tests.TestKitBaseTests.WithinTests.Within_should_increase_max_timeout_by_the_provided_epsilon_value

```
Failed: Block took 00:00:01.1694737, exceeding 00:00:01.

   at Akka.TestKit.TestKitBase.<WithinAsync>d__223`1.MoveNext() in D:\a\1\s\src\core\Akka.TestKit\TestKitBase_Within.cs:line 347
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Akka.TestKit.TestKitBase.Within(TimeSpan max, Action action, Nullable`1 epsilonValue, CancellationToken cancellationToken) in D:\a\1\s\src\core\Akka.TestKit\TestKitBase_Within.cs:line 51
   at Akka.TestKit.Tests.TestKitBaseTests.WithinTests.Within_should_increase_max_timeout_by_the_provided_epsilon_value() in D:\a\1\s\src\core\Akka.TestKit.Tests\TestKitBaseTests\WithinTests.cs:line 25
```

Guess I'm going to need to look at that test too - first time I've seen it fail in 30-something attempts

PR #7546, opened 2025-03-20
Akka.Streams: Improve logging and completion handling in `RestartWithBackoffFlow`

Aaronontheweb:
## Changes

Added debug logging when `RestartWithBackoffFlow` terminates due to max restarts being reached on upstream finish (in scenarios where we are restarting in response to graceful shutdowns, which is configurable.) This helps diagnose scenarios where flows terminate unexpectedly due to reaching their restart limits.

The flow has two distinct behaviors:
1. `WithBackoff` - restarts on both completion and failures
2. `OnFailuresWithBackoff` - only restarts on failures

When using WithBackoff mode, normal completion triggers a restart (via `ScheduleRestartTimer()`) unless:

- The flow is already finishing (`_finishing = true`)
- Max restarts has been reached

When max restarts is reached during upstream finish, we now log this state transition to help diagnose completion behavior. This is particularly useful for understanding why a flow terminated when using `WithBackoff` mode with` maxRestarts` configured.

The debug level was chosen since this is expected behavior rather than an error condition.

Test: `A_restart_with_backoff_flow_should_not_restart_on_completion_when_maxRestarts_is_reached`

## Checklist

### DIFF
--- a/src/core/Akka.Streams/Dsl/RestartFlow.cs
+++ b/src/core/Akka.Streams/Dsl/RestartFlow.cs
@@ -304,8 +304,6 @@ namespace Akka.Streams.Dsl
             sinkIn.SetHandler(new LambdaInHandler(
                 onPush: () =>
                 {
-                    if (IsAvailable(Out))
-                        Push(Out, sinkIn.Grab());
                     Push(Out, sinkIn.Grab());
                 },
                 onUpstreamFinish: () =>
@@ -342,8 +340,6 @@ namespace Akka.Streams.Dsl
             SetHandler(Out,
                 onPull: () =>
                 {
-                    if (!sinkIn.IsClosed)
-                        sinkIn.Pull();
                     sinkIn.Pull();
                 },
                 onDownstreamFinish: cause =>
--- a/src/core/Akka.Streams/Dsl/RestartFlow.cs
+++ b/src/core/Akka.Streams/Dsl/RestartFlow.cs
@@ -306,6 +306,7 @@ namespace Akka.Streams.Dsl
                 {
                     if (IsAvailable(Out))
                         Push(Out, sinkIn.Grab());
+                    Push(Out, sinkIn.Grab());
                 },
                 onUpstreamFinish: () =>
                 {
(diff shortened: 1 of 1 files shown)


src/core/Akka.Streams/Dsl/RestartFlow.cs
```diff
@@ -301,25 +301,35 @@ protected RestartWithBackoffLogic(
...
             sinkIn.SetHandler(new LambdaInHandler(
-                onPush: () => Push(Out, sinkIn.Grab()),
+                onPush: () =>
+                {
+                    Push(Out, sinkIn.Grab());
+                },
                 onUpstreamFinish: () =>
                 {
-                    if (_finishing || MaxRestartsReached() || _onlyOnFailures)
+                    if (_finishing || (_onlyOnFailures && !MaxRestartsReached()))
```
Aaronontheweb, line 311: `A_restart_with_backoff_flow_should_not_restart_on_completion_when_maxRestarts_is_reached` has been racy because we never checked to see if `_finishing` was true during a normal completion event - we'd wait for another restart attempt instead.

src/core/Akka.Streams/Dsl/RestartFlow.cs
```diff
@@ -301,25 +301,35 @@ protected RestartWithBackoffLogic(
...
                     }
                 },
-                /*
-                 * upstream in this context is the wrapped stage
-                 */
                 onUpstreamFailure: ex =>
                 {
                     if (_finishing || MaxRestartsReached())
+                    {
+                        _finishing = true;
```
Aaronontheweb, line 330: Make sure `_finishing` is set to `true` so we short-circuit next time.

Aaronontheweb (commented):
Detailed my changes - going to have this run on AzDo a few times to see if we can reproduce the failure.

Aaronontheweb:
Still didn't fix my issue: 

```
System.Exception : [ExpectNextN] expected 2 next elements but received 1 elements before an exception occured. Received: [out complete]
---- Failed: Timeout 00:00:03 while waiting for a message of type Akka.Streams.TestKit.TestSubscriber+OnNext`1[System.String]
```

Aaronontheweb:
The real issue is deeper than just the `RestartFlow` https://github.com/akkadotnet/akka.net/issues/7555

PR #7548, opened 2025-03-20
 Akka.Peristence.Query.InMemory: properly unwrap `Tagged` messages in all queries

### LINKED ISSUE #7528: Akka.Persistence.Query.InMemory does not remove `Tagged` class like other persistence providers
**Version Information**
In version 1.5.39 

**Describe the bug**
When replaying an event that was tagged via `EventsByPersistenceId` with an InMem journal, the behavior is different as compared to MongoDB.   When using MongoDB, the `EventEnvelope.Event` is the type that was persisted.  When using InMem for testing, the `EventEnvelope.Event` is type `Tagged` if the event was tagged when it was written to the journal.

**To Reproduce**
[Repro case](https://github.com/jaydeboer/AkkaInMemTaggedBugRepro) has a test that will fail without the workaround I added to make InMem work like MongoDB

**Expected behavior**
The `EventEnvelope.Event` would be the type that the application journaled.

**Actual behavior**
The `EventEnvelope.Event` is type `Tagged` for events that were tagged.

Aaronontheweb:
## Changes

close #7528

## Checklist

* [x] Design discussion issue #7528

### DIFF
--- a/src/contrib/persistence/Akka.Persistence.Query.InMemory/MemoryQueryJournalHelpers.cs
+++ b/src/contrib/persistence/Akka.Persistence.Query.InMemory/MemoryQueryJournalHelpers.cs
@@ -1,0 +1,32 @@
+// -----------------------------------------------------------------------
+//  <copyright file="MemoryQueryJournalHelpers.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2025 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2025 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System.Linq;
+using Akka.Persistence.Journal;
+
+namespace Akka.Persistence.Query.InMemory;
+
+/// <summary>
+/// INTERNAL API
+/// </summary>
+internal static class MemoryQueryJournalHelpers
+{
+    public static EventEnvelope PrepareEnventEnvelope(IPersistentRepresentation message, Offset? offsetHint = null)
+    {
+        // Bugfix for https://github.com/akkadotnet/akka.net/issues/7528
+        var payload = message.Payload is Tagged t ? t.Payload : message.Payload;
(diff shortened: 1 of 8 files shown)


src/contrib/persistence/Akka.Persistence.Query.InMemory/MemoryQueryJournalHelpers.cs
```diff
@@ -0,0 +1,32 @@
...
+using Akka.Persistence.Journal;
+
+namespace Akka.Persistence.Query.InMemory;
+
+/// <summary>
+/// INTERNAL API
+/// </summary>
+internal static class MemoryQueryJournalHelpers
+{
+    public static EventEnvelope PrepareEnventEnvelope(IPersistentRepresentation message, Offset? offsetHint = null)
```
Aaronontheweb, line 18: Added a unified handler for inspecting `IPersistentRepresentation`s and then doing "the right thing" on the other end

src/core/Akka.Persistence/Journal/MemoryJournal.cs
```diff
@@ -127,8 +127,7 @@ private Task<int> ReplayTaggedMessagesAsync(ReplayTaggedMessages replay)
                          .Skip(replay.FromOffset)
                          .Take(replay.ToOffset))
             {
-                var payload = (Tagged)persistence.Payload;
-                replay.ReplyTo.Tell(new ReplayedTaggedMessage(persistence.WithPayload(payload.Payload), replay.Tag, replay.FromOffset + index), ActorRefs.NoSender);
+                replay.ReplyTo.Tell(new ReplayedTaggedMessage(persistence, replay.Tag, replay.FromOffset + index), ActorRefs.NoSender);
```
Aaronontheweb, line 130: The reason why tagged message queries worked is we unpacked `Tagged` messages for those specifically here - we've moved that logic instead to all of the `abstract` persistence query implementations for the `InMemory` journal.

We preserve the `Tagged` bits in the messages here because this design also caused us to lose data about _all_ of the tags applicable to this event, not _just the ones we asked for_.

Aaronontheweb (commented):
Detailed my changes

PR #7279, opened 2024-07-10
Akka.Cluster: Fix AK1004 warnings

Aaronontheweb:
## Changes

resolve a few warnings from https://getakka.net/articles/debugging/rules/AK1004.html inside Akka.Cluster

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).

PR #7562, opened 2025-03-23
suppress `AK1004` warnings in Akka.Remote and Akka.Cluster

Aaronontheweb:
## Changes

These projects require fine-grained control over timer scheduling that goes beyond IWithTimers capabilities: need for multiple concurrent timers of the same type, complex timer management for network protocols and cluster coordination. Previous attempts to use IWithTimers (PR #7279) failed due to these requirements. Added XML comments in .csproj files explaining the rationale for suppression.


Aaronontheweb:
I realized that #7279 failed because we need to have multiple scheduled messages for the same "key" inflight concurrently for some things - disabling the warning here seemed like the lesser of two evils (the alternatives being ignoring the errors or messing with systems code)

Issue #7246, opened 2024-06-11
Epic: Full AOT Support for Akka.NET


Aaronontheweb:
**Is your feature request related to a problem? Please describe.**

Part of the goal for Akka.NET v1.6 is to be able to get Akka.NET to run _everywhere_:

-  Games: Unity / Godot
- Mobile: iOS / Android
- Embedded Systems
- x64 / x86
- ARM

One of the major advantages Akka.NET has over Orleans et al is our versatility - we're not dependent on external cloud infrastructure in order to run. Akka.NET's ability to run in a stand-alone .NET process with in-memory messaging is a tremendous source of value for developers. We should lean into this.

The future of .NET is to expand beyond Windows and even x64 architectures - Ahead-of-Time Compilation, "AOT," is what will open the door for Akka.NET to be able to run on an even larger number of platforms more efficiently than it can today. So for v1.6 we are putting "full AOT support" on the critical path.

## Requirements

Where are the problem areas that prevent full AOT support today? My AOT canary implementation on TurboMqtt highlights some of the big ones: https://github.com/petabridge/TurboMqtt/pull/25

- [x] #7430
- [ ] #7431
- [ ] ExpressionCompiler magic for `ReceiveActor`, `Props`, and more: see https://github.com/akkadotnet/akka.net/issues/2811 and other related issues
- [ ] Newtonsoft.Json and default `System.Object` serialization - we are already planning on moving to Roslyn Source-Generator powered serialization in v1.6, which should address this issue. See #7465
- [ ] `Props` and its use of `NewExpression` - this is a no-no. We might need to go to more basic `Func<TActor>` implementations or other ways of specifying `Props` without using a `NewExpression`. Worth noting: `NewExpression` has been a source of perf issues as well.

All of these will need to be addressed, gradually, by leveraging tools like the `Setup` class (replaces HOCON-driven `Type` loading with static-linking-friendly loading) and changing the way some of Akka.NET's built-in components attempt to load configuration.

## Things Users Need to Do To Prepare for AOT

If you care about AOT, please read the next sentences carefully:

* The default `ActorSystem.Create` experience, with no custom plugins / Akka.Remote / Akka.Cluster / Akka.Persistence et al, will be automatically AOT-friendly out of the box;
[...]

PR #7577, opened 2025-03-24
Props - Remove usage of expressions

mhbuck:
Investigating how props can remove the usage of expressions as part of https://github.com/akkadotnet/akka.net/issues/7246

### DIFF
--- a/src/core/Akka/Actor/Props.cs
+++ b/src/core/Akka/Actor/Props.cs
@@ -1,4 +1,4 @@
-﻿//-----------------------------------------------------------------------
+//-----------------------------------------------------------------------
 // <copyright file="Props.cs" company="Akka.NET Project">
 //     Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
 //     Copyright (C) 2013-2025 .NET Foundation <https://github.com/akkadotnet/akka.net>
@@ -8,13 +8,9 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Linq.Expressions;
-using Akka.Configuration;
 using Akka.Dispatch;
 using Akka.Routing;
 using Akka.Util;
-using Akka.Util.Internal;
-using Akka.Util.Reflection;
 using Newtonsoft.Json;
 
 namespace Akka.Actor
@@ -46,8 +42,6 @@ namespace Akka.Actor
         ///     </note>
         /// </summary>
         public static readonly Props None = null;
-        private Type _inputType;
-        private Type _outputType;
         private readonly IIndirectActorProducer _producer;
 
         /// <summary>
@@ -163,12 +157,11 @@ namespace Akka.Actor
         /// <param name="args">The arguments needed to create the actor.</param>
         /// <exception cref="ArgumentException">This exception is thrown if <paramref name="type" /> is an unknown actor producer.</exception>
         public Props(Deploy deploy, Type type, params object[] args)
-#pragma warning disable CS0618 // Type or member is obsolete
-            : this(CreateProducer(type, args), deploy, args) // have to preserve the "CreateProducer" call here to preserve backwards compat with Akka.DI.Core
-#pragma warning restore CS0618 // Type or member is obsolete
+            : this(new ActivatorProducer(type, args), deploy, args)
         {
 
         }
+
 
         /// <summary>
         ///     Initializes a new instance of the <see cref="Props" /> class using a specified <see cref="IIndirectActorProducer"/>.
@@ -182,7 +175,6 @@ namespace Akka.Actor
         internal Props(IIndirectActorProducer producer, Deploy deploy, params object[] args)
         {
             Deploy = deploy;
-            _inputType = producer.ActorType;
             Arguments = args ?? NoArgs;
             _producer = producer;
         }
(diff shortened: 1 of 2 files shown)


src/core/Akka/Actor/Props.cs
```diff
@@ -163,13 +157,12 @@ public Props(Deploy deploy, Type type, IEnumerable<object> args)
         /// <param name="args">The arguments needed to create the actor.</param>
         /// <exception cref="ArgumentException">This exception is thrown if <paramref name="type" /> is an unknown actor producer.</exception>
         public Props(Deploy deploy, Type type, params object[] args)
-#pragma warning disable CS0618 // Type or member is obsolete
-            : this(CreateProducer(type, args), deploy, args) // have to preserve the "CreateProducer" call here to preserve backwards compat with Akka.DI.Core
```
mhbuck, line 167: I have a feeling removing this is not supposed to happen. Still need to understand any test failures.

src/core/Akka/Actor/Props.cs
```diff
@@ -46,8 +42,6 @@ public class Props : IEquatable<Props>, ISurrogated
         ///     </note>
         /// </summary>
         public static readonly Props None = null;
-        private Type _inputType;
-        private Type _outputType;
```
mhbuck, line 50: I expect this removal needs to come back in mainly due to the serialisation comment below on the `TypeName` property

src/core/Akka/Actor/Props.cs
```diff
@@ -225,9 +215,8 @@ public string Dispatcher
         /// </summary>
         public string TypeName
         {
-            get => _inputType.AssemblyQualifiedName;
-            //for serialization
-            private set => _inputType = Type.GetType(value);
+            get => _producer.ActorType.AssemblyQualifiedName;
+            //private set => _producer.ActorType = Type.GetType(value);
```
mhbuck, line 219: Commented out for now

src/core/Akka/Actor/Props.cs
```diff
@@ -349,19 +337,9 @@ public override int GetHashCode()
...
-                return new DynamicProps<TActor>(factory.Compile());
-
-            var newExpression = factory.Body.AsInstanceOf<NewExpression>();
-            if (newExpression == null)
-                throw new ArgumentException("The create function must be a 'new T (args)' expression");
-
-            var args = newExpression.Arguments.Count > 0 ? newExpression.GetArguments() : NoArgs;
-
-            return new Props(new ActivatorProducer(typeof(TActor), args), DefaultDeploy, args){ SupervisorStrategy = supervisorStrategy };
+            return new Props(new FactoryProducer<TActor>(factory), DefaultDeploy, NoArgs) { SupervisorStrategy = supervisorStrategy };
```
mhbuck, line 342: Func being used in the factory producer

src/core/Akka/Actor/Props.cs
```diff
@@ -687,23 +645,22 @@ public ActivatorProducer(Type actorType, object[] args)
...
 
-
             public void Release(ActorBase actor)
             {
                 actor = null;
             }
         }
 
-        private class FactoryConsumer<TActor> : IIndirectActorProducer where TActor : ActorBase
+        private class FactoryProducer<TActor> : IIndirectActorProducer where TActor : ActorBase
```
mhbuck, line 649: Renamed to be more inline with the interface.

src/core/Akka/Actor/Props.cs
```diff
@@ -687,23 +645,22 @@ public ActivatorProducer(Type actorType, object[] args)
 
             public ActorBase Produce()
             {
-                return Activator.CreateInstance(ActorType, _args).AsInstanceOf<ActorBase>();
+                return (ActorBase)Activator.CreateInstance(ActorType, _args);
```
mhbuck, line 638: Not sure if there is a preference here.

src/core/Akka/Actor/Props.cs
```diff
@@ -499,12 +477,13 @@ public Props WithStashCapacity(int stashCapacity)
         public Props WithDeploy(Deploy deploy)
         {
             var copy = Copy();
-            var original = copy.Deploy;
+            //var original = copy.Deploy;
```
mhbuck, line 490: Not needed as the todo below has commented it out.

mhbuck (commented):
Still not made movement on the `Create<TActor>(params object[] args)` signature but interesting that so much is passing currently

src/core/Akka/Actor/Props.cs
Aaronontheweb: I removed it from https://github.com/akkadotnet/akka.net/pull/7903 - it's a legacy function call that we probably don't need any more.

Aaronontheweb:
Closing after measurement: under .NET 10 Native AOT the `Props.Create(Expression<Func<T>>)` path already runs on the LINQ interpreter with zero IL warnings (the only IL3050 was the non-generic `Expression.Lambda` fallback in `ExpressionExtensions`, fixed in #8605), so replacing the expression with a delegate is not needed for AOT — and doing so empties `Props.Arguments`, which remote deployment serializes. The AOT milestone-1 stack starts at #8601; the Props work there is annotations only. Thanks for the investigation, @mhbuck.

PR #7579, opened 2025-03-24
Fix #7578, Cancelled sinks are blocking other BroadcastHub consumers

### LINKED ISSUE #7578: Cancelled sinks are blocking other `BroadcastHub` consumers
**Version Information**
Version of Akka.NET? 1.5.40
Which Akka.NET Modules? Streams

**Describe the bug**

Cancelled sinks are blocking other `BroadcastHub` consumers. This is the same exact `BroadcastHub` problem as described here: https://github.com/akka/akka/issues/23205#issuecomment-338641510

**To Reproduce**

```csharp
[Fact]
public async Task BroadcastHub_must_handle_cancelled_Sink()
{
    await this.AssertAllStagesStoppedAsync(async () =>
    {
        var upstream = this.CreatePublisherProbe<int>();
        var hubSource = Source.FromPublisher(upstream).RunWith(BroadcastHub.Sink<int>(4), Materializer);
        var downstream = this.CreateSubscriberProbe<int>();
        
        hubSource.RunWith(Sink.Cancelled<int>(), Materializer);
        hubSource.RunWith(Sink.FromSubscriber(downstream), Materializer);
        
        await downstream.EnsureSubscriptionAsync();
        
        await downstream.RequestAsync(10);
        await upstream.ExpectRequestAsync();
        await upstream.SendNextAsync(1);
        await downstream.ExpectNextAsync(1);
        await upstream.SendNextAsync(2);
        await downstream.ExpectNextAsync(2);
        await upstream.SendNextAsync(3);
        await downstream.ExpectNextAsync(3);
        await upstream.SendNextAsync(4);
        await downstream.ExpectNextAsync(4);
        await upstream.SendNextAsync(5);
        await downstream.ExpectNextAsync(5);
        
        await upstream.SendCompleteAsync();
        await downstream.ExpectCompleteAsync();
    }, Materializer);
}
```

**Expected behavior**

`downstream` should receive the 5th element

**Actual behavior**

Timeout while waiting for message exception on the `await downstream.ExpectNextAsync(5);` line.

Arkatufus:
close #7578

## Changes

* Add reproduction unit test
* Port https://github.com/akka/akka/pull/23185 except for fixes to `QueueSource`
* Port https://github.com/akka/akka/pull/24073
* Port https://github.com/akka/akka/pull/23970
* Port unit tests


src/core/Akka.Streams.Tests/Dsl/HubSpec.cs
```diff
@@ -614,6 +614,38 @@ await Awaiting(async () =>
             }, Materializer);
         }
 
+        [Fact]
+        public async Task BroadcastHub_must_handle_cancelled_Sink()
```
Arkatufus, line 619: The actual reproduction

src/core/Akka.Streams.Tests/Implementation/Fusing/AsyncCallbackSpec.cs
```diff
@@ -0,0 +1,330 @@
...
+using Akka.TestKit;
+using Akka.Util;
+using Xunit;
+using Xunit.Abstractions;
+using FluentAssertions;
+using static FluentAssertions.FluentActions;
+
+namespace Akka.Streams.Tests.Implementation.Fusing;
+
+public class AsyncCallbackSpec: AkkaSpec
```
Arkatufus, line 24: These are specs for the new async callback behavior, two are currently skipped because the related new behavior hasn't been ported yet

src/core/Akka.Streams/Dsl/Hub.cs
```diff
@@ -542,67 +542,16 @@ private sealed class RegistrationPending : IHubEvent
...
-        {
-            public NeedWakeup(long id, int previousOffset, int currentOffset)
-            {
-                Id = id;
-                PreviousOffset = previousOffset;
-                CurrentOffset = currentOffset;
-            }
-
-            public long Id { get; }
+        private sealed record UnRegister(long Id, int PreviousOffset, int FinalOffset) : IHubEvent;
```
Arkatufus, line 548: Cleanup, convert classes to records if they're not a singleton

src/core/Akka.Streams/Dsl/Hub.cs
```diff
@@ -747,69 +659,101 @@ public override void OnPush()
 
             private void OnEvent(IHubEvent hubEvent)
             {
-                if (hubEvent == RegistrationPending.Instance)
+                switch (hubEvent)
```
Arkatufus, line 662: Refactor cascading if...else if statements to a switch statement

src/core/Akka.Streams/Dsl/Hub.cs
```diff
@@ -747,69 +659,100 @@ public override void OnPush()
...
+                        var open = (Open)State.GetAndSet(_noRegistrationState);
+                        foreach (var consumer in open.Registrations)
+                        {
+                            var startFrom = _head;
+                            _activeConsumer++;
+                            AddConsumer(consumer, startFrom);
+                            
+                            // in case the consumer is already stopped we need to undo registration
+                            consumer.Callback(new Initialize(startFrom))
+                                .ContinueWith(t =>
```
Arkatufus, line 675: New code, the `consumer.Callback()` uses the new API that returns a `Func<T, Task>` instead of the old `Action<T>` that allows us to respond to the result of the callback. In this case, if the stream is dead, we unregister the consumer.

src/core/Akka.Streams/Dsl/Hub.cs
```diff
@@ -747,69 +659,100 @@ public override void OnPush()
...
-                    if (_activeConsumer == 0)
+                        return;
+                    }
+                    case UnRegister unregister:
                     {
-                        if (IsClosed(_stage.In))
-                            CompleteStage();
-                        else if (_head != unregister.FinalOffset)
+                        if(FindAndRemoveConsumer(unregister.Id, unregister.PreviousOffset) is not null)
+                            _activeConsumer--;
```
Arkatufus, line 708: Fix active consumer counter tracking. We only decrease the active consumer count if it is actually registered.

src/core/Akka.Streams/Dsl/Hub.cs
```diff
@@ -989,7 +932,7 @@ public Logic(HubSourceLogic stage, long id) : base(stage.Shape)
 
                 public override void PreStart()
                 {
-                    var callback = GetAsyncCallback<IConsumerEvent>(OnCommand);
+                    var callback = GetAsyncCallbackAsync<IConsumerEvent>(OnCommand);
```
Arkatufus: New async callback API call that returns a `Func<T, Task>`

src/core/Akka.Streams/Implementation/Fusing/ActorGraphInterpreter.cs
```diff
@@ -694,36 +695,18 @@ public ExposedPublisher(GraphInterpreterShell shell, int id, IActorPublisher pub
             public GraphInterpreterShell Shell { get; }
         }
 
-        /// <summary>
-        /// TBD
-        /// </summary>
```
Arkatufus, line 699: TBD cleanup

src/core/Akka.Streams/Implementation/Fusing/ActorGraphInterpreter.cs
```diff
@@ -694,36 +695,18 @@ public ExposedPublisher(GraphInterpreterShell shell, int id, IActorPublisher pub
...
-            /// <param name="event">TBD</param>
-            /// <param name="handler">TBD</param>
-            /// <returns>TBD</returns>
-            public AsyncInput(GraphInterpreterShell shell, GraphStageLogic logic, object @event, Action<object> handler)
+            public AsyncInput(GraphInterpreterShell shell, GraphStageLogic logic, object @event, TaskCompletionSource<Done> promise, Action<object> handler)
             {
                 Shell = shell;
                 Logic = logic;
                 Event = @event;
+                Promise = promise;
```
Arkatufus, line 709: We now include the `TaskCompletionSource` "promise" inside the `AsyncInput` struct.

src/core/Akka.Streams/Implementation/Fusing/GraphInterpreter.cs
```diff
@@ -819,9 +814,16 @@ public void RunAsyncInput(GraphStageLogic logic, object evt, Action<object> hand
                     try
                     {
                         handler(evt);
+                        if (!ReferenceEquals(promise, GraphStageLogic.NoPromise))
+                        {
+                            promise.TrySetResult(Done.Instance);
+                            logic.OnFeedbackDispatched(promise);
+                        }
```
Arkatufus, line 821: If an active promise is passed in, we set its completion here.

src/core/Akka.Streams/Implementation/Fusing/GraphInterpreter.cs
```diff
@@ -819,9 +814,16 @@ public void RunAsyncInput(GraphStageLogic logic, object evt, Action<object> hand
...
                     }
                     catch (Exception e)
                     {
+                        if (!ReferenceEquals(promise, GraphStageLogic.NoPromise))
+                            promise.TrySetException(e);
```
Arkatufus: If an active promise is passed in, we set its failure here.

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -423,6 +425,16 @@ public Timer(int id, ICancelable task)
     /// </summary>
     public abstract class GraphStageLogic : IStageLogging
     {
+        private List<Func<object, Task>>? CallbacksWaitingForInterpreter = null;
```
Arkatufus: Still need to implement this.

Technically, callbacks can be invoked early before the interpreter is initialized. This should cache these early callbacks and defer their invocation until the interpreter is initialized.

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -423,6 +425,16 @@ public Timer(int id, ICancelable task)
...
+        private AtomicReference<List<TaskCompletionSource<Done>>?> AsyncCallbacksInProgress = new(new List<TaskCompletionSource<Done>>());
```
Arkatufus: This holds the current async callback invocations that are currently in progress. Note that it is a nullable atomic reference to a list of `TaskCompletionSource`s. We use null to signal that the stream is shutting down or in the process of doing so, all callback invocations will automatically trigger a `StreamDetachedException` if it sees that this list is null. We need this because we need to guarantee cross-thread access safety (async callbacks are invoked from outside the stream)

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -423,6 +425,16 @@ public Timer(int id, ICancelable task)
...
+        public static readonly TaskCompletionSource<Done> NoPromise;
```
Arkatufus, line 430: Marker for callbacks that doesn't need to observe any task completion (fire and forget)

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -1683,8 +1695,77 @@ protected Action<T> GetAsyncCallback<T>(Action<T> handler)
         /// <param name="handler">TBD</param>
         /// <returns>TBD</returns>
         protected Action GetAsyncCallback(Action handler)
-            => () => Interpreter.OnAsyncInput(this, NotUsed.Instance, _ => handler());
+            => () => Interpreter.OnAsyncInput(this, NotUsed.Instance, NoPromise, _ => handler());
 
+        protected Func<T, Task<Done>> GetAsyncCallbackAsync<T>(Action<T> handler)
```
Arkatufus: New async callback API, instead of returning an `Action<T>`, this new API returns a `Func<T, Task<Done>>` where external callers can observe the result of any async callback invocations.

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -1683,8 +1695,77 @@ protected Action<T> GetAsyncCallback<T>(Action<T> handler)
...
+        {
+            return @event =>
+            {
+                var promise = new TaskCompletionSource<Done>();
+                if (AddToWaiting(promise))
+                {
+                    Interpreter.OnAsyncInput(this, @event, promise, x => handler((T)x));
+                }
```
Arkatufus: We create a new `TaskCompletionSource` "promise" for each async callback invocation. `AddToWaiting()` checks that the promise can be atomically added to `AsyncCallbacksInProgress`. If the add succeed, we enqueue its invocation in the interpreter, if not, we signal a `StreamDetachedException` on the promise.

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -1746,6 +1827,35 @@ protected internal virtual void AfterPostStop()
                 _stageActor.Stop();
                 _stageActor = null;
             }
+
+            var inProgress = AsyncCallbacksInProgress.GetAndSet(null);
+            if (inProgress is not null && inProgress.Count > 0)
```
Arkatufus: On stream stop, we set `AsyncCallbacksInProgress` to null to signal that we're not accepting any async callback anymore, then for each active callbacks, we signal a `StreamDetachedException` failure.

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -1746,6 +1827,35 @@ protected internal virtual void AfterPostStop()
...
+        {
+            switch (AsyncCallbacksInProgress.Value)
+            {
+                case null:
+                    // already finished, nothing to do here
+                    break;
+                case var x:
+                    x.Remove(p);
+                    break;
+            }
```
Arkatufus: When interpreter finished processing the callback, it signals the logic (this instance) through this callback. The logic then removes the `TaskCompletionSource` from the pending list.

Arkatufus (commented):
self-review

src/core/Akka.Streams/Stage/GraphStage.cs
Arkatufus: Implemented.

src/core/Akka.Streams.Tests/Implementation/Fusing/AsyncCallbackSpec.cs
Arkatufus: New behavior ported, both skipped unit test is working.

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -423,6 +425,16 @@ public Timer(int id, ICancelable task)
     /// </summary>
     public abstract class GraphStageLogic : IStageLogging
     {
+        private readonly Queue<ActorGraphInterpreter.AsyncInput> _callbacksWaitingForInterpreter = new();
```
Arkatufus: Queue to hold early callbacks that were invoked before the interpreter is initialized

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -423,6 +425,16 @@ public Timer(int id, ICancelable task)
...
+        private readonly AtomicReference<List<TaskCompletionSource<Done>>?> _asyncCallbacksInProgress = new(new List<TaskCompletionSource<Done>>());
```
Arkatufus: This holds the current async callback invocations that are currently in progress. Note that it is a nullable atomic reference to a list of TaskCompletionSources. We use null to signal that the stream is shutting down or in the process of doing so, all callback invocations will automatically trigger a StreamDetachedException if it sees that this list is null. We need this because we need to guarantee cross-thread access safety (async callbacks are invoked from outside the stream)

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -1669,7 +1689,14 @@ protected void PassAlong<TOut, TIn>(Inlet<TIn> from, Outlet<TOut> to, bool doFin
...
         protected Action<T> GetAsyncCallback<T>(Action<T> handler)
-            => @event => Interpreter.OnAsyncInput(this, @event, x => handler((T)x));
+            => @event =>
+            {
+                if(_interpreter == null)
+                    _callbacksWaitingForInterpreter.Enqueue(
+                        new ActorGraphInterpreter.AsyncInput(null, null, @event, NoPromise, x => handler((T)x)));
+                else
+                    Interpreter.OnAsyncInput(this, @event, NoPromise, x => handler((T)x));
+            };
```
Arkatufus: Example of how the invocation is stored using the existing `AsyncInput` data structure to be defered for later. The actual execution will be performed when `Interpreter` property is assigned.

src/core/Akka.Streams/Stage/GraphStage.cs
```diff
@@ -818,7 +830,15 @@ internal GraphInterpreter Interpreter
...
             }
-            set => _interpreter = value;
+            set
+            {
+                _interpreter = value;
+                while (_callbacksWaitingForInterpreter.Count > 0)
+                {
+                    var input = _callbacksWaitingForInterpreter.Dequeue();
+                    _interpreter.OnAsyncInput(this, input.Event, input.Promise, input.Handler);
+                }
```
Arkatufus: All defered callback invocation are executed here.

Arkatufus (commented):
self-review

Aaronontheweb (approved):
Reviewed this together on the call this morning, LGTM

Aaronontheweb (changes requested):
There are still test failures

Aaronontheweb:
Resolved via https://github.com/akkadotnet/akka.net/pull/7615

PR #7595, opened 2025-04-10
Feature/7582

### LINKED ISSUE #7582: Ability to specifically detect persistence journal and/or snapshot initialisation failures
There are various reasons why initialisation of a journal or snapshot store might fail. This could be due to external issues. For example, the service being unavailable. 
But it could also be due to local issues caused by the client application itself. Such as misconfigured connection details, or perhaps some wierd Assembly load issue if your still dealing with netframework or perhaps targetting specific platforms and are reliant on specific DB client assemblies for that platform. (Looking at you sqlite).

Currently when a journal fails to initialize. Its assumed its a transient problem and Initialisation is immediately retried.
Ofcourse if its a misconfigured connectionstring of sorts. It will never recover. 

Which was a scenario i ran into recently. It resulted in over 800 million trace log records in our logging infra. And warnings about a massive increase in garbage collector activity. Which, considering the journal is basically in an infinite do while loop, isn't strange.

I think it would be a worthwhile feature, to be able to have the client application specifically detect initialization failures of the journal and snapshot stores.

This will allow the application to make decisions about what it wants to do. Wait for it to recover automatically. Or do something else. Which in my case would be to let the error be logged (which it currently already does). and crash the application if i see its not a recoverable problem.
Reason for that is that if the journal gets in an unrecoverable state, there is no use trying to work with any persistent actor anyway and it possibly will just send me on a wild goose chase as to why an actor isn't working. While its the journal itself thats not working.

So this feature falls into the realm of: being able to verify that the application is configured properly and is expected to run accordingly.

**What would possibly look like**
It would be great if we can create a solution that means the persistence implementations do not have to accommodate for this, and its handled/provided out of the box by the persistence abstractions themselves. But im not sure if thats feasable.

In essense i also dont think that it would have to be a complicated feature. It could be as simple as publishing a dedicated message on the internal EventStream, and let clients subscribe to that. But maybe we can also provide a function callback api somewhere in the persistence hosting api surface. Im not sure what the challenges would be for that.

Ill add on more idea's or suggestions when i have them.

Danthar:
## Changes
Added a facility to provide your own supervision strategy that is used to govern the journal and snapshot-store plugins

This is a WIP. I need to probably add a configuration Spec, and some crash behavior tests at the very least.

Fixes #7582

## Checklist

* [x] Design discussion issue #7582

### Latest `dev` Benchmarks 

not relevant

### This PR's Benchmarks

not relevant


Danthar:
Although the way this supervision strategy configuration method will work. Sadly we dont enforce some kind of Init callback for our snapshot or journal implementations. If we did, we could force all implementations to handle their setup in there, stuff like initializing connections and whatnot. Then if that fails we could provide out of the box detection for init failures and use custom exceptions to explicitly communicate initialization failure. 

But that would be a major breaking change for persistence plugin implementations.

Arkatufus (approved):
LGTM, waiting for the CI/CD to go through

Aaronontheweb:
Well done @Danthar - ty for your contribution

PR #7596, opened 2025-04-11
Mark IDaemonMsg with public interface marker

### LINKED ISSUE #89: Remote deployment fails when there is a current Activity
For unknown reasons (maybe because the Akka.Remote.DaemonMsgCreate is wrapped in a Phobos.Tracing.SpanEnvelope?) remote deployment of an actor fails silently when using Phobos & when there is a current Activity.

The issue does not occur when there is no current activity (such as when manually setting Activity.Current = null).

A reproduction is provided here: [Link](https://github.com/Xeiotos/AkkaRemoteDeploymentDemo)

I suspect it's some kind of serialization issue where the EndpointWriter/EndpointReader does not realise this is a remote deployment, but I'm not sure.

Arkatufus:
Fixes https://github.com/petabridge/phobos-issues/issues/89

## Changes

Mark IDaemonMsg with public interface marker


Aaronontheweb:
@Arkatufus need API approval

Aaronontheweb (approved):
LGTM

PR #7604, opened 2025-04-22
Fix HTML content on homepage

Aaronontheweb:
DocFx upgrade started treating this sections as markdown rather than HTML


Aaronontheweb:
superseded via #7605

PR #7607, opened 2025-04-28
[Streams] Allow GroupBy with infinite output substreams

### LINKED ISSUE #7514: Akka.Streams: `GroupBy` shouldn't require users to pre-calculate max number of groups
**Version Information**
Version of Akka.NET? v1.5.38

**Describe the bug**

So referencing my code sample from https://github.com/akkadotnet/akka.net/issues/7512 again, which you can run here: https://share.linqpad.net/bkgp72uf.linq

```csharp
// Generate a stable set of data records.
var records = GenerateRecords(numEntities: 50, numPeriods: 10, recordsPerPeriod: 3);

// Build the stream graph.
// Group by EntityId with parallelism of 5.
var stream = Source
	.From(records)
	.GroupBy(5, r => r.EntityId)
```

This will fail with a `TooManySubsStreamsException` - because my test data set has 50 unique `EntityId`s but I've only specified a maximum of `5` substreams. This seems like a really poor design choice to me - why should users have to pre-calculate how many groups their streams are going to be processed in advance?

A much better approach would be for it to work the way my hack does:

```csharp
.GroupBy(10, r => r.EntityId.GetHashCode() % 5)
```

Rather than blow up with a stupid exception, just shove the entities into the appropriate grouping as a partitioning strategy instead.

**Additional context**

If people really depend on / want to keep the current `GroupBy` behavior - fine, we can keep it even though I think it sucks. Maybe I'll call this `.Partition` instead (and there's another equally bad stream stage with that name too.)

Arkatufus:
Fixes #7514

## Changes

* Allows GroupBy to have infinite substreams by passing negative values to `maxSubstreams` argument.
* Add new public `Flow` and `Source` fluent API that does not require the `maxSubstreams` argument.

## Checklist


src/core/Akka.Streams/Implementation/Fusing/StreamOfStreams.cs
```diff
@@ -419,9 +419,9 @@ public void OnPush()
                     }
                     else
                     {
-                        if (_activeSubstreams.Count + _closedSubstreams.Count == _stage._maxSubstreams)
+                        if (_stage._maxSubstreams > -1 && _activeSubstreams.Count + _closedSubstreams.Count == _stage._maxSubstreams)
```
Arkatufus, line 422: Actual fix. Only check substream overflow condition if `_maxSubstream` is greater than or equals to 0.

Arkatufus (commented):
Self-review

src/core/Akka.Streams/Dsl/FlowOperations.cs
```diff
@@ -1366,9 +1423,38 @@ public static SubFlow<TOut, TMat, Sink<TIn, TMat>> GroupBy<TIn, TOut, TMat, TKey
...
+        /// a new substream is opened and subsequently fed with all elements belonging to
+        /// that key.
+        /// <para>
+        /// WARNING: The stage keeps track of all keys of streams that have already been closed.
+        /// If you expect an infinite number of keys this can cause memory issues. Elements belonging
+        /// to those keys are drained directly and not send to the substream.
+        /// </para>
+        /// See <seealso cref="GroupBy{TIn, TOut, TMat, TKey}(Flow{TIn, TOut, TMat}, int, Func{TOut, TKey}, bool)"/>
+        /// </summary> 
+        /// <typeparam name="TIn">TBD</typeparam>
```
Aaronontheweb, line 1449: @Arkatufus just a general note going forward, just don't have a comment instead of a TBD.

Aaronontheweb (approved):
LGTM

PR #7608, opened 2025-04-29
Fix Akka.Persistence.Custom bug

### LINKED ISSUE #7599: Akka.Persistence.Custom.Tests have several failures
**Version Information**
Version of Akka.NET? current `dev`
Which Akka.NET Modules? Akka.Persistence.Custom

**Describe the bug**

Discovered while working on https://github.com/akkadotnet/akka.net/pull/7598 - tons of these tests are just non-operable and will fail 100% of the time locally. This test suite either needs to be updated or deleted. For now, we're just skipping these specs just like we were with FAKE previously.

**Expected behavior**

Tests should pass or be deleted if not relevant.

**Actual behavior**

Test fail violently, both locally and in CI. Had to skip them in Incrementalist, just like we did with FAKE.

**Environment**

Fails with constant DLL not found exceptions on .NET Framework / Windows, but has stable failures across all platforms

**Additional context**

It looks like this was an accident @Arkatufus added this library back in 2022 to create an example we could reference in the docs to demonstrate how to create custom persistence plugins. My guess is that the tests have never worked because I'm pretty sure the default globbing inside FAKE _always_ excluded the `/samples` project even before 2022.

Arkatufus:
Fixes #7599

## Changes

Fix bug in Akka.Persistence.Custom SQL string

### DIFF
--- a/src/examples/Akka.Persistence.Custom/Snapshot/SqliteSnapshotStore.cs
+++ b/src/examples/Akka.Persistence.Custom/Snapshot/SqliteSnapshotStore.cs
@@ -315,7 +315,7 @@ namespace Akka.Persistence.Custom.Snapshot
                     : (DateTime?)null;
                 
                 var sql = timestamp.HasValue
-                    ? DeleteSnapshotRangeSql + " AND { Configuration.TimestampColumnName} = @Timestamp"
+                    ? DeleteSnapshotRangeSql + " AND timestamp = @Timestamp"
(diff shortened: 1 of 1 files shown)


Aaronontheweb (approved):
LGTM - do you also want to re-enable these specs inside Incrementalist?

Arkatufus:
Nah, they're also considered as sample test project, they're not important for Akka.

PR #7612, opened 2025-04-30
Add xunit.v3 implementation

### LINKED ISSUE #7603: Akka.TestKit.Xunit Xunit3 support

Xunit3 is out - we should probably add support for it.

IMHO - bump the Akka.TestKit.Xunit package to support it and leave the Xunit2 package as-is.

We could ship a new Xunit3 package instead - that leaves the issue of what do about the old Xunit package though.

Arkatufus:
Fixes #7603

## Changes

* Add xunit.v3 support
* Akka.TestKit.Xunit now targets xunit.v3
* Akka.TestKit.Xunit2 still targets xunit v2

## Limitations

* Akka.TestKit.Xunit only supports net472 and net8.0 framework targets (xunit 3 limitation)
* Akka.TestKit.Xunit only works with xunit.runner.visualstudio v3.0.2 and above (xunit 3 limitation)


src/contrib/testkits/Akka.TestKit.Xunit.Tests/Akka.TestKit.Xunit.Tests.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="../../../xunitSettings.props" />
+
+  <PropertyGroup>
+    <TargetFrameworks>$(NetFrameworkTestVersion);net8.0</TargetFrameworks>
```
Aaronontheweb, line 5: Well, that's going to be a bit of a breaking change but you know, at some point we have to step into the future here.

Aaronontheweb (approved):
LGTM

PR #7614, opened 2025-05-01
leverage exception when

SimonCropp:
avoid catching exceptions, and hence stack trace, when they are re-thrown based on logic


Aaronontheweb:
Very cool - I didn't even know about this language feature.

PR #7623, opened 2025-05-06
Akka.IO: fix `TcpListener` connection queue problem

Aaronontheweb:
## Changes

close https://github.com/akkadotnet/akka.net/issues/5988

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [x] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [x] Design discussion issue #5988
* [x] Changes in public API reviewed, if any.

### Latest `dev` Benchmarks 

```                                                                                                                                       
                                                                                                                                          
BenchmarkDotNet v0.13.12, Pop!_OS 22.04 LTS                                                                                               
13th Gen Intel Core i7-1360P, 1 CPU, 16 logical and 12 physical cores                                                                     
.NET SDK 8.0.404                                                                                                                          
  [Host]  : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2                                                                                 
  LongRun : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2                                                                                 
                                                                                                                                          
Job=LongRun  Concurrent=True  Server=True                                                                                                 
InvocationCount=1  IterationCount=10  LaunchCount=3                                                                                       
RunStrategy=Monitoring  UnrollFactor=1  WarmupCount=3                                                                                     
                                                                                                                                          
```                                                                                                                                       
| Method                    | MessageLength | ClientsCount | Mean      | Error     | StdDev    | Req/sec    |                             
|-------------------------- |-------------- |------------- |----------:|----------:|----------:|-----------:|                             
| **ClientServerCommunication** | **10**            | **1**            | **27.982 μs** | **1.2878 μs** | **1.9275 μs** |  **35,737.37** | 
| **ClientServerCommunication** | **10**            | **3**            | **14.090 μs** | **2.1295 μs** | **3.1874 μs** |  **70,974.46** | 
| **ClientServerCommunication** | **10**            | **5**            | **10.736 μs** | **1.0059 μs** | **1.5055 μs** |  **93,147.14** | 
[...]

PR #7637, opened 2025-05-13
Akka.IO: redesign `TcpConnection`

### LINKED ISSUE #7634: Akka.IO: `TcpConnection` does not propagate closure events reliably
**Version Information**
Version of Akka.NET? v1.5.41

**Describe the bug**

Yet another Akka.IO bug - while working on https://github.com/akkadotnet/akka.net/pull/7633 I tried to tests to verify that we can track connection closures, but there's a small problem: the `TcpConnection` actor doesn't reliably transmit this information.

**Expected behavior**

If the remote socket gets closed, the local socket should close and report back to its handler immediately.

**Actual behavior**

Wildly different stuff depending on how the socket was closed. Sometimes an event, usually not.

**Additional context**

This code is wildly over-engineered and is in drastic need of cleanup.

Aaronontheweb:
## Changes

Addresses a significant number of reliability and performance issues with Akka.IO.Tcp

close #7634

Still need to add:

- [x] Pull mode for `TcpConnection` - Akka.Streams.IO blows up without this
- [x] Pull mode for `TcpListener` - same deal
- [x] Better buffer management code for `TcpConnection` writing - if someone sends a message larger than the max frame size we'll hang currently. Better to paginate it.
- [x] Need to restore the max-queue-size as byte-size semantics in our tests. I was against this initially but IMHO I think it's probably the more sensible "real measure" of resources

## Checklist

* [x] Design discussion issue #7634

### Latest `dev` Benchmarks 

```

BenchmarkDotNet v0.13.12, Windows 11 (10.0.26100.3775)
12th Gen Intel Core i7-1260P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 8.0.403
  [Host]  : .NET 8.0.14 (8.0.1425.11118), X64 RyuJIT AVX2
  LongRun : .NET 8.0.14 (8.0.1425.11118), X64 RyuJIT AVX2

Job=LongRun  Concurrent=True  Server=True  
InvocationCount=1  IterationCount=10  LaunchCount=3  
RunStrategy=Monitoring  UnrollFactor=1  WarmupCount=3  

```
| Method                    | MessageLength | ClientsCount | Mean      | Error     | StdDev    | Req/sec    |
|-------------------------- |-------------- |------------- |----------:|----------:|----------:|-----------:|
| **ClientServerCommunication** | **10**            | **1**            | **53.038 μs** | **2.3749 μs** | **3.5546 μs** |  **18,854.32** |
| **ClientServerCommunication** | **10**            | **3**            | **22.307 μs** | **0.9668 μs** | **1.4470 μs** |  **44,829.77** |
| **ClientServerCommunication** | **10**            | **5**            | **15.343 μs** | **1.7903 μs** | **2.6797 μs** |  **65,177.89** |
| **ClientServerCommunication** | **10**            | **7**            | **13.806 μs** | **1.5307 μs** | **2.2911 μs** |  **72,433.14** |
| **ClientServerCommunication** | **10**            | **10**           | **11.569 μs** | **0.9201 μs** | **1.3771 μs** |  **86,436.51** |
| **ClientServerCommunication** | **10**            | **20**           |  **9.188 μs** | **0.8901 μs** | **1.3322 μs** | **108,841.90** |
| **ClientServerCommunication** | **10**            | **30**           |  **9.048 μs** | **0.8720 μs** | **1.3051 μs** | **110,519.01** |
| **ClientServerCommunication** | **10**            | **40**           | **10.213 μs** | **0.4584 μs** | **0.6861 μs** |  **97,915.46** |
| **ClientServerCommunication** | **100**           | **1**            | **57.817 μs** | **5.1364 μs** | **7.6880 μs** |  **17,296.00** |
| **ClientServerCommunication** | **100**           | **3**            | **21.813 μs** | **0.9598 μs** | **1.4365 μs** |  **45,843.55** |
| **ClientServerCommunication** | **100**           | **5**            | **16.136 μs** | **1.3610 μs** | **2.0370 μs** |  **61,974.70** |
| **ClientServerCommunication** | **100**           | **7**            | **14.732 μs** | **2.3724 μs** | **3.5510 μs** |  **67,878.50** |
| **ClientServerCommunication** | **100**           | **10**           | **11.896 μs** | **1.4770 μs** | **2.2107 μs** |  **84,058.77** |
| **ClientServerCommunication** | **100**           | **20**           |  **9.574 μs** | **0.8134 μs** | **1.2174 μs** | **104,445.29** |
| **ClientServerCommunication** | **100**           | **30**           |  **9.879 μs** | **0.7771 μs** | **1.1631 μs** | **101,222.03** |
| **ClientServerCommunication** | **100**           | **40**           |  **9.455 μs** | **0.9147 μs** | **1.3691 μs** | **105,759.12** |

### This PR's Benchmarks

```
BenchmarkDotNet v0.13.12, Windows 11 (10.0.26100.3775)
12th Gen Intel Core i7-1260P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 8.0.403
  [Host]  : .NET 8.0.14 (8.0.1425.11118), X64 RyuJIT AVX2
  LongRun : .NET 8.0.14 (8.0.1425.11118), X64 RyuJIT AVX2

Job=LongRun  Concurrent=True  Server=True
InvocationCount=1  IterationCount=10  LaunchCount=3
RunStrategy=Monitoring  UnrollFactor=1  WarmupCount=3
```

| Method                    | MessageLength | ClientsCount | Mean      | Error     | StdDev    | Req/sec    |
|-------------------------- |-------------- |------------- |----------:|----------:|----------:|-----------:|
| ClientServerCommunication | 10            | 1            | 44.998 us | 1.9392 us | 2.9025 us |  22,223.07 |
| ClientServerCommunication | 10            | 3            | 20.735 us | 1.1877 us | 1.7777 us |  48,227.95 |
| ClientServerCommunication | 10            | 5            | 15.255 us | 1.2650 us | 1.8934 us |  65,550.70 |
| ClientServerCommunication | 10            | 7            | 13.271 us | 0.9893 us | 1.4807 us |  75,351.45 |
| ClientServerCommunication | 10            | 10           | 10.537 us | 1.0431 us | 1.5613 us |  94,905.86 |
| ClientServerCommunication | 10            | 20           |  8.821 us | 0.6686 us | 1.0007 us | 113,359.97 |
| ClientServerCommunication | 10            | 30           |  8.886 us | 0.5094 us | 0.7624 us | 112,542.02 |
| ClientServerCommunication | 10            | 40           |  9.362 us | 0.8531 us | 1.2769 us | 106,809.83 |
| ClientServerCommunication | 100           | 1            | 43.620 us | 3.1453 us | 4.7078 us |  22,925.43 |
| ClientServerCommunication | 100           | 3            | 20.511 us | 1.1551 us | 1.7288 us |  48,754.38 |
| ClientServerCommunication | 100           | 5            | 15.767 us | 1.9737 us | 2.9541 us |  63,425.26 |
| ClientServerCommunication | 100           | 7            | 13.081 us | 1.4413 us | 2.1573 us |  76,444.56 |
| ClientServerCommunication | 100           | 10           | 10.298 us | 0.7608 us | 1.1388 us |  97,105.06 |
| ClientServerCommunication | 100           | 20           |  9.168 us | 0.8832 us | 1.3219 us | 109,070.42 |
| ClientServerCommunication | 100           | 30           |  8.572 us | 0.6353 us | 0.9509 us | 116,655.03 |
| ClientServerCommunication | 100           | 40           |  9.378 us | 1.0401 us | 1.5568 us | 106,630.69 |


Aaronontheweb:
The most significant difference in performance here is the single client test, which accurately measures single socket read / write performance - we went from 18k to 22k operations per second there, an improvement of ~18%

Aaronontheweb:
Currently blocked on some original design issues - basically Akka.IO is missing some basic features you want for real TCP server applications, namely:

- Explicit system buffer sizes
- Explicit max frame sizes

The other issue I have with the settings code, in particular, is the question of how large the write buffer should be allowed to grow. In Akka.Remote we don't really constraint this at all outside of capping the number of `ISystemMessage`s that can reside in the reliable delivery queue.

Aaronontheweb:
So apparently `SocketOptions` exposes a way to configure send and receive buffer sizes already, but this is totally undocumented and not obvious. Not sure how we want to address that because nowhere are any system defaults of any kind documented - and these settings have to be squared against the maximum frame size too.

Aaronontheweb:
Now that https://github.com/akkadotnet/akka.net/pull/7651 has been merged, we need to get some new numbers

Aaronontheweb:
Benchmark numbers using the changes from #7651 

## `dev`

```
BenchmarkDotNet v0.13.12, Pop!_OS 22.04 LTS                                                                                                
13th Gen Intel Core i7-1360P, 1 CPU, 16 logical and 12 physical cores                                                                      
.NET SDK 8.0.404                                                                                                                           
  [Host]  : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2                                                                                  
  LongRun : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2                                                                                  
                                                                                                                                           
Job=LongRun  Concurrent=True  Server=True                                                                                                  
InvocationCount=1  IterationCount=10  LaunchCount=3                                                                                        
RunStrategy=Monitoring  UnrollFactor=1  WarmupCount=3                                                                                      
```
                                                                                                                                           
| Method                    | MessageLength | ClientsCount | Mean       | Error     | StdDev    | Req/sec      | Gen0   | Allocated |      
|-------------------------- |-------------- |------------- |-----------:|----------:|----------:|-------------:|-------:|----------:|      
| ClientServerCommunication | 10            | 1            | 5,135.3 ns | 236.85 ns | 354.51 ns |   194,731.42 | 0.0060 |    1206 B |      
| ClientServerCommunication | 10            | 3            | 1,478.6 ns | 111.37 ns | 166.69 ns |   676,327.57 | 0.0060 |    1221 B |      
| ClientServerCommunication | 10            | 5            | 1,124.8 ns | 128.99 ns | 193.06 ns |   889,068.01 | 0.0060 |    1170 B |      
| ClientServerCommunication | 10            | 7            |   825.4 ns | 104.25 ns | 156.03 ns | 1,211,471.51 | 0.0060 |    1150 B |      
| ClientServerCommunication | 10            | 10           |   639.7 ns | 113.02 ns | 169.16 ns | 1,563,153.36 | 0.0060 |    1102 B |      
| ClientServerCommunication | 10            | 20           |   371.2 ns |  31.36 ns |  46.94 ns | 2,694,289.82 | 0.0050 |     998 B |      
| ClientServerCommunication | 10            | 30           |   414.9 ns |  65.05 ns |  97.37 ns | 2,410,263.95 | 0.0050 |     963 B |      
| ClientServerCommunication | 10            | 40           |   421.7 ns |  68.71 ns | 102.84 ns | 2,371,095.25 | 0.0050 |     955 B |      
| ClientServerCommunication | 100           | 1            | 5,526.0 ns | 359.47 ns | 538.04 ns |   180,961.13 | 0.0070 |    1437 B |      
| ClientServerCommunication | 100           | 3            | 1,672.0 ns | 156.09 ns | 233.63 ns |   598,092.81 | 0.0080 |    1433 B |      
| ClientServerCommunication | 100           | 5            | 1,204.7 ns | 150.04 ns | 224.57 ns |   830,072.77 | 0.0070 |    1431 B |      
| ClientServerCommunication | 100           | 7            |   869.2 ns | 111.72 ns | 167.21 ns | 1,150,541.82 | 0.0070 |    1431 B |      
| ClientServerCommunication | 100           | 10           |   675.4 ns | 104.88 ns | 156.97 ns | 1,480,676.62 | 0.0070 |    1385 B |      
| ClientServerCommunication | 100           | 20           |   601.2 ns |  60.82 ns |  91.03 ns | 1,663,267.73 | 0.0070 |    1276 B |      
| ClientServerCommunication | 100           | 30           |   497.8 ns |  75.23 ns | 112.60 ns | 2,008,798.91 | 0.0070 |    1258 B |      
| ClientServerCommunication | 100           | 40           |   449.2 ns |  63.37 ns |  94.86 ns | 2,226,409.78 | 0.0060 |    1253 B |      

## This PR

```

BenchmarkDotNet v0.13.12, Pop!_OS 22.04 LTS
13th Gen Intel Core i7-1360P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 8.0.404
  [Host]  : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  LongRun : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2

Job=LongRun  Concurrent=True  Server=True  
InvocationCount=1  IterationCount=10  LaunchCount=3  
RunStrategy=Monitoring  UnrollFactor=1  WarmupCount=3  

```
| Method                    | MessageLength | ClientsCount | Mean       | Error     | StdDev    | Median     | Req/sec      | Gen0   | Allocated |
|-------------------------- |-------------- |------------- |-----------:|----------:|----------:|-----------:|-------------:|-------:|----------:|
| **ClientServerCommunication** | **10**            | **1**            | **4,545.8 ns** | **439.01 ns** | **657.09 ns** | **4,524.7 ns** |   **219,983.69** | **0.0030** |     **590 B** |
| **ClientServerCommunication** | **10**            | **3**            | **1,343.4 ns** | **119.77 ns** | **179.27 ns** | **1,240.0 ns** |   **744,359.46** | **0.0030** |     **588 B** |
| **ClientServerCommunication** | **10**            | **5**            | **1,080.3 ns** |  **58.26 ns** |  **87.20 ns** | **1,105.1 ns** |   **925,699.35** | **0.0030** |     **588 B** |
| **ClientServerCommunication** | **10**            | **7**            |   **797.5 ns** |  **90.12 ns** | **134.89 ns** |   **856.1 ns** | **1,253,969.27** | **0.0030** |     **589 B** |
| **ClientServerCommunication** | **10**            | **10**           |   **612.7 ns** |  **88.60 ns** | **132.61 ns** |   **611.6 ns** | **1,632,132.17** | **0.0030** |     **592 B** |
| **ClientServerCommunication** | **10**            | **20**           |   **465.1 ns** |  **68.30 ns** | **102.23 ns** |   **489.2 ns** | **2,150,250.47** | **0.0030** |     **581 B** |
| **ClientServerCommunication** | **10**            | **30**           |   **343.0 ns** |  **53.88 ns** |  **80.65 ns** |   **313.2 ns** | **2,915,801.87** | **0.0030** |     **578 B** |
| **ClientServerCommunication** | **10**            | **40**           |   **364.3 ns** |  **68.47 ns** | **102.49 ns** |   **344.5 ns** | **2,745,034.99** | **0.0030** |     **574 B** |
| **ClientServerCommunication** | **100**           | **1**            | **4,574.4 ns** | **166.00 ns** | **248.46 ns** | **4,594.4 ns** |   **218,606.52** | **0.0040** |     **772 B** |
| **ClientServerCommunication** | **100**           | **3**            | **1,430.3 ns** |  **96.81 ns** | **144.90 ns** | **1,378.0 ns** |   **699,156.00** | **0.0040** |     **771 B** |
| **ClientServerCommunication** | **100**           | **5**            | **1,073.9 ns** | **129.49 ns** | **193.82 ns** | **1,074.0 ns** |   **931,146.72** | **0.0040** |     **768 B** |
| **ClientServerCommunication** | **100**           | **7**            |   **788.6 ns** |  **96.11 ns** | **143.85 ns** |   **840.6 ns** | **1,268,041.70** | **0.0040** |     **768 B** |
| **ClientServerCommunication** | **100**           | **10**           |   **627.9 ns** |  **55.15 ns** |  **82.55 ns** |   **666.8 ns** | **1,592,704.94** | **0.0040** |     **770 B** |
| **ClientServerCommunication** | **100**           | **20**           |   **380.5 ns** |  **56.70 ns** |  **84.86 ns** |   **396.5 ns** | **2,628,316.78** | **0.0040** |     **763 B** |
| **ClientServerCommunication** | **100**           | **30**           |   **352.8 ns** |  **74.77 ns** | **111.91 ns** |   **310.3 ns** | **2,834,145.20** | **0.0040** |     **756 B** |
| **ClientServerCommunication** | **100**           | **40**           |   **397.1 ns** |  **60.68 ns** |  **90.83 ns** |   **387.2 ns** | **2,518,125.92** | **0.0040** |     **752 B** |

Aaronontheweb:
Cuts memory usage in half, just about, which is inline with what I'd expect

Aaronontheweb:
Still need to add:

- [x] Pull mode for `TcpConnection` - Akka.Streams.IO blows up without this
- [x] Pull mode for `TcpListener` - same deal
- [x] Better buffer management code for `TcpConnection` writing - if someone sends a message larger than the max frame size we'll hang currently. Better to paginate it.
- [x] Need to restore the max-queue-size as byte-size semantics in our tests. I was against this initially but IMHO I think it's probably the more sensible "real measure" of resources

Aaronontheweb:
Looking good on the tests - probably still need to tweak the default HOCON settings some - and also make a decision about the `INET` values

Aaronontheweb:
Updated benchmark figures:

```ini
BenchmarkDotNet v0.13.12, Pop!_OS 22.04 LTS
13th Gen Intel Core i7-1360P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 8.0.404
  [Host]  : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2
  LongRun : .NET 8.0.11 (8.0.1124.51707), X64 RyuJIT AVX2

Job=LongRun  Concurrent=True  Server=True
InvocationCount=1  IterationCount=10  LaunchCount=3
RunStrategy=Monitoring  UnrollFactor=1  WarmupCount=3
```

| Method                    | MessageLength | ClientsCount | Mean       | Error     | StdDev    | Median     | Req/sec      | Gen0   | Allocated |
|-------------------------- |-------------- |------------- |-----------:|----------:|----------:|-----------:|-------------:|-------:|----------:|
| ClientServerCommunication | 10            | 1            | 4,628.8 ns | 208.53 ns | 312.12 ns | 4,609.6 ns |   216,038.53 | 0.0020 |     495 B |
| ClientServerCommunication | 10            | 3            | 1,226.6 ns | 118.20 ns | 176.92 ns | 1,196.6 ns |   815,246.70 | 0.0020 |     494 B |
| ClientServerCommunication | 10            | 5            |   848.6 ns | 109.93 ns | 164.53 ns |   754.0 ns | 1,178,350.39 | 0.0020 |     493 B |
| ClientServerCommunication | 10            | 7            |   630.7 ns |  84.23 ns | 126.07 ns |   552.2 ns | 1,585,506.64 | 0.0020 |     491 B |
| ClientServerCommunication | 10            | 10           |   449.5 ns |  53.85 ns |  80.60 ns |   413.5 ns | 2,224,545.09 | 0.0020 |     489 B |
| ClientServerCommunication | 10            | 20           |   350.9 ns |  56.38 ns |  84.39 ns |   303.4 ns | 2,850,007.13 | 0.0020 |     468 B |
| ClientServerCommunication | 10            | 30           |   319.0 ns |  46.31 ns |  69.31 ns |   285.9 ns | 3,134,575.99 | 0.0020 |     459 B |
| ClientServerCommunication | 10            | 40           |   315.0 ns |  39.78 ns |  59.55 ns |   287.3 ns | 3,174,601.41 | 0.0020 |     458 B |
| ClientServerCommunication | 100           | 1            | 4,343.2 ns | 202.61 ns | 303.26 ns | 4,392.8 ns |   230,247.36 | 0.0030 |     675 B |
| ClientServerCommunication | 100           | 3            | 1,290.1 ns | 101.37 ns | 151.72 ns | 1,219.4 ns |   775,143.69 | 0.0030 |     674 B |
| ClientServerCommunication | 100           | 5            |   848.6 ns |  88.48 ns | 132.43 ns |   773.6 ns | 1,178,426.61 | 0.0030 |     674 B |
| ClientServerCommunication | 100           | 7            |   617.8 ns |  74.03 ns | 110.81 ns |   579.8 ns | 1,618,648.52 | 0.0030 |     673 B |
| ClientServerCommunication | 100           | 10           |   489.6 ns |  75.87 ns | 113.56 ns |   441.4 ns | 2,042,510.67 | 0.0030 |     666 B |
| ClientServerCommunication | 100           | 20           |   352.6 ns |  53.85 ns |  80.60 ns |   308.7 ns | 2,836,045.38 | 0.0030 |     648 B |
| ClientServerCommunication | 100           | 30           |   326.7 ns |  46.89 ns |  70.18 ns |   286.1 ns | 3,061,130.39 | 0.0030 |     638 B |
| ClientServerCommunication | 100           | 40           |   338.9 ns |  53.17 ns |  79.59 ns |   283.7 ns | 2,950,622.14 | 0.0030 |     641 B |

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -4192,6 +4199,7 @@ namespace Akka.IO
         public abstract class WriteCommand : Akka.IO.Tcp.Command
         {
             protected WriteCommand() { }
+            public abstract long Bytes { get; }
```
Aaronontheweb, line 4202: New field so we can instantly measure how long our output is - need this for buffer-overflow checking

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -4065,6 +4069,8 @@ namespace Akka.IO
             public System.Collections.Generic.IEnumerable<Akka.IO.Inet.SocketOption> Options { get; }
             public bool PullMode { get; }
             public System.Net.EndPoint RemoteAddress { get; }
+            [System.Runtime.CompilerServices.NullableAttribute(2)]
+            public Akka.IO.TcpSettings TcpSettings { get; set; }
```
Aaronontheweb, line 4073: Customize `TcpSettings` on per-`Connect` basis

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -4001,6 +4002,8 @@ namespace Akka.IO
             public System.Net.EndPoint LocalAddress { get; }
             public System.Collections.Generic.IEnumerable<Akka.IO.Inet.SocketOption> Options { get; }
             public bool PullMode { get; }
+            [System.Runtime.CompilerServices.NullableAttribute(2)]
+            public Akka.IO.TcpSettings TcpSettings { get; set; }
```
Aaronontheweb, line 4006: Customize `TcpSettings` on per-`Bind` basis. Set as a mutable property since I didn't want to manipulate the CTOR

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -4231,21 +4238,31 @@ namespace Akka.IO
         public static Akka.IO.Tcp.Command Unbind() { }
         public static Akka.IO.Tcp.Command Write(Akka.IO.ByteString data, Akka.IO.Tcp.Event ack = null) { }
     }
-    public class TcpSettings
+    public sealed class TcpSettings : System.IEquatable<Akka.IO.TcpSettings>
```
Aaronontheweb, line 4241: This is now a `record` type

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -4231,21 +4238,31 @@ namespace Akka.IO
...
-        public int BatchAcceptLimit { get; }
+        public int BatchAcceptLimit { get; set; }
+        [System.ObsoleteAttribute("This property is unused")]
         public string BufferPoolConfigPath { get; }
+        [System.ObsoleteAttribute("This property is unused")]
         public string FileIODispatcher { get; }
-        public int FinishConnectRetries { get; }
+        public int FinishConnectRetries { get; set; }
+        [System.ObsoleteAttribute("This property is unused")]
         public int InitialSocketAsyncEventArgs { get; }
```
Aaronontheweb, line 4253: We don't use this setting anymore - the `BatchAcceptLimit` does this and the value will not change at any point during execution.

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.Net.verified.txt
```diff
@@ -4195,7 +4202,6 @@ namespace Akka.IO
     public sealed class TcpExt : Akka.IO.IOExtension
     {
         public TcpExt(Akka.Actor.ExtendedActorSystem system) { }
-        public Akka.IO.Buffers.IBufferPool BufferPool { get; }
```
Aaronontheweb, line 4198: Not used

src/core/Akka.Streams.Tests/IO/TcpSpec.cs
```diff
@@ -29,9 +29,11 @@ namespace Akka.Streams.Tests.IO
 {
     public class TcpSpec : TcpHelper
     {
-        public TcpSpec(ITestOutputHelper helper) : base(@"
-akka.loglevel = DEBUG
-akka.stream.materializer.subscription-timeout.timeout = 2s", helper)
+        public TcpSpec(ITestOutputHelper helper) : base("""
+                                                        akka.io.tcp.trace-logging = on
```
Aaronontheweb, line 33: Turned trace logging on here to help me debug some issues with the design

src/core/Akka.Streams.Tests/IO/TcpSpec.cs
```diff
@@ -503,7 +506,7 @@ await WithinAsync(TimeSpan.FromSeconds(15), async () =>
                             await AwaitAssertAsync(async () =>
                             {
                                 // Getting rid of existing connection actors by using a blunt instrument
-                                system2.ActorSelection(system2.Tcp().Path / "$a" / "*").Tell(Kill.Instance);
+                                system2.ActorSelection(system2.Tcp().Path / "tcp-client-connection-*").Tell(Kill.Instance);
```
Aaronontheweb, line 509: `akka.io.tcp` actors now get clear names, so that broke the functionality of this test which relied on guessing the randomly assigned actor path name

src/core/Akka.Streams/Implementation/IO/TcpStages.cs
```diff
@@ -41,6 +43,7 @@ private sealed class ConnectionSourceStageLogic : TimerGraphStageLogic, IOutHand
             private readonly TaskCompletionSource<StreamTcp.ServerBinding> _bindingPromise;
             private readonly TaskCompletionSource<NotUsed> _unbindPromise = new();
             private bool _unbindStarted = false;
+            private readonly Queue<StreamTcp.IncomingConnection> _pendingConnections = new();
```
Aaronontheweb, line 46: We removed backpressure support for accepting incoming TCP connections from the `TcpListener` itself in https://github.com/akkadotnet/akka.net/pull/7623 for a couple of reasons:

1. The way we were doing it before was demonstrably unsafe: https://github.com/akkadotnet/akka.net/issues/5988
2. Doing accounting for non-TCP things (Akka.Streams backpressure) inside the `TcpListener` would introduce a lot of complexity around accounting for which `SocketAsyncEventArgs` can be in-use, which ones cannot, and so forth. On top of that, this setting is already redundant with the system-level `backlog` setting, which is how this problem should be managed at the system level.

So instead we're making the "rate at which Akka.Streams can process incoming connections" an Akka.Streams problem by queueing them up inside the stage itself here. That way we can avoid complex accounting code inside the `TcpListener` AND avoid missing valid connections when the system is busy (which is what would happen if the TCP backlog exceeded its max value - servicing connections as quickly as possible through `SocketAsyncEventArgs` usage prevents this.)

src/core/Akka.Streams/Implementation/IO/TcpStages.cs
```diff
@@ -53,8 +56,16 @@ public ConnectionSourceStageLogic(Shape shape, ConnectionSourceStage stage, Task
 
             public void OnPull()
             {
-                // Ignore if still binding
-                _listener?.Tell(new Tcp.ResumeAccepting(1), StageActor.Ref);
+                TryPush();
+            }
+
+            private void TryPush()
```
Aaronontheweb, line 62: Similar pattern to what we use everywhere else for safely pushing downstream

src/core/Akka.Streams/Implementation/IO/TcpStages.cs
```diff
@@ -53,8 +56,16 @@ public ConnectionSourceStageLogic(Shape shape, ConnectionSourceStage stage, Task
 
             public void OnPull()
             {
-                // Ignore if still binding
-                _listener?.Tell(new Tcp.ResumeAccepting(1), StageActor.Ref);
```
Aaronontheweb, line 57: `ResumeAccepting` is obsolete

src/core/Akka.Streams/Implementation/IO/TcpStages.cs
```diff
@@ -549,22 +564,39 @@ private void Connected((IActorRef, object) args)
...
 
-                if (msg is Terminated) FailStage(new StreamTcpException("The connection actor has terminated. Stopping now."));
-                else if (msg is Tcp.CommandFailed failed) FailStage(new StreamTcpException($"Tcp command {failed.Cmd} failed"));
-                else if (msg is Tcp.ErrorClosed closed) FailStage(new StreamTcpException($"The connection closed with error: {closed.Cause}"));
-                else if (msg is Tcp.Aborted) FailStage(new StreamTcpException("The connection has been aborted"));
-                else if (msg is Tcp.Closed) CompleteStage();
-                else if (msg is Tcp.ConfirmedClosed) CompleteStage();
-                else if (msg is Tcp.PeerClosed) Complete(_bytesOut);
-                else if (msg is Tcp.Received received)
+                switch (msg)
```
Aaronontheweb, line 567: Transformed into a `switch` statement but otherwise still the same as before

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -421,22 +416,22 @@ public async Task Should_fail_writing_when_buffer_is_filled()
...
-                    // First overflow data will be received anyway
+                    // First message will go through, second one will not
                     (await actors.ServerHandler.ReceiveWhileAsync(TimeSpan.FromSeconds(1), m => m as Tcp.Received).ToListAsync())
                         .Sum(m => m.Data.Count)
-                        .Should().Be(InternalConnectionActorMaxQueueSize + 1);
+                        .Should().Be(InternalConnectionActorMaxQueueSize);
                 
                     // Check that almost-overflow size does not cause any problems
-                    actors.ClientHandler.Send(actors.ClientConnection, Tcp.ResumeWriting.Instance); // Recover after send failure
+                    //actors.ClientHandler.Send(actors.ClientConnection, Tcp.ResumeWriting.Instance); // Recover after send failure
```
Aaronontheweb, line 429: Only need to send this message once - we don't "pause" writing after a send failure because a send failure means the socket is cooked. If we reject a `Write` because the buffer is full this has no impact on the `TcpConnection` at all (it's the application's problem.)

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -611,10 +610,16 @@ public class ConnectionDetail
 
             public async Task ExpectReceivedDataAsync(TestProbe handler, int remaining)
             {
-                if (remaining > 0)
+                while (true)
```
Aaronontheweb, line 613: Fixed this so it was no longer recursive

src/core/Akka/Configuration/Config.cs
```diff
@@ -149,6 +150,7 @@ public virtual bool GetBoolean(string path, bool @default = false)
         /// <param name="def">Default return value if none provided.</param>
         /// <exception cref="InvalidOperationException">This exception is thrown if the current node is undefined.</exception>
         /// <returns>The long value defined in the specified path.</returns>
+        [return: NotNullIfNotNull(nameof(def))]
```
Aaronontheweb, line 153: updated this method to specify when it will return a nullable long versus when not.

src/core/Akka/Configuration/akka.conf
```diff
@@ -773,57 +773,33 @@ akka {
...
       register-timeout = 5s
 
-      # The maximum number of bytes delivered by a `Received` message. Before
-      # more data is read from the network the connection actor will try to
-      # do other work.
-      # The purpose of this setting is to impose a smaller limit than the
-      # configured receive buffer size. When using value 'unlimited' it will
-      # try to read all from the receive buffer.
-      max-received-message-size = unlimited
+      # The maximum size of a message that can be flushed onto the wire at once.
```
Aaronontheweb, line 776: New settings designed to match what we do with Akka.Remote, although the behavior is slightly different:

* `max-frame-size` - this is the largest amount of bytes we will flush / read to or from the socket in any one go.
* `send-buffer-size` and `receive-buffer-size` - they mean exactly what you think they mean. They should always be at least 2x the `max-frame-size`.

There are some settings for configuring this today (didn't find them because they are part an extremely non-discoverable API I would like to get rid of in the future.) Will call those out as I get there.

src/core/Akka/Configuration/akka.conf
Aaronontheweb, line 1: None of the deleted settings in here were ever used.

src/core/Akka/IO/Tcp.cs
Aaronontheweb, line 1: Deleted junk that wasn't used at all before / isn't used now.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
     /// Similar approach can be found on other networking libraries (i.e. System.IO.Pipelines and EventStore).
     /// Both buffers and <see cref="SocketAsyncEventArgs"/> are pooled to reduce GC pressure.
     /// </summary>
-    internal abstract class TcpConnection : ActorBase, IRequiresMessageQueue<IUnboundedMessageQueueSemantics>
+    internal abstract class TcpConnection : ReceiveActor, IRequiresMessageQueue<IUnboundedMessageQueueSemantics>
```
Aaronontheweb, line 73: Made this into a `ReceiveActor` to reduce boilerplate in some areas. Not worried about perf hit especially with the changes we've made in v1.6.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
     {
-        [Flags]
-        enum ConnectionStatus
+        /// <summary>
+        /// Immutable flags – reference to the live Queue + byte counter **and any deferred half‑close**.
+        /// Moving every transient flag in here lets us reason over shutdown with a single value.
+        /// </summary>
+        private readonly record struct ConnState(
```
Aaronontheweb, line 79: Rather than having a bitmasking system that was impossible to reason about, I made a `readonly record struct` with several `bool` values for all of the states we have to track - most of this only gets used during shutdown.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
+            bool IsReceiving,
+            bool IsSending,
+            bool PeerClosed,
+            bool OutputShutdown,
+            bool ReadingSuspended,
+            bool WritingSuspended,
+            bool KeepOpenOnPeerClosed,
+            Queue<(Write Cmd, IActorRef Snd)> Queue,
```
Aaronontheweb, line 87: This is a reference to the write queue held by the actor itself - we're only referencing it here in order to make `HasPending` account for it.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
-
-        private readonly ConcurrentQueue<(IActorRef Commander, object Ack)> _pendingAcks = new();
-        private bool _peerClosed;
-        private IActorRef _interestedInResume;
-        private CloseInformation _closedMessage;  // for ConnectionClosed message in postStop
-
-        private IActorRef _watchedActor = Context.System.DeadLetters;
-
-        private readonly IOException droppingWriteBecauseWritingIsSuspendedException = new("Dropping write because writing is suspended");
+            public static ConnState Initial(Queue<(Write Cmd, IActorRef Snd)> q) =>
```
Aaronontheweb, line 94: Initial state for all default TCP connections is to have:

- No active receives or sends in-progress
- Peer is open
- Output is open
- Reading and writing are suspended (will be un-suspended once we get a `Tcp.Register`)
- `KeepOpenOnPeerClosed` off by default
- No writes in the buffer

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
+                new(false, false, false, false, true, true, false, q, 0);
+        }
 
-        private readonly IOException droppingWriteBecauseQueueIsFullException = new("Dropping write because queue is full");
+        #region Ack‑aware SAEA
 
-        protected TcpConnection(TcpExt tcp, Socket socket, bool pullMode, Option<int> writeCommandsBufferMaxSize)
+        private sealed class AckSocketAsyncEventArgs : SocketAsyncEventArgs, INoSerializationVerificationNeeded,
```
Aaronontheweb, line 100: `SocketAsyncEventArgs` is designed to be subclassed so custom state can be included in the event of a completed operation - which is what we're doing here. We're also sending these custom `SocketAsyncEventArgs` objects to ourselves directly as messages, hence why dead letter suppression is turned on and serialization verification is turned off

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
 
-            _pullMode = pullMode;
-            _writeCommandsQueue = new PendingSimpleWritesQueue(Log, writeCommandsBufferMaxSize);
-            _traceLogging = tcp.Settings.TraceLogging;
+        private sealed class ReadSocketAsyncEventArgs : SocketAsyncEventArgs, INoSerializationVerificationNeeded,
+            IDeadLetterSuppression;
 
-            Tcp = tcp;
-            Socket = socket;
+        private class CommanderDied : IDeadLetterSuppression
```
Aaronontheweb, line 110: No more relying on `DeathPatchException`s to cause the `TcpConnection` to die in the event that its handler or commander dies during the state machine. Instead we use `Context.WatchWith` and either a `CommanderDied` or `HandlerDied` signal to ourselves.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
-            [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get { return !HasStatus(ConnectionStatus.Sending) && !_writeCommandsQueue.IsEmpty; }
+            public static readonly HandlerDied Instance = new();
+
+            private HandlerDied()
+            {
+            }
         }
 
-        private Option<PendingWrite> GetAllowedPendingWrite() => IsWritePending ? GetNextWrite() : Option<PendingWrite>.None;
```
Aaronontheweb, line 132: An extremely goofy setting for pulling writes that we no longer need

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
-        }
+        protected readonly TcpSettings Settings;
+        protected readonly Socket Socket;
+        protected ILoggingAdapter Log { get; } = Context.GetLogger();
 
-        protected void UnsignDeathPact()
-        {
-            if (!ReferenceEquals(_watchedActor, Context.System.DeadLetters)) Context.Unwatch(_watchedActor);
-        }
+        private readonly ArrayPool<byte> _bufferPool = ArrayPool<byte>.Shared;
```
Aaronontheweb, line 134: We borrow our read buffer from the `ArrayPool` because all of the bytes we use get copied from the read buffer using a `new byte[size]` before being transmitted to the handler. So, we will rent a read buffer - use it for the entire lifespan of this actor, and then return it when we're done. This is mostly an optimization for scenarios where you have lots of short-lived TCP connections popping off all at once.

Honestly we could probably get rid of it but I tested it with 1m short-lived connections and it definitely didn't hurt anything.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
 
-        // STATES
+        private readonly Queue<(Write Cmd, IActorRef Sender)> _pendingWrites;
+        private readonly byte[] _receiveBuffer;
+        private readonly ReadSocketAsyncEventArgs _receiveArgs;
+        private readonly AckSocketAsyncEventArgs _sendArgs;
+        
+        private readonly int _maxQueuedBytes;
```
Aaronontheweb, line 141: Maximum outbound write buffer size in terms of total bytes - this is how many bytes we're allowed to queue inside `_pendingWrites` before we start failing new `Tcp.WriteCommand` operations. We had this setting before but it was named incorrectly.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
 
-                        var registerInfo = new ConnectionInfo(register.Handler, register.KeepOpenOnPeerClosed, register.UseResumeWriting);
+        // used by Akka.Streams
+        private readonly bool _pullMode;
 
-                        Context.SetReceiveTimeout(null);
-                        Context.Become(Connected(registerInfo));
+        private IActorRef? _commander;
+        private IActorRef? _handler;
+        private CloseInformation? _closeInformation;
```
Aaronontheweb, line 152: Store closure information and report it to all subscribers during `PostStop` - subscribers are stored in the `CloseInformation` type itself. Hardening this (as you'll see later in the file) was key to fixing https://github.com/akkadotnet/akka.net/issues/7634

src/core/Akka/IO/TcpConnection.cs
Aaronontheweb: BTW, this change - sending the `SocketAsyncArgs` directly to the actor without a wrapper message, cut down on allocations significantly.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
-                        }
+        private static readonly IOException DroppingWriteBecauseQueueIsFullException =
+            new("Dropping write because queue is full");
 
-                        return true;
-                    default: return false;
-                }
-            };
-        }
+        private int? _partialWriteOffset = null;
```
Aaronontheweb, line 163: If we have a partialWrite, i.e. a really large message that exceeds the `maxFrameSize`, or the first part of a smaller message that we could grab while preparing a write when the write buffer was quite full, we store that value here and keep grabbing chunks of it. We had this behavior before and two tests to cover it.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
-                }
-                if (message is ResumeReading) return true;
-                return false;
-            };
-        }
+            _traceLogging = Settings.TraceLogging;
+            _state = ConnState.Initial(_pendingWrites);
+            Socket = socket ?? throw new ArgumentNullException(nameof(socket));
+            _receiveBuffer = _bufferPool.Rent(settings.MaxFrameSizeBytes);
+            _receiveArgs = new ReadSocketAsyncEventArgs();
```
Aaronontheweb, line 176: We have exactly 1 `SocketAsyncEventArgs` for reads and 1 for writes. You don't need any more than that for a single TCP conneciton. UDP is a different story.

src/core/Akka/IO/TcpConnection.cs
Aaronontheweb, line 1: So apologies in advance - this file is so different now that reading the diff is basically impossible. We'll have to cover this on a live call or something probably.

src/core/Akka/IO/TcpListener.cs
```diff
@@ -88,9 +88,9 @@ private ConnectionTerminated()
             public static ConnectionTerminated Instance { get; } = new();
         }
 
-        private sealed record AcceptCompleted(SocketAsyncEventArgs EventArgs) : INoSerializationVerificationNeeded;
```
Aaronontheweb, line 91: Added dead letter suppression to these types since it came up in my tests

src/core/Akka/IO/TcpManager.cs
```diff
@@ -54,6 +54,22 @@ internal sealed class TcpManager : ActorBase
     {
         private readonly TcpExt _tcp;
         
+        const string TcpListenerNamePrefix = "tcp-listener";
```
Aaronontheweb, line 57: Added some code here to force the `TcpManager` to give understandable, intent-driven names to its child actors. Should be self-explanatory.

src/core/Akka/IO/TcpOutgoingConnection.cs
Aaronontheweb, line 1: Again, apologies for the diff, but mostly all I did was refactor this to be a `ReceiveActor` and get rid of some dumb code around `SocketAsyncEventArgs` usage. No real structural changes despite what the diffs say.

src/core/Akka/IO/TcpSettings.cs
```diff
@@ -12,9 +12,9 @@
 namespace Akka.IO
 {
     /// <summary>
-    /// TBD
+    /// Settings for Akka.IO.Tcp's outbound and inbound connection acvtors.
     /// </summary>
-    public class TcpSettings
+    public sealed record TcpSettings
```
Aaronontheweb, line 17: BREAKING CHANGE: made this into a `sealed record` - should be binary compatible as long as people weren't inheriting from it.

src/core/Akka/IO/TcpSettings.cs
```diff
@@ -38,38 +38,72 @@ public static TcpSettings Create(ActorSystem system)
...
-                    ? int.MaxValue
-                    : config.GetInt("file-io-transferTo-limit", 512 * 1024),
                 finishConnectRetries: config.GetInt("finish-connect-retries", 5),
                 outgoingSocketForceIpv4: config.GetBoolean("outgoing-socket-force-ipv4", false),
                 writeCommandsQueueMaxSize: config.GetInt("write-commands-queue-max-size", -1));
         }
+        
+        
+        // private so we can change the constructor in the future
+        private TcpSettings(
```
Aaronontheweb, line 64: Breaking change: made the CTOR private. No one could possibly `new` up this class anyway, given how complicated it is. So have them use `TcpSettings.Create` plus the record syntax to manipulate the values afterwards. This will make it much easier for us to change settings values again in the future if needed.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -46,375 +70,174 @@ namespace Akka.IO
...
-                    case SocketReceived _: DoRead(info, null); return true;
-                    case CloseCommand cmd: HandleClose(info, Sender, cmd.Event); return true;
-                    default: return false;
-                }
-            };
-        }
+            Settings = settings;
+            _maxQueuedBytes = settings.WriteCommandsQueueMaxSize; // –1 ⇒ unlimited;
+            _pendingWrites = new Queue<(Write Cmd, IActorRef Sender)>(16);
+            _pullMode = pullMode;
```
Aaronontheweb, line 170: If `_pullMode` is enabled, which it always is when we're running Akka.Streams.IO.Tcp, then we are only allowed to have 1 `Tcp.Received` in-flight at any given time until we get a `Tcp.ResumeReading`. I'm fine with Akka.Streams backpressuring the socket this way because the buffer settings we added allow for this, whereas doing this with accepting incoming connections is dumb and can result in connection loss.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -435,692 +258,523 @@ protected void CompleteConnect(IActorRef commander, IEnumerable<Inet.SocketOptio
...
+            _state = _state with { IsSending = true };
+            if (_traceLogging)
+                Log.Debug($"[TcpConnection] TrySend: sending {segs.Count} segments, total bytes={batchBytes}");
+            if (!Socket.SendAsync(_sendArgs)) Self.Tell(_sendArgs, Self);
+        }
+        
+        /* ====================================================================*/
+        /*  Shutdown decision                                                  */
+        /* ====================================================================*/
+        private void HandleClose(IActorRef closeSender, ConnectionClosed closeEvent)
```
Aaronontheweb, line 632: Probably the most complex / important piece of code in the machinery here: dealing with shutdowns. I mostly preserved this from the original code in spirit - because it actually deals with almost everything EXCEPT FOR REPORTING SOCKET CLOSURES correctly, which we do address now.

src/core/Akka/IO/TcpConnection.cs
```diff
@@ -435,692 +258,523 @@ protected void CompleteConnect(IActorRef commander, IEnumerable<Inet.SocketOptio
...
+        {
+            Receive<Register>(reg =>
+            {
+                _handler = reg.Handler;
+                if (_traceLogging) Log.Debug("[{0}] registered as connection handler", reg.Handler);
+                Context.WatchWith(_handler, HandlerDied.Instance);
+                Context.Unwatch(_commander);
+                _state = _state with { KeepOpenOnPeerClosed = reg.KeepOpenOnPeerClosed, ReadingSuspended = _pullMode, WritingSuspended = false };
+                // set a default close event - if someone hard-kills us we log an aborted
+                _closeInformation = CloseInformation.Single(_handler, Aborted.Instance);
```
Aaronontheweb, line 294: Key fix for https://github.com/akkadotnet/akka.net/issues/7634 - we always store an `Abort` + `CloseInformation`, so if the actor gets killed via `Context.Stop` or whatever we will _always_ report a reason now.

Aaronontheweb (commented):
Detailed my changes as best as I could - lots of work went into this PR but the numbers in the benchmarks speak for themselves.

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -4065,6 +4069,8 @@ namespace Akka.IO
             public System.Collections.Generic.IEnumerable<Akka.IO.Inet.SocketOption> Options { get; }
```
Aaronontheweb, line 4069: These really need to be obsoleted and deleted

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
Arkatufus: Is it possible to duplicate the constructor and replace the old `SocketOption` with `TcpSettings` instead? We can then mark the old .ctor as obsolete

Arkatufus (changes requested):
Just one question

Aaronontheweb:
@Arkatufus 

> Is it possible to duplicate the constructor and replace the old SocketOption with TcpSettings instead? We can then mark the old .ctor as obsolete

yep, we could do that - but would I also need to go through and disable the `SocketOption`s from actually being applied to the TCP connections too?

And to go even further: to completely remove this stuff we'd need to clean up the Akka.IO.Udp side of things too, which is probably a good project for v1.6.

Arkatufus:
Well, with the obsolete attribute, we'll be phasing out the old option, by making the option not available on the new setting, only one would be applied, so there wont be any conflict.

Arkatufus (approved):
LGTM

PR #7639, opened 2025-05-13
Add DistributedPubSub ping-pong benchmark

Arkatufus:


Aaronontheweb:
Think we could get this integrated into Akka.Cluster.Benchmarks instead @Arkatufus ?

Arkatufus:
I'll look into that

PR #7643, opened 2025-05-14
Akka.Actor: remove `TBD`s

Aaronontheweb:
## Changes

Working on cleaning up TBDs


src/core/Akka/Event/DefaultLogger.cs
```diff
@@ -19,10 +19,10 @@ public class DefaultLogger : ActorBase, IRequiresMessageQueue<ILoggerMessageQueu
         private MinimalLogger _stdoutLogger;
         
         /// <summary>
-        /// TBD
+        /// Handles incoming logger messages and events.
         /// </summary>
-        /// <param name="message">TBD</param>
-        /// <returns>TBD</returns>
+        /// <param name="message">The message to be processed.</param>
+        /// <returns>True if the message was handled, false otherwise.</returns>
```
Aaronontheweb, line 25: Eh, we should get rid of this comment - might do that on another PR

src/core/Akka/Util/StringFormat.cs
```diff
@@ -10,7 +10,7 @@
 namespace Akka.Util
 {
     /// <summary>
-    /// TBD
+    /// Utility methods for safely formatting strings with null values.
     /// </summary>
     public static class StringFormat
```
Aaronontheweb, line 15: Need to review where this class is being used - looks like a perf hog

src/core/Akka/Util/Vector.cs
```diff
@@ -12,16 +12,17 @@
...
-        /// TBD
+        /// Creates a function that fills a collection with a specified number of elements 
+        /// generated by the provided factory function.
         /// </summary>
-        /// <typeparam name="T">TBD</typeparam>
-        /// <param name="number">TBD</param>
-        /// <returns>TBD</returns>
+        /// <typeparam name="T">The type of elements to create.</typeparam>
+        /// <param name="number">The number of elements to create.</param>
+        /// <returns>A function that takes an element factory and returns a list of generated elements.</returns>
```
Aaronontheweb, line 25: This looks like a perfect waste of time - probably needs to be deleted.

PR #7648, opened 2025-05-15
Akka.IO: validate we don't lose data when processing `Tcp.Received`

Aaronontheweb:
## Changes

just making double-extra-sure that we can't lose data with how we do buffer processing

## Checklist

### DIFF
--- a/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
+++ b/src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
@@ -21,6 +21,7 @@ using Xunit;
 using Xunit.Abstractions;
 using FluentAssertions;
 using System.Runtime.InteropServices;
+using Akka.Util;
 
 namespace Akka.Tests.IO
 {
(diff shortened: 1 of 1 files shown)


src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -211,20 +212,45 @@ public async Task BugFix_3021_Tcp_Should_not_drop_large_messages()
                 var actors = await x.EstablishNewClientConnectionAsync();
 
                 // create a large-ish byte string
-                var str = Enumerable.Repeat("f", 567).Join("");
-                var testData = ByteString.FromString(str);
+                const int length = 2064;
```
Aaronontheweb, line 215: 1. Create 3 random messages that are all reasonably large
2. Transmit them sequentially
3. Validate that the received byte buffers all match the input messages, when laid out  sequentially in a new output buffer

src/core/Akka.Tests/IO/TcpIntegrationSpec.cs
```diff
@@ -211,20 +212,45 @@ public async Task BugFix_3021_Tcp_Should_not_drop_large_messages()
...
+                }
+                
+                var recv1 = bigBuffer.Slice(0, length).ToArray();
+                Assert.Equivalent(recv1, buffer1);
+                
+                var recv2 = bigBuffer.Slice(length, length).ToArray();
+                Assert.Equivalent(recv2, buffer2);
+                
+                var recv3 = bigBuffer.Slice(length * 2, length).ToArray();
+                Assert.Equivalent(recv3, buffer3);
```
Aaronontheweb, line 253: All of these pass, so we should be good to go

Aaronontheweb (commented):
Detailed my changes

Aaronontheweb:
Looks like something hung during execution on Linux, but there weren't any actual test failures.

PR #7652, opened 2025-05-15
Add wait-for-subscribers feature

### LINKED ISSUE #7627: `DistributedPubSub`: add a `wait-for-subscribers` value to `Publish` commands to delay publishing / discarding until subscribers are available

A common problem that occurs with `DistributedPubSub` is when both the publisher and the subscriber start at essentially the same time, but on different nodes, and it takes a moment for that subscription to replicate. You end up with an immediate `DeadLetter` per https://github.com/akkadotnet/akka.net/issues/7626 and the publisher is never notified that this happened unless the end-user implements their own explicit ACK-reply system from all subscribers.

This makes `DistributedPubSub` kind of unwieldy to use in cases that require a higher degree of consistency.

I think it would be worth doing two things:

1. Create a new `PublishWithAck` message that sends a reply back to the publisher, letting them know that the message was sent to `N` nodes with subscribers. That way the publisher can get feedback on whether or not their message was actually delivered to someone without having to write infrastructure to do this themselves in the subscriber actors.
2. Create a new `WaitForSubscribers` value - we shouldn't let the user do stuff like specifying the number of subscribers because users will 100% interpret that to mean "the number of subscribers" and not the "number of nodes with subscribers," the latter being what we actually can measure. This should also have a timeout property to prevent memory build up and the `DistributedPubSub` should have a max-buffer-size setting for the number of "waiting" messages it will allow per-topic.

The alternatives are forcing the user to do all of this themselves via the queries we added to support https://github.com/akkadotnet/akka.net/issues/3663 - which is very high-boilerplate, janky, and generally not good.

Arkatufus:
Fixes #7627

# DistributedPubSub wait-for-subscriber feature flag

## HOCON settings added

* akka.cluster.pub-sub.buffered-messages.max-per-topic = 1000
* akka.cluster.pub-sub.buffered-messages.timeout-check-interval = 1s

## Settings added to DistributedPubSubSettings

* MaxBufferedMessagePerTopic
* BufferedMessageTimeoutCheckInterval

## Buffer

* If the mediator received a `PublishWithAck` message and able to deliver it immediately, it will send back a `PublishSucceeded` message back to the original sender as a signal.
* If the mediator could not deliver a `PublishWithAck` message, it will store the message as a `BufferedMessage` inside the buffer.

### Buffer timeout

* The `PublishWithAck` message contains the timeout value for each messages
* Every `timeout-check-interval` time, the mediator will scan through all of the buffered message to check and see if it has been sitting in the buffer for more than timeout period. 
* Timed out buffered message will be removed from the buffer and either silently discarded or sent to the dead letter.
* A `PublishFailure` message will also be sent to the original sender to signal publish failures.

### Buffer delivery

* We're now tracking every new key insertion into the bucket registry.
* When keys are being inserted, the mediator send a `NewBucketKeysAdded` message to itself.
* When a `NewBucketKeysAdded` message is received, the mediator checks the buffer and re-send all waiting messages that matches the keys to itself.
* For each delivered message, the mediator will send a `PublishSucceeded` message to the original sender as a signal.

### Buffer overflow

When a new message is inserted and the topic buffer count exceeds `buffered-messages.max-per-topic`, the mediator will discard the oldest message in the buffer and send a `PublishFailed` to the original sender of that message.

## Checklist


src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/reference.conf
```diff
@@ -32,6 +32,32 @@ akka.cluster.pub-sub {
   # When a message is published to a topic with no subscribers send it to the dead letters.
   send-to-dead-letters-when-no-subscribers = on
 
+  # When set to 'true', mediator will buffer messages that are failed to be published or sent
+  # because there are no subscribers in the cluster. 
+  # Default: false
+  wait-for-subscribers = false
```
Aaronontheweb: Nah - this should be a `per-publish` operation, not a global setting. The publisher is going to also need an ACK / NACK back to confirm that it really did, or did not send a message to someone. So the message is also going have to specify a publish timeout - we can make that optional and have a default value somewhere in HOCON, but what we absolutely don't want is a binary "everything is done this way" or "nothing is done this way" behavior setting

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/OverflowStrategy.cs
```diff
@@ -0,0 +1,39 @@
...
+//      Copyright (C) 2013-2025 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+namespace Akka.Cluster.Tools.PublishSubscribe;
+
+/// <summary>
+/// Represents a strategy that decides how to deal with a buffer that is full but is about to receive a new message.
+/// </summary>
+public enum OverflowStrategy
```
Aaronontheweb: Make the buffer circular and just fail older messages with a NACK.

Aaronontheweb: Again, if users want something fancier - they should be setting up RabbitMQ instances

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
```diff
@@ -415,6 +415,20 @@ public override string ToString()
             return $"Publish<topic:{Topic}, sendOneToEachGroup:{SendOneMessageToEachGroup}, message:{Message}>";
         }
     }
+    
+    public sealed class PublishWithAck : Publish
```
Arkatufus: New publish class, if failed to publish, will be buffered

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
```diff
@@ -415,6 +415,20 @@ public override string ToString()
...
+    {
+        public PublishWithAck(string topic, object message, TimeSpan timeout, bool sendOneMessageToEachGroup = false) : base(topic, message, sendOneMessageToEachGroup)
+        {
+            Timeout = timeout;
+        }
+        
+        public TimeSpan Timeout { get; }
+    }
+    
+    public sealed record PublishFailed(PublishWithAck Message);
```
Arkatufus: `PublishWithAck` signal that publish failed

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
```diff
@@ -415,6 +415,20 @@ public override string ToString()
...
+    
+    public sealed record PublishSucceeded(PublishWithAck Message);
```
Arkatufus: `PublishWithAck` signal that publish succeeded

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -127,20 +129,19 @@ public static Props Props(DistributedPubSubSettings settings)
         private readonly string _topicPrefix;
         private readonly PubSubCache _cache;
 
+        private readonly int _maxBufferPerTopic;
+        private readonly TimeSpan _bufferedMessageTimeoutCheckInterval;
+        private readonly Dictionary<string, List<BufferedMessage>> _bufferedMessages = new();
```
Arkatufus, line 134: New fields, maximum buffer size per topic, timeout check interval, and the per topic buffer

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -127,20 +129,19 @@ public static Props Props(DistributedPubSubSettings settings)
...
         public IImmutableDictionary<Address, long> OwnVersions
-        {
-            get
-            {
-                return _registry
-                    .Select(entry => new KeyValuePair<Address, long>(entry.Key, entry.Value.Version))
-                    .ToImmutableDictionary(kv => kv.Key, kv => kv.Value);
-            }
-        }
+            => _registry
```
Arkatufus, line 143: No logic change, just modernization

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -320,14 +324,54 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
...
+                                continue;
+                            
+                            // Create a bucket diff, we're only interested in new items being added
+                            var keys = new HashSet<string>(bucket.Content.Keys);
+                            keys.ExceptWith(new HashSet<string>(myBucket.Content.Keys));
+                            if (keys.Count > 0)
+                            {
+                                // Send the diff to ourselves
+                                Self.Tell(new NewBucketKeysAdded(keys));
+                            }
```
Arkatufus: Capture `Delta` diff, we're interested in new keys being propagated to this cluster node from other nodes.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -320,14 +324,54 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
...
+                            
+                            // Merge remote bucket with ours
+                            var newBucket = new Bucket(myBucket.Owner, bucket.Version, myBucket.Content.SetItems(bucket.Content));
+                            _registry[bucket.Owner] = newBucket;
```
Arkatufus, line 357: Old code, no logic change

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -320,14 +324,54 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
...
                         }
                     }
                 }
             });
             Receive<GossipTick>(_ => HandleGossip());
             Receive<Prune>(_ => HandlePrune());
+            Receive<NewBucketKeysAdded>(subs =>
```
Arkatufus, line 364: New message handler, this handles new keys being added into the registry bucket

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -320,14 +324,54 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
...
+                    if (!_bufferedMessages.TryGetValue(key, out var buffer)) 
+                        continue;
+                    
+                    foreach (var bufferedMessage in buffer)
+                    {
+                        Self.Tell(bufferedMessage.Message, bufferedMessage.Sender);
+                    }
+
+                    _bufferedMessages.Remove(key);
+                }
```
Arkatufus, line 377: For each newly added keys, see if we have a matching buffer. If we do, resend all of them

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -320,14 +324,54 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
...
+            });
+            Receive<PruneBufferTick>(_ =>
```
Arkatufus, line 379: New message handler for timed-out message pruning

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -320,14 +324,54 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
...
+            {
+                foreach (var buffer in _bufferedMessages.Values)
+                {
+                    var removed = buffer.Where(bufferedMessage => bufferedMessage.Deadline.IsOverdue).ToArray();
+                    foreach (var removedMsg in removed)
+                    {
+                        buffer.Remove(removedMsg);
+                        IgnoreOrSendToDeadLetters(removedMsg.Message, removedMsg.Sender);
+                    }
+                }
```
Arkatufus, line 389: Scan the whole buffered messages and remove all timed-out messages

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -417,6 +461,32 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
             });
         }
 
+        private void BufferMessageOrDeadLetter(PublishWithAck message, IActorRef sender)
+        {
+            var topic = message.Topic;
+            var encodedTopic = _cache.EncodeName(topic);
+            var key = _cache.MakeKey(Self.Path, encodedTopic);
```
Arkatufus, line 483: Generate the key for the topic

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -417,6 +461,32 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
...
+            if(buffer.Count >= _maxBufferPerTopic)
+            {
+                _log.Warning("PublishWithAck buffer overflowed for topic [{0}]. New message inserted: {1}", topic, message);
+                while (buffer.Count >= _maxBufferPerTopic)
+                {
+                    var removed = buffer[0];
+                    buffer.RemoveAt(0);
+                    IgnoreOrSendToDeadLetters(removed.Message, removed.Sender);
+                }
+            }
```
Arkatufus, line 500: If buffer is full, dequeue the last message in the buffer and send a NACK to the sender

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -417,6 +461,32 @@ public DistributedPubSubMediator(DistributedPubSubSettings settings)
...
+            
+            buffer.Add(new BufferedMessage(message, new Deadline(DateTime.UtcNow + message.Timeout), sender));
```
Arkatufus, line 502: Store the new message into the buffer

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -493,15 +560,21 @@ private void HandleRegisterTopic(IActorRef actorRef)
...
-            if (!_registry.TryGetValue(_cluster.SelfAddress, out var bucket))
-                _registry.Add(_cluster.SelfAddress,
-                    new Bucket(_cluster.SelfAddress, v, ImmutableDictionary<string, ValueHolder>.Empty.Add(key, new ValueHolder(v, value))));
-            else
-                _registry[_cluster.SelfAddress] = new Bucket(bucket.Owner, v, bucket.Content.SetItem(key, new ValueHolder(v, value)));
+            var newBucket = _registry.TryGetValue(_cluster.SelfAddress, out var bucket)
+                ? new Bucket(bucket.Owner, v, bucket.Content.SetItem(key, new ValueHolder(v, value)))
+                : new Bucket(_cluster.SelfAddress, v, ImmutableDictionary<string, ValueHolder>.Empty.Add(key, new ValueHolder(v, value)));
+            
+            _registry[_cluster.SelfAddress] = newBucket;
```
Arkatufus, line 582: Old code, no logic changed

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -493,15 +560,21 @@ private void HandleRegisterTopic(IActorRef actorRef)
...
+            Self.Tell(new NewBucketKeysAdded([key]));
```
Arkatufus: Send a signal to self that a new key has been added

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -493,15 +560,21 @@ private void HandleRegisterTopic(IActorRef actorRef)
...
         }
 
-        private void IgnoreOrSendToDeadLetters(IWrappedMessage message)
+        private void IgnoreOrSendToDeadLetters(object message, IActorRef sender)
         {
+            if (message is PublishWithAck needAck)
+            {
+                sender.Tell(new PublishFailed(needAck));
+            }
```
Arkatufus: If the message is `PublishWithAck`, notify sender that it failed

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -527,9 +600,17 @@ private void PublishMessage(string path, IWrappedMessage publish, bool allButSel
                 if (r == null) continue;
                 r.Forward(publish.Message);
                 counter++;
+                if(publish is PublishWithAck needAck)
+                    Sender.Tell(new PublishSucceeded(needAck));
```
Arkatufus: If message is `PublishWithAck`, send a success signal to the sender

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -527,9 +600,17 @@ private void PublishMessage(string path, IWrappedMessage publish, bool allButSel
...
             }
 
-            if (counter == 0) IgnoreOrSendToDeadLetters(publish);
+            if (counter == 0)
+            {
+                if(publish is PublishWithAck needAck)
+                    BufferMessageOrDeadLetter(needAck, Sender);
```
Arkatufus, line 627: Buffer message if its a `PublishWithAck`

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -555,7 +636,10 @@ private void PublishToEachGroup(string path, Publish publish)
 
             if (groups.Count == 0)
             {
-                IgnoreOrSendToDeadLetters(publish);
+                if(publish is PublishWithAck needAck)
+                    BufferMessageOrDeadLetter(needAck, Sender);
```
Arkatufus, line 660: Buffer message if its a `PublishWithAck`

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -565,6 +649,9 @@ private void PublishToEachGroup(string path, Publish publish)
                     if (routees.Length != 0)
                         new Router(_settings.RoutingLogic, routees).Route(wrappedMessage, Sender);
                 }
+                
+                if(publish is PublishWithAck needAck)
+                    Sender.Tell(new PublishSucceeded(needAck));
```
Arkatufus, line 674: If message is `PublishWithAck`, send a success signal to the sender

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -582,8 +669,8 @@ private IEnumerable<KeyValuePair<string, Routee>> ExtractGroups(string prefix, s
 
         private void HandlePrune()
         {
-            var modifications = new Dictionary<Address, Bucket>();
-            foreach (var (owner, bucket) in _registry)
+            var modifications = new List<Bucket>();
+            foreach (var bucket in _registry.Values)
```
Arkatufus, line 693: Simplify logic, the bucket itself already stores the bucker owner, there is no need to scan the dictionary keys.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -634,6 +721,7 @@ protected override void PreStart()
             //Start periodic gossip to random nodes in cluster
             Timers.StartPeriodicTimer(GossipTimerKey, GossipTick.Instance, _settings.GossipInterval, _settings.GossipInterval, Self);
             Timers.StartPeriodicTimer(PruneTimerKey, Prune.Instance, _pruneInterval, _pruneInterval, Self);
+            Timers.StartPeriodicTimer(PruneBufferTimerKey, PruneBufferTick.Instance, _bufferedMessageTimeoutCheckInterval, _bufferedMessageTimeoutCheckInterval, Self);
```
Arkatufus, line 744: Start timed out message pruning timer

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
```diff
@@ -45,37 +45,28 @@ public static DistributedPubSubSettings Create(Config config)
...
-                default:
-                    throw new ArgumentException("Unknown routing logic is tried to be applied to the pub-sub mediator: " +
-                                                routingLogicName);
-            }
+                "random" => (RoutingLogic) new RandomLogic(),
+                "round-robin" => new RoundRobinRoutingLogic(),
+                "broadcast" => new BroadcastRoutingLogic(),
+                "consistent-hashing" => throw new ArgumentException("Consistent hashing routing logic cannot be used by the pub-sub mediator"),
+                var unknown => throw new ArgumentException($"Unknown routing logic is tried to be applied to the pub-sub mediator: {unknown}")
+            };
```
Arkatufus, line 55: Old code modernization, no logic changed

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
```diff
@@ -119,13 +120,47 @@ public static DistributedPubSubSettings Create(Config config)
         /// <param name="maxDeltaElements">The maximum number of delta elements that can be propagated in a single gossip tick.</param>
         /// <param name="sendToDeadLettersWhenNoSubscribers">When a message is published to a topic with no subscribers send it to the dead letters.</param>
         /// <exception cref="ArgumentException">Thrown if a user tries to use a <see cref="ConsistentHashingRoutingLogic"/> with routingLogic.</exception>
+        [Obsolete("Use .ctor that supports WaitForSubscribers instead. Since 1.4.42")]
```
Arkatufus, line 123: Old .ctor obsoleted

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
```diff
@@ -119,13 +120,47 @@ public static DistributedPubSubSettings Create(Config config)
...
+        /// <param name="routingLogic">Optional. The routing logic used for distributing messages for topic groups.</param>
+        /// <param name="gossipInterval">The gossip interval for propagating topic/subscriber data to other mediators.</param>
+        /// <param name="removedTimeToLive">The amount of time it takes to prune a deactivated subscriber from the network.</param>
+        /// <param name="maxDeltaElements">The maximum number of delta elements that can be propagated in a single gossip tick.</param>
+        /// <param name="sendToDeadLettersWhenNoSubscribers">When a message is published to a topic with no subscribers send it to the dead letters.</param>
+        /// <param name="maxBufferedMessagePerTopic">Maximum message buffer size for each topic</param>
+        /// <param name="bufferedMessageTimeoutCheckInterval">Buffered message timeout condition check interval</param>
+        /// <exception cref="ArgumentException">Thrown if a user tries to use a <see cref="ConsistentHashingRoutingLogic"/> with routingLogic.</exception>
+        public DistributedPubSubSettings(
```
Arkatufus, line 155: New backward compatible .ctor

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
```diff
@@ -138,66 +173,54 @@ public DistributedPubSubSettings(
...
-        public DistributedPubSubSettings WithRole(string role)
+        private DistributedPubSubSettings Copy(
+            string? role = null,
+            RoutingLogic? routingLogic = null,
+            TimeSpan? gossipInterval = null,
+            TimeSpan? removedTimeToLive = null,
+            int? maxDeltaElements = null,
+            bool? sendToDeadLettersWhenNoSubscribers = null,
+            int? maxBufferedMessagePerTopic = null,
+            TimeSpan? bufferedMessageTimeoutCheckInterval = null)
```
Arkatufus: Immutability modernization, implement `Copy()` pattern

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
```diff
@@ -316,6 +317,25 @@ internal sealed class GossipTick
         private GossipTick() { }
     }
 
+    /// <summary>
+    /// Internal event signalling that a new subscriber has been added to the registry
+    /// either locally using <see cref="Put"/>, <see cref="Subscribe"/>, or from a <see cref="Delta"/>.
+    /// </summary>
+    internal sealed record NewBucketKeysAdded(IEnumerable<string> Topics);
```
Arkatufus: New internal message, event signal that a new key was added to the registry bucket

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
```diff
@@ -316,6 +317,25 @@ internal sealed class GossipTick
...
+    
+    /// <summary>
+    /// Container for buffered <see cref="Publish"/> or <see cref="Send"/> messages
+    /// </summary>
+    /// <param name="Message">The original message being buffered</param>
+    /// <param name="Deadline">The deadline where this buffered message should be timed out</param>
+    internal sealed record BufferedMessage(IWrappedMessage Message, Deadline Deadline, IActorRef Sender);
```
Arkatufus: New internal class, buffer item

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
```diff
@@ -316,6 +317,25 @@ internal sealed class GossipTick
...
+
+    internal sealed class PruneBufferTick
```
Arkatufus: New timer message for pruning timed out messages

Arkatufus (commented):
Self-review

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
Aaronontheweb: Rather than subclass, I'd have these both be separate classes - this will avoid a litany of potential internal programming issues with Akka.NET (i.e. having to organize the `Receive` cases in the right order since pattern matching for `Publish` will also hit `PublishWithAck`), plus you _really don't want_ users to be able to subclass this either.

Aaronontheweb: I'm dealing with this same type of design smell in https://github.com/akkadotnet/akka.net/pull/7637 and wish we'd never gone down that road (having event / message types be subclassable) there too

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
Aaronontheweb: Should include a "reason" field in here - an `enum` with some values like:

- Timeout
- MediatorShuttingDown (we will should fail pending publish operations if the mediator crashes or dies)

We should also enable dead letter suppression for this type

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
Aaronontheweb: Need deadletter suppression here too

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Aaronontheweb: LGTM

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Aaronontheweb: Might as well make the `BufferedMessage` a `struct` to save on allocations

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Aaronontheweb: Doesn't this send a `PublishSucceeded` for _every_ subscriber? We only need to do this once per message.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
Aaronontheweb: A thing we did on https://github.com/akkadotnet/akka.net/pull/7637 that we should do here:

1. Don't even make the CTORs on these settings classes `public`. Force users to use the `static SettingsClass.Create` method. Eliminates a ton of binary compatibility issues long-term.
2. Make all of these settings classes into `record` types - this eliminates the `.WithValueName(` method spam.

Maybe we do that across the board in v1.6 - I'm doing it in v1.5 for the `TcpSettings` type on https://github.com/akkadotnet/akka.net/pull/7637. I'm fine doing that for `DistributedPubSub` too.

Aaronontheweb: Implication of this design change is that users have to call `DistributedPubSubSettings.Create(ActorSystem)` and then they can manipulate it from there.

Aaronontheweb: Changing this on v1.5 will probably blow a bunch of stuff up immediately, so we should probably save that for v1.6

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
Aaronontheweb: Add dead letter suppression for this - and might as well take an `IReadOnlyList<string>` instead of an `IEnumerable<string>` so we know that we're dealing with a full snapshot and not a partially enumerated / deferred enumeration list.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
Aaronontheweb: Make this a `struct`

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
Aaronontheweb: Where does this get handled? Tried looking for it

Aaronontheweb (changes requested):
Made some suggestions

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
Arkatufus: Handled on DistributedPubSubMediator.cs:363

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Aaronontheweb: An optimization here: you don't have the iterate through the entire collection. Because `List.Add` appends items to the back, you can do the following:

1. If the first item on the list is NOT overdue, you can skip the check entirely - because it's the oldest message.
2. If the first item on the list is overdue, just check until you hit the first `Overdue == false`  - you know everything after that is also not overdue

Aaronontheweb (changes requested):
Left you a suggestion on the pruning check

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
Aaronontheweb: Add dead letter suppression to this

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Arkatufus: This is not possible because the timeout value is _per message_, first message might have a 20 second timeout and the second can have a 10 millisecond timeout.

Aaronontheweb: Good point - I guess the alternative would be sorting the collection by how far out the deadline is but that's probably not worth the trouble.

Aaronontheweb (commented):
More comments

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
Arkatufus: Done

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
Arkatufus: Done

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
Arkatufus: Converted to record, Done.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Arkatufus: No, its not worth the trouble because its a queue, first in first out, order preserving.

If we want to do that, then we would have to have 2 list, one is the queue and the other is a sorted queue on the timeout property. Then we would have to remove from both list, while the sorted list removal would be optimized, the queue removal would not, making the optimization moot.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Arkatufus: Done

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/TopicMessages.cs
Arkatufus: Done

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedMessages.cs
```diff
@@ -415,6 +415,22 @@ public override string ToString()
             return $"Publish<topic:{Topic}, sendOneToEachGroup:{SendOneMessageToEachGroup}, message:{Message}>";
         }
     }
+    
+    public sealed record PublishWithAck(
+        string Topic,
+        object Message,
+        TimeSpan Timeout,
```
Aaronontheweb: I'd add a CTOR implementation just to quickly validate the timeout or you can do validation inside the mediator and send a failure message back. Probably better to throw an exception inside the publisher.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Aaronontheweb: Need to fail all of these during `PostStop` if non-empty.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Aaronontheweb: Yep - we'll stick with this as-is then.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -493,28 +588,38 @@ private void HandleRegisterTopic(IActorRef actorRef)
...
+            //       AFTER we send `SubscribeAck`
+            _newlyAddedKeys.Add(key);
         }
 
-        private void IgnoreOrSendToDeadLetters(IWrappedMessage message)
+        private void IgnoreOrSendToDeadLetters(object message, IActorRef sender)
         {
+            if (message is PublishWithAck needAck)
+            {
+                sender.Tell(new PublishFailed(needAck, PublishFailReason.Timeout));
```
Aaronontheweb, line 593: Do we want to do an early return here and skip logging the dead letter or do we want to do both? The way we have it does both and I actually think that's ok - since the `DeadLetter` monitoring will usually get universal attention.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
```diff
@@ -545,7 +659,7 @@ IEnumerable<IActorRef> Refs()
             }
         }
 
-        private void PublishToEachGroup(string path, Publish publish)
+        private void PublishToEachGroup(string path, IWrappedMessage publish)
```
Aaronontheweb, line 649: IMHO, do a common interface for both of the `Publish` message types and handle that instead.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
Aaronontheweb: Want to just make this into a `record` and handle it that way? You can still keep the methods for `With____`, but this avoids having to update the `Copy` method again in the future

Aaronontheweb (changes requested):
Left some more suggestions but I think overall the design looks good - what's the testing plan?

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Arkatufus: Done

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Arkatufus: I don't think its wise to change old behavior

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubMediator.cs
Arkatufus: This can't be done because `PublishMessage` is also used to send `SendAll`

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/DistributedPubSubSettings.cs
Arkatufus: Done

src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubConfigSpec.cs
Aaronontheweb, line 1: LGTM

src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubPublishWithAckSpec.cs
Aaronontheweb, line 1: LGTM

Aaronontheweb (approved):
LGTM

PR #7653, opened 2025-05-16
remove TestValue output

SimonCropp:
Added here https://github.com/akkadotnet/akka.net/commit/7ba59b4d52c2e72de89297a3082aa0885a030e5a

i assume this was for debugging purposes and is no longer required


Aaronontheweb:
> Added here [7ba59b4](https://github.com/akkadotnet/akka.net/commit/7ba59b4d52c2e72de89297a3082aa0885a030e5a)
> 
> i assume this was for debugging purposes and is no longer required

Yep, nice catch - can't recall what it was for but it's definitely no longer needed.

Aaronontheweb (approved):
LGTM

PR #7654, opened 2025-05-16
fix some xUnit1031 "use await instead of .Result"  warnings

SimonCropp:


Aaronontheweb:
Looks like there's some compilation issues here

PR #7670, opened 2025-05-27
Allow sharding delivery consumer to passivate self

### LINKED ISSUE #7664: Akka.Delivery at least once in cluster sharding does not respect sharding passivation convention
**Version Information**
Version of Akka.NET? dev

**Describe the bug**

From Discord.

Normally, a remember entity sharding entity can terminate itself by using `Context.Parent.Tell(new Passivate(PoisonPill.Instance)`. This can't be done when using Akka.Cluster.Sharding.Delivery because there is an extra layer of actor between the `Shard` actor and the actual user code actor. These actors should just pass the `Passivate` message to their parents to allow users to use the old passivation convention.

Arkatufus:
Fixes #7664

Allows user Akka.Cluster.Sharding.Delivery remember entity consumer actor to use `Context.Parent.Tell(new Passivate())` message convention to passivate itself.

## Changes

* Add `Passivate` message handler in `ShardingConsumerController"

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/Delivery/Internal/ShardingConsumerControllerImpl.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/Delivery/Internal/ShardingConsumerControllerImpl.cs
@@ -69,6 +69,11 @@ internal class ShardingConsumerController<T> : ReceiveActor, IWithStash
             _log.Debug("Consumer terminated before initialized.");
             Context.Stop(Self);
         });
+
+        Receive<Passivate>(_ => Sender.Equals(_consumer), p =>
+        {
+            Context.Parent.Tell(p);
+        });
         
         ReceiveAny(msg =>
         {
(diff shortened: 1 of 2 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/Delivery/Internal/ShardingConsumerControllerImpl.cs
```diff
@@ -141,6 +160,25 @@ private void Active()
                 }
             }
         });
+
+        Receive<Passivate>(_ => Sender.Equals(_consumer), p =>
+        {
+            // only consume message if remember entities is enabled
+            if(ClusterSharding.Get(Context.System).Settings.RememberEntities)
```
Aaronontheweb: I would just always forward this to the parent and skip the permissions checks et al - no reason to overthink it here

Arkatufus: done

Aaronontheweb (approved):
LGTM

PR #7680, opened 2025-06-06
Add shutdown reason to exit code propagation

### LINKED ISSUE #7517: CoordinatedShutdown: return non-zero (error) exit code when ActorSystem is removed due to DOWNing

This issue was raised by a support customer - TL;DR; if Akka.NET could force the process to exit when the Split Brain Resolver (SBR) kicks in using a non-zero exit code in situations where the node is downed, that would allow Windows Services to automatically detect a failure and restart the service more easily.

We'd need some way of passing into a parameter like `int exitCode` when we invoke the `CoordinatedShutdown` routine if one doesn't already exist. That could be propagated all the way through until the `clr-exit` stage if it's enabled - and the return code could be set there.

N/A

**Additional context**

This might be useful in non-Windows contexts too, like K8s supervision

Arkatufus:
Fixes #7517

## Changes

* Add exit code output to ActorSystem `WhenTerminated` property and `.Terminate()` method task.
* If the `CoordinatedShutdown` stack was run, its reason will be propagated to the `WhenTerminated` task.
* If "exit-clr" is set, the exit code will be passed to the `Environment.Exit()` call.
* Exit codes conforms to the Linux exit code convention

See: https://manpages.ubuntu.com/manpages/lunar/man3/sysexits.h.3head.html
See: https://manpages.ubuntu.com/manpages/noble/man3/EXIT_SUCCESS.3const.html

### Possible Exit Codes

| Exit code | Description |
|---:|---|
| 0 | `ActorSystem` shutdown cleanly due to `.Terminate()`, `.Abort()`, or normal cluster events |
| 1 | Unknown shutdown cause |
| 3 | [CLUSTER] `ActorSystem` terminated because it is downed |
| 4 | [CLUSTER] `ActorSystem` terminated because it failed to join the cluster |

### DIFF
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
@@ -400,6 +400,7 @@ namespace Akka.Actor
         public abstract Akka.Actor.IScheduler Scheduler { get; }
         public abstract Akka.Serialization.Serialization Serialization { get; }
         public abstract Akka.Actor.Settings Settings { get; }
+        public abstract Akka.Actor.CoordinatedShutdown.Reason ShutdownReason { get; }
         public System.TimeSpan StartTime { get; }
         public System.TimeSpan Uptime { get; }
         public abstract System.Threading.Tasks.Task WhenTerminated { get; }
@@ -579,30 +580,41 @@ namespace Akka.Actor
         public class ActorSystemTerminateReason : Akka.Actor.CoordinatedShutdown.Reason
         {
             public static readonly Akka.Actor.CoordinatedShutdown.Reason Instance;
+            public override int ExitCode { get; }
         }
         public class ClrExitReason : Akka.Actor.CoordinatedShutdown.Reason
         {
             public static readonly Akka.Actor.CoordinatedShutdown.Reason Instance;
+            public override int ExitCode { get; }
         }
         public class ClusterDowningReason : Akka.Actor.CoordinatedShutdown.Reason
         {
             public static readonly Akka.Actor.CoordinatedShutdown.Reason Instance;
+            public override int ExitCode { get; }
         }
         public class ClusterJoinUnsuccessfulReason : Akka.Actor.CoordinatedShutdown.Reason
         {
             public static readonly Akka.Actor.CoordinatedShutdown.Reason Instance;
+            public override int ExitCode { get; }
         }
         public class ClusterLeavingReason : Akka.Actor.CoordinatedShutdown.Reason
         {
             public static readonly Akka.Actor.CoordinatedShutdown.Reason Instance;
+            public override int ExitCode { get; }
         }
-        public class Reason
+        public abstract class Reason
         {
             protected Reason() { }
+            public abstract int ExitCode { get; }
+            public static string GetReason(int code) { }
+            public static void RegisterReason(int code, string reason) { }
+            public override string ToString() { }
+            public static bool TryGetReason(int code, out string reason) { }
         }
         public class UnknownReason : Akka.Actor.CoordinatedShutdown.Reason
         {
             public static readonly Akka.Actor.CoordinatedShutdown.Reason Instance;
+            public override int ExitCode { get; }
         }
     }
     public sealed class CoordinatedShutdownExtension : Akka.Actor.ExtensionIdProvider<Akka.Actor.CoordinatedShutdown>
@@ -2025,6 +2037,7 @@ namespace Akka.Actor.Internal
         public override Akka.Actor.IScheduler Scheduler { get; }
         public override Akka.Serialization.Serialization Serialization { get; }
         public override Akka.Actor.Settings Settings { get; }
+        public override Akka.Actor.CoordinatedShutdown.Reason ShutdownReason { get; }
         public override Akka.Actor.IInternalActorRef SystemGuardian { get; }
         public override System.Threading.Tasks.Task WhenTerminated { get; }
         public override void Abort() { }
--- a/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
@@ -214,10 +214,10 @@ namespace Akka.Tests.Actor
         }
 
         [Fact]
-        public async Task ActorSystem_exit_code_must_be_OK_on_terminate()
+        public async Task ActorSystem_shutdown_reason_must_be_Ok()
         {
-            var shutdownResult = await Sys.Terminate();
-            shutdownResult.Should().Be((int)CoordinatedShutdown.CommonExitCodes.Ok);
+            await Sys.Terminate();
+            Sys.ShutdownReason.ExitCode.Should().Be(0);
         }
 
         [Fact]
--- a/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
@@ -221,14 +221,6 @@ namespace Akka.Tests.Actor
         }
 
         [Fact]
-        public async Task ActorSystem_exit_code_must_be_ABORT_on_abort()
-        {
-            ((ExtendedActorSystem)Sys).Abort();
-            var shutdownResult = await Sys.WhenTerminated;
-            shutdownResult.Should().Be((int)CoordinatedShutdown.CommonExitCodes.Abort);
-        }
-
-        [Fact]
         public async Task Reliably_create_waves_of_actors()
         {
             var timeout = Dilated(TimeSpan.FromSeconds(20));
--- a/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
+++ b/src/core/Akka.Tests/Actor/ActorSystemSpec.cs
@@ -23,6 +23,7 @@ using Akka.Event;
 using Akka.TestKit.Extensions;
 using FluentAssertions.Execution;
 using Akka.Tests.Util;
+using FluentAssertions;
 
 namespace Akka.Tests.Actor
 {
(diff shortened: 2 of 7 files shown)


src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -156,20 +156,69 @@ public static CoordinatedShutdown Get(ActorSystem sys)
...
+            UnknownReason = 1,
+            // Exit code 2 is reserved for Linux Bash for "Incorrect Usage"
+            Abort = 3,
+            ClrExit = 4,
+            ClusterDowned = 5,
+            ClusterLeft = 6,
+            ClusterJoinFailed = 7,
+            // Exit codes 64-78 is reserved by Linux sysexits.h
+            // Exit codes 126 and above is reserved by Linux shell
+        }
```
Arkatufus: Currently supported common exit codes. Are there any other that needs to be considered?

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -156,20 +156,69 @@ public static CoordinatedShutdown Get(ActorSystem sys)
...
+                ReasonMap = new Dictionary<int, string>
+                {
+                    [(int)CommonExitCodes.Ok] = "ActorSystem shutdown successfully",
+                    [(int)CommonExitCodes.UnknownReason] = "Unknown shutdown reason",
+                    [(int)CommonExitCodes.Abort] = "ActorSystem aborted",
+                    [(int)CommonExitCodes.ClrExit] = "CLR exit",
+                    [(int)CommonExitCodes.ClusterDowned] = "Cluster node downed",
+                    [(int)CommonExitCodes.ClusterLeft] = "Cluster node left the cluster",
+                    [(int)CommonExitCodes.ClusterJoinFailed] = "Cluster join unsuccessful",
+                };
```
Arkatufus: Human readable string mapping for the error codes

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -646,32 +700,31 @@ internal static void InitPhaseActorSystemTerminate(ActorSystem system, Config co
                         {
                             if (!system.WhenTerminated.Wait(timeout) && !coord._runningClrHook)
                             {
-                                Environment.Exit(0);
+                                // TODO: We're forcefully stopping the process due to timeout, should this return 0?
+                                Environment.Exit(coord.ShutdownReason?.ExitCode ?? 0);
```
Arkatufus: Should we return non-zero for shutdown timeout?

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -646,32 +700,31 @@ internal static void InitPhaseActorSystemTerminate(ActorSystem system, Config co
...
 
                     if (terminateActorSystem)
                     {
-                        return system.FinalTerminate().ContinueWith(_ =>
+                        return system.FinalTerminate(coord.ShutdownReason?.ExitCode ?? 0).ContinueWith(_ =>
                         {
                             if (exitClr && !coord._runningClrHook)
                             {
-                                Environment.Exit(0);
+                                Environment.Exit(coord.ShutdownReason?.ExitCode ?? 0);
```
Arkatufus, line 708: We propagate the exit code to the process exit code, if it is requested

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -646,32 +700,31 @@ internal static void InitPhaseActorSystemTerminate(ActorSystem system, Config co
...
+                    
+                    if (exitClr)
                     {
-                        Environment.Exit(0);
-                        return TaskEx.Completed;
-                    }
-                    else
-                    {
-                        return TaskEx.Completed;
+                        Environment.Exit(Get(system).ShutdownReason?.ExitCode ?? 0);
```
Arkatufus, line 717: We propagate the exit code to the process exit code, if it is requested

src/core/Akka/Actor/Internal/ActorSystemImpl.cs
```diff
@@ -648,23 +649,33 @@ string PrintNode(IActorRef node, string indent)
...
+        /// <param name="system">The ActorSystem parent of this instance</param>
+        public TerminationCallbacks(Task upStreamTerminated, ActorSystem system)
         {
-            _atomicRef = new AtomicReference<Task>(new Task(() => { }));
+            _atomicRef = new AtomicReference<Task?>(new Task(() =>
+            {
+                // CoordinatedShutdown must be fetched in a lazy way because it might not be initialized yet.
+                var exitCode = CoordinatedShutdown.Get(system).ShutdownReason?.ExitCode ?? FinalExitCode;
+                _terminationPromise.TrySetResult(exitCode);
+            }));
```
Arkatufus: `TerminationCallbacks` now uses a `TaskCompletionSource` for the final task. This is cleaner than relying on task continuation chain result.

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -156,20 +156,69 @@ public static CoordinatedShutdown Get(ActorSystem sys)
...
+                    [(int)CommonExitCodes.UnknownReason] = "Unknown shutdown reason",
+                    [(int)CommonExitCodes.Abort] = "ActorSystem aborted",
+                    [(int)CommonExitCodes.ClrExit] = "CLR exit",
+                    [(int)CommonExitCodes.ClusterDowned] = "Cluster node downed",
+                    [(int)CommonExitCodes.ClusterLeft] = "Cluster node left the cluster",
+                    [(int)CommonExitCodes.ClusterJoinFailed] = "Cluster join unsuccessful",
+                };
+            }
+            
+            public abstract int ExitCode { get; }
```
Arkatufus, line 198: Each `Reason` implementation will need to declare their exit codes.

Arkatufus (commented):
self-review

src/core/Akka/Actor/CoordinatedShutdown.cs
Aaronontheweb: ClusterLeft and ClrExit should all return 0. These are normal error codes. Screwing this up and returning non-zero exit codes for "normal" terminations will set a lot of runtime environments on fire.

Aaronontheweb: Heuristic:

- If you don't know why the application exited, return 0. Better safe than sorry. Preserves current behavior.
- If you know the application exited for a not-good reason, like being downed, return non-zero. 
- If you know the application exited for a normal reason, return 0.

Aaronontheweb (changes requested):
Feedback

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -156,20 +156,63 @@ public static CoordinatedShutdown Get(ActorSystem sys)
...
+        internal enum CommonExitCodes
+        {
+            Ok = 0,
+            UnknownReason = 1,
+            // Exit code 2 is reserved for Linux Bash for "Incorrect Usage"
+            ClusterDowned = 3,
+            ClusterJoinFailed = 4,
+            // Exit codes 64-78 is reserved by Linux sysexits.h
+            // Exit codes 126 and above is reserved by Linux shell
+        }
```
Arkatufus, line 175: Updated error code list

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -156,20 +156,63 @@ public static CoordinatedShutdown Get(ActorSystem sys)
...
+            private static readonly Dictionary<int, string> ReasonMap;
+
+            static Reason()
+            {
+                ReasonMap = new Dictionary<int, string>
+                {
+                    [(int)CommonExitCodes.Ok] = "ActorSystem shutdown successfully",
+                    [(int)CommonExitCodes.UnknownReason] = "Unknown shutdown reason",
+                    [(int)CommonExitCodes.ClusterDowned] = "Cluster node downed",
+                    [(int)CommonExitCodes.ClusterJoinFailed] = "Cluster join unsuccessful",
```
Arkatufus, line 194: Updated human readable code mapping

src/core/Akka/Actor/CoordinatedShutdown.cs
Aaronontheweb: That looks good to me

Aaronontheweb (changes requested):
LGTM, but why are the multi-node tests exiting in under 8 minutes? Did we introduce an error here that causes those not to run reliably? Or could that have happened as a result of earlier binary updates we made?

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
```diff
@@ -400,6 +400,7 @@ namespace Akka.Actor
         public abstract Akka.Actor.IScheduler Scheduler { get; }
         public abstract Akka.Serialization.Serialization Serialization { get; }
         public abstract Akka.Actor.Settings Settings { get; }
+        public abstract Akka.Actor.CoordinatedShutdown.Reason ShutdownReason { get; }
```
Aaronontheweb, line 403: I don't believe we need this property here any longer, no?

src/core/Akka.Tests/Actor/ActorSystemSpec.cs
```diff
@@ -212,6 +213,13 @@ public async Task Throw_exception_when_register_callback_after_shutdown()
             Assert.Equal("ActorSystem already terminated.", ex.Message);
         }
 
+        [Fact]
+        public async Task ActorSystem_shutdown_reason_must_be_Ok()
+        {
+            await Sys.Terminate();
+            Sys.ShutdownReason.ExitCode.Should().Be(0);
```
Aaronontheweb, line 220: Yeah I think we want to keep the `ActorSystem` decoupled from this - totally a `CoordinatedShutdown` concern.

src/core/Akka/Actor/CoordinatedShutdown.cs
```diff
@@ -156,20 +156,63 @@ public static CoordinatedShutdown Get(ActorSystem sys)
...
+            }
+            
+            public abstract int ExitCode { get; }
+            
             protected Reason()
             {
 
             }
+            
+            public static void RegisterReason(int code, string reason)
```
Aaronontheweb, line 205: This should not be necessary - passing the reason into the `CoordinatedShutdown.Run` method should be sufficient.

Aaronontheweb (changes requested):
I think this needs to be redesigned in a fresh PR - concerns got mixed up here, namely trying to jam `ActorSystem.Terminate()` into the exit code infrastructure, and even after trying to clean that up the concerns are still really messy here.

IMHO - we should:

1. Just add an `ExitCode` property to the `Reason` type
2. Set the application's ExitCode to that value in the place where we do it currently.
3. Have this value default to `0` when unset, preserving current behavior.
4. Not touch the `ActorSystem` class at all.
5. I wouldn't even worry about writing tests for this - not really necessary or practical IMHO.

PR #7684, opened 2025-06-10
Propagate CoordinatedShutdown reason to application exit code

### LINKED ISSUE #7517: CoordinatedShutdown: return non-zero (error) exit code when ActorSystem is removed due to DOWNing

Arkatufus:
Fixes #7517
Superseeds #7680

## Changes

* If "exit-clr" HOCON options is set and the `CoordinatedShutdown` stack was run, its reason will be propagated to the `Environment.Exit()` call.
* Exit codes conforms to the Linux exit code convention

See: https://manpages.ubuntu.com/manpages/lunar/man3/sysexits.h.3head.html
See: https://manpages.ubuntu.com/manpages/noble/man3/EXIT_SUCCESS.3const.html

### Possible Exit Codes

| Exit code | Description |
|---:|---|
| 0 | `ActorSystem` shutdown cleanly due to `.Terminate()`, `.Abort()`, or normal cluster events |
| 1 | Unknown shutdown cause |
| 3 | [CLUSTER] `ActorSystem` terminated because it is downed |
| 4 | [CLUSTER] `ActorSystem` terminated because it failed to join the cluster |

### DIFF
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.DotNet.verified.txt
@@ -579,30 +579,37 @@ namespace Akka.Actor
         public class ActorSystemTerminateReason : Akka.Actor.CoordinatedShutdown.Reason
         {
             public static readonly Akka.Actor.CoordinatedShutdown.Reason Instance;
+            public override int ExitCode { get; }
         }
         public class ClrExitReason : Akka.Actor.CoordinatedShutdown.Reason
         {
             public static readonly Akka.Actor.CoordinatedShutdown.Reason Instance;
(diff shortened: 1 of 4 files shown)


Aaronontheweb (approved):
LGTM

PR #7693, opened 2025-06-12
Make sure that `EventSourced` timer is cancelled when persistent actor is stopped

### LINKED ISSUE #7692: `EventSourced` actors failed to clean up its timer on death
**Version Information**
Version of Akka.NET? dev

**Describe the bug**

The timer in question is in this line: https://github.com/akkadotnet/akka.net/blob/5161195633341d0e6e863ac32079e474d04b9e6a/src/core/Akka.Persistence/Eventsourced.Recovery.cs#L185

If the actor is forcefully killed while inside this state, it fails to call `timeoutCancelable.Cancel()` and will cause infinite `RecoveryTick` messages to be dead-lettered until the actor system is restarted.

Arkatufus:
Fixes #7692

## Changes

Keep track of the currently running timer `ICancelable` in a private field instead of a closure variable and make sure that it is being cancelled inside `PostStop()`


src/core/Akka.Persistence/Eventsourced.Recovery.cs
```diff
@@ -274,13 +285,21 @@ private EventsourcedState Recovering(Receive recoveryBehavior, TimeSpan timeout)
...
-                    timeoutCancelable.Cancel();
+                    _timeoutCancelable?.Cancel();
+                    _timeoutCancelable = null;
                     ReturnRecoveryPermit();
                     throw;
                 }
             });
         }
 
+        protected override void PostStop()
```
Aaronontheweb: Is there an aroundpoststop or something we can call that can't be overridden by end-users?

Arkatufus: good idea, though user can still override `AroundPostStop()` if they want to since it isn't sealed

Aaronontheweb (approved):
LGTM

PR #7717, opened 2025-06-24
[PubSub] Push ack responsibility to topic actors

Arkatufus:
Make `DistributedPubSub` `PublishWithAck` response more responsive to subscription changes by pushing the buffering and ack responsibility to the `Topic` actor.

The current `PublishWithAck` logic have a problem in that, for it to switch back to sending `PublishFailed` messages, the actual topic has to be pruned from all nodes.

## Changes

Push buffering and ack responsibility to the `Topic` actor.

> [!NOTE]
>
> This change is not the silver bullet, it introduces another problem. Lets say we have a 3 node cluster, (**Pub**, **SubA**, and **SubB** nodes).
> * An actor subscribes to topic in **SubA** node, `Topic` actor gets created in **SubA**
> * **SubA** subscriber is downed or unsubscribed itself
> * An actor subscribes to topic in **SubB** node, another `Topic` actor gets created in **SubB**
> * An actor sends a `PublishWithAck` to the mediator in **Pub**
> * The actor receives `PublishFailed` from **SubA** and a `PublishSucceeded` from **SubB**

## Checklist

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
@@ -43,7 +43,8 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Internal
     /// </summary>
     internal abstract class TopicLike : ActorBase, IWithTimers
     {
-        private const string PruneTimerKey = "PruneTimer";
+        private const string PruneTimerKey = nameof(PruneTimerKey);
+        private const string PruneBufferTimerKey = nameof(PruneBufferTimerKey);
         
         /// <summary>
         /// TBD
@@ -70,18 +71,31 @@ namespace Akka.Cluster.Tools.PublishSubscribe.Internal
         /// </summary>
         protected readonly bool SendToDeadLettersWhenNoSubscribers;
 
+        protected readonly TimeSpan PruneBufferInterval;
+
+        private readonly ILoggingAdapter _log;
+        
+        protected int MaxBufferSize { get; }
+        
+        private readonly List<BufferedMessage> _buffer = new();
+
         /// <summary>
         /// Creates a new instance of a topic or group actor.
         /// </summary>
         /// <param name="emptyTimeToLive">The TTL for how often this actor will be removed.</param>
         /// <param name="sendToDeadLettersWhenNone">When set to <c>true</c>, this actor will
         /// publish a <see cref="DeadLetter"/> for each message if the total number of subscribers == 0.</param>
-        protected TopicLike(TimeSpan emptyTimeToLive, bool sendToDeadLettersWhenNone)
+        /// <param name="pruneBufferInterval">Time interval to check for timed out PublishWithAck</param>
+        /// <param name="maxBufferSize">Maximum buffer size for each topic</param>
+        protected TopicLike(TimeSpan emptyTimeToLive, bool sendToDeadLettersWhenNone, TimeSpan pruneBufferInterval, int maxBufferSize)
         {
             Subscribers = new HashSet<IActorRef>();
             EmptyTimeToLive = emptyTimeToLive;
             SendToDeadLettersWhenNoSubscribers = sendToDeadLettersWhenNone;
             PruneInterval = new TimeSpan(emptyTimeToLive.Ticks / 2);
+            PruneBufferInterval = pruneBufferInterval;
+            MaxBufferSize = maxBufferSize;
+            _log = Context.GetLogger();
         }
 
         public ITimerScheduler Timers { get; set; }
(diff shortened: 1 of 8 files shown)


src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
```diff
@@ -152,20 +187,64 @@ private bool DefaultReceive(object message)
...
+        private void BufferMessageOrDeadLetter(PublishWithAck message, IActorRef sender)
+        {
+            if(_buffer.Count >= MaxBufferSize)
+            {
+                _log.Warning("PublishWithAck buffer overflowed for topic [{0}]. New message inserted: {1}", Self.Path.Name, message);
+                while (_buffer.Count >= MaxBufferSize)
+                {
+                    var removed = _buffer[0];
+                    _buffer.RemoveAt(0);
+                    IgnoreOrSendToDeadLetters(removed.Message, removed.Sender);
```
Aaronontheweb, line 223: So this will trigger a `PublishedFailed` message if the original message was a `PublishWithAck`

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
```diff
@@ -152,20 +187,64 @@ private bool DefaultReceive(object message)
...
+                sender.Tell(new PublishFailed(needAck, PublishFailReason.Timeout));
+            }
+            
+            if (SendToDeadLettersWhenNoSubscribers)
+            {
+                // Use the specialized NoSubscribersDeadLetter struct to clearly indicate
+                // that the message was not delivered because there were no subscribers
+                var noSubs = new NoSubscribersDeadLetter(Context.Self.Path.Name, message);
+                var deadLetter = new DeadLetter(noSubs, Sender, Self);
+                Context.System.EventStream.Publish(deadLetter);
```
Aaronontheweb, line 243: you can just publish the `NoSubs` here, no?

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
```diff
@@ -152,20 +187,64 @@ private bool DefaultReceive(object message)
 
                 default:
                     foreach (var subscriber in Subscribers)
-                        subscriber.Forward(message);
+                    {
+                        if(message is PublishWithAck needAck)
+                        {
+                            subscriber.Forward(needAck.Message);
+                            Sender.Tell(new PublishSucceeded(needAck));
```
Aaronontheweb: you will get an ack-per-subscriber under this design. Probably better to move all of this into an `else` statement below the `Subscribers.Count` check so it only happens once.

Aaronontheweb:
> An actor subscribes to topic in SubA node, Topic actor gets created in SubA
> SubA subscriber is downed or unsubscribed itself
> An actor subscribes to topic in SubB node, another Topic actor gets created in SubB
> An actor sends a PublishWithAck to the mediator in Pub
> The actor receives PublishFailed from SubA and a PublishSucceeded from SubB

ah, is this because the forwarding mediator is going to re-run the entire `PublishWithAck` procedure again?

Aaronontheweb (changes requested):
Found some bugs - please also address my comment-reply

Arkatufus:
> ah, is this because the forwarding mediator is going to re-run the entire PublishWithAck procedure again?

No, because we push the ack responsibility to the `Topic` actor. There will be 2 `Topic` actors in the cluster, one with a subscriber and the other with 0 subscribers. The one with no subscriber will report a `PublishFailed` after the `PublishWithAck` timeout expired. The only way for the `Topic` actor to be removed is by the pruning process, which happens every 120 seconds ("akka.cluster.pub-sub.remove-time-to-live" setting)

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
Arkatufus: Yes, it will. A message that got knocked out off the buffer due to buffer overflow will cause a `PublishFailed`

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
Arkatufus: ~~That is the same behavior as before the change, we wrap the message with `NoSubscriberDeadletter` context so that it would not show up as a regular dead letter.~~

correction, that was the exact same deadletter code as before, I just moved it to a new location.

src/contrib/cluster/Akka.Cluster.Tools/PublishSubscribe/Internal/Topics.cs
Arkatufus: Ah, good catch, I'll fix that

Arkatufus: Fixed

Aaronontheweb:
In the scenario you outlined one of the two nodes is going down anyway so I don't see a problem here.

PR #7720, opened 2025-07-02
Stop failing remembered entity if supervisor strategy failed

### LINKED ISSUE #7629: Akka.Cluster.Sharding: dealing with `remember-entities` and actors who can't start up correctly
**Version Information**
Version of Akka.NET? v1.5.41

**Describe the bug**

So an interesting scenario:

1. Actor gets created via Akka.Cluster.Sharding, but is misconfigured (i.e. missing DI registration that can't be resolved when the `resolver.Props` gets fired up) - crashes immediately;
2. Actor hits its "10 crashes within 1s and die" limit on the `SupervisionStrategy`, gets killed off by the `Shard`
3. `Shard` notices that the remembered-entity actor is dead, re-creates it; and
4. The cycle repeats, creating millions of errors until human intervention arrives.

**Expected behavior**

IMHO, the `remember-entities` system should not bypass the supervision system - if the supervisor recommends perma-death to a repeatedly failing actor, we need some way of honoring that and un-remembering the actor.

**Actual behavior**

Infinite churn between the `remember-entities` system and the restarting actor.

Arkatufus:
Fixes #7629

## Changes

* Add specialized shard supervision strategy with feedback mechanism to signal excessive failures
* Add new `SupervisorStrategy` settings to `ShardSupervisionStrategy` (accessible only via C# fluent API)

## Checklist

* [x] Design discussion issue #7629

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ShardSupervisionStrategy.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ShardSupervisionStrategy.cs
@@ -59,9 +59,10 @@ public class ShardSupervisionStrategy: OneForOneStrategy
             RestartChild(child, cause, suspendFirst: false);
         else
         {
-            // Only send failure feedback if restart was requested
-            if(restart)
-                context.Self.Tell(new ExcessiveSupervisorRestartPassivation(child, WithinTimeRangeMilliseconds, MaxNumberOfRetries, cause));
+            var reason = restart 
+                ? $"entity failed repeatedly within {WithinTimeRangeMilliseconds} ms, exceeding the supervisor strategy maximum restart count of {MaxNumberOfRetries}" 
+                : "entity stopped by Directive.Stop decision";
+            context.Self.Tell(new SupervisorStopDirectivePassivation(child, reason, cause));
             context.Stop(child);
         }
     }
--- a/src/contrib/cluster/Akka.Cluster.Sharding/ShardSupervisionStrategy.cs
+++ b/src/contrib/cluster/Akka.Cluster.Sharding/ShardSupervisionStrategy.cs
@@ -1,0 +1,68 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="ShardSupervisionStrategy.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2025 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using Akka.Actor;
+using Akka.Actor.Internal;
+
+namespace Akka.Cluster.Sharding;
+
+public class ShardSupervisionStrategy: OneForOneStrategy
+{
+    public ShardSupervisionStrategy(int? maxNrOfRetries, TimeSpan? withinTimeRange, Func<Exception, Directive> localOnlyDecider) 
+        : base(maxNrOfRetries, withinTimeRange, localOnlyDecider)
+    {
+    }
+
+    public ShardSupervisionStrategy(int? maxNrOfRetries, TimeSpan? withinTimeRange, IDecider decider) 
+        : base(maxNrOfRetries, withinTimeRange, decider)
+    {
+    }
+
+    public ShardSupervisionStrategy(int maxNrOfRetries, int withinTimeMilliseconds, Func<Exception, Directive> localOnlyDecider, bool loggingEnabled = true) 
+        : base(maxNrOfRetries, withinTimeMilliseconds, localOnlyDecider, loggingEnabled)
+    {
+    }
+
+    public ShardSupervisionStrategy(int maxNrOfRetries, int withinTimeMilliseconds, IDecider decider, bool loggingEnabled = true) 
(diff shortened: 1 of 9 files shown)


src/contrib/cluster/Akka.Cluster.Sharding/ShardSupervisionStrategy.cs
```diff
@@ -0,0 +1,68 @@
...
+// -----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using Akka.Actor;
+using Akka.Actor.Internal;
+
+namespace Akka.Cluster.Sharding;
+
+public class ShardSupervisionStrategy: OneForOneStrategy
```
Arkatufus, line 15: This is the custom supervisor strategy, only for the `Shard` actor

src/contrib/cluster/Akka.Cluster.Sharding/ShardSupervisionStrategy.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    public override void ProcessFailure(IActorContext context, bool restart, IActorRef child, Exception cause, ChildRestartStats stats, IReadOnlyCollection<ChildRestartStats> children)
+    {
+        if (restart && stats.RequestRestartPermission(MaxNumberOfRetries, WithinTimeRangeMilliseconds))
+            RestartChild(child, cause, suspendFirst: false);
+        else
+        {
+            // Only send failure feedback if restart was requested
+            if(restart)
+                context.Self.Tell(new ExcessiveSupervisorRestartPassivation(child, WithinTimeRangeMilliseconds, MaxNumberOfRetries, cause));
```
Arkatufus: The `ProcessFailure` code is pretty much the same as the `OneToOneStrategy` with these additional lines, we send the `Shard` actor a warning message that this failing child is due for termination because it is thrashing.

src/contrib/cluster/Akka.Cluster.Sharding/ShardingMessages.cs
```diff
@@ -61,6 +61,8 @@ public Passivate(object stopMessage)
         public object StopMessage { get; }
     }
 
+    internal sealed record ExcessiveSupervisorRestartPassivation(IActorRef Child, int TimeWindowInMilliseconds, int MaxRestartCount, Exception LastCause) : IShardRegionCommand;
```
Arkatufus: New internal local only message from supervisor strategy to `Shard` actor

Arkatufus (commented):
Self review

Arkatufus:
A note here, while this design works in a very quiet system, it might still somehow fail on a very busy system.

This scheme would not be as responsive as what the unit test shows if the `ExcessiveSupervisorRestartPassivation` message somehow got burried in the `Shard` mailbox (busy system).

src/contrib/cluster/Akka.Cluster.Sharding/ShardSupervisionStrategy.cs
Aaronontheweb: This is not quite right - we also need to handle scenarios where the `SupervisorStrategy` decided to issue a `Stop` directive. Reason being: if the actor failed in such a way that it has to be stopped, it's by definition an irrecoverable exception. Continuing to remember the entity after we get this type of signal back is net-destructive.

Aaronontheweb (changes requested):
Needs some changes in the way the supervision strategy is implemented

src/contrib/cluster/Akka.Cluster.Sharding/ShardSupervisionStrategy.cs
Aaronontheweb: Also: we need to make sure this only applies to entity actors, not to any other children of the `Shard`, such as the RE infrastructure itself. You can determine this by checking the actor paths.

Aaronontheweb: > Also: we need to make sure this only applies to entity actors, not to any other children of the Shard, such as the RE infrastructure itself. You can determine this by checking the actor paths.

nevermind, this gets handled inside the `Shard` message handlers by the looks of things

Arkatufus:
OK, all fixed

src/contrib/cluster/Akka.Cluster.Sharding/ShardingMessages.cs
```diff
@@ -61,6 +61,8 @@ public Passivate(object stopMessage)
         public object StopMessage { get; }
     }
 
+    internal sealed record SupervisorStopDirectivePassivation(IActorRef Child, string Reason, Exception LastCause) : IShardRegionCommand;
```
Aaronontheweb, line 64: LGTM

src/contrib/cluster/Akka.Cluster.Sharding/ShardSupervisionStrategy.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    public ShardSupervisionStrategy()
+    {
+    }
+
+    public override void ProcessFailure(IActorContext context, bool restart, IActorRef child, Exception cause, ChildRestartStats stats, IReadOnlyCollection<ChildRestartStats> children)
+    {
+        if (restart && stats.RequestRestartPermission(MaxNumberOfRetries, WithinTimeRangeMilliseconds))
+            RestartChild(child, cause, suspendFirst: false);
+        else
```
Aaronontheweb, line 60: LGTM

Aaronontheweb (approved):
LGTM

PR #7721, opened 2025-07-02
Fix `Settings.InjectTopLevelFallback` race condition

### LINKED ISSUE #7426: `Settings.InjectTopLevelFallback` race condition can cause failures
**Version Information**
Version of Akka.NET? v1.5.15
Which Akka.NET Modules? Akka.Cluster.Tools, others too presumably

**Describe the bug**

Consider the following method in the `ClusterSingletonManagerSettings` class:

https://github.com/akkadotnet/akka.net/blob/55644b12d08d700419f52f18fcde0d82b8c80497/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManagerSettings.cs#L28-L36

It uses ` system.Settings.InjectTopLevelFallback` in order to make sure that its serialization definitions are available for Akka.Remote so all of the Cluster.Singleton-specific messages can be deserialized.

However, the way ` system.Settings.InjectTopLevelFallback` works is inherently racy - multiple components all starting up at the same time might inject their configs all at once and we can end up "missing" a config injection, which causes the following line to fail occasionally:

https://github.com/akkadotnet/akka.net/blob/55644b12d08d700419f52f18fcde0d82b8c80497/src/contrib/cluster/Akka.Cluster.Tools/Singleton/ClusterSingletonManagerSettings.cs#L33-L34

**Expected behavior**

Plugins should be able to add their configurations to Akka.NET in a thread-safe, reliable manner that is also `CTOR`-friendly (no-`Task`). 

**Actual behavior**

There is a small chance that the plugin configuration never gets properly registered and it crash - it's also possible for the default HOCON to get quite large as a result of many plugins multiplicatively registering their configuration values.

**Additional context**

We had talked about solving this problem in v1.4, when we attempted to replace our built-in HOCON implementation with [https://github.com/akkadotnet/HOCON](https://github.com/akkadotnet/HOCON). The suggestion at the time was to have some means looking for known configuration sections and then staging them altogether inside the `ActorSystem`.

Given that https://github.com/akkadotnet/akka.net/issues/7246 is a huge goal for v1.6, any sort of assembly-scanning solution is off the table - what we probably need is something functionally similar to `InjectTopLevelFallback` now but with thread-safety.

Arkatufus:
Fixes #7426

## Changes

* Use `AtomicReference<T>` to enforce thread safety

### DIFF
--- a/src/core/Akka/Actor/Settings.cs
+++ b/src/core/Akka/Actor/Settings.cs
@@ -13,6 +13,7 @@ using Akka.Configuration;
 using Akka.Dispatch;
 using Akka.Event;
 using Akka.Routing;
+using Akka.Util;
 using ConfigurationFactory = Akka.Configuration.ConfigurationFactory;
 
 namespace Akka.Actor
@@ -27,7 +28,7 @@ namespace Akka.Actor
     {
         private readonly Config _userConfig;
         //internal static readonly Config AkkaDllConfig = ConfigurationFactory.FromResource<Settings>("Akka.Configuration.Pigeon.conf");
-        private Config _fallbackConfig;
+        private readonly AtomicReference<Config> _fallbackConfig;
 
         private void RebuildConfig()
         {
@@ -42,13 +43,20 @@ namespace Akka.Actor
         /// <summary>
         /// Injects a system config at the top of the fallback chain
         /// </summary>
-        /// <param name="config">TBD</param>
+        /// <param name="config">The latest config to be added to the front of the <see cref="Settings.Config"/> fallback chain</param>
         public void InjectTopLevelFallback(Config config)
         {
             if (Config.Contains(config)) 
                 return;
 
-            _fallbackConfig = config.SafeWithFallback(_fallbackConfig);
+            while(true)
+            {
+                var oldConfig = _fallbackConfig.Value;
+                var newConfig = config.SafeWithFallback(oldConfig);
+                if (_fallbackConfig.CompareAndSet(oldConfig, newConfig))
+                    break;
+            }
+    
             RebuildConfig();
         }
 
@@ -77,7 +85,7 @@ namespace Akka.Actor
         {
             Setup = setup;
             _userConfig = config;
-            _fallbackConfig = ConfigurationFactory.Default();
+            _fallbackConfig = new AtomicReference<Config>(ConfigurationFactory.Default());
             RebuildConfig();
 
             System = system;


src/core/Akka/Actor/Settings.cs
```diff
@@ -42,13 +43,20 @@ private void RebuildConfig()
...
                 return;
 
-            _fallbackConfig = config.SafeWithFallback(_fallbackConfig);
+            while(true)
+            {
+                var oldConfig = _fallbackConfig.Value;
+                var newConfig = config.SafeWithFallback(oldConfig);
+                if (_fallbackConfig.CompareAndSet(oldConfig, newConfig))
+                    break;
+            }
```
Arkatufus, line 58: Fix: Do not assign value to naked field directly, this can cause race condition if multiple Akka extensions running in multiple threads tries to start at the same time.

Scenario:
* Time 0
  * Extension "A" reads `_fallbackConfig` into `configA`
  * Extension "B" reads `_fallbackConfig` into `configB`
* Time 1
  * Extension "A" injects "akka.cluster.tools.singleton" into `configA`
  * Extension "B" injects "akka.persistence.sql" into `configB`
* Time 2
  * Extension "A" writes `configA` into `_fallbackConfig`
* Time 3
  * Extension "B" writes `configB` into `_fallbackConfig`
* Time 4
  * Extension "A" reads "akka.cluster.tools.singleton" from `Settings.Config`, gets `NullReferenceException` because it got clobbered by Extension "B"

Arkatufus (commented):
Self-review

src/core/Akka/Actor/Settings.cs
```diff
@@ -27,7 +28,7 @@ public class Settings
     {
         private readonly Config _userConfig;
         //internal static readonly Config AkkaDllConfig = ConfigurationFactory.FromResource<Settings>("Akka.Configuration.Pigeon.conf");
-        private Config _fallbackConfig;
+        private readonly AtomicReference<Config> _fallbackConfig;
```
Aaronontheweb, line 31: yeah, this is a much simpler fix than what I was imagining

src/core/Akka/Actor/Settings.cs
Aaronontheweb: LGTM

Aaronontheweb (approved):
LGTM - that was simple. Let's see how the testing system does.

PR #7736, opened 2025-07-29
Fix EventStreamSpec racy unit test

### LINKED ISSUE #7709: Racy Test: Akka.Tests.Event.EventStreamSpec.Be_able_to_log_unhandled_messages
Assertion failure:

```
Akka.Tests.Event.EventStreamSpec.Be_able_to_log_unhandled_messages

Assert.True() Failure
Expected: True
Actual:   **False**
```

Stack Trace:

```
   at Akka.TestKit.AkkaSpecExtensions.ShouldBeTrue(Boolean b, String message) in /home/vsts/work/1/s/src/core/Akka.Tests.Shared.Internals/AkkaSpecExtensions.cs:line 169
   at Akka.Tests.Event.EventStreamSpec.Be_able_to_log_unhandled_messages() in /home/vsts/work/1/s/src/core/Akka.Tests/Event/EventStreamSpec.cs:line 116
```

Arkatufus:
Fixes #7709

## Changes

Replace custom `MyLog` logger with the built-in test kit event filter

### DIFF
--- a/src/core/Akka.Tests/Event/EventStreamSpec.cs
+++ b/src/core/Akka.Tests/Event/EventStreamSpec.cs
@@ -16,6 +16,7 @@ using Akka.Util.Internal;
 using Xunit;
 using System.Threading.Tasks;
 using FluentAssertions;
+using FluentAssertions.Extensions;
 using Xunit.Abstractions;
 
 namespace Akka.Tests.Event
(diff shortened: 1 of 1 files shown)


Aaronontheweb (changes requested):
Still prefer to keep the `Async` test method - we should consider the synchronous testkit methods as deprecated going forward.

src/core/Akka.Tests/Event/EventStreamSpec.cs
```diff
@@ -300,9 +304,7 @@ private static string GetDebugUnhandledMessagesConfig()
                     log-dead-letters = off
                     stdout-loglevel = DEBUG
                     loglevel = DEBUG
-                    loggers = [""%logger%""]
-                }
-                ".Replace("%logger%", typeof(MyLog).AssemblyQualifiedName);
```
Aaronontheweb, line 305: where is this class and do we still need it?

Aaronontheweb (approved):
LGTM

PR #7739, opened 2025-08-05
Change Delivery CreateWithFuzzing to public for testing

Arkatufus:


src/core/Akka/Delivery/ProducerController.cs
```diff
@@ -47,7 +48,8 @@ public static Props Create<T>(IActorRefFactory actorRefFactory, string producerI
         return p;
     }
 
-    internal static Props CreateWithFuzzing<T>(IActorRefFactory actorRefFactory, string producerId,
+    [InternalApi]
+    public static Props CreateWithFuzzing<T>(IActorRefFactory actorRefFactory, string producerId,
```
Aaronontheweb, line 57: Add an XML-DOC explaining that this internal and for testing purposes; discourage off-the-street users from calling this.

Arkatufus: Done.

Issue #7644, opened 2025-05-15
Akka.Actor: remove setter from `IWithTimers`


Aaronontheweb:
**Version Information**
Version of Akka.NET? v1.5.52
Which Akka.NET Modules? Akka

**Describe the bug**

Users should not be able to modify `IWithTimers`

https://github.com/akkadotnet/akka.net/blob/1f7ffa7479152b13f201434a6791156f7f18d213/src/core/Akka/Actor/Scheduler/IWithTimers.cs#L13-L20

The reason why is we automatically set this in the `ActorBase` constructor

https://github.com/akkadotnet/akka.net/blob/1f7ffa7479152b13f201434a6791156f7f18d213/src/core/Akka/Actor/ActorBase.cs#L122-L131

Problems are:

- If user sets this, the `ActorBase` class is going to reset it, which will cause any timers scheduled in the CTOR to get lost possibly
- VS / any other IDE is going to complain about nullability / et al having this unset `Timers` property. 

## Solution

- Force users to set the timer scheduler themselves
- Make it a built-in property of `ActorBase` and not require users to decorate. Have timers built-in to all actors automatically. There will be a small perf hit from this on the message receiving pipeline but it's not the end of the world.
- Add a Roslyn Analyzer warning telling users not to set it, which is dumb because it means the users are going to get complained at either way.
- In newer versions of .NET, maybe this can be solved with default interface implementations.

## Additional Context

Stashing also has this problem!

PR #7741, opened 2025-08-07
[POC] Streamline IWithTimers API

Arkatufus:
Discussing #7644

A POC to address IWithTimers

## Changes

* Remove IWithTimers
* Timers are accessed directly from the ActorContext (ActorCell)
* Timers are created lazily
* Instead of using `obj is IWithTimers` check, we use `Context.HaveTimers` instead (mechanism can always be changed)


src/core/Akka/Actor/ActorCell.cs
```diff
@@ -153,7 +154,19 @@ internal static ActorCell? Current
...
+
+        public ITimerScheduler Timers
+        {
+            get
+            {
+                return _timers ??= new TimerScheduler(this);
+            }
+        }
+
+        public bool HaveTimers => _timers is not null;
```
Arkatufus, line 168: Actual change, we move the timer into `ActorCell` itself, which is what being used as `ActorContext` inside actor message handlers. The timers instace is created lazily to save memory consuption.

The `HaveTimers` property is needed for cleanup purposes, we can always change this to `internal`

Arkatufus (commented):
self-review

Arkatufus:
DocFX test failing because we delete `IWithTimers`, which is still being referenced in the documentation

PR #7745, opened 2025-08-11
Fix TestKit startup failures under load with SpinWait

### LINKED ISSUE #7744: TestKit: Replace Thread.Sleep with SpinWait to reduce startup failures under load
## Problem
TestKit's `WaitUntilTestActorIsReady` uses `Thread.Sleep(10)` in a tight loop, which:
- Always waits minimum 10ms even if TestActor is ready in microseconds
- Blocks thread pool threads unnecessarily
- Exacerbates thread starvation under high CPU load (e.g., CI with 2 CPUs)
- Causes "Timeout waiting for test actor to be ready" failures

## Root Cause
The TestActor (a simple system actor) typically starts in microseconds, but `Thread.Sleep(10)` forces a minimum 10ms wait. Under resource pressure, this blocking wait makes thread pool starvation worse, leading to timeouts.

## Proposed Solution
Replace `Thread.Sleep(10)` with `SpinWait` which:
- Spins for ~10-20 microseconds (detects ready TestActor almost instantly)
- Then yields to other threads progressively
- Eventually sleeps if needed
- Adapts to system load automatically

## Benefits
- Faster test startup (microseconds vs 10ms minimum)
- Better behavior under load (proper yielding)
- No breaking changes
- Reduces CI timeout failures

Related to #7259, #7423

Aaronontheweb:
## Summary
Replaces `Thread.Sleep(10)` with `SpinWait` in TestKit initialization to reduce startup failures and improve performance.

Fixes #7744

## Changes
- Replace `Thread.Sleep(10)` with `SpinWait` in `WaitUntilTestActorIsReady`
- SpinWait provides adaptive waiting that detects TestActor readiness in microseconds
- Better thread yielding under resource pressure

## Performance Impact
- **Before**: Always waits minimum 10ms even if TestActor ready in microseconds
- **After**: Detects readiness almost instantly, yields properly under load
- Should significantly reduce "Timeout waiting for test actor to be ready" failures in CI

## Compatibility
This is a non-breaking internal optimization. All existing tests continue to work exactly as before.

### DIFF
--- a/src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
+++ b/src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
@@ -142,8 +142,12 @@ namespace Akka.TestKit.Xunit2
             {
                 var extSystem = (ExtendedActorSystem)system;
                 var logger = extSystem.SystemActorOf(Props.Create(() => new TestOutputLogger(Output)), "log-test");
-                logger.Ask<LoggerInitialized>(new InitializeLogger(system.EventStream), TestKitSettings.TestKitStartupTimeout)
-                    .ConfigureAwait(false).GetAwaiter().GetResult();
+                // Start the logger initialization task but don't wait for it yet
+                var loggerTask = logger.Ask<LoggerInitialized>(new InitializeLogger(system.EventStream), TestKitSettings.TestKitStartupTimeout);
+                
+                // By the time TestActor is ready (which happens in base constructor), 
+                // the logger is likely ready too. Now we can safely wait.
+                loggerTask.ConfigureAwait(false).GetAwaiter().GetResult();
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
```diff
@@ -142,8 +142,12 @@ protected void InitializeLogger(ActorSystem system)
             {
                 var extSystem = (ExtendedActorSystem)system;
                 var logger = extSystem.SystemActorOf(Props.Create(() => new TestOutputLogger(Output)), "log-test");
-                logger.Ask<LoggerInitialized>(new InitializeLogger(system.EventStream), TestKitSettings.TestKitStartupTimeout)
-                    .ConfigureAwait(false).GetAwaiter().GetResult();
+                // Start the logger initialization task but don't wait for it yet
```
Aaronontheweb, line 145: The idea behind this was to be able to wrap it around the `WaitUntilTestActorReady` call but that's a bit more work than I thought

Arkatufus (approved):
LGTM

PR #7747, opened 2025-08-12
Fix excessive exception nesting when cancelling ExpectMsgAsync

### LINKED ISSUE #7743: Cancelling ExpectMsgAsync (Xunit) throws excessively nested exception type
**Version Information**
Akka.TestKit.Xunit: 1.5.46

**Describe the bug**
When async message expectation operation is cancelled, the exception has one more level of nesting than expected.

**To Reproduce**
1. Create a test class:
```csharp
using System;
using System.Threading;
using System.Threading.Tasks;

using Akka.TestKit.Xunit;

using Shouldly;

using Xunit;

namespace Actors.Routers.Tests;

[Trait("Category", "Example")]
public class ExampleTests : TestKit
{
    [Fact]
    public async Task SynchronousExample()
    {

        using var stopper = new CancellationTokenSource();
        stopper.Cancel();

        var task = probe.ExpectMsgAsync<int>(
            cancellationToken: stopper.Token).AsTask();

        await Task.Delay(TimeSpan.FromSeconds(1)); // default timeout is 3 seconds

        var inner = task.Exception.ShouldBeOfType<AggregateException>();
        var superInner = inner.InnerException.ShouldBeOfType<AggregateException>();
        superInner.InnerException.ShouldBeOfType<TaskCanceledException>();
    }

    [Fact]
    public async Task AsynchronousExample()
    {

        using var stopper = new CancellationTokenSource();
        stopper.Cancel();

        try
        {
            await probe.ExpectMsgAsync<int>(
                cancellationToken: stopper.Token);
        }
        catch (Exception caught)
        {
            var superInner = caught.ShouldBeOfType<AggregateException>();
            superInner.InnerException.ShouldBeOfType<TaskCanceledException>();
        }
    }
}
```
2. Run the tests.

**Expected behavior**
I expected the tests to fail. I expected the operation to complete with some `OperationCanceledException` wrapped in `AggregateException` as with the most of .Net APIs.

**Actual behavior**
The tests pass. For some reason the method completes with `TaskCanceledException` wrapped in `AggregateException` wrapped in yet another `AggregateException`.

**Environment**
On Windows, dotnet 9.0.303, but I believe it does not matter.

Aaronontheweb:
## Summary
- Fixes #7743 
- Prevents double AggregateException wrapping when ExpectMsgAsync is cancelled
- Ensures cancellation exceptions have minimal, predictable nesting

## Changes
- Modified `InternalTryReceiveOneAsync` to handle cancellation properly without extra exception wrapping
- Added comprehensive test coverage for cancellation scenarios

## Test Plan
- Added `TestKitAsyncCancellationSpec` with 4 test cases covering various cancellation scenarios
- All existing TestKit tests pass (291 passed, 1 skipped)
- Verified the original bug report scenario now works correctly

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Metrics.Tests/Base/AkkaSpecWithCollector.cs
+++ b/src/contrib/cluster/Akka.Cluster.Metrics.Tests/Base/AkkaSpecWithCollector.cs
@@ -66,6 +66,29 @@ namespace Akka.Cluster.Metrics.Tests.Base
             } while (!HasRequiredMetrics(metrics.Metrics, requiredMetrics));
             return metrics;
         }
+        
+        protected async Task<NodeMetrics> CreateTestDataAsync(TimeSpan timeout, string[] requiredMetrics)
+        {
+            using var cts = new CancellationTokenSource(timeout);
+            NodeMetrics metrics;
+            
+            do
+            {
+                cts.Token.ThrowIfCancellationRequested();
+                metrics = Collector.Sample();
+                
+                if (HasRequiredMetrics(metrics.Metrics, requiredMetrics))
+                {
+                    return metrics;
+                }
+                
+                // Small delay between attempts to avoid tight loop
+                await Task.Delay(100, cts.Token);
+                
+            } while (!cts.Token.IsCancellationRequested);
+            
+            throw new OperationCanceledException($"Could not collect required metrics {string.Join(", ", requiredMetrics)} within {timeout}");
+        }
 
         private static bool HasRequiredMetrics(ImmutableHashSet<NodeMetrics.Types.Metric> metrics, string[] requiredMetrics)
         {
--- a/src/contrib/cluster/Akka.Cluster.Metrics.Tests/MetricsCollectorSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Metrics.Tests/MetricsCollectorSpec.cs
@@ -47,12 +47,12 @@ namespace Akka.Cluster.Metrics.Tests
         }
 
         [Fact]
-        public void MetricsCollector_should_collector_accurate_metrics_for_node()
+        public async Task MetricsCollector_should_collector_accurate_metrics_for_node()
         {
             NodeMetrics sample;
             try
             {
-                sample = CreateTestData(10.Seconds(), [
+                sample = await CreateTestDataAsync(30.Seconds(), [
                     StandardMetrics.MemoryAvailable,
                     StandardMetrics.MemoryUsed
                 ]);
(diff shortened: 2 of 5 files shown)


Aaronontheweb (commented):
## Self-Review

### Problem Analysis
The issue (#7743) was that when `ExpectMsgAsync` was cancelled, it threw exceptions with excessive nesting - `AggregateException` containing another `AggregateException` containing the actual `OperationCanceledException`. This made exception handling unpredictable and inconsistent with standard .NET async patterns.

### Root Cause
The excessive nesting occurred in `InternalTryReceiveOneAsync` when using `Task.WhenAny()` with tasks that could be cancelled. The original implementation didn't properly handle the distinction between timeout and user cancellation, leading to multiple layers of exception wrapping.

### Solution Approach

#### Key Changes in `TestKitBase_Receive.cs`:
1. **Replaced `Task.WhenAny` pattern with direct `WaitToReadAsync`**: This eliminates one source of potential exception wrapping
2. **Used linked cancellation token**: Combines the timeout and user cancellation into a single token, making the logic cleaner
3. **Distinguished timeout from cancellation**: 
   - Timeout (normal case) → returns `(false, null)` following the Try pattern
   - User cancellation (exceptional case) → throws `OperationCanceledException` as expected in async methods

The key insight is that `InternalTryReceiveOneAsync` follows a "Try" pattern but should still throw for cancellation (as is standard in async methods). The refined implementation:
```csharp
catch (OperationCanceledException) when (\!cancellationToken.IsCancellationRequested)
{
    // This was a timeout, not user cancellation - return false
    take = (false, null);
}
// If cancellationToken.IsCancellationRequested is true, let the exception propagate
```

### Test Coverage
Added `TestKitAsyncCancellationSpec.cs` with 4 comprehensive tests that verify:
- No double-nested `AggregateException` wrapping
- Proper exception types (`OperationCanceledException` or subclasses)
- Consistent behavior between sync and async APIs
- The exact scenario from the original bug report now works correctly

### Validation
- All 291 existing TestKit tests pass
- New tests specifically validate the fix
- The solution maintains backward compatibility while fixing the issue

This approach ensures cancellation exceptions now have minimal, predictable nesting consistent with standard .NET async patterns.

quixoticaxis:
I assume it would fix not only my issue, but lots of subtle inconsistencies like cancellation being reported as timeout, which currently manifestins as strange-ish error messages along the lines "timout of 100 days reached after 0.03 seconds".

Aaronontheweb:
This is having some unintended consequences elsewhere in the test suite, so we might have to rethink how we're doing this.

Aaronontheweb:
Ah, I think it's the `ConfigureAwait(false)` - which is a major no-no during testing.

Aaronontheweb:
@Arkatufus I think this is ready to review now

Arkatufus (approved):
LGTM

PR #7762, opened 2025-08-16
Bump .NET Core lib target to .NET 8.0

Arkatufus:


Directory.Build.props
MattKotsenas, line 60: SourceLink is built into the .NET 8 SDK: https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/source-link

If you're targeting the 8 TFM, then it's safe to assume you're using the 8+ SDK, so you can remove this reference.

PR #7771, opened 2025-08-19
Resolve: `TestActor` deadlock issue

### LINKED ISSUE #7770: TestKit deadlock during parallel test execution with TestActor initialization
## Description

When running tests in parallel using Akka.TestKit (especially with Akka.Hosting.TestKit), a deadlock can occur during TestActor initialization. This causes tests to fail intermittently when run in parallel.

## Root Cause

The deadlock occurs due to the following sequence:

1. TestActor is created using `SystemActorOf` which calls `AttachChild(props, true, name)` with `async=true`
2. This creates a `RepointableActorRef` that calls `Initialize(async: true)`
3. The `Initialize(true)` method sends a `Supervise` system message but does NOT immediately call `Point()`
4. The TestActor runs on the `CallingThreadDispatcher`
5. The `Supervise` message needs to be processed on the calling thread to complete initialization
6. In Akka.Hosting.TestKit, this happens inside a `WithActors` callback which blocks the startup thread
7. **Deadlock**: The thread is blocked waiting for initialization, but initialization needs that same thread to process the `Supervise` message

## Reproduction

See reproduction test in Akka.Hosting PR: https://github.com/akkadotnet/Akka.Hosting/pull/643

The issue manifests as:
```
System.Exception : Timeout waiting for test actor to be ready
  at TestKitBase.InitializeTest()
```

## Symptoms

- Tests fail randomly when run locally with parallel execution enabled
- Tests may pass in CI/CD pipelines with different threading characteristics
- Changing XUnit's `parallelAlgorithm` setting doesn't fully resolve the issue
- Issue is more likely to occur with high test parallelism (e.g., 40+ concurrent tests)

## Technical Details

The core issue is in `TestKitBase.CreateTestActor`:
- Uses `SystemActorOf` which creates a `RepointableActorRef` with async initialization
- The TestActor uses `CallingThreadDispatcher` (configured in Reference.conf as `akka.test.test-actor.dispatcher`)
- During async initialization, the `Supervise` message must be processed on the calling thread
- If that thread is blocked waiting for initialization, we have a circular dependency

## Affected Components

- Akka.TestKit
- Akka.TestKit.Xunit2
- Akka.Hosting.TestKit
- Any test framework using TestKitBase with parallel execution

## Environment

- Akka.NET: 1.5.45+
- Affects all platforms (Windows, Linux, macOS)
- More pronounced with high CPU core counts and aggressive parallelism

## Workaround

Current workaround is to disable parallel test execution in xunit.runner.json:
```json
{
  "parallelizeTestCollections": false
}
```

However, this significantly increases test execution time.

Aaronontheweb:
## Changes

Fixes #7770

## Checklist

* [x] Design discussion issue #7770

We will need to backport this to the `v1.5` branch also.

### DIFF
--- a/src/core/Akka.TestKit/TestKitBase.cs
+++ b/src/core/Akka.TestKit/TestKitBase.cs
@@ -170,7 +170,7 @@ namespace Akka.TestKit
             if (string.IsNullOrEmpty(testActorName))
                 testActorName = "testActor" + _testActorId.IncrementAndGet();
 
-            var testActor = CreateTestActor(system, testActorName);
+            var testActor = CreateInitialTestActor(system, testActorName);
 
             // Wait for the testactor to start
             WaitUntilTestActorIsReady(testActor, _testState.TestKitSettings);
@@ -718,12 +718,12 @@ namespace Akka.TestKit
             return CreateTestActor(_testState.System, name);
         }
 
-        private IActorRef CreateTestActor(ActorSystem system, string name)
+        private IActorRef CreateInitialTestActor(ActorSystem system, string name)
         {
             var testActorProps = Props.Create(() => new InternalTestActor(_testState.Queue))
                 .WithDispatcher("akka.test.test-actor.dispatcher");
             
-            // IMPORTANT: We need to create the TestActor with special handling to avoid deadlocks
+            // IMPORTANT: For the initial TestActor, we need special handling to avoid deadlocks
             // when running tests in parallel. The default SystemActorOf uses async initialization
             // which can deadlock when the TestActor runs on CallingThreadDispatcher.
             //
@@ -731,9 +731,6 @@ namespace Akka.TestKit
             // Initialize(async) method sends a Supervise message but doesn't immediately call Point().
             // This Supervise message needs to be processed on the CallingThreadDispatcher thread,
             // but that thread may be blocked waiting for initialization to complete.
-            //
-            // For now, we still use SystemActorOf but with enhanced waiting logic below
-            // that can handle the initialization better.
             var systemImpl = system.AsInstanceOf<ActorSystemImpl>();
             var guardian = systemImpl.Provider.SystemGuardian;
             var path = guardian.Path / name;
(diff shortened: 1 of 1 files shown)


src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -710,10 +718,41 @@ public IActorRef CreateTestActor(string name)
             return CreateTestActor(_testState.System, name);
         }
 
+        private IActorRef CreateInitialTestActor(ActorSystem system, string name)
```
Aaronontheweb, line 721: we have a separate method for creating our implicit / default `TestActor` that doesn't share a code path with the others

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -710,10 +718,41 @@ public IActorRef CreateTestActor(string name)
...
+            var path = guardian.Path / name;
+    
+            var dispatcher = systemImpl.Dispatchers.Lookup("akka.test.test-actor.dispatcher");
+            var mailboxType = systemImpl.Mailboxes.GetMailboxType(testActorProps, dispatcher.Configurator.Config);
+    
+            var repointableRef = new RepointableActorRef(
+                systemImpl, testActorProps, dispatcher, mailboxType, guardian, path);
+    
+            // Initialize synchronously by passing async=false
+            repointableRef.Initialize(async: false);
```
Aaronontheweb: this is the key fix - forcing synchronous initialization of the `TestActor` so the `CallingThreadDispatcher` doesn't have to get poked to do anything while we're waiting for the actor to start up.

Aaronontheweb (commented):
Detailed my changes

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -710,10 +718,35 @@ public IActorRef CreateTestActor(string name)
...
         }
 
+        private IActorRef CreateInitialTestActor(ActorSystem system, string name)
+        {
+            var testActorProps = Props.Create(() => new InternalTestActor(_testState.Queue))
+                .WithDispatcher("akka.test.test-actor.dispatcher");
+            
+            var testActor = system.AsInstanceOf<ActorSystemImpl>().SystemActorOf(testActorProps, name);
+            
+            // Force the TestActor to start synchronously to avoid deadlock
```
Aaronontheweb: Took a different tact after the test failures from the first fix, caused by the `TestActor` being incorrectly registered to the `/system` guardian actor. This time around we just force the `CallingThreadDispatcher` to run by sending a dummy message to the `TestActor` - this should cause the queued `Supervise` system message for the `TestActor` to get processed before we even start waiting for test initialization.

PR #7772, opened 2025-08-19
TestKit: synchronous `TestActor` start

### LINKED ISSUE #7770: TestKit deadlock during parallel test execution with TestActor initialization

Aaronontheweb:
## Changes

fixes #7770 - but this is just a prototype
supersedes, closes #7771 

## Checklist

* [x] Design discussion issue #7770


src/contrib/testkits/Akka.TestKit.Xunit/TestKit.cs
```diff
@@ -170,22 +171,39 @@ protected void InitializeLogger(ActorSystem system)
...
             
-        var extSystem = (ExtendedActorSystem)system;
-        var logger = extSystem.SystemActorOf(Props.Create(() => new TestOutputLogger(Output)), "log-test");
-        logger.Ask<LoggerInitialized>(new InitializeLogger(system.EventStream), TestKitSettings.TestKitStartupTimeout)
-            .ConfigureAwait(false).GetAwaiter().GetResult();
+        var systemImpl = system as ActorSystemImpl ?? throw new InvalidOperationException("Expected ActorSystemImpl");
+        
+        // Create logger actor synchronously to avoid deadlock during parallel test execution
+        // Use AttachChildWithAsync with isAsync:false to create LocalActorRef instead of RepointableActorRef
+        var logger = systemImpl.Provider.SystemGuardian.Cell.AttachChildWithAsync(
```
Aaronontheweb, line 178: Using the new `AttachChildWithAsync` method, which allows a top-level actor to start synchronously. Should really only be used during testing. But this means we don't have to wait for the logger to reply back - it should "just work"

Aaronontheweb: This eliminates 1 source of deadlock / contention pressure when running lots of tests in parallel

src/contrib/testkits/Akka.TestKit.Xunit2/TestKit.cs
```diff
@@ -140,30 +141,39 @@ protected void InitializeLogger(ActorSystem system)
...
-                // Start the logger initialization task but don't wait for it yet
-                var loggerTask = logger.Ask<LoggerInitialized>(new InitializeLogger(system.EventStream), TestKitSettings.TestKitStartupTimeout);
+                var systemImpl = system as ActorSystemImpl ?? throw new InvalidOperationException("Expected ActorSystemImpl");
                 
-                // By the time TestActor is ready (which happens in base constructor), 
-                // the logger is likely ready too. Now we can safely wait.
-                loggerTask.ConfigureAwait(false).GetAwaiter().GetResult();
+                // Create logger actor synchronously to avoid deadlock during parallel test execution
+                // Use AttachChildWithAsync with isAsync:false to create LocalActorRef instead of RepointableActorRef
+                var logger = systemImpl.Provider.SystemGuardian.Cell.AttachChildWithAsync(
```
Aaronontheweb, line 148: Same category of fix from before

src/core/Akka.API.Tests/verify/CoreAPISpec.ApproveCore.Net.verified.txt
```diff
@@ -27,6 +27,8 @@
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("Akka.Streams.Tests")]
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("Akka.TestKit")]
 [assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("Akka.TestKit.Tests")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute("Akka.TestKit.Xunit")]
```
Aaronontheweb, line 30: Needed to make the xUnit testkits friend assemblies in order to access the `AttachChildWithAsync`

src/core/Akka.TestKit.Tests/TestActorRefTests/ParallelTestActorDeadlockSpec.cs
```diff
@@ -0,0 +1,89 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.TestKit.Tests.TestActorRefTests
+{
+    public class ParallelTestActorDeadlockSpec
```
Aaronontheweb, line 10: Repurposed @Arkatufus 's reproduction from https://github.com/akkadotnet/Akka.Hosting/pull/643 - got it to pass with these changes. Both the logger and the test initialization changes were required.

src/core/Akka.TestKit/TestKitBase.cs
```diff
@@ -710,10 +673,31 @@ public IActorRef CreateTestActor(string name)
...
+        {
+            // Fix both serialization and deadlock issues:
+            // 1. Use isSystemService=true to skip serialization checks
+            // 2. Use isAsync=false to create LocalActorRef synchronously (avoids RepointableActorRef deadlock)
+            var testActorProps = Props.Create(() => new InternalTestActor(_testState.Queue))
+                .WithDispatcher("akka.test.test-actor.dispatcher");
+            
+            var systemImpl = system.AsInstanceOf<ActorSystemImpl>();
+            // Use the new AttachChildWithAsync method to create TestActor synchronously
+            var testActor = systemImpl.Provider.SystemGuardian.Cell.AttachChildWithAsync(
```
Aaronontheweb, line 686: Key fix: start the actor with this method so it's automatically initialized with an `ActorCell` - no need to spin and check afterwards

src/core/Akka/Actor/ActorCell.Children.cs
```diff
@@ -98,6 +98,22 @@ public virtual IActorRef AttachChild(Props props, bool isSystemService, string?
...
+        /// Used by TestKit to create TestActors synchronously to avoid deadlocks.
+        /// </summary>
+        /// <param name="props">The <see cref="Props"/> this child actor will use.</param>
+        /// <param name="isSystemService">If <c>true</c>, then this actor is a system actor and skips serialization checks.</param>
+        /// <param name="isAsync">If <c>true</c>, creates RepointableActorRef with async init. If <c>false</c>, creates LocalActorRef synchronously.</param>
+        /// <param name="name">The name of the actor being started. Can be <c>null</c> for auto-generated name.</param>
+        /// <returns>A reference to the initialized child actor.</returns>
+        internal IActorRef AttachChildWithAsync(Props props, bool isSystemService, bool isAsync, string? name = null)
+        {
+            return MakeChild(props, name == null ? GetRandomActorName() : CheckName(name), isAsync, isSystemService);
```
Aaronontheweb, line 115: The key fix - the ability to attach a system service actor and have it start synchronously. 

This, very likely, screws with actor supervision if there are any crashes at startup, hence why this API is `internal` and should not be used outside of this one narrow case.

Aaronontheweb (commented):
Detailed my changes

Arkatufus (approved):
LGTM

PR #7789, opened 2025-08-22
Port #7769 to 1.5 branch, remove FluentAssertions dependency from all TestKit

Arkatufus:
## Changes

* Remove FluentAssertions dependency from all TestKit
* Fix assertion logic

(cherry picked from commit d924e57e205527d2e2c263792f52af290b22709a)

> [!NOTE]
> Would need to be propagated to other TestKit implementation if released.


Aaronontheweb:
We'll leave this as a v1.6-only change for now.

PR #7790, opened 2025-08-25
[TestKit] Enable nullability for Akka.TestKit.TestKitBase

Arkatufus:
## Changes

* Enable nullability
* Remove `DynamicInvoke()` reflection call
* Wrap all NRE prone property calls with clear exceptions


Aaronontheweb:
@Arkatufus lots of test failures here

Arkatufus:
it should be fixed now

PR #7791, opened 2025-08-26
Parameterize Incrementalist base branch for Azure DevOps pipelines

Aaronontheweb:
## Summary
- Adds dynamic base branch detection for Incrementalist in Azure DevOps pipelines
- Prevents unnecessary full builds when PRs target version branches (e.g., v1.5) instead of dev

## Changes
- Added `incrementalist.baseBranch` variable to pipeline templates that:
  - Uses `System.PullRequest.TargetBranch` for PR builds
  - Defaults to 'dev' for non-PR builds
- Updated all Incrementalist commands to use `--branch $(incrementalist.baseBranch)` parameter

## Impact
When a PR targets v1.5, Incrementalist will now correctly compare against v1.5 instead of dev, avoiding full builds when only specific projects have changed between v1.5 and the PR branch.

This solves the issue where PRs to version branches would always trigger full builds because Incrementalist was comparing against dev (which contains v1.6 code) instead of the actual target branch.

## Test Plan
The pipeline changes will be validated when this PR runs through Azure DevOps.

### DIFF
--- a/build-system/azure-pipeline.mntr-template.yaml
+++ b/build-system/azure-pipeline.mntr-template.yaml
@@ -14,19 +14,25 @@ jobs:
     timeoutInMinutes: ${{ parameters.timeoutInMinutes }}
     pool:
       vmImage: ${{ parameters.vmImage }}
-    variables:
-      # Determine the base branch for Incrementalist based on the target branch
-      - name: incrementalist.baseBranch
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
This should, ideally, prevent any future PRs being sent to the `v1.5` branch from getting compared to `dev`, which will trigger a full rebuild each time

PR #7793, opened 2025-09-02
Fix IIS/Windows Service console race condition (#7691)

Aaronontheweb:
## Changes

- Detect when running in IIS/Windows Service environments where Console.Out and Console.Error are redirected to the same StreamWriter.Null singleton
- Skip console output entirely in these environments to prevent race conditions that cause IndexOutOfRangeException and cascade failures
- Improve DefaultLogger error handling to prevent feedback loops
- Add unit tests for non-console scenarios

The race condition occurs because:
1. IIS/Services redirect both Console.Out and Console.Error to StreamWriter.Null
2. StreamWriter.Null is a singleton, not thread-safe for concurrent access
3. Multiple threads writing to both streams cause IndexOutOfRangeException
4. Console output goes nowhere in these environments anyway

Fixes #7691

## Checklist

For significant changes, please ensure that the following have been completed (delete if not relevant):

* [x] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).
* [x] This change follows the [Akka.NET Wire Compatibility Guidelines](https://getakka.net/community/contributing/wire-compatibility.html).
* [ ] I have [reviewed my own pull request](https://getakka.net/community/contributing/index.html#review-your-own-pull-requests).
* [x] Design discussion issue #7691

PR #7796, opened 2025-09-03
Fix `StreamOfStreams` with atomic CompareExchange

Aaronontheweb:
## Summary

Fixes race condition in `FlowPrefixAndTailSpec.PrefixAndTail_must_throw_if_tail_is_attempted_to_be_materialized_twice` that was causing intermittent test failures on CI.

**Error**: Test expected `OnError` but received `OnNext(2)` instead.

## Root Cause

The issue was a read-then-CAS race condition in `SubSource<T>.Logic.SetCallback()`:

1. **Thread 1**: reads status (null) → `CompareAndSet` succeeds → starts data flow
2. **Thread 2**: reads status (stale null) → `CompareAndSet` fails → does separate read for retry
3. **Race window**: Between Thread 2's CAS failure and subsequent read, Thread 1's `OnNext(2)` flows to wrong subscriber

## Solution

1. **Added `CompareExchange` method to `AtomicReference<T>`**: Returns actual previous value instead of just boolean, following standard .NET `Interlocked` patterns

2. **Replaced read-then-CAS with single atomic operation**:
   ```csharp
   // Before (racy)
   var status = _stage._status.Value; 
   if (status == null) { 
       if (\!_stage._status.CompareAndSet(null, callback)) // Race window here
   }
   
   // After (atomic)
   var previous = _stage._status.CompareExchange(null, callback);
   switch (previous) { ... }
   ```

This eliminates all race conditions by using a single atomic operation that both attempts the change AND returns what was actually there.

## Testing

- Test now passes consistently (ran 50+ times without failure)
- No other tests affected by `AtomicReference` changes

## Files Changed
- `src/core/Akka/Util/AtomicReference.cs` - Added `CompareExchange` method
- `src/core/Akka.Streams/Implementation/Fusing/StreamOfStreams.cs` - Fixed race condition

Fixes intermittent CI failures from PR #7793.

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/FlowIdleInjectSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/FlowIdleInjectSpec.cs
@@ -137,7 +137,8 @@ namespace Akka.Streams.Tests.Dsl
                 .RunWith(Sink.FromSubscriber(downstream), Materializer);
 
             await downstream.RequestAsync(10);
-            downstream.ExpectNextN(Enumerable.Range(1, 10));
+            var received = downstream.ExpectNextN(10);
+            received.OrderBy(x => x).Should().BeEquivalentTo(Enumerable.Range(1, 10));
             
             await downstream.ExpectNoMsgAsync(TimeSpan.FromSeconds(1.5));
             await downstream.RequestAsync(1);
@@ -181,7 +182,8 @@ namespace Akka.Streams.Tests.Dsl
                 .RunWith(Sink.FromSubscriber(downstream), Materializer);
 
             await downstream.RequestAsync(10);
-            downstream.ExpectNextN(Enumerable.Range(1, 10));
+            var received = downstream.ExpectNextN(10);
+            received.OrderBy(x => x).Should().BeEquivalentTo(Enumerable.Range(1, 10));
             
             await downstream.ExpectNoMsgAsync(TimeSpan.FromSeconds(1.5));
             await upstream.SendNextAsync(1);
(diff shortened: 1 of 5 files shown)


src/core/Akka.Streams/Implementation/Fusing/StreamOfStreams.cs
```diff
@@ -1242,31 +1242,44 @@ public Logic(SubSource<T> stage) : base(stage.Shape)
 
             private void SetCallback(Action<IActorSubscriberMessage> callback)
             {
-                var status = _stage._status.Value;
-
-                if (status == null)
+                // Single atomic operation that both attempts the change AND returns the previous value
+                var previous = _stage._status.CompareExchange(null, callback);
```
Aaronontheweb, line 1246: This is the real fix - only read the value once, as the result of the `CompareExchange`

Aaronontheweb:
This is, for sure, a bug that may be affecting some other tests too - looks like this was the worst offender though.

src/core/Akka.Streams.Tests/Dsl/FlowIdleInjectSpec.cs
```diff
@@ -137,7 +137,8 @@ public async Task KeepAlive_must_work_if_timer_fires_before_initial_request_afte
                 .RunWith(Sink.FromSubscriber(downstream), Materializer);
 
             await downstream.RequestAsync(10);
-            downstream.ExpectNextN(Enumerable.Range(1, 10));
+            var received = downstream.ExpectNextN(10);
+            received.OrderBy(x => x).Should().BeEquivalentTo(Enumerable.Range(1, 10));
```
Aaronontheweb, line 141: Made the ordering deterministics

Aaronontheweb:
This will need a backport to `v1.5`

Arkatufus (approved):
LGTM, waiting for CI/CD

PR #7801, opened 2025-09-04
Fix race condition in `ClusterLogVerboseDefaultSpec` test

Aaronontheweb:
## Changes

The test was using a convoluted pattern of expecting a timeout exception when verbose logging was disabled. This created a race condition with nested timeouts.

Changed to directly assert that 0 verbose log messages are emitted when verbose logging is disabled, making the test more deterministic and clear.

- Use E`ventFilter.ExpectAsync(0, ...)` instead of expecting timeout failure
- Remove nested timeout complexity with `WaitAsync(11.Seconds())`
- Test now directly asserts the expected behavior (no verbose logs)

## Checklist

### DIFF
--- a/src/core/Akka.Cluster.Tests/ClusterLogSpec.cs
+++ b/src/core/Akka.Cluster.Tests/ClusterLogSpec.cs
@@ -125,9 +125,35 @@ namespace Akka.Cluster.Tests
         public async Task A_cluster_must_not_log_verbose_cluster_events_by_default()
         {
             _cluster.Settings.LogInfoVerbose.ShouldBeFalse();
-            await AssertThrowsAsync<XunitException>(() => JoinAsync(upLogMessage)).WaitAsync(11.Seconds());
+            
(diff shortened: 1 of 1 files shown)


src/core/Akka.Cluster.Tests/ClusterLogSpec.cs
```diff
@@ -125,9 +125,35 @@ public ClusterLogVerboseDefaultSpec(ITestOutputHelper output)
         public async Task A_cluster_must_not_log_verbose_cluster_events_by_default()
         {
             _cluster.Settings.LogInfoVerbose.ShouldBeFalse();
-            await AssertThrowsAsync<XunitException>(() => JoinAsync(upLogMessage)).WaitAsync(11.Seconds());
+            
+            // Join the cluster but verify verbose log messages are NOT emitted
+            await EventFilter
```
Aaronontheweb, line 130: Avoids the "double negative" testing pattern we were using before - we now expect the cluster to join to succeed and simply not log anything verbosely.

PR #7808, opened 2025-09-04
GraphStageLogic.ConcurrentAsyncCallback throws NRE when used with ChannelSource

### LINKED ISSUE #7794: Null Exception at Akka.Streams.Stage.GraphStageLogic.ConcurrentAsyncCallback with Generic channel types
**Version Information**
1.5.46
Akka.DependencyInjection
Akka.Hosting
Akka.Streams

**Describe the bug**
The following error is encountered:

``` 
at Akka.Streams.Stage.GraphStageLogic.ConcurrentAsyncCallback`1.<>c__DisplayClass3_0.<.ctor>b__0(Object obj)
at Akka.Streams.Implementation.Fusing.GraphInterpreter.RunAsyncInput(GraphStageLogic logic, Object evt, TaskCompletionSource`1 promise, Action`1 handler)
```
 

**To Reproduce**

 ```

using var system = ActorSystem.Create("ChannelExample");
using var actorSystemMaterializer = system.Materializer();
var channel = Channel.CreateBounded<Message<string, string>>(new BoundedChannelOptions(capacity: 100)
{
  FullMode = BoundedChannelFullMode.Wait,
  SingleReader = true,
  SingleWriter = true,
  AllowSynchronousContinuations = false
});


var streamRes = ChannelSource.FromReader(channel.Reader)
    .Select(e => e)
    .RunWith(Sink.Ignore<Message<string, string>>(), actorSystemMaterializer);

_ = Task.Run(async () =>
{
  await Task.Delay(100);
  channel.Writer.Complete();
});

await streamRes;

```
`Message` is defined as:
```
public class Message<TKey, TValue>
{
    public TKey Key { get; set; }
    public TValue Value { get; set; }
}
```

The problem is not reproduced when sending on `Channel<string>` i.e., in this example:
```
using var system = ActorSystem.Create("ChannelExample");
using var actorSystemMaterializer = system.Materializer();
var channel = Channel.CreateBounded<string>(new BoundedChannelOptions(capacity: 100)
{
 FullMode = BoundedChannelFullMode.Wait,
 SingleReader = true,
 SingleWriter = true,
 AllowSynchronousContinuations = false
});


var streamRes = ChannelSource.FromReader(channel.Reader)
   .Select(e => e)
   .RunWith(Sink.Ignore<Message<string, string>>(), actorSystemMaterializer);

_ = Task.Run(async () =>
{
 await Task.Delay(100);
 channel.Writer.Complete();
});

await streamRes;
```

**Expected behavior**
No error should be encountered when using object types in channels

**Actual behavior**
Error encountered
 
**Environment**
Linux, Dotnet 9

### PR DESCRIPTION
**Pull request by Arkatufus**
Fixes #7794

## Changes

Make sure ConcurrentAsyncCallback can handle null events properly.

## Actual Bug

* `ChannelSource.OnReaderComplete` accepts a nullable `Exception` as an argument (null sentinel) and was passed in as a logic async callback. 
* When this callback was invoked with null value, `ConcurrentAsyncCallback` would not accept null as a value and considered it as an invalid event and tried to raise an `ArgumentException` exception.
* `ConcurrentAsyncCallback` throws a `NullReferenceException` because it tried to call `obj.GetType()` inside string extrapolation because `obj` is null.

### DIFF
--- a/src/core/Akka.Streams/Stage/GraphStage.cs
+++ b/src/core/Akka.Streams/Stage/GraphStage.cs
@@ -885,11 +885,18 @@ namespace Akka.Streams.Stage
                 _ownedStage = ownedStage;
                 _wrappedHandler = obj =>
                 {
-                    if (obj is T e)
-                        handler1(e);
-                    else
-                        throw new ArgumentException(
-                            $"Expected {nameof(obj)} to be of type {typeof(T)}, but was {obj.GetType()}");
+                    switch (obj)
+                    {
+                        // Always assume that T can be null and the handler will handle null values
+                        case null:
+                            handler1(default);
+                            break;
+                        case T e:
+                            handler1(e);
+                            break;
+                        default:
+                            throw new ArgumentException($"Expected {nameof(obj)} to be of type {typeof(T)}, but was {obj.GetType()}");
+                    }
                 };
             }
 
--- a/src/core/Akka.Streams.Tests/Issue7794Spec.cs
+++ b/src/core/Akka.Streams.Tests/Issue7794Spec.cs
@@ -1,0 +1,55 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="Issue7794Spec.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2025 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Akka.Streams.Dsl;
+using Akka.TestKit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Akka.Streams.Tests.Implementation;
+
+public class Issue7794Spec: AkkaSpec
+{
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/core/Akka.Streams/Stage/GraphStage.cs [hunk 1]
```diff
@@ -885,11 +885,18 @@ public ConcurrentAsyncCallback(Action<T> handler, GraphStageLogic ownedStage)
                 _ownedStage = ownedStage;
                 _wrappedHandler = obj =>
                 {
-                    if (obj is T e)
-                        handler1(e);
-                    else
-                        throw new ArgumentException(
-                            $"Expected {nameof(obj)} to be of type {typeof(T)}, but was {obj.GetType()}");
```
Arkatufus, line 892: The NRE was thrown here by `obj.GetType()` when obj is null.

The actual fix was to make sure that this wrapper also handles null and passes it to the handler delegate

Arkatufus (commented):
Self-review

src/core/Akka.Streams/Stage/GraphStage.cs
Arkatufus: Note, this should be fixed using `#nullable enable` later to make it explicit that `T` can be null

Aaronontheweb (approved):
Changes look tentatively good to me, so long as the test suite agrees

PR #7814, opened 2025-09-08
[Tests] Harden Persistence.InMemory event Tests

Arkatufus:
# Test bed for testing Persistence.Query.InMemory flaky unit tests


Aaronontheweb (approved):
LGTM

Arkatufus:
The original fix did not fix the underlying issue, still exploring for other possible causes

Arkatufus:
I'm turning this PR into a test bed to probe for other possible causes, DO NOT MERGE

Aaronontheweb (changes requested):
I think this is over-complicated and touches some dangerous stuff like `RepointableActorRef`. Going to close it.

PR #7820, opened 2025-09-12
POC - Persistence Test Kit for xunit v3

mhbuck:
I was working on a project and upgrading test kits and I came across the issue with the Persistence test kit being xunit 2 only

This is a proof of concept of how I think potentially can handle the other test kits similar to how the base testkit has been created. I have only tried this out on the Persistence test kit so far just to see if the approach is what should be used for the others.


src/core/Akka.Persistence.TestKit.Xunit/Akka.Persistence.TestKit.Xunit.csproj
```diff
@@ -0,0 +1,20 @@
...
+
+  <PropertyGroup>
+    <Description>TestKit for writing tests for Akka.NET Persistance module using xUnit</Description>
+    <TargetFrameworks>$(NetStandardLibVersion);$(NetLibVersion)</TargetFrameworks>
+    <PackageTags>$(AkkaPackageTags);testkit;persistance;xunit</PackageTags>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\contrib\testkits\Akka.TestKit.Xunit\Akka.TestKit.Xunit.csproj" />
```
mhbuck, line 11: Pulling the reference in here brings in xunit v3

src/core/Akka.Persistence.TestKit.Xunit/PersistenceTestKit.cs
```diff
@@ -0,0 +1,367 @@
...
+
+
+namespace Akka.Persistence.TestKit
+{
+    using System;
+    using System.Threading.Tasks;
+    using Actor;
+    using Akka.TestKit.Xunit;
+    using Configuration;
+    using Xunit;
```
mhbuck, line 19: The only changes to this file are the xunit references the rest is a full copy of the file. What theoretically could be done is creating a base class of the `PersistenceTestKit` in the base library and only have the constructors in the two xunit specific projects to minimise the copied code.

src/core/Akka.Persistence.TestKit/Akka.Persistence.TestKit.csproj
```diff
@@ -12,7 +12,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\contrib\testkits\Akka.TestKit.Xunit2\Akka.TestKit.Xunit2.csproj" />
```
mhbuck, line 15: I think this would be a breaking change someone could be referencing this project and only using the `Akka.TestKit` and not the `Akka.Persistence.TestKit` Not sure if it would be a problem but just calling that out

src/core/Akka.Persistence.TestKit.Xunit/PersistenceTestKit.cs
mhbuck: There have been no other ways around duplicating the code that I can think of here without the references

mhbuck:
I wanted to see if I could get the Persistence tests all running using the new project and I was able to hack out quite a bit of data to get parts of the tests in a position to run with xunit3 and it has mostly worked. There are quite a few things that rely on some xunit v2 references but it does "prove" the new project added works.
image

Arkatufus:
Already addressed with the xUnit 3 ports

PR #7821, opened 2025-09-12
Fix missing error when SSL private key is missing

Arkatufus:
## Changes

* Add fail fast path if SSL certificate doesn't have a private key


Aaronontheweb:
Superseded via https://github.com/akkadotnet/akka.net/pull/7824

PR #7826, opened 2025-09-16
Harden Akka.Streams.Tests.FlowLogSpec

Arkatufus:
## Summary
- Fixes a racy/flaky test in `Akka.Streams.Tests` (`FlowLogSpec`) observed on constrained CI runners.
- Root cause: test captured all EventStream traffic and then casted blindly to `Debug`, which intermittently included non-`Debug` messages (e.g., `SuppressedDeadLetter`).
- Resolution: scope log capture to the specific stage messages by tag and type, using `FishForMessageAsync`-based helpers. No timeouts changed.

## Root Cause
- The test previously subscribed a probe to `typeof(object)` and used `ReceiveN(n).Cast<Debug>()` to read the next N logs.
- On busy/slow CI, unrelated EventStream events (e.g., `Akka.Event.SuppressedDeadLetter`) interleaved and caused `InvalidCastException` when casting to `Debug`.
- Additionally, asserting “next message is X” (`ExpectMsgAsync<T>`) was fragile under DEBUG-level noise.

## Changes
- Probe now subscribes to `typeof(LogEvent)` only, excluding non-log traffic.
- Added helpers:
  - `LogEvents<T>(n, tag)` and `LogMessages(n, tag)` that use `FishForMessageAsync` with a prefix tag (e.g., `[my-debug]`) to capture only the stage’s log messages, in order.
- Replaced brittle `ReceiveN(n).Cast<Debug>()` and raw `ExpectMsgAsync<T>` uses with tag- and type-filtered fishing across the spec.
- Kept test logic and expected messages intact; only the capture mechanism changed.

## Why Deterministic
- Filters by `LogEvent` type and `[tag]` prefix ensure the probe ignores unrelated EventStream messages.
- `FishForMessageAsync` waits for the exact expected message without draining needed entries or depending on head-of-queue type.
- Unique tags per test isolate assertions even in noisy DEBUG environments.

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/FlowLogSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/FlowLogSpec.cs
@@ -195,4 +195,3 @@ namespace Akka.Streams.Tests.Dsl
         }
     }
 }
-
--- a/src/core/Akka.Streams.Tests/Dsl/FlowLogSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/FlowLogSpec.cs
@@ -6,6 +6,7 @@
 //-----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Akka.Event;
@@ -40,9 +41,23 @@ namespace Akka.Streams.Tests.Dsl
             _logProbe = p;
         }
 
-        private ValueTask<string[]> LogMessages(int n)
-            => _logProbe.ReceiveWhileAsync<Debug>(shouldContinue:_ => true, msgs: n)
-                .Select(x => x.Message.ToString()).ToArrayAsync();
+        private async Task<string[]> LogMessages(int n, string tag)
+        {
+            return (await LogEvents<Debug>(n, tag)).Select(m => m.Message.ToString()).ToArray();
+        }
+
+        private async Task<T[]> LogEvents<T>(int n, string tag) where T : LogEvent
+        {
+            var count = 0;
+            var messages = new List<T>();
+            while (count < n)
+            {
+                var msg = (T) await _logProbe.FishForMessageAsync(m => m is T d && d.Message.ToString().StartsWith(tag));
+                count++;
+                messages.Add(msg);
+            }
+            return messages.ToArray();
+        }
 
         private static readonly Type LogType = typeof (IMaterializer);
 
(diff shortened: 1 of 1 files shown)


src/core/Akka.Streams.Tests/Dsl/FlowLogSpec.cs
```diff
@@ -26,145 +28,169 @@ public FlowLogSpec(ITestOutputHelper helper) : base("akka.loglevel = DEBUG", hel
...
+        private TestProbe _logProbe;
 
+        protected override void AtStartup()
+        {
+            base.AtStartup();
+            
             var p = CreateTestProbe();
-            Sys.EventStream.Subscribe(p.Ref, typeof(object));
-            LogProbe = p;
+            Sys.EventStream.Subscribe(p.Ref, typeof(LogEvent));
```
Arkatufus, line 40: Don't subscribe to everything, we're only interested in `Debug`, `Info`, `Warning` and `Error`

src/core/Akka.Streams.Tests/Dsl/FlowLogSpec.cs
```diff
@@ -26,145 +28,169 @@ public FlowLogSpec(ITestOutputHelper helper) : base("akka.loglevel = DEBUG", hel
...
+            var count = 0;
+            var messages = new List<T>();
+            while (count < n)
+            {
+                var msg = (T) await _logProbe.FishForMessageAsync(m => m is T d && d.Message.ToString().StartsWith(tag));
+                count++;
+                messages.Add(msg);
+            }
+            return messages.ToArray();
+        }
```
Arkatufus, line 60: custom "ReceiveN" helper, instead of blindly pulling messages, limit the messages to the explicit type and tags that we know must be emitted

src/core/Akka.Streams.Tests/Dsl/FlowLogSpec.cs
```diff
@@ -26,145 +28,169 @@ public FlowLogSpec(ITestOutputHelper helper) : base("akka.loglevel = DEBUG", hel
...
-            Source.Single(42)
+            _ = Source.Single(42)
                 .Log("flow-6")
                 .WithAttributes(Attributes.CreateLogLevels(LogLevel.WarningLevel, LogLevel.InfoLevel,
                     LogLevel.DebugLevel))
                 .RunWith(Sink.Ignore<int>(), Materializer);
 
-            LogProbe.ExpectMsg<Warning>().Message.ToString().Should().Be("[flow-6] Element: 42");
-            LogProbe.ExpectMsg<Info>().Message.ToString().Should().Be("[flow-6] Upstream finished.");
+            var warnings = await LogEvents<Warning>(1, "[flow-6]");
```
Arkatufus, line 157: `ExpectMsg` replaced with the helper function `LogEvents`. We want to make sure that we're not reading any event bus noise

Arkatufus (commented):
Self-review

Aaronontheweb (approved):
LGTM - seems much more deterministic than before.

PR #7829, opened 2025-09-16
Harden Akka.Streams.Tests.Dsl.HubSpec.MergeHub_must_keep_working_even_if_one_of_the_producers_fail

Arkatufus:
# Fix flaky HubSpec: MergeHub_must_keep_working_even_if_one_of_the_producers_fail

## Summary
- Stabilizes a racy test that intermittently timed out in CI (AzDO).
- Replaces EventFilter-based log assertion with deterministic EventStream subscription and enforces failure-before-cancel ordering.
- No production code or timeouts changed.

## Theory
- Flake stems from two races: (1) EventFilter installed asynchronously can miss fast error logs; (2) downstream `Take(10)` may complete and cancel producers before a failing producer signals, which (by design) produces no error log.

## Assumptions
- Preserve MergeHub semantics: producer failure logs `MergeHub.ProducerFailed`; cancellation does not log.
- Do not increase timeouts; make ordering deterministic.
- EventStream subscription on `TestActor` is synchronous; `TestPublisher.Probe` lets us drive failure explicitly.

## Fix
- Replace EventFilter with `Sys.EventStream.Subscribe(TestActor, typeof(Error))`.
- Attach a controllable failing producer (`TestPublisher.Probe<int>`), ensure subscription, then `SendError(new TestException("failing"))`.
- Fish for the `Error` and assert `Cause` is `MergeHub.ProducerFailed`.
- Start the healthy producer after the failure and assert downstream result.
- Test-only change; no runtime code or timeouts modified.

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/HubSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/HubSpec.cs
@@ -1013,3 +1013,4 @@ namespace Akka.Streams.Tests.Dsl
         }
     }
 }
+
--- a/src/core/Akka.Streams.Tests/Dsl/HubSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/HubSpec.cs
@@ -1013,4 +1013,3 @@ namespace Akka.Streams.Tests.Dsl
         }
     }
 }
-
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Superseded via #7830

PR #7830, opened 2025-09-18
Akka.Streams: Fix `Source.Failed` race condition by implementing as `GraphStage`

Aaronontheweb:
## Description

Fixes the intermittent test failure in `MergeHub_must_keep_working_even_if_one_of_the_producers_fail` by addressing the root cause: a fundamental design difference between .NET and Scala's implementation of `Source.Failed`.

## Root Cause Analysis

The race condition occurred because:

1. **.NET's implementation** used `ErrorPublisher` which immediately calls `subscriber.OnError()` **synchronously** during `Subscribe()`
2. This bypassed the normal GraphInterpreter event queue and could occur before the graph was properly initialized
3. When the `MergeHub.ProducerFailed` exception was thrown from `OnUpstreamFailure`, the GraphInterpreter's `ReportStageError` method would check if `ActiveStage == null`
4. Since the error happened so early, `ActiveStage` was often null, causing the exception to be **rethrown instead of logged**
5. The EventFilter in the test would timeout waiting for an error log that never appeared

## The Solution

Align with Scala's implementation by replacing `ErrorPublisher` with a proper `GraphStage`:

### Scala's Approach (No Race Condition)
```scala
def failed[T](cause: Throwable): Source[T, NotUsed] =
  Source.fromGraph(new FailedSource[T](cause))

class FailedSource[T](failure: Throwable) extends GraphStage[SourceShape[T]] {
  override def preStart(): Unit = {
    failStage(failure)  // Fails during PreStart lifecycle
  }
}
```

### .NET's New Implementation (Fixed)
```csharp
public static Source<T, NotUsed> Failed<T>(Exception cause)
{
    return Source.FromGraph(new Implementation.FailedSource<T>(cause, "FailedSource"));
}

internal sealed class FailedSource<T> : GraphStage<SourceShape<T>>
{
    public override void PreStart()
    {
        FailStage(_failure); // Now fails during PreStart like Scala
    }
}
```

## Key Changes

1. **Added `FailedSource` GraphStage** (`src/core/Akka.Streams/Implementation/FailedSource.cs`)
   - Implements failure during `PreStart()` lifecycle hook
   - Ensures GraphInterpreter is properly initialized when the error occurs
   - `ActiveStage` is properly set, so errors are logged correctly

2. **Modified `Source.Failed`** to use `FailedSource` instead of `ErrorPublisher`
   - `ErrorPublisher` is still retained for internal use in `IOSources` and `StreamLayout`

3. **Reverted all defensive logging workarounds** in `Hub.cs`
   - No longer needed since we fixed the root cause

## Testing

The fix has been verified by running the test multiple times without any failures. The race condition is completely eliminated because failures now go through the proper GraphInterpreter lifecycle.

## Related Issues

- Supersedes PR #7829 which attempted to work around the issue by modifying the test
- The issue was specific to .NET's implementation and didn't exist in Scala Akka

## Breaking Changes

None. This is an internal implementation change that maintains the same public API and behavior.


Aaronontheweb:
This fix isn't right yet - just experimenting

src/core/Akka.Streams/Dsl/Source.cs
```diff
@@ -844,10 +844,7 @@ public static Source<T, TaskCompletionSource<T>> Maybe<T>()
         /// <returns>TBD</returns>
         public static Source<T, NotUsed> Failed<T>(Exception cause)
         {
-            return new Source<T, NotUsed>(new PublisherSource<T>(
-                new ErrorPublisher<T>(cause, "FailedSource"),
-                DefaultAttributes.FailedSource,
-                Shape<T>("FailedSource")));
+            return Source.FromGraph(new Implementation.FailedSource<T>(cause, "FailedSource"));
```
Aaronontheweb, line 847: `ErrorPublisher` is still used in other places, but not in the same "early failure" mode that it was used for here inside `Source.Failed<T>` - so we aren't going to replace it entirely.

src/core/Akka.Streams/Implementation/FailedSource.cs
```diff
@@ -0,0 +1,67 @@
...
+                _source = source;
+                SetHandler(source.Out, this);
+            }
+
+            public override void PreStart()
+            {
+                // Fail the stage during the PreStart lifecycle hook
+                // This ensures the failure happens after the GraphInterpreter
+                // is properly initialized and ActiveStage is set
+                FailStage(_source._failure);
```
Aaronontheweb, line 57: See the comment above.

Aaronontheweb (commented):
Detailed my changes

Arkatufus (approved):
LGTM

PR #7831, opened 2025-09-18
Fix race condition in ClusterLogSpec tests

Aaronontheweb:
## Summary
- Fixed intermittent test failures in `ClusterLogVerboseEnabledSpec` caused by race condition
- Added proper synchronization using Identify/ActorIdentity pattern
- Applied fix consistently across all ClusterLogSpec test classes

## Problem
The `ClusterLogVerboseEnabledSpec.A_cluster_must_log_verbose_cluster_events_when_log_info_verbose_is_on` test was failing intermittently with timeout errors when waiting for the "event MemberUp" log message.

Root cause: The `EventBusListener` actor (which logs verbose cluster events) wasn't fully subscribed to cluster events before the `MemberUp` event was published during cluster join.

## Solution
Added `EnsureEventBusListenerReadyAsync()` method that:
1. Sends an `Identify` message to the EventBusListener actor
2. Waits for an `ActorIdentity` response
3. Ensures the actor is ready before proceeding with tests

This approach is more reliable than using arbitrary delays and follows Akka best practices for actor synchronization.

## Test Results
- All ClusterLogSpec tests pass consistently
- Ran tests multiple times to verify stability
- No flaky behavior observed

## Related Issues
Addresses ongoing CI failures with this test that have been happening recently.

### DIFF
--- a/src/core/Akka.Cluster.Tests/ClusterLogSpec.cs
+++ b/src/core/Akka.Cluster.Tests/ClusterLogSpec.cs
@@ -22,7 +22,7 @@ namespace Akka.Cluster.Tests
 {
     public abstract class ClusterLogSpec : AkkaSpec
     {
-        public const string Config = @"    
+        public const string Config = @"
             akka.cluster {
               auto-down-unreachable-after = 0s
               publish-stats-interval = 0s # always, when it happens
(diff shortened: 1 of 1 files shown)


src/core/Akka.Cluster.Tests/ClusterLogSpec.cs
```diff
@@ -48,6 +48,18 @@ protected ClusterLogSpec(ITestOutputHelper output, Config config = null)
             _cluster = Cluster.Get(Sys);
         }
 
+        /// <summary>
+        /// Ensures the EventBusListener is ready by sending an Identify message and waiting for response.
+        /// This prevents race conditions where cluster events are published before the listener is subscribed.
+        /// </summary>
+        protected async Task EnsureEventBusListenerReadyAsync()
```
Aaronontheweb, line 55: XML-DOC Comment explains the key fix.

src/core/Akka.Cluster.Tests/ClusterLogSpec.cs
```diff
@@ -169,6 +188,10 @@ public ClusterLogVerboseEnabledSpec(ITestOutputHelper output)
         public async Task A_cluster_must_log_verbose_cluster_events_when_log_info_verbose_is_on()
         {
             _cluster.Settings.LogInfoVerbose.ShouldBeTrue();
+
+            // Ensure EventBusListener is ready and subscribed before joining
+            await EnsureEventBusListenerReadyAsync();
```
Aaronontheweb, line 193: Ensure the listener is running before we start any of the cluster operations.

Aaronontheweb (commented):
Detailed my changes

PR #7833, opened 2025-09-18
Fix race condition in KeepAliveConcatSpec test

Aaronontheweb:
## Summary
Fixed a race condition in the `KeepAliveConcat_should_emit_buffered_elements_when_upstream_completed` test that was causing intermittent failures with timeout errors.

## Problem
The test was failing with:
```
Failed: Timeout 00:00:03 while waiting for a message of type Akka.Streams.TestKit.TestPublisher+RequestMore
```

## Root Cause
The `TestPublisher.SendNext` method enforces reactive streams backpressure - it waits for demand before sending elements. The test was calling `SendNext` before the `KeepAliveConcat` stage's initial pull had propagated to the publisher, causing a timeout.

## Solution
- Added `await upstream.ExpectRequestAsync()` to wait for the initial pull from KeepAliveConcat
- Updated all test operations to use async/await consistently
- Ensured the test respects reactive streams protocol (no sends without demand)

## Notes
- This test was added in the .NET port and doesn't exist in the Scala Akka version
- All Scala tests follow the pattern of creating demand before sending elements
- The fix ensures proper sequencing without violating reactive streams semantics

## Test Results
All 6 tests in `KeepAliveConcatSpec` pass after the fix.

Fixes #[issue number if applicable]

### DIFF
--- a/src/core/Akka.Streams.Tests/Dsl/KeepAliveConcatSpec.cs
+++ b/src/core/Akka.Streams.Tests/Dsl/KeepAliveConcatSpec.cs
@@ -134,7 +134,7 @@ namespace Akka.Streams.Tests.Dsl
         }
 
         [Fact]
-        public void KeepAliveConcat_should_emit_buffered_elements_when_upstream_completed()
+        public async Task KeepAliveConcat_should_emit_buffered_elements_when_upstream_completed()
         {
             var upstream = this.CreatePublisherProbe<int>();
             var downstream = this.CreateSubscriberProbe<int>();
@@ -143,15 +143,19 @@ namespace Akka.Streams.Tests.Dsl
                 .Via(new KeepAliveConcat<int>(5, TimeSpan.FromSeconds(60), x => new[] { x }))
                 .RunWith(Sink.FromSubscriber(downstream), Sys.Materializer());
 
-            upstream.SendNext(1);
-            upstream.SendNext(2);            
-            upstream.SendComplete();
+            // Wait for KeepAliveConcat to pull from upstream (respecting reactive streams protocol)
+            await upstream.ExpectRequestAsync();
 
-            downstream.Request(2);
-            downstream.ExpectNextN(2).Should().BeEquivalentTo(new[] { 1, 2 }, o => o.WithStrictOrdering());
+            // Now send elements in response to the demand
+            await upstream.SendNextAsync(1);
+            await upstream.SendNextAsync(2);
+            await upstream.SendCompleteAsync();
 
-            downstream.Request(1);
-            downstream.ExpectComplete();
+            await downstream.RequestAsync(2);
+            await downstream.ExpectNextNAsync(new[] { 1, 2 });
+
+            await downstream.RequestAsync(1);
+            await downstream.ExpectCompleteAsync();
         }
     }
 


src/core/Akka.Streams.Tests/Dsl/KeepAliveConcatSpec.cs
```diff
@@ -143,15 +143,19 @@ public void KeepAliveConcat_should_emit_buffered_elements_when_upstream_complete
                 .Via(new KeepAliveConcat<int>(5, TimeSpan.FromSeconds(60), x => new[] { x }))
                 .RunWith(Sink.FromSubscriber(downstream), Sys.Materializer());
 
-            upstream.SendNext(1);
-            upstream.SendNext(2);            
-            upstream.SendComplete();
+            // Wait for KeepAliveConcat to pull from upstream (respecting reactive streams protocol)
+            await upstream.ExpectRequestAsync();
```
Aaronontheweb, line 147: This is the key fix - `SendNext` will not work if demand is not detected on the stream.

src/core/Akka.Streams.Tests/Dsl/KeepAliveConcatSpec.cs
Aaronontheweb, line 1: The rest of the changes here are just `async` / `await` test conversion.

Aaronontheweb (commented):
Detailed my changes

Aaronontheweb:
This is a bit of a work-around for what I suspect is a structural weakness in the way `SendNext` is designed inside the Akka.Streams.TestKit. We expect to be able to interact with the stream immediately after it's materialized, but if you call `SendNext` before downstream demand is detected, there's a chance the `Request` message the `TestProbe` is looking for inside `SendNext` could have already passed by.

Aaronontheweb:
> This is a bit of a work-around for what I suspect is a structural weakness in the way `SendNext` is designed inside the Akka.Streams.TestKit. We expect to be able to interact with the stream immediately after it's materialized, but if you call `SendNext` before downstream demand is detected, there's a chance the `Request` message the `TestProbe` is looking for inside `SendNext` could have already passed by.

Nope, this is wrong 

```csharp
internal static async Task SendNextTask(Probe<T> probe, T element, CancellationToken cancellationToken)
{
    await EnsureSubscriptionTask(probe, cancellationToken);
    var sub = probe.Subscription;
    if (probe.Pending == 0)
        probe.Pending = await sub.ExpectRequestAsync(cancellationToken);
    probe.Pending--;
    sub.SendNext(element);
}
```

We are already doing the correct thing internally. That also means this fix is wrong too - deck chair shuffling.

Aaronontheweb:
Actually, I think the issue is more subtle than that. The problem isn't with  itself, but with the timing of when the subscription object is established versus when KeepAliveConcat's  sends its initial Pull request.

Here's what I believe is happening:

1. Stream materializes and KeepAliveConcat's  immediately calls  
2. This Pull creates a  message that gets sent to the TestPublisher
3. TestPublisher receives the subscription and creates a  object
4. When  calls , it's looking for a  that has **this specific subscription object** attached

The race condition: If the Pull from step 1 arrives before the subscription is fully established, the  might have a null or different subscription reference. The  inside  is checking  - it's looking for a RequestMore with the correct subscription object.

By adding the explicit  before , we're ensuring we catch that initial Pull (even if it has a different subscription reference) and increment , so  doesn't need to wait for another request that will never come.

The real issue is that KeepAliveConcat calls Pull in PreStart before the upstream subscription is fully wired, causing the RequestMore to be sent with an incorrect/null subscription reference.

Aaronontheweb:
Actually, I think the issue is more subtle than that. The problem isn't with `SendNext` itself, but with the timing of when the subscription object is established versus when KeepAliveConcat's `PreStart()` sends its initial Pull request.

Here's what I believe is happening:

1. Stream materializes and KeepAliveConcat's `PreStart()` immediately calls `Pull(In)` 
2. This Pull creates a `RequestMore` message that gets sent to the TestPublisher
3. TestPublisher receives the subscription and creates a `PublisherProbeSubscription` object
4. When `SendNext` calls `sub.ExpectRequestAsync()`, it's looking for a `RequestMore` that has **this specific subscription object** attached

The race condition: If the Pull from step 1 arrives before the subscription is fully established, the `RequestMore` might have a null or different subscription reference. The `ExpectRequestAsync` inside `SendNext` is checking `isMessage: x => Equals(this, x.Subscription)` - it's looking for a RequestMore with the correct subscription object.

By adding the explicit `ExpectRequestAsync` before `SendNext`, we're ensuring we catch that initial Pull (even if it has a different subscription reference) and increment `Pending`, so `SendNext` doesn't need to wait for another request that will never come.

The real issue is that KeepAliveConcat calls Pull in PreStart before the upstream subscription is fully wired, causing the RequestMore to be sent with an incorrect/null subscription reference.

Aaronontheweb:
Ok, I think this fix is correct after all.

PR #7834, opened 2025-09-18
[Streams] Fix StreamRefSettings.FinalTerminationSignalDeadline never initialized

Arkatufus:
The `FinalTerminationSignalDeadline` property of `StreamRefSettings` never get initialized from HOCON settings, it is always TimeSpan.Zero.

## Changes

* Make sure `FinalTerminationSignalDeadline` is initialized.
* Clean up reference.conf, convert all tabs to spaces


Aaronontheweb:
@Arkatufus is this setting actually used anywhere currently?

Arkatufus:
Its being used here: https://github.com/akkadotnet/akka.net/blob/cc9435c388a0237e2059caa5de1cd5ccfdd023fc/src/core/Akka.Streams/Implementation/StreamRef/SourceRefImpl.cs#L250-L258

Its also used as a stream attribute, though the attribute isn't being used anywhere: https://github.com/akkadotnet/akka.net/blob/cc9435c388a0237e2059caa5de1cd5ccfdd023fc/src/core/Akka.Streams/Attributes.cs#L927-L931

Issue #7840, opened 2025-09-24
Akka.Persistence HealthCheck API


Aaronontheweb:
**Is your feature request related to a problem? Please describe.**

Since we shipped https://github.com/akkadotnet/Akka.Hosting/releases/tag/1.5.47-beta1, we've essentially moved Akka.NET's health check implementation from https://github.com/petabridge/akkadotnet-healthcheck to Akka.Hosting - a net positive change that greatly reduces the amount of configuration overhead and installed NuGet packages needed to power Akka.NET's health check system.

However, we are missing one beloved feature from Akka.HealthChecks - the Akka.Persistence checks, which were implemented using the infamous `SuicideProbe`:

https://github.com/petabridge/akkadotnet-healthcheck/blob/d482d5399be4b9ccd1d39f759cfb6be21384f8bf/src/Akka.HealthCheck.Persistence/AkkaPersistenceLivenessProbe.cs#L248-L295

We need to bring _some semblance_ of this functionality back into the picture in Akka.Hosting's health check implementation.

**Describe the solution you'd like**

The `SuicideProbe`, despite its amusing and delightful name, was a bit problematic:

1. Accidentally polluted journals / snapshot stores - took several rounds of bug-fixing to get right;
2. Resulted in writeable / billable units for customers running on cloud providers - health checks should run as close to zero cost as possible; 
3. Was a rather complex and somewhat fragile piece of infrastructure; and
4. Like most of the health checks in Akka.HealthCheck, it was too aggressive - a single persist / recover failure could trigger a liveness check failure. We tried adjusting this via how its parent, the `AkkaPersistenceLivenessProbe` handled retires, but that proved to be a bit unwieldy too.

So what I'm proposing is we add a new `virtual` method to the `AsyncWriteJournal` and `SnapshotStore` base classes:

```csharp
enum HealthCheckResult{
 Healthy = 0,
 Degraded = 1, // transient failures
 Unhealthy = 2 // irrecoverable failures
}

public virtual Task<HealthCheckResult> CheckHealthAsync(CancellationToken ct = default);
```

I think the default base class implementation could just use the `CircuitBreaker`'s Open/Closed status, since that would be a reliable method for determining whether or not the plugin was struggling to perform its work over a recent period of time - AND we could reset the healthcheck status from `Degraded` --> `Healthy` when the `CircuitBreaker` resets.

**Describe alternatives you've considered**

[...]

PR #7842, opened 2025-09-24
Akka.Persistence HealthChecks

Aaronontheweb:
## Changes

Implements #7840

## Checklist

* [x] Design discussion issue #7840

### Latest `dev` Benchmarks 

### This PR's Benchmarks

### DIFF
--- a/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.DotNet.verified.txt
+++ b/src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.DotNet.verified.txt
@@ -113,6 +113,18 @@ namespace Akka.Persistence
         public override int GetHashCode() { }
         public override string ToString() { }
     }
+    public sealed class CheckJournalHealth : Akka.Actor.INoSerializationVerificationNeeded, Akka.Persistence.IJournalMessage, Akka.Persistence.IJournalRequest, Akka.Persistence.IPersistenceMessage
+    {
+        public CheckJournalHealth(System.Threading.CancellationToken cancellationToken) { }
+        public System.Threading.CancellationToken CancellationToken { get; }
+        public override string ToString() { }
+    }
+    public sealed class CheckSnapshotStoreHealth : Akka.Actor.INoSerializationVerificationNeeded, Akka.Persistence.IPersistenceMessage, Akka.Persistence.ISnapshotMessage, Akka.Persistence.ISnapshotRequest
+    {
+        public CheckSnapshotStoreHealth(System.Threading.CancellationToken cancellationToken) { }
+        public System.Threading.CancellationToken CancellationToken { get; }
+        public override string ToString() { }
+    }
     public sealed class DeleteMessagesFailure : Akka.Actor.INoSerializationVerificationNeeded, System.IEquatable<Akka.Persistence.DeleteMessagesFailure>
     {
         public DeleteMessagesFailure(System.Exception cause, long toSequenceNr) { }
@@ -318,6 +330,12 @@ namespace Akka.Persistence
     {
         Akka.Persistence.IStashOverflowStrategy Create(Akka.Configuration.Config config);
     }
+    public sealed class JournalHealthCheckResponse : Akka.Actor.INoSerializationVerificationNeeded, Akka.Persistence.IJournalMessage, Akka.Persistence.IJournalResponse, Akka.Persistence.IPersistenceMessage
+    {
+        public JournalHealthCheckResponse(Akka.Persistence.PersistenceHealthCheckResult result) { }
+        public Akka.Persistence.PersistenceHealthCheckResult Result { get; }
+        public override string ToString() { }
+    }
     public sealed class LoadSnapshot : Akka.Actor.INoSerializationVerificationNeeded, Akka.Persistence.IPersistenceMessage, Akka.Persistence.ISnapshotMessage, Akka.Persistence.ISnapshotRequest, System.IEquatable<Akka.Persistence.LoadSnapshot>
     {
         public LoadSnapshot(string persistenceId, Akka.Persistence.SnapshotSelectionCriteria criteria, long toSequenceNr) { }
@@ -377,11 +395,36 @@ namespace Akka.Persistence
         public Akka.Persistence.IStashOverflowStrategy DefaultInternalStashOverflowStrategy { get; }
         public Akka.Persistence.PersistenceSettings Settings { get; }
         public Akka.Persistence.Journal.EventAdapters AdaptersFor(string journalPluginId) { }
+        public System.Threading.Tasks.Task<Akka.Persistence.PersistenceHealthCheckResult> CheckJournalHealthAsync(string journalPluginId, System.Threading.CancellationToken cancellationToken = null) { }
+        public System.Threading.Tasks.Task<Akka.Persistence.PersistenceHealthCheckResult> CheckSnapshotStoreHealthAsync(string snapshotStorePluginId, System.Threading.CancellationToken cancellationToken = null) { }
         [Akka.Annotations.InternalStableApiAttribute()]
         public Akka.Actor.IActorRef JournalFor(string journalPluginId) { }
         public string PersistenceId(Akka.Actor.IActorRef actor) { }
         [Akka.Annotations.InternalStableApiAttribute()]
         public Akka.Actor.IActorRef SnapshotStoreFor(string snapshotPluginId) { }
+    }
+    [System.Runtime.CompilerServices.IsReadOnlyAttribute()]
+    [System.Runtime.CompilerServices.NullableAttribute(0)]
+    public struct PersistenceHealthCheckResult : System.IEquatable<Akka.Persistence.PersistenceHealthCheckResult>
+    {
+        public PersistenceHealthCheckResult(Akka.Persistence.PersistenceHealthStatus Status, string Description = null, System.Exception Exception = null, [System.Runtime.CompilerServices.NullableAttribute(new byte[] {
+                2,
+                0,
+                0})] System.Collections.Generic.IReadOnlyDictionary<string, object> Data = null) { }
+        [System.Runtime.CompilerServices.NullableAttribute(new byte[] {
+                2,
+                0,
+                0})]
+        public System.Collections.Generic.IReadOnlyDictionary<string, object> Data { get; set; }
+        public string Description { get; set; }
+        public System.Exception Exception { get; set; }
+        public Akka.Persistence.PersistenceHealthStatus Status { get; set; }
+    }
+    public enum PersistenceHealthStatus
+    {
+        Healthy = 0,
+        Degraded = 1,
+        Unhealthy = 2,
     }
     public sealed class PersistenceSettings : Akka.Actor.Settings
     {
@@ -627,6 +670,12 @@ namespace Akka.Persistence
         public override int GetHashCode() { }
         public override string ToString() { }
     }
+    public sealed class SnapshotStoreHealthCheckResponse : Akka.Actor.INoSerializationVerificationNeeded, Akka.Persistence.IPersistenceMessage, Akka.Persistence.ISnapshotMessage, Akka.Persistence.ISnapshotResponse
+    {
+        public SnapshotStoreHealthCheckResponse(Akka.Persistence.PersistenceHealthCheckResult result) { }
+        public Akka.Persistence.PersistenceHealthCheckResult Result { get; }
+        public override string ToString() { }
+    }
     public sealed class StashingHandlerInvocation : Akka.Persistence.IPendingHandlerInvocation
     {
         public StashingHandlerInvocation(object evt, System.Action<object> handler) { }
@@ -857,6 +906,7 @@ namespace Akka.Persistence.Journal
     {
         protected readonly bool CanPublish;
         protected AsyncWriteJournal() { }
+        public virtual System.Threading.Tasks.Task<Akka.Persistence.PersistenceHealthCheckResult> CheckHealthAsync(System.Threading.CancellationToken cancellationToken = null) { }
         protected abstract System.Threading.Tasks.Task DeleteMessagesToAsync(string persistenceId, long toSequenceNr, System.Threading.CancellationToken cancellationToken);
         public abstract System.Threading.Tasks.Task<long> ReadHighestSequenceNrAsync(string persistenceId, long fromSequenceNr, System.Threading.CancellationToken cancellationToken);
         protected virtual bool Receive(object message) { }
@@ -1215,6 +1265,7 @@ namespace Akka.Persistence.Snapshot
     public abstract class SnapshotStore : Akka.Actor.ActorBase
     {
         protected SnapshotStore() { }
+        public virtual System.Threading.Tasks.Task<Akka.Persistence.PersistenceHealthCheckResult> CheckHealthAsync(System.Threading.CancellationToken cancellationToken = null) { }
         protected abstract System.Threading.Tasks.Task DeleteAsync(Akka.Persistence.SnapshotMetadata metadata, System.Threading.CancellationToken cancellationToken);
         protected abstract System.Threading.Tasks.Task DeleteAsync(string persistenceId, Akka.Persistence.SnapshotSelectionCriteria criteria, System.Threading.CancellationToken cancellationToken);
         protected abstract System.Threading.Tasks.Task<Akka.Persistence.SelectedSnapshot> LoadAsync(string persistenceId, Akka.Persistence.SnapshotSelectionCriteria criteria, System.Threading.CancellationToken cancellationToken);
--- a/src/core/Akka.Persistence.Tests/JournalHealthCheckSpec.cs
+++ b/src/core/Akka.Persistence.Tests/JournalHealthCheckSpec.cs
@@ -19,7 +19,7 @@ using Xunit.Abstractions;
 
 namespace Akka.Persistence.Tests;
 
-public class PersistenceHealthCheckSpec : PersistenceSpec
+public class JournalHealthCheckSpec : PersistenceSpec
 {
     private static Config HealthCheckConfig()
     {
@@ -49,7 +49,7 @@ public class PersistenceHealthCheckSpec : PersistenceSpec
             .WithFallback(Configuration("PersistenceHealthCheckSpec", extraConfig: extraConfig));
     }
 
-    public PersistenceHealthCheckSpec(ITestOutputHelper output) : base(HealthCheckConfig(), output)
+    public JournalHealthCheckSpec(ITestOutputHelper output) : base(HealthCheckConfig(), output)
     {
     }
 
--- a/src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
+++ b/src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
@@ -106,7 +106,8 @@ namespace Akka.Persistence.Journal
             if(_breaker.IsOpen)
                 return Task.FromResult(new PersistenceHealthCheckResult(PersistenceHealthStatus.Degraded, 
                     $"Circuit breaker is open, some operations may be failing intermittently with error: with error: {_breaker.LastCaughtException?.Message ?? "N/A"}"));
-            return Task.FromResult(_hasFatalError ? new PersistenceHealthCheckResult(PersistenceHealthStatus.Unhealthy, "Fatal error has occurred, some operations may be failing intermittently.") : new PersistenceHealthCheckResult(PersistenceHealthStatus.Healthy, "Healthy"));
+            return Task.FromResult(_hasFatalError ? new PersistenceHealthCheckResult(PersistenceHealthStatus.Unhealthy, "Fatal error has occurred. The ActorSystem must be restarted.") 
+                : new PersistenceHealthCheckResult(PersistenceHealthStatus.Healthy));
         }
 
         /// <inheritdoc/>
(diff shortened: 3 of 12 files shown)


src/core/Akka.Persistence.Tests/JournalHealthCheckSpec.cs
```diff
@@ -0,0 +1,32 @@
...
+public class PersistenceHealthCheckSpec : PersistenceSpec
+{
+    public PersistenceHealthCheckSpec(ITestOutputHelper output) : base(Configuration("PersistenceHealthCheckSpec"), output)
+    {
+    }
+
+    [Theory]
+    [InlineData(null)] // default plugin
+    [InlineData("akka.persistence.journal.inmem")]
+    public async Task JournalHealthCheck_should_default_to_Healthy(string? pluginId)
```
Aaronontheweb, line 59: Basic end to end health check for Akka.Persistence health checks on the journal. Will add some for the SnapshotStore once I get some validation on the design for this.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -25,6 +24,12 @@ public abstract class AsyncWriteJournal : WriteJournalBase, IAsyncRecovery
     {
         protected readonly bool CanPublish;
         private readonly CircuitBreaker _breaker;
+        
+        /// <summary>
+        /// Set to <c>true</c> when a fatal error has occurred, i.e., the Akka.Persistence configuration is illegal
+        /// </summary>
+        private bool _hasFatalError;
```
Aaronontheweb: If we have fatal errors, such as configuration errors, in the CTOR we should try to capture those.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -86,6 +93,23 @@ protected AsyncWriteJournal()
             _resequencer = Context.ActorOf(Props.Create(() => new Resequencer()), "resequencer");
         }
 
+        /// <summary>
+        /// Health check for the journal.
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation token for the health check invocation.</param>
+        /// <returns>A <see cref="PersistenceHealthCheckResult"/> with a health status and optional error message.</returns>
+        public virtual Task<PersistenceHealthCheckResult> CheckHealthAsync(CancellationToken cancellationToken = default)
```
Aaronontheweb, line 100: Overridable health check implementation - this base one looks at the state of the `CircuitBreaker` and uses that to determine the health of the overall plugin.

A question I had wondered about but believe I have answered - what if there's a large burst of writes that trigger an `OpenCircuitBreakerException`, which will cause this health check to show "degraded", but then the writes stop? Will this health check eventually heal?

The answer to that question is "yes" - due to the `CircuitBreaker`'s use of the `IScheduler`:

https://github.com/akkadotnet/akka.net/blob/5290189a4f01ca0814df1b21693dff6104c8a614/src/core/Akka/Pattern/CircuitBreakerState.cs#L120-L131

So the health check can be written simply like this - the circuit breaker already does the difficult work of re-testing to see if we're healthy again.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -205,6 +229,14 @@ protected bool ReceiveWriteJournal(object message)
                 case DeleteMessagesTo deleteMessagesTo:
                     HandleDeleteMessagesTo(deleteMessagesTo);
                     return true;
+                case CheckHealth checkHealth:
```
Aaronontheweb: Messaging handler for the health check inside the journal.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -385,11 +421,11 @@ private void ProcessResults(IImmutableList<Exception> results, int atomicWriteCo
                 : new WriteMessageRejected(x, exception, writeMessage.ActorInstanceId), results, resequencerCounter, writeMessage, resequencer, writeJournal);
         }
         
-        private void Resequence(Func<IPersistentRepresentation, Exception, object> mapper,
+        private static void Resequence(Func<IPersistentRepresentation, Exception, object> mapper,
             IImmutableList<Exception> results, long resequencerCounter, WriteMessages msg, IActorRef resequencer, IActorRef writeJournal)
         {
             var i = 0;
-            var enumerator = results?.GetEnumerator();
+            using var enumerator = results?.GetEnumerator();
```
Aaronontheweb, line 430: Fixed a memory leak here.

src/core/Akka.Persistence/Snapshot/SnapshotStore.cs
```diff
@@ -49,6 +51,23 @@ protected SnapshotStore()
             
             _log = Context.GetLogger();
         }
+        
+        /// <summary>
+        /// Health check for the journal.
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation token for the health check invocation.</param>
+        /// <returns>A <see cref="PersistenceHealthCheckResult"/> with a health status and optional error message.</returns>
+        public virtual Task<PersistenceHealthCheckResult> CheckHealthAsync(CancellationToken cancellationToken = default)
```
Aaronontheweb, line 64: Snapshot store implementation of the health check - currently incomplete. I still need to add

- [ ] Messaging protocol
- [ ] Handling of messaging protocol
- [ ] Method of accessing the health check from the `PersistenceExtension`
- [ ] Tests

src/core/Akka.Persistence/JournalProtocol.cs
```diff
@@ -180,6 +181,42 @@ public override int GetHashCode()
...
     }
 
+    /// <summary>
+    /// Invokes a health check on the journal plugin.
+    /// </summary>
+    public sealed class CheckHealth : IJournalRequest
+    {
+        public CheckHealth(CancellationToken cancellationToken)
+        {
+            CancellationToken = cancellationToken;
```
Aaronontheweb, line 191: Health check messaging protocol - these messages are all tagged with `INoSerializationVerificationNeeded` so it's safe to pass the `CancellationToken` around via them.

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -252,6 +253,23 @@ public IActorRef JournalFor(string journalPluginId)
             return PluginHolderFor(configPath, JournalFallbackConfigPath).Ref;
         }
 
+        /// <summary>
+        /// Shortcut for invoking journal health checks.
+        /// </summary>
+        /// <param name="journalPluginId">The HOCON id of the Akka.Persistence plugin./</param>
+        /// <param name="cancellationToken">An optional cancellation token.</param>
+        /// <returns>A <see cref="PersistenceHealthCheckResult"/> with health status and possibly a descriptive message.</returns>
+        public async Task<PersistenceHealthCheckResult> CheckJournalHealthAsync(string journalPluginId,
```
Aaronontheweb, line 262: Convenience method for invoking the health check - if the `journalPluginId` is not found this method will err.

Aaronontheweb (commented):
Detailed my changes, so far. Looking for feedback before I move on with implementing the `SnapshotStore` checks.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -86,6 +93,23 @@ protected AsyncWriteJournal()
...
+        /// <summary>
+        /// Health check for the journal.
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation token for the health check invocation.</param>
+        /// <returns>A <see cref="PersistenceHealthCheckResult"/> with a health status and optional error message.</returns>
+        public virtual Task<PersistenceHealthCheckResult> CheckHealthAsync(CancellationToken cancellationToken = default)
+        {
+            if(_breaker.IsHalfOpen)
+                return Task.FromResult(new PersistenceHealthCheckResult(PersistenceHealthStatus.Degraded, 
+                    $"Circuit breaker is half-open, some operations may be failing intermittently with error: {_breaker.LastCaughtException?.Message ?? "N/A"}"));
```
to11mtm: Two thoughts looking at this;

First, IMO It would be nice if we had some way to track 'last success' (And maybe even have auto-polling as a config option to track)

Second is a question, could we have a more 'structured' output instead of just a formatted string? It at least makes it easier to handle such events from a parsing standpoint (esp if we are able to add time or possibly other things, need to keep looking at this to know what's up)

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -371,7 +407,7 @@ private async Task ExecuteBatch(WriteMessages message, int atomicWriteCount, IAc
             }
         }
 
-        private void ProcessResults(IImmutableList<Exception> results, int atomicWriteCount, WriteMessages writeMessage, IActorRef resequencer,
+        private static void ProcessResults(IImmutableList<Exception> results, int atomicWriteCount, WriteMessages writeMessage, IActorRef resequencer,
```
to11mtm, line 412: Soooo this is safe and I get why it may have happened, I would however suggest double-checking this against Persistence SQL (or another plugin of choice that uses AsyncJournal with some extra logic behind it) just in case the jumps bt methodtables cause an issue.

(To be clear I'm probably overthinking this BUUUUUT It would still be good to know the difference just in case 😇🤷‍♂️)

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
to11mtm: Would it be worth it to have an enum to distinguish the types of known errors?

Thinking with a devops hat, a fatal error due to config vs a fatal error due to a condition that forces a journal shutdown/restart (Is that a thing or am I misremembering? 😇) would be useful to track... but maybe I'm forgetting things here (probably?).

As an additional thought... IDK if it's a requirement for all plugins buuuut a simple check like getting the max ordering ID would be nice.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -86,6 +93,23 @@ protected AsyncWriteJournal()
...
+        /// <returns>A <see cref="PersistenceHealthCheckResult"/> with a health status and optional error message.</returns>
+        public virtual Task<PersistenceHealthCheckResult> CheckHealthAsync(CancellationToken cancellationToken = default)
+        {
+            if(_breaker.IsHalfOpen)
+                return Task.FromResult(new PersistenceHealthCheckResult(PersistenceHealthStatus.Degraded, 
+                    $"Circuit breaker is half-open, some operations may be failing intermittently with error: {_breaker.LastCaughtException?.Message ?? "N/A"}"));
+            if(_breaker.IsOpen)
+                return Task.FromResult(new PersistenceHealthCheckResult(PersistenceHealthStatus.Degraded, 
+                    $"Circuit breaker is open, some operations may be failing intermittently with error: with error: {_breaker.LastCaughtException?.Message ?? "N/A"}"));
+            return Task.FromResult(_hasFatalError ? new PersistenceHealthCheckResult(PersistenceHealthStatus.Unhealthy, "Fatal error has occurred. The ActorSystem must be restarted.")
```
to11mtm: nitpick: Formatting here makes this a PITA to read, probably best to split the ternary into more lines.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -76,6 +82,7 @@ protected AsyncWriteJournal()
                     _replayFilterMode = ReplayFilterMode.Warn;
                     break;
                 default:
+                    _hasFatalError = true;
                     throw new ConfigurationException($"Invalid replay-filter.mode [{replayFilterMode}], supported values [off, repair-by-discard-old, fail, warn]");
```
to11mtm, line 81: Stupid question... if the ctor throws is there even a chance this variable will be read?

src/core/Akka.Persistence.Tests/JournalHealthCheckSpec.cs
to11mtm: needs some failure specs too plz, especially if plugins want to implement so that people and genAI can use as example for scaffolding.

to11mtm (commented):
Left some thoughts. Might poke at it more but these were the hard hitters.

src/core/Akka.Persistence.Tests/JournalHealthCheckSpec.cs
Aaronontheweb: I'll probably need to code a custom journal with a configurable `CircuitBreaker` or some-such to make that happen, but I will try.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
Aaronontheweb: > Second is a question, could we have a more 'structured' output instead of just a formatted string? It at least makes it easier to handle such events from a parsing standpoint (esp if we are able to add time or possibly other things, need to keep looking at this to know what's up)

As long as it can fit into a [`HealthCheckResult`](https://github.com/dotnet/aspnetcore/blob/6c99deea2314accdd79af07c937f8b00ca318263/src/HealthChecks/Abstractions/src/HealthCheckResult.cs#L24-L30):

```csharp
public HealthCheckResult(HealthStatus status, string? description = null, Exception? exception = null, IReadOnlyDictionary<string, object>? data = null)
    {
        Status = status;
        Description = description;
        Exception = exception;
        Data = data ?? _emptyReadOnlyDictionary;
    }
```

Then that should be fine.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
Aaronontheweb: This method was always private so it can't have side-effects in other plugins.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
Aaronontheweb: I could just capture the `Exception` itself in a nullable field - that's probably easiest. 

There's only a handful of "fatal" exceptions - we treat the rest of the exceptions as "eventually recoverable"

Aaronontheweb: > As an additional thought... IDK if it's a requirement for all plugins buuuut a simple check like getting the max ordering ID would be nice.

that might be a stretch just given some of the limitations across plugins

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
Aaronontheweb: That's how I originally had it (the way you suggested) until Rider nitpicked me. Never doubt your vibe I guess :shrug:

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
Aaronontheweb: That's a Jon Skeet question - I honestly have no idea.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
to11mtm: oh wow I'm dumb I meant to say the benchmarks.

Main side effect I'm thinking of is code locality.

But, again, probably overthinking it.

src/core/Akka.API.Tests/verify/CoreAPISpec.ApprovePersistence.DotNet.verified.txt
Aaronontheweb, line 1: These are all of the new public message and enum types we added to support health checks.

src/core/Akka.Persistence.Tests/JournalHealthCheckSpec.cs
```diff
@@ -0,0 +1,132 @@
...
+                                                       max-failures = 1
+                                                       call-timeout = 1s
+                                                       reset-timeout = 1s
+                                                   }
+                                               }
+                                               # Disable message serialization for circuit breaker tests to avoid serialization issues
+                                               akka.actor.serialize-messages = off
+
+                                   """;
+        return TestConfigs.TestSchedulerConfig
```
Aaronontheweb, line 48: We use the `TestScheduler` here to drive the `CircuitBreaker` resets.

src/core/Akka.Persistence.Tests/JournalHealthCheckSpec.cs
Aaronontheweb: Done

src/core/Akka.Persistence.Tests/JournalHealthCheckSpec.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        // Advance time to let the write fail and circuit breaker open
+        var testScheduler = (TestScheduler)Sys.Scheduler;
+        testScheduler.Advance(TimeSpan.FromSeconds(2));
+
+        using var cts = new CancellationTokenSource(RemainingOrDefault);
+        var pluginHealth = await Extension.CheckJournalHealthAsync("akka.persistence.journal.failing-open", cts.Token);
+
+        Assert.Equal(PersistenceHealthStatus.Degraded, pluginHealth.Status);
+        Assert.Contains("Circuit breaker is open", pluginHealth.Description);
```
Aaronontheweb, line 87: Probably not necessary to be this specific, but wanted to illustrate that leveraging the `CircuitBreaker` is a good default health heuristic.

src/core/Akka.Persistence.Tests/JournalHealthCheckSpec.cs
```diff
@@ -0,0 +1,132 @@
...
+        testScheduler.Advance(TimeSpan.FromSeconds(1));
+
+        // Give the async operations time to complete
+        await Task.Delay(100);
+
+        // Advance time past reset-timeout to transition to half-open
+        testScheduler.Advance(TimeSpan.FromSeconds(1));
+
+        // Give the transition time to complete
+        await Task.Delay(100);
```
Aaronontheweb, line 113: Even though we can use the `TestScheduler` to advance time, the circuit breaker still has to asynchronously perform its update.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
Aaronontheweb: As it turns out: https://www.youtube.com/watch?v=BxgaXOSS2t4

So I got rid of this.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
Aaronontheweb: Implemented this.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -84,6 +86,26 @@ protected AsyncWriteJournal()
...
+        /// <summary>
+        /// Health check for the journal.
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation token for the health check invocation.</param>
+        /// <returns>A <see cref="PersistenceHealthCheckResult"/> with a health status and optional error message.</returns>
+        public virtual Task<PersistenceHealthCheckResult> CheckHealthAsync(CancellationToken cancellationToken = default)
+        {
+            if(_breaker.IsHalfOpen)
+                return Task.FromResult(new PersistenceHealthCheckResult(PersistenceHealthStatus.Degraded, 
+                    $"Circuit breaker is half-open, some operations may be failing intermittently", _breaker.LastCaughtException, _defaultHealthCheckTags));
```
Aaronontheweb, line 104: We record the `PersistenceHealthStatus`, a description, the last exception caught by the `CircuitBreaker`, and our default tags - which just include the plugin id.

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -205,6 +227,18 @@ protected bool ReceiveWriteJournal(object message)
                 case DeleteMessagesTo deleteMessagesTo:
                     HandleDeleteMessagesTo(deleteMessagesTo);
                     return true;
+                case CheckJournalHealth checkHealth:
+                    var sender = Sender;
+                    CheckHealthAsync(checkHealth.CancellationToken)
```
Aaronontheweb, line 232: Actual handling of the health check invocation

src/core/Akka.Persistence/Journal/AsyncWriteJournal.cs
```diff
@@ -256,16 +290,6 @@ private void HandleReplayMessages(ReplayMessages message)
 
             async Task ExecuteHighestSequenceNr()
             {
-                void CompleteHighSeqNo(long highSeqNo)
```
Aaronontheweb, line 259: Moved this below the `return`

src/core/Akka.Persistence/Persistence.cs
```diff
@@ -252,6 +253,40 @@ public IActorRef JournalFor(string journalPluginId)
...
+            return r.Result;
+        }
+
+        /// <summary>
+        /// Shortcut for invoking snapshot store health checks.
+        /// </summary>
+        /// <param name="snapshotStorePluginId">The HOCON id of the Akka.Persistence plugin.</param>
+        /// <param name="cancellationToken">An optional cancellation token.</param>
+        /// <returns>A <see cref="PersistenceHealthCheckResult"/> with health status and possibly a descriptive message.</returns>
+        public async Task<PersistenceHealthCheckResult> CheckSnapshotStoreHealthAsync(string snapshotStorePluginId,
```
Aaronontheweb, line 279: Convenience method for checking the snapshot store plugin health.

Aaronontheweb (commented):
Detailed my changes

Aaronontheweb:
All of the failing unit tests are racy - not relation to these changes.

Arkatufus (approved):
LGTM

Aaronontheweb:
I'll need to port this to `dev`

PR #7839, opened 2025-09-24
Fix TLS handshake error handling

Arkatufus:
Fixes #7838

# Enforce ActorSystem shutdown on TLS handshake failures

## Summary
- Any TLS handshake failure now triggers CoordinatedShutdown with a clear reason (exit code 79), preventing a node from remaining "half‑healthy" via reverse connections.
- Tests cover server/client TLS failures and verify the misconfigured side shuts down.

## Changes
- Transport (DotNetty)
  - Handle `TlsHandshakeCompletionEvent` failures in `TcpTransport` and run `CoordinatedShutdown` using a new `TlsHandshakeFailureReason` (exit code 79). Logs include channel endpoints and id.

## New shutdown reason
- `Akka.Remote.Transport.DotNetty.TlsHandshakeFailureReason` (exit code 79): used by `TcpTransport` and `EndpointManager` to terminate the misconfigured system.

## Tests (Akka.Remote.Tests)
- `Tls_handshake_failure_should_be_logged_and_shutdown_server`
- `Server_side_tls_handshake_failure_should_shutdown_server`
- `Client_side_tls_handshake_failure_should_shutdown_client`

## Rationale
- Akka.NET remote is peer‑to‑peer: any TLS misconfiguration in either role must cause the association to fail hard. Shutting down the misconfigured ActorSystem avoids split‑brain‑like symptoms where nodes can still send outbound but fail inbound.

## Impact / Migration
- Behavior change: nodes with TLS handshake failures now terminate (exit code 79). Ensure certificates are valid and accessible.
- No public API changes; wire protocol unaffected.

PR #7824, opened 2025-09-15
Propagate error from DotNetty TLS handshake failure to Akka.Remote

Arkatufus:
Fixes #7823
Supercedes #7821

## Summary

- Surface TLS handshake failures from DotNetty in Akka.Remote.
- Log the actual SSL/private-key error at ERROR and fail outbound associations fast with a clear exception instead of silent disassociations.
- Adds unit test.

## Changes

- DotNetty transport
  - Handle `TlsHandshakeCompletionEvent` in `TcpHandlers.UserEventTriggered`:
    - Log failure at ERROR with the underlying exception.
    - Notify via `UnderlyingTransportError` when possible.
    - Close the channel.
  - Client association (`AssociateInternal`)
    - Wait for TLS handshake completion when SSL is enabled.
    - Throw `InvalidAssociationException` with the TLS failure as inner exception on error.
- Tests
  - Added `DotNettyTlsHandshakeFailureSpec` to assert that server logs “TLS handshake failed …” at ERROR when cert lacks a private key and a client connects.

## Behavioral impact

- On TLS failures:
  - Server logs ERROR with the underlying exception (e.g., private key access denied).
  - Client `Associate` now fails fast with `InvalidAssociationException` containing the TLS error.
- No public API changes. Normal operation unaffected.

PR #7847, opened 2025-10-02
Fix: Validate SSL certificate private key access at server startup

Aaronontheweb:
## Problem

Akka.Remote server starts successfully even when the application lacks permissions to access the SSL certificate's private key. The server appears healthy but fails when clients attempt to connect, causing:
- Hard-to-diagnose TLS handshake failures during runtime
- Silent failures that only appear when connections arrive
- Poor operational experience for administrators

### Customer Impact
A production customer reported this issue where:
- Node with certificate but no private key permissions can join an existing cluster
- Server appears to start successfully
- TLS errors only appear when other nodes try to connect
- Misleading error messages make diagnosis difficult

## Root Cause

Certificate loading in `DotNettyTransportSettings` only validates that the certificate EXISTS in the Windows certificate store, not whether the application can ACCESS the private key. 

```csharp
// Current code - vulnerable
private SslSettings(string certificateThumbprint, ...)
{
    var find = store.Certificates.Find(X509FindType.FindByThumbprint, ...);
    Certificate = find[0];  // ← NO PRIVATE KEY VALIDATION!
}
```

**Key insight**: Windows separates certificate storage from private key storage with individual ACL permissions. `Certificate.HasPrivateKey` may return `true` even when the application lacks ACL permissions to use the private key.

### 1. Add ValidateCertificate() method to SslSettings
- Checks `Certificate.HasPrivateKey`
- Actually tests private key access with `GetRSAPrivateKey()` (not just presence)
- Throws `ConfigurationException` with clear error message on failure

### 2. Call validation in Listen() before server bind
- Ensures fail-fast behavior at startup
- Prevents server from running in broken state
- Provides clear error messages for administrators

### 3. Comprehensive test coverage
- Server fails at startup with inaccessible private key ✅
- Server starts successfully with valid certificate ✅
- Server starts successfully without SSL ✅
- Updated existing tests to validate fail-fast (removed incorrect runtime failure tests)

### Files Modified
1. `Akka.Remote/Transport/DotNetty/DotNettyTransportSettings.cs`
   - Added `ValidateCertificate()` method to `SslSettings` class

2. `Akka.Remote/Transport/DotNetty/DotNettyTransport.cs`
   - Added validation call in `Listen()` before server socket binds

3. `Akka.Remote.Tests/Transport/DotNettyCertificateValidationSpec.cs`
   - New test file with comprehensive validation tests

4. `Akka.Remote.Tests/Transport/DotNettyTlsHandshakeFailureSpec.cs`
   - Updated to test fail-fast behavior
   - Removed tests that validated incorrect behavior

### Breaking Change (Expected)
Existing misconfigured deployments will now fail at startup instead of silently starting with broken TLS. This is **correct behavior** - fail-fast is better than silent failure.

### Migration Guide
If ActorSystem fails to start with:
```
ConfigurationException: SSL certificate private key exists but cannot be accessed.
Verify application user has permissions to the private key in certificate store.
```

**Fix**: Grant the application user read permissions to the certificate's private key:
```powershell
# Find the private key file
$cert = Get-ChildItem Cert:\LocalMachine\My\<thumbprint>
$keyPath = $cert.PrivateKey.CspKeyContainerInfo.UniqueKeyContainerName
$keyFile = Get-ChildItem "$env:ProgramData\Microsoft\Crypto\RSA\MachineKeys\$keyPath"

# Grant permissions (replace DOMAIN\AppUser with your app identity)
icacls $keyFile.FullName /grant "DOMAIN\AppUser:R"
```

## Test Results
```
DotNettyCertificateValidationSpec:
✅ Server_should_fail_at_startup_with_certificate_without_private_key
✅ Server_should_start_successfully_with_valid_certificate
✅ Server_should_start_successfully_without_ssl

DotNettyTlsHandshakeFailureSpec:
✅ Server_should_fail_at_startup_with_certificate_without_private_key (updated)
✅ Client_side_tls_handshake_failure_should_shutdown_client
```

## Related

- Related to PR #7839 (added shutdown on TLS failure - incomplete fix)
- Related to PR #7824 (v1.5.50 - added TLS error detection)
- Related to PR #7821 (first attempt - insufficient)

## Checklist

- [x] Code changes implemented
- [x] Tests added/updated
- [x] All tests passing
- [x] Clear error messages for administrators
- [x] Documentation in commit messages
- [x] Backward compatible (fail-fast only affects misconfigured systems)

## Next Steps

Optional enhancement (separate PR):
- Mutual TLS enforcement to require client certificates with private keys
- Defense-in-depth security configuration

### DIFF
--- a/src/core/Akka.Remote.Tests/Transport/DotNettyTlsHandshakeFailureSpec.cs
+++ b/src/core/Akka.Remote.Tests/Transport/DotNettyTlsHandshakeFailureSpec.cs
@@ -68,122 +68,32 @@ namespace Akka.Remote.Tests.Transport
 
 
         [Fact]
-        public async Task Tls_handshake_failure_should_be_logged_and_shutdown_server()
+        public async Task Server_should_fail_at_startup_with_certificate_without_private_key()
         {
             CreateCertificateWithoutPrivateKey();
 
-            ActorSystem server = null;
-            ActorSystem client = null;
-
             try
             {
-                // Start TLS server with a cert that has no private key
+                // Server with cert that has no private key should FAIL TO START
                 var serverConfig = CreateConfig(true, NoKeyCertPath, null, suppressValidation: true);
 
-                server = ActorSystem.Create("ServerSystem", serverConfig);
-                InitializeLogger(server, "[SERVER] ");
-
-                // Server started - add an echo actor and subscribe to errors
-                server.ActorOf(Props.Create(() => new EchoActor()), "echo");
-
-                var errorProbe = CreateTestProbe(server);
-                server.EventStream.Subscribe(errorProbe.Ref, typeof(Event.Error));
-
-                // Start client with valid TLS cert
-                var clientConfig = CreateConfig(true, ValidCertPath, Password, suppressValidation: true);
-                client = ActorSystem.Create("ClientSystem", clientConfig);
-                InitializeLogger(client, "[CLIENT] ");
-
-                var serverAddress = RARP.For(server).Provider.DefaultAddress;
-                var echoPath = new RootActorPath(serverAddress) / "user" / "echo";
-                var echoSel = client.ActorSelection(echoPath);
-
-                // Trigger association attempt
-                var probe = CreateTestProbe(client);
-                echoSel.Tell("ping", probe.Ref);
-
-                // Expect server to log TLS handshake failure promptly
-                var err = errorProbe.ExpectMsg<Event.Error>(TimeSpan.FromSeconds(10));
-                var msg = err.ToString();
-                Assert.Contains("TLS handshake failed", msg, StringComparison.OrdinalIgnoreCase);
-
-                // Server should shutdown due to TLS failure
-                await AwaitAssertAsync(async () =>
+                // ActorSystem.Create should throw during startup due to certificate validation
+                var aggregateEx = Assert.Throws<AggregateException>(() =>
                 {
-                    Assert.True(server.WhenTerminated.IsCompleted);
-                    await Task.CompletedTask;
-                }, TimeSpan.FromSeconds(10), TimeSpan.FromMilliseconds(100));
+                    using var server = ActorSystem.Create("ServerSystem", serverConfig);
+                });
+
+                // Unwrap to find the ConfigurationException
+                var innerEx = aggregateEx.InnerException ?? aggregateEx;
+                while (innerEx is AggregateException agg && agg.InnerException != null)
+                    innerEx = agg.InnerException;
+
+                // Should be ConfigurationException about private key
+                Assert.IsType<ConfigurationException>(innerEx);
+                Assert.Contains("private key", innerEx.Message, StringComparison.OrdinalIgnoreCase);
             }
             finally
             {
-                if (client != null) 
-                    Shutdown(client, TimeSpan.FromSeconds(10));
-                if (server != null) 
-                    Shutdown(server, TimeSpan.FromSeconds(10));
-                try
-                {
-                    if (File.Exists(NoKeyCertPath)) 
-                        File.Delete(NoKeyCertPath);
-                } catch { /* ignore */ }
-            }
-            await Task.CompletedTask;
-        }
-
-        [Fact]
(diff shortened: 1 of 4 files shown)


src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
```diff
@@ -180,6 +180,13 @@ protected async Task<IChannel> NewServer(EndPoint listenAddress)
 
         public override async Task<(Address, TaskCompletionSource<IAssociationEventListener>)> Listen()
         {
+            // Validate SSL certificate before starting server
+            // This ensures fail-fast behavior if private key is inaccessible
+            if (Settings.EnableSsl)
+            {
+                Settings.Ssl.ValidateCertificate();
```
Aaronontheweb, line 187: Validate the private key during server binding, which should trigger shutdown of the `ActorSystem` if it fails

src/core/Akka.Remote.Tests/Transport/DotNettyTlsHandshakeFailureSpec.cs
Aaronontheweb, line 1: This spec is either somewhat redundant or unnecessary given the fail fast implementation - so it's been simplified.

Aaronontheweb (commented):
Detailed my changes

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransportSettings.cs
```diff
@@ -342,6 +342,48 @@ public SslSettings(X509Certificate2 certificate, bool suppressValidation)
...
+            {
+                throw new ConfigurationException(
+                    "SSL certificate does not have a private key. " +
+                    "Ensure certificate is installed with private key permissions.");
+            }
+
+            // Actually test private key access (not just presence)
+            try
+            {
+                using (var rsa = Certificate.GetRSAPrivateKey())
```
Arkatufus: We also need to check to see if `.GetECDsaPrivateKey()` returns null, need to monitor these in the future to make sure that these two private key types are the only ones that are supported by SSL.

Arkatufus (changes requested):
I think this needs to be tightened.

Aaronontheweb:
@Arkatufus Good catch on the ECDSA keys! I've updated the validation to check both RSA and ECDSA private keys.

## Changes Made

### 1. Added ECDSA Key Validation
Now checking both key types that SslStream supports:
```csharp
using (var rsaKey = Certificate.GetRSAPrivateKey())
using (var ecdsaKey = Certificate.GetECDsaPrivateKey())
{
    if (rsaKey == null && ecdsaKey == null)
        throw new ConfigurationException(...);
}
```

**Behavior**:
- **RSA cert**: `GetRSAPrivateKey()` succeeds, `GetECDsaPrivateKey()` returns null ✅
- **ECDSA cert**: `GetECDsaPrivateKey()` succeeds, `GetRSAPrivateKey()` returns null ✅
- **Neither accessible**: Both return null → validation fails with clear error ✅
- **Permission error**: `CryptographicException` caught → clear error message ✅

### 2. Improved Disposal Pattern
Changed from manual `Dispose()` calls to `using` statements to prevent resource leaks if an exception is thrown.

## Relationship to TLS Handler
The TLS handler uses `TlsHandler.Server(Settings.Ssl.Certificate)` (line 378 of DotNettyTransport.cs), which internally uses `SslStream`. SslStream will extract either RSA or ECDSA private keys depending on the certificate type. Our validation now matches this behavior exactly.

## Note on Future Key Types
Per .NET documentation, `GetRSAPrivateKey()` and `GetECDsaPrivateKey()` return `null` (not throw) when the key type doesn't match. If a certificate uses a different key type (like deprecated DSA), both methods return null and our validation correctly fails. If .NET adds new key types in the future, we'll need to update this validation.

Tests still passing ✅

Arkatufus (approved):
I guess we can call this "best effort" for a fail fast detection.

Aaronontheweb:
This is going into the `dev` branch but we'll need to back port it to v1.5

PR #7859, opened 2025-10-04
Fix race condition in UnfoldResourceAsyncSource callback invocation

Aaronontheweb:
## Summary
Fixed a race condition in `UnfoldResourceSourceAsync` where resource cleanup was skipped when streams were cancelled quickly (e.g., with `Sink.Cancelled`).

## Changes
- Changed `UnfoldResourceSourceAsync` to use `InvokeWithFeedback` instead of direct callback invocation
- Used `GetTypedAsyncCallback` to obtain `IAsyncCallback<T>` for proper feedback handling
- Implemented async local function pattern to properly catch `StreamDetachedException`
- Ensures resource cleanup occurs even when callback cannot be processed due to stream completion

## Root Cause
The previous implementation used `Action<T>` invocation which silently drops callbacks after stage shutdown. This caused the close callback to never execute when the stream stopped before the resource creation callback could be invoked.

## Technical Details
The fix matches the Scala Akka implementation's use of `invokeWithFeedback`, which returns a Future/Task that fails with `StreamDetachedException` when the stage has stopped. This allows proper cleanup in the exception handler.

## Test Results
- ✅ All 20 tests in `UnfoldResourceAsyncSourceSpec` pass
- ✅ Specifically fixes the flaky test: `A_UnfoldResourceAsyncSource_must_close_resource_when_stream_is_quickly_cancelled_reproducer_2`
- ✅ Verified stable across 10 consecutive runs

## Files Changed
- `src/core/Akka.Streams/Implementation/Sources.cs`

### DIFF
--- a/src/core/Akka.Streams/Implementation/Sources.cs
+++ b/src/core/Akka.Streams/Implementation/Sources.cs
@@ -564,6 +564,7 @@ namespace Akka.Streams.Implementation
         {
             private readonly UnfoldResourceSourceAsync<TOut, TSource> _stage;
             private readonly Lazy<Decider> _decider;
+            private readonly Lazy<IAsyncCallback<Try<TSource>>> _createdCallback;
             private Option<TSource> _state = Option<TSource>.None;
 
             public Logic(UnfoldResourceSourceAsync<TOut, TSource> stage, Attributes inheritedAttributes)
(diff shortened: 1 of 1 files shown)


src/core/Akka.Streams/Implementation/Sources.cs
```diff
@@ -576,18 +577,21 @@ public Logic(UnfoldResourceSourceAsync<TOut, TSource> stage, Attributes inherite
                     return strategy != null ? strategy.Decider : Deciders.StoppingDecider;
                 });
 
+                _createdCallback = new Lazy<IAsyncCallback<Try<TSource>>>(() =>
```
Aaronontheweb, line 580: This is another stage that really needs the `InvokeWithFeedback` changes we introduced in #7578

src/core/Akka.Streams/Implementation/Sources.cs
```diff
@@ -697,24 +701,31 @@ private void CreateResource()
...
+                            }
+                            else
+                            {
+                                // failed to open but stream is stopped already
+                                throw resource.Failure.Value;
+                            }
                         }
                     }
+
+                    _ = InvokeCallback();
```
Aaronontheweb, line 728: Invokes the callback and ensures that the stage gets closed afterwards in the event of failure, so that way we can't miss / fail resource clean-up.

Aaronontheweb (commented):
Detailed my changes

PR #7860, opened 2025-10-04
Fix race condition in DistributedPubSubDeadLetterSpec

Aaronontheweb:
## Summary
- Fixed intermittent test failure in `DistributedPubSubDeadLetterSpec.DistributedPubSubMediator_should_send_specialized_dead_letter_message_when_no_subscribers`
- Added warmup check to ensure `DeadLetterListener` is active before testing

## Root Cause
The test was racing against the initialization of the `DeadLetterListener` actor. The listener may not have fully started and subscribed to the `EventStream` before the test attempted to verify dead letter logging, causing the test to timeout waiting for a log message that would never arrive.

## Changes
Added an `AwaitAssertAsync` block that sends a warmup dead letter message and waits (with retries) for it to be logged. This ensures the `DeadLetterListener` is fully initialized and processing messages before running the actual test assertion.

## Test plan
- [x] Ran the test 20 consecutive times - all passed
- [x] Verified the fix eliminates the race condition

### DIFF
--- a/src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubDeadLetterSpec.cs
+++ b/src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubDeadLetterSpec.cs
@@ -5,6 +5,7 @@
 // </copyright>
 //-----------------------------------------------------------------------
 
+using System;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Cluster.Tools.PublishSubscribe;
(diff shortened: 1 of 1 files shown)


src/contrib/cluster/Akka.Cluster.Tools.Tests/PublishSubscribe/DistributedPubSubDeadLetterSpec.cs
```diff
@@ -37,6 +37,16 @@ public async Task DistributedPubSubMediator_should_send_specialized_dead_letter_
             var mediator = DistributedPubSub.Get(Sys).Mediator;
             var testMessage = "test-message";
 
+            // Ensure DeadLetterListener is active by retrying until it logs a warmup message
+            await AwaitAssertAsync(async () =>
```
Aaronontheweb, line 42: Comment explains it all, but the issue here was the test executing so quickly that the `DeadLetterListener` couldn't complete its `DeadLetter` subscription to the `EventStream`

Aaronontheweb: So now we `AwaitAssert` that the subscription is intact before proceeding with the real test

PR #7866, opened 2025-10-04
Fix race condition in MemorySnapshotStore

Aaronontheweb:
## Summary
Fixes a race condition in `MemorySnapshotStore` where concurrent access to a non-thread-safe `List<SnapshotEntry>` caused `InvalidOperationException` during enumeration.

## Changes
- Added lock synchronization (`_snapshotsLock`) around all `Snapshots` collection access
- Protected `DeleteAsync`, `LoadAsync`, and `SaveAsync` methods with locks
- Added comprehensive documentation explaining why locks were chosen over thread-safe collections

## Why Locks Instead of Thread-Safe Collections?
1. Each persistence ID can have multiple snapshots at different sequence numbers, requiring range queries
2. `LoadAsync` needs to enumerate and sort to find the highest sequenceNr matching criteria  
3. `SaveAsync` requires atomic check-then-update-or-add operations (FirstOrDefault + mutation/Add)
4. `ConcurrentDictionary` keyed by persistenceId would still require a non-thread-safe List/Bag per value

The lock ensures atomicity of compound operations and consistent enumeration during LINQ queries.

## Root Cause
The `MemorySnapshotStore` uses `TaskContinuationOptions.ExecuteSynchronously` (inherited from `SnapshotStore` base class), which means async operations run on ThreadPool threads, not the actor's mailbox thread. This caused concurrent modifications to the `List<SnapshotEntry>` collection.

## Testing
- ✅ Flaky test `MemorySnapshotStore_is_threadsafe` now passes consistently (ran 5 times, all passed)
- ✅ All Akka.Persistence.TCK.Tests pass (46 tests, ran 3 iterations)
- ✅ Verified fix eliminates race condition under high concurrency

Fixes flaky test: `Akka.Persistence.TCK.Tests.MemorySnapshotStoreSpec.MemorySnapshotStore_is_threadsafe`

### DIFF
--- a/src/core/Akka.Persistence/Snapshot/MemorySnapshotStore.cs
+++ b/src/core/Akka.Persistence/Snapshot/MemorySnapshotStore.cs
@@ -23,6 +23,19 @@ namespace Akka.Persistence.Snapshot
     public class MemorySnapshotStore : SnapshotStore
     {
         /// <summary>
+        /// Lock for thread-safe access to the Snapshots collection.
+        ///
+        /// Note: We use locks instead of thread-safe collections (e.g., ConcurrentDictionary) because:
+        /// 1. Each persistence ID can have multiple snapshots at different sequence numbers, requiring range queries
+        /// 2. LoadAsync needs to find the highest sequenceNr matching criteria via enumeration and sorting
+        /// 3. SaveAsync requires atomic check-then-update-or-add operations (FirstOrDefault + mutation/Add)
+        /// 4. ConcurrentDictionary keyed by persistenceId would still require a non-thread-safe List/Bag per value
+        ///
+        /// The lock ensures atomicity of compound operations and consistent enumeration during LINQ queries.
+        /// </summary>
+        private readonly object _snapshotsLock = new();
+
+        /// <summary>
         /// This is available to expose/override the snapshots in derived snapshot stores
         /// </summary>
         protected virtual List<SnapshotEntry> Snapshots { get; } = new();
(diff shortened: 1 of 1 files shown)


src/core/Akka.Persistence/Snapshot/MemorySnapshotStore.cs
```diff
@@ -22,6 +22,19 @@ namespace Akka.Persistence.Snapshot
...
+        ///
+        /// Note: We use locks instead of thread-safe collections (e.g., ConcurrentDictionary) because:
+        /// 1. Each persistence ID can have multiple snapshots at different sequence numbers, requiring range queries
+        /// 2. LoadAsync needs to find the highest sequenceNr matching criteria via enumeration and sorting
+        /// 3. SaveAsync requires atomic check-then-update-or-add operations (FirstOrDefault + mutation/Add)
+        /// 4. ConcurrentDictionary keyed by persistenceId would still require a non-thread-safe List/Bag per value
+        ///
+        /// The lock ensures atomicity of compound operations and consistent enumeration during LINQ queries.
+        /// </summary>
+        private readonly object _snapshotsLock = new();
```
Aaronontheweb, line 36: I had previously reviewed and blocked a PR that @Arkatufus submitted that used `lock`s to fix a similar issue inside the `MemoryJournal`. After spending quite a bit of time investigating this issue AND some of the ones in the `MemoryJournal`, I think he might have been right after all.

PR #7875, opened 2025-10-05
Fix race condition in QueueSource offer handling

Aaronontheweb:
Fixed a race condition where `OfferAsync()` calls could hang indefinitely when called during stream shutdown. The issue occurred when offers were accepted between `CompleteStage()`/`FailStage()` and `PostStop()`.

## Root Cause

The existing `_terminating` flag was only used to defer completion until buffer draining, but wasn't checked when processing new offers. This allowed offers to be accepted during shutdown but never completed, causing the `TaskCompletionSource` to hang indefinitely.

## Solution

- Check `_terminating` at callback entry and immediately drop offers with `QueueOfferResult.Dropped`
- Set `_terminating=true` in `OnDownstreamFinish` before `CompleteStage()`
- Set `_terminating=true` in `Failure` handler before `FailStage()`  
- Set `_terminating=true` in buffer overflow `Fail` cases before `FailStage()`

This aligns the .NET implementation with the upstream pattern where terminating offers are rejected immediately.

## Verification

- Test passed 30 consecutive runs
- All 26 QueueSourceSpec tests passed

Fixes flaky test: `QueueSourceSpec.QueueSource_should_buffer_when_needed`


Aaronontheweb:
## Relationship to Previous Fix

This fix is closely related to #7865 (QueueSink race condition fix from Oct 4, 2025), which used a `_closed` flag to prevent the same race condition pattern in `QueueSink`.

### Why #7865 Wasn't a Complete Fix

The QueueSink fix only addressed **one side** of the queue pattern - it fixed the consumer side (`QueueSink.PullAsync()`), but didn't address the producer side (`QueueSource.OfferAsync()`).

Both components share the same underlying race condition pattern:
1. User calls operation (`PullAsync()` or `OfferAsync()`)
2. Creates a `TaskCompletionSource` and queues an async callback
3. Stream shutdown begins - calls `CompleteStage()` or `FailStage()`
4. **Race window**: Callback is still active until `PostStop()` executes
5. If operation arrives during this window, it's accepted but never completed
6. Task hangs indefinitely

### Why This Fix Uses a Different Approach

While #7865 added a new `_closed` flag, this fix reuses the **existing `_terminating` flag** that was already present in QueueSource but wasn't being checked properly.

The key insight: `_terminating` was only used to defer completion until buffer draining, but it should have been rejecting **all new offers** immediately when set. By checking `_terminating` at the callback entry point and setting it in all shutdown paths, we achieve the same protection without adding another flag.

This aligns the .NET implementation with the upstream pattern where terminating offers are rejected immediately with `Dropped` status.

PR #7879, opened 2025-10-05
Fix race condition in GroupedWeightedWithin timer-based emission

Aaronontheweb:
## Summary
Fixes a race condition in the `GroupedWeightedWithin` operator where groups could be emitted prematurely when downstream pulled before the timer interval elapsed.

## Problem
The test `FlowGroupedWithinSpec.A_GroupedWithin_must_reset_time_window_when_max_elements_reached` was flaky, sometimes receiving element [4] after ~900ms-2s instead of the expected ~2s after the timer reset.

### Root Cause
When the stream starts, a timer is scheduled. If elements arrive after a delay and trigger max-count emission, the timer is rescheduled. However, the original timer could fire before the new interval elapses, setting `_pushEagerly = true`. When downstream then pulls, `OnPull()` would emit the group immediately, bypassing the timer interval.

Timeline:
1. T=0s: Stream starts, timer scheduled for T=2s
2. T=1s: Elements 1,2,3 arrive, max reached, timer rescheduled for T=3s, element 4 buffered
3. T=2s: Original timer fires (only 1s after reschedule), sets `_pushEagerly = true`
4. When downstream pulls, `OnPull()` emits element 4 prematurely

## Solution
Added `_groupClosed` flag that tracks whether the timer has fired for the current group:
- Set to `false` in `StartNewGroup()` when a new group begins
- Set to `true` in `OnTimer()` when the timer fires
- Checked in `OnPull()` to only emit if `_groupClosed` is true

This ensures groups only emit after the full timer interval elapses, regardless of when downstream pulls.

## Testing
- Fixed test now passes 50/50 consecutive runs
- Verified the fix addresses the race condition without breaking existing functionality

### DIFF
--- a/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
+++ b/src/core/Akka.Streams/Implementation/Fusing/Ops.cs
@@ -3136,6 +3136,7 @@ namespace Akka.Streams.Implementation.Fusing
             private long _totalWeight = 0L;
             private int _totalNumber = 0;
             private bool _hasElements;
+            private bool _groupClosed;
 
             public Logic(GroupedWeightedWithin<T> stage)
                 : base(stage.Shape)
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
This one is not ready to go yet

PR #7885, opened 2025-10-06
Fix TickSourceSpec timing assertions for scheduler catch-up behavior

Aaronontheweb:
## Summary
Relaxes timing assertions in TickSourceSpec to tolerate the HashedWheelTimer scheduler's catch-up behavior on .NET 6+.

## Changes
- Modified `A_Flow_based_on_a_tick_publisher_must_be_possible_to_cancel` test to use `ExpectNextNAsync()` instead of precise timing checks
- Test now verifies 2 ticks are received within 3 seconds, rather than enforcing exact 1-second intervals

## Problem
The test was failing intermittently on slower CI machines with:
```
Expected no messages during 00:00:00.2000000, instead we received <TestSubscriber.OnNext(tick)> 
after 00:00:00.0791451
```

The HashedWheelTimer scheduler on .NET 6+ (introduced in c5b132065) includes catch-up logic that fires multiple timer callbacks rapidly when it detects it has fallen behind schedule. On slower/loaded CI machines:
1. Scheduler detects clock drift (e.g., thread was starved)
2. Enters catch-up loop to process missed ticks
3. Fires timer callbacks in rapid succession (79ms apart instead of 1000ms)
4. TickSource receives multiple OnTimer calls quickly
5. Since there's already demand (Request(2)), both ticks get pushed immediately

## Root Cause
The test's strict timing assertions (`ExpectNoMsgAsync(200ms)` between ticks) were incompatible with the scheduler's catch-up behavior, which is a legitimate performance optimization for handling thread starvation.

## Solution
Relaxed the test to focus on what it actually needs to verify (cancellation works correctly) rather than acting as a scheduler timing precision test:

**Before:**
```csharp
sub.Request(2);
await c.ExpectNoMsgAsync(TimeSpan.FromMilliseconds(600));
await c.ExpectNextAsync("tick");
await c.ExpectNoMsgAsync(TimeSpan.FromMilliseconds(200)); // ← Failed here on slow CI
await c.ExpectNextAsync("tick");
```

**After:**
```csharp
sub.Request(2);
var ticks = await c.ExpectNextNAsync(2, TimeSpan.FromSeconds(3)).ToListAsync();
ticks.Should().HaveCount(2);
ticks.Should().OnlyContain(t => t == "tick");
```

This approach:
- ✅ Still verifies 2 ticks are received
- ✅ Tolerates scheduler catch-up behavior
- ✅ Focuses on the test's purpose (cancellation, not timing)
- ✅ Works on both fast and slow machines

## Test Results
✅ Passed 20/20 consecutive local runs
✅ Should now pass consistently on CI

## Related
- Previous fixes: #7868, commit eaf304c0e, commit 0c247d7ed
- HashedWheelTimer catch-up logic: commit c5b132065 (.NET 6+ PeriodicTimer support)


Aaronontheweb:
Disabling auto-merge - it looks like this issue isn't fully resolved yet (most recent commit failed one of the very tests that should be passing now)

PR #7897, opened 2025-10-09
Fix TLS hostname validation bug and add configurable validation

### LINKED ISSUE #7893: TLS client validates against wrong hostname, breaks per-node certificates
## Problem

TLS client-side validation uses the **wrong hostname**. In `DotNettyTransport.cs` line 355:

```csharp
var host = certificate.GetNameInfo(X509NameType.DnsName, false);  // BUG: Uses CLIENT's cert
tlsHandler = TlsHandler.Client(host, certificate);
```

This extracts the DNS name from the **client's own certificate** instead of using the **remote server's address**. Result: `RemoteCertificateNameMismatch` errors when using distinct per-node certificates.

## Example

- Client A has cert with `CN=node-a.example.com`
- Client A connects to Server B at `node-b.example.com`
- Client A expects server cert to match `node-a.example.com` ❌
- Validation fails even though both certs are valid

## Fix

```csharp
var host = remoteAddress.Host;  // Use the address we're connecting TO
```

## Additional: Configuration Option Needed

Even with the fix, hostname validation breaks legitimate mutual TLS scenarios:
- IP-based connections (`10.0.0.1` vs cert has `DNS:node1.example.com`)
- Service discovery (dynamic addresses)
- Per-node certificates in P2P clusters

Add configuration to disable hostname validation for mutual TLS:

```hocon
akka.remote.dot-netty.tcp.ssl {
  validate-certificate-name = false  # Default for backward compat
}
```

When `false`: Validates cert chain against CA, ignores `RemoteCertificateNameMismatch`  
When `true`: Traditional TLS hostname validation

## Testing

Extend `DotNettyMutualTlsSpec.cs`:
- Different certs + validation disabled → Pass
- Different certs + validation enabled → Fail (name mismatch)
- Same cert + validation enabled → Pass
- Config unspecified → Pass (default disabled)

## Acceptance Criteria

- [ ] Fix line 355 to use `remoteAddress.Host`
- [ ] Add `validate-certificate-name` config (default: false)
- [ ] Custom validation callback when disabled
- [ ] Test coverage for both modes
- [ ] Update docs

Aaronontheweb:
## Summary

This PR fixes a critical TLS hostname validation bug (issue #7893) and introduces a configurable, type-safe validation system for DotNetty SSL/TLS connections.

### 🐛 Bug Fix (DotNettyTransport.cs:355-356)
**Problem**: TLS client was validating against the client's own certificate DNS name instead of the remote server address.

**Fix**: Changed from `certificate.GetNameInfo(X509NameType.DnsName, false)` to `remoteAddress.Host`

This bug caused incorrect hostname validation in TLS connections, particularly affecting mutual TLS scenarios.

### ⚙️ New Configuration Option
Added `validate-certificate-hostname` setting to `Remote.conf`:
- **Default**: `false` (disabled for backward compatibility)
- **When enabled**: Traditional TLS hostname validation (certificate CN/SAN must match target hostname)
- **When disabled**: Only validates certificate chain against CA, ignores hostname mismatches
- **Use cases**: Mutual TLS with per-node certificates, IP-based connections, dynamic service discovery

### 🔒 Type-Safe Validation System
Introduced enums to prevent primitive confusion in security-critical code:
- `ChainValidationMode` enum (`ValidateChain`, `IgnoreChainErrors`)
- `HostnameValidationMode` enum (`ValidateHostname`, `IgnoreHostnameMismatch`)
- `TlsValidationCallbacks` static factory class with convenience methods:
  - `Create(chainValidation, hostnameValidation, log)` - Main factory
  - `ValidateFull()` - Full validation (chain + hostname)
  - `ValidateChainOnly()` - Chain validation only
  - `ValidateHostnameOnly()` - Hostname validation only (for testing)
  - `AcceptAll()` - Accept all certificates (dangerous, test-only)

**Benefits**:
- Makes validation flags independent and composable
- Prevents accidental parameter swapping
- Self-documenting API
- Reduces ~35 lines of inline callback code to 3 lines

### 🔌 Programmatic API Enhancement
Extended `DotNettySslSetup` to expose all SSL/TLS settings:
- Added `RequireMutualAuthentication` property (was missing from programmatic API)
- Added `ValidateCertificateHostname` property (new setting)
- Backward-compatible constructors (2, 3, and 4 parameters)
- Comprehensive XML documentation

### ✅ Test Coverage

**DotNettyMutualTlsSpec** - 4 new end-to-end tests:
- Hostname validation disabled with different certificates (should succeed)
- Hostname validation enabled with different certificates (should fail)
- Same certificate mutual TLS scenario (typical use case)
- Default configuration backward compatibility check

**DotNettySslSetupSpec** - 3 integration tests:
- Verify 2-parameter setup configures effective settings with defaults
- Verify 3-parameter setup configures effective settings
- Verify 4-parameter setup configures all settings correctly

All tests verify the actual consumption path through `DotNettyTransportSettings.Create()`.

## Technical Details

Chain validation and hostname validation are now fully independent:
- `suppressValidation=true` disables chain/CA validation (for self-signed certs)
- `validateCertificateHostname=true/false` controls hostname matching

This allows testing hostname validation with self-signed certificates using:
```
suppressValidation = true
validateCertificateHostname = true
```

## Breaking Changes

None - all defaults preserve existing behavior.

## Related Issues

Fixes #7893

## Testing

All new tests pass:
- ✅ DotNettyMutualTlsSpec: 10/10 tests passed
- ✅ DotNettySslSetupSpec: 5/5 tests passed

### DIFF
--- a/docs/articles/remoting/security.md
+++ b/docs/articles/remoting/security.md
@@ -122,7 +122,9 @@ The validation process follows [RFC 5280 (X.509 PKI Certificate and CRL Profile)
 
 **New in v1.5.52+:** Controls whether the certificate CN/SAN must match the target hostname.
 
-#### Validate-Certificate-Hostname = False (DEFAULT)
+**IMPORTANT: This setting defaults to `false` (disabled).** Hostname validation is NOT performed by default to support common Akka.NET deployment patterns like mutual TLS with per-node certificates and IP-based connections.
+
+#### Validate-Certificate-Hostname = False (DEFAULT - Disabled)
 
 **What it does:**
 
@@ -445,7 +447,7 @@ akka.remote.dot-netty.tcp {
   ssl {
     suppress-validation = false  # Validates all certificates (default when SSL enabled)
     require-mutual-authentication = true  # Requires client certs (default when SSL enabled since v1.5.52)
-    validate-certificate-hostname = false  # Skips hostname validation (default, suitable for P2P with per-node certs)
+    validate-certificate-hostname = false  # DEFAULT: Hostname validation disabled (suitable for P2P with per-node certs)
     certificate {
       use-thumbprint-over-file = true
       thumbprint = "2531c78c51e5041d02564697a88af8bc7a7ce3e3"
@@ -583,7 +585,8 @@ The best practice for network security is to make the network itself secure. Run
 ### Error: "RemoteCertificateNameMismatch" - Hostname Validation Failure
 
 **Full error message:**
-```
+
+```text
 TLS certificate validation failed (full validation):
   - Certificate name mismatch
     - RemoteCertificateNameMismatch: The hostname being connected to does not match
@@ -602,17 +605,18 @@ Connection target: 192.168.1.100:4053
 **Common scenarios:**
 
 1. **Connecting via IP but certificate has DNS name**
-   - Connecting to: `192.168.1.100`
-   - Certificate CN: `node1.example.com`
+   * Connecting to: `192.168.1.100`
+   * Certificate CN: `node1.example.com`
 
 2. **Per-node certificates in P2P cluster**
-   - Node A cert CN: `node-a.cluster.local`
-   - Node B cert CN: `node-b.cluster.local`
-   - Each node's certificate doesn't match the other node's hostname
+   * Node A cert CN: `node-a.cluster.local`
+   * Node B cert CN: `node-b.cluster.local`
+   * Each node's certificate doesn't match the other node's hostname
 
 **Fix:**
 
 Option 1 (Recommended for P2P clusters): Disable hostname validation
+
 ```hocon
 akka.remote.dot-netty.tcp.ssl {
   validate-certificate-hostname = false  # Allow per-node certs
@@ -620,6 +624,7 @@ akka.remote.dot-netty.tcp.ssl {
 ```
 
 Option 2: Use certificates with matching CN/SAN
+
 ```bash
 # Ensure certificate CN matches connection target
 # For IP connections, add IP SAN to certificate:
@@ -629,6 +634,7 @@ New-SelfSignedCertificate -Subject "CN=node1" `
 ```
 
 Option 3: Connect via DNS names that match certificate CN
+
 ```hocon
 akka.remote.dot-netty.tcp {
   hostname = "node1.example.com"  # Must match cert CN
@@ -638,7 +644,8 @@ akka.remote.dot-netty.tcp {
 ### Error: "UntrustedRoot" - Certificate Chain Validation Failure
 
 **Full error message:**
-```
+
+```text
 TLS/SSL certificate validation failed:
   - Certificate chain validation errors
     - UntrustedRoot: A certificate chain processed, but terminated in a root
@@ -654,6 +661,7 @@ Certificate Details:
 **Fix:**
 
 Option 1 (Development only): Suppress chain validation
+
 ```hocon
 akka.remote.dot-netty.tcp.ssl {
   suppress-validation = true  # WARNING: Development only!
@@ -661,6 +669,7 @@ akka.remote.dot-netty.tcp.ssl {
 ```
 
 Option 2 (Recommended): Trust the CA certificate
+
 ```powershell
 # Windows: Import CA to Trusted Root store
 Import-Certificate -FilePath ca.cer -CertStoreLocation Cert:\LocalMachine\Root
@@ -681,7 +690,8 @@ Since v1.5.52, TLS handshake failures provide detailed diagnostic information in
 * **Actionable recommendations**
 
 **Example comprehensive error:**
-```
+
+```text
 TLS handshake failed on channel [127.0.0.1:4053->127.0.0.1:54321](Id=...)
 
 Detailed TLS Error:
--- a/docs/articles/remoting/security.md
+++ b/docs/articles/remoting/security.md
@@ -18,19 +18,19 @@ title: Network Security
 
 For many deployments, TLS is not strictly necessary:
 
-* ✅ **Internal networks only** - If your cluster runs entirely within a trusted network boundary
-* ✅ **Development/staging environments** - Where data sensitivity is low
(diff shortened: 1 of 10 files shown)


docs/articles/remoting/security.md
Aaronontheweb, line 1: Cleaned up the docs and mentioned the new settings introduced in this PR, along with where certs are sourced from in the store / how they're used.

src/core/Akka.Remote.Tests/Transport/DotNettyMutualTlsSpec.cs
```diff
@@ -275,6 +280,165 @@ await Assert.ThrowsAsync<AskTimeoutException>(async () =>
             }
         }
 
+        [Fact(DisplayName = "Different certificates with hostname validation disabled should connect successfully")]
+        public async Task Hostname_validation_disabled_should_allow_different_certificates()
```
Aaronontheweb, line 284: It's a tad difficult for us to robustly test this functionality without doing things like installing custom CAs, so we're doing this all with self-signed certificates with hostname validation and CA validation both disabled.

src/core/Akka.Remote.Tests/Transport/DotNettySslSetupSpec.cs
```diff
@@ -105,6 +105,96 @@ await Assert.ThrowsAsync<RemoteTransportException>(async () =>
             });
         }
 
+        [Fact(DisplayName = "DotNettySslSetup with 2 parameters should configure effective DotNettyTransportSettings with defaults (RequireMutualAuth=true, ValidateHostname=false)")]
```
Aaronontheweb, line 108: Updated the `DotNettySslSetup` so this stuff can get incorporated into that configuration path as well

src/core/Akka.Remote/Configuration/Remote.conf
```diff
@@ -565,6 +565,18 @@ akka {
...
+        # When true: Traditional TLS hostname validation is performed (certificate CN/SAN must match target hostname)
+        # When false: Only validates certificate chain against CA, ignores hostname mismatches
+        #
+        # Set to false for scenarios such as:
+        #   - Mutual TLS with per-node certificates in P2P clusters
+        #   - IP-based connections where certificates use DNS names
+        #   - Service discovery with dynamic addresses
+        #
+        # Default: false (disabled for backward compatibility and mutual TLS flexibility)
+        validate-certificate-hostname = false
```
Aaronontheweb, line 579: New setting, for validating the hostnames on the provided certificates. This should default to false so IP-based connections can still function correctly, but it can be enabled for zero-trust environments.

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
```diff
@@ -352,42 +352,39 @@ private void SetClientPipeline(IChannel channel, Address remoteAddress)
...
             {
                 var certificate = Settings.Ssl.Certificate;
-                var host = certificate.GetNameInfo(X509NameType.DnsName, false);
+                // Use the remote address host for TLS validation, not the client's certificate name
+                var host = remoteAddress.Host;
 
                 IChannelHandler tlsHandler;
 
-                if (Settings.Ssl.SuppressValidation)
+                // Build validation callback using type-safe factory methods
```
Aaronontheweb, line 360: Simplified the SSL callback construction using some enums to signify which behaviors we want to use - this should help reduce branching in here and make it easier to debug the SSL configuration.

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransportSettings.cs
```diff
@@ -424,17 +447,145 @@ private SslSettings(string certificateThumbprint, string storeName, StoreLocatio
...
+        IgnoreHostnameMismatch
+    }
+
+    /// <summary>
+    /// INTERNAL API
+    ///
+    /// Factory for creating TLS certificate validation callbacks with different security policies.
+    /// Provides type-safe, self-documenting methods for configuring certificate validation behavior.
+    /// </summary>
+    internal static class TlsValidationCallbacks
```
Aaronontheweb, line 517: Builder for creating the cert validation callbacks - covered with our tests

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransportSettings.cs
```diff
@@ -424,17 +447,145 @@ private SslSettings(string certificateThumbprint, string storeName, StoreLocatio
...
+        /// Use for traditional client-server TLS with CA-signed certificates and DNS names.
+        /// </summary>
+        public static RemoteCertificateValidationCallback ValidateFull(ILoggingAdapter log)
+            => Create(ChainValidationMode.ValidateChain, HostnameValidationMode.ValidateHostname, log);
+
+        /// <summary>
+        /// Creates validation callback that validates chain but ignores hostname mismatches.
+        /// Use for: Mutual TLS with per-node certificates, IP-based connections, dynamic service discovery.
+        /// </summary>
+        public static RemoteCertificateValidationCallback ValidateChainOnly(ILoggingAdapter log)
```
Aaronontheweb, line 573: This is the default setting

Aaronontheweb (commented):
Detailed my changes

src/core/Akka.Remote/Transport/DotNetty/DotNettyTransport.cs
```diff
@@ -352,42 +352,39 @@ private void SetClientPipeline(IChannel channel, Address remoteAddress)
             if (Settings.EnableSsl)
             {
                 var certificate = Settings.Ssl.Certificate;
-                var host = certificate.GetNameInfo(X509NameType.DnsName, false);
+                // Use the remote address host for TLS validation, not the client's certificate name
+                var host = remoteAddress.Host;
```
Aaronontheweb, line 356: Another important fix - the host is determined by the address of the remote host, not the client's certificate name (because client != server host names, due to how Akka.NET mesh networking works.)

Arkatufus (approved):
LGTM

PR #7906, opened 2025-10-13
Fix UnfoldFlow initialization race condition causing test failures

Arkatufus:
## Summary
Fixes a race condition in the UnfoldFlow GraphStage that was causing intermittent test failures with "Timeout while waiting for OnSubscribe" errors.

## The Problem
The UnfoldFlow GraphStage had a circular dependency during initialization where:
- The output handler waited for the feedback outlet to be available before pushing the seed
- The feedback handler waited for the output outlet to be available before pushing the seed
- If neither outlet was available when pulled, the seed was never pushed
- This prevented the subscription handshake from completing, causing test timeouts

## The Solution
Modified the initialization logic to:
1. Add a `_waitingForInitialDemand` flag to track initialization state
2. Create a `TryPushInitialSeed()` method to centralize seed pushing logic
3. During initialization: Push the seed as soon as the feedback outlet has demand (don't wait for both outlets)
4. After initialization: Preserve the original logic requiring both outlets to be available
5. Eliminate code duplication between OnPull handlers

## Test Results
✅ **All UnfoldFlowSpec tests pass consistently**
- Ran the entire test suite 30 consecutive times with **zero failures**
- 19 tests × 30 runs = 570 successful test executions
- The previously failing test `UnfoldFlow_should_increment_integers_and_handle_KillSwitch_and_fail_after_timeout_when_aborted` now passes consistently

## Impact
This fix eliminates the race condition that was causing tests marked with `[LocalFact(SkipLocal = "Racy on Azure DevOps")]` to fail intermittently, improving CI reliability.

Fixes #7899 (if there's a related issue)

### DIFF
--- a/src/core/Akka.Streams/Dsl/UnfoldFlow.cs
+++ b/src/core/Akka.Streams/Dsl/UnfoldFlow.cs
@@ -22,6 +22,7 @@ namespace Akka.Streams.Dsl
 
         protected TState _pending;
         protected bool _pushedToCycle;
+        private bool _waitingForInitialDemand;
 
         protected UnfoldFlowGraphStageLogic(FanOutShape<TIn, TState, TOut> shape, TState seed, TimeSpan timeout) : base(shape)
         {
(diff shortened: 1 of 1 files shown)


Aaronontheweb (approved):
LGTM

PR #7911, opened 2025-10-16
Fix LWWDictionary.Delta ArgumentNullException when underlying delta is null

### LINKED ISSUE #7910: [BUG] LWWDictionary.Delta throws ArgumentNullException when underlying ORDictionary.Delta is null
## Description

`LWWDictionary` throws an `ArgumentNullException` with message "Delta operation can not be null" when the Replicator attempts to access the `Delta` property on a dictionary where the underlying `ORDictionary.Delta` is `null`.

This occurs during normal DData operations and results in `ModifyFailure` messages being delivered as dead letters.

## Versions & Operating System

- **Akka.NET version**: 1.5.53
- **OS**: Windows
- **.NET Runtime**: .NET 9.0

## Expected behavior

The `LWWDictionary.Delta` property should handle the case where `ORDictionary.Delta` returns `null` (which is valid according to the `ORDictionary` implementation after initialization or `ResetDelta()`).

## Actual behavior

An `ArgumentNullException` is thrown, causing the update operation to fail with a `ModifyFailure` message:

```
Message [ModifyFailure] from [akka://DrawTogether/user/ddataReplicatorSupervisor/ddataReplicator#729325827] to [akka://DrawTogether/user/all-drawings-publisher#860097492] was unhandled. 

Message content: ModifyFailure(all-drawings-index, System.ArgumentNullException: Delta operation can not be null (Parameter 'underlying')
   at Akka.DistributedData.LWWDictionary`2.LWWDictionaryDelta..ctor(IDeltaOperation underlying)
   at Akka.DistributedData.LWWDictionary`2.get_Delta()
   at Akka.DistributedData.LWWDictionary`2.Akka.DistributedData.IDeltaReplicatedData.get_Delta()
   at Akka.DistributedData.Replicator.ReceiveUpdate(IKey key, Func`2 modify, IWriteConsistency consistency, Object request))
```

## Reproduction

```csharp
public sealed class AllDrawingsPublisherActor : UntypedActor
{
    private readonly IActorRef _replicator = DistributedData.Get(Context.System).Replicator;
    private readonly IWriteConsistency _writeConsistency = new WriteMajority(TimeSpan.FromSeconds(3));
    private readonly Cluster _cluster = Cluster.Get(Context.System);

    {
        switch (message)
        {
            case DrawingActivityUpdate update:
            {
                var key = ClusterConstants.AllDrawingsIndexKey;
                var updateOp = new Update(key, LWWDictionary<string, DrawingActivityUpdate>.Empty, _writeConsistency,
                    dictionary =>
                    {
                        var dict = (LWWDictionary<string, DrawingActivityUpdate>)dictionary;
                        dict = dict.SetItem(_cluster, update.DrawingSessionId.SessionId, update);
                        return dict;
                    });
                _replicator.Tell(updateOp);
                break;
            }
        }
    }
}
```

The error occurs when the Replicator internally accesses the `Delta` property on the modified dictionary.

## Root Cause

In `LWWDictionary.cs`, the `Delta` property is implemented as:

```csharp
public IDeltaOperation Delta => new LWWDictionaryDelta(Underlying.Delta);
```

However, `ORDictionary.Delta` can legitimately return `null` in two cases:
1. After initial construction when no modifications have been made
2. After `ResetDelta()` is called

The `LWWDictionaryDelta` constructor throws when passed `null`:

```csharp
public LWWDictionaryDelta(ORDictionary<TKey, LWWRegister<TValue>>.IDeltaOperation underlying)
{
    Underlying = underlying ?? throw new ArgumentNullException(
        nameof(underlying), "Delta operation can not be null");
    // ...
}
```

## Proposed Fix

The `LWWDictionary.Delta` property should handle the null case:

```csharp
public IDeltaOperation Delta => Underlying.Delta == null ? null : new LWWDictionaryDelta(Underlying.Delta);
```

Or alternatively, the interface contract for `IDeltaReplicatedData.Delta` should clearly specify whether `null` is an acceptable return value, and all implementations should be consistent.

## Links to source

- [LWWDictionary.cs Delta property](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.DistributedData/LWWDictionary.cs)
- [ORDictionary.cs Delta property](https://github.com/akkadotnet/akka.net/blob/dev/src/contrib/cluster/Akka.DistributedData/ORDictionary.cs)

Aaronontheweb:
## Summary
Fixed a bug where `LWWDictionary.Delta` would throw `ArgumentNullException` when the underlying `ORDictionary.Delta` is `null`, which is a legitimate state after initialization or calling `ResetDelta()`.

## Changes
- Modified `LWWDictionary.Delta` property to return `null` when `Underlying.Delta` is `null`
- Added test `Bugfix_7910_LWWDictionary_Delta_should_handle_null_underlying_delta` to verify the fix

## Root Cause
The `LWWDictionary.Delta` property was unconditionally wrapping `Underlying.Delta` in a `LWWDictionaryDelta` constructor, which throws `ArgumentNullException` when passed `null`. However, `ORDictionary.Delta` can legitimately return `null` after construction or `ResetDelta()`.

The Replicator expects deltas to be nullable (uses `??` operator at lines 665 and 685 in `Replicator.cs`), so this fix aligns `LWWDictionary` with the expected interface contract.

## Test Evidence
All 6 `LWWDictionarySpec` tests pass:
- 5 existing tests (no regressions)
- 1 new test verifying null delta handling

Fixes #7910

### DIFF
--- a/src/contrib/cluster/Akka.DistributedData.Tests/LWWDictionarySpec.cs
+++ b/src/contrib/cluster/Akka.DistributedData.Tests/LWWDictionarySpec.cs
@@ -161,5 +161,33 @@ namespace Akka.DistributedData.Tests
             merged1.Entries["b"].Should().BeEquivalentTo("B2");
             merged1.Entries["c"].Should().BeEquivalentTo("C");
         }
+
+        /// <summary>
+        /// Bug reproduction: https://github.com/akkadotnet/akka.net/issues/7910
+        /// LWWDictionary.Delta should return null when underlying ORDictionary.Delta is null
+        /// </summary>
+        [Fact]
+        public void Bugfix_7910_LWWDictionary_Delta_should_handle_null_underlying_delta()
+        {
+            // Empty dictionary has no delta
+            var empty = LWWDictionary<string, string>.Empty;
+            empty.Delta.Should().BeNull("empty dictionary should have null delta");
+
+            // After ResetDelta(), delta should be null
+            var m1 = LWWDictionary<string, string>.Empty
+                .SetItem(_node1, "a", "A")
+                .SetItem(_node1, "b", "B");
+
+            m1.Delta.Should().NotBeNull("dictionary with modifications should have a delta");
+
+            var m2 = m1.ResetDelta();
+            m2.Delta.Should().BeNull("after ResetDelta(), delta should be null");
+
+            // Verify the dictionary still contains the data
+            m2.ContainsKey("a").Should().BeTrue();
+            m2.ContainsKey("b").Should().BeTrue();
+            m2["a"].Should().Be("A");
+            m2["b"].Should().Be("B");
+        }
     }
 }
--- a/src/contrib/cluster/Akka.DistributedData/LWWDictionary.cs
+++ b/src/contrib/cluster/Akka.DistributedData/LWWDictionary.cs
@@ -402,8 +402,8 @@ namespace Akka.DistributedData
         }
 
         // TODO: optimize this so it doesn't allocate each time it's called
-        public ORDictionary<TKey, LWWRegister<TValue>>.IDeltaOperation Delta => 
-            new LWWDictionaryDelta(Underlying.Delta);
+        public ORDictionary<TKey, LWWRegister<TValue>>.IDeltaOperation Delta =>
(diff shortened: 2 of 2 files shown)


Arkatufus (approved):
LGTM

PR #7920, opened 2025-10-23
Fix race condition in ProducerController when producer restarts with unconfirmed messages

Aaronontheweb:
## Summary
Fixes a race condition in `ProducerController` where the `ResendFirst` timer (1-second interval) was not being cancelled when a producer restarted, causing aggressive resends of old unconfirmed messages to race with newly sent messages from the restarted producer.

## Changes
- Added `Timers.Cancel(ResendFirst.Instance)` in `ProducerControllerImpl.ReceiveStart()` when the producer reference changes
- New messages from restarted producers now take priority over aggressive resends
- Unconfirmed messages are still preserved and will be redelivered via the parent controller's slower idle timeout (10s)
- Maintains at-least-once delivery semantics without aggressive 1-second retry behavior during producer restarts

## Root Cause
The test `ReliableDeliveryShardingSpec.ReliableDelivery_with_Sharding_must_allow_restart_of_producer` was flaky because:
1. Test sends msg-1 and msg-2 but doesn't confirm them
2. Child `ProducerController` starts a 1-second `ResendFirst` timer for msg-1
3. Producer "restarts" (new `Start` message with different producer reference)
4. Timer continues running because child actor doesn't restart
5. Timer fires ~1 second after msg-1, triggering resend
6. Resent msg-1 races with newly sent msg-3
7. Sometimes msg-1 arrives first → test fails (expected msg-3)

## Test Results
Verified with extensive testing:
- **Akka.Delivery tests**: 20 consecutive runs, 1,120 total tests, 0 failures
- **Akka.Cluster.Sharding.Delivery tests**: 20 consecutive runs (both .NET 4.8 and .NET 8.0), 480 total tests, 0 failures
- **Grand total**: 1,600 tests executed across 40 runs with 100% success rate

## Modified Files
- `src/core/Akka/Delivery/Internal/ProducerControllerImpl.cs`

### DIFF
--- a/src/core/Akka/Delivery/Internal/ProducerControllerImpl.cs
+++ b/src/core/Akka/Delivery/Internal/ProducerControllerImpl.cs
@@ -638,6 +638,12 @@ internal sealed class ProducerController<T> : ReceiveActor, IWithTimers
     {
         AssertLocalProducer(start.Producer);
         _log.Debug("Registering new producer [{0}], currentSeqNr [{1}]", start.Producer, CurrentState.CurrentSeqNr);
+
+        // Cancel aggressive resend timer on producer restart
+        // New messages from restarted producer should take priority over aggressive resends of old unconfirmed messages
+        // The parent ShardingProducerController's idle timeout will handle eventual redelivery if needed
+        Timers.Cancel(ResendFirst.Instance);
+
         if (CurrentState is { Requested: true, RemainingChunks.IsEmpty: true })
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Re-assessing this fix - analysis revealed that the change doesn't actually address the race condition in the sharding case because child ProducerControllers don't receive the second Start message. Need to reproduce the original issue and implement the correct fix.

PR #7939, opened 2025-11-24
Fix: Akka.Remote should not shutdown on invalid TLS traffic

### LINKED ISSUE #7938: Akka.Remote: When TLS is enabled, unexpected requests sent to the remoting port can cause the application to shut down
**Version Information**
Version of Akka.NET? 1.5.55
Which Akka.NET Modules? Akka.Remoting

**Describe the bug**
When TLS is enabled, a HTTP request sent to the Akka.Remote port will cause the app to shut down.

**To Reproduce**
1. Enable Akka.Remoting with mutual TLS & pinned certificate validation. (See screenshot for the exact config used).
2. Send a HTTP OPTIONS request to the Akka.Remote port. (e.g. with `curl -XGET https://localhost:5055`)
3. See the application shut down

**Expected behavior**
Application continues running, possibly logging the unhandled request on that port.

**Actual behavior**
Application shuts down.

**Screenshots**

Akka.Remoting config used:

Image

Example of logs from when a HTTP OPTIONS request is received & subsequent shutdown:

Image

Image

**Environment**
.NET 8
Kubernetes (Linux, amd64, kernel version 5.10.245-241.976.amzn2.x86_64)

**Additional context**
Akka.Remoting otherwise works perfectly - I am running 6 nodes in the cluster, and they are all able to communicate with each other using Akka.Remote with TLS.

Aaronontheweb:
## Summary
- Fixed issue where invalid traffic (like HTTP requests) hitting a TLS-enabled Akka.Remote port would cause the entire ActorSystem to shut down
- Server now rejects invalid connections gracefully without terminating
- Client-side TLS failures still trigger shutdown as expected for security

## Changes
- Modified `TcpTransport.UserEventTriggered()` to distinguish between client and server-side TLS failures
- Only client-side TLS handshake failures trigger `CoordinatedShutdown`
- Server-side failures just log a warning and close the connection

## Test Plan
- [x] Added test `Server_side_invalid_traffic_should_not_shutdown_server` to verify fix
- [x] Verified existing client-side TLS failure test still passes
- [x] Ran new test 20 times locally to ensure stability
- [x] All tests in `DotNettyTlsHandshakeFailureSpec` pass

Fixes #7938

### DIFF
--- a/src/core/Akka.Remote/Transport/DotNetty/TcpTransport.cs
+++ b/src/core/Akka.Remote/Transport/DotNetty/TcpTransport.cs
@@ -92,9 +92,20 @@ namespace Akka.Remote.Transport.DotNetty
                     context.Channel.LocalAddress, context.Channel.RemoteAddress,
                     context.Channel.Id, detailedError);
 
-                // Shutdown the ActorSystem on TLS handshake failure
-                var cs = CoordinatedShutdown.Get(Transport.System);
-                cs.Run(new TlsHandshakeFailureReason($"TLS handshake failed on channel [{context.Channel.LocalAddress}->{context.Channel.RemoteAddress}](Id={context.Channel.Id})"));
+                // Only shutdown the ActorSystem if this is a client-side failure
+                // Server-side failures (incoming connections) should just reject the connection
+                if (isClient)
+                {
+                    // Client-side: We initiated the connection and TLS failed - this is critical
+                    var cs = CoordinatedShutdown.Get(Transport.System);
+                    cs.Run(new TlsHandshakeFailureReason($"TLS handshake failed on outbound connection to [{context.Channel.RemoteAddress}]"));
+                }
+                else
+                {
+                    // Server-side: Someone connected to us with invalid TLS - just reject them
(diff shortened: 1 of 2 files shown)


src/core/Akka.Remote/Transport/DotNetty/TcpTransport.cs
```diff
@@ -92,9 +92,20 @@ public override void UserEventTriggered(IChannelHandlerContext context, object e
                     context.Channel.LocalAddress, context.Channel.RemoteAddress,
                     context.Channel.Id, detailedError);
 
-                // Shutdown the ActorSystem on TLS handshake failure
-                var cs = CoordinatedShutdown.Get(Transport.System);
-                cs.Run(new TlsHandshakeFailureReason($"TLS handshake failed on channel [{context.Channel.LocalAddress}->{context.Channel.RemoteAddress}](Id={context.Channel.Id})"));
+                // Only shutdown the ActorSystem if this is a client-side failure
+                // Server-side failures (incoming connections) should just reject the connection
+                if (isClient)
```
Aaronontheweb, line 97: Fix is simple - just don't shut the server down when clients with invalid certs connect

Arkatufus (approved):
LGTM

